PR #1866, opened 2020-01-13
Introduce workspaces to Pipelines.

sbwsg:
A workspace can be threaded through the tasks of a Pipeline, allowing easier
reuse and customisation of shared storage between tasks.

Workspace types have been added to Pipeline structs:

* A Pipeline declares a list of workspace names that it requires be provided
by a PipelineRun.
* A PipelineTask declares a mapping from a Pipeline's workspace name to a
workspace name declared by the Task it references (or embeds).
* A PipelineRun declares the concrete workspace implementations - whether
they be a PVC, emptyDir, configMap or secret - and passes those in with
the name declared by the Pipeline it references (or embeds).

An example PipelineRun has been added to demonstrate usage of
workspaces in pipelines but I still need to add docs and tests.

TODOs:
- [x] Document that pipeline task ordering is not yet enforced when tasks share workspaces. Must use `runAfter` for now.
- [x] Document emptyDir not being shared across tasks in pipelinerun
- [x] Document configmap and secret workspace support in the pipelineruns doc.
- [x] Document configmap size limit.  (Does secret have same size limit? Yes)
- [x] E2E test: check that validation fails when workspace provided more than once by pipelinerun
- [x] E2E test: check that submitting pipelinerun with missing workspace results in error

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Pipelines and PipelineRuns can now declare workspaces to more easily thread shared storage through the tasks in a pipeline
```


docs/pipelineruns.md
```diff
@@ -286,9 +286,47 @@ spec:
...
+
+```yaml
+workspaces:
+- name: myworkspace
+  persistentVolumeClaim:
+    claimName: mypvc
+  subPath: my-subdir
+```
+
+Or to use [`emptyDir`](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) for the same `workspace`:
```
chmouel: can emptyDir be passed across Tasks in a Pipeline? I was under the impression they are bound to a task/pod?

sbwsg: You're right - they don't get passed across.  I'll clarify this in the docs.

chmouel: configmap looks good as an alternative for folks that don't have a PVC available but i think there is limit to how much we can store in a configmap, we may want to mention it? It seems that etcd limits it to 1MB ie: (old issue tho) https://github.com/kubernetes/kubernetes/issues/19781

sbwsg: have updated docs with mention of emptyDir being newly created for each task & linked to kubernetes code showing the MaxSecretSize check (used for both configmap and secrets).  I couldn't find this max size documented anywhere in k8s' docs unfortunately.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -553,6 +553,22 @@ func (c *Reconciler) createTaskRun(rprt *resources.ResolvedPipelineRunTask, pr *
...
 
+	for _, w := range rprt.PipelineTask.Workspaces {
+		taskWorkspaceName := w.Name
+		pipelineWorkspaceName := w.Workspace
+		var binding *v1alpha1.WorkspaceBinding
+		for i := range pr.Spec.Workspaces {
+			if pr.Spec.Workspaces[i].Name == pipelineWorkspaceName {
+				binding = pr.Spec.Workspaces[i].DeepCopy()
+				binding.Name = taskWorkspaceName
+			}
```
pierretasci: If I'm not mistaken, there isn't any validation that a pipeline run's workspaces are uniquely named like there is on the pipeline. This could lead to a weird foot-gun where the last similarly named workspace wins out. Unlikely but easily avoidable with a map of the pipeline run's workspace instead of iterating over both. 

Just a suggestion. I think this would work almost all of the time as is.

sbwsg: I've added some validation around the case you've mentioned and updated this code to use a map.  I've also improved on the error message that gets returned here to be more correct.

sbwsg:
OK, barring any last build failures or reviewer items I think this PR is in a finished state.

chmouel:
nicely done @sbwsg 👍 

/lgtm

sbwsg:
:pray: thank you!

Issue #1867, opened 2020-01-13
PullRequest Resource e2e tests


wlynch:
I thought there was an issue for this already, but I couldn't find one. If there is another, feel free to close this as a duplicate. :smiley: 

We should add e2e tests for the pull request resource. This should test:

- GitHub, GitLab providers (incl custom domains)
- Should be able to work with the real providers if desired.
- Should have a local, non-authenticated mode as an approximation for local presubmit testing.

/assign wlynch

sbwsg:
Moving to the beta milestone.  We may well introduce this in an RC instead.

PR #1874, opened 2020-01-14
Pull request GitHub resource e2e test.

wlynch:
# Changes

Adds new E2E test that can runs the pull request resource on both a fake
and real SCM servers.

Also slightly modifies the behavior of GitHub client creation to allow
http URLs for the default GitHub host in order to allow for HTTP
proxying in for the fake SCM case.

First step towards #1867. More providers and GHE tests to come later.

# Submitter Checklist

review them:

## Reviewer Notes


pkg/pullrequest/e2e/e2e_test.go
```diff
@@ -0,0 +1,155 @@
+package e2e
```
dibyom: Should these files be in test/ package/folder like the rest of our e2e tests?

vdemeester: I would definitely prefer it to be in `test/pullrequest` or something indeed :angel:

pkg/pullrequest/e2e/e2e_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/jenkins-x/go-scm/scm"
+)
+
+var (
+	proxy = flag.Bool("proxy", true, "whether to use fake in-memory proxy instead of using real SCM servers")
```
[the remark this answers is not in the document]
wlynch: Done.

test/pullrequest/e2e_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/jenkins-x/go-scm/scm"
+)
+
+var (
+	proxy = flag.Bool("use-fake-scm-proxy", true, "whether to use fake in-memory proxy instead of using real SCM servers")
```
bobcatfish: im not a huge fan of this being global - i cant figure out a better option but i feel like putting it in an `init` function is slightly better? it's effectively the same :( i dont have anything super useful to say here; i feel like we should be defining all the test flags in one place but then that makes it harder to track down what is doing :woman_shrugging: 

We do have init_test.go which is which is where the other flags are handled.

test/pullrequest/e2e_test.go
```diff
@@ -0,0 +1,170 @@
...
+					t.Fatalf("net.Listen: %v", err)
+				}
+				t.Logf("starting HTTP proxy on %s", l.Addr().String())
+				go func() {
+					if err := http.Serve(l, h); err != nil {
+						fmt.Println("http.Serve:", err)
+					}
+				}()
+				os.Setenv("http_proxy", l.Addr().String())
+			}
```
bobcatfish: separate function for this?

test/pullrequest/e2e_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+			t.Log("Environment vars:", os.Environ())
+
+			banner(t, "download")
```
bobcatfish: hmmmm are you sure you want the banner like this? seems like the output from this test is gonna look pretty different from the rest that we have - maybe knowing what it looks like would change my mind.

maybe this is an indication that this test has more output than is needed?

test/pullrequest/e2e_test.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		log.Fatalf("ioutil.TempDir: %v", err)
+	}
+	t.Logf("Test dir: %s", tmpdir)
+	defer os.RemoveAll(tmpdir)
+
+	// Make sure command is installed locally to resolve any dependencies.
+	// We need to do this first since the run might try to pull in new
+	// dependencies through the proxy.
+	install := exec.Command("go", "get", "-u", "github.com/tektoncd/pipeline/cmd/pullrequest-init")
```
bobcatfish: im confused by why this and the below `go run` are needed - could you explain a bit more? i thought the proxy was just receiving HTTP requests and replying, not sure why extra dependencies are needed?

test/pullrequest/e2e_test.go
```diff
@@ -0,0 +1,170 @@
...
+				fmt.Sprintf("-path=%s", dir),
+				fmt.Sprintf("-url=%s", tc.url),
+			}
+			upload := exec.Command("go", uploadFlags...)
+			t.Log(upload)
+			out, err := upload.CombinedOutput()
+			if err != nil {
+				t.Fatal(string(out))
+			}
+		})
```
bobcatfish: i think this test would be a bit easier to read if some of this was broken into separate functions - probably exactly in the spots where you have comments to explain the code below, those are great candidates for separate functions

i dunno if youve ever read the book Clean Code (https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29#comments-rules has a good but probably too short summary of the contents) - ive only skimmed it myself!! but one thing that really stuck with me was to always try to explain myself in code and not comments - and often that means making a function call instead of a comment followed by a bunch of code

test/pullrequest/e2e_test.go
bobcatfish: > but we should aim to consolidate these flags into a single "main.go" - in this case e2e_test.go to keep with test file semantics and to keep the flag definitions as close to their usage as possible.

I think this would make sense to me for separate binaries, but afaik every end to end test will pull in this flag as a result of it being declared globally.

The only other option I can think of is to add a build flag around the whole test, in which case it really would be a separate binary, but I think that makes things more confusing in the long run.

I dunno, I don't feel too strongly.

test/pullrequest/e2e_test.go
bobcatfish: nice!! i like it :D

test/pullrequest/e2e_test.go
```diff
@@ -0,0 +1,170 @@
...
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			dir := filepath.Join(tmpdir, tc.name)
+			if err := os.Mkdir(dir, 0755); err != nil {
+				t.Fatalf("os.Mkdir: %v", err)
+			}
+
+			if tc.token != "" {
+				os.Setenv("AUTH_TOKEN", tc.token)
+				defer os.Unsetenv("AUTH_TOKEN")
```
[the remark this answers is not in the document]
bobcatfish: _ooooooooohhhhhhhh_ you're running the pull request init binary directly!!! Everything makes a lot more sense now. (What did you think it was doing beforehand you might ask? I guess I assumed you were running an entire Task or something but clearly that isn't happening XD)

I think there are 2 things that could be done to help an easily confused skimmer like me:
1. Mention in that readme that you are running the pull request binary directly (i saw that there was an http server running and that was the only potential other "binary" i could wrap my head around for a while)
2. maybe instead of `runCommand` you could call the function something like `runPRBinary` or something? (even if runCommand is technically correct)?

bobcatfish:
Okay I think I have one last request around making it more clear to folks who are skimming like me that you're running the pull request binary (since most end to end tests tend to execute Tekton types, vs. running a binary that we publish in a container) and once we have that I think we are gtg!

vdemeester:
@wlynch needs a rebase :angel:

test/pullrequest/e2e_test.go
```diff
@@ -0,0 +1,200 @@
...
+	}()
+	os.Setenv("http_proxy", l.Addr().String())
+}
+
+// updateDeps makes sure command is installed locally to resolve any
+// dependencies. We need to do this first since the run might try to
+// pull in new dependencies through the proxy.
+func updateDeps(t *testing.T) {
+	t.Helper()
+	install := exec.Command("go", "get", "-u", pkg)
```
vdemeester: This can't be done in the `CI` (because of `-mod=vendor`). Our tests should do as little "network" as possible. A `exec.Command("go", "build", pkg)` should work though :wink:

vdemeester: All the dependencies are already there (in the `vendor` folder) and even if we didn't have the `vendor` folder, doing a `go build` (without `-mod=vendor`) would get those dependencies. Doing a `go get -u` here scares me to death as I don't have the guarantee it will use the code that is in the PR (for `pullrequest-init`)

test/pullrequest/e2e_test.go
```diff
@@ -0,0 +1,200 @@
+// Package pullrequest tests the pullrequest-init binary end-to-end by invoking
+// the binary and observing the resulting filesystem state.
+package pullrequest
```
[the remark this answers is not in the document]
wlynch: I intentionally wrote this in a way so that it could be executed as a unit test. :D 

The idea was that by default the proxy is spun up and all traffic stays local by default. Only if the proxy flag is disabled should this actually make external calls.

That said, I don't feel strongly about this so if we want to take the stance that if it shells out it should have an e2e label that sounds fine. Let me know what you think!

test/pullrequest/e2e_test.go
wlynch: Talked about this a bit offline -

For now this is fine because it's the only test in the package. We should be mindful of this flag spreading to other binaries / libraries in the future though.

test/pullrequest/e2e_test.go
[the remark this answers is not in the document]
wlynch: Done.

test/pullrequest/e2e_test.go
wlynch: Good call. Done.

PR #1884, opened 2020-01-16
Use `tekton` instead of all as categories 🎿

### LINKED ISSUE #1246: What should be the CRDs categories for Tekton ?
As of today, Tekton's CRDs are using `tektoncd-pipeline` and `all` as categories. This means you can do `kubectl get all` or `kubectl get tektoncd-pipeline` and it will display those resources (and others if in the same categorie). For `all`, this might pose a problem if the user has not sufficient rights, see quote below.

>_The problem with this is that if you are operating under expectation that 'all' will only include the original core builtin resources from Kubernetes, if you are using 'oc' from a service account with a basic set of roles which only covers builtin resource types, then you start getting errors:
>
> Error from server (Forbidden): images.caching.internal.knative.dev is forbidden: User "system:serviceaccount:homeroom:portal-homeroom-user1" cannot list resource "images" in API group "caching.internal.knative.dev" in the namespace "portal-homeroom-user1-2"
> Error from server (Forbidden): clustertasks.tekton.dev is forbidden: User "system:serviceaccount:homeroom:portal-homeroom-user1" can not list resource "clustertasks" in API group "tekton.dev" at the cluster scope
>
> Can someone explain the rationale as to why knative/tekton resources are being shoved under the set of resources that 'all' returns when using 'oc get all'. It seems to go against the idea that these are optional resources for an operator that may not even be installed, they are not core builtin resources.

kubectl get all *or* oc get all displays all resources that are in the "all" category.

There used to be a small "best practice" doc on when to add resources to all or not (see https://github.com/thtanaka/kubernetes/blob/master/docs/devel/kubectl-conventions.md#rules-for-extending-special-resource-alias---all) ; which were
- No cluster scoped resources
- No namespace admin level resources (limits, quota, policy, authorization rules)
- No resources that are potentially unrecoverable (secrets and pvc)
- Resources that are considered "similar" to #3 should be grouped the same (configmaps)

Somehow, I can't find this document anymore. I don't think the expectation was "only original core builtin", as a lot of CRDs are adding themselves to "all" (part of istio, …).

That said, I understand that having those errors are not really user friendly. It might also make sense to groups some of those resources in other categories.

I wonder if we really need to have our CRDs adding to the `all` categorie or if we can remove that from the categories and use more *accurate* ones (like `ci`, `pipelines`, …) ?

/kind question

vdemeester:
# Changes

This means Tekton struct won't show in `kubectl get all` but in
`kubectl get tekton` (or `kubectl get tekton-pipelines`) 😉.

If we do this for other tektoncd project (like [`tektoncd/triggers`](https://github.com/tektoncd/triggers)), it would be nice :angel: (we would be able to list all tekton related resource at once, not polluting or being polluted by `all`). And each tekton components can define their own (like `tekton-pipelines` here, `tekton-triggers` for triggers, …)

Fix #1246

/cc @sbwsg @dibyom @bobcatfish @ImJasonH 
/ping @siamaksade 

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Remove the category `all` for our CRD and have a `tekton` categorie that is shared with other tektoncd projects
```

### DIFF
--- a/config/300-clustertask.yaml
+++ b/config/300-clustertask.yaml
@@ -21,7 +21,7 @@ spec:
     kind: ClusterTask
     plural: clustertasks
     categories:
-    - all
+    - tekton
     - tekton-pipelines
   scope: Cluster
   # Opt into the status subresource so metadata.generation
--- a/config/300-condition.yaml
+++ b/config/300-condition.yaml
@@ -21,7 +21,7 @@ spec:
     kind: Condition
     plural: conditions
     categories:
-      - all
+      - tekton
       - tekton-pipelines
   scope: Namespaced
   # Opt into the status subresource so metadata.generation
--- a/config/300-pipeline.yaml
+++ b/config/300-pipeline.yaml
@@ -21,7 +21,7 @@ spec:
     kind: Pipeline
     plural: pipelines
     categories:
-    - all
+    - tekton
     - tekton-pipelines
   scope: Namespaced
   # Opt into the status subresource so metadata.generation
--- a/config/300-pipelinerun.yaml
+++ b/config/300-pipelinerun.yaml
@@ -21,7 +21,7 @@ spec:
     kind: PipelineRun
     plural: pipelineruns
     categories:
-    - all
+    - tekton
     - tekton-pipelines
     shortNames:
     - pr
--- a/config/300-resource.yaml
+++ b/config/300-resource.yaml
@@ -21,7 +21,7 @@ spec:
     kind: PipelineResource
     plural: pipelineresources
     categories:
-    - all
+    - tekton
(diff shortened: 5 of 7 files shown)


sbwsg:
Could it support both `all` and `tekton`?

sbwsg:
oh sorry, nevermind, i see the original issue.

/lgtm

chmouel:
yeah `all` is a pretty bad place for user level stuff...

there was some dicussions about it sometime ago in this operator-framework  issue : 

https://github.com/operator-framework/operator-lifecycle-manager/pull/631#issuecomment-448590464

/lgtm

PR #1887, opened 2020-01-16
feat: creds-init, add ability to filter secrets for a specific GitHub…

### LINKED ISSUE #1886: creds-init: filter a kubernetes secret for a matching pipeline request + GitHub org
# Expected Behavior
When using GitHub Apps we have multiple tokens (and therefore k8s secrets), one for each GitHub Organisation the App is installed into.  When triggering a pipeline we want to mount the correct token that matches the GitHub org with the git owner of the pipeline request.

# Actual Behavior
Multiple secrets are matched when the creds-init container spec is created so we can have the wrong token mounted for that pipeline.

Have multiple git secrets in a namespace with different tokens used for different GitHub Orgs

# Additional Info
PR that shows an approach that is working well for us on it's way, it might be easier to explain a proposed solution using that

rawlingsj:
… organisation when using GitHub Apps

fixes #1886

# Changes

Adds ability to filter secrets for a specific GitHub organisation when using GitHub Apps.

# Submitter Checklist

review them:

# Release Notes

```
- Fixes #1886
- Filter multiple git secrets when using GitHub Apps
```

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -175,6 +175,32 @@ generated containing the credentials configured in the `Secret`, and these
 credentials are then used to authenticate when retrieving any
 `PipelineResources`.
 
+## GitHub Apps
+
+[GitHub Apps](https://developer.github.com/apps/about-apps/) are the preferred way for automated operations say from
+within a pipeline to interact with GitHub, rather than managing a 'bot' user and personal access token.
+
+A GitHub App is installed into a GitHub Organisation or users account, an API is then called to retrieve a token that can
+then be used to authenticate Git requests.  This means we now have many tokens, one for each Organisation.  For Tekton
+to mount the correct Kubernetes secret for a given Organisation we need to annotate each secret with the owner that
+matches the Organisation, so we can filter out other secrets during the `creds-init` pipeline step.
+
+   ```yaml
+   apiVersion: v1
+   kind: Secret
+   metadata:
+     name: basic-user-pass
+     annotations:
+       tekton.dev/git-0: https://github.com
+       tekton.dev/githubapp-owner: <github organisation>
+   type: kubernetes.io/basic-auth
+   stringData:
+     username: <username>
+     password: <password>
+   ```
+
+__NOTE__ GitHub Apps issue a short lived token which needs to be continually refreshed.
+
 ## Basic authentication (Docker)
 
 1. Define a `Secret` containing the username and password that the build should
--- a/pkg/pod/creds_init.go
+++ b/pkg/pod/creds_init.go
@@ -36,7 +36,7 @@ import (
 //
 // If it finds secrets, it also returns a set of Volumes to attach to the Pod
 // to provide those secrets to this initialization.
-func credsInit(credsImage string, serviceAccountName, namespace string, kubeclient kubernetes.Interface, volumeMounts []corev1.VolumeMount, implicitEnvVars []corev1.EnvVar) (*corev1.Container, []corev1.Volume, error) {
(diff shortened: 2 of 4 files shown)


docs/auth.md
```diff
@@ -175,6 +175,32 @@ generated containing the credentials configured in the `Secret`, and these
...
+so we can filter out other secrets during the `creds-init` pipeline step.
+
+   ```yaml
+   apiVersion: v1
+   kind: Secret
+   metadata:
+     name: basic-user-pass
+     annotations:
+       tekton.dev/git-0: https://github.com # Described below
+       tekton.dev/githubapp-owner: <github organisation>
```
wlynch: We should also include App ID and Installation ID, this way one cluster could potentially reference multiple Apps.

docs/auth.md
```diff
@@ -175,6 +175,32 @@ generated containing the credentials configured in the `Secret`, and these
 credentials are then used to authenticate when retrieving any
 `PipelineResources`.
 
+## GitHub Apps
+
+[GitHub Apps](https://developer.github.com/apps/about-apps/) are the preferred way for automated operations say from
+within a pipeline to interact with GitHub, rather than managing a 'bot' user and personal access token.
+
+A GitHub App is installed into a GitHub Organisation and an API is used to retrieve a token that can then be used to
```
wlynch: nit: GitHub Apps can be installed on orgs or users. We should clean up the docs/examples to reflect this.

rawlingsj: good call - fixed

docs/auth.md
rawlingsj: Ok cool, I've not looked at using multiple GitHub Apps for a single Tekton deployment yet.  I wonder if we could do that as a follow up PR so that we keep this one a bit simpler?

bobcatfish:
Whoops looks like this PR really fell by the wayside :( @sbwsg might be relevant to your current creds investigations?

rawlingsj:
@bobcatfish np! :) If there's a better way of doing things being worked out then that's fine by me.  We're happy to use whatever approach.

pkg/pod/creds_init.go
```diff
@@ -55,7 +55,10 @@ func credsInit(credsImage string, serviceAccountName, namespace string, kubeclie
 		if err != nil {
 			return nil, nil, err
 		}
-
+		// if there's a github app annotation on the secret let's match that to the github org this pipeline is for
+		if owner != "" && secret.Annotations[gitHubAppAnnotationKey] != "" && secret.Annotations[gitHubAppAnnotationKey] != owner {
```
sbwsg, line 59: If there's a github app annotation on the secret but no gh org on the pipeline, should the secret still be mounted?  It looks like it would be at the moment.

pkg/pod/creds_init_test.go
```diff
@@ -118,7 +118,123 @@ func TestCredsInit(t *testing.T) {
...
+		}},
+	}
+
+	for _, c := range []struct {
+		desc  string
+		want  *corev1.Container
+		objs  []runtime.Object
+		owner string
+	}{{
+		desc: "github organisation is foo so my-creds1 should be mounted into the container",
```
sbwsg, line 152: worth adding a testcase for `owner: ""` to regression test the expected behaviour for that situation?

bobcatfish:
@rawlingsj @sbwsg with the current work being done to overhaul creds init it feels like it might make sense to close this for now?

or @sbwsg do you think it would still make sense to resolve the conflicts and merge this in the meantime?

sbwsg:
Let's close for now, but happy to reopen if @rawlingsj wants to revisit.

Issue #1273, opened 2019-09-04
Feature Request: Task Output Type of `Param`


kbruner:
It would be eminently useful for Tasks to have an output type of 'param', which could then be passed to the next task in a PipelineRun as an input param.  

While other people might envision other use cases, a single type of `string` should be fine.  Other data could be put into string form via JSON or however, or Base64-encoded by the user.  I'm also not sure what a reasonable size cap would be, although for the use cases I picture, 1kb should be reasonable?

sbwsg:
In principle I agree that this could be useful but I can't think of concrete examples - could you describe a bit more the use case you have for this?  What's the problem that you're trying to solve & in what way is it not served by, for example, PipelineResources?

kbruner:
The only PipelineResource type that's really suitable for this use case would be a bucket.  However, I have issues with how using a bucket for caching data between steps actually works.

Also note that our current use case is to use tekton to create and destroy temporary infrastructure based mainly on terraform.  We're not doing git-based builds, so making caching work with a git repository somehow would be, well, a lot of work.

1. Even though I have defined an artifact bucket for the pipelines controller to use, work outputs don't get pushed to it unless I *also* create a storage PipelineResource.  That makes no sense to me.  If /workingdir/output could reliably (or at least as a configuration option) by saved to the artifact bucket and then pulled into to /workingdir for the next task, yes, that would solve 99% of use cases.
2. Using a storage PipelineResource as a cache between tasks in a PipelineRun has some other nuisances:
  a. Unlike the artifact bucket, it only supports absolute paths, so it requires some magic in the tasks to make sure concurrent runs of the same pipeline don't clobber/mix up files.
  b. It doesn't clean up interim files between runs.  The tasks would have to implement the same PipelineRun/Task file structure for input and outputs that the artifact bucket already does.
3. Params can be plugged into multiple Task attributes, not just accessed after the actual container runtime environment has begun.  That can't be done easily/at all if the value needed is in a bucket blob.

bobcatfish:
I think this makes a lot of sense and I've run into some similar issues myself, for example in the nightly release pipeline (adding in https://github.com/tektoncd/pipeline/pull/1274) I need to generate a version tag, then pass that between steps. Between steps it isnt so bad b/c I can write it to a file on disk and read it into an environment variable but:

- it would be even harder between tasks
- even between steps it would be good to have a better way to do it

skaegi:
[...]

PR #1888, opened 2020-01-16
Changes to add results to a task

othomann:
# Changes
This is the first set of changes that modifies the task spec to add results. It also modifies the entrypoint to feed the termination log with the expected results.
It also adds a new cli argument to the entrypoint cmd (-results).

Part of #1273 

# Submitter Checklist

review them:

- [x] Includes [docs] Updated developer documentation

## Reviewer Notes

# Release Notes

```
Add results to the task spec and handle all changes in the entrypoint code.
```


pkg/apis/pipeline/v1alpha1/resource_types.go
```diff
@@ -123,6 +123,7 @@ type PipelineResourceResult struct {
 	Key         string              `json:"key"`
 	Value       string              `json:"value"`
 	ResourceRef PipelineResourceRef `json:"resourceRef,omitempty"`
+	ResultType  string              `json:"type,omitempty"`
```
eddie4941: Im not super familiar with go lang, but perhaps something like and enum type is useful here. 

I think in go you can do something like this: 

```
	ResourceRef PipelineResourceRef `json:"resourceRef,omitempty"`
	ResultType  ResultType          `json:"type,omitempty"`
}

type ResultType string

const (
	PipelineResourceResultType = "PipelineResourceResult"
	TaskRunResultType          = "TaskRunResult"
	UnknownResultType          = ""
)
```

That way we don't have to hard code the strings everywhere and can use the constants instead.

pkg/entrypoint/entrypointer.go
```diff
@@ -18,13 +18,19 @@ package entrypoint
...
 	"time"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/logging"
 	"github.com/tektoncd/pipeline/pkg/termination"
 )
 
+const (
+	// DefaultResultPath is the path for task result
+	DefaultResultPath = "/tekton/results/"
```
eddie4941: We probably need to create this directory on the file system somewhere so that its available for task containers. 

Im not sure if theres any place in the code where tekton performs the filesystem setup. I found this line https://github.com/tektoncd/pipeline/blob/c55d311ffc4b277f5a64a3787341d68d0288ff00/pkg/pod/workingdir_init.go#L71 that initializes the workspace. Im not sure if thats the right place for stuff under `/tekton` though.

pkg/pod/entrypoint.go
```diff
@@ -109,6 +110,18 @@ func orderContainers(entrypointImage string, steps []corev1.Container) (corev1.C
 				"-post_file", filepath.Join(mountPoint, fmt.Sprintf("%d", i)),
 				"-termination_path", terminationPath,
 			}
+			if results != nil && len(steps) == 1 {
```
eddie4941: Maybe wrap the logic to append the results command line flags in a function and reuse?

pkg/apis/pipeline/v1alpha1/resource_types.go
othomann: @sbwsg any thoughts?

pkg/entrypoint/entrypointer.go
othomann: @bobcatfish @sbwsg any suggestions?

pkg/apis/pipeline/v1alpha1/resource_types.go
sbwsg: It seems reasonable to use an enum type so that we don't hard code the strings.  I don't think we necessarily need the `UnknownResultType` though since the contract here is entirely internal to tekton.  Technically any string that isn't "PipelineResourceResult" or "TaskRunResult" is unknown.  And if we saw a result type that we didn't know about then I think it'd just be an error.

pkg/entrypoint/entrypointer.go
sbwsg: I think the entrypoint should create that directory when it starts up if it doesn't exist already.  I don't see a need for that folder to be a shared volume (yet).

pkg/entrypoint/entrypointer.go
```diff
@@ -100,12 +109,39 @@ func (e Entrypointer) Go() error {
...
+			logger.Fatalf("Error while handling results: %s", err)
+		}
+	}
 	if wErr := termination.WriteMessage(e.TerminationPath, output); wErr != nil {
 		logger.Fatalf("Error while writing message: %s", wErr)
 	}
 	return err
 }
 
+func (e Entrypointer) handleResultFiles() error {
```
sbwsg: nit: suggest `readResultsFromDisk` as more explicit func name.

pkg/entrypoint/entrypointer.go
```diff
@@ -100,12 +109,39 @@ func (e Entrypointer) Go() error {
...
+	output := []v1alpha1.PipelineResourceResult{}
+	for _, resultFile := range e.Results {
+		// check result file
+		fileContents, err := ioutil.ReadFile(DefaultResultPath + resultFile)
+		if err != nil {
+			return err
```
sbwsg: I think it's ok if the file is missing:

```go
if os.IsNotExist(err) {
  continue
} else if err != nil {
  return err
}
```

pkg/pod/entrypoint.go
sbwsg: I'd like to propose that the results arg is always appended to step containers if results != nil.  This way we don't need to worry about the distinction between `1` and `len - 1` cases.

This will mean that every step's entrypointer will search for results and write them to their termination log.  It will then be the TaskRun controller's responsibility to aggregate the results of all containers into the TaskRun Results field, performing any validation (such as checking whether a result is missing) as needed.

pkg/entrypoint/entrypointer.go
```diff
@@ -100,12 +109,39 @@ func (e Entrypointer) Go() error {
...
 		logger.Fatalf("Error while writing message: %s", wErr)
 	}
 	return err
 }
 
+func (e Entrypointer) handleResultFiles() error {
+	output := []v1alpha1.PipelineResourceResult{}
+	for _, resultFile := range e.Results {
+		// check result file
+		fileContents, err := ioutil.ReadFile(DefaultResultPath + resultFile)
```
sbwsg: suggest using `filepath.Join(DefaultResultPath, resultFile)` here to account for operating system specific separators.

pkg/apis/pipeline/v1alpha1/task_types.go
```diff
@@ -64,6 +64,18 @@ type TaskSpec struct {
 
 	// Workspaces are the volumes that this Task requires.
 	Workspaces []WorkspaceDeclaration `json:"workspaces,omitempty"`
+
+	// Results are all possible results of this Task
```
sbwsg: Suggest `Results are values that this Task can output`

pkg/entrypoint/entrypointer.go
```diff
@@ -100,12 +107,44 @@ func (e Entrypointer) Go() error {
...
+		if os.IsNotExist(err) {
+			continue
+		} else if err != nil {
+			return err
+		}
+		// if the file doesn't exist, ignore it
+		output = append(output, v1alpha1.PipelineResourceResult{
+			Key:        resultFile,
+			Value:      string(fileContents),
+			ResultType: "result",
```
eddie4941: Can we use the new `ResultType` here instead of hard coding this string?

bobcatfish:
```
I0117 21:16:30.477] 2020/01/17 21:16:30 main.go:734: provider gke, will acquire project type gke-project from boskos
I0117 21:21:30.475] 2020/01/17 21:21:30 main.go:316: Something went wrong: failed to prepare test environment: --provider=gke boskos failed to acquire project: resources not found
```

yikes! twice in a row :thinking:

vdemeester:
/area api
/kind api-change

@othomann this will also need some docs update :angel:

othomann:
@vdemeester What kind of doc update do you want?

cmd/entrypoint/main.go
```diff
@@ -51,6 +52,12 @@ func main() {
 		Waiter:          &realWaiter{},
 		Runner:          &realRunner{},
 		PostWriter:      &realPostWriter{},
+		Results:         strings.Split(*results, ","),
+	}
+	if len(e.Results) >= 1 && e.Results[0] != "" {
```
sbwsg: I am curious - what is the purpose of the `e.Results[0] != ""` check?

sbwsg:
Since this change is mostly internal to Tekton (for now... it will be exposed to users in an upcoming PR based on this work) I think the best place for documentation would be in `docs/developers/README.md`.  We could add a section called "How task results are defined and outputted by a task" or something like that.  This could then explain the -results flag for the entrypoint, the directory that is created for results, the files that are written for each result, and finally the way this data is formatted in the termination log of each step.

Does that make sense?

cmd/entrypoint/main.go
othomann: This is because strings.Split("", ",") returns an array with an empty string in it. So an empty string means we try to read the folder as a file and it fails.

othomann:
@sbwsg should the doc update be part of this PR?

sbwsg:
@othomann yeah I think so.

othomann:
@sbwsg ok done. Commits squashed and tests are running again. Let me know if there is anything else that is needed to get this approved.

docs/developers/README.md
```diff
@@ -195,3 +195,34 @@ but an Error when the sidecar exits with an error. This is only apparent when
...
+    - name: print-date
+      image: bash:latest
+      args:
+        - "-c"
+        - |
+          date > /tekton/results/current-date
+```
+
+The result is added to a file name with the specified result's name into the `/tekton/results` folder. This is then added to the
+task run status.
```
vdemeester: nit: can be done later, it would be nice to show how that part of the TaskRun status looks :angel:

vdemeester:
```
I0122 10:31:14.503] pkg/pod/status.go:259:13: unslice: could simplify podInstance.Status.ContainerStatuses[:] to podInstance.Status.ContainerStatuses (gocritic)
I0122 10:31:14.503] 	sort.Slice(podInstance.Status.ContainerStatuses[:], func(i, j int) bool {
I0122 10:31:14.503] 	           ^
I0122 10:31:14.503] cmd/pullrequest-init/main.go:21: File is not `goimports`-ed (goimports)
I0122 10:31:14.504] 	"fmt"
```

Interesting, the rebase made the lint fail :sweat_smile:

docs/developers/README.md
othomann: This is not part of this PR. The next PR once this one is merged will update the status.

othomann:
@vdemeester I haven't changed the files `pkg/pod/status.go` or `cmd/pullrequest-init/main.go` so I don't see why this is reported against this PR.
Once this is merged, I'll work with Eduardo to get the next PR for updating the task run status.

docs/developers/README.md
vdemeester: oh right, then forget what I just said :joy_cat:

vdemeester:
@othomann it was some previous PR merged that did that, it got fixed :wink:

PR #1894, opened 2020-01-17
Bump kubernetes to 1.16.5 and knative/pkg to release-0.12

### LINKED ISSUE #1113: Make sure the webhook is always registered before allowing user to create resources
As of today, `knative/pkg`'s code is the one responsible to register the webhook on kubernetes. This means, the following are possible :
- The webhook fails to start (for any reason), it does not get registered, thus it's not applied
- The webhook takes time to start (slow network, …), it's not registered quickly enough, a user can create resources

In all those cases, the user can end up creating resources that are either not valid or not mutated (with defaults values)

There is two way to fix these cases:

1. Move the webhook registration in the `yaml`s, so that the webhook is registered as soon as the CRDs are created, *before* it started, and thus it will fail if the webhook is not running
2. Having the controller validate

I have a **very very high** preference for solution `1.` .

vdemeester:
# Changes

This updates `cmd/webhook/main.go` to use latest knative/pkg code, most simplified

This also changes the following:
- Add validating and mutating webhook definition *in* yaml (so it gets
  registered before the webhook starts)
- <del>Rename the controller service/deployment from `tekton-pipelines-controller` to `controller`</del>
- <del>Rename the webhook service/deployment from `tekton-pipelines-webhook` to `webhook`</del>
- Use `pipeline.tekton.dev/release` instead of `tekton.dev/release` ;
  mainly to let other project use their own (`trigger.tekton.dev/release`).

/hold waiting for 0.10 to be out to get this merge

Also, this moves the min k8s server version to 1.15.x

Closes #1113

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Bump kubernetes dependency to 1.16 :angel: 
```


cmd/webhook/main.go
```diff
@@ -18,29 +18,127 @@ package main
...
+		// The path on which to serve the webhook.
+		"/defaulting",
+
+		// The resources to validate and default.
+		types,
+
+		// A function that infuses the context passed to Validate/SetDefaults with custom metadata.
+		func(ctx context.Context) context.Context {
+			// FIXME(vdemeester) uncomment that for auto-conversion
+			// return v1alpha2.WithUpgradeViaDefaulting(store.ToContext(ctx))
```
mattmoor: Conversion webhooks are imminent in knative/pkg btw cc @dprotaso

vdemeester: :angel:

pkg/reconciler/pipelinerun/metrics_test.go
```diff
@@ -101,7 +111,7 @@ func TestRecordPipelineRunDurationCount(t *testing.T) {
 			err = metrics.DurationAndCount(test.taskRun)
 			assertErrIsNil(err, "DurationAndCount recording recording got an error", t)
 			metricstest.CheckDistributionData(t, "pipelinerun_duration_seconds", test.expectedTags, 1, test.expectedDuration, test.expectedDuration)
-			metricstest.CheckCountData(t, "pipelinerun_count", test.expectedTags, test.expectedCount)
+			metricstest.CheckCountData(t, "pipelinerun_count", test.expectedCountTags, test.expectedCount)
```
vdemeester: @mattmoor @hrishin wasn't sure about this… I don't fully understand the behavior change there :sweat_smile:

mattmoor: This was a while ago, but I think maybe done by @anniefu?

vdemeester:
I think it's gonna be green :tada: :green_heart: 
@mattmoor if you have a little bit of time to review the changes :angel: :stuck_out_tongue_closed_eyes:

mattmoor:
Have a parent/teacher thing now, but will TAL after.  Thanks for doing this!!! 🙏👍

bobcatfish:
Just double checking, but knative is also requiring 1.15 or greater now? I think the GKE default is still 1.13 but 1.15 is available :thinking: 

Anyway assuming that:
/lgtm

pkg/reconciler/pipelinerun/metrics_test.go
anniefu: Hmm, I created the metricstest helper package a while ago, but didn't do this particular test change. 

A cursory glance implies that someone added a new tag to the "pipelinerun_count" metric so it has a different set of metrics tags as the "pipelinerun_duration_seconds" metric now, which isn't a problem.

config/500-webhooks.yaml
```diff
@@ -0,0 +1,120 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: Secret
+metadata:
+  name: webhook-certs
+  namespace: tekton-pipelines
+  labels:
+    tekton.pipeline.dev/release: devel
```
mattmoor:
```suggestion
    pipeline.tekton.dev/release: devel
```

config/controller.yaml
```diff
@@ -14,27 +14,27 @@
...
     app.kubernetes.io/name: tekton-pipelines
     app.kubernetes.io/component: controller
-    tekton.dev/release: "devel"
+    pipeline.tekton.dev/release: "devel"
 spec:
   replicas: 1
   selector:
     matchLabels:
-      app: tekton-pipelines-controller
+      app: controller
```
mattmoor: I didn't think changing this was upgrade-safe 🤔

config/webhook.yaml
```diff
@@ -18,44 +18,78 @@ metadata:
...
-          value: tekton-pipelines-webhook
-      volumes:
-        - name: config-logging
-          configMap:
-            name: config-logging
+          value: webhook
+        - name: METRICS_DOMAIN
+          value: tekton.dev/pipeline
+        securityContext:
+          allowPrivilegeEscalation: false
```
mattmoor: You should change `.ko.yaml` to match the defaultBaseImage from serving, which uses `:nonroot`

tekton/pipeline-release-run-5bgpg.log
```diff
@@ -0,0 +1,10128 @@
+[precheck : git-source-tekton-pipelines-git-v0-10-0-fxv9c] {"level":"info","ts":1579772695.226866,"logger":"fallback-logger","caller":"logging/config.go:69","msg":"Fetch GitHub commit ID from kodata failed: open /var/run/ko/refs/heads/master: no such file or directory"}
```
mattmoor: Did you mean to include this gigantic file?

vdemeester: I did not 🧐😆

config/controller.yaml
vdemeester: Good point :sweat_smile: I could (and probably should revert that change), but I though, if we wanna break people, let's break them now instead of real soon when we are beta & co…

/cc @sbwsg @bobcatfish @afrittoli

mattmoor:
This LGTM

config/controller.yaml
sthaha: Just noting that this change would break operator as well; see https://github.com/tektoncd/operator/blob/master/pkg/controller/setup/constants.go

vdemeester:
I'll split the `:nonroot` in a separate PR (and I'll revert the controller/webhook name change :stuck_out_tongue: )

vdemeester:
It is green :tada: :man_dancing:

bobcatfish:
We should probably update some docs when we bump the k8s dependency also, e.g.:
- https://github.com/tektoncd/pipeline/blob/master/DEVELOPMENT.md#kubernetes-cluster
- https://github.com/tektoncd/pipeline/blob/master/docs/install.md#pre-requisites

bobcatfish:
And the release notes too, ill add something

mattmoor:
It is probably useful to also formulate some guidance around how these versions are bumped.  We recently went through this exercise, and I'm happy to share if it's something you think it'd be worth aligning on.

PR #1927, opened 2020-01-23
Avoid a panic when sorting container statuses

ImJasonH:
If a step fails, subsequent steps might not have a terminated time yet,
which led to a panic and crashloop backoff in the controller.

The only place where `sortContainerStatuses` is used is when deriving the failure message (the first `Message` of any terminated step, i.e., where `Terminated != nil`, so sorting these to the beginning of the list has no practical effect.

# Submitter Checklist

review them:

/assign @sbwsg

### DIFF
--- a/pkg/pod/status.go
+++ b/pkg/pod/status.go
@@ -257,7 +257,14 @@ func areStepsComplete(pod *corev1.Pod) bool {
 
 func sortContainerStatuses(podInstance *corev1.Pod) {
 	sort.Slice(podInstance.Status.ContainerStatuses, func(i, j int) bool {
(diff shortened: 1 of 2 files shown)


sbwsg:
oof.  this seems like it might be worth doing a point release for?

/lgtm
/approve

ImJasonH:
> oof. this seems like it might be worth doing a point release for?

Might be worth it. FWIW it was introduced in https://github.com/tektoncd/pipeline/commit/fc349ee5ec3f203fee1662f7c1dcbb39814ccda7#diff-9f8b8a92c4961f27c24bcfee2d137a59

PR #1736, opened 2019-12-11
Script mode set -xe by default when no shebang used

sbwsg:
# Changes

Fixes #1735

When I use script mode without a #! the current behaviour is to insert
one for me - `#!/bin/sh`. This is helpful and removes an unneccessary
line from my script. In addition I'd like two things by default: First
I want the script to error out if any of the commands error out. Second
I'd like to see the commands the script runs before their output. These
two features are baked into `sh` with the `set -xe` flags.

After this change the default behaviour of script mode when no shebang
is provided is to insert both the default shebang as well as the flags
required to make shell errors and shell echos happen.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Updates [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

# Release Notes

```
Script mode will now insert `set -xe` to scripts that don't include a shebang #!, which will in turn fail scripts with commands that error our and also print the command that is about to run. This default behaviour can be disabled by simply specifying your own #! at the top of the script.
```

PR #1936, opened 2020-01-24
Workspace "from" clauses ensure task ordering

sbwsg:
# Changes

Workspaces allow users to wire PVCs through the tasks of theirpipelines. Unfortunately, multiple tasks all trying to mount a PVC at once can result in unpleasant conflicts.
    
To combat the situation where multiple Task pods are fighting over a PVC, "from" clauses have been introduced for workspaces. One task can explicitly declare that it will use a workspace from a previous task. When Tekton sees a from clause linking one task's workspace to another it will ensure that the sequence of those tasks is enforced. 

Here's a relevant snippet of YAML to show how this PR currently operates:

```yaml
  tasks:
    - name: task1
      taskRef:
        name: write-to-workspace
      workspaces:
        - name: output
          workspace: pipeline-ws1
    - name: task2
      taskRef:
        name: read-from-workspace
      workspaces:
        - name: src
          from:
            task: task1
            name: output
```

Here, task2 will run after task1 because task2 declares that its src workspace should be populated using whatever volume was used as the output workspace from task1.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Pipeline Tasks that need to share a workspace (for example a PVC to be written to and read from by multiple tasks) can now explicitly indicate the order in which they should run (and therefore access the workspace) by using "from" clauses.
```


pritidesai:
How would it behave if a pipeline has three or more tasks?

```
tasks:
    - name: task1
      taskRef:
        name: write-to-workspace
      workspaces:
        - name: output
          workspace: pipeline-ws1
    - name: task2
      taskRef:
        name: read-from-workspace
      workspaces:
        - name: src
          workspace: $(task1.output)
    - name: task3
      taskRef:
        name: process-workspace
      workspaces:
        - name: src
          workspace: $(task1.output)
```

Would it enforce `task2` execution before `task3`? Or this use case is out of scope of this discussion and falls under `runAfter`?

IMO, `$()` is so far conceived as parameter interpolation and does not guarantee task execution order. It might create confusion in the beginning, if we choose to go with `$()`. I think `from` under `workspaces` makes ideal solution but we are adding one more field in the YAML.

sbwsg:
> Would it enforce task2 is executed before task3?

Hm, the way to ensure that currently would be to have task2 use `workspace: $(task3.src)`.  Then task2 would wait until task3 is done.

> IMO, $() is so far conceived as parameter interpolation and does not guarantee task execution order. It might create confusion in the beginning, if we choose to go with $().

Yeah that's a good point, I agree.  It would also be confusing because variables are normally strings, but here the user isn't able to interpolate anything but other workspace names (i.e. they can't use a param to inject a workspace name).  I think I'm going to change this to use a From field.

sbwsg:
I've rewritten this to use "from" syntax instead and I think it's largely ready for review now.  Removed WIP label.

pritidesai:
thanks @sbwsg, validated one more use case with workspace from second task `read-from-workspace`:

```
kind: Pipeline
metadata:
  name: fetch-and-print-recipe
spec:
  workspaces:
  - name: password-vault
  - name: recipe-store
  - name: shared-data
  tasks:
  - name: fetch-the-recipe
    taskRef:
      name: fetch-secure-data
    workspaces:
    - name: super-secret-password
      workspace: password-vault
    - name: secure-store
      workspace: recipe-store
    - name: filedrop
      workspace: shared-data
  - name: print-the-recipe
    taskRef:
      name: print-data
    params:
    - name: filename
      value: recipe.txt
    workspaces:
    - name: storage
      from:
        task: fetch-the-recipe
        name: filedrop
  - name: one-more-print-the-recipe
    taskRef:
      name: print-data
    params:
      - name: filename
        value: recipe.txt
    workspaces:
      - name: storage
        from:
          task: print-the-recipe
          name: storage
```

/lgtm

sbwsg:
ah, thank you for checking it out @pritidesai! I think I might add an example that does this as well to make sure multiple "from"s chain correctly.

pritidesai:
/lgtm

Would also like to hear @skaegi's comments on this which we couldn't get to during the demo in working group.

pritidesai:
@sbwsg I came across one more use case, `runAfter` lets us specify a list of tasks, is it possible for such task to consume workspace from two or more different tasks? Should the dependent tasks run sequentially?

sbwsg:
> is it possible for such task to consume workspace from two or more different tasks? Should the dependent tasks run sequentially?

Good point - I've added another example to cover this.  When workspaces of a task are "from" multiple previous tasks each of those should complete before the task relying on them runs.  Does [this example](https://github.com/tektoncd/pipeline/pull/1936/files#diff-c1a436b34e7fbd6f7f972bfe785777fd) describe what you're expecting here @pritidesai ?

examples/pipelineruns/workspaces.yaml
```diff
@@ -104,15 +124,25 @@ spec:
   - name: print-the-recipe
     taskRef:
       name: print-data
-    # Note: this is currently required to ensure order of write / read on PVC is correct.
-    runAfter:
-      - fetch-the-recipe
```
bobcatfish: neat!

pkg/apis/pipeline/v1alpha1/pipeline_types.go
```diff
@@ -35,7 +35,7 @@ type PipelineSpec struct {
 	// Workspaces declares a set of named workspaces that are expected to be
 	// provided by a PipelineRun.
 	// +optional
-	Workspaces []WorkspacePipelineDeclaration `json:"workspaces,omitempty"`
+	Workspaces []PipelineWorkspaceDeclaration `json:"workspaces,omitempty"`
```
bobcatfish: good call

pkg/apis/pipeline/v1alpha1/pipeline_types.go
```diff
@@ -152,6 +152,12 @@ func (pt PipelineTask) Deps() []string {
 			deps = append(deps, rd.From...)
 		}
 	}
+	// Add any dependents declared via workspace from clauses
```
bobcatfish: this isnt your doing but it's not clear to me what this `Deps` function is for exactly - any chance you'd add a docstring for it? 🙏  no worries if not

pkg/apis/pipeline/v1alpha1/pipeline_validation_test.go
```diff
@@ -310,7 +310,7 @@ func TestPipeline_Validate(t *testing.T) {
 		p: tb.Pipeline("name", "namespace", tb.PipelineSpec(
 			tb.PipelineWorkspaceDeclaration("foo"),
 			tb.PipelineTask("taskname", "taskref",
-				tb.PipelineTaskWorkspaceBinding("taskWorkspaceName", "pipelineWorkspaceName")),
+				tb.PipelineTaskWorkspaceBinding("taskWorkspaceName", tb.PipelineTaskWorkspaceBindingWorkspace("pipelineWorkspaceName"))),
```
bobcatfish: i think maybe there should be some more unit test cases added here? or are they caught somewhere else? (i remember the validation unit tests being a bit hard to puzzle out what is tested where...)

pkg/reconciler/pipelinerun/resources/workspaces_test.go
```diff
@@ -0,0 +1,166 @@
...
+		t.Run(tc.description, func(t *testing.T) {
+			_, err := ResolveWorkspaces(tc.in.Spec.DeepCopy())
+			if err == nil {
+				t.Errorf("no error received")
+			} else if err.Error() != tc.errorMessage {
+				t.Errorf("expected error message %q received %q", tc.errorMessage, err.Error())
+			}
+		})
+	}
+}
```
bobcatfish: 🎉

pkg/reconciler/pipelinerun/resources/workspaces_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+func TestResolveWorkspaces_Invalid(t *testing.T) {
+	for _, tc := range []struct {
+		description  string
+		in           *v1alpha1.Pipeline
+		errorMessage string
```
bobcatfish: this is mostly personal preference but i usually avoid testing against error messages - feels like there's not a lot of value if they change a bit and then the test case needs to change too (similar feelings for log messages)

but im guessing the reason might actually be that you want to make sure the error happened for hte right reason? if so, then i have 2 alternatives to suggest:
1. return typed errors and assert agains the type
2. break up the funcitonality into multiple functions, which each fail or not for their own reasons

pkg/reconciler/pipelinerun/resources/workspaces_test.go
```diff
@@ -0,0 +1,166 @@
...
+			),
+		)),
+	}} {
+		t.Run(tc.description, func(t *testing.T) {
+			result, err := ResolveWorkspaces(tc.in.Spec.DeepCopy())
+			if err != nil {
+				t.Errorf("error resolving workspace: %v", err)
+			}
+			if d := cmp.Diff(&tc.expected.Spec, result); d != "" {
+				t.Errorf("ResolveWorkspaces() got diff %s", d)
```
bobcatfish: i find it really handy to include the `-want +got` thing in the error

pkg/reconciler/pipelinerun/resources/workspaces.go
```diff
@@ -0,0 +1,56 @@
+package resources
+
+import (
+	"fmt"
+
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+)
+
+func ResolveWorkspaces(p *v1alpha1.PipelineSpec) (*v1alpha1.PipelineSpec, error) {
```
bobcatfish: can has docstring explaining purpose, inputs and outputs?

pkg/reconciler/pipelinerun/resources/workspaces.go
```diff
@@ -0,0 +1,56 @@
...
+	spec := p.DeepCopy()
```
bobcatfish: maybe a comment explaining why a deepcopy is needed? (im guessing b/c we're going to start making changes to it?)

pkg/reconciler/pipelinerun/resources/workspaces.go
```diff
@@ -0,0 +1,56 @@
...
+					Name:      pt.Workspaces[wsIdx].Name,
+					Workspace: resolvedName,
+				}
+			}
+		}
+	}
+	return spec, nil
+}
+
+func getResolvedWorkspaceName(pts map[string]*v1alpha1.PipelineTask, ws *v1alpha1.PipelineTaskWorkspaceBinding) (string, error) {
```
bobcatfish: i have a strong feeling this could be exported and unit tested on its own - maybe we want a `workspaces` package

pkg/reconciler/pipelinerun/resources/workspaces.go
```diff
@@ -0,0 +1,56 @@
...
+SEARCH_FOR_WORKSPACE:
+	for i := 0; i < len(pts); i++ {
+		if ws.Workspace != "" {
+			return ws.Workspace, nil
+		} else if ws.From.Task != "" {
+			if nextTask, ok := pts[ws.From.Task]; ok {
+				for wsi := range nextTask.Workspaces {
+					if nextTask.Workspaces[wsi].Name == ws.From.Name {
+						ws = &nextTask.Workspaces[wsi]
+						continue SEARCH_FOR_WORKSPACE
```
bobcatfish: im really not sure why Go allows gotos, but imo we should avoid them - maybe we can find a way to make this work with a function call instead?

pkg/reconciler/pipelinerun/resources/workspaces.go
```diff
@@ -0,0 +1,56 @@
...
+					}
+				}
+				return "", fmt.Errorf("pipeline task %q has no workspace named %q", ws.From.Task, ws.From.Name)
+			}
+			return "", fmt.Errorf("pipeline task %q does not exist", ws.From.Task)
+		}
+		break
```
bobcatfish: this break is so far away from the if statement that gates it i cant really tell why its happening

bobcatfish:
Oh yeah one more thought, maybe you have an example that handles this, but what happens when you fan-out, i.e. use the same workspace in 2+ parallel tasks? can all the tasks safely write to it? in PipelineResources we actually copied the data to and from a local directory vs having ppl write to the PVC directly. (still would have had race conditions tho if multiple parallel tasks are trying to update the same resource :S)

examples/pipelineruns/workspaces-multiple-from.yaml
```diff
@@ -0,0 +1,180 @@
...
+  workspaces:
+  - name: ws1
+    readOnly: true
+  - name: ws2
+    readOnly: true
+  - name: result
+  steps:
+  - name: calculate-sum
+    image: ubuntu
+    script: |
```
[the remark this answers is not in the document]
sbwsg: We get `set -xe` for free now ever since https://github.com/tektoncd/pipeline/pull/1736 :tada:

examples/pipelineruns/workspaces-multiple-from.yaml
```diff
@@ -0,0 +1,180 @@
...
+---
+apiVersion: tekton.dev/v1alpha1
+kind: Pipeline
+metadata:
+  name: horrible-fibonacci
+spec:
+  workspaces:
+  - name: A
+  - name: B
+  - name: C
```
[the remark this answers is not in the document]
sbwsg: Good point, I've added a description field to the type and included them here.

sbwsg:
Thanks for the review @bobcatfish I've addressed all comments I think.

sbwsg:
Ah no, I missed the reconciler test.  Adding today!

sbwsg:
> Oh yeah one more thought, maybe you have an example that handles this, but what happens when you fan-out, i.e. use the same workspace in 2+ parallel tasks? can all the tasks safely write to it? in PipelineResources we actually copied the data to and from a local directory vs having ppl write to the PVC directly. (still would have had race conditions tho if multiple parallel tasks are trying to update the same resource :S)

Support for this is dictated by the volume I think.  Some types don't support multiple simultaneous writers while some do.  GKE's persistent volume implementation [doesn't support multiple simultaneous writers](https://cloud.google.com/kubernetes-engine/docs/concepts/persistent-volumes#access_modes).  We don't perform any copies of the data as we shuffle the PVCs around.  I believe the platform will complain through Pod errors if one tries to do that.

sbwsg:
aaand reconciler tests have been added

examples/pipelineruns/workspaces-multiple-from.yaml
```diff
@@ -0,0 +1,195 @@
...
+        task: second-add
+        name: result
+---
+apiVersion: tekton.dev/v1alpha1
+kind: PipelineRun
+metadata:
+  generateName: fib-
+spec:
+  pipelineRef:
+    name: horrible-fibonacci
```
pritidesai: @sbwsg I think you need to change this to `worlds-slowest-but-greatest-fibonacci` 🤔

pritidesai: this pipelinerun is failing with:

```
 kubectl get pipelinerun
NAME              SUCCEEDED   REASON               STARTTIME   COMPLETIONTIME
fib-pipelinerun   False       CouldntGetPipeline   26s         26s
```

pritidesai: @sbwsg I might not have pulled all your code to test with (because of the conflict), please ignore it if you are not hitting this but the pods are stuck in pending: 

```
kubectl get pods
NAME                                     READY   STATUS    RESTARTS   AGE
fib-pipelinerun-init-1-5rglh-pod-zvznr   0/1     Pending   0          19m
fib-pipelinerun-init-2-nnq6v-pod-bq8xs   0/1     Pending   0          19m
```

with a message:

```
"message": "pod status \"PodScheduled\":\"False\"; message: \"pod has unbound immediate PersistentVolumeClaims\"",
```

sbwsg: amazing, great catch.  I updated the pipelineref.

There are some remaining issues related to using regional PVCs in the example yaml.  I'm getting volume node affinity errors when I run it locally and it appears that similar issues are occurring here in the CI cluster as well.  Can't quite put my finger on what's going wrong but still looking into it.

sbwsg:
/hold

This feature is still a bit contentious and so I'm still in the process of discussing whether it's desirable / needed / a future-us problem / etc...

sbwsg:
Given that we're no longer going with "from" syntax for task results, and that the requirement for this feature remains under debate I'm going to close the pull request.  We can reopen at some point if we decide this is more than a nice-to-have.

sbwsg:
OK we're going to revisit this PR but using a variable interpolation syntax instead.

sbwsg:
Closing for now, intend to revisit at a later date.

examples/pipelineruns/workspaces-multiple-from.yaml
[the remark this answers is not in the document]
bobcatfish: oooo cool i didnt know that!

Issue #1856, opened 2020-01-13
Script support for sidecar


chmouel:
# Expected Behavior

Same as main steps, `script` working on `sidecars` steps

# Actual Behavior

Not supported

```
Error from server (InternalError): error when creating "sidecar.yaml": Internal error occurred: admission webhook "webhook.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "script"
```

# Steps to Reproduce the Problem

```yaml
  sidecars:
    - image: centos
      name: uploader
      script: |
        #!/usr/bin/env bash
        while :;do
            ls /workspace/output
            sleep 5
        done
```

# Additional Info

tomgeorge:
If no one is already working on this, I have a preliminary PR that I hope to open in the next few days.

vdemeester:
This is now the case as #1987 is merged :wink: 
/close

PR #1987, opened 2020-01-30
Add Script Support to Sidecars

tomgeorge:
# Changes

This makes a sidecar behave like a Step, and allows the ability to pass a script argument.

This enables #1856.

To implement this, I created a `Sidecar` type that was the same as `Step`.  

I then changed `convertScripts` to accept an image, a slice of steps, and a slice of sidecars, and return the init container, and two lists of containers: one for steps, and one for sidecars.  

I also refactored `convertScripts` to call a helper function, `convertListOfSteps`, that takes a list of steps to convert, a pointer to the init container to add to, a pointer to `placeScripts`, and a format string for writing out the heredoc and temp file names.  It returns the converted list of containers.  `convertListOfSteps` does most of the work that `convertScripts` does.  It might be a little ugly but it seemed better than putting everything in `convertScripts`.  __Unfortunately this doesn't translate very well to the Github PR viewer, but git diff works fine__

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Add script support to sidecars
```

### DIFF
--- a/examples/taskruns/sidecar-ready-script.yaml
+++ b/examples/taskruns/sidecar-ready-script.yaml
@@ -16,8 +16,6 @@ spec:
     steps:
     - name: check-ready
       image: ubuntu
-      # The step will only succeed if the sidecar has written this file, which
-      # it does 5s after it starts, before it reports Ready.
       script: cat /shared/message
       volumeMounts:
       - name: shared
--- a/examples/taskruns/sidecar-ready-script.yaml
+++ b/examples/taskruns/sidecar-ready-script.yaml
@@ -1,0 +1,30 @@
+apiVersion: tekton.dev/v1alpha1
+kind: TaskRun
+metadata:
+  generateName: sidecar-ready-
+spec:
+  taskSpec:
+    sidecars:
+    - name: slow-sidecar
+      image: ubuntu
+      script: |
+        echo "hello from sidecar" > /shared/message
+      volumeMounts:
+      - name: shared
+        mountPath: /shared
+
+    steps:
+    - name: check-ready
+      image: ubuntu
+      # The step will only succeed if the sidecar has written this file, which
+      # it does 5s after it starts, before it reports Ready.
+      script: cat /shared/message
+      volumeMounts:
+      - name: shared
+        mountPath: /shared
+
+    # Sidecars don't have /workspace mounted by default, so we have to define
+    # our own shared volume.
+    volumes:
+    - name: shared
+      emptyDir: {}
--- a/pkg/pod/pod_test.go
+++ b/pkg/pod/pod_test.go
@@ -355,10 +355,14 @@ func TestMakePod(t *testing.T) {
 				Image:   "primary-image",
 				Command: []string{"cmd"}, // avoid entrypoint lookup.
 			}}},
-			Sidecars: []corev1.Container{{
-				Name:  "sc-name",
-				Image: "sidecar-image",
-			}},
+			Sidecars: []v1alpha1.Sidecar{
+				{
+					Container: corev1.Container{
+						Name:  "sc-name",
+						Image: "sidecar-image",
+					},
+				},
+			},
 		},
 		wantAnnotations: map[string]string{},
 		want: &corev1.PodSpec{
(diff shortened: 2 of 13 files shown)


chmouel:
Nice! thank you! I think one thing missing would be a yaml in the examples directory which are used in CI for validation too..

tomgeorge:
I'll add an example this evening

examples/taskruns/sidecar-ready-script.yaml
```diff
@@ -0,0 +1,30 @@
...
+        echo "hello from sidecar" > /shared/message
+      volumeMounts:
+      - name: shared
+        mountPath: /shared
+
+    steps:
+    - name: check-ready
+      image: ubuntu
+      # The step will only succeed if the sidecar has written this file, which
+      # it does 5s after it starts, before it reports Ready.
```
sbwsg: I think I'm missing something here - where is the 5s delay happening?  And what is meant here by the sidecar reporting Ready?

pkg/pod/pod_test.go
```diff
@@ -394,6 +398,76 @@ func TestMakePod(t *testing.T) {
...
+		desc: "sidecar container with script",
+		ts: v1alpha1.TaskSpec{
+			Steps: []v1alpha1.Step{{Container: corev1.Container{
+				Name:    "primary-name",
+				Image:   "primary-image",
+				Command: []string{"cmd"}, // avoid entrypoint lookup.
+			}}},
+			Sidecars: []v1alpha1.Sidecar{
+				{
+					Container: corev1.Container{
```
sbwsg: Suggest collapsing these brackets to match format with the Step on line 404.  Similarly for the init container on line 422.

pkg/pod/script.go
```diff
@@ -61,7 +61,21 @@ func convertScripts(shellImage string, steps []v1alpha1.Step) (*corev1.Container
...
+		return &placeScriptsInit, convertedStepContainers, sidecarContainers
+	}
+	return nil, convertedStepContainers, sidecarContainers
+}
+
+// convertListOfSteps does the heavy lifting for convertScripts.
+//
+// It iterates through the list of steps (or sidecars), generates the script file name and heredoc termination string,
+// adds an entry to the init container args, sets up the step container to run the script, and sets the volume mounts.
+func convertListOfSteps(steps []v1alpha1.Step, initContainer *corev1.Container, placeScripts *bool, format string) []corev1.Container {
```
sbwsg: suggest "namePrefix" instead of "format" here, since the value isn't expected to be a format string.

pkg/pod/script.go
```diff
@@ -45,12 +45,12 @@ var (
...
 )
 
-// convertScripts converts any steps that specify a Script field into a normal Container.
+// convertScripts converts any steps and sidecars that specify a Script field into a normal Container.
 //
 // It does this by prepending a container that writes specified Script bodies
 // to executable files in a shared volumeMount, then produces Containers that
 // simply run those executable files.
-func convertScripts(shellImage string, steps []v1alpha1.Step) (*corev1.Container, []corev1.Container) {
+func convertScripts(shellImage string, steps []v1alpha1.Step, sidecars []v1alpha1.Sidecar) (*corev1.Container, []corev1.Container, []corev1.Container) {
```
sbwsg: Rather than amending the signature and adding the extra func to account for Sidecars, couldn't the caller (MakePod) simply execute this function twice?  Once for Steps and once for Sidecars?

examples/taskruns/sidecar-ready-script.yaml
tomgeorge: That was me copy/pasting another example without deleting the comments.  It isn't really waiting for anything to be ready.

pkg/pod/script.go
tomgeorge: Yes you can definitely do the same thing that way.  I think in order to do that you need to change the signature of `convertScripts` anyways to include context about the namePrefix.  Would you prefer I changed it to that?

sbwsg: That would be excellent, thank you @tomgeorge !

tomgeorge:
@sbwsg I made the formatting/comment changes you requested.  I need to think a little bit more about the best way to make the `convertScript` change, because if you call it twice in `MakePod` you have to keep track of the init container, instead of `convertScripts`.  Currently, `convertScripts` handles the init container and appends to it appropriately.

sbwsg:
@tomgeorge building up the script in the init container definitely complicates things. Looking at it again I think your original solution is good. I was initially wary of the passing a pointer to the placeScripts bool and init container but given the way the code worked before this change I think yours is a sound approach.  So I'm happy to leave things as-is with the two formatting changes.

/lgtm

PR #1991, opened 2020-02-01
Check for LimitRange Minimum for TaskRun Container Requests

### LINKED ISSUE #1045: Leave Resource Request Empty Instead of Requesting 0
I recently went through the [pipelines-tutorial](https://github.com/openshift/pipelines-tutorial) in the OpenShift org and encountered the following error message:

```
Missing or invalid Task pipelines-tutorial/s2i-java-8: pods
        "petclinic-deploy-pipelinerun-qwq6k-build-hnq2g-pod-f9e7df" is
        forbidden: [minimum memory usage per Container is 6Mi, but request is
        0., minimum memory usage per Container is 6Mi, but request is 0.,
        minimum memory usage per Container is 6Mi, but request is 0., minimum
        memory usage per Container is 6Mi, but request is 0., minimum memory
        usage per Container is 6Mi, but request is 0., minimum memory usage per
        Container is 6Mi, but request is 0., minimum memory usage per Container
        is 6Mi, but request is 0.]
```

This occurred because my OpenShift project had a `LimitRange` with the following specifications:

```yaml
kind: LimitRange
apiVersion: v1
metadata:
  name: pipelines-tutorial-core-resource-limits
  namespace: pipelines-tutorial
  selfLink: >-
    /api/v1/namespaces/pipelines-tutorial/limitranges/pipelines-tutorial-core-resource-limits
  uid: 492a15c6-9d90-11e9-ae59-0a8e513d9bd0
  resourceVersion: '1063283'
  creationTimestamp: '2019-07-03T12:44:23Z'
spec:
  limits:
    - type: Container
      max:
        memory: 6Gi
      min:
        memory: 6Mi
      default:
        cpu: 500m
        memory: 1536Mi
      defaultRequest:
        cpu: 50m
        memory: 256Mi
    - type: Pod
      max:
        memory: 12Gi
      min:
        memory: 6Mi
```

Since there were no specifications on the container cpu of resource request for the pipeline, the resource request was set to 0. This caused the `pipelinerun` as [part of the tutorial](https://github.com/openshift/pipelines-tutorial#trigger-pipeline) to fail.

If no amount had been requested, the `LimitRange`'s specifications would have been used for the `task`s' containers, and I think this should be the approach moving forward. I believe the source in question can be found [here](https://github.com/tektoncd/pipeline/blob/d494344ebf338a2869a4446d2e571d4006971a06/pkg/reconciler/v1alpha1/taskrun/resources/pod.go#L422).

In general, I think the error message should be improved to better reflect the situation. `Missing or invalid Task pipelines-tutorial/s2i-java-8` is not what was wrong with the `pipelinerun`. The issue was due to the cpu resource request being 0, which was less than the `LimitRange`'s min value.

More on how I ran into this issue can be found [here](https://github.com/openshift/pipelines-tutorial/issues/6).

# Expected Behavior

Error message should not have `Missing or invalid Task pipelines-tutorial/s2i-java-8` as reason `pipelinerun` failed. It should focus more on the `LimitRange` aspect of the issue.

The default value for a resource request should be the minimum of a `LimitRange` instead of 0.

# Actual Behavior

Error message in issue description.

1. Set a `LimitRange` minimum of greater than 0 on an OpenShift 4 cluster namespace and trigger a `pipelinerun` with the default resource request.

# Additional Info

This was run on a Red Hat Product Demo System (RHPDS) OpenShift 4.1 cluster.

### LINKED ISSUE #1920: Specify LimitRange Minimum as Part of Task Definition
Related to #1045 

As documented in #1045, Tekton pipeline does not work when a LimitRange minimum is specified. This is due to how container resource requests are "zeroed out" in [`resolveResourceRequests(containers)`](https://github.com/tektoncd/pipeline/blob/020137ee2f93559fa9d560bdfdaf1b3e222e374a/pkg/pod/resource_request.go#L47). 

The proposal in #1045 is to have a way for a user to set a minimum for container resource requests. I am wondering if it makes sense to allow a user to define this as part of a Task definition. If the user defines this field, `resolveResourceRequests` will accept the LimitRange minimum values passed in. If no minimum is specified, the current behavior of zeroing out the resource requests would be applied. 

This would allow the user control to honor a LimitRange minimum set within a namespace. The values could be pulled from a TaskRun and passed and used instead of `allZeroQty()`.

Also open to alternative suggestions, but I think this provides the user an option. LimitRange minimums are probably somewhat of an edge case, but we should still honor the request.

/kind question
/kind feature

danielhelfand:
Closes #1045
Closes #1920

# Changes

This pull request adds the ability to specify a LimitRange name as part of how a TaskRun handles container resource requests. Currently, pipelines/tasks do not work when a LimitRange minimum is specified because all container requests that do not have the max values are set to 0. Since these requests are 0, they fail to meet the minimum specified and the TaskRun fails.

In order to specify a LimitRange name, a LimitRangeName property has been added to TaskRun/PipelineRun specs. The LimitRange name is then used to get the LimitRange, which is passed to `resolveResourceRequests`. LimitRanges were added to the 200-clusterrole.yaml to allow LimitRanges to be accessed. 

Open to suggestions on more robust testing and can also include docs/examples if helpful.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Allow pipelines/tasks to work when LimitRange minimum is set by specifying a LimitRangeName as part of TaskRun/PipelineRun specs
```

### DIFF
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -20,6 +20,7 @@ Creation of a `PipelineRun` will trigger the creation of
 - [Cancelling a PipelineRun](#cancelling-a-pipelinerun)
 - [Examples](https://github.com/tektoncd/pipeline/tree/master/examples/pipelineruns)
 - [Logs](logs.md)
+- [LimitRange Name](#limitrange-name)
 
 ## Syntax
 
@@ -36,7 +37,6 @@ following fields:
     your `PipelineRun` resource object.
     - [`pipelineRef` or `pipelineSpec`](#specifiying-a-pipeline) - Specifies the [`Pipeline`](pipelines.md) you want to run.
 - Optional:
-
   - [`resources`](#resources) - Specifies which
     [`PipelineResources`](resources.md) to use for this `PipelineRun`.
   - [`serviceAccountName`](#service-account) - Specifies a `ServiceAccount` resource
@@ -51,6 +51,10 @@ following fields:
     follow the instruction [here](taskruns.md#Configuring-default-timeout) to configure the
     default timeout, the same way as `TaskRun`.
   - [`podTemplate`](#pod-template) - Specifies a [pod template](./podtemplates.md) that will be used as the basis for the `Task` pod.
+  - [`limitRangeName`](#limitrange-name) - Specifies the name of a LimitRange that exists in the namespace of the `PipelineRun`. This LimitRange's minimum 
+    for container resource requests will be used as part of requesting the appropriate amount of CPU, memory, and ephemeral storage for containers that are 
+    part of the `TaskRuns` of a `PipelineRun`. This property only needs to be specified if the `PipelineRun` is happening in a namespace with a LimitRange 
+    minimum specified for container resource requests.
 
 [kubernetes-overview]:
   https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#required-fields
@@ -371,6 +375,37 @@ metadata:
 spec:
   # […]
   status: "PipelineRunCancelled"
+```
+
+## LimitRange Name
+
+In order to request the minimum amount of resources needed to support the containers 
+for `steps` that are part of a `TaskRun`, Tekton only requests the maximum values for CPU, 
+memory, and ephemeral storage from the `steps` that are part of a TaskRun. Only the max 
+resource request values are needed since `steps` only execute one at a time in `TaskRun` pod. 
+All requests that are not the max values are set to zero as a result. 
+
+When a [LimitRange](https://kubernetes.io/docs/concepts/policy/limit-range/) is present in a namespace 
+with a minimum set for container resource requests (i.e. CPU, memory, and ephemeral storage) where `PipelineRuns` 
+are attempting to run, the `limitRangeName` property must be specified to appropriately apply the LimitRange's 
+minimum values to containers that are part of a `TaskRun` associated with a `PipelineRun`. This property helps 
+prevent failures of `TaskRuns` not meeting the minimum requirements specified by a LimitRange for containers.
+
+In the example below, the LimitRange `limit-mem-cpu-per-container` will be applied to all `TaskRuns` associated 
+with the `PipelineRun`:
+
+```yaml
+apiVersion: tekton.dev/v1alpha1
+kind: PipelineRun
+metadata:
+  creationTimestamp: null
+  generateName: deploy-pipeline-run-
+  namespace: default
+spec:
+  pipelineRef:
+    name: deploy-pipeline-hello
+  limitRangeName: "limit-mem-cpu-per-container"
+status: {}
 ```
 
 ---
--- a/pkg/apis/pipeline/v1alpha1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1alpha1/pipelinerun_types.go
@@ -66,6 +66,12 @@ type PipelineRunSpec struct {
 	// with those declared in the pipeline.
 	// +optional
 	Workspaces []WorkspaceBinding `json:"workspaces,omitempty"`
+	// Used to specify name of LimitRange that exists in namespace
+	// where PipelineRun will run so that the LimitRange's minimum for
+	// container requests can be used by containers of TaskRuns associated
+	// with PipelineRun
+	// +optional
+	LimitRangeName string `json:"limitRangeName"`
 }
 
 // PipelineRunSpecStatus defines the pipelinerun spec status the user can provide
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -119,8 +119,18 @@ func MakePod(images pipeline.Images, taskRun *v1alpha1.TaskRun, taskSpec v1alpha
 	initContainers = append(initContainers, entrypointInit)
 	volumes = append(volumes, toolsVolume, downwardVolume)
 
+	// If present on TaskRunSpec, use LimitRangeName to get LimitRange
+	// so it can be used in resolveResourceRequests
+	var limitRange *corev1.LimitRange
+	if taskRun.Spec.LimitRangeName != "" {
+		limitRange, err = kubeclient.CoreV1().LimitRanges(taskRun.Namespace).Get(taskRun.Spec.LimitRangeName, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	// Zero out non-max resource requests.
-	stepContainers = resolveResourceRequests(stepContainers)
+	stepContainers = resolveResourceRequests(stepContainers, limitRange)
 
 	// Add implicit env vars.
 	// They're prepended to the list, so that if the user specified any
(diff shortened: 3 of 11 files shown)


pkg/pod/pod.go
```diff
@@ -119,8 +119,26 @@ func MakePod(images pipeline.Images, taskRun *v1alpha1.TaskRun, taskSpec v1alpha
 	initContainers = append(initContainers, entrypointInit)
 	volumes = append(volumes, toolsVolume, downwardVolume)
 
+	// Get LimitRange by name if present in ConfigMap tekton-limitrange-config.
+	// Otherwise, pass empty request to resolveResourceRequests.
+	limitRangeName := ""
+	limitRangeConfig, err := kubeclient.CoreV1().ConfigMaps(taskRun.Namespace).Get("tekton-limitrange-config", metav1.GetOptions{})
```
danielhelfand: Would be open to changing `tekton-limitrange-config` to something else in case this approach could be used to support other namespaced default values.

pkg/pod/resource_request.go
```diff
@@ -50,16 +52,42 @@ func resolveResourceRequests(containers []corev1.Container) []corev1.Container {
 		}
 	}
 
+	// Get limitrange minimum for container requests so they won't
+	// be zeroed out if minimum is specified in namespace
+	limitRangeItems := limitRange.Spec.Limits
```
sbwsg: this could potentially panic, since limitRange is a pointer.

pkg/pod/resource_request.go
```diff
@@ -50,16 +52,42 @@ func resolveResourceRequests(containers []corev1.Container) []corev1.Container {
...
+	min := allZeroQty()
+	if limitRange != emptyLimitRange {
```
sbwsg: won't this compare the pointer addresses rather than the struct values?  I suspect that these addresses will always be different since we're allocing `emptyLimitRange` and `limitRange` in different places.  If that's not the case I'm a bit confused on what's happening.

pkg/pod/pod.go
```diff
@@ -119,8 +119,26 @@ func MakePod(images pipeline.Images, taskRun *v1alpha1.TaskRun, taskSpec v1alpha
...
+	if err == nil {
+		limitRangeName = limitRangeConfig.Data["default-limitrange-name"]
+		if limitRangeName == "" {
+			return nil, fmt.Errorf("tekton-limitrange-config is present but default-limitrange-name key is missing")
+		}
+	}
+	limitRange := &corev1.LimitRange{}
```
sbwsg: suggest leaving this as nil by default.  we won't need to compare against the empty limit range in resolveResourceRequests if we do that since we can just assume that if the limitRange is non nil then we should use its values.

pkg/pod/resource_request.go
danielhelfand: Yes, you're correct and thanks for pointing this out. I misinterpreted the limitrange here. Let me push up a change to have the default case be that limitrange is nil.

pkg/pod/resource_request.go
```diff
@@ -50,16 +51,45 @@ func resolveResourceRequests(containers []corev1.Container) []corev1.Container {
...
 
+	// Get limitrange minimum for container requests so they won't
+	// be zeroed out if minimum is specified in namespace
+	var limitRangeItems []corev1.LimitRangeItem
+	if limitRange != nil {
+		limitRangeItems = limitRange.Spec.Limits
+	}
+	min := allZeroQty()
+	if limitRange != nil {
+		for _, limitRangeItem := range limitRangeItems {
```
sbwsg: nit: you can safely `range` over `nil` here, no need for the outer `if limiteRange  != nil` check.

pkg/pod/resource_request.go
```diff
@@ -50,16 +51,45 @@ func resolveResourceRequests(containers []corev1.Container) []corev1.Container {
...
+			if limitRangeItem.Type == "Container" {
```
sbwsg: minor: suggest `== corev1.LimitTypeContainer` rather than the literal string "Container" here.

danielhelfand:
> Looks good! Would definitely like to see the configmap name and purpose documented as well.

Thanks! I'll add the last changes you added and a doc.

docs/tekton-limitrange-config.md
```diff
@@ -0,0 +1,45 @@
...
+All requests that are not the max values are set to zero as a result. 
+
+## tekton-limitrange-config Example
+
+The [behavior above](#How-Tekton-Handles-Container-Requests) is how Tekton carries out 
+TaskRuns by default, but, in the event you are working in a namespace that has a LimitRange 
+minimum defined for container requests, a ConfigMap named `tekton-limitrange-config` can be 
+created that holds the name of the LimitRange in the namespace where your TaskRuns are to be ran. 
+
+An example `tekton-limitrange-config` is shown below:
```
sbwsg: nice :+1:

danielhelfand:
@sbwsg Sorry to do this, but I changed this up a bit. While I was hesitant to do this because I am not sure how commonly used this will be, I think it makes more sense to add the `LimitRangeName` to the pipelinerun/taskrun specs and avoid using a separate config. 

If it is preferred to not add this field to the spec, I can go back to the config implementation. Otherwise, I can squash these commits and go with this new approach.

pkg/apis/pipeline/v1alpha1/pipelinerun_types.go
```diff
@@ -66,6 +66,8 @@ type PipelineRunSpec struct {
 	// with those declared in the pipeline.
 	// +optional
 	Workspaces []WorkspaceBinding `json:"workspaces,omitempty"`
+	// +optional
```
sbwsg: could use a comment here describing the field's purpose.  `// LimitRangeName specifies the limit range to use for minimum resource values in task pods` or similar.

Same comment for taskrun_types in v1alpha1 and 2

danielhelfand: Thanks! Added along with some docs for pipelineruns/taskruns.

docs/pipelineruns.md
```diff
@@ -51,6 +51,10 @@ following fields:
     follow the instruction [here](taskruns.md#Configuring-default-timeout) to configure the
     default timeout, the same way as `TaskRun`.
   - [`podTemplate`](#pod-template) - Specifies a [pod template](./podtemplates.md) that will be used as the basis for the `Task` pod.
+  - [`limitRangeName`](#limitrange-name) - Specifies the name of a LimitRange that exists in the namespace of the `PipelineRun`. This LimitRange's minimum 
+    for container resource requests will be used as part of requesting the appropriate amount of CPU, memory, and ephemeral storage for containers that are 
+    part of the `TaskRuns` of a `PipelineRun`. This proerty only needs to be specified if the `PipelineRun` is happening in a namespace with a LimitRange
```
sbwsg: nit: `proerty` -> `property`

pkg/apis/pipeline/v1alpha1/pipelinerun_types.go
sbwsg: :+1: nice

PR #2003, opened 2020-02-05
golang format tools

mattmoor:
Produced via:
  `gofmt -s -w $(find -path './vendor' -prune -o -path './third_party' -prune -o -type f -name '*.go' -print)`
  `goimports -w $(find -name '*.go' | grep -v vendor | grep -v third_party)`
/assign ImJasonH
/cc ImJasonH

### DIFF
--- a/pkg/entrypoint/entrypointer.go
+++ b/pkg/entrypoint/entrypointer.go
@@ -18,11 +18,12 @@ package entrypoint
 
 import (
 	"fmt"
-	"go.uber.org/zap"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"time"
+
+	"go.uber.org/zap"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/entrypoint/entrypointer.go
```diff
@@ -18,12 +18,13 @@ package entrypoint
...
 import (
 	"fmt"
-	"go.uber.org/zap"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"time"
 
+	"go.uber.org/zap"
+
```
sbwsg: odd spacing choice.  do we separate like this anywhere else in the codebase?

ImJasonH: Yeah the preferred style is to keep stdlib imports separate.

sbwsg: I meant the separation between zap and our internal packages.  We don't do that anywhere else.

ImJasonH: Oh yeah. Some people do that, some don't. I don't really care except I make it a policy never to disagree with formatting tools. 🤖

sbwsg: it's over.  the robots have won.

PR #2004, opened 2020-02-05
Add LimitRange Examples for PipelineRun and TaskRun

danielhelfand:
Adding some examples to illustrate #1991 for specifying a LimitRangeName as part of pipelineruns/taskruns.

# Submitter Checklist

review them:

# Release Notes

```
Add examples for specifying a LimitRangeName as part of a pipelinerun/taskrun
```

### DIFF
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -408,6 +408,8 @@ spec:
 status: {}
 ```
 
+An example `PipelineRun` using `limitRangeName` is available [here](../examples/pipelineruns/no-ci/limitrange.yaml).
+
 ---
 
 Except as otherwise noted, the content of this page is licensed under the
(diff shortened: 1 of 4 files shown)


danielhelfand:
Unfortunately, the examples here create a LimitRange which causes all other examples to fail since they don't take into account `limitRangeName`. Moving the examples to `no-ci` and will make them more visible by including in documentation.

sbwsg:
oh cool, i didn't know `no-ci` existed!

Would be great to get these limitranges running on every push at some point though.  I'll create an issue on plumbing to address this - maybe we need to run each example in its own namespace?

/lgtm

danielhelfand:
> - maybe we need to run each example in its own namespace?
> 

Yeah, if these examples are excluded from the others in a different namespace, they would work. But would be great if we could create different namespaces for certain examples.

PR #2020, opened 2020-02-07
Don't Require User Input to Account for LimitRange Min

danielhelfand:
Opening this as a proposal to improve upon #1991. This pull request removes the requirement of having to use `limitRangeName` to specify a LimitRange minimum to account for with TaskRuns.

This approach may be easier for users who do not have permissions to query LimitRanges or do not know which LimitRange has container requests defined in a namespace. 

# Submitter Checklist

## Reviewer Notes

# Release Notes

```
Remove limitRangeName and have TaskRun account for LimitRange on behalf of user
```

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -119,18 +119,13 @@ func MakePod(images pipeline.Images, taskRun *v1alpha1.TaskRun, taskSpec v1alpha
 	initContainers = append(initContainers, entrypointInit)
 	volumes = append(volumes, toolsVolume, downwardVolume)
 
-	// If present on TaskRunSpec, use LimitRangeName to get LimitRange
-	// so it can be used in resolveResourceRequests
-	var limitRange *corev1.LimitRange
-	if taskRun.Spec.LimitRangeName != "" {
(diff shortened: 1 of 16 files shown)


sbwsg:
:+1: Nice improvement! This makes sense to me.

danielhelfand:
> 👍 Nice improvement! This makes sense to me.

Great! I am comfortable with this as stands and think it makes more sense to not include as part of taskrun/pipelinerun specs at least for initial release of this feature. 

If we find there is reason for allowing some level of user control on this, it can be added down the line. But this makes more sense to me as a starting point.

pkg/pod/pod.go
```diff
@@ -262,3 +257,33 @@ func MakeLabels(s *v1alpha1.TaskRun) map[string]string {
...
+		return nil, err
+	}
+
+	min := allZeroQty()
+	for _, lr := range limitRanges.Items {
+		lrItems := lr.Spec.Limits
+		for _, lrItem := range lrItems {
+			if lrItem.Type == corev1.LimitTypeContainer {
+				if lrItem.Min != nil {
+					for k, v := range lrItem.Min {
```
danielhelfand: Open to not having this approach, which is to search for the max LimitRange container min in the event there are conflicting container mins. 

An alternative approach would just be to break the loop after the first container min is found and let it potentially fail.

PR #2030, opened 2020-02-10
Rewrite the Tekton Pipelines overview for clarity and flow

spomorski:
# Changes

Rewrites the Tekton Pipelines overview for clarity and reorganizes the information for better flow.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,7 +1,7 @@
 # Tekton Pipelines
 
-Tekton Pipelines is a custom Kubernetes extension that installs and runs on your Kubernetes cluster.
-It defines a set of Kubernetes [Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) that act as building blocks from which you can assemble custom CI/CD pipelines. Once installed,
+Tekton Pipelines is a Kubernetes extension that installs and runs on your Kubernetes cluster.
+It defines a set of Kubernetes [Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) that act as building blocks from which you can assemble CI/CD pipelines. Once installed,
 Tekton Pipelines becomes available via the Kubernetes CLI (kubectl) and via API calls, just
 like pods and other resources. Tekton is open-source and part of the [CD Foundation](https://cd.foundation/),
 a [Linux Foundation](https://www.linuxfoundation.org/projects/) project.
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1,67 +1,63 @@
...
-
-## Related info
-
-If you are interested in contributing to the Tekton Pipeline project, see the
-[Tekton Pipeline contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md).
+Tekton Pipelines is a custom Kubernetes extension that installs and runs on your Kubernetes cluster.
+It defines a set of Kubernetes [Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) that act as building blocks from which you can assemble custom CI/CD pipelines. Once installed,
+Tekton Pipelines becomes available via the Kubernetes CLI (kubectl) and via API calls, just
+like pods and other resources. Tekton is open-source and part of the [CD Foundation](https://cd.foundation/),
+a [Linux Foundation](https://www.linuxfoundation.org/projects/) project.
```
sbwsg: We use the word "custom" several times in this paragraph but it doesn't add a ton of meaning except when it's the name of a specific Kubernetes type (Custom Resource).  Suggest dropping its usage.

docs/README.md
```diff
@@ -1,67 +1,63 @@
...
+
+## Tekton Pipelines entities
+
+Tekton Pipelines defines the following entities:
+
+<table>
```
sbwsg: Github supports formatting tables without HTML ([see their help docs here](https://help.github.com/en/github/writing-on-github/organizing-information-with-tables)) which is a little more readable than the raw HTML I think.

docs/README.md
spomorski: Agreed, I was being over-specific as a newbie to the space. Removed.

docs/README.md
spomorski: I'm aware of markdown tables, but in my experience I've found them cumbersome and always had better visual results with HTML when a table was more complex than 2-3 columns. If we prefer markdown tables in our docs as a rule, though, I'll convert it, of course.

sbwsg: I have no strong feeling either way, just wanted to see if you knew of them already.  If your prefer authoring with the html then that's cool by me! :+1:

spomorski: Cool, I'll stick with HTML then. :)

PR #2038, opened 2020-02-11
Add Timeout to PipelineTask

### LINKED ISSUE #1568: Need a way to specify a timeout for a pipeline task
# Actual Behavior

I configured a Tekton trigger event listener + trigger + pipeline for building some complex software that happens to exceeds the global task run timeout configuration.

With `tkn pr describe <pipeline run id>`, I see that several task runs failed with `TaskRunTimeout`.
Since these `TaskRuns` are generated by the tekton trigger event listener, I can't specify a timeout.

# Expected Behavior

Proposal to augment the `Pipeline` syntax such that a `PipelineTask` could specify a timeout.

For example, adapting the [pipeline doc example](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#usage):

```
kind: Pipeline
metadata:
  name: pipeline-with-parameters
spec:
  params:
    - name: context
      type: string
      description: Path to context
      default: /some/where/or/other
  tasks:
    - name: build-skaffold-web
      taskRef:
        name: build-push
      # Same syntax as TaskRunSpec.timeout: https://golang.org/pkg/time/#ParseDuration
      timeout: 5m
      params:
        - name: pathToDockerFile
          value: Dockerfile
        - name: pathToContext
          value: "$(params.context)"
```

The proposal is that running this pipeline would pass the `PipelineTask.timeout` as a `TaskRunSpec.timeout`. In the adapted example above, it would mean that the `TaskRun` for `build-skaffold-web` would timeout after 5 minutes regardless of the global default task run timeout configuration setting.

1. A tekton trigger event listener with a trigger for a pipeline that executes a task that exceeds the default task run timeout.

# Additional Info

### LINKED ISSUE #2037: Timeout Code Comment for v1alpha2 TaskRunSpec Says Default Timeout is 10 Minutes
The code comment for Timeout as part of the TaskRunSpec appears to state the default time of 1 hour incorrectly.

# Expected 

The default timeout is 1 hour as documented [here](https://github.com/tektoncd/pipeline/blob/master/docs/taskruns.md#configuring-default-timeout). 

# Actual 

The [comment](https://github.com/tektoncd/pipeline/blob/52babda84f585fbe4ad6a8dce7a1802d2de513dd/pkg/apis/pipeline/v1alpha2/taskrun_types.go#L46) states the timeout default is 10 minutes.

/kind documentation

danielhelfand:
Closes #1568 
Closes #2037 

# Changes

This pull request adds a Timeout property to PipelineTasks to allow TaskRuns associated with a PipelineRun to define Timeouts. 

```yaml
kind: Pipeline
metadata:
  name: pipeline
spec:
  tasks:
  - name: build
    taskRef:
      name: build
    timeout: 1m
  - name: deploy
    taskRef:
      name: deploy
    runAfter:
      - build
```

If the TaskRun times out, the TaskRun will fail and the PipelineRun will fail as well. 

This pull request also changes a code comment for TaskRun timeouts that incorrectly states the default for TaskRun timeouts is 10 minutes. It is actually 1 hour. 

# Submitter Checklist

## Reviewer Notes

# Release Notes

```
Add Timeout property for PipelineTasks
```


vdemeester:
@danielhelfand you need to run `./hack/update-codegen.sh` to update the generated file with this new field :angel: 

/area api
/kind api-cĥang

danielhelfand:
> @danielhelfand you need to run `./hack/update-codegen.sh` to update the generated file with this new field 👼

Thanks! The script seems to have an issue with a script from plumbing. Any thoughts on it would be appreciated, but I'll look into it too.

```
/github.com/tektoncd/pipeline/vendor/github.com/tektoncd/plumbing/scripts/library.sh: line 325: go-licenses: command not found
```

danielhelfand:
I see, `go-licenses` needs to be installed.

sbwsg:
Looks good to me - just want to check before I lgtm tho: Are the additions to third_party intentional?  What was the process that introduced those changes?  Was it a result of the update-codegen.sh script?

That directory is only used so we have a place to store license info for the libraries we use, is that right?

danielhelfand:
> Looks good to me - just want to check before I lgtm tho: Are the additions to third_party intentional? What was the process that introduced those changes? Was it a result of the update-codegen.sh script?

The updates occurred as part of running `update-codegen.sh`. The build-tests were failing until it was updated. 

> That directory is only used so we have a place to store license info for the libraries we use, is that right?

That was my thought, but I'm not completely sure. If someone else can confirm, would appreciate it.

sbwsg:
Awesome, cheers!

/lgtm

PR #2042, opened 2020-02-13
Adding pipeline results

othomann:
# Changes
Adding the ability to reuse task results into other tasks within the same pipeline

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Add pipeline results. This makes it possible to a user to use results from another task as parameters to a task within the same pipeline.
```


sbwsg:
Something that occurs to me reading through this: It looks like we allow the user to only use a single isolated variable substitution in a given parameter.  In other words they can't do something like:

```
params:
- name: ephemeral-env-name
  value: "ci-$(tasks.git-push.results.short-commit-sha)-ephemeral"
```

Where the expected value here would end up looking something like "ci-991f807-ephemeral" and in this example would be the name of an ephemeral namespace or something that is deployed to when a PR is being tested.

I need to check what variable substitution looks like with other fields. I have a hunch that this kind of thing is supported with others and we'd probably want the same kind of support here as well.  Will look into it and update this PR with a comment.

pkg/apis/pipeline/v1alpha2/param_types.go
```diff
@@ -124,3 +125,77 @@ func (arrayOrString *ArrayOrString) ApplyReplacements(stringReplacements map[str
...
+	return &ResultReference{
+		PipelineTask: pipelineTask,
+		Result:       result,
+	}, nil
+}
+
+// HasResultReference is similar to NewResultReference, but it is used
+// when the error is not relevant. i.e. we simply want to check if param contains
+// a result reference
+func HasResultReference(param Param) (*ResultReference, bool) {
```
sbwsg: While I understand the impulse to add sugar for common use cases I'd argue that this doesn't add a ton of value over simply doing something like this directly in the 2 places which use `HasResultReference`:

```go
if resultRef, _ := NewResultReference(param); resultRef != nil {
  // The param is a result reference
}
```

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -428,7 +428,17 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1alpha1.PipelineRun) er
 		c.Logger.Errorf("Error getting potential next tasks for valid pipelinerun %s: %v", pr.Name, err)
 	}
 
-	rprts := pipelineState.GetNextTasks(candidateTasks)
+	nextRprts := pipelineState.GetNextTasks(candidateTasks)
+	if err := resources.ResolveParams(pipelineState, nextRprts); err != nil {
```
sbwsg: We have some existing code to perform variable substitutions in `pkg/reconciler/pipelinerun/resources/apply.go` and I am wondering if we can use some of that to reduce the amount of new machinery we're introducing with `ResolveParams`.

The way that `ApplyParameters` works, for example, is to build up a table of string keys of the format "params.foo" to the value which that param is set to.  So you might end up with a map looking something like this:
```go
map[string]string {
  "params.branch-name": "master",
  "params.repo-url": "https://github.com/tektoncd/pipeline",
}
```

and then ApplyParameters will use the `ApplyReplacements` func to walk through all the tasks in the pipelinespec and perform substitutions on all those task's params.

Would it be possible to reuse some of that code to write an additional func like `ApplyResults` that performs similar work?  It looks like this same approach could map well to the problem we're solving here and would reduce the amount of new code required.  It may also allow us to support the use case I mentioned in another comment of embedding a result's value inside a larger string such as "ci-$(tasks.git-push.results.short-sha)-ephemeral-env".

eddie4941:
> params:
> -  name: ephemeral-env-name
> value: "ci-$(tasks.git-push.results.short-commit-sha)-ephemeral"
> Where the expected value here would end up looking something like "ci-991f807-ephemeral" and > in this example would be the name of an ephemeral namespace or something that is deployed to > when a PR is being tested.

Oh yeah. Didn't think about that. I don't think it would be too hard to implement that behavior. though theres something to think about that is the desired behavior.

Since results are taken verbatim form the contents of the result file that the task lays on disk, we might run into a bunch of issues with newlines when performing the substitutions in embedded cases like the one you mentioned.

In that example, if the task laid down the SHA like so (newline at the end):
```
f3af3eddd5047d2e7eb795affac54cb6952d22b9\n
```
```
"ci-$(tasks.git-push.results.short-commit-sha)-ephemeral"
```
would turn into
```
ci-f3af3eddd5047d2e7eb795affac54cb6952d22b9\n-ephemeral
```

We could strip newline to prevent this, but that might be a bad idea as we likely don't want to alter the users output.

skaegi:
@sbwsg I'm trying to rework all the variable substitution right now to use the jsonpath.Expand func from PR #1951 everywhere so that kind of logic isn't reimplemented a zillion times. I just hadn't realized it was such a huge job. @eddie4941 we're so close to the beta that if it's easy to implement in the current way it would be good even if we re-implement a little bit later.

sbwsg:
> I need to check what variable substitution looks like with other fields. I have a hunch that this kind of thing is supported with others

OK, I've run a test this morning to see what happens if you wrap a pipeline's param into a larger string being passed into a Task's param and it looks like it operates as mentioned.  The pipeline's param is interpolated into the rest of the string contents. [1]

> we might run into a bunch of issues with newlines when performing the substitutions in embedded cases like the one you mentioned.

Great point.  I had not thought about newlines being included here.  My feeling is we should clearly document the behaviour initially and start considering changes such as trimming the whitespace only if users really struggle with it. My guess right now is that so long as we are passing the result values around verbatim that users will quickly learn to remove extra whitespace from the result values they emit from their tasks.  This also sounds like a good thing to watch out for when we start updating tasks in the Catalog to emit their results.

[1] Here's the quick example pipeline I wrote up to check the params behaviour:

```yaml
kind: Task
metadata:
  name: print-message
spec:
  inputs:
    params:
    - name: message
      type: string
      description: "The message to print"
      default: "hello world"
  steps:
  - name: print
    image: ubuntu
    script: |
      #!/usr/bin/env bash
      echo "$(inputs.params.message)"
---
kind: PipelineRun
metadata:
  name: add-octopus-arms
spec:
  params:
  - name: message
    value: "o.O"
  pipelineSpec:
    params:
    - name: message
      type: string
      description: "The figure to wrap in octopus arms"
    tasks:
    - name: add-octopus-arms
      taskRef:
        name: print-message
      params:
      - name: message
        value: "~~$(params.message)~~"
```

sbwsg:
> This also sounds like a good thing to watch out for when we start updating tasks in the Catalog to emit their results.

Example from a catalog entry of trimming newline when storing to a result here: https://github.com/tektoncd/catalog/pull/182/files#diff-1239eb0a6238e01666691e444b6367b7R55

It would be cool if this were handled for the user but it's not a huuuuge headache as it stands.  Just `echo -n $FOO > $(results.foo.path)`

akihikokuroda:
HI!  I'm interested in how I can use this.  Are there any design doc for this?

bobcatfish:
Just a quick drive by comment: I wanted to try this out but couldn't find any docs or examples - definitely want to have those before merging! Makes it much easier to review too cuz it provides context for how this will actually look in practice :racing_car:

jbarrick-mesosphere:
This is really great! I spent some time today wrapping my head around it and trying it and it works perfectly - however, I don't think there's a way to use the `$(tasks.<task>.results.<result>)` variables in a Pipeline Resource, is there?

othomann:
@bobcatfish Here is an example I used to give it a try.
```
kind: Pipeline
metadata:
  name: add-octopus-arms
spec:
  params:
    - name: first_message
      type: string
      default: "o.O"
    - name: second_message
      type: string
      default: "o.O"
  tasks:
    - name: first-message
      taskRef:
        name: print-message
      params:
        - name: message
          value: "$(params.first_message)"
    - name: all-messages
      taskRef:
        name: print-message
      params:
        - name: message
          value: "$(tasks.first-message.results.message)$(params.second_message)"
    - name: octopus-message
      taskRef:
        name: print-message
      params:
        - name: message
          value: "~~$(tasks.all-messages.results.message)~~"
```
```
kind: Task
metadata:
  name: print-message
spec:
  inputs:
    params:
    - name: message
      type: string
      description: "The message to print"
      default: "hello world"
  results:
    - name: message
      description: The printed message
  steps:
  - name: print
    image: ubuntu
    script: |
      #!/usr/bin/env bash
      echo -n "$(inputs.params.message)" | tee $(results.message.path)
```

othomann:
Another one:
```
kind: Pipeline
metadata:
  name: sum-and-multiply-pipeline-3
spec:
  params:
    - name: a
      type: string
      default: "1"
    - name: b
      type: string
      default: "1"
  tasks:
    - name: sum-inputs
      taskRef:
        name: sum
      params:
        - name: a
          value: "$(params.a)"
        - name: b
          value: "$(params.b)"
    - name: multiply-by-twenty
      taskRef:
        name: sum
      params:
        - name: a
          value: "$(tasks.sum-inputs.results.sum)$(tasks.sum-inputs.results.sum)"
        - name: b
          value: "$(tasks.sum-inputs.results.sum)$(tasks.sum-inputs.results.sum)"
```
```
kind: Task
metadata:
  name: sum
  annotations:
    description: |
      A simple task that sums the two provided integers
spec:
  inputs:
    params:
      - name: a
        type: string
        default: "1"
        description: The first integer
      - name: b
        type: string
        default: "1"
        description: The second integer        
  results:
    - name: sum
      description: The sum of the two provided integers
  steps:
    - name: sum
      image: bash:latest
      script: |
        #!/usr/bin/env bash
        echo -n $(( "$(inputs.params.a)" + "$(inputs.params.b)" )) | tee $(results.sum.path)
```

This will be added to the documentation once the code itself is reviewed.

othomann:
> This is really great! I spent some time today wrapping my head around it and trying it and it works perfectly - however, I don't think there's a way to use the `$(tasks.<task>.results.<result>)` variables in a Pipeline Resource, is there?

Do you have an example of what you have in mind ? Thanks.

jbarrick-mesosphere:
@othomann I would like to be able to do something like this: https://gist.githubusercontent.com/jbarrick-mesosphere/76bb4a083f724e5160c55fb86006b1b8/raw/189497ede646472d612560b4993e1414d9ac3a8c/gistfile1.txt

But after making the example, I'm not sure if there's an API that would make sense for it.

eddie4941:
@jbarrick-mesosphere 

The example you posted, wouldn't be possible with the implementation here. These results are only substituted in `pipeline.spec.tasks[].params[]`. However, what you are after can be achieved by using task results and workspaces. i.e.  

```---
kind: PipelineRun
metadata:
  name: run
spec:
  pipelineRef:
    name: pipeline
  params: 
    ...
  worksapces:
    - name: git-repo
      persistentVolumeClaim:
        claimName: pvc-to-hold-git-repo

---
kind: Pipeline
metadata:
  name: pipeline
spec:
  params:
    name: git-url
  workspaces:
    - name: git-repo
  tasks:
    - name: version
      taskRef:
        name: version
      params:
        - name: url
          value: $(params.url)
      workspaces:
        - name: output
          worksapce: git-repo      
    - name: build-image
      taskRef:
        name: build-image
      params:
        - name: image-url
          value: example.com/mydockerimage
        - name: image-tag
          value: $(tasks.version.results.version)
      workspaces:
        - name: repo
          workspace: git-repo
---
kind: Task
metadata:
  name: version
spec:
  workspaces:
    - name: output
      description: The git repo will be cloned onto the volume backing this workspace
  inputs:
    params:
      - name: url
        description: gir url to clone
  results:
    - name: version
      description: The version of the git repo that was cloned
  steps:
    - name: version
      image: alpine/git
      workingDir: /workspace/git/
      script: |
        #!/usr/bin/env bash
        git clone $(inputs.params.url) $(workspaces.output.path) # Clone repo into worksapce
        ...
        git describe --tags | tee $(results.version.path)
---
kind: Task
metadata:
  name: build-image
spec:
  imputs:
    params:
      - name: image-tag
      - name: image-repo
  workspaces:
    - name: repo
      description: The volume backing this workspace will contain the git repo
  steps:
    - name: build-image
      image: gcr.io/kaniko-project/executor:latest
      args:
      - --context=$(workspaces.repo.path)
      - --destination=$(inputs.params.image-repo):$(inputs.params.image-tag)
      - --oci-layout-path=/workspace/output/dockerimage/
```

This provides even more flexibility then using resources, since you can choose how the workspaces are backed.

othomann:
I rebased on top of master.

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
```diff
@@ -0,0 +1,153 @@
...
+	return deduped
+}
+
+// resolveResultRefs converts all params of the resolved pipeline run task
+func (resolver *resultRefResolver) resolveResultRefs(target *ResolvedPipelineRunTask) (ResolvedResultRefs, error) {
+	var resolvedParams ResolvedResultRefs
+	for _, param := range target.PipelineTask.Params {
+		resolvedResultRefs, err := resolver.ResolveResultRefs(param)
+		if err != nil {
+			return nil, err
```
sbwsg: Suggest wrapping this err to indicate the pipeline task and param that caused the error.  Something like:

```go
return nil, fmt.Errorf("unable to find result referenced by param %q in pipeline task %q: %w", param.Name, target.PipelineTask.Name, err)
```

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+type resultRefResolver struct {
+	resolvedTaskByName map[string]*ResolvedPipelineRunTask
+}
+
+func newResultRefResolver(pipelineRunState PipelineRunState) *resultRefResolver {
+	resolvePipelineRunTaskByName := make(map[string]*ResolvedPipelineRunTask)
+	for _, resolvedPipelineRunTask := range pipelineRunState {
+		resolvePipelineRunTaskByName[resolvedPipelineRunTask.PipelineTask.Name] = resolvedPipelineRunTask
```
sbwsg: PipelineRunState has a `toMap()` func that performs this work.  I suggest making that method public and then reusing it throughout this file.  I also think we could then remove the `resultRefResolver` type completely and simply call `toMap` in the places where we need to look up by pipeline task name.

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
othomann: Done.

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
```diff
@@ -0,0 +1,153 @@
...
+			return nil, err
+		}
+		allResolvedResultRefs = append(allResolvedResultRefs, resolvedResultRefs...)
+	}
+	return removeDup(allResolvedResultRefs), nil
+}
+
+// resolveResultRef resolves any ResultReference that are found in param
+// Returns nil if none are found
+func (resolver *resultRefResolver) ResolveResultRefs(param v1alpha2.Param) (ResolvedResultRefs, error) {
```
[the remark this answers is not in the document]
othomann: sounds good. I might remove it completely based on your last comment.

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -53,6 +54,20 @@ func ApplyParameters(p *v1alpha1.PipelineSpec, pr *v1alpha1.PipelineRun) *v1alph
...
+func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResultRefs) {
+	stringReplacements := map[string]string{}
+
+	for _, resolvedResultRef := range resolvedResultRefs {
+		replaceTarget := v1alpha2.GetVariableSubExpression(resolvedResultRef.ResultReference)
+		stringReplacements[replaceTarget] = resolvedResultRef.Value.StringVal
+	}
+
+	for _, resolvedPipelineRunTask := range targets {
+		resolvedPipelineRunTask.PipelineTask = applyTaskResult(resolvedPipelineRunTask.PipelineTask, stringReplacements)
```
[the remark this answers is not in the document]
othomann: We wanted to keep the same logic with the apply funcs. If you want that to be inlined, this is doable.

pkg/apis/pipeline/v1alpha2/param_types.go
```diff
@@ -124,3 +126,96 @@ func (arrayOrString *ArrayOrString) ApplyReplacements(stringReplacements map[str
...
+		}
+	}
+	return false
+}
+
+func looksLikeResultRef(expression string) bool {
+	return strings.HasPrefix(expression, "task") && strings.Contains(expression, ".result")
+}
+
+// GetVariableSubExpression returns the variable substitution expression fopr a result reference
```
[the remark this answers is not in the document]
othomann: Fixed.

pkg/reconciler/pipelinerun/resources/apply.go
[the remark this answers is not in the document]
othomann: ok fixed.

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, resolvedPipelineRunTask := range pipelineRunState {
+		resolvePipelineRunTaskByName[resolvedPipelineRunTask.PipelineTask.Name] = resolvedPipelineRunTask
+	}
+	return &resultRefResolver{
+		resolvedTaskByName: resolvePipelineRunTaskByName,
+	}
+}
+
+// ResolveResultRefs resolves any ResultReference that are found in the target ResolvedPipelineRunTask
+func ResolveResultRefs(pipelineRunState PipelineRunState, targets PipelineRunState) (ResolvedResultRefs, error) {
```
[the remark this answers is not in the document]
othomann: Done.

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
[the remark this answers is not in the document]
othomann: Done.

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
othomann: Ok, I am removing resultRefResolver completely. Let's see if the resulting code looks better.

othomann:
Got weird failures in integration tests. Restarting them.

docs/developers/README.md
```diff
@@ -280,3 +280,127 @@ status:
...
 
 Instead of hardcoding the path to the result file, the user can also use a variable. So `/tekton/results/current-date-unix-timestamp` can be replaced with: `$(results.current-date-unix-timestamp.path)`. This is more flexible if the path to result files ever changes.
+
+## How task results can be used in pipeline's tasks
+
+Now that we have tasks that can return a result, the user can refer to a task result in a pipeline by using the syntax
+`$(tasks.<task name>.results.<result name)`. This will substitute the task result at the location of the variable.
+
+```yaml
+apiVersion: tekton.dev/v1alpha1
```
sbwsg: I suggest copying this into an example in `examples/pipelineruns` and linking there from here.  Maybe include just the most relevant parts as snippets to illustrate your text description (maybe just the "params" and "results" portions?)  This full pipeline is quite long I think.

Another nice benefit will be that the example file will work as an e2e test as well.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -125,7 +125,8 @@ func (t ResolvedPipelineRunTask) IsCancelled() bool {
 	return c.IsFalse() && c.Reason == v1alpha1.TaskRunSpecStatusCancelled
 }
 
-func (state PipelineRunState) toMap() map[string]*ResolvedPipelineRunTask {
+// ToMap returns a map that maps pipeline task name to the resolved pipeline run task
```
sbwsg: :+1: nice!

docs/developers/README.md
```diff
@@ -280,3 +280,127 @@ status:
 ```
 
 Instead of hardcoding the path to the result file, the user can also use a variable. So `/tekton/results/current-date-unix-timestamp` can be replaced with: `$(results.current-date-unix-timestamp.path)`. This is more flexible if the path to result files ever changes.
+
+## How task results can be used in pipeline's tasks
+
+Now that we have tasks that can return a result, the user can refer to a task result in a pipeline by using the syntax
+`$(tasks.<task name>.results.<result name)`. This will substitute the task result at the location of the variable.
```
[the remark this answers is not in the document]
othomann: Done

docs/developers/README.md
othomann: done.

pkg/apis/pipeline/v1alpha2/param_types.go
```diff
@@ -124,3 +126,96 @@ func (arrayOrString *ArrayOrString) ApplyReplacements(stringReplacements map[str
...
+	}
+	return false
+}
+
+func looksLikeResultRef(expression string) bool {
+	return strings.HasPrefix(expression, "task") && strings.Contains(expression, ".result")
+}
+
+// GetVariableSubExpression returns the variable substitution expression for a result reference
+func GetVariableSubExpression(resultRef ResultRef) string {
```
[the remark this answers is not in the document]
othomann: done.

othomann:
@sbwsg merged your reviews, squashed the commits and waiting for the tests to complete.

docs/developers/README.md
```diff
@@ -280,3 +280,52 @@ status:
...
+kind: Pipeline
+metadata:
+  name: sum-and-multiply-pipeline
+#...
+  tasks:
+    - name: sum-inputs
+#...
+    - name: multiply-inputs
+#...
+- name: sum-and-multiply
```
sbwsg: indent here to match line 298

docs/developers/README.md
```diff
@@ -280,3 +280,52 @@ status:
...
+```shell
+tkn pipelinerun logs sum-and-multiply-pipeline-run-rgd9j -f -n default
+[multiply-inputs : product] 150
+
+[sum-inputs : sum] 25
+
+[sum-and-multiply : sum] 30050
+```
+
+As you can see, you can define multiple tasks in the same pipeline and use the result of more than one task inside another task parameter. The substitution is only done inside `pipeline.spec.tasks[].params[]`.
```
sbwsg: Add a link to the example:
```
For a complete example demonstrating Task Results in a Pipeline see the [pipelinerun example](../examples/pipelineruns/task_results_example.yaml).
```

pkg/reconciler/pipelinerun/resources/apply_test.go
```diff
@@ -19,6 +19,8 @@ package resources
 import (
 	"testing"
 
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha2"
```
sbwsg: group import with other tektoncd imports below

docs/developers/README.md
othomann: Done.

docs/developers/README.md
othomann: Done

pkg/reconciler/pipelinerun/resources/apply_test.go
othomann: Done.

sbwsg:
@vdemeester @dibyom @bobcatfish @afrittoli does anyone have any objection to my lgtm'ing this in time for the 0.11 cut?

vdemeester:
@sbwsg it's gonna make me rebase the v1beta1 PR again, but no, no objection :wink: 
/lgtm
/meow

othomann:
I am doing a rebase on top of the v1beta1 change. This will require a `lgtm`tag again once I am done.

othomann:
unit tests are passing locally, but they fail when run as automated tests. Investigating.

othomann:
Since the move to v1beta1, the test `github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun.TestReconcileWithTaskResults` no longer works when run as part of the automated tests, but works perfectly fine locally. If someone has any clue how to debug that, please let me know.

othomann:
ok problem came from the validation of names. Fix pushed.

sbwsg:
re-adding the earlier lgtm

/lgtm

:crossed_fingers:

sbwsg:
:tada:

Issue #2013, opened 2020-02-06
Phase out overwriting of image's HOME variable. Remove /tekton/home.


sbwsg:
# Expected Behavior

Images with an expectation that HOME is set to something specific should work as intended.

# Actual Behavior

Those images do not work as expected - Pipelines overwrites the HOME variable of an image with `/tekton/home`.

See https://github.com/tektoncd/pipeline/issues/1836 for the original report on this issue.

So the plan is as follows:

- [x] ASAP add an opt-out flag to prevent $HOME being set to /tekton/home. This should be configurable through a ConfigMap.
- [x] Perform a release, announce deprecation.
- [ ] Wait 9 months for beta deprecation policy to expire (~4 Dec 2020)
- [ ] Update the code to make the flag opt-in instead of opt-out.
- [ ] Perform a release, announcing the change in flag behaviour.
- [ ] Wait 9 months for beta deprecation policy to expire.
- [ ] Remove feature flag completely.
- [ ] Perform a release, announcing the complete removal.

sbwsg:
Adding a configmap field to allow the HOME overwrite to be disabled: https://github.com/tektoncd/pipeline/compare/master...sbwsg:opt-out-home-overwrite?expand=1

chanseokoh:
Thanks for doing this! This has been very annoying for Java people on Knative and GCB, because almost all the time, most Java applications get the genuine, actual user home through a [platform-independent way](https://stackoverflow.com/questions/585534/what-is-the-best-way-to-find-the-users-home-directory-in-java/586345) that does not read `$HOME`. For example, Tekton pipelines run as `root`, and because the OS declares that the home of the user `root` is `/root`, the Maven build system creates a local artifact repository under `/root/.m2` on Tekton (and GCB). BTW, the local Maven repository is crucial to cache in the Java world. (Gradle has the same problem for, e.g., `~/.gradle/caches`.)

After all, changing the `$HOME` environment variable for a certain set of processes or shell sessions doesn't alter the fact that the user home of `root` declared by OS is `/root`. Some tools and programs (e.g., `cd`, understandably) make use of `$HOME`, but most Java applications don't. (Aside, `cd` will go into `$HOME`, but `cd ~` will rightfully go into `/root`.)

[...]

Issue #1836, opened 2020-01-09
Problems with images that assume a particular workingDir and/or HOME environment variable


skaegi:
I've been integrating with a few pen testing tool images that are relying on a very specific workingDir being set. The problem I've been having is that Tekton overrides this setting and then I had to go back into the image to reverse engineer its original value. 

Could we provide a way to not override this setting and do we even need to do this anymore?

bobcatfish:
> The problem I've been having is that Tekton overrides this setting and then I had to go back into the image to reverse engineer its original value.

I'm surprised that Tekton overrides the workingDir - can you provide an example of this happening?

If that's the case I agree that we shouldn't be arbitrarily setting it (what is it getting set to?)

skaegi:
Sure any task will do -- just `echo $(pwd)` and it is always `/workspace`

vdemeester:
Indeed, we are setting it to `/workspace` if there isn't any. I wonder if we should do the same as the `entrypoint` magic where we read the image configuration before doing that to see if any is set.

Also, maybe it should be the other way around, by default we do not set it if empty, but we have an option somewhere that tells Tekton to override it (per task or pipeline, or globally ?)

skaegi:
I want to expand this issue as the HOME env (which gets set to '/tekton/home')variable is the other common problem we see e.g. where an image has explicitly set HOME. We're currently asking our users to explicitly set HOME and workingDir.

I <3 the way the entrypoint works as it is mostly transparent and wonder if other things could work the same way.

bobcatfish:
> Also, maybe it should be the other way around, by default we do not set it if empty

:+1: for not setting it if it's empty. I propose we warn in 0.10 and change it in the following release, whaddaya think @vdemeester @skaegi ?

> but we have an option somewhere that tells Tekton to override it (per task or pipeline, or globally ?)

At the moment i think you can do this with the stepTemplate, maybe that's good enough for now?

> I want to expand this issue as the HOME env (which gets set to /tekton/home)variable is the other common problem we see e.g. where an image has explicitly set HOME. We're currently asking our users to explicitly set HOME and workingDir.

This is a bit tricker imo b/c we're trying to create a "home" that is shared across all containers - ( + @ImJasonH ) - I'd be up for adding logic like we do with entrypoint, so that we have a hierarchy like:

- If the Step specifies HOME, use that
- If not but the stepTemplate does, use that
- If not but the Image does, use that
- If none of the above, use /tekton/home

?

[...]

PR #2044, opened 2020-02-13
Allow disabling implicit HOME overwrite

sbwsg:
# Changes

Tekton currently overwrites the $HOME variables in task containers. This behaviour is problematic when an image explicitly sets the HOME env var and expects it to be set to that value upon running.

This PR introduces a `feature-flags` ConfigMap with a single flag - `disable-home-env-overwrite` - that, when set to "true" will prevent Tekton from overriding the $HOME variable in Task containers.  This behaviour will be kept for one release which must include a deprecation warning that the behaviour will be changing soon. In the release after that the flag will be flipped so that the default behaviour becomes for Tekton to NOT override the $HOME variable.

Contributes to issues #2013 and #1836

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Tekton currently overwrites the $HOME variables in Step containers. This behaviour is problematic when an image explicitly sets the HOME env var and expects it to be set to that value upon running. You can now disable this behaviour in Tekton by using the feature-flags ConfigMap. See docs/install.md for further details on disabling this behaviour.
```

### DIFF
--- a/config/config-feature-flags.yaml
+++ b/config/config-feature-flags.yaml
@@ -1,0 +1,30 @@
+# Copyright 2019 The Tekton Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


sbwsg:
okey dokey, I think this PR is ready for review now.

skaegi:
/lgtm
Just re-validated...

PR #2076, opened 2020-02-20
Added conditional check tests

praveen4g0:
Added condition check Test on pipeline Run resources
# Changes

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```

### DIFF
--- a/test/conditional_check_test.go
+++ b/test/conditional_check_test.go
@@ -1,0 +1,96 @@
+// +build e2e
+
+/*
+Copyright 2019 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import (
+	"testing"
+
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+	tb "github.com/tektoncd/pipeline/test/builder"
+	knativetest "knative.dev/pkg/test"
+)
+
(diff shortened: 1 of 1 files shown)


test/conditional_check_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	knativetest.CleanupOnInterrupt(func() { tearDown(t, c, namespace) }, t.Logf)
+	defer tearDown(t, c, namespace)
+
+	condResource := tb.Condition("file-exists", namespace,
+		tb.ConditionSpec(
+			tb.ConditionParamSpec("path", v1alpha1.ParamTypeString),
+			tb.ConditionResource("workspace", v1alpha1.PipelineResourceTypeGit),
+			tb.ConditionSpecCheck("", "alpine"),
+			tb.ConditionSpecCheckScript("cat $(resources.workspace.path)/$(params.path)| grep \"Cloud Native\""),
```
sbwsg: I dunno about this bit - we're kind of assuming that README.md will always contain the words "Cloud Native".  Changing a README file in the future shouldn't result in a e2e test failure and will be super confusing if we ever do.

How about checking whether this very test file exists?  If you want to keep the "grep" then you could add a comment here like `// This comment is used by this test` and then grep for that with `cat $(resources.workspace.path)/$(params.path) | grep \"This comment is used by this test\"`

bobcatfish:
Hey @praveen4g0 ! Thanks for taking the time to add this :D

Is there an error you ran into that motivated you to add this?

I'm wondering b/c one thing that's odd about the end to end tests is that we actually have our tests split across 2 locations:
* The examples folder <-- anything that we can verify by writing and executing a pipeline/task directly we add here (basically the Run will fail if there is a bug
* The end to end go tests <-- we add coverage for things we can't easily express just with the examples

In the case of conditionals we have these 2 examples which test them:
* https://github.com/tektoncd/pipeline/blob/master/examples/pipelineruns/conditional-pipelinerun.yaml
* https://github.com/tektoncd/pipeline/blob/master/examples/pipelineruns/conditional-pipelinerun-with-optional-resources.yaml

So I'm wondering if maybe we don't need to add another end to end test after all?

bobcatfish:
Closing this for now but please reopen if you feel strongly that we should have another test here!

PR #2084, opened 2020-02-21
Remove linting from release pipelines 🧹

bobcatfish:
# Changes
This might be a controversial choice but I think our release Pipelines
should only include Tasks that we expect to give us useful feedback
about the release.

Linting should never fail on a release Pipeline since any issues should
be caught in the PR and
tektoncd/plumbing#241 is making linting
flakey.

I also made it so build and unit test can run in parallel since neither depends
on the other.

I tested this by manually applying the pipelines and manually triggering
the nightly cron:

```
 k --context dogfood create job --from cronjob/nightly-cron-trigger-pipeline-nightly-release nightly-cron-trigger-pipeline-nightly-release-manual-03232020
```

@afrittoli @vdemeester 

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/tekton/README.md
+++ b/tekton/README.md
@@ -186,7 +186,6 @@ Install Task from plumbing too:
 
 ```bash
 # Apply the Tasks we are using from the catalog
-kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/golang/lint.yaml
 kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/golang/build.yaml
 kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/master/golang/tests.yaml
 kubectl apply -f https://raw.githubusercontent.com/tektoncd/plumbing/master/tekton/resources/release/
--- a/tekton/release-pipeline-nightly.yaml
+++ b/tekton/release-pipeline-nightly.yaml
@@ -39,22 +39,7 @@ spec:
   - name: notification
     type: cloudEvent
   tasks:
-    - name: lint
-      taskRef:
-        name: golangci-lint
-      params:
-        - name: package
-          value: $(params.package)
-        - name: version
-          value: v1.24
-        - name: flags
-          value: "-v --timeout 10m"
-      resources:
-        inputs:
-          - name: source
-            resource: source-repo
     - name: unit-tests
-      runAfter: [lint]
       taskRef:
         name: golang-test
       params:
@@ -65,7 +50,6 @@ spec:
           - name: source
             resource: source-repo
     - name: build
-      runAfter: [lint]
       taskRef:
         name: golang-build
       params:
--- a/tekton/release-pipeline.yaml
+++ b/tekton/release-pipeline.yaml
@@ -57,23 +57,7 @@ spec:
             resource: bucket
           - name: source-to-release
             resource: source-repo
-    - name: lint
-      runAfter: [precheck]
-      taskRef:
-        name: golangci-lint
-      params:
-        - name: package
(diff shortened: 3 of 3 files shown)


bobcatfish:
p.s. related im just noticing the dogfood cluster has `pipeline-release-nightly-no-unit-triggers` and `pipeline-release-no-unit`. Anyone know what these are? Can we delete them?

vdemeester:
I am ok with removing linting, but not ok removing unit tests.

afrittoli:
I'm happy to remove the linting on release, or we could run the lint step, but do not fail the release on failure of lint.
About unit tests, I agree that if the tests are the ones being flaky we could release, however we have no way of knowing before hand, so I'm not sure we should skip unit tests.

bobcatfish:
@vdemeester @afrittoli could you explain about about what scenario could cause the unit tests to fail on a release that we'd be concerned about not catching?

in the interests of https://github.com/tektoncd/plumbing/issues/241 I can separate this into 2 PRs, one to remove linting and one to remove unit tests, but I don't fully understand the scenario we are trying to catch by running unit tests when doing a release

vdemeester:
> @vdemeester @afrittoli could you explain about about what scenario could cause the unit tests to fail on a release that we'd be concerned about not catching?

1. We shouldn't have *any* flaky unit tests. A flaky unit-test should be consider like a bug to be fixed as soon as possible (and before any release). Flaky e2e tests is different and I can see us doing releases (at least nightly) without running those (or better, ignore failures if known flaky).
2. A failing unit-test during a release might be the case of a unit test that was written in a weird way and depend on an environment or something that is not always true — or could be a legit bug that doesn't show in our CI but during release (for some reason)
3. I don't feel confident at all to do a release where the unit tests didn't run. For me, unit tests are as important as *production* code, so a failing unit tests is like a compilation failure (and a compilation failure means no release).

> in the interests of [tektoncd/plumbing#241](https://github.com/tektoncd/plumbing/issues/241) I can separate this into 2 PRs, one to remove linting and one to remove unit tests, but I don't fully understand the scenario we are trying to catch by running unit tests when doing a release

I would like having 2 separates PR to discuss this more indeed. As stated above, for me, unit tests are as important as *production* code, so a failing unit tests is like a compilation failure.

bobcatfish:
> I would like having 2 separates PR to discuss this more indeed. As stated above, for me, unit tests are as important as production code, so a failing unit tests is like a compilation failure.

kk I'll split this up! I'll try to ASAP update the deployed pipeline too b/c right now i think ive disabled both.

@vdemeester (1) and (2) that you described make sense to me - I'm wondering, what if instead of tying this to releases we had a nightly CI job that ran unit tests and maybe even system tests too? That way we could catch flakes without blocking releases.

I'm less sure about (3) but open to discussing.

bobcatfish:
The more I think about it, there's probably no reason for me to argue my position so strongly since we haven't even run into a case where unit tests block a release, I think it might just be a principle thing for me that isn't very important so I think I should drop it!

@vdemeester @afrittoli what do you think about adding a nightly CI job anyway, to try to catch flakes?

vdemeester:
> I'm wondering, what if instead of tying this to releases we had a nightly CI job that ran unit tests and maybe even system tests too?

Yes and no. Having a daily (or twice a day) test run (unit and e2e) is something we need for sure (but we need report for this :angel:). But it shouldn't replace the nightly release though (as the nightly is also there for ppl to be able to quickly try master at a given point of time)

bobcatfish:
PTAL @afrittoli @vdemeester @sbwsg

Issue #1530, opened 2019-11-06
Handling parameter interpolation in fields not designed for it


skaegi:
When authoring Tekton resources that use parameter interpolation we can get in trouble when either the value type is not a `string` or if there is additional validation like for #1514 . For now I'm just pointing out the problem but it would be good if we could somehow defer validation in Tasks and Pipelines until after parameter substitution occurred.

skaegi:
Two approaches...
1) parse selected fields of the Task and Pipeline `spec` identified as supporting interpolation into interface{} or similar until after fields have been interpolated or perhaps not until they are translated into a runtime object.
2) introduce a new field to Task and Pipeline called `patches` that lets you target a path and provide a value that will be added at that location. (semantics of JSONPatch "add"). For example...
```
spec:
  params:
    - name: my-boolean-param
       default: true
  patches:
    - path: somePlace.thatNeeds.ABoolean
      value: $(params.my-boolean-param)
  somePlace:
    thatNeeds:
      ABoolean: false # if the value is $(params.my-boolean-param) we will fail because until interpolated it is a string
```

skaegi:
I'm looking at -- https://github.com/kubernetes-sigs/kustomize/blob/master/examples/patchMultipleObjects.md -- to see if it might be an applicable approach. One difference is that the patch would be inlined (and parameterized) instead of in a separate file.

skaegi:
So... to construct an example...
```
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: echo-hello-world
spec:
  steps:
    - name: echo
      image: ubuntu
      command:
        - echo
      args:
        - "hello world"
```
Could be written as...
```
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: echo-hello-world
spec:
  patchesStrategicMerge:
    spec:
      steps:
        - name: echo
           args:
             - "hello strategic merge world"
  steps:
    - name: echo
      image: ubuntu
      command:
        - echo
```
or with params as...
```
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: echo-hello-world
spec:
  params:
    - name: helloworld
      default: "hello strategic merge world"
  patchesStrategicMerge:
    spec:
      steps:
        - name: echo
           args:
             - $(params.helloworld)
  steps:
    - name: echo
      image: ubuntu
      command:
        - echo
```
---
[...]

PR #2085, opened 2020-02-21
Adds Basic Support for Star Array Notation

### LINKED ISSUE #2041: Provide short term solution for array "star" substitution
We want to eventually just use JSONPath for variable expansion but in the short-term we should just add an extra star expansion for array types for our string-prefix based expansion. e.g. `params.myarray` will cause the create "both" a `params.myarray` and `params.myarray[*]` map key for `ApplyArrayReplacements` and the various validate methods.

This will let us update our tests to use the star-expansion in advance of jsonpath support.

skaegi:
# Changes
When referring to an array reference that we want to expand inside an existing array we should use star syntax.
For example: ["a", "b", "$(params.myArray[*])", "d"] to avoid confusion when in the future we want to add the array itself as an array element.

Basic support changes are in substitution.go
All other changes are for unit tests

Fixes #2041
# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -250,8 +250,8 @@ can be passed to the `Task` from a `TaskRun`.
 Input parameters in the form of `$(inputs.params.foo)` are replaced inside of
 the [`steps`](#steps) (see also [variable substitution](#variable-substitution)).
 
-The following `Task` declares an input parameter called 'flags', and uses it in
-the `steps.args` list.
+The following `Task` declares two input parameters named 'flags' (array) and 'someURL' (string), and uses them in
+the `steps.args` list. Array parameters like 'flags' can be expanded inside of an existing array by using star expansion syntax by adding `[*]` to the named parameter as we do below using `$(inputs.params.flags[*])`. 
 
 ```yaml
 apiVersion: tekton.dev/v1alpha1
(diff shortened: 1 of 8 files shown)


skaegi:
Whoops -- didn't see this til right now. Yep you bet.

docs/tasks.md
```diff
@@ -576,7 +576,7 @@ Param values from resources can also be accessed using [variable substitution](.
 
 ##### Variable Substitution with Parameters of Type `Array`
 
-Referenced parameters of type `array` will expand to insert the array elements in the reference string's spot.
+Referenced parameters of type `array` can be expanded using 'star-expansion' by adding `[*]` to the named parameter to insert the array elements in the reference string's spot.
```
dibyom: What happens if I don't use the * with an array param? Is that an error or will it simply not expand the values out?

skaegi: Yes with caveats. Right now the behaviour is identical as the star expansion is effectively an alias, but longer term when we add proper type support if you just reference the array directly then the array itself will be added. Whether or not that is an error is going to depend on the type we unmarshal into (see #1530 for where we're heading).

skaegi: As it turn out right now you can't assign an array parameter to an actual array field, so afaik in all cases how we are currently use arrays we should just use star-expansion. Talking to @afrittoli earlier, I'll also update the catalog tasks, but that is not desperately urgent for now other than to illustrate good practices.

Issue #1684, opened 2019-12-04
Design: Failure Strategy for TaskRuns in a PipelineRun


dibyom:
The goal is to come up with a design to handle failing task runs in a pipelinerun. Today, we simply fail the entire pipelinerun if a single taskrun fails.

# Current Status

Summary in this comment: https://github.com/tektoncd/pipeline/issues/1684#issuecomment-611016087

# Ideas
Here are a couple of ideas from @sbwsg and me:
1. Introduce an `errorStrategy` field in `PipelineTasks` similar to the idea in #1573 
2. The `errorStrategy` could be under the `runAfter` field.
3. To start off, we could have two error strategies : `FailPipeline` which is the default for today, and `ContinuePipeline` which will continue running the whole pipeline
4. Later on, we could add branch based error strategies e.g. fail one one branch of the graph but continue running the remaining pipelines

# Additional Info
@sbwsg has some strawperson YAMLs:
[RunNextTasks for an integration test cleanup scenario](https://gist.github.com/sbwsg/83b6999cd30c079fe0a77e229b3efbd3)
[FailPipeline(default) for a unit test failing before a deploy task](https://gist.github.com/sbwsg/731917ef3bb3da28849f256dbb8208b6)

# Use Cases
* Unit test fails but integration still run
* Rollbacks for CD  e.g. Canaries - rollback if canary analysis fails
* Cleanup task if integration test fails
* Always run a step/task at the end e.g. to  Report results
* Run on conditional failures #1023 

# Related Issues
The Epic #1376 has all the related issues

bobcatfish:
Thanks for getting this started!!!

```yaml
    - name: integration
      taskRef:
        name: run-integration-tests
      runAfter: uts
      errorStrategy: RunNextTasks # allow cleanup to occur
    - name: cleanup
      taskRef:
        name: cleanup-integration-test-junk
      runAfter: integration
```

A concern about this is for these two use cases:
* Always run a step/task at the end e.g. to Report results
* The example above: cleanup after a integration test

It seems like it doesn't work if there is one more Task in the Pipeline, e.g. (totally contrived?) but something like:

```yaml
    - name: integration
      runAfter: uts
      errorStrategy: RunNextTasks # allow cleanup to occur
    - name: integration2 # pretend there was another set of tests?
      runAfter: uts
      errorStrategy: RunNextTasks # allow cleanup to occur
    - name: cleanup
      taskRef:
        name: cleanup-integration-test-junk
[...]

Issue #1023, opened 2019-06-26
Conditionals:  Configure what happens when a task is skipped due to condition failure


dibyom:
# Expected Behavior
Users can decide to either run or skip tasks that are dependent on the task that was skipped due to a condition check failure.

# Actual Behavior
Tasks that are dependent on a skipped task are always skipped

# Steps to Reproduce the Problem

1. Create a pipeline like similar to this:
```
apiVersion: tekton.dev/v1alpha1
kind: Pipeline
metadata:
  name: two-tasks-one-condition
spec:
  tasks:
  - name: conditional-task
    taskRef:
      name: task1
    conditions:
      - conditionRef: always-fail
  - name: run-if-1-or-1-skipped
    taskRef:
      name: task2
    runAfter:
      - conditional-task
```
2. Run the pipeline through a PipelineRun
3. task2 is not executed

# Additional Info
1. Might be worth enumerating the use cases for this feature before implementing this.
2. Based on discussion from #27: https://docs.google.com/a/google.com/document/d/1iTb0rDDlgQcPbiYIGJ-atkaML8aCyPCyhfucCPJmNLs/edit?disco=AAAADD1TnxU

dibyom:
ref: https://github.com/tektoncd/pipeline/issues/1137

bobcatfish:
One thought after looking at #1137 : 

- If a Task is getting a resource `from` a Task that has a failing condition, I'm not sure it makes any sense to run that dependent Task: e.g. if a Task is building an image as an output, and another task wants to get the built image (digest) `from` that Task, it wouldn't make sense to run it anyway
- `runAfter` is fuzzier - if it's just a matter of ordering, it feels like the user could potentially choose whether the tasks "after" should be skipped or not

dibyom:
From @afrittoli in #1264:

> In the example pipeline I used runAfter but I believe the behaviour would be the same if the dependency was introduced using from on a resource.

afrittoli:
> If a Task is getting a resource `from` a Task that has a failing condition, I'm not sure it makes any sense to run that dependent Task: e.g. if a Task is building an image as an output, and another task wants to get the built image (digest) `from` that Task, it wouldn't make sense to run it anyway

I'm not sure we have a good solution for this case specifically. I've been trying to build a pipeline that builds a docker image only if needed - which I encoded in a condition. The build task has the condition attached and the next Task deploys a cloud function that uses that image plus some code on top.
[...]

PR #2094, opened 2020-02-24
WIP - failure strategies using runOn 🎉

pritidesai:
# Changes

This is the first subset of changes implementing failure strategies.

This PR implements `runOn` with two states `success` and `failure`:

```
task4:
    runOn:
        task: task1
        states: ["success"]
        task: task2
        states: ["failure"]
        task: task3
        states: ["success", "failure"]
```

Execute `task4` if:
- task1 was successful AND
- task2 was failure AND
- task3 was either successful OR failure

Pipeline Status with these changes: **Succeeded (Completed)**
**Pipeline Status: Succeeded and Tasks Completed: 3, Skipped: 0, Failed: 1**

These changes should work for a pipeline with any number of branches i.e.

Pipeline with single branch:

```
 Pipeline:
         task1
           | failure
         task2 (runOn task1 failure)
           | success
         task3 (runOn task2 success)
```

Pipeline with multiple branches:

```
 Pipeline:
             task1
    success /     \ failure
         task2   task3
```

synonym to:

```
if task1:
  task2
else
  task3
```

Pipeline Status with these changes: **Succeeded (Completed)**
**Pipeline Status: Succeeded and Tasks Completed: 2, Skipped: 1, Failed: 0**

By default (without any runOn), Pipeline execution fails on very first task failure.

But `runOn` confirms the state of the task specified with the actual state.

`runOn` Success is defined by TaskRunStatus with `Condition.Type` set to `Succeeded` and `Condition.Status` set to `True`.
`runOn` Failure is defined by TaskRunStatus with `Condition.Type` set to `Succeeded`, `Condition.Status` set to `False` and Condition.Reason set to `Failed`.

With `runOn`, pipelineRun attempts to execute all the specified tasks. When a parent task does not match the state specified by its dependent task's runOn section, dependent task is marked as conflicted and continues executing rest of the pipeline.  To clarify, in the above example, if task1 succeeds, task2 is executed since it depends on task1's success state and task3 is skipped and marked with `SkippedAsStateConflicted` but execution continues if there are more tasks and pipeline ends Successfully with `Completed`.

**Note: There are still two more states needs to be added as part of runOn, `skip` and `always` as per the design document.**

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Adds the ability to be able to specify Tasks to run when other Tasks have failed
```


pkg/apis/pipeline/v1alpha1/pipeline_types.go
```diff
@@ -138,6 +139,10 @@ type PipelineTask struct {
 	// Refer Go's ParseDuration documentation for expected format: https://golang.org/pkg/time/#ParseDuration
 	// +optional
 	Timeout *metav1.Duration `json:"timeout,omitempty"`
+
+	// RunOn declares a list of dependent tasks and their states that they have to be in to run this task.
+	// +optional
+	RunOn []PipelineTaskRunOn `json:"runOn,omitempty"`
```
pritidesai: Adding a new field `runOn` as part of `PipelineTask`

pkg/apis/pipeline/v1alpha2/pipeline_types.go
```diff
@@ -227,3 +232,29 @@ type PipelineList struct {
...
+// 	- task: task1
+//    states: ["success", "failure"]
+//  - task: task2
+//    states: ["success"]
+type PipelineTaskRunOn struct {
+	// Task is the name of the PipelineTask as specified in a pipeline
+	Task string `json:"task"`
+	// States is a list of states of a PipelineTask which is specified in the Task above
+	States []PipelineTaskState `json:"states"`
+}
```
pritidesai: `runOn` is a map with key as a task name and value as a list of states

pkg/apis/pipeline/v1alpha2/pipeline_types.go
```diff
@@ -227,3 +232,29 @@ type PipelineList struct {
...
+
+// PipelineTaskState represents the state of a parent PipelineTask that it has to be in for it to run
+type PipelineTaskState = string
+
+const (
+	// PipelineTaskStateSuccess indicates that a PipelineTask has finished execution successfully
+	PipelineTaskStateSuccess PipelineTaskState = "success"
+	// PipelineTaskStateFailure indicates that a PipelineTask has finished execution and has failed
+	PipelineTaskStateFailure PipelineTaskState = "failure"
+)
```
pritidesai: Adding two possible states to begin with: (1) success (2) failure, remaining two states (skip and always) will go here

pkg/apis/pipeline/v1alpha2/pipeline_validation.go
```diff
@@ -194,6 +194,10 @@ func (ps *PipelineSpec) Validate(ctx context.Context) *apis.FieldError {
 		return err
 	}
 
+	// Validate PipelineTask runOn and runAfter
+	if err := validatePipelineTaskRunOn(ps.Tasks); err != nil {
+		return err
+	}
```
pritidesai: validating `runOn` for not having any state other than whats allowed and making sure doesnt have duplicates

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -422,8 +422,12 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1alpha1.PipelineRun) er
...
 		return err
 	}
-
-	candidateTasks, err := dag.GetSchedulable(d, pipelineState.SuccessfulPipelineTaskNames()...)
+	// Get the next tasks to be executed which doesnt't have any TaskRuns created yet
+	// ExecutedPipelineTaskNames gets the list of finished tasks (either successful or failed or skipped)
+	// GetSchedulable parses the whole graph and get a list of candidates based on already executed tasks
+	// e.g. for this Pipeline task1 -> task2 (runAfter task1) -> task3 (runAfter task2)
+	// GetSchedulable returns task1 followed by task2 (after task1 finishes) followed by task3 (after task2 finishes)
+	candidateTasks, err := dag.GetSchedulable(d, pipelineState.ExecutedPipelineTaskNames()...)
```
pritidesai: This was little tricky to change, `SuccessfulPipelineTaskNames` returned list of tasks which were finished executing successfully. Now, with `runOn` failed tasks should also be included in the returned list.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -150,17 +169,67 @@ func (state PipelineRunState) IsDone() (isDone bool) {
 // GetNextTasks will return the next ResolvedPipelineRunTasks to execute, which are the ones in the
 // list of candidateTasks which aren't yet indicated in state to be running.
 func (state PipelineRunState) GetNextTasks(candidateTasks map[string]struct{}) []*ResolvedPipelineRunTask {
```
pritidesai: this is where the most of the changes happened, for a task before adding it to the queue, check if `runAfter` is specified, check if `runOn` is specified, match the state in case `runOn` is specified. Mark the task as conflicted when `runOn` states does not match with the actual state, etc

pkg/apis/pipeline/v1alpha2/pipeline_types.go
dibyom: might be a good time to bikeshed on the `runOn` format i.e. map vs list. And I think @skaegi has some ideas around here as well :)

pkg/apis/pipeline/v1alpha2/pipeline_types.go
dibyom: I think other states we'd need are 1. skip for condition failure 2. skip due to a previously failing taskrun

pkg/reconciler/pipelinerun/pipelinerun.go
dibyom: Perhaps `pipelineState.CompletedPipelineTaskNames()` ?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -58,6 +60,8 @@ const (
 	// ReasonConditionCheckFailed indicates that the reason for the failure status is that the
 	// condition check associated to the pipeline task evaluated to false
 	ReasonConditionCheckFailed = "ConditionCheckFailed"
+
+	ReasonRunOnStateConflicted = "RunOnStateConflicted"
```
dibyom: a bit confused by this...could you add a comment explaining this state and why/when it can occur?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -150,17 +169,67 @@ func (state PipelineRunState) IsDone() (isDone bool) {
...
-		if _, ok := candidateTasks[t.PipelineTask.Name]; ok && t.TaskRun == nil {
-			tasks = append(tasks, t)
-		}
-		if _, ok := candidateTasks[t.PipelineTask.Name]; ok && t.TaskRun != nil {
-			status := t.TaskRun.Status.GetCondition(apis.ConditionSucceeded)
-			if status != nil && status.IsFalse() {
-				if !(t.TaskRun.IsCancelled() || status.Reason == v1alpha1.TaskRunSpecStatusCancelled || status.Reason == ReasonConditionCheckFailed) {
-					if len(t.TaskRun.Status.RetriesStatus) < t.PipelineTask.Retries {
-						tasks = append(tasks, t)
+		// for each candidate task, check if its TaskRun is initialized or not
```
dibyom: The comments are super helpful! Thanks! 
The nesting here is quite deep. Can we extract some bits onto helper functions? e.g  the `if t.TaskRun == nil` branch can be its own function probably..

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: thanks @dibyom for suggestion, replaced with `CompletedPipelineTaskNames`

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: Done :)

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: Done :)

pkg/apis/pipeline/v1alpha2/pipeline_types.go
pritidesai: humm, `skip` for sure, and may be `always`?

`skip` would run when parent tasks skipping due to condition failure.

Not sure what (2) signify? By default, `runOn` task is skipped if the states conflicts and that task is marked as conflicted (`RunOnStateConflicted`), any task dependent on task with `RunOnStateConflicted` is also skipped and marked `RunOnStateConflicted` instead of failing the whole pipeline, please check pipeline.md for more details.

pritidesai:
an attempt to rebase and rename v1alpha2 to v1beta1 🏃‍♀ 😅 🤞

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -227,3 +232,29 @@ type PipelineList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []Pipeline `json:"items"`
 }
+
+// PipelineTaskRunOn represents runOn of a pipelineTask which is a map of task and a list of states, e.g.
+// runOn:
+// 	- task: task1
+//    states: ["success", "failure"]
+//  - task: task2
+//    states: ["success"]
```
vdemeester: nit: incoherent spaces :stuck_out_tongue: :angel:

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -306,3 +310,53 @@ func validatePipelineNoArrayReferenced(name, value, prefix string, vars map[stri
...
+		// compare the list of tasks in runAfter and in runOn
+		// runOn should not have any extra task specified outside of runAfter
+		missing := list.DiffLeft(dependentTasks, pt.RunAfter)
+		if len(missing) > 0 {
+			return apis.ErrInvalidValue(
+				fmt.Sprintf("pipeline task %q has a list of tasks under runOn which does not exists in runAfter: %s", pt.Name, missing),
+				fmt.Sprintf("spec.tasks[%d].%q.runOn", ptIdx, pt.Name),
+			)
+		}
+	}
```
vdemeester: Should we support implicit `runAfter` if `runOn` is specified on a task not present in `runAfter` ? (to reduce "duplication")

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -240,6 +248,16 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun) error
 		return err
 	}
 
+	if tr.IsSkipped() {
+		tr.Status.SetCondition(&apis.Condition{
+			Type:    apis.ConditionSucceeded,
+			Status:  corev1.ConditionFalse,
+			Reason:  "RunOnStateConflicted",
```
vdemeester: Shouldn't it be `Reason:  resources.ReasonRunOnStateConflicted,` ?

pritidesai:
> Looking good, few comments:
> 
> * I am not sure of the reason `RunOnStateConflicted`, it doesn't really click to me 😅 (as it doesn't convey to me that this Task go skipped because of the `runOn`), but I am not sure what to use instead… but looking at the rest of the reason, I don't see something else to add instead 🤔
It does deserve an explanation 😉`RunOnStateConflicted` is very important state when a pipeline has a task `task2` which is dependent on failure of its parent `task1` but `task1` succeeds, in this case, TaskRun for `task2` is skipped with state `RunOnStateConflicted` and pipeline is concluded as `Completed` instead of  just `Succeeded` (Tasks Completed: 1, Skipped: 1, Failed: 0).

This new state is also very crucial for a pipeline with multiple branches, such as:

```
               task1
success /      \ failure
        task2    task3
 failure |          | success
         task4   task5
```

In this example, `task2` depending on success of `task1` whereas `task3` depending on failure of `task1`, and in turn these tasks (`task2` and `task3`) having dependencies. Here, Pipeline completes execution and marks `TaskRun` for `task3` and `task5` as skipped with `RunOnStateConflicted` state since `task1` was successful.

> * We should add an example with multiple `states` on `runOn` (docs and example folders) shouldn't we ?
yup, added one example with multiple states 😄

pkg/reconciler/taskrun/taskrun.go
pritidesai: `ReasonRunOnStateConflicted` is declared in pipeline resources, should we introduce pipeline resources to taskrun resources? Not sure, might result in some conflicts down the road, may be, may not 🤔 OR declare it once again under taskrun resources and use it from there 😕

pkg/apis/pipeline/v1beta1/pipeline_validation.go
pritidesai: We could but wouldn't work for a task having more than one tasks in `runAfter` 🤔

pkg/apis/pipeline/v1beta1/pipeline_types.go
pritidesai: Done :)

pritidesai:
thanks @vdemeester for reviewing 👍 please refer to my comments above ⬆️

pritidesai:
Just replaced `RunOnStateConflicted` with `SkippedAsStateConflicted` to notify that the task was skipped due to state conflicts, for example:

```
tkn pr describe pipelinerun-runon-success-failure
...
🗂  Taskruns

 NAME                                                 TASK NAME   STARTED          DURATION     STATUS
 ∙ pipelinerun-runon-success-failure-task2-wxn9n   task2       ---              ---          Failed(SkippedAsStateConflicted)
 ∙ pipelinerun-runon-success-failure-task3-dszrf   task3       10 minutes ago   12 seconds   Succeeded
 ∙ pipelinerun-runon-success-failure-task1-sn5gg   task1       10 minutes ago   30 seconds   Failed

```

pritidesai:
its ready for review, will add support for skip and always after this PR is merged 😲 🤞

pkg/apis/pipeline/v1beta1/pipeline_validation.go
afrittoli: Why not? We could even extend the runAfter syntax to support both an array of strings (like today) and an array of maps:
```
runAfer["taskA", "taskB"]
```
or
```
runAfter["taskA": ['success'], "taskB": ['success', 'failure']]
```

where the first syntax becomes a shorthand format for the ['success'] case.

pritidesai:
> Thank you for this! It looks great :)
> I'm taking a bit of time to review the code - so I'll come back with proper feedback soon.
> 
> One thing that I feel is missing, but we may add on top, is the ability to have a "finally" type of syntax. With this is kind of possible, but one has to list all possible outcomes of a task in the `runOn`, and still there would be no guarantee that it is executed as the pipeline might timeout in the meanwhile.

Yes, we need support for `finally` type of syntax which I foresee as a bool field in PipelineTask. `finally` needs more discussion and would be ideal to add after we add `skip` support as part of `runOn`.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
pritidesai: We have few different options to handle this validation:

1. Keep it the way its implemented here, i.e. `runAfter` as a list and derives order of execution. `runOn` as a map of `task` and a list of states. Tasks specified in `runOn` must exist in `runAfter` otherwise results in validation failure. Pros: (1) Explicit and clear, task ordering is driven by `runAfter` only they way we have it today and introducing `runOn` to deal with states of tasks and a decision factor whether to execute next tasks. Cons (2) Verbose, multiple sections in a PipelineTask to handle execution of a task on parent task failure or skipping of a parent.   

2. Like @vdemeester mentioned, in absence of `runAfter`, have `runOn` decide the order of execution in addition to monitoring states of those tasks in absence of `runAfter`. Giving more thought on this, I will correct myself, I think its achievable and will not be limited to number of tasks like I said before. Some how I read it as "not having task names in `runOn`" 😞 In absence of `runAfter`, `runOn` decides order and in presence of `runAfter` and `runOn`, execution order is driven by the union of both (`runOn` and `runAfter`). Pros: Reduce duplication of specifying same tasks under `runOn` and `runAfter`. Cons: One more section `runOn` driving the execution order in addition to `runAfter` and `from`.

3. Replace `runOn` with `runAfter` and have `runAfter` take a map in addition to list. Will have to validate this to make sure it doesnt break backward compatibility.

pkg/apis/pipeline/v1alpha1/taskrun_types.go
```diff
@@ -68,6 +68,8 @@ const (
 	// TaskRunSpecStatusCancelled indicates that the user wants to cancel the task,
 	// if not already cancelled or terminated
 	TaskRunSpecStatusCancelled = v1beta1.TaskRunSpecStatusCancelled
+	// TaskRunSpecStatusSkipped indicates that the task should be skipped and not executed
```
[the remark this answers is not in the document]
pritidesai: With these changes, `TaskRun` is created for a task which depends on a certain state of a parent but parent ends up with different state, for example, `TaskB` depends on failure of `TaskA` but `TaskA` finishes with success. In this scenario, a taskrun is created for `TaskB` with this state, more specifically `reason`.  This state is very crucial (1) to determine the status of PipelineRun, Completed with a skipped task (2) tasks which are dependent on `TaskB` are also marked with the same state instead of failing the whole pipeline

Note these two examples and how `runOn` handles them:

https://github.com/tektoncd/pipeline/blob/64d326f29cee582d7f5a75ae6e9e73fe691cc289/docs/pipelines.md#state-of-a-parent-task-specified-in-runon-section-conflicts-with-the-actual-state

https://github.com/tektoncd/pipeline/blob/64d326f29cee582d7f5a75ae6e9e73fe691cc289/docs/pipelines.md#task-with-multiple-dependencies-based-on-different-states

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -435,8 +435,12 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1alpha1.PipelineRun) er
 		reconciler.EmitEvent(c.Recorder, before, after, pr)
 		return err
 	}
-
-	candidateTasks, err := dag.GetSchedulable(d, pipelineState.SuccessfulPipelineTaskNames()...)
+	// Get the next tasks to be executed which doesnt't have any TaskRuns created yet
+	// ExecutedPipelineTaskNames gets the list of finished tasks (either successful or failed or skipped)
```
afrittoli: IIUC the skipped ones does not include those skipped because of a condition?
Having a runOn(["skipped"])  could be a trick to provide a way to express an `else` statement for a condition, which is something that we miss today, we need to define to opposite conditions for that.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -466,20 +470,31 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1alpha1.PipelineRun) er
 		if rprt == nil {
 			continue
 		}
+		if rprt.IsRunOnConflicted {
+			rprt.TaskRun, err = c.createRunOnConflictedTaskRun(rprt, pr)
```
afrittoli: Do we really need to create the TaskRun here? 
This TaskRun does not meet the precondition for being scheduled because it depends on a condition (as defined in runOn) that did not happen. And TaskRuns that depended on this one will never be scheduled because their predecessor in the DAG did not run.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -73,6 +77,12 @@ type ResolvedPipelineRunTask struct {
 	ResolvedTaskResources *resources.ResolvedTaskResources
 	// ConditionChecks ~~TaskRuns but for evaling conditions
 	ResolvedConditionChecks TaskConditionCheckState // Could also be a TaskRun or maybe just a Pod?
+	// IsFailurePermitted is driven by the runOn section of the dependent tasks, runOn:  state: ["failure"]
+	// it is set to true if any of the tasks within the pipeline has dependency on this task's failure state
+	// by default, IsFailurePermitted is set to false
+	IsFailurePermitted bool
```
afrittoli: I would like to propose a change in how we handle this logic here.
I'm not sure I've thought enough about this, but these are the basic ideas:
- any taskrun is "permitted" to fail, it simply fails
- we include failed taskruns in the list of executed tasks
- we include the filters defined by the `runOn` in the logic that calculates which task should be included in the list of schedulable ones at a certain point in time
- we change the logic that decides when a pipelinerun is finished. The following conditions should be met: the method that gets the next schedulable task does not return anything, no taskrun is in a transient state. We need to make sure that we don't meet this condition when for instance taskruns are waiting for a pod to be scheduled, but I think we should be safe there

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -73,6 +77,12 @@ type ResolvedPipelineRunTask struct {
...
+	// IsRunOnConflicted is set to true for a task with runOn states conflicting with actual states of parent task
+	IsRunOnConflicted bool
```
afrittoli: I'm not completely convinced that we need to store this state here. 
Could the function that returns the list of schedulable TaskRuns use the filters defined in runOn and the state of TaskRuns executed until now to decide if a taskrun should be schedulable or not?

NIT: the "conflicted" term does not resonate with me, but I'm miserably failing to offer an alternative that sounds good. Perhaps using the opposite - `isRunOnSatisfied` - but this does not sound very good to me either.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -129,6 +139,18 @@ func (t ResolvedPipelineRunTask) IsCancelled() bool {
 	return c.IsFalse() && c.Reason == v1alpha1.TaskRunSpecStatusCancelled
 }
 
+// isSkippedAsStateConflicted return true only if the taskRun has failed with reason SkippedAsStateConflicted
```
afrittoli: I think it would be semantically more correct to actually skip the taskrun, and not fail it with a "skipped" status.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: I commented this already, but instead of marking the TaskRun as conflicted, could we simply not return it here?  Is the "conflicted" status used to avoid having to match the runOn filters in various reconcile cycles?

Doesn't this pose the problem that if a taskrun depended on a conflicted one, with a runOn(failure), it would then get execute, instead of being skipped too?

pkg/reconciler/pipelinerun/pipelinerun.go
dibyom: Yup, I was thinking that could be a solution to #1023

docs/pipelines.md
```diff
@@ -547,6 +548,348 @@ build-app  build-frontend
...
+      runOn:
+        - task: task1
+          states: ["success"]
+    - name: task3
+      taskRef:
+        Name: successful-task
+      runAfter: ["task1"]
+      runOn:
+        - task: task1
+          states: ["failure"]
```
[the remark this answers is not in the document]
dibyom: I think this means that task3 runs if and only if task1 has failed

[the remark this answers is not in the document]
afrittoli: Oh, I see, so the `runOn` overrides `runAfter` in a way.
The meaning of runAfter to today is basically runOn['success']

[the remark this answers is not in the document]
pritidesai: yup, it does, and `runAfter` today is `runOn` with `success`.

pritidesai:
> Again, thank you for this, it's great work!

Thank you @afrittoli, very excited for this, its been fun!

> I have now reviewed most of the code and I have two main questions that I would like to clarify:
> 
> * why do we need to mark tasks as conflicted, run them with failure, instead of skipping them and not returning them as candidates for execution?

Because we want to continue with an execution until there are no tasks left, and the biggest reason being, from an audit perspective, it would be convenient to know what happened to a task, was task executed, not executed, if not then why not. Also, to design a generic solution for simple use cases with single branch and more complex with multiple branches, for example, 

```
          task1 (finishes with success)
 failure |
            v
          task2
```

`task2` is marked as `conflicted` and PipelineRun finishes with `Completed` and 1 `skipped` task. In this case, we can skip `task2` instead of returning it as a candidate for execution. But to apply the same logic to more complex use case is challenging. Complex use case includes Pipeline with two ore more branches, we need a way to distinguish when a task actually failed vs when a conflict happens,  for example,

```
                  task1
    success /     \ failure
         task2   task3
   failure |       | success
         task4   task5
```

If `task1` is successful, `task3` and `task5` both are marked as `conflicted` whereas `task2` and `task4` finishes execution and PipelineRun is marked as `Completed` with two tasks `skipped`.

> * would it be possible to collapse the "runAfter" and "runOn" into one? they are strongly related, "runAfter" is a special case of "runOn"

Sure, we can, I am open to collapsing both, like @vdemeester also suggested.

pkg/apis/pipeline/v1alpha1/taskrun_types.go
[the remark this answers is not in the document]
afrittoli: OK, I see, thanks for the clarification!

I think this calls for a bit of extra dicussion, since the proposed behaviour diverges from what we do today:
- if a task fails, the pipeline is failed. Any task that was running until that point finishes, and no new task is scheduled. Tasks that are not executed have no taskrun associated and are not tracked in the pipelinerun status either
- if a task if not executed because of a `runAfter` condition that is not met, the taskrun is not created and not tracked in the pipelinerun status
- if a task does not pass a condition, it is not executed, so no taskrun is created for it. However it is tracked in the pipelinerun status (see example below), because the conditions are part of it. tasks that depend on a task skipped because of a condition are not executed and not tracked in the pipelinerun status

```
      status:
        conditions:
        - lastTransitionTime: "2020-03-12T21:02:42Z"
          message: ConditionChecks failed for Task tekton-random-check-xsz9w-check-reset-dr9hr
            in PipelineRun tekton-random-check-xsz9w
          reason: ConditionCheckFailed
          status: "False"
          type: Succeeded
        podName: ""
```

For `runOn`, to be consistent with today's behaviour for `runAfter`, we should not create a taskrun and not track it in the pipelinerun status. 
If we wanted to be consisted with the conditions instead, we could track the evaluation of the runOn in the pipelinerun status, but not create a taskrun.

This said I'm not necessarily advocating for our current approach. It is straight-forward and fairly readable, but it has a downside: anyone (CLI, Dashboard, human user) trying to understand what happened in the pipeline, need to consult the pipeline definition, infer the DAG and compare with the known states to fully understand what happened. For fairly simple pipelines this is easy to do, but it may not always be so obvious for more complex ones, especially in case of humans looking at API output directly. 

One thing that we may consider doing in future is to track all taskruns in the pipelinerun status; if we do so we should for all cases, and I think the change should not happen in this PR.

Something I would not do in any case is to actually submit new taskruns resources to the API only to mark them as skipped. That information should be inferred from the parent object, the pipelinerun.

@dibyom @vdemeester @sbwsg @bobcatfish thoughts?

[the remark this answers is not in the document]
pritidesai: Sure, it would be great to discuss this further and get everyone's input 👍 

This is certainly different than the way we handle conditional failures today. The biggest difference with `runOn` is pipeline continues even after a conflict is discovered.

Today, we stop pipelinerun as soon as a task has failed or condition resulted in failure. In case when a task fails, `pipelinerun` would have the `taskrun` of the failed task with the failure reason. 

```
tkn pr describe 01-pipelinerun-without-runon

💌 Message

TaskRun 01-pipelinerun-without-runon-task1-ln8zz has failed ("step-fail" exited with code 1 (image: "docker-pullable://ubuntu@sha256:04d48df82c938587820d7b6006f5071dbbffceb7ca01d2814f81857c631d44df"); for logs run: kubectl -n default logs 01-pipelinerun-without-runon-task1-ln8zz-pod-pjgmg -c step-fail)

🗂  Taskruns

 NAME                                         TASK NAME   STARTED          DURATION    STATUS
 ∙ 01-pipelinerun-without-runon-task1-ln8zz   task1       13 seconds ago   8 seconds   Failed
```

Now, its possible to retrieve the `taskrun`:

```
tkn tr describe 01-pipelinerun-without-runon-task1-ln8zz

Message

"step-fail" exited with code 1 (image: "docker-pullable://ubuntu@sha256:04d48df82c938587820d7b6006f5071dbbffceb7ca01d2814f81857c631d44df"); for logs run: kubectl -n default logs 01-pipelinerun-without-runon-task1-ln8zz-pod-pjgmg -c step-fail

 NAME     STATUS
 ∙ fail   Error
```

Whereas for a task with a condition and when that condition fails, `pipelinerun` contains `taskrun` for the task:

```
tkn pr describe pipelinerun-condition-failures
🗂  Taskruns

 NAME                                           TASK NAME   STARTED   DURATION   STATUS
 ∙ pipelinerun-condition-failures-task1-j46vj   task1       ---       ---        Failed(ConditionCheckFailed)
```

And can not retrieve that `taskrun`:

```
tkn tr describe pipelinerun-condition-failures-task1-j46vj
Error: failed to find taskrun "pipelinerun-condition-failures-task1-j46vj"
``` 
But at least `taskrun` for the condition exists and can be retrieved:

```
tkn tr list
NAME                                                              STARTED          DURATION    STATUS
pipelinerun-condition-failures-task1-j46vj-false-conditio-gkllw   8 minutes ago    7 seconds   Failed
```

In case of `runOn`, if a `taskrun` is listed in the `pipelinerun` as conflicted, providing an ability to retrieve that `taskrun` would be helpful. But we can certainly drop it out of this PR if needed. 

```
tkn pr describe pipelinerun-runon-conflicted
🌡️  Status

STARTED          DURATION   STATUS
17 minutes ago   1 minute   Succeeded(Completed)

🗂  Taskruns

 NAME                                                            TASK NAME   STARTED          DURATION     STATUS
 ∙ pipelinerun-runon-conflicted-task2-s942v   task2       16 minutes ago   23 seconds   Failed
 ∙ pipelinerun-runon-conflicted-task3-8mhh9   task3       16 minutes ago   ---          Failed(SkippedAsStateConflicted)
 ∙ pipelinerun-runon-conflicted-task1-trxwl   task1       17 minutes ago   22 seconds   Succeeded
```

```
 tkn tr describe pipelinerun-runon-conflicted-task3-8mhh9
🌡️  Status

STARTED          DURATION    STATUS
17 minutes ago   ---         Failed(SkippedAsStateConflicted)

Message

TaskRun "pipelinerun-runon-conflicted-task3-8mhh9" was skipped
```

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -231,3 +236,29 @@ type PipelineList struct {
...
+}
+
+// PipelineTaskState represents the state of a parent PipelineTask that it has to be in for it to run
+type PipelineTaskState = string
+
+const (
+	// PipelineTaskStateSuccess indicates that a PipelineTask has finished execution successfully
+	PipelineTaskStateSuccess PipelineTaskState = "success"
+	// PipelineTaskStateFailure indicates that a PipelineTask has finished execution and has failed
+	PipelineTaskStateFailure PipelineTaskState = "failure"
```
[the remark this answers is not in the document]
pritidesai: yes and no, `failure` is equivalent to `failed`, i.e. not success and not unknown. `skip` is equivalent to `failed` with `conditionCheckFailed`. Should we include `cancel` or `timeout` here in `skip`? (`skip` is not implemented in this PR)

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: yup, `runOn: [skip]` is equivalent to run this task if parent was not executed because of a condition failure i.e.

```
if condition then
    task1
else
    task2
```

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: Answered above ⬆️ , once again, we can drop creating taskrun here but would be helpful to have one.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: > I would like to propose a change in how we handle this logic here.
> I'm not sure I've thought enough about this, but these are the basic ideas:
> 
> * any taskrun is "permitted" to fail, it simply fails

not sure if I am following this, on a high level here not any taskrun is permitted to fail, after task is executed and if it fails, pipelinerun determines to exit with failure or continue executing rest of the tasks based on this flag `IsFailurePermitted`. This flag is set while parsing the pipelinerun.

> * we include failed taskruns in the list of executed tasks

Yes, we do include failed taskruns in the list of executed tasks.

> * we include the filters defined by the `runOn` in the logic that calculates which task should be included in the list of schedulable ones at a certain point in time

Yup, based on the states specified in `runOn` and actual states of a task, a list of schedulable tasks is generated.

> * we change the logic that decides when a pipelinerun is finished. The following conditions should be met: the method that gets the next schedulable task does not return anything, no taskrun is in a transient state. We need to make sure that we don't meet this condition when for instance taskruns are waiting for a pod to be scheduled, but I think we should be safe there

Again, not sure, if I am following this, but the logic which decides if pipeline is finished has changed and now includes, (1) check on failed task if failure is permitted (2) skipped tasks due to conflicts are counted as skipped

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: > I'm not completely convinced that we need to store this state here.
> Could the function that returns the list of schedulable TaskRuns use the filters defined in runOn and the state of TaskRuns executed until now to decide if a taskrun should be schedulable or not?

Yup, that's exactly how its implemented. A taskrun should be scheduled or not is determined solely using the filters (states) defined in `runOn` and actual states of taskruns.
These flags are mainly used to determine (1) create skipped state taskrun if `IsRunOnConflicted` is set (2) pipelinerun decides whether to continue when a task fails if `IsFailurePermitted` is set for that task. Note here with DAG, a list of schedulable tasks is determined by looking at the N-1 level of the tree whereas `IsFailurePermitted` is set for a parent task based on its children `runOn` states.  

> NIT: the "conflicted" term does not resonate with me, but I'm miserably failing to offer an alternative that sounds good. Perhaps using the opposite - `isRunOnSatisfied` - but this does not sound very good to me either.

Haha 😂I am having same issue 😫, cannot come up with better option to signify conflict between a state specified in `runOn` vs actual taskrun state.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: we need a way to signify that a parent task was skipped and hence its children should be skipped too and that the task is visited and was skipped

pritidesai:
@afrittoli I think 🤔I have responded to all of your comments and conclude it with the following two items which might need further discussion:

1) Should `runOn` derive task execution order in addition to `runAfter`? like you and @vdemeester suggested here https://github.com/tektoncd/pipeline/pull/2094#discussion_r387505539

2) Whether to create a taskrun in case when a task has specified certain `runOn` states but actual states are different i.e. conflict. If we decide not to create a skipped taskrun i.e. change the way its implemented in this PR, we will need a way to distinguish a task is visited but was skipped vs a task hasn't attempted, this is tricky from code perspective as `GetNextTasks` function does not have a pointer receiver and `PipelineRunState` does not convey that message to reconciler.

pkg/apis/pipeline/v1alpha1/taskrun_types.go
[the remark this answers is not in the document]
afrittoli: > Today, we stop pipelinerun as soon as a task has failed or condition resulted in failure. In case when a task fails, `pipelinerun` would have the `taskrun` of the failed task with the failure reason.

What do you mean by "condition resulted in failure"?
If the check in a condition is false, so the condition fails, the associated task is not executed, however the execution of the pipelinerun continues.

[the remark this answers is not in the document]
afrittoli: > And can not retrieve that `taskrun`:
> 
> ```
> tkn tr describe pipelinerun-condition-failures-task1-j46vj
> Error: failed to find taskrun "pipelinerun-condition-failures-task1-j46vj"
> ```

Personally I think this correct, the task was not executed at all and so it was not created.
The information about what happened is available in the pipelinerun. 
If we create a taskrun, even as a failed one, it will trigger the taskrun controller, which will emit events and do some work.

pkg/reconciler/pipelinerun/pipelinerun.go
afrittoli: On second thoughts, runOn: [skip] might be not a bit loose, at least for handling conditionals. If we had this setup (sorry about the approximated YAML syntax):
```
task1:
  conditions:
    - condA
taskElseCondA:
    runOn: [ task1, skip ]

task2:
  conditions:
    - condB
taskElseCondB:
    runOn: [ task2, skip ]

```

If `condA` fails, task1 willl be skipped.  As a consequence `task2` fill be skipped too.
`taskElseCondA` will be executed, but what happens with `taskElseCondB`? 
Since `task2` was skipped, it could be executed. However the intention of the pipeline author there is probaly that `taskElseCondB` is executed if `condB` is false, which we haven't evaluated. 

I think it would be best to have the `else` task bound to the conditional rather than the task associated to the conditional.

pkg/apis/pipeline/v1alpha1/taskrun_types.go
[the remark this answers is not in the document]
vdemeester: > This said, I'm not necessarily advocating for our current approach. It is straight-forward and fairly readable, but it has a downside: anyone (CLI, Dashboard, human user) trying to understand what happened in the pipeline, need to consult the pipeline definition, infer the DAG and compare with the known states to fully understand what happened. For fairly simple pipelines this is easy to do, but it may not always be so obvious for more complex ones, especially in case of humans looking at API output directly.

@afrittoli I think we need to capture this in an issue and work towards having a `status` on `PipelineRun` that is *self descriptive* in some ways (aka making it easier for cli, dashboard, .) to get information without having to query the `Pipeline` spec. 

cc @bobcatfish

pritidesai:
> 2. Whether to create a taskrun in case when a task has specified certain `runOn` states but actual states are different i.e. conflict. If we decide not to create a skipped taskrun i.e. change the way its implemented in this PR, we will need a way to distinguish a task is visited but was skipped vs a task hasn't attempted, this is tricky from code perspective as `GetNextTasks` function does not have a pointer receiver and that's why `PipelineRunState` does not convey that message to reconciler, which can be changed but something to note.

Without registering `taskRun` in the reconciler for skipped tasks and only having the status in the PipelineRunStatus, retrieving tasks dependent on these skipped tasks is failing (tried couple of options today 😞), I can look into it more after we decide the best way to represent skipped tasks.

bobcatfish:
Hey @pritidesai and all! I've come into this late and I have to admit it's a bit hard to follow what's going on at this point 😅  Which to me points toward us needing to do a bit more to flesh out the design before we would merge this entirely. 

The biggest question on my mind is:

* What is the smallest step we can take toward this right now, as we continue to hammer out the details?

I've tried to identify a few themes in the discussions so far:

1. Creating taskrun objects for tasks that have not executed
2. The syntax for runAfter + runOn
3. The "conflicted" state
4. Handling "skipped" tasks

I suggest we try to tease these apart and tackle them individually:

* (1) <-- i strongly feel we should NOT be creating taskrun objects for Tasks that are not executing. If the core issue is being able to see the state of all of the tasks in the PipelineRun, I think we can do that by updating the status of the pipelinerun and don't need to create actual taskruns (i think @dibyom has been discussing this a bit already). if the current code is making this hard, maybe a great first step here would be to refactor the code in prepartion? i.e. make it so that the state of completed TaskRuns is available in teh status of the PipelineRun and doesnt need to be retrieved
* (2) <-- im not sure we should use the map because think it's hard to understand what it means, e.g.:
image
Does this mean when task1 is success OR failure AND when task2 is success? or is there an AND between task1 and task2? <-- and this comes back to: what is the simplest thing we can do next?
* (3) I don't think there should be a conflicted state; not executing tasks that are meant to execute when when something fails is a totally normal and fine thing
* (4) can we avoid (skip?) handling skip for now and add it later?

My recommendations for next steps:
- Start doing any refactoring that can be done now
- Write up 2-4 in more detail, either in the existing design doc, in the pipeline.md in this PR or in brand new docs so we can look athem. I would prioritize (2) b/c nailing down the syntax will unblock some incremental changes

We're going to be having a beta/workflow meeting on monday at 12pm EST - if you were able to attend @pritidesai we could discuss in more detail then, but if not let's find another time when we could meet and work out the details, if that works for you? If not then writing this down clearly will be more important than ever

Anyway that's my 2 cents and again I am SO EXCITED to get this in - let's figure out what incremental steps we can start making so we can start merging PRs and not leave this big one open for a long time. ❤️

pritidesai:
> Hey @pritidesai and all! I've come into this late and I have to admit it's a bit hard to follow what's going on at this point 😅 Which to me points toward us needing to do a bit more to flesh out the design before we would merge this entirely.
 
👍

> The biggest question on my mind is:
> 
> * What is the smallest step we can take toward this right now, as we continue to hammer out the details?
 
> I've tried to identify a few themes in the discussions so far:
> 
> 1. Creating taskrun objects for tasks that have not executed
> 2. The syntax for runAfter + runOn
> 3. The "conflicted" state
> 4. Handling "skipped" tasks

Yup, we need a discussion on these topics, sooner the better.

> I suggest we try to tease these apart and tackle them individually:
> 
> * (1) <-- i strongly feel we should NOT be creating taskrun objects for Tasks that are not executing. If the core issue is being able to see the state of all of the tasks in the PipelineRun, I think we can do that by updating the status of the pipelinerun and don't need to create actual taskruns (i think @dibyom has been discussing this a bit already). if the current code is making this hard, maybe a great first step here would be to refactor the code in prepartion? i.e. make it so that the state of completed TaskRuns is available in teh status of the PipelineRun and doesnt need to be retrieved

I have no strong preference for one over the other. The changes in this PR has it implemented such that `TaskRun` does get created for tasks that are not executing. The core issue otherwise is not being able to retrieve a list of tasks to be executed next for which the parent was not executed. The task will be skipped if parent was skipped but we should be able to add that status in PipelineRun status.

If we change it to not create `taskRun`, the implementation will change in the following area:
`GetNextTasks` operates on `PipelineRunState` which holds the entire `PipelineRun` after resolving tasks, resources, conditions, etc and most importantly `TaskRun` registered in reconciler (actual running `taskRun`).  In case of `runOn` with skipped tasks, `GetNextTasks` would need to read `PipelineRun.Status` in addition to `PipelineRunState`. `PipelineRunState` has `runOn` section with tasks and states. And `PipelineRun.Status` holds `taskRun` object indicating that the task was skipped. They key issue here is identifying when to retrieve a task from `PipelineRunState` vs `PipelineRun.Status`. We could dedicate `PipelineRun.State` to only return skipped tasks and ignore successful/failed tasks. But I feel, it makes the whole process very complicated iterating over two different data structure and making sure we dont include task which was not suppose to 😞 

One example here is:

```
                  task1
    success /     \ failure
         task2     task3
   failure |          | success
         task4     task5
```

Let's say, `task1` was successful which means `task3` and `task5` should be marked as `skipped` since it depends on failure of `task1`. `PipelineRunState` wouldn't carry `taskRun` for `task3` and `task5` vs `PipelineRun.Status` would have all five tasks and their status. 

> * (2) <-- im not sure we should use the map because think it's hard to understand what it means, e.g.:
>   image
>   Does this mean when task1 is success OR failure AND when task2 is success? or is there an AND between task1 and task2? <-- and this comes back to: what is the simplest thing we can do next?

😢 yup, there is an AND between `task1` and `task2`. It means execute this task:

```
if (task1 == success OR task1 == failure) AND task2 == success {
....
}
```

We could simplify by going back to list but will loose an ability to support this same example 😜: 
 
```
runAfter: ["task1", "task2"]
runOn: ["success", "failure"]
```

this means, execute this task if both `task1` and `task2` are either successful or both have failed 🙃   

> * (3) I don't think there should be a conflicted state; not executing tasks that are meant to execute when when something fails is a totally normal and fine thing

humm 🤔, `conflicted` state is to indicate that the task should be skipped since the states didnt match, I think it needs to rename to something more meaningful, `conflicted` is used internally to create a skip taskRun (actual taskRun or just object). 

> * (4) can we avoid (skip?) handling skip for now and add it later?

Yup, I have it implemented but not added in this PR, its already becoming complicated 😫 
 
> My recommendations for next steps:
> 
> * Start doing any refactoring that can be done now
> * Write up 2-4 in more detail, either in the existing design doc, in the pipeline.md in this PR or in brand new docs so we can look at hem. I would prioritize (2) b/c nailing down the syntax will unblock some incremental changes

Yup, syntax is the key, lets nail it down first. This PR has doc here: 
https://github.com/tektoncd/pipeline/blob/64d326f29cee582d7f5a75ae6e9e73fe691cc289/docs/pipelines.md#failure-strategies
I will either create a separate PR or update existing one with more details.

> We're going to be having a beta/workflow meeting on monday at 12pm EST - if you were able to attend @pritidesai we could discuss in more detail then, but if not let's find another time when we could meet and work out the details, if that works for you? If not then writing this down clearly will be more important than ever

Yup, I will see you during beta/working meeting on Monday 😄 

> Anyway that's my 2 cents and again I am SO EXCITED to get this in - let's figure out what incremental steps we can start making so we can start merging PRs and not leave this big one open for a long time. ❤️

Thanks a bunch @bobcatfish 😍, yup super excited 💃

pritidesai:
I feel like creating a separate PR without creating taskRun for skipped tasks before we meet on Monday to better understand the changes, hopefully 🤞I can get it through with my three kids home 👶👶👧

pritidesai:
I have one more PR created to avoid creating Task Run here, PR #2261 🏃‍♀ 🏃

afrittoli:
> I have one more PR created to avoid creating Task Run here, PR #2261 🏃‍♀ 🏃
> 
> It still shows in the PipelineRun Status but actual TaskRun is not created, this [example](https://github.com/tektoncd/pipeline/blob/4a9ab268abd3f1d328e2f91fb4a908ed4d2ea566/docs/pipelines.md#pipeline-with-multiple-dependencies-based-on-different-states) has the details.

Thank you! 🙏

afrittoli:
> > Hey @pritidesai and all! I've come into this late and I have to admit it's a bit hard to follow what's going on at this point 😅 Which to me points toward us needing to do a bit more to flesh out the design before we would merge this entirely.
> 
> 👍
> 
> > The biggest question on my mind is:
> > 
> > * What is the smallest step we can take toward this right now, as we continue to hammer out the details?
> 
> > I've tried to identify a few themes in the discussions so far:
> > 
> > 1. Creating taskrun objects for tasks that have not executed
> > 2. The syntax for runAfter + runOn
> > 3. The "conflicted" state
> > 4. Handling "skipped" tasks
> 
> Yup, we need a discussion on these topics, sooner the better.
> 
> > I suggest we try to tease these apart and tackle them individually:
> > 
> > * (1) <-- i strongly feel we should NOT be creating taskrun objects for Tasks that are not executing. If the core issue is being able to see the state of all of the tasks in the PipelineRun, I think we can do that by updating the status of the pipelinerun and don't need to create actual taskruns (i think @dibyom has been discussing this a bit already). if the current code is making this hard, maybe a great first step here would be to refactor the code in prepartion? i.e. make it so that the state of completed TaskRuns is available in teh status of the PipelineRun and doesnt need to be retrieved

+1 

> 
> I have no strong preference for one over the other. The changes in this PR has it implemented such that `TaskRun` does get created for tasks that are not executing. The core issue otherwise is not being able to retrieve a list of tasks to be executed next for which the parent was not executed. The task will be skipped if parent was skipped but we should be able to add that status in PipelineRun status.
> 
> If we change it to not create `taskRun`, the implementation will change in the following area:
> `GetNextTasks` operates on `PipelineRunState` which holds the entire `PipelineRun` after resolving tasks, resources, conditions, etc and most importantly `TaskRun` registered in reconciler (actual running `taskRun`). In case of `runOn` with skipped tasks, `GetNextTasks` would need to read `PipelineRun.Status` in addition to `PipelineRunState`. `PipelineRunState` has `runOn` section with tasks and states. And `PipelineRun.Status` holds `taskRun` object indicating that the task was skipped. They key issue here is identifying when to retrieve a task from `PipelineRunState` vs `PipelineRun.Status`. We could dedicate `PipelineRun.State` to only return skipped tasks and ignore successful/failed tasks. But I feel, it makes the whole process very complicated iterating over two different data structure and making sure we dont include task which was not suppose to 😞
> 
> One example here is:
> 
> ```
>                   task1
>     success /     \ failure
>          task2     task3
>    failure |          | success
>          task4     task5
> ```
> 
> Let's say, `task1` was successful which means `task3` and `task5` should be marked as `skipped` since it depends on failure of `task1`. `PipelineRunState` wouldn't carry `taskRun` for `task3` and `task5` vs `PipelineRun.Status` would have all five tasks and their status.
> 
> > * (2) <-- im not sure we should use the map because think it's hard to understand what it means, e.g.:
> >   image
> >   Does this mean when task1 is success OR failure AND when task2 is success? or is there an AND between task1 and task2? <-- and this comes back to: what is the simplest thing we can do next?
> 
> 😢 yup, there is an AND between `task1` and `task2`. It means execute this task:
> 
> ```
> if (task1 == success OR task1 == failure) AND task2 == success {
> ....
> }
> ```
> 
> We could simplify by going back to list but will loose an ability to support this same example 😜:
> 
> ```
> runAfter: ["task1", "task2"]
> runOn: ["success", "failure"]
> ```
> 
> this means, execute this task if both `task1` and `task2` are either successful or both have failed 🙃
> 

Personally I really feel that `runOn` is a special case of `runAfter` - specifically if both `runOn` and `runAfter` were specified on a task it would not be very intuitive to understand the resulting rule to be applied.

One alternative would be to leave the `runAfter` rules as they are today, and change this so that the failure strategy is defined on the task that fails: `Pipeline` would define what is the failure behaviour for each `Task`. i.e. fail or continue (with default == fail like today).

> > * (3) I don't think there should be a conflicted state; not executing tasks that are meant to execute when when something fails is a totally normal and fine thing
> 
> humm 🤔, `conflicted` state is to indicate that the task should be skipped since the states didnt match, I think it needs to rename to something more meaningful, `conflicted` is used internally to create a skip taskRun (actual taskRun or just object).

My point on this was that we don't explicitly track skipped taskruns today, we simply do not run them, and the initial implementation of `runOn` could do the same.  We could add tracking of skipped taskruns into the `PipelineRun` status as a follow-up PR - but we should then do that for all cases, even when `runOn` is not used at all, and also for skip in case of conditionals. 

> 
> > * (4) can we avoid (skip?) handling skip for now and add it later?
> 
> Yup, I have it implemented but not added in this PR, its already becoming complicated 😫
> 
> > My recommendations for next steps:
> > 
> > * Start doing any refactoring that can be done now
> > * Write up 2-4 in more detail, either in the existing design doc, in the pipeline.md in this PR or in brand new docs so we can look at hem. I would prioritize (2) b/c nailing down the syntax will unblock some incremental changes
> 
> Yup, syntax is the key, lets nail it down first. This PR has doc here:
> https://github.com/tektoncd/pipeline/blob/64d326f29cee582d7f5a75ae6e9e73fe691cc289/docs/pipelines.md#failure-strategies
> I will either create a separate PR or update existing one with more details.
> 
> > We're going to be having a beta/workflow meeting on monday at 12pm EST - if you were able to attend @pritidesai we could discuss in more detail then, but if not let's find another time when we could meet and work out the details, if that works for you? If not then writing this down clearly will be more important than ever
> 
> Yup, I will see you during beta/working meeting on Monday 😄
> 
> > Anyway that's my 2 cents and again I am SO EXCITED to get this in - let's figure out what incremental steps we can start making so we can start merging PRs and not leave this big one open for a long time. ❤️
> 
> Thanks a bunch @bobcatfish 😍, yup super excited 💃

pkg/apis/pipeline/v1alpha1/taskrun_types.go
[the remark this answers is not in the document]
dibyom: Big +1 to tracking the entire state using the pipelinerun status instead of looking up taskruns

pkg/reconciler/pipelinerun/pipelinerun.go
dibyom: You are right task 2 will be skipped. One way to do what you want would be for task2 to say `runAfter: [task1: [`success`, `skip`]

pritidesai:
/hold
We are revisiting the syntax of `runOn` (list vs map) with the combination of `runAfter`. Will update this PR once we have the changes in.

bobcatfish:
Closing this for now as we are revisiting the design, see https://github.com/tektoncd/pipeline/issues/1684#issuecomment-611016087 for discussion!

PR #2115, opened 2020-02-26
Add handling of working directory

othomann:
# Changes

Tekton currently overwrites the working directory in task containers if a specific working directory is not specified. This behaviour is problematic when an image explicitly sets a working directory and expects it to be set to that value upon running.

This PR introduces a feature-flags ConfigMap with a single flag - disable-working-directory-overwrite - that, when set to "true" will prevent Tekton from overriding the working directory in Task containers. This behaviour will be kept for one release which must include a deprecation warning that the behaviour will be changing soon. In the release after that the flag will be flipped so that the default behaviour becomes for Tekton to NOT override the working directory.

Contributes to issue #1836

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Tekton currently overwrites the working directory in Step containers if not explicitly specified by the user. This behaviour is problematic when an image explicitly sets the working directory and expects it to be set to that value upon running. You can now disable this behaviour in Tekton by using the feature-flags ConfigMap. See docs/install.md for further details on disabling this behaviour.
```

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -34,8 +34,9 @@ import (
 const (
 	homeDir = "/tekton/home"
 
-	featureFlagConfigMapName     = "feature-flags"
-	featureFlagDisableHomeEnvKey = "disable-home-env-overwrite"
+	featureFlagConfigMapName        = "feature-flags"
+	featureFlagDisableHomeEnvKey    = "disable-home-env-overwrite"
+	featureFlagDisableWorkingDirKey = "disable-working-directory-overwrite"
 
 	taskRunLabelKey = pipeline.GroupName + pipeline.TaskRunLabelKey
 )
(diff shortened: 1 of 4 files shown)


skaegi:
/lgtm
I just tested with my kube-hunter image and now it works without explicitly setting workingDir

pkg/pod/pod.go
```diff
@@ -167,7 +168,7 @@ func MakePod(images pipeline.Images, taskRun *v1alpha1.TaskRun, taskSpec v1alpha
 	// TODO(#1605): Remove this loop and make each transformation in
 	// isolation.
 	for i, s := range stepContainers {
-		if s.WorkingDir == "" {
+		if s.WorkingDir == "" && shouldOverrideWorkingDir(kubeclient) {
```
skaegi: You should pre-compute this outside the for loop

sbwsg:
@vdemeester good point!  I took a very quick look at the configmap watcher and got :fearful: because I couldn't easily trace where the actual configmaps to watch were being configured.  I can take another look at this.  Issue to track it: https://github.com/tektoncd/pipeline/issues/2117

pkg/pod/pod.go
othomann: done.

Issue #1839, opened 2020-01-09
 Feature: Versioning on Tasks/Pipelines


pierretasci:
# Abstract
A way to define a specific version of a named pipeline or task so that the spec of that pipeline or task can be referenced at a specific moment in time. For example, the way docker containers work today with tags whereby I am able to run `container:tag` for any valid tag of the same container name. 

# Use Cases
- In a production system, it is likely a project's Tekton pipeline will evolve its definition over time. If I wanted to run the pipeline as it existed a month ago, I have no good way to know what the spec was if it has since changed.

- The ability to test a change to a pipeline's spec without affecting the pipeline runs of other users that share the same spec. 

- Improve the ability to share and reuse tasks so I can depend on a pinned version of a Task from a catalog while the actual spec changes.

# Details
I could imagine this being addressed as part of the Tekton spec. For now, our workaround has been to publish tasks and pipelines with the "version" in the name of the task. This does introduce a discoverability problem as well as clutter.

A solution that is part of the Tekton spec would make a lot of sense. Eg:

```
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: my-unique-tests
spec:
  - tag: v1
     spec: 
       steps:
         - name: run-test
            image: ubuntu
            command: ["foo"]
---
apiVersion: tekton.dev/v1alpha1
kind: TaskRun
metadata:
  name: my-unique-tests-run
spec:
  taskRef:
    name: my-unique-task
    tag: v1
```

An easy addition to this is a default "latest" to keep the existing functionality, again, similar to docker.

vdemeester:
I think the main question is, is this in scope for tektoncd/pipeline or some higher level component (from Tekton or not) ? versionning could be done using different name and labels, … and managed by an higher level component, keeping the tekton api as simple (relatively speaking) as possible.

This is also where experiment like [`catalgos`](https://github.com/tektoncd/experimental/tree/master/catalogs) (expect some code in the next week :angel:), from https://github.com/tektoncd/community/issues/53) comes into place (on the discoverability side of things.

/area api
/kind question
/kind design

pierretasci:
I completely agree that the main question comes down to whether this lives in tekton or externally. I think that both are viable. Since I brought up the issue, I will make a case in favor of making it a part of the Task spec and that is to aid in task reuse. If the goal is to make tasks that are generic and reusable with a well-defined interface, they will need to be versioned somehow to prevent the definition being changed while an old version is running.

[...]

PR #2137, opened 2020-03-03
 Adds resolver to provide task resolution for images

sthaha:
# Changes
This partially addresses the desire to fetch tasks
 from an OCI image artifact.

 Issue: https://github.com/tektoncd/pipeline/issues/1839

Signed-off-by: Sunil Thaha <sthaha@redhat.com>

# Submitter Checklist

review them:

- [🙅‍♂ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

## Reviewer Notes

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```

### DIFF
--- a/pkg/remote/oci.go
+++ b/pkg/remote/oci.go
@@ -1,0 +1,109 @@
+/*
+Copyright 2019 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package remote
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	imgname "github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned/scheme"
+)
+
+// KeychainProvider is an input to the OCIResolver which returns a keychain for fetching remote images with
+// authentication.
+type KeychainProvider func() (authn.Keychain, error)
+
+// OCIResolver will attempt to fetch Tekton resources from an OCI compliant image repository.
+type OCIResolver struct {
+	imageReference   string
+	keychainProvider KeychainProvider
+}
+
+// GetTask will retrieve the specified task from the resolver's defined image and return its spec. If it cannot be
+// retrieved for any reason, an error is returned.
+func (o OCIResolver) GetTask(taskName string) (*v1beta1.TaskSpec, error) {
+	taskContents, err := o.readImageLayer("task", taskName)
+	if err != nil {
+		return nil, err
+	}
+
+	// Deserialize the task into a valid task spec.
+	var task v1beta1.Task
+	_, _, err = scheme.Codecs.UniversalDeserializer().Decode(taskContents, nil, &task)
+	if err != nil {
+		return nil, fmt.Errorf("Invalid remote task %s: %w", taskName, err)
+	}
+
+	return &task.Spec, nil
+}
+
+func (o OCIResolver) readImageLayer(kind string, name string) ([]byte, error) {
+	imgRef, err := imgname.ParseReference(o.imageReference)
+	if err != nil {
+		return nil, fmt.Errorf("%s is an unparseable task image reference: %w", o.imageReference, err)
+	}
+
+	// Create a keychain for use in authenticating against the remote repository.
+	keychain, err := o.keychainProvider()
+	if err != nil {
+		return nil, err
+	}
+
+	img, err := remote.Image(imgRef, remote.WithAuthFromKeychain(keychain))
+	if err != nil {
+		return nil, fmt.Errorf("Error pulling image %q: %w", o.imageReference, err)
+	}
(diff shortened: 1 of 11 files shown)


pkg/remote/oci.go
```diff
@@ -0,0 +1,110 @@
...
+	_, _, err = scheme.Codecs.UniversalDeserializer().Decode(taskContents, nil, &task)
+	if err != nil {
+		return nil, fmt.Errorf("Invalid remote task %s: %w", taskName, err)
+	}
+
+	return &task.Spec, nil
+}
+
+func (o OCIResolver) readImageLayer(kind string, name string) ([]byte, error) {
+	// TODO: When this is moved into the Tekton controller, authorize this
```
sthaha: This will be addressed in a follow up patch as the change seems rather large

vdemeester: Do worry too much about it, we've seen way larger changes :wink:

pkg/remote/oci.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+func init() {
+	// Add Tekton's resources to the K8s deserializer
+	schemeBuilder := runtime.NewSchemeBuilder(v1beta1.AddToScheme)
+	err := schemeBuilder.AddToScheme(scheme.Scheme)
+	if err != nil {
+		log.Panic(err)
+	}
+}
```
vdemeester: why is it required ? :thinking:

pkg/remote/oci.go
pierretasci: So should this be all done in one PR or in multiple?

vdemeester: depend what you mean by "all". But having a full working `pkg/remote/oci.go` file (aka at least with the serviceAccount support, the validation too) works, then other PR to start integrating it into the reconcilier would be others.

pkg/remote/oci.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned/scheme"
+)
+
+// KeychainProvider is an input to the OCIResolver which returns a keychain for fetching remote images with
+// authentication.
+type KeychainProvider func() (authn.Keychain, error)
+
+// ImageResolver will attempt to fetch Tekton resources from an OCI compliant image repository.
```
sthaha: OCIResolver

pkg/remote/oci_test.go
```diff
@@ -0,0 +1,124 @@
...
+	s := httptest.NewServer(registry.New())
+	defer s.Close()
+	u, err := url.Parse(s.URL)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	imgRef, err := name.ParseReference(fmt.Sprintf("%s/test/ociresolver", u.Host))
+	if err != nil {
+		t.Errorf("undexpected error producing image reference %s", err.Error())
```
sbwsg: nit: undexpected -> unexpected

pkg/remote/oci.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned/scheme"
+)
+
+// KeychainProvider is an input to the OCIResolver which returns a keychain for fetching remote images with
+// authentication.
+type KeychainProvider func() (authn.Keychain, error)
+
+// OCIResolver will attempt to fetch Tekton resources from an OCI compliant image repository.
+type OCIResolver struct {
```
sbwsg: What's a typical usage of this type going to look like?  If it's most often going to be instantiated and immediately used like this:

```go
{
  myresolver := remote.OCIResolver{imageRef, kp}
  mytask := myresolver.GetTask(foo)
}
```

then I suggest simply exposing a `GetOCITask` func that takes `(imageReference string, keychainProvider KeychainProvider, taskName string)`.  WDYT?

pierretasci: yeah that makes a ton of sense. I guess it comes down to how this evolves when we add a resolver for Git or other remote sources. I could see an argument for either route. 

I think the idea here is that it might be easier for a `reconciler` to use a factory to instantiate a `Resolver` and call it instead of having to know itself whether it should use an OCIResolver or a GitResolver or whatever else we add. We might even add a "local" resolver and embed the local task resolver into this format. Could be overkill or could be a nice abstraction. Thoughts?

pierretasci:
> * support for multiple api (right now it's `…v1beta1`, at some point we will need to support more than one version at the same time)
> * support for specific version of tekton, or minimum version required

Question for @vdemeester, do you mean that the user should specify a min-supported version of Tekton when referencing the task as in 

```
taskRef:
  name: my-task
  image: docker.com/my-image
  minVersion: v1alpha1
```

or do you mean that we should ensure the version of the fetched task is not incompatible with the version of the `taskrun` or `pipelinerun` that references it?

```
apiVersion: v1beta1
...
taskRef:
  name: my-task
  image: docker.com/my-image # apiVersion: v1alpha1
```

pkg/remote/oci.go
sbwsg: I can definitely see a factory fitting that use case.  I'm not sure it follows that we'd therefore need a `remote.Resolver` interface and creation of structs like this anywhere outside the factory's internals though. 

I think it'll ultimately depend on both the kind of configuration we expect to be available and the signature we want to expose for resolution. Examples of config might be that we allow specific resolvers to be turned on/off/configured via configmap / cli flags.  Examples of public signature might be a func like `ResolveTaskFromRef(taskRef)` that figures out the kind of resolution it needs to make without publicly exposing the Resolver type.

Not to say we need to commit to those kinds of design choices here, more that by introducing the Resolver machinery before we get to that stage we're already kind of leaning towards a design. And as often happens we might end up needing to unwind those kinds of choices based on work discovered as we dig into that stage more thoroughly.  And in turn that can lead to a some extra work reviewing in future as well.

None of this is a blocker at all from my POV, just raising it from my read-through. 👍 

/lgtm

pierretasci: Yup. Makes a ton of sense. We can always wrap the resolver in a function like you mention to give a single interface to caller but I guess the next PR that embeds this in the reconciler will show if it makes it more or less "gross" and we can go from there 😄

Issue #2131, opened 2020-03-02
Tasks with results fail if when docker containers do not run as root


kh-sonrai:
The /tekton/results directory is only writable by root, so images that do not run as root fail to create the directory. If any such docker image is included in a task, the task will fail with the following error if a results section is included in the task spec:
`Error creating the results directory: mkdir /tekton/results: permission denied`
 https://github.com/tektoncd/pipeline/blob/97a0c45e2db51942f6b4cb05ced30521977edd7b/cmd/entrypoint/main.go#L60

othomann:
I will be pleased to fix it as soon as I am done with the PR for pipeline results.

othomann:
I have a fix for it, but I am not sure how to add a regression test for it.

sbwsg:
Closing as fixed by #2143

PR #2143, opened 2020-03-03
Make results folder writable (Fix for #2131)

othomann:
# Changes

Make sure the results folder is writable by any user. See #2131.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Make sure the results folder is writable by any user.
```

### DIFF
--- a/pkg/pod/pod_test.go
+++ b/pkg/pod/pod_test.go
@@ -65,6 +65,14 @@ func TestMakePod(t *testing.T) {
 		VolumeMounts: []corev1.VolumeMount{toolsMount},
 	}
 
+	resultsInit := corev1.Container{
+		Name:         "tekton-results-folder-writable",
+		Image:        images.ShellImage,
+		Command:      []string{"sh"},
+		Args:         []string{"-c", "chmod 777 /tekton/results"},
+		VolumeMounts: implicitVolumeMounts,
+	}
+
 	runtimeClassName := "gvisor"
 	automountServiceAccountToken := false
 	dnsPolicy := corev1.DNSNone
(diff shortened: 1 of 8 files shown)


othomann:
This solves the problem by adding a new init container that sets the results folder to be writable by all users. It doesn't seem to work when the user is using the podTemplate to set the permissions. @sbwsg can give more details on this.

othomann:
> The code looks good to me. We should probably include a description of this behaviour in `docs/developers/README.md`. Up to you if you want to include that in this PR or I can follow up with another.

Done in the latest commit. All commits were squashed and tests are running.

sbwsg:
I'm adding lgtm in addition to my approve earlier since I think the bug in the linked issue is a bit of a showstopper and the fix is relatively small.  I'd like to have this included in the initial 0.11 release.

/lgtm

pkg/pod/pod_test.go
```diff
@@ -65,6 +65,14 @@ func TestMakePod(t *testing.T) {
 		VolumeMounts: []corev1.VolumeMount{toolsMount},
 	}
 
+	resultsInit := corev1.Container{
```
vdemeester: So… this is a *kinda **big*** problem on OpenShift as… it fails (running as non-root, …)

vdemeester: Btw, I don't think this one is actually needed (at least, from the OpenShift point of view, the `emptyDir` should make the path writable)

sbwsg: Sorry about the extra churn from the chmod.  You're right - emptyDir gets 777 perms: https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/emptydir/empty_dir.go#L38-L43

Issue #701, opened 2019-03-31
Is Tekton compatible with injected sidecars? (E.g. istio)


dicarlo2:
How does Tekton handle injected sidecars? Off the top of my head, potential issues:

1. Are containers run by Tekton (git-init, creds-init, etc.) resilient to an unavailable network on pod startup. Use case: Service mesh (istio, linkerd) sidecars can take some time on startup to configure the proxy. During that time, all network connections will fail.
2. Will Tekton kill injected sidecars (or delete the Pod) once the main container exits?

dicarlo2:
Looking through the codebase, here are the answers:

1. No. Containers will not wait for injected sidecars nor are they resilient to network failure. E.g. git-init is a very simple container which just [runs](https://github.com/tektoncd/pipeline/blob/9428c431bcef5d7c70246160fb6d3c210b6c831c/cmd/git-init/main.go#L57) a few git commands without retries. Containers defined in steps, either added automatically or by users will wait for previous steps to complete by [watching](https://github.com/tektoncd/pipeline/blob/9428c431bcef5d7c70246160fb6d3c210b6c831c/pkg/entrypoint/entrypointer.go#L64) for a `wait_file` indicating the [completion](https://github.com/tektoncd/pipeline/blob/9428c431bcef5d7c70246160fb6d3c210b6c831c/pkg/reconciler/v1alpha1/taskrun/entrypoint/entrypoint.go#L132) of the previous step.
2. No. The [TaskRun controller](https://github.com/tektoncd/pipeline/blob/9428c431bcef5d7c70246160fb6d3c210b6c831c/pkg/reconciler/v1alpha1/taskrun/taskrun.go#L317) watches the pod status and waits for it to exit. 

[...]

PR #2158, opened 2020-03-04
Optimize start time for TaskRuns with no sidecars

### LINKED ISSUE #2080: Set Ready annotation during pod creation for Taskruns with no Sidecars
## Context

Jason has a really nice description of this in the Problem section of #1569. But in short, today a step in a Taskrun waits to start until it sees a "READY" value that is populated by the Taskrun controller using the Downward API. The controller waits for all sidecars to be up an running setting the annotation.

### Suggestion 

For taskruns without sidecars, the controller should be able to set the ready value when creating the pod. 

This is a performance optimization and should help the taskrun start running faster and the taskrun controller have less overall load.

dibyom:
# Changes

When no sidercars are present **and** the cluster does not use
injected sidecars, the pipeline controller can optimize the
TaskRun Pod creation process by setting the `tekton.dev/ready`
annotation before pod creation itself instead of setting it after
the pod has been created.

This commit adds an option to a new `config-features` ConfigMap
called `running-in-environment-with-injected-sidecars`. Enabling
this option will decrease the time it takes for a TaskRun to start
running(when no sidecars are present). However, for clusters that
use injected sidecars e.g. istio enabling this option can lead to
unexpected behavior.By default, the option is set to "true" for
backwards compatibility.

Fixes #2080

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Users can optimize the start time for TaskRuns without sidecars 
running in clusters that do not use injected sidecars (e.g. istio) 
by enabling the following option in the `feature-flags` ConfigMap:
`running-in-environment-with-injected-sidecars`: "false"
```

### DIFF
--- a/config/config-feature-flags.yaml
+++ b/config/config-feature-flags.yaml
@@ -41,3 +41,11 @@ data:
   # See https://github.com/tektoncd/pipeline/issues/1836 for more
   # info.
   disable-working-directory-overwrite: "false"
+  # This option should be set to false when Pipelines is running in a
+  # cluster that does not use injected sidecars such as Istio. Setting
+  # it to false should decrease the time it takes for a TaskRun to start
+  # running. For clusters that use injected sidecars, setting this
+  # option to false can lead to unexpected behavior.
+  #
+  # See https://github.com/tektoncd/pipeline/issues/2080 for more info.
+  running-in-environment-with-injected-sidecars: "true"
--- a/docs/install.md
+++ b/docs/install.md
@@ -278,6 +278,12 @@ The default value is `false`, which causes Tekton to override the working direct
 for each `Step` that does not have its working directory explicitly set with `/workspace`.
 For more information, see the [associated issue](https://github.com/tektoncd/pipeline/issues/1836).
 
+- `running-in-environment-with-injected-sidecars`: set this flag to `"true"` to allow the
+Tekton controller to set the `tekon.dev/ready` annotation at pod creation time for 
+TaskRuns with no Sidecars specified. Enabling this option should decrease the time it takes for a TaskRun to
+start running. However, for clusters that use injected sidecars e.g. istio
+enabling this option can lead to unexpected behavior.
+
 For example:
 
 ```yaml
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -45,8 +45,11 @@ const (
 	// ResultsDir is the folder used by default to create the results file
 	ResultsDir = "/tekton/results"
 
-	featureFlagDisableHomeEnvKey    = "disable-home-env-overwrite"
-	featureFlagDisableWorkingDirKey = "disable-working-directory-overwrite"
+	featureInjectedSidecar                   = "running-in-environment-with-injected-sidecars"
+	featureFlagDisableHomeEnvKey             = "disable-home-env-overwrite"
+	featureFlagDisableWorkingDirKey          = "disable-working-directory-overwrite"
+	featureFlagConfigMapName                 = "feature-flags"
+	featureFlagSetReadyAnnotationOnPodCreate = "enable-ready-annotation-on-pod-create"
 
 	taskRunLabelKey = pipeline.GroupName + pipeline.TaskRunLabelKey
 )
@@ -238,6 +241,10 @@ func MakePod(images pipeline.Images, taskRun *v1beta1.TaskRun, taskSpec v1beta1.
 	podAnnotations := taskRun.Annotations
 	podAnnotations[ReleaseAnnotation] = ReleaseAnnotationValue
 
+	if shouldAddReadyAnnotationOnPodCreate(taskSpec.Sidecars, kubeclient) {
+		podAnnotations[readyAnnotation] = readyAnnotationValue
+	}
+
 	return &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			// We execute the build's pod in the same namespace as where the build was
@@ -350,4 +357,22 @@ func shouldOverrideWorkingDir(kubeclient kubernetes.Interface) bool {
 		return false
 	}
 	return true
+}
+
+// shouldAddReadyAnnotationonPodCreate returns a bool indicating whether the
+// controller should add the `Ready` annotation when creating the Pod. We cannot
+// add the annotation if Tekton is running in a cluster with injected sidecars
+// or if the Task specifies any sidecars.
+func shouldAddReadyAnnotationOnPodCreate(sidecars []v1beta1.Sidecar, kubeclient kubernetes.Interface) bool {
+	// If the TaskRun has sidecars, we cannot set the READY annotation early
+	if len(sidecars) > 0 {
+		return false
+	}
+	// If the TaskRun has no sidecars, check if we are running in a cluster where sidecars can be injected by other
+	// controllers.
+	configMap, err := kubeclient.CoreV1().ConfigMaps(system.GetNamespace()).Get(GetFeatureFlagsConfigName(), metav1.GetOptions{})
+	if err == nil && configMap != nil && configMap.Data != nil && configMap.Data[featureInjectedSidecar] == "false" {
+		return true
+	}
+	return false
 }
--- a/pkg/pod/pod_test.go
+++ b/pkg/pod/pod_test.go
@@ -24,7 +24,9 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/pipeline/pkg/system"
 	"github.com/tektoncd/pipeline/test/diff"
@@ -41,11 +43,13 @@ var (
 		CredsImage:      "override-with-creds:latest",
 		ShellImage:      "busybox",
 	}
+
+	ignoreReleaseAnnotation = func(k string, v string) bool {
+		return k == ReleaseAnnotation
+	}
 )
 
 func TestMakePod(t *testing.T) {
-	names.TestingSeed()
-
 	implicitEnvVars := []corev1.EnvVar{{
 		Name:  "HOME",
 		Value: pipeline.HomeDir,
@@ -58,14 +62,12 @@ func TestMakePod(t *testing.T) {
 		Name:         "tekton-internal-secret-volume-multi-creds-9l9zj",
 		VolumeSource: corev1.VolumeSource{Secret: &corev1.SecretVolumeSource{SecretName: "multi-creds"}},
 	}
-
 	placeToolsInit := corev1.Container{
 		Name:         "place-tools",
 		Image:        images.EntrypointImage,
 		Command:      []string{"cp", "/ko-app/entrypoint", "/tekton/tools/entrypoint"},
 		VolumeMounts: []corev1.VolumeMount{toolsMount},
 	}
-
 	runtimeClassName := "gvisor"
 	automountServiceAccountToken := false
 	dnsPolicy := corev1.DNSNone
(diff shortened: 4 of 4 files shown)


docs/install.md
```diff
@@ -273,6 +273,12 @@ The default value is `false`, which causes Tekton to override the working direct
 for each `Step` that does not have its working directory explicitly set with `/workspace`.
 For more information, see the [associated issue](https://github.com/tektoncd/pipeline/issues/1836).
 
+- `enable-ready-annotation-on-pod-create`: set this flag to `"true"` to allowq the
```
sbwsg: nit: `allowq` -> `allow`

docs/install.md
```diff
@@ -273,6 +273,12 @@ The default value is `false`, which causes Tekton to override the working direct
...
+Tekton controller to set the `tekon.dev/ready` flag at pod creation time for
```
sbwsg: suggest "annotation" in place of "flag" here.

dibyom:
>  I intend to replace these configmap lookups with the configmap watcher but that will happen in a future PR

Yeah, I was just thinking about that -- feature falgs watching can probably be extracted out to its own package or something

pkg/pod/pod_test.go
```diff
@@ -780,6 +884,12 @@ script-heredoc-randomly-generated-78c5n
 			if d := cmp.Diff(c.want, &got.Spec, resourceQuantityCmp); d != "" {
 				t.Errorf("Diff(-want, +got):\n%s", d)
 			}
+
+			if c.wantAnnotations != nil {
```
dibyom: One thing I realized while adding the tests, is that we were never using the wantAnnotations filed at all (it was only specified for the sidecar tests). I added this block and started ignoring the releaseAnnotation that gets added by default.

docs/install.md
dibyom: done

docs/install.md
dibyom: fixed

config/config-feature-flags.yaml
```diff
@@ -35,6 +34,15 @@ data:
...
+
+  # Setting this flag to "true" will set the `tekton.dev/ready` annotation
+  # at pod creation time for Taskruns without sidecars.
+  #
+  # Enabling this option should decrease the time it takes for a TaskRun to
+  # start running. However, for clusters that use injected sidecars e.g. istio
+  # enabling this option can lead to unexpected behavior.
+  #
+  # See https://github.com/tektoncd/pipeline/issues/2080 for more info.
+  enable-ready-annotation-on-pod-create: "false"
```
bobcatfish: are we 100% sure we want to use a feature flag for this? i feel like we're quickly going to get to a point where most new features use feature flags - maybe this is what we want!

either way, I assume the idea is that in a future release we'd turn this to "true" by default, and maybe remove the flag? If so can we create an issue to track this and make sure we do it?

pkg/pod/pod_test.go
```diff
@@ -888,3 +998,82 @@ func TestShouldOverrideWorkingDir(t *testing.T) {
 		})
 	}
 }
+
+func TestShouldAddReadyAnnotationonPodCreate(t *testing.T) {
```
bobcatfish: 🎉

pkg/pod/pod.go
```diff
@@ -336,3 +342,22 @@ func shouldOverrideWorkingDir(kubeclient kubernetes.Interface) bool {
 	}
 	return true
 }
+
+// shouldAddReadyAnnotationonPodCreate returns a bool indicating whether the
+// controller should add the `Ready` annotation when creating the Pod. The
+// default behavior is to add the Ready annotation after  Pod has been created
+// and all of its sidecars are in a Ready state.
```
bobcatfish: totally minor point but I find when I find myself writing docstrings and tests for a private/unexported function, it's often an indication that it would make senes to move that function into a separate/new function and make it exported

pkg/pod/pod.go
```diff
@@ -336,3 +342,22 @@ func shouldOverrideWorkingDir(kubeclient kubernetes.Interface) bool {
...
+//
+// For further reference see https://github.com/tektoncd/pipeline/issues/2080
```
bobcatfish: i'm not a huge fan of referencing issues in docstrings and comments, unless it's a TODO. id rather duplicate the info here if we think it's important. Also if we link the issue from the commit message we dont need to mention it here too

pkg/pod/pod.go
```diff
@@ -224,6 +226,10 @@ func MakePod(images pipeline.Images, taskRun *v1alpha1.TaskRun, taskSpec v1alpha
 	podAnnotations := taskRun.Annotations
 	podAnnotations[ReleaseAnnotation] = ReleaseAnnotationValue
 
+	if shouldAddReadyAnnotationonPodCreate(taskSpec.Sidecars, kubeclient) {
+		podAnnotations[readyAnnotation] = readyAnnotationValue
```
bobcatfish: it feels a bit odd to me that we are using `readyAnnotation` directly here AND we're using it in `UpdateReady` in a totally different file - could it make sense to call `UpdateReady` here, or in some way refactor this a bit so we can share the code that accesses `readyAnnotation`?

pkg/pod/pod_test.go
```diff
@@ -40,41 +41,46 @@ var (
...
+		dnsPolicy                    = corev1.DNSNone
+		enableServiceLinks           = false
+		priorityClassName            = "system-cluster-critical"
+	)
 
 	for _, c := range []struct {
 		desc            string
 		trs             v1alpha1.TaskRunSpec
 		ts              v1alpha1.TaskSpec
+		featureFlags    map[string]string
```
bobcatfish: this might be a subject for a larger discussion but a couple thoughts here:
- the more feature flags we add, the larger the matrix of test cases we're going to need
- does it make sense for the `pod` package to know about feature flags, or could it be that we tell the `pod` package something more specific like a bool for "update status early if no sidecars"?

pkg/pod/pod_test.go
bobcatfish: thanks for catching this! might also be a sign that the test cases are a bit too complex

config/config-feature-flags.yaml
dibyom: The reason we need this in a feature-flag is due to us supporting injected sidecars like Istio. If the cluster uses injected sidecars, the controller can not set the Ready annotation early since the sidecar injector might inject a sidecar container that the controller does not know about during pod creation time.

If we decide that injected sidecars are something we do not support injected sidecars, then yes, we do not need to feature-flag this. Though if I recall correctly, the feature was initially designed this way because we explicitly wanted to support the injected sidecar use case (correct me if I'm wrong @sbwsg!)

pkg/pod/pod_test.go
dibyom: > the more feature flags we add, the larger the matrix of test cases we're going to need
Indeed though I can't think of a good way around adding more tests at the moment! 

> does it make sense for the pod package to know about feature flags, or could it be that we tell the pod package something more specific like a bool for "update status early if no sidecars"?

I think we should refactor the feature flags bit into its own package (#2363 is somewhat related)

pkg/pod/pod.go
dibyom: Agreed. I think pkg/pod/pod.go has a few of these feature flags functions now. We have #2363 open for making this better!

pkg/pod/pod.go
dibyom: removed and added a line about why we are doing this (i.e. the injected sidecar resource)

pkg/pod/pod.go
dibyom: It is a different file but in the same package though. I don't think we can call `UpdateReady` here because that is updating an already existing Pod while the usage here is adding to a Pod's definition before it is created.
One thing we could do is move this and UpdateReady to its own file in the same package...but that puts UpdateReady in a different file from all the other Sidecar functions in that file.

dibyom:
> I have lots of minor feedback, one more major thing: I'm not convinced we definitely need a feature flag for this, it's very hard to imagine why anyone would have been relying on the previous functionality, and feature flags make testing + maintenance more complex

Sorry for the delay on this. I hope the explanation in https://github.com/tektoncd/pipeline/pull/2158#discussion_r406933018 makes sense. Happy to discuss whether or not we should actually support injected sidecars. If not, we can get rid of this feature-flag!

config/config-feature-flags.yaml
bobcatfish: hm interesting - we definitely want to support injected sidecars!

My understanding of the term "feature flag" is that it's used to control rollouts of features that we eventually intend to make the default - does that match other ppl's understanding? If so, I think this would make sense as a command line config param maybe? or it's own config map? i.e. something separate from feature flags

Or I'm totally wrong and it's normal to use "feature flags" to refer to functionality you intend to retain over the long run.

>  If the cluster uses injected sidecars, the controller can not set the Ready annotation early since the sidecar injector might inject a sidecar container that the controller does not know about during pod creation time.

I'm a bit sad that we can't find a solution that handles both cases well. 🤔  Have we brainstormed any ideas around this? e.g. what if we did something like:
- set the annotation to ready immediately
- be notified if the pod is modified (i.e. an admission controller has injected a sidecar), set the annotation to not ready <-- is it possible to do this before the pod actually started running? i wonder how crazy it would be to have our own admission controller for this

Is it possible we could name the flag after after exactly what it's intended for? e.g. something like `running-in-environment-with-injected-sidecars`? I think it would even be reasonable to default it to false in that case eventually, i.e. apply this optimization by default

sbwsg:
@dibyom @bobcatfish How do we want to proceed here?  Seems like we're at an impasse on the fact that this is a feature flag?  Does it satisfy if we just rename it for now to `running-in-environment-with-injected-sidecars` and then decide later if we want to add another separate configmap for runtime-configurations-that-arent-feature-flags (or do something else entirely)?

If not, do we want to explore other concrete solutions to this problem as part of this PR?  Or, finally, do we want to close this PR for now and explore them out-of-band?

FWIW I've worked on teams that used the term "feature flags" interchangeably for both 1) limiting exposure to in-development features and 2) long-lived configuration options that give users control over an application's behaviour or layout. I don't feel super strongly either way about Tekton's notion of "feature flags" but I am wondering if we can make some decisions around where this PR is heading next.

/kind misc

dibyom:
@sbwsg Sorry, forgot about this PR. I'm ok with both renaming the flag. I don't feel too strongly about adding this in a features configMap vs a feature-flag one - I'm ok with adding a new one.
I think this is a useful feature until the sidecar KEP becomes a reality. I won't get to this today but I'll fix it up next Tuesday!

config/config-feature-flags.yaml
dibyom: > My understanding of the term "feature flag" is that it's used to control rollouts of features that we eventually intend to make the default - does that match other ppl's understanding? 

Will add a new `options` or `features` configMap for this.

> I'm a bit sad that we can't find a solution that handles both cases well. 🤔 Have we brainstormed any ideas around this? e.g. what if we did something like:

Some discussion in https://github.com/tektoncd/pipeline/issues/2080 and https://github.com/tektoncd/pipeline/issues/701. Short of the Sidecar KEP becoming a reality, I think this is the best we can do at this moment

>  <-- is it possible to do this before the pod actually started running?
 No way to guarantee that unfortunately.

> Is it possible we could name the flag after after exactly what it's intended for? e.g. something like running-in-environment-with-injected-sidecars? I think it would even be reasonable to default it to false in that case eventually, i.e. apply this optimization by default

I'll change the flag name for now. Maybe we can set the default to false in a later change.

dibyom:
@sbwsg @bobcatfish I've updated the PR (finally!)....I've both renamed the feature/option name (`running-in-environment-with-injected-sidecars`) as well as put it in a different "features" configMap. 
While updating the docs, however, I realized that having two configMaps to point users to in the `Customizing Pipeline Controller` behavior is confusing - you have to know both the option to configure as well as the right configMap (`feature-flags` vs `features`) it belongs in. One option is to revert to using a single configMap like before. The other, perhaps, is to maybe put this option in the [`config-defaults` configMap](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#customizing-the-pipelines-controller-behavior)
What do you think?

sbwsg:
> One option is to revert to using a single configMap like before. The other, perhaps, is to maybe put this option in the config-defaults configMap

I'm totally fine with either of these options, with a very very slight preference towards putting it in `feature-flags`. No harm no foul if it goes the other way though.

dibyom:
Ok, updated PR. Keeping the `feature-flags` configMap for now!

bobcatfish:
Let's do it!

/lgtm

/hold cancel

PR #2180, opened 2020-03-06
Creds-init writes to fixed location when HOME override is disabled

### LINKED ISSUE #2165: TaskRun fails during initialization when disable-home-env-overwrite=true
This is closely related to the on-going Tekton `$HOME` issue (https://github.com/tektoncd/pipeline/issues/2013#issuecomment-585908031). I am testing `disable-home-env-overwrite` before it gets flipped.

This [comment](https://github.com/tektoncd/pipeline/issues/2013#issuecomment-588423688) says

> With this new flag Tekton will no longer interfere with HOME - it will be whatever you expect it to be when the container runs in a Pod.
>
> Previously $HOME would have been set to /tekton/home but now it won't be. So I would expect `$HOME/.docker/config.json` to be written to `/root/.docker/config.json` if the user is root and the image doesn't specify its own HOME.

I don't think this is the case. I am testing `gcr.io/cloud-builders/gradle`, but Tekton fails as it tries to create a directory `/.docker`.

```json
"level":"fatal",
"ts":1583431818.4164164,
"caller":"creds-init/main.go:41",
"msg":"Error initializing credentials: mkdir /.docker: permission denied",
"stacktrace":
main.main
    github.com/tektoncd/pipeline/cmd/creds-init/main.go:41
runtime.main
    runtime/proc.go:203
```

Note the "permission denied" error is not the issue here. The issue is that it is `/.docker` instead of `/root/.docker`.

sbwsg:
# Changes

Fixes #2165
Relates to https://github.com/tektoncd/pipeline/issues/2013

When the disable-home-env-overwrite flag is set to "true" each Step in a Task can conceivably have its own HOME directory. The concept of "HOME" is further muddied in systems that randomize the UID of containers.

When the disanle-home-env-overwite flag is set to "true", creds-init will write its credentials to `/tekton/creds` instead of `/tekton/home`.  Then the entrypoint binary will copy the credentials from `/tekton/creds` into `$HOME` before the Task's Steps are allowed to run.  This change should be completely transparent to the user.

For those users who were for some very specific reason actually relying on the credentials being in `/tekton/home` we'll now expose a new variable substitution: `$(credentials.path)` will get replaced with the directory that creds-init writes to - either `/tekton/home` or `/tekton/creds` depending on the state of the disable-home-env-overwrite flag.

This is a stop-gap measure as we work towards beta. It seems that we will probably need to revisit our recommendations around auth and the contract we're making with creds-init.

- [x] Variable replacement for `$(credentials.path)`

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Users who disable Tekton's HOME env overwrite will now find that git and docker credentials are written to a volumeMount, /tekton/creds, before they're copied into $HOME by the entrypoint. A new variable replacement has been introduced, $(credentials.path), which expands to the directory where creds-init writes its credentials, either /tekton/home or /tekton/creds.
```

### DIFF
--- a/cmd/entrypoint/main.go
+++ b/cmd/entrypoint/main.go
@@ -25,6 +25,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/tektoncd/pipeline/pkg/credentials"
 	"github.com/tektoncd/pipeline/pkg/entrypoint"
 )
 
@@ -53,6 +54,13 @@ func main() {
 		PostWriter:      &realPostWriter{},
 		Results:         strings.Split(*results, ","),
 	}
+
+	// Copy any creds injected by creds-init into the $HOME directory of the current
+	// user so that they're discoverable by git / ssh.
+	if err := credentials.CopyCredsToHome(credentials.CredsInitCredentials); err != nil {
+		log.Printf("non-fatal error copying credentials: %q", err)
+	}
+
 	if err := e.Go(); err != nil {
 		switch t := err.(type) {
 		case skipError:
--- a/pkg/reconciler/taskrun/resources/apply.go
+++ b/pkg/reconciler/taskrun/resources/apply.go
@@ -125,6 +125,15 @@ func ApplyTaskResults(spec *v1alpha1.TaskSpec) *v1alpha1.TaskSpec {
 	return ApplyReplacements(spec, stringReplacements, map[string][]string{})
 }
 
+// ApplyCredentialsPath applies a substitution of the key $(credentials.path) with the path that the creds-init
+// helper will write its credentials to.
+func ApplyCredentialsPath(spec *v1alpha1.TaskSpec, path string) *v1alpha1.TaskSpec {
+	stringReplacements := map[string]string{
+		"credentials.path": path,
+	}
+	return ApplyReplacements(spec, stringReplacements, map[string][]string{})
+}
+
 // ApplyReplacements replaces placeholders for declared parameters with the specified replacements.
 func ApplyReplacements(spec *v1alpha1.TaskSpec, stringReplacements map[string]string, arrayReplacements map[string][]string) *v1alpha1.TaskSpec {
 	spec = spec.DeepCopy()
--- a/pkg/apis/pipeline/v1alpha1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1alpha1/task_validation_test.go
@@ -256,6 +256,15 @@ func TestTaskSpecValidate(t *testing.T) {
 			}}},
 		},
 	}, {
+		name: "valid creds-init path variable",
+		fields: fields{
+			Steps: []v1alpha1.Step{{Container: corev1.Container{
+				Name:  "mystep",
+				Image: "echo",
+				Args:  []string{"$(credentials.path)"},
+			}}},
+		},
+	}, {
 		name: "step template included in validation",
 		fields: fields{
 			Steps: []v1alpha1.Step{{Container: corev1.Container{
--- a/pkg/apis/pipeline/v1alpha1/taskrun_validation_test.go
+++ b/pkg/apis/pipeline/v1alpha1/taskrun_validation_test.go
@@ -191,6 +191,19 @@ func TestTaskRunSpec_Validate(t *testing.T) {
 				}}},
 			}},
 		},
+	}, {
+		name: "task spec with credentials.path variable",
+		spec: v1alpha1.TaskRunSpec{
+			TaskSpec: &v1alpha1.TaskSpec{TaskSpec: v1beta1.TaskSpec{
+				Steps: []v1alpha1.Step{{
+					Container: corev1.Container{
+						Name:  "mystep",
+						Image: "myimage",
+					},
+					Script: `echo "creds-init writes to $(credentials.path)"`,
+				}},
+			}},
+		},
 	}}
 	for _, ts := range tests {
 		t.Run(ts.name, func(t *testing.T) {
--- a/pkg/apis/pipeline/v1beta1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation_test.go
@@ -161,6 +161,15 @@ func TestTaskSpecValidate(t *testing.T) {
 			}}},
 		},
 	}, {
+		name: "valid creds-init path variable",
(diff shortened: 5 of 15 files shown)


sbwsg:
/hold

Want to look at this again with fresh eyes on Monday before committing.

docs/auth.md
```diff
@@ -24,6 +24,12 @@ To solve this, before any `PipelineResources` are retrieved, all `pods` execute
 a credential initialization process that accesses each of its secrets and
 aggregates them into their respective files in `$HOME`.
 
+Please note: If the `disable-home-env-overwrite` feature flag has been set to
```
vdemeester: So, in the case of `disable-home-env-overwrite` (and by default in the future), do we decide that `/tekton/home` is where we write anything that "would be in `$HOME`" ? (like `.gitconfig`, …) I am +:100: on this as we will document it, and we are sure it's always written at the same place.

sbwsg: Interesting idea!  I think this might make sense.  I was considering using a more explicit name like `/tekton/creds` since we'd originally planned for /tekton/home to be removed in future.  But if we see this as reusable then I think it makes sense to leave it as-is.

I'm going to write up a one-page design doc and will present it to the WG on wednesday just to quickly get community feedback on that plan.

vdemeester: +:100: on both your comments :wink: I really like `/tekton/creds` and having a variable for it is :heart_eyes_cat:

sbwsg:
I've written up a very short design doc covering the solution in this PR and a slight alternative.  Doc here: https://docs.google.com/document/d/1SVuDt-SXPHymz41dveSXFSPrx5Z-Wzb9eHliJAyYg4o

sbwsg:
Note to self: This would probably break PipelineResources when the `disable-home-env-overwrite` flag is enabled (and eventually flipped to on by default in the following release).  Might need to update PipelineResources to account for this.

sbwsg:
Updated to write to /tekton/creds instead of /tekton/home.

Still looking at:
- Variable replacement for `$(credentials.path)`
- Fixing `PipelineResources` to work with the credentials path.
- Some kind of integration testing (though at the moment I might have to postpone this until after the flag gets flipped to be default true... not sure)

sbwsg:
I've added variable replacement from `$(credentials.path)` to `/tekton/home` or `/tekton/creds` depending on what the disable-home-env-overwrite flag is set to.

I've updated the `git` PipelineResource to deal with the possibility that creds are in /tekton/creds instead of /tekton/home.

sbwsg:
This change should be pretty much invisible to users now, no extra copying in their Steps should be necessary.  The entrypoint binary now copies the creds into $HOME instead of creds-init (when the disable-home-env-overwrite feature flag is true) so the net result is the same regardless of the feature flag: creds end up in container HOME dirs.

sbwsg:
/hold cancel

I think this is probably ready for review now.

docs/auth.md
sbwsg: 👍 awesome, i've implemented both of these now.

cmd/entrypoint/main.go
```diff
@@ -53,6 +54,13 @@ func main() {
 		PostWriter:      &realPostWriter{},
 		Results:         strings.Split(*results, ","),
 	}
+
+	// Copy any creds injected by creds-init into the $HOME directory of the current
+	// user so that they're discoverable by git / ssh.
+	if err := credentials.CopyCredsToHome(credentials.CredsInitCredentials); err != nil {
```
vdemeester: any reason why it is required here ? (in the `entrypoint` wrapper binary)

pkg/reconciler/taskrun/resources/apply.go
```diff
@@ -125,6 +125,15 @@ func ApplyTaskResults(spec *v1alpha1.TaskSpec) *v1alpha1.TaskSpec {
 	return ApplyReplacements(spec, stringReplacements, map[string][]string{})
 }
 
+// ApplyCredentialsPath applies a substitution of the key $(credentials.path) with the path that the creds-init
+// helper will write its credentials to.
+func ApplyCredentialsPath(spec *v1alpha1.TaskSpec, path string) *v1alpha1.TaskSpec {
```
vdemeester: Is this only usable as internal ? (because I think it's gonna be invalid if using directly in a spec)

cmd/entrypoint/main.go
sbwsg: `entrypoint` is the only tool we have that for-sure runs in the Step container, with access to (and knowledge of) the Step's HOME dir.

So we can only really use entrypoint to perform this copy I think.  And the nice benefit is that the behaviour now matches precisely before and after the feature flag: before the flag the credentials end up in $HOME (creds-init writes them there) and after the flag the credentials end up in $HOME (entrypoint writes them there).

pkg/reconciler/taskrun/resources/apply.go
sbwsg: Could you tell me more about it being invalid if used directly in spec?  It's intended for users to be able to work with as well.

cmd/entrypoint/main.go
vdemeester: ah very good point :+1: gotcha :wink:

pkg/reconciler/taskrun/resources/apply.go
vdemeester: I think this might fail at validation saying `credential.path` is an invalid variable (not 100% sure but my feeling is that not having anything about it in the validation code means it will fail — maybe I am wrong though, a tests showing this would be nice :stuck_out_tongue_closed_eyes: )

sbwsg: Aaaaah.  Lemme check.  Great catch.

sbwsg: Initial manual test proved positive - I was able to use `$(credential.path)` in a TaskRun w/ embedded TaskSpec.  I didn't hit a validation error and the Task ran successfully.

I'm going to add a test for this right now to make sure it doesn't fail validation in future.

sbwsg: Updated with tests for task and taskrun validations in both v1beta1 and v1alpha1 packages.

PR #2194, opened 2020-03-09
Handle the case of multiple versions of a status

### LINKED ISSUE #2188: Broken logic for pull request statuses in PR resource
# Expected Behavior
When a check is updated in the PR pipeline resource, the corresponding check should be updated on github side.

# Actual Behavior
The update may happen or not depending on the case.

# Additional Info
The code in https://github.com/tektoncd/pipeline/blob/8f427a84c7a975d3e5d9dea5b046dc583d8a5a33/pkg/pullrequest/api.go#L275-L278 makes the assumption that the `label`, or `context` in case of GitHiub, is a unique key to identify a "check".  That is not the case for GitHub - the API model maintains a list of checks, which may have the same `context`, but different `state` and `id`.

The reason why this issue happens is that in some cases the [code](https://github.com/tektoncd/pipeline/blob/8f427a84c7a975d3e5d9dea5b046dc583d8a5a33/pkg/pullrequest/api.go#L275-L278 ) in the pullrequest/api packages makes a map of the checks based on the `label`, so the `stated` stored in the map may be that of any of the status available.

Even if it was considered responsibility of the CI system to manage these multiple states, the pull request resource does not make that possible by hiding the check `id` away.

When a process in the CI system sends an update to a status, it should get back the ID from github. That ID can then be used by the CI system to update the correct status once the CI job is complete.  

Unfortunately the GitHub API docs don't really explain more in which cases a new status is created even if one exists already with the same name. It might be that that happens every time an update is sent without the ID. Or it may be that the API allows for multiple statues against the same job to allow for parallel checks. I would say the former is more likely, but I struggle to understand how we could end up then with 7 different IDs for the same `context` like in the [example](https://gist.github.com/4f846feed1a2a1fd610547d57f8ce6fc) from the gist.

afrittoli:
# Changes

The GitHub API keeps old versions of a status for historical
reasons. The statuses are returned in reverse chronological
order. This patch pulls-in a version of go-scm that handles
this behaviour and only returns the latest status.

Fixes #2188

# Submitter Checklist

review them:

## Reviewer Notes


afrittoli:
@tomgeorge

pkg/pullrequest/api.go
```diff
@@ -272,15 +272,22 @@ func (h *Handler) uploadStatuses(ctx context.Context, statuses []*scm.Status, sh
...
+	csMap := map[string]scm.Status{}
 	for _, s := range cs {
-		csMap[s.Label] = s.State
+		// Statuses are returned by the API in reverse chronological order:
+		// https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref
+		// We only consider the most recent status with a give label.
+		if _, ok := csMap[s.Label]; ok {
+			continue
+		}
+		//
```
wlynch: delete

pkg/pullrequest/api.go
```diff
@@ -272,15 +272,22 @@ func (h *Handler) uploadStatuses(ctx context.Context, statuses []*scm.Status, sh
 	}
 
 	// Index the statuses so we can avoid sending them if they already exist.
-	csMap := map[string]scm.State{}
+	csMap := map[string]scm.Status{}
 	for _, s := range cs {
-		csMap[s.Label] = s.State
+		// Statuses are returned by the API in reverse chronological order:
+		// https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref
+		// We only consider the most recent status with a give label.
```
wlynch:
```suggestion
		// We only consider the most recent status with a given label.
```

pkg/pullrequest/api.go
```diff
@@ -272,15 +272,22 @@ func (h *Handler) uploadStatuses(ctx context.Context, statuses []*scm.Status, sh
 	}
 
 	// Index the statuses so we can avoid sending them if they already exist.
-	csMap := map[string]scm.State{}
+	csMap := map[string]scm.Status{}
 	for _, s := range cs {
-		csMap[s.Label] = s.State
+		// Statuses are returned by the API in reverse chronological order:
+		// https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref
```
wlynch: We shouldn't tie semantics directly to the GitHub API (e.g. does this also hold true for GitLab?)

We may need another mechanism to determine "most recent".

pkg/pullrequest/api.go
```diff
@@ -272,15 +272,22 @@ func (h *Handler) uploadStatuses(ctx context.Context, statuses []*scm.Status, sh
...
+		}
+		//
+		csMap[s.Label] = *s
 	}
 
 	var merr error
 
 	for _, s := range statuses {
-		if csMap[s.Label] == s.State {
+		if csMap[s.Label].State == s.State && csMap[s.Label].Target == s.Target {
```
wlynch: IIUC, this is to allow updates of statuses event if their state is the same. Do we also want to compare the description for the same reason?

afrittoli: I care about the target because that changes between runs, but yeah, I will add description too

pkg/pullrequest/api.go
afrittoli: Good point. I will check what gitlab does here

afrittoli: The problem here is that the go-scm API does not expose any other mean of selecting the best version of a status:
https://github.com/jenkins-x/go-scm/blob/838e28f71c1eb2c8b4486e2b50bc498c8a5cfd0f/scm/repo.go#L91-L96

I will file an issue on go-scm side, I think the best solution would be for them to do the filtering in the GitHub driver. I would like to keep this code until the patches merges on go-scm side, with an updated comment.

afrittoli: https://github.com/jenkins-x/go-scm/issues/79

afrittoli: and https://github.com/jenkins-x/go-scm/pull/80

pkg/pullrequest/api.go
afrittoli: done

pkg/pullrequest/api.go
afrittoli: done

pkg/pullrequest/api.go
afrittoli: done

afrittoli:
The `go-scm` patch is merged, and we have a new release, https://github.com/jenkins-x/go-scm/releases/tag/v1.5.79, so the only change in this patch should be the update to `go.mod` now.

pkg/pullrequest/api.go
afrittoli: Done

afrittoli:
Oh, I need to sync the vendor folder

sbwsg:
@bobcatfish just reminded me that we've got the pullrequest Catalog Task now which is pinned to a specific version of the pullrequest-init binary.  It's going to be a bit painful until we transition the pullrequest Task away from the binary here but we should probably keep that pinned version up to date as well.  PR Task is only on the v1beta1-wip branch at the moment in the catalog ([see it here](https://github.com/tektoncd/catalog/tree/v1beta1-wip/pullrequest))

afrittoli:
> @bobcatfish just reminded me that we've got the pullrequest Catalog Task now which is pinned to a specific version of the pullrequest-init binary. It's going to be a bit painful until we transition the pullrequest Task away from the binary here but we should probably keep that pinned version up to date as well. PR Task is only on the v1beta1-wip branch at the moment in the catalog ([see it here](https://github.com/tektoncd/catalog/tree/v1beta1-wip/pullrequest))

Ack - once this is in the release I will update the catalog task.

afrittoli:
> @vdemeester: cat image

owww, cute

afrittoli:
> > @bobcatfish just reminded me that we've got the pullrequest Catalog Task now which is pinned to a specific version of the pullrequest-init binary. It's going to be a bit painful until we transition the pullrequest Task away from the binary here but we should probably keep that pinned version up to date as well. PR Task is only on the v1beta1-wip branch at the moment in the catalog ([see it here](https://github.com/tektoncd/catalog/tree/v1beta1-wip/pullrequest))
> 
> Ack - once this is in the release I will update the catalog task.

https://github.com/tektoncd/catalog/pull/218
@bobcatfish @sbwsg

PR #2216, opened 2020-03-11
config: prefixes image names with ko:// scheme

poy:
# Changes

Prefixes image names with ko:// scheme. This allows `ko` to improve the experience with Go modules. See https://github.com/google/ko/pull/58 for more information.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -42,7 +42,7 @@ spec:
       serviceAccountName: tekton-pipelines-controller
       containers:
       - name: tekton-pipelines-controller
-        image: github.com/tektoncd/pipeline/cmd/controller
+        image: ko://github.com/tektoncd/pipeline/cmd/controller
         args: [
           "-kubeconfig-writer-image", "github.com/tektoncd/pipeline/cmd/kubeconfigwriter",
           "-creds-image", "github.com/tektoncd/pipeline/cmd/creds-init",
--- a/config/webhook.yaml
+++ b/config/webhook.yaml
@@ -48,7 +48,7 @@ spec:
       - name: webhook
         # This is the Go import path for the binary that is containerized
         # and substituted here.
-        image: github.com/tektoncd/pipeline/cmd/webhook
+        image: ko://github.com/tektoncd/pipeline/cmd/webhook
         env:
         - name: SYSTEM_NAMESPACE
           valueFrom:


config/controller.yaml
```diff
@@ -42,7 +42,7 @@ spec:
       serviceAccountName: tekton-pipelines-controller
       containers:
       - name: tekton-pipelines-controller
-        image: github.com/tektoncd/pipeline/cmd/controller
+        image: ko://github.com/tektoncd/pipeline/cmd/controller
         args: [
           "-kubeconfig-writer-image", "github.com/tektoncd/pipeline/cmd/kubeconfigwriter",
```
vdemeester: Shouldn't thoses (all starting with `github.com/tektoncd/pipeline`) also be prefixed ?

ImJasonH:
/ok-to-test

The improvement is from making typos less difficult to debug.

vdemeester:
@ImJasonH could there be a `strict` way with this, as : in a "strict mode" you would get an error if a `ko://` reference is missing from what `ko` gathered to build the image for. Would it be possible to have that ? (because otherwise, it will fail silently, right ? and will have an ErrPull… on the pod(s))

ImJasonH:
If I'm understanding, yes that's exactly what ko does with this prefix. If it sees the special prefix in normal mode it strips it (for recent versions of ko in the last 6? months). If you pass `--strict` then it strips the ko:// prefix _and_ fails if the import path isn't recognized, instead of passing it through, which leads to unpullable image references.

This is step one to adopting ko strict more. After this we can run ko with --strict to avoid bad refs.

vdemeester:
> If I'm understanding, yes that's exactly what ko does with this prefix. If it sees the special prefix in normal mode it strips it (for recent versions of ko in the last 6? months). If you pass `--strict` then it strips the ko:// prefix _and_ fails if the import path isn't recognized, instead of passing it through, which leads to unpullable image references.
> 
> This is step one to adopting ko strict more. After this we can run ko with --strict to avoid bad refs.

Alright, if it works like this, it sounds good to me :+1: 
/lgtm

/cc @afrittoli @bobcatfish

vdemeester:
https://github.com/tektoncd/pipeline/pull/2216/files/f7442117ba4f54a884173bebb06ff542e7bc6ffe#diff-d56e306ce931c2991f18ef68b6b501a3 needs to be taken into account :angel:

vdemeester:
/hold cancel
I'll follow-up on this

Issue #2029, opened 2020-02-10
Broken "message" on last transition in case of failure


afrittoli:
# Expected Behavior
The message shall point to the first step that failed and how to get its logs.

# Actual Behavior
The message points to the last step that failed. 
Additionally the message to get the logs includes an extra "new line" that breaks the copy/paste.

When a `taskrun` has output resources, steps are appended to process those. If a step during the task fails, all the appended resource steps are marked as failed e.g.
```
    state:
      terminated:
        containerID: docker://35d721aed3233f6217a63f97661afdd8d344b33718744ab374992c19c6cf8487
        exitCode: 1
        finishedAt: "2020-02-10T13:16:57Z"
        message: |
          {"level":"info","ts":1581340399.0176625,"logger":"fallback-logger","caller":"logging/config.go:69","msg":"Fetch GitHub commit ID from kodata failed: open /var/run/ko/HEAD: no such file or directory"}
          2020/02/10 13:16:57 Skipping step because a previous step failed
        reason: Error
        startedAt: "2020-02-10T13:13:19Z"
``

# Steps to Reproduce the Problem

1. https://dashboard.dogfooding.tekton.dev/#/namespaces/default/pipelineruns/triggers-release-nightly-sk54v

# Additional Info
One option here is to fix the logic to get the first failed stepped instead of the last - however I think the current logic relies on the fact that only one step may be failed, which - at least for now - seems reasonable. I wonder if we should leave these steps added by resources as "unknown" instead of marking them as "failed" when they are skipped.

GregDritschler:
I have been looking into this issue.  There are two problems reported.
1) The PipelineRun "for logs run..." message does not point to the first failing step.
2) The message has an extra newline.

For the first problem, pkg/pod/status.go does sort the container status by finish time (State.Terminated.FinishedAt) to try to find the first failure.  However this field has a resolution of seconds.  In my experiments the steps that are skipped usually finish in the same second as the step that failed.  This defeats the intention of the sort.  By the way this problem can be observed without using output resources.   Any user steps that follow the failed step get the skipped message too.

My initial thought to fix this would be to store a higher-precision finish time in the pod's termination message and copy that to the termination status, imitating what happens with the start time but with a higher-precision timestamp.  One side effect is that the higher-precision timestamp would show up externally in the finishedAt field.  It might be desirable to change startedAt have the same precision so the formats look alike.

[...]

PR #2222, opened 2020-03-13
Panic in controller when step fails before image digest exporter

### LINKED ISSUE #2220: panic when submitting TaskRun
# Expected Behavior
TaskRun to complete successfully

# Actual Behavior

Controller panics

1. This happens when I create a TaskRun that references a ClusterTask.
2.
3.

# Additional Info

- Kubernetes version:

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:54Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.10-gke.17", GitCommit:"bdceba0734835c6cb1acbd1c447caf17d8613b44", GitTreeState:"clean", BuildDate:"2020-01-17T23:10:13Z", GoVersion:"go1.12.12b4", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

v0.10.1

The panic from the controller logs:

```
panic: runtime error: index out of range [5] with length 5

goroutine 258 [running]:
github.com/tektoncd/pipeline/pkg/pod.(*stepStateSorter).changeIndex(...)
	github.com/tektoncd/pipeline/pkg/pod/status.go:386
github.com/tektoncd/pipeline/pkg/pod.(*stepStateSorter).Swap(0xc000caad40, 0x2, 0x1)
	github.com/tektoncd/pipeline/pkg/pod/status.go:393 +0x514
sort.insertionSort(0x1b7c1a0, 0xc000caad40, 0x0, 0x5)
	sort/sort.go:28 +0x57
sort.quickSort(0x1b7c1a0, 0xc000caad40, 0x0, 0x5, 0x6)
	sort/sort.go:209 +0x201
sort.Sort(0x1b7c1a0, 0xc000caad40)
	sort/sort.go:218 +0x79
github.com/tektoncd/pipeline/pkg/pod.sortTaskRunStepOrder(0xc000fea240, 0x5, 0x8, 0xc000ab8000, 0x7, 0x8, 0x2, 0x0, 0x0)
	github.com/tektoncd/pipeline/pkg/pod/status.go:357 +0xb5
github.com/tektoncd/pipeline/pkg/pod.MakeTaskRunStatus(0x0, 0x0, 0x0, 0x0, 0xc000828630, 0x2c, 0x0, 0x0, 0xc0008286f0, 0x29, ...)
	github.com/tektoncd/pipeline/pkg/pod/status.go:177 +0x4f7
github.com/tektoncd/pipeline/pkg/reconciler/taskrun.(*Reconciler).reconcile(0xc000106750, 0x1b8e3a0, 0xc0001f01e0, 0xc000aadb80, 0xed5fc68ab, 0x27380c0)
	github.com/tektoncd/pipeline/pkg/reconciler/taskrun/taskrun.go:360 +0x115e
github.com/tektoncd/pipeline/pkg/reconciler/taskrun.(*Reconciler).Reconcile(0xc000106750, 0x1b8e3a0, 0xc0001f01e0, 0xc0009bfc80, 0x56, 0xc0000b9e00, 0x1b8e3a0)
	github.com/tektoncd/pipeline/pkg/reconciler/taskrun/taskrun.go:153 +0x841
knative.dev/pkg/controller.(*Impl).processNextWorkItem(0xc0002e4600, 0x0)
	knative.dev/pkg@v0.0.0-20191111150521-6d806b998379/controller/controller.go:335 +0x654
knative.dev/pkg/controller.(*Impl).Run.func1(0xc0009224f0, 0xc0002e4600)
	knative.dev/pkg@v0.0.0-20191111150521-6d806b998379/controller/controller.go:285 +0x53
created by knative.dev/pkg/controller.(*Impl).Run
	knative.dev/pkg@v0.0.0-20191111150521-6d806b998379/controller/controller.go:283 +0x1ac
```

I've managed to modify the `./pkg/pod/status_test.go` file (`TestSortTaskRunStepOrder`) to recreate the error:

```diff
diff --git a/pkg/pod/status_test.go b/pkg/pod/status_test.go
index 9215e116..3f96a41a 100644
--- a/pkg/pod/status_test.go
+++ b/pkg/pod/status_test.go
@@ -635,6 +635,10 @@ func TestSidecarsReady(t *testing.T) {
 func TestSortTaskRunStepOrder(t *testing.T) {
 	steps := []v1alpha1.Step{{Container: corev1.Container{
 		Name: "hello",
+	}}, {Container: corev1.Container{
+		Name: "extra-1",
+	}}, {Container: corev1.Container{
+		Name: "extra-2",
 	}}, {Container: corev1.Container{
 		Name: "exit",
 	}}, {Container: corev1.Container{
```

```
go test ./pkg/pod --run TestSortTaskRunStepOrder -v
```

sbwsg:
# Changes

Fixes #2220 

The image digest exporter (part of the Image Output Resource) is configured with `"terminationMessagePolicy": "FallbackToLogsOnError",`.

When a previous step has failed in a Task our entrypoint wrapping the exporter emits a log line like `2020/03/13 12:03:26 Skipping step because a previous step failed`. Because the image digest exporter is set to `FallbackToLogsOnError` Kubernetes slurps up this log line as the termination message.

That line gets read by the Tekton controller, which is looking for JSON in the termination message. It fails to parse and stops trying to read step statuses.

That in turn results in a mismatch in the length of the list of steps and the length of the list of step statuses. Finally we attempt to sort the list of step statuses alongside the list of steps. This method panics with an out of bounds error because it assumes the lengths of the two lists are the same.

So, this PR does the following things:

1. The image digest exporter has the `FallbackToLogsOnError` policy removed. I can't think of a reason that we need this anymore.
2. The Tekton controller no longer breaks out of the loop while it's parsing step statuses and instead simply ignores non-JSON termination messages.

# Submitter Checklist

review them:

### DIFF
--- a/pkg/pod/status.go
+++ b/pkg/pod/status.go
@@ -24,8 +24,8 @@ import (
 	"time"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
-	"github.com/tektoncd/pipeline/pkg/logging"
 	"github.com/tektoncd/pipeline/pkg/termination"
+	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/pkg/apis"
@@ -99,7 +99,7 @@ func SidecarsReady(podStatus corev1.PodStatus) bool {
 }
 
 // MakeTaskRunStatus returns a TaskRunStatus based on the Pod's status.
-func MakeTaskRunStatus(tr v1alpha1.TaskRun, pod *corev1.Pod, taskSpec v1alpha1.TaskSpec) v1alpha1.TaskRunStatus {
+func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1alpha1.TaskRun, pod *corev1.Pod, taskSpec v1alpha1.TaskSpec) v1alpha1.TaskRunStatus {
 	trs := &tr.Status
 	if trs.GetCondition(apis.ConditionSucceeded) == nil || trs.GetCondition(apis.ConditionSucceeded).Status == corev1.ConditionUnknown {
 		// If the taskRunStatus doesn't exist yet, it's because we just started running
@@ -114,39 +114,12 @@ func MakeTaskRunStatus(tr v1alpha1.TaskRun, pod *corev1.Pod, taskSpec v1alpha1.T
 	trs.PodName = pod.Name
 	trs.Steps = []v1alpha1.StepState{}
 	trs.Sidecars = []v1alpha1.SidecarState{}
-	logger, _ := logging.NewLogger("", "status")
-	defer func() {
-		_ = logger.Sync()
-	}()
 
 	for _, s := range pod.Status.ContainerStatuses {
 		if IsContainerStep(s.Name) {
 			if s.State.Terminated != nil && len(s.State.Terminated.Message) != 0 {
-				msg := s.State.Terminated.Message
-				r, err := termination.ParseMessage(msg)
-				if err != nil {
-					logger.Errorf("Could not parse json message %q because of %w", msg, err)
-					break
-				}
-				for index, result := range r {
-					if result.Key == "StartedAt" {
-						t, err := time.Parse(time.RFC3339, result.Value)
-						if err != nil {
-							logger.Errorf("Could not parse time: %q: %w", result.Value, err)
-							break
-						}
-						s.State.Terminated.StartedAt = metav1.NewTime(t)
-						// remove the entry for the starting time
-						r = append(r[:index], r[index+1:]...)
-						if len(r) == 0 {
-							s.State.Terminated.Message = ""
-						} else if bytes, err := json.Marshal(r); err != nil {
-							logger.Errorf("Error marshalling remaining results: %w", err)
-						} else {
-							s.State.Terminated.Message = string(bytes)
-						}
-						break
-					}
+				if err := updateStatusStartTime(&s); err != nil {
+					logger.Errorf("error setting the start time of step %q in taskrun %q: %w", s.Name, tr.Name, err)
 				}
 			}
 			trs.Steps = append(trs.Steps, v1alpha1.StepState{
@@ -178,6 +151,36 @@ func MakeTaskRunStatus(tr v1alpha1.TaskRun, pod *corev1.Pod, taskSpec v1alpha1.T
 	trs.Steps = sortTaskRunStepOrder(trs.Steps, taskSpec.Steps)
 
 	return *trs
+}
+
+// updateStatusStartTime searches for a result called "StartedAt" in the JSON-formatted termination message
+// of a step and sets the State.Terminated.StartedAt field to this time if it's found. The "StartedAt" result
+// is also removed from the list of results in the container status.
+func updateStatusStartTime(s *corev1.ContainerStatus) error {
+	r, err := termination.ParseMessage(s.State.Terminated.Message)
+	if err != nil {
+		return fmt.Errorf("termination message could not be parsed as JSON: %w", err)
+	}
+	for index, result := range r {
+		if result.Key == "StartedAt" {
+			t, err := time.Parse(time.RFC3339, result.Value)
+			if err != nil {
+				return fmt.Errorf("could not parse time value %q in StartedAt field: %w", result.Value, err)
+			}
+			s.State.Terminated.StartedAt = metav1.NewTime(t)
+			// remove the entry for the starting time
+			r = append(r[:index], r[index+1:]...)
+			if len(r) == 0 {
+				s.State.Terminated.Message = ""
+			} else if bytes, err := json.Marshal(r); err != nil {
+				return fmt.Errorf("error marshalling remaining results back into termination message: %w", err)
+			} else {
+				s.State.Terminated.Message = string(bytes)
+			}
+			break
+		}
+	}
+	return nil
 }
 
 func updateCompletedTaskRun(trs *v1alpha1.TaskRunStatus, pod *corev1.Pod) {
--- a/pkg/pod/status_test.go
+++ b/pkg/pod/status_test.go
@@ -22,6 +22,8 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"github.com/tektoncd/pipeline/pkg/logging"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/pkg/apis"
@@ -40,6 +42,7 @@ func TestMakeTaskRunStatus(t *testing.T) {
 	for _, c := range []struct {
 		desc      string
 		podStatus corev1.PodStatus
+		taskSpec  v1alpha1.TaskSpec
 		want      v1alpha1.TaskRunStatus
 	}{{
 		desc:      "empty",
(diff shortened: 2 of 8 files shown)


pkg/pod/status.go
```diff
@@ -125,8 +125,8 @@ func MakeTaskRunStatus(tr v1alpha1.TaskRun, pod *corev1.Pod, taskSpec v1alpha1.T
 				msg := s.State.Terminated.Message
 				r, err := termination.ParseMessage(msg)
 				if err != nil {
-					logger.Errorf("Could not parse json message %q because of %w", msg, err)
-					break
+					// step generated non-JSON termination message; ignore
+					continue
```
bobcatfish: this change seems a bit inconsistent with the other error handling in this function, which will still be logging and `break`ing, 2 thoughts:

1. are we sure we want to continue processing messages when this happens? altho we dont want to panic, maybe it's reasonably to assume that encountering a badly formatted message indicates the rest are bogus, or is it that we could have some steps with good termination messages, and some with bad?
2. maybe logging here still makes sense? if you want to track down what happened later?

bobcatfish: side note / scope creep: why are we creating a new logger in this function? if this is in the controller, can we pass in the (properly configured) logger instead? <-- probably the material for a separate issue

pkg/pod/status.go
```diff
@@ -353,59 +353,20 @@ func getWaitingMessage(pod *corev1.Pod) string {
...
+			if iStep.Name == specStep.Name {
+				return true
+			}
+			if jStep.Name == specStep.Name {
+				return false
+			}
+		}
+		return false
+	})
+	return taskRunSteps
```
bobcatfish: is this part of fixing the issue or is it a separate improvement? if the latter, maybe a separate PR makes sense?

pkg/pod/status.go
sbwsg: > this change seems a bit inconsistent with the other error handling in this function, which will still be logging and breaking

This is true, though the other `break`s in this function, lines 136 & 148, are inside another nested `for` loop.  That means when those lines are hit the outer `for` loop is allowed to continue on processing the rest of the steps.

> are we sure we want to continue processing messages when this happens? altho we dont want to panic, maybe it's reasonably to assume that encountering a badly formatted message indicates the rest are bogus, or is it that we could have some steps with good termination messages, and some with bad?

My feeling is that this issue is evidence of `some steps with good termination messages, and some with bad`. There are other ways that a Step might inadvertently break the termination message and I think we can afford to be a little more robust in that scenario. Off the top of my head one example where this could happen again would be that a process spawned inside the container could race with the entrypoint to write the termination log and clobber it. That would be some awful user error but it isn't inconceivable, and I think Tekton should be able to handle that gracefully without abandoning the rest of the steps.

> maybe logging here still makes sense? if you want to track down what happened later?

Yeah that's totally fair.  I'll add this back in and look at threading the Logger through from the caller.

pkg/pod/status.go
sbwsg: So, the panic originates here.  I have two thoughts related to this :-
1. we shouldn't really be leaving a sort func lying around that panics in the face of a length mismatch.
2. this sorting code felt very complicated when compared to its intended purpose. I spent a good amount of time just trying to puzzle out what it was trying to achieve when the out-of-bounds error occurred. I would guess that the number of steps in a task is always going to be relatively small and I'm not convinced that the performance optimization this func was implementing was really worthwhile (even if the number of steps went into the hundreds or thousands... I don't think this would have changed the runtime really dramatically).

I guess we could approach this a couple of different ways (that I can think of, happy to go a different way if there are more I don't mention):
- as done here, sort the statuses and push any that don't appear in taskRunSteps to the back of the list.  I should probably also add a comment describing this behaviour.
- check if len(taskRunSteps) and len(taskSpecSteps) are equal.  If not, return the list unsorted.  Leave the stepStateSorter in place. comment about the constraint.
- document the panic with a comment.
- leave alone for now and revisit in a separate PR.

I'm kinda happy with the change as it stands (though it needs a comment) but I also don't mind if you think we should tackle this in a separate PR or a totally different way.  WDYT?

pkg/pod/status.go
sbwsg: I've threaded the logger through from the controller and added the log line back in.

pkg/pod/status.go
```diff
@@ -114,7 +115,6 @@ func MakeTaskRunStatus(tr v1alpha1.TaskRun, pod *corev1.Pod, taskSpec v1alpha1.T
 	trs.PodName = pod.Name
 	trs.Steps = []v1alpha1.StepState{}
 	trs.Sidecars = []v1alpha1.SidecarState{}
-	logger, _ := logging.NewLogger("", "status")
 	defer func() {
 		_ = logger.Sync()
```
sbwsg: I have no idea if this needs to stick around or not now that we're passing in the logger from the controller.

pkg/pod/status.go
sbwsg: i've added a comment describing the new sort behaviour.  still happy to take this change out or modify if preferred.

pkg/pod/status.go
sbwsg: I've factored the inner for loop into its own function.  I think this makes the control flow a little clearer.  I've also added a test for the panic-inducing case.

afrittoli:
I wonder it https://github.com/tektoncd/pipeline/issues/2029 was a different manifestation of this issue. In that case the overall Task failure status was set to that of the skipped  image digest exporter step, which is incorrect since the failure was in a different step, and the digest exporter step should have been only skept - see the [full yaml](https://gist.github.com/afrittoli/25a4fb15cb0e52a28a2c3f8dbbf67878).

pkg/pod/status.go
vdemeester: Nit (and note): `sort.Slice` uses reflection (whereas `sort.Sort` doesn't, rely on the `interface` contract). This means, it's gonna be slower — might be a premature optimization thinking that though :stuck_out_tongue:

sbwsg: I've put the original sort func back.  Don't want to hold up this PR if that change is controversial.

pkg/pod/status.go
```diff
@@ -180,6 +153,36 @@ func MakeTaskRunStatus(tr v1alpha1.TaskRun, pod *corev1.Pod, taskSpec v1alpha1.T
 	return *trs
 }
 
+// updateStatusStartTime searches for a result called "StartedAt" in the JSON-formatted termination message
+// of a step and sets the State.Terminated.StartedAt field to this time if it's found. The "StartedAt" result
+// is also removed from the list of results in the container status.
+func updateStatusStartTime(s *corev1.ContainerStatus) error {
```
bobcatfish: i love that this is a separate function! i dont mind if you want to ignore this, but since its separate id expect to see unit tests for it, and also i find that once we have a private/unexported function that 1) can have meaningful tests and 2) has good reason to have a docstring, it often makes sense to move it into its own package and make it public/exported

pkg/pod/status_test.go
```diff
@@ -600,6 +603,94 @@ func TestMakeTaskRunStatus(t *testing.T) {
 				}},
 			},
 		},
+	}, {
+		desc: "non-json-termination-message-with-steps-afterwards-shouldnt-panic",
```
bobcatfish: 🎉

pkg/pod/status_test.go
```diff
@@ -624,7 +715,8 @@ func TestMakeTaskRunStatus(t *testing.T) {
 				},
 			}
 
-			got := MakeTaskRunStatus(tr, pod, v1alpha1.TaskSpec{})
+			logger, _ := logging.NewLogger("", "status")
```
bobcatfish: thanks for threading the logger through! its kind of a pain but i think it's right way to do it :D

pkg/pod/status.go
```diff
@@ -114,39 +114,12 @@ func MakeTaskRunStatus(tr v1alpha1.TaskRun, pod *corev1.Pod, taskSpec v1alpha1.T
...
-							s.State.Terminated.Message = ""
-						} else if bytes, err := json.Marshal(r); err != nil {
-							logger.Errorf("Error marshalling remaining results: %w", err)
-						} else {
-							s.State.Terminated.Message = string(bytes)
-						}
-						break
-					}
+				if err := updateStatusStartTime(&s); err != nil {
+					logger.Errorf("error setting the start time of step %q in taskrun %q: %w", s.Name, tr.Name, err)
```
bobcatfish: ah okay so we're just logging an error but continuing onward - i could see including a comment here about why, since it might look like an oversight if just reading the code

PR #2233, opened 2020-03-16
Fetch remote task refs from images in taskruns

sthaha:
This includes several smaller refactors that came to light while
incorporating the fetching into the taskruns.

1. The OCIResolver is edited to return a TaskInterface. This is so that
   OCIResolver.GetTask can be used 1:1 with the `resources.GetTask` and
   remote task fetching can be substituted for local task fetching.

2. Instead of using a Keychain provider, we instead directly instaniate
   the keychain we need. This is exactly how the entrypoint impl works as
   well. The reason for doing this is to avoid the "in-cluster" calls that
   the previous API had. We already are piping through a
   `kubernetes.Interface` so we can just reuse that for in-cluster access
   and to mock it out in testing.
3. We allow  `GetTask` to return either a Task or ClusterTask
4. The testbuilder is updated to have an option to set the `TypeMeta` on
   the resources so that they can be marshalled properly.
5. A new test helper is created to take a url for a local registry and
   to upload a task definition as an image and return the reference.
6. The TaskRun reconciler is updated to return the proper `GetTask`
   func. If the `taskRef` specifies an `image` field (which has been added
   to the API with docs), then a remote resolver is constructed and it's
   `GetTask` is returned. Test cases are added for both NamespacedTasks and
   ClusterTasks. Specifically, this introduces no changes to the
   `resources/taskspec.go` which calls `GetTask` and returns the task's
   metadata and spec.

Part of #1839 

review them:

## Release Notes

```
You can also reference a `Task` contained in an OCI-compliant `Image`
```

### DIFF
--- a/docs/taskruns.md
+++ b/docs/taskruns.md
@@ -83,7 +83,18 @@ spec:
     name: read-task
 ```
 
-Or you can embed the spec of the `Task` directly in the `TaskRun`:
+You can also reference a `Task` contained in an OCI-compliant `Image`.
+
+```yaml
+spec:
+  taskRef:
+    name: read-task
+    image: docker.com/my-repo/my-image
+```
+
+These images can be generated using [this](https://github.com/tektoncd/experimental/tree/master/oci) cli and uploaded to most registries. You can also specify a digest or tag on the image. If access is required to read the image, a [service account](#service-account) must be specified on the `TaskRun` with an `ImagePullSecret` to properly fetch the image.
+
+You can also embed the spec of the `Task` directly in the `TaskRun`:
 
 ```yaml
 spec:
--- a/pkg/apis/pipeline/v1beta1/task_types.go
+++ b/pkg/apis/pipeline/v1beta1/task_types.go
@@ -138,6 +138,9 @@ type TaskRef struct {
 	// API version of the referent
 	// +optional
 	APIVersion string `json:"apiVersion,omitempty"`
+	// Image reference that contains the task referenced.
+	// +optional
+	Image string `json:"image,omitempty"`
 }
 
 // Check that Pipeline may be validated and defaulted.
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -36,6 +36,7 @@ import (
 	"github.com/tektoncd/pipeline/pkg/reconciler"
 	"github.com/tektoncd/pipeline/pkg/reconciler/taskrun/resources"
 	"github.com/tektoncd/pipeline/pkg/reconciler/taskrun/resources/cloudevent"
+	"github.com/tektoncd/pipeline/pkg/remote"
 	"github.com/tektoncd/pipeline/pkg/termination"
 	"github.com/tektoncd/pipeline/pkg/workspace"
 	"go.uber.org/zap"
@@ -203,28 +204,40 @@ func (c *Reconciler) updateStatusLabelsAndAnnotations(tr, original *v1alpha1.Tas
 	return nil
 }
 
-func (c *Reconciler) getTaskFunc(tr *v1alpha1.TaskRun) (resources.GetTask, v1alpha1.TaskKind) {
-	var gtFunc resources.GetTask
-	kind := v1alpha1.NamespacedTaskKind
+func (c *Reconciler) getTaskFunc(tr *v1alpha1.TaskRun) (resources.GetTask, v1alpha1.TaskKind, error) {
+	if tr.Spec.TaskRef != nil && tr.Spec.TaskRef.Image != "" {
+		taskFunc, err := remote.ImageTaskResolver(c.KubeClientSet, tr.Spec.TaskRef.Image, tr.GetNamespace(), tr.Spec.ServiceAccountName)
+		// This is the only real error case. If we know we need a GetTask func that works on Images but we couldn't create
+		// one, the Task lookup will never suceed
+		if err != nil {
+			return nil, "", fmt.Errorf("Could not lookup remote task: %w", err)
+		}
+
+		kind := v1alpha1.NamespacedTaskKind
+		if tr.Spec.TaskRef.Kind == v1alpha1.ClusterTaskKind {
+			kind = v1alpha1.ClusterTaskKind
+		}
+
+		return taskFunc, kind, err
+	}
+
 	if tr.Spec.TaskRef != nil && tr.Spec.TaskRef.Kind == v1alpha1.ClusterTaskKind {
-		gtFunc = func(name string) (v1alpha1.TaskInterface, error) {
+		return func(name string) (v1alpha1.TaskInterface, error) {
 			t, err := c.PipelineClientSet.TektonV1alpha1().ClusterTasks().Get(name, metav1.GetOptions{})
 			if err != nil {
 				return nil, err
 			}
 			return t, nil
-		}
-		kind = v1alpha1.ClusterTaskKind
-	} else {
-		gtFunc = func(name string) (v1alpha1.TaskInterface, error) {
-			t, err := c.PipelineClientSet.TektonV1alpha1().Tasks(tr.Namespace).Get(name, metav1.GetOptions{})
-			if err != nil {
-				return nil, err
-			}
-			return t, nil
-		}
+		}, v1alpha1.ClusterTaskKind, nil
 	}
-	return gtFunc, kind
+
+	return func(name string) (v1alpha1.TaskInterface, error) {
+		t, err := c.PipelineClientSet.TektonV1alpha1().Tasks(tr.Namespace).Get(name, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
+		return t, nil
+	}, v1alpha1.NamespacedTaskKind, nil
 }
 
 func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun) error {
@@ -249,7 +262,18 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun) error 
 		return err
 	}
 
-	getTaskFunc, kind := c.getTaskFunc(tr)
+	getTaskFunc, kind, err := c.getTaskFunc(tr)
+	if err != nil {
+		c.Logger.Errorf("Failed to determine Task spec to use for taskrun %s: %v", tr.Name, err)
+		tr.Status.SetCondition(&apis.Condition{
+			Type:    apis.ConditionSucceeded,
+			Status:  corev1.ConditionFalse,
+			Reason:  podconvert.ReasonFailedResolution,
+			Message: err.Error(),
+		})
+		return nil
+	}
+
 	taskMeta, taskSpec, err := resources.GetTaskData(ctx, tr, getTaskFunc)
 	if err != nil {
 		if ce, ok := err.(*v1beta1.CannotConvertError); ok {
@@ -303,6 +327,7 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun) error 
 		inputs = tr.Spec.Resources.Inputs
 		outputs = tr.Spec.Resources.Outputs
 	}
+
 	rtr, err := resources.ResolveTaskResources(taskSpec, taskMeta.Name, kind, inputs, outputs, c.resourceLister.PipelineResources(tr.Namespace).Get)
 	if err != nil {
 		c.Logger.Errorf("Failed to resolve references for taskrun %s: %v", tr.Name, err)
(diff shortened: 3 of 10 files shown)


vdemeester:
Putting a hold just to make sure we take time to review this :angel: 

/kind design
/area api
/hold

pierretasci:
* Ping * to get some eyes on this.

pkg/apis/pipeline/v1beta1/task_types.go
```diff
@@ -138,6 +138,9 @@ type TaskRef struct {
 	// API version of the referent
 	// +optional
 	APIVersion string `json:"apiVersion,omitempty"`
+	// Image reference that contains the task referenced.
+	// +optional
+	Image string `json:"image,omitempty"`
```
bobcatfish: i dont see any validation added for this - we'll want to add some at least simple validation (i think https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1beta1/taskrun_validation.go is the place to) and  tests for it; e.g. we'll need to make sure that only one of spec, ref, oci image are specified

https://github.com/tektoncd/pipeline/blob/9e0e8a1a06166124fdf047febf92ebbd9ff58957/pkg/apis/pipeline/v1beta1/taskrun_validation.go#L45-L48

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -203,28 +204,40 @@ func (c *Reconciler) updateStatusLabelsAndAnnotations(tr, original *v1alpha1.Tas
 	return nil
 }
 
-func (c *Reconciler) getTaskFunc(tr *v1alpha1.TaskRun) (resources.GetTask, v1alpha1.TaskKind) {
-	var gtFunc resources.GetTask
-	kind := v1alpha1.NamespacedTaskKind
+func (c *Reconciler) getTaskFunc(tr *v1alpha1.TaskRun) (resources.GetTask, v1alpha1.TaskKind, error) {
```
bobcatfish: this is scope creep from me but something that could make this function way better would be to move it into another package and not having it use the reconciler. this would mean passing c.kubeclientset in directly, but it would also mean we could have unit tests for it directly which would be great! `GetResourceFromBinding` is a very (very) similar example:

https://github.com/tektoncd/pipeline/blob/2e7e68f2fba58f90955c369123b228ab5654190e/pkg/reconciler/taskrun/resources/taskresourceresolution.go#L78-L80

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -203,28 +204,40 @@ func (c *Reconciler) updateStatusLabelsAndAnnotations(tr, original *v1alpha1.Tas
...
+		// This is the only real error case. If we know we need a GetTask func that works on Images but we couldn't create
+		// one, the Task lookup will never suceed
+		if err != nil {
+			return nil, "", fmt.Errorf("Could not lookup remote task: %w", err)
+		}
+
+		kind := v1alpha1.NamespacedTaskKind
+		if tr.Spec.TaskRef.Kind == v1alpha1.ClusterTaskKind {
+			kind = v1alpha1.ClusterTaskKind
+		}
```
bobcatfish: total nit but i wonder if this logic could be moved out of this if statement and work for all 3 cases somehow

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -440,6 +445,24 @@ func TestReconcile_ExplicitDefaultSA(t *testing.T) {
...
+	// Create images using an example task.
+	simpleTaskRef, err := test.CreateTaskImage(u.Host, remoteOnlyTask)
+	if err != nil {
+		t.Errorf("unexpected error pushing task image %w", err)
+	}
+	clusterTaskRef, err := test.CreateTaskImage(u.Host, remoteOnlyClusterTask)
+	if err != nil {
+		t.Errorf("unexpected error pushing cluster task image %w", err)
+	}
+
```
bobcatfish: maybe we could wrap this up in a function? e.g. `setupRemoteTaskRegistry()`

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -440,6 +445,24 @@ func TestReconcile_ExplicitDefaultSA(t *testing.T) {
 }
 
 func TestReconcile(t *testing.T) {
+	// For some test cases, we need an image registry to push up an image with a task to reference.
+	s := httptest.NewServer(registry.New())
+	defer s.Close()
+	u, err := url.Parse(s.URL)
+	if err != nil {
+		t.Fatal(err)
```
bobcatfish: I'm not sure what it would mean for there to be an error here, what would cause that? maybe just a bit more info in the error message

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -888,7 +923,84 @@ func TestReconcile(t *testing.T) {
...
+						"--",
+					),
+					tb.WorkingDir(workspaceDir),
+					tb.EnvVar("HOME", "/tekton/home"),
+					tb.VolumeMount("tekton-internal-tools", "/tekton/tools"),
+					tb.VolumeMount("tekton-internal-downward", "/tekton/downward"),
+					tb.VolumeMount("tekton-internal-workspace", workspaceDir),
+					tb.VolumeMount("tekton-internal-home", "/tekton/home"),
+					tb.VolumeMount("tekton-internal-results", "/tekton/results"),
+					tb.TerminationMessagePath("/tekton/termination"),
```
bobcatfish: i cant quite wrap my head around why this existing test case changed also?

pkg/remote/resolver.go
```diff
@@ -17,22 +17,27 @@ limitations under the License.
...
-// TODO: Right now, there is only one resolver type. When more are added, this will need to be updated.
-func NewResolver(imageReference string, serviceAccountName string) Resolver {
-	return OCIResolver{
-		imageReference: imageReference,
-		keychainProvider: func() (authn.Keychain, error) {
-			return k8schain.NewInCluster(k8schain.Options{ServiceAccountName: serviceAccountName})
-		},
+// ImageTaskResolver will return a GetTask function capable of returning a Task from an OCI compliant image using the
+// image pull secrets of the included service account.
+func ImageTaskResolver(k8s kubernetes.Interface, imgRef, ns, serviceAccountName string) (resources.GetTask, error) {
```
bobcatfish: nice!!!

pkg/remote/resolver.go
```diff
@@ -17,22 +17,27 @@ limitations under the License.
...
+	kc, err := k8schain.New(k8s, k8schain.Options{
+		Namespace:          ns,
+		ServiceAccountName: serviceAccountName,
+	})
+	if err != nil {
+		return nil, err
```
bobcatfish: a bit more context on this error?

bobcatfish:
image

something is wrong with the coverage report again 😩

pkg/remote/oci.go
```diff
@@ -25,36 +25,67 @@ import (
...
-	_, _, err = scheme.Codecs.UniversalDeserializer().Decode(taskContents, nil, &task)
+	// Deserialize the contents into a valid task spec.
+	obj, kind, err := scheme.Codecs.UniversalDeserializer().Decode(taskContents, nil, nil)
 	if err != nil {
 		return nil, fmt.Errorf("Invalid remote task %s: %w", taskName, err)
 	}
 
-	return &task.Spec, nil
+	if t, ok := obj.(*v1alpha1.Task); ok {
+		return t, err
```
bobcatfish: there are a few places in here where a bare `err` is being returned, itd be great to update them to include some context about what caused the error

pkg/remote/oci.go
```diff
@@ -25,36 +25,67 @@ import (
...
+			},
+		}, nil
+	}
+
+	if t, ok := obj.(*v1alpha1.ClusterTask); ok {
+		return t, err
+	}
+
+	if t, ok := obj.(*v1beta1.ClusterTask); ok {
+		return &v1alpha1.ClusterTask{
```
bobcatfish: why are we converting everything to a v1alpha1 version? maybe a comment to explain

test/remote.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	remoteimg "github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+)
+
+// CreateTaskImage will push a new OCI image artifact with the provided task as a layer and return the full image
+// reference with a digest to fetch the image.
+func CreateTaskImage(registryHost string, task v1alpha1.TaskInterface) (string, error) {
```
bobcatfish: nice!!!

pkg/remote/oci.go
```diff
@@ -25,36 +25,67 @@ import (
...
+			TypeMeta:   t.TypeMeta,
+			ObjectMeta: t.TaskMetadata(),
+			Spec: v1alpha1.TaskSpec{
+				TaskSpec: t.TaskSpec(),
+			},
+		}, nil
+	}
```
bobcatfish: could we add some more unit test cases for the new code?

test/remote.go
```diff
@@ -0,0 +1,54 @@
...
+		return "", fmt.Errorf("could not push example image to registry")
+	}
+
+	digest, err := img.Digest()
+	if err != nil {
+		return "", fmt.Errorf("could not read image digest: %w", err)
+	}
+
+	return imgRef.Context().Digest(digest.String()).String(), nil
+}
```
bobcatfish: i could almost see having a unit test for this function itself

docs/taskruns.md
```diff
@@ -83,7 +83,18 @@ spec:
...
 ```
 
-Or you can embed the spec of the `Task` directly in the `TaskRun`:
+You can also reference a `Task` contained in an OCI-compliant `Image`.
+
+```yaml
+spec:
+  taskRef:
+    name: read-task
+    image: docker.com/my-repo/my-image
```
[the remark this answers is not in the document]
vdemeester: @bobcatfish as an image could (and would in the current experiments) be able to store more than one resource (aka several tasks, pipelines, …), `name` is required **as** it would be the way to get the right resource from the image. 

That said, we could also go the way of having only one resource per image — but I do like the idea to store multiples tasks in a single oci artifact :angel:

docs/taskruns.md
```diff
@@ -83,7 +83,18 @@ spec:
...
+```
+
+These images can be generated using [this](https://github.com/tektoncd/experimental/tree/master/oci) cli and uploaded to most registries. You can also specify a digest or tag on the image. If access is required to read the image, a [service account](#service-account) must be specified on the `TaskRun` with an `ImagePullSecret` to properly fetch the image.
+
```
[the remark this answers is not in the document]
vdemeester: > How to generate and upload the image using the CLI - https://github.com/tektoncd/experimental/tree/master/oci has no docs, we should either add docs there, or here, or both, including: how to use it, what the expected structure of the OCI image is, etc.

Yeah it needs some docs :angel: (it's on my todo if nobody picks it up before me)

pkg/remote/oci.go
vdemeester: I guess this is because those are not *passed* through the webhook. That said, we should use the `Convert*` function instead of converting it manually (mainly to re-use the code)

vdemeester:
>     * the docs recommend using the experimental CLI, we should see about making it not experimental ASAP

I don't think we need to rush this, I want to make sure the format of oci image we will use is well defined before marking the tool that allow to create tasks/pipelines in an OCI image as not experimental.

We need a well defined spec of what are tekton OCI image, what is store where, and how, etc… before being able to have this usable for the user (at least this is my thinking). **But** we could hide this being a feature-flag (aka supporting only if the feature-flag is there in the configmap) and mark it as experimental (and iterate over it).

bobcatfish:
> We need a well defined spec of what are tekton OCI image, what is store where, and how, etc… before being able to have this usable for the user (at least this is my thinking)

I agree! I think you're right @vdemeester , this is more important than whether the tool itself is experimental. If the .md file in this PR could be updated with more details on the expected structure that'd be a big step toward that, maybe it'll be some thing we can agree on quickly and we won't need to hide it behind a feature flag.

bobcatfish:
@pierretasci has expanded on the proposed format at https://docs.google.com/document/d/1lXF_SvLwl6OqqGy8JbpSXRj4hWJ6CSImlxlIl4V9rnM/edit

vdemeester:
See https://github.com/tektoncd/pipeline/pull/2395 for the current work

/close

PR #2260, opened 2020-03-21
Make defaults config map name customizable

eddycharly:
This PR makes the defaults config map name customizable.

It allows users to override the defaults config map name by specifying a `CONFIG_DEFAULTS_NAME` environment variable.

This is already the case for artifacts configurations :

https://github.com/tektoncd/pipeline/blob/fbde4aa579102b7664b6f7fb3d22919776e05efd/pkg/artifacts/artifacts_storage.go#L69-L83

If the environment variable is not defined, it uses the `config-defaults` name.

I would like to have this to be able to generate a unique config map name in a helm chart (see https://github.com/tektoncd/experimental/pull/494).

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -115,9 +115,9 @@ func NewConfigValidationController(ctx context.Context, cmw configmap.Watcher) *
 
 		// The configmaps to validate.
 		configmap.Constructors{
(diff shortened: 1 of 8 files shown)


eddycharly:
Strange, it doesn't work with webhook...

eddycharly:
But when I build webhook locally it works as expected....

PR #2277, opened 2020-03-24
Optimize TaskRun and PipelineRun controller reconciliation

jbarrick-mesosphere:
The Tekton controller will try running cleanup tasks for TaskRuns and
PipelineRuns everytime they are reconciled, which on controller startup
means that the controller spends around 200ms per TaskRun and
PipelineRun that it reconciles.

This adds a new status condition to the TaskRun and PipelineRun statuses
that indicates that the Tekton controller has completed all cleanup
tasks. If it is set, then we skip reconciliation of the objects. As the
objects are already in cache, the reconciler only takes around
200nanoseconds to get the object from cache and check if it needs to be
reconciled.

On my Kubernetes cluster with ~3500 TaskRuns and ~300 PipelineRuns, this drops controller start time (the time for the initial reconcile of all objects to complete) from 10-15 minutes to ~1 minute.

Signed-off-by: jbarrick@mesosphere.com <jbarrick@mesosphere.com>

# Changes

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
- Optimized reconcilation of TaskRun and PipelineRun objects to improve controller start time.
```

### DIFF
--- a/pkg/apis/pipeline/v1alpha1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1alpha1/pipelinerun_types.go
@@ -29,6 +29,8 @@ import (
 	"knative.dev/pkg/apis"
 )
 
+const ConditionCleanedUp apis.ConditionType = "CleanedUp"
+
 var (
 	groupVersionKind = schema.GroupVersionKind{
 		Group:   SchemeGroupVersion.Group,
@@ -157,6 +159,11 @@ func (pr *PipelineRun) GetTaskRunRef() corev1.ObjectReference {
 // GetOwnerReference gets the pipeline run as owner reference for any related objects
 func (pr *PipelineRun) GetOwnerReference() metav1.OwnerReference {
 	return *metav1.NewControllerRef(pr, groupVersionKind)
+}
+
+// IsCleanedUp returns true if the PipelineRun's status indicates that it is cleaned up.
+func (pr *PipelineRun) IsCleanedUp() bool {
+	return !pr.Status.GetCondition(ConditionCleanedUp).IsUnknown()
 }
 
 // IsDone returns true if the PipelineRun's status indicates that it is done.
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -133,6 +133,12 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 	// Snapshot original for the label/annotation check below.
 	original := pr.DeepCopy()
 
+	// If the PipelineRun has already been cleaned up, we return early to skip reconciliation.
+	if pr.IsCleanedUp() {
+		c.Logger.Info("Not reconciling cleaned up PipelineRun")
+		return nil
+	}
+
 	if !pr.HasStarted() {
 		pr.Status.InitializeConditions()
 		// In case node time was not synchronized, when controller has been scheduled to other nodes.
@@ -170,6 +176,11 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 			logger.Errorf("Failed to update TaskRun status for PipelineRun %s: %v", pr.Name, err)
 			return err
 		}
+		pr.Status.SetCondition(&apis.Condition{
+			Type:   v1beta1.ConditionCleanedUp,
+			Status: corev1.ConditionTrue,
+			Reason: "Cleaned Up",
+		})
 		go func(metrics *Recorder) {
 			err := metrics.DurationAndCount(pr)
 			if err != nil {
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -135,7 +135,7 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 
 	// If the PipelineRun has already been cleaned up, we return early to skip reconciliation.
 	if pr.IsCleanedUp() {
-		c.Logger.Info("Not reconciling cleaned up PipelineRun")
+		logger.Info("Not reconciling cleaned up PipelineRun")
 		return nil
 	}
 
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -107,6 +107,12 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, tr *v1beta1.TaskRun) pkg
 		events.Emit(recorder, nil, afterCondition, tr)
 	}
 
+	// If the TaskRun has already been cleaned up, return early to save reconciliation time.
+	if tr.IsCleanedUp() {
+		c.Logger.Info("Not reconciling cleaned up TaskRun")
+		return nil
+	}
+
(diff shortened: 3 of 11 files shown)


pkg/reconciler/taskrun/taskrun.go
```diff
@@ -111,6 +111,11 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 		tr.Status.StartTime = &tr.CreationTimestamp
 	}
 
+	if tr.IsCleanedUp() {
+		c.Logger.Info("Not reconciling cleaned up TaskRun")
```
sbwsg, line 115: Suggest making this a tad more explicit wrt the specific condition.  Something like:

```go
c.Logger.Infof("Skipping reconcile of TaskRun %q: condition %s is True", tr.Name, v1beta1.ConditionCleanedUp)
```

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -166,6 +172,11 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 			c.Logger.Errorf("Failed to update TaskRun status for PipelineRun %s: %v", pr.Name, err)
 			return err
 		}
+		pr.Status.SetCondition(&apis.Condition{
+			Type:   v1beta1.ConditionCleanedUp,
+			Status: corev1.ConditionTrue,
+			Reason: "Cleaned Up",
```
sbwsg, line 178: The `Reasons` in other packages are typically CamelCased with no spaces.  See for example `pkg/pod/status.go` with `CouldntGetTask`, `TaskRunResolutionFailed`, `TaskRunValidationFailed`, etc...  Suggest moving this into a named constant and matching format to existing reasons.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -139,6 +139,12 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 
 	// Don't modify the informer's copy.
 	pr := original.DeepCopy()
+
+	if pr.IsCleanedUp() {
+		c.Logger.Info("Not reconciling cleaned up PipelineRun")
```
sbwsg, line 144: Similarly to my comment on taskruns, I suggest including the PipelineRun's name and the name of the specific condition that is leading to the skip:

```go
c.Logger.Infof("Skipping reconcile of PipelineRun %q: condition %s is True", tr.Name, v1beta1.ConditionCleanedUp)
```

pkg/apis/pipeline/v1alpha1/pipelinerun_types.go
```diff
@@ -29,6 +29,8 @@ import (
 	"knative.dev/pkg/apis"
 )
 
+const ConditionCleanedUp apis.ConditionType = "CleanedUp"
```
afrittoli, line 32: NIT: `CleanedUp` does not resonate with me very much as a description for this condition. 

BTW I was working on doing something similar for the initialization steps, i.e. setting start time,  setting the initializing the `Succeeded` condition, the cloud events status. I wanted to store the specs in the status so that we only fetch / validate them once and then mark the initialization as successful.

So perhaps we could have three conditions:
- `PreRunSucceeded` (before steps for Task and before DAG for PipelineRuns)
- `Succeeded`
- `PostRunSucceeded` (after steps for Task and after DAG for PipelineRuns)

pkg/reconciler/pipelinerun/pipelinerun.go
afrittoli: IIUC this will happen in a few cases:
- the updates done during `isDone` (including setting this condition) will trigger the informer and get this key re-queued
- the periodic re-sync 
- a restart of the controller

The last two may take a while when there are a lot of items in etcd, so it's good to stop reconcile as early as possible :)

pkg/reconciler/taskrun/taskrun.go
afrittoli: I think this could move to right after the `DeepCopy` in L103

afrittoli: I think this could move to right after the `DeepCopy` in L103

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -155,7 +160,12 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 			}
 		}(c.metrics)
 
-		return merr.ErrorOrNil()
+		tr.Status.SetCondition(&apis.Condition{
+			Type:   v1beta1.ConditionCleanedUp,
+			Status: corev1.ConditionTrue,
+			Reason: "Cleaned Up",
+		})
+		return multierror.Append(merr, c.updateStatusLabelsAndAnnotations(tr, original)).ErrorOrNil()
```
afrittoli, line 168: I think we only need to invoke `updateStatusLabelsAndAnnotations` once per reconcile cycle, and we already do it in L126. It's fine to do it at the end, so as long as we capture and accumulate all possible errors between L126 and here we could remove the call on L126.

sbwsg:
@jbarrick-mesosphere just checking in here; do you want to close this PR for the time being and revisit later?  if you want to continue with this, could you rebase against latest master?

afrittoli:
@jbarrick-mesosphere this looks like a good feature to have, are you still pursuing it?

ImJasonH:
/hold

Before we make a user-visible API change, I think we should clarify in the PR description, and in user-facing docs, what exactly "cleaned up" means. Completed TaskRuns don't delete underlying Pods for instance (unless they're cancelled), so I'm not sure off the top of my head what would need to be cleaned up.

bobcatfish:
> This adds a new status condition to the TaskRun and PipelineRun statuses that indicates that the Tekton controller has completed all cleanup tasks.

This sounds like it could be a good candidate for a TEP! https://github.com/tektoncd/community/tree/master/teps

afrittoli:
> /hold
> 
> Before we make a user-visible API change, I think we should clarify in the PR description, and in user-facing docs, what exactly "cleaned up" means. Completed TaskRuns don't delete underlying Pods for instance (unless they're cancelled), so I'm not sure off the top of my head what would need to be cleaned up.

My interpretation is that once a taskrun/pipeline is complete, we still need to do a few things (update results, update metrics, stop timeout handler: https://github.com/tektoncd/pipeline/blob/e6c91d2135fee9199d9c2e2915fc8a957b734079/pkg/reconciler/pipelinerun/pipelinerun.go#L160-L186
All the methods invoked in there must be idempotent, and they will attempt to run on every reconcile, and do nothing from the 2nd time on. Even doing nothing takes time tough, and restarting the controller on a cluster with thousands of completed taskruns and pipelines means that it will take quite a few seconds for the reconcile to go through all and start reconciling new ones.

In that sense, having a dedicated condition would help further shorten the time, since it would be a single check for a condition.

> 
> edit: This kind of API change might be a good candidate for a [TEP](https://github.com/tektoncd/community/blob/master/teps/0001-tekton-enhancement-proposal-process.md) since it would give the community more dedicated time to discuss and iterate on the plan before sending code changes. This PR's been open long enough that the TEP process wasn't defined when it was opened, so I'm sorry if that's a moving goalpost. 😅

sbwsg:
Given that this PR hasn't moved since July and there's been some push-back on the original intent and idea I think we should close this for now. Interested parties should follow up with, at minimum, an issue clearly describing the problem and ideally a TEP describing the proposed solution for the community to discuss.

PR #2295, opened 2020-03-26
Correct missing field(s) error

### LINKED ISSUE #1844: missing fields `error` doesn't give much informations
# Expected Behavior

Add information about the missing fields

# Actual Behavior

No missing fields displayed..

```
Error from server (InternalError): error when creating "push-images-release.yaml": Internal error occurred: admission webhook "webhook.tekton.dev" denied the request: mutation failed: missing field(s):
```

Create a yaml task with missing fields and try to apply it with `kubectl`

/kind bug

### LINKED ISSUE #2235: It's weird that it says "missing field(s):" and no fields
# Expected Behavior
To see what fields are required of me (when omitting content from `spec`).

# Actual Behavior

```
$ oc apply -f no-task-pipeline.yaml
Error from server (BadRequest): error when creating "no-task-pipeline.yaml": admission webhook "webhook.tekton.dev" denied the request: mutation failed: missing field(s):
```

1. Apply an empty `spec` Pipeline to the OpenShift Pipelines Operator

# Additional Info

- Kubernetes version:

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T12:36:28Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.1", GitCommit:"f17a1cc", GitTreeState:"clean", BuildDate:"2020-03-15T08:08:33Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

```
$ tkn version
Client version: 0.8.0
Pipeline version: unknown
```

The file I used:
```
$ cat no-task-pipeline.yaml
kind: Pipeline
metadata:
  name: no-task-pipeline
spec:
```

Also tried it in the OpenShift Console UI (with & without `tasks:`)
image

GregDritschler:
Fixes #2235

Fixes #1844

# Changes

As I mentioned in the issue, it is debatable whether a completely empty pipeline spec is an error since all fields are optional.  On the other hand there is value in trying to catch mistakes.  It's unlikely anyone wants to define a pipeline with nothing in it.  So I changed the code to return the following error.

```
	if equality.Semantic.DeepEqual(ps, &PipelineSpec{}) {
		return apis.ErrGeneric("expected at least one, got none", "spec.description", "spec.params", "spec.resources", "spec.tasks", "spec.workspaces")
	}
```

field_error.go doesn't have any methods with canned text for this situation.  I found uses of ErrGeneric with the text "expected at least one, got none" in some places under vendor/knative.dev and decided that was good text for this case.  It produces the following error:

```
Error from server (BadRequest): error when creating "pipelinemissingspec1.yaml": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: expected at least one, got none: spec.description, spec.params, spec.resources, spec.tasks, spec.workspaces
```

I looked for other resource validation methods that use apis.ErrMissingField(apis.CurrentField).

* task_validation.go:
    Immediately after checking for an empty spec, the validation method checks if any steps are defined.  Therefore the empty spec check is superfluous and I removed it.
* pipelineresource_validation.go:
    I changed this one to specify the path that's required ("spec.type").
* condition_validation.go:
    I could not reproduce the ErrMissingField error with condition.  I left the code alone to be safe.
* workspace_validation.go:
    I could not reproduce the ErrMissingField error with workspace.  I left the code alone to be safe.

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/pkg/apis/pipeline/v1alpha1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1alpha1/task_validation_test.go
@@ -368,7 +368,7 @@ func TestTaskSpecValidateError(t *testing.T) {
 		name: "nil",
 		expectedError: apis.FieldError{
 			Message: `missing field(s)`,
-			Paths:   []string{""},
+			Paths:   []string{"steps"},
 		},
 	}, {
 		name: "no build",
--- a/pkg/apis/pipeline/v1alpha1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1alpha1/task_validation_test.go
@@ -365,7 +365,7 @@ func TestTaskSpecValidateError(t *testing.T) {
 		fields        fields
 		expectedError apis.FieldError
 	}{{
-		name: "nil",
+		name: "empty spec",
 		expectedError: apis.FieldError{
 			Message: `missing field(s)`,
 			Paths:   []string{"steps"},
--- a/pkg/apis/pipeline/v1beta1/task_validation.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation.go
@@ -25,7 +25,6 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/validate"
 	"github.com/tektoncd/pipeline/pkg/substitution"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"knative.dev/pkg/apis"
 )
@@ -40,9 +39,6 @@ func (t *Task) Validate(ctx context.Context) *apis.FieldError {
 }
 
 func (ts *TaskSpec) Validate(ctx context.Context) *apis.FieldError {
-	if equality.Semantic.DeepEqual(ts, &TaskSpec{}) {
-		return apis.ErrMissingField(apis.CurrentField)
-	}
 
 	if len(ts.Steps) == 0 {
 		return apis.ErrMissingField("steps")
--- a/pkg/apis/pipeline/v1beta1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation_test.go
@@ -285,7 +285,7 @@ func TestTaskSpecValidateError(t *testing.T) {
 		name: "nil",
 		expectedError: apis.FieldError{
 			Message: `missing field(s)`,
-			Paths:   []string{""},
+			Paths:   []string{"steps"},
 		},
 	}, {
 		name: "no step",
--- a/pkg/apis/pipeline/v1beta1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation_test.go
@@ -282,7 +282,7 @@ func TestTaskSpecValidateError(t *testing.T) {
 		fields        fields
 		expectedError apis.FieldError
 	}{{
-		name: "nil",
+		name: "empty spec",
 		expectedError: apis.FieldError{
 			Message: `missing field(s)`,
 			Paths:   []string{"steps"},
--- a/pkg/apis/pipeline/v1alpha1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1alpha1/pipeline_validation.go
@@ -154,7 +154,7 @@ func validateParamResults(tasks []PipelineTask) error {
 // of Tasks expressed in the Pipeline makes sense.
 func (ps *PipelineSpec) Validate(ctx context.Context) *apis.FieldError {
 	if equality.Semantic.DeepEqual(ps, &PipelineSpec{}) {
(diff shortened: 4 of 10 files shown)


vdemeester:
/cc @sbwsg 
Do we want this in 0.11 ? :angel:

sbwsg:
sure thing! added cherry pick label.

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -285,7 +285,7 @@ func TestTaskSpecValidateError(t *testing.T) {
 		name: "nil",
```
sbwsg: i know this isn't part of the diff but it's incorrectly named - the Task Spec does not end up being nil in this test.  It ends up being a valid pointer to an empty TaskSpec object.

It doesn't look like we exercise the nil case.

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -40,9 +39,6 @@ func (t *Task) Validate(ctx context.Context) *apis.FieldError {
 }
 
 func (ts *TaskSpec) Validate(ctx context.Context) *apis.FieldError {
-	if equality.Semantic.DeepEqual(ts, &TaskSpec{}) {
-		return apis.ErrMissingField(apis.CurrentField)
-	}
 
 	if len(ts.Steps) == 0 {
```
sbwsg: What will happen if ts is a nil pointer? Is there protection against that elsewhere in the codebase that will prevent it from happening?

pkg/apis/pipeline/v1beta1/task_validation_test.go
GregDritschler: I changed the name.

pkg/apis/pipeline/v1beta1/task_validation.go
GregDritschler: The way the code is now, the callers of the  validate methods appear to be responsible for checking for nil.  For example in taskrun_validation:

```
	// Validate TaskSpec if it's present
	if ts.TaskSpec != nil {
		if err := ts.TaskSpec.Validate(ctx); err != nil {
			return err
		}
	}
```
I haven't seen a nil check inside any of the validation methods I've looked at.

sbwsg: I'm not sure this is always the case though - see Task.Validate a few lines above this one:

```go
func (t *Task) Validate(ctx context.Context) *apis.FieldError {
	if err := validate.ObjectMetadata(t.GetObjectMeta()); err != nil {
		return err.ViaField("metadata")
	}
	return t.Spec.Validate(ctx)
}
```

I'm fine to remove the check in TaskSpec.Validate if we're absolutely sure that there won't be usages.  But if we're cherry-picking this into the 0.11 branch just before we cut the beta and we're not 100% sure then I'd prefer we postpone this change to another PR.  WDYT?

sbwsg: (Also, I believe - though I could be completely wrong! - that `equality.Semantic.DeepEqual` will actually consider a nil pointer and &TaskSpec{} as semantically equal, so this would return an error if there was a nil pointer.  We should have unit testing around this.)

pkg/apis/pipeline/v1alpha1/task_validation_test.go
```diff
@@ -365,10 +365,10 @@ func TestTaskSpecValidateError(t *testing.T) {
 		fields        fields
 		expectedError apis.FieldError
 	}{{
-		name: "nil",
+		name: "empty spec",
```
sbwsg: nice 🙏 thankyou!

pkg/apis/pipeline/v1beta1/task_validation.go
GregDritschler: No, the first thing I did was to check if the DeepEqual check was catching a nil.  I put the check back and forced ts to be nil.  It fell through the DeepEqual check and panicked on the ts.Steps reference.  So I don't believe I'm introducing an exposure.  It was there before.

GregDritschler: (fwiw I don't think this is fixing an urgent problem so it's fine to postpone merging it to post 0.11)

sbwsg: 👍 understood, thanks for testing the nil case and correcting my assumption here.

PR #2297, opened 2020-03-26
Updated taskspec with nodeselector to allow for different nodeselector for each task. 

NikeNano:
The goal with this PR is to allow for different task in a Pipeline to have different node selectors.

Currently it is only possible(correct my if I am wrong) to set the nodeslector on the podTemplate spec, which will result in that all task have the same node selector. This limites the node selection to one choice and also limits the workflow usage. 

My first contribution to Tekton, super happy for feedback and any suggestions. 

# Changes

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

### DIFF
--- a/pkg/apis/pipeline/v1alpha1/pipeline_types.go
+++ b/pkg/apis/pipeline/v1alpha1/pipeline_types.go
@@ -142,6 +142,10 @@ type PipelineTask struct {
 	// Refer Go's ParseDuration documentation for expected format: https://golang.org/pkg/time/#ParseDuration
 	// +optional
 	Timeout *metav1.Duration `json:"timeout,omitempty"`
+
+	// Set the NodeSelector on PipelineTask level to allow
+	// for different NodeSelectors for different steps.
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 }
 
 func (pt PipelineTask) HashKey() string {
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -587,6 +587,10 @@ func (c *Reconciler) createTaskRun(rprt *resources.ResolvedPipelineRunTask, pr *
 			PodTemplate:        pr.Spec.PodTemplate,
 		}}
 
+	if len(rprt.PipelineTask.NodeSelector) > 0 {
+		tr.Spec.PodTemplate.NodeSelector = rprt.PipelineTask.NodeSelector
+	}
+
 	if rprt.ResolvedTaskResources.TaskName != "" {
 		tr.Spec.TaskRef = &v1alpha1.TaskRef{
 			Name: rprt.ResolvedTaskResources.TaskName,


dibyom:
I think of the ideas with the PodTemplate is that it is part of the Run object types and not the Task/Pipeline. This makes the Task/Pipeline types reusable across clusters. The use case is interesting though so I'll put this on the Working Group agenda to see what others think of this

pkg/apis/pipeline/v1alpha1/pipeline_types.go
```diff
@@ -142,6 +142,10 @@ type PipelineTask struct {
 	// Refer Go's ParseDuration documentation for expected format: https://golang.org/pkg/time/#ParseDuration
 	// +optional
 	Timeout *metav1.Duration `json:"timeout,omitempty"`
+
+	// Set the NodeSelector on PipelineTask level to allow
+	// for different NodeSelectors for different steps.
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
 }
```
jlpettersson: I think there need to be a validation so that `NodeSelector` and `Workspace` is not used together. Tasks using Workspace need to be scheduled to where the e.g. PVC (volume) is bound.

NikeNano: Thx, I will check it out :). New to tekton so need to read up on workspaces!

NikeNano:
> I think of the ideas with the PodTemplate is that it is part of the Run object types and not the Task/Pipeline. This makes the Task/Pipeline types reusable across clusters. The use case is interesting though so I'll put this on the Working Group agenda to see what others think of this

Would it be better suited to move the functionality to the run objects using the names/tags  to set nodeselectors to specific tasks during run? This PR is related to working on a PR for kubeflow pipelines related to tekton https://github.com/kubeflow/kfp-tekton.

vdemeester:
> > I think of the ideas with the PodTemplate is that it is part of the Run object types and not the Task/Pipeline. This makes the Task/Pipeline types reusable across clusters. The use case is interesting though so I'll put this on the Working Group agenda to see what others think of this
> 
> Would it be better suited to move the functionality to the run objects using the names/tags to set nodeselectors to specific tasks during run? This PR is related to working on a PR for kubeflow pipelines related to tekton https://github.com/kubeflow/kfp-tekton.

@NikeNano You can already specficy `NodeSelectors` on the *Run object using `podTemplate` : see [`podTemplate`](https://github.com/tektoncd/pipeline/blob/master/docs/podtemplates.md)

vdemeester:
Ah I see, what you need is a "by pipelinetask" podtemplate, a little bit like what we have for `serviceAccount` and `serviceAccountNames` :thinking:

NikeNano:
> Ah I see, what you need is a "by pipelinetask" podtemplate, a little bit like what we have for `serviceAccount` and `serviceAccountNames` 🤔

Yeah, I would guess this is a use case that could be relevant for several uses cases outside the efforts of Kubeflow pipeline + Tekton.

vdemeester:
Actually this is a bit trickier than I initially thought. Putting a hold to discuss this more.

The *general* needs is : being able to override runtime info (`serviceAccountName`, `podTemplate`) for a task or a set of task in the Pipeline execution (`PipelineRun`).

Right now, we do this *only* for `serviceAccountName` with a `serviceAccounNames` field : 
```yaml
spec:
  serviceAccountName: sa-1
  serviceAccountNames:
    - taskName: build-task
      serviceAccountName: sa-for-build
```

Adding a `podTemplates` on the `PipelineRun` spec would make it weird, I feel we may want a better way to do this.

```yaml

spec:
  serviceAccountName: sa-1
  podTemplate: 
    securityContext:
      runAsNonRoot: true
    # […]
  taskOverride: # <- need a better name
  - name: build-task
    serviceAccountName: sa-for-build
    podTemplate:    
      securityContext:
        runAsNonRoot: true
      # […]
```

@bobcatfish @afrittoli @sbwsg @NikeNano wdyt ? 

I would really *love* to keep those k8s specific *and runtime* field in `PipelineRun` and in a `podTemplate` as much as possible — mainly to reduce our implicit dependency to k8s structs.

/hold

NikeNano:
I don't follow do you mean that the following: 

> ```yaml
> spec:
>   serviceAccountName: sa-1
>   podTemplate: 
>     securityContext:
>       runAsNonRoot: true
>     # […]
>   taskOverride: # <- need a better name
>   - name: build-task
>     serviceAccountName: sa-for-build
>     podTemplate:    
>       securityContext:
>         runAsNonRoot: true
>       # […]
> ```'

is good or bad? I am new to Tekton so don't have that much of an opinion, but the above looks good to me.  Have use argo a little bit and some times get the feeling that Tekton is more verbose and even simpler pipelines becomes quite verbose. 

I would be happy to contribute if changes are needed in the controller or similar to handle the changes!

NikeNano:
Any progress on this @dibyom ? :)

bobcatfish:
> I would really love to keep those k8s specific and runtime field in PipelineRun and in a podTemplate as much as possible — mainly to reduce our implicit dependency to k8s structs.

I completely agree @vdemeester 

Before we go ahead with anything @NikeNano could you describe in a bit more detail the use case where you need this? I think it will help us to figure out the best solution

NikeNano:
@bobcatfish I made this initial PR as parts of the efforts for [Kubeflow Pipelines ](https://github.com/kubeflow/pipelines)to support both argo and [Tekton as the underlaying infrastructure](https://github.com/kubeflow/kfp-tekton/issues/29) for execution. For Kubeflow Pipelines node selectors can have several use cases one of which is to allow for the usage of GPU resources [example](https://www.kubeflow.org/docs/gke/pipelines/enable-gpu-and-tpu/) from the docs. I guess some users also want to have the controller of which node which workload should go to and thus utalize node selector. For the use case with GPU:s not all steps of a pipeline usually have the possibility to utilise GPU resources and thus the current setup where one PodTemplateSpec is used is not sufficient.

bobcatfish:
Thanks for the explanation @NikeNano !

skaegi:
Sorry... a little late to the party here. We "kustomize" our Tasks and Pipelines to support things like serviceName, namespace, runtimeclass, add bonus annotations etc. to support these kinds of flows. Maybe one option to consider is to somehow provide first-class support for kustomize in our CRDs instead of adding our own override syntax.

NikeNano:
Could you point me to where in the code base this is done @skaegi? Would like to take a look to see how it is done :) To me it sounds like this "kustomize" is very similar, how would this be specified?

skaegi:
I'm embarrassed to say our agent is proprietary right now but kustomize is a technology built into kubectl and described here -- https://github.com/kubernetes-sigs/kustomize

NikeNano:
> I'm embarrassed to say our agent is proprietary right now but kustomize is a technology built into kubectl and described here -- https://github.com/kubernetes-sigs/kustomize

Ahh .... thanks for the link!

skaegi:
I opened #2362 to look specifically at adding per task override that @vdemeester first mentioned as I think it might be the cleanest approach. The kustomization we do in our agent is different in that we are also altering non-runtime definitions. It's interesting but I think more flexible than what is needed.

NikeNano:
Based upon your comment @skaegi, should we close this PR? I will start to work on #2362 in the mean time :)

vdemeester:
@NikeNano it would make sense yes :wink: 
/close

PR #2309, opened 2020-03-31
Update the CloudEvents version.

mattmoor:
cc @n3wscott @vdemeester @afrittoli


n3wscott:
@mattmoor  you have to also change how the return from send is handled. See the examples in knative. At least, the return check use to be `err != nil` check, but now the return is a result and you need to check: `cloudevents.IsACK(result)`

mattmoor:
cc @bobcatfish @ImJasonH @dlorenc

Issue #1843, opened 2020-01-10
git-init ssh symlink is failing


chmouel:
# Expected Behavior

Create the SSH file properly

# Actual Behavior

git-init step is failing when trying to symlink `/root/.ssh` to `/tekton/home/.ssh`

```
[git-source-tektoncd-pipeline-git-qlcdk] {"level":"warn","ts":1578650329.4586325,"logger":"fallback-logger","caller":"git/git.go:149","msg":"Unexpected error: creating symlink: symlink /tekton/home/.ssh /root/.ssh: file exists"}
```
# Steps to Reproduce the Problem

I am not so sure yet how to reproduce this but the image has definitively no /root/.ssh already and has no volume mount on /root/

# Additional Info

Just capturing it before I forget and will investigate a bit further and update this issue later on..

/kind bug

vincent-pli:
@chmouel 
The  https://github.com/tektoncd/pipeline/blob/3ae789c25831b3240721437d6570b58097cd55d9/pkg/git/git.go#L134
is invoked twice in 
https://github.com/tektoncd/pipeline/blob/3ae789c25831b3240721437d6570b58097cd55d9/pkg/git/git.go#L57
and 
https://github.com/tektoncd/pipeline/blob/3ae789c25831b3240721437d6570b58097cd55d9/pkg/git/git.go#L114

That's why we get `file exists` in second invoke.

ad22:
fyi, #2320 performs a refactor of the fetch+submodule fetch and combines it into a single call. Subsequently, this can be fixed w/o much change.

sbwsg:
This issue should be resolved now that we're only making the `ensureHomeEnv` call once in `git-init`. Closing as fixed but we can reopen if the problem rears its head again.

PR #2320, opened 2020-04-01
Introduce optional "refspec" to git PipelineResource, refactor

### LINKED ISSUE #2282: Git Resources are always on "master" branch.
# Expected Behavior
If I provide a branch or commit to a Git PipelineResource, running `git status` would indicate that I am not on the master branch.

# Actual Behavior
The branch is always master, even if the revision provided was not the master branch.

1. Create a task which takes in a git resource and a param which is a target branch to push any changes made by the task to.
2. Pass a Git resource to the Task created from `$branch` and also pass `$branch` to the param.
3. The TaskRun will fail because of branch mismatch.

I also ran a Task that executes `git rev-parse --abbrev-ref HEAD` against a non-master revision, which indicate that it was on the `master` branch.

# Additional Info

- Kubernetes version:

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:41:55Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
```

- Tekton Pipeline version:

Client version: 0.7.1

ad22:
# Changes
The git pipeline resource does not allow for any use case beyond a basic ref fetch, and is especially problematic for servers that don't support direct SHA fetches, which means that
the user cannot check out specific commit hashes not pointed to by a named ref. The user
is also not able to check out other refs alongside (example, the tag history). Current implementation also has some unexpected behavior, such as checking out/reset --hard 
on the master branch after git-init (see #2282), running git pull without depth when checkout fails and so on . The end result is that several users might end up creating their own custom containers for fetching and checking out what they need. This change makes the git resource much more extensible, via an optional refspec parameter.

Addresses #2282

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
This change should be fully backward compatible w.r.t to revision parameter
EXCEPT the ability to fetch git commit short hashes. This was only possible
because the original implementation did a full repository pull when git fetch 
for the short SHA (unsupported) failed

- Adds ability to checkout older commit hashes on a particular ref
  chain (when fetch via commit SHA is not enabled on the server)
- Adds ability to fetch several other refs (for example, refs/tags)
  alongside the main revision
- Provides resolution for #2282 - If the user needs to checkout a branch,
  they must specify the appropriate refspec and revision as the ref of the
  branch. There was no way to do this in the original implementation
- Fixes bug reported in #2282 by ensuring the revision is always checked
  out on a detached HEAD when specified w/o refspec
- Fixes bug - depth parameter is not passed to git submodule update, leading to
  unnecessary data transfer
- Fixes bug --recurse-submodules is appended to the fetch command even when
  the user specified submodule: "false", leading to unnecessary data transfer
```

This is my first time contributing to this repository, please let me know if I missed something. Thanks!


ad22:
@googlebot I signed it!

pkg/git/git.go
```diff
@@ -84,53 +86,92 @@ func Fetch(logger *zap.SugaredLogger, spec FetchSpec) error {
...
+	fetchParam := []string{spec.Revision}
+	checkoutParam := "FETCH_HEAD"
+
+	if spec.Refspec != "" {
+		// if refspec is specified, fetch the refspec and verify with provided revision
+		fetchParam = strings.Split(spec.Refspec, " ")
+		checkoutParam = spec.Revision
+	}
+
+	fetchArgs = append(fetchArgs, "origin", "--update-head-ok")
```
sbwsg, line 107: I didn't recognize this flag and looked it up in the docs, which say:

> By default git fetch refuses to update the head which corresponds to the current branch. This flag disables the check. This is purely for the internal use for git pull to communicate with git fetch, and unless you are implementing your own Porcelain you are not supposed to use it.

I'm not sure what Porcelain is in this context but it would be great to include a note here on what the flag does and why we need to use it given that it's mostly intended for internal-to-git usage.

test/git_checkout_test.go
```diff
@@ -53,7 +53,61 @@ func TestGitPipelineRun(t *testing.T) {
...
+		})
+	}
+}
+
+// Test revision fetching with refspec specified
+func TestGitPipelineRunWithRefspec(t *testing.T) {
+	t.Parallel()
+
+	// Pair of revision, refspec
+	var pairs = []string{
```
sbwsg, line 88: The approach we take elsewhere in our testing is to build a table of test cases with descriptions, inputs, and expected outputs (if any).

For example, I'd approach this test as follows:

```go
for _, tc := range []struct{
	description string
	revision string
	refspec string
}{{
	description: "Fetch refs/tags/v0.1.0 alongside master and checkout the master branch",
	revision: "master",
	refspec: "refs/tags/v0.1.0:refs/tags/v0.1.0 refs/heads/master:refs/heads/master",
}, {
	description: "Checkout specific revision from refs/pull/1009/head's commit chain",
	revision: ""968d5d37a61bfb85426c885dc1090c1cc4b33436"",
	refspec: "refs/pull/1009/head",
}, {
	// ...and so on...
}} {
	t.Run(tc.description, func(t *testing.T) {
		// test body
	})
}
```

The benefit is that this moves the comments into test names, which get logged on failure, and it adds a clean field name for each of the entries in your current pairs list.  It would also match more closely what we're doing elsewhere, which is nice from a pure code-style perspective.

test/v1alpha1/git_checkout_test.go
```diff
@@ -51,7 +51,61 @@ func TestGitPipelineRun(t *testing.T) {
...
+		})
+	}
+}
+
+// Test revision fetching with refspec specified
+func TestGitPipelineRunWithRefspec(t *testing.T) {
+	t.Parallel()
+
+	// Pair of revision, refspec
+	var pairs = []string{
```
sbwsg, line 86: Same comment here as in the other test file: it would be great to see this turned into a table with field names.

pkg/git/git.go
ad22: I've added a comment w/ explanation. Required because git init checks out the master branch by default.

test/git_checkout_test.go
ad22: @sbwsg Thanks for the explanation. That's a great way to organize the test cases. I've updated the PR.

test/v1alpha1/git_checkout_test.go
```diff
@@ -78,6 +78,61 @@ func TestGitPipelineRun(t *testing.T) {
...
+		description: "Fetch refs/pull/1009/head into a named master branch and then check it out",
+		revision:    "master",
+		refspec:     "refs/pull/1009/head:refs/heads/master",
+	}} {
+		t.Run(tc.description, func(t *testing.T) {
+			c, namespace := setup(t)
+			knativetest.CleanupOnInterrupt(func() { tearDown(t, c, namespace) }, t.Logf)
+			defer tearDown(t, c, namespace)
+
+			t.Logf("Creating Git PipelineResource %s", gitSourceResourceName)
```
sbwsg, line 107: I think it makes sense to log in the failure case with the `t.Fatalf` but these kinds of status lines can make test logs quite noisy to trace through when looking at verbose output.  I suggest removing the use of `t.Logf` unless absolutely necessary.  What do you think?

ad22: I've removed the log entries in the latest squashed commit.

ad22:
@afrittoli I've squashed my changes into a single commit

pkg/git/git.go
```diff
@@ -84,53 +86,99 @@ func Fetch(logger *zap.SugaredLogger, spec FetchSpec) error {
...
+	}
+
+	// git-init always creates and checks out an empty master branch. When the user requests
+	// "master" as the revision, git-fetch will refuse to update the HEAD of the branch it is
+	// currently on. The --update-head-ok parameter tells git-fetch that it is ok to update
+	// the current (empty) HEAD on initial fetch.
+	// The --force parameter tells git-fetch that its ok to update an existing HEAD in a
+	// non-fast-forward manner (though this cannot be possible on initial fetch, it can help
+	// when the refspec specifies the same destination twice)
+	fetchArgs = append(fetchArgs, "origin", "--update-head-ok", "--force")
```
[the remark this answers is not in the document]
ad22: I don't see the value in putting an additional conditional for that case. This is (guaranteed to be) the first fetch command run after git-init, so its implied that 
a. it will always be the master branch, unless git-init changes behavior 
b. we are not expecting any unintended updates to existing refs

cmd/git-init/main.go
```diff
@@ -28,16 +28,16 @@ import (
...
 )
 
 func init() {
 	flag.StringVar(&fetchSpec.URL, "url", "", "Git origin URL to fetch")
 	flag.StringVar(&fetchSpec.Revision, "revision", "", "The Git revision to make the repository HEAD")
+	flag.StringVar(&fetchSpec.Refspec, "refspec", "", "The Git refspec to fetch the revision from (optional)")
 	flag.StringVar(&fetchSpec.Path, "path", "", "Path of directory under which Git repository will be copied")
 	flag.BoolVar(&fetchSpec.SSLVerify, "sslVerify", true, "Enable/Disable SSL verification in the git config")
-	flag.BoolVar(&submodules, "submodules", true, "Initialize and fetch Git submodules")
+	flag.BoolVar(&fetchSpec.Submodules, "submodules", true, "Initialize and fetch Git submodules")
```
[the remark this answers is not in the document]
ad22: This is actually needed for the patch, to pass the submodule setting to git.go and prevent  --recurse-submodules from being appended to the fetch command when
the user specified submodule: "false"

vdemeester:
@ad22 needs a rebase, sorry :pray:

ad22:
Rebased and added minor fix for #1843

sbwsg:
I'm going to merge this now and if there are any outstanding comments that need to be addressed we can tackle them in a follow-up.  Thanks for the PR @ad22 !

/lgtm

PR #2321, opened 2020-04-02
Limit namespaces the webhook queries for secrets

sbwsg:
# Changes

The webhook uses informers to watch secrets in order to react when its
self-created certificates get updated. The webhook currently requests
cluster-wide access to secrets in order to do so.  This is unnecessary
however - the certifactes secret always lives alongside the webhook in
the tekton-pipelines namespace.

This commit reduces the scope of the webhook's informers so that they
only look for secrets in the tekton-pipelines namespace. This change
should be transparent to users but results in a relatively large drop
in the permissions the webhook needs to be granted.

In addition, this commit splits the webhook's service account from the
controller's and binds a reduced set of permissions vs those of the
controller via new roles and clusterroles.

Hat-tip to @eddie4941 for designing these changes.

# Submitter Checklist

review them:

### DIFF
--- a/config/201-rolebinding.yaml
+++ b/config/201-rolebinding.yaml
@@ -11,14 +11,17 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: tekton-pipelines-controller
-  namespace: tekton-pipelines
----
-apiVersion: v1
-kind: ServiceAccount
+
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: RoleBinding
 metadata:
   name: tekton-pipelines-webhook
   namespace: tekton-pipelines
+subjects:
+  - kind: ServiceAccount
+    name: tekton-pipelines-webhook
+    namespace: tekton-pipelines
+roleRef:
+  kind: Role
+  name: tekton-pipelines-webhook
+  apiGroup: rbac.authorization.k8s.io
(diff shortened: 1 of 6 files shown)


ImJasonH:
Can this change also drop the permissions requested in [the ClusterRole](https://github.com/tektoncd/pipeline/blob/2b933047eea67cc1a2d4f9b53161f886fad8a68b/config/200-clusterrole.yaml#L21)?

eddie4941:
@ImJasonH 

Correct. This would allow us to drop some of the permissions in the ClusterRole but not without some reworking of how the Roles/ClusterRoles are setup. We are working on a updated set of RBAC rules that will greatly reduce the permissions that ship with the `tekton-pipelines-admin` ClusterRole but were not done quite yet. This is one thing that fell out of doing that and seems like a pretty low risk change that can go in before the RBAC rules are complete refactored.

sbwsg:
@ImJasonH Good catch!  We've added a new service account specific to the webhook and bind a much reduced set of permissions through a new role and clusterrole.

sbwsg:
I'm seeing a lot of timeouts in e2e tests this morning.

sbwsg:
Integration failed due to a failed condition on a Task.  Doesn't appear to be related to the RBAC changes so re-running.

config/201-rolebinding.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright 2019 The Tekton Authors
```
savitaashture, line 1: I assume this file added in this PR so I think this should be 2020 
```suggestion
# Copyright 2020 The Tekton Authors
```

sbwsg: Good catch 😅 I'll make a PR to update this.

PR #2329, opened 2020-04-05
Emit events for all TaskRun lifecycle events

### LINKED ISSUE #2328: TaskRun emits no event when it fails on timeout
# Expected Behavior
When the task finishes, even in case of timeout, an event is emitted.

# Actual Behavior
There's no event generated

```
Status:
  Completion Time:  2020-04-05T09:21:31Z
  Conditions:
    Last Transition Time:  2020-04-05T09:21:31Z
    Message:               TaskRun "taskrun-that-runs-for-longz4wqz" failed to finish within "30s"
    Reason:                TaskRunTimeout
    Status:                False
    Type:                  Succeeded
  Pod Name:                taskrun-that-runs-for-longz4wqz-pod-lm8lx
  Start Time:              2020-04-05T09:21:01Z
  Steps:
    Container:  step-this-step-does-something-it-sleeps-30-seconds
    Image ID:   docker.io/library/busybox@sha256:b26cd013274a657b86e706210ddd5cc1f82f50155791199d29b9e86e935ce135
    Name:       this-step-does-something-it-sleeps-30-seconds
    Running:
      Started At:  2020-04-05T09:21:06Z
    Container:     step-fail
    Image ID:      docker.io/library/busybox@sha256:b26cd013274a657b86e706210ddd5cc1f82f50155791199d29b9e86e935ce135
    Name:          fail
    Running:
      Started At:  2020-04-05T09:21:07Z
    Container:     step-never-runs
    Image ID:      docker.io/library/busybox@sha256:b26cd013274a657b86e706210ddd5cc1f82f50155791199d29b9e86e935ce135
    Name:          never-runs
    Running:
      Started At:  2020-04-05T09:21:07Z
Events:
  Type    Reason   Age   From                Message
  ----    ------   ----  ----                -------
  Normal  Started  110s  taskrun-controller
```

1. `TaskRun`:

```
kind: TaskRun
metadata:
  generateName: taskrun-that-runs-for-long
spec:
  timeout: 30s
  taskSpec:
    steps:
    - name: this-step-sleeps-for-10m
      image: busybox
      script: |
        #!/bin/sh
        set -ex
        echo "Wait for a bit"
        sleep 600
```

# Additional Info

- Kubernetes version:

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:41:55Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.11+IKS", GitCommit:"0562ba8a2dfdd05f7f8721ab4952c02fe1605860", GitTreeState:"clean", BuildDate:"2020-03-13T14:45:42Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version:

`master`,  after `v0.11.0`

### PR DESCRIPTION
**Pull request by afrittoli**
# Changes

Emit events for additional TaskRun lifecyle events:
- taskrun started
- taskrun running

Fix the logic in events.go to compare semantic equality as
opposed to raw pointer equality.
Fix broken EmitEvents unit tests and extend them to cover new
functionality.

Extend reconcile test to verify new events are sent. To do so,
get the event recorder from the context when creating the
controller - if avaialble. This allows using the fake recorder
for testing instead of having to look for event related actions
in the fake client go action list.

Add documentation on events.

Fixes #2328
Work towards #2082

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
We now emit kubernetes events for additional TaskRun lifecyle events:
- taskrun started
- taskrun running
```

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -1,0 +1,39 @@
+<!--
+---
+linkTitle: "Events"
+weight: 2
+---
+-->
+# Events
+
+Tekton runtime resources, specifically `TaskRuns` and `PipelineRuns`,
+emit events when they are executed, so that users can monitor their lifecycle
+and react to it. Tekton emits [kubernetes events](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#event-v1-core), that can be retrieve from the resource via
+`kubectl describe [resource]`.
+
+No events are emitted for `Conditions` today (https://github.com/tektoncd/pipeline/issues/2461).
+
+## TaskRuns
+
+`TaskRun` events are generated for the following `Reasons`:
+- `Started`: this is triggered the first time the `TaskRun` is picked by the
+  reconciler from its work queue, so it only happens if web-hook validation was
+  successful. Note that this event does not imply that a step started executing,
+  as several conditions must be met first:
+  - task and bound resource validation must be successful
+  - attached conditions must run successfully
+  - the `Pod` associated to the `TaskRun` must be successfully scheduled
+- `Succeeded`: this is triggered once all steps in the `TaskRun` are executed
+  successfully, including post-steps injected by Tekton.
+- `Failed`: this is triggered if the `TaskRun` is completed, but not successfully.
+  Causes of failure may be: one the steps failed, the `TaskRun` was cancelled or
+  the `TaskRun` timed out.
+
+## PipelineRuns
+
+`PipelineRun` events are generated for the following `Reasons`:
+- `Succeeded`: this is triggered once all `Tasks` reachable via the DAG are
+  executed successfully.
+- `Failed`: this is triggered if the `PipelineRun` is completed, but not
+  successfully. Causes of failure may be: one the `Tasks` failed or the
+  `PipelineRun` was cancelled.
--- a/pkg/reconciler/event.go
+++ b/pkg/reconciler/event.go
@@ -17,6 +17,7 @@ package reconciler
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 	"knative.dev/pkg/apis"
@@ -27,20 +28,38 @@ const (
 	EventReasonSucceded = "Succeeded"
 	// EventReasonFailed is the reason set for events about unsuccessful completion of TaskRuns / PipelineRuns
 	EventReasonFailed = "Failed"
+	// EventReasonStarted is the reason set for events about the start of TaskRuns / PipelineRuns
+	EventReasonStarted = "Started"
 )
 
-// EmitEvent emits success or failed event for object
-// if afterCondition is different from beforeCondition
+// EmitEvent emits an event for object if afterCondition is different from beforeCondition
+//
+// Status "ConditionUnknown":
+//   beforeCondition == nil, emit EventReasonStarted
+//   beforeCondition != nil, emit afterCondition.Reason
+//
+//  Status "ConditionTrue": emit EventReasonSucceded
+//  Status "ConditionFalse": emit EventReasonFailed
+//
 func EmitEvent(c record.EventRecorder, beforeCondition *apis.Condition, afterCondition *apis.Condition, object runtime.Object) {
-	if beforeCondition != afterCondition && afterCondition != nil {
-		// Create events when the obj result is in.
+	if !equality.Semantic.DeepEqual(beforeCondition, afterCondition) && afterCondition != nil {
+		// If the condition changed, and the target condition is not empty, we send an event
 		switch afterCondition.Status {
 		case corev1.ConditionTrue:
 			c.Event(object, corev1.EventTypeNormal, EventReasonSucceded, afterCondition.Message)
-		case corev1.ConditionUnknown:
-			c.Event(object, corev1.EventTypeNormal, afterCondition.Reason, afterCondition.Message)
 		case corev1.ConditionFalse:
 			c.Event(object, corev1.EventTypeWarning, EventReasonFailed, afterCondition.Message)
+		case corev1.ConditionUnknown:
+			if beforeCondition == nil {
+				// If the condition changed, the status is "unknown", and there was no condition before,
+				// we emit the "Started event". We ignore further updates of the "unknown" status.
+				c.Event(object, corev1.EventTypeNormal, EventReasonStarted, "")
+			} else {
+				// If the condition changed, the status is "unknown", and there was a condition before,
+				// we emit an event that matches the reason and message of the condition.
+				// This is used for instance to signal the transition from "started" to "running"
+				c.Event(object, corev1.EventTypeNormal, afterCondition.Reason, afterCondition.Message)
+			}
 		}
 	}
 }
--- a/pkg/reconciler/event_test.go
+++ b/pkg/reconciler/event_test.go
@@ -21,38 +21,45 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/record"
 	"knative.dev/pkg/apis"
 )
 
 func TestEmitEvent(t *testing.T) {
 	testcases := []struct {
-		name        string
-		before      *apis.Condition
-		after       *apis.Condition
-		expectEvent bool
+		name          string
+		before        *apis.Condition
+		after         *apis.Condition
+		expectEvent   bool
+		expectedEvent string
 	}{{
-		name: "unknown to true",
+		name: "unknown to true with message",
 		before: &apis.Condition{
 			Type:   apis.ConditionSucceeded,
 			Status: corev1.ConditionUnknown,
 		},
 		after: &apis.Condition{
-			Type:   apis.ConditionSucceeded,
-			Status: corev1.ConditionTrue,
+			Type:    apis.ConditionSucceeded,
+			Status:  corev1.ConditionTrue,
+			Message: "all done",
 		},
-		expectEvent: true,
+		expectEvent:   true,
+		expectedEvent: "Normal Succeeded all done",
 	}, {
 		name: "true to true",
 		before: &apis.Condition{
-			Type:   apis.ConditionSucceeded,
-			Status: corev1.ConditionTrue,
+			Type:               apis.ConditionSucceeded,
+			Status:             corev1.ConditionTrue,
+			LastTransitionTime: apis.VolatileTime{Inner: metav1.NewTime(time.Now())},
 		},
 		after: &apis.Condition{
-			Type:   apis.ConditionSucceeded,
-			Status: corev1.ConditionTrue,
+			Type:               apis.ConditionSucceeded,
+			Status:             corev1.ConditionTrue,
+			LastTransitionTime: apis.VolatileTime{Inner: metav1.NewTime(time.Now().Add(5 * time.Minute))},
 		},
-		expectEvent: false,
+		expectEvent:   false,
+		expectedEvent: "",
 	}, {
 		name: "false to false",
 		before: &apis.Condition{
@@ -63,7 +70,24 @@ func TestEmitEvent(t *testing.T) {
 			Type:   apis.ConditionSucceeded,
 			Status: corev1.ConditionFalse,
 		},
-		expectEvent: false,
+		expectEvent:   false,
+		expectedEvent: "",
+	}, {
+		name: "unknown to unknown",
+		before: &apis.Condition{
+			Type:    apis.ConditionSucceeded,
+			Status:  corev1.ConditionUnknown,
+			Reason:  "",
+			Message: "",
+		},
+		after: &apis.Condition{
+			Type:    apis.ConditionSucceeded,
+			Status:  corev1.ConditionUnknown,
+			Reason:  "foo",
+			Message: "bar",
+		},
+		expectEvent:   true,
+		expectedEvent: "Normal foo bar",
 	}, {
 		name:  "true to nil",
 		after: nil,
@@ -71,7 +95,8 @@ func TestEmitEvent(t *testing.T) {
 			Type:   apis.ConditionSucceeded,
 			Status: corev1.ConditionTrue,
 		},
-		expectEvent: true,
+		expectEvent:   false,
+		expectedEvent: "",
 	}, {
 		name:   "nil to true",
 		before: nil,
@@ -79,7 +104,40 @@ func TestEmitEvent(t *testing.T) {
 			Type:   apis.ConditionSucceeded,
 			Status: corev1.ConditionTrue,
 		},
-		expectEvent: true,
+		expectEvent:   true,
+		expectedEvent: "Normal Succeeded ",
+	}, {
+		name:   "nil to unknown with message",
+		before: nil,
+		after: &apis.Condition{
+			Type:    apis.ConditionSucceeded,
+			Status:  corev1.ConditionUnknown,
+			Message: "just starting",
+		},
+		expectEvent:   true,
+		expectedEvent: "Normal Started ",
+	}, {
+		name: "unknown to false with message",
+		before: &apis.Condition{
+			Type:   apis.ConditionSucceeded,
+			Status: corev1.ConditionUnknown,
+		},
+		after: &apis.Condition{
+			Type:    apis.ConditionSucceeded,
+			Status:  corev1.ConditionFalse,
+			Message: "really bad",
+		},
+		expectEvent:   true,
+		expectedEvent: "Warning Failed really bad",
+	}, {
+		name:   "nil to false",
+		before: nil,
+		after: &apis.Condition{
+			Type:   apis.ConditionSucceeded,
+			Status: corev1.ConditionFalse,
+		},
+		expectEvent:   true,
+		expectedEvent: "Warning Failed ",
 	}}
 
 	for _, ts := range testcases {
@@ -90,12 +148,21 @@ func TestEmitEvent(t *testing.T) {
 
 		select {
 		case event := <-fr.Events:
-			if ts.expectEvent && event == "" {
-				t.Errorf("Expected event but got empty for %s", ts.name)
+			if event == "" {
+				// The fake recorder reported empty, it should not happen
+				t.Fatalf("Expected event but got empty for %s", ts.name)
+			}
+			if !ts.expectEvent {
+				// The fake recorder reported an event which we did not expect
+				t.Errorf("Unxpected event \"%s\" but got one for %s", event, ts.name)
+			}
+			if !(event == ts.expectedEvent) {
+				t.Errorf("Expected event \"%s\" but got \"%s\" instead for %s", ts.expectedEvent, event, ts.name)
 			}
 		case <-timer.C:
-			if !ts.expectEvent {
-				t.Errorf("Unexpected event but got for %s", ts.name)
+			if ts.expectEvent {
+				// The fake recorder did not report, the timer timeout expired
+				t.Errorf("Expected event but got none for %s", ts.name)
 			}
 		}
 	}
--- a/pkg/reconciler/taskrun/controller.go
+++ b/pkg/reconciler/taskrun/controller.go
@@ -44,6 +44,7 @@ const (
 	resyncPeriod = 10 * time.Hour
 )
 
+// NewController instantiates a new controller.Impl from knative.dev/pkg/controller
 func NewController(images pipeline.Images) func(context.Context, configmap.Watcher) *controller.Impl {
 	return func(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
 		logger := logging.FromContext(ctx)
@@ -66,6 +67,7 @@ func NewController(images pipeline.Images) func(context.Context, configmap.Watch
 			ConfigMapWatcher:  cmw,
 			ResyncPeriod:      resyncPeriod,
 			Logger:            logger,
+			Recorder:          controller.GetEventRecorder(ctx),
 		}
 
 		entrypointCache, err := pod.NewEntrypointCache(kubeclientset)
--- a/pkg/reconciler/taskrun/resources/cloudevent/cloud_event_controller.go
+++ b/pkg/reconciler/taskrun/resources/cloudevent/cloud_event_controller.go
@@ -66,8 +66,7 @@ func cloudEventDeliveryFromTargets(targets []string) []v1alpha1.CloudEventDelive
 }
 
 // SendCloudEvents is used by the TaskRun controller to send cloud events once
-// the TaskRun is complete. `tr` is used to obtain the list of targets but also
-// to construct the body of the
+// the TaskRun is complete. `tr` is used to obtain the list of targets
 func SendCloudEvents(tr *v1alpha1.TaskRun, ceclient CEClient, logger *zap.SugaredLogger) error {
 	logger = logger.With(zap.String("taskrun", tr.Name))
 
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -103,11 +103,16 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 
 	// If the TaskRun is just starting, this will also set the starttime,
 	// from which the timeout will immediately begin counting down.
-	tr.Status.InitializeConditions()
-	// In case node time was not synchronized, when controller has been scheduled to other nodes.
-	if tr.Status.StartTime.Sub(tr.CreationTimestamp.Time) < 0 {
-		c.Logger.Warnf("TaskRun %s createTimestamp %s is after the taskRun started %s", tr.GetRunKey(), tr.CreationTimestamp, tr.Status.StartTime)
-		tr.Status.StartTime = &tr.CreationTimestamp
+	if !tr.HasStarted() {
+		tr.Status.InitializeConditions()
+		// In case node time was not synchronized, when controller has been scheduled to other nodes.
+		if tr.Status.StartTime.Sub(tr.CreationTimestamp.Time) < 0 {
+			c.Logger.Warnf("TaskRun %s createTimestamp %s is after the taskRun started %s", tr.GetRunKey(), tr.CreationTimestamp, tr.Status.StartTime)
+			tr.Status.StartTime = &tr.CreationTimestamp
+		}
+		// Emit events
+		afterCondition := tr.Status.GetCondition(apis.ConditionSucceeded)
+		reconciler.EmitEvent(c.Recorder, nil, afterCondition, tr)
 	}
 
 	// If the TaskRun is complete, run some post run fixtures when applicable
@@ -200,7 +205,6 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 	if err = c.reconcile(ctx, tr, taskSpec, rtr); err != nil {
 		c.Logger.Errorf("Reconcile error: %v", err.Error())
 	}
-
 	// Emit events (only when ConditionSucceeded was changed)
 	after := tr.Status.GetCondition(apis.ConditionSucceeded)
 	reconciler.EmitEvent(c.Recorder, before, after, tr)
@@ -337,6 +341,9 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun,
 		if k8serrors.IsNotFound(err) {
 			// Keep going, this will result in the Pod being created below.
 		} else if err != nil {
+			// This is considered a transient error, so we return error, do not update
+			// the task run condition, and return an error which will cause this key to
+			// be requeued for reconcile.
 			c.Logger.Errorf("Error getting pod %q: %v", tr.Status.PodName, err)
 			return err
 		}
@@ -401,7 +408,6 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun,
 	}
 
 	c.Logger.Infof("Successfully reconciled taskrun %s/%s with status: %#v", tr.Name, tr.Namespace, tr.Status.GetCondition(apis.ConditionSucceeded))
-
 	return nil
 }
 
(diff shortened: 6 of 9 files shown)

### CODE REVIEW

**Comment by afrittoli** [comment]
One of the reason why certain tests fail here, is that we test on the number of actions executed in certain error cases, and this patch adds one "emit event" for the start case.
This got me to think about when do we want to consider the `TaskRun` as started, and thus emit an event for it.

This is roughly the ordering of events in the controller today. 
This is not 100% complete / correct, but it should be enough to highlight the issues.

main method|action|re-queue|set condition|events|notes
---|---|---|---|---|---
`Reconcile` | `hasStarted` | no | if `true`, set start time | start event | start timeout
`Reconcile` | `isDone` | maybe | no | | metrics, cloud events, stop timeout
`reconcile` | set defaults | no | no | | 
`reconcile` | convert | no | fail task on failure | | 
`reconcile` | `isCancelled` | no | fail task if `true` | end event | 
`reconcile` | get task | no | fail task on failure | | also convert when fetch from API 
`reconcile` | labels and annotations | no | no | | 
`reconcile` | check timeout | no | fail task if expired | end event | 
`reconcile` | bindings | no | fail task on failure | | pipeline resources / params
`reconcile` | create pod if needed | maybe | | | some errors are handled, some re-queue
`reconcile` | update from pod | maybe | end if pod end | end event | may be success of failure

Issues that I see:
1. Currently the start event is before we validate the conversion, task and bindings. In case of validation failures we have a start event but no end event.

We could move the start event to later on, but then it would not match with the start-time set in the taskrun anymore. Also later on it is a bit less obvious when a a `TaskRun` has to be considered as new, since start time has been set already. We could use the fact that the pod does not exists yet as an indication. Alternatively we could send event on other error conditions, like when `TaskRun` fails validation of fetching the task or so, but that feels a bit artificial. I think it would be consistent with metrics thought, not 100% sure.
I would like to get more opinion on this, I might bring it up at the WG. 
@vdemeester @bobcatfish @ImJasonH thoghts?

2. The timeout check is a bit in the middle of validations 

This is maybe a cosmetic thing, but I think we should move the timeout check somewhere else, maybe at the beginning of `reconcile`. I can file a separate issue for this.

3. We re-run fetching the task, doing conversion and validation on every reconcile cycle that hits `reconcile`. This is a bit `inefficient` and also may cause issues if the `Task` definition was to change in the middle of the `TaskRun`.

This is not new, and not something that we should change in this PR, but the fact that we might associate events to failures on getting the task / validations adds more to the problem.

**Comment by afrittoli** [comment]
The more I think about this, the more I tend to think that we should emit events at the very start of the TaskRun i.e. when the controller first looks at it, and emit failures in case of validation failure - since that is something that task/pipeline editors could look for when testing their work.

If there is no opposing opinion this week I'll move forward under this assumption, as I'd really like to move on with this series of PR swiftly now.

File: docs/events.md [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+---
+-->
+# Events
+
+Tekton runtime resources, specifically `TaskRuns` and `PipelineRuns`,
+emit events when they are executed, so that users can monitor their lifecycle
+and react to it. Tekton emits [kubernetes events](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#event-v1-core), that can be retrieve from the resource via
+`kubectl describe [resource]`.
+
+No events are emitted for `Conditions` today.
```
bobcatfish, line 14: maybe we could link to an issue to add this?

docs/events.md
```diff
@@ -0,0 +1,39 @@
...
+## TaskRuns
+
+`TaskRun` events are generated for the following `Reasons`:
+- `Started`: this is triggered the first time the `TaskRun` is picked by the
+  reconciler from its work queue, so it only happens if web-hook validation was
+  successful. Note that this event does not imply that a step started executing,
+  as several conditions must be met first:
+  - task and bound resource validation must be successful
+  - attached conditions must run successfully
+  - the `Pod` associated to the `TaskRun` must be successfully scheduled
```
bobcatfish, line 25: i think that's reasonable, we could later add events for each step potentially if we wanted to

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -106,11 +106,15 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
 	// If the TaskRun is just starting, this will also set the starttime,
 	// from which the timeout will immediately begin counting down.
-	tr.Status.InitializeConditions()
-	// In case node time was not synchronized, when controller has been scheduled to other nodes.
-	if tr.Status.StartTime.Sub(tr.CreationTimestamp.Time) < 0 {
-		c.Logger.Warnf("TaskRun %s createTimestamp %s is after the taskRun started %s", tr.GetRunKey(), tr.CreationTimestamp, tr.Status.StartTime)
-		tr.Status.StartTime = &tr.CreationTimestamp
+	if !tr.HasStarted() {
+		tr.Status.InitializeConditions()
+		// In case node time was not synchronized, when controller has been scheduled to other nodes.
```
bobcatfish, line 111: is there anywhere else we make this kind of comparision? this seems like the sort of thing that might be well suited to a separate package with unit tests (e.g. a package that allows us to compare times safely in cases of rescheduling)

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -191,6 +193,26 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+
+// Run any async logic that may be required once the tr is successfully reconciled
+// This method is used to emit events, notifications or any other async operation
+func (c *Reconciler) postRunAsyncHook(ctx context.Context, tr *v1alpha1.TaskRun, beforeCondition *apis.Condition) {
+	c.Logger.Infof("postRunAsyncHook: %s", tr.Name)
+
+	// Emit event
+	afterCondition := tr.Status.GetCondition(apis.ConditionSucceeded)
+	reconciler.EmitEvent(c.Recorder, beforeCondition, afterCondition, tr)
+}
```
bobcatfish, line 214: do these need to be functions on the reconciler? maybe we could make a new struct and initialize it with the function to emit events?

bobcatfish:
I like the table a lot but im not quite sure how to interpret the "requeue" and "notes" columns, maybe a bit more detail on those? 

i wonder if there is some way to isolate the logic that is used to emit events to make it a bit simpler, e.g. treat reconciling as having phases (i havent looked at your other PR @afrittoli which reorganizes reconciling so maybe you're doing this already!) but something like:

1. validation / fetching everything
2. making changes we need to make (e.g. creating TaskRuns, pods)
3. updating status + emitting events

Maybe we could look at the status the Runs and use that to decide what events to emit, i.e. isolate the logic for this all in one place instead of throughout the reconcile

Sounds like the logic around the start time is a bit more subtle than what above is capturing tho 🤔

docs/events.md
afrittoli: I created the issue, I will add the link https://github.com/tektoncd/pipeline/issues/2461

docs/events.md
afrittoli: Yeah, I started separating reconcile in prepare + reconcile  in https://github.com/tektoncd/pipeline/pull/2421, so we could add intermediate events and even intermediate conditions.

pkg/reconciler/taskrun/taskrun.go
afrittoli: We do the same in the pipeline controller.
I agree with your comment, but I would prefer to do that change in a PR on top.

pkg/reconciler/taskrun/taskrun.go
afrittoli: I put them on the reconciler to give them access to anything that they may need from it, but indeed I was thinking of changing the implementation there.

afrittoli:
> I like the table a lot but im not quite sure how to interpret the "requeue" and "notes" columns, maybe a bit more detail on those?
> 
> i wonder if there is some way to isolate the logic that is used to emit events to make it a bit simpler, e.g. treat reconciling as having phases (i havent looked at your other PR @afrittoli which reorganizes reconciling so maybe you're doing this already!) but something like:
> 
>     1. validation / fetching everything
> 
>     2. making changes we need to make (e.g. creating TaskRuns, pods)
> 
>     3. updating status + emitting events
> 
> 
> Maybe we could look at the status the Runs and use that to decide what events to emit, i.e. isolate the logic for this all in one place instead of throughout the reconcile
> 
> Sounds like the logic around the start time is a bit more subtle than what above is capturing tho 🤔

Yeah, I've been trying to organize things a bit more in the other PR - I would welcome feedback there. Looking at how other resources behave I think I need some changes in the PR related to the start event:
- only use `Succeeded` condition for the end event
- emit an event for the `Created` condition when the reconciler first meets the resource
- emit an event for the `Started` condition when validation is passed and we are about to attempt to create the pod
- emit an event for the `Running` condition once the pod is created successfully

For pipeline we will similarly have:
- only use `Succeeded` condition for the end event
- emit an event for the `Created` condition when the reconciler first meets the pipeline
- emit an event for the `Started` condition when validation is passed and we have a DAG 
- emit an event for the `Running` condition once the first TaskRun is applied successfully

afrittoli:
> > I like the table a lot but im not quite sure how to interpret the "requeue" and "notes" columns, maybe a bit more detail on those?
> > i wonder if there is some way to isolate the logic that is used to emit events to make it a bit simpler, e.g. treat reconciling as having phases (i havent looked at your other PR @afrittoli which reorganizes reconciling so maybe you're doing this already!) but something like:
> > ```
> > 1. validation / fetching everything
> > 
> > 2. making changes we need to make (e.g. creating TaskRuns, pods)
> > 
> > 3. updating status + emitting events
> > ```
> > 
> > 
> > Maybe we could look at the status the Runs and use that to decide what events to emit, i.e. isolate the logic for this all in one place instead of throughout the reconcile
> > Sounds like the logic around the start time is a bit more subtle than what above is capturing tho 🤔
> 
> Yeah, I've been trying to organize things a bit more in the other PR - I would welcome feedback there. Looking at how other resources behave I think I need some changes in the PR related to the start event:
> 
>     * only use `Succeeded` reason for the end event
> 
>     * emit an event for the `Created` reason when the reconciler first meets the resource
> 
>     * emit an event for the `Started` reason when validation is passed and we are about to attempt to create the pod
> 
>     * emit an event for the `Running` reason once the pod is created successfully
> 
> 
> For pipeline we will similarly have:
> 
>     * only use `Succeeded` reason for the end event
> 
>     * emit an event for the `Created` reason when the reconciler first meets the pipeline
> 
>     * emit an event for the `Started` reason when validation is passed and we have a DAG
> 
>     * emit an event for the `Running` reason once the first TaskRun is applied successfully

@bobcatfish going back on your point of organizing the reconcile in phases, I think it might be nice to have use different conditions during reconcile to identify them, and send events on transitions. The only downside might be more updates to the status. This is something I would prefer to do after this patch though.

We should also add a doc page about different conditions we have today and their meaning.

afrittoli:
> The following is the coverage report on the affected files.
> Say `/test pull-tekton-pipeline-go-coverage` to re-run this coverage report
> File 	Old Coverage 	New Coverage 	Delta
> [pkg/reconciler/event.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/2329/pull-tekton-pipeline-go-coverage/1255132491145023490/artifacts/line-cov.html#file0) 	80.0% 	100.0% 	20.0
> [pkg/reconciler/taskrun/taskrun.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/2329/pull-tekton-pipeline-go-coverage/1255132491145023490/artifacts/line-cov.html#file3) 	74.5% 	75.1% 	0.6

Nice increase in coverage, I only need to fix the data race now...

pkg/reconciler/taskrun/taskrun.go
afrittoli: I decided to remove these functions - at least until I actually need to use them.
Emitting events is already non-blocking - down in the implementation of `generateEvent`.
So there was little value in adding functions here.
Besides, fetching the afterCondition in a goroutine triggered a data race in the unit tests, which is a good sign that I should not do that.

pkg/reconciler/event.go
```diff
@@ -27,20 +28,38 @@ const (
...
+	EventReasonStarted = "Started"
 )
 
-// EmitEvent emits success or failed event for object
-// if afterCondition is different from beforeCondition
+// EmitEvent emits an event for object if afterCondition is different from beforeCondition
+//
+// Status "ConditionUnknown":
+//   beforeCondition == nil, emit EventReasonStarted
+//   beforeCondition != mil, emit afterCondition.Reason
```
vdemeester, line 39: `s/mil/nil` :stuck_out_tongue_closed_eyes:

docs/events.md
```diff
@@ -0,0 +1,39 @@
+<!--
+---
+linkTitle: "Events"
+weight: 2
+---
+-->
+# Events
```
bobcatfish, line 7: 🎉

docs/events.md
```diff
@@ -0,0 +1,39 @@
...
+  successful. Note that this event does not imply that a step started executing,
+  as several conditions must be met first:
+  - task and bound resource validation must be successful
+  - attached conditions must run successfully
+  - the `Pod` associated to the `TaskRun` must be successfully scheduled
+- `Succeeded`: this is triggered once all steps in the `TaskRun` are executed
+  successfully, including post-steps injected by Tekton.
+- `Failed`: this is triggered if the `TaskRun` is completed, but not successfully.
+  Causes of failure may be: one the steps failed, the `TaskRun` was cancelled or
+  the `TaskRun` timed out.
```
bobcatfish, line 30: what if it was skipped? (maybe to be added?)

docs/events.md
bobcatfish: would it make sense there to mention those plans here, maybe link to an issue?

pkg/reconciler/event.go
```diff
@@ -27,20 +28,38 @@ const (
...
 		case corev1.ConditionTrue:
 			c.Event(object, corev1.EventTypeNormal, EventReasonSucceded, afterCondition.Message)
-		case corev1.ConditionUnknown:
-			c.Event(object, corev1.EventTypeNormal, afterCondition.Reason, afterCondition.Message)
 		case corev1.ConditionFalse:
 			c.Event(object, corev1.EventTypeWarning, EventReasonFailed, afterCondition.Message)
+		case corev1.ConditionUnknown:
+			if beforeCondition == nil {
+				// If the condition changed, the status is "unknown", and there was no condition before,
+				// we emit the "Started event". We ignore further updates of the "uknown" status.
```
bobcatfish, line 55: typo-> unknown

pkg/reconciler/event_test.go
```diff
@@ -90,12 +148,21 @@ func TestEmitEvent(t *testing.T) {
...
 		select {
 		case event := <-fr.Events:
-			if ts.expectEvent && event == "" {
-				t.Errorf("Expected event but got empty for %s", ts.name)
+			if event == "" {
+				// The fake recorder reported empty, it should not happen
+				t.Fatalf("Expected event but got empty for %s", ts.name)
 			}
-		case <-timer.C:
 			if !ts.expectEvent {
```
bobcatfish, line 155: i'm a bit confused by the cases that don't require an event, do you think it might make sense to have an event every time the condition changes?

pkg/reconciler/event_test.go
```diff
@@ -90,12 +148,21 @@ func TestEmitEvent(t *testing.T) {
...
-				t.Errorf("Unexpected event but got for %s", ts.name)
+				// The fake recorder reported an event which we did not expect
+				t.Errorf("Unxpected event \"%s\" but got one for %s", event, ts.name)
+			}
+			if !(event == ts.expectedEvent) {
+				t.Errorf("Expected event \"%s\" but got \"%s\" instead for %s", ts.expectedEvent, event, ts.name)
+			}
+		case <-timer.C:
+			if ts.expectEvent {
+				// The fake recorder did not report, the timer imeout expired
```
bobcatfish, line 164: typo timeout

pkg/reconciler/taskrun/controller.go
```diff
@@ -44,6 +44,7 @@ const (
 	resyncPeriod = 10 * time.Hour
 )
 
+// NewController instantiates a new controller.Impl from knative.dev/pkg/controller
```
bobcatfish, line 47: 🎉

pkg/reconciler/taskrun/controller.go
```diff
@@ -66,6 +67,7 @@ func NewController(images pipeline.Images) func(context.Context, configmap.Watch
 			ConfigMapWatcher:  cmw,
 			ResyncPeriod:      resyncPeriod,
 			Logger:            logger,
+			Recorder:          controller.GetEventRecorder(ctx),
```
bobcatfish, line 70: is this just for testing? that's a bit unfortunate if so :( probably worth it tho?

pkg/reconciler/taskrun/resources/cloudevent/cloud_event_controller.go
```diff
@@ -66,8 +66,7 @@ func cloudEventDeliveryFromTargets(targets []string) []v1alpha1.CloudEventDelive
 }
 
 // SendCloudEvents is used by the TaskRun controller to send cloud events once
-// the TaskRun is complete. `tr` is used to obtain the list of targets but also
-// to construct the body of the
```
bobcatfish, line 70: 😅

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -337,6 +341,9 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun,
 		if k8serrors.IsNotFound(err) {
 			// Keep going, this will result in the Pod being created below.
 		} else if err != nil {
+			// This is considered a transient error, so we return error, do not update
+			// the task run condition, and return an error which will cause this key to
+			// be requeued for reconcile.
```
bobcatfish, line 346: 🎉

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -103,11 +103,16 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+	if !tr.HasStarted() {
+		tr.Status.InitializeConditions()
+		// In case node time was not synchronized, when controller has been scheduled to other nodes.
+		if tr.Status.StartTime.Sub(tr.CreationTimestamp.Time) < 0 {
+			c.Logger.Warnf("TaskRun %s createTimestamp %s is after the taskRun started %s", tr.GetRunKey(), tr.CreationTimestamp, tr.Status.StartTime)
+			tr.Status.StartTime = &tr.CreationTimestamp
+		}
+		// Emit events
+		afterCondition := tr.Status.GetCondition(apis.ConditionSucceeded)
+		reconciler.EmitEvent(c.Recorder, nil, afterCondition, tr)
```
bobcatfish, line 115: i'm a bit confused why this call to `EmitEvent` is needed in addition to the other one a bunch of lines down in the same function

is it possible we can get away with one? i do like the idea of only having to call this logic once instead of being aware of 2 places it is called

or if it is required maybe a comment explaining why?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -1038,15 +1089,9 @@ func TestReconcile(t *testing.T) {
...
-			for _, action := range actions {
-				if action.GetVerb() == "create" && action.GetResource().Resource == "events" {
-					eventCount++
-				}
-			}
-			if d := cmp.Diff(tc.wantEvents, eventCount); d != "" {
-				t.Errorf("Event count does not match (-want, +got): %s. ", d)
+			err = checkEvents(fr, tc.name, tc.wantEvents)
+			if !(err == nil) {
+				t.Errorf(err.Error())
```
bobcatfish, line 1094: another option here would be to let checkEvents actually fail the test by passing in t and using t.Helper

normally i like the clarity of handling the error, but in this case it looks like our error handling is really not adding any additional useful context, and i think the effect of using t.Helper might be the same as doing this?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -271,6 +272,29 @@ func getTaskRunController(t *testing.T, d test.Data) (test.Assets, func()) {
 	}, cancel
 }
 
+func checkEvents(fr *record.FakeRecorder, testName string, wantEvents []string) error {
+	timer := time.NewTimer(1 * time.Second)
```
bobcatfish, line 276: whats the timer for and why 1 second? maybe a comment to explain? (im guessing that since this is a "unit"-ish test that this timeout is _extremely_ unlikely to occur)

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -271,6 +272,29 @@ func getTaskRunController(t *testing.T, d test.Data) (test.Assets, func()) {
...
+	foundEvents := []string{}
+	for ii := 0; ii < len(wantEvents)+1; ii++ {
```
bobcatfish, line 278: i find it a bit confusing that this function is both looping and receiving events via channel at the same time, but then after thinking through the alternative, this seems better - do you think it might be worth a comment explaining what is happening here and why?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -271,6 +272,29 @@ func getTaskRunController(t *testing.T, d test.Data) (test.Assets, func()) {
...
+		case event := <-fr.Events:
+			foundEvents = append(foundEvents, event)
+			if ii > len(wantEvents)-1 {
+				return fmt.Errorf("Received event \"%s\" for %s but not more expected", event, testName)
+			}
+			wantEvent := wantEvents[ii]
+			if !(strings.HasPrefix(event, wantEvent)) {
+				return fmt.Errorf("Expected event \"%s\" but got \"%s\" instead for %s", wantEvent, event, testName)
+			}
+		case <-timer.C:
```
bobcatfish, line 289: if you ran just one test that used this helper, and all events were received in < 1 second, would this go routine stay running and force the test to execute for at least 1 second?

maybe not important, just wondering if there's a way around it

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -1213,24 +1267,33 @@ func TestReconcileInvalidTaskRuns(t *testing.T) {
...
 			clients := testAssets.Clients
-			err := c.Reconciler.Reconcile(context.Background(), getRunName(tc.taskRun))
+			reconciler := c.Reconciler.(*Reconciler)
+			fr := reconciler.Recorder.(*record.FakeRecorder)
+			err := reconciler.Reconcile(context.Background(), getRunName(tc.taskRun))
+
 			// Events are sent in a goroutine, let's sleep a bit to make sure they're
 			// captured by the fake client-go action list
-			time.Sleep(100 * time.Millisecond)
+			time.Sleep(2000 * time.Millisecond)
```
bobcatfish, line 1276: 2 * time.Second?

I'd really like to take some time to think about whether or not we can avoid having sleeps in these tests, they just always seem to lead to pain later on

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -1213,24 +1267,33 @@ func TestReconcileInvalidTaskRuns(t *testing.T) {
...
+			// Check actions and events
 			actions := clients.Kube.Actions()
-			if len(actions) != 2 ||
-				actions[0].GetVerb() != "list" ||
-				actions[0].GetResource().Resource != "namespaces" ||
-				actions[1].GetVerb() != "create" ||
-				actions[1].GetResource().Resource != "events" {
-				t.Errorf("expected two actions (list namespaces + event) created by the reconciler, got %+v", actions)
+			if len(actions) != 1 ||
+				actions[0].Matches("namespaces", "list") {
```
bobcatfish, line 1288: one line maybe?

THE MOST MINOR COMMENT OF ALL TIME

docs/events.md
afrittoli: A TaskRun is something that may only happen when a Pipeline is involved, so I did not look into that on this patch, where I focuses on events generated by the TaskRun controller.
There are two kind of skips:
- skip because of a condition not met - which is easy, and we definitely can emit an event for that. I opened an issue which is related https://github.com/tektoncd/pipeline/issues/2461
- skip because it was not reached in the DAG: we don't do anything today we tasks that were not reached in the DAG because of previous failures, but it should be possible to process them as a post-run step in the pipelinerun reconciler

pkg/reconciler/taskrun/controller.go
afrittoli: Yes, injecting fakes in the context is the same mechanism that we use to inject fake k8s client-go and other fakes.

pkg/reconciler/event_test.go
afrittoli: The only cases where we do not emit an event are:
- there was no change in condition (except for LastTransitionTime) 
- afterCondition == nil

For the former I think we do not want an event, it may be we wrote the condition again and updated LastTransitionTime, but nothing really happened to the Condition, so no event.
For the latter, I could not think of a use case where afterCondition is nil, so I saw no point in adding that case now.

pkg/reconciler/taskrun/taskrun.go
afrittoli: You need to emit the event every time the condition changes, because - as things are today - you could have more than one condition change in one reconcile cycle.

That could change if we added extra conditions in future, i.e. instead of only using Succeeded (unknown/true/false) we could use Started and Running too. In any case, I think it's worth sending the event as early as possible. This reconcile cycle will go on to create a pod and run several API calls, so we may be losing a few hundreds milliseconds if we wait. 

If something is looking for the event to update the status of a check in github, half a second later makes a difference.

pkg/reconciler/taskrun/taskrun_test.go
afrittoli: it would not run for 1s, because of the exit condition in the loop `ii < len(wantEvents)+1`

pkg/reconciler/taskrun/taskrun_test.go
afrittoli: the fake recorder runs in a go routine, so the timeout is just there to avoid waiting on the channel forever if fewer than expected events are received. I guess it could be reduced, but it will only be reached in case of test failure, so I don't think it matters really.

pkg/reconciler/taskrun/taskrun_test.go
afrittoli: yeah, I can add one.
afaik the select case is the easiest way to multiplex N channels, so to handle the timeout which is needed in failure cases (fewer events then expected).

pkg/reconciler/taskrun/taskrun_test.go
afrittoli: I can change it if you prefer so. Personally I prefer to see in the body of the test where t.Errorf (or t.Fatalf) is invoked, without having to look into other functions. But it's a minor thing.
Another minor aspect is that the caller of `checkEvents` still has the choice of t.Errorf and t.Fatalf depending on whether this check is blocking or not for further checks.

pkg/reconciler/taskrun/taskrun_test.go
afrittoli: So, actually this was needed with the fake client-go and checking of actions, and it's one of the reason I looked for another way of running the test.

Now, since I switched to the fake recorder and the `checkEvents` approach, my bet is that this is not required anymore. I will try removing it.

pkg/reconciler/taskrun/taskrun_test.go
afrittoli: 😅 and I even thought about it, not sure why I did not do it?

afrittoli:
> All of my feedback is super minor! Mostly comments and a couple typos.
> 
> The only piece of feedback that's at all substantial is that I'd really like to find away to avoid the sleep in the reconciler test. I don't mind merging this as is but if we could find an alternative to that I think it's worth the effort (or at least a detailed comment explaining why it's unavoidable - imo sleeps, especially in unit tests, should not be the norm)

Yeah, good catch - this is actually one of the reasons I changed how we run event related unit tests - and finally I forgot to drop the sleep. My bet is that this is not needed anymore, so I will re-spin this without it and with fixes to typos.

pkg/reconciler/event.go
afrittoli: Done

pkg/reconciler/event.go
afrittoli: I'm a professional typo maker :D
Fixed now

pkg/reconciler/event_test.go
afrittoli: Done

docs/events.md
bobcatfish: ah right, there's no taskrun at all, makes sense. those 2 skips are totally valid also. like you said, conditions aren't emitting events anyway, maybe we can revisit this then.

thanks!!

pkg/reconciler/event.go
bobcatfish: hahaha me too XD

pkg/reconciler/taskrun/taskrun.go
bobcatfish: >  you could have more than one condition change in one reconcile cycle.

sounds like that's only in the started case? going from started to running - and shouldn't be expected from any other states? 

anyway makes sense! i do think this is worth a comment explaining why the 2 calls but also wouldnt block on this

pkg/reconciler/taskrun/taskrun_test.go
bobcatfish: i dont feel strongly enough to push back! i think t.Helper has pros and cons and that kind of thing is sort of an antipattern outside of tests

good point also re. error or fatal

pkg/reconciler/taskrun/taskrun_test.go
bobcatfish: sounds reasonable! i do think a comment might make sense tho?

pkg/reconciler/taskrun/taskrun_test.go
bobcatfish: nice!!

pkg/reconciler/taskrun/taskrun.go
afrittoli: Yeah, it's only in that case. I promise I'll add a comment in a follow-up 👼

bobcatfish:
I think there were 1 or 2 comments that might make sense to add but nothing worth blocking over!

Thanks for this @afrittoli !! :D 🎉 

/lgtm

bobcatfish:
ps

image

nice!!! 👍 👍 👍

PR #2344, opened 2020-04-07
[master] Auto-update dependencies

mattmoor:
Produced via:
  `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh`
/assign ImJasonH vdemeester
/cc ImJasonH vdemeester

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,6 @@ go 1.13
 
 require (
 	cloud.google.com/go v0.47.0 // indirect
-	cloud.google.com/go/storage v1.0.0
 	contrib.go.opencensus.io/exporter/stackdriver v0.12.8 // indirect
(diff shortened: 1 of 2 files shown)


mattmoor:
Ignore this.  I need to add `vendor/modules.txt` to the list of files to ignore when considering whether a PR has any meaningful changes.

This is up, though, so as you want to start floating deps forward tweak `./hack/update-deps.sh`.  We plan to freeze `release-0.14` of knative/pkg today.

cc @vdemeester

vdemeester:
@mattmoor also, it should add the diff in `vendor` :angel: I wasn't sure if that was the case or not :stuck_out_tongue:

mattmoor:
@vdemeester Right, but there are times when only `go.mod` changes because of `replace` rules, which wouldn't touch vendor.

vdemeester:
ah :sweat_smile:

PR #2346, opened 2020-04-07
Split controller RBAC into cluster-wide and tenant roles

sbwsg:
# Changes

The controller currently operates with a single ClusterRole that spans a very broad set of access permissions.  In multi-tenant scenarios this kind of RBAC configuration can be quite dangerous.  In order to better support potential multi-tenant configurations this PR splits the roles that the controller receives into two:

1. `tekton-pipelines-controller-cluster-access`: those permissions needed cluster-wide by the controller.
2. `tekton-pipelines-controller-tenant-access` those permissions needed on a namespace-by-namespace basis.

This PR does not actually change the level of access afforded to the controller.  Instead, the roles are split but remain cluster-scoped by default.  **There should be no noticeable change in behaviour from the existing RBAC configuration in master.**

If a team wanted to start running a multi-tenant service they would be able to bind `tekton-pipelines-controller-tenant-access` using a `RoleBinding` instead of a `ClusterRoleBinding`, thereby limiting the access that the controller has to specific tenant namespaces.

Hat-tip to @eddie4941 for designing these changes.

# Submitter Checklist

review them:

- [x] Includes documentation

### DIFF
--- a/config/200-clusterrole.yaml
+++ b/config/200-clusterrole.yaml
@@ -15,20 +15,17 @@
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: tekton-pipelines-admin
+  name: tekton-pipelines-controller-cluster-access
 rules:
   - apiGroups: [""]
-    resources: ["pods", "pods/log", "namespaces", "secrets", "events", "serviceaccounts", "configmaps", "persistentvolumeclaims", "limitranges"]
-    verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
-  - apiGroups: ["apps"]
-    resources: ["deployments"]
-    verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
-  - apiGroups: ["apps"]
-    resources: ["deployments/finalizers"]
-    verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
-  - apiGroups: ["admissionregistration.k8s.io"]
-    resources: ["mutatingwebhookconfigurations", "validatingwebhookconfigurations"]
-    verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
+    # Namespace access is required because the controller timeout handling logic
+    # iterates over all namespaces and times out any PipelineRuns that have expired.
+    # Pod access is required because the taskrun controller wants to be updated when
+    # a Pod underlying a TaskRun changes state.
+    resources: ["namespaces", "pods"]
+    verbs: ["list", "watch"]
+    # Controller needs cluster access to all of the CRDs that it is responsible for
+    # managing.
   - apiGroups: ["tekton.dev"]
     resources: ["tasks", "clustertasks", "taskruns", "pipelines", "pipelineruns", "pipelineresources", "conditions"]
     verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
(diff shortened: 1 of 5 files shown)


sbwsg:
Good catch!  This PR was branched from the webhook one.  I'll rebase on top of master now that it's merged.

Also, good call on the doc, I'll add some supporting information.

sbwsg:
I've added a short description to the `docs/developers/README.md` just to document the existence of these two roles and their intended usage.  I think there's an opportunity to provide much stronger guidelines on building multi-tenant services with Tekton but I'm not knowledgeable enough to do that right now.

I've also updated this PR to be based off of master, which has removed a couple of files from the diff since they were already merged in an earlier commit.

config/200-clusterrole.yaml
```diff
@@ -45,6 +42,24 @@ rules:
...
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  # This is the access that the controller needs on a per-namespace basis.
+  name: tekton-pipelines-controller-tenant-access
+rules:
+  - apiGroups: [""]
+    resources: ["pods", "pods/log", "namespaces", "secrets", "events", "serviceaccounts", "configmaps", "persistentvolumeclaims", "limitranges"]
+    verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
+    # Unclear if this access is actually required.  Simply a hold-over from the previous
```
dibyom, line 52: Hmm, yeah this one is suspect...maybe we can get rid of it in a followup

config/200-role.yaml
```diff
@@ -12,6 +12,21 @@
...
+  namespace: tekton-pipelines
+rules:
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["list", "watch"]
+  # The controller needs access to these configmaps for logging information and runtime configuration.
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["get"]
+    resourceNames: ["config-logging", "config-observability", "config-artifact-bucket", "config-artifact-pvc", "feature-flags"]
```
sbwsg, line 28: Note to self: these configmap names are configurable - overrideable with environment variables in the controller pod.  We should probably document these being named in the Role and tell users to make sure they change them here as well if they modify the defaults.

eddie4941: Yup thats correct. The names are listed out here: https://github.com/tektoncd/pipeline/blob/master/config/controller.yaml#L66-L75

Alternatively, if we are worried about keeping these in sync, we can have the default role thats included in release.yaml not list resourceNames. We can then add a line in the docs saying that its possible to limit the names further if we want by added resourceNames that match those that are passed to the controller via env vars. Though being explicitly is nicer from rbac perspective as you now exactly what the controller is doing.

config/200-clusterrole.yaml
```diff
@@ -45,6 +42,24 @@ rules:
 ---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  # This is the access that the controller needs on a per-namespace basis.
+  name: tekton-pipelines-controller-tenant-access
+rules:
+  - apiGroups: [""]
+    resources: ["pods", "pods/log", "namespaces", "secrets", "events", "serviceaccounts", "configmaps", "persistentvolumeclaims", "limitranges"]
```
eddie4941, line 50: Just noticed this here. Were listing `namespaces` even though they are not a namespaces resource. So we can drop it form here. This is already covered in the `tekton-pipelines-controller-cluster-access` `ClusterRole` anyway. 
This shouldn't cause any issues and should be treated as a no-op, but it would be nice to clean it up.

Issue #2267, opened 2020-03-23
Document all variables available in Tekton in one place


afrittoli:
# Expected Behavior
Have a single reference document with all variables `$()` available in different context to Tekton resource authors.

# Actual Behavior
Today we have reference to different variables available to resource authors, but there's no clear reference anywhere of what variables can be used where.

sbwsg:
As an exercise I just had a look through to find all the substitutions I could.  These are they:

### Available in Tasks
- `params.<param name>`
- `resources.inputs.<resource name>.path`
- `resources.outputs.<resource name>.path`
- `results.<result name>.path`
- `workspaces.<workspace name>.path`
- `workspaces.<workspace name>.volume`
- `credentials.path`

Each PipelineResource type also defines its own variables that we seem to only really have in code.  Each Resource exposes a `Replacements()` func that returns them:

- [Git PipelineResource Variables](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/resource/v1alpha1/git/git_resource.go#L130)
- [PullRequest Resource Variables](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/resource/v1alpha1/pullrequest/pull_request_resource.go#L102)
- [Image Resource Variables](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/resource/v1alpha1/image/image_resource.go#L70)
- [GCS Resource Variables](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/resource/v1alpha1/storage/gcs.go#L103)
- [GCS Build Resource Variables](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/resource/v1alpha1/storage/build_gcs.go#L123)
- [Cluster Resource Variables](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/resource/v1alpha1/cluster/cluster_resource.go#L124)

(These are all accessed through `resources.inputs.<resource name>.<variable>` or `resources.outputs.<resource name>.<variable>`)

### Available in Pipelines
- `params.<param name>`
- `tasks.<task name>.results.<result name>`

PR #2358, opened 2020-04-09
Document Tekton's Many Variables

sbwsg:
# Changes

Closes https://github.com/tektoncd/pipeline/issues/2267

Tekton exposes a number of variables in Tasks and Pipelines that are substituted for their runtime values. The complete set of variables isn't documented in any single place.

This commit introduces a doc to try and capture all the variables in one place.

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -1,0 +1,108 @@
+<!--
+---
+linkTitle: "Variable Substitutions"
+weight: 15
+---
+-->
+# Variable Substitutions
+
+This doc aggregates the complete set of variable substitions available
+in `Tasks` and `Pipelines`.
+
+## Variables Available in a Pipeline
+
+| Variable | Description |
+| -------- | ----------- |
+| `params.<param name>` | The value of the param at runtime. |
+| `tasks.<task name>.results.<result name>` | The value of a Task's result (**Note**: Affects Task ordering in a Pipeline!) |
+
+## Variables Available in a Task
+
+| Variable | Description |
+| -------- | ----------- |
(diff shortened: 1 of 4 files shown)


docs/variables.md
```diff
@@ -0,0 +1,108 @@
...
+in `Tasks` and `Pipelines`.
+
+## Task Variables
+
+| Variable | Description |
+| -------- | ----------- |
+| `params.<param name>` | The value of the param at runtime. |
+| `resources.inputs.<resource name>.path` | The path to the input resource's directory. |
+| `resources.outputs.<resource name>.path` | The path to the output resource's directory. |
+| `results.<result name>.path` | The path to the file where a Task's result should be written. |
```
afrittoli, line 19: I thought it was possible for a Task to refer to another task results now?

afrittoli: Something like:

```
tasks.[task-name].results.[result-name]
```

sbwsg: Ah, this is a Pipeline Task variable, which got pushed down under the PipelineResource variables since they're Task-only i think.  I'll move the Pipeline variables up higher.

sbwsg:
excellent point on linking from another doc, i'll do that.

docs/variables.md
afrittoli: Oh, right, I missed that, thank you!

sbwsg:
I've updated with links from other docs and also moved the Pipeline Variables to the top of the doc.

Issue #2337, opened 2020-04-06
PipelineRun Cancellation is not working


sbwsg:
# Expected Behavior

Cancelling a PipelineRun should result in its TaskRuns being put into a cancelled state.

Our integration test, `test/cancel_test.go` should pass.

# Actual Behavior

Cancelling a PipelineRun no longer seems to cancel the underlying TaskRun.  It appears that each attempt to cancel the TaskRuns hit errors like this:

```
error(s) from cancelling TaskRun(s) from PipelineRun pear: Operation cannot be fulfilled on taskruns.tekton.dev \"pear-foo-cb9rg\": the object has been modified; please apply your changes to the latest version and try again
```

The e2e `test/cancel_test.go` fails.

# Steps to Reproduce the Problem

1. `go test -v -count=1 -tags=e2e -timeout=20m -run ".*PipelineRunCancel.*" ./test`
2. Watch the PipelineRuns - they are cancelled almost immediately.
3. Look at the TaskRuns - these never get cancelled and the test times out.

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

I've tested against a 1.18 server and a 1.15 server.  Both exhibit the same issue.

  ```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.5", GitCommit:"20c265fef0741dd71a66480e35bd69f18351daea", GitTreeState:"clean", BuildDate:"2019-10-15T19:16:51Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:50:46Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version:

Latest master

vincent-pli:
@sbwsg I checked the code:
https://github.com/tektoncd/pipeline/blob/8a059fe34a923f23a2a88b46239b9c9efb3a3a65/pkg/reconciler/pipelinerun/cancel.go#L48-L54

It update the `status` then the `spec`, it will raise the error you mentioned.
Why we need update the `status` of `taskrun` which will be cancelled? since I'm not found other such update in `reconcile` of pipelinerun.

sbwsg:
[...]

PR #2359, opened 2020-04-09
Mark PipelienResourceBinding as inlined 🏑

vdemeester:
# Changes

Without this, using `Unstructured` to unmarshall the structs will
result in empty/nil resources.

This is a problem for `tektoncd/cli` v1beta1 support :sweat:. Should we consider it for a 0.11.2 ?

/cc @sbwsg @bobcatfish 

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Fix embedded PipelineResourceBinding json unmarshalling
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/taskrun_types.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_types.go
@@ -75,7 +75,7 @@ type TaskRunInputs struct {
 // TaskResourceBinding points to the PipelineResource that
 // will be used for the Task input or output called Name.
 type TaskResourceBinding struct {
-	PipelineResourceBinding
+	PipelineResourceBinding `json:",inline"`
 	// Paths will probably be removed in #1284, and then PipelineResourceBinding can be used instead.
 	// The optional Path field corresponds to a path on disk at which the Resource can be found
 	// (used when providing the resource via mounted volume, overriding the default logic to fetch the Resource).


sbwsg:
I don't fully understand what's going on here.  Unstructured is [this type from k8s](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/unstructured/unstructured.go#L41) right?  And `inline` I found documented in the `bson` package but not `json`, it means that json fields don't need to be nested in order for their values to end up in the embedded struct, is that right?

So at the moment when cli unmarshals `TaskResourceBinding` into an `Unstructured` variable, we lose `PipelineResourceBinding` - it gets set to nil.  Is that right?

/approve

vdemeester:
> I don't fully understand what's going on here. Unstructured is [this type from k8s](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/unstructured/unstructured.go#L41) right? And `inline` I found documented in the `bson` package but not `json`, it means that json fields don't need to be nested in order for their values to end up in the embedded struct, is that right?

`inline` for some of the go marshalling json library, means this is a embedded type, the fields from it are at the same level as the field. Which is the case here (and for other field in the API).

> So at the moment when cli unmarshals `TaskResourceBinding` into an `Unstructured` variable, we lose `PipelineResourceBinding` - it gets set to nil. Is that right?
> 
> /approve

yes

piyush-garg:
This will need https://github.com/tektoncd/pipeline/pull/2359 to work properly

sbwsg:
I'm pretty sure these failures are https://github.com/tektoncd/pipeline/issues/2337

sbwsg:
looks like this should pass after a rebase!  the integration failed on the pipelinerun cancel e2e test we merged earlier today.

PR #2365, opened 2020-04-10
Consolidate cancel and timeout logic

afrittoli:
Cancel and timeout do very similar things when they happen: they
update the status of the taskrun, set the completion time and try
and delete the pod.

Today this is done for the two cases in different places, the code
structured differently and the behaviour slightly different:
- log levels of the messages are different
- cancel does not set the completion time
- cancel does not check if the error on pod deletion is a NotFound

This commit introduces "HasTimedOut" to tasktun_types, which
matches what "IsCancelled" does. It introduces a "killTaskRun"
function that can be used by both cancel and timeout, with the
only different being the "Reason" and termination message.
The timeout_check module is not necessary anymore.

The check for IsCancelled and HasTimedOut are move out of
"reconcile" into "Reconcile", so that now "Reconcile" checks:
- HasStarted
- isDone
- IsCancelled
- HasTimedOut
and finally, if applicable, it invokes "reconcile".

# Changes

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```


pkg/reconciler/taskrun/taskrun.go
```diff
@@ -168,74 +191,15 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 	return multierror.Append(merr, c.updateStatusLabelsAndAnnotations(tr, original)).ErrorOrNil()
 }
 
-func (c *Reconciler) updateStatusLabelsAndAnnotations(tr, original *v1alpha1.TaskRun) error {
```
afrittoli, line 171: This just moved down, no changes

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -465,67 +498,38 @@ func (c *Reconciler) handlePodCreationError(tr *v1alpha1.TaskRun, err error) {
 	c.Logger.Errorf("Failed to create build pod for task %q: %v", tr.Name, err)
 }
 
-func updateTaskRunResourceResult(taskRun *v1alpha1.TaskRun, podStatus corev1.PodStatus) error {
```
afrittoli, line 468: This just moved down - no changes.
I consolidated all func (c *Reconciler) together.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -159,6 +160,28 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 
 		return merr.ErrorOrNil()
 	}
+
+	// If the TaskRun is cancelled, kill resources and update status
+	if tr.IsCancelled() {
+		before := tr.Status.GetCondition(apis.ConditionSucceeded)
+		message := fmt.Sprintf("TaskRun %q was cancelled", tr.Name)
+		err := c.failTaskRun(tr, "TaskRunCancelled", message)
```
sbwsg, line 168: nit: suggest using the constant `TaskRunSpecStatusCancelled` here instead of literal string `"TaskRunCancelled"`.  Or putting it into a named constant if you want to keep `task.spec.status` conceptually separate from `task.status`.

afrittoli: Good idea, I'll fix that, thank you

afrittoli:
I'm hitting two data races in the unit tests - not tests that I've touched - but perhaps the new logic exposes or creates a new issue.

pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -120,6 +121,17 @@ func (trs *TaskRunStatus) MarkResourceNotConvertible(err *CannotConvertError) {
 	})
 }
 
+// MarkResourceFailed sets the ConditionSucceeded condition to ConditionFalse
+// based on an error that occurred and a reason
+func (trs *TaskRunStatus) MarkResourceFailed(reason string, err error) {
```
afrittoli, line 126: Not sure if this should be in alpha too...

vdemeester:
```
// TaskRunStatus defines the observed state of TaskRun
type TaskRunStatus = v1beta1.TaskRunStatus
```

It doesn't as it's just an alias in `v1alpha1` :wink:

afrittoli:
The failure in the integration tests is on pipeline-task retry. 
It doesn't seem to be related to this patch and I've not been able to reproduce the issue locally, so retesting. I wonder where the extra pod comes from though, there may be a race condition somewhere.

pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -333,6 +349,28 @@ func (tr *TaskRun) IsCancelled() bool {
...
+	timeout := tr.GetTimeout()
+	// If timeout is set to 0 or defaulted to 0, there is no timeout.
+	if timeout == apisconfig.NoTimeoutDuration {
+		return false
+	}
+	runtime := time.Since(tr.Status.StartTime.Time)
+	return runtime > timeout
+}
+
+func (tr *TaskRun) GetTimeout() time.Duration {
```
bobcatfish, line 366: hey @afrittoli are you sure we need this function? looking at the taskrun defaulting logic it looks like we default the value of the timeout:

https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1beta1/taskrun_defaults.go#L47

also we don't have the same function for pipelineruns, and finally finally `apisconfig.DefaultTimeoutMinutes` is a default value afaik; but the user can actually override it with a config map so we probably want to use that instead (there is some similar logic in the timeout logic that i think is being overly cautious - its always hard to know when you can rely on the defaulting to be called)

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -465,67 +470,33 @@ func (c *Reconciler) handlePodCreationError(tr *v1alpha1.TaskRun, err error) {
...
 	}
-	if !reflect.DeepEqual(tr.ObjectMeta.Labels, newTr.ObjectMeta.Labels) || !reflect.DeepEqual(tr.ObjectMeta.Annotations, newTr.ObjectMeta.Annotations) {
-		newTr.ObjectMeta.Labels = tr.ObjectMeta.Labels
-		newTr.ObjectMeta.Annotations = tr.ObjectMeta.Annotations
-		return c.PipelineClientSet.TektonV1alpha1().TaskRuns(tr.Namespace).Update(newTr)
+
+	// tr.Status.PodName will be empty if the pod was never successfully created. This condition
+	// can be reached, for example, by the pod never being schedulable due to limits imposed by
+	// a namespace's ResourceQuota.
+	err := c.KubeClientSet.CoreV1().Pods(tr.Namespace).Delete(tr.Status.PodName, &metav1.DeleteOptions{})
```
danielhelfand, line 493: I am wondering if some kind of if statement should have been placed here to guard against scenarios when a TaskRun reason is `v1beta1.TaskRunReasonTimedOut` or `v1beta1.TaskRunCancelled`. As of now, this is deleting pods for TaskRuns that have timed out or been cancelled, which amongst other things, deletes the TaskRun logs.

PR #2369, opened 2020-04-10
PipelineRun Cancellation is not working

sbwsg:
# Changes

Cancelling a PipelineRun is supposed to also cancel the TaskRuns spawned by that PipelineRun. The way that we do this is to issue `Update` and `UpdateStatus` calls on each TaskRun. Unfortunately this can (and does) fail often because modifications to the TaskRun race each other.  This has resulted in many many failed integration tests, giving the PipelineRun cancellation e2e tests the appearance of being flakey.  In fact they were actually catching real problems!

This commit updates the PipelineRun reconciler's behaviour to PATCH TaskRuns associated with a PipelineRun. This updates the TaskRun's `spec.status` regardless of its current resourceVersion / generation.

A similar related issue was happening in the PipelineRun cancel test itself. When submitting the cancellation status to the test's PipelineRun the API server was sometimes rejecting that update and failing the test. This has also been replaced with a PATCH.

The PipelineRun cancellation e2e tests now appear to be passing consistently.

Co-authored by @bobcatfish 

# Submitter Checklist

review them:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	golang.org/x/sys v0.0.0-20200212091648-12a6c2dcc1e4 // indirect
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
 	golang.org/x/tools v0.0.0-20200214144324-88be01311a71 // indirect
-	gomodules.xyz/jsonpatch/v2 v2.1.0 // indirect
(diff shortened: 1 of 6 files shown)


vincent-pli:
```
This updates the TaskRun's spec.status regardless of its current resourceVersion / generation.
```
thanks.
/lgtm

PR #2383, opened 2020-04-13
Update API compatibility policy for beta 📝

### LINKED ISSUE #1030: Audit and document "paths" in Tekton Pipelines API
# Expected Behavior

1. It should be clear to users of Tekton Pipelines what special directories will be available to executing containers, what the contents of those directories are, and what the consequences are of editing them.
2. It should be clear to contributors as well, and all directories should be defined in one location in the code

# Actual Behavior

There are a number of special directories, but I'm not sure how many or how they relate to each other. There is no one place in the docs to find this info, nor one place in the code (paths like `/builder` and `/workspace` are often hardcoded directly).

# Additional Info

* These directories should be considered part of the Tekton Pipelines API (https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md)
* We may want to consider randomizing some of these dir names (e.g. in case someone tries to use a container that has a `/workspace` already)
* While doing this investigation we may discover that we want to actually _change_ some of these paths

bobcatfish:
# Changes

Update our API compatibility policy as discussed and outlined in the
beta policy and plan doc (https://docs.google.com/document/d/1H8I2Rk4kLdQaR4mV0A71Qbk-1FxXFrmvisEAjLKT6H0/edit#
which is visible to members of
[the tekton-dev mailing
list](https://github.com/tektoncd/community/blob/master/contact.md#mailing-list)).

As a bonus, since we consider the directories used by Tekton as part of
our API, this also fixes #1030.

PTAL @vdemeester @sbwsg @afrittoli @ImJasonH @abayer @skaegi @sergetron 

# Submitter Checklist

review them:

## Reviewer Notes


docs/developers/README.md
```diff
@@ -143,16 +143,21 @@ with the binary and file(s) is mounted.
...
-- Various tools like the entrypoint are placed in `/tekton/tools`
-- The termination log message is written to `/tekton/termination`
-- Sequencing step containers is done using both `/tekton/downward/ready`
-and numbered files in `/tekton/tools`
+* `/workspace` - This directory is where [resources](#resources) and [workspaces](#workspaces)
+  are mounted.
+* `/tekton` - This directory is used for Tekton specific functionality:
+  * These folders are [part of the Tekton API](../api_compatibility_policy.md):
+    * `/tekton/results` is where [results](#results) are written to
+  * These folders are [implementation details of Tekton](developers/README.md):
```
ImJasonH, line 156: "...and users should not rely on this specific behavior, as it may change in the future."

vdemeester: We should document variables for those path — and if there is not, we should add some :upside_down_face:

sbwsg: Good thinking. `$(results.foo.path)` was added so that users don't have to hard-code `/tekton/results/foo`.

sbwsg:
I read through the policy and plan doc and noticed we don't mention the labels that are attached to the resources we create (such as `tekton.dev/task`), which are propagated to runs / pods.

Given that functionality in other Tekton projects relies on those labels (e.g. CLI using them to query for taskruns related to a task) should we start thinking about adding a policy around those too?  Or is this kind of thing captured elsewhere than the API policy?

It seems pretty likely that platform builders, devops teams, and individual users might be relying on those labels for their own tooling.

Totally understand if this is an out-of-scope discussion for this PR specifically but wondering if it's worth following up on.

vdemeester:
I think it is, at least, worth following up on

sbwsg:
I've created https://github.com/tektoncd/pipeline/issues/2388 to discuss the label policy issue further.

vdemeester:
@bobcatfish ping :angel:

bobcatfish:
I've added the caveat that users should not rely on the reserved paths and also recommended that folks use variable substitution for the paths, PTAL @vdemeester @ImJasonH @sbwsg 

(I'm assuming we can merge this without the label policy and add that in another PR)

PR #2270, opened 2020-03-23
Refactor remote.Resolver interface and implementation

sthaha:
# Changes

The existing implementation has been refactored to better integrate with
the existing logic in the Reconcilers. (This work was started in
https://github.com/tektoncd/pipeline/pull/2233 and pulled out to here).

Included in this is a move for the interface to implement `GetTask`
which is what gets passed into the TaskRun and Pipeline reconciler and
require less changes in the embedding of the task itself into the pod
spec. A couple of test helpers have been modified and a new remote test
helper is added to make pushing to images in a test easier. Tests are
included for both.

Signed-off-by: Sunil Thaha <sthaha@redhat.com>

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [🙅‍♂  ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

Double check this list of stuff that's easy to miss:

- If you are adding [a new binary/image to the `cmd` dir](../cmd), please update
  [the release Task](../tekton/publish.yaml) to build and release this image.

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

# Release Notes

```
Describe any user facing changes here, or delete this block.

Examples of user facing changes:
- API changes
- Bug fixes
- Any changes in behavior

```

PR #2395, opened 2020-04-14
Refactor to the remote image library and task references

sthaha:
This change is part of larger effort to support OCI image references in
task and pipeline refs. This intial change is a refactor of existing
logic for clarity and in preparation of adding the new API fields into
the Tekton objects.

For more details see the design: https://docs.google.com/document/d/1lXF_SvLwl6OqqGy8JbpSXRj4hWJ6CSImlxlIl4V9rnM/edit#

Principle changes:
- Refactor the remote library to operate more like an independent API.
  It allows the user to Get and List objects in a remote image. Decoupling
  the API of fetching and inspecting images allows this to be reused by
  external clients (like `tkn`).
- Create a new taskref library resolver that knows how to resolve local
  references. This mirrors the remote resolver and now establishes an
  ability for the next PR to create a factory function that assigns the
  correct resolver based on the TaskRun. It also allows us to test this
  logic by putting it into a function.
- Likewise, we have created a new pipelineref library for all the same reasons above.

Signed-off-by: Sunil Thaha <sthaha@redhat.com>

# Changes

# Submitter Checklist

review them:

- [🙅 ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

## Reviewer Notes

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```


test/remote.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	remoteimg "github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// CreateImage will push a new OCI image artifact with the provided raw data object as a layer and return the full image
+// reference with a digest to fetch the image. Key must be specified as [lowercase kind]/[object name]. The image ref
+// with a digest is returned.
+func CreateImage(ref string, objs ...runtime.Object) (string, error) {
```
vdemeester, line 37: https://github.com/tektoncd/experimental/tree/master/oci should be update accordingly I think.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -207,28 +207,18 @@ func (c *Reconciler) updateStatusLabelsAndAnnotations(tr, original *v1alpha1.Tas
 	return nil
 }
 
-func (c *Reconciler) getTaskFunc(tr *v1alpha1.TaskRun) (resources.GetTask, v1alpha1.TaskKind) {
-	var gtFunc resources.GetTask
+// TODO: Use factory func instead of hard-coding this once OCI images are supported.
+func (c *Reconciler) getTaskResolver(tr *v1alpha1.TaskRun) (*resources.LocalTaskRefResolver, v1alpha1.TaskKind) {
+	resolver := &resources.LocalTaskRefResolver{
```
vdemeester, line 212: Why not creating the `resolver` after the `if` ? the you could inline all.

```go
resolver := &resources.LocalTaskRefResolver{
	Namespace:    tr.Namespace,
	Tektonclient: c.PipelineClientSet,
	Kind: kind,
}
```

pkg/remote/resolver.go
```diff
@@ -17,22 +14,20 @@ limitations under the License.
...
-	return OCIResolver{
-		imageReference: imageReference,
-		keychainProvider: func() (authn.Keychain, error) {
-			return k8schain.NewInCluster(k8schain.Options{ServiceAccountName: serviceAccountName})
-		},
-	}
+// Resolver defines a generic API to retrieve Tekton resources from remote locations. It allows 2 principle operations:
+//   - List:     retrieve a flat set of Tekton objects in this remote location
+//   - Get:      retrieves a specific object with the given Kind and name.
+type Resolver interface {
```
vdemeester, line 30: This interface would be better defined where we use it (and we *might* not even need to be exported in that case).
One reason for this is the dependencies it brings if we want to use it — aka any package the `pkg/remote` package depends on.

This means at least:
```
	"github.com/google/go-containerregistry/pkg/authn"
	imgname "github.com/google/go-containerregistry/pkg/name"
	v1 "github.com/google/go-containerregistry/pkg/v1"
	"github.com/google/go-containerregistry/pkg/v1/remote"
	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned/scheme"
	"k8s.io/apimachinery/pkg/runtime"
```

- Using this interface *shouldn't* require any dependency on `go-containerregistry` (using an impl. might).
- I wonder about the `runtime.Object` return case too :thinking:

test/remote.go
pierretasci: The goal is to abandon the experimental tool so we can go in and remove it.

pkg/reconciler/taskrun/taskrun.go
pierretasci: True. This function is actually being deleted in the next PR. Both are very long and do different things so we wanted to split them up. We can still inline it if you want or we can wait until the next one.

test/remote.go
vdemeester: The `experimental` tool will still have value for a while, to be able to test this feature we will need to be able to create such images, having the `oci` tool in experimental is the best way to achieve that while we work on this.

To re-iterate a little bit what `tektoncd/experimental` is for : quicker *dev to in `master`* cycle. It's way easier to iterate quickly in experimental, demo it and later on integrating it in the other projects (here, on cli, …)

vdemeester:
Quick question @sthaha @pierretasci, how does it relates to 
- https://github.com/tektoncd/pipeline/pull/2270
- https://github.com/tektoncd/pipeline/pull/2233

If it carry / override those previous PRs, we should close the previous PR :angel: :pray:

pierretasci:
> Quick question @sthaha @pierretasci, how does it relates to
> 
> * #2270
> * #2233
> 
> If it carry / override those previous PRs, we should close the previous PR 👼 🙏

You are correct, those are abandoned for this new path. I will close them.

test/remote.go
pierretasci: Discussed offline. Our hope is to get in a "spec" document in the next PR to act as the stopgap so users will be able to create their own solution while waiting for the official Tekton solution.

pkg/remote/resolver.go
pierretasci: Re: `runtime.Object` it was the most generic, universal object I could think of to return either a Pipeline, Task, or ClusterTask out of the same API.

On the larger question of where to put the interface. I was thinking it would make sense to co-locate it with the definitions since that is where users would expect it to be. You are correct that it would require bringing in the container dependencies and if we add a git implementation, git dependencies as well. I guess we could put the `resolver` interface under the `remote` package to centralize it and then create sub-packages with each implementation so that you only import the ones you care about? Does that make sense?

vdemeester: > On the larger question of where to put the interface. I was thinking it would make sense to co-locate it with the definitions since that is where users would expect it to be. You are correct that it would require bringing in the container dependencies and if we add a git implementation, git dependencies as well. I guess we could put the `resolver` interface under the `remote` package to centralize it and then create sub-packages with each implementation so that you only import the ones you care about? Does that make sense?

I would work — it's not the best from my point of view (aka I am not a huge fan of exported interface), but I think it's a good start :wink:

pierretasci: Done

pierretasci:
* friendly ping * this is ready for re-review

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -798,19 +798,20 @@ func TestReconcileOnCompletedPipelineRun(t *testing.T) {
...
+	if len(clients.Pipeline.Actions()) != 2 {
 		t.Fatalf("Expected client to have updated the TaskRun status for a completed PipelineRun, but it did not")
 	}
 
-	actual := clients.Pipeline.Actions()[0].(ktesting.UpdateAction).GetObject().(*v1alpha1.PipelineRun)
+	actual := clients.Pipeline.Actions()[1].(ktesting.UpdateAction).GetObject().(*v1alpha1.PipelineRun)
 	if actual == nil {
 		t.Errorf("Expected a PipelineRun to be updated, but it wasn't.")
 	}
+	t.Log(clients.Pipeline.Actions())
```
ImJasonH, line 809: Is this logging statement necessary?

pkg/reconciler/pipelinerun/resources/pipelineref.go
```diff
@@ -0,0 +1,41 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"fmt"
+
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
```
ImJasonH, line 22: Will this still work if the PipelineRun is a v1beta1 PipelineRun object?

pkg/reconciler/taskrun/resources/taskref.go
```diff
@@ -0,0 +1,50 @@
...
+	Namespace    string
+	Kind         v1alpha1.TaskKind
+	Tektonclient clientset.Interface
+}
+
+// GetTask will resolve either a Task or ClusterTask from the local cluster using a versioned Tekton client. It will
+// return an error if it can't find an appropriate Task for any reason.
+func (l *LocalTaskRefResolver) GetTask(name string) (v1alpha1.TaskInterface, error) {
+	if l.Kind == v1alpha1.ClusterTaskKind {
+		task, err := l.Tektonclient.TektonV1alpha1().ClusterTasks().Get(name, metav1.GetOptions{})
```
ImJasonH, line 38: style nit: This can just be `return l.Tektonclient...ClusterTasks().Get(...)`

ImJasonH:
/lgtm

I'll leave it to @vdemeester to approve

pkg/reconciler/pipelinerun/resources/pipelineref.go
vdemeester: Right, with https://github.com/tektoncd/pipeline/pull/2413 and https://github.com/tektoncd/pipeline/issues/2410, we will be able to use `v1beta1` everywhere :wink:

vdemeester:
arf, needs a rebase @sthaha @pierretasci :sweat_smile:

PR #2397, opened 2020-04-15
Update install.md to add --cluster-version=latest

hamelsmu:
# Changes

added flag `--cluster-version=latest` so that people following the tutorial will not get the error that the latest version of Tekton is not compatible with whatever version of Kubernetes (for me it was v1.14 that was installed) that is installed by Default

## Reviewer Notes

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```

cc: @jlewi

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -26,7 +26,7 @@ This guide explains how to install Tekton Pipelines. It covers the following top
    ```bash
    #Example command for creating a cluster on GKE
    gcloud container clusters create $CLUSTER_NAME \
-     --zone=$CLUSTER_ZONE --cluster-version=latest
+     --zone=$CLUSTER_ZONE --cluster-version=1.15.11-gke.5
    ```
 
 3. Grant `cluster-admin` permissions to the current user:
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -26,7 +26,7 @@ This guide explains how to install Tekton Pipelines. It covers the following top
    ```bash
    #Example command for creating a cluster on GKE
    gcloud container clusters create $CLUSTER_NAME \
-     --zone=$CLUSTER_ZONE
+     --zone=$CLUSTER_ZONE --cluster-version=latest
```
afrittoli, line 29: Thanks for this.

It does say version 1.15 or later on L24, but it is better to have a working command here.
NIT: `latest` here will change over time, depending on what GKE runs, so perhaps we should use `--cluster-version`. On the other side a specific version may not be available anymore at some point in time, but still I slightly  prefer a specific version.

dibyom: yeah, I think a specific version would be nice!

hamelsmu: ok can someone suggest a version?

afrittoli: I would say the latest available now `1.15.11-gke.5`

hamelsmu:
@afrittoli  thanks!  I made the changes and tested that it works 🙇

PR #2413, opened 2020-04-16
Add support for conversion webhook 🍸

vdemeester:
# Changes

This adds a new webhook controller : ConversionController.
The `HubVersion` is the stored version, and `Zygotes` stores the types
of supported versions.

The CRDs are updated to setup the conversion strategy to webhook,
pointing to the tekton-pipelines-webhook.

An update in the webhook clusterrole is required for customresourcedefinitions.

Closes #2047 

/cc @sbwsg @bobcatfish @mattmoor 

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

Double check this list of stuff that's easy to miss:

- If you are adding [a new binary/image to the `cmd` dir](../cmd), please update
  [the release Task](../tekton/publish.yaml) to build and release this image.

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

# Release Notes

```
Add support for conversion webhook for tektoncd/pipeline CRDs
```

Issue #2410, opened 2020-04-16
Change the storage version to v1beta1 types


vdemeester:
Once conversion webhook are in place (see #2047), we should be able to switch to `v1beta1` to storage types (and when we go to `v1` we will be able to switch to `v1` as storage right away).

This should help cleaning up the code (both in `pkg/apis` *and* `pkg/reconcilier`) and will make the API easier to reason about (aka `v1alpha1` won't need to have `v1beta1` fields anymore).

The target for this should be one release after #2047 (so tentatively adding it to a 0.13 milestone)

/kind cleanup
/assign

PR #2421, opened 2020-04-16
Split reconcile into prepare and reconcile

afrittoli:
# Changes

The current "reconcile" function does a lot of things, which means
it's hard to test properly, it's hard to follow the logic and
it's hard to handle all possible errors and events properly.

This splits reconcile into two parts, the first one that deals
preparing and validating the taskrun and all the resources it
depends on, the second one that actually reconciles the TaskRun
by creating the pod if required and updating the TaskRun from
the pod.

This adds events that were missing for error situation that
happens during preparation and validation.

# Submitter Checklist

review them:

- [WIP] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)

## Reviewer Notes

# Release Notes

```
The TaskRun controller now emits events also when error occurs during the validation of the TaskRun or the resources it depends on.
```

### DIFF
--- a/pkg/reconciler/reconciler.go
+++ b/pkg/reconciler/reconciler.go
@@ -97,7 +97,13 @@ func NewBase(opt Options, controllerAgentName string, images pipeline.Images) *B
 	if recorder == nil {
 		// Create event broadcaster
 		logger.Debug("Creating event broadcaster")
-		eventBroadcaster := record.NewBroadcaster()
+
+		correlatorOptions := record.CorrelatorOptions{
+			// The default burst size is 25
+			BurstSize: 50,
+			QPS:       1,
+		}
+		eventBroadcaster := record.NewBroadcasterWithCorrelatorOptions(correlatorOptions)
 		eventBroadcaster.StartLogging(logger.Named("event-broadcaster").Infof)
 		eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{Interface: opt.KubeClientSet.CoreV1().Events("")})
 
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -79,9 +79,6 @@ var _ controller.Reconciler = (*Reconciler)(nil)
 // converge the two. It then updates the Status block of the Task Run
 // resource with the current status of the resource.
 func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
-	// In case of reconcile errors, we store the error in a multierror, attempt
-	// to update, and return the original error combined with any update error
-	var merr error
 
 	// Convert the namespace/name string into a distinct namespace and name
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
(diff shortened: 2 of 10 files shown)


pkg/reconciler/taskrun/taskrun.go
```diff
@@ -182,38 +179,50 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
 	}
-	return multierror.Append(merr, c.updateStatusLabelsAndAnnotations(tr, original)).ErrorOrNil()
+	return multierror.Append(err, c.updateStatusLabelsAndAnnotations(tr, original)).ErrorOrNil()
 }
 
-func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun) error {
+// TODO(afrittoli) for now we return spec and resources. In future we should store
+// them in the TaskRun.Status so we don't need to re-run `prepare` at every
+// reconcile.
+func (c *Reconciler) prepare(ctx context.Context, tr *v1alpha1.TaskRun) (*v1alpha1.TaskSpec, *resources.ResolvedTaskResources, error) {
```
afrittoli, line 204: Should these be *v1beta1 somehow?

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -275,8 +284,15 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun) error
 	}
 	cloudevent.InitializeCloudEvents(tr, prs)
 
+	return taskSpec, rtr, nil
+}
+
+func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun,
```
afrittoli, line 290: Should these be *v1beta1 somehow? After `prepare` they should be converted to beta?

Also I think I should move the emit event bit out of `reconcile` to `Reconcile`, so that all events are generated there.

vdemeester: Nope. The are converted to `beta` and then to `alpha` to make sure they are in `beta` shape — but the internal (and stored) representation is still `v1alpha1`. But the `v1alpha1` struct you have after `prepare` are **sure** to be in the `v1beta1` shape, so you can/should/have-to use *only* `v1beta1` fields :angel: 

I know it's a bit confusing right now. This will change with #2413 and #2410 :angel:

afrittoli: ack, thanks

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -182,38 +179,60 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 		return multierror.Append(err, c.updateStatusLabelsAndAnnotations(tr, original)).ErrorOrNil()
 	}
 
+	// prepare fetches all required resources, validates them together with the
```
sbwsg, line 182: suggest this comment describing `prepare`'s behaviour be attached to the `func prepare(...)` on line 214.  But I do think it makes sense to leave a comment here about errors coming back from prepare being considered permanent.

pkg/reconciler/taskrun/taskrun.go
sbwsg: agree about event emitting being centralized into one area of the file 👍 

Also, could we add a comment about the conditions under which an error _should_ be returned by `func reconcile(...)`?  So often I see errors caught in the code but we `return nil` regardless.  I've never totally understood the logic of when to return an error from reconcile and when not to.

afrittoli: I'm trying to make the logic about errors and events more consistent in the controller, so I will add some comments about this. 
One thing that I'd like to do eventually is also to start using `PermanentError`s, from knative/pkg. When `Reconcile` returns a `PermanentError` the key is not re-queued.

pkg/reconciler/taskrun/taskrun.go
afrittoli: Good point, will do

afrittoli:
I cannot reproduce any of the integration test failures on my test cluster, even after multiple runs. I think that something in the logic that waits for events is not going well, not waiting long enough or perhaps starting to wait too soon.

afrittoli:
> I cannot reproduce any of the integration test failures on my test cluster, even after multiple runs. I think that something in the logic that waits for events is not going well, not waiting long enough or perhaps starting to wait too soon.

I changed the cleanup logic so that namespaces are not cleaned up for failed tests. 
I can see all taskrun events, but the pipelinerun completion event is missing - I did not change that in my PR though - and I cannot reproduce the issue locally - with the only difference being that I only run one test at the time in my local environment.

afrittoli:
I was finally able to reproduce the CI failures on my cluster. 
From local debugging, I see that after exactly 127 events, the API stops reporting events to clients. After 1h exactly, the initial 127 events expire, and magically the remaining events are returned to the client. I'm trying to find out if it is possible to change this behavior, else we might have to change the test to validate events only in unit tests.

afrittoli:
> I was finally able to reproduce the CI failures on my cluster.
> From local debugging, I see that after exactly 127 events, the API stops reporting events to clients. After 1h exactly, the initial 127 events expire, and magically the remaining events are returned to the client. I'm trying to find out if it is possible to change this behavior, else we might have to change the test to validate events only in unit tests.

We'll have to use the `NewBroadcasterWithCorrelatorOptions` in the reconciler to tweak burst/qps options for the event broadcaster.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -445,33 +467,31 @@ func (c *Reconciler) getTaskFunc(tr *v1alpha1.TaskRun) (resources.GetTask, v1alp
...
+			Type:    apis.ConditionSucceeded,
+			Status:  corev1.ConditionUnknown,
+			Reason:  podconvert.ReasonExceededResourceQuota,
+			Message: fmt.Sprintf("%s: %v", msg, err),
+		})
 	} else {
-		succeededStatus = corev1.ConditionFalse
-		reason = podconvert.ReasonCouldntGetTask
+		// The pod creation failed, not because of quota issues. The most likely
+		// reason is that something is wrong with the spec of the Task
```
dibyom, line 485: > // reason is that something is wrong with the spec of the Task

shouldn't that be caught by validation beforehand?

test/init_test.go
```diff
@@ -67,6 +67,16 @@ func setup(t *testing.T, fn ...func(*testing.T, *clients, string)) (*clients, st
 	return c, namespace
 }
 
+func setupStatic(t *testing.T) (*clients, string) {
+	t.Helper()
+	namespace := "faninfanout"
```
dibyom, line 72: I'm guessing this is for debugging only?

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -182,38 +179,60 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+
+	// Emit events (only when ConditionSucceeded was changed)
+	after := tr.Status.GetCondition(apis.ConditionSucceeded)
+	reconciler.EmitEvent(c.Recorder, before, after, tr)
+
+	return multierror.Append(err, c.updateStatusLabelsAndAnnotations(tr, original)).ErrorOrNil()
 }
 
-func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun) error {
+// TODO(afrittoli) for now we return spec and resources. In future we should store
```
dibyom, line 211: Would be nice for the TODO to reference a Github issue instead!

pkg/reconciler/taskrun/taskrun.go
afrittoli: Yeah, I did not change this logic here, and I had the same question.
My guess since we can have pod templates in taskrun, if something goes wrong in the pod creation if may be due to that... but tbh I don't think we can really know for sure.

test/init_test.go
afrittoli: upz, forgot to remove this, thanks!

pkg/reconciler/taskrun/taskrun.go
afrittoli: I will add a link

pkg/reconciler/reconciler.go
```diff
@@ -97,7 +97,13 @@ func NewBase(opt Options, controllerAgentName string, images pipeline.Images) *B
 	if recorder == nil {
 		// Create event broadcaster
 		logger.Debug("Creating event broadcaster")
-		eventBroadcaster := record.NewBroadcaster()
+
+		correlatorOptions := record.CorrelatorOptions{
+			// The default burst size is 25
+			BurstSize: 50,
```
dibyom, line 103: Probably fine for now but should be configurable right?

pkg/reconciler/reconciler.go
```diff
@@ -97,7 +97,13 @@ func NewBase(opt Options, controllerAgentName string, images pipeline.Images) *B
...
+			QPS:       1,
```
dibyom, line 104: Is this the number of queries per second that we can emit? I couldn't figure out after reading the godoc for this

pkg/reconciler/reconciler.go
afrittoli: Yes, I'll follow up on this

pkg/reconciler/reconciler.go
afrittoli: Yes, the way I understood this is the QPS we allow after a burst

Issue #2416, opened 2020-04-16
Unexpected start times reported for skipped steps


AlanGreene:
# Expected Behavior
Step start times should be sequential regardless of whether they actually ran or not.

# Actual Behavior
If a TaskRun fails, any remaining skipped steps retain their original container start time instead of being updated by the entrypoint binary.

# Steps to Reproduce the Problem

```
kubectl create -f - << EOF
---
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  generateName: test-start-time-correct-
spec:
  taskSpec:
    steps:
    - image: ubuntu
      script: |
        #!/usr/bin/env bash
        sleep 3 && echo step 1
    - image: ubuntu
      script: |
        #!/usr/bin/env bash
        sleep 3 && echo step 2
    - image: ubuntu
      script: |
        #!/usr/bin/env bash
        sleep 3 && echo step 3
    - image: ubuntu
      script: |
        #!/usr/bin/env bash
        sleep 3 && echo step 4
---
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  generateName: test-start-time-wrong-
spec:
  taskSpec:
    steps:
    - image: ubuntu
      script: |
        #!/usr/bin/env bash
        sleep 3 && echo step 1
    - image: ubuntu
      script: |
        #!/usr/bin/env bash
        sleep 3 && echo step 2
        false
    - image: ubuntu
      script: |
        #!/usr/bin/env bash
        sleep 3 && echo step 3
    - image: ubuntu
      script: |
        #!/usr/bin/env bash
        sleep 3 && echo step 4
EOF
```

The step start times for `test-start-time-correct-` are recorded and reported correctly, reflecting the time the step actually executed.

The step start times for the first 2 steps in `test-start-time-wrong-` are correct, but the last 2 skipped steps still report their original container start times. This means that if I order the step status by start time they appear before the steps that actually ran.

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  Client Version: v1.16.3
  Server Version: v1.15.5
  ```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

  v0.11.1

/kind bug

Related issue: https://github.com/tektoncd/pipeline/issues/1512
This fixed the start times for steps that are actually executed.

[...]

PR #2429, opened 2020-04-17
Set "startedAt" in skipped case

vincent-pli:
Fix issue:  #2416 

The root cause is in `skipped` case, `entrypoint` do not write `startedAt` to `Termination file`.
That's means when `reconcile` of `taskrun` try to update `startedAt`, they found nothing, and the result is the `startedAt` will keep as the `startedAt` in `pod`.
# Changes

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```

### DIFF
--- a/pkg/entrypoint/entrypointer.go
+++ b/pkg/entrypoint/entrypointer.go
@@ -82,15 +82,25 @@ type PostWriter interface {
 func (e Entrypointer) Go() error {
 	prod, _ := zap.NewProduction()
 	logger := prod.Sugar()
+
+	output := []v1alpha1.PipelineResourceResult{}
 	defer func() {
+		if wErr := termination.WriteMessage(e.TerminationPath, output); wErr != nil {
+			logger.Fatalf("Error while writing message: %s", wErr)
+		}
 		_ = logger.Sync()
 	}()
-	output := []v1alpha1.PipelineResourceResult{}
+
 	for _, f := range e.WaitFiles {
 		if err := e.Waiter.Wait(f, e.WaitFileContent); err != nil {
 			// An error happened while waiting, so we bail
 			// *but* we write postfile to make next steps bail too.
 			e.WritePostFile(e.PostFile, err)
+			output = append(output, v1alpha1.PipelineResourceResult{
+				Key:   "StartedAt",
+				Value: time.Now().Format(time.RFC3339),
+			})
+
 			return err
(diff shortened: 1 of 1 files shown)


sbwsg:
@vincent-pli Could you update your commit message to match the PR message?  It makes browsing our git history much more helpful in the future.

pkg/entrypoint/entrypointer.go
```diff
@@ -91,6 +91,14 @@ func (e Entrypointer) Go() error {
 			// An error happened while waiting, so we bail
 			// *but* we write postfile to make next steps bail too.
 			e.WritePostFile(e.PostFile, err)
+			output = append(output, v1alpha1.PipelineResourceResult{
+				Key:   "StartedAt",
+				Value: time.Now().Format(time.RFC3339),
+			})
+
+			if wErr := termination.WriteMessage(e.TerminationPath, output); wErr != nil {
```
sbwsg, line 99: could we do this once in a `defer`-ed func instead of duplicating it?

pkg/entrypoint/entrypointer.go
```diff
@@ -91,6 +91,14 @@ func (e Entrypointer) Go() error {
 			// An error happened while waiting, so we bail
 			// *but* we write postfile to make next steps bail too.
 			e.WritePostFile(e.PostFile, err)
+			output = append(output, v1alpha1.PipelineResourceResult{
```
sbwsg, line 94: how about just doing this once, when `output` is initialized?

```go
// line 88
output := []v1alpha1.PipelineResourceResult{
  {
    Key: "StartedAt",
    Value: time.Now().Format(time.RFC3339),
  }
}
```

Could also remove lines 109-112 then as well.

vincent-pli:
@sbwsg 
Commit message is update, thanks

pkg/entrypoint/entrypointer.go
vincent-pli: Done, thanks

pkg/entrypoint/entrypointer.go
vincent-pli: I think we can not initialize it at one place, the `time.Now()` should after `wait` "wait files",
and the `return`s are in different code block, so the `time.Now()` is different.

PR #2437, opened 2020-04-18
WIP - implementing finally at the pipeline level 🤞🏼💥😱

pritidesai:
# Changes

We can now specify a list of tasks that needs to be executed just before
pipeline exits (either after finishing all non-final tasks successfully or after
a single failure)

Most common final tasks could be (1) report test results, (2) cleanup cluster resources, etc

```
kind: Pipeline
metadata:
  name: pipeline-with-final-tasks
spec:
  tasks:
    - name: pre-work
      taskRef:
        Name: some-pre-work
    - name: unit-test
      taskRef:
        Name: run-unit-test
      runAfter:
        - pre-work
    - name: integration-test
      taskRef:
        Name: run-integration-test
      runAfter:
        - unit-test
  finally:
    - name: cleanup-test
      taskRef:
        Name: cleanup-cluster
    - name: report-results
      taskRef:
        Name: report-test-results
```

Design doc: https://docs.google.com/document/d/1lxpYQHppiWOxsn4arqbwAFDo4T0-LCqpNa6p-TJdHrw/edit#

Part of https://github.com/tektoncd/pipeline/issues/1684
Fixes https://github.com/tektoncd/pipeline/issues/2446

Depends on: https://github.com/tektoncd/pipeline/pull/2504

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Users can now specify Tasks within a Pipeline that will always execute, even if Tasks fail, via the new `finally` clause
```

### DIFF
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -21,6 +21,7 @@ weight: 3
   - [Configuring execution results at the `Pipeline` level](#configuring-execution-results-at-the-pipeline-level)
   - [Configuring the `Task` execution order](#configuring-the-task-execution-order)
   - [Adding a description](#adding-a-description)
+  - [Adding `Finally` to the `Pipeline`](#adding-finally-to-the-pipeline)
   - [Code examples](#code-examples)
 
 ## Overview
@@ -515,6 +516,203 @@ In particular:
 ## Adding a description
 
 The `description` field is an optional field and can be used to provide description of the `Pipeline`.
+
+## Adding `Finally` to the `Pipeline`
+
+You can specify a list of one or more final tasks under `finally` section. Final tasks are guaranteed to be executed
+after all `PipelineTasks` under `tasks` have completed regardless of success or error. Final tasks are very similar to
+`PipelineTasks` under `tasks` section and follow the same syntax. Each final task must have a
+[valid](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names) `name` and a `taskRef` or
+`taskSpec`. For example:
+
+```yaml
+spec:
+  tasks:
+    - name: tests
+      taskRef:
+        Name: integration-test
+  finally:
+    - name: cleanup-test
+      taskRef:
+        Name: cleanup
+```
+
+### Specifying `Resources` in Final Tasks
+
+Similar to `tasks`, you can use [PipelineResources](#specifying-resources) as inputs and outputs for
+final tasks in the Pipeline. The only difference here is, final tasks with an input resource can have a `from` clause
+but that resource has to rely on output of a `PipelineTask` from `tasks` section. For example:
+
+```yaml
+spec:
+  tasks:
+    - name: tests
+      taskRef:
+        Name: integration-test
+      resources:
+        inputs:
+          - name: source
+            resource: tektoncd-pipeline-repo
+        outputs:
+          - name: workspace
+            resource: my-repo
+  finally:
+    - name: clear-workspace
+      taskRef:
+        Name: clear-workspace
+      resources:
+        inputs:
+          - name: workspace
+            resource: my-repo
+            from:
+              - tests
+```
+
+### Specifying `Workspaces` in Final Tasks
+
+It is possible that the final tasks might require access to the same workspaces which `PipelineTasks` might have utilized
+e.g. a mount point for credentials held in Secrets. To support that requirement, you can specify one or more
+`Workspaces` in the `workspaces` field for the final tasks similar to `tasks`.
+
+```yaml
+spec:
+  resources:
+    - name: app-git
+      type: git
+  workspaces:
+    - name: shared-workspace
+  tasks:
+    - name: clone-app-source
+      taskRef:
+        name: clone-app-repo-to-workspace
+      workspaces:
+        - name: shared-workspace
+          workspace: shared-workspace
+      resources:
+        inputs:
+          - name: app-git
+            resource: app-git
+  finally:
+    - name: cleanup-workspace
+      taskRef:
+        name: cleanup-workspace
+      workspaces:
+        - name: shared-workspace
+          workspace: shared-workspace
+```
+
+### Specifying `Parameters` in Final Tasks
+
+Again, similar to `tasks`, you can specify [`Parameters`](tasks.md#specifying-parameters):
+
+```yaml
+spec:
+  tasks:
+    - name: tests
+      taskRef:
+        Name: integration-test
+  finally:
+    - name: report-results
+      taskRef:
+        Name: report-results
+      params:
+        - name: url
+          value: $(params.url)
+```
+
+### Cannot configure the Final Task execution order
+
+It's not possible to configure or modify the execution order of the final tasks. Unlike `Tasks` in a `Pipeline`,
+all final tasks run simultaneously and starts executing once all `PipelineTasks` under `tasks` have settled which means
+no `runAfter` can be specified in final tasks.
+
+### Cannot specify execution `Conditions` in Final Tasks
+
+`Tasks` in a `Pipeline` can be configured to run only if some conditions are satisfied using `conditions`. But the
+final tasks are guaranteed to be executed after all `PipelineTasks` therefore no `conditions` can be specified in
+final tasks.
+
+### `PipelineRun` Status with `finally`
+
+With `finally`, `PipelineRun` status is calculated based on `PipelineTasks` under `tasks` section and final tasks.
+
+Without `finally`:
+
+| `PipelineTasks` under `tasks` | `PipelineRun` status | Reason |
+| ----------------------------- | -------------------- | ------ |
+| all `PipelineTasks` successful | `true` | `Succeeded` |
+| one or more `PipelineTasks` skipped and rest successful | `true` | `Completed` |
+| single failure of `PipelineTask` | `false` | `failed` |
+
+With `finally`:
+
+| `PipelineTasks` under `tasks` | Final Tasks | `PipelineRun` status | Reason |
+| ----------------------------- | ----------- | -------------------- | ------ |
+| all `PipelineTask` successful | all final tasks successful | `true` | `Succeeded` |
+| all `PipelineTask` successful | one or more failure of final tasks | `false` | `Failed` |
+| one or more `PipelineTask` skipped and rest successful | all final tasks successful | `true` | `Completed` |
+| one or more `PipelineTask` skipped and rest successful | one or more failure of final tasks | `false` | `Failed` |
+| single failure of `PipelineTask` | all final tasks successful | `false` | `failed` |
+| single failure of `PipelineTask` | one or more failure of final tasks | `false` | `failed` |
+
+### Known Limitations
+
+#### Cannot configure `Task` execution results with `finally`
+
+Final tasks can not be configured to consume `Results` of `PipelineTask` from `tasks` section i.e. the following
+example is not supported right now but we are working on adding support for the same.
+
+```yaml
+spec:
+  tasks:
+    - name: count-comments-before
+      taskRef:
+        Name: count-comments
+    - name: add-comment
+      taskRef:
+        Name: add-comment
+    - name: count-comments-after
+      taskRef:
+        Name: count-comments
+  finally:
+    - name: check-count
+      taskRef:
+        Name: check-count
+      params:
+        - name: before-count
+          value: $(tasks.count-comments-before.results.count)
+        - name: after-count
+          value: $(tasks.count-comments-after.results.count)
+```
+
+In this example, container logs for `check-count` might have something like this: 
+
+```
+/tekton/scripts/script-0-zsg5v: 3: tasks.sum-inputs.results.sum-result: not found
+/tekton/scripts/script-0-2g2dr: 3: tasks.count-comments-before.results.count: not found
+```
+
+#### Cannot configure `Pipeline` result with `finally`
+
+Final tasks can emit `Results` but results emitted from the final tasks can not be configured in the
+[Pipeline Results](#configuring-execution-results-at-the-pipeline-level). We are working on adding support for this.
+
+```yaml
+  results:
+    - name: comment-count-validate
+      value: $(finally.check-count.results.comment-count-validate)
+```
+
+In this example, `PipelineResults` is set to:
+
+```
+"pipelineResults": [
+  {
+    "name": "comment-count-validate",
+    "value": "$(finally.check-count.results.comment-count-validate)"
+  }
+],
+```
 
 ## Code examples
 
--- a/examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
+++ b/examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
@@ -1,0 +1,149 @@
+# Task producing success by exiting with 0
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: successful-task
+spec:
+  steps:
+    - name: success
+      image: ubuntu
+      script: |
+        exit 0
+---
+
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: pipeline-with-final-tasks
+spec:
+  tasks:
+    - name: pre-work
+      taskRef:
+        Name: successful-task
+    - name: unit-test
+      taskRef:
+        Name: successful-task
+      runAfter:
+        - pre-work
+    - name: integration-test
+      taskRef:
+        Name: successful-task
+      runAfter:
+        - unit-test
+  finally:
+    - name: cleanup-test
+      taskRef:
+        Name: successful-task
+    - name: report-results
+      taskRef:
+        Name: successful-task
+---
+
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: pipelinerun-with-final-tasks
+spec:
+  pipelineRef:
+    name: pipeline-with-final-tasks
+---
+
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: shared-workspace
+spec:
+  resources:
+    requests:
+      storage: 16Mi
+  volumeMode: Filesystem
+  accessModes:
+    - ReadWriteOnce
+---
+
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: clone-app-repo-to-workspace
+spec:
+  workspaces:
+    - name: shared-workspace
+  resources:
+    inputs:
+      - name: app-git
+        type: git
+        targetPath: application
+  steps:
+    - name: clone-app-repo-to-workspace
+      image: ubuntu
+      script: |
+        #!/usr/bin/env bash
+        set -xe
+        cp -avr $(resources.inputs.app-git.path)/ $(workspaces.shared-workspace.path)/
+---
+
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: cleanup-workspace
+spec:
+  workspaces:
+    - name: shared-workspace
+  steps:
+    - name: cleanup-workspace
+      image: ubuntu
+      script: |
+        #!/usr/bin/env bash
+        set -xe
+        rm -rf $(workspaces.shared-workspace.path)/application/
+---
+
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: clone-into-workspace-and-cleanup-workspace
+spec:
+  resources:
+    - name: app-git
+      type: git
+  workspaces:
+    - name: shared-workspace
+  tasks:
+    - name: clone-app-source
+      taskRef:
+        name: clone-app-repo-to-workspace
+      workspaces:
+        - name: shared-workspace
+          workspace: shared-workspace
+      resources:
+        inputs:
+          - name: app-git
+            resource: app-git
+  finally:
+    - name: cleanup-workspace
+      taskRef:
+        name: cleanup-workspace
+      workspaces:
+        - name: shared-workspace
+          workspace: shared-workspace
+---
+
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: write-and-cleanup-workspace
+spec:
+  pipelineRef:
+    name: clone-into-workspace-and-cleanup-workspace
+  workspaces:
+    - name: shared-workspace
+      persistentVolumeClaim:
+        claimName: shared-workspace
+  resources:
+    - name: app-git
+      resourceSpec:
+        type: git
+        params:
+          - name: url
+            value: https://github.com/tektoncd/pipeline.git
+---
--- a/pkg/apis/pipeline/v1alpha1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1alpha1/pipeline_validation.go
@@ -66,6 +66,16 @@ func validateDeclaredResources(ps *PipelineSpec) error {
 			}
 		}
 	}
+	for _, t := range ps.Finally {
+		if t.Resources != nil {
+			for _, input := range t.Resources.Inputs {
+				required = append(required, input.Resource)
+			}
+			for _, output := range t.Resources.Outputs {
+				required = append(required, output.Resource)
+			}
+		}
+	}
 
 	provided := make([]string, 0, len(ps.Resources))
 	for _, resource := range ps.Resources {
@@ -90,7 +100,7 @@ func isOutput(outputs []PipelineTaskOutputResource, resource string) bool {
 
 // validateFrom ensures that the `from` values make sense: that they rely on values from Tasks
 // that ran previously, and that the PipelineResource is actually an output of the Task it should come from.
-func validateFrom(tasks []PipelineTask) error {
+func validateFrom(tasks []PipelineTask, finalTasks []PipelineTask) *apis.FieldError {
 	taskOutputs := map[string][]PipelineTaskOutputResource{}
 	for _, task := range tasks {
 		var to []PipelineTaskOutputResource
@@ -114,10 +124,32 @@ func validateFrom(tasks []PipelineTask) error {
 			for _, pt := range rd.From {
 				outputs, found := taskOutputs[pt]
 				if !found {
-					return fmt.Errorf("expected resource %s to be from task %s, but task %s doesn't exist", rd.Resource, pt, pt)
+					return apis.ErrInvalidValue(fmt.Sprintf("expected resource %s to be from task %s, but task %s doesn't exist", rd.Resource, pt, pt),
+						"spec.tasks.resources.inputs.from")
 				}
 				if !isOutput(outputs, rd.Resource) {
-					return fmt.Errorf("the resource %s from %s must be an output but is an input", rd.Resource, pt)
+					return apis.ErrInvalidValue(fmt.Sprintf("the resource %s from %s must be an output but is an input", rd.Resource, pt),
+						"spec.tasks.resources.inputs.from")
+				}
+			}
+		}
+	}
+	for _, t := range finalTasks {
+		inputResources := []PipelineTaskInputResource{}
+		if t.Resources != nil {
+			inputResources = append(inputResources, t.Resources.Inputs...)
+		}
+
+		for _, rd := range inputResources {
+			for _, pt := range rd.From {
+				outputs, found := taskOutputs[pt]
+				if !found {
+					return apis.ErrInvalidValue(fmt.Sprintf("expected resource %s to be from task %s, but task %s doesn't exist under spec.tasks", rd.Resource, pt, pt),
+						"spec.finally.resources.inputs.from")
+				}
+				if !isOutput(outputs, rd.Resource) {
+					return apis.ErrInvalidValue(fmt.Sprintf("the resource %s from %s must be an output but is an input", rd.Resource, pt),
+						"spec.finally.resources.inputs.from")
 				}
 			}
 		}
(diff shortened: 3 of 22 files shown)


bobcatfish:
Hey @pritidesai ! Very excited to see this moving along!! :D :D

I'm going to update the PR description to include the release notes, design doc and related issue.

Is it possible to include the docs on how to use this in the PR? This makes reviewing the PR waaaay easier, otherwise one has to read the code to understand how this is intended to work

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -135,6 +139,35 @@ type PipelineTask struct {
 	Timeout *metav1.Duration `json:"timeout,omitempty"`
 }
 
+// FinalPipelineTask defines a final task in a Pipeline, passing inputs from both
+// Params and from the output of non-final tasks.
+type FinalPipelineTask struct {
```
bobcatfish, line 144: What are the differences between this struct and the existing PipelineStruct task? can we use the same struct but apply different validation?

bobcatfish: A bit more context: im thinking that if we use the exact same PipelineTask, but have some internal notion of whether the PipelineTask is a final task or not, they could share a lot of the same execution logic (and then youd get things like timeouts and retries for free like @skaegi is pointing out, in his words "treat it no differently from a regular Task but the reordering rules are different"

another way to do it in Go would be to have types such as:
- PipelineTask <-- which has the bare minimum fields shared by both PipelineTask (today) and FinalPipelineTask
- DAGPiplineTask (or something) <-- which also includes runAfter, etc.

pritidesai: Yup, certainly we can, the idea to create this new struct was to keep the implementation clean vs adding `if finally` at the validation level which would make it even more complicated when we add `except` 😢  Also, one more reason was to enforce the fact that `tasks` and `finally` tasks are actually different from the perspective of specifying dependencies - no `runAfter` supported under `finally`. 

After having this discussed in the API workflow as well, I will change it to use the same `PipelineTask` struct.

pritidesai: > another way to do it in Go would be to have types such as:
> 
> * PipelineTask <-- which has the bare minimum fields shared by both PipelineTask (today) and FinalPipelineTask
> * DAGPiplineTask (or something) <-- which also includes runAfter, etc.

I had thought about this but discarded the idea as it resulted in embedding structure into structure (I am fine either ways, dont have strong preference):

```
type PipelineTask struct {
    name string
}

type DAGPipelineTask struct {
    PipelineTask  //embedding
    runAfter string
}
``` 

This results in changing all of the unit tests from:

```
Tasks: []PipelineTask{{						
		Name: "task1",
```
to

```
Tasks: []DAGPipelineTask{{
   PipelineTask:
     PipelineTask{
	Name: "task1",
```

Referencing those fields (Name and etc) would still remain the same.

pritidesai: Whatever type we decide to go with, should not impact execution logic since we are adding an additional field `[FinalPipelineTask](https://github.com/tektoncd/pipeline/pull/2437/files#diff-4e301064fef70f3c0db810267c0959f2R73)` to maintain the list of final tasks which would not interfere with non-final tasks.

pritidesai: hey @bobcatfish I have listed pros and cons of refactoring `PipelineTask` [here](https://gist.github.com/pritidesai/54ea8efa60692cdc2582cfc60eb12a8b) 😱

docs/pipelines.md
```diff
@@ -516,6 +517,137 @@ In particular:
...
+        - name: after-count
+          value: $(tasks.count-comments-after.results.count)
+```
+
+In this example, final task `check-count` consumes results from two `PipelineTask` under `tasks` section,
+(1) `count-comments-before` and (2) `count-comments-after`.
+
+*Note:* Final tasks can emit `Results` but can not be consumed by any other final task nor by any `PipelineTask`
+from `tasks` section for two reasons (1) All final tasks run in parallel (2) Final tasks are executed after all
+`PipelineTask` from `tasks` section.
```
bobcatfish, line 623: 👍

docs/pipelines.md
```diff
@@ -516,6 +517,137 @@ In particular:
...
+
+```yaml
+  results:
+    - name: comment-count-validate
+      value: $(finally.check-count.results.comment-count-validate)
+```
+
+*Note:* In the above example, if adding the comment to PR fails and task `add-comment` results in `Pipeline` failure,
+the final task `check-count` is not executed since the task result resolution fails for
+$(tasks.count-comments-after.results.count).
```
bobcatfish, line 636: ah good point!

docs/pipelines.md
```diff
@@ -516,6 +517,137 @@ In particular:
...
+all final tasks run simultaneously and starts executing once all `PipelineTasks` under `tasks` have settled which means
+no `runAfter` can be specified in final tasks.
+
+### Specifying execution `Conditions` in Final Tasks
+
+`Tasks` in a `Pipeline` can be configured to run only if some conditions are satisfied using `conditions`. But the
+final tasks are guaranteed to be executed after all `PipelineTasks` therefore no `conditions` can be specified in
+final tasks.
+
+
```
bobcatfish, line 650: these docs are very thorough and really help me!

the only thing I can think of that I'd want to also see (and let me know if I just missed it above!) is how the failure/success of finally Tasks effects the failure/success of the overall Pipeline.

my assumption is that it would work something like this:
- Pipeline succeeds, finally task fails = overall status is failure
- Pipeline succeeds, finally tasks succeed = overall status is success
- Pipeline fails, finally tasks succeed = overall status is failure
- Pipeline fails, finally task fails = overall status is failure

examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
```diff
@@ -0,0 +1,236 @@
...
+        - name: shared-workspace
+          workspace: shared-workspace
+      resources:
+        inputs:
+          - name: app-git
+            resource: app-git
+  finally:
+    - name: cleanup-workspace
+      taskRef:
+        name: cleanup-workspace
```
bobcatfish, line 212: nice example!

bobcatfish:
I can't seem to comment on the original thread anymore so continuing down here:

> hey @bobcatfish I have listed pros and cons of refactoring `PipelineTask` [here](https://gist.github.com/pritidesai/54ea8efa60692cdc2582cfc60eb12a8b) 😱 

Hey great weighing of pros and cons! I think you hit the nail on the head and your solution is way better than what I was suggesting. I tried to see if I could find any ways around the cons you mentioned in option 2 and came up with nothing.

And the more I think about it, the more option 1 makes sense: i bet we'll end up allowing conditions in finally tasks and I think we may even allow runAfter eventually.

I like option 1! Thanks for your thorough examination and patience 🙏

docs/pipelines.md
pritidesai: deleting this for now, will add this back when we add support for task result in final tasks.

docs/pipelines.md
pritidesai: nope it was not covered in here 😞 just added it 😸

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -141,45 +173,13 @@ func (ps *PipelineSpec) Validate(ctx context.Context) *apis.FieldError {
...
-				return apis.ErrMultipleOneOf(fmt.Sprintf("spec.tasks[%d].name", i))
-			}
-			taskNames[t.Name] = struct{}{}
-		}
+	if err := validateFinalTasks(ps); err != nil {
+		return err
+	}
+
+	// PipelineTask must have a valid unique label, at least one of taskRef or taskSpec should be specified
+	if err := validatePipelineTasks(ctx, ps.Tasks, ps.Finally); err != nil {
```
pritidesai, line 181: Have moved this refactoring to separate PR: https://github.com/tektoncd/pipeline/pull/2504

Please review https://github.com/tektoncd/pipeline/pull/2504.

pritidesai:
this is ready for review now, thanks 🙏

docs/pipelines.md
```diff
@@ -516,6 +517,184 @@ In particular:
...
+        - name: after-count
+          value: $(tasks.count-comments-after.results.count)
+```
+
+In this example, container logs for `check-count` might have something like this: 
+
+```
+/tekton/scripts/script-0-zsg5v: 3: tasks.sum-inputs.results.sum-result: not found
+/tekton/scripts/script-0-2g2dr: 3: tasks.count-comments-before.results.count: not found
+```
```
vdemeester, line 623: The header of this part make me think it's possible. I think this part should be rewritten differently, like having a `know limitation` header and list what won't work

docs/pipelines.md
```diff
@@ -516,6 +517,184 @@ In particular:
...
+In this example, `PipelineResults` is set to:
+
+```
+"pipelineResults": [
+  {
+    "name": "comment-count-validate",
+    "value": "$(finally.check-count.results.comment-count-validate)"
+  }
+],
+```
```
vdemeester, line 645: Same here

docs/pipelines.md
vdemeester: Also, I'm sure this has been discussed before, but what is the reason it wouldn't/couldn't support using results from previous tasks ?

docs/pipelines.md
```diff
@@ -516,6 +517,184 @@ In particular:
...
+
+It's not possible to configure or modify the execution order of the final tasks. Unlike `Tasks` in a `Pipeline`,
+all final tasks run simultaneously and starts executing once all `PipelineTasks` under `tasks` have settled which means
+no `runAfter` can be specified in final tasks.
+
+### Specifying execution `Conditions` in Final Tasks
+
+`Tasks` in a `Pipeline` can be configured to run only if some conditions are satisfied using `conditions`. But the
+final tasks are guaranteed to be executed after all `PipelineTasks` therefore no `conditions` can be specified in
+final tasks.
```
vdemeester, line 657: I would also put those under `know limitation` or something

ImJasonH:
/hold 

There's still some ongoing discussion in the design doc that I'd like to see resolved before we merge this.

docs/pipelines.md
```diff
@@ -516,6 +517,184 @@ In particular:
...
+```none
+apis.Condition{
+  Type:    apis.ConditionSucceeded,
+  Status:  corev1.ConditionTrue,
+  Reason:  Succeeded, # or Completed
+  Message: fmt.Sprintf("Tasks Completed: %d, Skipped: %d", len(tasks)+len(finally), skipTasks),
+}
+```
+
+Note that it could be possible that one or all final tasks fail but `PipelineRun` status does not reflect
```
vdemeester, line 694: I am not sure about this. If I have a pipeline that create a cluster (`openshift-install`), does things, and have a finally to remove delete the cluster. If the whole pipeline succeeded **but** the finally fails, I would consider this as a failure (even though it's at the cleaning part) because this means I may have a *dangling* cluster running around.

vdemeester: or I misunderstood the status part. I feel a table showing the different state and result would be nice :angel:

docs/pipelines.md
pritidesai: yup, we did discuss in the [API workflow group](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit) 😄 

Sure no worries @vdemeester 

Its not supported right now in this PR, we discussed a few options to address this:

**Option 1:** Should `after-count` be assigned a default value (task result default) in case `count-comments-after` fails to execute?

**Option 2:** Allow specifying explicit default, e.g.:

```
     params:
        - name: after-count
          value: $(tasks.count-comments-after.results.count)
          default: "0"
```

**Option 3:** Do not execute `check-count` but continue executing rest of the final tasks.

docs/pipelines.md
```diff
@@ -516,6 +517,184 @@ In particular:
...
+    - name: tests
+      taskRef:
+        Name: integration-test
+  finally:
+    - name: cleanup-test
+      taskRef:
+        Name: cleanup
+```
+
+### Specifying `Resources` in Final Tasks
```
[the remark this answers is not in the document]
pritidesai: yup, added workspace [here](https://github.com/tektoncd/pipeline/pull/2437/files#diff-db9ac78ce08367ea2b7db8c1412049b6R571)

docs/pipelines.md
pritidesai: added a section `Known Limitations` with task results and pipeline results, [here](https://github.com/tektoncd/pipeline/pull/2437/files#diff-db9ac78ce08367ea2b7db8c1412049b6R658)

docs/pipelines.md
pritidesai: yup, added under `known limitations`, [here](https://github.com/tektoncd/pipeline/pull/2437/files#diff-db9ac78ce08367ea2b7db8c1412049b6R695)

docs/pipelines.md
pritidesai: change the title to `Cannot specify execution ...` [here](https://github.com/tektoncd/pipeline/pull/2437/files#diff-db9ac78ce08367ea2b7db8c1412049b6R623)

docs/pipelines.md
pritidesai: replaced the explanation with table, [here](https://github.com/tektoncd/pipeline/blob/634777213080a469b254cfb2c315a8174e997549/docs/pipelines.md#pipelinerun-status-with-finally)

pritidesai:
> Few comments 👼

thanks @vdemeester I have addressed all of your comments, let me know if any more changes needed 🙏

docs/pipelines.md
```diff
@@ -516,6 +517,203 @@ In particular:
 
 The `description` field is an optional field and can be used to provide description of the `Pipeline`.
 
+## Adding `Finally` to the `Pipeline`
+
+You can specify a list of one or more final tasks under `finally` section. Final tasks are guaranteed to be executed
```
afrittoli, line 522: It may be worth saying that all tasks specified here will be executed in parallel

docs/pipelines.md
```diff
@@ -516,6 +517,203 @@ In particular:
...
+    - name: cleanup-test
+      taskRef:
+        Name: cleanup
+```
+
+### Specifying `Resources` in Final Tasks
+
+Similar to `tasks`, you can use [PipelineResources](#specifying-resources) as inputs and outputs for
+final tasks in the Pipeline. The only difference here is, final tasks with an input resource can have a `from` clause
+but that resource has to rely on output of a `PipelineTask` from `tasks` section. For example:
```
afrittoli, line 544: What happens if the task we take the resource from failed?
I think we should not allow finally tasks to depend on things that may not be there, or we could allow that only if the input resource is an optional one, so that the task can still be executed even if the input is missing.

pritidesai: Irrespective of whether task succeed or fails, referenced resource container does get created, lets say if its image resource, image might not have created and might not exist at the registry but from tekton perspective image resource still exists. 

How do we distinguish between failure to create and push image v/s task not creating or pushing image even after executing successfully?

Please point me to a specific scenario if I am missing anything 🙏

afrittoli: > Irrespective of whether task succeed or fails, referenced resource container does get created, 

Do you mean the resource folder in the artifact storage?
The task with the resource may also be skipped because of previous failures.

> lets say if its image resource, image might not have created and might not exist at the registry but from tekton perspective image resource does exist.
> 
> How do we distinguish between failure to create and push image v/s task not creating or pushing image even after executing successfully?
> 
> Please point me to a specific scenario if I am missing anything 🙏

I think I see your point. Using from enforces ordering, which is ignored in case of `finally` tasks. The task will look for the content of the input resource in the artifact storage and it will have to be able to cope with both cases, content is there or is not.
This is actually a very good behavior I think, but we might want to add a comment in the docs here to say that the input resource may not be populated there, and so a finally task will have to handle that.

pkg/apis/pipeline/v1alpha1/pipeline_validation.go
```diff
@@ -237,6 +319,16 @@ func validatePipelineWorkspaces(wss []WorkspacePipelineDeclaration, pts []Pipeli
 			}
 		}
 	}
+	for tIdx, t := range finalTasks {
+		for wsIdx, ws := range t.Workspaces {
```
bobcatfish, line 323: do we really need to use these indexes? maybe we could ignore them with _? not sure how much the adde to the error messages

pkg/apis/pipeline/v1alpha1/pipeline_validation.go
```diff
@@ -204,15 +204,97 @@ func (ps *PipelineSpec) Validate(ctx context.Context) *apis.FieldError {
...
+	for _, f := range ps.Finally {
+		if len(f.RunAfter) != 0 {
+			return apis.ErrInvalidValue(fmt.Sprintf("no runAfter allowed under spec.finally, final task %s has runAfter specified", f.Name), "spec.finally")
+		}
+		if len(f.Conditions) != 0 {
+			return apis.ErrInvalidValue(fmt.Sprintf("no conditions allowed under spec.finally, final task %s has conditions specified", f.Name), "spec.finally")
+		}
+	}
+	return nil
+}
```
bobcatfish, line 237: do we want something in here to look for "results" use as well or is that covered in other code?

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -245,6 +327,16 @@ func validatePipelineWorkspaces(wss []WorkspacePipelineDeclaration, pts []Pipeli
...
+	for tIdx, t := range finalTasks {
+		for wsIdx, ws := range t.Workspaces {
+			if _, ok := wsTable[ws.Workspace]; !ok {
+				return apis.ErrInvalidValue(
+					fmt.Sprintf("pipeline task %q expects workspace with name %q but none exists in pipeline spec", t.Name, ws.Workspace),
+					fmt.Sprintf("spec.finally[%d].workspaces[%d]", tIdx, wsIdx),
+				)
+			}
+		}
+	}
```
bobcatfish, line 339: hmmm, is it possible to share any of this logic b/w v1alpha1 and v1beta1? using separate logic for results led to some serious bugs and confusion so i think it's worth investing some time into finding ways for them both to call the exact same code

(also a great chance to break up some of this validation with unit tests, vs testing it all by passing in complete pipelines)

this is a big ask but also a good candidate for a separate PR that could be merged before this one?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -506,13 +506,23 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1alpha1.PipelineRun) er
 		return nil
 	}
 
-	if pipelineState.IsDone() && pr.IsDone() {
+	if pipelineState.IsDone() && pipelineState.IsFinalTasksDone() && pr.IsDone() {
```
bobcatfish, line 509: what about having IsDone check the final tasks? feels odd that IsDone would return true when the pipeline isnt done?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -506,13 +506,23 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1alpha1.PipelineRun) er
...
 	}
 
-	for _, rprt := range pipelineState {
+	pipelineTasks := resources.PipelineRunState{}
+	finalTasks := resources.PipelineRunState{}
+	for _, task := range pipelineState {
+		if task.PipelineTask != nil {
+			pipelineTasks = append(pipelineTasks, task)
+		} else if task.FinalPipelineTask != nil {
+			finalTasks = append(finalTasks, task)
```
bobcatfish, line 521: would it make sense for PipelineRunState to hold both final and regular tasks so you dont need to use 2 different objects, and you dont need to iterate over them separately?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -597,16 +622,49 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1alpha1.PipelineRun) er
...
+		// set the PipelineRun Status if still executing non-final tasks
+		pr.Status.SetCondition(pipelineTasksStatusAfter)
+		reconciler.EmitEvent(c.Recorder, pipelineTasksStatusBefore, pipelineTasksStatusAfter, pr)
+	}
 	pr.Status.TaskRuns = getTaskRunsStatus(pr, pipelineState)
 	c.Logger.Infof("PipelineRun %s status is being set to %s", pr.Name, pr.Status.GetCondition(apis.ConditionSucceeded))
 	return nil
 }
 
+func (c *Reconciler) executeFinalTasks(pipelineState resources.PipelineRunState, finalTasks resources.PipelineRunState, pr *v1alpha1.PipelineRun, pipelineSpec *v1alpha1.PipelineSpec, pipelineTasksStatusAfter *apis.Condition, storageBasePath string) error {
```
bobcatfish, line 642: when possible we should try to avoid putting functions on the reconciler - its easier to test and maintain over time if we put them in another package and pass in any attributes of the reconciler they need

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -597,16 +622,49 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1alpha1.PipelineRun) er
...
-	before := pr.Status.GetCondition(apis.ConditionSucceeded)
-	after := resources.GetPipelineConditionStatus(pr, pipelineState, c.Logger, d)
-	pr.Status.SetCondition(after)
-	reconciler.EmitEvent(c.Recorder, before, after, pr)
-
+	pipelineTasksStatusBefore := pr.Status.GetCondition(apis.ConditionSucceeded)
+	pipelineTasksStatusAfter := resources.GetPipelineConditionStatus(pr, pipelineTasks, c.Logger, d)
+	// begin executing final tasks when all non-final tasks are completed successfully or any one of them have failed
+	if pipelineTasksStatusAfter.IsTrue() || (pipelineTasksStatusAfter.IsFalse() && pipelineTasksStatusAfter.Reason == resources.ReasonFailed) {
+		if err := c.executeFinalTasks(pipelineState, finalTasks, pr, pipelineSpec, pipelineTasksStatusAfter, as.StorageBasePath(pr)); err != nil {
```
bobcatfish, line 629: do we want to treat finally task execution as completely different or is it possible that the dag logic could be updated to understand finally tasks?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -677,6 +735,29 @@ func getTaskRunsStatus(pr *v1alpha1.PipelineRun, state []*resources.ResolvedPipe
 	return status
 }
 
+func getFinalTaskRunsStatus(pr *v1alpha1.PipelineRun, state []*resources.ResolvedPipelineRunTask) map[string]*v1alpha1.PipelineRunTaskRunStatus {
```
bobcatfish, line 738: good candidate for unit tests? (and maybe being in a separate package?)

bobcatfish:
I was just also thinking about ways to make incremental progress on this - I wonder if it would work to have a PR to add the types and their validation before adding the functionality.

docs/pipelines.md
```diff
@@ -516,6 +517,203 @@ In particular:
...
+| `PipelineTasks` under `tasks` | Final Tasks | `PipelineRun` status | Reason |
+| ----------------------------- | ----------- | -------------------- | ------ |
+| all `PipelineTask` successful | all final tasks successful | `true` | `Succeeded` |
+| all `PipelineTask` successful | one or more failure of final tasks | `false` | `Failed` |
+| one or more `PipelineTask` skipped and rest successful | all final tasks successful | `true` | `Completed` |
+| one or more `PipelineTask` skipped and rest successful | one or more failure of final tasks | `false` | `Failed` |
+| single failure of `PipelineTask` | all final tasks successful | `false` | `failed` |
+| single failure of `PipelineTask` | one or more failure of final tasks | `false` | `failed` |
+
+### Known Limitations
```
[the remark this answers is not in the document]
pritidesai: sure since we might add `runAfter` and `conditions` to final tasks as well in future 😱

docs/pipelines.md
```diff
@@ -516,6 +517,203 @@ In particular:
...
+
+#### Cannot configure `Task` execution results with `finally`
+
+Final tasks can not be configured to consume `Results` of `PipelineTask` from `tasks` section i.e. the following
+example is not supported right now but we are working on adding support for the same.
```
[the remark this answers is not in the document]
pritidesai: [Here](https://docs.google.com/document/d/10iEJqVstY6k3KNvAXgffIJLcHRbPQ-GIAfQk5Dlrf3c/edit) is the design doc, I will create an issue as well. How do we generally function when a feature is in the discussion which is part of the new functionality (not implemented/lives in PR)? (1) task results (2) pipeline results (3) `from` pipeline resources (4) potentially post refactoring, etc

examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
```diff
@@ -0,0 +1,149 @@
...
+kind: PersistentVolumeClaim
+metadata:
+  name: shared-workspace
+spec:
+  resources:
+    requests:
+      storage: 16Mi
+  volumeMode: Filesystem
+  accessModes:
+    - ReadWriteOnce
```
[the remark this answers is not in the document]
pritidesai: Great point, will change it to use volume template 🙏

examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
```diff
@@ -0,0 +1,149 @@
...
+      script: |
+        #!/usr/bin/env bash
+        set -xe
+        rm -rf $(workspaces.shared-workspace.path)/application/
+---
+
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: clone-into-workspace-and-cleanup-workspace
```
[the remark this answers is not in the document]
pritidesai: this example file has two separate pipelines and pipelineruns, good point on embedding the spec into pipelinerun 👍

pkg/apis/pipeline/v1alpha1/pipeline_validation.go
pritidesai: yup, index is used to notify which workspace in the `PipelineTask` is not declared in Pipeline spec in case its missing from Pipeline's workspace list.

pkg/apis/pipeline/v1alpha1/pipeline_validation.go
pritidesai: `results` are part of tasks, we are validating `PipelineTask` here. Also, no validation done on whether a final task is looking for result. In case a final task references a task result in a param e.g. `$(tasks.count-comments-after.results.count)`, that param value is not resolved and results in[ not found](https://github.com/tektoncd/pipeline/pull/2437/files#diff-db9ac78ce08367ea2b7db8c1412049b6R691) during the pipeline execution.

I think its a good point to catch it and fail at the validation, will add it.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
pritidesai: I am sure it is possible to share logic between v1alpha1 and v1beta1 also love your idea of breaking this validation and testing them piece by piece instead of creating complete pipeline.

I am trying to think about possible fix here, one option could be keep the local functions to validate graph, workspaces, resources, etc in `v1beta1` and make them public so that the same functions can be used by `v1alpha1` after converting `v1alpha1` resources to `v1beta1` (mainly `PipelineTask`) 🤭  

Also, looked at moving those local functions to a common utility 🤔

pritidesai: something like [this](https://github.com/tektoncd/pipeline/compare/master...pritidesai:refactor-pipeline-validation) 🤔

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: yes, it does feel odd, was trying to keep them separate but I guess we are still talking about the same pipeline so `isDone` can check final tasks too 😕

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: yup its treated completely different for now, I wouldnt mix it up with dag logic, the existing dag logic doesnt handle tasks which are skipped. If a condition results in failure, a task is skipped along with all the tasks from that branch. This list of skipped tasks is determined while retrieving [PipelineRunStatus](https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L446) and kept limited to the status message with its count. DAG logic is lacking means of knowing if a given task was skipped on purpose or haven't scheduled for execution yet.

Would like to propose something like [PipelineRunPipelineTaskState](https://github.com/pritidesai/pipeline/blob/skip/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L295) which holds a task with its execution status and more. This proposal will simplify dag logic, the way a list of skipped task is derived and also be beneficial for CLI and dashboard.

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: yup, lacking lot of unit testing, love your proposal of splitting the implementation into separate PRs, can add more unit tests instead of having one huge XX...XL PR 😱

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: > This is looking great! I think we're all almost totally agreed about the feature itself so most of my comments are about the implementation.

🎉 🎉 🎉 
 
> Main themes:
> 
> * I'm not sure about allowing "from" in finally tasks, feels a bit confusing to me when we don't allow results (e.g. imagine a git resource that is "from" a task that builds - if the build task executes, the data on disc for the resource may be different than if it doesnt)

Yup, I agree, task results and `from` fall under the same category and can be added incrementally. 

> * I think its worth trying to avoid duplicating the validation logic as much as possible, since it leads to some nasty bugs, for example we ran into that with results (#2471) - we're still duplicating some logic there so it might not be completely avoidable, but if we can avoid it its worth the effort

yup, on it, have commented below as well, how about something like [this](https://github.com/tektoncd/pipeline/compare/master...pritidesai:refactor-pipeline-validation), any ideas or feedback is appreciated, I am still learning `v1alpha1` and `v1beta1` APIs.

> * Better to add functions in separate packages than add more functions to the reconciler

Sure, still learning about reconcilers 🤪

> * If we get the structure of PipelineRunState right, i bet we can avoid some of the duplicated code in the reconciler
> * Is it possible for the logic that determines what runs next to become aware of finally? If the logic that returns the next tasks to run can handle Finally tasks, i bet we need a lot less special casing for finally

Yup, see my comment [below](https://github.com/tektoncd/pipeline/pull/2437#discussion_r420534851), I am working on proposal which can be part of separate PR and is applicable without finally as well.

> It looks like the coverage generally went down, and we want it to generally go up or stay the same so it might be worth digging into that a bit once we're more sure about the details of the implementation:
> 
> image

yup, not happy at all, I have lot more unit testing which is not part of this PR, was nervous with the sizing and growing number of LOC. Splitting is the key, will have separate PRs.

docs/pipelines.md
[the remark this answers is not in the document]
pritidesai: https://github.com/tektoncd/pipeline/issues/2557

pritidesai:
I have split this PR into two separate `wip` PRs:
(1) adding `finally` type and its validation - PR #2650 
(2) implementing this new feature - PR #2651 

will close this PR once all the comments from this PR are addressed in new PRs.

docs/pipelines.md
pritidesai: `from` is not being supported and being tracked in https://github.com/tektoncd/pipeline/issues/2711

docs/pipelines.md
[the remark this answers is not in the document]
pritidesai: https://github.com/tektoncd/pipeline/issues/2711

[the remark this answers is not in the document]
pritidesai: https://github.com/tektoncd/pipeline/issues/2710

pkg/apis/pipeline/v1alpha1/pipeline_validation.go
pritidesai: Added, https://github.com/tektoncd/pipeline/pull/2650/files#diff-5b096c7641765902d79ee2d5e5f7d213R465

pkg/apis/pipeline/v1beta1/pipeline_validation.go
pritidesai: Have done huge refactoring by separating positive tests and negative tests, also by renaming test package to `v1beta1` to test individual functions, will create validation routine as part of `PipelineTask` like @bobcatfish mentioned here: https://github.com/tektoncd/pipeline/pull/2650/files#r428820876

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: have merged it in the latest PR #2661

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: have separated PRs and added more unit tests to make sure coverage doesnt go down 😄

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: merged it in the latest PR #2661

docs/pipelines.md
pritidesai: Added in the latest implementation PR: https://github.com/tektoncd/pipeline/pull/2661/files#diff-db9ac78ce08367ea2b7db8c1412049b6R535

examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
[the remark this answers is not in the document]
pritidesai: Changed it to VolumeClaimTemplate in replacement [PR](https://github.com/tektoncd/pipeline/pull/2661/files#diff-56c052a6b622ca906e10799a94023df4R175)

examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
```diff
@@ -0,0 +1,149 @@
...
+    - name: shared-workspace
+      persistentVolumeClaim:
+        claimName: shared-workspace
+  resources:
+    - name: app-git
+      resourceSpec:
+        type: git
+        params:
+          - name: url
+            value: https://github.com/tektoncd/pipeline.git
```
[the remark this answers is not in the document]
pritidesai: yup replaced pipeline resource with git task and copy pasted `git-clone` 😞 https://github.com/tektoncd/pipeline/pull/2661/files#diff-56c052a6b622ca906e10799a94023df4R4

examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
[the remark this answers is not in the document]
pritidesai: Dropped the first pipelineRun from examples, converted it into integration test comparing final task start time and DAG task completion time [here](https://github.com/tektoncd/pipeline/pull/2661/files#diff-a80234191e535ce50a353f68f8fbb21dR35).

pritidesai:
I have addressed all the comments from this PR. Please review the two (new) separate PRs:

* PR #2650 
* PR #2661 

Closing this PR 👋

Issue #279, opened 2018-11-29
feature request: snapshot pipeline data in pipelineRun for historical and repeatability purposes


nader-ziada:
## Expected Behavior
When a PipelineRun is created and it expands the pipeline reference, it should include the complete copy of the pipeline yaml so that if the pipeline changes, a user can tell which actual pipeline the results came from.

## Actual Behavior
No information is about the Pipeline is added in the PipelineRun

## Steps to Reproduce the Problem

  1. kubectl get "your-pipeline-run" -o yaml 
will show the pipelineRun yaml with no information about the contents of the pipeline itself

## Additional Info
 * same applied to task and taskRun

bobcatfish:
> When a PipelineRun is created and it expands the pipeline reference, it should include the complete copy of the pipeline yaml so that if the pipeline changes, a user can tell which actual pipeline the results came from.

@pivotal-nader-ziada would this include also caching this `Pipeline` and using it in future reconcile loops? @tanner-bruce @mattmoor and @dwnusbaum were discussing this a bit in the slack chat, about wanting to make sure that changing a `Pipeline` midway through a run shouldn't affect runs in progress

(no worries if it makes more sense to tackle that separately)

Also @mattmoor 's suggestion about having a similar `Revision` model here ([similar to knative/serving](https://github.com/knative/serving/blob/master/docs/spec/overview.md#revision)) could be relevant here.

Any other ideas?

pivotal-nader-ziada:
you mean creating a new version of the pipeline every time its changed? might be too much versions and we then the pipelineRun needs to have the ability to select which version to use if not latest. 

To clarify, yes, I would expect the PipelineRun to make a copy of the Pipeline and use that in any reconcile until this run is finished.

dwnusbaum:
I think that having the ability to select the version of a Pipeline/Task to run could be useful in some cases. For example, if you have a Task that is used in a few Pipelines, and you want to update it, you will probably want to test your updates on a single Pipeline to make sure it works before you force all Pipelines to use the new version of the Task for new builds. You can always do this with a kind of poor man's versioning in the Pipeline/Task name manually (i.e. `PipelineNameV2`, `TaskName@scm-revision`), but if this is a common use case then supporting it directly could be helpful.

[...]

PR #2399, opened 2020-04-15
Add the command and arguments run in each step to the TaskRun.Status

dlorenc:
# Changes

The TaskRun.Status currently serves as the best log of what happened in a TaskRun.
It contains the container digests for each step, as well as their exit code and
termination message. However, it is currently missing the Command and Arguments,
making it impossible to verify what was run inside of the container after the
pod completes.

This commit adds those fields to the API, and copies them over from the Pod
when the rest of the Status is populated.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [X] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [X] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [X] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

# Release Notes

```
TaskRun.Status.Steps now includes the Command and Args run inside each container.
```

PR #2444, opened 2020-04-20
Dereference the TaskSpec into TaskRun.Status.

dlorenc:
# Changes

The Task definition used for a TaskRun can change after the TaskRun
has started. This poses problems for auditability post-run. Rather
than chase down every part of a Task that we might like to audit later,
let's just add the entire thing here.

This is a replacement for https://github.com/tektoncd/pipeline/pull/2399

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
The TaskRun.Status now contains a TaskSpec field, including the full dereferenced Task used to instantiate the run.
```

### DIFF
--- a/docs/taskruns.md
+++ b/docs/taskruns.md
@@ -316,9 +316,9 @@ If multiple `Steps` are defined in the `Task` invoked by the `TaskRun`, you can 
 status in the `steps.results` field using the following command, where `<name>` is the name of the target
 `TaskRun`:
 
-```
-kubectl get taskrun <name>
-```
+The exact Task Spec used to instantiate the TaskRun is also included in the Status for full auditability.
+
+### Steps
 
 The corresponding statuses appear in the `status.steps` list in the order in which the `Steps` have been
 specified in the `Task` definition.
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -216,6 +216,11 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1alpha1.TaskRun) error 
 		return nil
 	}
 
+	// Store the fetched TaskSpec on the TaskRun for auditing
+	if err := storeTaskSpec(ctx, tr, taskSpec); err != nil {
+		c.Logger.Errorf("Failed to store TaskSpec on TaskRun.Statusfor taskrun %s: %v", tr.Name, err)
+	}
+
 	// Propagate labels from Task to TaskRun.
 	if tr.ObjectMeta.Labels == nil {
 		tr.ObjectMeta.Labels = make(map[string]string, len(taskMeta.Labels)+1)
@@ -697,4 +702,13 @@ func applyVolumeClaimTemplates(workspaceBindings []v1alpha1.WorkspaceBinding, ow
 		taskRunWorkspaceBindings = append(taskRunWorkspaceBindings, b)
 	}
 	return taskRunWorkspaceBindings
+}
+
+func storeTaskSpec(ctx context.Context, tr *v1alpha1.TaskRun, ts *v1alpha1.TaskSpec) error {
+	// Only store the TaskSpec once, if it has never been set before.
+	if tr.Status.TaskSpec == nil {
+		tr.Status.TaskSpec = &v1beta1.TaskSpec{}
+		return ts.ConvertTo(ctx, tr.Status.TaskSpec)
+	}
+	return nil
 }
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
 	golang.org/x/tools v0.0.0-20200214144324-88be01311a71 // indirect
 	gomodules.xyz/jsonpatch/v2 v2.1.0
-	google.golang.org/api v0.15.0
+	google.golang.org/api v0.15.0 // indirect
 	google.golang.org/appengine v1.6.5 // indirect
 	k8s.io/api v0.17.3
 	k8s.io/apiextensions-apiserver v0.17.3 // indirect
--- a/pkg/apis/pipeline/v1beta1/taskrun_types.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_types.go
@@ -177,6 +177,9 @@ type TaskRunStatusFields struct {
 	// The list has one entry per sidecar in the manifest. Each entry is
 	// represents the imageid of the corresponding sidecar.
 	Sidecars []SidecarState `json:"sidecars,omitempty"`
+
+	// TaskSpec contains the Spec from the dereferenced Task definition used to instantiate this TaskRun.
+	TaskSpec *TaskSpec `json:"taskSpec,omitempty"`
 }
 
 // TaskRunResult used to describe the results of a task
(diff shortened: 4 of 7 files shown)


afrittoli:
Nice, I was going to work on this after I finished https://github.com/tektoncd/pipeline/pull/2421 :)
This is both a benefit to auditability, but it also should improve the reconcile performance, since we now fetch the task on every reconcile cycle. 
Once we start fetching tasks from an OCI registry, it will be even more relevant.

One though I have is that we should also store resources in the status.

pkg/pod/status.go
```diff
@@ -116,6 +118,12 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1alpha1.TaskRun, pod *core
 	trs.Steps = []v1alpha1.StepState{}
 	trs.Sidecars = []v1alpha1.SidecarState{}
 
+	ts := v1beta1.TaskSpec{}
```
afrittoli, line 121: We might want to store this into the TaskRun status when we first fetch it, i.e. https://github.com/tektoncd/pipeline/blob/9344e4f2291bb5b9a35f50f7549c520e02c8bf56/pkg/reconciler/taskrun/taskrun.go#L207-L208

We could then modify that could to fetch the task only if not available in the status, and else use that from status

dlorenc: Whoa, yeah thanks. Do you think it's enough to only set this once, the first time?

dlorenc:
> One though I have is that we should also store resources in the status.

I hesitated on this one for now. We get some of the resource info in the status now (under containers):

```json
{
      "container": "step-git-source-skaffold-git-build-push-kaniko-p28nk",
      "imageID": "docker-pullable://gcr.io/dlorenc-vmtest2/git-init-4874978a9786b6625dd8b6ef2a21aa70@sha256:99f410cc098b4300384ad056c8febf5bb79e4749417b5e9621fc529514c4a0c3",
      "name": "git-source-skaffold-git-build-push-kaniko-p28nk",
      "terminated": {
        "containerID": "docker://b96396af17dea856526547ef5e140478148d4d672c8bda8a95e6010380e7cfab",
        "exitCode": 0,
        "finishedAt": "2020-04-20T15:32:20Z",
        "message": "[{\"key\":\"commit\",\"value\":\"6ed7aad5e8a36052ee5f6079fc91368e362121f7\",\"resourceRef\":{\"name\":\"skaffold-git-build-push-kaniko\"}}]",
        "reason": "Completed",
        "startedAt": "2020-04-20T15:32:18Z"
      }
    },
```

This doesn't include commands/arguments, but it's close. We also get `resourcesResults`. 

Should we add them to the v1beta1 API, given we don't know the future of them yet? https://github.com/tektoncd/pipeline/pull/2447

pkg/pod/status.go
afrittoli: Yes, I think that would be great! Especially when the task spec is not embedded in the taskrun spec, this avoids issues when the task definition is updated during the run.

The only case where is use the change to the spec for something is `cancel`, so we would have to find an alternative for that. Perhaps the `tkn` client could reset the task definition in status to force a reload? @vdemeester what do you think?

pkg/pod/status.go
```diff
@@ -116,6 +118,15 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1alpha1.TaskRun, pod *core
...
 
+	// Only set this once, the first time.
+	if trs.TaskSpec == nil {
+		ts := v1beta1.TaskSpec{}
+		if err := taskSpec.ConvertTo(context.Background(), &ts); err != nil {
+			logger.Errorf("error setting taskrun.Status.taskSpec in taskrun %s: %w", tr.Name, err)
+		}
+		trs.TaskSpec = &ts
+	}
+
```
afrittoli, line 129: I till think that https://github.com/tektoncd/pipeline/blob/9344e4f2291bb5b9a35f50f7549c520e02c8bf56/pkg/reconciler/taskrun/taskrun.go#L207-L208 would be a better place for this.
Some tasks will never have a pod successfully created, and for them there will be no taskspec available for audit.

Still this works fine for many other cases, so I don't mind go ahead with this and adding on top if you prefer so.

Something that I would like to do is to use the spec stored in there during reconcile as well, but since that would break `cancel`, we need to find a solution to that first.

dlorenc: Ah, that makes a lot of sense. Thanks for explaining! I made the change.

pkg/pod/status.go
afrittoli: Oh, actually it would not break `cancel` since that goes in the spec of the taskrun, not of the task - so fetching the task only once would work just fine!

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -699,3 +704,12 @@ func applyVolumeClaimTemplates(workspaceBindings []v1alpha1.WorkspaceBinding, ow
 	}
 	return taskRunWorkspaceBindings
 }
+
+func storeTaskSpec(ctx context.Context, tr *v1alpha1.TaskRun, ts *v1alpha1.TaskSpec) error {
```
afrittoli, line 708: 🎉

skaegi:
Hmm... I get the idea here and we care deeply about audit capabilities, however I think this is insufficient for that purpose. For example, what about the configmaps and the secrets and any of the other resources involved in running a pipeline. All of those also need to be snapshotted for audit purposes. At least for us this just makes the TaskRun bigger.

dlorenc:
> Hmm... I get the idea here and we care deeply about audit capabilities, however I think this is insufficient for that purpose. For example, what about the configmaps and the secrets and any of the other resources involved in running a pipeline. All of those also need to be snapshotted for audit purposes. At least for us, this just makes the TaskRun bigger.

Definitely insufficient, but still a good step forward IMO. With this we'd capture the secret and configmap references, but not their values. I don't think we'd want to snapshot secret values, but could probably be convinced the other way.

Is size a concern? As @afrittoli pointed out this also has other uses beyond auditing.

pkg/pod/status.go
vdemeester: yeah

afrittoli:
We discussed about this [during the WG](https://docs.google.com/document/d/1rPR7m1Oj0ip3bpd_bcS1sjZyPgGi_g9asF5YrExeESc/edit#bookmark=id.lqgyn5x5ir84). 
The concerned raised about this was the size limits in etcd, but generally it seemed that the benefits would outweigh the concerns.

So I think it should be fine to move forward with this, we can always decide to store the audit data somewhere else in future. I also created https://github.com/tektoncd/pipeline/issues/2473 to track the work on resolving tasks and pipelines only once - which is related to this in the sense that it could use the same info stored for auditing purposes.

afrittoli:
> Hmm... I get the idea here and we care deeply about audit capabilities, however I think this is insufficient for that purpose. For example, what about the configmaps and the secrets and any of the other resources involved in running a pipeline. All of those also need to be snapshotted for audit purposes. At least for us, this just makes the TaskRun bigger.
> 
> @ImJasonH had a [proposal for a Results API](https://docs.google.com/document/d/1-XBYQ4kBlCHIHSVoYAAf_iC01_by_KoK2aRVO0t8ZQ0/edit#heading=h.xgjl2srtytjt). Maybe that's something we could improve

It's an interesting idea - the only concern I would have with that is that it would make the execution history API a mandatory component for running tekton pipelines.

docs/taskruns.md
```diff
@@ -316,9 +316,9 @@ If multiple `Steps` are defined in the `Task` invoked by the `TaskRun`, you can
 status in the `steps.results` field using the following command, where `<name>` is the name of the target
 `TaskRun`:
 
-```
-kubectl get taskrun <name>
-```
+The exact Task Spec used to instantiate the TaskRun is also included in the Status for full auditability.
```
bobcatfish, line 319: 🎉

skaegi:
Yep. Apart from my own selfish utility concerns ;) size in etcd is the only real concern. This is still fairly safe and not going to break anything for us.

bobcatfish:
I believe this and #2489 fixed #279

PR #2447, opened 2020-04-20
Document PipelineResource's Alpha State

### LINKED ISSUE #1369: Beta documentation: write up alternatives to PipelineResources if they remain Alpha
# Expected Behavior

Clearly documented approach to passing files, folders, docker images, cluster information, auth information, repo information etc etc without relying on PipelineResources.

# Actual Behavior

At the moment our docs assume that PipelineResources are "the answer" to modelling meaningful resources external to the CI/CD system. In the most recent PipelineResource working group we landed on the idea of keeping PipelineResources as Alpha when the rest of the Tekton resources move to Beta. So with that approach we'll need to make sure that our docs correctly reflect the "right" way to work with external resources.

sbwsg:
# Changes

Fixes #1369 

During the release of Pipelines' Beta we communicated the
fact that PipelineResources are not being promoted along
with other types but we didn't include any extensive
commentary to support that decision.

This commit adds a description of the problems that remain
with PipelineResources, along with possible next steps
that the project might take with them.

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/docs/resources.md
+++ b/docs/resources.md
@@ -19,6 +19,19 @@ For example:
     deployed in a cluster.
 -   A `Task`'s output can be a jar file to be uploaded to a storage bucket.
 
+> **Note**: PipelineResources have not been promoted to Beta in tandem with Pipeline's
+> other CRDs. This means that the level of support for `PipelineResources`
+> remains Alpha and there are effectively no guarantees about the type's future.
+> There remain a lot of known issues with the type that have caused Tekton's
+> developers to reassess it.
+>
+> For Beta-supported alternatives to PipelineResources see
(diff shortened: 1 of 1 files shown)


docs/resources.md
```diff
@@ -1048,6 +1056,70 @@ Data,
   }
 ```
```
dlorenc, line 1058: It would also be helpful to include a section on how to use the API in the meantime. What are the recommended ways to use Tasks and Pipelines now, if you were relying on PipelineResources before?

Maybe this is already covered in a v1alpha to beta migration guide.

sbwsg: Good catch, it's covered there and I've added links to it.

PR #2462, opened 2020-04-21
PR upload validates comment file extension

### LINKED ISSUE #2168: pullrequest-init comment interface is inconsistent b/w upload and download
# Expected Behavior

The expected contents of the comments dir [e.g. example](https://github.com/tektoncd/pipeline/tree/master/cmd/pullrequest-init/example/comments) should be the same whether you are modifying a PR or getting it's state.

# Actual Behavior

Using `-mode upload` with a json formatted comment such as those seen [in the example dir](https://github.com/tektoncd/pipeline/tree/master/cmd/pullrequest-init/example/comments) results in the entire json content being used as a comment, e.g.: https://github.com/bobcatfish/catservice/pull/20#issuecomment-594836608

```
{
"Body": "How about right now? It's 2020-03-04T20:59:01Z",
"Author": {
"Login": "robotcatfish",
},
}
```

But when using `-mode download`, the comments dir is populated with files such as:

```json
{"ID":594836608,"Body":"{\n  \"Body\": \"How about right now? It's 2020-03-04T20:59:01Z\",\n  \"Author\": {\n    \"Login\": \"robotcatfish\",\n  },\n}\n","Author":{"Login":"bobcatfish","Name":"","Email":"","Avatar":"https://avatars2.githubusercontent.com/u/432502?v=4","Link":"","Created":"0001-01-01T00:00:00Z","Updated":"0001-01-01T00:00:00Z"},"Link":"https://github.com/bobcatfish/catservice/pull/20#issuecomment-594836608","Created":"2020-03-04T20:59:09Z","Updated":"2020-03-04T20:59:09Z"}}
``` 

# Additional Info

Note the format is consistent for the `status` dir: https://github.com/tektoncd/pipeline/tree/master/cmd/pullrequest-init/example/status

adshmh:
# Changes

pullrequest upload now validates the extension of each comment file:
the only valid extensions are none and .json. This allows the upload to
fail if a .json comment file cannot be parsed, rather than posting all
the contents as the comment. Fixes #2168

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Comment files under a PullRequest need to have either:
- no extension: for comment files that contain comment as plain text
- .json extension: for comment files in JSON format
```

### DIFF
--- a/pkg/pullrequest/disk_test.go
+++ b/pkg/pullrequest/disk_test.go
@@ -232,7 +232,7 @@ func TestFromDiskWithoutComments(t *testing.T) {
 	writeFile(filepath.Join(d, "base.json"), &base)
 	writeFile(filepath.Join(d, "head.json"), &head)
 
-	rsrc, err := FromDisk(d)
+	rsrc, err := FromDisk(d, false)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -341,7 +341,7 @@ func TestFromDisk(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	rsrc, err := FromDisk(d)
+	rsrc, err := FromDisk(d, false)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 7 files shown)


dibyom:
/ok-to-test
cc @wlynch

sbwsg:
If we go ahead with this we'd need to document it somewhere; maybe in the pipeline resources doc?  Also bear in mind this binary is used by the [pullrequest task in the catalog](https://github.com/tektoncd/catalog/tree/v1beta1/pullrequest), so we'd need to update docs and tagged version there too.

I have some reservations about basing the behaviour on the file extension. But it might be fine in the context of the pullrequest resource and match user expectations?  I've never used it so I don't really have a strong idea.  @bobcatfish @wlynch does this seem like a good approach from your pov?  Basing the behaviour on the file extension?

/kind feature

wlynch:
The original purpose of this was so that tools could do something like `cat "asdf" > comment` to create comments without worrying about JSON validation, since the API only allows you to include the `body` field anyway. `.json` extensions were allowed so the same behavior could be used for updating comments.

I think the use case/concern given here is valid, though I think I'd prefer if we enable strict JSON validation on only `.json` extensions, and allow text for any other extension, this way users could also include things like `.md` extensions for markdown.
We should also probably make the extension check case insensitive(e.g. `.json`, `.JSON`, etc.) via strings.EqualFold.

We probably also want to control this behavior via a flag. This way existing users can have a mechanism for users to revert to the old behavior if this change breaks their workflow.

adshmh:
> If we go ahead with this we'd need to document it somewhere; maybe in the pipeline resources doc? Also bear in mind this binary is used by the [pullrequest task in the catalog](https://github.com/tektoncd/catalog/tree/v1beta1/pullrequest), so we'd need to update docs and tagged version there too.

Thank you for the review. I have updated the PR to include docs. I will also submit a PR against catalog.

adshmh:
> I think the use case/concern given here is valid, though I think I'd prefer if we enable strict JSON validation on only `.json` extensions, and allow text for any other extension, this way users could also include things like `.md` extensions for markdown.

Thank you for the review. I have updated the PR to address this comment.

> We should also probably make the extension check case insensitive(e.g. `.json`, `.JSON`, etc.) via strings.EqualFold.
> 
> We probably also want to control this behavior via a flag. This way existing users can have a mechanism for users to revert to the old behavior if this change breaks their workflow.

Thank you for the review. I have updated the PR and added a flag to allow overriding this behavior.

adshmh:
failures seem unrelated.
/retest

pkg/pullrequest/disk_test.go
```diff
@@ -420,6 +420,131 @@ func readAndUnmarshal(t *testing.T, p string, v interface{}) {
...
+				t.Fatal("expected error, got nil")
+			}
+			if err != nil && !tt.expectErr {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if len(comments) != len(tt.expectedComments) {
+				t.Fatalf("Comments length does not match expected length. expected %d, got %d: %+v", len(tt.expectedComments), len(comments), comments)
+			}
+			for i, c := range tt.expectedComments {
+				if diff := cmp.Diff(comments[i], c); diff != "" {
```
sbwsg, line 540: I think the order of arguments is incorrect here; should be `(expected, received)`, which here would manifest as:

```go
if diff := cmp.Diff(c, comments[i]); diff != "" {
```

pkg/pullrequest/disk_test.go
```diff
@@ -420,6 +420,131 @@ func readAndUnmarshal(t *testing.T, p string, v interface{}) {
...
+					t.Errorf("Get comments: -want +got: %s", diff)
```
sbwsg, line 541: This can be replaced with `diff.PrintWantGot(diff)`.  The `diff` package can be imported from `github.com/tektoncd/pipeline/test/diff`

pkg/pullrequest/disk_test.go
adshmh: Thank you for the review. Fixed.

pkg/pullrequest/disk_test.go
adshmh: Thank you for the review. Fixed.

adshmh:
> @adshmh I noticed one more fix to be made in a test but otherwise I think this PR is ready to go. Let me know if you're still interested in getting this merged. Will need a rebase and commits to be squashed as well then I'll lgtm and hopefully we can get this in ASAP!

Thank you for the review. Rebased and squashed commits.

Issue #2079, opened 2020-02-20
Embed condition spec into pipeline


siamaksade:
# Expected Behavior
Allow embedding conditions into the pipeline spec so that a pipeline could contain all relevant task, resource and conditions specs that it uses.

# Actual Behavior
Conditions are currently referenced from a pipeline.

bobcatfish:
We're already embedding everything else so this seems reasonable!

@siamaksade out of curiousity, are you doing a lot of spec embedding? imo for "best practices" (whatever that means: here I think it just means practices Christie likes XD) embedding should be the exception, not the norm (less reusability!)

siamaksade:
@bobcatfish We do. The rationale is that some conditions are intimate details of how a particular task should execute in a particular pipeline. For those, an referenced object is not attractive since it will never be re-used by any other pipeline.

The other aspect of it is `PipelineRun` which should be a snapshot of the execution of a particular pipeline at some point in time. Referenced resources make it difficult to get that snapshot since they can change after the pipeline run is completed and make it impossible to have an accurate view of past pipeline runs. `PipelineRun` can embed the pipeline to solve this but if the `Condition` is still not included in the pipelinerun (in pipeline spec really), then still one won't know what the pipelinerun actually executed if the condition gets modified.

bobcatfish:
Update: @jerop started to work on this and then after discussion with @vdemeester we decided to put it on hold until we decide how to get conditions to beta

jerop:
update: https://github.com/tektoncd/community/blob/master/teps/0007-conditions-beta.md

PR #2465, opened 2020-04-22
Allow results to work when pipeline spec is embedded ⛏️

### LINKED ISSUE #2463: Results require runAfter when embedding Pipeline spec
# Expected Behavior

Embedding a Pipeline spec into a PipelineRun should not change the behaviour of results.

# Actual Behavior

If you remove the `runAfter` clause from [this example](https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun-results-with-params.yml), you get an error caused by the controller mistakenly starting both Tasks simultaneously.

```yaml
kind: PipelineRun
metadata:
  generateName: test-case
  name: test-case-run
spec:
  params:
  - name: prefix
    value: prefix
  pipelineSpec:
    params:
    - name: prefix
    tasks:
    - name: generate-suffix
      taskSpec:
        results:
        - name: suffix
        steps:
        - name: generate-suffix
          image: alpine
          script: |
            echo -n "suffix" > $(results.suffix.path)
    - name: do-something
      taskSpec:
        params:
        - name: arg
        steps:
        - name: do-something
          image: alpine
          script: |
            echo "$(params.arg)"
      params:
      - name: arg
       # this should be enough to imply to the controller that this Task should run
       # after generate-suffix, but with runAfter removed, the controller starts it at
       # the same time as generate-suffix
        value: "$(params.prefix):$(tasks.generate-suffix.results.suffix)"
```

1. Apply the above yaml and look at the pipelinerun

# Additional Info

Interestingly this doesnt happen when not embedding the pipelinespec, e.g. this other example does not have the same problem https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun-results.yaml

- Tekton Pipeline version: HEAD @ 1346656111decda59d4c36280b2cc828127a37e1

bobcatfish:
# Changes

Yesterday I thought I would just quickly update our existing build +
deploy pipelinerun example to use a git task and a kaniko task instead
of PipelineResources and I immediately started stumbling over some very
strange bugs.

I'm not sure I've gotten 100% to the bottm of why these bugs happen, but
part of it seems to be that depending on whether a pipelinespec is
embedded or not, sometimes the PipelineTask instances are v1beta1 and
sometimes v1alpha1. The logic to construct the graph is intentionally
using a "Task" interface instead of dealing with either of thse
directly, so it seems that in order to work properly, we need to have
the same logic for the Deps function for both. In the long run I hope we
can use exactly the same code instead of duplicating it, but in the
short term, now both functions are the same. To do this I had to make it
so that both v1alpha1 and v1beta1 could use the same Params and Results
types. I think this is a nice way to organize the code anyway, and
as a bonus, moving Params into a package outside of v1alpha1 and v1beta1
will make it so that we can more easily embed Conditions as part of #2079.

Fixes #2463

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Fixes a bug which led to incorrect behavior when using results to order Tasks and embedding Pipeline specs.
```


pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -152,6 +142,25 @@ func (pt PipelineTask) Deps() []string {
...
+	}
+	// Add any dependents from task results
+	for _, param := range pt.Params {
+		expressions, ok := results.GetVarSubstitutionExpressionsForParam(param)
+		if ok {
+			resultRefs := results.NewResultRefs(expressions)
+			for _, resultRef := range resultRefs {
+				deps = append(deps, resultRef.PipelineTask)
+			}
+		}
```
vdemeester, line 163: This is the real fix, right ?

bobcatfish:
Spoke with @vdemeester at length about this - gonna see if i can fix it without moving the params + results types into their own packages, but might still do that in a separate PR!

pkg/apis/pipeline/v1beta1/pipeline_types.go
othomann: This is the change to handle the dependencies.

pkg/apis/pipeline/v1alpha1/pipeline_conversion.go
```diff
@@ -44,6 +44,7 @@ func (source *PipelineSpec) ConvertTo(ctx context.Context, sink *v1beta1.Pipelin
 	sink.Workspaces = source.Workspaces
 	sink.Description = source.Description
 	if len(source.Tasks) > 0 {
+		// todo: why is this failing?
```
bobcatfish, line 47: lol

Issue #2387, opened 2020-04-14
Cannot assign params with task results to other task value


r0bj:
# Expected Behavior

I can assign params with task results to other task value:
```
      params:
      - name: arg
        value: "$(params.prefix):$(tasks.generate-suffix.results.suffix)"
```

# Actual Behavior

Cannot assign params with task results to other task value. Error message is generated:
```
validation failed: invalid value: Invalid result reference expression: Must be of the form \"tasks.<taskName>.results.<resultName>\": spec.tasks.params.value
```

# Steps to Reproduce the Problem

pipeline.yaml:
```
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  generateName: test-
spec:
  params:
  - name: prefix
    value: prefix
  pipelineSpec:
    params:
    - name: prefix
    tasks:
    - name: generate-suffix
      taskSpec:
        results:
        - name: suffix
        steps:
        - name: generate-suffix
          image: alpine
          script: |
            echo "suffix" > $(results.suffix.path)
    - name: do-something
      runAfter:
      - generate-suffix
      taskSpec:
        params:
        - name: arg
        steps:
        - name: do-something
          image: alpine
          script: |
            echo "$(params.arg)"
      params:
      - name: arg
        value: "$(params.prefix):$(tasks.generate-suffix.results.suffix)"
```

Executing this pipelinerun generates error message:
```
{"level":"warn","logger":"tekton.pipeline-controller","caller":"pipelinerun/pipelinerun.go:193","msg":"Failed to update PipelineRun status{error 25 0  admission webhook \"validation.webhook.pipeline.tekton.dev\" denied the request: validation failed: invalid value: Invalid result reference expression: Must be of the form \"tasks.<taskName>.results.<resultName>\": spec.tasks.params.value}","commit":"82e2758","knative.dev/controller":"pipeline-controller"}
[...]

PR #2471, opened 2020-04-22
Fix 3 bugs with results

### LINKED ISSUE #2464: Results regex should be the same as other substitution regexes
# Expected Behavior

We should permit the same characters in substitution for Task results as we do for other subsitutions.

# Actual Behavior

The results variable replacement logic defines its own regex:

https://github.com/tektoncd/pipeline/blob/c03323ece1af7ab8c3598d91aa988294d9b6b36b/pkg/apis/pipeline/v1beta1/resultref.go#L37

Which is different from the one used for other substitution:

https://github.com/tektoncd/pipeline/blob/0452a0f9fcc3098ccd7ca81b9c7f16b80ba3a26f/pkg/substitution/substitution.go#L27

This means the results regex, doesn't allow for the same set of characters, for example it does not allow the name of a variable to contain `_`

bobcatfish:
# Changes

Okay it's one PR with 3 fixes BUT they are in different commits! And also - this is much more reasonable than my first attempt (https://github.com/tektoncd/pipeline/pull/2465)

* Me 🐱: I fixed it! Review my giant PR where I also moved params and results and touched nearly every file!
* Vincent 😺: I think you can fix it without all that refactoring, maybe merge that first?
* Me 😡: Then I will have to duplicate everything! NOOOOOOOO
* Vincent 😺: Sometimes a bit of duplication is worth it
* Me 😡:  FINE I WILL TRY IT RELUCTANTLY
* *tries it*
* *actually needs no additional duplication at all*
* Me 😳: ... perhaps I should have tried that first

**Lesson: try the quicker fix first and THEN refactor**

## Okay so here are the 3 things I'm trying to address:

(To review separately plz look at the individual commits https://github.com/tektoncd/pipeline/pull/2471/commits 🙏 )

**1. Use the same Deps logic in v1alpha1 and v1beta1**

Sometimes the PipelineTask that Deps is being executed on is actually
v1beta1 instead of v1alpha1 and the old Deps function, which doesn't
account for Results, is being called.

This PR duplicates the logics so the Deps function is the same.
After #2410 we'll be able to assume we're always using the v1beta1 types
and will not need the logic in both places and can avoid bugs like this.

It also duplicates the DAG test logic which is the closest thing Deps()
currently has to a set of unit tests.

**2. Allow results to contain underscores**

When results parsing was implemented, the set of allowed characters
wasn't made to match the rest of our substitutions. This duplicates the
set of characters though hopefully we can eventually consolidate this
logic.

I discovered this when trying to use the kaniko task in the catalog
https://github.com/tektoncd/catalog/blob/v1beta1/kaniko/kaniko.yaml

**3. Determine correct graph when params and results are mixed 🥄**

This was partially addressed in #2387 however the example from the
bug (#2387) was using runAfter, and it should be possible to determine
the correct graph without runAfter. The logic that was determining the
graph was returning an error if any of the substitutions present in a
string along with a result were not results. Now we will instead ignore
these substitutions. We might want to revisit the validation to make
sure that invalid results substitutions are still caught.

Fixes #2464

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Fix 3 bugs with results:
- Determine correct graph when results variable replacement is mixed with other types
- Allow results replacement to work with all valid characters
- Use the same graph logic when using embedded Pipeline specs vs referring to Pipeline CRDs
```


bobcatfish:
Fixes #2463

PR #1370, opened 2019-09-30
Only mount artifact bucket volume once, even with multiple inputs.

jbarrick-mesosphere:
# Changes

This fixes an issue where an artifact bucket secret can be mounted multiple times if there are multiple inputs that copy
from another task.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

# Release Notes

```
TaskRuns no longer error when using artifact buckets and multiple sources.
```

PR #2472, opened 2020-04-23
taskrun/resources: allow input and output storage resources

gpaul:
# Changes

This fixes an issue where a Task that uses a secret for both input and output resources leads to duplicate volumes in the generated Pod. 

This happens because both `resources.AddInputResource()` and `resources.AddOutputResources()` mount the necessary secrets to load artifacts. However, when multiple resources use the same secret the secret volume is added twice and leads to an invalid PodSpec.

This fix extends the work done in https://github.com/tektoncd/pipeline/pull/1370

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
TaskRuns no longer error when using artifact buckets as input and output resources.
```

### DIFF
--- a/pkg/reconciler/taskrun/resources/volume.go
+++ b/pkg/reconciler/taskrun/resources/volume.go
@@ -29,3 +29,42 @@ func GetPVCVolume(name string) corev1.Volume {
 		},
 	}
 }
+
+// appendNewSecretsVolumes takes a variadic list of secret volumes and a list of volumes, and
+// appends any secret volumes that aren't already present. The secret volumes are volumes whose
+// VolumeSource is set to *corev1.SecretVolumeSource with only the SecretName field filled in.
+// Specifically, they have the following structure as defined by
+// (pkg/apis/resource/v1alpha1/storage.ArtifactBucket).GetSecretsVolumes():
+//
+// corev1.Volume{
+//   Name: fmt.Sprintf("volume-bucket-%s", sec.SecretName),
+//   VolumeSource: corev1.VolumeSource{
+//     Secret: &corev1.SecretVolumeSource{
+//       SecretName: sec.SecretName,
+//     },
+//   },
+// }
+//
+// Any new volumes that don't match this structure are added regardless of whether they are already
+// present in the list of volumes.
+func appendNewSecretsVolumes(vols []corev1.Volume, newVols ...corev1.Volume) []corev1.Volume {
+	for _, newv := range newVols {
+		alreadyExists := false
+		for _, oldv := range vols {
+			if newv.Name != oldv.Name {
+				continue
+			}
+			if oldv.Secret == nil {
+				continue
+			}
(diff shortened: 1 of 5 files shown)


gpaul:
I'm perfectly happy with that. I'm no fan of reflect.DeepEqual either. I'll push an update.

gpaul:
Are these failures flakes? I don't see how my PR could lead to these failures.

sbwsg:
> Are these failures flakes? I don't see how my PR could lead to these failures.

There's definitely a possibility of that, yeah.  Something seems to be up with our CI recently.

gpaul:
It seems that this strategy also fixes the case where multiple storage output resources are used.

pkg/reconciler/taskrun/resources/volume.go
```diff
@@ -29,3 +29,45 @@ func GetPVCVolume(name string) corev1.Volume {
...
+//   VolumeSource: corev1.VolumeSource{
+//     Secret: &corev1.SecretVolumeSource{
+//       SecretName: sec.SecretName,
+//     },
+//   },
+// }
+func appendNewSecretsVolumes(vols []corev1.Volume, newVols ...corev1.Volume) []corev1.Volume {
+	for _, newv := range newVols {
+		if newv.Secret == nil {
+			panic("appendNewSecretsVolumes: VolumeSource is not a secret")
```
sbwsg, line 50: Would definitely prefer not to panic in the controller.  Can we just silently add non-secret volumes? And / or maybe return a `multierror`?

pkg/reconciler/taskrun/resources/volume.go
```diff
@@ -29,3 +29,45 @@ func GetPVCVolume(name string) corev1.Volume {
...
+		}
+		if newv.Secret.DefaultMode != nil || len(newv.Secret.Items) != 0 {
+			panic("appendNewSecretsVolumes: unexpected structure for VolumeSource.Secret")
```
sbwsg, line 53: same comment here regarding panics.

pkg/reconciler/taskrun/resources/volume.go
gpaul: Good idea. Adding non-secret volumes, even if that means there are duplicates, is still strictly better than the status quo. That makes this PR de-duplicate only secret volumes with the expected structure, which is targets the problem more specifically.

pkg/reconciler/taskrun/resources/volume.go
gpaul: Addressed in my latest commit :+1:

sbwsg:
One further request: could you squash your commits into 1 and make the commit message the same as your PR description?  Thanks again!

gpaul:
> One further request: could you squash your commits into 1 and make the commit message the same as your PR description? Thanks again!

Certainly, doing so now.

gpaul:
Bump

gpaul:
@sbwsg would you mind also adding a /lgtm?

PR #2475, opened 2020-04-23
Fix artifact log

eddycharly:
Fix a call to the logger that should be placed inside a if (artifact_storage).


eddycharly:
TestHelmDeployPipelineRun is failing again ?

eddycharly:
Ok, because github was down

PR #2480, opened 2020-04-23
Support overriding the webhook names in order to support multiple Tekton installations

jbarrick-mesosphere:
# Changes

Previously the webhook names were hardcoded, which caused issues if the webhook name was changed:

```
{"level":"error","logger":"webhook","caller":"controller/controller.go:376","msg":"Reconcile error","commit":"82e2758","error":"error retrieving webhook: mutatingwebhookconfiguration.admissionregistration.k8s.io \"webhook.pipeline.tekton.dev\" not found","stacktrace":"github.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller.(*Impl).handleErr\n\tgithub.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller/controller.go:376\ngithub.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller.(*Impl).processNextWorkItem\n\tgithub.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller/controller.go:362\ngithub.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller.(*Impl).Run.func2\n\tgithub.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller/controller.go:310"}
```
# Submitter Checklist

review them:

# Release Notes

```
The webhook names are now configurable via the `MUTATING_WEBHOOK_NAME`, `VALIDATION_WEBHOOK_NAME`, and `CONFIG_WEBHOOK_NAME` environment variables.
```

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -60,6 +60,14 @@ var types = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
 	v1beta1.SchemeGroupVersion.WithKind("PipelineRun"): &v1beta1.PipelineRun{},
 }
 
+func defaultEnv(envVar string, defaultValue string) string {
+	envValue := os.Getenv(envVar)
+	if envValue == "" {
+		envValue = defaultValue
+	}
+	return envValue
+}
+
 func newDefaultingAdmissionController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
 	// Decorate contexts with the current state of the config.
 	store := defaultconfig.NewStore(logging.FromContext(ctx).Named("config-store"))
@@ -68,7 +76,7 @@ func newDefaultingAdmissionController(ctx context.Context, cmw configmap.Watcher
 	return defaulting.NewAdmissionController(ctx,
 
 		// Name of the resource webhook.
-		"webhook.pipeline.tekton.dev",
+		defaultEnv("MUTATING_WEBHOOK_NAME", "webhook.pipeline.tekton.dev"),
 
 		// The path on which to serve the webhook.
 		"/defaulting",
@@ -93,7 +101,7 @@ func newValidationAdmissionController(ctx context.Context, cmw configmap.Watcher
 	return validation.NewAdmissionController(ctx,
 
 		// Name of the resource webhook.
-		"validation.webhook.pipeline.tekton.dev",
+		defaultEnv("VALIDATION_WEBHOOK_NAME", "validation.webhook.pipeline.tekton.dev"),
 
 		// The path on which to serve the webhook.
 		"/resource-validation",
@@ -115,7 +123,7 @@ func newConfigValidationController(ctx context.Context, cmw configmap.Watcher) *
 	return configmaps.NewAdmissionController(ctx,
 
 		// Name of the configmap webhook.
-		"config.webhook.pipeline.tekton.dev",
+		defaultEnv("CONFIG_WEBHOOK_NAME", "config.webhook.pipeline.tekton.dev"),
 
 		// The path on which to serve the webhook.
 		"/config-validation",
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -76,7 +76,7 @@ func newDefaultingAdmissionController(ctx context.Context, cmw configmap.Watcher
 	return defaulting.NewAdmissionController(ctx,
 
 		// Name of the resource webhook.
-		defaultEnv("MUTATING_WEBHOOK_NAME", "webhook.pipeline.tekton.dev"),
+		os.Getenv("MUTATING_WEBHOOK_NAME"),
 
 		// The path on which to serve the webhook.
 		"/defaulting",
@@ -101,7 +101,7 @@ func newValidationAdmissionController(ctx context.Context, cmw configmap.Watcher
 	return validation.NewAdmissionController(ctx,
 
 		// Name of the resource webhook.
-		defaultEnv("VALIDATION_WEBHOOK_NAME", "validation.webhook.pipeline.tekton.dev"),
+		os.Getenv("VALIDATION_WEBHOOK_NAME"),
 
 		// The path on which to serve the webhook.
 		"/resource-validation",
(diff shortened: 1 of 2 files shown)


sbwsg:
Could you add entries to the env section of `config/webhook.yaml` with these values?  I think this would remove the need for the `defaultEnv` func since we'll know what the values are just from looking at the webhook deployment / pod.

jbarrick-mesosphere:
> Could you add entries to the env section of `config/webhook.yaml` with these values? I think this would remove the need for the `defaultEnv` func since we'll know what the values are just from looking at the webhook deployment / pod.

That seems good to me (especially for documentation reasons), but if we only include it in the manifest won't we break users who just bump their images (idk if this is something we care to support, but it's worth considering)?

sbwsg:
> if we only include it in the manifest won't we break users who just bump their images (idk if this is something we care to support, but it's worth considering)?

I haven't seen this kind of env defaulting elsewhere in our codebase.  Aren't those same users already going to be broken if they just rely on the images and change values like these around?  I genuinely don't know the answer to this so would be interested to hear the use-case and how it's already catered for.

jbarrick-mesosphere:
I can only speak for myself, but: we have a heavily modified chart so updates are always a matter of manually reconciling the upstream changes. The less there are in the manifests, the easier it is for me. But happy to drop it, it's not a big deal!

sbwsg:
The flip side to this might be that we introduce this kind of env defaulting everywhere to support the use case more explicitly.  I would definitely support that being a PR unto itself if it made life easier for downstream platforms / users!

sbwsg:
I was off base with my comment about not providing default env var values.  We do so even in the same webhook main.go file: https://github.com/tektoncd/pipeline/blob/master/cmd/webhook/main.go#L128-L131

```go
	serviceName := os.Getenv("WEBHOOK_SERVICE_NAME")
	if serviceName == "" {
		serviceName = "tekton-pipelines-webhook"
	}
```

So my feedback re: the `defaultEnv` func is different now - if you want to include it I suggest matching style with what we already have elsewhere.  Otherwise I think this PR just needs to be squashed into a single commit and it will be ready when someone comes along to lgtm.  Thanks for the PR!

eddycharly:
Shall I ask why we use env variables everywhere instead of regular command line arguments ?
It makes it a lot harder to find and document what is supported.

sbwsg:
I don't really know the answer other than: it was like that when I got here. 🤷

vdemeester:
> Shall I ask why we use env variables everywhere instead of regular command line arguments ?
> It makes it a lot harder to find and document what is supported.

@eddycharly a bit as @sbwsg, if we could rewrite history, I would go for `flags` but, it is there now (`no is temporary, yes is forever` :sweat_smile:)

eddycharly:
@vdemeester i wonder if it would be acceptable to use `flags` when writing new code or should we stick with env variables ?

vdemeester:
> @vdemeester i wonder if it would be acceptable to use `flags` when writing new code or should we stick with env variables ?

I am voting for consistency :wine_glass: so we should stick with env variables for now, adding flags if we feel it make senses… Or as put differently, as it is consistent with what we are doing elsewhere, I am :+1: :stuck_out_tongue:

vdemeester:
ping @jbarrick-mesosphere for the CLA :pray:

jbarrick-mesosphere:
Hrm, I can't seem to get EasyCLA to work, I select my company and nothing happens. Any ideas?

vdemeester:
> Hrm, I can't seem to get EasyCLA to work, I select my company and nothing happens. Any ideas?

You need to check with your company (I guess mesosphere) if you need to sign the individual CLA or if they need to setup the corporate CLA. Depending on the answer you can either sign the individual one or wait for them to setup the corporate one (see https://www.linuxfoundation.org/blog/2019/07/easycla-beta/ and https://github.com/communitybridge/easycla/tree/master/getting-started#corporate-cla-signatory)

cmd/webhook/main.go
```diff
@@ -68,7 +76,7 @@ func newDefaultingAdmissionController(ctx context.Context, cmw configmap.Watcher
 	return defaulting.NewAdmissionController(ctx,
 
 		// Name of the resource webhook.
-		"webhook.pipeline.tekton.dev",
+		os.Getenv("MUTATING_WEBHOOK_NAME"),
```
sbwsg, line 79: looks like `defaultEnv` is back but unused atm.  If you want to use that here and on the modified lines below then +1  otherwise remove `defaultEnv` and I think this is ready to lgtm!  Thanks and sorry for all the delays @jbarrick-mesosphere

sbwsg:
ah also just noticed it needs to be squashed, then good to go!

ImJasonH:
/hold

I'm curious about your use case. I don't think it's possible to meaningfully have two copies of the webhook running, since they'll always both be notified on every change, and they'll both duplicate any validation/conversion/etc.

This change by itself seems innocuous -- just making the webhook name configurable -- but the underlying use case of being able to run multiple webhooks simultaneously will require more work.

vdemeester:
ping @jbarrick-mesosphere

chhsia0:
Here's our use case:
We want to deploy two Tekton instances monitoring two different namespaces. Each instance has its own configurations, so the defaulting webhooks should only mutate the CRD objects in their respective namespaces based on their own config maps.

Meanwhile, I'm proposing that we can have a way to deploy a cluster-wide webhook in "conversion-only" mode for CRD conversion: https://github.com/tektoncd/pipeline/issues/2696

jbarrick-mesosphere:
@ImJasonH you can set a namespace selector on a webhook configuration so that it only applies to certain namespaces, so we had it configured so that each installation of Tekton had its own webhook that was matched based on labels we set on the namespaces.

sbwsg:
@ImJasonH given response from @jbarrick-mesosphere (to reiterate: webhooks can be scoped to specific namespaces which is what they're doing), are you happy for this PR to move ahead?

ImJasonH:
I'm fine with the change, as before, but I think there will be more work needed to fully support the use case.

If we decide it's a use case Tekton should commit to supporting we'll also need tests to guard against future breakages.

But as far as this change I have no concerns.

sbwsg:
Cheers @ImJasonH !

/hold cancel

sbwsg:
@jbarrick-mesosphere do you want to carry on with this PR?  If so it looks like `defaultEnv()` either needs to be removed or kept and put to use instead of the `os.GetEnv` calls.

sbwsg:
I'm going to close this for now.  Please feel free to re-open if you'd like to pick it up again and finish it off.  Thanks for the PR!

PR #2491, opened 2020-04-24
Add subPath to WorkspacePipelineTaskBinding

### LINKED ISSUE #2490: Add subPath to PipelineTask so instances of tasks can write to different paths of ws
# Expected Behavior

That I can use multiple instances of a task that writes to the workspace volume, but to different paths.

E.g. one **git-clone** task that clone the _app-code-repository_ and another **git-clone** task that clone the _app-config-repository_ and that I in a later task can use files from both git-clone tasks.

Another use case is to **git-clone** code into a directory of workspace volume, and in addition fetch cached dependencies from a GCS bucket into **another directory** of the workspace volume.

# Actual Behavior

This can currently be done using two workspaces of type `PersistentVolumeClaim` in combination with different values of `subPath`, e.g.

```
workspaces:
- name: workspace-1
  persistentVolumeClaim:
    claimName: mypvc
  subPath: subdir-1
- name: workspace-2
  persistentVolumeClaim:
    claimName: mypvc
  subPath: subdir-2
```

but when using a workspace of type `volumeClaimTemplate` there is currently no way of doing this, since providing **two templates** gives you **two different volumes**

```
workspaces:
- name: workspace-1
  volumeClaimTemplate:
    spec:
      accessModes: 
      - ReadWriteOnce
      resources:
        requests:
          storage: 1Gi
- name: workspace-2
  volumeClaimTemplate:
    spec:
      accessModes: 
      - ReadWriteOnce
      resources:
        requests:
          storage: 1Gi
```

# Proposed Solution

If we add the field `subPath` to the `WorkspacePipelineTaskBinding` - different instances of a task can be bound to different _subdirectories_ of the _workspace volume_.  Then this is done when the pipeline binds the workspace to the **instance of the task** in the `Pipeline`.

Example `Pipeline`

```
kind: Pipeline
metadata:
  name: pipeline
spec:
  tasks:
  - name: writer-1
    taskRef:
      name: writer
    workspaces:
    - name: task-ws
      workspace: ws
      subPath: dir-1
  - name: writer-2
    taskRef:
      name: writer
    workspaces:
    - name: task-ws
      workspace: ws
      subPath: dir-2
  - name: read-all
    runAfter:
    - writer-1
    - writer-2
    taskSpec:
      steps:
      - name: read-1
        image: ubuntu
        script: cat $(workspaces.local-ws.path)/dir-1/foo | grep bar
      - name: read-2
        image: ubuntu
        script: cat $(workspaces.local-ws.path)/dir-2/foo | grep bar
      workspaces:
      - name: local-ws
    workspaces:
    - name: local-ws
      workspace: ws
  workspaces:
  - name: ws
---
kind: PipelineRun
metadata:
  generateName: pr-
spec:
  pipelineRef:
    name: pipeline
  workspaces:
  - name: ws
    volumeClaimTemplate:
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 1Gi
```

/kind feature

jlpettersson:
# Changes

**Use case:** Use two instances of a task that writes to the same workspace volume - but
they write to different path of the volume. A typical use case is two git-clone tasks
that clone two git repositories, but the files should be located in two different
directories on the workspace volume.

This can currently be solved when using `PersistentVolumeClaim` based workspace, by declaring multiple workspaces using the same PVC but different `subPath` (in the WorkspaceBinding of *Run). But that does not work when using a `VolumeClaimTemplate` based workspace.

**This commit** solves this by adding the `subPath` field to the WorkspacePipelineTaskBinding.

If the WorkspaceBinding in `PipelineRun` also has a `subPath` set, they will be appended, e.g. `subPath/subPath`. This was discussed in the API WG.

Example log output from the example:

```
$ tkn pr logs pr-gj4l9
[writer-1 : write] + echo bar

[writer-2 : write] + echo bar

[read-all : read-1] + cat /workspace/local-ws/dir-1/foo
[read-all : read-1] + grep bar
[read-all : read-1] bar

[read-all : read-2] + cat /workspace/local-ws/dir-2/foo
[read-all : read-2] + grep bar
[read-all : read-2] bar

```

Fixes #2490

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -19,6 +19,7 @@ package pipelinerun
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"reflect"
 	"strings"
 	"time"
@@ -733,9 +734,9 @@ func (c *Reconciler) createTaskRun(rprt *resources.ResolvedPipelineRunTask, pr *
 		pipelineRunWorkspaces[binding.Name] = binding
 	}
 	for _, ws := range rprt.PipelineTask.Workspaces {
-		taskWorkspaceName, pipelineWorkspaceName := ws.Name, ws.Workspace
+		taskWorkspaceName, pipelineTaskSubPath, pipelineWorkspaceName := ws.Name, ws.SubPath, ws.Workspace
 		if b, hasBinding := pipelineRunWorkspaces[pipelineWorkspaceName]; hasBinding {
-			tr.Spec.Workspaces = append(tr.Spec.Workspaces, taskWorkspaceByWorkspaceVolumeSource(b, taskWorkspaceName, pr.GetOwnerReference()))
+			tr.Spec.Workspaces = append(tr.Spec.Workspaces, taskWorkspaceByWorkspaceVolumeSource(b, taskWorkspaceName, pipelineTaskSubPath, pr.GetOwnerReference()))
 		} else {
 			return nil, fmt.Errorf("expected workspace %q to be provided by pipelinerun for pipeline task %q", pipelineWorkspaceName, rprt.PipelineTask.Name)
 		}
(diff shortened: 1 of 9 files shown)


jlpettersson:
This should probably be on hold until it is discussed

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -766,6 +767,18 @@ func taskWorkspaceByWorkspaceVolumeSource(wb v1alpha1.WorkspaceBinding, taskWork
...
 
+// combinedSubPath returns the combined value of the optional subPath from workspaceBinding and the optional
+// subPath from pipelineTask. If both is set, they are joined with a slash.
+func combinedSubPath(workspaceSubPath string, pipelineTaskSubPath string) string {
+	if workspaceSubPath == "" {
+		return pipelineTaskSubPath
+	} else if pipelineTaskSubPath == "" {
+		return workspaceSubPath
+	} else {
+		return fmt.Sprintf("%s/%s", workspaceSubPath, pipelineTaskSubPath)
```
sbwsg, line 778: suggest [`filepath.Join` here](https://golang.org/pkg/path/filepath/#Join), which will also clean the paths in case extra slashes get included at the end / beginning of either the strings.

jlpettersson:
This was discussed in the API WG today, and the code is now updated to reflect the discussion, e.g. append subPath from PipelineRun subPath and PipelineTask subPath if both is set.

I think we can remove the hold now.

pkg/reconciler/pipelinerun/pipelinerun.go
jlpettersson: Thanks, sounds like a good suggestion! I didn't know about it.

Issue #2497, opened 2020-04-27
Proposal: Use standard k8s labels


eddycharly:
## What
Use standard/recommended k8s labels to identify applications/components resources (https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)

## Why
Tekton applications need to find resources belonging to other applications (dashboard needs to talk to pipelines and triggers, triggers needs to talk to pipelines, etc.).

The current process for identifying resources in dependent applications is lacking flexibility, often relying on resource names (see https://github.com/tektoncd/dashboard/blob/83f7c6a9746470f5fb391a160779d855f8db8e00/pkg/endpoints/dashboard.go#L110-L130 for example).

Also, version labels are not completely clear to me:https://github.com/tektoncd/pipeline/blob/c11c6afa4904216d7b1c7a7a44c0b703e6093ccd/config/controller.yaml#L23-L24
What is the difference between `pipeline.tekton.dev/release` and `version` ?

## How
We could establish a clear list of applications/components labels and document them.
These labels should be set on all resources and we should use them to lookup the necessary resources instead of relying on names.

Such a list could look something like this:
| application | component | app.kubernetes.io/name | app.kubernetes.io/component | app.kubernetes.io/version |
| --- | --- | --- | --- | --- |
| pipelines | controller | `tekton-pipelines` | `controller` | `v0.11.3` |
| pipelines | webhook | `tekton-pipelines` | `webhook` | `v0.11.3` |
| triggers | controller | `tekton-triggers` | `controller` | `v0.4.0` |
| triggers | webhook | `tekton-triggers` | `webhook` | `v0.4.0` |
| dashboard | controller | `tekton-dashboard` | `controller` | `v0.6.1` |
| webhooks-extension | extension | `tekton-webhooks-extension` | `extension` | `v0.6.1` |
| webhooks-extension | validator | `tekton-webhooks-extension` | `validator` | `v0.6.1` |

Other labels (`app.kubernetes.io/instance`, `app.kubernetes.io/part-of`, and `app.kubernetes.io/managed-by`) could be used to reflect the deployment instance and the technology used to deploy (kubectl/helm).

## Benefits
The benefits would be:
- Normalized labelling
- More flexible lookup process
- Easy to document/test

vdemeester:
In general, I am +:100: on this :wink: 
 
> Also, version labels are not completely clear to me:
> 
[...]

PR #2501, opened 2020-04-28
Add recommended k8s labels

eddycharly:
This PR adds recommended k8s labels on resources.
Implementation for proposal #2497 

Config: added labels and updated selectors for all resources in `/config`
Doc: added `/docs/developers/resources-labelling.md`

**NOTE**: this will introduce a breaking change when trying to update deployments as label selectors have changed.

### DIFF
--- a/config/101-podsecuritypolicy.yaml
+++ b/config/101-podsecuritypolicy.yaml
@@ -16,6 +16,9 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: tekton-pipelines
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipelines
 spec:
   privileged: false
   allowPrivilegeEscalation: false
--- a/config/100-namespace.yaml
+++ b/config/100-namespace.yaml
@@ -16,3 +16,6 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: tekton-pipelines
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipelines
--- a/config/200-clusterrole.yaml
+++ b/config/200-clusterrole.yaml
@@ -16,6 +16,10 @@ kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: tekton-pipelines-controller-cluster-access
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipelines
 rules:
   - apiGroups: [""]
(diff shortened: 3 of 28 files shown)


config/101-podsecuritypolicy.yaml
```diff
@@ -16,6 +16,11 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
   name: tekton-pipelines
+  labels:
+    app.kubernetes.io/name: psp
```
vdemeester, line 20: Any reason for `psp` instead of a longer name ?

eddycharly: Nope, no particular reason, do you prefer a longer name ?
I don't think anyone will need to look this up anyway, i have set a name on every resource for consistency though.
Not completely sure if making the `app.kubernetes.io/name` mandatory for every resource makes sense to be honest.

eddycharly:
/hold
I feel like there are probably some superfluous labels

config/101-podsecuritypolicy.yaml
eddycharly: I removed the label where it was not necessary

eddycharly:
/hold cancel
I removed `app.kubernetes.io/name` and `app.kubernetes.io/version` when they had no value.

@vdemeester how about using `app.kubernetes.io/version` everywhere instead of `pipeline.tekton.dev/release`, `triggers.tekton.dev/release`, `dashboard.tekton.dev/release`, etc ?

vdemeester:
> /hold cancel
> I removed `app.kubernetes.io/name` and `app.kubernetes.io/version` when they had no value.
> 
> @vdemeester how about using `app.kubernetes.io/version` everywhere instead of `pipeline.tekton.dev/release`, `triggers.tekton.dev/release`, `dashboard.tekton.dev/release`, etc ?

yeah we could, but we will need to keep those (at least for a while) there as tools depend on them.

eddycharly:
I know the dashboard depends on it, i am adding support for `app.kubernetes.io/veersion` there.
Any other tool ?

vdemeester:
> I know the dashboard depends on it, i am adding support for `app.kubernetes.io/veersion` there.
> Any other tool ?

- `tkn`
- OpenShift Pipeline
- most likely a bunch of other tools that we do not control :stuck_out_tongue_closed_eyes: 

We should add the new one, mark the old one as "deprecated" (in release note) and decide on a time where it gets removed (if we want to remove it :stuck_out_tongue: )

eddycharly:
@vdemeester do you want to include changes from https://github.com/tektoncd/pipeline/pull/2508 in this PR ?

vdemeester:
> @vdemeester do you want to include changes from #2508 in this PR ?

oh yeah we can :+1:

eddycharly:
Done, i took the `objectSelector` approach, hope it's ok for you.

PR #2516, opened 2020-04-30
Tighten up the language for clarity and flow.

sergetron:
Improves and standardizes the wording throughout the document.

### DIFF
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -4,108 +4,107 @@ linkTitle: "Variable Substitutions"
 weight: 15
 ---
 -->
-# Variable Substitutions
+# Variable Substitutions Supported by `Tasks` and `Pipelines`
 
-This doc aggregates the complete set of variable substitions available
-in `Tasks` and `Pipelines`.
+This page documents the variable substitions supported by `Tasks` and `Pipelines`.
 
-## Variables Available in a Pipeline
-
-| Variable | Description |
-| -------- | ----------- |
-| `params.<param name>` | The value of the param at runtime. |
-| `tasks.<task name>.results.<result name>` | The value of a Task's result (**Note**: Affects Task ordering in a Pipeline!) |
-
-## Variables Available in a Task
+## Variables available in a `Pipeline`
 
 | Variable | Description |
 | -------- | ----------- |
-| `params.<param name>` | The value of the param at runtime. |
-| `resources.inputs.<resource name>.path` | The path to the input resource's directory. |
-| `resources.outputs.<resource name>.path` | The path to the output resource's directory. |
-| `results.<result name>.path` | The path to the file where a Task's result should be written. |
-| `workspaces.<workspace name>.path` | The path to the mounted workspace. |
-| `workspaces.<workspace name>.claim` | The name of the PersistentVolumeClaim used as a volume source for the workspace or empty string if a volume source other than PersistentVolumeClaim was used. |
-| `workspaces.<workspace name>.volume` | The name of the volume populating the workspace. |
-| `credentials.path` | The path to the location of credentials written by the `creds-init` init container. |
+| `params.<param name>` | The value of the parameter at runtime. |
+| `tasks.<taskName>.results.<resultName>` | The value of the `Task's` result. Can alter `Task` execution order within a `Pipeline`.) |
 
-### PipelineResource Variables
-
-Each PipelineResource exposes its own set of variables. Below the variables are grouped by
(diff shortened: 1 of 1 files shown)


docs/variables.md
```diff
@@ -4,108 +4,106 @@ linkTitle: "Variable Substitutions"
...
-| `params.<param name>` | The value of the param at runtime. |
+| `params.<param name>` | The value of the parameter at runtime. |
 | `resources.inputs.<resource name>.path` | The path to the input resource's directory. |
 | `resources.outputs.<resource name>.path` | The path to the output resource's directory. |
-| `results.<result name>.path` | The path to the file where a Task's result should be written. |
-| `workspaces.<workspace name>.path` | The path to the mounted workspace. |
-| `workspaces.<workspace name>.claim` | The name of the PersistentVolumeClaim used as a volume source for the workspace or empty string if a volume source other than PersistentVolumeClaim was used. |
-| `workspaces.<workspace name>.volume` | The name of the volume populating the workspace. |
-| `credentials.path` | The path to the location of credentials written by the `creds-init` init container. |
+| `results.<result name>.path` | The path to the file that stores the `Task's` result. |
```
sbwsg, line 25: Ideally what we'd communicate here is that a Step in a Task needs to write a file at this path in order to get some data exposed as a result in the TaskRun. It's kind of a call-to-action, "Write a file here in your Step to expose its contents as a result in the TaskRun."

Maybe this already describes that and I'm just reading it with the wrong emphasis.  I think the problem might be that "The path to the file" is the focus of the sentence whereas ideally the action to be taken, Writing to the file at that path, would be the focus.  Does this make any sense?

Either way, not a blocker at all.

docs/variables.md
```diff
@@ -4,108 +4,106 @@ linkTitle: "Variable Substitutions"
...
-| `sslVerify` | The value of the resource's `sslVerify` param: `"true"` or `"false"`. |
-| `httpProxy` | The value of the resource's `httpProxy` param. |
-| `httpsProxy` | The value of the resource's `httpsProxy` param. |
-| `noProxy` | The value of the resource's `noProxy` param. |
-
-#### PullRequest PipelineResource
+| `name` | The name of the resource. |
+| `type` | Type value of `"git"`. |
+| `url` | The URL of the Git repository. |
+| `revision` | The revision to check out. |
```
sbwsg, line 45: There's another one to add here which was only recently introduced in 0.12: `refspec`.  Description would be something like "The value of the resource's `refspec` parameter."

docs/variables.md
```diff
@@ -4,108 +4,106 @@ linkTitle: "Variable Substitutions"
...
 
 | Variable | Description |
 | -------- | ----------- |
-| `name` | The resource's name |
-| `type` | `"image"` |
+| `name` | The name of the resource. |
+| `type` | Type value of `"image"`. |
 | `url` | The complete path to the image. |
-| `digest` | The image's digest. |
+| `digest` | The digest for the image. |
```
sbwsg, line 69: v minor: suggest `of` instead of `for` here.  It's an artifact of the image's content (or, more accurately, a hash of the hashes of each layer of content in the image).

docs/variables.md
```diff
@@ -4,108 +4,106 @@ linkTitle: "Variable Substitutions"
...
+#### Variables for the `GCS` type
 
 | Variable | Description |
 | -------- | ----------- |
-| `name` | The resource's name |
-| `type` | `"gcs"` |
-| `location` | The location of the blob storage. |
+| `name` | The name of the resource. |
+| `type` | Type value of `"gcs"`. |
+| `location` | The path to the blob storage. |
```
sbwsg, line 77: V minor again: technically it'll be a fully-qualified address like `gs://my-bucket/foo/bar`

docs/variables.md
sergetron: No problem. Reworded to "The path to the file where the `Task` writes its results data." Let me know if that works.

docs/variables.md
sergetron: Added.

docs/variables.md
sergetron: Reworded. :)

docs/variables.md
sbwsg: That's great, thankyou!

docs/variables.md
sergetron: Fixed. Missed that one in my last commit.

PR #2521, opened 2020-05-01
Document how PVC access modes affect pipeline execution

jlpettersson:
# Changes

The access mode configured for a PVC used as a workspace volume source
may affect how a pipeline is executed, this is not well documented. This commit
intend to improve the documentation about this and also provide an example
on how to run parallel tasks when using PVC with ReadWriteOnce access mode.

- Document how access mode affect task ordering in pipeline under "Specifying Workspace order in a Pipeline"
- Provide a full pipeline example of how to use parallel tasks when using a PVC with access mode ReadWriteOnce
- We do not provide a "PipelineRun example" under "Example PipelineRun definitions using Workspaces". This commit provide a full PipelineRun example using a workspace. Instead of providing examples of different volume sources here, examples on how to use different VolumeSources is moved to "Specifying VolumeSources in Workspaces"
- The VolumeSources persistentVolumeClaim and volumeClaimTemplate both is a PVC, and PVCs has its own peculiarities, e.g. access mode, that we now document. Both PVC volume sources is moved to a section so we can document the common peculiarities in a single place
- Add the workspace variable introduced in #2506

# Submitter Checklist

review them:

## Reviewer Notes

/kind documentation

### DIFF
--- a/docs/workspaces.md
+++ b/docs/workspaces.md
@@ -17,8 +17,10 @@ weight: 5
   - [Using `Workspaces` in `Pipelines`](#using-workspaces-in-pipelines)
     - [Specifying `Workspace` order in a `Pipeline`](#specifying-workspace-order-in-a-pipeline)
     - [Specifying `Workspaces` in `PipelineRuns`](#specifying-workspaces-in-pipelineruns)
-    - [Example `PipelineRun` definitions using `Workspaces`](#example-pipelinerun-definitions-using-workspaces)
+    - [Example `PipelineRun` definition using `Workspaces`](#example-pipelinerun-definitions-using-workspaces)
   - [Specifying `VolumeSources` in `Workspaces`](#specifying-volumesources-in-workspaces)
+    - [Using `PersistentVolumeClaims` as `VolumeSource`](#using-persistentvolumeclaims-as-volumesource)
+    - [Using other types of `VolumeSources`](#using-other-types-of-volumesources)
 - [More examples](#more-examples)
 
 ## Overview
@@ -118,6 +120,8 @@ The following variables make information about `Workspaces` available to `Tasks`
 
 - `$(workspaces.<name>.path)` - specifies the path to a `Workspace`
    where `<name>` is the name of the `Workspace`.
+- `$(workspaces.<name>.claim)` - specifies the name of the `PersistentVolumeClaim` used as a volume source for the `Workspace` 
+   where `<name>` is the name of the `Workspace`. If a volume source other than `PersistentVolumeClaim` is used, an empty string is returned.
 - `$(workspaces.<name>.volume)`- specifies the name of the `Volume`
    provided for a `Workspace` where `<name>` is the name of the `Workspace`.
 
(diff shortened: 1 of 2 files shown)


docs/workspaces.md
```diff
@@ -118,6 +120,8 @@ The following variables make information about `Workspaces` available to `Tasks`
 
 - `$(workspaces.<name>.path)` - specifies the path to a `Workspace`
    where `<name>` is the name of the `Workspace`.
+- `$(workspaces.<name>.claim)` - specifies the name of the `PersistentVolumeClaim` used as a volume source for the `Workspace` 
+   where `<name>` is the name of the `Workspace`. If the volume source other than `PersistentVolumeClaim` is used, an empty string is returned.
```
sbwsg, line 124: nit: `If the volume` -> `If a volume`

docs/workspaces.md
```diff
@@ -244,20 +248,28 @@ The `subPath` specified in a `Pipeline` will be appended to any `subPath` specif
...
-single `Task` writing to it at once.
-
-**Warning:** You *must* ensure that this order is correct. Incorrectly ordering can result
-in a deadlock where multiple `Task` `Pods` are attempting to mount a `PersistentVolumeClaim`
-for writing at the same time, which would cause the `Tasks` to time out.
-
-To define this order, use the `runAfter` field in your `Pipeline` definition. For more
+for concurrent reads and writes. For example, a `PersistentVolumeClaim` with
+[access mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes)
+`ReadWriteOnce` only allow `Tasks` on a the same node writing to it at once.
```
sbwsg, line 253: `on a the same` -> `on the same`

docs/workspaces.md
```diff
@@ -244,20 +248,28 @@ The `subPath` specified in a `Pipeline` will be appended to any `subPath` specif
...
+[access mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes)
+`ReadWriteMany` or `ReadOnlyMany` but you must ensure that those are available for your storage class.
+When using `PersistentVolumeClaims` with access mode `ReadWriteOnce` for parallel `Tasks`, you can configure a
+workspace with it's own `PersistentVolumeClaim` for each parallel `Task`. See a full example of a 
+[Pipeline with parallel tasks using PersistentVolumeClaims](../examples/v1beta1/pipelineruns/pipelinerun-with-parallel-tasks-using-pvc.yaml).
+
+Use the `runAfter` field in your `Pipeline` definition to define when a `Task` should be executed. For more
 information, see the [`runAfter` documentation](pipelines.md#runAfter).
 
+**Warning:** You *must* ensure that this order is compatible with your configured access modes for your `PersistentVolumeClaim`.
```
sbwsg, line 265: Suggestion: `with your configured access modes for your` -> `with the configured access modes of your`

docs/workspaces.md
```diff
@@ -244,20 +248,28 @@ The `subPath` specified in a `Pipeline` will be appended to any `subPath` specif
...
+Parallel `Tasks` using the same `PersistentVolumeClaim` with access mode `ReadWriteOnce`, may execute on
+different nodes and be forced to execute sequentially which may cause `Tasks` to time out.
```
sbwsg, line 267: Does the k8s scheduler force this sequential ordering automatically or is it something the _user_ is forced to do by adding `runAfter`?

docs/workspaces.md
```diff
@@ -269,12 +281,52 @@ The entry must also include one `VolumeSource`. See [Using `VolumeSources` with
...
+      volumeClaimTemplate:
+        spec:
+          accessModes:
+            - ReadWriteOnce # access mode may affect how you can use this volume in parallel tasks
+          resources:
+            requests:
+              storage: 1Gi
+```
+
+For examples on how to use other types of volume sources, see [Specifying `VolumeSources` in `Workspaces`](#specifying-volumesources-in-workspaces).
```
sbwsg, line 309: nit: `For examples on how to use other` -> `For examples of using other`

docs/workspaces.md
```diff
@@ -269,12 +281,52 @@ The entry must also include one `VolumeSource`. See [Using `VolumeSources` with
...
+### Specifying `VolumeSources` in `Workspaces`
+
+You can only use a single type of `VolumeSource` per `Workspace` entry. The configuration
+options differ for each type. `Workspaces` support the following fields:
+
+#### Using `PersistentVolumeClaims` as `VolumeSource`
+
+`PersistentVolumeClaim` volumes are a good choice for sharing data among `Tasks` within a `Pipeline`.
+Beware that the [access mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes)
+configured for the `PersinstentVolumeClaim` affect how you can use the volume for parallel `Tasks` in a `Pipeline`. See
```
sbwsg, line 321: nit: `affect` -> `effects`

examples/v1beta1/pipelineruns/pipelinerun-using-different-subpaths-of-workspace.yaml
```diff
@@ -46,6 +46,8 @@ spec:
           workspace: ws
           subPath: dir-1
     - name: writer-2
+      runAfter:
+        - writer-1
```
sbwsg, line 50: Interesting! This example was working already I think?  So here we are just kinda documenting the ordering which the PVC accessMode already forced?

docs/workspaces.md
jlpettersson: The k8s scheduler puts pods on nodes - immediately, they all mount the PVC. But the mount only succeeds on **one node** at the time. Pods on other nodes is stuck in state "Pending" - but they eventually start to run when the volume is not used on other nodes anymore.

And when the pods in the "other nodes" gets into "Running" - the Task is already started - so it may have "timed out" as was already documented.

But on my single node cluster - all tasks can run concurrently, so I did not see that parallel tasks on other nodes may not start until the volume is available.

I did not know how this was working, so I had to create a bigger test to validate this. Me and @skaegi discussed about it in the last API WG.

examples/v1beta1/pipelineruns/pipelinerun-using-different-subpaths-of-workspace.yaml
jlpettersson: Yes, it was "working" but the tasks was declared "parallel". So:

- the two tasks will successfully run concurrently **if scheduled to the same node**
- the two tasks may successfully run sequentially if scheduled to different nodes **and the second task does not time out**
- the second task _may_ time out if it is scheduled to another node and it takes too long time before it can execute/finish (I dont know how we implement time out)

**Consequentially:** If this is used in regression testing - the test may be **flaky** sometimes.

But this fixes so the two tasks always run in a sequence. For parallel tasks we should do as in the example provided in this commit. Unless we know that it will not time out.

sbwsg: Ah, I see, great catch!

jlpettersson:
I don't see why the integration tests fails, but I'll give it a new try...

jlpettersson:
>Error from server (AlreadyExists): error when creating "STDIN": tasks.tekton.dev "writer" already exists

There was actually a problem with naming in my example - name collision with another example file. I changed name on my example task.

PR #2526, opened 2020-05-02
Emit events when we fail to update the taskrun

afrittoli:
# Changes

Similarly to how the pipeline controller does, we should emit events
when we fail to update the taskrun.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
We now emit warning events from the taskrun controller if the controller fails to sync the updated taskrun back.
```

### DIFF
--- a/pkg/reconciler/event.go
+++ b/pkg/reconciler/event.go
@@ -30,6 +30,8 @@ const (
 	EventReasonFailed = "Failed"
 	// EventReasonStarted is the reason set for events about the start of TaskRuns / PipelineRuns
 	EventReasonStarted = "Started"
+	// EventReasonError is the reason set for events related to TaskRuns / PipelineRuns reconcile errors
+	EventReasonError = "Error"
 )
 
 // EmitEvent emits an event for object if afterCondition is different from beforeCondition
@@ -61,5 +63,12 @@ func EmitEvent(c record.EventRecorder, beforeCondition *apis.Condition, afterCon
 				c.Event(object, corev1.EventTypeNormal, afterCondition.Reason, afterCondition.Message)
 			}
 		}
+	}
+}
+
+// EmitErrorEvent emits a failure associated to an error
+func EmitErrorEvent(c record.EventRecorder, err error, object runtime.Object) {
+	if err != nil {
+		c.Event(object, corev1.EventTypeWarning, EventReasonError, err.Error())
 	}
 }
--- a/pkg/reconciler/event_test.go
+++ b/pkg/reconciler/event_test.go
@@ -17,6 +17,9 @@ limitations under the License.
 package reconciler
 
 import (
+	"errors"
+	"fmt"
+	"strings"
 	"testing"
 	"time"
 
@@ -28,11 +31,10 @@ import (
 
 func TestEmitEvent(t *testing.T) {
 	testcases := []struct {
-		name          string
-		before        *apis.Condition
-		after         *apis.Condition
-		expectEvent   bool
-		expectedEvent string
+		name      string
+		before    *apis.Condition
+		after     *apis.Condition
+		wantEvent string
 	}{{
 		name: "unknown to true with message",
 		before: &apis.Condition{
@@ -44,8 +46,7 @@ func TestEmitEvent(t *testing.T) {
 			Status:  corev1.ConditionTrue,
 			Message: "all done",
 		},
-		expectEvent:   true,
-		expectedEvent: "Normal Succeeded all done",
+		wantEvent: "Normal Succeeded all done",
 	}, {
 		name: "true to true",
 		before: &apis.Condition{
@@ -58,8 +59,7 @@ func TestEmitEvent(t *testing.T) {
 			Status:             corev1.ConditionTrue,
 			LastTransitionTime: apis.VolatileTime{Inner: metav1.NewTime(time.Now().Add(5 * time.Minute))},
 		},
-		expectEvent:   false,
-		expectedEvent: "",
+		wantEvent: "",
 	}, {
 		name: "false to false",
 		before: &apis.Condition{
@@ -70,8 +70,7 @@ func TestEmitEvent(t *testing.T) {
 			Type:   apis.ConditionSucceeded,
 			Status: corev1.ConditionFalse,
 		},
-		expectEvent:   false,
-		expectedEvent: "",
+		wantEvent: "",
 	}, {
 		name: "unknown to unknown",
 		before: &apis.Condition{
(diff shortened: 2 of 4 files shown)


pkg/reconciler/event.go
```diff
@@ -63,3 +65,10 @@ func EmitEvent(c record.EventRecorder, beforeCondition *apis.Condition, afterCon
 		}
 	}
 }
+
+// EmitErrorEvent emits a failure associated to an error
+func EmitErrorEvent(c record.EventRecorder, err error, object runtime.Object) {
+	if err != nil {
```
bobcatfish, line 71: feels a bit odd to have this function take an error as an argument vs checking the error in the place where it originates - is this a pattern we do other places as well?

pkg/reconciler/event_test.go
```diff
@@ -127,43 +122,70 @@ func TestEmitEvent(t *testing.T) {
...
-				t.Errorf("Expected event but got none for %s", ts.name)
-			}
+		err := checkEvents(fr, ts.name, ts.wantEvent)
+		if err != nil {
+			t.Errorf(err.Error())
 		}
 	}
 }
+
+func TestEmitErrorEvent(t *testing.T) {
```
bobcatfish, line 148: 🎉 thanks for splitting these up!

pkg/reconciler/event_test.go
```diff
@@ -127,43 +122,70 @@ func TestEmitEvent(t *testing.T) {
...
+	}}
+
+	for _, ts := range testcases {
+		fr := record.NewFakeRecorder(1)
+		tr := &corev1.Pod{}
+		EmitErrorEvent(fr, ts.err, tr)
+
+		err := checkEvents(fr, ts.name, ts.wantEvent)
+		if err != nil {
+			t.Errorf(err.Error())
```
bobcatfish, line 170: might be nice to add some context, e.g. "expected error when blah blah"

pkg/reconciler/event_test.go
```diff
@@ -127,43 +122,70 @@ func TestEmitEvent(t *testing.T) {
...
+		if !(strings.HasPrefix(event, wantEvent)) {
+			return fmt.Errorf("expected event \"%s\" but got \"%s\" instead for %s", wantEvent, event, testName)
+		}
+	case <-timer.C:
+		if wantEvent != "" {
+			return fmt.Errorf("received no events for %s but %s expected", testName, wantEvent)
+		}
+	}
+	return nil
+}
```
bobcatfish, line 191: when this fails, what line number does the test failure produce? would it be useful to use `t.Helper` here?

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -206,10 +200,15 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
-	reconciler.EmitEvent(c.Recorder, before, after, tr)
+	return c.finishReconcileUpdateEmitEvents(tr, original, before, err)
+}
 
-	return multierror.Append(err, c.updateStatusLabelsAndAnnotations(tr, original)).ErrorOrNil()
+func (c *Reconciler) finishReconcileUpdateEmitEvents(tr, original *v1alpha1.TaskRun, beforeCondition *apis.Condition, previousError error) error {
+	afterCondition := tr.Status.GetCondition(apis.ConditionSucceeded)
+	reconciler.EmitEvent(c.Recorder, beforeCondition, afterCondition, tr)
+	err := c.updateStatusLabelsAndAnnotations(tr, original)
+	reconciler.EmitErrorEvent(c.Recorder, err, tr)
```
bobcatfish, line 210: personally i think id rather see an explicit "if err" here vs having EmitErrorEvent handle it, just cuz it doesnt seem consistent with how we usually do error handling - maybe you have a bigger plan around this tho that im not seeing

test/cancel_test.go
```diff
@@ -171,14 +172,22 @@ func TestTaskRunPipelineRunCancel(t *testing.T) {
...
 			if len(events) != expectedNumberOfEvents {
-				t.Fatalf("Expected %d number of successful events from pipelinerun and taskrun but got %d; list of received events : %#v", expectedNumberOfEvents, len(events), events)
+				collectedEvents := ""
+				for i, event := range events {
+					collectedEvents += fmt.Sprintf("%#v", event)
+					if i < (len(events) - 1) {
+						collectedEvents += ", "
+					}
+				}
+				t.Fatalf("Expected %d number of successful events from pipelinerun and taskrun but got %d; list of received events : %#v", expectedNumberOfEvents, len(events), collectedEvents)
```
bobcatfish, line 190: nice!

pkg/reconciler/event.go
afrittoli: Well, it's a bit different because in this case the error is the event that we want to send.

On knative they went a bit further, they have now an event object that also implements the error interface.
They have a generated `Reconcile` that invokes a `ReconcileKind` which returns an event that may be an error - at least that's how I understood it - so this goes a bit in that direction. 

I could check if the `err` is nil at the source, but it felt safer to do here.

afrittoli: https://github.com/knative/pkg/tree/master/injection#generated-reconciler-responsibilities

pkg/reconciler/taskrun/taskrun.go
afrittoli: I discovered https://github.com/knative/pkg/tree/master/injection#generated-reconciler-responsibilities after I wrote this patch. Maybe I should just try and adopt that :P

pkg/reconciler/event.go
afrittoli: My plan for upcoming patches was to centralize the event logic in `event.go` i.e. to receive all requests events / error notifications from the taskrun/pipelinerun controller in there, and to decide there when to emit what, k8s event and/or cloudevent.

pkg/reconciler/event_test.go
afrittoli: Yes, it would be nice, I'll add `t` to the method interface then

pkg/reconciler/event_test.go
afrittoli: The context is added in the error by `checkEvents` already.
I could add `unexpected error when checking for events: %#v", err.Error()` but I'm not sure it would add much.

pkg/reconciler/taskrun/taskrun.go
afrittoli: I'm trying to centralize the logic about what event to send when of what type in events.go, so I included there the check for nil too

bobcatfish: I havent seen https://github.com/knative/pkg/tree/master/injection#generated-reconciler-responsibilities before! Glancing at it I'm not 100% sure what implications it has for this PR, im really glad to see some docs for pkg tho.

> I'm trying to centralize the logic about what event to send when of what type in events.go, so I included there the check for nil too

okay, maybe it will make more sense to me as it evolves

afrittoli:
@bobcatfish I had to rebase because of the stored v1beta1 version :)

PR #2532, opened 2020-05-04
Track deprecated Beta features w/ earliest date of removal

sbwsg:
# Changes

We don't currently have a centralized place to list the
features that Pipelines has deprecated.

This commit adds a deprecations doc which shows the feature
being deprecated, a link to the release where its deprecation
was announced, and the earliest date by which the feature
will be removed entirely in relation to the announcement.

At the moment I've only listed Beta features that have been
announced as deprecated. I haven't included things which are
still in Alpha or were announced as deprecated prior to
v0.11.0-rc1, since the policy for alpha code was simply that
we'd give 1 release of forewarning.

# Submitter Checklist

# Release Notes

```
A list of feature deprecations along with the earliest date of their removal has been added to Pipelines' docs.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,9 @@ a cluster with **Kubernetes version 1.15 or later***.
 _See [our API compatibility policy](api_compatibility_policy.md) for info on the
 stability level of the API._
 
+_See [our Deprecations table](docs/deprecations.md) for features that have been
+deprecated and the earliest date they'll be removed._
+
 ## Migrating
 
 ### v1alpha1 to v1beta1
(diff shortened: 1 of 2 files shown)


bobcatfish:
/approve
/lgtm

Didn't seem worth blocking on but would it makes sense for the deprecation doc to link to https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md and maybe vice versa?

sbwsg:
> Didn't seem worth blocking on but would it makes sense for the deprecation doc to link to https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md and maybe vice versa?

Good call.  We already do link to the policy from this doc (the API Compatibility Policy heading in the table links to it), but we can also add a link from the policy doc to this one.

PR #2537, opened 2020-05-04
Versions for builders and internal package 📦

vdemeester:
# Changes

This adds versions for the builder: v1alpha1 and v1beta1.

This also migrate the builder to `internal/builder` while keeping an
alias on `test/builder` for user of this package (marking it as
deprecated).

Those are in the `internal` package so only this project can refer them (making it easier to remove later on as nobody from the outside will depend on them).
This is in preparation of #2410  as… well… builders are gonna be needed for v1beta1 (or… it will take *ages* to do :joy_cat: ).

This feels big but there is not much net new code:
- It's moving files around (`test/builder` -> `internal/builder/v1alpha1`)
- It's adding a whole new builder package (`internal/builder/v1beta1`) based on `internal/builder/v1alpha1` but with the v1beta1 API

/cc @afrittoli @dibyom @bobcatfish @ImJasonH 

# Submitter Checklist

review them:

## Reviewer Notes


bobcatfish:
@vdemeester i think you tried to explain this to me in chat but I don't quite understand why we need 2 versions of the builders? esp. if we're going to be able to rely on having only 1 type to interact with in the tekton controller, which is my understanding of the finished state of updating the storage types.

> builders are gonna be needed for v1beta1 (or… it will take ages to do 😹 ).

Or is the idea that the next step would be to delete the v1alpha1 builders?

ImJasonH:
I would like to take this opportunity to commit to a goal of phasing out test builders entirely, after these changes are in place. Phasing out test builders shouldn't block "real" work like updating the storage version, but since this will move them to `internal` packages, we should just commit to removing them, with a reasonable deadline.

If this seems reasonable I can write a real issue to track it, with justifications and timelines, and assign it to myself. If not, I'd like to convince you that it _is_ reasonable and still do it. 😄

afrittoli:
> I would like to take this opportunity to commit to a goal of phasing out test builders entirely, after these changes are in place. Phasing out test builders shouldn't block "real" work like updating the storage version, but since this will move them to `internal` packages, we should just commit to removing them, with a reasonable deadline.
> 
> If this seems reasonable I can write a real issue to track it, with justifications and timelines, and assign it to myself. If not, I'd like to convince you that it _is_ reasonable and still do it. 😄

I think test code will be more transparent without the builders. It may be non-obvious in some cases how to build an object, but a good reference example should be enough to cover that.

vdemeester:
> @vdemeester i think you tried to explain this to me in chat but I don't quite understand why we need 2 versions of the builders? esp. if we're going to be able to rely on having only 1 type to interact with in the tekton controller, which is my understanding of the finished state of updating the storage types.
> 
> > builders are gonna be needed for v1beta1 (or… it will take ages to do joy_cat ).
> 
> Or is the idea that the next step would be to delete the v1alpha1 builders?

@bobcatfish There is multiple answer to the "2 versions". There is 2 versions of the APIs, if we do builder for the api it make sense to have 2 version of the builder — I mean, why not, especially if we use them in tests (we are testing v1alpha1 and v1beta1 in e2e for example). But, yes the idea is the "next" step (after #2410) will be to remove the `v1alpha1`  builders (and in general most likely all).


> I would like to take this opportunity to commit to a goal of phasing out test builders entirely, after these changes are in place. Phasing out test builders shouldn't block "real" work like updating the storage version, but since this will move them to `internal` packages, we should just commit to removing them, with a reasonable deadline.
> 
> If this seems reasonable I can write a real issue to track it, with justifications and timelines, and assign it to myself. If not, I'd like to convince you that it _is_ reasonable and still do it. smile

@ImJasonH Yes :+1: We may need alternative for verbose cases (a `fromYaml` methods or whatever), but in a gist I am fine with that.

The main goal here is:
1. make it so that people don't depend on it (hence the `internal`) — we need to put a deprecation notice on `test/builder` so that we can remove it in a few releases.
2. make it more easy to do #2410 – and more quick ; and also more easy to review. 
3. Once #2410 is done, the `v1alpha1` builder should be only used in a few places (most likely only in `test/v1alpha1`)
4. From there on we can phase out the `builder`

> I think test code will be more transparent without the builders. It may be non-obvious in some cases how to build an object, but a good reference example should be enough to cover that.

So the goal is to have test readable and with the least "noise". That said, builders didn't really achieve what I had in mind initially… :sweat:

vdemeester:
The *linting* should be fixed :crossed_fingers:

chmouel:
Hard to review the full thing but from what I understand it is just a migration between alpha to beta until we get rid of the `builders` 

so that sounds good to me,

/lgtm

Issue #1251, opened 2019-08-28
Run CI's YAML tests as a `go test` instead of bash scripts


sbwsg:
# Expected Behavior

YAML tests can be run as part of a `go test` suite, maybe with their own build tag. Failures in the tests should be correctly highlighted in the log output / prow error highlighting.

# Actual Behavior

We currently run YAML tests through a series of bash scripts. Their failure isn't reported accurately by the log output / prow error highlighting - the integration suite will show "0 / 29 failures" but grepping through the log output will indicate that the YAML tests failed or did not complete.

# Steps to Reproduce the Problem

1. Make a PR against Tekton Pipelines
2. Wait for the integration suite to fail
3. Grep through the integration logs for "YAML" and find errors

thomaschandler:
I'd like to work on this! This will be my first contribution, so it will take me some time to ramp up!

bobcatfish:
yaaaaaassss!!!!

![image](https://user-images.githubusercontent.com/432502/85412064-2d88f300-b537-11ea-9b11-93b8ad3adce2.png)

Thanks again for this @thomaschandler and for your patience with me duplicating your efforts and the review being prolonged. Really excited to have this in!!!

Issue #2540, opened 2020-05-04
yaml tests seem to be consistently timing out


bobcatfish:
# Expected Behavior

"yaml tests" should only fail if something is actually wrong

# Actual Behavior

All of the runs for #2531 have failed:

https://tekton-releases.appspot.com/builds/tekton-prow/pr-logs/pull/tektoncd_pipeline/2531/pull-tekton-pipeline-integration-tests/

![image](https://user-images.githubusercontent.com/432502/81013432-1a0bb680-8e29-11ea-9b65-9f00197d712e.png)

And recent runs across PRs it seems most are failing too
https://tekton-releases.appspot.com/builds/tekton-prow/pr-logs/directory/pull-tekton-pipeline-integration-tests

# Steps to Reproduce the Problem

Not sure what's going on yet

# Additional Info

I can't decipher 1..60 sleep 10 for the life of me:

https://github.com/tektoncd/pipeline/blob/a4065de3e0b434c10bfa1fc2edd64eda5d13387a/test/e2e-common.sh#L65-L77

jlpettersson:
I added [example pipelinerun-with-parallel-tasks-using-pvc.yaml](https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun-with-parallel-tasks-using-pvc.yaml) in #2521, few days ago.

Things looks worse after that. But I don't really understand what is causing that. Maybe volumes takes time and there are some time outs? I find it hard to see what test is causing trouble.

vdemeester:
> I added [example pipelinerun-with-parallel-tasks-using-pvc.yaml](https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun-with-parallel-tasks-using-pvc.yaml) in #2521, few days ago.
> 
> Things looks worse after that. But I don't really understand what is causing that. Maybe volumes takes time and there are some time outs? I find it hard to see what test is causing trouble.

Yeah, that's my guess :sweat: 

> I can't decipher 1..60 sleep 10 for the life of me:

it's gonna do 60 loops of 10s to check the status of the pipelinerun (or taskrun), meaning it times out after 10min.

vdemeester:
There is few ways to fix:
- the quick one, add more time (aka do 90 loops)
- the *longer* migrate to go to run those tests

#2541 does the later.

vdemeester:
[...]

PR #2541, opened 2020-05-05
Execute yaml examples via go tests

bobcatfish:
# Changes

In #2540 we are seeing that some yaml tests are timing out, but it's
hard to see what yaml tests are failing. This commit moves the logic out
of bash and into individual go tests - now we will run an individual go
test for each yaml example, completing all v1alpha1 before all v1beta1
and cleaning up in between. The output will still be challenging to read
since it will be interleaved, however the failures should at least
be associated with a specific yaml file.

This also makes it easier to run all tests locally, though if you
interrupt the tests you end up with your cluster in a bad state and it
might be good to update these to execute each example in a separate
namespace (in which case we could run all of v1alpha1 and v1beta1 at the
same time as well!)

I have a feeling this won't work on the first try and that I've still go a few issues to work out, not to mention that the code is a bit icky, esp. since im using t.Helper so profusely.

# Submitter Checklist

review them:

## Reviewer Notes


examples/yaml_test.go
```diff
@@ -0,0 +1,241 @@
...
+	cmd.Env = os.Environ()
+
+	if input != "" {
+		cmd.Stdin = strings.NewReader(input)
+	}
+
+	var stderr, stdout bytes.Buffer
+	cmd.Stderr = &stderr
+	cmd.Stdout = &stdout
+	if err := cmd.Run(); err != nil {
```
ImJasonH, line 61: Maybe we can use `CombinedOutput` instead?

examples/yaml_test.go
```diff
@@ -0,0 +1,241 @@
...
+		"taskruns.tekton.dev",
+		"pipelineruns.tekton.dev",
+		"services",
+		"pods",
+		"configmaps",
+		"secrets",
+		"serviceaccounts",
+		"persistentvolumeclaims",
+	}
+	for _, c := range crdTypes {
```
ImJasonH, line 88: Eventually we should be able to parse the YAML file and create the resource using a client, instead of involving kubectl.

For this specifically we can just use the CRD client to delete the types.

It'd be fine to have a TODO for that for now but that's a better final state I think.

examples/yaml_test.go
```diff
@@ -0,0 +1,241 @@
...
+		if matches, _ := filepath.Match("*.yaml", f.Name()); matches {
+			yamls = append(yamls, f.Name())
+		}
+	}
+	return yamls
+}
+
+// replaceDockerRepo will look in the content f and replace the hard coded docker
+// repo with the on provided via the KO_DOCKER_REPO environment variable
+func replaceDockerRepo(t *testing.T, f string) string {
```
ImJasonH, line 115: Ugh, this is kinda gross. I think we could just drop any test that requires ko-building a package?

examples/yaml_test.go
```diff
@@ -0,0 +1,241 @@
...
+	t.Helper()
+	yaml, err := cmd(t, "kubectl", []string{"get", run, "-o", "yaml"}, "")
+	if err == nil {
+		t.Logf(yaml)
+	}
+}
+
+// pollRun will use kubectl to query the specified run to see if it
+// has completed. It will timeout after timeoutSeconds.
+func pollRun(t *testing.T, run string, wg *sync.WaitGroup) {
```
ImJasonH, line 139: This could Watch instead and maybe end up faster?

test/e2e-tests-yaml.sh
```diff
@@ -32,24 +32,7 @@ set +o pipefail
...
-    fi
-  done
-  # Clean resources
-  delete_pipeline_resources
-  for res in services pods configmaps secrets serviceaccounts persistentvolumeclaims; do
-    kubectl delete --ignore-not-found=true ${res} --all
-  done
-done
-
+failed=$(go test -timeout 15m ./examples)
```
ImJasonH, line 35: 🙌🙌🙌🙌🙌

examples/yaml_test.go
vdemeester: Yes and no. We will need to also test some `kubectl create` commands as this is something the user will do.

vdemeester:
```
level=error msg="Running error: context loading failed: failed to load program with go/packages: could not parse GOARCH and Go compiler in format \"<GOARCH> <compiler>\" from stdout of go command:\nGOROOT=/usr/local/go GOPATH=/home/prow/go GO111MODULE= PWD=/home/prow/go/src/github.com/tektoncd/pipeline go [list -f {{context.GOARCH}} {{context.Compiler}} -tags e2e -mod=vendor -- unsafe]\ndir: /home/prow/go/src/github.com/tektoncd/pipeline\nstdout: <<>>\nstderr: <<go: inconsistent vendoring in /home/prow/go/src/github.com/tektoncd/pipeline:\n\tgithub.com/google/go-containerregistry@v0.0.0-20200310013544-4fe717a9b4cb: is explicitly required in go.mod, but vendor/modules.txt indicates github.com/google/go-containerregistry@v0.0.0-20200115214256-379933c9c22b\n\tgithub.com/pkg/errors@v0.9.1: is explicitly required in go.mod, but vendor/modules.txt indicates github.com/pkg/errors@v0.8.1\n\nrun 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory\n>>"
```

The build failure is interesting :stuck_out_tongue:

examples/yaml_test.go
bobcatfish: TIL! thanks :D

I'm not sure I do want to use combinedoutput in this case tho, cuz im using stderr if the command fails, and if not, im returning stdout. in the error case, dumping both seems fine, but in the success case, not including stderr seems like it makes sense to me. what do you think?

examples/yaml_test.go
bobcatfish: we could - this is what the original yaml tests do tho, do you mind if i keep this in the context of this pull request and remove in another PR?

examples/yaml_test.go
bobcatfish: I'm not 100% sure which is better - if we want to use the clients, we need to load the yamls + find find the right client for the thing being created. i think im like 60% convinced this is better in the long run, but there's also something nice about just blasting it all out and watching it. i do think the part where i look for the word "run" in the ko output is pretty hacky, ill put a comment in about that for sure

examples/yaml_test.go
bobcatfish: hm could you give me an example of how that would work? ive never used kubectl watch - trying it out now it seems like id have to stream output from it which seems a bit more complicated given im calling kubectl with exec.Command - i think im being a bit lazy for sure but this seems pretty much fine for now? i can put in a comment to explore using watch

bobcatfish:
I dunno what was up with that weird prow error but it stopped :D

This should be ready for a real review now! Might still be some kinks to work out...

I'm hoping after this we could migrate these tests to a separate test triggered via prow maybe, tho it's a bit more complicated b/c we'll have to invoke boskos too...

sbwsg:
Looks like this would fix https://github.com/tektoncd/pipeline/issues/1251 ?

Also looks like there's a similarly-intentioned PR here: https://github.com/tektoncd/pipeline/pull/2685 ?

examples/admission/tasks.yaml
```diff
@@ -0,0 +1,83 @@
...
+          workspaces:
+            - name: input-files
+          steps:
+            - name: echo
+              image: ubuntu
+              script: |
+                #!/usr/bin/env bash
+
+                # TODO: assert something about the expected contents of $(params.output)
+                # TODO: assert something about the expected results in the cluster
```
bobcatfish, line 83: oh no

bobcatfish:
> Looks like this would fix #1251 ?

🤦‍♀️ my bad for not seeing that sooner! If this next round of tests passes, im suggesting (https://github.com/tektoncd/pipeline/pull/2685#issuecomment-634058108 ) we merge this and then add improvements from #2685 on top - means more work for @thomaschandler tho :(

test/e2e-tests.sh
```diff
@@ -30,8 +30,9 @@ install_pipeline_crd
 failed=0
 
 # Run the integration tests
-header "Running Go e2e tests"
-go_test_e2e -timeout=20m ./test/... || failed=1
+# TODO HACK HACK HACK HACK
```
sbwsg, line 33: Is this dead code? If not could the comment do a bit more to describe why these lines have been left in but commented out?

examples/examples_test.go
```diff
@@ -0,0 +1,256 @@
...
+	t.Helper()
+	r := os.Getenv("KO_DOCKER_REPO")
+	if r == "" {
+		t.Fatalf("KO_DOCKER_REPO must be set")
+	}
+	read, err := ioutil.ReadFile(f)
+	if err != nil {
+		t.Fatalf("couldnt read contents of %s: %v", f, err)
+	}
+	return strings.Replace(string(read), "gcr.io/christiewilson-catfactory", r, -1)
```
sbwsg, line 111: Would it be worth putting `"gcr.io/christiewilson-catfactory"` into a named constant?  I'm a bit confused why it appears here.

test/e2e-tests.sh
bobcatfish: WHOOPS

bobcatfish: This is the bit that runs our integration tests! Unfortunately they all run BEFORE the yaml tests so I had temporarily commented them out so I could check if the tests were succeeding without having to wait for these to complete 🤦‍♀️ 

In the long run hopefully we can run them in parallel! The main complication that stops us from doing that right away is that both tests require a boskos cluster

examples/examples_test.go
bobcatfish: good call!

bobcatfish:
can has lgtm?

examples/examples_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+	if input != "" {
+		cmd.Stdin = strings.NewReader(input)
+	}
+
+	var stderr, stdout bytes.Buffer
+	cmd.Stderr = &stderr
+	cmd.Stdout = &stdout
+	if err := cmd.Run(); err != nil {
+		logf("couldn't run command %s %v: %v, %s", c, args, err, stderr.String())
```
afrittoli, line 70: It looks to me like in case of error we do not get to see the stdout at all. 
I think we should print out both out and err, perhaps in two different log commands.

examples/examples_test.go
```diff
@@ -0,0 +1,261 @@
...
+	// the output as it returns
+	for i := 0; i < (timeoutSeconds / sleepBetween); i++ {
+		status, err := cmd(t.Logf, "kubectl", []string{"--namespace", n, "get", run, "--output=jsonpath={.status.conditions[*].status}"}, "")
+		if err != nil {
+			t.Fatalf("couldnt get status of %s: %v", run, err)
+			wg.Done()
+			return
+		}
+
+		switch status {
```
afrittoli, line 142: In future we might want to allow for example metadata of some kind (annotations?) to specify an expected target status and more details about it.

examples/examples_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+// eraseClusterTasks will erase all cluster tasks which do not get cleaned up by namespace.
+func eraseClusterTasks(logf formatLogger) {
+	if _, err := cmd(logf, "kubectl", []string{"delete", "--ignore-not-found=true", "clustertasks.tekton.dev", "--all"}, ""); err != nil {
+		logf("couldn't delete cluster tasks: %v", err)
+	}
+}
+
+// getYamls will look in the directory in examples indicated by version and run for yaml files
+func getYamls(t *testing.T, version, run string) []string {
```
afrittoli, line 85: I wonder if some of this helpers could use a unit test... we do run them as part of the tests anyways, so they most likely all do that they are expected to :)

afrittoli:
I'm afraid something is wrong, the YAML tests are timing out, logged as "FAILED" but still the CI job is marked as green:
```
panic: test timed out after 15m0s

(...)

goroutine 2833 [IO wait]:
internal/poll.runtime_pollWait(0x7f23b8a5d008, 0x72, 0xffffffffffffffff)
	/usr/local/go/src/runtime/netpoll.go:203 +0x55
internal/poll.(*pollDesc).wait(0xc00038a4f8, 0x72, 0x501, 0x5de, 0xffffffffffffffff)
	/usr/local/go/src/internal/poll/fd_poll_runtime.go:87 +0x45
internal/poll.(*pollDesc).waitRead(...)
	/usr/local/go/src/internal/poll/fd_poll_runtime.go:92
internal/poll.(*FD).Read(0xc00038a4e0, 0xc00027ec22, 0x5de, 0x5de, 0x0, 0x0, 0x0)
	/usr/local/go/src/internal/poll/fd_unix.go:169 +0x19b
os.(*File).read(...)
	/usr/local/go/src/os/file_unix.go:263
os.(*File).Read(0xc0005ea328, 0xc00027ec22, 0x5de, 0x5de, 0x22, 0x0, 0x0)
	/usr/local/go/src/os/file.go:116 +0x71
bytes.(*Buffer).ReadFrom(0xc000589ef0, 0x13c8460, 0xc0005ea328, 0x7f23b9c24660, 0xc000589ef0, 0xc0005f0701)
	/usr/local/go/src/bytes/buffer.go:204 +0xb1
io.copyBuffer(0x13c7120, 0xc000589ef0, 0x13c8460, 0xc0005ea328, 0x0, 0x0, 0x0, 0x406ca5, 0xc0004ae360, 0xc0005f07b0)
	/usr/local/go/src/io/io.go:391 +0x2fc
io.Copy(...)
	/usr/local/go/src/io/io.go:364
os/exec.(*Cmd).writerDescriptor.func1(0xc0004ae360, 0xc0005f07b0)
	/usr/local/go/src/os/exec/exec.go:310 +0x63
os/exec.(*Cmd).Start.func1(0xc0005e7600, 0xc00069bb00)
	/usr/local/go/src/os/exec/exec.go:436 +0x27
created by os/exec.(*Cmd).Start
	/usr/local/go/src/os/exec/exec.go:435 +0x608
FAIL	github.com/tektoncd/pipeline/examples	900.032s
FAIL'
+ ((  failed  ))
```

bobcatfish:
whoa, that's no good at all! thanks for noticing that @afrittoli 🙏

thomaschandler:
@bobcatfish I've managed to get tests passing. How would you feel about merging #2685 instead of this MR?

examples/examples_test.go
```diff
@@ -0,0 +1,261 @@
...
+	"fmt"
+	"io/ioutil"
+	"path"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
```
vdemeester, line 32: nit: extra space not needed :stuck_out_tongue_closed_eyes:

examples/examples_test.go
```diff
@@ -0,0 +1,261 @@
...
+)
+
+const (
+	timeoutSeconds = 600
+	sleepBetween   = 10
+
+	// we may want to consider either not running examples that require registry access
+	// or doing something more sophisticated to inject the right registry in when folks
+	// are executing the examples
+	horribleHardCodedRegistry = "gcr.io/christiewilson-catfactory"
```
vdemeester, line 47: nit: Why not running a local registry (in the test namespace) ? (in any case, it would be a follow-up)

examples/examples_test.go
vdemeester: I tend to agree, if the command failed, I tend to want to see `stderr` and `stdout`.
[`gotest.tools/v3/icmd`](https://pkg.go.dev/gotest.tools/v3@v3.0.2/icmd?tab=doc) would come handy there :stuck_out_tongue_closed_eyes:

examples/examples_test.go
```diff
@@ -0,0 +1,261 @@
...
+	return yamls
+}
+
+// replaceDockerRepo will look in the content f and replace the hard coded docker
+// repo with the on provided via the KO_DOCKER_REPO environment variable
+func replaceDockerRepo(t *testing.T, f string) string {
+	t.Helper()
+	r := os.Getenv("KO_DOCKER_REPO")
+	if r == "" {
+		t.Fatalf("KO_DOCKER_REPO must be set")
```
vdemeester, line 110: Should we error out or `skip` ? (like we do on the e2e go tests — mainly to not break openshift-pipelines CI :stuck_out_tongue_closed_eyes: )

examples/examples_test.go
```diff
@@ -0,0 +1,261 @@
...
+	read, err := ioutil.ReadFile(f)
+	if err != nil {
+		t.Fatalf("couldnt read contents of %s: %v", f, err)
+	}
+	return strings.Replace(string(read), horribleHardCodedRegistry, r, -1)
+}
+
+// logRun will retrieve the entire yaml of run in namespace n and log it
+func logRun(t *testing.T, n, run string) {
+	t.Helper()
```
vdemeester, line 121: `t.Helper` might no be needed here (as it is called by other func that are already calling `t.Helper`) — see https://github.com/stretchr/testify/pull/933

examples/examples_test.go
```diff
@@ -0,0 +1,261 @@
...
+// pollRun will use kubectl to query the specified run in namespace n
+// to see if it has completed. It will timeout after timeoutSeconds.
+func pollRun(t *testing.T, n, run string, wg *sync.WaitGroup) {
+	t.Helper()
+	// instead of polling it might be faster to explore using --watch and parsing
+	// the output as it returns
+	for i := 0; i < (timeoutSeconds / sleepBetween); i++ {
+		status, err := cmd(t.Logf, "kubectl", []string{"--namespace", n, "get", run, "--output=jsonpath={.status.conditions[*].status}"}, "")
+		if err != nil {
+			t.Fatalf("couldnt get status of %s: %v", run, err)
```
vdemeester, line 137: `t.Fatalf` does `t.FailNow` which calls `runtime.Goexit`, so `wg.Done()`  seems unecessary :upside_down_face:

vdemeester: This means it will quit the test on this error… If that's not what we want, we need to use `t.Errorf`.

examples/examples_test.go
```diff
@@ -0,0 +1,261 @@
...
+			// apply the yaml in its own namespace so it can run in parallel
+			// with similar tests (e.g. v1alpha1 + v1beta1) without conflicting with each other
+			// and we can easily cleanup.
+			n := createRandomNamespace(t)
+			knativetest.CleanupOnInterrupt(func() { deleteNamespace(t.Logf, n) }, t.Logf)
+			defer deleteNamespace(t.Logf, n)
+
+			t.Logf("Applying %s to namespace %s", y, n)
+			content := replaceDockerRepo(t, fmt.Sprintf("%s/%s/%s", version, run, y))
+			output, err := cmd(t.Logf, "ko", []string{"create", "--namespace", n, "-f", "-"}, content)
```
vdemeester, line 234: `s/ko/kubectl` :thinking: :interrobang:

examples/examples_test.go
```diff
@@ -0,0 +1,261 @@
...
+func eraseClusterTasks(logf formatLogger) {
+	if _, err := cmd(logf, "kubectl", []string{"delete", "--ignore-not-found=true", "clustertasks.tekton.dev", "--all"}, ""); err != nil {
+		logf("couldn't delete cluster tasks: %v", err)
+	}
+}
+
+// getYamls will look in the directory in examples indicated by version and run for yaml files
+func getYamls(t *testing.T, version, run string) []string {
+	t.Helper()
+	_, filename, _, _ := runtime.Caller(0)
```
vdemeester, line 87: Any reason to make the test dependent on the test file ? :thinking: (and use `runtime.Caller`)

bobcatfish:
> @bobcatfish I've managed to get tests passing on #2685. How would you feel about merging #2685 instead of this MR?

@thomaschandler good call, I think you're able to get to this faster than me! lemme go over to #2685 and review and we can merge your PR instead/first :D

bobcatfish:
Apologies for closing this after your careful review @vdemeester @afrittoli @ImJasonH but @thomaschandler is making much faster progress over in #2685 and he's using the client libs so it's a bit cleaner so im closing this PR in favor of it.

examples/examples_test.go
bobcatfish: haha probably!!!

examples/examples_test.go
bobcatfish: could be a good follow up! unless the test is deploying to a different namespace 🤔

examples/examples_test.go
bobcatfish: the existing scripts are using ko :O

PR #2543, opened 2020-05-05
Improvements on pipeline cancel

afrittoli:
# Changes

When a pipeline is cancelled, detect it early in the reconcile
cycle and cancel all taskruns defined in the pipelinerun status,
so that we don't spend time building the DAG and resolving all
resources.

The patch to cancel a TaskRun is the same for all TaskRuns, so
build it once and apply it to all taskruns.

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/pkg/reconciler/pipelinerun/cancel.go
+++ b/pkg/reconciler/pipelinerun/cancel.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	clientset "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
-	"github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun/resources"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -35,7 +34,7 @@ import (
 )
 
 // cancelPipelineRun marks the PipelineRun as cancelled and any resolved TaskRun(s) too.
-func cancelPipelineRun(logger *zap.SugaredLogger, pr *v1alpha1.PipelineRun, pipelineState []*resources.ResolvedPipelineRunTask, clientSet clientset.Interface) error {
+func cancelPipelineRun(logger *zap.SugaredLogger, pr *v1alpha1.PipelineRun, clientSet clientset.Interface) error {
 	pr.Status.SetCondition(&apis.Condition{
 		Type:    apis.ConditionSucceeded,
 		Status:  corev1.ConditionFalse,
@@ -45,23 +44,21 @@ func cancelPipelineRun(logger *zap.SugaredLogger, pr *v1alpha1.PipelineRun, pipe
 	// update pr completed time
 	pr.Status.CompletionTime = &metav1.Time{Time: time.Now()}
 	errs := []string{}
-	for _, rprt := range pipelineState {
-		if rprt.TaskRun == nil {
-			// No taskrun yet, pass
-			continue
-		}
 
-		logger.Infof("cancelling TaskRun %s", rprt.TaskRunName)
+	// Use Patch to update the TaskRuns since the TaskRun controller may be operating on the
+	// TaskRuns at the same time and trying to update the entire object may cause a race
+	b, err := getCancelPatch()
+	if err != nil {
+		return fmt.Errorf("couldn't make patch to update TaskRun cancellation: %v", err)
+	}
 
-		// Use Patch to update the TaskRuns since the TaskRun controller may be operating on the
-		// TaskRuns at the same time and trying to update the entire object may cause a race
-		b, err := getCancelPatch()
-		if err != nil {
-			errs = append(errs, fmt.Errorf("couldn't make patch to update TaskRun cancellation: %v", err).Error())
-			continue
-		}
-		if _, err := clientSet.TektonV1alpha1().TaskRuns(pr.Namespace).Patch(rprt.TaskRunName, types.JSONPatchType, b, ""); err != nil {
-			errs = append(errs, fmt.Errorf("Failed to patch TaskRun `%s` with cancellation: %s", rprt.TaskRunName, err).Error())
+	// Loop over the TaskRuns in the PipelineRun status.
+	// If a TaskRun is not in the status yet we should not cancel it anyways.
+	for taskRunName := range pr.Status.TaskRuns {
+		logger.Infof("cancelling TaskRun %s", taskRunName)
+
+		if _, err := clientSet.TektonV1alpha1().TaskRuns(pr.Namespace).Patch(taskRunName, types.JSONPatchType, b, ""); err != nil {
+			errs = append(errs, fmt.Errorf("Failed to patch TaskRun `%s` with cancellation: %s", taskRunName, err).Error())
 			continue
 		}
 	}
--- a/pkg/reconciler/pipelinerun/metrics.go
+++ b/pkg/reconciler/pipelinerun/metrics.go
@@ -48,6 +48,7 @@ var (
 		stats.UnitDimensionless)
 )
 
+// Recorder holds keys for Tekton metrics
 type Recorder struct {
 	initialized bool
 
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -158,9 +158,10 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 
 	// In case of reconcile errors, we store the error in a multierror, attempt
 	// to update, and return the original error combined with any update error
-	var merr error
+	var merr *multierror.Error
 
-	if pr.IsDone() {
+	switch {
+	case pr.IsDone():
 		// We may be reading a version of the object that was stored at an older version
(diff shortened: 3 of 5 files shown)


pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -45,23 +44,21 @@ func cancelPipelineRun(logger *zap.SugaredLogger, pr *v1alpha1.PipelineRun, pipe
...
-		// TaskRuns at the same time and trying to update the entire object may cause a race
-		b, err := getCancelPatch()
-		if err != nil {
-			errs = append(errs, fmt.Errorf("couldn't make patch to update TaskRun cancellation: %v", err).Error())
-			continue
-		}
-		if _, err := clientSet.TektonV1alpha1().TaskRuns(pr.Namespace).Patch(rprt.TaskRunName, types.JSONPatchType, b, ""); err != nil {
-			errs = append(errs, fmt.Errorf("Failed to patch TaskRun `%s` with cancellation: %s", rprt.TaskRunName, err).Error())
+	// Loop over the TaskRuns in the PipelineRun status.
+	// If a TaskRun is not in the status yet we should not cancel it anyways.
```
bobcatfish, line 56: nice!

bobcatfish: just double checking, can we guarantee that all started TaskRuns will be in the status? im guessing we can but wanted to check: i think this would mean that in the same reconcile loop that we create a TaskRun, we always update the status (i'm wondering also if there was an error partway through if we might end up in an inconsistent state? e.g. we create the taskrun, then a transient networking error prevents us from updating the status?)

pkg/reconciler/pipelinerun/metrics.go
```diff
@@ -48,6 +48,7 @@ var (
 		stats.UnitDimensionless)
 )
 
+// Recorder holds keys for Tekton metrics
```
bobcatfish, line 51: thanks for adding docs!

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -160,7 +160,8 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 	// to update, and return the original error combined with any update error
 	var merr error
 
-	if pr.IsDone() {
+	switch {
```
bobcatfish, line 163: nice :D

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -527,15 +534,6 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1alpha1.PipelineRun) er
 		}
 	}
 
-	// If the pipelinerun is cancelled, cancel tasks and update status
-	if pr.IsCancelled() {
-		before := pr.Status.GetCondition(apis.ConditionSucceeded)
-		err := cancelPipelineRun(c.Logger, pr, pipelineState, c.PipelineClientSet)
-		after := pr.Status.GetCondition(apis.ConditionSucceeded)
-		reconciler.EmitEvent(c.Recorder, before, after, pr)
-		return err
```
bobcatfish, line 536: in this version, the `reconcile` function would return at this point if cancelled, but looks like that is no longer the case - want to double check we want to keep executing `reconcile` in this case

pkg/reconciler/pipelinerun/cancel.go
afrittoli: Yeah, I thought about that. In my understanding the reconciler would not take the same key from the queue. When we start to reconcile we get the latest version of the pr from the API, that should have all running taskruns. It might be that a previous reconcile failed to update the status of the pipeline, but the same may be true for the taskrun status too, so I feel we are not introducing a new issue.

pkg/reconciler/pipelinerun/pipelinerun.go
afrittoli: In the new version we run even less reconcile than before. We hit this condition in the `Reconcile` and never enter `reconcile`.

pkg/reconciler/pipelinerun/cancel.go
afrittoli: The current implementation uses the `pipelineState` which requires fetching the pipeline and taskruns and tasks and pipeline resources and validating and binding all together... I was hoping to moving the catch of cancel earlier in the reconcile and thus avoid calculating the `pipelineState`.

bobcatfish: > It might be that a previous reconcile failed to update the status of the pipeline, but the same may be true for the taskrun status too, so I feel we are not introducing a new issue.

I guess in that case the next reconcile would notice that the status needed to be updated anyway maybe, then update it, then maybe the "is cancelled" block would catch any taskruns that need to be cancelled that werent before 🤔 

it _seems_ like eventually it would end up in the right state?

bobcatfish: That would only be if the reconcile actually continued tho, if we bail out as soon as we see is cancelled im not sure that it would.

I think I'm just rambling tho, I'll try to be a bit more clear and think this through:

> It might be that a previous reconcile failed to update the status of the pipeline, but the same may be true for the taskrun status too, so I feel we are not introducing a new issue.

Okay so let's say a taskrun status update fails, i think something like this would happen:

1. reconcile starts, determines status needs to be updated
2. status update fails due to network error, reconcile returns an error (i assume?), triggering re-queue of taskrun
3. reconcile starts again, hopefully updates status correctly

Let's say a pipelinerun status update fails, then it's cancelled:

1. reconcile starts, creates taskrun, determines status needs to be updated with new taskrun
2. status update fails due to network error, reconcile returns an error (i assume?), triggering re-queue of pipelinerun
3. user cancels pipelinerun
4. requeuing notices that pipelinerun is now cancelled, there are no taskruns in status, therefore not cancelled

I think it depends what happens next, for exmaple if we stop reconciling at that point, the taskrun won't ever get cancelled.

The only way around this i can think of is if we actually completed the rest of the reconcile function after this point, b/c i think the controller would notice that the status needs to be updated with the running taskrun, and update it, presumably triggering another reconcile, when the cancellation logic would get hit again and the taskrun would get cancelled

pkg/reconciler/pipelinerun/pipelinerun.go
bobcatfish: you're right! i was misreading, i thought that the reconcile still happened

pkg/reconciler/pipelinerun/cancel.go
afrittoli: Thank you, you make a very good point. 
I agree that the scenario you describe is not covered.

Looking deeper into the code, it looks like it is not covered today either 😅 
[ResolvePipelineRun](https://github.com/tektoncd/pipeline/blob/fe5981e4e75ffec7f50a4aaab5eb6899742dd0f2/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L288) is used to build the pipeline run state. That function behaves as follows:
- for each task in the spec:
- look in the pipelinerun status for a matching taskrun
- when not found, generate a new name
- try to fetch the taskruns, if error and error is `NotFound`, ignore

This means that if we fail to sync the pipelinerun status, we never go a list TaskRuns by owner, so the taskruns created during the reconcile cycle where the issue happened become orphaned. A new name is generated for the tasks so presumably `dag.GetSchedulable` will return the task again and it will end up running the same task(s) twice.

It is a rather unlikely case to happen, since it requires being able to create a taskrun and a few milliseconds later being unable to update the pipelinerun status, however it may well happen.

I think a way forward would be to ensure that the pipelinerun status is valid. 
To make sure that the status is in sync with the list of taskruns owned by the pipelinerun, we can run this check before we try and use the pipelinerun status as source of truth. To do we only need the pipelinerun status and the object ref of the pipelinerun, so we do not need to go through the resolution and validation and conversion and re-conversion madness. 

I will propose a patch that does that, keep this on hold, and once (if) the other patch is approved, it should then be fine to merge this roughly as it is today. Does it sound like a plan?

afrittoli: https://github.com/tektoncd/pipeline/issues/2558

afrittoli: @bobcatfish Since - given https://github.com/tektoncd/pipeline/issues/2558 - this patch does not change the situation compared to today - I wonder if we could go ahead with this and fix https://github.com/tektoncd/pipeline/issues/2558 then.

afrittoli: I setup a unit test to verify that we do not recover orphaned taskruns - assuming that they may happen in case the pipelinerun status sync fails.
This WIP fix shows how we could solve the issue: https://github.com/tektoncd/pipeline/issues/2558

bobcatfish: > This means that if we fail to sync the pipelinerun status, we never go a list TaskRuns by owner, so the taskruns created during the reconcile cycle where the issue happened become orphaned. A new name is generated for the tasks so presumably dag.GetSchedulable will return the task again and it will end up running the same task(s) twice.

yikes!! nice find :D

> To make sure that the status is in sync with the list of taskruns owned by the pipelinerun, we can run this check before we try and use the pipelinerun status as source of truth

kk sounds good to me!

Thanks for looking into this and opening the other issue 🙏

bobcatfish:
Thanks for the indepth back and forth :D 💃 

/lgtm

Issue #2474, opened 2020-04-23
[feature] Start using knative PermanentErrors in reconcilers


afrittoli:
# Expected Behavior
Distinguish explicitly between permanent errors and transient ones in the reconciler code.

# Actual Behavior
Today we capture errors that are considered permanent in `Reconcile` so that we can return `nil` and the key is not directly requeued. So we use the same semantic to indicate no error and permanent error.

[Knative pkg](https://github.com/knative/eventing/blob/9b0599f33985020f445820ce6ee843d5ab701122/vendor/knative.dev/pkg/controller/controller.go#L462-L466) defines `PermanentError`: when `Reconcile` is invoked on an object taken from the queue, [if it returns `PermanentError` the object is not requeued](https://github.com/knative/eventing/blob/9b0599f33985020f445820ce6ee843d5ab701122/vendor/knative.dev/pkg/controller/controller.go#L425). 
This allows being explicit in the reconcile code about which error is considered a permanent one.

/kind feature

Issue #2082, opened 2020-02-21
Optionally emit events during execution


bobcatfish:
# Expected Behavior

Although a user can use [the CloudEvent PipelineResource](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#cloud-event-resource) to emit events during a TaskRun or PipelineRun, some folks may want this to happen by default, for significant events such as:

* Starting execution
* Finishing execution
* Condition skipped
* ....?

# Actual Behavior

[The CloudEvent PipelineResource](https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#cloud-event-resource)  is the only option for this right now, but even that needs #1684 before it can be relied on (see #2053)

# Additional info

* One open question is _where_ it would be emitting events _to_, e.g. does the Run provide an endpoint address that CloudEvents are sent to via HTTP, or do the events go to some kind of message bus, or something else?
* Note that the keptn.sh folks are interested in creating standards around these kinds of events https://github.com/keptn/keptn/issues/1259

afrittoli:
@bobcatfish I've been thinking about this for a bit, and I think it would be great to have the ability to specify a sink in the `Task|PipelineRun` so that cloud events could be sent by a task without the need of altering the task to add the pipeline resource. This would greatly improve the re-usability of tasks, e.g. when using Tasks as CI Jobs.

It gets a bit more complicated than just a sink though, since we might need to:
1. send the event to multiple sinks
1. send the event to different sinks depending on the event
1. send the event to different sinks depending on the task in a pipeline

A lot of those might be avoided when using tekton triggers by using a single EventListener with multiple triggers and CEL filters to select the right thing to select.

This is highly related to the notification work too, as I think the basic infrastructure required in the controller to make this happen is the same as that needed for notifications.

afrittoli:
As a starting point, I think it's a fair assumption to let receivers / channels filter events, and err on the side of sending too many events, in favour of a lean syntax on Tekton side.
In practice this means that a `PipelineRun` and a `TaskRun` will get a new field that allows specifying a list of sinks. If sinks are specified, a cloud event will be sent for all events to all sinks.

Options for the field name: `sinks`, `cloudEvents`, `cloudEventSinks`, `notify`

```
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  name: pipeline-with-sinks
spec:
  serviceAccountName: bot
  cloudEvents:
    - http://my-event-listener
    - https://another-event-listener
[...]

PR #2545, opened 2020-05-05
Emit events from the PipelineRun controller

afrittoli:
# Changes

Emit events from the PipelineRun controller

Emit events:
- Pipeline Start
- Pipeline Started Running
- Various Error Conditions, Cancel, Timeout

Emit all events through the events.go module.
Align and simplify the reconcile structure to have clear points
for error handling and emitting events.

Added test for the normal reconcile to completion case.
Added more event checks to existing tests.

Partially addresses #2474
Partially addresses #2082

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
PipelineRuns emit an additional k8s event when starting to execute, to run and on different failure scenarios.
```

### DIFF
--- a/docs/events.md
+++ b/docs/events.md
@@ -27,13 +27,21 @@ No events are emitted for `Conditions` today (https://github.com/tektoncd/pipeli
   successfully, including post-steps injected by Tekton.
 - `Failed`: this is triggered if the `TaskRun` is completed, but not successfully.
   Causes of failure may be: one the steps failed, the `TaskRun` was cancelled or
-  the `TaskRun` timed out.
+  the `TaskRun` timed out. `Failed` events are also triggered in case the `TaskRun`
+  cannot be executed at all because of validation issues.
 
 ## PipelineRuns
 
 `PipelineRun` events are generated for the following `Reasons`:
+- `Started`: this is triggered the first time the `PipelineRun` is picked by the
+  reconciler from its work queue, so it only happens if web-hook validation was
+  successful. Note that this event does not imply that a step started executing,
+  as pipeline, task and bound resource validation must be successful first.
+- `Running`: this is triggered when the `PipelineRun` passes validation and
+  actually starts running.
 - `Succeeded`: this is triggered once all `Tasks` reachable via the DAG are
   executed successfully.
 - `Failed`: this is triggered if the `PipelineRun` is completed, but not
   successfully. Causes of failure may be: one the `Tasks` failed or the
-  `PipelineRun` was cancelled.
+  `PipelineRun` was cancelled or timed out. `Failed` events are also triggered
+  in case the `PipelineRun` cannot be executed at all because of validation issues.
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -18,7 +18,6 @@ package pipelinerun
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"path/filepath"
 	"reflect"
@@ -50,7 +49,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/configmap"
@@ -97,10 +95,6 @@ const (
 
 	// pipelineRunAgentName defines logging agent name for PipelineRun Controller
 	pipelineRunAgentName = "pipeline-controller"
-
-	// Event reasons
-	eventReasonFailed    = "PipelineRunFailed"
-	eventReasonSucceeded = "PipelineRunSucceeded"
 )
 
 type configStore interface {
@@ -141,7 +135,7 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
 		c.Logger.Errorf("invalid resource key: %s", key)
-		return nil
+		return controller.NewPermanentError(err)
 	}
 
 	ctx = c.configStore.ToContext(ctx)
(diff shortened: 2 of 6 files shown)


vdemeester:
```
history.go:52: [debug] getting history for release gohelloworld
        Release "gohelloworld" does not exist. Installing it now.
        install.go:158: [debug] Original chart version: ""
        install.go:175: [debug] CHART PATH: /workspace/gitsource/test/gohelloworld/gohelloworld-chart
        
        Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "Deployment" in version "extensions/v1beta1"
        helm.go:75: [debug] unable to recognize "": no matches for kind "Deployment" in version "extensions/v1beta1"
        unable to build kubernetes objects from release manifest
        helm.sh/helm/v3/pkg/action.(*Install).Run
        main.runInstall
        main.newUpgradeCmd.func1
        github.com/spf13/cobra.(*Command).execute
        github.com/spf13/cobra.(*Command).ExecuteC
        github.com/spf13/cobra.(*Command).Execute
        main.main
        runtime.main
        runtime.goexit
```

Something is "legit" wrong with the helm tests

docs/events.md
```diff
@@ -27,13 +27,21 @@ No events are emitted for `Conditions` today (https://github.com/tektoncd/pipeli
...
+  as pipeline, task and bound resource validation must be successful first.
+- `Running`: this is triggered when the `PipelineRun` passes validation and
+  actually starts running.
 - `Succeeded`: this is triggered once all `Tasks` reachable via the DAG are
   executed successfully.
 - `Failed`: this is triggered if the `PipelineRun` is completed, but not
   successfully. Causes of failure may be: one the `Tasks` failed or the
-  `PipelineRun` was cancelled.
+  `PipelineRun` was cancelled or timed out. `Failed` events are also triggered
+  in case the `PipelineRun` cannot be executed at all because of validation issues.
```
bobcatfish, line 47: thanks for adding docs!

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -239,6 +239,17 @@ func (pr *PipelineRunStatus) MarkResourceNotConvertible(err *CannotConvertError)
...
 }
 
+// MarkResourceFailed sets the ConditionSucceeded condition to ConditionFalse
+// based on an error that occurred and a reason
+func (pr *PipelineRunStatus) MarkResourceFailed(reason string, err error) {
+	pipelineRunCondSet.Manage(pr).SetCondition(apis.Condition{
+		Type:    apis.ConditionSucceeded,
+		Status:  corev1.ConditionFalse,
+		Reason:  reason,
+		Message: err.Error(),
```
bobcatfish, line 249: i wonder if we want to pass the error message through as is in the message field; it might be easier for the user to pass in a more specific message about what went wrong? often when we write error messages i think we have developers in mind more than the users

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -149,206 +145,142 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+		return controller.NewPermanentError(err)
 	} else if err != nil {
 		return err
 	}
 
 	// Don't modify the informer's copy.
 	pr := original.DeepCopy()
+
 	if !pr.HasStarted() {
+		// HasStarted is made false by InitializeConditions
```
bobcatfish, line 157: i'm kind of confused by this comment - especially since we just called HasStarted and it's already false - would it make sense to elaborate a bit more or remove this comment?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -149,206 +145,142 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
 		// In case node time was not synchronized, when controller has been scheduled to other nodes.
 		if pr.Status.StartTime.Sub(pr.CreationTimestamp.Time) < 0 {
 			c.Logger.Warnf("PipelineRun %s createTimestamp %s is after the pipelineRun started %s", pr.GetRunKey(), pr.CreationTimestamp, pr.Status.StartTime)
 			pr.Status.StartTime = &pr.CreationTimestamp
 		}
-		// start goroutine to track pipelinerun timeout only startTime is not set
+		// start goroutine to track pipelinerun timeout only if startTime is not set
 		go c.timeoutHandler.WaitPipelineRun(pr, pr.Status.StartTime)
-	} else {
-		pr.Status.InitializeConditions()
```
bobcatfish, line 169: why did we previously initializeconditions only if the pipelinerun had started and now we do the opposite (only if it has not started?)

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -149,206 +145,142 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+		// Emit events. During the first reconcile the status of the PipelineRun may change twice
+		// from not Started to Started and then to Running, so we need to sent the event here
+		// and at the end of 'Reconcile' again.
+		// We also want to send the "Started" event as soon as possible for anyone who may be waiting
+		// on the event to perform user facing initialisations, such has reset a CI check status
```
bobcatfish, line 170: these comments are really helpful, thanks!!

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -149,206 +145,142 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+		afterCondition := pr.Status.GetCondition(apis.ConditionSucceeded)
+		events.Emit(c.Recorder, nil, afterCondition, pr)
 	}
 
-	// In case of reconcile errors, we store the error in a multierror, attempt
-	// to update, and return the original error combined with any update error
-	var merr *multierror.Error
+	if pr.IsDone() {
+		c.Logger.Infof("taskrun done : %s \n", pr.Name)
```
bobcatfish, line 176: this log message feels a bit adhoc, is it intentional to include it? also i think the message should say "pipelinerun done"?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -149,206 +145,142 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
 		pr.SetDefaults(contexts.WithUpgradeViaDefaulting(ctx))
 
-		c.updatePipelineResults(ctx, pr)
-		if err := artifacts.CleanupArtifactStorage(pr, c.KubeClientSet, c.Logger); err != nil {
-			c.Logger.Errorf("Failed to delete PVC for PipelineRun %s: %v", pr.Name, err)
-			return err
+		// Completion time is set, so we don't need the timeout handler anymore
+		c.timeoutHandler.Release(pr)
+
+		// and sync the update to etcd
```
bobcatfish, line 186: is it possible to add a bit more info about why?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -149,206 +145,142 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+		// We also want to send the "Started" event as soon as possible for anyone who may be waiting
+		// on the event to perform user facing initialisations, such has reset a CI check status
+		afterCondition := pr.Status.GetCondition(apis.ConditionSucceeded)
+		events.Emit(c.Recorder, nil, afterCondition, pr)
 	}
 
-	// In case of reconcile errors, we store the error in a multierror, attempt
-	// to update, and return the original error combined with any update error
-	var merr *multierror.Error
+	if pr.IsDone() {
```
bobcatfish, line 175: the Reconcile function might be easier to read if the actions that happen in each case were in separate functions - i think cancelPipelineRun is a good example of this

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -149,206 +145,142 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
-			return err
 		}
-		c.timeoutHandler.Release(pr)
-		if err := c.updateTaskRunsStatusDirectly(pr); err != nil {
-			c.Logger.Errorf("Failed to update TaskRun status for PipelineRun %s: %v", pr.Name, err)
-			return err
+
+		// Update Pipeline results
+		c.updatePipelineResults(ctx, pr)
+		// Cleanup the artifact storage
```
bobcatfish, line 203: these 2 comments feel like they repeat what the line below them says - we can probably remove them without losing much? there are a few other comments like that in this PR also

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -149,206 +145,142 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
-			return err
+		// Completion time is set, so we don't need the timeout handler anymore
+		c.timeoutHandler.Release(pr)
+
+		// and sync the update to etcd
+		if err := c.updateTaskRunsStatusDirectly(pr); err != nil {
+			merr = multierror.Append(err, merr)
+			c.Logger.Errorf("Failed to update TaskRun status for PipelineRun %s: %v", pr.Name, err)
 		}
+		// Even if taskrun status updates failed, try to sync pipeline results at least
```
bobcatfish, line 191: i'm kind of confused by this comment, is this referring to the pipeline results feature or something else (it seems like something else since the function being called is about labels and annotations)? and i'm not sure why we would want to be trying to do the sync anyway, maybe a bit more info?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -149,206 +145,142 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
...
+		updateErr := c.updateStatusLabelsAndAnnotations(pr, original)
+		events.EmitError(c.Recorder, updateErr, pr)
```
bobcatfish, line 193: i think i mentioned this before, but i would expect to only call EmitError if updateErr is an error, its kinda confusing to me that this is called even when there isnt an error

a function name change might help, maybe EmitIfError?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -507,104 +406,210 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
...
+	// otherwise we return the original error, so we preserve the type (permanent or not)
+	if err != nil {
+		return multierror.Append(previousError, err).ErrorOrNil()
+	}
+	return previousError
+}
+
+func (c *Reconciler) updatePipelineResults(ctx context.Context, pr *v1beta1.PipelineRun) {
+
+	// TODO: Use factory func instead of hard-coding this once OCI images are supported.
```
bobcatfish, line 491: i know you didnt put this TODO in so feel free to ignore me but is there a github issue we can link to? otherwise TODOs tend to stick around forever

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -507,104 +406,210 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
...
+	resolver := &resources.LocalPipelineRefResolver{
+		Namespace:    pr.Namespace,
+		Tektonclient: c.PipelineClientSet,
+	}
+	pipelineMeta, pipelineSpec, err := resources.GetPipelineData(ctx, pr, resolver.GetPipeline)
+	if err != nil {
+		if ce, ok := err.(*v1beta1.CannotConvertError); ok {
```
bobcatfish, line 498: i often find that when im tempted to check for a specific type of error and handle it differently, it turns out to be cleaner to instead have 2 functions with different responsibilities: one which fails if conversion fails and one for the rest

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -507,104 +406,210 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
...
+			pr.Status.MarkResourceNotConvertible(ce)
```
bobcatfish, line 499: im confused by why this function seems to repeat a lot of what `prepare` does - is it possible for `prepare` to be called before this function instead of duplicating the logic?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -507,104 +406,210 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
...
+	// - pipelineSpec: the pipeline spec
+	//
+	// The structure of reconcile is:
+	// - Create any PVC that may be needed - when templates are used (only before the first task)
+	// - Find next Tasks from the DAG
+	// - Resolve results that may be input of Tasks to be started
+	// - Initialize (only when needed) the artifact storage
+	// - Loop through the tasks obtained from the DAG and run each of them
+	// - Update the overall PipelineRun status from pipelineState
+	// - Update the TaskRun status in PipelineRun status from the TaskRuns
```
bobcatfish, line 574: could this be a docstring? it seems like the kind of thing you'd have in a docstring

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -815,30 +968,42 @@ func TestReconcileOnCompletedPipelineRun(t *testing.T) {
...
+			case action.Matches("create", "taskruns"):
+				t.Errorf("Expected client to not have created a TaskRun, but it did")
+			case action.Matches("update", "pipelineruns"):
+				pipelineUpdates++
+			case action.Matches("get", "pipeline"):
+				// We expect one pipelinerun update and one pipeline get.
+				// We need to check the pipeline spec in case there are results to be calculated
+				fallthrough
+			default:
+				continue
```
bobcatfish, line 997: this seems pretty complicated for a unit test - should we have unit tests for this as well?

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
afrittoli: It was always pretty much the same in the original code, which is why I went this direction.
But it's ok to change it back to separated ones.

pkg/reconciler/pipelinerun/pipelinerun.go
afrittoli: Previously we ran it both when pipeline had started and when not (L160 and L169).
I remember I thought a lot about this but it was sometime ago, now I need to look back into it

pkg/reconciler/pipelinerun/pipelinerun_test.go
afrittoli: Perhaps we could have an helper to check actions and use it in all tests... and we could unit test the helper.

afrittoli: But I would do that in a different PR

afrittoli:
@bobcatfish this is just rebased on top of @mattmoor patch - I still have to address more of your comments.

pkg/reconciler/pipelinerun/pipelinerun.go
vdemeester: It's related to the OCI work, so not sure which issue to link onto

pkg/reconciler/pipelinerun/pipelinerun.go
vdemeester: yeah I should have probably written a function like `v1beta1.IsConvertError(…)` instead… I'll add that to my "small refactorings to do)

afrittoli:
> hey @afrittoli i started reviewing this and i've only gotten part way through - im wondering if it might be reasonable to break up some of the refactoring into a separate PR or 2? I know that's a huge pain to do but id like to be able to review this carefully and it's a bit tough - the description of the PR is that it's expanding the events we are emitting but the content feels like a pretty significant refactoring of the existing logic - it'd be easier to review if we could review the refactoring and understand what that is changing without new functionality.

I'll try to break this down small pieces out of this where possible.
Some of the refactoring was now done in a separate PR, even though in a slightly different and better way https://github.com/tektoncd/pipeline/pull/2749.


> 
> what do you think? if you think its unfeasible ill have to set aside a big block of time to review carefully as is
> 
> meanwhile I left some initial feedback!

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
afrittoli: This was not done otherwise in https://github.com/tektoncd/pipeline/pull/2749

pkg/reconciler/pipelinerun/pipelinerun.go
afrittoli: `InitializeConditions` is idempotent so it is ok to run it everytime, but we don't have to.
It is convenient for some unit tests as the start time is then set, but I think it's better to set the start time explicitly in the unit tests where needed.

pkg/reconciler/pipelinerun/pipelinerun.go
afrittoli: I removed it

pkg/reconciler/pipelinerun/pipelinerun.go
afrittoli: I linked https://github.com/tektoncd/pipeline/issues/1839

pkg/reconciler/pipelinerun/pipelinerun.go
afrittoli: thanks

pkg/reconciler/pipelinerun/pipelinerun_test.go
afrittoli: https://github.com/tektoncd/pipeline/issues/2772

afrittoli:
I addressed the initial comments, but I've not split this one yet.

afrittoli:
There are a few things happening:
- add a function for updating status and labels. That is going away with `ReconcileKind` so I might wait for that PR instead
- add a function to finalize the reconcile, that does the update and emit events. this is still needed, until at least we integrate with the genreconile event handling
- split reconcile in "prepare + reconcile"
- start using permanent errors
- change order of things in the `isDone` section

Roughly I can try to see if I can make a PR for each of this. @bobcatfish does it make sense?

bobcatfish:
> Roughly I can try to see if I can make a PR for each of this. @bobcatfish does it make sense?

Sounds great to me @afrittoli , apologies for the extra work.

I think in particular having these pieces separated out into 1 or 2 PRs would help a lot:
* split reconcile in "prepare + reconcile"
* start using permanent errors

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -208,8 +236,10 @@ func TestReconcile(t *testing.T) {
 	defer cancel()
 	c := testAssets.Controller
 	clients := testAssets.Clients
+	reconciler := c.Reconciler.(*Reconciler)
+	fr := reconciler.Recorder.(*record.FakeRecorder)
```
afrittoli, line 240: This should be updated the same way it was done for the taskrun reconciler tests

vdemeester:
@afrittoli what is the status here ?

afrittoli:
This PR has been broken down into pieces and merged.
The bit missing from this PR is the separation of the `reconcile` method into `prepare` + `reconcile` - however with the use the the gen reconciler and permanent errors I think there would be no benefit in that split other than making the method code a bit shorter.

PR #2562, opened 2020-05-06
Base entrypoint image on distroless

ImJasonH:
The only reason it was based on busybox was to have access to `cp`,
which it used to copy its binary to /tekton/tools for use in later
steps.

Instead of relying on `cp`, this adds a "cp mode" to the entrypoint
binary itself. When invoked with the positional args `cp <src> <dst>`,
it copies src to dst using Go's os and io packages.

/hold for discussion

# Submitter Checklist

review them:

# Release Notes

```
Entrypoint image is no longer based on busybox, now based on distroless like other images.
```

### DIFF
--- a/cmd/entrypoint/main.go
+++ b/cmd/entrypoint/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"flag"
+	"io"
 	"log"
 	"os"
 	"os/exec"
@@ -39,8 +40,40 @@ var (
 	waitPollingInterval = time.Second
 )
 
+func cp(src, dst string) error {
+	s, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer s.Close()
+
+	// Owner has permission to write and execute, and anybody has
+	// permission to execute.
+	d, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE, 0311)
+	if err != nil {
+		return err
+	}
+	defer d.Close()
+
+	_, err = io.Copy(d, s)
+	return err
+}
+
 func main() {
 	flag.Parse()
+
+	// If invoked in "cp mode" (`entrypoint cp <src> <dst>`), simply copy
+	// the src path to the dst path. This is used to place the entrypoint
+	// binary in the tools directory, without requiring the cp command to
+	// exist in the base image.
+	if len(flag.Args()) == 3 && flag.Args()[0] == "cp" {
+		src, dst := flag.Args()[1], flag.Args()[2]
+		if err := cp(src, dst); err != nil {
+			log.Fatal(err)
+		}
+		log.Println("Copied", src, "to", dst)
+		return
+	}
 
 	e := entrypoint.Entrypointer{
 		Entrypoint:      *ep,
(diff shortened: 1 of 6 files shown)


cmd/entrypoint/main.go
```diff
@@ -39,9 +40,39 @@ var (
...
 )
 
+func cp(src, dst string) error {
+	s, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer s.Close()
+
+	d, err := os.OpenFile(dst, os.O_RDWR|os.O_CREATE, 0777)
```
chmouel, line 50: is 777 what we want? isnt that 644 the default?

ImJasonH: The default for [`os.Create`](https://golang.org/pkg/os/#Create) is 0666. With that, or 0644, subsequent steps fail with `exec: \"/tekton/tools/entrypoint\": permission denied": unknown` because the other group doesn't have permission to execute.

0667 works, or as little as 0311 (owner can write, anybody can execute), but I didn't really feel strongly about limiting permissions on the executable. I can knock it down to 0311 though.

mattmoor:
If you were to just create `./cmd/tools/cp` with the needed CLI surface, then you could avoid the entrypoint juggling when using this, and I bet the image would be microscopic.

ImJasonH:
> If you were to just create `./cmd/tools/cp` with the needed CLI surface, then you could avoid the entrypoint juggling when using this, and I bet the image would be microscopic.

That would only work if I somehow got the `cp` binary into the same step as the `entrypoint` binary, to be able to copy it out to the shared volume. I guess that would be possible with (pseudocode):

```
initContainers:
# Copy `cp` to /tekton/tools
- image: ko://github.com/tektoncd/pipeline/cmd/cp
  command: ['cp', '/ko-app/cp', '/tekton/tools/cp']
  volumeMounts:
  - name: tekton-tools
    mountPath: /tekton/tools
- image: ko://github.com/tektoncd/pipeline/cmd/entrypoint
  command: ['/tekton/tools/cp', '/ko-app/entrypoint', '/tekton/tools/entrypoint']
  volumeMounts:
  - name: tekton-tools
    mountPath: /tekton/tools

containers:
- image: ubuntu
  command: ['/ko-app/entrypoint', 
```

mattmoor:
Yeah, sorry wires crossed in my head.  I thought this was to enable more direct usage of `cp`, but this is exactly what I did in warmimage for rewriting entrypoints 😇

ImJasonH:
> Yeah, sorry wires crossed in my head. I thought this was to enable more direct usage of `cp`, but this is exactly what I did in warmimage for rewriting entrypoints 😇

For future reference, warm-image's `sleeper` binary takes a `-mode` flag that tells it to either sleep ~forever, or copy the binary itself to a destination:

https://github.com/mattmoor/warm-image/blob/8f6fd7efd2ebc47745f0a679e7b327c50722c089/cmd/sleeper/main.go#L31

ImJasonH:
Closing in favor of #2606 for now.

ImJasonH:
With [`ko` gaining multi-arch support](https://github.com/google/ko/pull/38), this change is becoming a bit more relevant, since it makes every TaskRun require a `cp` binary in the base image, which makes it harder to support multi-arch support in Tekton.

I've rebased this on `master` and I'd like to see if we think this is still the right way forward.

/reopen

But still:
/hold

ImJasonH:
Can't reopen this, created https://github.com/tektoncd/pipeline/pull/3286 instead 🤷‍♂️

Issue #2523, opened 2020-05-01
Git PipelineResource fails when disable-home-env-overwrite=true and UID is non-zero


sbwsg:
# Expected Behavior

When `disable-home-env-overwrite` is "true" and UID is non-zero PipelineResources should continue to operate just as well as when UID is 0.

# Actual Behavior

PipelineResources fail copying credentials.  This happens because:

1. HOME is no longer `/tekton/home` (because feature flag).
2. PipelineResources don't have any HOME fallback.
3. So PipelineResources try to copy creds from `/tekton/creds` into `/`
    - more accurately, the entrypointer does this but it happens in PipelineResource injected steps.
4. UID is non-zero, non-root. Copying into `/` fails.  This is actually a non-fatal error but...
5. Git PipelineResource fails to lock `//.gitconfig`. This is a fatal error.  Task dies.

# Steps to Reproduce the Problem

Set `disable-home-env-overwrite` feature flag to "true". Redeploy tekton.  Apply taskrun:

```yaml
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  name: test
spec:
  resources:
    inputs:
    - name: source
      resourceSpec:
        type: git
        params:
        - name: url
          value: https://github.com/tektoncd/pipeline
  taskSpec:
    resources:
      inputs:
      - name: source
        type: git
    steps:
    - name: check-dirs
      image: ubuntu
      env:
      - name: HOME
        value: /workspace
      script: |
        # This script is never reached. TaskRun dies in PipelineResource injected steps.
        id
        echo $HOME
  podTemplate:
    securityContext:
      runAsUser: 1111
      runAsGroup: 2222
      fsGroup: 3333
```

See non-fatal error in logs.  Example: `unsuccessful cred copy: ".ssh" from "/tekton/creds" to "/": unable to create destination directory: mkdir /.ssh: permission denied`

[...]

PR #2575, opened 2020-05-07
Git PipelineResource errs when HOME != /tekton/home & UID is non-zero

sbwsg:
Fixes https://github.com/tektoncd/pipeline/issues/2523

# Changes

When the disable-home-env-overwrite flag is "true"...

Currently the git pipelineresource errors out if a securityContext
is set on a TaskRun. This happens because it will attempt to lock
$HOME/.gitconfig. If UID is non-zero and HOME is not set then it
tries to lock `//.gitconfig` which is in a root-owned directory.

This commit sets an explicit home directory of /tekton/home for the git
pipelineresource. This volume mount always exists, even if the user
has set the disable-home-env-overwrite flag to "true", and it's always
world-writeable since it's an emptyDir with no permissions set on it.

# Submitter Checklist

review them:

### DIFF
--- a/pkg/apis/pipeline/paths.go
+++ b/pkg/apis/pipeline/paths.go
@@ -21,4 +21,6 @@ const (
 	WorkspaceDir = "/workspace"
 	// DefaultResultPath is the path for task result
 	DefaultResultPath = "/tekton/results"
+	// HomeDir is the HOME directory of PipelineResources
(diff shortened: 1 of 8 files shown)


vdemeester:
ah needs a rebase :pray:

sbwsg:
Cheers, rebased!

PR #2580, opened 2020-05-07
Replace all the various (-want, +got) styles with a single format

sbwsg:
# Changes

Many, many of our tests use cmp.Diff to test whether two objects
match. Every time that func gets used the test-writer has to remember
to include "(-want, +got)" in the string they print. This is annoying
to remember the order and format and can be easily implemented in a
helper func instead.

The first commit in this PR introduces a helper func to print diff messages in a
consistent way. This means test writers dont have to manually type
this formatting out every single time. The func is called `diff.PrintWantGot()`.

The second commit in this PR replaces every single place where cmp.Diff output
is printed with a call to `diff.PrintWantGot()` so that diff formatting becomes consistent
everywhere.

# Submitter Checklist

review them:


bobcatfish:
AMAZING

test/diff/print.go
```diff
@@ -0,0 +1,25 @@
...
+limitations under the License.
+*/
+
+package diff
+
+import "fmt"
+
+// Print takes a diff string generated by cmp.Diff and returns it
+// in a consistent format for reuse across all of our tests.
+func Print(diff string) string {
```
bobcatfish, line 23: one detail that might be important to mention the docstring or maybe even in the function name is that this expects that cmp.Diff has been called with the arguments in a specific order (want vs. got)

bobcatfish: another option is to wrap cmp.diff entirely and also display the error - if you use https://golang.org/pkg/testing/#B.Helper you'll get very similar functionality to what we have now, e.g. something in the test like

```
diff.FailIfWantNotGot(tc.expectedTaskSpec, *ts, "Didn't get expected TaskSpec modifications")
```

And that function would actually call Errorf

This is straying dangerously into non-go-like territory tho, your way is clearer! I think it's clarify vs ease

pritidesai:
wow, this is a huge cleanup, its always confusing to remind myself which one means want `+` or `-` 🤔. This will help a lot, thanks @sbwsg 👍 

Build is failing on imports order 😢 :

```
-	"github.com/tektoncd/pipeline/test/names"
 	"github.com/tektoncd/pipeline/test/diff"
+	"github.com/tektoncd/pipeline/test/names"
```

test/diff/print.go
sbwsg: I did consider wrapping cmp.Diff and I do think that is a very good idea.  Two considerations that lead me away from doing that here:

1. Some tests use `t.Fatalf` while others use `t.Errorf`.  So I figured a diff wrapping func would probably need to account for that too.
2. With an 80 file diff I wanted to do the very minimum to make all the various call sites uniform and limit the amount of review required.

I figure we can write a wrapping func for cmp.Diff in a future PR and then progressively replace cmp.Diff usage, perhaps over the course of multiple PRs?  Excellent point about B.Helper, I did not know about it!

sbwsg: I really like the idea of a func with a name that points out the ordering ala `...WantGot(want, got)`; great idea!

sbwsg: Create https://github.com/tektoncd/pipeline/issues/2587 to follow up.

sbwsg: OK, I've updated the PR to use `diff.PrintWantGot` as the name and I've updated the docstring to make explicit the expectation of ordering.

sbwsg:
Thanks for reviewing @pritidesai !

> Build is failing on imports order 😢

Thank goodness for the Makefile @vdemeester introduced, I ran `make goimports` and it fixed them all (I hope!)

ImJasonH:
Can we have a presubmit check to ensure I don't add a new `cmp.Diff` call that bypasses the helper? Otherwise I'm going to make sure we're going to need one of these PRs every few quarters 😅.

vdemeester:
> Can we have a presubmit check to ensure I don't add a new `cmp.Diff` call that bypasses the helper? Otherwise I'm going to make sure we're going to need one of these PRs every few quarters sweat_smile.

Let's do that in a follow-up :wink: https://github.com/tektoncd/pipeline/issues/2594

Issue #2617, opened 2020-05-14
Can not get task output in pipeline


gsmlg:
# Expected Behavior
Get task results

# Actual Behavior
unable to find result referenced by param "lint" in "report": Could not find result with name output for task run lint

# Steps to Reproduce the Problem

1. Run pipeline
task yaml
```yaml
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: yarn-run
  namespace: cicd
spec:
  params:
    - name: cmd
      type: string
      default: lint
  results:
    - name: output
      description: task run output
    - name: result
      description: task run result
  steps:
    - name: run
      workingDir: $(workspaces.repo.path)
      image: node:12-alpine
      script: |
        #!/bin/sh
        REPO=$(workspaces.repo.path)
        yarn run $(params.cmd) | tee $(results.output.path)
        echo $? | tee $(results.result.path)
      volumeMounts:
        - name: yarnrc
          mountPath: /tekton/home/.yarnrc
          subPath: config
        - name: yarn-cache
          mountPath: /yarn-cache
  volumes:
    - name: yarnrc
      configMap:
        name: yarnrc
    - name: yarn-cache
      persistentVolumeClaim:
        claimName: yarn-cache
  workspaces:
    - name: repo
```
2. Can not get results of lint output when output is very long
pipeline describe:
```
Name:              web-push-pipeline-run-8h7kg
Namespace:         cicd
Pipeline Ref:      zddi-web-ci-pipeline
Service Account:   pipeline-run-service
Timeout:           1h0m0s
Labels:
 tekton.dev/pipeline=zddi-web-ci-pipeline
 triggers.tekton.dev/eventlistener=gitlab-listener-interceptor
 triggers.tekton.dev/trigger=push-pipeline
 triggers.tekton.dev/triggers-eventid=xzlb8

🌡️  Status

STARTED          DURATION     STATUS
57 minutes ago   23 minutes   Failed(PipelineValidationFailed)

💌 Message

unable to find result referenced by param "lint" in "report": Could not find result with name output for task run lint

📦 Resources

 No resources

⚓ Params

 NAME               VALUE
 ∙ project_name     -------
 ∙ repository_url   ------
 ∙ revision         93cfe7b3657ebbaf7d147afe5eeba0cc4b636f2f
 ∙ ref              -------
 ∙ user_name       ------
 ∙ user_username   ------
 ∙ user_email    ------

🗂  Taskruns

[...]

PR #2626, opened 2020-05-15
Invalid length of container termination message

vincent-pli:
Fix issue: #2617

`Tekton` use container termination message of kubenetes to pass additional information to the controller(by Tekton Result).

Kubenates use `MaxContainerTerminationMessageLength` which is 4096 to limit the read size from disk, the redundant content will be discard. See the below implements in `Kubenetes`:

https://github.com/kubernetes/kubernetes/blob/96e13de777a9eb57f87889072b68ac40467209ac/pkg/kubelet/kuberuntime/kuberuntime_container.go#L420

# Changes

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
Tasks will now fail if they emit termination messages larger than Kubernetes' termination message size. At time of writing this is 4096 bytes. Since Task Results are returned JSON-encoded through a container's termination message please take care that they do not bloat the message over 4096 bytes.

### DIFF
--- a/pkg/termination/write.go
+++ b/pkg/termination/write.go
@@ -23,6 +23,12 @@ import (
 	v1alpha1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 )
 
+const (
+	// MaxContainerTerminationMessageLength is the upper bound any one container may write to
+	// its termination message path. Contents above this length will cause a failure.
+	MaxContainerTerminationMessageLength = 1024 * 4
+)
+
 // WriteMessage writes the results to the termination message path.
 func WriteMessage(path string, pro []v1alpha1.PipelineResourceResult) error {
 	// if the file at path exists, concatenate the new values otherwise create it
@@ -41,6 +47,11 @@ func WriteMessage(path string, pro []v1alpha1.PipelineResourceResult) error {
 	if err != nil {
 		return err
 	}
+
+	if len(jsonOutput) > MaxContainerTerminationMessageLength {
+		return aboveMax
+	}
+
 	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0666)
 	if err != nil {
 		return err
@@ -54,4 +65,15 @@ func WriteMessage(path string, pro []v1alpha1.PipelineResourceResult) error {
 		return err
 	}
 	return nil
+}
+
+//MessageLengthError indicate the length of termination message of container is beyond 4096 which is the max length read by kubenates
+type MessageLengthError string
+
+const (
+	aboveMax MessageLengthError = "Termination message is above max allowed size 4096, caused by large task result."
+)
+
+func (e MessageLengthError) Error() string {
+	return string(e)
 }
--- a/pkg/termination/write_test.go
+++ b/pkg/termination/write_test.go
@@ -16,9 +16,11 @@ limitations under the License.
 package termination
 
 import (
+	"errors"
 	"io/ioutil"
 	"log"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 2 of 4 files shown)


pkg/termination/write.go
```diff
@@ -55,3 +67,10 @@ func WriteMessage(path string, pro []v1alpha1.PipelineResourceResult) error {
 	}
 	return nil
 }
+
+//MessageLengthError indicate the length of termination message of container is beyond 4096 which is the max length read by kubenates
+type MessageLengthError string
+
+func (e MessageLengthError) Error() string {
+	return string(e)
```
ImJasonH, line 75: nit: The contents of this string will always be the same, so its message can also be a const. I'm pretty sure the compiler will notice this so it's not so much for performance (which is negligible anyway) but more for readability.

This would let line 53 become `return MessageLengthError{}` and hide the specifics of the string down here.

pkg/termination/write.go
```diff
@@ -41,6 +48,11 @@ func WriteMessage(path string, pro []v1alpha1.PipelineResourceResult) error {
 	if err != nil {
 		return err
 	}
+
+	if len(jsonOutput) > MaxContainerTerminationMessageLength {
+		return MessageLengthError(fmt.Sprintf("Size of container termination message is above max size allowed: %d, cause by large task result.", MaxContainerTerminationMessageLength))
```
ImJasonH, line 53: How about: `Termination message is above max allowed size (%d), caused by large task result.`

pkg/termination/write_test.go
```diff
@@ -66,3 +66,29 @@ func TestExistingFile(t *testing.T) {
 		}
 	}
 }
+
+func TestMaxSizeFile(t *testing.T) {
+	value := ""
+	for n := 0; n < 4096; n++ {
```
sbwsg, line 72: optionally `value := strings.Repeat("a", 4096)` could be used in place of the loop here.

pkg/termination/write_test.go
```diff
@@ -66,3 +66,29 @@ func TestExistingFile(t *testing.T) {
...
+	if err != nil {
+		log.Fatal("Cannot create temporary file", err)
+	}
+	// Remember to clean up the file afterwards
+	defer os.Remove(tmpFile.Name())
+
+	logger, _ := logging.NewLogger("", "test termination")
+	defer func() {
+		_ = logger.Sync()
+	}()
```
sbwsg, line 85: I'm not sure this is required - what happens if it's removed?

pkg/termination/write_test.go
```diff
@@ -66,3 +66,29 @@ func TestExistingFile(t *testing.T) {
...
+	output := []v1alpha1.PipelineResourceResult{{
+		Key:   "key1",
+		Value: value,
+	}}
+
+	if err := WriteMessage(tmpFile.Name(), output); err == nil {
```
sbwsg, line 91: Suggest checking for specific error kind here so that other errors we introduce by mistake are also caught:

```go
if err := WriteMessage(tmpFile.Name(), output); !errors.Is(err, MessageLengthError) {
  t.Fatalf("Expected MessageLengthError, receved: %v", err)
}
```

sbwsg:
We should probably also document this change and mention it in release notes.

Could you update the Release Notes in the PR to something like this:

> Tasks will now fail if they emit termination messages larger than Kubernetes' termination message size. At time of writing this is 4096 bytes. Since Task Results are returned JSON-encoded through a container's termination message please take care that they do not bloat the message over 4096 bytes.

And update [the Results section in tasks.md](https://github.com/tektoncd/pipeline/blob/master/docs/tasks.md#storing-execution-results) to describe the limit that's imposed and the error that will be returned?

afrittoli:
@vincent-pli thanks for this. Could you either remove or feel-in the release notes section in the PR (as appropriate)? Thanks!

pkg/termination/write.go
vincent-pli: Thanks @ImJasonH , I enhance the implements, please let me know if it satisfy you.

pkg/termination/write.go
vincent-pli: Done

pkg/termination/write_test.go
vincent-pli: Great, thansk @sbwsg

pkg/termination/write_test.go
vincent-pli: Removed.

pkg/termination/write_test.go
vincent-pli: Done, thanks

vincent-pli:
@sbwsg @afrittoli 
Added the release notes and enhanced the documents.

Issue #2524, opened 2020-05-01
Steps can't use creds-init credentials when their UID is different from creds-init UID and disable-home-env-overwrite=true


sbwsg:
# Expected Behavior

When a Step has its own non-root `UID` which is different from `creds-init` any credentials should still be usable by that Step.

# Actual Behavior

The Step will fail to utilize credentials.

When `disable-home-env-overwrite` is "true" the Step's `entrypointer` tries to copy credentials out of `/tekton/creds` into `$HOME`.  But the UID of the Step is different from `creds-init`.  The files inside of `/tekton/creds` are owned by whatever user `creds-init` ran as.  The Step is disallowed from reading those files as it tries to copy them out.

# Steps to Reproduce the Problem

Set `disable-home-env-overwrite` feature flag to "true". Redeploy tekton. Apply taskrun:

```yaml
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  name: test
spec:
  taskSpec:
    steps:
    - name: check-dirs
      image: ubuntu
      env:
      - name: HOME
        value: /myhome
      securityContext:
        runAsUser: 1111
        runAsGroup: 2222
      script: |
        id
        echo $HOME
        ls -lahR /tekton/creds
```

See errors in step logs.  Example: `unsuccessful cred copy: ".ssh" from "/tekton/creds" to "/workspace": unable to open source: open /tekton/creds/.ssh/known_hosts: permission denied`

# Additional Info

Originally discussed here: https://github.com/tektoncd/pipeline/issues/2165

sbwsg:
Here's a slightly more in-depth illustration of the problem:

```
[check-dirs] 2020/05/08 16:33:24 unsuccessful cred copy: ".docker" from "/tekton/creds" to "/myhome": unable to create destination directory: mkdir /myhome: permission denied
[check-dirs] 2020/05/08 16:33:24 unsuccessful cred copy: ".gitconfig" from "/tekton/creds" to "/myhome": unable to open source: open /tekton/creds/.gitconfig: permission denied
[check-dirs] 2020/05/08 16:33:24 unsuccessful cred copy: ".git-credentials" from "/tekton/creds" to "/myhome": unable to open source: open /tekton/creds/.git-credentials: permission denied
[check-dirs] 2020/05/08 16:33:24 unsuccessful cred copy: ".ssh" from "/tekton/creds" to "/myhome": unable to create destination directory: mkdir /myhome: permission denied
[check-dirs] + id
[...]

Issue #1271, opened 2019-09-04
[Bug Report] Cannot use ssh private key in task steps


chengjingtao:
# Expected Behavior
We could use ssh private key in task steps in order to execute command like `git pull` or `git push`

# Actual Behavior
When using command like `git pull` in task steps , it shows errors

```
[config] Setting up libcurl3-gnutls:amd64 (7.58.0-2ubuntu3.7) ...
[config] Setting up git (1:2.17.1-1ubuntu0.4) ...
[config] Processing triggers for libc-bin (2.27-3ubuntu1) ...
[config] Processing triggers for ca-certificates (20180409) ...
[config] Updating certificates in /etc/ssl/certs...
[config] 0 added, 0 removed; done.
[config] Running hooks in /etc/ca-certificates/update.d...
[config] done.
[config] Host key verification failed.
[config] fatal: Could not read from remote repository.
[config]
[config] Please make sure you have the correct access rights
[config] and the repository exists.

container step-config has failed  : Error
```

# Steps to Reproduce the Problem
Flow example https://github.com/tektoncd/pipeline/blob/master/examples/taskruns/taskrun-git-ssh.yaml and just change the `args` of  `config step`

1. 
taskrun-git-ssh.yaml
```
---
apiVersion: v1
kind: Secret
metadata:
  name: test-git-ssh
  annotations:
    tekton.dev/git-0: github.com
type: kubernetes.io/ssh-auth
data:
  # Generated by:
  # cat id_rsa | base64 -w 0
  # This deploy key has read-only permissions on github.com/knative/build
[...]

PR #2671, opened 2020-05-21
Remove creds-init initContainer

sbwsg:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/2524 and puts us in a better position to flip the `disable-home-env-overwrite` feature flag with less fallout for users.

Fixes https://github.com/tektoncd/pipeline/issues/1271 when the feature-flag is flipped.  Also we have documentation on how to use SSH creds without the feature flag flip [in auth.md](https://github.com/tektoncd/pipeline/blob/master/docs/auth.md#using-ssh-authentication-in-your-own-git-tasks) now too.

**TLDR; The behaviour that used to be implemented by creds-init is now performed by
the entrypointer.**

Creds-init is an initContainer that writes credential files (like .ssh) to
/tekton/home. One side-effect of relying on an initContainer to write these
files is that they receive the UID / Group ID of the initContainer's process.
This side-effect breaks any Step in the Task that relies on these credentials
but runs with a different UID. An example of where this can happen is on
OpenShift, where the UID of the user is randomized for each container in
order to limit the fallout of malicious process breaking out.

This commit removes the creds-init initContainer from all TaskRun Pods. I
haven't removed the creds-init binary from our build process in this
changeset. Doing so generates a lot of extra line noise which distracts
from the core modification being presented here.

Further fallout from this change:

- Credential volume mounts from annotated Secrets used to be mounted only once,
in the `creds-init` initContainer, at `/tekcon/creds-secrets/...`. These volume mounts
are now given to every Step container so that they can be copied out by the entrypointer.
- Credentials used to be copied from `/tekton/creds-secrets/...` to `/tekton/home`
or `/tekton/creds`, depending on whether the `disable-home-env-overwrite` feature flag
was set. Now those credentials are always copied into `/tekton/creds`.
- The `$(credentials.path)` variable is now always `/tekton/creds` regardless of whether
the `disable-home-env-overwrite` feature flag is set or not.
- When the  `disable-home-env-overwrite` feature flag is set, credentials will be placed at
the actual home directory of the current user in the Step container, instead of being placed
in `/tekton/home`. If the container or Step explicitly sets a `$HOME` then credentials will
end up in there.
- Every single Step gets its own empty `/tekton/creds` volume. This allows `/tekton/creds` to be world-writable and empty for every Step to copy in to.  The directory needs to be world-writable so it's accessible even if the user's UID wouldn't normally have permissions to create a directory under `/tekton`. This is also why there are so many lines changed in the various `pkg/pod` tests.

I've put a hold on this while I add better testing and examples that exercise the cases that this PR deals with.

~~Need to figure out a safe way to introduce a credential that can be used for testing. Possibly a `read-only` Github Deploy Secret for a private test Github repo in the tektoncd org?~~

Edit: It took me way too long but I finally figured out how to spin up a git/ssh sidecar that works as an authenticated remote repo for testing creds-init.

- [x] Example that shows credentials being used in conjunction with different `securityContext`s per Step.
- [x] Example that shows using these credentials with vanilla `git` commands, not git-init or Git PipelineResource.
- [x] Documentation that describes the behaviour of credentials in the face of different `securityContext`s.

I think there also needs to be a bit of discussion about some of the more questionable design decisions.  Maybe next API WG.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Tekton's creds-init initContainer has been removed and will no longer show up in TaskRun Pods. Credentials from annotated Secrets will continue to be mounted and placed in Step containers' home directory.
```

/kind bug
/hold

### DIFF
--- a/cmd/entrypoint/main.go
+++ b/cmd/entrypoint/main.go
@@ -26,6 +26,8 @@ import (
 	"time"
 
 	"github.com/tektoncd/pipeline/pkg/credentials"
+	"github.com/tektoncd/pipeline/pkg/credentials/dockercreds"
+	"github.com/tektoncd/pipeline/pkg/credentials/gitcreds"
 	"github.com/tektoncd/pipeline/pkg/entrypoint"
 	"github.com/tektoncd/pipeline/pkg/termination"
 )
@@ -41,7 +43,21 @@ var (
 )
 
 func main() {
+	// Add credential flags originally used in creds-init.
+	gitcreds.AddFlags(flag.CommandLine)
+	dockercreds.AddFlags(flag.CommandLine)
+
 	flag.Parse()
+
+	// Copy creds-init credentials from secret volume mounts to /tekton/creds
+	// This is done to support the expansion of a variable, $(credentials.path), that
+	// resolves to a single place with all the stored credentials.
+	builders := []credentials.Builder{dockercreds.NewBuilder(), gitcreds.NewBuilder()}
+	for _, c := range builders {
+		if err := c.Write("/tekton/creds"); err != nil {
+			log.Printf("Error initializing credentials: %s", err)
+		}
+	}
 
 	e := entrypoint.Entrypointer{
 		Entrypoint:      *ep,
@@ -56,7 +72,7 @@ func main() {
 		Results:         strings.Split(*results, ","),
 	}
 
-	// Copy any creds injected by creds-init into the $HOME directory of the current
+	// Copy any creds injected by the controller into the $HOME directory of the current
 	// user so that they're discoverable by git / ssh.
 	if err := credentials.CopyCredsToHome(credentials.CredsInitCredentials); err != nil {
 		log.Printf("non-fatal error copying credentials: %q", err)
--- a/examples/v1beta1/taskruns/authenticating-git-commands.yaml
+++ b/examples/v1beta1/taskruns/authenticating-git-commands.yaml
@@ -1,0 +1,204 @@
+# This example demonstrates usage of creds-init credentials to issue
+# git commands without a Git PipelineResource or git-clone catalog task.
+#
+# In order to exercise creds-init a sidecar is used to run a
+# git server fronted by SSH. The sidecar does the following things:
+# - Generates a host key pair, providing the public key to Steps for their known_hosts file
+# - Accepts a public key generated from creds-init credentials and uses that for an authorized_keys file
+# - Creates a bare git repo for the test git commands to run against
+# - Starts sshd and tails its log, waiting for the git commands to come in over SSH
+#
+# Two separate Steps then perform authenticated git actions against the sidecar
+# git server using the credentials mounted by creds-init:
+
+# The first step makes a git clone of the bare repository and populates it
+# with a file.
+#
+# The second step makes a git clone of the populated repository and checks
+# the contents of the repo match expectations. This step runs as a non-root
+# user in order to exercise creds-init credentials when a securityContext
+# is set.
+#
+# Notice that in each Step there is different code for handling creds-init
+# credentials when the disable-home-env-overwrite flag is "false" and when
+# it's "true".
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/ssh-auth
+metadata:
+  name: ssh-key-for-git
+  annotations:
+    tekton.dev/git-0: localhost
+data:
+  # This key was generated for this test and isn't used for anything else.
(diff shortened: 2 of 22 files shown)


sbwsg:
I've put a hold on this while I add better testing and examples that exercise the cases that this PR deals with.

vdemeester:
I am +:100: for this change :heart_eyes_cat:

chmouel:
I wonder if we can use a [gitolite](https://hub.docker.com/r/jgiannuzzi/gitolite/) deployment spinned just before running the tests with dynamically generated ssh keys for testing, I mean instead of having to deal with private external github repositories?

sbwsg:
> I wonder if we can use a gitolite deployment spinned just before running the tests with dynamically generated ssh keys for testing, I mean instead of having to deal with private external github repositories?

I've managed to spin up an example which doesn't need the private external repo - see `examples/v1beta1/taskruns/authenticating-git-commands.yaml ` in the diff here.  I'll remove that comment from the description.

sbwsg:
/hold cancel

OK, I think this PR is ready for review.  It may be worth taking a look at the `example` YAML I've added to get a sense of how creds-init changes in the face of different securityContexts & feature flags.

I've also added some short docs to try and help users out navigating the quagmire. It's definitely not easy to wrap one's head around unfortunately.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -442,7 +442,7 @@ func (c *Reconciler) updateStatusLabelsAndAnnotations(tr, original *v1beta1.Task
 		go func(metrics *Recorder) {
 			err := metrics.RunningTaskRuns(c.taskRunLister)
 			if err != nil {
-				c.Logger.Warnf("Failed to log the metrics : %v", err)
+				// c.Logger.Warnf("Failed to log the metrics : %v", err)
```
vdemeester, line 445: I "guess" the log should still be there ? :thinking:

pkg/reconciler/taskrun/controller.go
```diff
@@ -58,7 +58,7 @@ func NewController(namespace string, images pipeline.Images) func(context.Contex
 		timeoutHandler := reconciler.NewTimeoutHandler(ctx.Done(), logger)
 		metrics, err := NewRecorder()
 		if err != nil {
-			logger.Errorf("Failed to create taskrun metrics recorder %v", err)
+			// logger.Errorf("Failed to create taskrun metrics recorder %v", err)
```
vdemeester, line 61: Same here ?

sbwsg:
> We will be able to remove creds-init image in a follow-up right ?

Yes, 💯, it just made the diff really extra noisy.  I plan to move straight on to that after this.

pkg/reconciler/taskrun/taskrun.go
sbwsg: :D oops, that was me trying to reduce noise while debugging unit tests.

sbwsg:
Have addressed comments and improved the coverage so it's a positive rather than negative change.

cmd/entrypoint/main.go
```diff
@@ -41,8 +43,20 @@ var (
...
+	// Add credential flags originally used in creds-init.
+	gitcreds.AddFlags(flag.CommandLine)
+	dockercreds.AddFlags(flag.CommandLine)
+
 	flag.Parse()
 
+	// Copy creds-init credentials from secret volume mounts to /tekton/creds
+	builders := []credentials.Builder{dockercreds.NewBuilder(), gitcreds.NewBuilder()}
+	for _, c := range builders {
+		if err := c.Write("/tekton/creds"); err != nil {
```
GregDritschler, line 55: Why do the credentials need to be written first to /tekton/creds and then copied to $HOME?  Is there a reason they can't be written directly to $HOME?

vdemeester: @GregDritschler kinda yes. `/tektoncd/creds` is the "safe" path, aka it will always be there, it is under a reserved folder, … `HOME` might not be defined in the image. Depending on the runtime, the image, … it might point to different things, it might not have the correct rights, … 

/cc @sbwsg

sbwsg: Totally right @vdemeester.  We have a variable expansion, `$(credentials.path)` that has previously expanded to either `/tekton/home` (when feature flag disabled - consistent $HOME) or `/tekton/creds` (when feature flag enabled - unknowable $HOME).  After this change it always resolves to `/tekton/creds` to ensure anyone using that variable still gets a valid value that points to the credentials.

This makes me wonder, though, whether we could mount all the credentials using a single projected volume into `/tekton/creds` without the extra copy steps.  I will investigate that in a follow-up after this PR.

sbwsg: And I'm going to add a comment explaining this to this file.

bobcatfish:
Looks like the coverage went down, is that something we want to address?

sbwsg:
> Looks like the coverage went down, is that something we want to address?

@bobcatfish You might be looking at an older report?  The latest numbers look positive to me:

Screen Shot 2020-06-23 at 10 59 07 AM

bobcatfish:
> @bobcatfish You might be looking at an older report? The latest numbers look positive to me:

Whoops, sorry about that @sbwsg , you're totally right!

ImJasonH:
> Edit: It took me way too long but I finally figured out how to spin up a git/ssh sidecar that works as an authenticated remote repo for testing creds-init.

:100: I love it.

/lgtm

vdemeester:
I just rebase it against master and fix the conflict on the `docs/variables.md` :upside_down_face: 
/lgtm

examples/v1beta1/taskruns/authenticating-git-commands.yaml
```diff
@@ -0,0 +1,204 @@
...
+      # This Step is only necessary as part of the test, it's not something you'll
+      # ever need in a real-world scenario involving an external git repo.
+      image: alpine/git:v2.24.3
+      securityContext:
+        runAsUser: 0
+      volumeMounts:
+      - name: messages
+        mountPath: /messages
+      script: |
+        #!/usr/bin/env ash
```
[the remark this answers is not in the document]
afrittoli: oh, never mind, I didn't know about `ash`

Issue #1680, opened 2019-12-04
(*PipelineRun).IsDone() is incorrect


gpaul:
The `(*PipelineRun).IsDone()` function returns true when the Success condition is no longer `Unknown`. However, as soon as the first task in the PipelineRun fails it sets the PipelineRun's Success condition to `False`, and `IsDone()` immediately returns `true`, despite several other tasks still `Running`.

Instead, `IsDone()` should check `Status.CompletionTime` or iterate through all tasks and check whether all tasks have finished executing.

Two new `HasSucceeded()` and `HasFailed()` functions can take the place of `IsDone()` if needed.

I'm happy to submit a PR if this sounds sensible.

vdemeester:
@gpaul thanks for the issue, I think this is by design "initially" — aka, we mark the PipelineRun as failed as soon as one TaskRun failed, even if other are running, as we know the PipelineRun is gonna fail anyway. 

That said, it might not be the best behaviour, and even more in relation to #1394, #1376 and #1023 :angel: 

/cc @bobcatfish @ImJasonH 
/kind question
/area api

gpaul:
> @gpaul thanks for the issue, I think this is by design "initially" — aka, we mark the PipelineRun as failed as soon as one TaskRun failed, even if other are running, as we know the PipelineRun is gonna fail anyway.

Yes, that makes total sense. The issue is really that we treat this state as `IsDone() == true`, which is not the case (we are merely saying that the pipelinerun has failed, not that it has completed.

vdemeester:
> Yes, that makes total sense. The issue is really that we treat this state as `IsDone() == true`, which is not the case (we are merely saying that the pipelinerun has failed, not that it has completed.)

That's fair :stuck_out_tongue_closed_eyes:.

dibyom:
> Two new HasSucceeded() and HasFailed() functions can take the place of IsDone() if needed.

Sounds reasonable!

> I'm happy to submit a PR if this sounds sensible.

That'd be awesome! Thanks :pray:

gpaul:
I've had a look and it's actually rather tricky.

Two strategies present themselves:

1. The first strategy is to use the presence of `PipelineRun.CompletionTime` to determine whether or not a PipelineRun is done:
```
func (pr *PipelineRun) IsDone() bool {
    return pr.Status.CompletionTime != nil && !pr.Status.CompletionTime.IsZero()
}
```
[...]

PR #2677, opened 2020-05-22
Add support for more cloud events

afrittoli:
# Changes

This PR is split into two commits, to help with the review. 

Add support for more cloud events

Change the event type string for taskruns from dev.tekton.event.task.*
to dev.tekton.event.taskrun.*. Since cloud events are only sent by
pipeline resources - which are alpha - we don't need to comply with
the beta deprecation policy for this.

Extend the cloudevents module to include more event types for TaskRuns
and event types for PipelineRuns, with unit test coverage.
This is achieved by introducing the RunsToCompletion and
RunsToCompletionStatus interface which are met by TaskRun, PipelineRun
and their respective status types.

At this stage there is no event producer that generates these new
events. This is preparing the stage for the next changes where
we will start to emit cloud events in addition to the k8s events
we emit today.

No extra docs yet since the new events are not produced by anything yet.

Partially addresses #2082
Partially addresses #2684

The second:

Make phase condition reasons part of the API

TaskRuns and PipelineRuns use the "Reason" field to complement the
value of the "Succeeded" condition. Those values are not part of
the API and are even owned by the underlying resource (pod) in
case of TaskRuns. This makes it difficult to rely on them to
understand that the state of the resource is.

In case of corev1.ConditionTrue, the reason can be used to
distinguish between:
- Successful
- Successful, some parts were skipped (pipelinerun only)

In case of corev1.ConditionFalse, the reason can be used to
distinguish between:
- Failed
- Failed because of timeout
- Failed because of cancelled by the user

In case of corev1.ConditionUnknown, the reason can be used to
distinguish between:
- Just started reconciling
- Validation done, running (or still running)
- Cancellation requested

This is implemented through the following changes:
- Bubble up reasons for taskrun and pipelinerun to the
  v1beta1 API, except for reason that are defined by the
  underlying resource
- Enforce the start reason to be set during condition init

This allows for an additional change in the eventing module: the
condition before and after can be used to decide whether to send
an event at all. If they are different, the after condition now
contains enough information to send the event.

The cloudevent module is extended with ability to send the correct
event based on both status and reason.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
The cloudevent message type strings change from "dev.tekton.event.task.*" to "dev.tekton.event.taskrun.*" for events produced by the `CloudEventPipelineResource`.
```

### DIFF
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -18,6 +18,7 @@ weight: 4
   - [Specifying `Workspaces`](#specifying-workspaces)
   - [Specifying `LimitRange` values](#specifying-limitrange-values)
   - [Configuring a failure timeout](#configuring-a-failure-timeout)
+- [Monitoring execution status](#monitoring-execution-status)
 - [Cancelling a `PipelineRun`](#cancelling-a-pipelinerun)
 - [Events](events.md#pipelineruns)
 
@@ -361,6 +362,71 @@ The `timeout` value is a `duration` conforming to Go's
 [`ParseDuration`](https://golang.org/pkg/time/#ParseDuration) format. For example, valid
 values are `1h30m`, `1h`, `1m`, and `60s`. If you set the global timeout to 0, all `PipelineRuns`
 that do not have an idividual timeout set will fail immediately upon encountering an error.
+
+## Monitoring execution status
+
+As your `PipelineRun` executes, its `status` field accumulates information on the execution of each `TaskRun`
+as well as the `PipelineRun` as a whole. This information includes the name of the pipeline `Task` associated
+to a `TaskRun`, the complete [status of the `TaskrRun`](taskruns.md#monitoring-execution-status) and details
+about `Conditions` that may be associated to a `TaskRun`.
+
+The following example shows an extract from the `status` field of a `PipelineRun` that has executed successfully:
+
+```yaml
+completionTime: "2020-05-04T02:19:14Z"
+conditions:
+- lastTransitionTime: "2020-05-04T02:19:14Z"
+  message: 'Tasks Completed: 4, Skipped: 0'
+  reason: Succeeded
+  status: "True"
+  type: Succeeded
+startTime: "2020-05-04T02:00:11Z"
+taskRuns:
+  triggers-release-nightly-frwmw-build-ng2qk:
+    pipelineTaskName: build
+    status:
+      completionTime: "2020-05-04T02:10:49Z"
+      conditions:
+      - lastTransitionTime: "2020-05-04T02:10:49Z"
+        message: All Steps have completed executing
+        reason: Succeeded
+        status: "True"
+        type: Succeeded
+      podName: triggers-release-nightly-frwmw-build-ng2qk-pod-8vj99
+      resourcesResult:
+      - key: commit
+        resourceRef:
+          name: git-source-triggers-frwmw
+        value: 9ab5a1234166a89db352afa28f499d596ebb48db
+      startTime: "2020-05-04T02:05:07Z"
+      steps:
+      - container: step-build
+        imageID: docker-pullable://golang@sha256:a90f2671330831830e229c3554ce118009681ef88af659cd98bfafd13d5594f9
+        name: build
+        terminated:
+          containerID: docker://6b6471f501f59dbb7849f5cdde200f4eeb64302b862a27af68821a7fb2c25860
+          exitCode: 0
+          finishedAt: "2020-05-04T02:10:45Z"
+          reason: Completed
+          startedAt: "2020-05-04T02:06:24Z"
+  ```
+
+The following tables shows how to read the overall status of a `PipelineRun`:
+
+`status`|`reason`|`completionTime` is set|Description
+:-------|:-------|:---------------------:|--------------:
+Unknown|Started|No|The `PipelineRun` has just been picked up by the controller.
+Unknown|Running|No|The `PipelineRun` has been validate and started to perform its work.
+Unknown|PipelineRunCancelled|No|The user requested the PipelineRun to be cancelled. Cancellation has not be done yet.
+True|Succeeded|Yes|The `PipelineRun` completed successfully.
+True|Completed|Yes|The `PipelineRun` completed successfully, one or more Tasks were skipped.
+False|Failed|Yes|The `PipelineRun` failed because one of the `TaskRuns` failed.
+False|\[Error message\]|No|The `PipelineRun` encountered an non-permanent error, but it's still running and it may ultimately succeed.
+False|\[Error message\]|Yes|The `PipelineRun` failed with a permanent error (usually validation).
+False|PipelineRunCancelled|Yes|The `PipelineRun` was cancelled successfully.
+False|PipelineRunTimeout|Yes|The `PipelineRun` timed out.
+
+When a `PipelineRun` changes status, [events](events.md#pipelineruns) are triggered accordingly.
 
 ## Cancelling a `PipelineRun`
 
--- a/docs/taskruns.md
+++ b/docs/taskruns.md
@@ -155,7 +155,7 @@ point for the `Pod` in which the container images specified in your `Task` will 
 customize the `Pod` configuration specifically for that `TaskRun`.
 
 In the following example, the `Task` specifies a `volumeMount` (`my-cache`) object, also provided by the `TaskRun`,
-using a `PersistentVolumeClaim` volume. A specific scheduler is also configured in the  `SchedulerName` field. 
+using a `PersistentVolumeClaim` volume. A specific scheduler is also configured in the  `SchedulerName` field.
 The `Pod` executes with regular (non-root) user permissions.
 
 ```yaml
@@ -281,7 +281,7 @@ For more information, see [`ServiceAccount`](auth.md).
 ## Monitoring execution status
 
 As your `TaskRun` executes, its `status` field accumulates information on the execution of each `Step`
-as well as the `TaskRun` as a whole. This information includes start and stop times, exit codes, the 
+as well as the `TaskRun` as a whole. This information includes start and stop times, exit codes, the
 fully-qualified name of the container image, and the corresponding digest.
 
 **Note:** If any `Pods` have been [`OOMKilled`](https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/)
@@ -310,6 +310,23 @@ steps:
     reason: Completed
     startedAt: "2019-08-12T18:22:54Z"
   ```
+
+The following tables shows how to read the overall status of a `TaskRun`:
+
+`status`|`reason`|`completionTime` is set|Description
+:-------|:-------|:---------------------:|--------------:
+Unknown|Started|No|The TaskRun has just been picked up by the controller.
+Unknown|Pending|No|The TaskRun is waiting on a Pod in status Pending.
+Unknown|Running|No|The TaskRun has been validate and started to perform its work.
+Unknown|TaskRunCancelled|No|The user requested the TaskRun to be cancelled. Cancellation has not be done yet.
+True|Succeeded|Yes|The TaskRun completed successfully.
+False|Failed|Yes|The TaskRun failed because one of the steps failed.
+False|\[Error message\]|No|The TaskRun encountered a non-permanent error, and it's still running. It may ultimately succeed.
+False|\[Error message\]|Yes|The TaskRun failed with a permanent error (usually validation).
+False|TaskRunCancelled|Yes|The TaskRun was cancelled successfully.
+False|TaskRunTimeout|Yes|The TaskRun timed out.
+
+When a `TaskRun` changes status, [events](events.md#taskruns) are triggered accordingly.
 
 ### Monitoring `Steps`
 
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
@@ -73,6 +73,21 @@ func (pr *PipelineRun) GetTaskRunRef() corev1.ObjectReference {
 	}
 }
 
+// GetTypeMeta returns the task run type meta
+func (pr *PipelineRun) GetTypeMeta() *metav1.TypeMeta {
+	return &pr.TypeMeta
+}
+
+// GetObjectMeta returns the task run type meta
+func (pr *PipelineRun) GetObjectMeta() *metav1.ObjectMeta {
+	return &pr.ObjectMeta
+}
+
+// GetStatus returns the task run status as a RunsToCompletionStatus
+func (pr *PipelineRun) GetStatus() RunsToCompletionStatus {
+	return &pr.Status
+}
+
 // GetOwnerReference gets the pipeline run as owner reference for any related objects
 func (pr *PipelineRun) GetOwnerReference() metav1.OwnerReference {
 	return *metav1.NewControllerRef(pr, groupVersionKind)
@@ -101,6 +116,11 @@ func (pr *PipelineRun) GetRunKey() string {
 
 // IsTimedOut returns true if a pipelinerun has exceeded its spec.Timeout based on its status.Timeout
 func (pr *PipelineRun) IsTimedOut() bool {
+	return pr.HasTimedOut()
+}
+
+// HasTimedOut returns true if a pipelinerun has exceeded its spec.Timeout based on its status.Timeout
+func (pr *PipelineRun) HasTimedOut() bool {
 	pipelineTimeout := pr.Spec.Timeout
 	startTime := pr.Status.StartTime
 
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
@@ -73,18 +73,8 @@ func (pr *PipelineRun) GetTaskRunRef() corev1.ObjectReference {
 	}
 }
 
-// GetTypeMeta returns the task run type meta
-func (pr *PipelineRun) GetTypeMeta() *metav1.TypeMeta {
-	return &pr.TypeMeta
-}
-
-// GetObjectMeta returns the task run type meta
-func (pr *PipelineRun) GetObjectMeta() *metav1.ObjectMeta {
-	return &pr.ObjectMeta
-}
-
-// GetStatus returns the task run status as a RunsToCompletionStatus
-func (pr *PipelineRun) GetStatus() RunsToCompletionStatus {
+// GetStatusCondition returns the task run status as a ConditionAccessor
+func (pr *PipelineRun) GetStatusCondition() apis.ConditionAccessor {
 	return &pr.Status
 }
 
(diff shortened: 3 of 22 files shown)


afrittoli:
I cannot reproduce the failure on the timeout test locally - the reason at the end of the pipeline is the wrong one - but it works normally for me.
My suspicion is that status may fail to sync back to etcd occasionally. Perhaps we should do like knative does and retry the update when it fails.

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -72,6 +72,21 @@ func (pr *PipelineRun) GetTaskRunRef() corev1.ObjectReference {
...
+// GetObjectMeta returns the task run type meta
+func (pr *PipelineRun) GetObjectMeta() *metav1.ObjectMeta {
+	return &pr.ObjectMeta
+}
+
+// GetStatus returns the task run status as a RunsToCompletionStatus
+func (pr *PipelineRun) GetStatus() RunsToCompletionStatus {
+	return &pr.Status
+}
+
```
jlpettersson, line 89: What is the benefit with adding these methods, compared to use the fields directly? e.g. `pr.TypeMeta`, `pr.ObjectMeta` and `pr.Status` ?

Update: Now, I understand - it is part of an interface. Maybe it doesn't have to be part of the interface? I don't know.

docs/pipelineruns.md
```diff
@@ -337,6 +338,70 @@ The `timeout` value is a `duration` conforming to Go's
...
+:-------|:-------|:---------------------:|--------------:
+Unknown|Started|No|The `PipelineRun` has just been picked up by the controller.
+Unknown|Running|No|The `PipelineRun` has been validate and started to perform its work.
+Unknown|PipelineRunCancelled|No|The user requested the PipelineRun to be cancelled. Cancellation has not be done yet.
+True|Succeeded|Yes|The `PipelineRun` completed successfully.
+False|Failed|Yes|The `PipelineRun` failed because one of the `TaskRuns` failed.
+False|\[Error message\]|Yes|The `PipelineRun` failed with a permanent error (usually validation).
+False|PipelineRunCancelled|Yes|The `PipelineRun` was cancelled successfully.
+False|PipelineRunTimeout|Yes|The `PipelineRun` timed out.
+False|\[Error message\]|No|The `PipelineRun` encountered an error, but it's still running.
```
jlpettersson, line 401: Could the two with `[Error message]` but placed next to eachother, it is harder to detect the difference when it is 2 items between them.

docs/taskruns.md
```diff
@@ -311,6 +311,22 @@ steps:
...
+The following tables shows how to read the overall status of a `TaskRun`:
+
+`status`|`reason`|`completionTime` is set|TaskRun status
+:-------|:-------|:---------------------:|--------------:
+Unknown|Started|No|The TaskRun has just been picked up by the controller.
+Unknown|Running|No|The TaskRun has been validate and started to perform its work.
+Unknown|TaskRunCancelled|No|The user requested the TaskRun to be cancelled. Cancellation has not be done yet.
+True|Succeeded|Yes|The TaskRun completed successfully.
+False|Failed|Yes|The TaskRun failed because one of the steps failed.
+False|\[Error message\]|Yes|The TaskRun failed with a permanent error (usually validation).
```
jlpettersson, line 323: Nit: Also here, I find it easier to read if this is moved down to the other `[Error message]`.

docs/pipelineruns.md
afrittoli: Good point, I will fix that

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
afrittoli: Indeed, it's part of the interface. The idea is that a `RunsToCompletion` has a `RunsToCompletionStatus` from which Conditions can be extracted

afrittoli:
I fixed a small an issue in `InitializeConditions` which was causing CI to fail. I added more unit test coverage for that. The last issue in CI was actually the nop image broken: https://github.com/tektoncd/pipeline/pull/2713

afrittoli:
Something is wrong with the cloudevents YAML test
```
ERROR: tests timed out
>> Checking test results for taskrun
ERROR: test poll-for-content-run=SucceededUnknown but should be succeededtrue
ERROR: one or more YAML tests failed
```

pkg/apis/pipeline/v1beta1/run_interface.go
```diff
@@ -0,0 +1,42 @@
...
+	GetTypeMeta() *metav1.TypeMeta
+	GetObjectMeta() *metav1.ObjectMeta
+	GetOwnerReference() metav1.OwnerReference
+	GetStatus() RunsToCompletionStatus
+	IsDone() bool
+	HasStarted() bool
+	IsCancelled() bool
+	HasTimedOut() bool
+	GetRunKey() string
+}
```
vdemeester: @afrittoli any reason for defining those in here and not where it is used ? (it, then, wouldn't need to be exported even :angel:)

afrittoli: Right now it is only used by events, but I was hoping in future to use this to generalise more functions, perhaps even have a base reconciler for `RunToCompletion`.

That would help if we wanted to have dedicated reconcilers for metrics, cloud events or else in the future.

We could have this private to start with and export it in the next PRs if you preferred so.

Since I was adding functions to pipeline run and taskrun types to match this interface, I thought it would make sense to have the interface in the same place, but we could also move it to the controller package.

vdemeester: > We could have this private to start with and export it in the next PRs if you preferred so.

That would be my *preferred* approach indeed :stuck_out_tongue_closed_eyes: I definitely prefer more unexported interfaces (to use the power of the *implicit* part of the interfaces) and small exported interfaces.

afrittoli:
@n3wscott ping - this may be relevant to you - or you may have feedback on the event type scheme.

docs/pipelineruns.md
```diff
@@ -362,6 +363,70 @@ The `timeout` value is a `duration` conforming to Go's
...
 that do not have an idividual timeout set will fail immediately upon encountering an error.
 
+## Monitoring execution status
+
+As your `PipelineRun` executes, its `status` field accumulates information on the execution of each `TaskRun`
+as well as the `PipelineRun` as a whole. This information includes the name of the pipeline `Task` associated
+to a `TaskRun`, the complete [status of the `TaskrRun`](taskruns.md#monitoring-execution-status) and details
+about `Conditions` that may be associated to a `TaskRun`.
+
+The following example shows an extract from tje `status` field of a `PipelineRun` that has executed successfully:
```
pritidesai, line 373: minor nit: typo here tje instead of the

docs/pipelineruns.md
```diff
@@ -362,6 +363,70 @@ The `timeout` value is a `duration` conforming to Go's
...
+Unknown|PipelineRunCancelled|No|The user requested the PipelineRun to be cancelled. Cancellation has not be done yet.
+True|Succeeded|Yes|The `PipelineRun` completed successfully.
+False|Failed|Yes|The `PipelineRun` failed because one of the `TaskRuns` failed.
+False|\[Error message\]|No|The `PipelineRun` encountered an error, but it's still running.
+False|\[Error message\]|Yes|The `PipelineRun` failed with a permanent error (usually validation).
+False|PipelineRunCancelled|Yes|The `PipelineRun` was cancelled successfully.
+False|PipelineRunTimeout|Yes|The `PipelineRun` timed out.
+
+When a `PipelineRun` changes status, [events](events.md#pipelineruns) are triggered accordingly.
+
```
pritidesai, line 429: thanks @afrittoli for the table, its very helpful and easier to understand, few comments here:

1) it will be easier to understand `PipelineRunCancelled` if both the rows with that reason are in sequence.
2) We are missing `Completed` with status `True` in this table. I think we should include it since this table represents the overall status of `PipelineRun`.
3) May be rename column heading `PipelineRun Status` to description or something similar.
4) This is misleading "The PipelineRun encountered an error, but it's still running." This is happening because [PipelineRun Status](https://github.com/tektoncd/pipeline/blob/be9ac3bf6b27f6beb3f4b7c828ef8388f9b323c3/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L437) is updated with one failure without waiting for all the tasks in `nextRprts` to finish. We can add little more to this description until we fix issue #1680, "The PipelineRun encountered an error, but waiting for already running tasks to finish."

docs/taskruns.md
```diff
@@ -311,6 +311,23 @@ steps:
...
+Unknown|TaskRunCancelled|No|The user requested the TaskRun to be cancelled. Cancellation has not be done yet.
+True|Succeeded|Yes|The TaskRun completed successfully.
+False|Failed|Yes|The TaskRun failed because one of the steps failed.
+False|\[Error message\]|No|The TaskRun encountered an error, but it's still running.
+False|\[Error message\]|Yes|The TaskRun failed with a permanent error (usually validation).
+False|TaskRunCancelled|Yes|The TaskRun was cancelled successfully.
+False|TaskRunTimeout|Yes|The TaskRun timed out.
+
+When a `TaskRun` changes status, [events](events.md#taskruns) are triggered accordingly.
+
```
pritidesai, line 330: Similar observation as before, may be add a little more description to "The TaskRun encountered an error, but it's still running.", change it "The TaskRun encountered an error, but it's still running since waiting for other tasks or steps (need to check 🔍 ) to finish executing." 🤔

pkg/reconciler/run_interface.go
```diff
@@ -0,0 +1,53 @@
...
+
+package reconciler
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/apis"
+)
+
+// RunsToCompletion is implemented by TaskRun and PipelineRun
+type RunsToCompletion interface {
```
vdemeester, line 25: Any reason for having it here and not in `events` ? It could then be unexported now ?

docs/pipelineruns.md
afrittoli: Nice catch, thanks

docs/pipelineruns.md
afrittoli: > thanks @afrittoli for the table, its very helpful and easier to understand, few comments here:
> 
>     1. it will be easier to understand `PipelineRunCancelled` if both the rows with that reason are in a sequence.

They are sorted by status. I could sort them by reason, that would make it more readable from a Reason pov, less readable from a status pov.... not sure which is best.
I thought status should go first as it's the primary indicator of the condition.

> 
>     2. We are missing `Completed` with status `True` in this table. I think we should include it since this table represents the overall status of `PipelineRun`.
> 
Nice catch, thank you

>     3. May be rename column heading `PipelineRun Status` to description or something similar.

Fixed
> 
>     4. This is misleading "The PipelineRun encountered an error, but it's still running." This is happening because [PipelineRun Status](https://github.com/tektoncd/pipeline/blob/be9ac3bf6b27f6beb3f4b7c828ef8388f9b323c3/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L437) is updated with one failure without waiting for all the tasks in `nextRprts` to finish. We can add little more to this description until we fix issue #1680, "The PipelineRun encountered an error, but waiting for already running tasks to finish."

That's not the case I was referring to here. 
If any non-permanent error happens during reconcile, the reason is set to the error description, the status is set to failed, but it may be switch back to succeeded if the next reconcile cycle clears the error.

pkg/apis/pipeline/v1beta1/run_interface.go
afrittoli: done

docs/taskruns.md
afrittoli: Updated, this is again about transient errors

docs/pipelineruns.md
afrittoli: Uhm, I checked this again, and there does not seem to be any transient error reported by the pipelinerun controller, so I will remove this one.
I think I will add the point you made as a general notice to the pipeline run docs.

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -200,6 +210,32 @@ type PipelineRunStatus struct {
 	PipelineRunStatusFields `json:",inline"`
 }
 
+// PipelineRunSucceededReason represents a reason for the pipeline run "Succeeded" condition
+type PipelineRunSucceededReason string
```
pritidesai, line 214: nit: I feel this is coming out too long, can we rename this `PipelineRunSucceededReason` to `PipelineRunReason`, drop `Succeeded`? I understand it represents the `Succeeded` condition which could have status as failure or success depending on the execution but it could be implicit and just call it `PipelineRunReason`. Similar comment to all below reasons 😲  Rename `PipelineRunSucceededReasonStarted` to `PipelineRunReasonStarted` ...

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -200,6 +210,32 @@ type PipelineRunStatus struct {
...
+	PipelineRunSucceededReasonCompleted PipelineRunSucceededReason = "Completed"
+	// PipelineRunSucceededReasonFailed is the reason set when the PipelineRun completed with a failure
+	PipelineRunSucceededReasonFailed PipelineRunSucceededReason = "Failed"
+	// PipelineRunSucceededReasonCancelled is the reason set when the PipelineRun cancelled by the user
+	// This reason may be found with a corev1.ConditionFalse status, if the cancellation was processed successfully
+	// This reason may be found with a corev1.ConditionUnknown status, if the cancellation is being processed or failed
+	PipelineRunSucceededReasonCancelled PipelineRunSucceededReason = "Cancelled"
+	// PipelineRunSucceededReasonTimedOut is the reason set when the PipelineRun has timed out
+	PipelineRunSucceededReasonTimedOut PipelineRunSucceededReason = "PipelineRunTimeout"
+)
```
pritidesai, line 233: I love the way its all in one place now and inside of `types` instead of part of `resources` 🙏

pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -102,6 +98,43 @@ type TaskRunStatus struct {
...
+	TaskRunSucceededReasonRunning TaskRunSucceededReason = "Running"
+	// TaskRunSucceededReasonSuccessful is the reason set when the TaskRun completed successfully
+	TaskRunSucceededReasonSuccessful TaskRunSucceededReason = "Succeeded"
+	// TaskRunSucceededReasonFailed is the reason set when the TaskRun completed with a failure
+	TaskRunSucceededReasonFailed TaskRunSucceededReason = "Failed"
+	// TaskRunSucceededReasonCancelled is the reason set when the Taskrun is cancelled by the user
+	TaskRunSucceededReasonCancelled TaskRunSucceededReason = "TaskRunCancelled"
+	// TaskRunSucceededReasonTimedOut is the reason set when the Taskrun has timed out
+	TaskRunSucceededReasonTimedOut TaskRunSucceededReason = "TaskRunTimeout"
+)
```
pritidesai, line 119: Similar NIT as above, if we can rename `TaskRunSucceededReason` to `TaskRunReason`, `TaskRunSucceededReasonStarted` to `TaskRunReasonStarted` ...

Nicely collected all reasons under `types` 🙏

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -210,13 +246,22 @@ func (pr *PipelineRunStatus) GetCondition(t apis.ConditionType) *apis.Condition
...
 	if pr.StartTime.IsZero() {
 		pr.StartTime = &metav1.Time{Time: time.Now()}
+		started = true
+	}
+	conditionManager := pipelineRunCondSet.Manage(pr)
+	conditionManager.InitializeConditions()
+	// Ensure the started reason is set for the "Succeeded" condition
+	if started {
+		initialCondition := conditionManager.GetCondition(apis.ConditionSucceeded)
+		initialCondition.Reason = TaskRunSucceededReasonStarted.String()
```
pritidesai, line 262: Should this be `PipelineRunSucceededReasonStarted` instead of `TaskRunSucceededReasonStarted` ? 🤔  which makes me think, should we combine common PipelineRunReasons and TaskRunReasons e.g. `Started`, `Failed`, `Completed` etc?

pkg/pod/status.go
```diff
@@ -114,7 +104,7 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev
 		trs.SetCondition(&apis.Condition{
```
pritidesai, line 104: Should we create MarkResource* routine for these conditions and use those?

pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -35,12 +35,32 @@ import (
 type TektonEventType string
 
 const (
+	// TektonTaskRunStartedV1 is sent for TaskRuns with "ConditionSucceeded" "Unknown"
+	// the first time they are picked up by the reconciler
+	TektonTaskRunStartedV1 TektonEventType = "dev.tekton.event.taskrun.started.v1"
```
pritidesai, line 40: NIT: rename `TektonTaskRunStarted` to `EventTaskRunStarted` 🤔  and same for all events below 😲

pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -71,27 +118,62 @@ func EventForTaskRun(taskRun *v1beta1.TaskRun) (*cloudevents.Event, error) {
...
 	case c.IsTrue():
-		event.SetType(TektonTaskRunSuccessfulV1.String())
+		switch runObject.(type) {
+		case *v1beta1.TaskRun:
+			eventType = TektonTaskRunSuccessfulV1
+		case *v1beta1.PipelineRun:
+			eventType = TektonPipelineRunSuccessfulV1
+		}
 	default:
 		return nil, fmt.Errorf("unknown condition for in TaskRun.Status %s", c.Status)
```
pritidesai, line 174: NIT: we are now including PipelineRun.Status here as well, may be change this error message to include PipelineRun as well

docs/taskruns.md
pritidesai: something like?

```
{"level":"info","logger":"tekton.taskrun-controller.event-broadcaster","caller":"record/event.go:274","msg":"Event(v1.ObjectReference{Kind:\"TaskRun\", Namespace:\"default\", Name:\"pipelinerun-one-failure-two-success-task-c-9fmnd\", UID:\"a788c615-9058-45ae-8dea-de1310343208\", APIVersion:\"tekton.dev/v1beta1\", ResourceVersion:\"4190777\", FieldPath:\"\"}): type: 'Warning' reason: 'Error' Operation cannot be fulfilled on taskruns.tekton.dev \"pipelinerun-one-failure-two-success-task-c-9fmnd\": the object has been modified; please apply your changes to the latest version and try again","commit":"ac439fb","knative.dev/controller":"taskrun-controller"}
```

pritidesai:
Thanks @afrittoli it looks great, couple of comments above and one more here:

Didnt see Event for PipelineRun Started in (logs)[https://gist.github.com/pritidesai/f90d9057e401bd2b36d84fd1353687eb] after fixing https://github.com/tektoncd/pipeline/pull/2677#discussion_r436158551

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
afrittoli: Yeah, I was debating about this one too. 
I thought we might have more conditions in future, and being on the API I wanted to have it unique... but no be fair there would be nothing wrong with having the same reason used in different conditions, so I'll go ahead and shorten the names :)

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
afrittoli: Nice catch, thank you!
I think I'd still keep them separated, even if the values are identical.

pkg/pod/status.go
afrittoli: We might do so to reduce boilerplate code, but I would not do it in this PR - https://github.com/tektoncd/pipeline/issues/2770

pkg/reconciler/events/cloudevent/cloudevent.go
afrittoli: Uhm, maybe `TaskRunStartedEvent`?

afrittoli: Good point, that Tekton in there is redundant

pkg/reconciler/events/cloudevent/cloudevent.go
afrittoli: done

afrittoli:
> Didnt see Event for PipelineRun Started in [logs](https://gist.github.com/pritidesai/f90d9057e401bd2b36d84fd1353687eb) after fixing [#2677 (comment)](https://github.com/tektoncd/pipeline/pull/2677#discussion_r436158551)

The start event requires https://github.com/tektoncd/pipeline/pull/2545 :)

afrittoli:
Rebased and addressed comments @pritidesai @vdemeester

docs/taskruns.md
afrittoli: Yep, failures on update are the only case. Hopefully with the genreconciler we'll see less of that

afrittoli:
yet another rebase :)

pritidesai:
thanks @afrittoli, it looks great 🙏 
/lgtm

PR #2693, opened 2020-05-26
WIP: add SAR check on SA for TaskRun/PipelineRun reference to ClusterTask;…

gabemontero:
… add Role/RoleBinding for default SA in project config; related test/doc changes

# Changes

This is the first (and simplest) of the set of[ subject access review (SAR)](https://kubernetes.io/docs/reference/access-authn-authz/authorization/#checking-api-access) checks that verify that the entity (the service account in this case) has permissions per the [role-based access control system (RBAC)](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) of k8s I brought up in recent WG sessions.

In particular, this change focuses on the read the cluster scoped `ClusterTask` references that can be done by either a namespace scoped `TaskRun` or namespaced scoped `PipelineRun`, and it verifies that the underlying Service Account (SA) of those namespaced scoped objects is allowed to access `ClusterTask` objects.

A new role and rolebinding so that the `default` service account in the `tekton-pipelines` namesapce can access `ClusterTask` is also defined, to preserve current behavior.  However, the permissions to do this are now explicitly declared, instead of inheriting the permissions of the Tekton Pipeline controller in a non-explicit fashion, which goes against recommendations and best practices around controllers manipulating API objects on behalf of other "users" (SAs most notably) in the system.

With these changes, cluster administrators can now control which namespaces have access to which `ClusterTasks` managed by a single Tekton Pipelines controller.

Also, as I indicated in the pipeline WG, I will be creating a design doc around the broader set of scenarios, but I thought it instructive to those unfamiliar with SAR/RBAC to have an actual example to cross reference.

And IMO, this change on its own *could* be merged independent of iterating over that doc and the more complicated scenarios.  But of course, let's see how the review and discussion goes, and decide on that.

# Submitter Checklist

review them:

- [/] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [/] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [/] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

## Reviewer Notes

# Release Notes

```

- API changes
- Bug fixes
- Any changes in behavior

```

### DIFF
--- a/config/200-clusterrole.yaml
+++ b/config/200-clusterrole.yaml
@@ -43,6 +43,9 @@ rules:
     resources: ["podsecuritypolicies"]
     resourceNames: ["tekton-pipelines"]
     verbs: ["use"]
+  - apiGroups: ["authorization.k8s.io"]
+    resources: ["subjectaccessreviews"]
+    verbs: ["create"]
 ---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
--- a/config/200-role.yaml
+++ b/config/200-role.yaml
@@ -59,3 +59,31 @@ rules:
     resources: ["secrets"]
     verbs: ["get", "update"]
     resourceNames: ["webhook-certs"]
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: tekton-view-clustertask
+  namespace: tekton-pipelines
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipelines
+rules:
+  - apiGroups: ["tekton.dev"]
+    resources: ["clustertasks"]
+    verbs: ["get", "list", "watch"]
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: tekton-view-clustertask
+  namespace: default
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipelines
+rules:
+  - apiGroups: ["tekton.dev"]
+    resources: ["clustertasks"]
+    verbs: ["get", "list", "watch"]
--- a/config/201-rolebinding.yaml
+++ b/config/201-rolebinding.yaml
@@ -47,3 +47,39 @@ roleRef:
   kind: Role
   name: tekton-pipelines-webhook
   apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: RoleBinding
+metadata:
+  name: tekton-view-clustertask
+  namespace: tekton-pipelines
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipelines
+subjects:
+  - kind: ServiceAccount
+    name: default
+    namespace: tekton-pipelines
+roleRef:
+  kind: Role
+  name: tekton-view-clustertask
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: RoleBinding
+metadata:
+  name: tekton-view-clustertask
+  namespace: default
+  labels:
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipelines
+subjects:
+  - kind: ServiceAccount
+    name: default
+    namespace: default
+roleRef:
+  kind: Role
+  name: tekton-view-clustertask
+  apiGroup: rbac.authorization.k8s.io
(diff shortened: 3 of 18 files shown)


gabemontero:
I'll remove the WIP once I clean up various CI hiccups

bobcatfish:
Hey @gabemontero ! Is there a design doc that goes along with this? In particular I'd like to see what alternatives we have to implementing this in the controller - one really robust approach would be to have some other component responsible for these checks such that the Tekton controller isn't aware at all. Is that a possibility?

gabemontero:
> Hey @gabemontero ! Is there a design doc that goes along with this? 

Hey @bobcatfish ! .... certainly understandable if it got lost in the tl;dr of my PR description, but there is no design doc *yet*, and that I chose to create this PR prior to getting the doc ready as either a) something we could merge without a doc, or b) a means to facilitate consumption of the upcoming doc ... i.e. a "reference implementation" if you will.

Certainly I am amenable to hold off (i.e. a hold label vs. the current WIP label) on merging this PR until the design doc is "processed" if the community ultimately deems that necessary.

> In particular I'd like to see what alternatives we have to implementing this in the controller - one really robust approach would be to have some other component responsible for these checks such that the Tekton controller isn't aware at all. Is that a possibility?

Doing these sort of checks have traditionally been something you do *within* a controller, as part of the controller's (perhaps primary) purpose of interacting with the k8s API server *on behalf* of another party.

But to your question (and to add a couple of logical follow ups):

- Is it possible:  depends on what is meant by "isn't aware at all"

- Could it lead to a fair amount of code duplication: quite possible, again, depending on how are far you go with "isn't aware at all"

- Does it break with k8s "convention":  yes

- Is breaking with k8s "convention" a "bad" thing: probably, but to be fair, who you ask may have some bearing on that answer

Lastly, aside from massaging this PR, and getting cycles to work on the doc, my thought was to see about getting this on the agenda for the API WG, to get some more detail discussion and table setting prior to any design document writing....would that seem warranted / appropriate to you?

thanks

bobcatfish:
Thanks for the additional context @gabemontero 

> a) something we could merge without a doc

I'd push for a design doc first - especially since this is not only a new feature, but imo a whole new area of features that we're opening up.

> Doing these sort of checks have traditionally been something you do within a controller, as part of the controller's (perhaps primary) purpose of interacting with the k8s API server on behalf of another party.

I'm wondering if this is something that could be done in a completely separate controller, with a couple of different ways this could look:

- A different controller, deployed with the existing tekton controller + webhook controller
- A different controller that is released completely separately (maybe even lives in a different repo)
- A different controller that uses some kind of policy engine such as OPA to express these kinds of rules <-- this would be my preference, and would ideally result in very little work on our part, we could trust systems that are built for this kind of policy enforcement to carry out the job vs. adding it ourselves (if there are any gaps to fill maybe we can provide the components to fill them? e.g. recommended tekton policies)

gabemontero:
> Thanks for the additional context @gabemontero
> 
> > a) something we could merge without a doc
> 
> I'd push for a design doc first - especially since this is not only a new feature, but imo a whole new area of features that we're opening up.

Sure thing.  I'll still iterate on it to clean up the remaining CI blockers, but will place a hold / keep the WIP as needed and ref the design doc once it is up.


> 
> > Doing these sort of checks have traditionally been something you do within a controller, as part of the controller's (perhaps primary) purpose of interacting with the k8s API server on behalf of another party.
> 
> I'm wondering if this is something that could be done in a completely separate controller, with a couple of different ways this could look:
> 
>     * A different controller, deployed with the existing tekton controller + webhook controller
> 
>     * A different controller that is released completely separately (maybe even lives in a different repo)
> 
>     * A different controller that uses some kind of policy engine such as OPA to express these kinds of rules <-- this would be my preference, and would ideally result in very little work on our part, we could trust systems that are built for this kind of policy enforcement to carry out the job vs. adding it ourselves (if there are any gaps to fill maybe we can provide the components to fill them? e.g. recommended tekton policies)

Yeah some of the details will be critical here IMO.  For example,
- we still do the processing in the existing controller to discern the `ClusterTask` we are referencing and the SA the `TaskRun` and `PipelineRun` are using, and then supply those as "parameters" to a "call" to the other controller.  Now, the parameters / call could be off the top of my head
a) a REST invocation between controllers
b) a new instance of a CR with spec/status that each controller operate over 
c) or additional ideas

where the pro / con list revolves around increased overhead vs. the "policy engine abstraction" you are getting at

But fodder for the upcoming discussions.

gabemontero:
OK, with the integration tests, my new e2e's passed, as well as the existing ones in explicit .go files, but the `RUNNING YAML E2E TESTS FOR V1ALPHA1 TASKRUNS` run using the clustertask related yamls in the `examples` subdir failed.

By comparison, the other three combos, v1beta1/taskrun, and v1alpha1 and v1beta1 pipelinerun passed.

I'm off sorting out the wiring of those yaml e2e tests along with v1beta1 vs. v1alpha1 diffs in my changes.

gabemontero:
OK all green tests.

Moving on to the design doc.  Will maintain this PR minimally for reference to go in conjunction with design doc.

Issue #2694, opened 2020-05-26
Conceptual bug: Wrong resource names used


dlorenc:
This is a conceptual bug in the way resource names are handled in Tekton today. I'm working on a fix, but wanted to discuss the actual issue here first.

PipelineResources right now can be referred to by two different names:

1. The PipelineResourceSpec.Params.Name field. This is not a top-level field in PipelineResources, but (almost?) all of the existing PipelineResource types take a parameter called "Name" that is then set on the object itself after unmarshalling.
This name can be thought of as a way to refer to the external object the PipelineResource refers to.
2. The TaskRun.Inputs|Outputs.Resources.Name field. This is a field on the ResourceBinding. This can be thought of as an internal "name" the TaskRun will use to refer to the resource.

Right now, we pass the external 'Name' field to the resource containers via the 'TEKTON_RESOURCE_NAME' env var. This is only used in a few locations right now:

- substitutions
- results ( to indicate which resource the result corresponds to )

We should be using the 'internal' name field here - the external names are not always set and can change/not be unique. Internal names are unique within a Task/TaskRun.

afrittoli:
I'm trying to understand but I'm not sure I understood the problem fully.
So here is my understanding of what the problem is.

The struct that holds all `PipelineResource` types has a field called "Name".  That field is not a `PipelineResource` parameter. For instance for the `git` resource, the struct includes name:
https://github.com/tektoncd/pipeline/blob/4d629484ca894c9ca03cd315bf6ac93d3a71a8d0/pkg/apis/resource/v1alpha1/git/git_resource.go#L37-L53

Name is not parsed from parameters: https://github.com/tektoncd/pipeline/blob/4d629484ca894c9ca03cd315bf6ac93d3a71a8d0/pkg/apis/resource/v1alpha1/git/git_resource.go#L68-L87

Name is taken instead from the PipelineResource CRD "Name" field. I think this is what you refer to as the "external" name.

A `Task` defines its own resource names, and a `TaskRun` binds the "external" name of a resource to the task resource. The code in the `TaskRun` controller should use the bound resource name only for fetching the resource spec, and the task resource name otherwise.

The `git` pipeline resource uses the `TEKTON_RESOURCE_NAME` to report results. 
The `image` resource a CLI flag instead to report results.

In both cases, a `v1beta1.PipelineResourceResult` is created, which holds a `v1beta1.PipelineResourceRef`. The resource ref points to the resource name and kind.
[...]

PR #2697, opened 2020-05-26
Fix PipelineResource name handling.

dlorenc:
# Changes

This commit changes the names used throughout PipelineResource handling.
PipelineResources have two names when used in a TaskRun:
- The name of the PipelineResource entity ("external name")
- The name given to the resource inside the Task definition ("internal name")

Task code is currently using the external names everywhere, but should be using
the internal names.

Ref: https://github.com/tektoncd/pipeline/issues/2694

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
The TaskRun.Status.ResourceResults field now contains a new field called "ResourceName" that points to the name of the PipelineResourceBinding.

The TaskRun.Status.ResourcesResults.ResourceRef.Name field points to this same object. This field is now redundant and will be removed in a future release.

The ClusterResource has been updated to write it's kubeconfig to the correct location on disk. Previously it was written to a location under /workspace/input/<resource ref name>, instead of the correct /workspace/input/<resource binding name>. If you hardcode resource paths in your tasks, you will need to update these. If you refer to paths by interpolation ( $(resources.inputs.<resource name>) ), you do not need to make any changes.
Both paths will work for this release, but you will need to update your Task definitions by a future release.
```


bobcatfish:
hi all! can we get the release notes field filled out? just cruising by this one and i want to understand the user facing impact (if any - it seems like at least a bug fix tho?)

dlorenc:
> hi all! can we get the release notes field filled out? just cruising by this one and i want to understand the user facing impact (if any - it seems like at least a bug fix tho?)

It's a bug fix that sort of requires an API change. Release Notes should be filled in now, we might want to discuss this in next week's meeting though before merging.

vdemeester:
@dlorenc needs a rebase :sweat_smile:

go.sum
```diff
@@ -1402,7 +1395,6 @@ sigs.k8s.io/kustomize v2.0.3+incompatible/go.mod h1:MkjgH3RdOWrievjo6c9T245dYlB5
 sigs.k8s.io/structured-merge-diff v0.0.0-20190525122527-15d366b2352e/go.mod h1:wWxsB5ozmmv/SG7nM11ayaAW51xMvak/t1r0CSlcokI=
 sigs.k8s.io/structured-merge-diff v1.0.1-0.20191108220359-b1b620dd3f06/go.mod h1:/ULNhyfzRopfcjskuui0cTITekDduZ7ycKN3oUT9R18=
 sigs.k8s.io/structured-merge-diff/v2 v2.0.1/go.mod h1:Wb7vfKAodbKgf6tn1Kl0VvGj7mRH6DGaRcixXEJXTsE=
-sigs.k8s.io/yaml v1.1.0 h1:4A07+ZFc2wgJwo8YNlQpr1rVlgUDlxXHhPJciaPY5gs=
```
bobcatfish, line 1405: i have given up on reviewing these files, i just assume they are fine b/c I don't understand when then change or why. someone plz let me know if there is something that makes sense to look for in these files.

test/v1alpha1/cluster_resource_test.go
```diff
@@ -140,13 +140,13 @@ clusters:
...
+  name: target-cluster
 contexts:
 - context:
-    cluster: helloworld-cluster
+    cluster: target-cluster
     user: test-user
-  name: helloworld-cluster
-current-context: helloworld-cluster
+  name: target-cluster
+current-context: target-cluster
```
bobcatfish, line 149: i dont totally understand why this changed - i thought we were changing the fields used in status, this seems to imply we might be making a backwards incompatible change to the behavior of the cluster resource? lemme know if im wrong tho

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -314,11 +314,7 @@ func (c *Reconciler) prepare(ctx context.Context, tr *v1beta1.TaskRun) (*v1beta1
 	// FIXME(afrittoli) This resource specific logic will have to be replaced
 	// once we have a custom PipelineResource framework in place.
 	c.Logger.Infof("Cloud Events: %s", tr.Status.CloudEvents)
-	prs := make([]*resourcev1alpha1.PipelineResource, 0, len(rtr.Outputs))
-	for _, pr := range rtr.Outputs {
-		prs = append(prs, pr)
-	}
-	cloudevent.InitializeCloudEvents(tr, prs)
+	cloudevent.InitializeCloudEvents(tr, rtr.Outputs)
```
bobcatfish, line 317: hah nice

go.sum
dlorenc: No idea...

dlorenc: No idea...

vdemeester: `go mod tidy` should keep those "kinda" clean :upside_down_face: but yeah, `go mod …` kinda tend to do a little bit of dark magic…

test/v1alpha1/cluster_resource_test.go
dlorenc: Yeah I guess this is a separate bug being fixed as a side effect - unsure of the best way to handle it. The resource in this case was named "target-cluster" in the task, the directory it was written to always should have been target-cluster. 

It looks like all the other resources determine where to write data using the "path" provided to GetInputTaskModifier, but the cluster resource used the incorrect name field.

People that referenced cluster resources using interpolation like `$(resources.inputs.target-cluster)` should see no change, but those that hardcoded it like this test did would see a breakage.

Any suggestions?

dlorenc:
> * If we want to deprecate the old field, could you add it to https://github.com/tektoncd/pipeline/blob/master/docs/deprecations.md - i also want to double check, do we definitely want to remove it? it might make sense to have both? (or maybe it ultimately would make more sense to embed the resourcespecs in the status in the same way as we do the task spec? in that case i think deprecating the field makes sense - though you still might want to be able to map the task's resource name to the actual resource name used? ... i guess that's already in the spec?)

Yes, we definitely want to delete this. It doesn't make semantic sense for the ResourceResult to contain a ResourceRef for a few reasons:
- The Ref could change later
- The Ref may never exist - the user could have specified it inline

Instead it should probably contain a fully dereferenced copy of the resource (like you've suggested), but that should wait until we figure out the plan for resources.

dlorenc:
> My one concern is that based on how the cluster end to end test changed, is this a backwards incompatible change for the behavior of the cluster pipelineresource, or was it just a quirk of how the test was setup? If the latter then I think we're all good, but if it is a backwards incompatible change we might want to do it a bit more slowly.

And yes, this is slightly backwards incompatible. I just changed the test to be more clear. The old behavior for ClusterResources was just plain wrong - they were writing to the wrong location on disk. Referring to the location via $(resources.inputs.<name>) still works and provides stable behavior.

I've explained this in the notes. I don't think there's a lot more we can do here...

dlorenc:
OK, this has now been changed (and the release notes are updated) to keep both paths in place for now. We can remove the old path in the next release.

docs/deprecations.md
```diff
@@ -23,3 +23,4 @@ being deprecated.
 | [`tekton.dev/task` label on ClusterTasks](https://github.com/tektoncd/pipeline/issues/2533) | [v0.12.0](https://github.com/tektoncd/pipeline/releases/tag/v0.12.0) | Beta | January 30 2021 |
 | [Step `$HOME` env var defaults to `/tekton/home`](https://github.com/tektoncd/pipeline/issues/2013) | [v0.11.0-rc1](https://github.com/tektoncd/pipeline/releases/tag/v0.11.0-rc1) | Beta | December 4 2020 |
 | [Step `workingDir` defaults to `/workspace`](https://github.com/tektoncd/pipeline/issues/1836) | [v0.11.0-rc1](https://github.com/tektoncd/pipeline/releases/tag/v0.11.0-rc1) | Beta | December 4 2020 |
+| [The `TaskRun.Status.ResourceResults.ResourceRef` field is deprecated and will be removed.](https://github.com/tektoncd/pipeline/issues/2694) | [v0.13.0](https://github.com/tektoncd/pipeline/releases/tag/v0.13.0) | Beta | December 4 2020 |
```
bobcatfish, line 26: * maybe we can point them at the field they should use instead as well
* since 0.13.0 is out, we could tentatively put 0.14.0 as the release with the warning
* with our policy [as it stands for beta (assuming this counts as beta? kind of a grey area, im open to treating this as alpha, but it is in the status field of the Task so safer to treat it like beta...)](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#alpha-beta-and-ga) you need at least 9 months warning, so if we do 0.14.0 in july, 9 months would be .... Apr 2021?

pkg/apis/resource/v1alpha1/cluster/cluster_resource.go
```diff
@@ -58,17 +59,23 @@ type Resource struct {
 	Secrets []resource.SecretParam `json:"secrets"`
 
 	KubeconfigWriterImage string `json:"-"`
+	ShellImage            string `json:"-"`
+
+	// Temproary field to hold the old, legacy value for name. See #2394
```
bobcatfish, line 64: typo: temporary

docs/deprecations.md
bobcatfish: scope creep, feel free to ignore me: you mentioned issue number 2694 in all the deprecation comments below, it'd probably be useful to have another column in this table to point folks at issues they can reference to track the progress of coming deprecations

test/cluster_resource_test.go
```diff
@@ -176,13 +176,13 @@ clusters:
...
+  name: target-cluster
 contexts:
 - context:
-    cluster: helloworld-cluster
+    cluster: target-cluster
     user: test-user
-  name: helloworld-cluster
-current-context: helloworld-cluster
+  name: target-cluster
+current-context: target-cluster
```
bobcatfish, line 185: is there a test that also uses the deprecated one, to make sure it works properly?

pkg/apis/resource/v1alpha1/cluster/cluster_resource.go
```diff
@@ -189,6 +196,9 @@ func (s *Resource) GetInputTaskModifier(ts *v1beta1.TaskSpec, path string) (v1be
 		Env: envVars,
 	}}
 	return &v1beta1.InternalTaskModifier{
-		StepsToPrepend: []v1beta1.Step{step},
+		StepsToPrepend: []v1beta1.Step{
+			step,
+			storage.CreateDirStep(s.ShellImage, s.Name, path),
```
bobcatfish, line 201: maybe a comment about 2694?

docs/deprecations.md
bobcatfish: would it make sense to also mention here that this is also making a dir at the legacy location?

dlorenc: Sorry so just to be clear on the policy - there are two changes we would be making.

- Removing the field from our API. The policy is a bit unclear - if we release a v1beta2 without this field, do we need to wait 9 months? https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md
- Writing directories to both locations. This one is more clearly alpha, since it's the behavior of Resources, not the API. I think we can remove this in one more release.

bobcatfish: > Writing directories to both locations. This one is more clearly alpha, since it's the behavior of Resources, not the API. I think we can remove this in one more release.

Ah, right, that makes sense - and I think we discussed that already, sorry for forgetting!

> Removing the field from our API. The policy is a bit unclear - if we release a v1beta2 without this field, do we need to wait 9 months? https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md

Interesting point - I always assumed this meant it would be 9 months at least until any kind of release went out which make the change in a backwards incompatible way, but I guess once v1beta2 is out, folks could keep using the field via v1beta1 presumably for long enough to meet the 9 month criteria.

I opened https://github.com/tektoncd/pipeline/issues/2790 to discuss. in the meantime i guess just use your best judgement in this PR :D (since there is another entry that says "January 30 2021" i do think other ppl have been assuming 9 months from when the change is introduced)

test/cluster_resource_test.go
dlorenc: Added one!

Issue #2587, opened 2020-05-08
Wrap `cmp.Diff` to make clear the (want, got) ordering of arguments.


sbwsg:
# Expected Behavior

In https://github.com/tektoncd/pipeline/pull/2580 we are trying to make test diff formatting uniform.  There is still room for user-error though: if the user did not call `cmp.Diff` with arguments in the expected order `(want, got)`, then the diff they print with `diff.Print` will be incorrect / reversed.

One possible improvement over this would be to wrap `cmp.Diff` with a [B.Helper](https://golang.org/pkg/testing/#B.Helper) called something like `diff.FailIfWantGot(want, got, interface{}, format string)` which clearly indicates the order of arguments in the func's name and could apply the correct formatting.  Usage could look something like this ([suggested by @bobcatfish in this comment](https://github.com/tektoncd/pipeline/pull/2580#discussion_r421788275)):

```go
diff.FailIfWantNotGot(tc.expectedTaskSpec, *ts, "Didn't get expected TaskSpec modifications")
```

# Additional Considerations

1. `cmp.Diff` takes additional options which some of tests utilize to ignore certain fields.

2. Some of our tests use cmp.Diff with `t.Fatalf` and some with `t.Errorf`.  This could be another parameter to the func?  Or we could have two helpers?  Or we could not worry about it and only support one of them?

```go
// Have separate funcs for errorf and fatalf
// Errorf
diff.ErrorIfWantNotGot(tc.expectedTaskSpec, *ts, "Didn't get expected modifications")
// Fatalf
diff.FatalIfWantNotGot(tc.expectedTaskSpec, *ts, "Didn't get expected modifications")
```

```go
// Pass format printer to func
diff.WantGot(tc.expectedTaskSpec, *ts, t.Fatalf, "Didn't get expected TaskSpec modifications")
```

2pai:
hi @sbwsg, this is my first time looking to tektoncd repository, 
is it okay for a first-timer to have a shot on this?

sbwsg:
@2pai Sure thing, thanks for your interest!  Since this is your first contribution I suggest you write the helper funcs first and make a PR.  Once code reviewers are happy with the helpers you could then replace usages of `cmp.Diff` in a separate commit or PR.

sbwsg:
Also, feel free to reach out to me on [Slack](https://github.com/tektoncd/community/blob/master/contact.md#slack) (user: Scott) if you need any pointers.

[...]

PR #2702, opened 2020-05-28
Add helper function ErrorIfWantNotGot & FatalIfWantNotGot to wrap cmp.Diff

2pai:
# Changes
This PR is trying to follow up #2587 
Including : 
- Adding a helper function called `FatalIfWantNotGot` & `ErrorIfWantNotGot` to wrap the cmp.Diff()
- Replace varying cmp.diff by using `FatalIfWantNotGot` & `ErrorIfWantNotGot` 

The implementation will be updated after the helper function is approved
 

# Submitter Checklist

review them:

### DIFF
--- a/test/diff/helper.go
+++ b/test/diff/helper.go
@@ -1,0 +1,44 @@
+/*
+Copyright 2020 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package diff
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+// ErrorWantGot Wrap cmp.Diff with a t.Helper(). It takes the testing.T
+// want,got, format message and cmp.option as an argument. Therefore, if diff was found
+// it will call t.Error of format and diff that wrapped by PrintWantGot func
+func ErrorWantGot(t *testing.T, want interface{}, got interface{}, format string, opts ...cmp.Option) {
+	t.Helper()
+	if d := cmp.Diff(want, got, opts...); d != "" {
+		t.Errorf(fmt.Sprintf("%s %s", format, PrintWantGot(d)))
+	}
+}
+
+// FatalWantGot Wrap cmp.Diff with a t.Helper(). It takes the testing.T
+// want,got, format message and cmp.option as an argument. Therefore, if diff was found
+// it will call t.Fatalf of format and diff that wrapped by PrintWantGot func
+func FatalWantGot(t *testing.T, want interface{}, got interface{}, format string, opts ...cmp.Option) {
+	t.Helper()
+	if d := cmp.Diff(want, got, opts...); d != "" {
+		t.Fatalf(fmt.Sprintf("%s %s", format, PrintWantGot(d)))
+	}
+}
--- a/pkg/apis/config/default_test.go
+++ b/pkg/apis/config/default_test.go
@@ -19,7 +19,6 @@ package config_test
 import (
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/pod"
 	test "github.com/tektoncd/pipeline/pkg/reconciler/testing"
@@ -187,9 +186,7 @@ func TestEquals(t *testing.T) {
 func verifyConfigFileWithExpectedConfig(t *testing.T, fileName string, expectedConfig *config.Defaults) {
 	cm := test.ConfigMapFromTestFile(t, fileName)
 	if Defaults, err := config.NewDefaultsFromConfigMap(cm); err == nil {
-		if d := cmp.Diff(Defaults, expectedConfig); d != "" {
-			t.Errorf("Diff:\n%s", diff.PrintWantGot(d))
-		}
+		diff.ErrorWantGot(t, Defaults, expectedConfig, "Diff:\n%s")
 	} else {
 		t.Errorf("NewDefaultsFromConfigMap(actual) = %v", err)
 	}
--- a/pkg/apis/config/store_test.go
+++ b/pkg/apis/config/store_test.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/tektoncd/pipeline/pkg/apis/config"
 	test "github.com/tektoncd/pipeline/pkg/reconciler/testing"
 	"github.com/tektoncd/pipeline/test/diff"
@@ -45,7 +44,5 @@ func TestStoreLoadWithContext(t *testing.T) {
 
 	cfg := config.FromContext(store.ToContext(context.Background()))
 
-	if d := cmp.Diff(cfg, expected); d != "" {
-		t.Errorf("Unexpected config %s", diff.PrintWantGot(d))
-	}
+	diff.ErrorWantGot(t, cfg, expected, "Unexpected config %s")
 }
--- a/pkg/apis/pipeline/v1alpha1/cluster_task_conversion_test.go
+++ b/pkg/apis/pipeline/v1alpha1/cluster_task_conversion_test.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	resource "github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1"
 	"github.com/tektoncd/pipeline/test/diff"
@@ -176,9 +175,7 @@ func TestClusterTaskConversion(t *testing.T) {
 					t.Errorf("ConvertFrom() = %v", err)
 				}
 				t.Logf("ConvertFrom() = %#v", got)
-				if d := cmp.Diff(test.in, got); d != "" {
-					t.Errorf("roundtrip %s", diff.PrintWantGot(d))
-				}
+				diff.ErrorWantGot(t, test.in, got, "roundtrip %s")
 			})
 		}
 	}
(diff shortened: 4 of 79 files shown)


2pai:
I'm sorry for late submitting this PR
This is my first time contributing to this project

sbwsg:
> I'm sorry for late submitting this PR
> This is my first time contributing to this project

Welcome!  Thanks for the contribution!

test/diff/helper.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// ErrorIfWantNotGot Wrap cmp.Diff with a t.Helper(). It takes the testing.T
+// want,got, format message and cmp.option as an argument. Therefore, if diff was found
+// it will call t.Error of format and diff that wrapped by PrintWantGot func
+func ErrorIfWantNotGot(t *testing.T, want interface{}, got interface{}, format string, opts ...cmp.Option) {
+	// t.Helper() used to marks the calling function as a test helper function
+	t.Helper()
+	if d := cmp.Diff(want, got, opts...); d != "" {
+		t.Errorf(format, PrintWantGot(d))
```
sbwsg, line 31: Here's a test that prints a diff currently in our codebase:

```go
for _, tt := range tests {
        t.Run(tt.name, func(t *testing.T) {
                got := resources.ApplyParameters(tt.args.ts, tt.args.tr, tt.args.dp...)
                if d := cmp.Diff(tt.want, got); d != "" {
                        t.Errorf("ApplyParameters() got diff %s", diff.PrintWantGot(d))
                }
        })
}
```

And here's what it would look like with this change:

```go
for _, tt := range tests {
        t.Run(tt.name, func(t *testing.T) {
                 got := resources.ApplyParameters(tt.args.ts, tt.args.tr, tt.args.dp...)
                 diff.ErrorIfWantNotGot(t, tt.want, got, "ApplyParameters() got diff %s")
        })
}
```

Notice that I must still include `"%s"`, otherwise the diff won't be printed.  I think it might be less surprising if the string didn't have to include `%s`?  So I could write:

```go
diff.ErrorIfWantNotGot(t, tt.want, got, "ApplyParameters() diff")
```

My suggestion is to accept a string message but don't use it for formatting.  Instead update this line to:

```go
t.Errorf(fmt.Sprintf("%s %s", message, PrintWantGot(d))
```

Same for `FatalIfWantNotGot`.  What do you think of this idea?

test/diff/helper.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"github.com/google/go-cmp/cmp"
+	"testing"
+)
+
+// ErrorIfWantNotGot Wrap cmp.Diff with a t.Helper(). It takes the testing.T
+// want,got, format message and cmp.option as an argument. Therefore, if diff was found
+// it will call t.Error of format and diff that wrapped by PrintWantGot func
+func ErrorIfWantNotGot(t *testing.T, want interface{}, got interface{}, format string, opts ...cmp.Option) {
+	// t.Helper() used to marks the calling function as a test helper function
```
sbwsg, line 28: I don't think this comment is needed; a developer can quickly check the golang docs.

test/diff/helper.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"github.com/google/go-cmp/cmp"
+	"testing"
+)
+
+// ErrorIfWantNotGot Wrap cmp.Diff with a t.Helper(). It takes the testing.T
+// want,got, format message and cmp.option as an argument. Therefore, if diff was found
+// it will call t.Error of format and diff that wrapped by PrintWantGot func
+func ErrorIfWantNotGot(t *testing.T, want interface{}, got interface{}, format string, opts ...cmp.Option) {
```
afrittoli, line 27: Thanks for this. May I suggest different names for the helpers?
What do you think about `diff.ErrorOnDiff` and `diff.FatalOnDiff` ? @sbwsg @vdemeester

NikeNano, line 27: Maybe it would be easier to just have one function and allow the user fo pass in the `t.Fatalf`or `t.Error`to the function. I think you could pass it in like this: 
```
FailIfWantGot( t *testing.T, want, got interface{}, format string, errorFunc func(string, ...interface{}), opts ...cmp.Option)
```

test/diff/helper.go
afrittoli: What about:
```
func ErrorOnDiff(t *testing.T, name string, want interface{}, got interface{}, opts ...cmp.Option)
```

and 

```
diff.ErrorOnDiff(t, "ApplyParameters()", tt.want, got)
```

afrittoli:
Thank you for your patch! 

Could you remove the merge commit from the PR?
NIT: since you're at it, it would be nice to wrap the text in the commit message, as per guidelines (https://chris.beams.io/posts/git-commit/#wrap-72).

test/diff/helper.go
2pai: Thanks for the suggestion @NikeNano  😁
I have some questions, is that okay to take many arguments? 
since testing.T is getting passed we can call it directly, but we have to separate the func.

ImJasonH:
Are we planning to take any steps to prevent raw `cmp.Diff` from being called in new tests? Without some check in place, we'll inevitably end up with `cmp.Diffs` sprouting back up in the future, possibly formatted incorrectly, and this change will just give us one more way to be inconsistent.

While we're at it, perhaps we could also use some kind of check that callers aren't passing `ErrorIfWantNotGot(got,want)` -- possibly with a simple grep that we don't find `WantNotGot.*got.*want`?

test/diff/helper.go
sbwsg: I think it would be nice if the name of the func made explicit the order of the params.  Want/Got, just like we have diff.PrintWantGot.  I'm not too fussed about what the overall names are but I do think it's worth preserving that pattern.

2pai: @sbwsg thanks for correcting me 
the explanation is very helpful
I've already updated the code

test/diff/helper.go
NikeNano: What d you mean by ` take many arguments`?

vdemeester:
> Are we planning to take any steps to prevent raw `cmp.Diff` from being called in new tests? Without some check in place, we'll inevitably end up with `cmp.Diffs` sprouting back up in the future, possibly formatted incorrectly, and this change will just give us one more way to be inconsistent.
> 
> While we're at it, perhaps we could also use some kind of check that callers aren't passing `ErrorIfWantNotGot(got,want)` -- possibly with a simple grep that we don't find `WantNotGot.*got.*want`?

Agreed :angel: That said, this PR only adds the function, does not yet migrate the calls, so we should do it as a follow-up.

2pai:
is my helper function was approved? if it's ok I will continue to 
Replace varying cmp.diff by using FatalIfWantNotGot & ErrorIfWantNotGot

sbwsg:
There is still comment feedback from other reviewers, it would be good to respond to any remaining open feedback to let people know what your decisions are, even if you disagree or do not know the answers.

It also looks like there are `gofmt` errors in [the build log](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/2702/pull-tekton-pipeline-build-tests/1266675652531064833/):

```
diff -u test/diff/helper.go.orig test/diff/helper.go
--- test/diff/helper.go.orig	2020-05-30 10:20:02.094970747 +0000
+++ test/diff/helper.go	2020-05-30 10:20:02.095970818 +0000
@@ -28,7 +28,7 @@
 func ErrorIfWantNotGot(t *testing.T, want interface{}, got interface{}, format string, opts ...cmp.Option) {
 	t.Helper()
 	if d := cmp.Diff(want, got, opts...); d != "" {
-		t.Errorf(fmt.Sprintf("%s %s",format, PrintWantGot(d)))
+		t.Errorf(fmt.Sprintf("%s %s", format, PrintWantGot(d)))
 	}
 }
 
@@ -38,6 +38,6 @@
 func FatalIfWantNotGot(t *testing.T, want interface{}, got interface{}, format string, opts ...cmp.Option) {
 	t.Helper()
 	if d := cmp.Diff(want, got, opts...); d != "" {
-		t.Fatalf(fmt.Sprintf("%s %s",format, PrintWantGot(d)))
+		t.Fatalf(fmt.Sprintf("%s %s", format, PrintWantGot(d)))
 	}
 }
```

You can use `make` to run gofmt against all files in the repo:

```bash
# from root of your git checkout
make fmt
```

test/diff/helper.go
sbwsg: @NikeNano I'm not sure what you mean by "it would be easier" here.  Could you maybe say why your approach is easier?  I'm not sure there's so much difference that either solution here is clearly better but I'm interested to hear your view on this.

test/diff/helper.go
```diff
@@ -0,0 +1,43 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package diff
+
+import (
+	"fmt"
+	"github.com/google/go-cmp/cmp"
```
sbwsg, line 21: Move this import down to its own section to match import style with other files in pipeline.  e.g.

```go
import (
  "fmt"
  "testing"

  "github.com/google/go-cmp/cmp"
)
```

test/diff/helper.go
2pai: okay, it's already removed 😄

test/diff/helper.go
2pai: so it has to be alphabetical order?

I already fix this.
but I doubt the change that I made, I'm looking forward to the review

sbwsg: > so it has to be alphabetical order?

We split the lists of packages into built in packages provided by go and those that are external libraries.  Thanks for fixing this!

test/diff/helper.go
sbwsg: I suggest renaming to `diff.ErrorWantGot` and `diff.FatalWantGot`.  This makes them shorter, removing the extra words "If" and "Not", and keeps the order of `WantGot` to help prompt the order of args.  Is that OK @afrittoli and @2pai ?

2pai: @NikeNano 
what I think is because this is a helper function it should make the function easier to use. and one of the approaches is to reduce the argument of the function. since testing.T is getting passed we can call it directly.

2pai: Sounds good, I agree with that @sbwsg

jerop:
@2pai do you plan to continue working on this? is there something we can do to move it along?

2pai:
hi @jerop , sorry for late fixing the PR. 
so I got an issue, I did the wrong way to implement the function that I've made. 
 
`diff.ErrorWantGot(t, tc.expectedTaskSpec, *ts, "Didn't get expected TaskSpec modifications %s")`

but in the helper function that I write is 
`t.Errorf(fmt.Sprintf("%s %s", format, PrintWantGot(d)))`

so I think I have to fix the implementation by replacing into 
`diff.ErrorWantGot(t, tc.expectedTaskSpec, *ts, "Didn't get expected TaskSpec modifications ")`

I already planned to rewrite it in this month, is that okay?

jerop:
@2pai sounds great, was just checking in 😊

sbwsg:
I'm going to close this PR as the associated issue has gone stale.  I'm not against reopening it but we should probably discuss whether we actually want what's being proposed first.

Issue #2513, opened 2020-04-30
Allow PipelineRuns to provide extra parameters


bobcatfish:
# Expected Behavior

Imagine you have a system creating PipelineRuns automatically, and each PipelineRun uses a different Pipeline which has a different set of parameters. The system has a number of parameters it knows how to provide, e.g. the service account and url of a bucket the Pipelines can push to (but not all Pipelines need to push to a bucket).

It would be very convenient if the system creating the PipelineRuns could provide all the parameters it has available to all PipelineRuns, and the Pipelines could use the ones they need and ignore the rest.

# Actual Behavior

If a PipelineRun provides extra params to a Pipeline, it will fail (99% sure!). This means that the system creating the PipelineRuns described above will need to look at each Pipeline, see what params it needs, and only provide the correct params. This feels like a level of complication that doesn't add much.

# Additional Info

* You can use TriggerBindings + TriggerTemplates to accomplish this today; TriggerBindings can provide params that TriggerTemplates do not use, but you have to involve an EventListener to make it work
* The specific use case which is motivating me to suggest this is designing the test infrastructure fo the Tekton catalog (some details [in this design doc](https://docs.google.com/document/d/1pZY7ROLuW47ymZYqUgAbxskmirrmDg2dd8VPATYXrxI/edit#heading=h.e9bh8qes6kh6)). The proposal is using TriggerBindings + TriggerTemplates because they allow this functionality, but beyond that there is no reason to use Triggers for this and it would be nice to simplify the design by removing that component.
* This might be an incremental improvement of the situation described in https://github.com/tektoncd/pipeline/issues/1294

bobcatfish:
Oh I think this is actually completed! Turns out there was NO validation at all XD in #2708  @jerop improved this so that we now check that AT LEAST the required params are provided, but the current state should be that you can provide extra

/assign jerop

jerop:
Similarly, @bobcatfish does it make sense to allow TaskRuns to provide extra parameters? 
We currently [disallow TasksRuns from providing extra parameters](https://github.com/tektoncd/pipeline/blob/4e99cb84ed624f6d48d9ff28249cda5364cd3358/pkg/reconciler/taskrun/validate_resources.go#L91-L94) -- can open a separate PR for it

bobcatfish:
> We currently disallow TasksRuns from providing extra parameters -- can open a separate PR for it

[...]

PR #2709, opened 2020-05-29
Validate PipelineRun parameters

### LINKED ISSUE #2708: Validate PipelineRun parameters
# Expected Behavior
There's validation that a PipelineRun has provided all the parameters expected by the Pipeline. 

# Actual Behavior
There's no validation that a PipelineRun has provided all the parameters expected by the Pipeline. 
When you run a Pipeline created by a PipelineRun that hasn't provided all the required parameters, you'll find that it mistakenly succeeds and you'll have to look into the logs to find that there was an issue.

1. Create a Pipeline with some params and a corresponding PipelineRun that doesn't provide one of the parameters. 
2. Run the pipeline, and check the results. 

# Additional Info
- While working on allowing PipelineRuns to provide extra parameters (#2513), we discovered that providing extra parameters was unintentionally allowed because, currently, there's no validation that all parameters expected 
by the Pipeline is provided by the PipelineRun. 
- When we add validation, we want allow PipelineRuns to provide extra parameters as described in (#2513).

jerop:
# Changes
While working on allowing PipelineRuns to provide extra parameters (#2513), we found that providing extra parameters was unintentionally allowed.That's because, currently, there's no validation that all parameters expected by the Pipeline is provided by the PipelineRun.

In this PR, we add validation for PipelineRun parameters by generating a list of provided parameters then iterating through the expected parameters to ensure they are in the list of provided parameters.

In the validation, we still allow PipelineRuns to provide extra parameters. If we disallow PipelineRuns from provides extra parameters, the Pipelines would fail. As a result, systems that autogenerate PipelineRuns need to look at each pipeline to see what parameters they need so it can provide only the required parameters. That means users would have to resort to more complex designs to solve this issue, as further described in (#2513). By allowing PipelineRuns to provide extra parameters, we make the process of autogenerating of PipelineRuns simpler.

Fixes #2708.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```
- PipelineRun parameters are validated to ensure that all the parameters expected by the Pipeline are provided by the PipelineRun.

- PipelineRun parameters validation allows PipelineRuns to provide extra parameters in addition to the expected parameters.
```


bobcatfish:
Great commit message and release notes @jerop !! :D 🎉

docs/pipelineruns.md
```diff
@@ -185,6 +185,10 @@ to all `persistentVolumeClaims` generated internally.
 You can specify `Parameters` that you want to pass to the `Pipeline` during execution,
 including different values of the same parameter for different `Tasks` in the `Pipeline`.
 
+You must specify all the `Parameters` that the `Pipeline` expects. 
+
+You can pass in extra `Parameters` if needed depending on you use case.
```
bobcatfish, line 190: typo: your use case

(you might want to elaborate on "your use case" here, e.g. give an example where a CI system sets up some infrastructure and wants to pass all the configuration values to all pipelineruns, or use #2513 for inspiration)

examples/v1beta1/pipelineruns/pipelinerun-with-extra-params.yaml
```diff
@@ -0,0 +1,57 @@
...
+kind: PipelineRun
+metadata:
+  generateName: pipelinerun-with-extra-params
+spec:
+  params:
+  - name: pl-param-x
+    value: "100"
+  - name: pl-param-y
+    value: "200"
+  - name: pl-param-z
```
bobcatfish, line 54: probably worth adding a comment to explain this is an extra param

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -80,6 +80,9 @@ const (
 	// ReasonCouldntGetCondition indicates that the reason for the failure status is that the
 	// associated Pipeline's Conditions couldn't all be retrieved
 	ReasonCouldntGetCondition = "CouldntGetCondition"
+	// ReasonCouldntGetParams indicates that the reason for the failure status is that the
+	// associated PipelineRun didn't provide all the required parameters
+	ReasonCouldntGetParams = "CouldntGetParams"
```
bobcatfish, line 85: i would call this something like "MissingParams" - "CouldntGet" is being used above when doing a GET on a resource is failing.

taskruns seem to use 'validation failed' for this case https://github.com/tektoncd/pipeline/blob/58c836d55ef750cd023bae901fec51389a97f318/pkg/reconciler/taskrun/taskrun.go#L300

Above there is a ParameterTypeMismatch reason as well, you could use that as a precedent and make it ParameterMissing

pkg/reconciler/pipelinerun/resources/validate_params_test.go
```diff
@@ -121,3 +121,83 @@ func TestValidateParamTypesMatching_Invalid(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateExpectedParametersProvided_Valid(t *testing.T) {
+	tcs := []struct {
+		name          string
+		p             *v1beta1.Pipeline
+		pr            *v1beta1.PipelineRun
+		errorExpected bool
```
bobcatfish, line 130: i highly recommend that instead of using a flag like this and mixing valid and invalid test cases, instead make two completely different tests, one with all the valid cases and one with all the invalid cases

its way easier to maintain over time and understand whats covered vs whats not, and what's valid vs what's not. it also avoids introducing bugs in the test itself around handling the flag

pkg/reconciler/pipelinerun/resources/validate_params_test.go
```diff
@@ -121,3 +121,83 @@ func TestValidateParamTypesMatching_Invalid(t *testing.T) {
...
+			tb.PipelineParamSpec("expected-array-param", v1beta1.ParamTypeArray))),
+		pr: tb.PipelineRun("a-pipelinerun", tb.PipelineRunSpec("test-pipeline",
+			tb.PipelineRunParam("expected-string-param", "somestring"))),
+		errorExpected: true,
+	}}
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			err := ValidateExpectedParametersProvided(&tc.p.Spec, tc.pr)
+			if (!tc.errorExpected) && (err != nil) {
+				t.Errorf("Pipeline.Validate() returned error: %v", err)
```
bobcatfish, line 166: i think it's helpful to add a bit more context here so if this fails the person running the tests has more info about whats going on, e.g. something like "didnt expect valid pipelinerun to return error but got %v"

pkg/reconciler/pipelinerun/resources/validate_params_test.go
bobcatfish: it actually looks like you did that down below already so you can remove the flag from this test probably :D (as long as the failure case in this test is covered in the other test also)

pkg/reconciler/pipelinerun/resources/validate_params.go
```diff
@@ -46,3 +47,26 @@ func ValidateParamTypesMatching(p *v1beta1.PipelineSpec, pr *v1beta1.PipelineRun
 	}
 	return nil
 }
+
+// ValidateExpectedParametersProvided validate that all the parameters expected by the Pipeline are provided by the PipelineRun.
```
bobcatfish, line 51: typo "validates that"

pkg/reconciler/pipelinerun/resources/validate_params.go
```diff
@@ -46,3 +47,26 @@ func ValidateParamTypesMatching(p *v1beta1.PipelineSpec, pr *v1beta1.PipelineRun
...
+		providedParams[param.Name] = param.Value.Type
+	}
+
+	// Build a list of parameter names in p that are missing from pr.
+	var missingParams []string
+	for _, param := range p.Params {
+		if _, ok := providedParams[param.Name]; !ok {
+			missingParams = append(missingParams, param.Name)
+		}
+	}
```
bobcatfish, line 65: you might be able to use DiffLeft from https://github.com/tektoncd/pipeline/blob/master/pkg/list/diff.go

pkg/reconciler/pipelinerun/resources/validate_params.go
bobcatfish: probably worth including in this comment that extra parameters are being allowed

pkg/reconciler/pipelinerun/resources/validate_params.go
```diff
@@ -24,17 +24,18 @@ import (
...
-	// Build a list of parameter names from pr that have mismatching types with the map created above.
-	var wrongTypeParamNames []string
-	for _, param := range pr.Spec.Params {
-		if paramType, ok := paramTypes[param.Name]; ok {
-			if param.Value.Type != paramType {
+	// Build a list of parameter names from p that have mismatching types from corresponding types from p.
+	var wrongTypeParamNames []string 
+	for _, param := range p.Params {
+		paramType, ok := providedParams[param.Name]; 
+		if ok {
```
bobcatfish, line 37: `if paramType, ok := paramTypes[param.Name]; ok` is the syntax we usually prefer whenever we can get away with it

pkg/reconciler/pipelinerun/resources/validate_params_test.go
bobcatfish: your error in the next test is much more descriptive!

pkg/reconciler/pipelinerun/resources/validate_params_test.go
```diff
@@ -121,3 +121,83 @@ func TestValidateParamTypesMatching_Invalid(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func TestValidateExpectedParametersProvided_Invalid(t *testing.T) {
+	tcs := []struct {
+		name string
+		p    *v1beta1.Pipeline
+		pr   *v1beta1.PipelineRun
```
bobcatfish, line 180: i think unit tests are a great tool to use to guide writing cleaner code and i think there is an interesting signal in this test: to setup the test, you have to create entire pipelines and pipelienruns, though all you really care about is the params.

it would be an easier test to read and maintain if you only had to have lists of the params required and the params provided.

what would it take to make it so that ValidateExpectedParametersProvided only takes the lists of params as arguments, instead of the entire pipeline and pipelinerun?

pkg/reconciler/pipelinerun/resources/validate_params.go
jerop: So useful, thanks!

PR #2713, opened 2020-05-29
Change the sha for the nop image

afrittoli:
# Changes

The sha for the nop image started returning 500 errors.
Pulling the image via the tag "latest" works and returns a different
sha, so updating our config to use the working one.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Changed the sha of the nop image since it changed on dockerhub side, but no actual change of content in the image.
```

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -67,7 +67,8 @@ spec:
           "-build-gcs-fetcher-image", "github.com/tektoncd/pipeline/vendor/github.com/GoogleCloudPlatform/cloud-builders/gcs-fetcher/cmd/gcs-fetcher",
 
           # These images are pulled from Dockerhub, by digest, as of May 19, 2020.
-          "-nop-image", "tianon/true@sha256:183cb5fd54142948ad88cc986a217465cfe8786cfdd89b1ed1fc49825da413a7",
+          # As of May 29, 2020 new sha for nop image
+          "-nop-image", "tianon/true@sha256:009cce421096698832595ce039aa13fa44327d96beedb84282a69d3dbcf5a81b",
           # This is google/cloud-sdk:293.0.0-slim
           "-gsutil-image", "google/cloud-sdk@sha256:37654ada9b7afbc32828b537030e85de672a9dd468ac5c92a36da1e203a98def",
           # The shell image must be root in order to create directories and copy files to PVCs.


vdemeester:
> Changed the sha of the nop image since it changed on dockerhub side, but no actual change of content in the image.

This is **really** weird (or bad…) as the digest should *never* change (nor disappear) in/from the registry :thinking:

sbwsg:
> Was it temporary ?

Yes, strangely we had the correct digest for tianon/true in 0.11 and 0.12.  We updated that on May 20th but then it seems that the digest for tianon/true was reverted back to its original value in docker hub or something else happened after that time?  Very odd.

vdemeester:
So it's a bit tricky, a image "by digest" that is not refered by a tag can disappear (garbage collected) at some point… My guess is `latest` moved (most likely back) and the image got garbage collected (*or* something happened on the hub and some images were reverted…). In any case, we should try to stick to non `latest` tag to get the digest :sweat_smile: (and in any case, the hub shouldn't error out 500 in that can but 404 or something..)

sbwsg:
I edited the issue here: https://github.com/tektoncd/pipeline/issues/2714 just to add that I think we should avoid external images for Tekton's core features.  It might not be a sustainable approach?  In the abstract it seems better to me to avoid this whole problem altogether by keeping things "in-house".

config/controller.yaml
```diff
@@ -67,7 +67,8 @@ spec:
           "-build-gcs-fetcher-image", "github.com/tektoncd/pipeline/vendor/github.com/GoogleCloudPlatform/cloud-builders/gcs-fetcher/cmd/gcs-fetcher",
 
           # These images are pulled from Dockerhub, by digest, as of May 19, 2020.
-          "-nop-image", "tianon/true@sha256:183cb5fd54142948ad88cc986a217465cfe8786cfdd89b1ed1fc49825da413a7",
```
tianon, line 70: This is interesting to me -- this digest can't have worked previously, because this is the digest of the image config, not an image manifest (unless Docker somehow does something clever when given a config object digest? but I don't think it does):

```json
digest: sha256:009cce421096698832595ce039aa13fa44327d96beedb84282a69d3dbcf5a81b
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
   "config": {
      "mediaType": "application/vnd.docker.container.image.v1+json",
      "size": 1511,
      "digest": "sha256:183cb5fd54142948ad88cc986a217465cfe8786cfdd89b1ed1fc49825da413a7"
   },
   "layers": [
      {
         "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
         "size": 147,
         "digest": "sha256:c53fb220cbad89d19e819673ea42ffcb938319825ef36f1b96df37af3d4665f5"
      }
   ]
}
```

vdemeester: huh… interesting… for some reason I managed to download this image (I do not remember if it was using `docker`, `podman` or `skopeo` lol.. :man_facepalming:). But that explain why it fails, mystery solved I'ld say.. (I also may have messed up when updating it… I need to proper check how I did it…)

manishtomar: Thanks for the info @tianon. This explains it is not working but it should still return 404 instead 500. We will investigate internally depending on priority.

PR #2719, opened 2020-05-29
Validate PipelineRun Parameters

### LINKED ISSUE #2708: Validate PipelineRun parameters

jerop:
# Changes

While working on allowing PipelineRuns to provide extra parameters (#2513), we found that providing extra parameters was unintentionally allowed. That's because, currently, there's no validation that all parameters expected by the Pipeline is provided by the PipelineRun.

In this PR, we add validation for PipelineRun parameters by generating a list of provided parameters then iterating through the expected parameters to ensure they are in the list of provided parameters. Note that parameters which have default values specified in Pipeline are not required to be provided by PipelineRun.

In the validation, we still allow PipelineRuns to provide extra parameters. If we disallow PipelineRuns from providing extra parameters, the Pipelines would fail. As a result, systems that autogenerate PipelineRuns would need to look at each pipeline to see what parameters they need so it can provide only the required parameters. That means users would have to resort to more complex designs to solve this issue, as further described in (#2513). 

By allowing PipelineRuns to provide extra parameters, we make the process of autogenerating of PipelineRuns simpler.

Fixes #2708.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
- PipelineRun parameters are validated to ensure that all the parameters required by the Pipeline are provided by the PipelineRun.

- PipelineRun parameters validation allows PipelineRuns to provide extra parameters in addition to the required parameters.

- Warning: backwards incompatible change that will force users to pass in default parameter values or provide the required parameters. 
```

### DIFF
--- a/examples/v1beta1/pipelineruns/pipelinerun-with-extra-params.yaml
+++ b/examples/v1beta1/pipelineruns/pipelinerun-with-extra-params.yaml
@@ -1,0 +1,56 @@
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: pipeline-with-extra-params
+spec:
+  params:
+  - name: pl-param-x
+    type: string
+  - name: pl-param-y
+    type: string
+  tasks:
+  - name: add-params
+    taskRef:
+      name: add-params
+    params:
+    - name: a
+      value: "$(params.pl-param-x)"
+    - name: b
+      value: "$(params.pl-param-y)"
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: add-params
+  annotations:
+    description: |
+      A simple task that sums the two provided integers
+spec:
+  params:
+  - name: a
+    type: string
+    description: The first integer
+  - name: b
+    type: string
+    description: The second integer
+  steps:
+  - name: sum
+    image: bash:latest
+    script: |
+      #!/usr/bin/env bash
+      echo -n $(( "$(inputs.params.a)" + "$(inputs.params.b)" ))
+---
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: pipelinerun-with-extra-params
+spec:
+  params:
+  - name: pl-param-x
+    value: "100"
+  - name: pl-param-y
+    value: "200"
+  - name: pl-param-z # the extra parameter
+    value: "300"
+  pipelineRef:
+    name: pipeline-with-extra-params
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -80,6 +80,9 @@ const (
 	// ReasonCouldntGetCondition indicates that the reason for the failure status is that the
 	// associated Pipeline's Conditions couldn't all be retrieved
 	ReasonCouldntGetCondition = "CouldntGetCondition"
+	// ReasonParameterMissing indicates that the reason for the failure status is that the
+	// associated PipelineRun didn't provide all the required parameters
+	ReasonParameterMissing = "ParameterMissing"
 	// ReasonFailedValidation indicates that the reason for failure status is
 	// that pipelinerun failed runtime validation
 	ReasonFailedValidation = "PipelineValidationFailed"
@@ -313,6 +316,14 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 		pr.Status.MarkFailed(ReasonCouldntGetResource,
 			"PipelineRun %s/%s can't be Run; it tries to bind Resources that don't exist: %s",
 			pipelineMeta.Namespace, pr.Name, err)
+		return nil
+	}
+	// Ensure that the PipelineRun provides all the parameters required by the Pipeline
+	if err := resources.ValidateRequiredParametersProvided(&pipelineSpec.Params, &pr.Spec.Params); err != nil {
+		// This Run has failed, so we need to mark it as failed and stop reconciling it
+		pr.Status.MarkFailed(ReasonParameterMissing,
+			"PipelineRun %s parameters is missing some parameters required by Pipeline %s's parameters: %s",
+			pr.Namespace, pr.Name, err)
 		return nil
 	}
 
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -186,6 +186,9 @@ to all `persistentVolumeClaims` generated internally.
 You can specify `Parameters` that you want to pass to the `Pipeline` during execution,
 including different values of the same parameter for different `Tasks` in the `Pipeline`.
 
+**Note:** You must specify all the `Parameters` that the `Pipeline` expects. Parameters 
+that have default values specified in Pipeline are not required to be provided by PipelineRun.
+
 For example:
 
 ```yaml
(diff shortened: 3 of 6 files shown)


examples/v1beta1/pipelineruns/pipelinerun-with-extra-params.yaml
```diff
@@ -0,0 +1,57 @@
...
+  - name: add-params
+    taskRef:
+      name: add-params
+    params:
+    - name: a
+      value: "$(params.pl-param-x)"
+    - name: b
+      value: "$(params.pl-param-y)"
+---
+apiVersion: tekton.dev/v1alpha1
```
pritidesai, line 21: please change the apiVersion to `v1beta1` here

examples/v1beta1/pipelineruns/pipelinerun-with-extra-params.yaml
```diff
@@ -0,0 +1,57 @@
...
+      type: string
+      description: The second integer
+  steps:
+  - name: sum
+    image: bash:latest
+    script: |
+      #!/usr/bin/env bash
+      echo -n $(( "$(inputs.params.a)" + "$(inputs.params.b)" ))
+---
+apiVersion: tekton.dev/v1alpha1
```
pritidesai, line 44: same as above, please change the apiVersion to `v1beta1` here since we are adding an example under `v1beta1` directory

examples/v1beta1/pipelineruns/pipelinerun-with-extra-params.yaml
```diff
@@ -0,0 +1,57 @@
...
+kind: Task
+metadata:
+  name: add-params
+  annotations:
+    description: |
+      A simple task that sums the two provided integers
+spec:
+  inputs:
+    params:
```
pritidesai, line 30: `params` will go directly under `spec` with `v1beta1`

pritidesai:
@jerop its a great catch, we somehow missed validating extra and missing pipelineRun parameters.

I ran a few pipelines on master:

1. [with extra parameters](https://gist.github.com/pritidesai/744f6065b472679560ea05bded6398d2) - this is supported, not validated but unintentionally ignored like you mentioned.
2. [with missing parameters](https://gist.github.com/pritidesai/0c938e4034dbe9e441fee147f1d637cb) - this is caught at the taskRun level when the task is trying to evaluate the parameter, it fails as the value is not set. This PR is catching this early and saving from unnecessary pod failure. 
3. [with missing default parameters](https://gist.github.com/pritidesai/889749f0735def3504b0180b0931c86f) - default are used whenever params are not specified.

Thanks for this PR, please address my comments above (changing apiVersion) and possibly add a test case to include params with default values otherwise `/lgtm`.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -440,6 +443,20 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 		})
 		return nil
 	}
+	// Ensure that the PipelineRun provides all the parameters required by the Pipeline
+	err = resources.ValidateRequiredParametersProvided(&pipelineSpec.Params, &pr.Spec.Params)
+	if err != nil {
+		// This Run has failed, so we need to mark it as failed and stop reconciling it
+		pr.Status.SetCondition(&apis.Condition{
```
pritidesai, line 450: please use `pr.Status.MarkFailed(` instead of `pr.Status.SetCondition` which @mattmoor added in PR #2741, something similar to: 
https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/pipelinerun/pipelinerun.go#L382

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -80,6 +80,9 @@ const (
 	// ReasonCouldntGetCondition indicates that the reason for the failure status is that the
 	// associated Pipeline's Conditions couldn't all be retrieved
 	ReasonCouldntGetCondition = "CouldntGetCondition"
+	// ParameterMissing indicates that the reason for the failure status is that the
+	// associated PipelineRun didn't provide all the required parameters
+	ParameterMissing = "ParameterMissing"
```
pritidesai, line 85: This is one of the reasons pipeline validation might fail with, we have been naming such const to begin with `Reason` 😲   please change `ParameterMissing` to `ReasonParameterMissing` 🙏

examples/v1beta1/pipelineruns/pipelinerun-with-extra-params.yaml
vdemeester: ah I missed that :sweat_smile:

jerop:
thanks for the review @vdemeester and @pritidesai! 

made the changes and the test case with default params is [here](https://github.com/tektoncd/pipeline/pull/2719/files#diff-e7a6f13b7414edc573c935d1f3db54f8R158-R173)

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -315,6 +318,15 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 			pipelineMeta.Namespace, pr.Name, err)
 		return nil
 	}
+	// Ensure that the PipelineRun provides all the parameters required by the Pipeline
+	err = resources.ValidateRequiredParametersProvided(&pipelineSpec.Params, &pr.Spec.Params)
+	if err != nil {
```
bobcatfish, line 323: looks like you could do this in one line since there are no other vars involved

bobcatfish:
@vdemeester if anyone was relying on this for Pipelines + PipelineRuns we definitely want to warn them, so I think it makes sense to put a warning in the release notes.

I think the catalog shouldn't need any updates b/c this PR is only updating Pipelines + PipelinesRuns, and it looks like Tasks + TaskRuns already have this validation, for example this Task:

```
kind: Task
metadata:
  name: hello
spec:
  params:
  - name: extraOptions
    type: string
  steps:
    - name: hello
      image: ubuntu
      command:
        - echo
      args:
        - "Hello World $(params.extraOptions)"
```

With this TaskRun:

```
kind: TaskRun
metadata:
  generateName: hello-
spec:
  taskRef:
    name: hello
```

The TaskRun fails with:

```
invalid input params: missing values for these params which have no default values: [extraOptions]
```

vdemeester:
> @vdemeester if anyone was relying on this for Pipelines + PipelineRuns we definitely want to warn them, so I think it makes sense to put a warning in the release notes.
> 
> I think the catalog shouldn't need any updates b/c this PR is only updating Pipelines + PipelinesRuns, and it looks like Tasks + TaskRuns already have this validation, for example this Task:
> 
> ```
> apiVersion: tekton.dev/v1beta1
> kind: Task
> metadata:
>   name: hello
> spec:
>   params:
>   - name: extraOptions
>     type: string
>   steps:
>     - name: hello
>       image: ubuntu
>       command:
>         - echo
>       args:
>         - "Hello World $(params.extraOptions)"
> ```
> 
> With this TaskRun:
> 
> ```
> apiVersion: tekton.dev/v1beta1
> kind: TaskRun
> metadata:
>   generateName: hello-
> spec:
>   taskRef:
>     name: hello
> ```
> 
> The TaskRun fails with:
> 
> ```
> invalid input params: missing values for these params which have no default values: [extraOptions]
> ```

Ahh good then ! It's a less scaring change :upside_down_face: 
/hold cancel

PR #2712, opened 2020-05-29
Fix failure running ./hack/update-deps.sh --upgrade

mattmoor:
This is just the change to `go.mod` and `./hack/update-deps.sh --upgrade`.

/assign @ImJasonH @bobcatfish @vdemeester

PR #2721, opened 2020-05-30
Update knative.dev/pkg

mattmoor:
This pulls forward the deps to 0.15

WIP until https://github.com/tektoncd/pipeline/pull/2712 lands.


sbwsg:
Does this dep update introduce any ConfigMaps or Secrets that we need to account for as part of [Pipeline's RBAC](https://github.com/tektoncd/pipeline/blob/master/config/200-role.yaml)?

mattmoor:
ugh, I hate that git on OSX doesn't handle file renames involving case in a portable way.

I'll have to fix this manually and repush.

mattmoor:
ok, this is RFAL.  macOS's case insensitive filesystem caused issues with `shoulder.md` -> `SHOULDER.md` 🙄

mattmoor:
@ImJasonH This needs another LGTM 😞

afrittoli:
@mattmoor Is it worth adding release notes?

pkg/client/injection/reconciler/pipeline/v1beta1/pipelinerun/controller.go
```diff
@@ -0,0 +1,118 @@
+/*
+Copyright 2019 The Tekton Authors
```
afrittoli, line 2: Should the generated code have 2020 in here?

mattmoor: I can change this and regenerate, but this is where that comes from: https://github.com/tektoncd/pipeline/blob/master/hack/boilerplate/boilerplate.go.txt#L2

afrittoli:
Thanks for this @mattmoor !
If I understand this correctly, we generate the controllers now and but still not use them.

I've been looking at the new controllers from knative/pkg, I think most of it is a good fit for Tekton, but it will take a bit of work to integrate them. So I wonder if we should only generated them once we start using them?

pkg/client/injection/reconciler/pipeline/v1beta1/pipelinerun/controller.go
afrittoli: Oh, I see - time to update that indeed :)

mattmoor:
> Is it worth adding release notes?

For this alone, I don't think so.  I am looking at picking up some of the stuff we've landed (e.g. generated reconcilers), but that's going to take some doing.

> So I wonder if we should only generated them once we start using them?

I could go either way on this.  I opted for this because script changes were needed for it to work and I had to dig around to find this patch from the last time I sat down to do this (thanks to @n3wscott for the help back then).

I could ofc keep the script changes and drop the `// +genreconciler` for now, but I wanted to make sure it worked.

pkg/client/injection/reconciler/pipeline/v1beta1/pipelinerun/controller.go
mattmoor: I pushed another commit with the timestamp changed and codegen rerun

vdemeester:
> I could ofc keep the script changes and drop the `// +genreconciler` for now, but I wanted to make sure it worked.

We may want to do that, wait for 0.13 to be done / "branched out" (later this week) and add them back :angel:

mattmoor:
New commit with changed timestamp, starting a thread on the codegen stuff.

mattmoor:
@vdemeester rerunning codegen now 😇

mattmoor:
Ok, should be RFAL!

vdemeester:
/cc @sbwsg @bobcatfish 
It would be nice to have this one before 0.13 :stuck_out_tongue_closed_eyes:

afrittoli:
> Ok, should be RFAL!

RFAL?

pkg/client/injection/reconciler/pipeline/v1beta1/pipelinerun/controller.go
afrittoli: thanks!

mattmoor:
/retest
RFAL = Ready For A Look (sorry, I love my acronyms)

PR #2722, opened 2020-05-30
Remove a bunch of dead logic in the base.

mattmoor:
I noticed this starting to look at what needs to be done to move to genreconciler.

I'm guessing this is a derivative of Serving and fwiw we've been able to completely eliminate these "Base" constructs everywhere with `// +genreconciler` 🤩 

/assign @ImJasonH @vdemeester


vdemeester:
oh nice, thanks @mattmoor

PR #2733, opened 2020-06-02
Refactoring in sidecar test

piyush-garg:
It will now run as parallel and also will create a
namespace
Use simple busybox image

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Refactor in sidecar e2e
```

### DIFF
--- a/test/v1alpha1/sidecar_test.go
+++ b/test/v1alpha1/sidecar_test.go
@@ -26,6 +26,7 @@ import (
 	tb "github.com/tektoncd/pipeline/internal/builder/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	knativetest "knative.dev/pkg/test"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


test/v1alpha1/sidecar_test.go
```diff
@@ -62,13 +67,13 @@ func TestSidecarTaskSupport(t *testing.T) {
 			task := tb.Task(sidecarTaskName,
 				tb.TaskSpec(
 					tb.Step(
-						"busybox:1.31.0-musl",
```
sbwsg, line 65: Thanks for this! Couple questions:

1. Could you add something to the commit message explaining _why_ this is preferred?
2. What is this doing now?  Pulling in ":latest"?  If so, should we simply pin to today's "latest" version / digest so that it remains a constant going forward for our tests?  If not, why?

piyush-garg: 1. There is no such difference to prefer, the job required can be done by `busybox` image and we are using that in many e2e tests, while `busybox image based on musl` is used only in this test so it will avoid this image pulling. I just used a simple busybox image as for the e2e test, it is just doing `echo` and doesn't have any scenario to use the alpine based image.

2. If you see all the e2e tests, in every test except one it is just`busybox` so I used it here also the same way.

sbwsg: Fantastic, thanks for explaining.  I didn't know that this was what we use everywhere else so it's good to know.

sbwsg: It would be great to get this in the commit message too.  Even when something is extremely obvious it really helps to have the context of the change described.  Otherwise I need to just "somehow" know.

Issue #2734, opened 2020-06-03
Tekton modifies objects in the informer cache


mattmoor:
# Expected Behavior

Tekton should `.DeepCopy()` any resource that is fetched from an informer's Lister, otherwise the resource in the underlying cache is altered and bad things result (TBH, I'm not even sure of the extent, but this is a "no no").

# Actual Behavior

In a few places Tekton updates the resource returned from the informer cache.

# Steps to Reproduce the Problem

I outlined this a bit here: https://github.com/tektoncd/pipeline/issues/2729#issuecomment-637673533, and also highlighted a few of the problem spots.

mattmoor:
/kind bug

I'm currently looking at this in spare cycles, but would be open to handing off to someone with more bandwidth to run down the issues this has dislodged.

mattmoor:
One complication here is that the single control loop is performing multiple Lister reads -> Client writes, but given that the fakes aren't updating the informers index during the client writes the second Lister read -> Client write reads a stale version of the object from the Lister cache.

mattmoor:
One partial solution to the above (since the same will happen during normal reconciliation) is to use a Patch like this, which bypasses the stale read and optimistic concurrency checks to update the labels and annotations:

```
-               newPr.ObjectMeta.Labels = pr.ObjectMeta.Labels
-               newPr.ObjectMeta.Annotations = pr.ObjectMeta.Annotations
-               return c.PipelineClientSet.TektonV1beta1().PipelineRuns(pr.Namespace).Update(newPr)
+               mergePatch := map[string]interface{}{
+                       "metadata": map[string]interface{}{
+                               "labels":      pr.ObjectMeta.Labels,
+                               "annotations": pr.ObjectMeta.Annotations,
+                       },
+               }
+               patch, err := json.Marshal(mergePatch)
+               if err != nil {
+                       return nil, err
+               }
+               return c.PipelineClientSet.TektonV1beta1().PipelineRuns(pr.Namespace).Patch(pr.Name, types.MergePatchType, patch)
```

I'm working through some other test failures as well, at least one of which is a test bug that was masked by this bug.

Issue #2729, opened 2020-06-02
RunningPipelineRun metric is racy


mattmoor:
I was looking at the PipelineRun controller, and noticed that there is a metric for the number of running PipelineRuns:

https://github.com/tektoncd/pipeline/blob/baf3c725c0f0e4409d935a6bee5c8f3a22695056/pkg/reconciler/pipelinerun/pipelinerun.go#L254-L261

This metric is updated when one of the reconciler's trailing updates modifies the resource (status or labels/annotations), but it bases its metric on the informer's lister cache.  Now, informers are fast, but in all likelihood the lister that is being used to drive this metric hasn't received the updated version of the resource from the API Server, so unless the informer cache's copy is being modified (a no no) then the cache won't reflect what it's clearly (given the `if updated` check) trying to capture.

Perhaps this metric could even be completely decoupled from this reconciler?  It seems like fairly separable logic that's needlessly coupled to reconciliation.  🤷

mattmoor:
I'd probably say `/kind bug`? 😇

vdemeester:
@mattmoor for my defense, I was on PTO, my brain is not completely wired back correctly it seems :joy_cat: 
/kind bug
/unkind feature

mattmoor:
> unless the informer cache's copy is being modified (a no no) 

I am seeing the informer's copy being modified in both `updateStatus` and `updateLabelsAndAnnotations`.

mattmoor:
This modifies a resource fetched from a lister (informer cache): https://github.com/tektoncd/pipeline/blob/be9ac3bf6b27f6beb3f4b7c828ef8388f9b323c3/pkg/reconciler/pipelinerun/pipelinerun.go#L916

mattmoor:
You can watch things explode by adding a line like `pr := pr.DeepCopy()` to the top of each of these loops:
https://github.com/tektoncd/pipeline/blob/be9ac3bf6b27f6beb3f4b7c828ef8388f9b323c3/test/controller.go#L108

mattmoor:
Alright, so https://github.com/tektoncd/pipeline/pull/2736 should address the modifications of the resources in the informer cache, but I didn't touch this code because no tests blew up (yet 🤞 ) and I feel like fixing this race needs a deeper design conversation.

Not sure how folks want to proceed, but that PR will almost certainly make this metric flakier until this issue is fixed.

mattmoor:
[...]

PR #2736, opened 2020-06-03
Avoid modifications to the informer's copy of resources.

mattmoor:
In each of the `{Pipeline,Task}Run` reconcilers the functions to update status and labels/annotations refetch the resource from the informer cache, check the field they want to update, and if an update is needed they set the field on the informer's copy and call the appropriate update method.

In pseudo-code:

```go
func update(fr *FooRun) {
  newFr := lister.Get(fr.Name)

  if reflect.DeepEqual(newFr.Field, fr.Field) {
    newFr.Field = fr.Field   // This modified the informer's copy!
    return client.Update(newFr)
  }
}
```

# Changes
I have worked around this in two different ways:

1. For the status updates I added a line like `newFr = newFr.DeepCopy()` immediately above the mutation to avoid writing to the informer's copy.

2. For the label/annotation updates, I changed the `Update` call to a `Patch` that bypasses optimistic concurrency checks.  This last bit is important because otherwise the update above will lead to the first reconciliation *always* failing due to `resourceVersion` skew caused by the status update.  This also works around some fun interactions with the test code (see fixed issue).

There are two other notable aspects to this change:

1. Test bugs! There were a good number of places that were assuming that the object stored in the informer was altered.  I changed most of these to refetch through the client.
2. D-Fence! I added some logic to some of the common test setup code to `DeepCopy()` resources before feeding them to the fake clients to try and avoid assumptions about "same object" creeping back in.

It is also worth calling out that this change will very likely destabilize the metric that I identified [here](https://github.com/tektoncd/pipeline/issues/2729) as racy, which is likely masked by the mutation of the informer copies.

Fixes: https://github.com/tektoncd/pipeline/issues/2734

# Submitter Checklist

review them:

- [ x ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ n/a ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [ x ] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

## Reviewer Notes

# Release Notes

N/A

/hold

I'm throwing a hold on this due to the lingering `t.Skip` / `DO NOT SUBMIT`, which I am asking about in slack, but I wanted to fling this at CI to see what falls out.

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -18,6 +18,7 @@ package pipelinerun
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"path/filepath"
 	"reflect"
@@ -49,6 +50,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/configmap"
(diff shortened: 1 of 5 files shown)


afrittoli:
This is great, thank you!
I started looking into it, it's not clear to me yet why the `TestReconcileWithPipelineResults` - whether it's a problem of the test or if it's a problem of the reconciler code.
If the former I think it might be ok to merge this, and go fix the test afterwards.

afrittoli:
Once we start fetching specs from OCI containers, we'll need an equivalent of the informers cache for that too!

PR #2743, opened 2020-06-03
Propagate NodeSelector and Tolerations to Affinity Assistant

### LINKED ISSUE #2679: Add option to specify PodTemplate for Affinity Assistant
From https://github.com/tektoncd/pipeline/pull/2630#issuecomment-629558587

# Expected Behavior

That Tekton is a flexible system that can be configured for custom needs.

# Actual Behavior

The Affinity Assistant is a static `StatefulSet` that can not be configured.

# Use cases

* For using Tekton only on a subset of Nodes, a `NodeSelector` can be set. This must now be set on the Affinity Assistant instead of on TaskRun pods

* When a platform has other requirements on specifying `ResourceRequests`/`Limits`. This should be set in the PodTemplate for the Affinity Assistant.

/kind feature

### LINKED ISSUE #2742: Support using Tekton on a subset of Nodes when Affinity Assistant is enabled
# Expected Behavior

That I can run Tekton workload only on a subset of my Nodes, e.g. on my build-nodes.

# Actual Behavior

This can be done by setting a [NodeSelector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) or by using [Tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) in the [default PodTemplate](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#customizing-basic-execution-parameters) or per `PipelineRun`  podTemplate.

But when using Affinity Assistant, there is no way to set those properties on the Affinity Assistant Pod - so e.g. my workload can not longer be kept to a subset of my nodes.

/kind feature

jlpettersson:
# Changes

To be able to run Tekton workload on a subset of the Nodes in a cluster,
a NodeSelector or Tolerations can be set on the pods, e.g. by specifying a
custom PodTemplate.

When using the Affinity Assistant, this was not supported to customize.

This commit, also propagate NodeSelector and Tolerations to the Affinity Assistant pod,
so that Tekton workload can be kept to a subset of the Nodes in a cluster, also when
using the Affinity Assistant.

My original idea https://github.com/tektoncd/pipeline/pull/2630#issuecomment-629558587 was that it could be useful to specify a specific PodTemplate for the AffinityAssistant.
But I think it is better to inherit some values from the already specified custom PodTemplate.

Fixes #2742
Fixes #2679
/kind feature

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -245,7 +245,7 @@ The example below customizes the following:
 - the default timeout from 60 minutes to 20 minutes.
 - the default `app.kubernetes.io/managed-by` label is applied to all Pods created to execute `TaskRuns`.
 - the default Pod template to include a node selector to select the node where the Pod will be scheduled by default.
-  For more information, see [`PodTemplate` in `TaskRuns`](./taskruns.md#pod-template) or [`PodTemplate` in `PipelineRuns`](./pipelineruns.md#pod-template).
+  For more information, see [`PodTemplate` in `TaskRuns`](./taskruns.md#specifying-a-pod-template) or [`PodTemplate` in `PipelineRuns`](./pipelineruns.md#specifying-a-pod-template).
 
 ```yaml
 apiVersion: v1
@@ -270,7 +270,7 @@ To customize the behavior of the Pipelines Controller, modify the ConfigMap `fea
 
 - `disable-affinity-assistant` - set this flag to disable the [Affinity Assistant](./workspaces.md#affinity-assistant-and-specifying-workspace-order-in-a-pipeline)
   that is used to provide Node Affinity for `TaskRun` pods that share workspace volume. 
-  The Affinity Assistant pods may be incompatible with NodeSelector and other affinity rules
+  The Affinity Assistant is incompatible with other affinity rules
   configured for `TaskRun` pods.
 
   **Note:** Affinity Assistant use [Inter-pod affinity and anti-affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity)
--- a/docs/workspaces.md
+++ b/docs/workspaces.md
@@ -209,13 +209,15 @@ The `subPath` specified in a `Pipeline` will be appended to any `subPath` specif
 
 Sharing a `Workspace` between `Tasks` requires you to define the order in which those `Tasks`
 write to or read from that `Workspace`. Use the `runAfter` field in your `Pipeline` definition
-to define when a `Task` should be executed. For more information, see the [`runAfter` documentation](pipelines.md#runAfter).
+to define when a `Task` should be executed. For more information, see the [`runAfter` documentation](pipelines.md#using-the-runafter-parameter).
 
 When a `PersistentVolumeClaim` is used as volume source for a `Workspace` in a `PipelineRun`,
 an Affinity Assistant will be created. The Affinity Assistant acts as a placeholder for `TaskRun` pods
 sharing the same `Workspace`. All `TaskRun` pods within the `PipelineRun` that share the `Workspace`
 will be scheduled to the same Node as the Affinity Assistant pod. This means that Affinity Assistant is incompatible
-with e.g. NodeSelectors or other affinity rules configured for the `TaskRun` pods. The Affinity Assistant
+with e.g. other affinity rules configured for the `TaskRun` pods. If the `PipepineRun` has a custom
+[PodTemplate](pipelineruns.md#specifying-a-pod-template) configured, the `NodeSelector` and `Tolerations` fields
+will also be set on the Affinity Assistant pod. The Affinity Assistant
 is deleted when the `PipelineRun` is completed. The Affinity Assistant can be disabled by setting the
 [disable-affinity-assistant](install.md#customizing-basic-execution-parameters) feature gate.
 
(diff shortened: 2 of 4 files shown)


docs/install.md
```diff
@@ -270,7 +270,7 @@ To customize the behavior of the Pipelines Controller, modify the ConfigMap `fea
 
 - `disable-affinity-assistant` - set this flag to disable the [Affinity Assistant](./workspaces.md#affinity-assistant-and-specifying-workspace-order-in-a-pipeline)
   that is used to provide Node Affinity for `TaskRun` pods that share workspace volume. 
-  The Affinity Assistant pods may be incompatible with NodeSelector and other affinity rules
+  The Affinity Assistant is incompatible with other affinity rules
   configured for `TaskRun` pods.
```
bobcatfish, line 274: would it make sense to document that specifically nodeselector and tolerations will be propogated from the podtemplate?

it feels like this is a significant enough feature that we might want to add a .md doc just to explain the affinity assistant and how to configure it, what do you think?

pkg/reconciler/pipelinerun/affinity_assistant_test.go
```diff
@@ -82,6 +82,55 @@ func TestCreateAndDeleteOfAffinityAssistant(t *testing.T) {
...
+	if len(stsWithTolerationsAndNodeSelector.Spec.Template.Spec.Tolerations) != 1 {
+		t.Errorf("expected Tolerations in the StatefulSet")
+	}
+
+	if len(stsWithoutTolerationsAndNodeSelector.Spec.Template.Spec.NodeSelector) != 0 {
+		t.Errorf("unexpected NodeSelector in the StatefulSet")
+	}
+
+	if len(stsWithoutTolerationsAndNodeSelector.Spec.Template.Spec.Tolerations) != 0 {
+		t.Errorf("unexpected Tolerations in the StatefulSet")
```
bobcatfish, line 130: it feels like this test include 2 test cases in one - is it possible to create two different Test functions, one for each case? (one for with tolerations, one for without)

jlpettersson: Sure, I can do that.

docs/install.md
jlpettersson: Yes, we have some [documentation, about what the Affinity Assistant](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#affinity-assistant-and-specifying-workspace-order-in-a-pipeline) **is**, and I can add docs about these two propagated fields there.

The only configuration the user can do currently, is to disable it. But we can also write more extensive docs about it. I can have a look about more extensive docs tomorrow.

jlpettersson: Thanks for the review, I added info about field propagation to the workspaces.md doc.

docs/workspaces.md
```diff
@@ -209,13 +209,15 @@ The `subPath` specified in a `Pipeline` will be appended to any `subPath` specif
...
-to define when a `Task` should be executed. For more information, see the [`runAfter` documentation](pipelines.md#runAfter).
+to define when a `Task` should be executed. For more information, see the [`runAfter` documentation](pipelines.md#using-the-runafter-parameter).
 
 When a `PersistentVolumeClaim` is used as volume source for a `Workspace` in a `PipelineRun`,
 an Affinity Assistant will be created. The Affinity Assistant acts as a placeholder for `TaskRun` pods
 sharing the same `Workspace`. All `TaskRun` pods within the `PipelineRun` that share the `Workspace`
 will be scheduled to the same Node as the Affinity Assistant pod. This means that Affinity Assistant is incompatible
-with e.g. NodeSelectors or other affinity rules configured for the `TaskRun` pods. The Affinity Assistant
+with e.g. other affinity rules configured for the `TaskRun` pods. If the `PipepineRun` has a custom
+[PodTemplate](pipelineruns.md#specifying-a-pod-template) configured, the `NodeSelector` and `Tolerations` fields
```
bobcatfish, line 219: thanks!

pkg/reconciler/pipelinerun/affinity_assistant_test.go
bobcatfish: my apologies but i have 1 more request around the tests! now that I see them separated, it feels to me like the original test was actually 4 different cases:

- node selector in pod template
- node selector not in pod template
- tolerations in pod template
- tolerations not in pod template

now that we have 2 test functions, we've had to duplicate the test data (with and without)

i suggest that we either:
1. have 1 test for prWithCustomPodTemplate only, and one test for prWithoutCustomPodTemplate. The with case can check both node selector and tolerations have progated, the without can check they havent (this will allow us to not need to duplicate the test input)
2. keep the separate tests the way they are but use table driven tests such that each of prWithCustomPodTemplate and prWithoutCustomPodTemplate are a separate test case (i.e. we end up with 4 tests total, 2 test functions) 

my preference is for (1)

apologies for nitpicking this test

jlpettersson: I agree. By not duplicating test data, it makes it more readable. Thanks, good feedback make me a better developer. The test is now updated.

pkg/reconciler/pipelinerun/affinity_assistant_test.go
```diff
@@ -82,6 +82,58 @@ func TestCreateAndDeleteOfAffinityAssistant(t *testing.T) {
...
+	}
+
+	if len(stsWithoutTolerationsAndNodeSelector.Spec.Template.Spec.NodeSelector) != 0 {
+		t.Errorf("unexpected NodeSelector in the StatefulSet")
+	}
+}
+
```
bobcatfish, line 136: 🎉  thanks for redoing these twice 🙏

Issue #2741, opened 2020-06-03
CompletionTime is not set if Succeeded is already True


mattmoor:
# Expected Behavior

CompletionTime is always updated to reflect the time the task/pipeline was observed to succeed.

# Actual Behavior

CompletionTime is only updated when something else in Status is being touched.  This is because `updateStatus` is guarded by an equivalence check, so if nothing else changes, then this doesn't get set.

# Steps to Reproduce the Problem

I saw this in a unit test where I set the condition but not CompletionTime, and was surprised it wasn't set.

# Additional Info

This is going to be blocking for a transition away from hand-rolled `updateStatus` to `// +genreconciler`.

/kind bug

afrittoli:
Interesting. 
I suppose we never want to have the case where we set the status "True" is set on the "Succeeded" Condition, and the `CompletionTime` is not set.
Would the `+genreconciler` require that separation?

mattmoor:
Well `updateStatus` is now generated, and doesn't do this 😅 

The way we manage this in Knative is to have higher-level functions for manipulating status conditions.  The equivalent here would be to have the function(s) that put Succeeded into a terminal state also update CompletionTime.

I haven't poked around to see how this is set throughout the reconcilers today, but would be happy to advice if someone wants to look 😇 

/good-first-issue

mattmoor:
I may take a look at this next.  It's all over and I think it'll clean things up a bit.

PR #2749, opened 2020-06-04
Use a helper for setting the Succeeded condition on PipelineRun.

mattmoor:
These helpers reduce a lot of the boilerplate and give us hooks
where we can eagerly set the CompletionTime field rather than waiting
for `updateStatus`.

Fixes: https://github.com/tektoncd/pipeline/issues/2741

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
@@ -227,6 +227,25 @@ func (pr *PipelineRunStatus) SetCondition(newCond *apis.Condition) {
 	}
 }
 
+// MarkSucceeded changes the Succeeded condition to True with the provided reason and message.
+func (pr *PipelineRunStatus) MarkSucceeded(reason, messageFormat string, messageA ...interface{}) {
+	pipelineRunCondSet.Manage(pr).MarkTrueWithReason(apis.ConditionSucceeded, reason, messageFormat, messageA...)
+	succeeded := pr.GetCondition(apis.ConditionSucceeded)
+	pr.CompletionTime = &succeeded.LastTransitionTime.Inner
+}
+
+// MarkFailed changes the Succeeded condition to False with the provided reason and message.
+func (pr *PipelineRunStatus) MarkFailed(reason, messageFormat string, messageA ...interface{}) {
+	pipelineRunCondSet.Manage(pr).MarkFalse(apis.ConditionSucceeded, reason, messageFormat, messageA...)
+	succeeded := pr.GetCondition(apis.ConditionSucceeded)
+	pr.CompletionTime = &succeeded.LastTransitionTime.Inner
+}
+
+// MarkRunning changes the Succeeded condition to Unknown with the provided reason and message.
+func (pr *PipelineRunStatus) MarkRunning(reason, messageFormat string, messageA ...interface{}) {
+	pipelineRunCondSet.Manage(pr).MarkUnknown(apis.ConditionSucceeded, reason, messageFormat, messageA...)
+}
+
 // MarkResourceNotConvertible adds a Warning-severity condition to the resource noting
 // that it cannot be converted to a higher version.
 func (pr *PipelineRunStatus) MarkResourceNotConvertible(err *CannotConvertError) {
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -279,26 +279,19 @@ func (c *Reconciler) updatePipelineResults(ctx context.Context, pr *v1beta1.Pipe
 			return
 		}
 		c.Logger.Errorf("Failed to determine Pipeline spec to use for pipelinerun %s: %v", pr.Name, err)
-		pr.Status.SetCondition(&apis.Condition{
-			Type:   apis.ConditionSucceeded,
-			Status: corev1.ConditionFalse,
-			Reason: ReasonCouldntGetPipeline,
-			Message: fmt.Sprintf("Error retrieving pipeline for pipelinerun %s: %s",
-				fmt.Sprintf("%s/%s", pr.Namespace, pr.Name), err),
-		})
+		pr.Status.MarkFailed(ReasonCouldntGetPipeline,
+			"Error retrieving pipeline for pipelinerun %s/%s: %s",
+			pr.Namespace, pr.Name, err)
 		return
 	}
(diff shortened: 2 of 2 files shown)


pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -227,6 +227,25 @@ func (pr *PipelineRunStatus) SetCondition(newCond *apis.Condition) {
...
+	pipelineRunCondSet.Manage(pr).MarkFalse(apis.ConditionSucceeded, reason, messageFormat, messageA...)
+	succeeded := pr.GetCondition(apis.ConditionSucceeded)
+	pr.CompletionTime = &succeeded.LastTransitionTime.Inner
+}
+
+// MarkRunning changes the Succeeded condition to Unknown with the provided reason and message.
+func (pr *PipelineRunStatus) MarkRunning(reason, messageFormat string, messageA ...interface{}) {
+	pipelineRunCondSet.Manage(pr).MarkUnknown(apis.ConditionSucceeded, reason, messageFormat, messageA...)
+}
+
```
vdemeester, line 248: :heart_eyes_cat:

pritidesai:
> The only concern left that I have is that we have an issue with setting the completion time today. We want to mark the pipeline failed as early as possible, however taskruns may still be running when that happens, and we need to keep them running until they finish, which means that the completion time does not really reflect the time when the last sub-resource completed.
> 
> To fix that issue with this change in place, it means we won't be able to set the pipeline as failed until all TaskRuns completed. Is this an acceptable approach?

yes its a valid concern. I havent looked into [this](https://github.com/tektoncd/pipeline/pull/2749/files#diff-0e0c01fdbdc46d2b6054a706f99516d0R563) yet (will look into it ASAP) but are we now setting pipeline competition time when a first taskRun failure is noted 🤔  

```
	after := resources.GetPipelineConditionStatus(pr, pipelineState, c.Logger, d)
	pr.Status.SetCondition(after)		switch after.Status {
	case corev1.ConditionTrue:
		pr.Status.MarkSucceeded(after.Reason, after.Message)
	case corev1.ConditionFalse:
		pr.Status.MarkFailed(after.Reason, after.Message)
```

In that case, we need a separate helper to not set the competition time on failure. The issue is `GetPipelineConditionStatus` returns `corev1.ConditionFalse` as soon as a failure is discovered without checking if any other taskRuns are still running. Ideally, `GetPipelineConditionStatus` should wait before returning failure but at the same time, reconciler should not pick any new tasks while in this waiting phase.

```
A    B     C
|      |      |
X    Y     Z
```

Let's say we have three tasks executing in parallel, A, B, and C. Task A fails but Task B and C are still running. Now, Task B finishes execution successfully but C is still running. While we are waiting for Task C to finish, Reconciler should not schedule (Task Y) subsequent node of Task B for execution otherwise we might break backward compatibility of failing pipeline on first task failure.

Trying to address this particular scenario in issue #1680

afrittoli:
TBH I'm more inclined now towards changing the logic to setting both status and completion time of the `PipelineRun` only once all subresources have finished their job. 
@pritidesai @vdemeester @bobcatfish @ImJasonH wdyt?

pritidesai:
I tested master with a pipeline having three tasks:

```
A(failure)    B(successful with sleep 10)     C(successful with sleep 20)
```

PipelineRun Completion Time is set to the task A competition time 😢   while task B and task C are still running:

```
kubectl get pr -o json | jq .items[].status.completionTime
"2020-06-05T19:52:03Z"

kubectl get pr -o json | jq '[ .items[].status.taskRuns[] | (.pipelineTaskName + " " + .status.completionTime) ]'
[
  "task-a 2020-06-05T19:52:03Z",
  "task-b 2020-06-05T19:52:16Z",
  "task-c 2020-06-05T19:52:27Z"
]
```

`tkn` CLI sets the duration to failed taskRun:

```
tkn pr describe pipelinerun-one-failure-two-success
Name:           pipelinerun-one-failure-two-success
Namespace:      default
Pipeline Ref:   pipeline-one-failure-two-success
Timeout:        1h0m0s
Labels:
 tekton.dev/pipeline=pipeline-one-failure-two-success

🌡️  Status

STARTED          DURATION     STATUS
18 minutes ago   10 seconds   Failed

💌 Message

TaskRun pipelinerun-one-failure-two-success-task-a-sgwh2 has failed ("step-fail" exited with code 1 (image: "docker-pullable://ubuntu@sha256:747d2dbbaaee995098c9792d99bd333c6783ce56150d1b11e333bbceed5c54d7"); for logs run: kubectl -n default logs pipelinerun-one-failure-two-success-task-a-sgwh2-pod-2xldc -c step-fail
)

📦 Resources

 No resources

⚓ Params

 No params

🗂  Taskruns

 NAME                                                 TASK NAME   STARTED          DURATION     STATUS
 ∙ pipelinerun-one-failure-two-success-task-a-sgwh2   task-a      18 minutes ago   10 seconds   Failed
 ∙ pipelinerun-one-failure-two-success-task-b-82xxr   task-b      18 minutes ago   23 seconds   Succeeded
 ∙ pipelinerun-one-failure-two-success-task-c-ztv7k   task-c      18 minutes ago   34 seconds   Succeeded
```

afrittoli:
Yeah, the completion time is not set along with setting success or failure, which is a change in behaviour. I think the next step on this should be to do so only once all resources spawned by the pipelinerun have completed their work.

@bobcatfish @vdemeester @pritidesai @mattmoor @ImJasonH thoughts?  
I would be happy to submit a PR for that if we agree on it. 
Personally I think it's ok to have v0.13.0 behaving like this.
Alternatively we could roll this back in a v0.13.1 branch (but not on master?) and continue towards the genreconciler for v0.14

mattmoor:
I'm a bit puzzled by why my change affected this unless the `Succeeded` condition is being set to failed and then unset before the `updateStatus`?  Let me take another look, but a workaround would be to have the `MarkUnknown` variant of this clear `CompletionTime` 🤔

mattmoor:
e.g. are we expecting this line to undo a `MarkFailed` earlier in the reconciliation? 
https://github.com/tektoncd/pipeline/blob/b51405aed08990db933ac13c4a4c12b3b7d7bfcb/pkg/reconciler/pipelinerun/pipelinerun.go#L566

afrittoli:
TBH I would prefer having stop setting the status to failed with no completion time, wait until the DAG is completed and then report how many failed / passed / skipped / cancelled. 
I started working on a patch for that, and I'll raise this question on our Monday API WG.

afrittoli:
https://github.com/tektoncd/pipeline/pull/2774

Issue #2640, opened 2020-05-18
Create a binary for Affinity Assistant used as a placeholder pod


jlpettersson:
This was requested in https://github.com/tektoncd/pipeline/pull/2630#discussion_r426566799

# Expected Behavior

A small, minimal binary, e.g. an Go http service.

# Actual Behavior

We currently use an Nginx image for this.

vdemeester:
@bobcatfish as quickly discussed today, the hardcoded nginx image is most likely gonna be a blocker for openshift-pipelines :sweat_smile: We can use a flag for this and still using nginx but allowing openshift-pipelines (and other integrator) to customize the image.

ImJasonH:
Does the image have to expose an HTTP service? Or just run indefinitely? Either way, a simple purpose-built Go binary + image sgtm.

Related #2562 #2714

jlpettersson:
> Or just run indefinitely?

Just run indefinitely. We could potentially use it for other things as well, if we want.

For this use case, it is just a pod with specific labels, so that taskRun-pods can configure podAffinity to attract that pod, so they are scheduled to the same Node.

ImJasonH:
It could (not sure) be nice to reuse the nop image (see #2714) and give it two modes: exit successfully immediately no matter what (nop), or run indefinitely no matter what (affinity assistant).

dlorenc:
> It could (not sure) be nice to reuse the nop image (see #2714) and give it two modes: exit successfully immediately no matter what (nop), or run indefinitely no matter what (affinity assistant).

The run indefinitely option sounds a lot like the k8s pause container: https://www.ianlewis.org/en/almighty-pause-container

we could maybe reuse that one - it is already optimized to be tiny and run forever with 0 CPU footprint.

dlorenc:
Source for it here: https://github.com/kubernetes/kubernetes/tree/master/build/pause

Issue #2738, opened 2020-06-03
Pipeline nightly build is broken


afrittoli:
# Expected Behavior
Pipeline nightly build works

# Actual Behavior
Pipeline nightly build is broken.
Building the base image fails with:
```
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/community/x86_64/APKINDEX.tar.gz
v3.12.0-30-g01407813ee [http://dl-cdn.alpinelinux.org/alpine/v3.12/main]
v3.12.0-29-gb310a5f576 [http://dl-cdn.alpinelinux.org/alpine/v3.12/community]
OK: 12726 distinct packages available
(1/1) Upgrading alpine-baselayout (3.2.0-r6 -> 3.2.0-r7)
Executing alpine-baselayout-3.2.0-r7.pre-upgrade
rm: can't remove '/var/run/secrets/kubernetes.io/serviceaccount/..data': Read-only file system
rm: can't remove '/var/run/secrets/kubernetes.io/serviceaccount/token': Read-only file system
rm: can't remove '/var/run/secrets/kubernetes.io/serviceaccount/namespace': Read-only file system
rm: can't remove '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt': Read-only file system
rm: can't remove '/var/run/secrets/kubernetes.io/serviceaccount/..2020_06_03_02_13_55.407209058/namespace': Read-only file system
rm: can't remove '/var/run/secrets/kubernetes.io/serviceaccount/..2020_06_03_02_13_55.407209058/ca.crt': Read-only file system
rm: can't remove '/var/run/secrets/kubernetes.io/serviceaccount/..2020_06_03_02_13_55.407209058/token': Read-only file system
Executing alpine-baselayout-3.2.0-r7.post-upgrade
ERROR: alpine-baselayout-3.2.0-r7: failed to rename var/.apk.f752bb51c942c7b3b4e0cf24875e21be9cdcd4595d8db384 to var/run.
Executing busybox-1.31.1-r16.trigger
1 error; 27 MiB in 25 packages
error building image: error building stage: failed to execute command: waiting for process to exit: exit status 1
```

# Steps to Reproduce the Problem

1. https://dashboard.dogfooding.tekton.dev/#/namespaces/default/pipelineruns/pipeline-release-nightly-tqgdd

# Additional Info
[...]

PR #2751, opened 2020-06-04
Externalize Affinity Assistant image

jlpettersson:
# Changes

Container images should not be hardcoded.
This commit externalize what image should be used for the Affinity Assistant.

/kind misc
Related to #2640

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -66,6 +66,11 @@ spec:
           "-pr-image", "github.com/tektoncd/pipeline/cmd/pullrequest-init",
           "-build-gcs-fetcher-image", "github.com/tektoncd/pipeline/vendor/github.com/GoogleCloudPlatform/cloud-builders/gcs-fetcher/cmd/gcs-fetcher",
 
+          # This image is used as a placeholder pod, the Affinity Assistant
+          # TODO(#2640) We may want to create a custom, minimal binary
+          # As of June 8, 2020, tag 1.19.0
+          "-affinity-assistant-image", "nginx@sha256:c870bf53de0357813af37b9500cb1c2ff9fb4c00120d5fe1d75c21591293c34d",
+
           # These images are pulled from Dockerhub, by digest, as of May 19, 2020.
           # As of May 29, 2020 new sha for nop image
           "-nop-image", "tianon/true@sha256:009cce421096698832595ce039aa13fa44327d96beedb84282a69d3dbcf5a81b",
--- a/pkg/reconciler/pipelinerun/affinity_assistant.go
+++ b/pkg/reconciler/pipelinerun/affinity_assistant.go
@@ -56,7 +56,8 @@ func (c *Reconciler) createAffinityAssistants(wb []v1alpha1.WorkspaceBinding, pr
 			claimName := getClaimName(w, pr.GetOwnerReference())
 			switch {
 			case apierrors.IsNotFound(err):
-				_, err := c.KubeClientSet.AppsV1().StatefulSets(namespace).Create(affinityAssistantStatefulSet(affinityAssistantName, pr, claimName))
+				affinityAssistantStatefulSet := affinityAssistantStatefulSet(affinityAssistantName, pr, claimName, c.Images.AffinityAssistantImage)
+				_, err := c.KubeClientSet.AppsV1().StatefulSets(namespace).Create(affinityAssistantStatefulSet)
 				if err != nil {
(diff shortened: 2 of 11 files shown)


config/controller.yaml
```diff
@@ -74,6 +74,9 @@ spec:
           # The shell image must be root in order to create directories and copy files to PVCs.
           # As of May 19, 2020
           "-shell-image", "gcr.io/distroless/base:debug@sha256:f79e093f9ba639c957ee857b1ad57ae5046c328998bf8f72b30081db4d8edbe4",
+
+          # This image is used as a placeholder pod, the Affinity Assistant
+          "-affinity-assistant-image", "nginx",
```
vdemeester, line 79: Should we "pin" a nginx image with digest here ? (as the above images)?

jlpettersson: I was thinking about it, but we don't depend on any functionality - just that it is a "service" (that does not terminate). But I don't know if it is better to pin?

ImJasonH: If it just needs to be an image that runs indefinitely, can it be `busybox` with `command: ['sleep', 'infinity']`?

jlpettersson: yes, but we want to externalize the image here, so that e.g. openshift-pipelines can use a custom image and registry. I think a custom Go-binary and built from "scratch" would be best, as in #2640

vdemeester: yeah, this PR is *just* about making sure we don't hardcode the image (in the release to come) :angel:

ImJasonH: sgtm, flagging is better than hard-coding, then we can work on making it a simpler Tekton-owned image.

cameronbraid:
What is wrong with using the tekton controller image with a new entrypoint that just sleeps indefinitely ?  Then a second image pull is not required, and will always be present.

config/controller.yaml
bobcatfish: > But I don't know if it is better to pin?

i think it makes sense to pin - without pinning images we depend on we sometimes end up in states where things are breaking, and often for reasons we don't really care about (e.g. https://github.com/tektoncd/pipeline/issues/2738) - pinning reduces the number of things that can change on us and cause us problems :D

pkg/reconciler/pipelinerun/affinity_assistant.go
```diff
@@ -130,10 +131,8 @@ func affinityAssistantStatefulSet(name string, pr *v1beta1.PipelineRun, claimNam
 	}
 
 	containers := []corev1.Container{{
-		Name: "affinity-assistant",
-
-		//TODO(#2640) We may want to create a custom, minimal binary
-		Image: "nginx",
```
bobcatfish, line 136: it seems like this doesn't completely resolve 2640 - would it make sense to keep this comment, maybe on the line where we default the value of `affinityAssistantImage`?

vdemeester: The comment might be put in `controller.yaml` yeah :upside_down_face:

config/controller.yaml
```diff
@@ -74,6 +74,11 @@ spec:
           # The shell image must be root in order to create directories and copy files to PVCs.
           # gcr.io/distroless/base:debug as of May 19, 2020
           "-shell-image", "gcr.io/distroless/base@sha256:f79e093f9ba639c957ee857b1ad57ae5046c328998bf8f72b30081db4d8edbe4",
+
+          # This image is used as a placeholder pod, the Affinity Assistant
+          # TODO(#2640) We may want to create a custom, minimal binary
+          # As of June 8, 2020, tag 1.19.0
```
bobcatfish, line 80: 👍 thanks!!

jlpettersson:
>Auto-merging config/controller.yaml
CONFLICT (content): Merge conflict in config/controller.yaml
Automatic merge failed; fix conflicts and then commit the result.

From the integration test. Yes, I had fixed the rebase-conflict. But what do I need to do now? Fix it in a different way? Manual merge?

jlpettersson:
I moved my changes in `controller.yaml` to avoid merge conflicts with https://github.com/tektoncd/pipeline/pull/2781

bobcatfish:
> I moved my changes in controller.yaml to avoid merge conflicts with #2781

nice!!! thanks for that extra attention to making this easy to cherry-pick

PR #2762, opened 2020-06-05
Disentangle metric reporting from the actual reconciler.

mattmoor:
This changes the metric reporting to happen periodically vs. being
triggered by reconciliation.  The previous method was prone to stale
data because it was driven by the informer cache immediately following
writes through the client, and might not fix itself since the status
may not change (even on resync every 10h).  With this change it should
exhibit the correct value within 30s + {informer delay}, where the 30s
is configurable on the Recorders.

Fixes: #2729

### DIFF
--- a/pkg/reconciler/pipelinerun/controller.go
+++ b/pkg/reconciler/pipelinerun/controller.go
@@ -40,10 +40,6 @@ import (
 	"knative.dev/pkg/tracker"
 )
 
-const (
-	resyncPeriod = 10 * time.Hour
-)
-
 // NewController instantiates a new controller.Impl from knative.dev/pkg/controller
 func NewController(namespace string, images pipeline.Images) func(context.Context, configmap.Watcher) *controller.Impl {
 	return func(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
@@ -102,6 +98,8 @@ func NewController(namespace string, images pipeline.Images) func(context.Contex
 		c.Logger.Info("Setting up ConfigMap receivers")
 		c.configStore = config.NewStore(images, c.Logger.Named("config-store"))
 		c.configStore.WatchConfigs(cmw)
+
+		go metrics.ReportRunningPipelineRuns(ctx, pipelineRunInformer.Lister())
 
 		return impl
 	}
--- a/pkg/reconciler/pipelinerun/metrics.go
+++ b/pkg/reconciler/pipelinerun/metrics.go
@@ -29,6 +29,7 @@ import (
 	"go.opencensus.io/tag"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"knative.dev/pkg/logging"
 	"knative.dev/pkg/metrics"
 )
 
@@ -56,6 +57,8 @@ type Recorder struct {
 	pipelineRun tag.Key
 	namespace   tag.Key
 	status      tag.Key
+
+	ReportingPeriod time.Duration
 }
 
 // NewRecorder creates a new metrics recorder instance
@@ -63,6 +66,9 @@ type Recorder struct {
 func NewRecorder() (*Recorder, error) {
 	r := &Recorder{
 		initialized: true,
+
+		// Default to 30s intervals.
+		ReportingPeriod: 30 * time.Second,
 	}
 
 	pipeline, err := tag.NewKey("pipeline")
@@ -183,4 +189,23 @@ func (r *Recorder) RunningPipelineRuns(lister listers.PipelineRunLister) error {
 	metrics.Record(ctx, runningPRsCount.M(float64(runningPRs)))
 
 	return nil
+}
+
+// ReportRunningPipelineRuns invokes RunningPipelineRuns on our configured PeriodSeconds
+// until the context is cancelled.
+func (r *Recorder) ReportRunningPipelineRuns(ctx context.Context, lister listers.PipelineRunLister) {
+	logger := logging.FromContext(ctx)
+	for {
+		select {
+		case <-ctx.Done():
+			// When the context is cancelled, stop reporting.
+			return
+
+		case <-time.After(r.ReportingPeriod):
+			// Every 30s surface a metric for the number of running pipelines.
+			if err := r.RunningPipelineRuns(lister); err != nil {
+				logger.Warnf("Failed to log the metrics : %v", err)
+			}
+		}
+	}
 }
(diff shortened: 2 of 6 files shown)


pkg/reconciler/pipelinerun/controller.go
```diff
@@ -40,10 +40,6 @@ import (
 	"knative.dev/pkg/tracker"
 )
 
-const (
-	resyncPeriod = 10 * time.Hour
-)
-
```
afrittoli, line 46: This was stale code?

pkg/reconciler/pipelinerun/metrics.go
```diff
@@ -56,13 +57,18 @@ type Recorder struct {
...
 }
 
 // NewRecorder creates a new metrics recorder instance
 // to log the PipelineRun related metrics
 func NewRecorder() (*Recorder, error) {
 	r := &Recorder{
 		initialized: true,
+
+		// Default to 30s intervals.
+		ReportingPeriod: 30 * time.Second,
```
hrishin, line 71: It could be nice if we keep this configurable? 🤔

pkg/reconciler/pipelinerun/controller.go
mattmoor: Yeah, I cleaned up it's one use previously but missed this.

pkg/reconciler/pipelinerun/metrics.go
mattmoor: It's a public field, so it can be changed after the resource is created, but I see that this is just consumed by NewController without really having a place to hook into.   I'm happy to whatever is conventional for Tekton here.

mattmoor:
> Processing o(n) resources every time could result in any unwanted bitter side effect for the controller deployment

O(n) isn't a change from before, but the way this is written now ensures that only one scan i happening at a time.  Suppose "n" is huge and it takes 10 minutes to scan the lister cache.

This approach would just take 10:30 to expose the metric.  The previous approach would leak go routines every time we update a single resources status.  So in addition to being O(n) in runtime, it also might leak O(n) go routines[1] if all n pipelines are actively running.

[1] - I'm assuming there are roughly a constant number of status updates, but this could be much higher given that there is no deduplication by key.

> If resync could be a problem, isn't it a general reconciliation problem? 

It is.  For global resyncs [we jitter](https://github.com/knative/pkg/blob/82fe339a5ea5196b455eec43d1eef4ac35b2a460/controller/controller.go#L481) to avoid starving real work for a period of time.  The default resync period is also every 10 hours to mitigate this, but that seems overly coarse for metrics resolution to be useful.

hrishin:
Thank you @mattmoor for the explanation.

> O(n) isn't a change from before, but the way this is written now ensures that only one scan i happening at a time. Suppose "n" is huge and it takes 10 minutes to scan the lister cache.

O(n) holds true, sorry to rephrase it, I meant every time vs only when something changes. :)

> The previous approach would leak go routines every time we update a single resources status.

True, as there wasn't serialization in place (frankly I was thinking to serialize it during the implementation using queue but later on felt that would be over-engineering) thus it could trigger the create stale metrics. However, the current PR approach deals better in that case. 

[2]
>The previous method was prone to stale
data because it was driven by the informer cache immediately following
writes through the client, and might not fix itself since the status
may not change (even on resync every 10h). With this change it should
exhibit the correct value within 30s + {informer delay}, where the 30s
is configurable on the Recorders.

On a different note, I'm still not able to get comment[2] and bit curious, would you like to elaborate on it?
As [Comment ](https://github.com/tektoncd/pipeline/issues/2729#issue-628901986)mentions
>Now, informers are fast, but in all likelihood the lister that is being used to drive this metric hasn't received the updated version of the resource from the API Server, so unless the informer cache's copy is being modified (a no no) then the cache won't reflect what it's clearly (given the if updated check) trying to capture.

However, I'm under the impression that either through the list or watch the updated copy of the resource is placed into the cache first then informer is notified about the updated resource which triggers reconciliation. Thus lister in the reconciler() refers the same copy from the cache which is up to the update. Is it right?

mattmoor:
>  I'm under the impression that either through the list or watch the updated copy of the resource is placed into the cache first then informer is notified about the updated resource which triggers reconciliation

**Yes**, the informer model uses watches, lists, and changes on those lists to *simulate* events by observing changes on those lists through the watch.  The premise of using the informer cache for reads during reconciliation is that the data is never a stale read of the change that motivated the reconciliation (because the event was generated by an update to the informer cache).  It could still be stale if subsequent reads came in, but this is why "in-flight" work isn't deduplicated with "queued" work (e.g. even if a key is being processed it can be duplicated in the work queue).

**However**, in the case at HEAD, the metrics logic is keying off of `if updated`, which is NOT an informer event (e.g. observer-side where lister caches have been updated).  So the sequence of events where this works is:
1. update status (`updated = true`)
2. informer QUICKLY observes the change and updates the lister cache (the key is also requeued)
3. metrics go routine is spawned and sees correct data
4. The key is processed again, nothing to do (`updated = false`)
5. metrics go routine is not spawned because nothing changed.

`2.` -> `3.` is the informer event happening first by chance, but there is no causal link here.  We aren't updating metrics because the informer cache reflects the updated state.

The sequence of events where this doesn't work is:
1. update status (`updated = true`)
2. metrics go routine is spawned and sees INCORRECT data
3. informer SLOWLY observes the change and updates the lister cache (the key is also requeued)
4. The key is processed again, nothing to do (`updated = false`)
5. metrics go routine is not spawned because nothing changed.

---
In the former case we get really lucky (informers are fast, but not "faster than a local go routine" fast).  There used to be a bug that meant we didn't have to get lucky, but that is fixed at HEAD.  Two other notable aspects to this:
1. The metrics data is never reevaluated on a nop resync, so bad data stays bad until there's a change (and we get lucky enough that there isn't new bad data).
2. When the controller process starts up, it won't surface this metric data until it observes a change.

Fundamentally what's at HEAD is "edge triggered" (keying off of deltas), vs. "level based" (changes that trigger more holistic reconsideration).  It's not that "level based" is the only correct way to do things and it can seem wasteful to reconsider everything when we think we know exactly what changed, but it can be much more forgiving of errors and dropped events.  I had some slides on this that it seems I never shared publicly and are locked up at Google somewhere 😞

pkg/reconciler/pipelinerun/metrics.go
hrishin: I'm not sure as well(I was away for some time). It's better if @afrittoli @vdemeester comment more on it.
In my opinion, we can have it as it is. If required in the future we can provide the relevant configuration option :)

However, maybe it's nice to document it? so users are aware of 30 seconds window to get updated metrics

afrittoli:
@mattmoor  I'm afraid this will need a rebase :)

mattmoor:
Rebased.

mattmoor:
(will need a fresh `/lgtm`)

I'll work on rebasing the others now as well.

Issue #1171, opened 2019-08-06
x509 certificate signed by unknown authority


pierretasci:
# Expected Behavior
I am trying to run a container from my private registry within Tekton. I would like Tekton to invoke the `entrypoint` defined in the container rather than specify a command. Per, https://github.com/tektoncd/pipeline/blob/master/docs/container-contract.md I have added an ImagePullSecret to the service account running my pipeline.

I would expect Tekton to be able to pull my image, read the entrypoint and execute that in the pipeline.

# Actual Behavior

In reality I am getting a certificate error from the entrypoint binary that is inserted into the execution. 

`Failed to create build pod "[my-pod-name]": couldn't create redirected TaskSpec: failed to add entrypoint to steps of TaskRun [my-pod-name]: Failed to fetch remote image [myprivateregistry.com]/pierretasci/tests:latest: Failed to get container image info from registry [myprivateregistry.com]/pierretasci/tests:latest: Get https://[myprivateregistry.com]/v2/: x509: certificate signed by unknown authority`

The weird thing is that if I specify the command in the entrypoint, then this works. It is able to pull the image and execute it. It is only within the task execution when I don't specify a command that this happens.

# Steps to Reproduce the Problem

1. Use a private docker repo like Artifactory
2. Create a k8s image pull secret for this and add it to your service account
3. Define a pipeline with a task that does not specify a command and references a container from the private docker

vdemeester:
@pierretasci Thanks for the issue. This is one pitfall of our `entrypoint` dark magic… WIthout a command, pipeline tries to get the image entrypoint from the registry, and, in your case fails because of certificates.

We need to find a way to use k8s image pull secret in pipeline (or do that in a separate container that would have this secret) when we try to fetch the image config from the registry. We will also need a way for the user to give that information (although if it's attach to the serviceaccount, we can manage)

/kind bug

pierretasci:
Yeah that is what I figured was going on so I am glad to see that I'm not crazy 😄 

Pardon my naivete, but is an image pull secret enough here? If the issue is that the image running the entrypoint "magic" doesn't trust the self-signed cert from my registry, being able to login to the registry wouldn't be enough. Is that about right?

vdemeester:
@pierretasci it should be enough, but it's the pipeline controller itself that also needs it (to get the image config)…

hougangliu:
I hit the same problem in my use case，any process for this issue?

JustinKuli:
[...]

PR #2764, opened 2020-06-05
Add registry cert config map to support self-signed private registry

FogDong:
# Changes

I met the same problem as tektoncd/pipeline#1171 when I use self-signed private registry.
Use config map to configure users' cert and inject it as an env in the controller could be a good way to solve it.

This commit
- Add a config map to store users' cert.
- Add doc to explain it.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Add registry cert config map to support self-signed private registry
```

### DIFF
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -18,6 +18,7 @@ weight: 4
   - [Specifying `Workspaces`](#specifying-workspaces)
   - [Specifying `LimitRange` values](#specifying-limitrange-values)
   - [Configuring a failure timeout](#configuring-a-failure-timeout)
+- [Monitoring execution status](#monitoring-execution-status)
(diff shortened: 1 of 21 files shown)


bobcatfish:
Hey @FogDong ! I'm wondering, do we also need to update the controller to use this configuration or is there something else in the cluster that uses it already?

Also adding this to the config/ dir would add it for all installations of Tekton - is this something we want by default always, or something that maybe users would want to optionally add?

FogDong:
Hey @bobcatfish !
It's an optionally add, but I still think it should be added into the config dir to install it, otherwise users can not pull image from self-signed registry, they need to insert cert as an env manually in the controller deployment.
I also edit the install doc, it's an optional configuration if users are using self-signed registry

PR #2771, opened 2020-06-06
Migrate the TaskRun controller to genreconciler.

mattmoor:
There are effectively two pieces to this change:
1. Generate and consume the base reconciler (dropping boilerplate)
2. Update tests to plumb certain things differently (the existing tests followed our bad example and used unsafe type cases to root around in the `controller.Reconciler.(*Reconciler)`)

The changes in the pipeline run files are mostly for symmetry, or because a common utility changed.

This is WIP until https://github.com/tektoncd/pipeline/pull/2760 lands.


mattmoor:
oh boy, my favorite.  dependency cycles!

mattmoor:
/assign @afrittoli @vdemeester @ImJasonH 

Alright, this should be RFAL.  There's some cruft to cleanup after this, but this has already started to creep a bit in the cleanup department. 😬

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -80,13 +83,15 @@ func getRunName(pr *v1beta1.PipelineRun) string {
 func getPipelineRunController(t *testing.T, d test.Data) (test.Assets, func()) {
 	//unregisterMetrics()
 	ctx, _ := ttesting.SetupFakeContext(t)
+	ctx, cancel := context.WithCancel(ctx)
```
vdemeester, line 86: ohhh :sweat:

mattmoor: Not sure I understand...

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -727,8 +662,7 @@ func updateStoppedSidecarStatus(pod *corev1.Pod, tr *v1beta1.TaskRun, c *Reconci
 			})
 		}
 	}
-	_, err := c.updateStatus(tr)
```
danielhelfand, line 730: Just making a note here in case this comes into question, but checked this out and seems to be fine without the call to `updateStatus`. Thanks for noticing and correcting.

test/controller.go
```diff
@@ -70,9 +73,10 @@ type Data struct {
 
 // Clients holds references to clients which are useful for reconciler tests.
 type Clients struct {
-	Pipeline *fakepipelineclientset.Clientset
-	Resource *fakeresourceclientset.Clientset
-	Kube     *fakekubeclientset.Clientset
+	Pipeline    *fakepipelineclientset.Clientset
+	Resource    *fakeresourceclientset.Clientset
+	Kube        *fakekubeclientset.Clientset
+	CloudEvents cloudeventclient.CEClient
```
afrittoli, line 79: I knew there was a better place to store this client :D only at the time I had no idea :P

pkg/reconciler/pipelinerun/pipelinerun_test.go
vdemeester: @mattmoor it was me realizing where I did put that `context.WithCancel` before :wink:

PR #2661, opened 2020-05-21
implementing pipeline level finally

pritidesai:
# Changes

We can now specify a list of tasks needs to be executed just before
pipeline exits (either after finishing all non-final tasks successfully or after
a single failure)

Most useful for tasks such as report test results, cleanup cluster resources, etc

### Summary of Changes:

1) `pipelines.md`: Documented final tasks including how to specify `workspaces` and `params`, documenting updates to pipeline run status as a result of  finally clause, and known limitations (task results, `from` clause, pipeline results, `runAfter`, and `conditions`).
2)  `pipelinerun-with-final-tasks.yaml`: Example pipeline demonstrating cloning a repo into shared workspace and cleaning up the workspace.
3) `pipelinerun.go`: Building graph for `final` tasks, build `pipelineState` for both DAG tasks and final tasks, invoke final tasks once all DAG tasks are done (execution queue is empty)
4) `pipelinerun_test.go`: Test three major test pipelines, (1) pipeline when a DAG task fails but final task succeeds, (2) pipeline when all DAG task succeeds but a final task fails, (3) pipeline when one of the DAG tasks fail and a final task fails
5) `pipelinerunresolution.go`: Replace `SuccessfulPipelineTaskNames` with `SuccessfulOrSkippedDAGTasks` like explained here. Two helper functions `isDAGTasksDone`, `isDAGTaskFailed `. `GetFinalTasks` to return final tasks once they are eligible to execute.
6) `pipelinefinally_test.go`: Validate pipelineRun Status and taskRun Status along with competition time making sure (1) Final tasks are executed after all DAG tasks succeeds (2) Pipeline results in failure since a DAG task fails but final tasks does get executed successfully (3) Pipeline exits with failure when final task execution fails (4) The DAG task is skipped due to condition failure, final tasks is still executed.

### Example:

```
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: pipeline-with-final-tasks
spec:
  tasks:
    - name: pre-work
      taskRef:
        Name: some-pre-work
    - name: unit-test
      taskRef:
        Name: run-unit-test
      runAfter:
        - pre-work
    - name: integration-test
      taskRef:
        Name: run-integration-test
      runAfter:
        - unit-test
  finally:
    - name: cleanup-test
      taskRef:
        Name: cleanup-cluster
    - name: report-results
      taskRef:
        Name: report-test-results
```

[...]

PR #2774, opened 2020-06-07
Set pipeline status when all tasks complete

### LINKED ISSUE #1680: (*PipelineRun).IsDone() is incorrect

afrittoli:
# Changes
We used to set the pipeline status to failed as soon as the first
task in the pipeline failed or was cancelled.

As soon as the first task in the pipeline fails or is cancelled, we
stop scheduling new tasks, as we did before, but we will report
status Unknown until all Tasks are complete, with reason "stopping".

This allows to:
- the completion time at the same time that the status is set
  and avoid inconsistencies
- wait until all tasks are complete before we cleanup the pipeline
  artifact storage, affinity assistant and record metrics
- report the correct number of failed / cancelled tasks, as there
  may be more than one. Other tasks that were already running
  when the first failure happened may fail too
- prepare the pipeline controller more complex workflows, where
  the controller may continue working scheduling after failures

Add test coverage for isSkipped and extend the pipelineresolution
module unit test coverage to capture more input pipeline state.

The DAG / GetNextTasks functions have not been touched at all.
The pipeline run reconciler, when the pipeline run is in stopping
mode, stops asking for next tasks to run. Once all running tasks
finish, the pipeline run finally gets in failed state.

When the pipeline run is in stopping mode, tasks that have not
been started yet are also counted as skipped in the status message
reported to the user.

Fixes #1680

# Submitter Checklist

review them:

- [WIP] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)

## Reviewer Notes

# Release Notes

```
In case of task run failure or cancellation, the pipeline run stops scheduling new tasks like before.
The status is marked as failed however only once all the task runs already scheduled are complete.

A new reason "PipelineRunStopping" indicate that the pipeline run has found a failure and is waiting for task runs to complete.
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -426,9 +426,51 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 		}
 	}
 
+	as, err := artifacts.InitializeArtifactStorage(c.Images, pr, pipelineSpec, c.KubeClientSet, logger)
+	if err != nil {
+		logger.Infof("PipelineRun failed to initialize artifact storage %s", pr.Name)
+		return err
+	}
+
+	// When the pipeline run is stopping, we don't schedule any new task and only
+	// wait for all running tasks to complete and report their status
+	if !pipelineState.IsStopping() {
+		err = c.runNextSchedulableTask(ctx, pr, d, pipelineState, as)
+		if err != nil {
+			return err
+		}
+	}
+
+	before := pr.Status.GetCondition(apis.ConditionSucceeded)
+	after := resources.GetPipelineConditionStatus(pr, pipelineState, logger, d)
+	switch after.Status {
+	case corev1.ConditionTrue:
+		pr.Status.MarkSucceeded(after.Reason, after.Message)
+	case corev1.ConditionFalse:
+		pr.Status.MarkFailed(after.Reason, after.Message)
+	case corev1.ConditionUnknown:
+		pr.Status.MarkRunning(after.Reason, after.Message)
+	}
+	// Read the condition the way it was set by the Mark* helpers
+	after = pr.Status.GetCondition(apis.ConditionSucceeded)
+	events.Emit(recorder, before, after, pr)
+
+	pr.Status.TaskRuns = getTaskRunsStatus(pr, pipelineState)
+	logger.Infof("PipelineRun %s status is being set to %s", pr.Name, after)
+	return nil
+}
+
+// runNextSchedulableTask gets the next schedulable Tasks from the dag based on the current
+// pipeline run state, and starts them
+func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.PipelineRun, d *dag.Graph, pipelineState resources.PipelineRunState, as artifacts.ArtifactStorageInterface) error {
+
+	logger := logging.FromContext(ctx)
+	recorder := controller.GetEventRecorder(ctx)
+
 	candidateTasks, err := dag.GetSchedulable(d, pipelineState.SuccessfulPipelineTaskNames()...)
 	if err != nil {
 		logger.Errorf("Error getting potential next tasks for valid pipelinerun %s: %v", pr.Name, err)
+		return nil
 	}
 
 	nextRprts := pipelineState.GetNextTasks(candidateTasks)
@@ -439,13 +481,6 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 		return nil
 	}
 	resources.ApplyTaskResults(nextRprts, resolvedResultRefs)
-
-	var as artifacts.ArtifactStorageInterface
-
-	if as, err = artifacts.InitializeArtifactStorage(c.Images, pr, pipelineSpec, c.KubeClientSet, logger); err != nil {
-		logger.Infof("PipelineRun failed to initialize artifact storage %s", pr.Name)
-		return err
-	}
 
 	for _, rprt := range nextRprts {
 		if rprt == nil {
@@ -467,20 +502,6 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 			}
 		}
 	}
-	before := pr.Status.GetCondition(apis.ConditionSucceeded)
-	after := resources.GetPipelineConditionStatus(pr, pipelineState, logger, d)
-	switch after.Status {
-	case corev1.ConditionTrue:
-		pr.Status.MarkSucceeded(after.Reason, after.Message)
-	case corev1.ConditionFalse:
-		pr.Status.MarkFailed(after.Reason, after.Message)
-	case corev1.ConditionUnknown:
-		pr.Status.MarkRunning(after.Reason, after.Message)
-	}
-	events.Emit(recorder, before, after, pr)
-
-	pr.Status.TaskRuns = getTaskRunsStatus(pr, pipelineState)
-	logger.Infof("PipelineRun %s status is being set to %s", pr.Name, after)
 	return nil
 }
 
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -127,6 +127,20 @@ func (t ResolvedPipelineRunTask) IsCancelled() bool {
 	return c.IsFalse() && c.Reason == v1beta1.TaskRunReasonCancelled.String()
 }
 
+// IsStarted returns true only if the PipelineRunTask itself has a TaskRun associated
+func (t ResolvedPipelineRunTask) IsStarted() bool {
+	if t.TaskRun == nil {
+		return false
+	}
+
+	c := t.TaskRun.Status.GetCondition(apis.ConditionSucceeded)
+	if c == nil {
+		return false
+	}
+
+	return true
+}
+
 // ToMap returns a map that maps pipeline task name to the resolved pipeline run task
 func (state PipelineRunState) ToMap() map[string]*ResolvedPipelineRunTask {
 	m := make(map[string]*ResolvedPipelineRunTask)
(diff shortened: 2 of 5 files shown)


pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -422,67 +440,78 @@ func GetPipelineConditionStatus(pr *v1beta1.PipelineRun, state PipelineRunState,
...
+			Reason: reason,
+			Message: fmt.Sprintf("Tasks Completed: %d (Failed: %d, Cancelled %d), Skipped: %d",
+				len(allTasks)-skipTasks, failedTasks, cancelledTasks, skipTasks),
 		}
 	}
 
-	// Hasn't timed out; no taskrun failed yet; and not all tasks have finished....
+	// Hasn't timed out; not all tasks have finished....
 	// Must keep running then....
+	if failedTasks > 0 || cancelledTasks > 0 {
```
GregDritschler, line 504: Isn't this point reached if there are dependent tasks that haven't been (and won't be) scheduled?  I don't see anything distinguishing running from not yet running.

afrittoli: Uhm, when I wrote this code I understood that isSkipped would mark descendants of failed tasks as skipped, but it doesn't, so I may need to revisit that, and add more unit tests because that should be caught by testing if it's not working as it should.

I thought the case you mentioned was covered but I'll check that again

afrittoli: Uhm, when I wrote this code I understood that isSkipped would mark descendants of failed tasks as skipped, but it doesn't, so I may need to revisit that, and add more unit tests because that should be caught by testing if it's not working as it should.

I thought the case you mentioned was covered but I'll check that again

pritidesai: @afrittoli havent looked all the changes yet but, skipped tasks are very tricky, please see my comment [here](https://github.com/tektoncd/pipeline/pull/2661#discussion_r428827199)

afrittoli: Thanks for catching this issue!

I've added now logic to distinguish the running and non running cases.
When the pipeline is in stopping mode, taskrun that have not been started yet will be counted as skipped. So if there is any task that is running the pipeline stays in stopping, else it switches to failed.

I've added more test cases for this too.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -421,9 +421,48 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
...
 
+	as, err := artifacts.InitializeArtifactStorage(c.Images, pr, pipelineSpec, c.KubeClientSet, c.Logger)
+	if err != nil {
+		c.Logger.Infof("PipelineRun failed to initialize artifact storage %s", pr.Name)
+		return err
+	}
+
+	// When the pipeline run is stopping, we don't schedule any new task and only
+	// wait for all running tasks to complete and report their status
+	if !pipelineState.IsStopping() {
```
afrittoli, line 432: The diff does not help very much, but the only changes here are:
- `if !pipelineState.IsStopping()`
- the code that used to be here moved "as-is" into `runNextSchedulableTask`

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
GregDritschler: 👍

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -486,7 +534,12 @@ func isSkipped(rprt *ResolvedPipelineRunTask, stateMap map[string]*ResolvedPipel
 	// if any of the parents have been skipped, skip as well
 	node := d.Nodes[rprt.PipelineTask.Name]
 	for _, p := range node.Prev {
-		skip := isSkipped(stateMap[p.Task.HashKey()], stateMap, d)
+		rcc := stateMap[p.Task.HashKey()]
```
pritidesai, line 537: Is this something avoidable? marking a task skipped if one of the parents is failed or cancelled, this might interfere with `except` similar to `finally` but is invoked only when a task fails 🤔 .`Pipeline` is anyways `stopped` or `paused` (may be better term), resume it where we left off and report failure. Just brainstorming here 💭 how does these changes impact pipeline run status?

pritidesai: Today, tasks are considered `skipped` (1) when its condition check fails and hence not executed and skipped (2) Or one of the parents is skipped. Tasks dependent on failed tasks are considered as `incomplete`.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -401,72 +429,92 @@ func GetPipelineConditionStatus(pr *v1beta1.PipelineRun, state PipelineRunState,
...
+		case !rprt.IsStarted() && isStopping:
+			// If the pipeline is in stopping mode, all tasks that are not running
+			// already will be skipped. Otherwise these tasks end up in the
+			// incomplete count.
 			skipTasks++
-			successOrSkipTasks = append(successOrSkipTasks, rprt.PipelineTask.Name)
+			withStatusTasks = append(withStatusTasks, rprt.PipelineTask.Name)
+		case rprt.IsSuccessful():
+			withStatusTasks = append(withStatusTasks, rprt.PipelineTask.Name)
+		case isSkipped(rprt, stateAsMap, dag):
```
vdemeester, line 464: nit: any reason for not having `rpr.IsSkipped(stateAsMap, dag)` here ?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: > Tasks are considered `skipped` (1) when its condition check fails and hence not executed and skipped (2) Or one of the parents is skipped. Tasks dependent on failed tasks are considered as `incomplete`.

Well, that's the logic I would like to change, i.e. recognize tasks that are going to be skipped as early as possible. This is useful when a pipeline is stopping, but it also gives a better indication to users about the actual incomplete. For instance finally tasks, would never go to skipped.

afrittoli: > Is this something avoidable? marking a task skipped if one of the parents is failed or cancelled, this might interfere with `except` similar to `finally` but is invoked only when a task fails 🤔 .

Well, I'm not really marking the tasks, this computation is only used to decide what the overall status of the pipeline is, but not extra info is added into the pipeline state.

> `Pipeline` is anyways `stopped` or `paused` (may be better term), resume it where we left off and report failure. Just brainstorming here 💭 how does these changes impact pipeline run status?

I used stopping because I think it better reflects the fact that some tasks are still running. No new tasks are scheduled but something is still going on.
The pipeline status, as soon as a failure/task cancel occurs, is updated as follows:
- if other tasks are still running, the status remains "unknown", the reason is changed to "stopping"  and in the message "incomplete" tasks are counted as "skipped" as they will never be executed
- if no other task is running, the status is set to "failed", the reason is set to the appropriate one (failed or cancelled) and the message shows the final count of completed / failed / cancelled / skipped tasks

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: No particular reason, it was like that before and I did not change it.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: But I think I should remove the check for "failed/cancelled" parent from the `isSkipped` anyways, since it does not serve any purpose right now (it did in my initial version of the code):
- if the pipeline is running normally (no failure, no cancel), this check will add anything
- if the pipeline is stopping, the actual number of skipped is given by:
```
		case !rprt.IsStarted() && isStopping:
			// If the pipeline is in stopping mode, all tasks that are not running
			// already will be skipped. Otherwise these tasks end up in the
			// incomplete count.
			skipTasks++
			withStatusTasks = append(withStatusTasks, rprt.PipelineTask.Name) 
```

afrittoli: Ok, I pushed a new version that removed the changes to `isSkipped` but kept the unit test coverage for that. That way we can keep better track of the behaviour of that function independently of the `GetPipelineConditionStatus`

afrittoli: > Is this something avoidable? 

Just to confirm, yes, it is avoidable, and I removed it now. Thank you!

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: There is a small advantage in having it like this because I can run `toMap` once, so I don't have to re-build the map for every single pipeline task.

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: May be this could be simplified and keep it as is, instead change `GetNextTasks()` such that no tasks are returned if pipeline is in stopping mode:

```
func (state PipelineRunState) GetNextTasks(candidateTasks map[string]struct{}) []*ResolvedPipelineRunTask {
	tasks := []*ResolvedPipelineRunTask{}
	if state.IsStopping() {
		return tasks
	}
```

With an empty execution queue, everything from `runNextSchedulableTask` is escaped except `InitializeArtificatStorage` (could be surrounded by the condition)

Dont worry if you are not comfortable changing it, I will have to change it in `finally` PR, as final tasks have to be executed after pipeline is stopping and once all running tasks are done.

pritidesai:
/lgtm

This PR will impact `finally` PR #2661, I am fine merging this before we merge `finally` PR, will incorporate necessary changes into that PR.

PR #2787, opened 2020-06-09
Add registry cert config map to support self-signed private registry

FogDong:
# Changes

I met the same problem as tektoncd/pipeline#1171 when I use self-signed private registry.
Use config map to configure users' cert and inject it as an env in the controller could be a good way to solve it.
There's a problem with my last PR tektoncd/pipeline#2764 so I opened a new one.
/kind misc

This commit
- Add a config map to store users' cert.
- Add doc to explain it.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Add registry cert config map to support self-signed private registry
```

### DIFF
--- a/config/200-role.yaml
+++ b/config/200-role.yaml
@@ -29,7 +29,7 @@ rules:
   - apiGroups: [""]
     resources: ["configmaps"]
     verbs: ["get"]
-    resourceNames: ["config-logging", "config-observability", "config-artifact-bucket", "config-artifact-pvc", "feature-flags", "config-leader-election"]
+    resourceNames: ["config-logging", "config-observability", "config-artifact-bucket", "config-artifact-pvc", "feature-flags", "config-leader-election", "config-registry-cert"]
   - apiGroups: ["policy"]
     resources: ["podsecuritypolicies"]
     resourceNames: ["tekton-pipelines"]
--- a/config/config-registry-cert.yaml
+++ b/config/config-registry-cert.yaml
@@ -1,0 +1,25 @@
+# Copyright 2020 Tekton Authors LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-registry-cert
+  namespace: tekton-pipelines
+  labels:
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/part-of: tekton-pipelines
+# data:
+#  # Registry's self-signed certificate
+#  cert: |
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -78,6 +78,8 @@ spec:
         volumeMounts:
         - name: config-logging
           mountPath: /etc/config-logging
+        - name: config-registry-cert
+          mountPath: /etc/config-registry-cert
         env:
         - name: SYSTEM_NAMESPACE
           valueFrom:
@@ -100,6 +102,10 @@ spec:
           value: feature-flags
         - name: CONFIG_LEADERELECTION_NAME
           value: config-leader-election
+        - name: SSL_CERT_FILE
+          value: /etc/config-registry-cert/cert
+        - name: SSL_CERT_DIR
+          value: /etc/ssl/certs
         - name: METRICS_DOMAIN
           value: tekton.dev/pipeline
         securityContext:
@@ -110,6 +116,9 @@ spec:
         - name: config-logging
           configMap:
             name: config-logging
+        - name: config-registry-cert
+          configMap:
+            name: config-registry-cert
 ---
 apiVersion: v1
 kind: Service
--- a/docs/install.md
+++ b/docs/install.md
@@ -262,6 +262,10 @@ data:
   default-cloud-events-sink: https://my-sink-url
 ```
 
+## Configuring self-signed cert for private registry
+
+The `SSL_CERT_DIR` is set to `/etc/ssl/certs` as the default cert directory. If you are using a self-signed cert for private registry and the cert file is not under the default cert directory, configure your registry cert in the `config-registry-cert` `ConfigMap` with the key `cert`. 
+
 ## Customizing basic execution parameters
 
 You can specify your own values that replace the default service account (`ServiceAccount`), timeout (`Timeout`), and Pod template (`PodTemplate`) values used by Tekton Pipelines in `TaskRun` and `PipelineRun` definitions. To do so, modify the ConfigMap `config-defaults` with your desired values.


sbwsg:
I'm a bit confused - how is the certificate used?  Is there some code in Pipelines or a dependency that leverages it somehow?

FogDong:
Hi, @sbwsg !
The certificate used as env in controller, based on google/ko#142

FogDong:
Sorry I just realized that I changed the env name while I was rechecking my code, just to keep the common env format like `CONFIG_....` 😅😥😣
Thanks for your review @sbwsg 😊

sbwsg:
Please squash commits.

This will also need someone else to review.  Thanks @FogDong !

bobcatfish:
Following the links from the ko issue:

> On UNIX systems the environment variables SSL_CERT_FILE and SSL_CERT_DIR can be used to override the system default locations for the SSL certificate file and SSL certificate files directory, respectively.

https://golang.org/pkg/crypto/x509/

I'm not very familiar with  SSL_CERT_FILE and SSL_CERT_DIR and the docs at https://golang.org/pkg/crypto/x509/ are a bit confusing - I think they are implying that the x509 lib will use these env vars automatically, is that right? Are we using those libs? i.e. how do we know for sure this works as expected? Maybe you could provide some links to some docs about how these env vars work?

/hold

remoe:
Why not support git resource as well with the same certificates? Because it throws:

```
git/git.go:41","msg":"Error running git [fetch --recurse-submodules=yes --depth=1 origin --update-head-ok --force master]: exit status 128\nfatal: unable to access 'https://example.local/test.git/': SSL certificate problem: self signed certificate in certificate chain\n","stacktrace":"github.com/tektoncd/pipeline/pkg/git.run\n\tgithub.com/tektoncd/pipeline/pkg/git/git.go:41\ngithub.com/tektoncd/pipeline/pkg/git.Fetch\n\tgithub.com/tektoncd/pipeline/pkg/git/git.go:116\nmain.main\n\tgithub.com/tektoncd/pipeline/cmd/git-init/main.go:52\nruntime.main\n\truntime/proc.go:203"}
```

remoe:
@FogDong , do you have tested this PR? I've tried it with v0.13.12 with following patch:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: tekton-pipelines-controller
  namespace: tekton-pipelines
spec:
  template:
    spec:
      serviceAccountName: tekton-pipelines-controller
      containers:
      - name: tekton-pipelines-controller
        volumeMounts:
        - name: ca-pemstore
          mountPath: /etc/config-registry-cert/certs.crt
          subPath: tls.crt
        env:
        - name: SSL_CERT_FILE
          value: /etc/config-registry-cert/certs.crt
      volumes:
        - name: ca-pemstore
          secret:
            secretName: all-certs
---
kind: Role
metadata:
  name: tekton-pipelines-controller
  namespace: tekton-pipelines
rules:
- apiGroups: [""]
  resources: ["secrets"]
  verbs: ["get"]
  resourceNames: ["all-certs"]
```

I still have the following error on TaskRun:

```example.local failed: Get https://example.local/v2/: x509: certificate signed by unknown authority```

**Comment by FogDong** [comment]
Hi, @bobcatfish 
I found source code [here](https://github.com/golang/go/blob/b13774691cfb22b1cb18263c9a35176c4a0e94b2/src/crypto/x509/root_unix.go)
On Unix systems other than macOS the environment variables SSL_CERT_FILE and SSL_CERT_DIR can be used to override the system default env. [link](https://github.com/golang/go/commit/b13774691cfb22b1cb18263c9a35176c4a0e94b2)

**Comment by chmouel** [comment]
/lgtm

Just need a rebase and i think we shoud be good to go!

**Comment by FogDong** [comment]
PTAL @chmouel 😉

**Comment by FogDong** [comment]
Can we cancel the hold label now? 🤔  @bobcatfish

**Comment by schue** [comment]
I am trying to test this pull request. I get 'net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x15\x03\x01\x00\x02\x02"' when referencing the local repository. I tried inlining the local cert both as plain text and as base64 encoded. What am I doing wrong?

**Comment by vdemeester** [comment]
> I'm not very familiar with SSL_CERT_FILE and SSL_CERT_DIR and the docs at https://golang.org/pkg/crypto/x509/ are a bit confusing - I think they are implying that the x509 lib will use these env vars automatically, is that right? Are we using those libs? i.e. how do we know for sure this works as expected? Maybe you could provide some links to some docs about how these env vars work?

Those are standard libraries, I'm pretty sure we can assume those are used.

@FogDong given that `SSL_CERT_FILE` *overrides* the system default env and that we set it (even if the configmap is empty), how does it act (in that case, where the configmap is empty) ? Is it still looking at the system (here container) certs ? If yes, I am full lgtm on this :wink:

**Comment by chmouel** [comment]
@vdemeester i think you right that would fail  : 


```
% curl https://icanhazip.com
2a01:e0a:252:1260:f842:40d6:9f26:eab2
% export SSL_CERT_FILE=/tmp/notpresent
% curl https://icanhazip.com
curl: (77) SSL: can't load CA certificate file /tmp/notpresent
```

FogDong:
> @vdemeester i think you right that would fail :
> 
> ```
> % curl https://icanhazip.com
> 2a01:e0a:252:1260:f842:40d6:9f26:eab2
> % export SSL_CERT_FILE=/tmp/notpresent
> % curl https://icanhazip.com
> curl: (77) SSL: can't load CA certificate file /tmp/notpresent
> ```

Seems like this `SSL_CERT_FILE` should be a optional env 🤔

schue:
I am using a self-signed cert and have this problem. This PR works for me.

vdemeester:
/lgtm cancel
We need a way to make this optional, and "merge" the user specified certificates and the "standards" one. I wonder how others do (like knative, …) ?

FogDong:
> We need a way to make this optional, and "merge" the user specified certificates and the "standards" one. I wonder how others do (like knative, …) ?

https://knative.dev/docs/serving/deploying/private-registry/

I'll give a try and re fix this PR.

vdemeester (commented):
@FogDong I think, if we also add `SSL_CERT_DIR` env variable to point to `/etc/ssl/certs/` it should probably work.

FogDong:
> @FogDong I think, if we also add `SSL_CERT_DIR` env variable to point to `/etc/ssl/certs/` it should probably work.

The `SSL_CERT_DIR` env will override the `SSL_CERT_FILE`, since the crt file will be under the `/etc/ssl/certs` so I think it's ok to accept the override and point that out in the doc. WDYT? @vdemeester

vdemeester:
> > @FogDong I think, if we also add `SSL_CERT_DIR` env variable to point to `/etc/ssl/certs/` it should probably work.
> 
> The `SSL_CERT_DIR` env will override the `SSL_CERT_FILE`, since the crt file will be under the `/etc/ssl/certs` so I think it's ok to accept the override and point that out in the doc. WDYT? @vdemeester

Looking at the [source code](https://golang.org/src/crypto/x509/root_unix.go#L43), it will take both (aka content of `SSL_CERT_FILE`, and any entries in the folders specified in `SSL_CERT_DIR`), so we can use both. `SSL_CERT_DIR` holds the default certs, `SSL_CERT_FILE` would allow the user to give its own.

FogDong:
Updated, add the `SSL_CERT_DIR` env and add some explanations in doc. PTAL @vdemeester 😉

sbwsg:
@bobcatfish @vdemeester are you happy for this to go in now that docs and changes have been made?

vdemeester (approved):
Indeed, I think that is a good start :angel: 
/lgtm

afrittoli:
@vdemeester @sbwsg @FogDong is there any reason for this to be held still?
It has bene approved and it looks to me like a great addition. This will only be used for self signed certificates, right?

sbwsg:
/hold cancel

Not as far as I'm aware.

bobcatfish:
Thanks again for adding those docs @FogDong !

vdemeester:
>  It has bene approved and it looks to me like a great addition. This will only be used for self signed certificates, right?

This could also be useful for proxy support.

PR #2802, opened 2020-06-10
Use permanent errors in the taskrun reconciler

### LINKED ISSUE #2474: [feature] Start using knative PermanentErrors in reconcilers

afrittoli:
# Changes

Instead of returning nil on error during reconcile, if the error
is transient return it, so that the key is requeue. If the error
is permanent return a permanent error, so that the key is not
requeued.

Fixes #2474

# Submitter Checklist

review them:

## Reviewer Notes

/kind cleanup


pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -1326,6 +1326,17 @@ func TestReconcile_SetsStartTime(t *testing.T) {
...
+
+	t.Logf("Creating SA %s in %s", "default", "foo")
+	if _, err := clients.Kube.CoreV1().ServiceAccounts("foo").Create(&corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "default",
+			Namespace: "foo",
+		},
+	}); err != nil {
+		t.Fatal(err)
+	}
```
vdemeester, line 1339: Why is this needed ? :thinking:

afrittoli: If you remove it the test will fail with missing service account.
It looks like returning the error exposes some hidden issues in existing tests.

vdemeester: ah ok :sweat_smile:

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -437,19 +438,21 @@ func (c *Reconciler) handlePodCreationError(ctx context.Context, tr *v1beta1.Tas
 			Reason:  podconvert.ReasonExceededResourceQuota,
 			Message: fmt.Sprintf("%s: %v", msg, err),
 		})
+		// return a transient error, so that the key is requeued
+		return err
```
zhujian7, line 442: Hi, @afrittoli  AFAIK, the key will be requeued by the `SetTaskRunTimer` function, I am curiously about if the key will be requeued twice if we returned a transient error here?

PR #2804, opened 2020-06-10
Use MarkStatus* helpers in pkg/pod/status

### LINKED ISSUE #2770: Make MarkResource* helpers for Pod similar to what we have for TaskRun and Pipelinerun
# Expected Behavior
Less boilerplate code in `pod/status.go`

# Actual Behavior

We repeat this code several times in small variations:
```
		trs.SetCondition(&apis.Condition{
			Status:  corev1.ConditionFalse,
			Reason:  v1beta1.TaskRunReasonFailed.String(),
			Message: msg,
		})
```
We could created helpers similar to what we did for TaskRun and PipelineRun, however since we do not own the `Pod` API we could just add them to `status.go` directly.

/kind cleanup

adshmh:
# Changes
Use MarkStatus* helper functions in pkg/pod/status.go
Closes #2770

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

### DIFF
--- a/pkg/pod/status.go
+++ b/pkg/pod/status.go
@@ -101,12 +101,7 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev
 	trs := &tr.Status
 	if trs.GetCondition(apis.ConditionSucceeded) == nil || trs.GetCondition(apis.ConditionSucceeded).Status == corev1.ConditionUnknown {
 		// If the taskRunStatus doesn't exist yet, it's because we just started running
-		trs.SetCondition(&apis.Condition{
-			Type:    apis.ConditionSucceeded,
-			Status:  corev1.ConditionUnknown,
-			Reason:  v1beta1.TaskRunReasonRunning.String(),
-			Message: "Not all Steps in the Task have finished executing",
-		})
+		MarkStatusRunning(trs, v1beta1.TaskRunReasonRunning.String(), "Not all Steps in the Task have finished executing")
 	}
 
 	trs.PodName = pod.Name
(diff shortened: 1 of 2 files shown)


pkg/pod/status.go
```diff
@@ -101,12 +101,7 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev
 	trs := &tr.Status
 	if trs.GetCondition(apis.ConditionSucceeded) == nil || trs.GetCondition(apis.ConditionSucceeded).Status == corev1.ConditionUnknown {
 		// If the taskRunStatus doesn't exist yet, it's because we just started running
-		trs.SetCondition(&apis.Condition{
-			Type:    apis.ConditionSucceeded,
-			Status:  corev1.ConditionUnknown,
-			Reason:  v1beta1.TaskRunReasonRunning.String(),
-			Message: "Not all Steps in the Task have finished executing",
-		})
+		MarkStatusRunning(trs, "", "")
```
vdemeester, line 104: We need to keep the message, don't we ?

pkg/pod/status.go
```diff
@@ -206,12 +191,7 @@ func updateCompletedTaskRun(trs *v1beta1.TaskRunStatus, pod *corev1.Pod) {
 func updateIncompleteTaskRun(trs *v1beta1.TaskRunStatus, pod *corev1.Pod) {
 	switch pod.Status.Phase {
 	case corev1.PodRunning:
-		trs.SetCondition(&apis.Condition{
-			Type:    apis.ConditionSucceeded,
-			Status:  corev1.ConditionUnknown,
-			Reason:  v1beta1.TaskRunReasonRunning.String(),
-			Message: "Not all Steps in the Task have finished executing",
-		})
+		MarkStatusRunning(trs, "", "")
```
vdemeester, line 194: Same here

pkg/pod/status.go
afrittoli: And the reason!

pkg/pod/status.go
adshmh: Thank you for the review. Fixed.

pkg/pod/status.go
adshmh: Thank you for the review. Fixed.

adshmh:
> Thanks for working on this!
> I agree we need to keep reason and message.
> 
> It seems worrying that all test pass with them missing, sounds like we lack some unit test coverage.
> As part of this, could you add some unit test coverage?

Thank you for the review. I reviewed the code again, the tests were passing because I was setting default values for `Reason` and `Message` fields in the helper function. I have updated the helper function's signature to fix this which was causing confusion. The tests seem to cover all cases involving helper functions.

pkg/pod/status.go
```diff
@@ -359,6 +334,33 @@ func getWaitingMessage(pod *corev1.Pod) string {
 	return "Pending"
 }
 
+func MarkStatusRunning(trs *v1beta1.TaskRunStatus, reason, message string) {
```
bobcatfish, line 337: as public exported functions, the convention is to include docstrings

adshmh: Thank you for the review. Fixed.

pritidesai:
This looks good and as an additional effort/PR, we could update many unit tests from the same test file `status_test.go` to use these helpers, e.g. running condition [here](https://github.com/tektoncd/pipeline/blob/master/pkg/pod/status_test.go#L36), failed condition [here](https://github.com/tektoncd/pipeline/blob/master/pkg/pod/status_test.go#L214),  etc instead of creating `apis.Condition` over and over.

/lgtm

PR #2805, opened 2020-06-11
Define a common function to exit ReconcileKind

afrittoli:
# Changes

Similar to what we did on the taskrun reconciler, define a function
finishReconcileUpdateEmitEvents that is always invoked when exiting
from ReconcileKind.

The function is responsible for dealing with label/annotation
updates as well as emitting events related to status changes.

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -129,9 +129,8 @@ var (
 // resource with the current status of the resource.
 func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun) pkgreconciler.Event {
 	logger := logging.FromContext(ctx)
-	recorder := controller.GetEventRecorder(ctx)
-	// Snapshot original for the label/annotation check below.
-	original := pr.DeepCopy()
+	// Read the initial condition
+	before := pr.Status.GetCondition(apis.ConditionSucceeded)
 
 	if !pr.HasStarted() {
 		pr.Status.InitializeConditions()
@@ -142,16 +141,9 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 		}
 		// start goroutine to track pipelinerun timeout only startTime is not set
 		go c.timeoutHandler.WaitPipelineRun(pr, pr.Status.StartTime)
-	} else {
-		pr.Status.InitializeConditions()
 	}
 
-	// In case of reconcile errors, we store the error in a multierror, attempt
-	// to update, and return the original error combined with any update error
-	var merr *multierror.Error
-
-	switch {
-	case pr.IsDone():
+	if pr.IsDone() {
 		// We may be reading a version of the object that was stored at an older version
 		// and may not have had all of the assumed default specified.
 		pr.SetDefaults(contexts.WithUpgradeViaDefaulting(ctx))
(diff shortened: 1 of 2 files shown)


sbwsg:
Looks good; suggest removing Release Notes section of PR but otherwise 👍 

/approve

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -142,80 +141,77 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
...
-			logger.Errorf("Reconcile error: %v", err.Error())
-			merr = multierror.Append(merr, err)
-		}
+	// Make sure that the PipelineRun status is in sync with the actual TaskRuns
+	err := c.updatePipelineRunStatusFromInformer(ctx, pr)
+	if err != nil {
+		// This should not fail. Return the error so we can re-try later.
+		logger.Errorf("Error while syncing the pipelinerun status: %v", err.Error())
+		return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
 	}
```
pritidesai, line 192: thanks @afrittoli, I am still trying to understand these changes, compared to `taskRun`,  we are missing `before` here before reconciling ... this `before` condition is used while updating emit events would it cause any conflicts in events states? 🤔 

```
	// Store the condition before reconcile
	before := tr.Status.GetCondition(apis.ConditionSucceeded)

	// Reconcile this copy of the task run and then write back any status
	if err = c.reconcile(ctx, tr, taskSpec, rtr); err != nil {
		logger.Errorf("Reconcile error: %v", err.Error())
	}
	// Emit events (only when ConditionSucceeded was changed)
	return c.finishReconcileUpdateEmitEvents(ctx, tr, before, err)
```

please feel free to ignore if its not relevant here.

afrittoli: There is also a `before` here, but it's earlier in the reconciler code. The reason being is that the code for sending the `start` even is not yet in for pipelineruns. Once I add that, I will need to move L133 further down in the reconcile code, similar to what taskrun does.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -142,80 +141,77 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
...
+}
+
+func (c *Reconciler) finishReconcileUpdateEmitEvents(ctx context.Context, pr *v1beta1.PipelineRun, beforeCondition *apis.Condition, previousError error) error {
+	recorder := controller.GetEventRecorder(ctx)
+	logger := logging.FromContext(ctx)
+
+	afterCondition := pr.Status.GetCondition(apis.ConditionSucceeded)
+	events.Emit(recorder, beforeCondition, afterCondition, pr)
+	_, err := c.updateLabelsAndAnnotations(pr)
+	if err != nil {
```
pritidesai, line 209: no issues here with these changes but this same check is missing from the taskrun `finishReconcileUpdateEmitEvents`:

```
	_, err := c.updateLabelsAndAnnotations(tr)
	events.EmitError(recorder, err, tr)
```

`taskrun` emits event without checking if err is `nil`, we can create a separate PR for that

pritidesai: Just read the [comments](https://github.com/tektoncd/pipeline/pull/2526#discussion_r424027814) from the taskrun PR, if `EmitError` has check for `nil`, should we drop the check from here like taskrun reconciler? Or do we still want to keep it?

pritidesai: if we drop `nil` check from here, no need for the `logger` as well ...

afrittoli: I added the check here because @bobcatfish mentioned before that she didn't like the check to be in `EmitError` only, so I was planning on changing that in the task run controller too (but in a different PR)

bobcatfish:
Looks like the release notes section should get removed

afrittoli:
Nice catch, thanks. It should be ok now

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: thanks @afrittoli

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -88,6 +88,8 @@ var _ taskrunreconciler.Interface = (*Reconciler)(nil)
 func (c *Reconciler) ReconcileKind(ctx context.Context, tr *v1beta1.TaskRun) pkgreconciler.Event {
 	logger := logging.FromContext(ctx)
 	recorder := controller.GetEventRecorder(ctx)
+	// Read the initial condition
+	before := tr.Status.GetCondition(apis.ConditionSucceeded)
```
bobcatfish, line 92: good call, no reason to be fetching that so many times if we can do it once :D

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -142,80 +141,77 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
...
+	// Reconcile this copy of the pipelinerun and then write back any status or label
+	// updates regardless of whether the reconciliation errored out.
+	if err = c.reconcile(ctx, pr); err != nil {
+		logger.Errorf("Reconcile error: %v", err.Error())
+	}
+
+	return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
+}
+
+func (c *Reconciler) finishReconcileUpdateEmitEvents(ctx context.Context, pr *v1beta1.PipelineRun, beforeCondition *apis.Condition, previousError error) error {
```
bobcatfish, line 202: i really like having one function to handle this!

PR #2808, opened 2020-06-11
Remove legacy Cluster Resource behavior left in place for the 0.14.0 …

### LINKED ISSUE #2694: Conceptual bug: Wrong resource names used

dlorenc:
…release.

# Changes

This is a cleanup commit that should go in after 0.14.0 is cut. It was introduced
in #2697, to fix #2694.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
The ClusterResource has been updated to write it's kubeconfig to the correct location on disk. Previously it was written to a location under /workspace/input/<resource ref name>, instead of the correct /workspace/input/<resource binding name>. If you hardcode resource paths in your tasks, you will need to update these. If you refer to paths by interpolation ( $(resources.inputs.<resource name>) ), you do not need to make any changes.

This changes was introduced in 0.13.0, but both paths were included. This removes the legacy path.
```

### DIFF
--- a/pkg/apis/resource/v1alpha1/cluster/cluster_resource.go
+++ b/pkg/apis/resource/v1alpha1/cluster/cluster_resource.go
@@ -20,11 +20,9 @@ import (
 	b64 "encoding/base64"
 	"encoding/json"
 	"fmt"
-	"path/filepath"
 	"strconv"
 	"strings"
 
-	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	resource "github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/names"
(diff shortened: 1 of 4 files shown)


bobcatfish:
Great, thanks @dlorenc !

/approve
/lgtm

Issue #2813, opened 2020-06-11
Completed steps within a pod will re-run if restarted for some reason


bobcatfish:
# Expected Behavior

Once a step in a TaskRun completes, it should not run again. If the entire pod was rescheduled before the TaskRun completed, perhaps all steps should start again, but if something triggers the re-creation of only containers within a pod, they should not run again.

# Actual Behavior

The entrypoint binary will wait for it's waitfile to exist, then run, then write the file that indicates it has completed. This means that if the container is restarted after completing, its waitfile will exist, and it will run again.

# Steps to Reproduce the Problem

I don't know how to reproduce it, but it happened while I was running the release pipeline for v0.13.0: https://dashboard.dogfooding.tekton.dev/#/namespaces/default/pipelineruns/pipeline-release-run-fwrpr

Check this out, it's crazy!!!

![image](https://user-images.githubusercontent.com/432502/84443498-b8035580-ac0d-11ea-9cee-edc674184a88.png)

[This is the task being run](https://github.com/tektoncd/pipeline/blob/master/tekton/publish.yaml)

Looking at the above, somehow _all steps have succeeded_ except `run-ko` and `git-source-tekton-pipelines-v0-13-1-v6pqm` - but `git-source-tekton-pipelines-v0-13-1-v6pqm` is one of the first containers to run!!! It's the PipelineResource injected container that gets the git source!! It's impossible that `build-push-base-images` could have succeeded for example without it running. Not to mention the logs for that container are vey telling:

```
k --context dogfood logs pipeline-release-run-fwrpr-publish-images-wwkf7-pod-g4lc5 -c step-git-source-tekton-pipelines-v0-13-1-v6pqm
{"level":"error","ts":1591815883.1945517,"caller":"git/git.go:41","msg":"Error running git [remote add origin https://github.com/tektoncd/pipeline]: exit status 128\nfatal: remote origin already exists.\n","stacktrace":"github.com/tektoncd/pipeline/pkg/git.run\n\tgithub.com/tektoncd/pipeline/pkg/git/git.go:41\ngithub.com/tektoncd/pipeline/pkg/git.Fetch\n\tgithub.com/tektoncd/pipeline/pkg/git/git.go:81\nmain.main\n\tgithub.com/tektoncd/pipeline/cmd/git-init/main.go:53\nruntime.main\n\truntime/proc.go:203"}
[...]

PR #2819, opened 2020-06-15
Fail step execution if the postFile already exists

ImJasonH:
This might indicate that the step was re-executed for some reason, which
indicates an internal error in step execution (#2813). Rather than risk
re-running a step in this case, we should fail loudly.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
Steps will fail if for some reason their postFile already exists before step is run; this might indicate some internal error where steps are re-run.
```

### DIFF
--- a/cmd/entrypoint/post_writer.go
+++ b/cmd/entrypoint/post_writer.go
@@ -20,3 +20,8 @@ func (*realPostWriter) Write(file string) {
 		log.Fatalf("Creating %q: %v", file, err)
 	}
 }
+
+func (*realPostWriter) Exists(file string) bool {
+	_, err := os.Stat(file)
+	return !os.IsNotExist(err)
+}
--- a/pkg/entrypoint/entrypointer.go
+++ b/pkg/entrypoint/entrypointer.go
@@ -78,6 +78,9 @@ type Runner interface {
 
 // PostWriter encapsulates writing a file when complete.
 type PostWriter interface {
+	// Exists returns whether the file exists.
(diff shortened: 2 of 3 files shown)


bobcatfish:
>  Rather than risk re-running a step in this case, we should fail loudly.

I'm not sure we need to fail loudly - at least in the scenario that happened to me, all of the steps completed successfully before getting "rescheduled" (or whatever mysterious thing happened) - and I ended up using the artifacts of that Run.

bobcatfish:
For https://github.com/tektoncd/pipeline/pull/2819#issuecomment-644111972 gonna add a hold in here:

/hold

I want to suggest an alternative that we silently exit and if possible log something. If some underlying thing decided to reschedule a container, but it has succeeded, since the entrypointer (afaik) controls execution, I think we could trust (famous last words?) all is well in the case that the file already exists.

I suppose something malicious could have written the file?

vdemeester:
> I want to suggest an alternative that we silently exit and if possible log something. If some underlying thing decided to reschedule a container, but it has succeeded, since the entrypointer (afaik) controls execution, I think we could trust (famous last words?) all is well in the case that the file already exists.

I tend to prefer to fail loudly at first and *then* see if this make sense or not. In that case, this is something we did not design around and/or we never anticipated that behavior, so… it feels like an error to me. Also, silently exit won't help us diagnose this (how much it happens, …) as we will never see it.

> I suppose something malicious could have written the file?

It could indeed.

PR #2821, opened 2020-06-15
Handle dag in pipelineresoultion

afrittoli:
# Changes

In the pipelinerun controller, today we follow this logic:
- build the dag from the spec, using the dag module
- resolve the pipelinerun state using the spec and status, using the
  resources module
- get a list of candidate next tasks from the dag module, passing part
  of the state
- get a list of next tasks from the resources module, using the list
  of candidates and the pipeline run status

The separation of concerns between the dag, resources and reconciler
modules feels a bit mixed up.

This is just a PoC that resolves part of the issue, by moving the
invocation of the dag building as well as obtaining the list of
candidates to the dag module, and aggregating the dag to the pipeline
state struct.

I did not update tests yet, this is for discussion purposes.
# Submitter Checklist

review them:

## Reviewer Notes

/kind cleanup

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation.go
@@ -173,7 +173,16 @@ func (ps *PipelineSpec) Validate(ctx context.Context) *apis.FieldError {
 
 	// Validate the pipeline task graph
 	if err := validateGraph(ps.Tasks); err != nil {
-		return apis.ErrInvalidValue(err.Error(), "spec.tasks")
+		ferr := apis.ErrInvalidValue(err.Error(), "spec.tasks")
+		ferr.Details = "Invalid Graph"
+		return ferr
+	}
+
+	// Validate the pipeline finally graph
+	if err := validateGraph(ps.Finally); err != nil {
+		ferr := apis.ErrInvalidValue(err.Error(), "spec.finally")
+		ferr.Details = "Invalid Graph"
+		return ferr
 	}
 
 	if err := validateParamResults(ps.Tasks); err != nil {
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -40,7 +40,6 @@ import (
 	"github.com/tektoncd/pipeline/pkg/contexts"
 	"github.com/tektoncd/pipeline/pkg/reconciler"
 	"github.com/tektoncd/pipeline/pkg/reconciler/events"
-	"github.com/tektoncd/pipeline/pkg/reconciler/pipeline/dag"
 	"github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun/resources"
 	"github.com/tektoncd/pipeline/pkg/reconciler/taskrun"
 	"github.com/tektoncd/pipeline/pkg/reconciler/volumeclaim"
@@ -284,31 +283,15 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 		pr.ObjectMeta.Annotations[key] = value
 	}
 
-	d, err := dag.Build(v1beta1.PipelineTaskList(pipelineSpec.Tasks))
-	if err != nil {
-		// This Run has failed, so we need to mark it as failed and stop reconciling it
-		pr.Status.MarkFailed(ReasonInvalidGraph,
-			"PipelineRun %s/%s's Pipeline DAG is invalid: %s",
-			pr.Namespace, pr.Name, err)
-		return controller.NewPermanentError(err)
-	}
-
-	// build DAG with a list of final tasks, this DAG is used later to identify
-	// if a task in PipelineRunState is final task or not
-	// the finally section is optional and might not exist
-	// dfinally holds an empty Graph in the absence of finally clause
-	dfinally, err := dag.Build(v1beta1.PipelineTaskList(pipelineSpec.Finally))
-	if err != nil {
-		// This Run has failed, so we need to mark it as failed and stop reconciling it
-		pr.Status.MarkFailed(ReasonInvalidGraph,
-			"PipelineRun %s's Pipeline DAG is invalid for finally clause: %s",
-			pr.Namespace, pr.Name, err)
-		return controller.NewPermanentError(err)
-	}
-
 	if err := pipelineSpec.Validate(ctx); err != nil {
 		// This Run has failed, so we need to mark it as failed and stop reconciling it
-		pr.Status.MarkFailed(ReasonFailedValidation,
+		var reason = ReasonFailedValidation
+		if err.Details == "Invalid Graph" {
+			// When the validation error reports tasks or finally in the error path
+			// we can use a more specific validation error reason
+			reason = ReasonInvalidGraph
+		}
+		pr.Status.MarkFailed(reason,
 			"Pipeline %s/%s can't be Run; it has an invalid spec: %s",
 			pipelineMeta.Namespace, pipelineMeta.Name, err)
 		return controller.NewPermanentError(err)
@@ -372,7 +355,7 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 	// pipelineState also holds a taskRun for each pipeline task after the taskRun is created
 	// pipelineState is instantiated and updated on every reconcile cycle
 	pipelineState, err := resources.ResolvePipelineRun(ctx,
-		*pr,
+		*pipelineSpec, *pr,
 		func(name string) (v1beta1.TaskInterface, error) {
 			return c.taskLister.Tasks(pr.Namespace).Get(name)
 		},
@@ -385,7 +368,7 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 		func(name string) (*v1alpha1.Condition, error) {
 			return c.conditionLister.Conditions(pr.Namespace).Get(name)
 		},
-		append(pipelineSpec.Tasks, pipelineSpec.Finally...), providedResources,
+		providedResources,
 	)
 
 	if err != nil {
@@ -413,7 +396,7 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 		return nil
 	}
 
-	for _, rprt := range pipelineState {
+	for _, rprt := range pipelineState.State() {
 		err := taskrun.ValidateResolvedTaskResources(rprt.PipelineTask.Params, rprt.ResolvedTaskResources)
 		if err != nil {
 			logger.Errorf("Failed to validate pipelinerun %q with error %v", pr.Name, err)
@@ -452,11 +435,11 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 		return controller.NewPermanentError(err)
 	}
 
-	if err := c.runNextSchedulableTask(ctx, pr, d, dfinally, pipelineState, as); err != nil {
+	if err := c.runNextSchedulableTask(ctx, pr, *pipelineState, as); err != nil {
 		return err
 	}
 
-	after := resources.GetPipelineConditionStatus(pr, pipelineState, logger, d, dfinally)
+	after := resources.GetPipelineConditionStatus(*pr, *pipelineState, logger)
 	switch after.Status {
 	case corev1.ConditionTrue:
 		pr.Status.MarkSucceeded(after.Reason, after.Message)
@@ -467,39 +450,36 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 	}
 	// Read the condition the way it was set by the Mark* helpers
 	after = pr.Status.GetCondition(apis.ConditionSucceeded)
-	pr.Status.TaskRuns = getTaskRunsStatus(pr, pipelineState)
+	pr.Status.TaskRuns = getTaskRunsStatus(pr, pipelineState.State())
 	logger.Infof("PipelineRun %s status is being set to %s", pr.Name, after)
 	return nil
 }
 
-// runNextSchedulableTask gets the next schedulable Tasks from the dag based on the current
-// pipeline run state, and starts them
-// after all DAG tasks are done, it's responsible for scheduling final tasks and start executing them
-func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.PipelineRun, d *dag.Graph, dfinally *dag.Graph, pipelineState resources.PipelineRunState, as artifacts.ArtifactStorageInterface) error {
-
+// runNextSchedulableTask gets the next schedulable Tasks based the pipeline run state and starts them
+func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.PipelineRun, pipelineState resources.ResolvedPipelineRun, as artifacts.ArtifactStorageInterface) error {
 	logger := logging.FromContext(ctx)
 	recorder := controller.GetEventRecorder(ctx)
 
-	var nextRprts []*resources.ResolvedPipelineRunTask
+	var (
+		nextRprts []*resources.ResolvedPipelineRunTask
+		err       error
+	)
 
 	// when pipeline run is stopping, do not schedule any new task and only
 	// wait for all running tasks to complete and report their status
-	if !pipelineState.IsStopping(d) {
-		// candidateTasks is initialized to DAG root nodes to start pipeline execution
-		// candidateTasks is derived based on successfully finished tasks and/or skipped tasks
-		candidateTasks, err := dag.GetSchedulable(d, pipelineState.SuccessfulOrSkippedDAGTasks(d)...)
+	if !pipelineState.IsStopping() {
+		// nextRprts holds a list of pipeline tasks which should be executed next
+		nextRprts, err = pipelineState.GetNextTasks()
 		if err != nil {
 			logger.Errorf("Error getting potential next tasks for valid pipelinerun %s: %v", pr.Name, err)
 			return controller.NewPermanentError(err)
 		}
-		// nextRprts holds a list of pipeline tasks which should be executed next
-		nextRprts = pipelineState.GetNextTasks(candidateTasks)
 	}
 
 	// GetFinalTasks only returns tasks when a DAG is complete
-	nextRprts = append(nextRprts, pipelineState.GetFinalTasks(d, dfinally)...)
+	nextRprts = append(nextRprts, pipelineState.GetFinalTasks()...)
 
-	resolvedResultRefs, err := resources.ResolveResultRefs(pipelineState, nextRprts)
+	resolvedResultRefs, err := resources.ResolveResultRefs(pipelineState.State(), nextRprts)
 	if err != nil {
 		logger.Infof("Failed to resolve all task params for %q with error %v", pr.Name, err)
 		pr.Status.MarkFailed(ReasonFailedValidation, err.Error())
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -451,8 +451,8 @@ func TestReconcile_InvalidPipelineRuns(t *testing.T) {
 		))),
 		tb.PipelineRun("pipeline-invalid-final-graph", tb.PipelineRunNamespace("foo"), tb.PipelineRunSpec("", tb.PipelineRunPipelineSpec(
 			tb.PipelineTask("dag-task-1", "taskName"),
-			tb.FinalPipelineTask("final-task-1", "taskName"),
-			tb.FinalPipelineTask("final-task-1", "taskName")))),
+			tb.FinalPipelineTask("final-task-1", "taskName", tb.RunAfter("dag-task-2")),
+		))),
 	}
 
 	d := test.Data{
@@ -1741,12 +1741,14 @@ func TestReconcileWithConditionChecks(t *testing.T) {
 		expectedConditionChecks[index] = makeExpectedTr(condition.Name, ccNames[condition.Name], condition.Labels, condition.Annotations)
 	}
 
-	// Check that the expected TaskRun was created
-	condCheck0 := clients.Pipeline.Actions()[1].(ktesting.CreateAction).GetObject().(*v1beta1.TaskRun)
-	condCheck1 := clients.Pipeline.Actions()[2].(ktesting.CreateAction).GetObject().(*v1beta1.TaskRun)
-	if condCheck0 == nil || condCheck1 == nil {
-		t.Errorf("Expected two ConditionCheck TaskRuns to be created, but it wasn't.")
+	// Check that the expected TaskRun were created
+	actions := clients.Pipeline.Actions()
+	if !actions[1].Matches("create", "taskruns") || !actions[2].Matches("create", "taskruns") {
+		t.Fatalf("Expected two ConditionCheck TaskRuns to be created, got instead %d actions: %#v", len(actions), actions)
 	}
+
+	condCheck0 := actions[1].(ktesting.CreateAction).GetObject().(*v1beta1.TaskRun)
+	condCheck1 := actions[2].(ktesting.CreateAction).GetObject().(*v1beta1.TaskRun)
 
 	actual := []*v1beta1.TaskRun{condCheck0, condCheck1}
 	if d := cmp.Diff(actual, expectedConditionChecks); d != "" {
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -19,7 +19,6 @@ package resources
 import (
 	"context"
 	"fmt"
-	"reflect"
 	"strconv"
 
 	"go.uber.org/zap"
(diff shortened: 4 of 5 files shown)


bobcatfish:
it feels like pipelinerunresolution already has so many responsibilities, im not sure that giving it another one will improve it - maybe we need a new package with new responsibilities?

I also think it might be worth stepping back and taking a look at the data structures we are using; i often find that when code is getting complex and confusing, changing the shapes of the data structures being used can really help (e.g. PipelineRunState might not be the best data structure to be passing around going foward)

bobcatfish:
sorry @afrittoli after looking at https://github.com/tektoncd/pipeline/pull/2661 some more, i do agree with moving the call to the DAG logic into the resolution logic

I do still think that 1) pipelineresolution does too much and needs to be broken up and that 2) the data structures need to change

but i agree this is an improvement!

afrittoli:
Thanks for the review. 
I also feel there is some space to simplify data structures. I also wonder if resolved resource could be abstracted out of task run and pipeline run. Maybe they're to different to share code? But it may help generalising main controller code.

bobcatfish:
> I also wonder if resolved resource could be abstracted out of task run and pipeline run. Maybe they're to different to share code? 

That sounds amazing!

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -73,9 +79,13 @@ type ResolvedPipelineRunTask struct {
 	ResolvedConditionChecks TaskConditionCheckState // Could also be a TaskRun or maybe just a Pod?
 }
 
-// PipelineRunState is a slice of ResolvedPipelineRunTasks the represents the current execution
+// PipelineRunState is the combination of the Pipeline dag and
+// a slice of ResolvedPipelineRunTasks the represents the current execution
 // state of the PipelineRun.
-type PipelineRunState []*ResolvedPipelineRunTask
+type PipelineRunState struct {
+	Dag		*dag.Graph
```
pritidesai, line 86: @afrittoli what do you think about adding a state matrix here in `PipelineRunState` which will help simplify `GetPipelineConditionStatus`?

I had implemented this some time back adding run state matrix in the `PipelineRunStatus`
https://github.com/tektoncd/pipeline/pull/2821/files#diff-4e301064fef70f3c0db810267c0959f2R85 
But now, it would fit well in this data structure.

afrittoli: @pritidesai the link points to your comment on this PR, perhaps you meant to share a link to a different PR?

pritidesai: 😲  sorry @afrittoli have updated the comment ⬆️

afrittoli: Thanks!

afrittoli:
@bobcatfish @pritidesai I've tried to keep this PR reasonable in size, so some of the refactors here are still not included.

The main objectives here are to move the dag handling into pipelinerunresolution, and to ensure the various helpers do what their name mean, so that the code hopefully is more readable and maintainable.

Things that are missing (maybe more), to be handled in separate PRs:
- We defined `IsStarted` and `IsCancelled` but through the code sometimes we use them, sometime we use part of the condition the include, sometimes we even use something different, which needs to be fixed (see `GetNextTasks`)
- `GetNextTasks` and `GetFinalTasks` can be combined, so that the pipelinerun controller only need to ask for the next tasks, and all the logic is encapsulated in `pipelinerunresource`

One thing that I noticed is that today we:
- build the pipeline state
- get schedulable tasks and start them
- build the pipeline condition from the original state
- build the pipeline status from the original state

I wonder if we should change the order to this instead:
- build the pipeline state
- build the pipeline condition from the original state
- build the pipeline status from the original state
- get schedulable tasks and start them

Just to make it clear that the new taskruns does not have an impact on the state reported in the current reconcile cycle.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -284,31 +283,15 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
...
-			"PipelineRun %s's Pipeline DAG is invalid for finally clause: %s",
-			pr.Namespace, pr.Name, err)
-		return controller.NewPermanentError(err)
-	}
-
 	if err := pipelineSpec.Validate(ctx); err != nil {
 		// This Run has failed, so we need to mark it as failed and stop reconciling it
-		pr.Status.MarkFailed(ReasonFailedValidation,
+		var reason = ReasonFailedValidation
+		if err.Details == "Invalid Graph" {
```
afrittoli, line 289: Because the graph is created in `pipelinerunresources`, validation happens first. In this context it was not clear that the error was an invalid graph one, so I added some extra context in the error to be able to catch it and use the specific reason. 
The alternatives are:
- skip validation here, but that's only a temp solution
- move validation into `ResolvePipelineRun` - which might make sense, but not in this PR
- simply return a validation error and rely on the error stack to provide more insight to the user
I wonder if we really need to run validation here, but I guess we will need to do once tasks / pipelines may come from a container registry or git url.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -451,8 +451,8 @@ func TestReconcile_InvalidPipelineRuns(t *testing.T) {
 		))),
 		tb.PipelineRun("pipeline-invalid-final-graph", tb.PipelineRunNamespace("foo"), tb.PipelineRunSpec("", tb.PipelineRunPipelineSpec(
 			tb.PipelineTask("dag-task-1", "taskName"),
-			tb.FinalPipelineTask("final-task-1", "taskName"),
-			tb.FinalPipelineTask("final-task-1", "taskName")))),
```
afrittoli, line 455: This error does not trigger an invalid dag when going through validation, it fails first on `validatePipelineTaskName`

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -1741,13 +1741,15 @@ func TestReconcileWithConditionChecks(t *testing.T) {
...
 	}
 
-	// Check that the expected TaskRun was created
-	condCheck0 := clients.Pipeline.Actions()[1].(ktesting.CreateAction).GetObject().(*v1beta1.TaskRun)
-	condCheck1 := clients.Pipeline.Actions()[2].(ktesting.CreateAction).GetObject().(*v1beta1.TaskRun)
-	if condCheck0 == nil || condCheck1 == nil {
-		t.Errorf("Expected two ConditionCheck TaskRuns to be created, but it wasn't.")
+	// Check that the expected TaskRun were created
+	actions := clients.Pipeline.Actions()
+	if !actions[1].Matches("create", "taskruns") || !actions[2].Matches("create", "taskruns") {
```
afrittoli, line 1746: Doing a cast without checking first leads to an un-managed failure in the tests when the test fails.
My code changes initially made this test failing and the error was not very informative.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -150,25 +192,58 @@ func (state PipelineRunState) ToMap() map[string]*ResolvedPipelineRunTask {
...
+		if !t.IsDone() {
+			return false
 		}
 	}
-	return
+	return true
+}
+
+// isFinallyDone returns true if all final tasks are done
+func (rpr ResolvedPipelineRun) isFinallyDone() bool {
```
afrittoli, line 230: This is only used as part of `IsDone` below, so I could move it in there... I thought it was more readable like this though.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -177,31 +252,36 @@ func (state PipelineRunState) IsBeforeFirstTaskRun() bool {
...
-			if t.IsFailure() {
-				return true
-			}
-		}
-	}
-	return false
+// at least one task already failed or was cancelled in the specified dag, but there are
+// still tasks to be run in the DAG
+func (rpr ResolvedPipelineRun) IsStopping() bool {
+	return rpr.DagState.HasFailedOrCancelled() && !rpr.isDAGDone()
```
afrittoli, line 258: There is a key change here - `IsStopping` will return true only as long at something is running in the main DAG.
As soon as the DAG is complete it will turn false, and will stay false while finally task are running.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -524,165 +524,6 @@ func DagFromState(state PipelineRunState) (*dag.Graph, error) {
 }
 
 func TestGetNextTasks(t *testing.T) {
-	tcs := []struct {
-		name         string
-		state        PipelineRunState
-		candidates   map[string]struct{}
-		expectedNext []*ResolvedPipelineRunTask
-	}{{
-		name:         "no-tasks-started-no-candidates",
```
afrittoli, line 533: Many of these use cases used an impossible state / candidates combination.
Now that candidates are calculated within `GetNextTasks` it's not possible anymore to force such impossible scenarios.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -741,76 +582,100 @@ func TestGetNextTaskWithRetries(t *testing.T) {
...
+		TaskRun:      withCancelled(makeFailed(trs[0])),
 		ResolvedTaskResources: &resources.ResolvedTaskResources{
 			TaskSpec: &task.Spec,
 		},
 	}}
 
+	// The cancelled by spec here has no impact on how this is treated by any
+	// of the functions in pipelineresolution. The `makeRetried` is what triggers
+	// the expected behaviour.
+	// TODO: remove withCancelledBySpec tests as misleading
```
afrittoli, line 678: I find this test misleading and it's not really testing anything to do with `withCancelledBySpec`.
It does test that a failure when retries are exhausted marks the task and DAG as done, so perhaps we could remove the `withCancelledBySpec` and rename the test?

afrittoli:
/hold
I want to add some more tests to ensure coverage is not going down.
As a refactor this does not change any behaviour, so I'd like the coverage level to stay where it was or improve.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -212,82 +292,37 @@ func (state PipelineRunState) GetNextTasks(candidateTasks map[string]struct{}) [
...
+		return tasks
+	}
+	// return list of tasks with all final tasks
+	for _, t := range rpr.FinalState {
+		// Final states cannot be skipped, so it's enough to check is a Task is not started
+		// and not done in case of failure
+		if t.TaskRun == nil {
+			tasks = append(tasks, t)
 		}
-		tasks = state.GetNextTasks(finalCandidates)
```
pritidesai, line 290: I haven't reviewed all the changes yet so I might be wrong but wanted to make sure we are covering cases where a final task fails and need to reschedule for another retry. Just checking `TaskRun` not nil here is not sufficient. If a final task has failed, it will not be rescheduled even if the retries are still left, please correct me if I am wrong 🙏

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -376,15 +411,57 @@ func ValidateServiceaccountMapping(p *v1beta1.PipelineSpec, pr *v1beta1.Pipeline
...
+	return &ResolvedPipelineRun{
+		Dag:        d,
+		Final:      final,
+		DagState:   state,
+		FinalState: finalState,
+		WontRun:    skippedTasks(state, d),
+	}, nil
+}
+
+func resolvePipelineRun(
```
pritidesai, line 455: We had this similar function implemented in first version of finally [PR](https://github.com/tektoncd/pipeline/pull/2437/files#diff-4e301064fef70f3c0db810267c0959f2R367).

This was simplified by merging both states (pipelineRunState) into one.

afrittoli: My impression after working on this refactor is that overall it's much simpler like this. 

Having the dag handled in `pipelinerunresolution` allows us to have the best of both worlds:
- we can expose a single state to the reconciler, via the `State()` function
- we can work internally with two separated states, which makes a lot of the functions more natural, it removes the need to check which state is in which dag which we had to do in many places. 

The unit test coverage still need some work, as it might be best to target tests to `resolvePipelineRun` and only add a couple of tests for logic in `ResolvePipelineRun`.

pritidesai: I understand @afrittoli the supporting functions are more natural this way but how will we address a scenario with a `finally` task having an input `PipelineResource` which refers to the output of a `DAG` task when resolved DAG tasks are stored in `DAGState` and resolved `finally` tasks are in `FinalState`. And same for task results.

Right now, `DAG` tasks and `finally` tasks are independent and have no reference from one to the other but at some point we need to address [these](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#known-limitations) limitations.

pritidesai: e.g.

```
func convertParamsToResultRefs(pipelineRunState PipelineRunState, target *ResolvedPipelineRunTask) (ResolvedResultRefs, error) {
```
and 

```
func resolveResultRef(pipelineState PipelineRunState, resultRef *v1beta1.ResultRef) (*ResolvedResultRef, error) {
```

pritidesai: I see, you have `State()` [here](https://github.com/tektoncd/pipeline/pull/2821/files#diff-0e0c01fdbdc46d2b6054a706f99516d0R482)

afrittoli: :)

Besides, I feel that the handling of results as it's implemented today kind of breaks the abstraction of resolved pipeline tasks and runs. It might be good to move the result resolution into the resources module too.
The fact that we need to reconcile pipeline results in the `isDone` after completion time is already set, is an example of the issue. When we start sending cloud events for pipeline run events, this will cause the pipeline results to be missing in the pipeline run spec sent in the event.

pritidesai: Yup I agree and we have the same issue with task results (task results are set after task run is declared successful, see PR #2538).

Moving the result resolution into the resources module could be next in refactoring queue.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -150,25 +195,55 @@ func (state PipelineRunState) ToMap() map[string]*ResolvedPipelineRunTask {
 	return m
 }
 
-// IsDone returns true when all pipeline tasks have respective taskRun created and
-// that taskRun has either succeeded or failed after all possible retry attempts
-func (state PipelineRunState) IsDone() (isDone bool) {
-	isDone = true
+// HsFailedOrCancelled returns true the state includes at least one failed or one cancelled task
```
vdemeester, line 198: `s/HsFailedOrCancelled/HasFailedOrCancelled`

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -173,7 +173,16 @@ func (ps *PipelineSpec) Validate(ctx context.Context) *apis.FieldError {
...
 	// Validate the pipeline task graph
 	if err := validateGraph(ps.Tasks); err != nil {
-		return apis.ErrInvalidValue(err.Error(), "spec.tasks")
+		ferr := apis.ErrInvalidValue(err.Error(), "spec.tasks")
+		ferr.Details = "Invalid Graph"
+		return ferr
+	}
+
+	// Validate the pipeline finally graph
+	if err := validateGraph(ps.Finally); err != nil {
```
pritidesai, line 182: This check should not be added here. `finally` section is a list of tasks not graph from users perspective. Tasks are validated to not include `runAfter` [here](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1beta1/pipeline_validation.go#L447). This check will result in confusion. I would rather have the checks be done explicitly for `runAfter` the way its implemented.

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: @afrittoli `pipelineSpec` is being validated here for many different checks in addition to `invalidGraph`. I think we do need to run validation here, I dont see validation happening else where, am I missing something? 🤔

pritidesai: We could certainly move the check on invalid graph to `ResolvePipelineRun`. 

Also, such reason could be generated where the error is getting created in `pipeline_validation.go` instead of calculating here. Its a great idea to set the reason to specific validation error in addition to returning `error`.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -413,7 +396,7 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 		return nil
 	}
 
-	for _, rprt := range pipelineState {
+	for _, rprt := range pipelineState.State() {
 		err := taskrun.ValidateResolvedTaskResources(rprt.PipelineTask.Params, rprt.ResolvedTaskResources)
```
pritidesai, line 400: This check `ValidateResolvedTaskResources` can be moved to `ResolvePipelineRun()`.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -467,39 +450,36 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
...
+		err       error
+	)
 
 	// when pipeline run is stopping, do not schedule any new task and only
 	// wait for all running tasks to complete and report their status
-	if !pipelineState.IsStopping(d) {
-		// candidateTasks is initialized to DAG root nodes to start pipeline execution
-		// candidateTasks is derived based on successfully finished tasks and/or skipped tasks
-		candidateTasks, err := dag.GetSchedulable(d, pipelineState.SuccessfulOrSkippedDAGTasks(d)...)
+	if !pipelineState.IsStopping() {
```
pritidesai, line 470: this check can be moved in the `GetNextTasks`

pkg/apis/pipeline/v1beta1/pipeline_validation.go
afrittoli: The only reason I added the check here was to be consistent with the current behaviour. We have a unit test right now that enforces emitting an `InvalidDagError` to the user in case the finally DAG is not valid, and the only way to keep that with the new code org was to add it here.

I'd be happy to drop this and drop the unit test and just bubble up any validation error as it's reported by the dag module directly.

pkg/reconciler/pipelinerun/pipelinerun.go
afrittoli: > @afrittoli `pipelineSpec` is being validated here for many different checks/conditions in addition to `invalidGraph`. I think we do need to run validation here, I dont see validation happening else where, am I missing something? 🤔

Validation is done by the webhook when resources are submitted to etcd.
However it will not be performed for resources that are not stored in etcd like tasks or pipelines coming from an external source (git/registry). I'm not sure about embedded specs - I'd expect to go through the same validation as normal resources.

afrittoli: > We could certainly move the check on invalid graph to `ResolvePipelineRun`.

That's the way I implemented it originally, but because we validate the spec here, we don't really catch the error in `ResolvePipelineRun`. The only way would be to move the whole validation of the spec into `ResolvePipelineRun`, which is something we could do eventually, but I didn't want to do it here as well.

> 
> Also, such reason could be generated where the error is getting created in `pipeline_validation.go` instead of calculating here. Its a great idea to set the reason to specific validation error in addition to returning `error`.

pkg/reconciler/pipelinerun/pipelinerun.go
afrittoli: Yeah, it sounds like something we good do. Perhaps in different PR?

pkg/reconciler/pipelinerun/pipelinerun.go
afrittoli: I think this check is already redundant now, but I need to try it out, and I was planning to do so in a next PR, along with giving a single NextTasks method for the pipeline controller to use.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: Heh, nice catch

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: You're right, the failed retries needs to be covered for finally too. 
If that is something we want to have - which I think it's the case - I'll add test coverage for it and make sure it's covered here.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
pritidesai: yes please, will add it back when we change `finally` to be graph instead of list, thanks 🙏

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: yup definitely, not everything has to be part of single PR

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: sounds good, thanks 👏

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -77,6 +76,29 @@ type ResolvedPipelineRunTask struct {
...
+
+	// WontRun is a hash of PipelineTask names that won't be executed because they are
+	// not reachable via the DAG due to the pipeline state, or had failed conditions
+	WontRun map[string]struct{}
+}
+
+// IsDone returns true if a PipelineTask was successful or failed all retries
+// NOTE: a PipelineTask that has been cancelled is not always treated as "done",
+// the condition must be checked separately
 func (t ResolvedPipelineRunTask) IsDone() (isDone bool) {
```
pritidesai, line 102: named return value? 😉

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -102,7 +124,7 @@ func (t ResolvedPipelineRunTask) IsSuccessful() bool {
 	return c.Status == corev1.ConditionTrue
 }
 
-// IsFailure returns true only if the taskrun itself has failed
+// IsFailure returns true only if the taskrun itself has failed and retries are exausted
```
pritidesai, line 127: s/exausted/exhausted

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -113,7 +135,18 @@ func (t ResolvedPipelineRunTask) IsFailure() bool {
 	return c.IsFalse() && retriesDone >= retries
 }
 
-// IsCancelled returns true only if the taskrun itself has cancelled
+// HasFailedWithRetriesLeft returns true only if the taskrun itself has failed and retries are not exausted
```
pritidesai, line 138: same as above: s/exausted/exhausted

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: yes we do if we don't have tests for them, please add or let me know if you want me to add them 🙏

pritidesai:
Thanks @afrittoli for the changes, I am still reviewing them, have left a few comments ⬆️ . Will resume review after the conflicts are resolved otherwise git diff is misleading.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: Since we have removed `GetNextTasks` from here, you will have to add separate tests for final tasks. I still think it would be easier to create a common function `getNextTasks` which we can unit test irrespective of DAG tasks or final tasks. `getNextTasks` can be used by both `GetNextTasks` and `GetFinalTasks`.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: I still have to rebase this on top of your last PRs

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: Good catch, thanks

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: 😿

bobcatfish:
@afrittoli do you have any strong feelings about keeping this open vs closing it for now?

afrittoli:
I really want to have time to hack on the controller again :D but it's ok to close for now

Issue #2748, opened 2020-06-03
Define compatibility policy for Go libs


bobcatfish:
# Expected Behavior

We try to minimize the impact of changes to our Go libs on other projects that are using them, for example in https://github.com/tektoncd/pipeline/pull/2717#issuecomment-637164266 when @pritidesai noted the impact of that change on the CLI.

It would make sense to have a clear policy around:
- How to make backwards incompatible changes to the libs and what expectations should be around this for users of the libs
- What is and is not part of the lib (people can currently import anything from tektoncd/pipeline except the builders in internal)

# Actual Behavior

We have our API compatibility policy outlined in https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md but it does not currently include our client libs.

# Additional Info

If we can identify which parts of the code we think it is reasonable for folks to import and use as libs, we can move the rest into [internal](https://golang.org/doc/go1.4#internalpackages) and that would make it very clear which code you can change without impacting other projects.

pritidesai:
/remove-lifecycle stale
Defining such policy is crucial as we are approaching `v1`.

wlynch:
/reopen

Wanted to re-raise this issue for discussion (especially as we near v1) - I think it would be useful to apply deprecation policies to client libraries.

For Go in particular, a breaking change to a minor version best case means that tools like dependabot can break and manual intervention is needed by humans to upgrade, and worst case means that tools build on top of Tekton may be incompatible with each other (i.e. if 2 separate libraries depend on different client versions, [minimal version selection](https://go.dev/ref/mod#minimal-version-selection) may force the selection of a non-compatible dependency). We can still make rapid changes to alpha APIs / clients, but for stable APIs it would be great to have more guarantees around stability.

dibyom:
Planning to incorporate this in #5633

PR #2826, opened 2020-06-17
adding metadata to taskSpec in PipelineTask

### LINKED ISSUE #2767: Add metadata to PipelineTask to specific additional labels and annotations
Update issue title and description on 6/9/2020:

---

> Metadata is not part of the taskspec, tasks contain both a taskspec to define the behavior and metadata to identify the task k8s resource. When a taskspec is embedded in a pipelinespec instead of referencing a task, their is no separate task k8s resource so we don't need task specific metadata since it wouldn't make sense.

We need to have support for setting different annotations or metadata on the TaskRun created from embedded TaskSpec define in Pipelines (from the Pipeline Label and Annotation), am I right ?

I could see having the following:
```YAML
kind: PipelineRun
metadata:
  name: pipelinerun-with-taskspec-to-echo-greetings
spec:
  pipelineSpec:
    params:
    - […]
    tasks:
    - name: echo-greetings
      medata:
        labels: [ …]  - name: GREETINGS
        description: "Greetings, default is Hello World!"
        type: string
        default: "Hello World!"
        annotations: […]
      taskSpec:
        […]
```
It should be added on all embedded spec types. It wouldn't break backward compatibility.

---

**Original Issue Question**: 
> Unsupported features for embedded pipelineSpec, taskSpec?

Can any Tekton YAML that contains separate "documents" for `Task`s, `Pipeline`, `PipelineRun` be converted to a YAML file containing only one `PipelineRun` "document" with embedded `pipelineSpec` which embeds multiple `taskSpec`s?

One thing that seems to get lost are Task `metadata` like pod label annotations.

**Related issues**:
https://github.com/tektoncd/pipeline/issues/872
https://github.com/tektoncd/pipeline/pull/1333
https://github.com/tektoncd/pipeline/pull/1554

**Answer from @vdemeester**:
> AFAIK there should be no lost capabilities from embedding resource specs instead of referencing separate resources. It changes which k8s resources the data is stored in, but all the same specs are provided.

pritidesai:
# Changes

Adding `metadata` to `taskSpec` within `PipelineTask` to allow specifying labels and annotations when a task is embedded using `taskSpec`. These labels and annotations are propagated to `taskRun` and then to the pods.

```
kind: PipelineRun
metadata:
  name: pipelinerun-with-taskspec-to-echo-greetings
spec:
  pipelineSpec:
    tasks:
      - name: echo-greetings
         taskSpec:
           metadata:
             labels: [ …]
             annotations: [...]
         steps:
...
```

Metadata is already supported as part of Tasks and Pipelines while respective CRDs are created. But was not possible to specify with embedded resources.

These labels are commonly used to identify and filter pods for further actions (such as collecting pod metrics, and cleaning up completed pod with certain labels, etc) even being part of one single Pipeline.

```
  pipelineSpec:
    tasks:
    - name: echo
      metadata:
        labels:
          pipeline-sdk-type: kfp
      taskSpec:
       ...
    - name: echo2
      metadata:
        labels:
          pipeline-sdk-type: tfx
      taskSpec:
       ...
```

Closes #2767 

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Pipeline authors can now specify metadata while embedding tasks (using taskSpec) into their pipeline.
```


examples/v1beta1/pipelineruns/pipelinerun-with-pipelinespec-and-taskspec.yaml
```diff
@@ -6,6 +6,9 @@ spec:
   pipelineSpec:
     tasks:
       - name: echo-good-morning
+        metadata:
```
jlpettersson, line 9: this is its own field now? so it will not necessarily be written "just above" `taskSpec`. If we could have the `Task` as a field, optional metadata would work out of the box, as with `volumeClaimTemplate` where I can declare a `PersistentVolumeClaim` **with** or **without** metadata, https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1beta1/workspace_types.go#L61
I would prefer that.

I also would like to be able to declare `PipelineRun` like that in `TriggerTemplate` ...so that I can set cutom labels, e.g. for commit-id.

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -93,6 +95,9 @@ type PipelineResult struct {
 // PipelineTask defines a task in a Pipeline, passing inputs from both
 // Params and from the output of previous tasks.
 type PipelineTask struct {
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
```
jlpettersson, line 100: If we had `Task *v1beta1.Task` instead of (or in addition to - for compatibility) the current `TaskSpec *TaskSpec`, then `metadata` would be included as optional - by design

...as how I understand from the `volumeClaimTemplate`.

jlpettersson: but this does not work for `TaskRef` if we want a "custom" metadata - but if we want to use the metdata as-is, I think it should work?

bobcatfish:
I think this might require a TEP - since it's an API change, and also this is introducing more k8s specific stuff into a PipelineTask - I'd at least like to discuss it before adding it

bobcatfish:
looks like @vdemeester didnt feel it required a TEP (https://github.com/tektoncd/pipeline/issues/2767#issuecomment-641416949) that's ok with me, but could we at least discuss this at the API working group and talk about why we need it?

pritidesai:
> looks like @vdemeester didnt feel it required a TEP ([#2767 (comment)](https://github.com/tektoncd/pipeline/issues/2767#issuecomment-641416949)) that's ok with me, but could we at least discuss this at the API working group and talk about why we need it?

yup, will add it for discussion in the API group 🙏

Tomcli:
Here is one of our use cases with this new API:
With the current API, we can have two tasks in our pipeline that use the same label key:
```yaml
kind: Task
metadata:
  labels:
    pipelines.kubeflow.org/pipeline-sdk-type: kfp
  name: echo
spec:
  steps:
  - args:
    - echo hello world
    command:
    - sh
    - -c
    image: library/bash:4.4.23
    name: main
---
kind: Task
metadata:
  labels:
    pipelines.kubeflow.org/pipeline-sdk-type: tfx
  name: echo2
spec:
  steps:
  - args:
    - echo hello world
    command:
    - sh
    - -c
    image: library/bash:4.4.23
    name: main
---
kind: Pipeline
metadata:
  name: sequential-pipeline
spec:
  tasks:
  - name: echo
    taskRef:
      name: echo
  - name: echo2
    runAfter:
    - echo
    taskRef:
      name: echo2
---
kind: PipelineRun
metadata:
  name: sequential-pipeline-run
spec:
  pipelineRef:
    name: sequential-pipeline
```
Here, we can filter these pods based on a label. e.g. `kubectl get pod -l pipelines.kubeflow.org/pipeline-sdk-type=kfp`. It is also useful with `labelSelector` when applying affinity on these tasks.

However, when we try to convert the above pipeline into a single pipelineRun kind, we don't have a way to express the labels or annotations on the task level. Therefore, we want to have the above API to represent the pipelineRun in this way.

```yaml
kind: PipelineRun
metadata:
  name: sequential-pipeline-run
spec:
  pipelineSpec:
    tasks:
    - name: echo
      metadata:
        labels:
          pipelines.kubeflow.org/pipeline-sdk-type: kfp
      taskSpec:
        steps:
        - args:
          - echo hello world
          command:
          - sh
          - -c
          image: library/bash:4.4.23
          name: main
    - name: echo2
      metadata:
        labels:
          pipelines.kubeflow.org/pipeline-sdk-type: tfx
      runAfter:
      - echo
      taskSpec:
        steps:
        - args:
          - echo hello world
          command:
          - sh
          - -c
          image: library/bash:4.4.23
          name: main
```

pritidesai:
thanks @Tomcli for the example, it looks great, would also like to know:

1) How are we utilizing these pods after they are filtered based on the `pipeline-sdk-type`?
2) Is it possible to use any other type instead of `metadata` to specify labels and annotations like `json`? 

I understand the expectation is accurate that the `labels` and `annotations` should propagate to each `taskRun` irrespective of how task definition is specified i.e. `taskRef` or `taskSpec`. The biggest drawback with this approach is:

Metadata specified with `taskRef` would cause conflicts with metadata coming from the task itself. We could restrict it at the pipeline task level which would avoid any conflict but trying to understand the use cases to better design this.

Tomcli:
> thanks @Tomcli for the example, it looks great, would also like to know:
> 
> 1. How are we utilizing these pods after they are filtered based on the `pipeline-sdk-type`?
> 2. Is it possible to use any other type instead of `metadata` to specify labels and annotations like `json`?
> 
> I understand the expectation is accurate that the `labels` and `annotations` should propagate to each `taskRun` irrespective of how task definition is specified i.e. `taskRef` or `taskSpec`. The biggest drawback with this approach is:
> 
> `PipelineTask` level `metadata` would cause conflicts with the `metadata` coming from the task itself (while using `taskRef`). We could restrict it at the pipeline task level which would avoid any conflict but trying to understand the use cases to better design this.

1. We have a metadata service that uses this label to filter out pods that need additional actions (such as collecting pod metrics or cleaning up completed pod with certain labels). K8S labels are useful for selecting objects like this and we want to use it using a single pipelineRun spec as well.

2. I'm not sure is there any other type that can reproduce the same behavior as k8s labels and annotations. If there's any alternative solution I can also give it a try.

I know there will be conflicts when using the `taskRef`. We only want to override the labels and annotations if the provided keys don't exist. It's similar to the conflicts we have for `volumeClaimTemplate` and `podTemplate`.

animeshsingh:
> `PipelineTask` level `metadata` would cause conflicts with the `metadata` coming from the task itself (while using `taskRef`)

This is similar to paradigms where a superset has "default" values, and unless a subset from that superset defines its own values which override the "default", the default gets applied to subset as well.

pkg/apis/pipeline/v1beta1/pipeline_types.go
pritidesai: thanks @jlpettersson, with `taskRef`, the referring task can have metadata which should be sufficient like we have today. My focus was to only add `metadata` with `taskSpec` as `pipeline` has no means of specifying it with `taskSpec`.

examples/v1beta1/pipelineruns/pipelinerun-with-pipelinespec-and-taskspec.yaml
pritidesai: @jlpettersson I think your proposal would help avoiding adding `k8s` specific types in the `PipelineTask`, we can introduce a new field `TaskTemplate` or `EmbeddedTask` in `PipelineTask` which can of type `*Task`:

```
type PipelineTask struct {
	TaskTemplate *Task `json:"taskTemplate,omitempty"`
```

`PipelineTask` can have one of `TaskRef`, `TaskSpec`, or `TaskTemplate`.

I understand we are looking over to `Custom Tasks` which would help solve many use cases but this would solve the issue in short term and could be extended to custom tasks if needed. I am looking for feedback on this 🙏 

@vdemeester @bobcatfish @jlpettersson thoughts?

jlpettersson: I think something like that is the most proper solution, yes.
We haven't used "Template" for declarations in other places, but it may work. Other alternatives are:

`Task *Task `json:"task, omitempty"`

or

`TaskDeclaration *Task `json:"task, omitempty"`

Most correct description about what this is, is probably: _inline task declaration_

I don't think this is related to `CustomTask` at all, it will be the exact same problem. I think `CustomTask` recently has become the most misunderstood thing about Tekton.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -608,14 +608,26 @@ func (c *Reconciler) createTaskRun(ctx context.Context, rprt *resources.Resolved
 		return c.PipelineClientSet.TektonV1beta1().TaskRuns(pr.Namespace).UpdateStatus(tr)
 	}
 
+	// Propagate labels and annotations from PipelineRun and TaskSpec to TaskRun.
+	labels := getTaskrunLabels(pr, rprt.PipelineTask.Name)
+	annotations := getTaskrunAnnotations(pr)
+	if rprt.PipelineTask.TaskSpec != nil {
```
vdemeester, line 614: One question here : which labels should take precedence ? My gut is that the `TaskRun` labels should take precedence over the one defined in the `Pipeline` ; otherwise a `Pipeline` can define a label that override `tekton.dev/Task` or something and it could break some UX/UI/…. This code does the other way around (labels define in the `Pipeline` tasks take over the `TaskRun` labels).

pritidesai: the way its implemented, labels defined in pipeline would take precedence. You are right, it is possible that pipeline defined `tekton.dev` label would break and takes over the TaskRun labels. I will change the order here.

Should we apply the same precedence order in [Conditions](https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/pipelinerun/pipelinerun.go#L812)? I can create a separate PR for conditions.

Also, I think labels with `tekton.dev/` should be reserved for the platform and we should not allow such labels, anything starting with `tekton.dev/`. What do you think? We could add this kind of validation in this PR as part of [ValidateTaskSpecMetadata](https://github.com/tektoncd/pipeline/pull/2826/files#diff-06d577762a729d5f494cfec3dd6771a6R157).

pkg/apis/pipeline/v1alpha1/pipeline_conversion.go
```diff
@@ -61,8 +63,8 @@ func (source *PipelineTask) ConvertTo(ctx context.Context, sink *v1beta1.Pipelin
 	sink.Name = source.Name
 	sink.TaskRef = source.TaskRef
 	if source.TaskSpec != nil {
-		sink.TaskSpec = &v1beta1.TaskSpec{}
-		if err := source.TaskSpec.ConvertTo(ctx, sink.TaskSpec); err != nil {
+		sink.TaskSpec = &v1beta1.EmbeddedTask{v1.ObjectMeta{}, &v1beta1.TaskSpec{}}
```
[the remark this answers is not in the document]
pritidesai: Awesome, thanks, Done 👍

pkg/reconciler/pipelinerun/pipelinerun.go
vdemeester: > Should we apply the same precedence order in [Conditions](https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/pipelinerun/pipelinerun.go#L812)? I can create a separate PR for conditions.

I would think so indeed :wink: 

> Also, I think labels with `tekton.dev/` should be reserved for the platform and we should not allow such labels, anything starting with `tekton.dev/`. What do you think? We could add this kind of validation in this PR as part of [ValidateTaskSpecMetadata](https://github.com/tektoncd/pipeline/pull/2826/files#diff-06d577762a729d5f494cfec3dd6771a6R157).

Indeed, we should probably add that kind of validation to `ValidateTaskSpecMetadata` (for some reason I thought that was already the case :stuck_out_tongue_winking_eye: )

examples/v1beta1/pipelineruns/pipelinerun-with-pipelinespec-and-taskspec.yaml
```diff
@@ -7,6 +7,9 @@ spec:
     tasks:
       - name: echo-good-morning
         taskSpec:
+          metadata:
```
pritidesai, line 10: updated an example to show the usage

examples/v1beta1/pipelineruns/pipelinerun-with-pipelinespec-and-taskspec.yaml
```diff
@@ -7,6 +7,9 @@ spec:
...
+            labels:
+              app: "example"
           steps:
```
pritidesai, line 13: updated an example to show the usage

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: Nope, in general `Metadata` is validated for valid `Name` (no special character `.` and max length of 63 characters). I have added validation on labels as part of `taskSpec` metadata to restrict using `tekton.dev/Task`. Not sure if we can block in general any label with `tekton.dev` (`git` and `docker` credentials use `tekton.dev` but they belong to k8s `Secrets`) in `taskSpec` metadata or restrict labels with `tekton.dev/` from any Tekton resource i.e. `Pipeline`, `Task`, etc

pritidesai: I have updated the order so that TaskRun labels takes higher precedence.

vdemeester:
test pull-tekton-pipeline-integration-tests/

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -104,7 +116,7 @@ type PipelineTask struct {
 
 	// TaskSpec is a specification of a task
 	// +optional
-	TaskSpec *TaskSpec `json:"taskSpec,omitempty"`
+	TaskSpec *EmbeddedTask `json:"taskSpec,inline,omitempty"`
```
pritidesai, line 119: Introducing one more change which might break CLI (see issue #2748)

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -139,6 +150,22 @@ type PipelineTask struct {
 	Timeout *metav1.Duration `json:"timeout,omitempty"`
 }
 
+func (pt *PipelineTask) TaskSpecMetadata() metav1.ObjectMeta {
+	return pt.TaskSpec.Metadata
+}
+
+// ValidateTaskSpecMetadata returns error if the name contains special character "."
+// or name exceeds more than 63 characters
+// or metadata contains label with reserved identifier tekton.dev
```
pritidesai, line 159: This validation is implemented in PR #2891

Tomcli:
Thanks @pritidesai for all the amazing work. Is there any work item needed for approving this PR?

pritidesai:
> Thanks @pritidesai for all the amazing work. Is there any work item needed for approving this PR?

@Tomcli there is no item pending, we need more eyes on this.

Also, this PR is introducing one key change (most likely breaking CLI):

Before:
```
TaskSpec *TaskSpec `json:"taskSpec,omitempty"`
```

After:
```
TaskSpec *EmbeddedTask `json:"taskSpec,inline,omitempty"`
```

where `EmbeddedTask` is defined with:

```
type EmbeddedTask struct {
	Metadata metav1.ObjectMeta `json:"metadata,omitempty"`
	*TaskSpec `json:",inline,omitempty"`
}
```

Which changes initialization of `TaskSpec` from:

```
TaskSpec: &v1beta1.TaskSpec{
```

To:

```
TaskSpec: &v1beta1.EmbeddedTask{TaskSpec: &v1beta1.TaskSpec{
```

@bobcatfish @jlpettersson How do we proceed on this PR? 🙏

vdemeester:
/cc @ImJasonH (in relation to custom tasks)
Overall, seems good to me.

ImJasonH:
Actually come to think of it, all the output-only fields like UID and creationTimestamp should also be rejected in validation, not just silently ignored.

pritidesai:
> Can a user specify `.metadata.namespace` now? I think we should validate that they don't, probably along with `name` and `generateName`?

yes but has no effect since the task is not created, users have freedom to add any metadata they want and the specification is posted in `status.pipelineSpec.tasks[].taskSpec`. 

And just like `taskRef`, labels and annotations are propagated to taskRun.

pritidesai:
> Actually come to think of it, all the output-only fields like UID and creationTimestamp should also be rejected in validation, not just silently ignored.

We could add such validation but these fields have no value since the resource is not created.

ImJasonH:
> We could add such validation but these fields have no value since the resource is not created.

Having fields that the user can supply and which will be accepted, but silently ignored, seems like a confusing but waiting to happen. I'd prefer we explicitly reject fields we don't intend on using, maybe even to the extent of not reusing the `ObjectMeta` type, and defining our own type that only supports `labels` and `annotations` -- this prevents future fields from being added in the future which we don't check for.

pritidesai:
> defining our own type that only supports `labels` and `annotations` -- this prevents future fields from being added in the future which we don't check for.

Creating our own type might contradict and cause more headache for folks (systems) who are auto-generating these specifications. The auto-generating scripts will now have to be intelligent enough to compose `metadata` with only labels and annotations.

I favor adding validation at the `taskSpec` level rather than introducing new type.

@vdemeester thoughts?

@Tomcli would your use case work with such restrictions on `metadata` or this kind of new type?

Tomcli:
@pritidesai yes, our use case only involves labels and annotations, so we are fine with the new type.

ImJasonH:
@pritidesai can you give more details about the kinds of failures you might expect to see from systems generating requests from (presumably) some other config? I'd expect they would handle a new type pretty well, unless they special-case ObjectMeta specifically.

pritidesai:
@ImJasonH, no failure as such, its more about inconsistency in `TaskRun` with `taskRef` vs with `taskSpec`.

Also with new type, labels and annotations themselves can not be validated and will be part of `TaskRun` and `pods` as is.

ImJasonH:
> @ImJasonH, no failure as such, its more about inconsistent `metadata` types in `TaskRun` with `taskRef` vs with `taskSpec`.
> 
> Also with new type, labels and annotations themselves can not be validated and will be pushed to `TaskRun` and `pods` as is.

I'm not sure what you mean by inconsistent `metadata` fields.

Users can specify either:

```yaml
spec:
  taskRef:
    name: task-name
    namespace: task-namespace
    kind: Task|ClusterTask
```

or

```yaml
spec:
  metadata:
    annotations: ...
    labels: ...
  taskSpec:
    steps: ...
```

If the `metadata` field is a type we define/own, that only includes `annotations` and `labels`, it's clearer I think that no other fields are supported. If we reuse the `ObjectMeta` type we have to validate that `name`, `generateName`, etc., aren't specified.

vdemeester:
@ImJasonH right, good point, we want want to define our own "objectmeta" type with the only field we support. I can see this as useful indeed :angel:

pritidesai:
@ImJasonH what I meant by inconsistency is having:

``` 
type TektonObjectMeta struct {
	Labels map[string]string
	Annotations map[string]string
}
```

And using this with `taskSpec`:

```
kind: TaskRun
metadata: #of type ObjectMeta
  generateName: generate-name-
spec:
  metadata: #of type TektonObjectMeta
    annotations: ...
    labels: ...
  taskSpec:
    steps: ...
```

And same With `taskRef`, metadata is of type [ObjectMeta](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#objectmeta-v1-meta):

```
kind: Task
metadata: #of type ObjectMeta
  name: task-name
spec:
  steps:
---

kind: TaskRun
metadata: #of type ObjectMeta
  generateName: generate-name-
spec:
  taskRef:
    name: task-name
``` 

Hope this is explains inconsistent types.

The general perception is `metadata` is of type `ObjectMeta`, it would be best to call it something else if we introduce our own type (field allowing labels and annotations).

vdemeester:
> The general perception is `metadata` is of type `ObjectMeta`, it would be best to call it something else if we introduce our own type (field allowing labels and annotations).

I see what you mean, not sure I fully agree :stuck_out_tongue_closed_eyes:. It needs to be properly documented, and if we had json schema it would make it "automatically" documented, but I don't think user *should* make the assumption `metadata` is of type `ObjectMeta`. It is a pretty common "noun", and the user should only care about what they can give — what type it is behind the scene is an implementation detail. `Step` is a good example of that (it looks like a `ContainerSpec` but it's not — although it's embedding it).

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -90,6 +90,23 @@ type PipelineResult struct {
 	Value string `json:"value"`
 }
 
+type Metadata struct {
+	// +optional
+	Labels map[string]string `json:"labels,omitempty" protobuf:"bytes,1,rep,name=labels"`
+
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty" protobuf:"bytes,2,rep,name=annotations"`
+}
```
pritidesai, line 99: @vdemeester @ImJasonH I have added a new type here, PTAL 🙇‍♀️

vdemeester: nit: I don't think we need the `protobuf:` part :thinking:

pritidesai: Done 😄

bobcatfish:
ps thanks for the updated context in https://github.com/tektoncd/pipeline/pull/2826#issuecomment-647824440 

is it possible to add more of this context (i.e. exactly what the labels and annotations are being used for) to https://github.com/tektoncd/pipeline/issues/2767 and maybe the PR description here as well? to make it easier for future explorers to understand why we need this

pritidesai:
> is it possible to add more of this context (i.e. exactly what the labels and annotations are being used for) to #2767 and maybe the PR description here as well? to make it easier for future explorers to understand why we need this

Sure. Done. Thanks @bobcatfish PTAL 🙏

pritidesai:
There is no rebase needed but the label `needs-rebase` not going away 😢 , deleting that label.

pritidesai:
@vdemeester @ImJasonH @bobcatfish @jlpettersson its ready for review, PTAL 🙏

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -90,6 +90,23 @@ type PipelineResult struct {
 	Value string `json:"value"`
 }
 
+type Metadata struct {
```
ImJasonH, line 93: Thanks for making this change! Would you be open to renaming it something like `PipelineTaskMetadata`? The name is a bit vague as it is.

The name of the type can be changed in the future without breaking our YAML-focused definition of backward-compatibility AIUI, so this might not be a big deal.

pritidesai: yup this can be changed in future without breaking anyone. I had named it `PipelineTaskMetadata` and then changed it to `Metadata` thinking this could be generic and used under any other section in future 💭

pritidesai: Done, since its in the `pipeline_types.go`, would be better to name it `PipelineTaskMetadata` 👍

ImJasonH:
/lgtm

🎉

vdemeester:
> 2020/08/19 08:26:57 main.go:319: Something went wrong: failed to prepare test environment: --provider=gke boskos failed to acquire project: resources not found

Hum, boskos is in trouble :sweat_smile:

pritidesai:
one more try 🙃 

PR #2833, opened 2020-06-18
Add bumping catalog dependency to process of creating a release

bobcatfish:
# Changes

Eventually we want to be able to have Tasks declare what versions they
are compatible with and test against those
(https://github.com/tektoncd/catalog/issues/373) but in the meantime, it
seems reasonable to bump the version used in catalog tests every time we
do a release (and even once we have https://github.com/tektoncd/catalog/issues/373,
we would want to be running tests against new versions as well - and this
could have the benefit of giving us information about issues in a release
early, before users notice them!)

Question: how does the person making the release ensure that this bump
didn't make anything start failing? I think we should still make this
part of the process before we figure this out entirely, but we might
want to create a cron or something that runs all the tests nightly, and
maybe have the build cop trigger the release manually to ensure there
aren't any problems?

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/tekton/README.md
+++ b/tekton/README.md
@@ -30,6 +30,9 @@ To make a new release:
    ([see one of way of doing that here](https://github.com/tektoncd/plumbing/tree/master/tekton/resources/release#create-draft-release)).
 1. Add an entry to [the README](../README.md) at `HEAD` for docs and examples for
    the new release ([README.md#read-the-docs](../README.md#read-the-docs)).
+1. Update [the catalog repo](https://github.com/tektoncd/catalog) test infrastructure to use the
+   new release by updating the `RELEASE_YAML` link in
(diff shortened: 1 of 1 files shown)


bobcatfish:
```
				if err := updateStatusStartTime(&s); err != nil {
				                                ^
pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go:237:46: G601: Implicit memory aliasing in for loop. (gosec)
		r, err := resources.GetResourceFromBinding(&resource, getResource)
		                                           ^
```

ohhhhhh this must be because of https://github.com/tektoncd/plumbing/pull/430 🤔  i wonder if we should be more careful about these kinds of bumps, e.g. try them out first. that would be tough b/c it might be doing stuff like this in other repos too.

PR #2837, opened 2020-06-18
Start emitting CloudEvents for TaskRuns

afrittoli:
# Changes

Add a new method 'SendCloudEventWithRetries' to the cloud events
controller. It that allows emitting cloud events asynchronously
(in a go routine), taking benefit of the retries capabilities of
the cloudevents go sdk.

Rework the fake client to allow for unit testing of
'SendCloudEventWithRetries'. The new implementation is similar
to that of the fake recorder from client-go, which allows to
unit test k8s events and cloud events in a similar fashion.

Add a new config option default-cloud-events-sink in the
defaults config map. This options allows setting a default sink
for cloud events. If the default sink is setup, cloud events are
sent, else they're disabled.

Invoke 'SendCloudEventWithRetries' from the events modules so
that we start sending cloud events in the same places we were
send k8s events (except for errors).

Most of the plumbing is in place to start emitting cloud events
for pipeline runs too, however there is a small refactor for
the config maps required before we can add that, so this commit
only enabled cloud events for task runs for now.

Partially addresses #2082

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Add a new configuration option `default-cloud-events-sink` in the defaults config map.
This options allows setting a default sink for cloud events. 
If the default sink is setup, cloud events are sent, else they're disabled.

Send cloud events (when enabled) from the task run and controller, for all non-error
events that are already notified via k8s events.
```

/cc @n3wscott 
/cc @bobcatfish 

/kind feature


afrittoli:
WIP because tests are missing for the config map part of the PR - I need to test that the config option is pulled in correctly in the context. It may be worth adding at least one reconciler test.
Adding an E2E test would require a custom config of Tekton for that.

pkg/reconciler/events/cloudevent/cloudeventclient.go
```diff
@@ -34,7 +35,16 @@ type CECKey struct{}
 
 func withCloudEventClient(ctx context.Context, cfg *rest.Config) context.Context {
 	logger := logging.FromContext(ctx)
+
+	// Set a cloud event target to the context if a default sink is defined
+	tektonConfig := config.FromContextOrDefaults(ctx)
+	defaultSink := tektonConfig.Defaults.DefaultCloudEventsSink
+	if defaultSink != "" {
+		ctx = cloudevents.ContextWithTarget(ctx, defaultSink)
```
afrittoli, line 43: I think this context is passed to the controller, but not all the way down to the reconciler, so the config option may get lost. I might have to move this code to the reconciler code, the same way I had to move add the client to the context.

docs/install.md
```diff
@@ -235,6 +235,25 @@ data:
   bucket.service.account.field.name: GOOGLE_APPLICATION_CREDENTIALS
 ```
 
+## Configuring CloudEvents notifications
+
+When configured so, Tekton can generate `CloudEVents` for `TaskRun` and `PipelineRun` lifecycle
```
sbwsg, line 240: nit: `EVents` -> `Events`

pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -133,6 +133,9 @@ func EventForPipelineRun(pipelineRun *v1beta1.PipelineRun) (*cloudevents.Event,
 
 func getEventType(runObject objectWithCondition) (*TektonEventType, error) {
 	c := runObject.GetStatusCondition().GetCondition(apis.ConditionSucceeded)
+	if c == nil {
+		return nil, fmt.Errorf("no condition for in %T", runObject)
```
sbwsg, line 137: nit: is the message missing a bit here? Looks like it should be `"no condition for %T"` or `"no condition for ConditionSucceeded in %T"` or similar?

afrittoli: Nice catch, thank you!

n3wscott:
Sweeeet! I will review in the morning. Just saw this. Until then I will be on the hype train!! 😅😂🤩

docs/install.md
afrittoli: Done

docs/events.md
```diff
@@ -39,3 +41,24 @@ No events are emitted for `Conditions` today (https://github.com/tektoncd/pipeli
 - `Failed`: this is triggered if the `PipelineRun` is completed, but not
   successfully. Causes of failure may be: one the `Tasks` failed or the
   `PipelineRun` was cancelled.
+
+# CloudEvents
+
+WHen a sink is [configured](../install.md#configuring-cloudevents-notifications), the following events
```
vdemeester, line 47: `s/WHen/When`

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -107,6 +115,12 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, tr *v1beta1.TaskRun) pkg
 		// on the event to perform user facing initialisations, such has reset a CI check status
 		afterCondition := tr.Status.GetCondition(apis.ConditionSucceeded)
 		events.Emit(recorder, nil, afterCondition, tr)
+		if sendCloudEvents {
```
vdemeester, line 118: Any reason for not putting this into `events.Emit` ? That way we wouldn't have to track all `events.Emit` to add that code ?

afrittoli: Yes, that was my original plan, but it would require changing the existing signature of Emit events, and I didn't want to make this PR larger than it was already. 
The way I would do it is to pass the context (instead of recorder)  `Emit` (or perhaps add a new method `EmitAll` that takes the `ctx`). The from the context extract the recorder for k8s events and the config for the cloud events.
I can do it in this PR or in a PR on top, wdyt?

vdemeester: I would rather do it in this PR as it won't add too much lines of code (aka ain't gonna add 1k :stuck_out_tongue:).

vdemeester:
@afrittoli needs a rebase :upside_down_face:

docs/events.md
afrittoli: done

pkg/reconciler/taskrun/taskrun.go
afrittoli: Ok, changing that implies that events will also be sent for pipeline runs. 
It also requires some changes to makes the method signature compatible.
I've done that, hopefully it looks good. When a sink is set, it will now start sending cloud events to that sink for all pipeline run and task run events!

afrittoli:
Should add unit tests to `events` to restore a good coverage. Also need to add cloud event testing to `pipelinerun_test`

PR #2846, opened 2020-06-22
Pick up the latest knative/pkg and K8s 0.17.x

mattmoor:
The only other manual changes are:
1. Call `Promote` in tests, otherwise our generated reconcilers won't do anything (they aren't the leader!)
2. Fill in namespaces in a few tests (I believe this is a change in the k8s 0.17.x libs)

The main change with respect to genreconciler is that they will now generate LeaderAware reconcilers.  Reconcilers may be sharded across multiple replicas by setting the `buckets` in `config-leaderelection`, which will subdivide the key space being reconciled across N buckets, which are individually leader elected.

Right now the sharded reconcilers only light up through our webhook entrypoint in shared main, but I hope to eliminate the split (likely through having mirrored entrypoints for a time) in the near future.

The broader design for this is here: https://docs.google.com/document/d/1i_QHjQO2T3SNv49xjZLWlivcc0UvZN1Tbw2NKxThkyM/edit

```release-notes
The minimum Kubernetes version is now 1.16.

The tekton webhook can now be horizontally scaled by enabling leader election via `enabledComponents: webhook` in `config-leader-election.yaml`
```

/hold

cc @vdemeester @ImJasonH


mattmoor:
ah the good old `shoulders.md` case change 🙄

vdemeester:
```
        >>> Container sidecar-registry:
        failed to try resolving symlinks in path "/var/log/pods/arendelle-8r9ph_kanikotask-run-pod-28v7p_98f0e76a-8ace-46d1-8572-01d1bc555635/sidecar-registry/0.log": lstat /var/log/pods/arendelle-8r9ph_kanikotask-run-pod-28v7p_98f0e76a-8ace-46d1-8572-01d1bc555635/sidecar-registry/0.log: no such file or directory
```

First time I see those log on a failure of the kaniko e2e test :thinking: :eyes:

mattmoor:
@vdemeester passing now!

afrittoli:
Does this mean that we would not rely on k8s 0.17.x?
If yes, I'd be concerned about restricting our users to a recent version of k8s.

vdemeester:
> Does this mean that we would not rely on k8s 0.17.x?
> If yes, I'd be concerned about restricting our users to a recent version of k8s.

I don't think so. @mattmoor might correct me but this is just an update of the client, we are not using anything k8s 1.17 specific that wouldn't be supported earlier, right ?

mattmoor:
@vdemeester @afrittoli Sorry, good catch.  Too focused on genreconciler.

This DOES pick up a new minimum K8s version as well: 1.16, sorry for missing that.

This will enable Tekton to pick up CRD v1 🤩 , and is available on all of the public clouds, but I should have caught it.  My bad.

I can update the release notes, but LMK if that's cool.  I put a `/hold` for good measure.

bobcatfish:
release notes update would be good, also tho we have some docs that talk about the min version, there's definitely https://github.com/tektoncd/pipeline/blob/e68c62700da03f407e5b5376d4bf69d473900a7b/DEVELOPMENT.md#kubernetes-cluster and there might be something at https://tekton.dev/ as well (hopefully its something pulled from this repo but not sure)

mattmoor:
Thanks @bobcatfish I will circle back this afternoon and at least try to hit those places (and release notes).

mattmoor:
/hold cancel

Ok, this is RFAL.  I updated every 1.15 reference outside of vendor 😉

afrittoli:
I feel like we should discuss with with a larger audience before moving forward? 
I've heard people having to stick to older Tekton versions because of the current k8s version constraint, so I wonder if we should force a new version already.

@mattmoor do you think it would be possible to control this via a feature flag? What I mean, if I'm not interested in scaling in the controller horizontally, be able to run on k8s 1.15.

mattmoor:
@afrittoli Generally the hard version dependency check is configurable via an environment variable on the controller, but I believe we also picked up some use of the v1 CRD API, which is only available in 1.16+

Note that [our K8s minimum version principle](https://github.com/knative/community/blob/250d8ad51a7ce5dec746e878fa0773a6437d523b/mechanics/RELEASE-VERSIONING-PRINCIPLES.md#k8s-minimum-version-principle) is based on what K8s actually supports upstream, but Ack the massive lag most cloud providers have here.  We are actually late picking up 1.16 because GKE was so far behind.

Ack the broader discussion, should I start one in slack?

vdemeester:
At least, let's hold that for after 0.14 (aka requiring 1.16 as minimum) ?

mattmoor:
@vdemeester When is that?  Is there a regular cycle?  Happy to mark it on my calendar to circle back and rebase/update.

vdemeester:
@mattmoor it's *almost* a regular cycle. It should happen next week (thursday :pray:)

mattmoor:
@vdemeester ok, great.  By then we should have pkg 0.16 snapped (Tuesday), so I will plan to update this shortly thereafter, and then it will hopefully just be good-to-go when the snap has happened.

mattmoor:
@vdemeester Ok, I think this is ready to go once the release cuts, it pins to `knative.dev/pkg@release-0.16`, which cut this morning.

mattmoor:
Thanks @vdemeester 🤩 

@afrittoli @bobcatfish LMK if there's anything else?

Issue #2844, opened 2020-06-22
Tekton metrics are reporting wrong values


RafaeLeal:
# Expected Behavior
Tekton metrics coherent with data gather with kubectl

# Actual Behavior
I have two examples, for the `tekton_pipelinerun_count` metric, it's currently reporting 22 failed and 58 success, but the cluster only have 20 PipelineRuns
![image](https://user-images.githubusercontent.com/9096450/85314522-09260b80-b490-11ea-825c-9e364eb301da.png)
```
$ kubectl get pipelinerun --all-namespaces -o json | jq '.items | length'
20
```
The second example, `tekton_pipelinerun_duration_seconds_sum` metric seems to increase indefinitely. I expected to raise until the pipelinerun is completed, than stabilize in the same value as the duration showed with `tkn pr describe <pipelinerun_name>`
![image](https://user-images.githubusercontent.com/9096450/85315102-d0d2fd00-b490-11ea-9161-aa61083169c7.png)

I also noticed that
1. It doesn’t raise at the same rate.
2. Always raises a multiple of the actual duration of the pipeline.

I suspect that every time the controller reconciles it’s adding the duration of the pipeline.

# Steps to Reproduce the Problem

1. Enable tekton metrics
2. Make Prometheus scrape tekton
3. Run some pipelineruns

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**
  ```
  $ kubectl version
  Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-26T06:17:09Z", GoVersion:"go1.14", Compiler:"gc",Platform:"darwin/amd64"}
  Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.11-eks-af3caf", GitCommit:"af3caf6136cd355f467083651cc1010a499f59b1", GitTreeState:"clean", BuildDate:"2020-03-27T21:51:36Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
  ```
- Tekton Pipeline version: v0.13.2

  **Output of `tkn version`**
  ```
  $ tkn version
  Client version: 0.8.0
  Pipeline version: unknown
  ```
[...]

PR #2848, opened 2020-06-23
Metrics: pipelinerun_count is not correct

vincent-pli:
Part fix issue: #2844 
In scenario: `PipelineResult` is not null, the Metrics: `pipelinerun_count` will not correct.
The root cause is that:
https://github.com/tektoncd/pipeline/blob/4fb10781528724d329120d9246da80fa39cb8641/pkg/reconciler/pipelinerun/pipelinerun.go#L149-L154

The `updatePipelineResults(ctx, pr)` will update `pr.Status.PipelineResults` if needed, this will make the `knative/pkg` update the `Pipelinerun.Status` and `reconcile` will be trigger again.

So the `pipelinerun_count` will count again.

# Changes

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -47,6 +47,7 @@ import (
 	"github.com/tektoncd/pipeline/pkg/workspace"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
+	equality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -149,6 +150,7 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 	if pr.IsDone() {
 		// We may be reading a version of the object that was stored at an older version
 		// and may not have had all of the assumed default specified.
+		original := pr.DeepCopy()
 		pr.SetDefaults(contexts.WithUpgradeViaDefaulting(ctx))
 
 		c.updatePipelineResults(ctx, pr)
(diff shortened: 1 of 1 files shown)


hrishin:
@vincent-pli thank you for the fix!
Let's move this complete logic into different place and consolidate all metrics reporting [in one place](https://github.com/tektoncd/pipeline/blob/c2d3ea6e0113627bb08befd9ae2862bdf8a272fa/pkg/reconciler/pipelinerun/metrics.go#L169). Wonders its better to decouple this logic from the reconciler? :thinking: 

Although in this case count and duration may need additional care. Like not to repeat the count and consolidate duration.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -165,12 +167,16 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
...
 			return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
 		}
-		go func(metrics *Recorder) {
-			err := metrics.DurationAndCount(pr)
-			if err != nil {
-				logger.Warnf("Failed to log the metrics : %v", err)
-			}
-		}(c.metrics)
+
+		if equality.Semantic.DeepEqual(original.Status, pr.Status) {
```
afrittoli, line 171: I don't think this is enough to prevent false metrics from being reported.
Not knowing much about opencensus, my expectation reading this code is that `DurationAndCount` was idempotent, and it would take care of not submitting twice.
Now, if the controller is restarted for instance, all metrics will be submitted again.

A way to solve this could perhaps be to write metrics where the completion time is set?

vincent-pli:
@hrishin thanks for comments, that's great if we can prevent repeat in `DurationAndCount`, but I guess it's difficult. as @afrittoli mentioned the `DurationAndCount` should be idempotent so the invoker should take care of it:
- when to call it (in this case, count the duration of `pipelinerun`, so should invoke when `pipelinerun` complete)
- prevent repeat
- maybe others

@afrittoli 
`completion time` is not work, it already existedevery time when access
`if pr.IsDone() {`

About "controller restart", I think that's ok, all the metrics will be report again but with different time series.

hrishin:
> completion time is not work, it already existed everytime when access
if pr.IsDone() {
About "controller restart", I think that's ok, all the metrics will be report again but with different time series.

@vincent-pli @afrittoli, yes this fix looks promising. Just wonders how metrics aggregators behave in case of deployment restarts? We shall test this?

vincent-pli:
@hrishin  
When `controller` restart, all metrics will be reload again unless you delete `pipelinerun` or other stuff at the restart moment.
For example: for `pipelinerun` we has metrics:
- pipelinerun_duration_seconds
- pipelinerun_count
- running_pipelineruns_count

All the metric list above will be recover just one thing: the `pipelinerun_count` become correct, since `if pr.IsDone() {` access only onece after `controller` restart : )

afrittoli:
I meant to submit the metric in a different place in the code, when the completion time is set, which happens only once, regardless of further reconciles or restarts.

Could we add test coverage for this?

vincent-pli:
@afrittoli 
The point `completion time is set` is only happen once, that's right, but at the point the `reconcile` is not complete since the final "status: success" `update` is not occurred, basically, it's not a real complete `pippelinerun` at that point,  I  think we could not launch metric to mark it's already complete.

https://github.com/tektoncd/pipeline/blob/08ac72c3c4be11330d9f909b7519a4788d46ad90/pkg/reconciler/pipelinerun/pipelinerun.go#L465-L466

PR #2859, opened 2020-06-24
Emit an event if overwriting PodTemplate affinity

### LINKED ISSUE #2678: Emit an Event when a set affinity property is overwritten
# Expected Behavior

That it is easy to detect misconfiguration or conflicting configurations

# Actual Behavior

The Affinity Assistant will set `affinity` on TaskRun pods, but not check or warning is done if the `affinity` filed already had a value set. An `Event` should be generated if this happens. Requested in https://github.com/tektoncd/pipeline/pull/2630#issuecomment-630320385

/kind feature
/kind cleanup

adshmh:
# Changes

The taskrun controller now emits a warning event if the affinity
specified by pod template will be overwritten with affinity assistant.
Closes #2678

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```
The taskrun controller emits a warning event if pod template affinity will be overwritten with affinity assistant.
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -629,8 +629,13 @@ func (c *Reconciler) createPod(ctx context.Context, tr *v1beta1.TaskRun, rtr *re
 		return nil, fmt.Errorf("translating TaskSpec to Pod: %w", err)
 	}
 
-	return c.KubeClientSet.CoreV1().Pods(tr.Namespace).Create(pod)
-
+	pod, err = c.KubeClientSet.CoreV1().Pods(tr.Namespace).Create(pod)
+	if err == nil && willOverwritePodSetAffinity(tr) {
+		if recorder := controller.GetEventRecorder(ctx); recorder != nil {
+			recorder.Eventf(tr, corev1.EventTypeWarning, "PodAffinityOverwrite", "Pod template affinity is overwritten by affinity assistant for pod %q", pod.Name)
+		}
+	}
+	return pod, err
 }
 
 type DeletePod func(podName string, options *metav1.DeleteOptions) error
@@ -730,4 +735,14 @@ func storeTaskSpec(ctx context.Context, tr *v1beta1.TaskRun, ts *v1beta1.TaskSpe
 		tr.Status.TaskSpec = ts
 	}
 	return nil
+}
+
+// willOverwritePodSetAffinity returns a bool indicating whether the
+// affinity for pods will be overwritten with affinity assistant.
+func willOverwritePodSetAffinity(taskRun *v1beta1.TaskRun) bool {
+	var podTemplate v1beta1.PodTemplate
+	if taskRun.Spec.PodTemplate != nil {
+		podTemplate = *taskRun.Spec.PodTemplate
+	}
+	return taskRun.Annotations[workspace.AnnotationAffinityAssistantName] != "" && podTemplate.Affinity != nil
 }
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -3026,3 +3026,68 @@ func Test_storeTaskSpec(t *testing.T) {
 		t.Fatalf(diff.PrintWantGot(d))
 	}
 }
+
+func TestWillOverwritePodAffinity(t *testing.T) {
+	affinity := &corev1.Affinity{
+		PodAffinity: &corev1.PodAffinity{
+			RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
+				{
+					Namespaces: []string{"tekton-pipelines"},
+				},
+			},
+		},
+	}
+	affinityAssistantName := "pipeline.tekton.dev/affinity-assistant"
+
+	tcs := []struct {
+		name                string
+		hasTemplateAffinity bool
+		annotations         map[string]string
+		expected            bool
+	}{
+		{
+			name:     "no settings",
+			expected: false,
+		},
+		{
+			name: "no PodTemplate affinity set",
+			annotations: map[string]string{
+				affinityAssistantName: "affinity-assistant",
+			},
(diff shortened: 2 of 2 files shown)


pkg/pod/pod.go
```diff
@@ -384,3 +384,13 @@ func shouldAddReadyAnnotationOnPodCreate(ctx context.Context, sidecars []v1beta1
 	cfg := config.FromContextOrDefaults(ctx)
 	return !cfg.FeatureFlags.RunningInEnvWithInjectedSidecars
 }
+
+// WillOverwritePodSetAffinity returns a bool indicating whether the
+// affinity for pods will be overwritten with affinity assistant.
+func WillOverwritePodSetAffinity(taskRun *v1beta1.TaskRun) bool {
```
jlpettersson, line 390: Is it possible to declare this in a package so that it can be _private_?

pkg/pod/status.go
```diff
@@ -67,6 +67,10 @@ const (
 
 	//timeFormat is RFC3339 with millisecond
 	timeFormat = "2006-01-02T15:04:05.000Z07:00"
+
+	// WarningOverwritingPodAffinity indicates that the affinity specified by
+	// the pod template will be overwritten with affinity assistant
+	WarningOverwritingPodAffinity = "PodAffinityOverwrite"
```
jlpettersson, line 73: Also this, is it possible to declare this in a context so that it can be _private_? I think it is only used in _a single place_  - so I feel it should be possible.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -360,11 +360,15 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1beta1.TaskRun,
 			tr.Spec.Workspaces = taskRunWorkspaces
 		}
 
+		willOverwriteAffinity := podconvert.WillOverwritePodSetAffinity(tr)
 		pod, err = c.createPod(ctx, tr, rtr)
```
jlpettersson, line 364: is it possible to move in the `willOverwritePodSetAffinity` to inside `c.createPod()` - not sure it is possible, but it would be closer to the place where the affinity is set.

pkg/pod/status.go
adshmh: Thank you for the review. Fixed.

pkg/reconciler/taskrun/taskrun.go
adshmh: Thank you for the review. Yes, it can be moved inside `createPod`, but then we need to pass the recorder to `createPod` as well, and it will need to start emitting events. I implemented it this way to avoid changing `createPod` signature, and to keep all the event emitting in the `reconciler` package. 

If moving this to the `pod` package is preferred, please let me know and I'll update the PR.

pkg/pod/pod.go
adshmh: Thank you for the review. I think this is related to the [comment below](https://github.com/tektoncd/pipeline/pull/2859#discussion_r445822195), i.e. if the `WillOverwritePodSetAffinity` is moved to the `reconciler` package, then it can be made private.

pkg/pod/pod.go
```diff
@@ -384,3 +384,13 @@ func shouldAddReadyAnnotationOnPodCreate(ctx context.Context, sidecars []v1beta1
...
+	var podTemplate v1beta1.PodTemplate
+	if taskRun.Spec.PodTemplate != nil {
+		podTemplate = *taskRun.Spec.PodTemplate
+	}
+	return taskRun.Annotations[workspace.AnnotationAffinityAssistantName] != "" && podTemplate.Affinity != nil
```
sbwsg, line 395: Is it possible that `taskRun.Annotations` is ever a nil map?  I think this might panic if so.

pkg/reconciler/taskrun/taskrun.go
sbwsg: It looks like you wouldn't have to change the signature of `createPod`, you can look up the recorder from the context:

```go
// in c.createPod
recorder := controller.GetEventRecorder(ctx)
if recorder != nil && willOverwritePodAffinity(tr) {
    recorder.Eventf(...)
}
```

adshmh: Thank you for the reviews. Moved `willOverwritePodSetAffinity` to `createPod`, no changes were made to the signature of `createPod`.

pkg/pod/pod.go
adshmh: Thank you for the review. I did not introduce a nil check here as it is only reading the map so there won't be a panic if the map is nil. Please let me know if I am missing something here regarding the nil check.

sbwsg:
@tektoncd/core-maintainers does anyone have bandwidth to review and lgtm if all looks ok?

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -629,8 +629,13 @@ func (c *Reconciler) createPod(ctx context.Context, tr *v1beta1.TaskRun, rtr *re
 		return nil, fmt.Errorf("translating TaskSpec to Pod: %w", err)
 	}
 
-	return c.KubeClientSet.CoreV1().Pods(tr.Namespace).Create(pod)
-
+	pod, err = c.KubeClientSet.CoreV1().Pods(tr.Namespace).Create(pod)
+	if err == nil && willOverwritePodSetAffinity(tr) {
+		if recorder := controller.GetEventRecorder(ctx); recorder != nil {
+			recorder.Eventf(tr, corev1.EventTypeWarning, "PodAffinityOverwrite", "Pod template affinity is overwritten by affinity assistant for pod %q", pod.Name)
```
pritidesai, line 635: I would introduced `const` here for `PodAffinityOverwrite` but going through the reviews from @jlpettersson on the same 🙃 , looks good for now, can be converted to `const` if needed in future

pritidesai:
/lgtm

thanks @adshmh @jlpettersson @sbwsg 🙏

PR #2869, opened 2020-06-26
 refactor GetPipelineConditionStatus - drop unexported function isdagstopped

pritidesai:
# Changes

 This is a minor refactoring of `GetPipelineConditionStatus`.

Adding an attribute (`IsSkipped`) to `PipelineRunState` to tell whether a PipelineTask will not be run because:

(1) its condition checks failed or
(2) one of the parents conditions failed or
(3) pipeline is in stopping state

Instead of explicitly checking (3) in `GetPipelineConditionStatus`, moved this check to the `IsSkipped` function which is used by both the scheduler and while determining pipeline status. 

Based on the discussion [here](https://github.com/tektoncd/pipeline/pull/2661#issuecomment-649843803).

One additional refactoring by adding a local function to check if a task exists in Graph and use that helper function instead of:

```
if _, ok := d.Nodes[t.PipelineTask.Name]; ok {
```

/cc @bobcatfish @afrittoli 

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -140,6 +140,43 @@ func (t ResolvedPipelineRunTask) IsStarted() bool {
 	return true
 }
 
+// IsSkipped returns true if a PipelineTask will not be run because
+// (1) its Condition Checks failed or
+// (2) one of the parent task's conditions failed or
+// (3) Pipeline is in stopping state (one of the PipelineTasks failed)
+// Note that this means IsSkipped returns false if a conditionCheck is in progress
+func (t ResolvedPipelineRunTask) IsSkipped(state PipelineRunState, d *dag.Graph) bool {
+	// it already has TaskRun associated with it - PipelineTask not skipped
+	if t.IsStarted() {
+		return false
+	}
+
+	// Check if conditionChecks have failed, if so task is skipped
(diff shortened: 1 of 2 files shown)


pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -292,6 +310,14 @@ func (state PipelineRunState) GetFinalTasks(d *dag.Graph, dfinally *dag.Graph) [
 	return tasks
 }
 
+// Check if a PipelineTask belongs to the specified Graph
+func isTaskInGraph(n string, d *dag.Graph) bool {
```
afrittoli, line 314: "n" is not a very informative name, since the type it's a string, perhaps we could use something like "pipelineTaskName"?

afrittoli:
Oh, it looks like there is a conflict now, possibly with your other PR that just merged.
Would you mind @pritidesai make this PR a single commit, and perhaps change the parameter name in `isTaskInGraph`?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: Done

pritidesai:
thanks @afrittoli I have renamed the variable and squashed two commits into one, PTAL 🙏

PR #2876, opened 2020-06-29
refactoring pipelinerun unit test

pritidesai:
# Changes

Creating a common function [reconcileRun](https://github.com/tektoncd/pipeline/pull/2876/files#diff-f7c8532cfe79319c0fff711a4979a6d6R3460) to:

* Create an instance of the `PipelineRun` controller - `getPipelineRunController`
* Reconcile specified `Data` - PipelineRun, Pipeline, Tasks, etc.
* Verify PipelineRun was reconciled
* Check events introduced in PR #2874 

All of the above steps were repeating in all the test functions 😢 

Updated all unit tests in `pipelinerun_test.go` to utilize this common function.

**Further possible refactoring:**

* Great suggestion from @afrittoli on using `Matches` instead of retrieving resources directly, [here](https://github.com/tektoncd/pipeline/pull/2821/files#r446544936), which can go in a separate PR.

* One more PR would be great to consolidate `namespace` instead of hardcoding `foo` everywhere.

* One test failure results in `metrics recorder` failure for all unit tests and very hard to spot what was the actual cause of failure. Possibly fix the way we register metrics view.

# Submitter Checklist

review them:

## Reviewer Notes

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -48,7 +48,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	k8stesting "k8s.io/client-go/testing"
 	ktesting "k8s.io/client-go/testing"
 	"k8s.io/client-go/tools/record"
 	"knative.dev/pkg/apis"
@@ -78,8 +77,11 @@ var (
 	ignoreResourceVersion = cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion")
 )
 
-func getRunName(pr *v1beta1.PipelineRun) string {
-	return strings.Join([]string{pr.Namespace, pr.Name}, "/")
+type PipelineRunTest struct {
+	test.Data  `json:"inline"`
+	Test       *testing.T
+	TestAssets test.Assets
+	Cancel     func()
 }
 
 func ensureConfigurationConfigMapsExist(d *test.Data) {
@@ -302,30 +304,22 @@ func TestReconcile(t *testing.T) {
 		ClusterTasks:      clusterTasks,
 		PipelineResources: rs,
 	}
+	prt := NewPipelineRunTest(d, t)
+	defer prt.Cancel()
 
-	testAssets, cancel := getPipelineRunController(t, d)
-	defer cancel()
-	c := testAssets.Controller
-	clients := testAssets.Clients
-
-	if err := c.Reconciler.Reconcile(context.Background(), "foo/test-pipeline-run-success"); err != nil {
-		t.Fatalf("Error reconciling: %s", err)
+	wantEvents := []string{
+		"Normal Started",
+		"Normal Running Tasks Completed: 0",
 	}
+	reconciledRun, clients := prt.reconcileRun("foo", "test-pipeline-run-success", wantEvents, false)
 
-	if len(clients.Pipeline.Actions()) == 0 {
-		t.Fatalf("Expected client to have been used to create a TaskRun but it wasn't")
-	}
-
-	t.Log("actions", clients.Pipeline.Actions())
-
-	// Check that the PipelineRun was reconciled correctly
-	reconciledRun, err := clients.Pipeline.TektonV1beta1().PipelineRuns("foo").Get("test-pipeline-run-success", metav1.GetOptions{})
-	if err != nil {
-		t.Fatalf("Somehow had error getting reconciled run out of fake client: %s", err)
+	actions := clients.Pipeline.Actions()
+	if len(actions) < 2 {
+		t.Fatalf("Expected client to have at least two action implementation but it has %d", len(actions))
 	}
 
 	// Check that the expected TaskRun was created
-	actual := clients.Pipeline.Actions()[1].(ktesting.CreateAction).GetObject()
+	actual := actions[1].(ktesting.CreateAction).GetObject()
 	expectedTaskRun := tb.TaskRun("test-pipeline-run-success-unit-test-1-mz4c7",
 		tb.TaskRunNamespace("foo"),
 		tb.TaskRunOwnerReference("PipelineRun", "test-pipeline-run-success",
@@ -394,15 +388,6 @@ func TestReconcile(t *testing.T) {
 
 	// A PVC should have been created to deal with output -> input linking
 	ensurePVCCreated(t, clients, expectedTaskRun.GetPipelineRunPVCName(), "foo")
-
-	wantEvents := []string{
-		"Normal Started",
-		"Normal Running Tasks Completed: 0",
-	}
-	err = checkEvents(t, testAssets.Recorder, "test-pipeline-run-success", wantEvents)
-	if !(err == nil) {
-		t.Errorf(err.Error())
-	}
 }
 
 func TestReconcile_PipelineSpecTaskSpec(t *testing.T) {
@@ -430,36 +415,25 @@ func TestReconcile_PipelineSpecTaskSpec(t *testing.T) {
 	}
 
 	d := test.Data{
-		PipelineRuns:      prs,
-		Pipelines:         ps,
-		Tasks:             nil,
-		ClusterTasks:      nil,
-		PipelineResources: nil,
+		PipelineRuns: prs,
+		Pipelines:    ps,
 	}
+	prt := NewPipelineRunTest(d, t)
+	defer prt.Cancel()
 
-	testAssets, cancel := getPipelineRunController(t, d)
-	defer cancel()
-	c := testAssets.Controller
-	clients := testAssets.Clients
-
-	if err := c.Reconciler.Reconcile(context.Background(), "foo/test-pipeline-run-success"); err != nil {
-		t.Fatalf("Error reconciling: %s", err)
+	wantEvents := []string{
+		"Normal Started",
+		"Normal Running Tasks Completed: 0",
 	}
+	reconciledRun, clients := prt.reconcileRun("foo", "test-pipeline-run-success", wantEvents, false)
 
-	if len(clients.Pipeline.Actions()) == 0 {
-		t.Fatalf("Expected client to have been used to create a TaskRun but it wasn't")
-	}
-
-	t.Log("actions", clients.Pipeline.Actions())
-
-	// Check that the PipelineRun was reconciled correctly
-	reconciledRun, err := clients.Pipeline.TektonV1beta1().PipelineRuns("foo").Get("test-pipeline-run-success", metav1.GetOptions{})
-	if err != nil {
-		t.Fatalf("Somehow had error getting reconciled run out of fake client: %s", err)
+	actions := clients.Pipeline.Actions()
+	if len(actions) < 2 {
+		t.Fatalf("Expected client to have at least two action implementation but it has %d", len(actions))
 	}
 
 	// Check that the expected TaskRun was created
-	actual := clients.Pipeline.Actions()[1].(ktesting.CreateAction).GetObject()
+	actual := actions[1].(ktesting.CreateAction).GetObject()
 	expectedTaskRun := tb.TaskRun("test-pipeline-run-success-unit-test-task-spec-9l9zj",
 		tb.TaskRunNamespace("foo"),
 		tb.TaskRunOwnerReference("PipelineRun", "test-pipeline-run-success",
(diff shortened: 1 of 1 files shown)


pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -45,7 +45,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	k8stesting "k8s.io/client-go/testing"
 	ktesting "k8s.io/client-go/testing"
```
pritidesai, line 48: the same module was imported twice with different label 😕

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -204,37 +203,15 @@ func TestReconcile(t *testing.T) {
...
-	t.Log("actions", clients.Pipeline.Actions())
+	reconciledRun, clients := getReconciledRun(t, prs, ps, ts, clusterTasks, nil, rs, nil, "foo", "test-pipeline-run-success")
+	actions := clients.Pipeline.Actions()
 
-	// Check that the PipelineRun was reconciled correctly
-	reconciledRun, err := clients.Pipeline.TektonV1beta1().PipelineRuns("foo").Get("test-pipeline-run-success", metav1.GetOptions{})
-	if err != nil {
-		t.Fatalf("Somehow had error getting reconciled run out of fake client: %s", err)
+	if len(actions) < 2 {
+		t.Fatalf("Expected client to have at least two action implementation but it has %d", len(actions))
```
pritidesai, line 210: Have added check on length before accessing that index in the list

pritidesai:
@vdemeester I rebased this PR but the label (`needs-rebase`) wasn't removed, weird 🤔

pritidesai:
Diff in this PR looks huge 🤦‍♀️. Its all about introducing a common function to build PipelineRun which is used by all unit tests in `pipelinerun_test.go` instead of repeating them in all the tests, would appreciate your review 🙏

bobcatfish:
> Diff in this PR looks huge 🤦‍♀️.

it's pretty amazing that it's `size/XL` and it's only one file! I'd love to find ways to have less tests in there one day 🙏  or maybe targeting the repetition is the way to go!! thanks for taking this on @pritidesai !

pkg/reconciler/pipelinerun/pipelinerun_test.go
bobcatfish: thanks for explaining! really helps when reviewing :D

and good call!!

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3759,3 +3305,112 @@ func getTaskRunStatus(t string, status corev1.ConditionStatus) *v1beta1.Pipeline
...
+	t *testing.T,
+	prs []*v1beta1.PipelineRun,
+	ps []*v1beta1.Pipeline,
+	ts []*v1beta1.Task,
+	clusterTasks []*v1beta1.ClusterTask,
+	trs []*v1beta1.TaskRun,
+	rs []*v1alpha1.PipelineResource,
+	conditions []*v1alpha1.Condition,
+	namespace, prName string,
+	wantEvents []string,
```
bobcatfish, line 3339: minor: using newlines like this is not very standard in our Go codebase, we're much more likely to put these together on one line. ways to improve this:

* Docstring
* Long param lists are often identifed as a "code smell" (https://refactoring.guru/smells/long-parameter-list https://www.codingrevolution.com/long-parameter-list/#:~:text=Long%20parameter%20list%20in%20a,four%20is%20considered%20too%20many) the solution is usually to group these args together into an object you pass in - in this case I could see maybe instantiating a struct with these args and calling functions on it - that might help with some of the ambiguity around the many responsibilities of this function as well (if you wanted to break it up into multiple functions)

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3759,3 +3305,112 @@ func getTaskRunStatus(t string, status corev1.ConditionStatus) *v1beta1.Pipeline
...
+		Pipelines:         ps,
+		TaskRuns:          trs,
+		Tasks:             ts,
+		ClusterTasks:      clusterTasks,
+		PipelineResources: rs,
+		Conditions:        conditions,
+	}
+}
+
+func getReconciledRun(
```
bobcatfish, line 3329: I think it's important to try to indicate with naming whether mutations / assertions are happening in functions - reading the name of `getReconciledRun` I would expect it to be only retrieving a run, but it seems to be doing several things:

- instantiating a controller
- reconciling
- asserting against events

Some alternative naming ideas: `reconcileRun` (dropping "get" makes it seem less like it's just retrieving) `assertRunSucceeds`

Another idea, mentioned below as well: create a struct that holds all the args below and expose several functions instead of just one - this will probably still be significantly easier to read and work with than the amount of repetition we have today

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3759,3 +3305,112 @@ func getTaskRunStatus(t string, status corev1.ConditionStatus) *v1beta1.Pipeline
...
+	d := buildTestData(prs, ps, ts, clusterTasks, trs, rs, conditions)
+
+	testAssets, cancel := getPipelineRunController(t, d)
+	defer cancel()
+	c := testAssets.Controller
+	clients := testAssets.Clients
+	recorder := testAssets.Recorder
+
+	if reconcileError := c.Reconciler.Reconcile(context.Background(), namespace+"/"+prName); reconcileError != nil {
+		t.Fatalf("Error reconciling: %s", reconcileError)
```
bobcatfish, line 3351: have you considered using t.Helper? (https://blog.golang.org/go1.9)

It might be worth comparing what it looks like when one of the tests using this helper fails using it vs not using it (it might make more sense if you were to break this function up into a few)

pkg/reconciler/pipelinerun/pipelinerun_test.go
pritidesai: renamed `getReconciledRun` to `reconcileRun`

Creating `struct` is a great suggestion, I created a struct [PipelineRunTest](https://github.com/tektoncd/pipeline/pull/2876/files#diff-f7c8532cfe79319c0fff711a4979a6d6R79) and made `reconcileRun` as an attribute to `PipelineRunTest`.

pkg/reconciler/pipelinerun/pipelinerun_test.go
pritidesai: Yup, very useful, instead of pointing back at the common utility (`reconcileRun`), it points back where the failure is originated 👍

pritidesai:
Based on @bobcatfish's comments, I have added these changes:

* Renamed `getReconciledRun` to `reconcileRun`
* added `t.Helper` to `reconcileRun`
* Created a new struct with all the arguments which were passed to `getReconciledRun` and define this common function as an attribute to `PipelineRunTest`

```
func (prt PipelineRunTest) reconcileRun() (*v1beta1.PipelineRun, test.Clients) {
```

pritidesai:
@bobcatfish its ready for review again, PTAL 😸  

There could be more refactoring done, mainly the way namespace and pipelineRun name are hard coded in the functions.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3753,3 +3456,40 @@ func getTaskRunStatus(t string, status corev1.ConditionStatus) *v1beta1.Pipeline
...
+		prt.Test.Fatalf("Error reconciling: %s", reconcileError)
+	}
+	// Check that the PipelineRun was reconciled correctly
+	reconciledRun, err := clients.Pipeline.TektonV1beta1().PipelineRuns(prt.Namespace).Get(prt.PRName, metav1.GetOptions{})
+	if err != nil {
+		prt.Test.Fatalf("Somehow had error getting reconciled run out of fake client: %s", err)
+	}
+
+	// Check generated events match what's expected
+	if len(prt.WantEvents) > 0 {
```
bobcatfish, line 3488: this is fairly minor but something feels weird to me about having WantEvents and PRName be attributes of PipelineRunTest - i think it's that this makes it so that if you want to call reconcileRun more than once in the same test, e.g. for two different pipelineruns, one reconciled after the other, you'd need to modify PipelineRunTest

I could see making 2 changes:
- Make WantEvents and PRName and Namespace arguments to this function
-  Have a NewPipelineRunTest function that returns an instance of PipelineRunTest and takes Test and Data as arguments

this way you can call NewPipelineRunTest to create your controller with the state it needs (store the testAssets as an attribute of PipelineRunTest) and you can call reconcileRun with as many runs as you want

pritidesai: Hey @bobcatfish, I have added [NewPipelineRunTest](https://github.com/tektoncd/pipeline/pull/2876/files#diff-f7c8532cfe79319c0fff711a4979a6d6R3633) and updated [reconcileRun](https://github.com/tektoncd/pipeline/pull/2876/files#diff-f7c8532cfe79319c0fff711a4979a6d6R3644) by adding function parameters (Namespace, PipelineRunName, wantEvents, and permanentError). 

Now, the PipelineRunTest has:

```
        test.Data  `json:"inline"`
	Test       *testing.T
	TestAssets test.Assets
	Cancel     func()
```

Also, have added `defer` on `Cancel` as part of each unit test instead of making it part of common function.

pritidesai: `Data` can be dropped from the `PipelineRunTest` struct since  `NewPipelineRunTest` is creating the controller and  instantiating `PipelineRunTest` with assets, `data` is not used/needed anywhere beyond `NewPipelineRunTest` - It makes `PipelineRunTest` complete (with inputs and assets) but can create separate PR to drop it if needed 😉

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3895,21 +3617,60 @@ func TestReconcile_CloudEvents(t *testing.T) {
...
-	err = checkCloudEvents(t, &ceClient, "reconcile-cloud-events", wantCloudEvents)
+	err := checkCloudEvents(t, &ceClient, "reconcile-cloud-events", wantCloudEvents)
 	if !(err == nil) {
 		t.Errorf(err.Error())
 	}
 }
+
+// NewPipelineRunTest returns PipelineRunTest with a new PipelineRun controller created with specified state through data
+// This PipelineRunTest can be reused for multiple PipelineRuns by calling reconcileRun for each pipelineRun
+func NewPipelineRunTest(data test.Data, t *testing.T) *PipelineRunTest {
```
bobcatfish, line 3633: i'm now just nitpicking, but i'd put the functions that are related to PipelineRunTest near the declaration of the struct. dont wanna block the PR tho!

pritidesai: thanks @bobcatfish I was planning on moving all helper functions including:

*  `PipelineRunTest` struct declaration and function related to it
* getPipelineRunController
* conditionCheckFromTaskRun
* checkEvents
* eventFromChannel

etc, to a separate file `/pkg/reconciler/pipelinerun/pipelinerun_helpers_test.go` as a follow up PR. `pkg/reconciler/pipelinerun/pipelinerun_test.go` is a combination of helper utilities and actual unit tests and helper functions are spread across this file 😞 Will be easier to introduce more tests and/or reviewing them with such separation.

bobcatfish: okay great!

A couple of thoughts:

* Somewhere in the top level [test](https://github.com/tektoncd/pipeline/tree/master/test) directory might be a good place to go, since that is where we are already putting our test libs there (might even make sense to work it into https://github.com/tektoncd/pipeline/blob/master/test/controller.go)
* I want to caution against naming and calling code "helpers" - this is totally a common practice but I think it's useful to get out of the habit, for a couple of reasons:
  * People tend to use this terminology about libs used for test code - sometimes they use it for the code under tests as well but i think that's less common - and it seems almost like it's putting the code down, i.e., "i need this code to help me" whereas I think that well factored test code is just as important as any other well factored code, so libraries for test code are just as important as other code < rant over >
  * The other (probably bigger and less my personal opinion) reason is that "helper" is not a very specific name - good naming and code organization requires thinking about what the code's responsibility is and naming and organizing it based on that; using names like "helper" "manager" "library" "util" is a short cut that prevents us from doing that. More on this at https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common and https://blog.golang.org/package-names in the section "bad package names":

> Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused. Over time, they accumulate dependencies that can make compilation significantly and unnecessarily slower, especially in large programs. And since such package names are generic, they are more likely to collide with other packages imported by client code, forcing clients to invent names to distinguish them.

Issue #2864, opened 2020-06-25
Validate conditions for affinity assistant to operate correctly


sbwsg:
# Expected Behavior

Tekton validates whether a TaskRun's or PipelineRun's volume configuration will be impossible to manage with the Affinity Assistant.

# Actual Behavior

Currently it's possible to get into a situation where [tasks wait forever to start](https://github.com/tektoncd/pipeline/issues/2829) due to use of multiple persistent volumes as workspaces. Ideally we should validate that this will happen and either a) reject the taskrun / pipelinerun with a clear validation error or b) fail the runs early with a clear error reason + message.

# Steps to Reproduce the Problem

1. See steps outlined in #2829

sbwsg:
I wonder if there is another option, c) decide that using the affinity assistant will not work and fallback to running without it, emitting a warning.

jlpettersson:
>using the affinity assistant for the multiple-workspace run

multiple-workspace run will work with the affinity assistant - as long as multiple workspaces is not used by the same task.

and multi-PVC-tasks should not be used in Tasks - they will most likely deadlock in regional clusters _sometimes_ as what happened in https://github.com/tektoncd/pipeline/pull/2546

sbwsg:
nice, thanks for clarifying!

jlpettersson:
I can have a look at this.

/assign

lmserrano:
I'm facing this issue for 2 nodes in a same single AZ and a pipelinerun which uses multiple pvcs and some tasks which also use multiple pvcs.

@jlpettersson , what is the suggested approach for avoiding multi-PVC tasks? In my use case, as part of a pipeline, I have a task which builds an application, and it does so based on 3 volumes: 1 with the git code and 2 with dependencies (which are retained across runs). Even if I put the dependencies together in a single pvc, I still can't make it work with less than 2, so I could use some guidance on how to best handle this. Thank you.

jlpettersson:
> @jlpettersson , what is the suggested approach for avoiding multi-PVC tasks?

As a generic solution, I suggests using subPaths of the same PVC volume. See the example with different subPaths: https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun-using-different-subpaths-of-workspace.yaml

[...]

Issue #2829, opened 2020-06-18
Task with more than one workspace waiting forever to start


r0bj:
# Expected Behavior

Task with more than one workspace starts.

# Actual Behavior

Task with more than one workspace waiting forever to start because it cannot attach volumes.

# Steps to Reproduce the Problem

1. There is a task (taskA) in pipeline which uses 2 workspaces
2. Pipelinerun starts, creates 2 different affinity-assistants
3. Because affinity-assistants use anti-affinity they are started on 2 different nodes
4. Each affinity-assistant mount volume on node where it runs, so volumes are mounted on different nodes
4. Task (taskA) cannot start because it uses affinity to one of the affinity-assistants and tries to attach 2 volumes which is impossible with `ReadWriteOnce` volumes
5. Task (taskA) waits forever to start

Is it possible to handle this situation with affinity-assistant enabled?

# Additional Info

- Tekton Pipeline version:
```
Client version: 0.10.0
Pipeline version: v0.13.2
Triggers version: v0.5.0
```

jlpettersson:
>Is it possible to handle this situation with affinity-assistant enabled?

No - by intention. When the Affinity Assistant was introduced, the docs was also updated so that we recommend _at most one_ writeable workspace per Task.

>Each affinity-assistant mount volume on node where it runs, so volumes are mounted on different nodes

And this will happen if e.g. a pipeline starts with two parallel tasks - even without the Affinity Assistant.

In the worst case - these two volumes are located in two different Availability Zones - and your Pipeline is deadlocked - with or without the Affinity Assistant. For this reason I had to remove an end-to-end-test using two volumes - that _sometimes_ ended up in to different Availability Zones in a Regional cluster: https://github.com/tektoncd/pipeline/pull/2546

If you really want to use two volumes in a single Task - you must disable the Affinity Assistant, but you will must likely run into problems in regional clusters - but it depends on how your storage work.

jlpettersson:
We should probably validate and emit an Event or use the Admission Webhook to validate these situation to inform the user in a better way.

jlpettersson:
@r0bj is there any specific reason to why you need two workspaces in the same Task?

It is possible to have a single workspace volume but mount different parts of it using `subPath`, there is an example of that: https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun-using-different-subpaths-of-workspace.yaml

Is that a way to solve your use case?

[...]

PR #2885, opened 2020-07-01
Validate TaskRun compatibility with the Affinity Assistant

jlpettersson:
# Changes

A TaskRun that mount more than one PVC-backed workspace is incompatible
with the Affinity Assistant. But there is no validation if the TaskRun
is compatible - so the TaskRun Pod is stuck with little information on why
to the user.

This commit adds validation of TaskRuns. When a TaskRun is associated with
an Affinity Assistant, it is checked that not more than one PVC workspace
is used - if so, the TaskRun will fail with a TaskRunValidationFailed condition.

Proposed in https://github.com/tektoncd/pipeline/issues/2829#issuecomment-645868118
Closes https://github.com/tektoncd/pipeline/issues/2864

/kind feature 
/release-note-none

# Submitter Checklist

review them:

## Reviewer Notes

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -294,6 +294,12 @@ func (c *Reconciler) prepare(ctx context.Context, tr *v1beta1.TaskRun) (*v1beta1
 		return nil, nil, controller.NewPermanentError(err)
 	}
 
+	if err := validateWorkspaceCompatibilityWithAffinityAssistant(tr); err != nil {
+		logger.Errorf("TaskRun %q workspaces are invalid: %v", tr.Name, err)
+		tr.Status.MarkResourceFailed(podconvert.ReasonFailedValidation, err)
+		return nil, nil, controller.NewPermanentError(err)
+	}
+
 	// Initialize the cloud events if at least a CloudEventResource is defined
 	// and they have not been initialized yet.
 	// FIXME(afrittoli) This resource specific logic will have to be replaced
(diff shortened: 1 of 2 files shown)


pkg/reconciler/taskrun/taskrun.go
```diff
@@ -700,3 +706,26 @@ func storeTaskSpec(ctx context.Context, tr *v1beta1.TaskRun, ts *v1beta1.TaskSpe
...
+
+	pvcWorkspaces := 0
+	for _, w := range tr.Spec.Workspaces {
+		if w.PersistentVolumeClaim != nil || w.VolumeClaimTemplate != nil {
+			pvcWorkspaces++
+		}
+	}
+
+	if pvcWorkspaces > 1 {
+		return fmt.Errorf("TaskRun mounts more than one PersistentVolumeClaim - that is forbidden when the Affinity Assistant is enabled")
```
sbwsg, line 728: Should this be "mounts more than one writeable PVC"?  I thought it was OK to receive multiple PVCs in read-only?  Or will that also trip into potential deadlock scenarios?

jlpettersson: That is a good question @sbwsg 

TLDR; as the message is written - it matches the current implementation of the Affinity Assistant.

The Affinity Assistant makes so that TaskRun pods are scheduled to the same Node. This actually helps to avoid two different problems:

1) All pods sharing a PVC (e.g. with accessMode ReadWriteOnce) can now execute in parallel - this may be possible with other accessModes e.g. ReadWriteMany even without the Affinity Assistant.
2) All pods sharing a PVC (e.g. with a zonal StorageClass) is scheduled to the same AZ, so it will not deadlock the pipeline - this may be possible with regional StorageClasses even without the Affinity Assistant - but on e.g. GCP regional storageClass volumes is only available on two of three AZs - so it may still be a problem. If using accessMode ReadWriteMany pods can execute in parallel within an AZ - but for regional clusters they additionally need to be regional StorageClass to avoid the problems that the Affinity Assistant solves.

The guidelines about "at most one writeable workspace" is for Tasks - e.g. they don't say if it is a PVC or emptyDir or regional StorageClass. With "writeable" I also meant PVCs - e.g. using a Secret or ConfigMap-workspace is still fine for Tasks - in addition to a PVC workspace.

I may add a new section to the workspace documentation to document more about this - it is a bit complicated.

jlpettersson: On the other hand - using a `ReadWriteOnce` in readOnly mode and regional storageClass - should work - but is currently not supported/implemented by the Affinity Assistant. We may want to implement it - but it is a bit complicated too - and probably a rarely used use case.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -297,6 +297,12 @@ func (c *Reconciler) prepare(ctx context.Context, tr *v1beta1.TaskRun) (*v1beta1
 		return nil, nil, controller.NewPermanentError(err)
 	}
 
+	if err := validateWorkspaceCompatibilityWithAffinityAssistant(tr); err != nil {
+		logger.Errorf("TaskRun %q workspaces are invalid: %v", tr.Name, err)
+		tr.Status.MarkResourceFailed(podconvert.ReasonFailedValidation, err)
+		return nil, nil, err
```
afrittoli, line 303: This should be a `PermanentError` otherwise the controller will requeue this.

sbwsg:
I'm going to add this to the 0.15 milestone as it would be great to get the validation in place in time for the next release.

sbwsg:
I'm also happy to pick this up @jlpettersson if you don't have time to finish.  let me know!

pkg/reconciler/taskrun/taskrun.go
jlpettersson: Thanks! Good catch!

jlpettersson:
> I'm also happy to pick this up @jlpettersson if you don't have time to finish. let me know!

@sbwsg I fixed the `PermanentError` - but haven't investigated if the tests can be refactored as @afrittoli suggested. Its up to you if you want to add those improvements in this PR or if it can be done in its own PR after this is merged.

Thanks @sbwsg and @afrittoli for review!

sbwsg:
Awesome, this looks great to me as it is but I'll leave it to @afrittoli to give the PermanentError another look.

vdemeester:
/lgtm
/hold
@afrittoli can you take a look :upside_down_face:

PR #2891, opened 2020-07-02
restricting tekton.dev labels in metadata

pritidesai:
# Changes

Tekton resources are automatically assigned certain labels including the ones with [tekton.dev](https://github.com/tektoncd/pipeline/blob/master/docs/labels.md#automatically-added-labels). The labels are created and propagated from one resource to another using this group name. We should restrict possibility of overwriting those labels otherwise it could result in conflicting resources.

Result of the discussion in PR [2826](https://github.com/tektoncd/pipeline/pull/2826#discussion_r448683708)

On hold until we confirm this is something we want to restrict 🤔  

This will break folks using these labels in metadata, see an [example](https://github.com/tektoncd/catalog/tree/master/github#usage) in catalog.

/cc @vdemeester 

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Restricting user defined tekton.dev/ labels.

```

### DIFF
--- a/pkg/apis/validate/metadata.go
+++ b/pkg/apis/validate/metadata.go
@@ -19,6 +19,8 @@ package validate
 import (
 	"strings"
 
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/pkg/apis"
 )
@@ -39,6 +41,16 @@ func ObjectMetadata(meta metav1.Object) *apis.FieldError {
 		return &apis.FieldError{
 			Message: "Invalid resource name: length must be no more than 63 characters",
 			Paths:   []string{"name"},
+		}
+	}
+
+	// metadata must not contain any label with tekton.dev/*
+	for k := range meta.GetLabels() {
+		if strings.HasPrefix(k, pipeline.GroupName) {
+			return &apis.FieldError{
+				Message: "Invalid label key: can not change tekton resource name",
+				Paths:   []string{"labels"},
+			}
 		}
 	}
 	return nil
--- a/pkg/apis/validate/metadata_test.go
+++ b/pkg/apis/validate/metadata_test.go
@@ -20,15 +20,19 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
+
 	"github.com/tektoncd/pipeline/pkg/apis/validate"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func TestMetadataInvalidLongName(t *testing.T) {
+func TestMetadata_Failure(t *testing.T) {
 
 	invalidMetas := []*metav1.ObjectMeta{
 		{Name: strings.Repeat("s", validate.MaxLength+1)},
 		{Name: "bad.name"},
+		{Labels: map[string]string{pipeline.GroupName: "mytekton"}},
+		{Labels: map[string]string{pipeline.GroupName + pipeline.TaskLabelKey: "mytask"}},
 	}
 	for _, invalidMeta := range invalidMetas {
 		if err := validate.ObjectMetadata(invalidMeta); err == nil {


pritidesai:
/kind misc
/hold

Need to confirm this is something we want to restrict 🤔

vdemeester:
> This will break folks using these labels in metadata, see an [example](https://github.com/tektoncd/catalog/tree/master/github#usage) in catalog.

Interesting, those catalog task shouldn't use that label anyway as it is supposed to be set by the controller :thinking:

pritidesai:
> > This will break folks using these labels in metadata, see an [example](https://github.com/tektoncd/catalog/tree/master/github#usage) in catalog.
> 
> Interesting, those catalog task shouldn't use that label anyway as it is supposed to be set by the controller 🤔

Yup, indeed, @bobcatfish @sbwsg WDYT?

dlorenc:
> Yup, indeed, @bobcatfish @sbwsg WDYT?

Ref https://github.com/tektoncd/pipeline/pull/2946 - we should probably document this there too

dlorenc:
> Yup, indeed, @bobcatfish @sbwsg WDYT?

Is it actually "used" there? The catalog reorg happened so I had to hunt them down (the links changed). Here it is now: https://github.com/tektoncd/catalog/blob/d7e5c6db536d79bd3c205caaf075a50f5c9dee7e/task/github-close-issue/0.1/README.md

They only really appear in the READMEs for examples on creating a TaskRun. My wild guess is that someone created the taskrun with the CLI or something, then dumped it as YAML to generate the docs and kept that label.

It's only in 3 places: 
image

pkg/apis/validate/metadata.go
```diff
@@ -41,5 +43,15 @@ func ObjectMetadata(meta metav1.Object) *apis.FieldError {
 			Paths:   []string{"name"},
 		}
 	}
+
+	// metadata must not contain any label with tekton.dev/*
+	for k := range meta.GetLabels() {
+		if strings.HasPrefix(k, pipeline.GroupName) {
```
dlorenc, line 49: Do we care/have an opinion on things like "*.tekton.dev" as well?

pritidesai: This is meant to be Tekton internal, created and utilized by the pipeline itself, better to enforce rather than relying on `Task/Pipeline/TaskRun/PipelineRun` authors to follow recommendations, thoughts?

vdemeester: @dlorenc good point. I would think we should reserve `tekton.dev` *and* most likely `pipelines.tekton.dev` (future-proof :sweat_smile: ), but given some [proposals](https://github.com/tektoncd/community/pull/147), we shouldn't restrict all the `*.tekton.dev` :upside_down_face:

dlorenc: +1 on saving the extra namespace. In chains right now we're using "chains.tekton.dev", but only in an annotation.

pritidesai: This PR is  restricting labels, annotations have all the freedom 😉

vdemeester:
cc @tektoncd/core-maintainers

dibyom:
> Tekton resources are automatically assigned certain labels including the ones with tekton.dev. The labels are created and propagated from one resource to another using this group name. We should restrict possibility of overwriting those labels otherwise it could result in conflicting resources.

Sounds good to me!

pritidesai:
no changes applied, rebased this branch with latest master to see if it can fix integration tests.

pkg/apis/validate/metadata.go
```diff
@@ -41,5 +43,15 @@ func ObjectMetadata(meta metav1.Object) *apis.FieldError {
...
+			return &apis.FieldError{
+				Message: "Invalid label key: can not change tekton resource name",
```
afrittoli, line 51: NIT: the error message may be confusing if someone is not trying to change the resource name. I think it would be better to say that the `tekton.dev/` labels are reserved for internal use.

afrittoli:
We do use `tekton.dev` labels in plumbing, so I need to change those before we install this :)

afrittoli (commented):
Thanks for this!
I would update the error message, but otherwise it looks good to me.
I think we can reserve extra namespace in further PRs if we decide to?
@pritidesai is the "hold" still needed on this one?

pritidesai:
thanks. Nope, no more hold needed.
/hold cancel

pritidesai:
> We do use `tekton.dev` labels in plumbing, so I need to change those before we install this :)

Could this be causing integration failure? 🤔 Integration tests failure is consistent with these changes in terms of number of failing Tests: 

```
162 failed / 9 succeeded
```

pritidesai:
16 test examples showing this error:

```
TestExamples/v1alpha1/pipelineruns/conditional-pipelinerun-with-same-condition-refer: examples_test.go:57: Failed waiting for pipeline run done: timed out waiting for the condition
```

pritidesai:
This is actually a legit failure 😢 not related to infra or timeout.

Taskrun and Pipelinerun controller injects these labels which are then going through the same validation routine which validates user specified labels. Will update this PR to separate this validation calls next week, putting it on hold until then.

/hold

vdemeester:
/remove-lifecycle stale
@tektoncd/core-maintainers what should we do here ?

vdemeester:
@pritidesai @tektoncd/core-maintainers gentle ping 👼🏼

pritidesai:
thanks @vdemeester its a long overdue, will look into it this week 🙏

pritidesai:
closing this for now, will open a new PR once I have a breather to implement this. I have created an issue capturing the PR discussion [here](https://github.com/tektoncd/pipeline/issues/4366).

PR #2915, opened 2020-07-08
Define a helper method pkg/controller.FilterRunRef

ImJasonH:
This will be useful to external Custom Task controllers to filter out
update notifications for Runs of irrelevant types.

This is used in (and copied from!) my example CEL task, for instance: https://github.com/ImJasonH/cel-task/blob/6c75268b6b49859f3680ceab0cbbf938191a9606/main.go#L50

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/controller/filter.go
+++ b/pkg/controller/filter.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2020 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package controller provides helper methods for external controllers for
+// Custom Task types.
+package controller
+
+import (
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/controller/filter.go
```diff
@@ -0,0 +1,52 @@
...
+// about.
+//
+// For example, a controller impl that wants to be notified of updates to Runs
+// which reference a Task with apiVersion "example.dev/v0" and kind "Example":
+//
+//     runinformer.Get(ctx).Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+//       FilterFunc: FilterRunRef("example.dev/v0", "Example"),
+//       Handler:    controller.HandleAll(impl.Enqueue),
+//     })
+func FilterRunRef(apiVersion, kind string) func(interface{}) bool {
```
pritidesai, line 36: Is this intended to be something specific for only `Run` or extend the same filtering to `taskRun` and/or `pipelineRun`? I am wondering since the package name is generic 🤔

ImJasonH: This method is only intended to be used for `Run`s, not `TaskRun`s or `PipelineRun`s. I was hoping the func name and doc comment would make that clearer, if you have ideas to improve that let me know.

pritidesai: nope, I have no suggestions at this point, seeking opinion from experts 😉 

Trying to understand the usage of this function and if we can create and apply same for `taskRun` and `pipelineRun`.

ImJasonH: There isn't really a similar need for taskruns and pipelineruns, since we generally expect our own controllers to be interested in receiving updates about those objects.

Or if some other controller cared to get notified of updates (the experimental Results watcher will do this, for instance), it will likely only care about getting _all_ updates to _all_ taskruns/pipelineruns, not just those that match some criteria, like we expect custom task authors to want.

PR #2918, opened 2020-07-08
Implement basic no-op reconciler for all Runs

ImJasonH:
This wires up a reconciler in the core controller binary that is
notified of changes to and creations of all Run objects of any type. It
currently does nothing in response to these events, but in the future
this will be used to enforce initial update timeouts, and possibly more
default behavior in the future.

# Submitter Checklist

review them:

- [n] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [n] Release notes block has been filled in or deleted (only if no user facing changes)

## Reviewer Notes

# Release Notes

```release-note
NONE
```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	"github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun"
+	"github.com/tektoncd/pipeline/pkg/reconciler/run"
 	"github.com/tektoncd/pipeline/pkg/reconciler/taskrun"
 	corev1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/injection"
@@ -74,5 +75,6 @@ func main() {
 	sharedmain.MainWithContext(injection.WithNamespaceScope(signals.NewContext(), *namespace), ControllerLogKey,
 		taskrun.NewController(*namespace, images),
 		pipelinerun.NewController(*namespace, images),
+		run.NewController,
 	)
 }
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/tektoncd/pipeline
 go 1.13
 
 require (
+	cloud.google.com/go/storage v1.6.0
 	contrib.go.opencensus.io/exporter/stackdriver v0.13.1 // indirect
 	github.com/GoogleCloudPlatform/cloud-builders/gcs-fetcher v0.0.0-20191203181535-308b93ad1f39
 	github.com/aws/aws-sdk-go v1.30.16 // indirect
@@ -25,6 +26,7 @@ require (
 	golang.org/x/crypto v0.0.0-20200323165209-0ec3e9974c59 // indirect
 	golang.org/x/oauth2 v0.0.0-20200107190931-bf48bf16ab8d
 	gomodules.xyz/jsonpatch/v2 v2.1.0
+	google.golang.org/api v0.20.0
 	google.golang.org/protobuf v1.22.0 // indirect
 	gopkg.in/yaml.v2 v2.3.0 // indirect
 	k8s.io/api v0.17.6
--- a/pkg/reconciler/run/controller.go
+++ b/pkg/reconciler/run/controller.go
@@ -1,0 +1,61 @@
+/*
+Copyright 2020 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package run
+
+import (
+	"context"
+
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
+	pipelineclient "github.com/tektoncd/pipeline/pkg/client/injection/client"
+	runinformer "github.com/tektoncd/pipeline/pkg/client/injection/informers/pipeline/v1alpha1/run"
+	runreconciler "github.com/tektoncd/pipeline/pkg/client/injection/reconciler/pipeline/v1alpha1/run"
+	"k8s.io/client-go/tools/cache"
+	"knative.dev/pkg/configmap"
+	"knative.dev/pkg/controller"
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/tracker"
+)
+
+// NewController instantiates a new controller.Impl from knative.dev/pkg/controller
+func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
+	logger := logging.FromContext(ctx)
+	pipelineclientset := pipelineclient.Get(ctx)
+	runInformer := runinformer.Get(ctx)
+
+	c := &Reconciler{
+		pipelineClientSet: pipelineclientset,
+	}
+	impl := runreconciler.NewImpl(ctx, c, func(impl *controller.Impl) controller.Options {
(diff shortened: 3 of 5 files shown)


cmd/controller/main.go
```diff
@@ -74,5 +75,6 @@ func main() {
 	sharedmain.MainWithContext(injection.WithNamespaceScope(signals.NewContext(), *namespace), ControllerLogKey,
 		taskrun.NewController(*namespace, images),
 		pipelinerun.NewController(*namespace, images),
+		run.NewController,
```
pritidesai, line 78: sorry coming late to the party 😢 but staring at this `run` here looks incomplete, is it possible to rename it to something descriptive? 😜

pkg/reconciler/run/controller.go
```diff
@@ -0,0 +1,61 @@
...
+	runreconciler "github.com/tektoncd/pipeline/pkg/client/injection/reconciler/pipeline/v1alpha1/run"
+	"k8s.io/client-go/tools/cache"
+	"knative.dev/pkg/configmap"
+	"knative.dev/pkg/controller"
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/tracker"
+)
+
+// NewController instantiates a new controller.Impl from knative.dev/pkg/controller
+func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
```
pritidesai, line 35: I am still trying to understand how controller initialization is happening in general. Here `NewController` is returning   implementation of the controller whereas `taskRun` and `pipelineRun` controllers are returning function:

https://github.com/tektoncd/pipeline/blob/732d2363b9df0b5c2e6b1de732a819b69c70bb3f/pkg/reconciler/taskrun/controller.go#L45-L46

https://github.com/tektoncd/pipeline/blob/732d2363b9df0b5c2e6b1de732a819b69c70bb3f/pkg/reconciler/pipelinerun/controller.go#L45-L46

Why does this defer from the `pipelineRun` and `taskRun`?

pkg/reconciler/run/run.go
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2019 The Tekton Authors
```
pritidesai, line 2: may be `2020` similar to one in `controller.go` ⬆️ ?

pkg/reconciler/run/run.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+const (
+	// runAgentName defines logging agent name for Run Controller
+	runAgentName = "run-controller"
+)
+
+// Reconciler implements controller.Reconciler for Run resources.
+type Reconciler struct {
+	pipelineClientSet clientset.Interface
```
pritidesai, line 37: `s/pipelineClientSet/runClientSet/` 🤔

pkg/reconciler/run/run.go
```diff
@@ -0,0 +1,54 @@
...
+	tracker           tracker.Interface
+}
+
+// Check that our Reconciler implements runreconciler.Interface
+var _ runreconciler.Interface = (*Reconciler)(nil)
+
+// Reconcile compares the actual state with the desired, and attempts to
+// converge the two. It then updates the Status block of the Task Run
```
pritidesai, line 45: s/Task Run/Run/

pkg/reconciler/run/controller.go
vdemeester: Good point. I think we may want to follow the same pattern. In PipelineRun and TaskRun reconciler, this is mainly because we need to pass arguments (and the func return to satisfy the knative/pkg controller contract). We should support at least the `namespace` for `Run` too I guess ? (I don't think we need to pass the `images`)

cmd/controller/main.go
ImJasonH: The `run` package is generated based on the name of the type, which is just `Run`. I can't really think of a better name for it, but ideas are welcome. An earlier iteration of the design called it `CustomTaskRun` but that got to be a mouthful/keyboardful...

pkg/reconciler/run/controller.go
ImJasonH: I had considered supporting namespaced-reconciliation for Runs in this PR but thought it would be simpler not to, to keep this change small. If you think it makes it easier to read here, though, I can do that.

pkg/reconciler/run/run.go
ImJasonH: This is actually named `pipeline` because it's the "Tekton Pipelines" client -- it provides (generated) methods to interact with _all_ "Tekton Pipelines" resources and API versions: https://godoc.org/github.com/tektoncd/pipeline/pkg/client/clientset/versioned

It's also named `pipelineClientSet` in the TaskRun controller too, for example: https://github.com/tektoncd/pipeline/blob/732d2363b9df0b5c2e6b1de732a819b69c70bb3f/pkg/reconciler/taskrun/taskrun.go#L64

ImJasonH:
```
==================================
==== INTEGRATION TESTS PASSED ====
==================================
```
/retest 
🤔

cmd/controller/main.go
pritidesai: is it distinguishable as a `Task` or a `Pipeline`?  How about `CustomRun`? 😜

pkg/reconciler/run/controller.go
pritidesai: Yup would be great to exercise the same contract and easier to follow. If it cannot be part of this PR, at least a comment would be helpful 🙏

pkg/reconciler/run/run.go
pritidesai: thanks for the explanation 👍

cmd/controller/main.go
ImJasonH: The general idea is that a `Run` reports the execution of a generic...something. What that is is up to the custom task author -- it could be running a remote build process, "running" a wait between two other Tasks, "running" an approval.

It's sort of like a generic container to express an ongoing execution of a thing. In the future we could experiment with having `Run<Task>` and `Run<Pipeline>` in addition to (eventually replacing?) `TaskRun` and `PipelineRun`. This would give us flexibility to experiment with new definitions of `Task` and `Pipeline` outside the `tektoncd/pipeline` repo, or to mix versions of these types (e.g., `Run<v1alpha1.Task>` alongside `Run<v1.Task>` in the same `PipelineRun`).

So I think even `CustomRun` is still a bit wordy for what it could end up supporting. What's a "custom" run in a possible future world where that's actually the _only_ kind of running thing? I personally actually kind of like the succintness of "`Run`"  in that context. 🤷‍♂️

pritidesai: cool, sounds good, thanks 👍

pkg/reconciler/run/controller.go
```diff
@@ -0,0 +1,61 @@
...
+
+	logger.Info("Setting up event handlers")
+	runInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    impl.Enqueue,
+		UpdateFunc: controller.PassNew(impl.Enqueue),
+	})
+	c.tracker = tracker.New(impl.EnqueueKey, controller.GetTrackerLease(ctx))
+
+	return impl
+}
```
bobcatfish, line 61: if the intention of this controller is for folks to take it and build on it (I think? side note im not quite sure what this controller is for from the PR description / commit message) then maybe it would make sense to include some comments for every few lines explaining whats going on?

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/tektoncd/pipeline
 go 1.13
 
 require (
+	cloud.google.com/go/storage v1.6.0
```
dlorenc, line 6: Any idea why this was added? It doesn't appear used anywhere.

vdemeester: I gets added and removed as will (or random :sweat_smile:). That is one of the *high* mistery of go modules with k8s I feel :sweat_smile:

pritidesai: no idea but leaves us with `git checkout -- go.mod` 😢

vdemeester:
@ImJasonH what is the status for this ?

ImJasonH:
I think there were questions about whether we wanted to have Tekton reconcile anything about Runs, which leaves this in limbo.

I'm fine to close now and reopen (or something like it) if we decide to have Tekton reconcile anything.

/close

Issue #1805, opened 2020-01-05
Config map names should be customizable, or at least prefixable


abayer:
# Expected Behavior

The various config maps used by Tekton Pipeline (`config-artifact-bucket`, `config-logging`, etc) should be able to be given custom names, to avoid potential collisions with other services.

# Actual Behavior

They can't. =)

# Steps to Reproduce the Problem

1. Change the `name` of one of the config maps.
2. Anything referring to that config map won't work.

# Additional Info

For the standard usage, this isn't a problem, since the config maps go into the `tekton-pipelines` namespace, but in Jenkins X, we put everything in the `jx` namespace along with...lots of other junk. It'd be really handy if we could at the least prefix the config map names to guarantee uniqueness. I haven't thought through the implementation details, but maybe something like an optional argument to the `controller` and `webhook` to provide a prefix for the various config map names would work.

abayer:
Looks like the prefix idea would be a challenge for `config-observability` since that config map is only referenced from `knative.dev/pkg` code...

bobcatfish:
I think we'd be open to this if anyone wants to do the work!

jerop:
@dibyom and I looked into this, and found that all config maps are customizable except for `feature-flags` config map which is fixed in #2363

PR #2930, opened 2020-07-10
Add Default TaskRun Workspace Bindings to config-default

### LINKED ISSUE #2398: Make emptyDir the default volume source for workspaces in TaskRun
In the case when we have a `Task` with a workspace, e.g.

```
piVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: task
spec:
      steps:
      - name: write
        image: ubuntu
        script: echo $(workspaces.task-ws.volume) > $(workspaces.task-ws.path)/foo
      workspaces:
      - name: task-ws
```
# Expected Behavior

That Tekton could provide me with good default values when possible and the convention is a good fit for the problem.

When I create a `TaskRun` _without_ defining a workspace volume, I can get a working solution by convention.

Example (YAML) but use case is useful for e.g. **cli** as well:

```
kind: TaskRun
metadata:
  generateName: tr
spec:
  taskRef:
    name: task
  workspaces:
  - name: task-ws
```

And the TaskRun should successfully run.

# Actual Behavior

A volume for the workspace _must_ be specified.

```
kubectl create -f tr.yaml 
Error from server (BadRequest): error when creating "tr.yaml": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: expected exactly one, got neither: workspace.configmap, workspace.emptydir, workspace.persistentvolumeclaim, workspace.secret, workspace.volumeclaimtemplate
```

A user can solve it by adding `emptyDir{}` for the workspace in the `TaskRun`

Example:

```
kind: TaskRun
metadata:
  generateName: tr
spec:
  taskRef:
    name: task
  workspaces:
  - name: task-ws
    emptyDir: {}
```

# Proposed solution

For a `TaskRun` an `emptyDir{}` is almost always possbile to use in most clusters (as what I know) and it is a fair use case, _if the workspace is used for passing files between steps_

However, for `PipelineRun` an `emptyDir{}` is less useful as a workspace since it is combined of multiple pods.

Providing an `emptyDir{}` volume as default was [suggested](https://github.com/tektoncd/pipeline/issues/1986#issuecomment-604971446) in the discussion about "Auto workspace"

When e.g. debugging a task in an existing pipeline, this is a useful, quick way to run a task.

### LINKED ISSUE #2595: Default workspace types for PipelineRuns and TaskRuns
# Expected Behavior

Following up on https://github.com/tektoncd/pipeline/issues/2398

It would be useful to be able to specify default `Workspace` types. The idea would be that users don't need to specify `workspaces` they don't care about (or whose configuration should be entirely in the hands of admins) and Tekton could just inject a default bit of `volume` configuration when it launches a `TaskRun` or `PipelineRun`.

It seems like one approach to this would be to introduce a `ConfigMap` (maybe per-namespace for different tenants?) that allows a user to set the default `WorkspaceBinding`. It probably makes sense to have separate configuration for `PipelineRun` and `TaskRun` since `TaskRuns` can execute happily with an `emptyDir` but `PipelineRuns` would probably want the option to use something like a `volumeClaimTemplate` by default.

If we were to also introduce [optional `Workspaces`](https://github.com/tektoncd/pipeline/issues/2119) then I'm not sure what the most desirable behaviour would be - would a default `WorkspaceBinding` be injected for optional `Workspaces` or would they be left empty?

# Actual Behavior

At the moment users have to completely specify `WorkspaceBindings` in their Runs.

Also relates to [the existing config-artifact-pvc](https://github.com/tektoncd/pipeline/issues/2330) which are used by PipelineResources.

jerop:
# Changes

Currently, users have to completely specify `Workspaces` they don't care 
about or whose configuration should be entirely in the hands of admins.

This PR enables users to specify default `Workspaces` for `TaskRuns`, 
for example they can use `emptyDir` by default.

The `WorkspaceBinding` can be set in the `config-defaults` ConfigMap 
in `default-task-run-workspace-binding`.

Fixes #2398 and partially fixes #2595.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```release-note
Users can now set a default `Workspace` configuration for any `Workspaces` that a Task declares but that a TaskRun does not explicitly provide. It can be set in the `config-defaults` ConfigMap in `default-task-run-workspace-binding`.
```

/cc sbwsg

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -84,6 +84,8 @@ spec:
         # If you are changing these names, you will also need to update
         # the controller's Role in 200-role.yaml to include the new
         # values in the "configmaps" "get" rule.
+        - name: CONFIG_DEFAULTS_NAME
+          value: config-defaults
         - name: CONFIG_LOGGING_NAME
           value: config-logging
         - name: CONFIG_OBSERVABILITY_NAME
--- a/pkg/apis/config/default_test.go
+++ b/pkg/apis/config/default_test.go
@@ -172,6 +172,26 @@ func TestEquals(t *testing.T) {
 			},
 			expected: true,
 		},
+		{
+			name: "different default workspace",
+			left: &config.Defaults{
+				DefaultTaskRunWorkspaceBinding: "emptyDir: {}",
+			},
+			right: &config.Defaults{
+				DefaultTaskRunWorkspaceBinding: "source",
+			},
+			expected: false,
+		},
+		{
+			name: "same default workspace",
+			left: &config.Defaults{
+				DefaultTaskRunWorkspaceBinding: "emptyDir: {}",
+			},
+			right: &config.Defaults{
+				DefaultTaskRunWorkspaceBinding: "emptyDir: {}",
+			},
+			expected: true,
+		},
 	}
 
 	for _, tc := range testCases {
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -24,7 +24,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ghodss/yaml"
 	"github.com/hashicorp/go-multierror"
+	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/pipeline/pkg/apis/resource"
@@ -288,6 +290,12 @@ func (c *Reconciler) prepare(ctx context.Context, tr *v1beta1.TaskRun) (*v1beta1
 		return nil, nil, controller.NewPermanentError(err)
 	}
 
+	if err := c.updateTaskRunWithDefaultWorkspaces(ctx, tr, taskSpec); err != nil {
+		logger.Errorf("Failed to update taskrun %s with default workspace: %v", tr.Name, err)
+		tr.Status.MarkResourceFailed(podconvert.ReasonFailedResolution, err)
+		return nil, nil, controller.NewPermanentError(err)
+	}
+
 	if err := workspace.ValidateBindings(taskSpec.Workspaces, tr.Spec.Workspaces); err != nil {
 		logger.Errorf("TaskRun %q workspaces are invalid: %v", tr.Name, err)
 		tr.Status.MarkResourceFailed(podconvert.ReasonFailedValidation, err)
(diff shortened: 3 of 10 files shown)


pkg/reconciler/taskrun/taskrun.go
```diff
@@ -289,6 +291,7 @@ func (c *Reconciler) prepare(ctx context.Context, tr *v1beta1.TaskRun) (*v1beta1
 		return nil, nil, err
 	}
 
+	c.updateTaskRunWithDefaultWorkspaces(ctx, tr, taskSpec)
```
sbwsg, line 298: This function could return an error so I wonder if we should log it and mark the taskrun as failed if that happens. What do you think?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -2727,6 +2727,51 @@ func TestReconcileWorkspaceMissing(t *testing.T) {
 	}
 }
 
+// TestReconcileDefaultWorkspace tests a reconcile of a TaskRun that does
+// not include a Workspace that the Task is expecting.
+func TestReconcileDefaultWorkspace(t *testing.T) {
```
jerop, line 2732: @sbwsg included this test for when default is added, and right above is a test for when workspace is missing and the default is not provided

pkg/apis/config/default_test.go
```diff
@@ -172,6 +172,26 @@ func TestEquals(t *testing.T) {
 			},
 			expected: true,
 		},
+		{
```
jerop, line 175: @sbwsg and this test too

pkg/reconciler/taskrun/taskrun.go
jerop: Agreed, will add that, thanks!

jerop: Done ✅

config/controller.yaml
```diff
@@ -91,6 +91,8 @@ spec:
         # If you are changing these names, you will also need to update
         # the controller's Role in 200-role.yaml to include the new
         # values in the "configmaps" "get" rule.
+        - name: CONFIG_DEFAULTS_NAME
+          value: config-defaults
```
vdemeester, line 95: Interesting, why wasn't this here before ? (or why is it needed now ?)
/cc @afrittoli

sbwsg: We have a fallback of "config-defaults" which is returned [here](https://github.com/tektoncd/pipeline/blob/0b345b56bda47c4542519102527889f7d0777a33/pkg/apis/config/default.go#L54-L59), which is likely why it's never caused problems not being here before.

I think I added this while Jerop and I were pairing on these changes. This is just a drive-by fix and not a strict requirement of this PR (I don't think - @jerop I might be misremembering?)

sbwsg:
I've reviewed the test changes and they look good to me!  I'm going to hold off on approving or lgtming the PR since I helped write a bit of it, would be good to get others' 👁️  on it for feedback.

config/controller.yaml
jerop: yes, that's right, we thought to fix it but not needed in this PR

why it's needed in general is that users may need to give the config maps custom names to avoid collisions with other services, as described in issue https://github.com/tektoncd/pipeline/issues/1805

afrittoli: The name of config maps used in the config store is configurable via environment variable today: https://github.com/tektoncd/pipeline/blob/93a4e5fa8d4ec8f7d099a5e9ad80bf79405c1507/pkg/apis/config/default.go#L54-L59 . Having the environment variable in the controller.yaml makes it easier for users who may want to change the default name. It's nice to have, but not related to this PR.
I don't mind having it in this PR though.

sbwsg:
Given that there have been two approvals now I'm going to lgtm it as well.

PR #2938, opened 2020-07-14
Add pipeline run support for cloud events

### LINKED ISSUE #2082: Optionally emit events during execution

afrittoli:
# Changes

Replace the pipeline run controller own config store with the
shared one used by the taskrun controller too. The pipeline run
config store is only useful to the artifact storage, however
the artifact storage loads the config by fetching the configmap
via the kube client, so it does not use the config store.

Attaching the shared config store to the controller, along with
the cloud events client, enables the pipeline run controller to
start sending cloud events for all events where we send k8s events
today (except for error ones).

Add a reconciler unit test to verify that events are sent when
the sink is configured.

Fixes #2082

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Add cloud events for pipeline runs via the  configuration option `default-cloud-events-sink`.
When the default sink is setup, cloud events are sent, else they're disabled.

Send cloud events (when enabled) from the pipeline run and controller, for all non-error
events that are already notified via k8s events.
```


pkg/reconciler/pipelinerun/controller.go
```diff
@@ -31,7 +32,7 @@ import (
 	pipelinerunreconciler "github.com/tektoncd/pipeline/pkg/client/injection/reconciler/pipeline/v1beta1/pipelinerun"
 	resourceinformer "github.com/tektoncd/pipeline/pkg/client/resource/injection/informers/resource/v1alpha1/pipelineresource"
 	"github.com/tektoncd/pipeline/pkg/reconciler"
-	"github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun/config"
```
vdemeester, line 34: Is this package still used ? (if not we should remove it :angel: )

afrittoli: Oh, I was sure I removed that... Will do now.
It looks like the first import wins, it wouldn't work when using the second. I'm surprised neither the compiler nor the linter catches that?

afrittoli: oh, never mind, I was looking on the phone, I thought this on the right side of the diff :P

afrittoli: I removed the package completely now.
It was not really used for a while, I couldn't find a place in the code where it was used going back in the code for a while. `InitializeArtifactStorage` always took a k8s client as input and no context, so there's no way it could access the store at any point.

Issue #2791, opened 2020-06-09
v0.13.0: unsuccessful cred copy: ".ssh" from "/tekton/creds" to "/root"


bitsofinfo:
# Expected Behavior

I should be able to **opt-out of the auto seeding of tekton creds** into each step and just mount my own secrets via workspaces under `/root`. Let the user have complete control and disable some of the automatic stuff tekton does around this. For steps which use images that rely on specific things in the user's home dir (in this case `/root`) I need to craft this directory in a specific way and this tekton cred stuff gets in the way.

# Actual Behavior

Task goes into `Error` state after 1st step. Each step in my task that does this ends up w/ this in logs. NOTE the serviceAccount this is running as has NO bound secrets.

```
2020/06/09 13:47:33 unsuccessful cred copy: ".ssh" from "/tekton/creds" to "/root": unable to open destination: open /root/.ssh/known_hosts: read-only file system
```

Overall in the tekton dashboard, the task just appears to be in a hung state.

![image](https://user-images.githubusercontent.com/1743646/84158226-df4cfd80-aa28-11ea-9f02-90947c8aa779.png)

# Steps to Reproduce the Problem

Note I have `"disable-home-env-overwrite":"true","disable-working-directory-overwrite":"true"`

1. Mount a secret backed workspace to a task at `/root/.ssh`
2. Start a pipelinerun for that task

My trigger workspaces:
```
        workspaces:
          - name: ssh-secrets
            secret:
              secretName: my-ssh-config 
              items:
                - key: ssh-privatekey
                  path: id_rsa
                  mode: 256
                - key: known_hosts
                  path: known_hosts
                - key: ssh-config
                  path: config
```

The task def workspaces:
```
    - name: ssh-secrets
      description: The workspace where we can find ssh secrets
      mountPath: /root/.ssh
      readOnly: true
```

# Additional Info

The `serviceAccount` that the pipelinerun uses, has no `secrets` attatched to it.

- Kubernetes version:

1.16.9

- Tekton Pipeline version: 0.13.0

sbwsg:
> this tekton cred stuff gets in the way.

Best, most succinct, description of creds-init I've read yet. Thanks for the issue report!

sbwsg:
In terms of the "hang" though - could you post the sanitized Pod YAML?  The errors you have in the issue report "shouldn't" cause hangs so I'd like to try reproducing to see if I can figure out what's going wrong.

bitsofinfo:
[...]

PR #2940, opened 2020-07-14
Dont write creds-init files if none of that type are mounted

sbwsg:
# Changes

This commit fixes the immediate bug described in https://github.com/tektoncd/pipeline/issues/2791, but does not implement the feature request. I'll leave that for a follow-up.

Currently our creds-init code builds files on disk even when a credential
of a given type doesn't exist. For example, a .ssh/known_hosts file
is written even if there are no git ssh keys mounted in the step container.
It will be an empty file. The same is true for .docker/config.json and
.gitconfig.

This commit changes our creds init code to only write files for credentials
that have actually been mounted from correctly annotated secrets.

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/examples/v1beta1/taskruns/creds-init-only-mounts-provided-credentials.yaml
+++ b/examples/v1beta1/taskruns/creds-init-only-mounts-provided-credentials.yaml
@@ -1,0 +1,46 @@
+# This example mounts a fake git ssh credential into a Task
+# and confirms that only the expected credential appears
+# on the Task container's disk.
+apiVersion: v1
+kind: Secret
+type: kubernetes.io/ssh-auth
+metadata:
+  name: phony-ssh-key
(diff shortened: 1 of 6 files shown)


sbwsg:
/hold

I will add some tests for this.

sbwsg:
/hold cancel

I've added an example YAML that exercises the expected behaviour.  Using an example YAML feels like the cleanest approach given that the directory contents inside the Task's container needs to be validated but I'm open to suggestions here if there's a better way.

bobcatfish:
@sbwsg if this fixes a bug does it make sense to include some release notes?

sbwsg:
Updated with release note and flipped labels on PR to indicate one is included.

Issue #2936, opened 2020-07-14
Tekton's Results don't get trimmed


soharaki:
# Expected Behavior

Tekton's Results have the line feeds.
Some error occurs when we want to pass parameters directly from the results, for example Kaniko.

# Actual Behavior

Tekton's Results have no line feeds.

# Example Test Codes

```yaml
apiVersion: tekton.dev/v1alpha1
kind: Pipeline
metadata:
  name: openshift-handson-line-feed-results
  namespace: handson-example
spec:
  tasks:
    - name: get-result
      taskSpec:
        steps:
          - name: set-results
            image: goreleaser/goreleaser # any images which have the shell
            script: |
              echo "123" > /tekton/results/pre
              echo "abc" > /tekton/results/post
        results:
        - name: pre
        - name: post
    - name: checkresults
      runAfter: ["get-result"]
      taskSpec:
        params:
          - name: IMAGETAG
        steps:
          - name: check-params
            image: goreleaser/goreleaser 
            script: |
              echo $(params.IMAGETAG)
      params:
        - name: IMAGETAG
          value: "$(tasks.get-result.results.pre):v$(tasks.get-result.results.post)-1.0"
```

and the console displays the follow:

```shell
$tkn pipelinerun logs openshift-handson-line-feed-results-run-7tsxl -f -n handson-example
[get-result : set-results] + echo 123
[get-result : set-results] + echo abc

[checkresults : check-params] + echo 123
[checkresults : check-params] + :vabc
[checkresults : check-params] 123
[checkresults : check-params] /tekton/scripts/script-0-q89sn: line 4: :vabc: not found
```

# Example Usage Codes

```yaml
    - name: push-container
      taskRef:
        name: kaniko
      workspaces:
        - name: source
          workspace: local-source
      runAfter: ["int-test", "gen-report"]
      params:
        - name: DOCKERFILE
          value: src/main/docker/Dockerfile.jvm
        - name: IMAGE
          value: "$(params.image-registry):v$(tasks.get-versions.results.version2)hhh" #this
````

and the console displays the follow:

```shell
[build-and-push] error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again: getting tag for destination: tag can only contain the runes `abcdefghijklmnopqrstuvwxyz0123456789_-.ABCDEFGHIJKLMNOPQRSTUVWXYZ`: v2
[build-and-push] hhh
```

# Additional Info

- OpenShift version:

```
[...]

PR #2942, opened 2020-07-14
Document that Task Results dont get trimmed

sbwsg:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/2936

The contents of task results are not processed by Tekton in any way - leading
and trailing whitespace characters are emitted verbatim from files written to
/tekton/results.

This commit adds documentation to make clear that Tekton does not process result
contents in any way.

In addition, I've added a short intro to the beginning of the results section
to give some focus on what results are useful for.

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -48,7 +48,7 @@ A `Task` declaration includes the following elements:
 - [Resources](#specifying-resources)
 - [Steps](#defining-steps)
 - [Workspaces](#specifying-workspaces)
-- [Results](#storing-execution-results)
+- [Results](#emitting-results)
 
 ## Configuring a `Task`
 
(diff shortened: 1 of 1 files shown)


docs/tasks.md
```diff
@@ -387,9 +387,22 @@ and the [`Workspaces` in a `TaskRun`](../examples/v1beta1/taskruns/workspace.yam
 
 ### Emitting results
```
pritidesai, line 388: /lgtm

one more section on `Consuming Results` might be helpful as well, emphasizing that the consumer task ends up having dependency on the emitting results task and executed after the dependencies are executed successfully.

(if not documented anywhere 🙏 )

sbwsg: There are a couple of places where it looks like we describe how results appear:

- The [TaskRun](https://github.com/tektoncd/pipeline/blob/master/docs/taskruns.md#monitoring-results) doc describes a bit about their appearance in the Status section of the resource.
- The [Pipeline](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#configuring-execution-results-at-the-task-level) doc describes a bit about referencing results between tasks and mentions the task ordering.

I think that Pipeline doc could use a bit of work to make it a tad clearer. I also notice that the link from there to this doc's results section is broken.  I'll make a PR for that.  Good call, thanks @pritidesai !

PR #2943, opened 2020-07-15
Add initial documentation for Runs

ImJasonH:
This is still incomplete as more fields and features will be added to
the Run type. This doc should also probably explain the responsibilities
of a custom task controller.

# Submitter Checklist

review them:

- [...yes] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [yep] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [yep] Release notes block has been filled in or deleted (only if no user facing changes)

## Reviewer Notes

# Release Notes

```release-note
Add docs for the v1alpha1 Run type
```

/assign @bobcatfish

### DIFF
--- a/docs/runs.md
+++ b/docs/runs.md
@@ -1,0 +1,231 @@
+<!--
+---
+linkTitle: "Runs"
+weight: 2
+---
+-->
+
+# `Run`s
+
+- [Overview](#overview)
+- [Configuring a `Run`](#configuring-a-run)
+  - [Specifying the target Custom Task](#specifying-the-target-custom-task)
+  - [Specifying `Parameters`](#specifying-parameters)
+- [Monitoring execution status](#monitoring-execution-status)
+  - [Monitoring `Results`](#monitoring-results)
+- [Code examples](#code-examples)
+  - [Example `Run` with a referenced custom task](#example-run-with-a-referenced-custom-task)
+  - [Example `Run` with an unnamed custom task](#example-run-with-an-unnamed-custom-task)
+  - [Example of specifying parameters](#example-of-specifying-parameters)
+
+# Overview
+
+A `Run` allows you to instantiate and execute a [Custom
+Task](https://github.com/tektoncd/community/blob/master/teps/0002-custom-tasks.md),
+which can be implemented by a custom task controller running on-cluster. Custom
+Tasks can implement behavior that doesn't correspond directly to running a
+workload in a `Pod` on the cluster. For Pod-based on-cluster workloads, you
+should use a [`TaskRun`](taskruns.md).
+
+In order for a `Run` to actually execute, there must be a custom task
+controller running on the cluster that is responsible for watching and updating
+`Run`s which reference their type. If no such controller is running, `Run`s
+will have no `.status` value and no further action will be taken.
+
+`Run`s are an **_experimental alpha feature_** and should be expected to change
+in breaking ways or even be removed.
+
+`Run`s are not currently integrated with Pipelines, and require a running
+third-party controller to actually perform any work. Without a third-party
+controller, `Run`s will just exist without a status indefinitely.
+
+## Configuring a `Run`
+
+A `Run` definition supports the following fields:
+
+- Required:
+  - [`apiVersion`][kubernetes-overview] - Specifies the API version, for example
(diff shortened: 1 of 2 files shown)


docs/runs.md
```diff
@@ -0,0 +1,227 @@
...
+spec:
+  ref:
+    apiVersion: example.dev/v1alpha1
+    kind: Example
+```
+
+This initiates the execution of a `Run` of a custom task of type `Example`, in
+the `example.dev` API group,  with the version `v1alpha1`.
+
+You can also specify the `name` and optional `namespace` (default is `default`)
```
dlorenc, line 73: This is more of a design question than a documentation one - but do we want to allow for cross-namespace access of runs like this? It differs from how TaskRuns can access Tasks - and makes RBAC on these harder.

ImJasonH: Yeah good question, and I'm not sure. I haven't looked at it yet. AFAIK traditional Tasks can only be referenced from the Pipeline's namespace, or as a `ClusterTask`, so that's probably a model worth following until someone asks for more.

docs/runs.md
```diff
@@ -0,0 +1,227 @@
...
+  - [Monitoring `Results`](#monitoring-results)
+- [Code examples](#code-examples)
+  - [Example `Run` with a referenced custom task](#example-run-with-a-referenced-custom-task)
+  - [Example `Run` with an unnamed custom task](#example-run-with-an-unnamed-custom-task)
+  - [Example of specifying parameters](#example-of-specifying-parameters)
+
+# Overview
+
+A `Run` allows you to instantiate and execute a [Custom
+Task](https://github.com/tektoncd/community/blob/master/teps/0002-custom-tasks.md),
```
bobcatfish, line 24: prolly makes sense to link to the TEP as we develop this but once the feature is ready to use it'd in important to have enough info in the docs on how this works that we don't direct ppl toward the proposal to understand how to use it

docs/runs.md
```diff
@@ -0,0 +1,227 @@
...
+
+To specify the custom task type you want to execute in your `Run`, use the
+`ref` field as shown below:
+
+```yaml
+spec:
+  ref:
+    apiVersion: example.dev/v1alpha1
+    kind: Example
+```
```
bobcatfish, line 68: i think we're going to end up wanting to repeat some of this in pipeline.md 🤔  or just move it over there and link here for an overview? i guess in that case we're "embedding" the spec in a pipeline? i think this just points to be it being important to have docs in pipeline.md that explain how the Run spec relates to using Runs in a pipeline

docs/runs.md
dlorenc: +1 to following that same model.

docs/runs.md
```diff
@@ -0,0 +1,227 @@
...
+Tasks can implement behavior that doesn't correspond directly to running a
+workload in a `Pod` on the cluster. For Pod-based on-cluster workloads, you
+should use a [`TaskRun`](taskruns.md).
+
+In order for a `Run` to actually execute, there must be a custom task
+controller running on the cluster that is responsible for watching and updating
+`Run`s which reference their type. If no such controller is running, `Run`s
+will have no `.status` value and no further action will be taken.
+
+`Run`s are an **_experimental alpha feature_** and should be expected to change
```
vdemeester, line 35: alpha feature should be enough shouldn't it ? :stuck_out_tongue_closed_eyes:

docs/runs.md
ImJasonH: Makes sense. As we add more functionality I'll make sure the docs stay up to date.

docs/runs.md
ImJasonH: Yeah that makes sense. When Pipelines are able to create Runs I'll update the docs accordingly.

docs/runs.md
ImJasonH: Yeah maybe, but I wanted to be _extra extra clear_ about the alpha-ness of it.

bobcatfish:
hey all! we just did a release with this functionality in it and no docs 😬 

/lgtm
/approve

Issue #2951, opened 2020-07-16
"ssh-keyscan": executable file not found in $PATH after upgrading from v0.13.2 to v0.14.1


Fabian-K:
Hi,

we have a pipeline running that is using ssh-based authentication to clone from a corporate git repository as described in https://github.com/tektoncd/pipeline/blob/master/docs/auth.md#ssh-authentication-git.

After upgrading from v0.13.2 to v0.14.1, the pipeline starts failing with the following log output:

```
invalid value "corp-git-secret=github.corp" for flag -ssh-git: ssh-keyscan error: exec: "ssh-keyscan": executable file not found in $PATH
Usage of /tekton/tools/entrypoint:
  -basic-docker value
    	List of secret=url pairs.
  -basic-git value
    	List of secret=url pairs.
  -docker-cfg string
    	Docker .dockercfg secret file.
  -docker-config string
    	Docker config.json secret file.
  -entrypoint string
    	Original specified entrypoint to execute
  -post_file string
    	If specified, file to write upon completion
  -results string
    	If specified, list of file names that might contain task results
  -ssh-git value
    	List of secret=url pairs.
  -termination_path string
    	If specified, file to write upon termination (default "/tekton/termination")
  -wait_file string
    	Comma-separated list of paths to wait for
  -wait_file_content
    	If specified, expect wait_file to have content
```

In the release notes of 0.14.1 I found the following paragraph:
The creds-init step is gone and its behavior is now handle by the entrypoint. This shouldn't have any impact except running as the same user as the step.

Do we now have to include ssh-keyscan (and others?) in the image that we are using for the task?

Thanks,
Fabian

vdemeester:
oh that ain't good :sweat: 
/cc @sbwsg 
(this might mean we need a 0.14.2 :upside_down_face: )

sbwsg:
Ah, yeah, that's fair.  Probably need to rollback the creds-init change.  I'll get a PR open.

sbwsg:
@Fabian-K can you confirm whether you include the `known_hosts` section in your Secrets (mentioned in [the auth doc](https://github.com/tektoncd/pipeline/blob/master/docs/auth.md#ssh-authentication-git)).

If not then you can add this known_hosts field to immediately sidestep the problem.  `ssh-keyscan` should only run if you don't include a known_hosts entry in the Secret.

sbwsg:
[...]

PR #2952, opened 2020-07-16
Rollback creds-init removal in 0.14 branch

sbwsg:
# Changes

In https://github.com/tektoncd/pipeline/pull/2671 I removed the creds-init initContainer from Task Pods so that credentials could be used by containers running with non-root users.  The intention was for this change to be free of any side-effects to end-users.  Unfortunately a [backwards incompatible issue](https://github.com/tektoncd/pipeline/issues/2951) has cropped up with this change:

When a user does not specify the `known_hosts` field in a creds-init Secret, the credential code will perform an `ssh-keyscan` of the remote server to get its public key.  The problem is that previously we could guarantee `ssh-keyscan` was available since the code ran in our own creds-init container with our own docker image. Since we've now moved that code into Steps' entrypoint a Step's container is required to provide `ssh-keyscan`.  This is a change in container contract and therefore backwards-incompatible.

In this PR I've reverted the creds-init change for the 0.14 branch rather than attempt to fix the `ssh-keyscan` issue and possibly introduce more problems.

Before 0.15 I'd like to get a better, backwards-compatible, fix organized.  So I plan to leave the creds-init change in place in the `master` branch for the time being.

# Submitter Checklist

review them:

# Release Notes

```release-note
The removal of the creds-init initContainer that occurred in 0.14.0 has been rolled back. The change introduced an unintended backwards-incompatibility that needs to be resolved.
```


afrittoli:
So, if I got it right, the plan is rollback on the branch only, and fix on master for v0.15 instead, right?

vdemeester:
That's my guess too, @sbwsg are we right ?

sbwsg:
> So, if I got it right, the plan is rollback on the branch only, and fix on master for v0.15 instead, right?

Yup, exactly right.  I'm working on a fix now that should hopefully be ready in time for 0.15.

vdemeester:
> > So, if I got it right, the plan is rollback on the branch only, and fix on master for v0.15 instead, right?
> 
> Yup, exactly right. I'm working on a fix now that should hopefully be ready in time for 0.15.

Let's add the refered issue in the milestone then :angel: 

/lgtm

sbwsg:
> Let's add the refered issue in the milestone then 👼

Good thinking, thanks for doing that!

PR #2953, opened 2020-07-16
Remove ssh-keyscan from git credential initialization

sbwsg:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/2951

Before this change an SSH Git Secret without a known_hosts would cause Tekton to perform an ssh-keyscan of the remote host to get its public key. This would happen in the entrypoint of the Step which meant that every Step container would have to have a copy of ssh-keyscan.

After this change the reponsibility for fetching missing host keys now falls on git-init, meaning that the Git PipelineResource and git-clone catalog task will fetch the missing keys at runtime. This is preferable because we can guarantee that the git-init image has ssh in it. We no longer rely on `ssh-keyscan` at all and instead use `ssh`'s `StrictHostKeyChecking=accept-new` option.

Need tests/examples to:

- [x] Confirm that omitting `known_hosts` does not result in `ssh-keyscan` erroring out a TaskRun.
- [x] Confirm that omitting `known_hosts` does result in `git-init` running with `StrictHostKeyChecking=accept-new`
- [x] Confirm that including `known_hosts` doesnt result in `StrictHostKeyChecking=accept-new`.

# Submitter Checklist

review them:

# Release Notes

```release-note
Tekton's creds-init initContainer has been removed and will no longer show up in TaskRun Pods. Credentials from annotated Secrets will continue to be mounted and placed in Step containers' home directory.

Further, previously a missing known_hosts entry from a Git SSH Secret would result in ssh-keyscan being executed to fetch a server's public key. That behaviour has been replaced with passing "-o StrictHostKeyChecking=accept-new" to ssh when using a Git PipelineResource or the git-clone Catalog Task. The ultimate result should be the same - the public key of the server is accepted without question - but the location where this work is performed has changed.
```


sbwsg:
/kind bug


/hold

needs tests

sbwsg:
/hold cancel

In order to test these changes I've updated an existing example and added a new one.  It's actually kinda hard to write unit tests for our credential code that calls out to `git`. As an example, our `git.Fetch` func is relatively long and calls out to the filesystem in lots of different ways that would be quite brittle to mock I think.  This is probably worth some deeper investigation so I've created https://github.com/tektoncd/pipeline/issues/2974 to get some work done on that.

sbwsg:
> Seems good to me. Should we have a "config" to disallow the "accept-new" (if the operator of tekton-pipelines wants to force users to provide ssh known_hosts) ?

Good thinking - I've created https://github.com/tektoncd/pipeline/issues/2981 to capture the feature request.

dlorenc:
Should we include something about the change in behavior of ssh-keyscan in the release notes?

sbwsg:
Updated the release note to call it out.

sbwsg:
ping; would be great to get this in as part of the 0.15 milestone.  if that's not feasible I'll rollback the creds-init change and try again in 0.16.

PR #2967, opened 2020-07-19
PodSecurityPolicy Config fixes to allow runnning in restricted envs.

### LINKED ISSUE #2966: PodSecurityPolicy Config problems running Tekton in a restricted environment
Until there is a better alternative we're using PodSecurityPolicy to lock down our environment and we did a quick audit of Tekton's default PSP config. We found a couple of fairly simple problems that should be fixed.

1. Tekton's PSP "tekton-pipelines" allows containers to `RunAsAny` user. It is applied to the "tekton-pipelines-controller" SA using a `ClusterRole`. A `ClusterRole` should not be used in this case as the PSP is only required in the "tekon-pipelines" namespace to allow the controller pod to be "created". Instead the PSP rule should be applied with a namespace-scoped `Role`. 

(As it stands, users able to create PipelineRuns and TaskRuns are indirectly permitted to "use" the "tekton-pipelines" PSP "in other namespaces" and in this case run as root since the controller is the "requesting user". See https://kubernetes.io/docs/concepts/policy/pod-security-policy/#authorizing-policies)

2. The  "tekton-pipelines-webhook" SA has no PSP applied to it so fails to start in a restricted environment. Similar to the controller the webhook should have a `RoleBinding` to permit use.

3. Both the controller and webhook pods run as root so need `runAsNonRoot: false` added to their deployment's template to help PSP find the right rule and avoid matching an alphabetically earlier PSP (like '001-restricted-psp') that will not permit them to run.

skaegi:
Fixes #2966

# Changes
1. Moves PSP rule from tekton-pipelines-controller-cluster-access ClusterRole to tekton-pipelines-controller Role. This reduces the scope of where the PSP can be used to prevent privilege escalation
2. Adds PSP rule to tekton-pipelines-webhook to allow the webhook to run when only restricted PSPs are available
3. Adds runAsNonRoot: false to the securityContext of both the controller and webhook deployments to help PSP choose a rule that will let the pods run-as-root

# Submitter Checklist

review them:

- [ N/A] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ N/A ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

## Reviewer Notes

# Release Notes
```release-note
The tekton-pipelines-controller and tekton-pipelines-controller are now configured to run as a non-root user. To match these reduced requirements, the tekton-pipelines PodSecurityPolicy updates its runAsUser rule to use MustRunAsNonRoot and is further tightened-up to only allow "use" in the tekton-pipelines namespace.
```

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -105,6 +105,9 @@ spec:
           value: config-leader-election
         - name: METRICS_DOMAIN
           value: tekton.dev/pipeline
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsUser: 1001
       volumes:
         - name: config-logging
           configMap:
--- a/config/101-podsecuritypolicy.yaml
+++ b/config/101-podsecuritypolicy.yaml
@@ -30,7 +30,7 @@ spec:
   hostIPC: false
   hostPID: false
   runAsUser:
-    rule: 'RunAsAny'
+    rule: 'MustRunAsNonRoot'
   seLinux:
     rule: 'RunAsAny'
   supplementalGroups:
--- a/config/200-clusterrole.yaml
+++ b/config/200-clusterrole.yaml
@@ -39,10 +39,6 @@ rules:
   - apiGroups: ["tekton.dev"]
     resources: ["tasks/status", "clustertasks/status", "taskruns/status", "pipelines/status", "pipelineruns/status", "pipelineresources/status"]
     verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
-  - apiGroups: ["policy"]
-    resources: ["podsecuritypolicies"]
-    resourceNames: ["tekton-pipelines"]
-    verbs: ["use"]
 ---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
--- a/config/200-role.yaml
+++ b/config/200-role.yaml
@@ -30,6 +30,10 @@ rules:
     resources: ["configmaps"]
     verbs: ["get"]
     resourceNames: ["config-logging", "config-observability", "config-artifact-bucket", "config-artifact-pvc", "feature-flags", "config-leader-election"]
+  - apiGroups: ["policy"]
+    resources: ["podsecuritypolicies"]
+    resourceNames: ["tekton-pipelines"]
+    verbs: ["use"]
 ---
 kind: Role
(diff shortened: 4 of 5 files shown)


vdemeester:
@skaegi I think this needs a `release-note` entry other than `NONE` though :stuck_out_tongue: 
/hold

config/controller.yaml
```diff
@@ -105,6 +105,8 @@ spec:
           value: config-leader-election
         - name: METRICS_DOMAIN
           value: tekton.dev/pipeline
+        securityContext:
+          runAsNonRoot: false
```
ImJasonH, line 109: Can we just run these as nonroot? There's nothing they need to do as root afaik.

dlorenc:
The build error should be fixed now

config/controller.yaml
skaegi: Yep -- totally agree. Trying out `runAsUser: 1001`

config/controller.yaml
```diff
@@ -105,6 +105,8 @@ spec:
           value: config-leader-election
         - name: METRICS_DOMAIN
           value: tekton.dev/pipeline
+        securityContext:
+          runAsUser: 1001
```
vdemeester, line 109: @skaegi shouldn't `allowPrivilegedEscalation: false` also be there ? :stuck_out_tongue_closed_eyes:

skaegi: In this case it probably does not matter however it's a step in the right direction and for consistency yes I think so. ;)

skaegi: done

skaegi:
updated release notes and squashed for enhanced beauty

gabemontero:
per the ask in AP WG today @skaegi I took another look  ... more questions came up than anything else

a) first the easy one ... so far all the changes so far look fine to me, and the needs for them that you cited seem valid to me
b) now, are we interested in how more restricted the policy could be made and still have the various tekton scenarios work?  
c) or have you already gone through that exercise @skaegi and this is where you landed?

If the answer to b) is "yes" and c) is "no", I might have some additional suggestions, but at first blush I'm hesitant to go down that path.  But if that was your intent when you asked me to look, let me know and we can discuss.

dlorenc:
/approve

This looks good to me! Thanks!

PR #2972, opened 2020-07-20
Remove the "baseBuildOverrides" flag from .ko.yaml

dlorenc:
# Changes

This section never actually worked in ko. See https://github.com/google/ko/issues/167

It doesn't appear that we're setting the ldflags anywhere else, so maybe we don't need them?

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
NONE
```

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -9,8 +9,3 @@ baseImageOverrides:
 
   # Our entrypoint image does not need root, it simply needs to be able to 'cp' the binary into a shared location.
   github.com/tektoncd/pipeline/cmd/entrypoint: gcr.io/distroless/base:debug-nonroot
-baseBuildOverrides:
-  github.com/tektoncd/pipeline/cmd/controller:
-    flags:
-    - name: ldflags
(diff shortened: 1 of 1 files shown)


bobcatfish:
This is also in the tasks we use to release: https://github.com/tektoncd/pipeline/blob/9b809ddda3d0600bb7bdd069ad0af2f179087c11/tekton/publish.yaml#L67-L83

bobcatfish:
I'm a bit confused, it seems like this was added intentionally in https://github.com/tektoncd/pipeline/pull/1650 - does that mean #1650 didn't work?

vdemeester:
@bobcatfish yes it never worked :stuck_out_tongue_closed_eyes: 
not sure a big deal because now the version is provided through the labels/annotations.. but yeah… :face_with_head_bandage:

bobcatfish:
Ah kk - should we walk back the rest of #1650 as well then @vdemeester ?

vdemeester:
I need to check something on cli, but overall, I think yes — except if we want to be able to run `docker run -ti --rm gcr.io/tektoncd-releases/…/controller --version` to get the binary version, but I don't think we care ?

PR #2982, opened 2020-07-21
Change podconvert.MakePod func into a configuration struct with methods

sbwsg:
# Changes

While working on a feature to optionally disable Tekton's built-in credentials
initialization process, it became apparent that the podconvert.MakePod func
would need an extra boolean argument in its signature. The signature is already
quite long which makes reading it difficult.

This commit separates the signature into configuration options and the task
spec / run arguments. In a follow-up commit I plan to add the boolean flag described
above as another configuration option in the PodBuilder struct.

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -1760,7 +1760,13 @@ func makePod(taskRun *v1beta1.TaskRun, task *v1beta1.Task) (*corev1.Pod, error) 
 		return nil, err
 	}
 
-	return podconvert.MakePod(context.Background(), images, taskRun, task.Spec, kubeclient, entrypointCache, true)
+	builder := podconvert.Builder{
+		Images:          images,
+		KubeClient:      kubeclient,
+		EntrypointCache: entrypointCache,
+		OverrideHomeEnv: true,
+	}
+	return builder.Build(context.Background(), taskRun, task.Spec)
 }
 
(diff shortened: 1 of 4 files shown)


pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -1760,7 +1760,13 @@ func makePod(taskRun *v1beta1.TaskRun, task *v1beta1.Task) (*corev1.Pod, error)
 		return nil, err
 	}
 
-	return podconvert.MakePod(context.Background(), images, taskRun, task.Spec, kubeclient, entrypointCache, true)
+	builder := podconvert.PodBuilder{
```
vdemeester, line 1763: nit: it bothers me it's not named the same as above :joy_cat: *but that's definitely a nit* :joy:

sbwsg: hm, i'm open to changing the naming, no big deal, but i'm not 100% clear what name you want - right now it's `podconvert.Builder` after golangci-lint corrected me.  What's your preferred name?

vdemeester: I was speaker about the variable name `builder` (and above `podBuilder`) :blush:

dlorenc:
I went back and forth in my head twice on whether or not the kubernetes client should be on this struct vs. left as a function paramter but think I settled on the way you did it here. Nice!

/lgtm

PR #2983, opened 2020-07-21
Rewrite the "Pod Templates" documentation for clarity and flow.

sergetron:
# Changes

Rewrites the "Pod Templates" documentation for clarity and flow.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

 ```release-note
NONE
 ```

### DIFF
--- a/docs/podtemplates.md
+++ b/docs/podtemplates.md
@@ -6,65 +6,91 @@ weight: 12
 -->
 # Pod templates
 
-A pod template specifies a subset of
-[`PodSpec`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#pod-v1-core)
-configuration that will be used as the basis for the `Task` pod.
+A Pod template defines a portion of a [`PodSpec`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#pod-v1-core)
+configuration that Tekton can use as "boilerplate" for a Pod that runs your `Tasks` and `Pipelines`.
 
-This allows for customizing some `Pod` specific field per `Task` execution, `TaskRun` or for all `Task` executions part of a `PipelineRun`.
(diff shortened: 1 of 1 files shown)


sergetron:
Needs approval - pretty please?

sbwsg:
Would you be open to rewriting the table HTML as markdown?  https://github.github.com/gfm/#tables-extension-

We use the markdown syntax, for example, in https://github.com/tektoncd/pipeline/blob/master/docs/variables.md

sergetron:
If there's a strong feeling against HTML tables, I can convert it to
markdown no problem. From my experience, markdown tables tend to get really
messy if you need more than a line of text or nested/bulleted content,
which is why I normally go for simple HTML that's understood by most
renderers.

sbwsg:
If you think the HTML tables are best for this then 👍 . Is it possible to use markdown for, e.g. the \<a\> links if the content is in an HTML table?  If we're sticking with raw HTML then I'd also suggest making the header \<td\>s into \<th\>s instead, which should allow you to drop the \<b\> tags in those cells since \<th\>s are typically bolded by default in stylesheets and/or the user agent.  Additionally suggest wrapping the top row in a \<thead\> rather than \<tbody\>.

sergetron:
From my experience working on past projects, Markdown tables are only good
for very simple 2-3 column layouts with 1-2 lines of text per cell. Beyond
that, especially if you want to do things like bulleted lists, hyperlinks,
code formatting, and so on, they fall apart.

And you're totally right, <td> and <tbody> makes a lot more sense than what
I did, let me update those.

sergetron:
*<th> instead of <td> that is!

sergetron:
All fixed.

sergetron:
Needs re-LGTM please!

PR #3010, opened 2020-07-24
Fixes  #2971, adding a section to the tutorial for using minikube when running locally

### LINKED ISSUE #2971: Tutorial: enhance "running this tutorial locally"
As of today, in the [tutorial](https://github.com/tektoncd/pipeline/blob/master/docs/tutorial.md#running-this-tutorial-locally), we assume the user is either running Windows or Mac OS(X) as we are only documenting the use of "Docker for Desktop" (which is only available on those).

We should enhance that part by having support for different tools and more Operating Systems (like GNU/Linux). We should at least document how to run the tutorial using a tool that is available on all platforms, aka [`minikube`](https://minikube.sigs.k8s.io/docs/) (and maybe [`kind`](https://kind.sigs.k8s.io/docs/) too).

/kind documentation

psschwei:
# Changes

This PR fixes #2971, adding a section to the tutorial for using minikube when running locally.

# Submitter Checklist

review them:

## Reviewer Notes

/kind documentation
/release-note-none

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -555,11 +555,14 @@ You can also see the statuses of the individual `TaskRuns`.
 
 ## Running this tutorial locally
 
-This section provides guidelines for completing this tutorial on your local workstation.
+This section provides guidelines for completing this tutorial on your local workstation on:
 
-### Prerequisites
+* [Docker for Desktop](#prerequisites-docker-for-desktop)
+* [Minikube](#prerequisites-minikube)
 
-Complete these prerequisites to run this tutorial locally:
+### Prerequisites: Docker for Desktop
+
+Complete these prerequisites to run this tutorial locally using Docker for Desktop:
 
 - Install the [required tools](https://github.com/tektoncd/pipeline/blob/master/DEVELOPMENT.md#requirements).
 - Install [Docker for Desktop](https://www.docker.com/products/docker-desktop) and configure it to use six CPUs,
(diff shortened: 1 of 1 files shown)


psschwei:
To whomever reviews this -- I used my best judgement for updating the format of the tutorial (by adding sections for Docker for Desktop and Minikube). But can revise if others have a more preferred layout.

docs/tutorial.md
```diff
@@ -595,6 +598,33 @@ You must reconfigure any `image` resource definitions in your `PipelineResources
...
+minikube start --memory 10240 --cpus 6 
+```
+- Point your shell to minikube's docker-daemon by running `eval $(minikube -p minikube docker-env)`
+- Set up a registry on minkube by running `minikube addons enable registry` and `minikube addons enable registry-aliases`
+
+### Reconfigure `image` resources
+
+You must reconfigure any `image` resource definitions in your `PipelineResources` as follows:
+
+- Set the URL to `example.com/<image_name>`
```
pritidesai, line 617: @psschwei I am not able to understand instructions in this section, what is this URL meant for? we enabled registry in the previous section. Also, since we are using `-L` flag, no need to setup `KO` related env (i.e. unlike `KO_DOCKER_REPO` for docker for desktop). What would this `example.com` be in a `minikube` installation?

psschwei: `example.com` is one of the aliases created automatically for the registry via the `registry-aliases` addon (others are `example.org`, `test.com`, and `test.org`): 

```
docker@minikube:~$ cat /etc/hosts
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::0	ip6-localnet
ff00::0	ip6-mcastprefix
ff02::1	ip6-allnodes
ff02::2	ip6-allrouters
172.17.0.3	minikube
172.17.0.1	host.minikube.internal
172.17.0.3	control-plane.minikube.internal
10.104.170.3	example.org
10.104.170.3	example.com
10.104.170.3	test.com
10.104.170.3	test.org
```

If we want to be able to push/pull to the minikube Docker registry, we'll need to use one of its aliases. If we don't, it will try to resolve to Docker Hub -- this is the error I got when running the `build-docker-image-from-git-source-task-run` TaskRun using just `leeroy-web` as the URL for the `skaffold-image-leeroy-web` PipelineResource:

```
error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again: checking push permission for "leeroy-web": POST https://index.docker.io/v2/library/leeroy-web/blobs/uploads/: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:library/leeroy-web Type:repository] map[Action:push Class: Name:library/leeroy-web Type:repository]]
```

@pritidesai I can add some clarifying remarks to the `Reconfigure `image` resources` section to explain why we're using `example.com` as the registry alias here. While we could use an alternative hostname (for example, `host.docker.local` like in the Docker for Desktop example), it would require editing `/etc/hosts` and the `CoreDNS` configmap, so I assumed the simpler option was better. :)

pritidesai: Thanks @psschwei for the explanation, you can tell how much I know `minikube` configuration 🤦‍♀️

This looks good, no need to add alternative hostname. It will be great to clarify why `example.com` and adding may be this command to get the list of available hostnames ` minikube ssh -- cat /etc/hosts`

psschwei: No worries -- I only learned this in the course of doing this PR (and from a very helpful suggestion from @kameshsampath ). Will make the update shortly.

psschwei: Update made

pritidesai:
thanks @psschwei 

/approve

docs/tutorial.md
```diff
@@ -555,11 +555,14 @@ You can also see the statuses of the individual `TaskRuns`.
 
 ## Running this tutorial locally
 
-This section provides guidelines for completing this tutorial on your local workstation.
+This section provides guidelines for completing this tutorial on your local workstation on:
 
-### Prerequisites
+* [Docker for Desktop](#prerequisites-docker-for-desktop)
```
sbwsg, line 560: minor nit, suggest matching bullet style with rest of document (using `-` instead of `*`)

docs/tutorial.md
```diff
@@ -595,6 +598,33 @@ You must reconfigure any `image` resource definitions in your `PipelineResources
...
 - To learn more about obtaining logs, see [Logs](logs.md).
 
+### Prerequisites: Minikube
+
+Complete these prerequisites to run this tutorial locally using Minikube:
+
+- Install the [required tools](https://github.com/tektoncd/pipeline/blob/master/DEVELOPMENT.md#requirements).
+- Install [minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/) and start a sesion as follows:
+```bash
+minikube start --memory 10240 --cpus 6
```
sbwsg, line 608: fwiw I develop Tekton in a minikube with 6GB RAM and 2 cpus and things run really nicely. (this is using the virtualbox driver on mac)

psschwei: Happy to lower this -- I just mirrored the specs that were listed from the Docker for Desktop section : "Install Docker for Desktop and configure it to use six CPUs, 10 GB of RAM and 2GB of swap space."

docs/tutorial.md
psschwei: @sbwsg was going to fix this, but looks like the PR already merged... I can open another one that'll clean this up (and fix the resource usage too), or wait and include it alongside something more meaningful if you think this is too trivial for a standalone PR.

sbwsg: Small PRs are good PRs :). Fixing in a follow-up is totally a good way to go.  Thanks @psschwei !

Issue #2905, opened 2020-07-06
The timeoutHandler is only instructed to wait when it creates pods


mattmoor:
/kind bug

Consider this logic in `taskrun.go`:

```go
		pod, err = c.createPod(ctx, tr, rtr)
		if err != nil {
			newErr := c.handlePodCreationError(ctx, tr, err)
			logger.Error("Failed to create task run pod for task %q: %v", tr.Name, newErr)
			return newErr
		}
		go c.timeoutHandler.WaitTaskRun(tr, tr.Status.StartTime)
```

If after creating the Pod the controller process is restarted (e.g. to rollout a new controller binary) then the timeout handler won't properly timeout task runs straddling this event because the controller's resync won't trigger the go routine.

The pattern we have used successfully in Knative is to have the library support idempotent invocation, so that `WaitTaskRun` can be called on *every* resync (as appropriate) and simply have it elide redundant invocations.  This is something that would shake out pretty quickly with chaos testing (https://github.com/knative/pkg/pull/1333 :eyes: ).

bobcatfish:
I was very excited to fix this bug but I'm actually no longer sure it's a bug:

When the controllers start, they call CheckTimeouts ([pipelinerun](https://github.com/tektoncd/pipeline/blob/abb677dbb23cdc305b0c73409679af980fa322c4/pkg/reconciler/pipelinerun/controller.go#L88), [taskrun](https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/taskrun/controller.go#L92) (they probably dont BOTH need to call it but oh well XD) which [looks for taskruns and pipelineruns](https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/timeout_handler.go#L187) and starts the waiting go routine or times them out as needed

Looks like this was originally addressed in https://github.com/tektoncd/pipeline/issues/456

That being said, having WaitTaskRun be idempotent makes a lot of sense and it's definitely not obvious that this is happening, so I think there's some improvement to make here, but I don't think it's a bug.

bobcatfish:
Thinking about this a bit more, even making the call idempotent won't handle the issue being described:

> If after creating the Pod the controller process is restarted (e.g. to rollout a new controller binary) then the timeout handler won't properly timeout task runs straddling this event because the controller's resync won't trigger the go routine.

[...]

PR #3011, opened 2020-07-24
Some timeout refactoring

bobcatfish:
# Changes

## Add more details about how the timeout handling works 🕒

While investigating #2905, I struggled to understand how the timeout
handling works, especially with TimeoutSet having very little comments,
so I've added some. I didn't add anything for backoffs yet because I'm
hoping we can separate that into a separate structure since it has a
very specific purpose that doesn't generalize to all timeouts.

Also changed the name "finished" to consistently use "done" so the
reader doesn't have to wonder about the difference between "finished"
and "done" (there isn't one)

## Move timeout handler into its own package 📦

I'd like to move the "backoff" logic into its own file, separate from
the other timeout logic, so it's clear which parts apply to what (i.e.
the timeout handler is being used for 2 purposes: timing out Runs which
take too long, and backing off when pod creation is failing - this is
totally fine but it's hard to understand when reading the code)

As a first step, I've moved the timeout handler into a separate package,
so we can have a file and tests dedicated to the backoff logic separate
from the other handling.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

 ```release-note
 NONE
 ```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -38,12 +38,12 @@ import (
 	listers "github.com/tektoncd/pipeline/pkg/client/listers/pipeline/v1beta1"
 	resourcelisters "github.com/tektoncd/pipeline/pkg/client/resource/listers/resource/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/contexts"
-	"github.com/tektoncd/pipeline/pkg/reconciler"
 	"github.com/tektoncd/pipeline/pkg/reconciler/events"
 	"github.com/tektoncd/pipeline/pkg/reconciler/pipeline/dag"
 	"github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun/resources"
 	"github.com/tektoncd/pipeline/pkg/reconciler/taskrun"
 	"github.com/tektoncd/pipeline/pkg/reconciler/volumeclaim"
+	"github.com/tektoncd/pipeline/pkg/timeout"
 	"github.com/tektoncd/pipeline/pkg/workspace"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
@@ -114,7 +114,7 @@ type Reconciler struct {
 	resourceLister    resourcelisters.PipelineResourceLister
 	conditionLister   listersv1alpha1.ConditionLister
 	tracker           tracker.Interface
-	timeoutHandler    *reconciler.TimeoutSet
+	timeoutHandler    *timeout.Handler
 	metrics           *Recorder
 	pvcHandler        volumeclaim.PvcHandler
 }
@@ -124,7 +124,7 @@ var (
 	_ pipelinerunreconciler.Interface = (*Reconciler)(nil)
 )
 
-// Reconcile compares the actual state with the desired, and attempts to
+// ReconcileKind compares the actual state with the desired, and attempts to
 // converge the two. It then updates the Status block of the Pipeline Run
 // resource with the current status of the resource.
 func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun) pkgreconciler.Event {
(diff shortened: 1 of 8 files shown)


pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -124,7 +124,7 @@ var (
 	_ pipelinerunreconciler.Interface = (*Reconciler)(nil)
 )
 
-// Reconcile compares the actual state with the desired, and attempts to
+// ReconcileKind compares the actual state with the desired, and attempts to
```
dlorenc, line 127: Nice catch :)

pkg/timeout/handler.go
```diff
@@ -58,24 +58,37 @@ type Backoff struct {
...
+// Handler knows how to track channels that can be used to communicate with go routines
+// which timeout, and the functions to call when that happens.
+type Handler struct {
+	logger *zap.SugaredLogger
+
+	// taskRunCallbackFunc is the function to call when a TaskRun has timed out.
+	// This is usually set to the function that enqueues the taskRun for reconciling.
+	taskRunCallbackFunc func(interface{})
+	// pipelineRunCallbackFunc is the function to call when a TaskRun has timed out
+	// This is usually set to the function that enqueues the taskRun for reconciling.
```
pritidesai, line 70: NIT: 
s/when a TaskRun/when a PipelineRun/
s/enqueues the taskRun/enqueues the pipelineRun/

pkg/timeout/handler.go
```diff
@@ -274,7 +286,7 @@ func (t *TimeoutSet) waitRun(runObj StatusKey, timeout time.Duration, startTime
 // the lifetime of the TaskRun no resources are released after the timer
 // fires. It is the caller's responsibility to Release() the TaskRun when
 // work with it has completed.
-func (t *TimeoutSet) SetTaskRunTimer(tr *v1beta1.TaskRun, d time.Duration) {
+func (t *Handler) SetTaskRunTimer(tr *v1beta1.TaskRun, d time.Duration) {
```
pritidesai, line 289: Random thought, `SetTaskRunTimer` but no `SetPipelineRunTimer` 🤔

pkg/timeout/handler.go
```diff
@@ -186,7 +198,7 @@ func (t *TimeoutSet) checkPipelineRunTimeouts(namespace string, pipelineclientse
 
 // CheckTimeouts function iterates through a given namespace or all namespaces
 // (if empty string) and calls corresponding taskrun/pipelinerun timeout functions
-func (t *TimeoutSet) CheckTimeouts(namespace string, kubeclientset kubernetes.Interface, pipelineclientset clientset.Interface) {
+func (t *Handler) CheckTimeouts(namespace string, kubeclientset kubernetes.Interface, pipelineclientset clientset.Interface) {
```
pritidesai, line 201: its interesting that we are scrapping through all possible namespaces (if not specified) and checking all `TaskRuns` and `PipelineRuns` in those namespaces or at least in one specified namespace 😲

pritidesai:
@bobcatfish I learnt some timeout handler with this PR 😜  and excited to see more changes ...

one minor [NIT](https://github.com/tektoncd/pipeline/pull/3011#discussion_r460334610) which can be addressed with next set of changes

/lgtm

pkg/timeout/handler.go
pritidesai: And the same check `CheckTimeouts` is done in TaskRun controller and PipelineRun controller 🤔

pkg/timeout/handler.go
bobcatfish: whoops, thanks @pritidesai !

pkg/timeout/handler.go
bobcatfish: yeah!! the second means we're probably doing this twice as much as we need to 😅 

and it sounds like we probably don't need to be doing it at all!! https://github.com/tektoncd/pipeline/issues/2905#issuecomment-663717912

Issue #1131, opened 2019-07-26
Allow sidecars to terminate gracefully


pierretasci:
# Expected Behavior
A sidecar might need to complete any pending work before being torn down (eg, uploading metadata).

# Actual Behavior
Sidecars are replaced with a `nop` image as soon as a TaskRun completes preventing any cleanup work.

# Steps to Reproduce the Problem

1. Create a TaskRun
2. Using an AdmissionsController to inject a sidecar that streams output from the Task to another service.

We would expect all of the output to be uploaded as the task runs but whatever chunk isn't uploaded before the sidecar is torn down is presumably lost.

# Additional Info

There are ways around this but it seems to me that the very least, a sidecar should receive the `kill` signal so it can handle shutdown however it needs to. A timeout can always been inserted to ensure it doesn't hang forever.

sbwsg:
There was some discussion of this problem on the [original design proposal](https://github.com/tektoncd/pipeline/issues/727). I think the main issue raised was that you can't guarantee an arbitrary sidecar has a `kill` binary available, so exec'ing into a container and running kill on a pid won't always work.  I'll do some reading on how kubernetes kills containers in a pod and see what it would take to do same for Tekton.

@pierretasci Do you ideas for ways this could be implemented that would support arbitrary containers?

pierretasci:
Yeah, I read through the discussion on the original proposal and I think it makes sense. I do think that it wouldn't be particularly harmful to send a `SIGTERM` or `SIGKILL` into the container arbitrarily with the contract that if you don't handle shutdown in XX seconds then we will perform our `nop` container swap.

If the container doesn't have shutdown logic, no harm done it seems. At the worst, it would introduce an XX second delay to completing each task.

I am of course open to other ideas but I do think stateful sidecars (logging is another use case that comes to mind) will want a shutdown mechanism.

sbwsg:
> I do think that it wouldn't be particularly harmful to send a SIGTERM or SIGKILL into the container

I totally agree, I'm just not entirely sure yet how this would be implemented if the container doesn't include a `kill` binary.

sbwsg:
I've been skimming through the k8s codebase looking for the spot where the container termination flow is implemented and [it looks as though](https://github.com/kubernetes/kubernetes/blob/42d7feee28bafaf91f328f9b115ad3c4c2184982/pkg/kubelet/kuberuntime/kuberuntime_container.go#L533) the precise implementation of the grace period is delegated to the container runtime.

[...]

PR #3013, opened 2020-07-26
Fixes: #1131, Enable graceful sidecar kill

shahnewazrifat:
# Changes

Fixes: https://github.com/tektoncd/pipeline/issues/1131

- Adds a boolean flag `WaitForSidecar` to the Sidecar struct to enable users to avoid sidecar termination when all the steps are finished in the pipeline.
- By default, it still does the forceful termination of SideCar using the Nop image.
- The idea is, there can be cases where the sidecar needs to do some processing and upload work when the Taskrun is done/completed. A sidecar should not be interrupted and the termination of the Sidecar can be delegated to the user injecting it in those cases.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```release-note
    - Added a new configuration to the Sidecar struct. The boolean flag `WaitForSidecar` can be set to true if you don't want your sidecar to be terminated automatically by Nop Image when all steps in a task run is complete. 
```


pkg/apis/pipeline/v1beta1/task_types.go
```diff
@@ -126,7 +126,16 @@ type Step struct {
...
-type Sidecar = Step
+type Sidecar struct {
+	corev1.Container `json:",inline"`
+
+	// Script is the contents of an executable file to execute.
+	// If Script is not empty, the Step cannot have an Command or Args.
+	Script string `json:"script,omitempty"`
+	// ForceTerminationDisabled is a bool flag to toggle sidecar termination when all the steps in the task are completed.
+	// +optional
+	ForceTerminationDisabled bool `json:"forceTerminationDisabled,omitempty,default:false"`
```
ImJasonH, line 137: Does the `default:false` here do anything? AFAIK the `json:""` struct tag only really supports things like `omitempty`, `inline`, etc.

pkg/pod/entrypoint.go
```diff
@@ -182,27 +182,26 @@ func UpdateReady(kubeclient kubernetes.Interface, pod corev1.Pod) error {
...
-				for j, c := range newPod.Spec.Containers {
-					if c.Name == s.Name && c.Image != nopImage {
-						updated = true
-						newPod.Spec.Containers[j].Image = nopImage
+			for _, containerItem := range *containers {
+				if s.Name == containerItem.Name && s.State.Running != nil {
+					for j, c := range newPod.Spec.Containers {
+						if c.Name == s.Name && c.Image != nopImage {
+							updated = true
+							fmt.Println(fmt.Sprintf("Attempting to terminate Sidecar container %s since it was not marked for exemption from ForceTermination in pod %s.", s.Name, newPod.Name))
```
ImJasonH, line 198: This should use a logger instead, probably derived from a provided context. This formats it so that it can be more easily indexed/searched, instead of just printing directly to stdout.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -709,6 +712,27 @@ func storeTaskSpec(ctx context.Context, tr *v1beta1.TaskRun, ts *v1beta1.TaskSpe
 	return nil
 }
 
+func getStoppableContainers(pod *corev1.Pod, sidecars []v1beta1.Sidecar) []corev1.Container {
+	stoppables := []corev1.Container{}
+	// Sidecars that does not have ForceTermination disabled should be terminated.
+	fmt.Println("Gathering Tekton Sidecars...")
```
ImJasonH, line 718: ditto about using a logger here.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -121,7 +121,10 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, tr *v1beta1.TaskRun) pkg
 		c.timeoutHandler.Release(tr)
 		pod, err := c.KubeClientSet.CoreV1().Pods(tr.Namespace).Get(tr.Status.PodName, metav1.GetOptions{})
 		if err == nil {
-			err = podconvert.StopSidecars(c.Images.NopImage, c.KubeClientSet, *pod)
+			fmt.Println("Stopping Sidecars that were not marked for exemption.")
```
ImJasonH, line 124: ditto about using a logger

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"gotest.tools/assert"
```
ImJasonH, line 23: I'd prefer not to add a dependency on `assert`-like libraries, since they don't always make the code any more readable, and developers end up having to learn the particular behaviors/quirks of the assert library. Pair that with the fact that they're not uniformly used across the codebase, and I think they end up causing more headaches than they solve.

pkg/apis/pipeline/v1beta1/task_types.go
shahnewazrifat: Worth testing. I thought it does 😄 
I think in this case `omitempty` makes sense then

pkg/pod/entrypoint.go
shahnewazrifat: Totally. I'll check what logger is being used in general and use it

pkg/reconciler/taskrun/taskrun_test.go
shahnewazrifat: Makes sense. Shall check what's being currently used for similar in this codebase

bobcatfish:
Thanks for this @shahnewazrifat ! I think this feature makes a lot of sense - ages ago with got some feedback from the Prow folks that they'd like this kind of functionality as well ( @ cjwagner ) 

I could see making a TEP for this as well, since it does feel like a new feature and there might be some scenarios to explain (e.g. timeouts) but since that's the only thing I can think of I could see going ahead with this without a TEP as well, does anyone have a preference @afrittoli @vdemeester @sbwsg @ImJasonH @dlorenc @pritidesai ? (and @skaegi had mentioned trying to mirror the proposed sidecar functionality as well)

One request I have is to change the name of the flag slightly, to make it so that the name is worded as a positive (vs negative - i.e. I think disabled: false is a bit confusing). maybe instead of `ForceTerminationDisabled` it could be something like `TerminateAfterSteps` / `ForceTerminate` or the opposite `WaitForSidecar` etc.

shahnewazrifat:
> Thanks for this @shahnewazrifat ! I think this feature makes a lot of sense - ages ago with got some feedback from the Prow folks that they'd like this kind of functionality as well ( @ cjwagner )
> 
> I could see making a TEP for this as well, since it does feel like a new feature and there might be some scenarios to explain (e.g. timeouts) but since that's the only thing I can think of I could see going ahead with this without a TEP as well, does anyone have a preference @afrittoli @vdemeester @sbwsg @ImJasonH @dlorenc @pritidesai ? (and @skaegi had mentioned trying to mirror the proposed sidecar functionality as well)
> 
> One request I have is to change the name of the flag slightly, to make it so that the name is worded as a positive (vs negative - i.e. I think disabled: false is a bit confusing). maybe instead of `ForceTerminationDisabled` it could be something like `TerminateAfterSteps` / `ForceTerminate` or the opposite `WaitForSidecar` etc.

Pleasure @bobcatfish ! Totally love this community.

Yes, I was even thinking about going a bit further and make it the default behaviour for sidecars. I.e., auto termination will only happen when specified. Otherwise, it will act like a normal container should. Keeps running. 

I don't think many users are aware of this force termination if they didn't deep dive into the docs. 

Re: the flag name 😄  I got confused myself couple of time, so yeah. `forcerTerminate` sounds reasonable to me.

Thank you so much for the feedback!

go.mod
```diff
@@ -27,6 +27,7 @@ require (
 	gomodules.xyz/jsonpatch/v2 v2.1.0
 	google.golang.org/api v0.25.0
 	gopkg.in/yaml.v2 v2.3.0 // indirect
+	gotest.tools v2.2.0+incompatible
```
vdemeester, line 30: This doesn't seems right :thinking:

pkg/reconciler/taskrun/taskrun_test.go
vdemeester: also, it should have been `gotest.tools/v3/assert` :upside_down_face: 
The real value of `gotest.tools` is the other packages (fs, env, golden, icmd, …) — *says one of the other of the lib* :grin:.

go.mod
shahnewazrifat: That's right. I'm replacing it to use what's already being used in this repo

vdemeester:
> I could see making a TEP for this as well, since it does feel like a new feature and there might be some scenarios to explain (e.g. timeouts) but since that's the only thing I can think of I could see going ahead with this without a TEP as well, does anyone have a preference @afrittoli @vdemeester @sbwsg @ImJasonH @dlorenc @pritidesai ? (and @skaegi had mentioned trying to mirror the proposed sidecar functionality as well)

I think I would definitely see value in having a TEP for this indeed :upside_down_face:

pkg/apis/pipeline/v1beta1/task_types.go
shahnewazrifat: It really doesn't work!
And I learned the hard way that there's no native good way to set a boolean flag in go struct to "true" by default 🤦‍♂️ 

Thanks for the catch!

shahnewazrifat:
> Thanks for this @shahnewazrifat ! I think this feature makes a lot of sense - ages ago with got some feedback from the Prow folks that they'd like this kind of functionality as well ( @ cjwagner )
> 
> I could see making a TEP for this as well, since it does feel like a new feature and there might be some scenarios to explain (e.g. timeouts) but since that's the only thing I can think of I could see going ahead with this without a TEP as well, does anyone have a preference @afrittoli @vdemeester @sbwsg @ImJasonH @dlorenc @pritidesai ? (and @skaegi had mentioned trying to mirror the proposed sidecar functionality as well)
> 
> One request I have is to change the name of the flag slightly, to make it so that the name is worded as a positive (vs negative - i.e. I think disabled: false is a bit confusing). maybe instead of `ForceTerminationDisabled` it could be something like `TerminateAfterSteps` / `ForceTerminate` or the opposite `WaitForSidecar` etc.

I chose one similar to `WaitForSidecar` 😄 
Using `waitForTermination` since we're already inside `Sidecar` struct. `Sidecar.WaitForTermination` felt good 
Let me know what you feel!

Thanks for the great idea. Almost missed it! The double negative was confusing to me even when writing tests 🤕

pkg/pod/entrypoint.go
shahnewazrifat: I'm using the logger now, but I'm not comfortable initialising empty context logger just for this :( Please suggest if there's a better way.

pkg/reconciler/taskrun/taskrun.go
shahnewazrifat: Done!

pkg/reconciler/taskrun/taskrun.go
shahnewazrifat: Done!

go.mod
shahnewazrifat: Removed 😄

pkg/reconciler/taskrun/taskrun_test.go
shahnewazrifat: Has been removed 😄

sbwsg:
I'm going to propose that we table this PR until the TEP, https://github.com/tektoncd/community/pull/191, is resolved and merged.  We can reopen once that's happened and pick up the review then.

If anyone disagrees, please feel free to come back and reopen.

Issue #3007, opened 2020-07-24
Image Mismatch


dlorenc:
While (briefly) debugging #2998 with @skaegi, I noticed that the images we set as default flag values in [cmd/controller](https://github.com/tektoncd/pipeline/blob/master/cmd/controller/main.go#L37) differ from the images we set in [config/controller.yaml](https://github.com/tektoncd/pipeline/blob/master/config/controller.yaml).

I'm not sure how much of a problem this is, but it can be confusing. We should probably keep these up to date and matching somehow, but I'm not sure the best way.

dlorenc:
An example of the confusion: the flag for "shell-image' defaults to busybox, while the deployment uses gcr.io/distroless/base:debug

ImJasonH:
I think we should default the flags to "" and log.Fatal if they're unset. Having values in two places is confusing and error prone.

ImJasonH:
Found another one!

[`.ko.yaml`](https://github.com/tektoncd/pipeline/blob/d3ba81d5590c6e5925caa80f5952f86bae438a37/.ko.yaml#L11) configures the base image for the `entrypoint` image to be `gcr.io/distroless/base:debug-nonroot`. But [`tekton/publish.yaml`](https://github.com/tektoncd/pipeline/blob/d3ba81d5590c6e5925caa80f5952f86bae438a37/tekton/publish.yaml#L78) writes a `.ko.yaml` file that configures its base to be `busybox`.

I'm not sure how to catch and fail on that kind of mismatch in the future, but should at least fix it for now.

dlorenc:
Thanks for taking this on! Deleting the other defaults makes sense to me.

dlorenc:
@ImJasonH I think this can be closed now, thanks for sending all those PRs!

PR #3016, opened 2020-07-27
Remove default flag values for image names

ImJasonH:
These values were previously defined in two places (here in flag
defaults, and in config/controller.yaml), which could lead to confusing
mismatches (#3007)

/cc @dlorenc 

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Image names passed as flags in config/controller.yaml are now required -- there are no longer default values defined,  and a missing flag value will result in a fatal error when starting the controller.
```

### DIFF
--- a/pkg/apis/pipeline/images.go
+++ b/pkg/apis/pipeline/images.go
@@ -16,6 +16,11 @@ limitations under the License.
 
 package pipeline
 
+import (
+	"fmt"
+	"sort"
+)
+
 // Images holds the images reference for a number of container images used
 // across tektoncd pipelines.
 type Images struct {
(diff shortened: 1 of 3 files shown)


dlorenc:
Hah, nice. Much better than my ideas of a weird test to make sure they match.

pkg/apis/pipeline/images.go
```diff
@@ -40,3 +42,25 @@ type Images struct {
 	// ImageDigestExporterImage is the container image containing our image digest exporter binary.
 	ImageDigestExporterImage string
 }
+
+// Validate returns an error if any image is not set.
+func (i Images) Validate() error {
```
vdemeester, line 47: This will only show the first image that is "unset". Couldn't we add in the error *all* the images that are unset instead ?

ImJasonH: Good call, done. `Validate` is a bit uglier now, but the map approach didn't support multiple `""` keys.

PR #3017, opened 2020-07-27
Add context variables for PipelineRun & TaskRun UIDs and add validation for all context variables

### LINKED ISSUE #2958: Support more context variables (uid, annotations, …)
### Feature request

Related to https://github.com/tektoncd/pipeline/issues/2957, it would be nice if we could expose more metadata values in the variables, under `context`. Currently we do have `pipelineRun.name`, …

Additionnal variables could benefit from being exposed:
- `metadata.uid` (on PipelineRun and TaskRun). As it is the unique id set by k8s (or whatever other impl. in the future). It could be used, for example, as a parameter for a task to tag an oci image, …
- `metadata.annotations` (on PipelineRun and TaskRun). This *could* be useful in conjunction with something like https://github.com/tektoncd/community/pull/147 where you could then pass some "runtime" information to the task execution. How it is presented (aka an array, etc…) is something to discuss though.

### Use case

Base on #2957, it could help having build number sequence feature (decoupled from tektoncd/pipeline but accessible for the tasks executions)

> Being able to use something like `$(context.pipelineRun.buildNumber)` in an environment variable would make this easier.
> 
> Presumably the user would control when the sequence resets, and arguably be able to set a specific value on the sequence (similar to database sequences).

jerop:
# Changes

A user may want to tag an oci image with the TaskRun or PipelineRun UIDs.
Currently, they can't do that because `metadata.uid` for TaskRuns and
PipelineRuns are not exposed.

In this PR, we add the UID context variable for TaskRuns and
PipelineRun. Users can now use `$(context.taskRun.uid)` and
`$(context.pipelineRun.uid)` to access and use UIDs.

In addition, we add validation for all context variables that are
supported so far -- `context.task.name`, `context.taskRun.name`,
`context.taskRun.namespace`, `context.taskRun.uid`,
`context.pipeline.name`, `context.pipelineRun.name`,
`context.pipelineRun.namespace`, `context.pipelineRun.uid`.

Partially fixes #2958

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
- User can access the uid of the `PipelineRun` that a `Pipeline` is running in using `context.pipelineRun.uid`.
- User can access the uid of the `TaskRun` that a `Task` is running in using `context.taskRun.uid`. 
- All context variables that are supported so far are now validated. 
```

### DIFF
--- a/pkg/reconciler/taskrun/resources/apply.go
+++ b/pkg/reconciler/taskrun/resources/apply.go
@@ -97,11 +97,15 @@ func ApplyResources(spec *v1beta1.TaskSpec, resolvedResources map[string]v1beta1
 }
 
 // ApplyContexts applies the substitution from $(context.(taskRun|task).*) with the specified values.
-// Currently supports only name substitution. Uses "" as a default if name is not specified.
+// Uses "" as a default if a value is not available.
 func ApplyContexts(spec *v1beta1.TaskSpec, rtr *ResolvedTaskResources, tr *v1beta1.TaskRun) *v1beta1.TaskSpec {
-	return ApplyReplacements(spec,
-		map[string]string{"context.taskRun.name": tr.Name, "context.task.name": rtr.TaskName, "context.taskRun.namespace": tr.Namespace},
-		map[string][]string{})
+	replacements := map[string]string{
+		"context.taskRun.name":      tr.Name,
+		"context.task.name":         rtr.TaskName,
+		"context.taskRun.namespace": tr.Namespace,
+		"context.taskRun.uid":       string(tr.ObjectMeta.UID),
+	}
+	return ApplyReplacements(spec, replacements, map[string][]string{})
 }
 
 // ApplyWorkspaces applies the substitution from paths that the workspaces in w are mounted to, the
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -16,6 +16,7 @@ This page documents the variable substitions supported by `Tasks` and `Pipelines
 | `tasks.<taskName>.results.<resultName>` | The value of the `Task's` result. Can alter `Task` execution order within a `Pipeline`.) |
 | `context.pipelineRun.name` | The name of the `PipelineRun` that this `Pipeline` is running in. |
 | `context.pipelineRun.namespace` | The namespace of the `PipelineRun` that this `Pipeline` is running in. |
+| `context.pipelineRun.uid` | The uid of the `PipelineRun` that this `Pipeline` is running in. |
 | `context.pipeline.name` | The name of this `Pipeline` . |
 
 
@@ -33,6 +34,7 @@ This page documents the variable substitions supported by `Tasks` and `Pipelines
 | `credentials.path` | The path to credentials injected from Secrets with matching annotations. |
 | `context.taskRun.name` | The name of the `TaskRun` that this `Task` is running in. |
 | `context.taskRun.namespace` | The namespace of the `TaskRun` that this `Task` is running in. |
+| `context.taskRun.uid` | The uid of the `TaskRun` that this `Task` is running in. |
 | `context.task.name` | The name of this `Task`. |
 
 ### `PipelineResource` variables available in a `Task`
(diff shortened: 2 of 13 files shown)


pkg/reconciler/taskrun/resources/apply.go
```diff
@@ -97,11 +97,19 @@ func ApplyResources(spec *v1beta1.TaskSpec, resolvedResources map[string]v1beta1
...
-		map[string]string{"context.taskRun.name": tr.Name, "context.task.name": rtr.TaskName, "context.taskRun.namespace": tr.Namespace},
-		map[string][]string{})
+	replacements := map[string]string{
+		"context.taskRun.name":      tr.Name,
+		"context.task.name":         rtr.TaskName,
+		"context.taskRun.namespace": tr.Namespace,
+		"context.taskRun.uid":       string(tr.ObjectMeta.UID),
+	}
+	for key, value := range tr.ObjectMeta.Annotations {
+		replacementKey := fmt.Sprintf("context.taskRun.annotations.%s", key)
```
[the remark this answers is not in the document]
jerop: removed it, thanks!

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -362,6 +365,14 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun) err
 		return controller.NewPermanentError(err)
 	}
 
+	// Ensure that the context variables in the PipelineRun parameters are valid.
+	if err := resources.ValidateContextVariables(pipelineSpec, pr); err != nil {
```
vdemeester, line 369: Any reason for doing it in the reconciler loop ? we didn't do it before :monocle_face:

jerop: thanks for the review @vdemeester!

yes, we run it in the need to do it in the reconciler look to access the pipelineruns to get the name, uid etc -- @sbwsg and I explored trying to do it in other ways and this was the only option we had

yes, we didn't have validation for context variables before

sbwsg: Ah, I think I may have just realised a mistake I made - I suggested doing this in the reconciler when we were intending to implement annotations-as-variables.  When annotations are used we have to do the validation in the reconciler because there's no way to validate the annotations referenced by a Pipeline without also having the PipelineRun object - e.g. a Pipeline can use `$(context.pipelineRun.annotations.foo-bar-baz)` and we can't validate `foo-bar-baz` is correct until we also have the PipelineRun available to us.

However we then decided to push the annotations-as-variables into a separate PR because we figured there'd be more design and discussion required.  So, now that we're only validating the use of uid and names I think we probably can push the validation into the webhook, since the set of available variables is known to us in advance.  What do you think @jerop?

vdemeester: Right, that is, indeed my initial thought on this. If we have annotations and/or labels in the variables, then there needs to be a "reconciler" validation, but for the rest (uid, …), it can be done solely on the webhook side :upside_down_face:

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -300,6 +300,12 @@ func (c *Reconciler) prepare(ctx context.Context, tr *v1beta1.TaskRun) (*v1beta1
 		return nil, nil, controller.NewPermanentError(err)
 	}
 
+	if err := validateContextVariables(taskSpec, tr); err != nil {
```
vdemeester, line 303: I am having trouble seeing this one as non-exported in `taskrun` and the other one *exported* in `pipelinerun/resources`. Could they share code even ? (because we are "validating" the same thing, be it taskrun or other objects)

pkg/reconciler/pipelinerun/pipelinerun.go
jerop: yes yes, the issue was with annotations...just updated the validation

Issue #2093, opened 2020-02-24
Tekton variable expansion seems to have issues.


n3wscott:
Found two things:

Given, 

```
params:
    - name: FOO
      value: foo
    - name: FOOBAR
      value: $(params.FOO)-bar
    - name: FOOBARBAZ
      value: $(params.FOOBAR)-baz
```

1. Templates that have nested refs do not fully resolve.

ie, the value of FOOBARBAZ will not fully resolve some times, and will be `$(params.FOOBAR)-baz`

2. Order seems to matter. Sometimes a run will have the value of FOOBAR as `foo-bar` and other times it will be `$(params.FOO)-bar` like `params.FOO` has not been processed yet.

vdemeester:
Thanks @n3wscott. 

The first one is definitely something we do not support at that point of time.
The second seems to be a bug as it should work. 

Few questions:
- which version of tekton are you using ? of the api (I guess `v1alpha1`) ?
- is this only on `Pipeline` / `PipelineRun` objects or also on `Task`/`TaskRun` ?

/kind bug
/kind feature

n3wscott:
I am using the latest release yaml, v1alpha1

It seems to be easer to hit when using Pipelines, but I can get it to happen with Task too but less.

vdemeester:
@n3wscott can you share your yamls to reproduce it ?

n3wscott:
yeah, I have reworked this a few times, but an older stapshot: https://github.com/n3wscott/chatbot/commit/67ac3c226fbfdc09e171ffb28a95c31f2a3040c4#diff-34cd834803c97715e915dd79c582e49cR54

and a workaround example: https://github.com/n3wscott/chatbot/blob/master/runs/release-0.13/run-update-toml-deps-sample-controller.yaml#L36

afrittoli:
> Thanks @n3wscott.
> 
> The first one is definitely something we do not support at that point of time.
> The second seems to be a bug as it should work.
> 
> Few questions:
> 
> * which version of tekton are you using ? of the api (I guess `v1alpha1`) ?
> * is this only on `Pipeline` / `PipelineRun` objects or also on `Task`/`TaskRun` ?
> 
> /kind bug
> /kind feature

Having the ability of doing in place modifications of params reduces the need for extra steps / tasks and increases the re-usability of tasks. I didn't know we supported even string interpolation (not-nested), that's good to know.

afrittoli:
[...]

PR #3024, opened 2020-07-28
Make string variable interpolation deterministic, and single-pass.

dlorenc:
# Changes

Make string variable interpolation deterministic, and single-pass.

Right now we replace string variables one at a time, in an order governed by
iteration through a map[string]string. This has two implications:
- We may expand variables twice. If one variable expands to another variable, that one may be expanded. And so on.
- Whether or not this secondary expansion occurs (and even how it occurs!) can vary from run to run.

This commit changes to replace everything at once, using the strings.Replacer package.
It also adds a test to ensure replacements are stable and non-recursive.

See https://github.com/tektoncd/pipeline/issues/2093

Note, we could go the other way this PR and instead embrace nested expansions. This one is simply my preference based on some reasons outlined in the issue.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
The variable interpolation algorithm has been changed. Nested expansion is no longer supported.
If you use replacements ( $(...) syntax) in your TaskRuns or PipelineRuns that resolve to secondary replacements, these will no longer work.

The algorithm is now deterministic and will result in the same expansion every run.
```

### DIFF
--- a/pkg/substitution/substitution.go
+++ b/pkg/substitution/substitution.go
@@ -118,10 +118,14 @@ func matchGroups(matches []string, pattern *regexp.Regexp) map[string]string {
 }
 
 func ApplyReplacements(in string, replacements map[string]string) string {
+	replacementsList := []string{}
 	for k, v := range replacements {
-		in = strings.Replace(in, fmt.Sprintf("$(%s)", k), v, -1)
+		replacementsList = append(replacementsList, fmt.Sprintf("$(%s)", k), v)
 	}
-	return in
+	// strings.Replacer does all replacements in one pass, preventing multiple replacements
+	// See #2093 for an explanation on why we need to do this.
+	replacer := strings.NewReplacer(replacementsList...)
+	return replacer.Replace(in)
 }
 
 // Take an input string, and output an array of strings related to possible arrayReplacements. If there aren't any
--- a/pkg/substitution/substitution_test.go
+++ b/pkg/substitution/substitution_test.go
@@ -167,6 +167,24 @@ func TestApplyReplacements(t *testing.T) {
 	}
 }
 
+func TestNestedReplacements(t *testing.T) {
+	replacements := map[string]string{
+		// Foo should turn into barbar, which could then expand into bazbaz depending on how this is expanded
+		"foo": "$(bar)$(bar)",
+		"bar": "baz",
+	}
+	input := "$(foo) is cool"
+	expected := "$(bar)$(bar) is cool"
+
+	// Run this test a lot of times to ensure the behavior is deterministic
+	for i := 0; i <= 1000; i++ {
+		got := substitution.ApplyReplacements(input, replacements)
+		if d := cmp.Diff(expected, got); d != "" {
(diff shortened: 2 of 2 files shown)


dlorenc:
Note that I'm not sure this is the best way to solve this issue. I could also easily be persuaded to correctly handle nested expansions, if we want to go that way. The important thing is that it's easy to describe and deterministic.

dlorenc:
I'm also not sure how much of a "breaking change" this is. Nested expansions were never documented as working, and only worked sometimes. That said, if you were using them and got lucky, this would break you.

pkg/substitution/substitution.go
```diff
@@ -118,10 +118,21 @@ func matchGroups(matches []string, pattern *regexp.Regexp) map[string]string {
 }
 
 func ApplyReplacements(in string, replacements map[string]string) string {
-	for k, v := range replacements {
-		in = strings.Replace(in, fmt.Sprintf("$(%s)", k), v, -1)
+	// We use a regular expression and ReplaceAll here to ensure we only expand once.
+	// With strings.Replace in a loop, we could expand variables multiple times.
+	// See https://github.com/tektoncd/pipeline/issues/2093
+	re := regexp.MustCompile(`\$\((.+?\))`)
```
dlorenc, line 124: I'm not quite sure if there's a better character class than `.*` here.

This function accepts any string map key, and previously did find/replace for any type of key...

dlorenc:

Not seeing these failures locally, retrying

dlorenc:
Now with fewer regular expressions, and more passing tests!

dlorenc:
cc @skaegi who has ideas as discussed in the community meeting

skaegi:
I'm not at all opposed to making it single-pass for now! The current way we have it just leads to confusion.

With this fix in place, in the future if we add order deterministic param expansion without breaking backwards compatibility.

dlorenc:
> With this fix in place, in the future we can add ordered deterministic param expansion without breaking backwards compatibility.

Ah, makes sense. Thanks for clarifying!

bobcatfish:
>  I'm not actually sure if this is an API change or not and would like to discuss it a bit before merging.

I'm not sure either! I guess it depends on whether you view it as a bug or not ( i think @ImJasonH has some kind of rule about that ). I think it is a bug, however it's also possible some folks are relying on this functionality even so.

My vote would be to go ahead and make the backwards incompatible change; if we wanted to be cautious we could have a release where we try to detect someone is using this functionality and warn them it's going to be removed?

dlorenc:
/hold cancel

Thanks!

dlorenc:
For my rationale on why I think this is ok:

- Secondary expansion is definitely unintended behavior, but I wouldn't call it a bug.
- The bug is the non-deterministic nature of the expansion.
- It's not really possible to depend on this, since it's basically random.
- Since it can't be depended on, we can change it.

ImJasonH:
I agree it's a bug and we should feel free to change it.

If we want to support multi-pass interpolation we'll need to have a use case first, then a short TEP. :)

Thanks for catching this Dan!

Issue #1307, opened 2019-09-13
Find a threadsafe way to log in timeout_handler


bobcatfish:
# Expected Behavior

We should be able to emit logs from the timeout_handler in a threadsafe manner. The logic in the timeout_handler is intended to be running in separate goroutines alongside the main reconcile loop execution, to manage actions like timing out and retrying.

# Actual Behavior

We were using the same zap sugared logger that we were using everywhere else for logging, e.g.:

https://github.com/tektoncd/pipeline/blob/5e3276d3df90c4d9b7c8d42c16c058a8d2348809/pkg/reconciler/timeout_handler.go#L276-L290

However when we started running tests with newer versions of go (e.g. 1.13 v 1.11 see #1124 ) and the `-race` flag, we found there were data races.

Not 100% sure what the root cause is but https://github.com/uber-go/zap/issues/99 suggests that zap logger writes are not protected and could have a race.

So in #1124 we're removing the logging, which is unfortunate b/c we used it previously to help us debug #731 and itd be great to have some logging for debugging any future issues.

# Additional Info

(Is "threadsafe" the right word to use when we're talking about goroutines??)

PR #3031, opened 2020-07-29
Move backoff logic into own struct and increase test coverage of timeout logic

bobcatfish:
# Changes

## Separate backoff logic from timeout logic ⏰

The timeout handler was trying to do two different jobs:
- Track timeouts for executing TaskRuns and PipelineRuns; when they
  timeout, they should be re-reconciled, at which point they would
  fail b/c they took to long
- Track backoffs when TaskRun pod creation fails due to resource quota
  issues. In this case we want to retry pod creation (by re-reconciling)
  after a backoff in case the resource issue has been resolved.

The logic for these 2 jobs was combined because they both share at their
heart very similar logic: the logic for creating go routines that wait
until a period of time has passed.

Now these two jobs are done by 2 different structs, which both use the
new Timer struct and logic, so they can share this logic, without having
to mix the code on top.

Future things we can do to make this even better:
- Add more test coverage for the new structs
- The "Handler" struct (which may have too generic a name at this point)
  probably doesn't need to have separate logic for pipelineruns +
  taskruns since each of these controllers will instanitate it
  separately
- As discussed in #2905, knative/pkg's controller logic will attempt
  to re-reconcile all the Runs on start up, so we probably don't need
  to be explicitly doing this as well

## Add more timeout test coverage and fix timeout "bug" 🐞

Add more tests to increase coverage of timeout package. This means
updating some code to be more unit testable, specifically injecting the
function used to determine the current time into Backoff.

This revealed a "bug" where the backoff logic wasn't using a default
value when the TaskRun timeout isn't set - however the defaulting logic
ensures that it is always set, so if anything, maybe we should actually
panic in this case.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

It almost seems like this is fixing a bug, but because of how the defaulting logic works, I don't think it really is.

```release-note
NONE
```


bobcatfish:
oooo a data race!

```
==================
WARNING: DATA RACE
Read at 0x00c000c3c198 by goroutine 103:
  reflect.typedmemmove()
      /usr/local/go/src/runtime/mbarrier.go:177 +0x0
  reflect.packEface()
      /usr/local/go/src/reflect/value.go:119 +0x126
  reflect.valueInterface()
      /usr/local/go/src/reflect/value.go:1023 +0x1b9
  reflect.Value.Interface()
      /usr/local/go/src/reflect/value.go:993 +0x3c27
  fmt.(*pp).printValue()
      /usr/local/go/src/fmt/print.go:726 +0x3c28
  fmt.(*pp).printValue()
      /usr/local/go/src/fmt/print.go:880 +0x2709
  fmt.(*pp).printArg()
      /usr/local/go/src/fmt/print.go:716 +0x25a
  fmt.(*pp).doPrintf()
      /usr/local/go/src/fmt/print.go:1030 +0x311
  fmt.Sprintf()
      /usr/local/go/src/fmt/print.go:219 +0x73
  go.uber.org/zap.(*SugaredLogger).log()
      /home/prow/go/src/github.com/tektoncd/pipeline/vendor/go.uber.org/zap/sugar.go:230 +0x175
  go.uber.org/zap.(*SugaredLogger).Errorf()
      /home/prow/go/src/github.com/tektoncd/pipeline/vendor/go.uber.org/zap/sugar.go:148 +0x331
  github.com/tektoncd/pipeline/pkg/timeout.(*Backoff).SetTimer()
      /home/prow/go/src/github.com/tektoncd/pipeline/pkg/timeout/backoff.go:142 +0x29a

Previous write at 0x00c000c3c198 by goroutine 79:
  knative.dev/pkg/apis/duck/v1beta1.(*Status).SetConditions()
      /home/prow/go/src/github.com/tektoncd/pipeline/vendor/knative.dev/pkg/apis/duck/v1beta1/status_types.go:84 +0x43
  github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1.(*TaskRunStatus).SetConditions()
      <autogenerated>:1 +0x61
  knative.dev/pkg/apis.conditionsImpl.SetCondition()
      /home/prow/go/src/github.com/tektoncd/pipeline/vendor/knative.dev/pkg/apis/condition_set.go:208 +0x3ee
  knative.dev/pkg/apis.(*conditionsImpl).SetCondition()
      <autogenerated>:1 +0x149
  github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1.(*TaskRunStatus).SetCondition()
      /home/prow/go/src/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1/taskrun_types.go:253 +0x43c
  github.com/tektoncd/pipeline/pkg/reconciler/taskrun.(*Reconciler).handlePodCreationError()
      /home/prow/go/src/github.com/tektoncd/pipeline/pkg/reconciler/taskrun/taskrun.go:470 +0x144
  github.com/tektoncd/pipeline/pkg/reconciler/taskrun.TestHandlePodCreationError.func1()
      /home/prow/go/src/github.com/tektoncd/pipeline/pkg/reconciler/taskrun/taskrun_test.go:2100 +0xd5
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:992 +0x1eb

Goroutine 103 (running) created at:
  github.com/tektoncd/pipeline/pkg/reconciler/taskrun.(*Reconciler).handlePodCreationError()
      /home/prow/go/src/github.com/tektoncd/pipeline/pkg/reconciler/taskrun/taskrun.go:467 +0x52a
  github.com/tektoncd/pipeline/pkg/reconciler/taskrun.TestHandlePodCreationError.func1()
      /home/prow/go/src/github.com/tektoncd/pipeline/pkg/reconciler/taskrun/taskrun_test.go:2100 +0xd5
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:992 +0x1eb

Goroutine 79 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1043 +0x660
  github.com/tektoncd/pipeline/pkg/reconciler/taskrun.TestHandlePodCreationError()
      /home/prow/go/src/github.com/tektoncd/pipeline/pkg/reconciler/taskrun/taskrun_test.go:2099 +0x12bd
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:992 +0x1eb
```

bobcatfish:
I think this race condition might be another instance of #1307 🤔

pkg/timeout/backoff.go
```diff
@@ -0,0 +1,148 @@
...
+		jittered = maxBackoffSeconds
+	}
+	return time.Duration(jittered) * time.Second
+}
+
+// SetTimer creates a blocking function to wait for
+// 1. Stop signal, 2. completion or 3. a given Duration to elapse.
+func (b *Backoff) SetTimer(runObj StatusKey, d time.Duration) {
+	if b.timeoutCallback == nil {
+		b.logger.Errorf("attempted to set a timer for %q but no callback has been assigned", runObj)
```
pritidesai, line 142: Sounds very odd (at least to me 😉), commenting out this `logger` from here fixes race condition. This function invocation within [handlePodCreationError](https://github.com/tektoncd/pipeline/pull/3031/files?diff=split&w=1#diff-14481606a469dcd50f6cb9f5540e3b06R467) and the following [SetCondition](https://github.com/tektoncd/pipeline/pull/3031/files?diff=split&w=1#diff-14481606a469dcd50f6cb9f5540e3b06R470) are somehow causing this.

pritidesai:
```
WARNING: DATA RACE
Read at 0x00c000c3c198 by goroutine 103:
...
go.uber.org/zap.(*SugaredLogger).log()
...

Previous write at 0x00c000c3c198 by goroutine 79:
  knative.dev/pkg/apis/duck/v1beta1.(*Status).SetConditions()
...
```

sbwsg: Ah, well spotted!  It looks like swapping out the `%q` for a `%p` in this log statement also removes the error.  Using `%q` results in a string-formatted struct being printed, which probably walks the entire object and accesses all the fields.  My hunch is that `%q` is not threadsafe in that case?

pritidesai: nice catch, yup it does walk the entire object. The [next](https://github.com/tektoncd/pipeline/pull/3031/files#diff-fc314ed1e1edf4a1e359bff0f4f2c47aR145) log statement (using `GetRunKey()` on `runObj`) looks safe and just prints the address of `taskRun` object instead.

bobcatfish:
While trying to figure this out I think I found a bigger problem, I'm going to close this PR for now b/c I'll need to rebase it anyway.

> My hunch is that the %q formatter is not threadsafe in that case? In hindsight this may be why we have the StatusKey interface - the GetRunKey() func of TaskRun returns a thread-safe descriptor.

In fact I think this is a bug that I introduced - the address of the Run might be thread safe, but it's not going to be the same on every reconcile! So I think that actually _every single Run is timing out_ because the done channel can never be signaled 😅

PR #2547, opened 2020-05-05
Allow for setting imagePullSecrets

NikeNano:
# Changes

NOTE: I am not assigned to the [issue](https://github.com/tektoncd/pipeline/issues/1779), but fixed it since I saw that it was stale for some time and thought it would be fun to try to fix it. 

This PR updates the [PodTemplate](https://github.com/tektoncd/pipeline/blob/a7fb3894310227cfc9e4e4303b8d0322beb6d95b/pkg/apis/pipeline/pod/template.go#L27) with ImagePullSecret to allow for setting image pull secrets. This PR aims to fix issue https://github.com/tektoncd/pipeline/issues/1779. 

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

Double check this list of stuff that's easy to miss:

- If you are adding [a new binary/image to the `cmd` dir](../cmd), please update
  [the release Task](../tekton/publish.yaml) to build and release this image.

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

# Release Notes

```
Allow to specify ImagePullSecrets in the PodTemplate.

```

Issue #17, opened 2018-09-07
Add `+optional` to optional fields


bobcatfish:
## Expected Behavior
[The k8s API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#optional-vs-required) explain that optional fields should have both the json `omitempty` directive _and_ should have the `+optional` comment.

## Actual Behavior
We are currently using only "omitempty" to indicate which fields are optional, e.g. a [Pipeline's](https://github.com/knative/build-pipeline#pipeline) [Task](https://github.com/knative/build-pipeline#task) source bindings: https://github.com/knative/build-pipeline/blob/ef113e1d513c67b692eb1096904789be0e9616b6/pkg/apis/pipeline/v1beta1/pipeline_types.go#L55

## Additional Info
* [k8s API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#optional-vs-required)

PR #3032, opened 2020-07-30
Add omitempty for SchedulerName, ImagePullSecrets, HostNetwork

danielhelfand:
# Changes 

This pull request adds `omitempty` for the following pod template fields:
* SchedulerName
* ImagePullSecrets
* HostNetwork

As documented in #17, `SchedulerName` and `HostNetwork` are designated as optional fields and as such should be omitted if not declared. Why ImagePullSecrets is not considered optional seems interesting to me, and perhaps should be discussed a bit further before being introduced. But overall, it might have been overlooked when introducing #2547.

Please let me know if there should be further discussions with regard to this being a breaking change.

# Submitter Checklist

## Reviewer Notes

# Release Notes

```release-note
Omit SchedulerName, HostNetwork, and ImagePullSecrets if not declared for PodTemplate
```

### DIFF
--- a/pkg/apis/pipeline/pod/template.go
+++ b/pkg/apis/pipeline/pod/template.go
@@ -92,14 +92,15 @@ type Template struct {
 	PriorityClassName *string `json:"priorityClassName,omitempty" protobuf:"bytes,7,opt,name=priorityClassName"`
 	// SchedulerName specifies the scheduler to be used to dispatch the Pod
 	// +optional
-	SchedulerName string `json:"schedulerName"`
+	SchedulerName string `json:"schedulerName,omitempty"`
 
 	// ImagePullSecrets gives the name of the secret used by the pod to pull the image if specified
-	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets"`
+	// +optional
+	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
 
 	// HostNetwork specifies whether the pod may use the node network namespace
 	// +optional
-	HostNetwork bool `json:"hostNetwork"`
+	HostNetwork bool `json:"hostNetwork,omitempty"`
 }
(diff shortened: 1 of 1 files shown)


pkg/apis/pipeline/pod/template.go
```diff
@@ -92,14 +92,14 @@ type Template struct {
 	PriorityClassName *string `json:"priorityClassName,omitempty" protobuf:"bytes,7,opt,name=priorityClassName"`
 	// SchedulerName specifies the scheduler to be used to dispatch the Pod
 	// +optional
-	SchedulerName string `json:"schedulerName"`
+	SchedulerName string `json:"schedulerName,omitempty"`
 
 	// ImagePullSecrets gives the name of the secret used by the pod to pull the image if specified
-	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets"`
+	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
```
pritidesai, line 98: @danielhelfand please add `// +optional` for this field if @ImJasonH agrees to this change 😉

ImJasonH:
/lgtm

This doesn't qualify as a breaking change because users shouldn't have been depending on specifics like `""` being returned in JSON responses from the server.

danielhelfand:
> /lgtm
> 
> This doesn't qualify as a breaking change because users shouldn't have been depending on specifics like `""` being returned in JSON responses from the server.

Thanks for the clarification. I guess my concern is more about the request for developers working around the API. Those fields are included even when not declared (e.g., when I tried to introduce pod template support in `tkn`). That being said, I would consider that to be unexpected.

pkg/apis/pipeline/pod/template.go
danielhelfand: Added the comment. Thanks!

PR #3033, opened 2020-07-30
Remove mention of ssh-keygen from auth doc

sbwsg:
# Changes

The auth docs mention that tekton generates a known_hosts file using
ssh-keygen but this isn't quite right - we used to fetch the key
using ssh-keyscan but now accept it by updating git's core.sshCommand.

This PR updates the auth doc to describe the implementation we use to
populate known_hosts when its missing.

# Submitter Checklist

# Release Notes

```release-note
Fix how we describe the way Tekton populates known_hosts files when they're missing from Secrets.
```

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -494,9 +494,10 @@ Host url2.com
 ...
 ```
 
-Note: Because `known_hosts` is a non-standard extension of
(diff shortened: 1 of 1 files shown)


pritidesai:
This looks great, thanks @sbwsg for a quick response 👍 

/approve

Issue #3051, opened 2020-08-03
Timed Out or Cancelled TaskRun Pods are Deleted


danielhelfand:
# Expected Behavior

After a TaskRun is timed out or cancelled, I should still be able to view the logs of the failed or cancelled TaskRun.

# Actual Behavior

It appears that the pods for TaskRuns are being deleted due to a change implemented in #2365. In the func, [`failTaskRun`](https://github.com/tektoncd/pipeline/blob/7f71aa8e42a4ead15cd64861fabc1dc65f8c535d/pkg/reconciler/taskrun/taskrun.go#L495) in `taskrun.go`, the pod associated with the TaskRun is being deleted as noted [here](https://github.com/tektoncd/pipeline/pull/2365/files#r464649272). 

Assuming this is not an expected feature, I would suggest checking for the failure reason before deleting the TaskRun pod:

```go
if reason != v1beta1.TaskRunReasonTimedOut && reason != v1beta1.TaskRunReasonCancelled {
		err := c.KubeClientSet.CoreV1().Pods(tr.Namespace).Delete(tr.Status.PodName, &metav1.DeleteOptions{})

		if err != nil && !k8serrors.IsNotFound(err) {
			logger.Infof("Failed to terminate pod: %v", err)
			return err
		}
}
```

If this is expected, it would help to document this behavior for TaskRuns as well as recommended best practices for preserving logs.

# Steps to Reproduce the Problem

1. Create a TaskRun with a timeout such that the timeout will be violated or cancel a running TaskRun
2. Run `kubectl get pods` to find if the pod of the TaskRun has been deleted

# Additional Info

- Kubernetes version:

N/A

- Tekton Pipeline version:

Using what is in latest as of v0.15.0.

danielhelfand:
/assign

Will wait for confirmation on if this is expect for TaskRuns, but can pick this up if it is unintended.

vdemeester:
> It appears that the pods for TaskRuns are being deleted due to a change implemented in #2365. In the func, [`failTaskRun`](https://github.com/tektoncd/pipeline/blob/7f71aa8e42a4ead15cd64861fabc1dc65f8c535d/pkg/reconciler/taskrun/taskrun.go#L495) in `taskrun.go`, the pod associated with the TaskRun is being deleted as noted [here](https://github.com/tektoncd/pipeline/pull/2365/files#r464649272).

[...]

PR #3053, opened 2020-08-04
Include More Details on Pod Deletion for Cancelled/Timed Out TaskRuns

danielhelfand:
Part of #3051 

This pull request updates documentation on pod deletion for both timed out/cancelled TaskRuns to mention that TaskRun pod deletion is part of the process. It also provides a brief amount of information on why this takes place. 

In addition to the updates above, there are some minor changes also made to reorganize the content. 

# Submitter Checklist

## Reviewer Notes

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/taskruns.md
+++ b/docs/taskruns.md
@@ -254,18 +254,23 @@ For more information, see the [`LimitRange` code example](../examples/v1beta1/ta
 
 ## Configuring the failure timeout
 
-You can use the `timeout` field to set the `TaskRun's` desired timeout value in minutes.
-If you do not specify this value in the `TaskRun`, the global default timeout value applies.
-If you set the timeout to 0, the `TaskRun` fails immediately upon encountering an error.
+You can use the `timeout` field to set the `TaskRun's` desired timeout value. If you do not specify this 
+value for the `TaskRun`, the global default timeout value applies. If you set the timeout to 0, the `TaskRun` will 
+have no timeout and will run until it completes successfully or fails from an error.
 
 The global default timeout is set to 60 minutes when you first install Tekton. You can set
 a different global default timeout value using the `default-timeout-minutes` field in
-[`config/config-defaults.yaml`](./../config/config-defaults.yaml).
+[`config/config-defaults.yaml`](./../config/config-defaults.yaml). If you set the global timeout to 0, 
+all `TaskRuns` that do not have a timeout set will have no timeout and will run until it completes successfully 
+or fails from an error.
 
 The `timeout` value is a `duration` conforming to Go's
 [`ParseDuration`](https://golang.org/pkg/time/#ParseDuration) format. For example, valid
-values are `1h30m`, `1h`, `1m`, and `60s`. If you set the global timeout to 0, all `TaskRuns`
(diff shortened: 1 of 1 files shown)


docs/taskruns.md
```diff
@@ -254,18 +254,23 @@ For more information, see the [`LimitRange` code example](../examples/v1beta1/ta
 
 ## Configuring the failure timeout
 
-You can use the `timeout` field to set the `TaskRun's` desired timeout value in minutes.
-If you do not specify this value in the `TaskRun`, the global default timeout value applies.
-If you set the timeout to 0, the `TaskRun` fails immediately upon encountering an error.
+You can use the `timeout` field to set the `TaskRun's` desired timeout value. If you do not specify this 
+value for the `TaskRun`, the global default timeout value applies. If you set the timeout to 0, the `TaskRun` will 
+have no timeout and will fail immediately upon encountering an error.
```
bobcatfish, line 259: i'm nitpicking wording that isn't yours but "the `TaskRun` will have no timeout and will fail immediately upon encountering an error" confused me a bit b/c "fail immediately upon encountering an error" is true regardless of whether there is a timeout or not

i think what this might want to say is something like "the `TaskRun` will  have no timeout and will run until it completes either successfully or with an error"?

docs/taskruns.md
```diff
@@ -254,18 +254,23 @@ For more information, see the [`LimitRange` code example](../examples/v1beta1/ta
...
 
 The global default timeout is set to 60 minutes when you first install Tekton. You can set
 a different global default timeout value using the `default-timeout-minutes` field in
-[`config/config-defaults.yaml`](./../config/config-defaults.yaml).
+[`config/config-defaults.yaml`](./../config/config-defaults.yaml). If you set the global timeout to 0, 
+all `TaskRuns` that do not have a timeout set will have no timeout and will fail immediately upon 
+encountering an error.
```
bobcatfish, line 265: same comment about the error

docs/taskruns.md
```diff
@@ -254,18 +254,23 @@ For more information, see the [`LimitRange` code example](../examples/v1beta1/ta
...
 
 The `timeout` value is a `duration` conforming to Go's
 [`ParseDuration`](https://golang.org/pkg/time/#ParseDuration) format. For example, valid
-values are `1h30m`, `1h`, `1m`, and `60s`. If you set the global timeout to 0, all `TaskRuns`
-that do not have an individual timeout set will fail immediately upon encountering an error.
+values are `1h30m`, `1h`, `1m`, and `60s`.
```
bobcatfish, line 269: is 0 not a valid value here? maybe worth mentioning that explicitly

docs/taskruns.md
danielhelfand: Thanks! I changed to the following:

```
If you set the global timeout to 0, all `TaskRuns` that do not have a timeout set 
will have no timeout and will run until it completes successfully or fails from an error.
```

docs/taskruns.md
danielhelfand: Yes, it is valid. I have added it in.

Issue #3054, opened 2020-08-04
Sharing pullrequest metadata from input to output PR resource results in permission error


sbwsg:
# Expected Behavior

One usecase for pull request resources is to have an input PR resource pull down the current state of a PR and then an output PR resource sync the updated state back to github.  In order to make this work a step needs to be added to copy the input resource's metadata to the output resource's workspace directory (e.g. `/workspace/output/pr`). Once the metadata is copied into the output workspace the output resource should then be able to sync those changes back to github.

# Actual Behavior

At the moment this does not appear to work - during pullrequest upload the pullrequest resource attempts to read from /workspace/output/pr/pr.json and gets a permission denied error like this:

```
[pr-source-pr-sng8f] {"level":"info","ts":1596551207.2666166,"caller":"pullrequest-init/main.go:68","msg":"RUNNING UPLOAD!","resource_type":"pullrequest","mode":"upload"}
[pr-source-pr-sng8f] {"level":"fatal","ts":1596551207.2668488,"caller":"pullrequest-init/main.go:71","msg":"open /workspace/output/pr/pr.json: permission denied","resource_type":"pullrequest","mode":"upload","stacktrace":"main.main\n\tgithub.com/tektoncd/pipeline/cmd/pullrequest-init/main.go:71\nruntime.main\n\truntime/proc.go:203"}
```

# Steps to Reproduce the Problem

The following YAML reproduces the problem on current master:

```yaml
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  name: pr-test
spec:
  taskSpec:
    resources:
      inputs:
        - name: pr
          type: pullRequest
      outputs:
        - name: pr
          type: pullRequest
    steps:
    - name: copy-pr-to-output
      image: busybox
      script: |
        #!/bin/sh
        mkdir -p $(outputs.resources.pr.path)
        cp -r $(inputs.resources.pr.path)/* $(outputs.resources.pr.path)/
  resources:
    inputs:
    - name: pr
      resourceSpec:
        type: pullRequest
        params:
        - name: url
          value: https://github.com/tektoncd/pipeline/pull/100
    outputs:
    - name: pr
      resourceSpec:
        type: pullRequest
        params:
        - name: url
          value: https://github.com/tektoncd/pipeline/pull/100
```

# Additional Info

This may be related to the fact that we build the entrypoint using a nonroot base image.  Still investigating.

PR #3018, opened 2020-07-27
Align release-generated .ko.yaml with repo config

ImJasonH:
This resolves disparities between the .ko.yaml normally used to build
images during development with the one generated and used during a
release (to ensure git-init and creds-init are built on the just-built
build-base image)

This won't prevent future drift.

#3007 

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [n] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [n] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [y] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [y] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

Double check this list of stuff that's easy to miss:

- If you are adding [a new binary/image to the `cmd` dir](../cmd), please update
  [the release Task](../tekton/publish.yaml) to build and release this image.

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

# Release Notes

```release-note
- Released entrypoint image is based on gcr.io/distroless/base:debug-nonroot instead of busybox
- Released gcs-fetcher image is based on gcr.io/distroless/static:latest
```

/cc @dlorenc @vdemeester

PR #2606, opened 2020-05-12
Run tekton containers as nonroot 🐰

vdemeester:
# Changes

This changes a slew of containers that Tekton runs to use non-root base images.

This carry #2435 (discussed with @mattmoor)
Closes #2435

Signed-off-by: Vincent Demeester <vdemeest@redhat.com>

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [ ] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

Double check this list of stuff that's easy to miss:

- If you are adding [a new binary/image to the `cmd` dir](../cmd), please update
  [the release Task](../tekton/publish.yaml) to build and release this image.

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

# Release Notes

```
Switch many of the Tekton images (e.g. controllers) to non-root by default.
```

PR #3055, opened 2020-08-04
Ensure pullrequest-init is based on a root image

sbwsg:
fixes https://github.com/tektoncd/pipeline/issues/3054

# Changes

The PullRequest Resource, when used as an output, is able to
read in a pr.json to determine if there have been any changes
that require syncing to github. pr.json may have been written
by any prior Step with any ownership settings. If pr.json
was written with root permissions then the PullRequest Resource
needs to be have permissions to read that file.

The PullRequest Resource image has been based on a nonroot
image in our `.ko.yaml` since 0.13 of Tekton Pipelines ([`.ko.yaml` was
updated here](https://github.com/tektoncd/pipeline/pull/2606)).

However, the published images did not match the configuration in the
`.ko.yaml` until 0.15.0 ([our `tekton/publish.yaml` was brought into line
with `.ko.yaml` here](https://github.com/tektoncd/pipeline/pull/3018)).

Given that copying or writing pr.json in a Step can result in the file
being owned by root using a nonroot image is not a suitable choice
of base image - the output PullRequest attempts to open pr.json and
hits a permissions error.

This commit updates the PullRequest image to be based on
distroless static instead of nonroot and adds an example yaml
file that should exercise the behaviour of copying the file from
an input to output pullrequest resource.

# Submitter Checklist

review them:

# Release Notes

```release-note
Fix an issue where an output pullrequest resource could fail because a pr.json file was written by a Step with root ownership.
```

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -6,6 +6,9 @@ baseImageOverrides:
   github.com/tektoncd/pipeline/cmd/git-init: gcr.io/tekton-nightly/github.com/tektoncd/pipeline/build-base:latest
   # GCS fetcher needs root due to workspace permissions
   github.com/tektoncd/pipeline/vendor/github.com/GoogleCloudPlatform/cloud-builders/gcs-fetcher/cmd/gcs-fetcher: gcr.io/distroless/static:latest
+  # PullRequest resource needs root because in output mode it needs to access pr.json
+  # which might have been copied or written with any level of permissions.
+  github.com/tektoncd/pipeline/cmd/pullrequest-init: gcr.io/distroless/static:latest
 
   # Our entrypoint image does not need root, it simply needs to be able to 'cp' the binary into a shared location.
   github.com/tektoncd/pipeline/cmd/entrypoint: gcr.io/distroless/base:debug-nonroot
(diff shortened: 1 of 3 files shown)


dlorenc:
> However, the published images did not match the configuration in the
> `.ko.yaml` until 0.15.0 ([our `tekton/publish.yaml` was brought into line with `.ko.yaml` here](https://github.com/tektoncd/pipeline/pull/3018)).

Ohhhh! Nice debugging.

.ko.yaml
```diff
@@ -6,6 +6,9 @@ baseImageOverrides:
   github.com/tektoncd/pipeline/cmd/git-init: gcr.io/tekton-nightly/github.com/tektoncd/pipeline/build-base:latest
   # GCS fetcher needs root due to workspace permissions
   github.com/tektoncd/pipeline/vendor/github.com/GoogleCloudPlatform/cloud-builders/gcs-fetcher/cmd/gcs-fetcher: gcr.io/distroless/static:latest
+  # PullRequest resource needs root because in output mode it needs to access pr.json
+  # which might have been copied or written with any level of permissions.
+  github.com/tektoncd/pipeline/cmd/pullrequest-init: gcr.io/distroless/static:latest
```
bobcatfish, line 11: is it true that this would apply to any container that needs to operate on any file written by a previous step? if so this feels like it might be a slightly larger problem since it feels like communicating b/w steps is a big part of our design 🤔  (maybe im over exaggerating that?)

sbwsg: Yeah, I do think that's a risk we might want to spend a bit more time analysing.  Here I'm fixing the immediate issue we observed in dogfooding wrt the PR pipeline resource.

PR #3069, opened 2020-08-07
Allow skipping some YAML tests

chmouel:
Allow skipping some of the yaml tests if the environment variable
`TEST_EXAMPLES_IGNORES` is set and match the regexp :

For example :

`export TEST_EXAMPLES_IGNORES=".*clustertasks.*"`

will skip all the kankiko examples yamls.

We have a need for this because we want to run the go yaml examples testsuite
but need to skip some of those tests that are not compatible with openshift due
of some restrictions on that system.

/kind misc
/area testing

# Changes

# Submitter Checklist

review them:

# Release Note

```release-note
NONE
```

### DIFF
--- a/test/examples_test.go
+++ b/test/examples_test.go
@@ -181,6 +181,18 @@ func getExamplePaths(t *testing.T, dir string) []string {
 			return filepath.SkipDir
 		}
 		if info.IsDir() == false && filepath.Ext(info.Name()) == ".yaml" {
+			// Ignore test matching the regexp in the TEST_EXAMPLES_IGNORES
+			// environement variable.
+			val, ok := os.LookupEnv("TEST_EXAMPLES_IGNORES")
+			if ok {
+				re := regexp.MustCompile(val)
+				submatch := re.FindSubmatch([]byte(path))
+				if submatch != nil {
+					t.Logf("Skipping test %s", path)
+					return nil
+				}
+			}
+			t.Logf("Adding test %s", path)
 			examplePaths = append(examplePaths, path)
 			return nil
 		}


test/examples_test.go
```diff
@@ -181,6 +181,18 @@ func getExamplePaths(t *testing.T, dir string) []string {
 			return filepath.SkipDir
 		}
 		if info.IsDir() == false && filepath.Ext(info.Name()) == ".yaml" {
+			// Ignore test matching the regexp in the TEST_EXAMPLES_IGNORES
+			// environement variable.
```
dlorenc, line 185: Is there a way to do this with the "-run" flag to go test instead? Or would that be too difficult since you would need to basically specify a negative regular expression.

chmouel: Tha'ts right, that would be quite difficult and probably error prone, I would need to list all tests, filter them out and feed them back to go_test_e2e

dlorenc: Makes sense.

test/examples_test.go
```diff
@@ -181,6 +181,18 @@ func getExamplePaths(t *testing.T, dir string) []string {
...
+			val, ok := os.LookupEnv("TEST_EXAMPLES_IGNORES")
+			if ok {
+				re := regexp.MustCompile(val)
```
dlorenc, line 188: nit: it looks like this val check and re compilation could be moved up out of the filepath.Walk.

chmouel: I think if we move out of the `filepath.Walk` we are only getting the directories ? And we need to have that condition to filter only the yaml files : 

```go
if info.IsDir() == false && filepath.Ext(info.Name()) == ".yaml" {
```

dlorenc: Ah right! My mistake. Sorry!

test/examples_test.go
dlorenc: Typo: environment

test/examples_test.go
```diff
@@ -181,6 +181,18 @@ func getExamplePaths(t *testing.T, dir string) []string {
...
+				submatch := re.FindSubmatch([]byte(path))
+				if submatch != nil {
+					t.Logf("Skipping test %s", path)
```
dlorenc, line 191: Does t.Skip work here?

vdemeester: @dlorenc it would skip all the tests I think :upside_down_face:

vdemeester (approved):
/meow

PR #3073, opened 2020-08-07
Fix an error formatting in taskrun reconciler

NavidZ:
# Changes
Fix an error formatting in taskrun reconciler

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -376,7 +376,7 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1beta1.TaskRun,
 		pod, err = c.createPod(ctx, tr, rtr)
 		if err != nil {
 			newErr := c.handlePodCreationError(ctx, tr, err)
-			logger.Error("Failed to create task run pod for task %q: %v", tr.Name, newErr)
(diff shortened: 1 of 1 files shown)


pkg/reconciler/taskrun/taskrun.go
```diff
@@ -376,7 +376,7 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1beta1.TaskRun,
 		pod, err = c.createPod(ctx, tr, rtr)
 		if err != nil {
 			newErr := c.handlePodCreationError(ctx, tr, err)
-			logger.Error("Failed to create task run pod for task %q: %v", tr.Name, newErr)
+			logger.Errorf("Failed to create task run pod for task %q: %v", tr.Name, newErr)
```
danielhelfand, line 379: Another issue with this appears to be that the error is stating that the added name is for the task, but the actual value is `tr.Name`. Should it not be the following:

```suggestion
			logger.Errorf("Failed to create task run pod for taskrun %q: %v", tr.Name, newErr)
```

NavidZ: You are right. Fixed!

PR #3078, opened 2020-08-07
Actually stop trying to time out finished Runs ⏰

### LINKED ISSUE #3047: "Done" channel (for tracking timeouts) never fires for Runs
# Expected Behavior

When we start a TaskRun or a PipelineRun, we start a go routine that will wait until either the Run completes, the controller exits, or a timeout occurs. When a timeout occurs, the Run is triggered for re-reconciliation and at that point the controller would notice it has run for too long and stop it.

Specific to this bug: when the Run completes, the controller should use the "done" channel to indicate to the go routine that it can stop. *poignant pause*

# Actual Behavior

In https://github.com/tektoncd/pipeline/commit/10b64272fd2fabbf022acd634d4ddb45866528d2 I got really enthusiastic about making sure even our reads were threadsafe and so I thought I would be clever and, instead of accessing attributes of a PipelineRun or TaskRun in a go routine, use a value that wouldn't change - specifically the address.

But unfortunately the address can change between reconciles 🤣   - so what happens is:

1. The Run is created, the go routine is started, the handler stores as a key for the "done" channel the address of the Run
2. Things continue
3. At some point the Run completes: this triggers a new reconcile loop, a Run object is instantiated: A NEW RUN OBJECT WITH A NEW ADDRESS MIND YOU
4. The controller calls "Release" to indicate the Run is done but there is nothing in the handler's map with the key of this new Run object
5. Sometime later, the go routine (WHICH IS STILL RUNNING) times out, it triggers a re-reconcile
6. This additional reconcile for absolutely no good reason should (hopefully) do nothing b/c the Run has already completed

Look through the logs of a running controller!

While following up on https://github.com/tektoncd/pipeline/pull/1335#issuecomment-666874624 I decided to look at the logs in our dogfooding cluster (shout out to the dogfooding cluster being super useful! + @afrittoli  ❤️ ) to see if the logs were ever interleaved.

I was shocked to see 0 logs indicating that a Run had completed successfully and the done channel had fired - instead what I saw was that every single Run had timed out. But when I looked at the statuses via tkn, none of them had timed out XD

# Additional Info

While this sounds dire, I think this timeout handling logic still works, it just has these problems:

* We track an extra goroutine for each Run, each of which is just waiting on a timer
* We do an extra reconcile (long) after a Run has finished
* The log messages for each Run make it look like every single Run times out <-- 🤣 

Tekton Pipeline version: This affects all versions from 0.7.0 to now - hopefully 0.16.0 will have a fix? XD

bobcatfish:
# Changes

In 10b6427 I got really enthusiastic about making sure even our
reads were threadsafe and so I thought I would be clever and,
instead of accessing attributes of a PipelineRun or TaskRun in
a go routine, use a value that wouldn't change - specifically the address.

But the address will change between reconcile loops, because the
reconcile logic will create a new instance of the Run object every time!
🤦‍♀️

Fortunately this doesn't cause any serious problems, it just makes
things slightly less efficient: for every Run you start, a go routine
will remain open until the timeout occurs, and when it fires, it will be
reconciled an extra time, even if it has completed. (In fact keeping
this functionality completed and dropping the "done" map might be a
reasonable option!)

With this change, we now return to using the namespace + name as a key
in the map that tracks the done channels; we pass these by value so that
reads will be threadsafe.

Instead of fixing this separately for the TaskRun and PipelineRun
functions, I've collapsed these and the callback into one. Each handler
instantiates its own timeout handler so there is no reason for the
timeout handler to have special knowledge of one vs the other.

Fixes #3047

### Test

I tried several different approaches to add a test case that would
reveal the underlying problem but I now feel like it's more hassle than
it's worth. Approaches:

1. instantiate the controller in the reconciler tests with a custom
   timeout handler that has been overridden to use a custom logger,
   so we can check for the log indicating the timeout handler completed
2. Similar to (1) but instead of checking logs, just pass in a custom
   done channel and wait for it to close

Both 1 + 2 require changing the way that NewController works, i.e. the
way we always instantiate controllers. I tried working around this by
taking the same approach as `TestHandlePodCreationError` and
instantiating my own Reconciler but it a) wasn't instantiated properly
no matter what I tried (trying to use it created panics) and b) had a
confusingly different interface, exposing ReconcileKind instead of
Reconcile

I tried some other approaches but these went nowhere either; I don't
think it's worth adding a test to cover this, but if folks feel strongly
I don't mind opening an issue at least to continue to explore it? I feel
that this bug is one that is very specific to the implementation and I'm
not sure how valuable a test that covers it would be. If we do pursue
it, we might want to do it at the level of an end to end test that
actually checks the logs from a real running controller.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
When a TaskRun or PipelineRun completes, the go routine waiting for it to timeout will now stop (as it was designed to do!) instead of always re-reon
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
@@ -17,7 +17,6 @@ limitations under the License.
 package v1beta1
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/tektoncd/pipeline/pkg/apis/config"
@@ -25,6 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/pkg/apis"
 	duckv1beta1 "knative.dev/pkg/apis/duck/v1beta1"
 )
@@ -98,10 +98,9 @@ func (pr *PipelineRun) IsCancelled() bool {
 	return pr.Spec.Status == PipelineRunSpecStatusCancelled
 }
 
-// GetRunKey return the pipelinerun key for timeout handler map
-func (pr *PipelineRun) GetRunKey() string {
-	// The address of the pointer is a threadsafe unique identifier for the pipelinerun
-	return fmt.Sprintf("%s/%p", pipeline.PipelineRunControllerName, pr)
+// GetNamespacedName returns a k8s namespaced name that identifies this PipelineRun
+func (pr *PipelineRun) GetNamespacedName() types.NamespacedName {
+	return types.NamespacedName{Namespace: pr.Namespace, Name: pr.Name}
 }
 
 // IsTimedOut returns true if a pipelinerun has exceeded its spec.Timeout based on its status.Timeout
@@ -231,7 +230,7 @@ const (
 	PipelineRunReasonCancelled PipelineRunReason = "Cancelled"
 	// PipelineRunReasonTimedOut is the reason set when the PipelineRun has timed out
 	PipelineRunReasonTimedOut PipelineRunReason = "PipelineRunTimeout"
-	// ReasonStopping indicates that no new Tasks will be scheduled by the controller, and the
+	// PipelineRunReasonStopping indicates that no new Tasks will be scheduled by the controller, and the
 	// pipeline will stop once all running tasks complete their work
 	PipelineRunReasonStopping PipelineRunReason = "PipelineRunStopping"
 )
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_types_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_types_test.go
@@ -17,12 +17,10 @@ limitations under the License.
 package v1beta1_test
 
 import (
-	"fmt"
 	"testing"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
-	tb "github.com/tektoncd/pipeline/internal/builder/v1beta1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/pipeline/test/diff"
 	corev1 "k8s.io/api/core/v1"
@@ -167,11 +165,12 @@ func TestPipelineRunHasVolumeClaimTemplate(t *testing.T) {
 	}
 }
 
-func TestPipelineRunKey(t *testing.T) {
-	pr := tb.PipelineRun("prunname")
-	expectedKey := fmt.Sprintf("PipelineRun/%p", pr)
-	if pr.GetRunKey() != expectedKey {
-		t.Fatalf("Expected taskrun key to be %s but got %s", expectedKey, pr.GetRunKey())
+func TestGetNamespacedName(t *testing.T) {
+	pr := &v1beta1.PipelineRun{ObjectMeta: metav1.ObjectMeta{Namespace: "foo", Name: "prunname"}}
+	n := pr.GetNamespacedName()
+	expected := "foo/prunname"
+	if n.String() != expected {
+		t.Fatalf("Expected name to be %s but got %s", expected, n.String())
 	}
 }
 
--- a/pkg/apis/pipeline/v1beta1/taskrun_types_test.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_types_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package v1beta1_test
 
 import (
-	"fmt"
 	"testing"
 	"time"
 
@@ -172,14 +171,11 @@ func TestTaskRunHasVolumeClaimTemplate(t *testing.T) {
 }
 
 func TestTaskRunKey(t *testing.T) {
-	tr := &v1beta1.TaskRun{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "taskrunname",
-		},
-	}
-	expectedKey := fmt.Sprintf("TaskRun/%p", tr)
-	if tr.GetRunKey() != expectedKey {
-		t.Fatalf("Expected taskrun key to be %s but got %s", expectedKey, tr.GetRunKey())
+	tr := &v1beta1.TaskRun{ObjectMeta: metav1.ObjectMeta{Namespace: "foo", Name: "trunname"}}
+	n := tr.GetNamespacedName()
+	expected := "foo/trunname"
+	if n.String() != expected {
+		t.Fatalf("Expected name to be %s but got %s", expected, n.String())
 	}
 }
 
--- a/pkg/apis/pipeline/v1beta1/taskrun_types.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_types.go
@@ -25,6 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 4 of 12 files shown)


bobcatfish:
image

coverage dropped a bit here but note I'm next planning to resurrect #3031 and significantly increase coverage!

bobcatfish:
```
pkg/timeout/handler.go:40:2: `defaultFunc` is unused (deadcode)
	defaultFunc        = func(i interface{}) {}
```

oooo nice thanks linter 🙏

bobcatfish:
image

well that seems apt :D

bobcatfish:
Weird that the v1alpha1 test failed and v1beta1 succeeded 🤔

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -98,10 +98,9 @@ func (pr *PipelineRun) IsCancelled() bool {
 	return pr.Spec.Status == PipelineRunSpecStatusCancelled
 }
 
-// GetRunKey return the pipelinerun key for timeout handler map
-func (pr *PipelineRun) GetRunKey() string {
-	// The address of the pointer is a threadsafe unique identifier for the pipelinerun
-	return fmt.Sprintf("%s/%p", pipeline.PipelineRunControllerName, pr)
+// GetNamespacedName returns a k8s namespaced name that identifies this PipelineRun
+func (pr *PipelineRun) GetNamespacedName() types.NamespacedName {
+	return types.NamespacedName{Namespace: pr.Namespace, Name: pr.Name}
```
vdemeester, line 103: Should we include the `kind` in there too ? We can't have 2 pipelinerun with the same name, but nothing prevent us from having a pipelinerun and a taskrun with the same name (which could be a problem then, as the key would be similar)

bobcatfish: we could - in the current implementation we have a timeout handler per controller, i.e. one for taskruns and one for pipelineruns. maybe it's better to be safe, but if we go this route we can use the existing NamespacedName type - maybe im just lazy :D

vdemeester: @bobcatfish oh I see, right, then I think it's ok for now :wink:

bobcatfish:
i dont want to introduce new flakiness but i do want to see if this happens consistently, havent been able to repro yet:

bobcatfish:
I looked into that test failure more - I'm pretty disturbed a test _directly related_ to the changes im making _just to happened to flake_ but it kinda seems like that happened.

The test that fails runs a pipeline with a task that is expected to succeed and a task that is expected to time out; the task that failed was actually the one that is supposed to succeed with:

```
        ---
        kind: TaskRun
        metadata:
          annotations:
            pipeline.tekton.dev/release: devel
          creationTimestamp: "2020-08-07T21:38:02Z"
          generation: 1
          labels:
            app.kubernetes.io/managed-by: tekton-pipelines
            tekton.dev/pipeline: pipelinetasktimeout
            tekton.dev/pipelineRun: prtasktimeout
            tekton.dev/pipelineTask: pipelinetask1
            tekton.dev/task: success
          name: prtasktimeout-pipelinetask1-ljfbw
          namespace: arendelle-fml4b
          ownerReferences:
          - apiVersion: tekton.dev/v1beta1
            blockOwnerDeletion: true
            controller: true
            kind: PipelineRun
            name: prtasktimeout
            uid: a6f55b45-9020-49aa-8129-cc3ce44fc9e9
          resourceVersion: "4998"
          selfLink: /apis/tekton.dev/v1alpha1/namespaces/arendelle-fml4b/taskruns/prtasktimeout-pipelinetask1-ljfbw
          uid: 237e86b7-d275-4a73-953b-81ada233c195
        spec:
          resources: {}
          serviceAccountName: ""
          taskRef:
            kind: Task
            name: success
          timeout: 1m0s
        status:
          completionTime: "2020-08-07T21:39:22Z"
          conditions:
          - lastTransitionTime: "2020-08-07T21:39:22Z"
            message: TaskRun "prtasktimeout-pipelinetask1-ljfbw" failed to finish within "1m0s"
            reason: TaskRunTimeout
            status: "False"
            type: Succeeded
          podName: prtasktimeout-pipelinetask1-ljfbw-pod-nxbns
          startTime: "2020-08-07T21:38:10Z"
          steps:
          - container: step-unnamed-0
            name: unnamed-0
            waiting:
              reason: PodInitializing
          taskSpec:
            steps:
            - args:
              - 1s
              command:
              - sleep
              image: busybox
              name: ""
              resources: {}
```

The task is supposed to sleep for 1s (with a 1 min timeout) and then succeed, but the pod was stuck "PodInitializing" for reasons now lost in the sands of time.

So it doesn't SEEM like this is a real problem 🤞 🤞 🤞

bobcatfish:
> I'm pretty disturbed a test directly related to the changes im making just to happened to flake but it kinda seems like that happened.

Thinking about it further, i think if an e2e test is gonna flake, the probability that a timeout related test would flake might actually be pretty high - there are a couple of them, run for both v1alpha1 and v1beta1

pkg/apis/pipeline/v1beta1/pipelinerun_types_test.go
```diff
@@ -167,11 +166,12 @@ func TestPipelineRunHasVolumeClaimTemplate(t *testing.T) {
 	}
 }
 
-func TestPipelineRunKey(t *testing.T) {
-	pr := tb.PipelineRun("prunname")
-	expectedKey := fmt.Sprintf("PipelineRun/%p", pr)
-	if pr.GetRunKey() != expectedKey {
-		t.Fatalf("Expected taskrun key to be %s but got %s", expectedKey, pr.GetRunKey())
+func TestGetNamespacedName(t *testing.T) {
+	pr := tb.PipelineRun("prunname", tb.PipelineRunNamespace("foo"))
```
ImJasonH, line 170:
```go
pr := &v1beta1.PipelineRun{
  ObjectMeta: metav1.ObjectMeta{Namespace: "foo", Name: "prunname"},
}
```

Please 🙏

pkg/apis/pipeline/v1beta1/taskrun_types_test.go
```diff
@@ -172,14 +172,11 @@ func TestTaskRunHasVolumeClaimTemplate(t *testing.T) {
...
 func TestTaskRunKey(t *testing.T) {
-	tr := &v1beta1.TaskRun{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: "taskrunname",
-		},
-	}
-	expectedKey := fmt.Sprintf("TaskRun/%p", tr)
-	if tr.GetRunKey() != expectedKey {
-		t.Fatalf("Expected taskrun key to be %s but got %s", expectedKey, tr.GetRunKey())
+	tr := tb.TaskRun("trunname", tb.TaskRunNamespace("foo"))
```
ImJasonH, line 175:
```go
pr := &v1beta1.TaskRun{
  ObjectMeta: metav1.ObjectMeta{Namespace: "foo", Name: "prunname"},
}
```

🙏

ImJasonH (commented):
I think there's a chance we can simplify a lot of this by using `impl.EnqueueAfter` to enqueue a reconciliation in the future when the timeout should be elapsed. But this fix lgtm even without that cleanup.

bobcatfish:
> I think there's a chance we can simplify a lot of this by using impl.EnqueueAfter to enqueue a reconciliation in the future when the timeout should be elapsed.

iiiiinteresting - i can try that out as part of #2905. i guess the only downside is that we get 1 extra reconcile for every Run? (which ironically is the current state, which this PR is undoing) seems worth it to me tho!

pkg/apis/pipeline/v1beta1/pipelinerun_types_test.go
bobcatfish: kk i can update as per https://github.com/tektoncd/pipeline/issues/3178

thanks @ImJasonH !

ImJasonH (commented):
/lgtm

Issue #3076, opened 2020-08-07
Create a separate location for YAML e2e tests


sbwsg:
### Feature request

Create a directory for "e2e tests" that are just YAML files which exercise specific behaviours of Tekton Pipelines. These differ from the YAML files in the `examples` directory because they aren't intended to document _correct_ usage but rather to exercise edge cases.

### Use case

As a dev I often want to create some YAML files that exercise a bug fix or new feature I'm working on. Some of these YAML files are suitable as user-facing examples because they demonstrate a feature's usecases or show a specific approach to a problem. Other YAMLs are less suitable as examples: they might make the Tekton controller behave in a certain way that is only useful to test correctness of implementation. It doesn't make a ton of sense for these groups of YAMLs to be in a shared directory because:

1. The kinds of comments and description useful in each group likely has a different audience.
2. The YAMLs testing controller behaviour may include bad practices or incorrect usage to ensure these cases are handled properly.
3. We may at some point want test-only YAMLs that fail as well as succeed. These could be checked using something like "golden files" where alongside a test YAML is also a description of the expected result. The test runner submits the test YAML and then compares the resulting resource in k8s with the expected description.

I've already created some of these test-only YAMLs but I've ended up putting them in the `examples/v1beta1` directory because it's the only place we have right now that will be executed as part of our integration suite. I'd like to introduce a directory like `test/e2e-yaml` that I can move those YAMLs into. The test runner will need to be updated to also run these YAMLs as part of our PRs.

sbwsg:
This was discussed as part of the productivity WG. It was mentioned there that we may conceivably be able to translate some of our existing e2e tests written in Go into this YAML + golden files format.  Though that's not a requirement as part of this issue.

PR #3087, opened 2020-08-11
Step timeout

### LINKED ISSUE #1690: Proposal: step timeout
It could be nice to augment the `Step` type with a timeout duration, after which time the step fails with some unique status/message.

To support this we could pass the timeout value to the entrypoint binary, which would start a timer when step execution starts, and stop executing and raise a timeout error if execution doesn't finish in time.

If this is interesting to people it should be a pretty good first issue for a new contributor.

/kind feature

Peaorl:
# Changes

This feature allows a Task author to specify a Step timeout in the TaskSpec.

An example use case is when a Task author would like to execute a Step for setting up an execution environment. One may expect this Step to execute within a few seconds. If the execution time takes longer than expected one may rather want to fail fast than wait for the TaskRun timeout to abort the TaskRun (example by @ImJasonH).

Closes #1690
Corresponding [TEP](https://github.com/tektoncd/community/blob/master/teps/0014-step-timeout.md)

Major thanks to @sbwsg for his guidance.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Task authors can now specify a timeout for a Step in the TaskSpec.
```

### DIFF
--- a/cmd/entrypoint/main.go
+++ b/cmd/entrypoint/main.go
@@ -41,6 +41,7 @@ var (
 	terminationPath     = flag.String("termination_path", "/tekton/termination", "If specified, file to write upon termination")
 	results             = flag.String("results", "", "If specified, list of file names that might contain task results")
 	waitPollingInterval = time.Second
+	timeout             = flag.Duration("timeout", time.Duration(0), "If specified, sets timeout for step")
 )
 
 func cp(src, dst string) error {
@@ -103,6 +104,7 @@ func main() {
 		Runner:          &realRunner{},
 		PostWriter:      &realPostWriter{},
 		Results:         strings.Split(*results, ","),
+		Timeout:         timeout,
 	}
 
 	// Copy any creds injected by the controller into the $HOME directory of the current
--- a/cmd/entrypoint/runner.go
+++ b/cmd/entrypoint/runner.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -19,7 +20,7 @@ type realRunner struct {
 
 var _ entrypoint.Runner = (*realRunner)(nil)
 
-func (rr *realRunner) Run(args ...string) error {
+func (rr *realRunner) Run(ctx context.Context, args ...string) error {
 	if len(args) == 0 {
 		return nil
 	}
@@ -33,7 +34,7 @@ func (rr *realRunner) Run(args ...string) error {
 	signal.Notify(rr.signals)
 	defer signal.Reset()
 
-	cmd := exec.Command(name, args...)
+	cmd := exec.CommandContext(ctx, name, args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	// dedicated PID group used to forward signals to
@@ -42,6 +43,9 @@ func (rr *realRunner) Run(args ...string) error {
 
 	// Start defined command
 	if err := cmd.Start(); err != nil {
+		if ctx.Err() == context.DeadlineExceeded {
+			return context.DeadlineExceeded
+		}
 		return err
 	}
 
@@ -57,6 +61,9 @@ func (rr *realRunner) Run(args ...string) error {
 
 	// Wait for command to exit
 	if err := cmd.Wait(); err != nil {
+		if ctx.Err() == context.DeadlineExceeded {
+			return context.DeadlineExceeded
+		}
 		return err
 	}
 
--- a/cmd/entrypoint/runner_test.go
+++ b/cmd/entrypoint/runner_test.go
@@ -1,9 +1,11 @@
 package main
 
 import (
+	"context"
 	"os"
 	"syscall"
 	"testing"
+	"time"
 )
 
 // TestRealRunnerSignalForwarding will artificially put an interrupt signal (SIGINT) in the rr.signals chan.
@@ -14,9 +16,24 @@ func TestRealRunnerSignalForwarding(t *testing.T) {
 	rr := realRunner{}
 	rr.signals = make(chan os.Signal, 1)
 	rr.signals <- syscall.SIGINT
-	if err := rr.Run("sleep", "3600"); err.Error() == "signal: interrupt" {
+	if err := rr.Run(context.Background(), "sleep", "3600"); err.Error() == "signal: interrupt" {
 		t.Logf("SIGINT forwarded to Entrypoint")
 	} else {
 		t.Fatalf("Unexpected error received: %v", err)
+	}
+}
+
+// TestRealRunnerTimeout tests whether cmd is killed after a millisecond even though it's supposed to sleep for 10 milliseconds.
+func TestRealRunnerTimeout(t *testing.T) {
+	rr := realRunner{}
+	timeout := time.Millisecond
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	if err := rr.Run(ctx, "sleep", "0.01"); err != nil {
+		if err != context.DeadlineExceeded {
+			t.Fatalf("unexpected error received: %v", err)
+		}
+	} else {
+		t.Fatalf("step didn't timeout")
 	}
 }
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -12,6 +12,7 @@ weight: 1
   - [Defining `Steps`](#defining-steps)
     - [Reserved directories](#reserved-directories)
     - [Running scripts within `Steps`](#running-scripts-within-steps)
+    - [Specifying a timeout](#specifying-a-timeout)
   - [Specifying `Parameters`](#specifying-parameters)
   - [Specifying `Resources`](#specifying-resources)
   - [Specifying `Workspaces`](#specifying-workspaces)
@@ -241,7 +242,27 @@ steps:
     #!/usr/bin/env bash
     /bin/my-binary
 ```
+#### Specifying a timeout
 
+A `Step` can specify a `timeout` field.
+If the `Step` execution time exceeds the specified timeout, the `Step` kills
+its running process and any subsequent `Steps` in the `TaskRun` will not be
+executed. The `TaskRun` is placed into a `Failed` condition.  An accompanying log
+describing which `Step` timed out is written as the `Failed` condition's message.
+
+The timeout specification follows the duration format as specified in the [Go time package](https://golang.org/pkg/time/#ParseDuration) (e.g. 1s or 1ms).
+
+The example `Step` below is supposed to sleep for 60 seconds but will be canceled by the specified 5 second timeout.
+```yaml
+steps:
+  - name: sleep-then-timeout
+    image: ubuntu
+    script: | 
+      #!/usr/bin/env bash
+      echo "I am supposed to sleep for 60 seconds!"
+      sleep 60
+    timeout: 5s
+``` 
 ### Specifying `Parameters`
 
 You can specify parameters, such as compilation flags or artifact names, that you want to supply to the `Task` at execution time.
--- a/pkg/apis/pipeline/v1beta1/task_types.go
+++ b/pkg/apis/pipeline/v1beta1/task_types.go
@@ -125,10 +125,12 @@ type Step struct {
 	//
 	// If Script is not empty, the Step cannot have an Command or Args.
 	Script string `json:"script,omitempty"`
+	// Timeout is the time after which the step times out. Defaults to never.
+	// Refer to Go's ParseDuration documentation for expected format: https://golang.org/pkg/time/#ParseDuration
+	Timeout *metav1.Duration `json:"timeout,omitempty"`
 }
 
-// Sidecar embeds the Container type, which allows it to include fields not
-// provided by Container.
+// Sidecar has nearly the same data structure as Step, consisting of a Container and an optional Script, but does not have the ability to timeout.
 type Sidecar struct {
 	corev1.Container `json:",inline"`
 
--- a/pkg/entrypoint/entrypointer.go
+++ b/pkg/entrypoint/entrypointer.go
@@ -17,6 +17,7 @@ limitations under the License.
 package entrypoint
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -63,6 +64,8 @@ type Entrypointer struct {
 
 	// Results is the set of files that might contain task results
 	Results []string
+	// Timeout is an optional user-specified duration within which the Step must complete
+	Timeout *time.Duration
 }
 
 // Waiter encapsulates waiting for files to exist.
@@ -73,7 +76,7 @@ type Waiter interface {
 
 // Runner encapsulates running commands.
 type Runner interface {
-	Run(args ...string) error
+	Run(ctx context.Context, args ...string) error
 }
 
 // PostWriter encapsulates writing a file when complete.
@@ -106,7 +109,6 @@ func (e Entrypointer) Go() error {
 				Value:      time.Now().Format(timeFormat),
 				ResultType: v1beta1.InternalTektonResultType,
 			})
-
 			return err
 		}
 	}
@@ -114,13 +116,34 @@ func (e Entrypointer) Go() error {
 	if e.Entrypoint != "" {
 		e.Args = append([]string{e.Entrypoint}, e.Args...)
 	}
+
 	output = append(output, v1beta1.PipelineResourceResult{
 		Key:        "StartedAt",
 		Value:      time.Now().Format(timeFormat),
 		ResultType: v1beta1.InternalTektonResultType,
 	})
 
-	err := e.Runner.Run(e.Args...)
+	var err error
+	if e.Timeout != nil && *e.Timeout < time.Duration(0) {
+		err = fmt.Errorf("negative timeout specified")
+	}
+
+	if err == nil {
+		ctx := context.Background()
+		var cancel context.CancelFunc
+		if e.Timeout != nil && *e.Timeout != time.Duration(0) {
+			ctx, cancel = context.WithTimeout(ctx, *e.Timeout)
+			defer cancel()
+		}
+		err = e.Runner.Run(ctx, e.Args...)
+		if err == context.DeadlineExceeded {
+			output = append(output, v1beta1.PipelineResourceResult{
+				Key:        "Reason",
+				Value:      "TimeoutExceeded",
+				ResultType: v1beta1.InternalTektonResultType,
+			})
+		}
+	}
 
 	// Write the post file *no matter what*
 	e.WritePostFile(e.PostFile, err)
--- a/pkg/pod/entrypoint.go
+++ b/pkg/pod/entrypoint.go
@@ -85,9 +85,8 @@ var (
 // Containers must have Command specified; if the user didn't specify a
 // command, we must have fetched the image's ENTRYPOINT before calling this
 // method, using entrypoint_lookup.go.
-//
-// TODO(#1605): Also use entrypoint injection to order sidecar start/stop.
-func orderContainers(entrypointImage string, extraEntrypointArgs []string, steps []corev1.Container, results []v1beta1.TaskResult) (corev1.Container, []corev1.Container, error) {
+// Additionally, Step timeouts are added as entrypoint flag.
+func orderContainers(entrypointImage string, commonExtraEntrypointArgs []string, steps []corev1.Container, taskSpec *v1beta1.TaskSpec) (corev1.Container, []corev1.Container, error) {
 	initContainer := corev1.Container{
 		Name:  "place-tools",
 		Image: entrypointImage,
@@ -121,8 +120,13 @@ func orderContainers(entrypointImage string, extraEntrypointArgs []string, steps
 				"-termination_path", terminationPath,
 			}
 		}
-		argsForEntrypoint = append(argsForEntrypoint, extraEntrypointArgs...)
-		argsForEntrypoint = append(argsForEntrypoint, resultArgument(steps, results)...)
+		argsForEntrypoint = append(argsForEntrypoint, commonExtraEntrypointArgs...)
+		if taskSpec != nil {
+			if taskSpec.Steps != nil && len(taskSpec.Steps) >= i+1 && taskSpec.Steps[i].Timeout != nil {
+				argsForEntrypoint = append(argsForEntrypoint, "-timeout", taskSpec.Steps[i].Timeout.Duration.String())
+			}
+			argsForEntrypoint = append(argsForEntrypoint, resultArgument(steps, taskSpec.Results)...)
+		}
 
 		cmd, args := s.Command, s.Args
 		if len(cmd) == 0 {
@@ -239,6 +243,6 @@ func isContainerSidecar(name string) bool { return strings.HasPrefix(name, sidec
 // trimStepPrefix returns the container name, stripped of its step prefix.
 func trimStepPrefix(name string) string { return strings.TrimPrefix(name, stepPrefix) }
 
-// trimSidecarPrefix returns the container name, stripped of its sidecar
+// TrimSidecarPrefix returns the container name, stripped of its sidecar
 // prefix.
 func TrimSidecarPrefix(name string) string { return strings.TrimPrefix(name, sidecarPrefix) }
--- a/pkg/pod/entrypoint_test.go
+++ b/pkg/pod/entrypoint_test.go
@@ -108,13 +108,15 @@ func TestOrderContainers(t *testing.T) {
 }
 
 func TestEntryPointResults(t *testing.T) {
-	results := []v1beta1.TaskResult{{
-		Name:        "sum",
-		Description: "This is the sum result of the task",
-	}, {
-		Name:        "sub",
-		Description: "This is the sub result of the task",
-	}}
+	taskSpec := v1beta1.TaskSpec{
+		Results: []v1beta1.TaskResult{{
+			Name:        "sum",
+			Description: "This is the sum result of the task",
+		}, {
+			Name:        "sub",
+			Description: "This is the sub result of the task",
+		}},
+	}
 
 	steps := []corev1.Container{{
 		Image:   "step-1",
@@ -172,7 +174,7 @@ func TestEntryPointResults(t *testing.T) {
 		VolumeMounts:           []corev1.VolumeMount{toolsMount},
 		TerminationMessagePath: "/tekton/termination",
 	}}
-	_, got, err := orderContainers(images.EntrypointImage, []string{}, steps, results)
+	_, got, err := orderContainers(images.EntrypointImage, []string{}, steps, &taskSpec)
 	if err != nil {
 		t.Fatalf("orderContainers: %v", err)
 	}
@@ -182,13 +184,15 @@ func TestEntryPointResults(t *testing.T) {
 }
 
 func TestEntryPointResultsSingleStep(t *testing.T) {
-	results := []v1beta1.TaskResult{{
-		Name:        "sum",
-		Description: "This is the sum result of the task",
-	}, {
-		Name:        "sub",
-		Description: "This is the sub result of the task",
-	}}
+	taskSpec := v1beta1.TaskSpec{
+		Results: []v1beta1.TaskResult{{
+			Name:        "sum",
+			Description: "This is the sum result of the task",
+		}, {
+			Name:        "sub",
+			Description: "This is the sub result of the task",
+		}},
+	}
 
 	steps := []corev1.Container{{
 		Image:   "step-1",
@@ -210,7 +214,7 @@ func TestEntryPointResultsSingleStep(t *testing.T) {
 		VolumeMounts:           []corev1.VolumeMount{toolsMount, downwardMount},
 		TerminationMessagePath: "/tekton/termination",
 	}}
-	_, got, err := orderContainers(images.EntrypointImage, []string{}, steps, results)
+	_, got, err := orderContainers(images.EntrypointImage, []string{}, steps, &taskSpec)
 	if err != nil {
 		t.Fatalf("orderContainers: %v", err)
 	}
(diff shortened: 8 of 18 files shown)


cmd/entrypoint/runner_test.go
```diff
@@ -14,9 +16,23 @@ func TestRealRunnerSignalForwarding(t *testing.T) {
...
 }
+
+// TestRealRunnerTimeout tests whether cmd is killed after time.Second even though it's supposed to sleep for an hour.
+func TestRealRunnerTimeout(t *testing.T) {
+	rr := realRunner{}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	// For seconds and higher time units, err.Error() returns "signal: killed".
+	// For microseconds and lower time units, err.Error() returns "context deadline exceeded".
+	if err := rr.Run(ctx, "sleep", "3600"); err.Error() == "signal: killed" || err.Error() == "context deadline exceeded" {
```
ImJasonH, line 33: Could we change this to "sleep 10" or something similarly short? This would cause it to fail faster in case timeout logic is broken by a future change.

docs/tasks.md
```diff
@@ -241,7 +241,24 @@ steps:
     #!/usr/bin/env bash
     /bin/my-binary
 ```
+#### Specifying a timeout
 
+A step can specify a `timeout` field. If the `Step` execution time exceeds the specified timeout, the `Step` and consequently the entire `TaskRun` is canceled.
+An accompanying error message is output under `status.conditions.message`.
+
+The format for a timeout is a duration string as specified in the [Go time package](https://golang.org/src/time/format.go?s=40541:40587#L1364) (e.g. 1s or 1ms).
```
ImJasonH, line 249: Could we link to https://golang.org/pkg/time/#ParseDuration instead? That has roughly the same information, and is less likely to fall out of sync than a link to specific lines of code in the Go stdlib.

pkg/pod/entrypoint.go
```diff
@@ -86,7 +86,7 @@ var (
 // method, using entrypoint_lookup.go.
 //
 // TODO(#1605): Also use entrypoint injection to order sidecar start/stop.
```
ImJasonH, line 88: nit: this TODO seems to be out of date and could be deleted.

pkg/pod/entrypoint.go
```diff
@@ -120,6 +120,7 @@ func orderContainers(entrypointImage string, extraEntrypointArgs []string, steps
 		}
 		argsForEntrypoint = append(argsForEntrypoint, extraEntrypointArgs...)
 		argsForEntrypoint = append(argsForEntrypoint, resultArgument(steps, results)...)
+		argsForEntrypoint = append(argsForEntrypoint, "-timeout", stepTimeouts[i])
```
ImJasonH, line 123: It might be easier to only append the arg if it's value is !=0 -- in particular, that would remove the need for a lot of (all of?) the changes in `entrypoint_test.go` below.

pkg/pod/pod.go
```diff
@@ -139,9 +139,12 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec
 		return nil, err
 	}
 
+	// Create Step timeout arguments
+	stepTimeoutArgs := returnTimeoutArgs(taskSpec.Steps)
+
 	// Rewrite steps with entrypoint binary. Append the entrypoint init
 	// container to place the entrypoint binary.
-	entrypointInit, stepContainers, err := orderContainers(b.Images.EntrypointImage, credEntrypointArgs, stepContainers, taskSpec.Results)
+	entrypointInit, stepContainers, err := orderContainers(b.Images.EntrypointImage, credEntrypointArgs, stepContainers, stepTimeoutArgs, taskSpec.Results)
```
ImJasonH, line 147: The args to `orderContainers` are becoming a bit of a smell. 👃 

Perhaps it should take a `[]Step` instead of `[]Container` and a parallel slice of timeouts?

(Not necessarily required for this change, but something to think about)

pkg/pod/script.go
```diff
@@ -62,13 +62,15 @@ func convertScripts(shellImage string, steps []v1beta1.Step, sidecars []v1beta1.
...
 
 	convertedStepContainers := convertListOfSteps(steps, &placeScriptsInit, &placeScripts, "script")
+
 	// convertListOfSteps operates on overlapping fields across Step and Sidecar, hence a conversion
 	// from Sidecar into Step
 	sideCarSteps := []v1beta1.Step{}
 	for _, step := range sidecars {
 		sidecarStep := v1beta1.Step{
 			step.Container,
 			step.Script,
```
ImJasonH, line 72: This should be:

```
sidecarStep := v1beta1.Step{
  Container: step.Container,
  Script:    step.Script,
}
```

(This makes it clearer what fields these values are being passed to, rather than depending on the field ordering in the struct, which could change subtly in the future.)

pkg/pod/pod_test.go
```diff
@@ -108,6 +108,8 @@ func TestPodBuild(t *testing.T) {
 					"/tekton/tools/0",
 					"-termination_path",
 					"/tekton/termination",
+					"-timeout",
```
ImJasonH, line 111: Looking through these test cases it doesn't seem like there are any that check that a step that specifies `timeout` results in a pod container that specifies the `-timeout` flag with the same value -- all the tests have `"-timeout", ""`. Could you add one, to guard against future breakages?

docs/tasks.md
```diff
@@ -241,7 +241,24 @@ steps:
     #!/usr/bin/env bash
     /bin/my-binary
 ```
+#### Specifying a timeout
```
bobcatfish, line 244: thanks for adding docs! could you add a link to this in the TOC at the top of https://github.com/tektoncd/pipeline/blob/master/docs/tasks.md ?

examples/v1beta1/taskruns/step-timeout.yaml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  taskSpec:
+    steps:
+      - name: wait
+        image: ubuntu
+        script: | 
+          #!/usr/bin/env bash
+          echo "I am supposed to sleep for 60 seconds!"
+          sleep 60
+        timeout: 5s
```
bobcatfish, line 14: examples can't currently fail b/c this will cause the integration tests to fail - also the examples are (currently) meant to be examples of successful tasks + pipelines (might change via https://github.com/tektoncd/pipeline/issues/3076 !)

so this would need to be an end to end test instead

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -1075,6 +1087,8 @@ func TestReconcile(t *testing.T) {
 						"/tekton/tools/0",
 						"-termination_path",
 						"/tekton/termination",
+						"-timeout",
+						"",
```
bobcatfish, line 1091: what do you think about having this be the default and not provide it when it's not needed? (no strong feelings but feels a bit unnecessary)

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -3185,3 +3209,48 @@ func Test_storeTaskSpec(t *testing.T) {
 		t.Fatalf(diff.PrintWantGot(d))
 	}
 }
+
+func TestGetResultsShouldFilterInternalTektonResultType(t *testing.T) {
```
bobcatfish, line 3213: i'm a bit confused by this test case, if this is unrelated to timeouts maybe we can merge it separately?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -3185,3 +3209,48 @@ func Test_storeTaskSpec(t *testing.T) {
...
+		v1beta1.TaskRunResult{
+			Name:  "",
+			Value: "",
+		},
+	}
+	expectedPipelineResourceResults := []v1beta1.PipelineResourceResult{
+		input[2],
+	}
+
+	gotTaskRunResults, gotPipelineResourceResults := getResults(input)
```
bobcatfish, line 3243: i'm also wondering about testing this private method - often testing private methods is a sign that the method under test could be factored differently, e.g. be an exported part of a different package, or that the testing belongs at a different level (e.g. the thing calling getResults)

im usually more inclined to the first option :D

docs/tasks.md
```diff
@@ -241,7 +241,24 @@ steps:
...
+The example `Step` below is supposed to sleep for 60 seconds but will be canceled by the specified 5 second timeout.
+```yaml
+steps:
+  - name: willTimeout
+    image: ubuntu
+    script: | 
+      #!/usr/bin/env bash
+      echo "I am supposed to sleep for 60 seconds!"
+      sleep 60
+    timeout: 5s
```
bobcatfish, line 260: could this talk a bit about what happens when the timeout occurs? e.g. i assume the following steps don't execute, and is there anything in the logs to indicate that a timeout occurred?

cmd/entrypoint/runner_test.go
bobcatfish: the shortest you can get away with plz!

bobcatfish: also: why are there two different possibilities for the error message?

i think it might be nicer to catch this error and return our own, e.g. something like "killed process since deadline %d exceeded"

cmd/entrypoint/main.go
```diff
@@ -78,7 +80,16 @@ func main() {
 		log.Printf("non-fatal error copying credentials: %q", err)
 	}
 
-	if err := e.Go(); err != nil {
+	// Add timeout to context if a non-zero timeout is specified for a step
+	ctx := context.Background()
+	timeoutContext, _ := time.ParseDuration(*timeout)
```
bobcatfish, line 85: is the other return value here an error? we probably want to fail if the specified timeout is invalid

(maybe all of this could also happen in a function which we can unit test?)

vivi2050:
test







At 2020-08-12 23:15:24, "Christie Wilson" <notifications@github.com> wrote:

@bobcatfish commented on this pull request.

Thanks for taking this on!

In docs/tasks.md:

> @@ -241,7 +241,24 @@ steps:
     #!/usr/bin/env bash
     /bin/my-binary
 ```
+#### Specifying a timeout


thanks for adding docs! could you add a link to this in the TOC at the top of https://github.com/tektoncd/pipeline/blob/master/docs/tasks.md ?

In examples/v1beta1/taskruns/step-timeout.yaml:

> @@ -0,0 +1,14 @@
+kind: TaskRun
+metadata:
+  generateName: timeout-test-
+spec:
+  taskSpec:
+    steps:
+      - name: wait
+        image: ubuntu
+        script: | 
+          #!/usr/bin/env bash
+          echo "I am supposed to sleep for 60 seconds!"
+          sleep 60
+        timeout: 5s


examples can't currently fail b/c this will cause the integration tests to fail - also the examples are (currently) meant to be examples of successful tasks + pipelines (might change via #3076 !)

so this would need to be an end to end test instead

In pkg/reconciler/taskrun/taskrun_test.go:

> @@ -1075,6 +1087,8 @@ func TestReconcile(t *testing.T) {
 						"/tekton/tools/0",
 						"-termination_path",
 						"/tekton/termination",
+						"-timeout",
+						"",


what do you think about having this be the default and not provide it when it's not needed? (no strong feelings but feels a bit unnecessary)

In pkg/reconciler/taskrun/taskrun_test.go:

> @@ -3185,3 +3209,48 @@ func Test_storeTaskSpec(t *testing.T) {
 		t.Fatalf(diff.PrintWantGot(d))
 	}
 }
+
+func TestGetResultsShouldFilterInternalTektonResultType(t *testing.T) {


i'm a bit confused by this test case, if this is unrelated to timeouts maybe we can merge it separately?

In pkg/reconciler/taskrun/taskrun_test.go:

> +			Value:      "",
+			ResultType: v1beta1.PipelineResourceResultType,
+		},
+	}
+
+	expectedTaskRunResults := []v1beta1.TaskRunResult{
+		v1beta1.TaskRunResult{
+			Name:  "",
+			Value: "",
+		},
+	}
+	expectedPipelineResourceResults := []v1beta1.PipelineResourceResult{
+		input[2],
+	}
+
+	gotTaskRunResults, gotPipelineResourceResults := getResults(input)


i'm also wondering about testing this private method - often testing private methods is a sign that the method under test could be factored differently, e.g. be an exported part of a different package, or that the testing belongs at a different level (e.g. the thing calling getResults)

im usually more inclined to the first option :D

In docs/tasks.md:

>  
+A step can specify a `timeout` field. If the `Step` execution time exceeds the specified timeout, the `Step` and consequently the entire `TaskRun` is canceled.
+An accompanying error message is output under `status.conditions.message`.
+
+The format for a timeout is a duration string as specified in the [Go time package](https://golang.org/src/time/format.go?s=40541:40587#L1364) (e.g. 1s or 1ms).
+
+The example `Step` below is supposed to sleep for 60 seconds but will be canceled by the specified 5 second timeout.
+```yaml
+steps:
+  - name: willTimeout
+    image: ubuntu
+    script: | 
+      #!/usr/bin/env bash
+      echo "I am supposed to sleep for 60 seconds!"
+      sleep 60
+    timeout: 5s


could this talk a bit about what happens when the timeout occurs? e.g. i assume the following steps don't execute, and is there anything in the logs to indicate that a timeout occurred?

In cmd/entrypoint/main.go:

> @@ -78,7 +80,16 @@ func main() {
 		log.Printf("non-fatal error copying credentials: %q", err)
 	}
 
-	if err := e.Go(); err != nil {
+	// Add timeout to context if a non-zero timeout is specified for a step
+	ctx := context.Background()
+	timeoutContext, _ := time.ParseDuration(*timeout)


is the other return value here an error? we probably want to fail if the specified timeout is invalid

(maybe all of this could also happen in a function which we can unit test?)

In cmd/entrypoint/runner_test.go:

>  		t.Logf("SIGINT forwarded to Entrypoint")
 	} else {
 		t.Fatalf("Unexpected error received: %v", err)
 	}
 }
+
+// TestRealRunnerTimeout tests whether cmd is killed after time.Second even though it's supposed to sleep for an hour.
+func TestRealRunnerTimeout(t *testing.T) {
+	rr := realRunner{}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	// For seconds and higher time units, err.Error() returns "signal: killed".
+	// For microseconds and lower time units, err.Error() returns "context deadline exceeded".
+	if err := rr.Run(ctx, "sleep", "3600"); err.Error() == "signal: killed" || err.Error() == "context deadline exceeded" {


the shortest you can get away with plz!

In cmd/entrypoint/runner_test.go:

>  		t.Logf("SIGINT forwarded to Entrypoint")
 	} else {
 		t.Fatalf("Unexpected error received: %v", err)
 	}
 }
+
+// TestRealRunnerTimeout tests whether cmd is killed after time.Second even though it's supposed to sleep for an hour.
+func TestRealRunnerTimeout(t *testing.T) {
+	rr := realRunner{}
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+	// For seconds and higher time units, err.Error() returns "signal: killed".
+	// For microseconds and lower time units, err.Error() returns "context deadline exceeded".
+	if err := rr.Run(ctx, "sleep", "3600"); err.Error() == "signal: killed" || err.Error() == "context deadline exceeded" {


also: why are there two different possibilities for the error message?

i think it might be nicer to catch this error and return our own, e.g. something like "killed process since deadline %d exceeded"

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub, or unsubscribe.

pkg/reconciler/taskrun/taskrun_test.go
Peaorl: Indeed this test case is not directly related to timeouts but a consequence of the necessity for an `InternalTektonResultType` in order to digest timeout status messages.

Currently `TaskRunResultType`, `PipelineResourceResultType` and `UnknownResultType` exist as a `ResultType` for a `PipelineResourceResult` or `TaskRunResult` (which is a `PipelineResourceResult`). All results that conform to one of these `ResultTypes` are exposed after a taskrun has completed.

The `InternalTektonResultType` adds the ability to process a `PipelineResourceResult` internally without exposing it.
This test confirms that this `ResultType` is filtered from the external `PipelineResourceResults` and `TaskRunResults`.

Would you suggest I create an entirely separate PR that introduces the `InternalTektonResultType` then?

cmd/entrypoint/main.go
Peaorl: Indeed the other return value is an error.

However, the validity of a specified timeout is already checked in [task_validation.go](https://github.com/tektoncd/pipeline/blob/1cfb4d554b94e525f5788ac597a629d0e1a080c2/pkg/apis/pipeline/v1beta1/task_validation.go#L193) which causes the webhook to deny the creation of a taskrun in case an invalid timeout is specified which is why it's not checked here again.

Would you still recommend I check it here again and/or possibly add a unit test?

cmd/entrypoint/runner_test.go
Peaorl: I have changed it so that the process spawned by a command is now supposed to sleep for 10 milliseconds while the execution time deadline is set to 1 milliseconds.

As a sidenote, it seems like the context deadline is accurate up to about 3 milliseconds.
I.e., if the process is supposed to sleep for 1 second, a 997 millisecond deadline may or may not cause a timeout.

I'll change the error handling.
What happens is that if a context deadline is exceeded, the running process is killed which is why `Run()` would return `"signal: killed"` as an error.
I think that if the deadline was super short (i.e. microseconds or lower), the process did not even execute yet which is why `"context deadline exceeded"` was returned.

I now changed it such that if `Run()` returns an error, first the context error is checked to find out whether the context deadline has been exceeded. If the context error is nil, the error returned by Run() will be displayed.

docs/tasks.md
Peaorl: The first line [here](https://github.com/tektoncd/pipeline/blob/0c2ff5cd1a7fb47b2dc012fd99ddd19d9e3bf3d6/docs/tasks.md#specifying-a-timeout) mentions that `Step` and subsequently the entire `TaskRun` are canceled.

I'll elaborate and explicitly specify subsequent `Steps` won't execute too. 
I can elaborate a bit more on the timeout log as well.

Furthermore I'll amend the doc by mentioning that the process spawned by a `Step` is killed when a timeout occurs.

pkg/pod/pod.go
Peaorl: Resolved this by adding the timeout to the stepContainers earlier on in the code.

pkg/reconciler/taskrun/taskrun_test.go
Peaorl: Please see #3138 for a cleanup.

pkg/reconciler/taskrun/taskrun_test.go
Peaorl: See #3138

test/timeout_test.go
```diff
@@ -163,6 +163,67 @@ func TestPipelineRunTimeout(t *testing.T) {
...
+				{Container: corev1.Container{
+					Name:    "albatros",
+					Image:   "busybox",
+					Command: []string{"/bin/sh"},
+					Args:    []string{"-c", "sleep 1"},
+				}},
+			},
+		},
+	}
+	if _, err := c.TaskClient.Create(task); err != nil {
```
ImJasonH, line 196: It might be a bit easier to read this test if you just create a TaskRun with the `taskSpec` declared inline, rather than creating a Task then running it.

test/timeout_test.go
```diff
@@ -163,6 +163,67 @@ func TestPipelineRunTimeout(t *testing.T) {
...
+	knativetest.CleanupOnInterrupt(func() { tearDown(t, c, namespace) }, t.Logf)
+	defer tearDown(t, c, namespace)
+
+	t.Logf("Creating Task with Step step-penguin and Step step-albatros in namespace %s", namespace)
+	taskName := "timeout"
+	task := &v1beta1.Task{
+		ObjectMeta: metav1.ObjectMeta{Name: taskName, Namespace: namespace},
+		Spec: v1beta1.TaskSpec{
+			Steps: []v1beta1.Step{
+				{Container: corev1.Container{
```
ImJasonH, line 180: Readability nit: I think it might be a bit easier to scan this if you change the brackets and newlines a bit, and use script mode instead of `sh -c`, to something like:

```
Steps: []v1beta1.Step{{
  Container: corev1.Container{
    Image: "busybox",
  },
  Script: "sleep 1",
  Timeout: "1ms",
}, {
  Container: corev1.Container{
    Image: "busybox",
  },
  Script: "sleep 1",
}}
```

vdemeester:
I know this is pretty late in the review process, but shouldn't this need a TEP ? (cc @pritidesai @bobcatfish @ImJasonH @sbwsg )

ImJasonH:
I don't have a problem with that. It should be a fairly straightforward TEP, and good practice for writing TEPs 🙂

/hold

Peaorl:
Awaiting result of #3239

Peaorl:
I updated the PR with a [TEP](https://github.com/tektoncd/community/blob/master/teps/0014-step-timeout.md)
Other then that I'd say the feature is finished and ready for (further) review.

pkg/entrypoint/entrypointer.go
```diff
@@ -63,6 +64,8 @@ type Entrypointer struct {
 
 	// Results is the set of files that might contain task results
 	Results []string
+	//Timeout is an optional user-specified flag for timing out Steps
```
sbwsg, line 67: nit: space after `//`. Suggest slightly different wording: `Timeout is an optional user-specified duration within which the Step must complete`

sbwsg:
Looking good!  Are you planning to move the TEP into `implementable` state?

cmd/entrypoint/runner_test.go
```diff
@@ -14,9 +15,22 @@ func TestRealRunnerSignalForwarding(t *testing.T) {
...
 	} else {
 		t.Fatalf("Unexpected error received: %v", err)
 	}
 }
+
+// TestRealRunnerTimeout tests whether cmd is killed after a millisecond even though it's supposed to sleep for 10 milliseconds.
+func TestRealRunnerTimeout(t *testing.T) {
+	rr := realRunner{}
+	timeout := "1ms"
+	if err := rr.Run(timeout, "sleep", "0.01"); err != nil {
```
sbwsg, line 29: It might be worth adding a couple of tests to codify how the function deals with edge cases like: invalid timeout string, negative timeout string, empty string.

Peaorl: Agreed on the empty string.
As for the invalid/negative timeout string, I'm not sure it adds much considering that the [webhook](https://github.com/tektoncd/pipeline/blob/d7cfc6671dae95c935331a20189d521b1a8379ac/pkg/apis/pipeline/v1beta1/task_validation.go#L164) would already bar such strings in the first place, what do you think?

sbwsg: Hm, I think there are still edge cases that wouldn't be caught by the validating webhook - the controller could have a bug in it that mangles the strings it sends in the entrypoint args, another developer could accidentally introduce a "-" to the duration value.

Edge cases like this can seem a bit silly but it's better (in my view anyway) for the code to fail loudly if we introduce any bugs like that than for it to start quietly exhibiting inexplicable behaviour that we don't have testing around.

Peaorl: I  understand. As you pointed out offline, this will be resolved in #3304 .

Peaorl: Additionally I added unit tests for the webhook.

sbwsg (approved):

cmd/entrypoint/runner.go
```diff
@@ -19,7 +22,7 @@ type realRunner struct {
 
 var _ entrypoint.Runner = (*realRunner)(nil)
 
-func (rr *realRunner) Run(args ...string) error {
+func (rr *realRunner) Run(timeout string, args ...string) error {
```
ImJasonH, line 25: I might suggest this method taking a ctx argument instead of a string timeout, or if there's some reason not to take a ctx, have it take a time.Duration instead.

The reason is that because of the variadic arg, it'll be a bit confusing to see it called like `Run("30s", "echo", "hello")`, compared to `Run(ctx, "echo", "hello")` or slightly worse,`Run(30*time.Second, "echo", "hello")`

pkg/pod/entrypoint_test.go
```diff
@@ -87,7 +87,7 @@ func TestOrderContainers(t *testing.T) {
 		VolumeMounts:           []corev1.VolumeMount{toolsMount},
 		TerminationMessagePath: "/tekton/termination",
 	}}
-	gotInit, got, err := orderContainers(images.EntrypointImage, []string{}, steps, nil)
+	gotInit, got, err := orderContainers(images.EntrypointImage, []string{}, make([][]string, len(steps)), steps, nil)
```
ImJasonH, line 90: Can we pass a `nil` slice here instead of an empty one? That might be easier to read here.

pkg/pod/entrypoint.go
```diff
@@ -84,9 +84,7 @@ var (
 // Containers must have Command specified; if the user didn't specify a
 // command, we must have fetched the image's ENTRYPOINT before calling this
 // method, using entrypoint_lookup.go.
-//
-// TODO(#1605): Also use entrypoint injection to order sidecar start/stop.
-func orderContainers(entrypointImage string, extraEntrypointArgs []string, steps []corev1.Container, results []v1beta1.TaskResult) (corev1.Container, []corev1.Container, error) {
+func orderContainers(entrypointImage string, commonExtraEntrypointArgs []string, extraEntrypointArgs [][]string, steps []corev1.Container, results []v1beta1.TaskResult) (corev1.Container, []corev1.Container, error) {
```
ImJasonH, line 87: Maybe instead of `extraEntrypointArgs`, which at least for now will always be a slice of at-most-two-element slices like `[]string{"-timeout", "30s"}`, we just make this `timeouts []*time.Duration`, where a `nil` means "no timeout"?

The addition of timeouts args, or extra args in general, feels like it belongs outside a method called `orderContainers`, which is so far primarily focused on setting `-wait_file`/`-post_file` flags. Should we rename the method to describe its new scope? Or split it into separate methods?

cmd/entrypoint/runner.go
Peaorl: Originally I did pass the context here, the reason that I'm now passing the timeout argument is that as soon as you add a deadline to a context, time starts running. The closer to the execution of the actual command the timeout is added to the context, the less timeout time is wasted on overhead before the command executes.

In reality this overhead (which technically could vary between systems) is probably in the order of milliseconds which may not be significant for the timeouts we expect to be set for a `Step`. If we accept that "wasting of time" it could also be done as you suggest.

ImJasonH: I think that's an acceptable skew from enforced timeout to "actual" timeout. If a user notices their timeouts being enforced 100 microseconds earlier, I'd be surprised :)

cmd/entrypoint/main.go
```diff
@@ -41,6 +41,7 @@ var (
 	terminationPath     = flag.String("termination_path", "/tekton/termination", "If specified, file to write upon termination")
 	results             = flag.String("results", "", "If specified, list of file names that might contain task results")
 	waitPollingInterval = time.Second
+	timeout             = flag.String("timeout", "", "If specified, sets timeout for step")
```
ImJasonH, line 44: So there's https://golang.org/pkg/flag/#Duration which handles parsing for us, and will let us abstract away another parse...

pkg/apis/pipeline/v1beta1/task_types.go
```diff
@@ -125,10 +125,11 @@ type Step struct {
 	//
 	// If Script is not empty, the Step cannot have an Command or Args.
 	Script string `json:"script,omitempty"`
+	// If step times out after Timeout, pod is terminated
+	Timeout string `json:"timeout,omitempty"`
```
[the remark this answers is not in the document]
Peaorl: Indeed much better!

pkg/entrypoint/entrypointer.go
```diff
@@ -63,6 +64,8 @@ type Entrypointer struct {
 
 	// Results is the set of files that might contain task results
 	Results []string
+	// Timeout is an optional user-specified duration within which the Step must complete
```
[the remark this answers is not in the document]
Peaorl: Made it `*time.Duration` in this case because it'll already be parsed as such.

cmd/entrypoint/runner.go
```diff
@@ -42,6 +43,9 @@ func (rr *realRunner) Run(args ...string) error {
 
 	// Start defined command
 	if err := cmd.Start(); err != nil {
+		if ctx.Err() == context.DeadlineExceeded {
```
ImJasonH, line 46: What error does cmd.Start return when the context's deadline is exceeded? Can we just return that? Having to check for ctx.Err() in each of these is cumbersome and we're likely to forget one in future changes.

pkg/apis/pipeline/v1beta1/task_types.go
```diff
@@ -125,10 +125,11 @@ type Step struct {
 	//
 	// If Script is not empty, the Step cannot have an Command or Args.
 	Script string `json:"script,omitempty"`
+	// If step times out after Timeout, pod is terminated
```
ImJasonH, line 128: Let's generalize this comment to describe what the user will see, instead of how it will work.

```
    // Timeout is the time after which the step times out. Defaults to never.
    // Refer to Go's ParseDuration documentation for expected format: https://golang.org/pkg/time/#ParseDuration
```

cmd/entrypoint/runner.go
Peaorl: cmd.Start() returns some generic error about the command being killed, this could be for any reason. Therefore we need to check ctx.Err() to know whether this was due to exceeding the timeout so we can have appropriate error reporting.

ImJasonH (commented):
/lgtm

ImJasonH (approved):

vdemeester (commented):
/lgtm

sbwsg:
Looks like all commits have been merged into 1.  Given approval and lgtms I'm going to cancel the hold.

/hold cancel

Peaorl:
Automatic merge/rebase into master didn't go smoothly after a new [commit](https://github.com/tektoncd/pipeline/commit/b9a21a6d476f3a2b2b8336c29f40b87cebf23142) yesterday so the code has been updated.
Additionally, I increased the timeout limit from 30s to 60s in [examples/v1beta1/taskrun/workspace-in-sidecar.yaml](https://github.com/tektoncd/pipeline/blob/17e7f90ba2f67ba5c9ee434208737771814ef5c5/examples/v1beta1/taskruns/workspace-in-sidecar.yaml) because it seems like the cluster is too resource constrained for it to complete in time.

Issue #3085, opened 2020-08-10
0.15.0:  "TaskRun mounts more than one PersistentVolumeClaim" despite single PVC w/ workspace subDirs


bitsofinfo:
I have a task `theTask` with 2 named workspaces:
```
...
  workspaces:

    - name: app-source
      description: The workspace where the app source code resides
      mountPath: /workspace/app-source
      readOnly: false

    - name: secondary-source
      description: The workspace where the secondary source code resides
      mountPath: /workspace/secondary-source
      readOnly: false
...
```

I have a pipeline with 1 workspace that is mapped to the 2 task workspaces via the diffentiating `subPath` flag:
```
    - name: theTask
      taskRef:
        name: theTask

      runAfter: 
        - init

      workspaces:
        - name: secondary-source
          workspace: build-ops
          subPath: secondary-source
        - name: app-source
          workspace: build-ops
          subPath: app-source
    
```

This worked totally fine up to 0.14.3 but now fails with this error in 0.15.0 and broke several functioning pipelines

```
 TaskRun mounts more than one PersistentVolumeClaim - that is forbidden
      when the Affinity Assistant is enabled
```

Related I assume: https://github.com/tektoncd/pipeline/pull/2885

Seems one should still be above to mount a single PVC, which is what this is, its just logically divided by the different subPaths, vs truly separate PVCs which I am not attempting to do.

PR #3091, opened 2020-08-11
Add support for repeated PVC-claim but using subPath in AA-validation

jlpettersson:
# Changes

The validation for compatibility with the Affinity Assistant does not
support the same PVC repeated, but using different subPaths.

This patch adds support for this case.

Closes https://github.com/tektoncd/pipeline/issues/3085
/kind bug

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Fixes a bug with validation for the Affinity Assistant when the same PVC is used for
multiple workspaces but with different subPaths
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -760,14 +760,17 @@ func validateWorkspaceCompatibilityWithAffinityAssistant(tr *v1beta1.TaskRun) er
 		return nil
 	}
 
-	pvcWorkspaces := 0
+	workspaceVolumes := make(map[string]bool)
 	for _, w := range tr.Spec.Workspaces {
-		if w.PersistentVolumeClaim != nil || w.VolumeClaimTemplate != nil {
-			pvcWorkspaces++
+		if w.PersistentVolumeClaim != nil {
+			workspaceVolumes[w.PersistentVolumeClaim.ClaimName] = true
+		}
+		if w.VolumeClaimTemplate != nil {
+			workspaceVolumes[w.Name] = true
 		}
 	}
 
-	if pvcWorkspaces > 1 {
+	if len(workspaceVolumes) > 1 {
 		return fmt.Errorf("TaskRun mounts more than one PersistentVolumeClaim - that is forbidden when the Affinity Assistant is enabled")
 	}
 	return nil


bitsofinfo:
will take it with or without a test! :) don't want to be stuck in 0.14.3

sbwsg:
> I'll leave this as a draft for now - since I have not implemented any test case for this case. I feel it is a bit heavy to add a complete new test case for this case - so the test may need refactoring. Feel free to add test case if anyone feel this fix is in hurry.

My suggestion would be to move the validation bit into `pkg/workspace/validate.go` and move the annotation check into the `prepare()` func of the reconciler.  So that would look something like this:

```go
	// pkg/reconciler/taskrun/taskrun.go
	// after workspace.ValidateBindings() is called in prepare()
	_, isAssociatedWithAnAffinityAssistant := tr.Annotations[workspace.AnnotationAffinityAssistantName]
	if isAssociatedWithAnAffinityAssistant {
		if err := workspace.ValidateOnlyOnePVCBindingUsed(tr.Spec.Workspaces); err != nil {
			logger.Errorf("TaskRun %q workspaces are invalid: %v", tr.Name, err)
			tr.Status.MarkResourceFailed(podconvert.ReasonFailedValidation, err)
		}
	}
```

```go
// pkg/workspace/validate.go
func ValidateOnlyOnePVCBindingUsed(wbs []v1beta1.WorkspaceBinding) error {
	workspaceVolumes := make(map[string]bool)
	for _, w := range wbs {
		if w.PersistentVolumeClaim != nil {
			workspaceVolumes[w.PersistentVolumeClaim.ClaimName] = true
		}
		if w.VolumeClaimTemplate != nil {
			workspaceVolumes[w.Name] = true
		}
	}

	if len(workspaceVolumes) > 1 {
		return fmt.Errorf("TaskRun mounts more than one ...") // cut for brevity
	}
	return nil
}
```

Then `ValidateOnlyOnePVCBindingUsed` can be tested in isolation in `pkg/workspace/validate_test.go`. @jlpettersson are you happy with this suggestion?  I can implement it and push to this PR tomorrow if you agree with the direction.

jlpettersson:
@sbwsg sounds good to me

sbwsg:
Updated as described and opened #3099

Issue #3089, opened 2020-08-11
Able to start pipelinerun from pipelinerun spec without providing value for workspace and the pipelinerun stays is `Running(Started)` state


VeereshAradhya:
# Expected Behavior
Pipelinerun should fail with some validation error

# Actual Behavior
Pipelinerun starts and stays in `Running(Started)` state

# Steps to Reproduce the Problem

1. Create a task with workspace
2. Create a pipeline with workspace using the above task
3. Create pipelinerun spec file to start the pipelinerun for above pipeline and don't provide any value for the workspace
4. start the pipelinerun using `kubectl apply -f`

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  Client Version: version.Info{Major:"1", Minor:"10+", GitVersion:"v1.10.0+b3b92b2", GitCommit:"b3b92b2", GitTreeState:"clean", BuildDate:"2020-07-15T09:27:21Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"linux/amd64"}
  Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.3+012b3ec", GitCommit:"012b3ec", GitTreeState:"clean", BuildDate:"2020-07-24T07:23:10Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}

  ```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**
  ```
  $ tkn version
  Client version: 0.11.0
  Pipeline version: v0.14.3
  Triggers version: v0.6.1
  ```

**Command logs**
```
[varadhya@localhost workspace-testing]$ cat workspace-pipelinerun.yaml 
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  name: test-workspace-pipeline-run-7
spec:
  params:
  - name: first-param
    value: simple value
  pipelineRef:
    name: test-workspace-pipeline
  serviceAccountName: pipeline
  timeout: 1h0m0s
  workspaces:
  - name: shared-workspace
[varadhya@localhost workspace-testing]$ 
[varadhya@localhost workspace-testing]$ oc apply -f workspace-pipelinerun.yaml 
pipelinerun.tekton.dev/test-workspace-pipeline-run-7 created
[...]

PR #3096, opened 2020-08-12
Validate missing volume configs in Workspace Bindings

sbwsg:
# Changes

Contributes to https://github.com/tektoncd/pipeline/issues/3089 but doesn't completely fix it. There's another bug hidden in there that a PipelineRun submitting a TaskRun with invalid workspace is not entering a failed state even though the TaskRun reconciler is loudly complaining about the workspace config.

When a PipelineRun specifies a Workspace Binding, it is required to
submit the volume configuration as part of that Binding. Currently
the volume details aren't validated and so a PipelineRun with a Workspace
Binding that only includes a "name" field will validate. The end
result is that the Workspace is passed to the TaskRun and fails
validation there instead.

This commit adds validation to PipelineRuns to ensure that any
Workspaces passed include some correct volume configuration data.

# Submitter Checklist

review them:

```release-note
Fix an issue where PipelineRuns would pass validation even when a workspace binding was missing required volume info.
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/workspace_validation.go
+++ b/pkg/apis/pipeline/v1beta1/workspace_validation.go
@@ -26,11 +26,11 @@ import (
 // allVolumeSourceFields is a list of all the volume source field paths that a
 // WorkspaceBinding may include.
 var allVolumeSourceFields []string = []string{
-	"workspace.persistentvolumeclaim",
-	"workspace.volumeclaimtemplate",
-	"workspace.emptydir",
-	"workspace.configmap",
-	"workspace.secret",
+	"persistentvolumeclaim",
+	"volumeclaimtemplate",
+	"emptydir",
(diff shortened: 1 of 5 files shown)


pkg/apis/pipeline/v1beta1/workspace_validation.go
```diff
@@ -26,11 +26,11 @@ import (
 // allVolumeSourceFields is a list of all the volume source field paths that a
 // WorkspaceBinding may include.
 var allVolumeSourceFields []string = []string{
-	"workspace.persistentvolumeclaim",
-	"workspace.volumeclaimtemplate",
-	"workspace.emptydir",
-	"workspace.configmap",
-	"workspace.secret",
+	"persistentvolumeclaim",
```
dlorenc, line 29: nice catch!

PR #3099, opened 2020-08-13
Add support for repeated PVC-claim but using subPath in AA-validation

### LINKED ISSUE #3085: 0.15.0:  "TaskRun mounts more than one PersistentVolumeClaim" despite single PVC w/ workspace subDirs

sbwsg:
# Changes

Picks up from #3091 and adds some tests for the validation. Original commit message:

The validation for compatibility with the Affinity Assistant does not
support the same PVC repeated, but using different subPaths.

This patch adds support for this case and tests for the validation.

Co-authored-by: Scott <sbws@google.com>

Closes #3085

Closes #3091 

/kind bug

# Submitter Checklist

review them:

# Release Notes

```release-note
Fixes a bug with validation for the Affinity Assistant when the same PVC is used for multiple workspaces but with different subPaths
```

### DIFF
--- a/pkg/workspace/validate.go
+++ b/pkg/workspace/validate.go
@@ -48,3 +48,25 @@ func ValidateBindings(w []v1beta1.WorkspaceDeclaration, wb []v1beta1.WorkspaceBi
 	}
 	return nil
 }
+
+// ValidateOnlyOnePVCIsUsed checks that a list of WorkspaceBinding uses only one
+// persistent volume claim.
+//
+// This is only useful to validate that WorkspaceBindings in TaskRuns are compatible
+// with affinity rules enforced by the AffinityAssistant.
+func ValidateOnlyOnePVCIsUsed(wb []v1beta1.WorkspaceBinding) error {
+	workspaceVolumes := make(map[string]bool)
+	for _, w := range wb {
+		if w.PersistentVolumeClaim != nil {
+			workspaceVolumes[w.PersistentVolumeClaim.ClaimName] = true
+		}
+		if w.VolumeClaimTemplate != nil {
+			workspaceVolumes[w.Name] = true
+		}
+	}
+
+	if len(workspaceVolumes) > 1 {
+		return fmt.Errorf("more than one PersistentVolumeClaim is bound")
+	}
+	return nil
+}
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -302,10 +302,12 @@ func (c *Reconciler) prepare(ctx context.Context, tr *v1beta1.TaskRun) (*v1beta1
 		return nil, nil, controller.NewPermanentError(err)
 	}
 
-	if err := validateWorkspaceCompatibilityWithAffinityAssistant(tr); err != nil {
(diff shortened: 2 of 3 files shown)


pkg/workspace/validate.go
```diff
@@ -48,3 +48,23 @@ func ValidateBindings(w []v1beta1.WorkspaceDeclaration, wb []v1beta1.WorkspaceBi
 	}
 	return nil
 }
+
+// ValidateOnlyOnePVCIsUsed checks that a list of WorkspaceBinding uses only
+// one persistent volume claim. This is useful to validate that WorkspaceBindings
+// are compatible with affinity rules enforced by the AffinityAssistant.
+func ValidateOnlyOnePVCIsUsed(wb []v1beta1.WorkspaceBinding) error {
```
jlpettersson, line 55: nit: maybe add to the comment that this condition (multiple pvcs) is valid for `WorkspaceBinding` in a PipelineRun, but not for a `TaskRun`

sbwsg: good thinking; I've updated the comment to include that detail.

jlpettersson: >This is useful to validate that WorkspaceBindings
// in PipelineRuns are compatible with affinity rules enforced by the
// AffinityAssistant.

It's the other way around :) . We only call this function from `taskrun.go` because this condition is ok when declared in a `PipelineRun`

jlpettersson: Maybe 

// This is only useful to validate that WorkspaceBindings
// in TaskRuns are compatible with affinity rules enforced by the
// AffinityAssistant.

sbwsg: oops, hah, thanks I've corrected it with your wording.

ImJasonH:
Do we want this PR in the v0.16 release milestone?

jlpettersson:
> Do we want this PR in the v0.16 release milestone?

I think so, unless we want it in a patch. It should be in a v0.15.x also.

vdemeester (commented):
/lgtm

PR #3107, opened 2020-08-17
Add multiarch specific fixes to the pipeline tests

barthy1:
# Changes
Code to run the tests has many amd64 specific images.
Tests fail on non amd64 architecture because of that.
This code provides one place to keep multiarch
specific updates, separated for each architecture.

Multiarch utils do following:
- replace image names to arch specific ones where it is applicable
- skip the tests, which fail for non amd64 arch now, based on test name
 and architecture
- skipping will be applied for e2e and examples tests

# Submitter Checklist

review them:

# Release Notes

```release-note
Add multiarch specific fixes to the tests
 ```

### DIFF
--- a/test/examples_test.go
+++ b/test/examples_test.go
@@ -116,6 +116,8 @@ type waitFunc func(t *testing.T, c *clients, name string)
 
 func exampleTest(path string, waitValidateFunc waitFunc, kind string) func(t *testing.T) {
 	return func(t *testing.T) {
+		SkipIfExcluded(t)
+
 		t.Parallel()
 
 		// Setup unique namespaces for each test so they can run in complete
--- a/test/git_checkout_test.go
+++ b/test/git_checkout_test.go
@@ -41,6 +41,8 @@ const (
 // TestGitPipelineRun is an integration test that will verify the source code is either fetched or pulled
 // successfully under different revision inputs (branch, commitid, tag, ref)
 func TestGitPipelineRun(t *testing.T) {
+	SkipIfExcluded(t)
+
 	t.Parallel()
 
 	revisions := []string{"master", "c15aced0e5aaee6456fbe6f7a7e95e0b5b3b2b2f", "release-0.1", "v0.1.0", "refs/pull/347/head"}
@@ -82,6 +84,8 @@ func TestGitPipelineRun(t *testing.T) {
 
 // Test revision fetching with refspec specified
 func TestGitPipelineRunWithRefspec(t *testing.T) {
+	SkipIfExcluded(t)
+
 	t.Parallel()
 
 	for _, tc := range []struct {
@@ -303,6 +307,8 @@ func TestGitPipelineRunFail_HTTPS_PROXY(t *testing.T) {
 // successfully under different revision inputs (default branch, branch)
 // This test will run on spring-petclinic repository which does not contain a master branch as the default branch
 func TestGitPipelineRunWithNonMasterBranch(t *testing.T) {
+	SkipIfExcluded(t)
+
 	t.Parallel()
 
 	revisions := []string{"", "main"}
--- a/test/init_test.go
+++ b/test/init_test.go
@@ -51,6 +51,8 @@ func init() {
 }
 
 func setup(t *testing.T, fn ...func(*testing.T, *clients, string)) (*clients, string) {
+	SkipIfExcluded(t)
+
 	t.Helper()
 	namespace := names.SimpleNameGenerator.RestrictLengthWithRandomSuffix("arendelle")
 
--- a/test/multiarch_utils.go
+++ b/test/multiarch_utils.go
@@ -1,0 +1,136 @@
+/*
+Copyright 2019 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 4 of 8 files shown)

### CODE REVIEW

File: test/multiarch_utils.go [hunk 1]
```diff
@@ -0,0 +1,131 @@
...
+		mapping[Registry] = "ibmcom/registry:2.6.2.5"
+
+	default:
+		mapping[BusyboxSha] = "busybox@sha256:895ab622e92e18d6b461d671081757af7dbaa3b00e3e28e12505af7817f73649"
+		mapping[Registry] = "registry"
+	}
+	return mapping
+}
+
+func initExceludedTests() map[string]bool {
```
vdemeester, line 48: typo: `initExcludedTests`

vdemeester: Also, not sure why the `init` prefix :stuck_out_tongue_closed_eyes:

barthy1: thanks for the catch! fixed it
init is used because the func initiates list of excluded tests. Do you think it better to use another name?

test/init_test.go
```diff
@@ -50,7 +51,26 @@ func init() {
 	flag.BoolVar(&skipRootUserTests, "skipRootUserTests", false, "Skip tests that require root user")
 }
 
+func excludeForArchitecture(callerLevel int) bool {
```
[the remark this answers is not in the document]
barthy1: added comment to explain what is going on

test/examples_test.go
```diff
@@ -21,16 +21,16 @@ package test
...
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
-
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	knativetest "knative.dev/pkg/test"
 )
```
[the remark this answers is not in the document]
barthy1: done

test/init_test.go
```diff
@@ -50,7 +51,27 @@ func init() {
 	flag.BoolVar(&skipRootUserTests, "skipRootUserTests", false, "Skip tests that require root user")
 }
 
+// excludeForArchitecture checks if name of calling function(name of the test) is in in the excluded list
```
dlorenc, line 54: nit: "is in in"

test/multiarch_utils.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// get test image based on unique id
+func GetTestImage(image int) string {
+	return imageNames[image]
+}
+
+// check if test name is in the excluded list based on current hardware architecture
+func IsExcluded(testName string) bool {
+	return excludedTests[testName]
```
dlorenc, line 130: Do we ever expect this to be false? Would it be easier to instead make this a map[string]struct{} and use an existence check?

test/init_test.go
```diff
@@ -50,7 +51,27 @@ func init() {
...
+	}
+	lastDot := strings.LastIndexByte(funcFullName[lastSlash:], '.') + lastSlash
+	funcName := funcFullName[lastDot+1:]
+
+	//check if func name, which equal to test name is excluded for special architecture
+	return IsExcluded(funcName)
+}
+
 func setup(t *testing.T, fn ...func(*testing.T, *clients, string)) (*clients, string) {
+	if excludeForArchitecture(1) || excludeForArchitecture(2) {
```
dlorenc, line 71: nit: the integers here don't have much context. I would find it more clear to move the 1, 2 and || check up into the excludeForArchitecture function itself.

test/init_test.go
```diff
@@ -50,7 +51,27 @@ func init() {
 	flag.BoolVar(&skipRootUserTests, "skipRootUserTests", false, "Skip tests that require root user")
 }
 
+// excludeForArchitecture checks if name of calling function(name of the test) is in in the excluded list
+func excludeForArchitecture(callerLevel int) bool {
+	//get the Caller and the func name after
+	pc, _, _, _ := runtime.Caller(callerLevel)
```
dlorenc, line 57: Could we use t.Name() instead? https://golang.org/src/testing/testing.go?s=22385:22415#L639

(I'm not actually sure, but it looks like it returns the current test name)

test/multiarch_utils.go
barthy1: yes, we expect this to be false for some cases. Following https://blog.golang.org/maps `map[string]bool` returns false if key doesn't exist in the map.
You also can have a look in the integration test, no tests from the "arch specific list" were skipped

test/init_test.go
barthy1: Thank you so much! yes, you are right, t.Name() solves all my challenges with tests and subtests. Don't know how I lost it in my mind.. :(

test/init_test.go
barthy1: not a topic any more, as per your great suggestion this code was removed

test/git_checkout_test.go
```diff
@@ -303,6 +310,10 @@ func TestGitPipelineRunFail_HTTPS_PROXY(t *testing.T) {
 // successfully under different revision inputs (default branch, branch)
 // This test will run on spring-petclinic repository which does not contain a master branch as the default branch
 func TestGitPipelineRunWithNonMasterBranch(t *testing.T) {
+	if IsExcluded(t.Name()) {
+		t.Skipf("skip for %s architecture", runtime.GOARCH)
+	}
+
```
vdemeester, line 316: nit: As this is duplicated a bunch of times, this could even be a function : `skipIfExcluded(t.Name())`.

barthy1: good point! will do

barthy1: however it will be `skipIfExcluded(t)`, as we need `t` to skip it

vdemeester: true :grin:

test/multiarch_utils.go
dlorenc: Ah, thanks for clarifying.

PR #2573, opened 2020-05-07
Sync the pipelinerun status from the informers

afrittoli:
# Changes

When we reconcile a pipelinerun, we should ensure that the
pipelinerun status is always in sync with the actual list of taskruns
that can be provided by the taskrun informer.

The only way to filter taskruns is by labels tekton.dev/pipelinerun.
In case an orphaned taskrun is found, we can use the other labels
on the taskrun to reconstruct the missing entry in the pipelinerun
status.

Fixes #2558 

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

Double check this list of stuff that's easy to miss:

- If you are adding [a new binary/image to the `cmd` dir](../cmd), please update
  [the release Task](../tekton/publish.yaml) to build and release this image.

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

PR #3115, opened 2020-08-19
Create Run resources from PipelineTask references

ImJasonH:
# Changes

Progress toward #3133

This change allows  `TaskRef`s in `PipelineTask`s to refer to a non-Tekton ("custom") task type. When such a Pipeline is run, the controller will create a new `Run` resource that embeds that reference, at which point a custom task controller can take over and begin reconciling that execution, eventually to completion.

Followups for future PRs:
- documentation about specifying Custom Task refs in a `Pipeline`
- `RunStatus` is not yet surfaced in the  `PipelineRunStatus`  since it currently introduces an import cycle 💣  -- need to move `RunStatus` to `v1beta1` even though `Run` is still in `v1alpha1`.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Allow Custom Task references in PipelineTask references, which result in Run resources being created.
```


pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -272,6 +278,14 @@ func validatePipelineTaskName(ctx context.Context, prefix string, i int, t Pipel
 		}
 		taskNames[t.Name] = struct{}{}
```
GregDritschler, line 279: The check for duplicate pipeline task names occurs inside this block when a taskref name is used.  So the validation hasn't been done if taskSpec is used, and now it also won't be done if a custom task has no taskRef name.  Should this duplicate name check be outside this block?  

I ran a test to confirm this.  If taskRef is used, a duplicate pipeline task name is caught here.  If taskSpec is used, a duplicate pipeline task name is not caught here, but it is caught by the reconciler when it builds the DAG:

```
Error from server (BadRequest): error when creating "C:\\dbg\\tekton\\samples\\paralleltasks.yaml": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: invalid value: task task1 is already present in Graph, can't add it again: duplicate pipeline task: spec.tasks
```

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -129,6 +136,27 @@ func TestPipeline_Validate_Failure(t *testing.T) {
 		p: &Pipeline{
 			ObjectMeta: metav1.ObjectMeta{Name: "pipeline"},
 		},
+	}, {
+		name: "pipelinetask without name",
+		p: &Pipeline{
+			Spec: PipelineSpec{
+				Tasks: []PipelineTask{{Name: "", TaskRef: &TaskRef{Name: "valid-task"}}},
```
GregDritschler, line 143: This test is already done in `TestValidatePipelineTasks_Failure` (line 320).  Perhaps your other two tests should go in that method too.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -564,9 +577,42 @@ func getPipelineRunResults(pipelineSpec *v1beta1.PipelineSpec, resolvedResultRef
...
+			prrs = &v1beta1.PipelineRunRunStatus{
+				PipelineTaskName: rprt.PipelineTask.Name,
+			}
+		}
+
+		if rprt.Run != nil {
+			// TODO(jasonhall): Support RunStatus.
+			//prrs.Status = &rprt.Run.Status
+		}
+		status[rprt.RunName] = prrs
```
GregDritschler, line 605: Add another TODO here to handle the status of the pipeline task's condition checks (the same way as `getTaskRunsStatus` does)?  Or are Condition checks not supported with custom tasks?

ImJasonH: Conditions should be supported, TODO added. Thanks!

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
ImJasonH: Good catch! Fixed.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
ImJasonH: Good catch! Fixed, and updated the test so it would have caught this.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -267,11 +279,15 @@ func validatePipelineTaskName(ctx context.Context, prefix string, i int, t Pipel
...
-			return apis.ErrMultipleOneOf(fmt.Sprintf(prefix+"[%d].name", i))
-		}
-		taskNames[t.Name] = struct{}{}
 	}
+
+	if hasTaskRef && !isCustomTask && t.TaskRef.Name == "" {
+		return apis.ErrInvalidValue(t.TaskRef, "taskRef must specify name")
+	}
+	if isCustomTask && t.TaskRef.Kind == "" {
+		return apis.ErrInvalidValue(t.TaskRef, "custom task ref must specify apiVersion and kind")
```
GregDritschler, line 288: There are a bunch of things that aren't supported initially with custom tasks:
* retries
* pipelineresources
* workspaces

There should be validation if the pipeline task tries to use these.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -422,54 +458,58 @@ func ResolvePipelineRun(
...
-		if err != nil {
-			if !errors.IsNotFound(err) {
+			taskRun, err := getTaskRun(rprt.TaskRunName)
+			if err != nil && !errors.IsNotFound(err) {
 				return nil, fmt.Errorf("error retrieving TaskRun %s: %w", rprt.TaskRunName, err)
 			}
-		}
-		if taskRun != nil {
 			rprt.TaskRun = taskRun
 		}
```
GregDritschler, line 515: There should be a TODO to update `resolveConditionChecks` to understand `PipelineRunRunStatus` as well.

GregDritschler:
There are a couple of other TaskRun-related functions in pipelinerun.go that might need equivalents for Run.  I think these are for corner case situations so it's not needed immediately but maybe just put TODOs to revisit them.

`updateTaskRunsStatusDirectly`:  This is called on the pr.isDone() path to re-sync all of the TaskRun statuses.  I'm not exactly sure why that's necessary;  I would have thought they all had to be synced to make a decision the PR is done and can't change after that.  But I guess there may be corner cases I guess where that's not true.

`updatePipelineRunStatusFromInformer`:  This looks for existing TaskRuns associated with the PR and makes sure the PR has their statuses.  It's intended to cover a window where the reconciler creates a TR but then fails when updating the PR status.  Without this, the next reconcile would create another TR because it lost track of it.  Similar logic is needed for Runs but it's definitely a corner case.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -66,79 +66,96 @@ func (e *ConditionNotFoundError) Error() string {
 // ResolvedPipelineRunTask contains a Task and its associated TaskRun, if it
 // exists. TaskRun can be nil to represent there being no TaskRun.
 type ResolvedPipelineRunTask struct {
-	TaskRunName           string
-	TaskRun               *v1beta1.TaskRun
+	TaskRunName string
+	TaskRun     *v1beta1.TaskRun
+
+	// If the PipelineTask is a Custom Task, RunName and Run will be set.
+	RunName string
```
pierretasci, line 73: This is more of an FYI than a comment but I noticed that our changes overlap here quite a bit. Overall, I think our approach to resolving pipeline tasks could use a cleaning up since now its OCI images and CustomTasks but who knows what else we will need to resolve in the future.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -696,6 +745,23 @@ func (c *Reconciler) createTaskRun(ctx context.Context, rprt *resources.Resolved
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            rprt.RunName,
+			Namespace:       pr.Namespace,
+			OwnerReferences: []metav1.OwnerReference{pr.GetOwnerReference()},
+			Labels:          getTaskrunLabels(pr, rprt.PipelineTask.Name),
+			Annotations:     getTaskrunAnnotations(pr),
+		},
+		Spec: v1alpha1.RunSpec{
+			Ref:    rprt.PipelineTask.TaskRef,
+			Params: rprt.PipelineTask.Params,
```
popcor255, line 759: Should the spec include the ServiceAccountName?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
ImJasonH: Yeah, absolutely! I was mostly trying to change as little as possible with this change, and try to keep track of cleanups for future PRs. If you have things in mind could you open an issue to help track them? I can add some smells I've noticed too.

pkg/reconciler/pipelinerun/pipelinerun.go
ImJasonH: `Run`s don't (currently) specify a `serviceAccountName`, since that idea is inhererntly tied to the cluster, and `Run`s are designed to be more separated from the cluster.

We might decide after using this for a while that allowing some cluster-specific configs to leak in, but I think I'd like to hold off on that until we have an explicit use case.

GregDritschler: Since the API WG voted to explore looping constructs via custom tasks for now, I think we do have an explicit use case and all TaskRun-related options should be reconsidered.  I know we've discussed using parameters for some things like service account name but then you have the following problems:

- The parameter list diverges from the Task's parameter list.  The custom task has to filter out those extra parameters.

- There is a disconnect between the configuration of the service account name for TaskRuns and Runs in the same pipeline. 

There also will be things needed by Tasks such as workspaces which can't be passed by parameters.

Obviously this doesn't need to be resolved in this PR but it will need to be resolved in one way or another in order to prototype any custom TaskRun functionality.

cc: @bobcatfish

ImJasonH:
> There are a couple of other TaskRun-related functions in pipelinerun.go that might need equivalents for Run. I think these are for corner case situations so it's not needed immediately but maybe just put TODOs to revisit them.
> 
> `updateTaskRunsStatusDirectly`: This is called on the pr.isDone() path to re-sync all of the TaskRun statuses. I'm not exactly sure why that's necessary; I would have thought they all had to be synced to make a decision the PR is done and can't change after that. But I guess there may be corner cases where that's not true.

Yeah I agree, AIUI in theory this shouldn't be necessary -- it's updating TaskRun statuses to match the PipelineRun's view, which shouldn't be necessary.

It's also kind of gross that TaskRuns are given names before they're ever run, so `errors.IsNotFound` is treated as expected (indicating not-run-yet) instead of a real error. We might want to take this opportunity to rethink this behavior.

> 
> `updatePipelineRunStatusFromInformer`: This looks for existing TaskRuns associated with the PR and makes sure the PR has their statuses. It's intended to cover a window where the reconciler creates a TR but then fails when updating the PR status. Without this, the next reconcile would create another TR because it lost track of it. Similar logic is needed for Runs but it's definitely a corner case.

This also seems like a smell to me, we're supposed to be enqueueing PipelineRun reconciliations when TaskRun statuses change already, as well as doing periodic reconciliations in case we miss something. (This was introduced in #2573 for context)

I'll add a TODO in both cases, but before I do them I'll make sure it's actually necessary at all.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
ImJasonH: Done. All of this is likely going to change with When Expressions too, so it'll be a race! 😄

pkg/apis/pipeline/v1beta1/pipeline_validation.go
ImJasonH: Added validation and tests, thanks for catching!

GregDritschler:
> > `updatePipelineRunStatusFromInformer`: This looks for existing TaskRuns associated with the PR and makes sure the PR has their statuses. It's intended to cover a window where the reconciler creates a TR but then fails when updating the PR status. Without this, the next reconcile would create another TR because it lost track of it. Similar logic is needed for Runs but it's definitely a corner case.
> 
> This also seems like a smell to me, we're supposed to be enqueueing PipelineRun reconciliations when TaskRun statuses change already, as well as doing periodic reconciliations in case we miss something. (This was introduced in #2573 for context)

 `updatePipelineRunStatusFromInformer` does cover a potential exposure.  The pipelinerun reconciler might fail to update the PR status after having created a TR.  On the next reconcile it won't have any record of having created a TR so it will create it again, resulting in two TRs for the same pipeline task.  There is an open real-life report of this happening:  #3126.  It's happening despite this function being there, so either there's a hole in it or there's something else going on.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -230,11 +249,19 @@ func (state PipelineRunState) IsStopping(d *dag.Graph) bool {
 func (state PipelineRunState) GetNextTasks(candidateTasks sets.String) []*ResolvedPipelineRunTask {
 	tasks := []*ResolvedPipelineRunTask{}
 	for _, t := range state {
-		if _, ok := candidateTasks[t.PipelineTask.Name]; ok && t.TaskRun == nil {
+		if _, ok := candidateTasks[t.PipelineTask.Name]; ok && t.TaskRun == nil && t.Run == nil {
 			tasks = append(tasks, t)
 		}
-		if _, ok := candidateTasks[t.PipelineTask.Name]; ok && t.TaskRun != nil {
-			status := t.TaskRun.Status.GetCondition(apis.ConditionSucceeded)
+		if _, ok := candidateTasks[t.PipelineTask.Name]; ok {
```
vdemeester, line 255: I find the following easier to follow that current code.

```go
if _, ok := candidateTasks[t.PipelineTask.Name]; ok {
	if t.TaskRun == nil && t.Run == nil {
		tasks = append(tasks, t)
	}
	// […]
}
```

or at least, extracting `_, ok := candidateTasks[t.PipelineTask.Name]` into its own variable and using it in the if or a switch/case :upside_down_face:

test/init_test.go
```diff
@@ -34,13 +34,10 @@ import (
...
-	knativetest "knative.dev/pkg/test"
-	"knative.dev/pkg/test/logging"
-
-	// Mysteriously by k8s libs, or they fail to create `KubeClient`s from config. Apparently just importing it is enough. @_@ side effects @_@. https://github.com/kubernetes/client-go/issues/242
-	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
-	// Mysteriously by k8s libs, or they fail to create `KubeClient`s when using oidc authentication. Apparently just importing it is enough. @_@ side effects @_@. https://github.com/kubernetes/client-go/issues/345
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" // Mysteriously by k8s libs, or they fail to create `KubeClient`s when using oidc authentication. Apparently just importing it is enough. @_@ side effects @_@. https://github.com/kubernetes/client-go/issues/345
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
+	knativetest "knative.dev/pkg/test"
+	"knative.dev/pkg/test/logging" // Mysteriously by k8s libs, or they fail to create `KubeClient`s from config. Apparently just importing it is enough. @_@ side effects @_@. https://github.com/kubernetes/client-go/issues/242
```
vdemeester, line 40: This comment seems misplaced — same goes for the one above.. I wonder how it passes the linter :thinking:

vdemeester (commented):
I think extracting the `Run` part of the go api to it's own package (like `resources`) will ease the import cycle & co pain.

Few comments but overall looks good.

ImJasonH:
Closing in favor of https://github.com/tektoncd/pipeline/pull/3463

PR #3128, opened 2020-08-24
Allow to specify target cluster architecture for tests

barthy1:
# Changes
When e2e or examples tests are started, hardware architectutere of the host
is always used. However test cases can be exectuted on cluster with another architecture,
using `--kubeconfig` parameter.
`TEST_RUNTIME_ARCH` allows to specify target cluster architecture to be able to use
correct restrictions and modifications, specific to concrete arch.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```release-note
Add TEST_RUNTIME_ARCH environment variable to specify target architecture of the tests
```

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -156,6 +156,10 @@ go test -v -count=1 -tags=e2e -timeout=20m ./test
 go test -v -count=1 -tags=e2e -timeout=20m ./test --kubeconfig ~/special/kubeconfig --cluster myspecialcluster
 ```
 
+If tests are applied to the cluster with hardware architecture different to the base one
+(for instance `go test` starts on amd64 architecture and `--kubeconfig` points to s390x Kubernetes cluster),
+use `TEST_RUNTIME_ARCH` environment variable to specify the target hardware architecture(amd64, s390x, ppc64le, arm, arm64, etc)
+
 You can also use
 [all of flags defined in `knative/pkg/test`](https://github.com/knative/pkg/tree/master/test#flags).
 
--- a/test/multiarch_utils.go
+++ b/test/multiarch_utils.go
@@ -17,6 +17,7 @@ limitations under the License.
 package test
 
 import (
+	"os"
 	"runtime"
 	"testing"
 )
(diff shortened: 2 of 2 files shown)


test/README.md
```diff
@@ -156,6 +156,10 @@ go test -v -count=1 -tags=e2e -timeout=20m ./test
 go test -v -count=1 -tags=e2e -timeout=20m ./test --kubeconfig ~/special/kubeconfig --cluster myspecialcluster
 ```
 
+If tests are applied to the cluster with hardware architecture different to the base one
+(for instance `go test` starts on amd64 architecture and `--kubeconfig` points to s390x Kubernetes cluster),
+use `TEST_RUNTIME_ARCH` environment variable to specify the target hardware architecture.
```
dlorenc, line 161: Could we add an example here of what valid TEST_RUNTIME_ARCH values are?

test/multiarch_utils.go
```diff
@@ -33,10 +34,20 @@ const (
...
 )
 
+// return architecture of the cluster where test suites will be executed.
+// default value is similar to build architecture, TEST_RUNTIME_ARCH is used when test target cluster has another architecture
+func getTestArch() string {
+	val, ok := os.LookupEnv("TEST_RUNTIME_ARCH")
+	if ok {
+		return val
+	}
+	return runtime.GOARCH
```
dlorenc, line 44: Defaulting to runtime.GOARCH would default to the runtime of where the tests binary runs, not the architecture of the cluster, right?

That might be fine, but could cause trouble as more development workstations move to arm.

barthy1: Some value should be default, and architecture similarity between dev workstation and dev/test/prod k8s cluster is still most valid case, I believe.
For arm workstation and, for instance, amd64 k8s test cluster `TEST_RUNTIME_ARCH` variable is here :)

test/README.md
barthy1: done

test/multiarch_utils.go
barthy1: Anyway this approach is just to tune the tests in arch specific manner. Most of the tests will run on any architecture(s) as is

dlorenc: Thanks! Yeah this should be fine for now. My initial worry is that this value would also include GOOS (linux, darwin, etc.) and break everyone running on Mac. Now we just have to wait for the arm Macs for that :)

vdemeester (commented):
/lgtm

PR #3131, opened 2020-08-24
Add minimal initial API spec document

ImJasonH:
First part of [TEP-0012](https://github.com/tektoncd/community/pull/172)

[Rendered](https://github.com/imjasonh/pipeline/blob/api-spec/docs/api-spec.md)

# Submitter Checklist

review them:

- [only] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [yes] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)

```release-note
Add initial minimal API specification document
```

/assign @bobcatfish 
/assign @vdemeester

### DIFF
--- a/docs/api-spec.md
+++ b/docs/api-spec.md
@@ -1,0 +1,321 @@
+# Tekton Pipelines API Specification
+
+<!-- toc -->
+- [Abstract](#abstract)
+- [Background](#background)
+- [Modifying This Specification](#modifying-this-specification)
+- [Resource Overview](#resource-overview)
+  * [`TaskRun`](#-taskrun-)
+- [Detailed Resources - v1beta1](#detailed-resources---v1beta1)
+  * [`TaskRun`](#-taskrun--1)
+    + [Metadata](#metadata)
+    + [Spec](#spec)
+    + [Status](#status)
+- [Status Signalling](#status-signalling)
+- [Listing Resources](#listing-resources)
+- [Detailed Resource Types - v1beta1](#detailed-resource-types---v1beta1)
+  * [`ArrayOrString`](#arrayorstring)
+  * [`ContainerStateRunning`](#containerstaterunning)
+  * [`ContainerStateWaiting`](#containerstatewaiting)
+  * [`ContainerStateTerminated`](#containerstateterminated)
+  * [`EnvVar`](#envvar)
+  * [`Param`](#param)
+  * [`ParamSpec`](#paramspec)
+  * [`Step`](#step)
+  * [`StepState`](#stepstate)
+  * [`TaskResult`](#taskresult)
+  * [`TaskRunResult`](#taskrunresult)
+  * [`TaskSpec`](#taskspec)
+  * [`WorkspaceBinding`](#workspacebinding)
+  * [`WorkspaceDeclaration`](#workspacedeclaration)
+<!-- /toc -->
+
+## Abstract
+
+The Tekton Pipelines platform provides common abstractions for describing and executing container-based, run-to-completion workflows, typipcally in service of CI/CD scenarios. This document describes the structure, lifecycle and management of Tekton resources in the context of the [Kubernetes Resource Model](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/resource-management.md). An understanding of the Kubernetes API interface and the capabilities of [Kubernetes Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) is assumed.
+
+This document does not define the [runtime contract](https://tekton.dev/docs/pipelines/container-contract/) nor prescribe specific implementations of supporting services such as access control, observability, or resource management.
+
+This document makes reference in a few places to different profiles for Tekton installations. A profile in this context is a set of operations, resources, and fields that are accessible to a developer interacting with a Tekton installation. Currently, only a single (minimal) profile for Tekton Pipelines is defined, but additional profiles may be defined in the future to standardize advanced functionality. A minimal profile is one that implements all of the “MUST”, “MUST NOT”, and “REQUIRED” conditions of this document.
+
+## Background
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “NOT RECOMMENDED”, “MAY”, and “OPTIONAL” are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+There is no formal specification of the Kubernetes API and Resource Model. This document assumes Kubernetes 1.16 behavior; this behavior will typically be supported by many future Kubernetes versions. Additionally, this document may reference specific core Kubernetes resources; these references may be illustrative (i.e. an implementation on Kubernetes) or descriptive (i.e. this Kubernetes resource MUST be exposed). References to these core Kubernetes resources will be annotated as either illustrative or descriptive.
+
+## Modifying This Specification
+
+This spec is a living document, meaning new resources and fields may be added, and may transition from being OPTIONAL to RECOMMENDED to REQUIRED over time. In general a resource or field should not be added as REQUIRED directly, as this may cause unsuspecting previously-conformant implementations to suddenly no longer be conformant. These should be first OPTIONAL or RECOMMENDED, then change to be REQUIRED once a survey of conformant implementations indicates that doing so will not cause undue burden on any implementation.
+
+## Resource Overview
+
+The Tekton Pipelines API provides a set of API resources to manage run-to-completion workflows. Those are expressed as [Kubernetes Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/)
+
+### `TaskRun`
+
+A `TaskRun` represents an instantiation of a single execution of a `Task`. It can describe the steps of the Task directly.
+
+Its HTTP API endpoint is `/apis/tekton.dev/v1beta1/[parent]/taskruns`,  where  `[parent]` can refer to any string identifying a grouping of `TaskRun`s.
+
+For example, in the Kubernetes implementation `[parent]` refers to a Kubernetes namespace. Other implementations could interpret the string differently, including enabling hierarchies of resources containing `TaskRun`s, for example the API endpoint `/apis/tekton.dev/v1beta1/folders/my-folder/project/my-project/taskruns` could represent a hierarchy of "projects" that own `TaskRun`s.
+
+| HTTP Verb                 | Requirement      |
+|---------------------------|------------------|
+| Create (POST)             | REQUIRED         |
+| Patch (PATCH)*            | RECOMMENDED      |
+| Replace (PUT)**           | RECOMMENDED      |
+| Delete (DELETE)           | OPTIONAL         |
+| Read (GET)                | REQUIRED         |
+| List (GET)                | REQUIRED         |
+| Watch (GET)               | OPTIONAL         |
+| DeleteCollection (DELETE) | OPTIONAL         |
+
+\* Kubernetes only allows JSON merge patch for CRD types. It is recommended that if allowed, at least JSON Merge patch be made available. [JSON Merge Patch Spec (RFC 7386)](https://tools.ietf.org/html/rfc7386)
+
+\** NB: Support for cancellation depends on `Replace`.
+
+## Detailed Resources - v1beta1
+
+The following schema defines a set of REQUIRED or RECOMMENDED resource fields on the Tekton resource types. Whether a field is REQUIRED or RECOMMENDED is denoted in the "Requirement" column.
+
+Additional fields MAY be provided by particular implementations, however it is expected that most extension will be accomplished via the `metadata.labels` and `metadata.annotations` fields, as Tekton implementations MAY validate supplied resources against these fields and refuse resources which specify unknown fields.
+
+Tekton implementations MUST NOT require `spec` fields outside this implementation; to do so would break interoperability between such implementations and implementations which implement validation of field names.
+
+**NB:** All fields and resources not listed below are assumed to be **OPTIONAL**, not RECOMMENDED or REQUIRED. For example, at this time, support for `PipelineRun`s and for CRUD operations on `Task`s or `Pipeline`s is **OPTIONAL**.
+
+### `TaskRun`
+
+#### Metadata
+
+Standard Kubernetes [meta.v1/ObjectMeta](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#objectmeta-v1-meta) resource.
+
+| Field Name           | Field Type         | Requirement |
+|----------------------|--------------------|-------------|
+| `name`               | string             | REQUIRED    |
+| `labels`             | map<string,string> | REQUIRED    |
+| `annotations`        | map<string,string> | REQUIRED    |
+| `creationTimestamp`* | string             | REQUIRED    |
+| `uid`                | string             | RECOMMENDED |
+| `resourceVersion`    | string             | RECOMMENDED |
+| `generation`         | int64              | RECOMMENDED |
+| `namespace`          | string             | RECOMMENDED |
+| `generateName`**     | string             | RECOMMENDED |
+
+\* `creationTimestamp` MUST be populated by the implementation, in [RFC3339](https://tools.ietf.org/html/rfc3339).
+
+** If `generateName` is supported by the implementation, when it is specified at creation, it MUST be prepended to a random string and set as the `name`, and not set on the subsequent response.
+
+#### Spec
+
+| Field Name            | Field Type           | Requirement |
+|-----------------------|----------------------|-------------|
+| `params`              | `[]Param`            | REQUIRED    |
+| `taskSpec`            | `TaskSpec`           | REQUIRED    |
+| `workspaces`          | `[]WorkspaceBinding` | REQUIRED    |
+| `status`              | Enum:<br>- `""` (default)<br>- `"TaskRunCancelled"` | RECOMMENDED |
+| `timeout`             | string (duration)    | RECOMMENDED |
+| `serviceAccountName`^ | string               | RECOMMENDED |
+
+^ In the Kubernetes implementation, `serviceAccountName` refers to a Kubernetes `ServiceAccount` resource that is assumed to exist in the same namespace. Other implementations MAY interpret this string differently, and impose other requirements on specified values.
+
+#### Status
+
+| Field Name            | Field Type             | Requirement |
+|-----------------------|------------------------|-------------|
+| `conditions`          | see [#error-signaling] | REQUIRED    |
+| `startTime`           | string                 | REQUIRED    |
+| `completionTime`*     | string                 | REQUIRED    |
+| `steps`               | `[]StepState`          | REQUIRED    |
+| `observedGeneration`  | int64                  | RECOMMENDED |
+
+\* `startTime` and `completionTime` MUST be populated by the implementation, in [RFC3339](https://tools.ietf.org/html/rfc3339).
+
+## Status Signalling
+
+The Tekton Pipelines API uses the [Kubernetes Conditions convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) to communicate status and errors to the user.
+
+`TaskRun`'s `status` field MUST have a `conditions` field, which must be a list of `Condition` objects of the following form:
+
+| Field                 | Type   | Requirement |
+|-----------------------|--------|-------------|
+| `type`                | string | REQUIRED    |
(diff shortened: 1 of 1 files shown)


docs/api-spec.md
```diff
@@ -0,0 +1,322 @@
...
+
+This document considers two users of a given Tekton Pipelines environment, and is particularly concerned with the expectations of developers (and language and tooling developers, by extension) deploying applications to the environment.
+
+* **Developers** write code against which Tekton Pipelines executes workflows. Developers may also author these workflow configurations.
+* **Operators** (also known as platform providers) provision the compute resources and manage the software configuration of Tekton Pipelines and the underlying abstractions (for example: Linux, Kubernetes, etc).
+
+## Modifying This Specification
+
+This spec is a living document, meaning new resources and fields may be added, and may transition from being OPTIONAL to RECOMMENDED to REQUIRED over time. In general a resource or field should not be added as REQUIRED directly, as this may cause unsuspecting previously-conformant implementations to suddenly no longer be conformant. These should be first OPTIONAL or RECOMMENDED, then change to be REQUIRED once a survey of conformant implementations indicates that doing so will not cause undue burden on any implementation.
```
dlorenc, line 54: Do we expect to have any expected lifetime or mandatory wait period as features move through here? I think wed' want to give platform implementors some time to add support for things so they don't fall out of compliance.

ImJasonH: I think in practice we'll end up having one, maybe even enforcing and documenting it, but it seems unnecessary at this early stage.

When we reach the point of first making an OPTIONAL surface REQUIRED we'll reach out to platform implementors and get their feedback about how soon they can support it, and work out a timeline. Over time I expect we'll find it normally takes N weeks/months/releases to achieve this, and we can document that as the default upgrade timeline.

docs/api-spec.md
```diff
@@ -0,0 +1,322 @@
...
+| Field                 | Type   |
+|-----------------------|--------|
+| `type`                | string |
+| `status`              | Enum:<br>- `"True"`<br>- `"False"`<br>- `"Unknown"` (default) |
+| `reason`              | string |
+| `message`             | string |
+| `severity`            | Enum:<br>- `""` (default)<br>- `"Warning"`<br>- `"Info"` |
+| `lastTransitionTime`* | string |
+
+\* If `lastTransitionTime` is populated by the implementation, it must be in [RFC3339](https://tools.ietf.org/html/rfc3339). Reporting `lastTransitionTime` is OPTIONAL.
```
bobcatfish, line 151: should we list REQUIRED, etc for the fields above as well? (e.g. i have no idea what severity is about, seems less important)

docs/api-spec.md
```diff
@@ -0,0 +1,322 @@
...
+|--------------|------------|-------------|
+| `name`       | string     | REQUIRED    |
+| `image`      | string     | REQUIRED    |
+| `args`       | []string   | REQUIRED    |
+| `command`    | []string   | REQUIRED    |
+| `workingDir` | []string   | REQUIRED    |
+| `env`        | `[]EnvVar` | REQUIRED    |
+| `script`     | string     | REQUIRED    |
+
+**NB:** All other fields inherited from the [core.v1/Container](https://godoc.org/k8s.io/api/core/v1#Container) type supported by the Kubernetes implementation are **OPTIONAL** for the purposes of this spec.
```
bobcatfish, line 260: nice, thanks for whittling this down! 

is StepState similar (i.e. based on pod status) or is this something we created ourselves?

docs/api-spec.md
```diff
@@ -0,0 +1,322 @@
...
+| `name`        | string     | REQUIRED    |
+| `description` | string     | REQUIRED    |
+| `type`        | Enum: <br>- `"string"` (default) <br>- `"array"` | REQUIRED |
+| `default`     | `ArrayOrString` | REQUIRED |
+
+### `Step`
+
+| Field Name   | Field Type | Requirement |
+|--------------|------------|-------------|
+| `name`       | string     | REQUIRED    |
```
bobcatfish, line 252: am i crazy or is it possible to have a step without a name

docs/api-spec.md
```diff
@@ -0,0 +1,322 @@
...
+### `StepState`
+
+| Field Name    | Field Type                 | Requirement |
+|---------------|----------------------------|-------------|
+| `name`        | string                     | REQUIRED    |
+| `imageID`     | string                     | REQUIRED    |
+| `waiting`*    | `ContainerStateWaiting`    | REQUIRED    |
+| `running`*    | `ContainerStateRunning`    | REQUIRED    |
+| `terminated`* | `ContainerStateTerminated` | REQUIRED    |
+| `taskResults` | `TaskRunResult`            | REQUIRED    |
```
bobcatfish, line 271: []TaskRunResult?

bobcatfish: ( dont let this fool you into thinking i was this thorough with everything i looked at XD )

docs/api-spec.md
```diff
@@ -0,0 +1,322 @@
...
+| `description` | string                   | REQUIRED    |
+
+### `WorkspaceBinding`
+
+| Field Name | Field Type   | Requirement |
+|------------|--------------|-------------|
+| `name`     | string       | REQUIRED    |
+| `emptyDir` | empty struct | REQUIRED    |
+
+**NB:** All other Workspace types supported by the Kubernetes implementation are **OPTIONAL** for the purposes of this spec.
```
bobcatfish, line 307: do we want to say anything more about this? in order to support tasks in a pipeline properly, the system would need to support _something_ besides emptyDir

bobcatfish (commented):
metaquestion: what do you see as the process to add more fields here going forward? e.g. how to add when expressions into here. "Modifying This Specification" starts to talk about it but i wonder if it'll make sense to create a process around this to guide contributors

overall looks good, just lots of minor feedback from me!

docs/api-spec.md
```diff
@@ -0,0 +1,322 @@
...
+  * [`TaskSpec`](#taskspec)
+  * [`WorkspaceBinding`](#workspacebinding)
+  * [`WorkspaceDeclaration`](#workspacedeclaration)
+<!-- /toc -->
+
+## Abstract
+
+The Tekton Pipelines platform provides common abstractions for describing and executing container-based, run-to-completion workflows, typipcally in service of CI/CD scenarios. This document describes the structure, lifecycle and management of Knative resources in the context of the [Kubernetes Resource Model](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/resource-management.md). An understanding of the Kubernetes API interface and the capabilities of [Kubernetes Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) is assumed.
+
+This document does not define the [runtime contract](https://tekton.dev/docs/pipelines/container-contract/) nor prescribe specific implementations of supporting services such as access control, observability, or resource management.
```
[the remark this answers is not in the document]
ImJasonH: I don't mind it being in a separate document. It's not currently phrased in terms of "platform implementations must satisfy this contract" so much as "users should expect this contract", but maybe that's just fine since the vast majority of readers will be users, and very few will be prospective platform providers.

In the meantime it's probably enough to just say "this isn't the runtime contract, that's over there" and when/if we get any confusion we can resolve it. Launch and iterate 😄

docs/api-spec.md
```diff
@@ -0,0 +1,322 @@
...
+
+## Background
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “NOT RECOMMENDED”, “MAY”, and “OPTIONAL” are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+There is no formal specification of the Kubernetes API and Resource Model. This document assumes Kubernetes 1.13 behavior; this behavior will typically be supported by many future Kubernetes versions. Additionally, this document may reference specific core Kubernetes resources; these references may be illustrative (i.e. an implementation on Kubernetes) or descriptive (i.e. this Kubernetes resource MUST be exposed). References to these core Kubernetes resources will be annotated as either illustrative or descriptive.
```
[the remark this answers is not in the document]
ImJasonH: Oh probably just because the original draft was written when 1.13 was the minimum required. 👴 

AFAIK no relevant fields have been removed/changed since 1.13, so it's technically still true. 🤷

docs/api-spec.md
```diff
@@ -0,0 +1,322 @@
...
+
+## Resource Overview
+
+The Tekton Pipelines API provides a set of [Kubernetes Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) to manage run-to-completion workflows.
+
+### `TaskRun`
+
+A `TaskRun` represents an instantiation of a single execution of a `Task`. It can describe the steps of the Task directly.
+
+Its HTTP API endpoint is `/apis/tekton.dev/v1beta1/[parent]/taskruns`
```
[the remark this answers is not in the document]
ImJasonH: Good question, one I should answer better in the doc 😄 

Basically, in the scope of the API spec, it's an arbitrary string that represents a container of taskruns. In the k8s implementation it'll be a namespace, but we should be meaningfully vague here to allow other platforms to collect resources however they want, including enabling hierarchies (e.g., something like `/apis/tekton.dev/v1beta1/projects/my-project/locations/us-east4/taskruns` as a total hypothetical)

docs/api-spec.md
ImJasonH: Yeah StepState has the same problem, but it's slightly better since there at least aren't tons of extra inherited fields. But K8s could add more any time they want, so we should spell it out defensively.

docs/api-spec.md
ImJasonH: This doc is saying it's "required" that platform implementations _support_ naming steps, not that the user's request _must_ specify one.

If there's a good clear way to describe that somewhere I'd love to clear that up for future readers.

docs/api-spec.md
ImJasonH: Yeah, this version of the doc is just to get a minimal "anything" spec out there, we can add more workspace types that platforms must support as we go.

It is _conceivable_ that a platform would only support emptyDir, though that would be a pretty serious limitation. So emptyDir should probably be the only REQUIRED option.

docs/api-spec.md
ImJasonH: Had me fooled! 😄

ImJasonH:
Ping @vdemeester @afrittoli or anybody else who may have thoughts?

docs/api-spec.md
```diff
@@ -0,0 +1,322 @@
...
+
+There is no formal specification of the Kubernetes API and Resource Model. This document assumes Kubernetes 1.13 behavior; this behavior will typically be supported by many future Kubernetes versions. Additionally, this document may reference specific core Kubernetes resources; these references may be illustrative (i.e. an implementation on Kubernetes) or descriptive (i.e. this Kubernetes resource MUST be exposed). References to these core Kubernetes resources will be annotated as either illustrative or descriptive.
+
+This document considers two users of a given Tekton Pipelines environment, and is particularly concerned with the expectations of developers (and language and tooling developers, by extension) deploying applications to the environment.
+
+* **Developers** write code against which Tekton Pipelines executes workflows. Developers may also author these workflow configurations.
+* **Operators** (also known as platform providers) provision the compute resources and manage the software configuration of Tekton Pipelines and the underlying abstractions (for example: Linux, Kubernetes, etc).
```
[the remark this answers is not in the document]
afrittoli: Right now they are grouped under "Developers" (Developers may also author these workflow configurations.) but they might be called out as workflow authors.
We do not mention these groups anywhere in the document though.  What is the purpose of having these groups here in the API spec?

docs/api-spec.md
```diff
@@ -0,0 +1,324 @@
...
+
+## Modifying This Specification
+
+This spec is a living document, meaning new resources and fields may be added, and may transition from being OPTIONAL to RECOMMENDED to REQUIRED over time. In general a resource or field should not be added as REQUIRED directly, as this may cause unsuspecting previously-conformant implementations to suddenly no longer be conformant. These should be first OPTIONAL or RECOMMENDED, then change to be REQUIRED once a survey of conformant implementations indicates that doing so will not cause undue burden on any implementation.
```
afrittoli, line 54: Conformance could be associated to version numbers. 

We do not have a Tekton wide version number as such; I think it's something that we might introduce in future, at least for the pipeline + triggers couple. That would allow us to have conformance associated to an overall Tekton version - which would also be helpful for operator.

In any case having to introduce new fields as OPTIONAL matches the model of adding features behind features flags that we should follow for the API.

docs/api-spec.md
```diff
@@ -0,0 +1,324 @@
...
+
+## Resource Overview
+
+The Tekton Pipelines API provides a set of [Kubernetes Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) to manage run-to-completion workflows.
```
afrittoli, line 58: NIT: it would be nice to phrase this from an API spec POV first and then detail the K8s specific implementation. E.g. The Tekton API (or a Tekton conformant API) provides a set of API resources to manage run-to-completion workflows.
Those are implemented by Tekton as [Kubernetes Custom Resources]...

docs/api-spec.md
```diff
@@ -0,0 +1,324 @@
...
+
+### `TaskRun`
+
+A `TaskRun` represents an instantiation of a single execution of a `Task`. It can describe the steps of the Task directly.
+
+Its HTTP API endpoint is `/apis/tekton.dev/v1beta1/[parent]/taskruns`,  where  `[parent]` can refer to any string identifying a grouping of `TaskRun`s.
+
+For example, in the Kubernetes implementation `[parent]` refers to a Kubernetes namespace. Other implementations could interpret the string differently, including enabling hierarchies of resources containing `TaskRun`s, for example the API endpoint `/apis/tekton.dev/v1beta1/folders/my-folder/project/my-project/taskruns` could represent a hierarchy of "projects" that own `TaskRun`s.
```
afrittoli, line 66: I'm a bit confused about the "Kubernetes implementation" terminology.
My understanding is that this documents abstracts the Tekton API with the aim to define what a "Tekton conform" API should look like.  
The reference implementation is the Tekton implementation, and it's based on k8s. There may be other implementation in future, which will not be called Tekton, and they may or may not be k8s based.
One might decide to create another k8s version of Tekton  written in Rust for instance (probably not).

docs/api-spec.md
```diff
@@ -0,0 +1,324 @@
...
+
+| HTTP Verb                 | Developer Access |
+|---------------------------|------------------|
+| Create (POST)             | REQUIRED         |
+| Patch (PATCH)*            | RECOMMENDED      |
+| Replace (PUT)             | RECOMMENDED      |
```
afrittoli, line 72: `Cancel` relies on PATCH/REPLACE - so this excludes `cancel` from the functionality required for conformance. It's probably OK, but I wanted to highlight it anyways.

docs/api-spec.md
```diff
@@ -0,0 +1,324 @@
...
+
+## Detailed Resources - v1beta1
+
+The following schema defines a set of REQUIRED or RECOMMENDED resource fields on the Tekton resource types. Whether a field is REQUIRED or RECOMMENDED is denoted in the "Requirement" column.
+
+Additional fields MAY be provided by particular implementations, however it is expected that most extension will be accomplished via the `metadata.labels` and `metadata.annotations` fields, as Tekton implementations MAY validate supplied resources against these fields and refuse resources which specify unknown fields.
+
+Tekton implementations MUST NOT require `spec` fields outside this implementation; to do so would break interoperability between such implementations and implementations which implement validation of field names.
+
+**NB:** All fields and resources not listed below are assumed to be **OPTIONAL**, not RECOMMENDED or REQUIRED. For example, at this time, support for `PipelineRun`s and for CRUD operations on `Task`s or `Pipeline`s is **OPTIONAL**.
```
afrittoli, line 89: Why are `PipelineRuns` not included? They are part of the beta API and I see for reason for them to be excluded. 
I also have mixed feeling about having `Tasks` as OPTIONAL.
While it is possible to re-use Tasks (from the public or a private catalog) by injecting them on the fly in a `TaskRun`'s `taskSpec`, doing so would require a non-trivial amount of work for a Tekton use to move some of their workflows to such Tekton conformant platform.

docs/api-spec.md
```diff
@@ -0,0 +1,324 @@
...
+\* `creationTimestamp` MUST be populated by the implementation, in [RFC3339](https://tools.ietf.org/html/rfc3339).
+
+** If `generateName` is supported by the implementation, when it is specified at creation, it MUST be prepended to a random string and set as the `name`, and not set on the subsequent response.
+
+#### Spec
+
+| Field Name            | Field Type           | Requirement |
+|-----------------------|----------------------|-------------|
+| `params`              | `[]Param`            | REQUIRED    |
+| `taskSpec`            | `TaskSpec`           | REQUIRED    |
```
afrittoli, line 118: This means that support for `taskSpec` is required for conformance, right?

docs/api-spec.md
```diff
@@ -0,0 +1,324 @@
...
+| `workspaces`          | `[]WorkspaceBinding` | REQUIRED    |
+| `status`              | Enum:<br>- `""` (default)<br>- `"TaskRunCancelled"` | RECOMMENDED |
+| `timeout`             | string (duration)    | RECOMMENDED |
```
afrittoli, line 121: I have mixed feelings about this. Timeouts are key to the usability of the platform... if you want to keep this optional we could mention that the platform should at least provide some timeout mechanism (which would not be workflow specific).

afrittoli (commented):
Thank you for starting this!

Having everything not mentioned defaulting to OPTIONAL makes sense, it makes it harder to review though as it is not obvious what we are excluding :P I think it should be fine to iterate on this and add things incrementally as we discover them?

docs/api-spec.md
```diff
@@ -0,0 +1,324 @@
...
+  * [`TaskResult`](#taskresult)
+  * [`TaskRunResult`](#taskrunresult)
+  * [`TaskSpec`](#taskspec)
+  * [`WorkspaceBinding`](#workspacebinding)
+  * [`WorkspaceDeclaration`](#workspacedeclaration)
+<!-- /toc -->
+
+## Abstract
+
+The Tekton Pipelines platform provides common abstractions for describing and executing container-based, run-to-completion workflows, typipcally in service of CI/CD scenarios. This document describes the structure, lifecycle and management of Knative resources in the context of the [Kubernetes Resource Model](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/resource-management.md). An understanding of the Kubernetes API interface and the capabilities of [Kubernetes Custom Resources](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) is assumed.
```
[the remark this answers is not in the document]
ImJasonH: 😊

docs/api-spec.md
ImJasonH: This is a really interesting concept I'd like to think on more. I do like the idea of an implementation being able to say it's "conformant as of Tekton 0.14", but that also adds a lot of complexity, for potential implementors, for conformance tests, for users just trying to build on Tekton.

How would you feel about punting on this until after this initial spec lands, and circling back?

docs/api-spec.md
ImJasonH: ❤️

docs/api-spec.md
ImJasonH: That's useful, thanks. Noted.

docs/api-spec.md
ImJasonH: PipelineRuns are excluded purely because this is the minimal set I thought we could get away with. It's _definitely_ coming, but this PR is already huge :)

docs/api-spec.md
ImJasonH: That's correct. Implementations _can_ support task references (including someday Tekton Bundles!), but they _must_ support inline-defined task specs.

docs/api-spec.md
ImJasonH: I agree! I wanted to have the initial spec be as minimally-scoped as possible, and timeout was on the bubble.

I could add a note that platforms should consider enforcing timeouts in some way, but honestly that kinda seems like their business. If they don't want to support timeout, I'm not sure we need to make them? I'm on the fence. 🤷‍♂️

docs/api-spec.md
ImJasonH: Yeah, I agree that's a confusing distinction. Words are hard 😅 

I'd like to avoid calling _anything_ "_the_ Tekton implementation", since the rest of this doc tries to draw the distinction that Tekton is primarily an _API_, with potentially many implementations.

I'd love to find a good term for "the implementation at https://github.com/tektoncd/pipeline", I've been referring to this as "the Kubernetes implementation", but you're right, that's not necessarily a unique identifier. Suggestions welcome, and we can also punt on finding the best term and just merge this and update it later when we find it.

vdemeester (approved):
Let's get this in and iterate over it :angel: 
/meow

PR #3135, opened 2020-08-25
Conditions Beta: When Expressions

jerop:
Adding `WhenExpressions` used to efficiently specify guarded execution
of `Tasks`, without spinning up new pods. We use `WhenExpressions` to
avoid adding an opinionated and complex expression language to the Tekton
API to ensure Tekton can be supported by as many systems as possible.
Further details about the design are in [Conditions Beta TEP](https://github.com/tektoncd/community/blob/master/teps/0007-conditions-beta.md).

The components of `WhenExpressions` are `Input`, `Operator`
and `Values`:
- `Input` is the input for the `Guard` checking which can be static
inputs or variables, such as `Parameters` or `Results`.
- `Operator` represents an `Input`'s relationship to a set of `Values`.
`Operators` we will use in `WhenExpressions` are `In` and `NotIn`.
- `Values` is an array of string values. The `Values` array must be
non-empty. It can contain static values or variables (`Parameters` or
`Results`).

The declared `WhenExpressions` are evaluated before the `Task` is run.
If all the `WhenExpressions` evaluate to `True`, the `Task` is run.
If any of the `WhenExpressions` evaluate to `False`, the `Task` is
skipped.

# Changes

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
- To run a Task only when certain criteria are met, it is now possible to guard task execution using
the when field, which allows you to list a series of references to WhenExpressions that contain
an Input, an Operator and Values. The valid Operators are in and notin. The WhenExpressions are ANDed.
- Conditions are deprecated, use WhenExpressions instead.
 ```

### CODE REVIEW

File: pkg/apis/pipeline/v1beta1/pipeline_validation.go [hunk 1]
```diff
@@ -488,6 +509,9 @@ func validateFinalTasks(finalTasks []PipelineTask) *apis.FieldError {
 		if len(f.Conditions) != 0 {
 			return apis.ErrInvalidValue(fmt.Sprintf("no conditions allowed under spec.finally, final task %s has conditions specified", f.Name), "spec.finally")
 		}
+		if len(f.WhenExpressions) != 0 {
```
jerop, line 512: @pritidesai confirming that we don't want to have when expressions in finally at this point...

pritidesai: Yes, no `when` expression in finally yet 👍  `finally` does not support/allow specifying conditions and will keep them as is until we find any use cases

pritidesai:
Sorry @jerop have been slow on reviewing your changes. Just started playing with them. Ran a simple PipelineRun with one task with `when` expression evaluating to `false` resulted in a TaskRun with `failed` similar to what happens when the condition fails for a task `failed` with `ConditionCheckFailed`.

```
kind: PipelineRun
metadata:
  name: pipelinerun-to-skip-task
spec:
  pipelineSpec:
    tasks:
      - name: skip-this-task
        when:
          - input: "foo"
            operator: notin
            values: ["foo"]
        taskSpec:
          steps:
            - name: echo
              image: ubuntu
              script: |
                echo "Good Night!"
```

```
$ tkn pr describe pipelinerun-to-skip-task
Name:        pipelinerun-to-skip-task
Labels:
 tekton.dev/pipeline=pipelinerun-to-skip-task

🌡️  Status

STARTED        DURATION    STATUS
1 minute ago   0 seconds   Succeeded(Completed)

🗂  Taskruns

 NAME                                              TASK NAME        STARTED   DURATION   STATUS
 ∙ pipelinerun-to-skip-task-skip-this-task-25tgt   skip-this-task   ---       ---        Failed(WhenExpressionsEvaluatedToFalse)
```

But the way users might start using `when` expressions would be to skip or ignore the task based on a param which was not possible before without writing conditions. For a ignore use case, the status `failed` might not make sense. Also, the task run which is listed here is just the `taskrun` object created as part of the `pipelinerun` status not actual `taskrun` which does not exist:

```
tkn tr describe pipelinerun-to-skip-task-skip-this-task-25tgt
Error: failed to get TaskRun pipelinerun-to-skip-task-skip-this-task-25tgt: taskruns.tekton.dev "pipelinerun-to-skip-task-skip-this-task-25tgt" not found
```

@bobcatfish how about changing this status `failed` to something different to signify that the task was `skipped`? thoughts? But then in that case, we have to also think about the tasks which are dependent on this skip task are also skipped but not part of the overall `pipelinerun` status and does not show up in the list of `Taskruns`.

jerop:
@pritidesai thanks for looking into this! currently, knative only provides `ConditionTrue`, `ConditionFalse`, and `ConditionUnknown` for `ConditionSucceeded`. Had gone with `ConditionFalse` but open to hearing arguments for using the other options. 

I've now updated it to our own status like this:
```yaml
prtrs.Status.SetCondition(&apis.Condition{
        Status:  corev1.ConditionStatus("Skipped"),
        Reason:  resources.ReasonWhenExpressionsEvaluatedToFalse,
        Message: fmt.Sprintf("When Expressions for Task %s in PipelineRun %s evaluated to False", rprt.TaskRunName, pr.Name),
}) 
```

That gives this result in your example:
```
$ kubectl describe pipelineruns.tekton.dev pipelinerun-to-skip-task
Name:         pipelinerun-to-skip-task
Namespace:    default
Labels:       tekton.dev/pipeline=pipelinerun-to-skip-task
API Version:  tekton.dev/v1beta1
Kind:         PipelineRun
Spec:
  Pipeline Spec:
    Tasks:
      Name:  skip-this-task
      Task Spec:
        Metadata:
        Steps:
          Image:  ubuntu
          Name:   echo
          Resources:
          Script:  echo "Good Night!"

      When:
        Input:     foo
        Operator:  notin
        Values:
          foo
  Timeout:  1h0m0s
Status:
  Completion Time:  2020-08-26T12:02:08Z
  Conditions:
    Last Transition Time:  2020-08-26T12:02:08Z
    Message:               Tasks Completed: 0 (Failed: 0, Cancelled 0), Skipped: 1
    Reason:                Completed
    Status:                True
    Type:                  Succeeded
  Pipeline Spec:
    Tasks:
      Name:  skip-this-task
      Task Spec:
        Metadata:
        Steps:
          Image:  ubuntu
          Name:   echo
          Resources:
          Script:  echo "Good Night!"

      When:
        Input:     foo
        Operator:  notin
        Values:
          foo
  Start Time:  2020-08-26T12:02:08Z
  Task Runs:
    Pipelinerun - To - Skip - Task - Skip - This - Task - Phpfz:
      Pipeline Task Name:  skip-this-task
      Status:
        Conditions:
          Last Transition Time:  2020-08-26T12:02:08Z
          Message:               When Expressions for Task pipelinerun-to-skip-task-skip-this-task-phpfz in PipelineRun pipelinerun-to-skip-task evaluated to False
          Reason:                WhenExpressionsEvaluatedToFalse
          Status:                Skipped
          Type:                  Succeeded
        Pod Name:                
      When Expressions Status:
        Evaluation Results:
          Expression:
            Input:     foo
            Operator:  notin
            Values:
              foo
Events:
  Type    Reason     Age    From         Message
  ----    ------     ----   ----         -------
  Normal  Started    3m32s  PipelineRun  
  Normal  Succeeded  3m32s  PipelineRun  Tasks Completed: 0 (Failed: 0, Cancelled 0), Skipped: 1
```

```
$ tkn pr describe pipelinerun-to-skip-task
Name:        pipelinerun-to-skip-task
Namespace:   default

🌡️  Status

STARTED          DURATION    STATUS
18 seconds ago   0 seconds   Succeeded(Completed)

📦 Resources

 No resources

⚓ Params

 No params

🗂  Taskruns

 NAME                                              TASK NAME        STARTED   DURATION   STATUS
 ∙ pipelinerun-to-skip-task-skip-this-task-phpfz   skip-this-task   ---       ---        (WhenExpressionsEvaluatedToFalse)
```
@pritidesai @bobcatfish what do you think?

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -0,0 +1,139 @@
...
+	for _, we := range wes {
+		if isTrue := we.isTrue(); !isTrue {
+			return false
+		}
+	}
+	return true
+}
+
+// HaveVariables indicates whether When Expressions have variables in the Inputs or Values.
+func (wes WhenExpressions) HaveVariables() bool {
```
bobcatfish, line 93: i think we usually use pointer recievers (https://tour.golang.org/methods/8)

i guess it has some advantages to not do that, cuz we're indicating clearly we're not mutating

looking at https://stackoverflow.com/questions/27775376/value-receiver-vs-pointer-receiver and https://flaviocopes.com/golang-methods-receivers/ im wondering if we should be using value recievers more often than we do. i thought there was some kind of problem with not using pointer receivers and using interfaces but i cant remember what it is 😓

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -0,0 +1,139 @@
...
+// ReplaceWhenExpressionsVariables interpolates variables in When Expressions.
+func (wes WhenExpressions) ReplaceWhenExpressionsVariables(replacements map[string]string) WhenExpressions {
+	var replaced []WhenExpression
+	for _, we := range wes {
+		replaced = append(replaced, we.applyReplacements(replacements))
+	}
+	return replaced
+}
+
+// WhenExpressionEvaluationResult is the result from evaluating a When Expression.
```
bobcatfish, line 111: a general comment about docstrings: its worth including different information than what you get from the name of the thing you are documenting. e.g. in this case, WhenExpressionEvaluationResult kinda already contains the info "is the result from evaluating a when expression"

you could use this docstring to go into more detail, e.g. what does "evaluating" mean exactly? what is the result for?

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -0,0 +1,139 @@
...
+	// IsTrue is a boolean from evaluating the Expression.
+	IsTrue bool `json:"result,omitempty"`
+}
+
+// WhenExpressionsEvaluationStatus are the results from evaluating When Expressions.
+type WhenExpressionsEvaluationStatus struct {
+	// EvaluationResults are the results from evaluating When Expressions.
+	EvaluationResults []*WhenExpressionEvaluationResult `json:"evaluationResults,omitempty"`
+	// Executed is a boolean that represents an AND of all results from evaluating When Expressions.
+	Executed bool `json:"execute,omitempty"`
```
bobcatfish, line 124: are you certain we need Executed in addition to the IsTrue values in EvaluationResults?

i also think this might duplicate information from the Condition, which i assume will also contain information about the fact that this was skipped and why?

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -0,0 +1,139 @@
...
+	Input string `json:"input,omitempty"`
+	// Operator that represents an Input's relationship to the values
+	Operator selection.Operator `json:"operator,omitempty"`
+	// Values is an array of strings, which is compared against the input, for guard checking
+	// It must be non-empty
+	Values []string `json:"values,omitempty"`
+}
+
+// GetVarSubstitutionExpressionsForWhenExpression extracts all the values between "$(" and ")"" for a when expression
+func (we *WhenExpression) GetVarSubstitutionExpressionsForWhenExpression() ([]string, bool) {
```
bobcatfish, line 36: if this is happening on a WhenExpression object you can pretty safely remove WhenExpression from the name, e.g. WhenExpression.GetVarSubstitutionExpressions vs WhenExpression.GetVarSubstitutionExpressionsForWhenExpression

maybe even shorten to GetSubstitutionExpressions? or GetSubstitutions?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -159,6 +161,15 @@ func (t ResolvedPipelineRunTask) IsSkipped(state PipelineRunState, d *dag.Graph)
 		}
 	}
 
+	// Check if the when expressions have been
+	if len(t.PipelineTask.WhenExpressions) > 0 {
+		if haveVariables := t.PipelineTask.WhenExpressions.HaveVariables(); !haveVariables {
+			if areTrue := t.PipelineTask.WhenExpressions.AreTrue(); !areTrue {
+				return true
```
bobcatfish, line 168: this will be slightly clearer as:

```go
if ! t.PipelineTask.WhenExpressions.AreTrue() {
```

bobcatfish: I'm also really tempted to suggest a different name for AreTrue, something like: `ShouldExecute` or `AreSatisfied`, (maybe that second one isnt any better, i couldnt think of much besides ShouldExecute XD)

they're already called "whenexpressions" so the "when" part kind indicates they are for something specific, i.e. determining WHEN to execute something, so i think it makes sense to name their functions a bit more after what they're for vs. the details of what they do

i.e. maybe AreTrue does check if all the expressions are true, but is it for? its for indicating if something should execute

so then this code could be:

```
if ! t.PipelineTask.WhenExpressions.ShouldExecute() {
```

oh oh oh or here's another idea, invert the logic and make the function `BlockExecution` or `ShouldBlock` or `Blocks`, then you' get something like:

```
if t.PipelineTask.WhenExpressions.Blocks() {
```

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
```diff
@@ -72,6 +72,14 @@ func extractResultRefsForParam(pipelineRunState PipelineRunState, param v1beta1.
 	return nil, nil
 }
 
+func extractResultRefsForWhenExpression(pipelineRunState PipelineRunState, whenExpression v1beta1.WhenExpression) (ResolvedResultRefs, error) {
+	expressions, ok := whenExpression.GetVarSubstitutionExpressionsForWhenExpression()
+	if ok {
+		return extractResultRefs(expressions, pipelineRunState)
+	}
+	return nil, nil
+}
```
bobcatfish, line 81: this is kind of a nit pick but this doesnt seem worth an entire function to me, the code might be clearer if you called GetVarSubstitutionExpressionsForWhenExpression(), what do you think?

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
```diff
@@ -173,6 +187,21 @@ func convertParams(params []v1beta1.Param, pipelineRunState PipelineRunState, na
 	return resolvedParams, nil
 }
 
+// convertWhenExpressions converts all When Expressions of the resolved pipeline run task
```
bobcatfish, line 190: we usually dont require docstrings for unexported functions but if we are going to have one here can you explain what "converts" means?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -835,6 +836,72 @@ func TestUpdateTaskRunsState(t *testing.T) {
 
 }
 
+func TestUpdateTaskRunsWithWhenExpressions(t *testing.T) {
+	// TestUpdateTaskRunsWithWhenExpressions runs "getTaskRunsStatus" and verifies how it updates a PipelineRun status
+	// from a TaskRun with When Expressionsassociated to the PipelineRun
```
bobcatfish, line 841: i think this would make sense as a docstring on the function? or maybe this is a convention other tests have used? kinda confuses me tho, doesn't seem like the right style for go

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -835,6 +836,72 @@ func TestUpdateTaskRunsState(t *testing.T) {
...
+	pipelineRunState := resources.PipelineRunState{{
+		PipelineTask: &pipelineTask,
+		TaskRunName:  "test-pipeline-run-success-unit-test-1",
+		TaskRun:      nil,
+		ResolvedTaskResources: &taskrunresources.ResolvedTaskResources{
+			TaskSpec: &task.Spec,
+		},
+	}}
+	pr.Status.InitializeConditions()
+	status := getTaskRunsStatus(pr, pipelineRunState)
```
bobcatfish, line 899: hmmm interesting that you're testing getTaskRunsStatus directly (cuz it's unexported)

in this case i wonder about calling Reconcile explicitly in the test?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -835,6 +836,72 @@ func TestUpdateTaskRunsState(t *testing.T) {
...
+	// TestUpdateTaskRunsWithWhenExpressions runs "getTaskRunsStatus" and verifies how it updates a PipelineRun status
+	// from a TaskRun with When Expressionsassociated to the PipelineRun
+	pr := tb.PipelineRun("test-pipeline-run", tb.PipelineRunNamespace("foo"), tb.PipelineRunSpec("test-pipeline"))
+	pipelineTask := v1beta1.PipelineTask{
+		Name:    "unit-test-1",
+		TaskRef: &v1beta1.TaskRef{Name: "unit-test-task"},
+		WhenExpressions: []v1beta1.WhenExpression{{
+			Input:    "foo",
+			Operator: selection.NotIn,
+			Values:   []string{"foo", "bar"},
```
bobcatfish, line 849: what do you think about including some variable replacement here as well to get the full picture?

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -356,30 +360,47 @@ func validatePipelineParameterVariables(tasks []PipelineTask, params []ParamSpec
...
+	}
+	return nil
+}
+
+func validatePipelineArrayVariable(name, value, prefix string, stringVars sets.String, arrayVars sets.String) *apis.FieldError {
+	if err := validatePipelineVariable(name, value, prefix, stringVars); err != nil {
+		return err
+	}
+	if err := validatePipelineArraysIsolated(name, value, prefix, arrayVars); err != nil {
+		return err
```
bobcatfish, line 403: ooo are these refactors that are unrelated to your changes? if so i highly recommend submitting them in a separate PR

bobcatfish:
@pritidesai @jerop that's a good point about "failed". 

one option would be implementing this PR with the same behavior as we use for existing Conditions, and then separately deciding how we want to handle this in the future? that would mean the potential for a backwards incompatible change though.

i dont want to change the behavior of our existing condition - one reason we implemented it the way we did was to satisfy knative duck typing

if anything, id like to see the addition of a new condition specifically for skipped. and perhaps in that case we do not include ConditionSucceeded at all? the fact that we have a few options here makes me want to suggest that we separate that discussion from this PR if we can - i could also see making a separate TEP about it to make sure we explore the options well

pkg/apis/pipeline/v1beta1/when_types.go
jerop: WhenExpressions is a slice, and my understanding is that slices are a pointer already so I don't need to use a pointer to the slice...https://golang.org/doc/effective_go.html#slices

pritidesai:
> one option would be implementing this PR with the same behavior as we use for existing Conditions, and then separately deciding how we want to handle this in the future? that would mean the potential for a backwards incompatible change though.

I am afraid of the backwards incompatibility playing role here specially if the current design lands in a release as is (without  any updates to the status)
 
> i dont want to change the behavior of our existing condition - one reason we implemented it the way we did was to satisfy knative duck typing

Yup, I am on same page, keep the condition behavior as is.

> if anything, id like to see the addition of a new condition specifically for skipped. and perhaps in that case we do not include ConditionSucceeded at all? the fact that we have a few options here makes me want to suggest that we separate that discussion from this PR if we can - i could also see making a separate TEP about it to make sure we explore the options well

Making separate TEP make sense but again concerned about backward compatibility. One quicker way to address this would be to not include the `taskrun` at all in the `pipelinerun` status (drop this [if condition](https://github.com/tektoncd/pipeline/pull/3135/files#diff-0e0c01fdbdc46d2b6054a706f99516d0R588)). This would exclude the task for which when expression evaluated to `false` along with all its dependencies. And results in a consistent behavior across the whole branch which was skipped (no `taskrun` object created at all).

Pipelinerun status has a count of that skipped task:

```
"status": {
                "completionTime": "2020-08-26T16:38:11Z",
                "conditions": [
                    {
                        "lastTransitionTime": "2020-08-26T16:38:11Z",
                        "message": "Tasks Completed: 0 (Failed: 0, Cancelled 0), Skipped: 1",
                        "reason": "Completed",
                        "status": "True",
                        "type": "Succeeded"
                    }
                ],
``` 

 And describing PR does not list any taskruns:

```
 tkn pr describe pipelinerun-to-skip-task
Name:        pipelinerun-to-skip-task
Labels:
 tekton.dev/pipeline=pipelinerun-to-skip-task

🌡️  Status

STARTED          DURATION    STATUS
28 seconds ago   0 seconds   Succeeded(Completed)

🗂  Taskruns

 NAME   TASK NAME   STARTED   DURATION   STATUS
```

And one drawback to this approach is nothing included in the `pipelinerun` status for that task:

```
"taskRuns": {
                    "pipelinerun-to-skip-task-skip-this-task-2jdzf": {
                        "pipelineTaskName": "skip-this-task"
                    }
                }
```

bobcatfish:
Maybe we can discuss in the API WG on Monday and find a way to unblock this PR at least - @ImJasonH had suggested that having Succeeded:True in this case, regardless of other info we might provide (e.g. a Reason and/or an additional Condition) might be a good path forward, since Succeeded:False feels like it's at the root of the current problems with Conditions

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -0,0 +1,139 @@
...
+// WhenExpression allows a PipelineTask to declare expressions to be evaluated before the Task is run
+// to determine whether the Task should be executed or skipped
+type WhenExpression struct {
+	// Input is the string for guard checking which can be a static input or an output from a parent Task
+	Input string `json:"input,omitempty"`
+	// Operator that represents an Input's relationship to the values
+	Operator selection.Operator `json:"operator,omitempty"`
+	// Values is an array of strings, which is compared against the input, for guard checking
+	// It must be non-empty
+	Values []string `json:"values,omitempty"`
```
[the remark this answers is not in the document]
pritidesai: yes please, no `omitempty` here, `when` itself is optional but I believe none of its fields are optional, we need all three to evaluate an expression

docs/pipelines.md
```diff
@@ -316,8 +317,60 @@ tasks:
...
+- Checking if the `Result` of a previous `Task` is as expected
+- Checking if a git file has changed in the previous commits
+- Checking if an image exists in the registry
+- Checking if the name of a CI job matches
+
+The components of `WhenExpressions` are `Input`, `Operator` and `Values`:
+
+- `Input` is the input for the `WhenExpression` which can be static inputs or variables (`Parameters` or `Results`).
+- `Operator` represents an `Input`'s relationship to a set of `Values`. `Operators` we will use in `WhenExpressions` are `in` and `notin`.
+- `Values` is an array of string values. The `Values` array must be non-empty. It can contain static values or variables (`Parameters` or `Results`.
```
pritidesai, line 336: NIT: missing `)` here in the end

examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml
```diff
@@ -0,0 +1,239 @@
...
+      taskSpec:
+        workspaces:
+          - name: source
+            description: The workspace to create the readme file in.
+        steps:
+          - name: write-new-stuff
+            image: ubuntu
+            script: 'touch $(workspaces.source.path)/README.md'
+    - name: check-file
+      when:
```
pritidesai, line 180: this `when` is not needed here since its always `true`, we can drop this from here. And instead we could add this in the doc to demonstrate different `when` expressions

docs/pipelines.md
```diff
@@ -316,8 +317,60 @@ tasks:
       name: build-push
 ```
 
+### Guard `Task` execution using `WhenExpressions`
```
pritidesai, line 320: we need more doc on the order of tasks execution (from the TEP) i.e. `runAfter` is not necessary when an `input` is a task result otherwise `runAfter` is must and no task ordering guaranteed .

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -0,0 +1,139 @@
...
+	return WhenExpression{Input: replacedInput, Operator: we.Operator, Values: replacedValues}
+}
+
+// WhenExpressions is an array of When Expressions which are used to specify whether a Task should be executed or skipped.
+type WhenExpressions []WhenExpression
+
+// AreTrue indicates whether the When Expression are true.
+func (wes WhenExpressions) AreTrue() bool {
+	for _, we := range wes {
+		if isTrue := we.isTrue(); !isTrue {
```
pritidesai, line 85: NIT: I don't think `isTrue` is required here, this `if` condition can be simplified to:

```
if !we.isTrue() {
  return false
}
```

And similar `if` condition on line 95.

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -0,0 +1,139 @@
...
+func (we *WhenExpression) GetVarSubstitutionExpressionsForWhenExpression() ([]string, bool) {
+	var allExpressions []string
+	allExpressions = append(allExpressions, validateString(we.Input)...)
+	for _, value := range we.Values {
+		allExpressions = append(allExpressions, validateString(value)...)
+	}
+	return allExpressions, len(allExpressions) != 0
+}
+
+func (we *WhenExpression) valuesContainsInput() bool {
```
pritidesai, line 45: NIT: I would expect to have a check if an `input` is in `Value` the way when expressions are defined. This function remains same in either case, just the method name threw me off. But feel free to ignore it if you would like.

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}
+	return false
+}
+
+func (we *WhenExpression) isTrue() bool {
+	if we.Operator == selection.In {
+		return we.valuesContainsInput()
+	}
+	return !we.valuesContainsInput()
```
pritidesai, line 58: NIT: we are assuming here the operator is `notin` which is fine since right now only two operators are allowed `in` and `notin`. But might be easier to build on this to allow additional operators if we explicitly check for `notin` operator as well.

pkg/apis/pipeline/v1beta1/when_types.go
pritidesai: We are storing all the when expressions specified and their results in [status.taskruns[].whenExpressionsStatus](https://gist.github.com/pritidesai/cbc7caadbd80f02745383acf524e3186#file-pull-3135-pipelinerun-json-L200) but do we really need to? It would be beneficial for debugging to see the resolved when expressions but at the same time makes the status too verbose. Instead, if we can store the resolved when expressions in the [status.pipelineSpec.tasks[].when](https://gist.github.com/pritidesai/cbc7caadbd80f02745383acf524e3186#file-pull-3135-pipelinerun-json-L125), we can avoid these structs and simplify the following function `GetWhenExpressionStatus`.

pritidesai: `GetWhenExpressionStatus` then needs to iterate over when expressions and return false if any of the expression evaluates to false i.e. `!wes[i].isTrue()`.

pritidesai: **This is separate from this PR**

I was having trouble seeing unresolved when expressions in `status.pipelineSpec.tasks[]` and was trying to troubleshoot and discovered we actually store the `pipelineSpec` as is even before resolving params and it's never updated for auditing purpose. But might be clearer if its updated once after resolving all params (and pipeline resources).

pritidesai: In that case, it make sense to store resolved when expressions in the status 😢

pritidesai:
Just to be aware of (no blocker as such), it is possible that a when expression operates on a param without that task specifying any param. In this case, when expression is reading value from the `Pipeline` param it possible because `when` expression is at the same level as `taskRef` or `taskSpec`:

```
spec:
  pipelineSpec:
    params:
      - name: PARAM
        default: "foo"
    tasks:
      - name: echo-hi
        when:
          - input: $(params.PARAM)
            operator: in
            values: ["foo"]
        taskSpec:
          steps:
            - name: echo
              image: ubuntu
              script: echo "Hi!"
```

pkg/apis/pipeline/v1beta1/pipeline_validation.go
jerop: they are related because I use these functions to validate parameters in when expressions

pkg/reconciler/pipelinerun/pipelinerun_test.go
jerop: yes, the other tests for pipelineruns have the docstrings -- can exclude it for this test case

docs/pipelines.md
pritidesai: nope, sorry if I am causing any confusion, what I meant to add here was "When expression with task results introduces resource dependency where as when expression with params does not change task execution order. A pipeline task with `runAfter` and `When` expression with task results both play role in execution order." does it read well? 🤔

jerop: yes yes, that makes sense...thanks for the clarification!

docs/pipelines.md
```diff
@@ -316,8 +317,60 @@ tasks:
...
+        values: ["exists"]
+    taskRef:
+        name: echo-file-exists
+```
+
+For an end-to-end example, see [PipelineRun with WhenExpressions](../examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml).
+
+When `WhenExpressions` are specified in a `Task`, [`Conditions`](#guard-task-execution-using-conditions) should not be speficied in the same `Task`. 
+The `Pipeline` will be rejected as invalid if both `WhenExpressions` and `Conditions` are included.
+
```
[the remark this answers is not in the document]
jerop: sounds good, added it here: https://github.com/tektoncd/pipeline/pull/3135/files#diff-80b68f938b349a4fc92d7629360bb25eR460-R524 as discussed in https://github.com/tektoncd/pipeline/issues/3139#issuecomment-682000189

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
jerop: these are great ideas...we could also use `Allows` for allows execution of the guarded `Task`
```
if t.PipelineTask.WhenExpressions.Allows() {
```

pkg/apis/pipeline/v1beta1/when_types.go
jerop: Changed the method name to `isInputInValues` 

```
func (we *WhenExpression) isInputInValues() bool {
```

pkg/apis/pipeline/v1beta1/when_types.go
pritidesai: After reading #3140, we might want to reconsider storing `when` expressions in the status (not trying to be a blocker here, this can be updated in next cycle which I believe we should, thanks @jerop for creating an issue #3145 to track this) 🙏

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -356,34 +360,72 @@ func validatePipelineParameterVariables(tasks []PipelineTask, params []ParamSpec
 		}
 	}
 
-	return validatePipelineVariables(tasks, "params", parameterNames, arrayParameterNames)
+	return validatePipelineParametersVariables(tasks, "params", parameterNames, arrayParameterNames)
```
pritidesai, line 363: this would be misleading for invalid params in when expressions, would produce the following error for the test case in [pipeline_validation_test.go](https://github.com/tektoncd/pipeline/pull/3135/files#diff-d1d41b803d92fc5570cfcefb5e355959R865) in which `baz` is not defined as a pipeline param but used in `when` expression:

```
validation failed: non-existent variable in "$(params.baz)" for task parameter whenInput[$(params.baz)]: pipelinespec.params.whenInput[$(params.baz)] 
```

This function is part of validating Pipeline Params and the prefix `params` is used for validating params and also when expression. Even if this prefix is fixed,[ the nested function](https://github.com/tektoncd/pipeline/pull/3135/files#diff-5b096c7641765902d79ee2d5e5f7d213R430) has it hardcoded as `task parameter` and `pipelineSpec.params` which is wouldn't be valid for `pipelineSpec.tasks.when`.

This calls for a huge refactoring of pipeline validation 😢   Again not blocking this PR here but worth following up in separate PR.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -356,34 +360,72 @@ func validatePipelineParameterVariables(tasks []PipelineTask, params []ParamSpec
...
+		if err := validateParametersVariablesInWhenExpressions(task.WhenExpressions, prefix, paramNames, arrayParamNames); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func validateParametersVariablesInWhenExpressions(wes WhenExpressions, prefix string, paramNames sets.String, arrayParamNames sets.String) *apis.FieldError {
+	for _, whenExpression := range wes {
+		if err := validatePipelineStringVariable(fmt.Sprintf("whenInput[%s]", whenExpression.Input), whenExpression.Input, prefix, paramNames, arrayParamNames); err != nil {
```
pritidesai, line 380: s/whenInput/when.input/ we follow the dot notation

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -356,34 +360,72 @@ func validatePipelineParameterVariables(tasks []PipelineTask, params []ParamSpec
...
+			return err
+		}
+		for _, whenExpressionValue := range whenExpression.Values {
+			if err := validatePipelineStringVariable(fmt.Sprintf("whenValue[%s]", whenExpressionValue), whenExpressionValue, prefix, paramNames, arrayParamNames); err != nil {
```
pritidesai, line 384: s/whenValue/when.values/

pkg/apis/pipeline/v1beta1/when_validation.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+	}
+	return nil
+}
+
+func (we *WhenExpression) validateWhenExpressionFields() *apis.FieldError {
+	if equality.Semantic.DeepEqual(we, &WhenExpression{}) || we == nil {
+		return apis.ErrMissingField(apis.CurrentField)
+	}
+	if err := validateWhenExpressionOperator(&we.Operator); err != nil {
```
pritidesai, line 57: missing validation on input just like we have it on Operator and Values here, this expression is considered valid otherwise:

```
		name: "missing input",
		wes: []WhenExpression{{
			Operator: selection.In,
			Values:   []string{"foo"},
		}},
```

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -159,6 +162,15 @@ func (t ResolvedPipelineRunTask) IsSkipped(state PipelineRunState, d *dag.Graph)
 		}
 	}
 
+	// Check if the when expressions are false, based on the input's relationship to the values
+	if len(t.PipelineTask.WhenExpressions) > 0 {
+		if !t.PipelineTask.WhenExpressions.HaveVariables() {
```
pritidesai, line 167: What is the significance of `HaveVariables` here?

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -0,0 +1,142 @@
...
+
+// GetStatus iterates through the When Expressions, and evaluates an Input's
+// relationship to an array of Values based on the Operator and produces a Status update
+// for Tasks with When Expressions.
+func (wes WhenExpressions) GetStatus() []*WhenExpressionsResult {
+	var status []*WhenExpressionsResult
+	for i := range wes {
+		status = append(status, &WhenExpressionsResult{
+			Expression: &wes[i],
+			IsTrue:     wes[i].isTrue(),
```
pritidesai, line 138: `IsTrue` is not necessary here, at this point (with resolved expressions), its very clear that `foo` is in `["foo", "bar"]` and `zoo` is not in `["foo", "bar"]`, its consuming unnecessary space which might be fine for a pipeline with few tasks but might not be fine running it repeatedly or for a pipeline with thousands of tasks 😜 

```
    "skip-this-task": [
                        {
                            "expression": {
                                "Input": "zoo",
                                "Operator": "in",
                                "Values": [
                                    "foo"
                                ]
                            },
                            "isTrue": false
                        }
                    ]
```

jerop: I was initially thinking that the values could be a long list and someone might find it tedious to look through the list, but I think that's a good point...

But that also makes me question whether we need the expressions because inputs & values in the pipelinespec could be sufficient because:
- the non-variables are right there in the spec
- the param variables in the when expressions in the pipeline spec can be looked up in the pipeline spec
- the result variables in the when expressions in the pipeline spec can be looked up in result section of the relevant taskrun

so maybe I can get remove the When Expressions Status from TaskRuns and Skipped, since all the details they need are already somewhere in the status 🤔

bobcatfish: im a bit torn! if we want, we could leave this out entirely for now and add it later, but im slightly inclined toward including it because folks will want to understand why something got skipped - i guess they can reverse engineer it from the information that is already there, but one difference between when expressions and tasks is that the logic in a task is up to the task itself, while the logic for a when expression is in the controller, so i can see being a bit more verbose about "this is exactly what we computed and this was the result"

pkg/apis/pipeline/v1beta1/pipeline_validation.go
jerop: good catch, made the location name and path variables passed into the nested functions

pkg/reconciler/pipelinerun/pipelinerun_test.go
jerop: added them below, with parameter and results replacements, with both evaluating to true and evaluating to false

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
jerop: validating that we're not determining that a task is skipped based on an unreplaced variable in a when expression

pkg/apis/pipeline/v1beta1/when_validation.go
jerop: in that case, the input will default to an empty string -- added validation that it's not an empty string but wondering if a user may intentionally want to pass in an empty string

pkg/apis/pipeline/v1beta1/when_types.go
pritidesai: @bobcatfish yup I can understand your concern, how about storing these details in controller logs? Do we expect users would have access to the logs in standard CI/CD setup?

jerop:
@tektoncd/core-maintainers please review this PR -- it in includes deprecating Conditions CRD and we'll [need at least half of the owners to sign off](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes)

sbwsg:
/approve
/hold until we get sign off from enough folks

This LGTM!

sbwsg:
/hold

... until we get sign off from enough folks

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -356,44 +360,84 @@ func validatePipelineParameterVariables(tasks []PipelineTask, params []ParamSpec
...
+			return err
+		}
+	}
+	return nil
+}
+
+func validateParametersVariablesInWhenExpressions(whenExpressions WhenExpressions, prefix string, paramNames sets.String, arrayParamNames sets.String) *apis.FieldError {
+	name, path := "task when expression", "pipelinespec.when"
+	for _, we := range whenExpressions {
+		if err := validatePipelineStringVariable(fmt.Sprintf("when.input[%s]", we.Input), we.Input, prefix, name, path, paramNames, arrayParamNames); err != nil {
```
pritidesai, line 381: Thanks a bunch @jerop for fixing it in this PR 🙏 and sorry for circling back, but you have done an amazing job separating and refactoring param validation and when expression validation, I would like to see one more change here. This `when` is not needed here, it adds extra `when` in the overall message: `pipelinespec.when.when.input[...])` and on line 385 causes `pipelinespec.when.when.values[...])`

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -356,44 +360,84 @@ func validatePipelineParameterVariables(tasks []PipelineTask, params []ParamSpec
...
+		}
+	}
+	return nil
+}
+
+func validateParametersVariablesInParameters(params []Param, prefix string, paramNames sets.String, arrayParamNames sets.String) *apis.FieldError {
+	name, path := "task parameter", "pipelinespec.params"
+	for _, param := range params {
+		if param.Value.Type == ParamTypeString {
+			if err := validatePipelineStringVariable(fmt.Sprintf("param[%s]", param.Name), param.Value.StringVal, prefix, name, path, paramNames, arrayParamNames); err != nil {
```
pritidesai, line 397: similar to the when expression, `param` here is not needed, results in `pipelinespec.params.param[b-param]` and on line 402.

docs/pipelines.md
```diff
@@ -316,8 +317,75 @@ tasks:
...
+from previous `Tasks` (that is, not through `Parameters` in the guarded `Task`).
+Using `Results` in a `WhenExpression`in a guarded `Task` introduces a resource
+dependency on the previous `Task` that produced the `Result`, thus `RunAfter`
+is not necessary. Otherwise, include `runAfter` to specify `Task` ordering. 
+
+The declared `WhenExpressions` are evaluated before the `Task` is run. 
+If all the `WhenExpressions`evaluate to `True`, the `Task` is run. 
+If any of the `WhenExpressions` evaluate to `False`, the `Task` is
+not run and the `TaskRun` status field `ConditionSkipped` is set to `True` 
+with the reason set to `WhenExpressionsEvaluatedToFalse`.
```
pritidesai, line 353: @jerop please make sure to update this section to exclude referring to the `taskrun` and `WhenExpressionsEvaluatedToFalse`.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
jerop: good catch, fixed it :)

docs/pipelines.md
jerop: updated the docs, thanks :)

docs/pipelines.md
```diff
@@ -316,8 +317,74 @@ tasks:
...
+      - input: "$(params.path)"
+        operator: in
+        values: ["README.md"]
+    taskRef:
+      name: create-readme-file
+---
+tasks:
+  - name: echo-file-exists
+    when:
+      - input: "$(tasks.check-file.results.path)"
```
afrittoli, line 371: I think this should be `$(tasks.check-file.results.status)` ?

jerop: yes, thanks :)

dibyom:
> @tektoncd/core-maintainers please review this PR -- it includes deprecating Conditions CRD and we'll [need at least half of the owners to sign off](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes)

Deprecating Conditions SGTM!

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -356,44 +360,84 @@ func validatePipelineParameterVariables(tasks []PipelineTask, params []ParamSpec
...
+	if err := validatePipelineVariable(name, value, prefix, locationName, path, stringVars); err != nil {
+		return err
+	}
+	if err := validatePipelineArraysIsolated(name, value, prefix, locationName, path, arrayVars); err != nil {
+		return err
+	}
+	return nil
+}
+
+func validatePipelineVariable(name, value, prefix string, locationName string, path string, vars sets.String) *apis.FieldError {
```
bobcatfish, line 431: there are 2 "smells" here that i'd like to point out:

1. passing variables through that end up being part of constructing an error <-- this makes me think that it would make sense for there to be a function and/or struct that could know what kind of thing it's validating and control that kind of information vs passing it around
2. there are a lot of layers of functions calling functions. the more layers we add, the more cognitive load we put on folks to understand the code, and also it makes the unit testing story very string: which of these functions should we unit test?

im not sure what the ideal looks like here and if you want to leave it like this its not the end of the world here, but i have a feeling there's a much clearer way to organize this - and one step might be just collapsing some of these functions and tolerating the code repetition

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -331,6 +331,10 @@ type PipelineRunStatusFields struct {
 
 	// PipelineRunSpec contains the exact spec used to instantiate the run
 	PipelineSpec *PipelineSpec `json:"pipelineSpec,omitempty"`
+
+	// list of tasks that were skipped due to when expressions evaluating to false
+	// +optional
+	SkippedTasks []string `json:"skippedTasks,omitempty"`
```
bobcatfish, line 337: one of the options for skipped tasks in [the tep](https://github.com/tektoncd/community/pull/192) involves adding more information under "skipped tasks" - maybe it makes sense to make this a list of a more complex object so we can add more fields later without making a backwards incompatible change?

depends on how likely you think it is that we'll want that extra information.

pkg/apis/pipeline/v1beta1/resultref_test.go
```diff
@@ -433,3 +434,343 @@ func TestLooksLikeResultRef(t *testing.T) {
...
+					PipelineTask: "sumTask1",
+					Result:       "sumResult",
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			expressions, ok := tt.args.whenExpression.GetVarSubstitutionExpressions()
+			if !ok && tt.want != nil {
```
bobcatfish, line 563: what if ok is true but want is nil, should that fail too?

im wondering if you need to check want here at all, or if !ok is enough? or do we also expect an error in the case where want is nil? if so i suggest adding a different test for that case since it might be clearer

pkg/apis/pipeline/v1beta1/resultref_test.go
```diff
@@ -433,3 +434,343 @@ func TestLooksLikeResultRef(t *testing.T) {
...
+				}
+			}
+		})
+	}
+}
+
+func TestHasResultReferenceWhenExpression(t *testing.T) {
+	type args struct {
+		whenExpression v1beta1.WhenExpression
+	}
```
bobcatfish, line 578: could you use whenExpression directly instead of wrapping it in the args struct?

pkg/apis/pipeline/v1beta1/resultref_test.go
```diff
@@ -433,3 +434,343 @@ func TestLooksLikeResultRef(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			expressions, ok := tt.args.whenExpression.GetVarSubstitutionExpressions()
+			if !ok {
+				t.Fatalf("expected to find expressions but didn't find any")
+			}
+			got := v1beta1.NewResultRefs(expressions)
+			sort.Slice(got, func(i, j int) bool {
```
bobcatfish, line 667: might be worth a comment to explain why the order is non deterministic (do we want it to be non deterministic?)

pkg/apis/pipeline/v1beta1/resultref_test.go
```diff
@@ -433,3 +434,343 @@ func TestLooksLikeResultRef(t *testing.T) {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			expressions, ok := tt.args.whenExpression.GetVarSubstitutionExpressions()
+			if ok {
+				if got := v1beta1.LooksLikeContainsResultRefs(expressions); got != tt.want {
+					t.Errorf("LooksLikeContainsResultRefs() = %v, want %v", got, tt.want)
+				}
+			} else if tt.want {
```
bobcatfish, line 771: this kinda looks like a candidate for a separate set of tests to me also

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -0,0 +1,115 @@
...
+	var replacedValues []string
+	for _, val := range we.Values {
+		replacedValues = append(replacedValues, ApplyReplacements(val, replacements))
+	}
+
+	return WhenExpression{Input: replacedInput, Operator: we.Operator, Values: replacedValues}
+}
+
+// ReplaceWhenExpressionsVariables interpolates variables, such as Parameters and Results, in
+// the Input and Values.
```
bobcatfish, line 108: probably worth mentioning it returns a new instance as well

(also: nice! :D)

pkg/apis/pipeline/v1beta1/when_validation.go
```diff
@@ -0,0 +1,85 @@
...
+		message := fmt.Sprintf("operator %q is not recognized. valid operators: %s", we.Operator, strings.Join(validWhenOperators, ","))
+		return apis.ErrInvalidValue(message, "spec.task.when")
+	}
+	if len(we.Values) == 0 {
+		return apis.ErrInvalidValue("expecting non-empty values field", "spec.task.when")
+	}
+	return nil
+}
+
+func (wes WhenExpressions) validateTaskResultsVariables() *apis.FieldError {
```
bobcatfish, line 70: it's too bad these are all unexported - i do think they might be worth unit testing regardless?

pkg/apis/pipeline/v1beta1/when_validation.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+	}
+	return nil
+}
+
+func (we *WhenExpression) validateWhenExpressionFields() *apis.FieldError {
+	if equality.Semantic.DeepEqual(we, &WhenExpression{}) || we == nil {
+		return apis.ErrMissingField(apis.CurrentField)
+	}
+	if we.Input == "" { //TODO(jerop): could someone use "" as a valid input?
```
bobcatfish, line 57: whenever we include TODOs we should reference an open issue to track the work - or deal with it before merging and remove the TODO

feels like it could be an okay input to me? i could imagine having a task produce a result that is an empty string (for some reason)

bobcatfish (commented):
Looks like its nearly there!

pkg/apis/pipeline/v1beta1/when_validation.go
jerop: `Input` is a string so its zero value is empty string -- `Input` defaults to empty string when not specified. Looked into disallowing defaulting to an empty string and I didn't find a good solution, please let me know if there's a good way to do this. The alternatives are:
1) documenting that if `Input` is not specified, it'd default to empty string and we'll compare that against the `Values`
2) disallowing empty strings as `Inputs` -- although some use cases could need this to be valid 

So maybe we can go with option 1 instead?

pkg/apis/pipeline/v1beta1/pipeline_validation.go
jerop: replaced it with code repetition, doesn't look good, we might have to refactor pipeline validation soon

pkg/apis/pipeline/v1beta1/when_validation.go
jerop: already added in `when_validation_tests.go`

pkg/apis/pipeline/v1beta1/pipeline_validation.go
pritidesai: Pipeline validation refactoring is a long overdue, I would recommend to have it in a separate PR (candidate for `good-first-issue`, nah may be overkill for fresh eyes 🤔 )

pritidesai: some of these functions could be moved to `param_types.go`

pkg/apis/pipeline/v1beta1/when_validation.go
bobcatfish: whoops, sorry for missing that! thanks :D

pkg/apis/pipeline/v1beta1/when_validation.go
bobcatfish: good catch, very thorough analysis!! 🤩

docs/pipelineruns.md
```diff
@@ -457,6 +457,26 @@ False|PipelineRunTimeout|Yes|The `PipelineRun` timed out.
...
+  Status:                True
+  Type:                  Succeeded
+Skipped Tasks:
+  Name:       skip-this-task
+Task Runs:
+  pipelinerun-to-skip-task-run-this-task-r2djj:
+    Pipeline Task Name:  run-this-task
+    Status:
+      ...
+```
```
bobcatfish, line 478: nice, thanks for making this example shorter!!

docs/pipelineruns.md
```diff
@@ -457,6 +457,26 @@ False|PipelineRunTimeout|Yes|The `PipelineRun` timed out.
...
 When a `PipelineRun` changes status, [events](events.md#pipelineruns) are triggered accordingly.
 
+When a `PipelineRun` has `Tasks` with [WhenExpressions](pipelines.md#guard-task-execution-using-whenexpressions):
+- If the `WhenExpressions` evaluate to `true`, the `Task` is executed and the `TaskRun` will be listed in the `Task Runs` section of the `status` of the `PipelineRun`.
+- If the `WhenExpressions` evaluate to `false`, the `Task` is skipped and it is listed in the `Skipped Tasks` section of the `status` of the `PipelineRun`. 
+
+```yaml
+Conditions:
+  Last Transition Time:  2020-08-27T15:07:34Z
+  Message:               Tasks Completed: 1 (Failed: 0, Cancelled 0), Skipped: 1
```
bobcatfish, line 467: this makes a lot of sense 👍

docs/pipelines.md
```diff
@@ -316,8 +317,56 @@ tasks:
...
+- `Operator` represents an `Input`'s relationship to a set of `Values`. A valid `Operator` must be provided, which can be either `in` or `notin`.
+- `Values` is an array of string values. The `Values` array must be provided and be non-empty. It can contain static values or variables ([`Parameters`](#specifying-parameters) or [`Results`](#using-results)).
+
+The [`Parameters`](#specifying-parameters) are read from the `Pipeline` and [`Results`](#using-results) are read directly from previous [`Tasks`](#adding-tasks-to-the-pipeline). Using [`Results`](#using-results) in a `WhenExpression` in a guarded `Task` introduces a resource dependency on the previous `Task` that produced the `Result`. 
+
+The declared `WhenExpressions` are evaluated before the `Task` is run. If all the `WhenExpressions` evaluate to `True`, the `Task` is run. If any of the `WhenExpressions` evaluate to `False`, the `Task` is not run and the `Task` is listed in the [`Skipped Tasks` section of the `PipelineRunStatus`](pipelineruns.md#monitoring-execution-status).
```
bobcatfish, line 331: thanks for adding all these links!

examples/v1beta1/pipelineruns/pipelinerun-with-resourcespec.yaml
```diff
@@ -20,7 +20,8 @@ spec:
         - |
           ls -al $(resources.inputs.pipeline-git.path)
 ---
-
+# `Conditions` are deprecated, use `WhenExpressions` instead
+# https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#guard-task-execution-using-whenexpressions
```
bobcatfish, line 24: thanks for adding all of these also 🙏

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -331,6 +331,17 @@ type PipelineRunStatusFields struct {
...
 	PipelineSpec *PipelineSpec `json:"pipelineSpec,omitempty"`
+
+	// list of tasks that were skipped due to when expressions evaluating to false
+	// +optional
+	SkippedTasks []SkippedTask `json:"skippedTasks,omitempty"`
+}
+
+// SkippedTask is used to describe the Tasks that were skipped due to their When Expressions
+// evaluating to False
+type SkippedTask struct {
```
bobcatfish, line 342: nice 👍

bobcatfish: might be worth a comment explaining why this is a struct - or part of the commit message? nbd tho

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -2004,6 +2006,230 @@ func ensurePVCCreated(t *testing.T, clients test.Clients, name, namespace string
...
+	if err != nil {
+		t.Fatalf("Could not get a dag from the state %#v: %v", pipelineRunState, err)
+	}
+	skippedtasks := getSkippedTasks(pr, pipelineRunState, dag)
+	if d := cmp.Diff(expectedSkippedTasks, skippedtasks); d != "" {
+		t.Fatalf("Expected PipelineRun status to match Skipped status, but got a mismatch: %s", diff.PrintWantGot(d))
+	}
+}
+
+func DagFromState(state resources.PipelineRunState) (*dag.Graph, error) {
```
bobcatfish, line 2225: probably doesnt need to be exported

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -2004,6 +2006,230 @@ func ensurePVCCreated(t *testing.T, clients test.Clients, name, namespace string
...
+	if len(actual.Items) != 1 {
+		t.Fatalf("Expected 1 TaskRuns got %d", len(actual.Items))
+	}
+	actualTaskRun := actual.Items[0]
+	if d := cmp.Diff(&actualTaskRun, expectedTaskRun, ignoreResourceVersion); d != "" {
+		t.Errorf("expected to see TaskRun %v created. Diff %s", expectedTaskRunName, diff.PrintWantGot(d))
+	}
+}
+
+func TestUpdateTaskRunsWithWhenExpressions(t *testing.T) {
```
bobcatfish, line 2175: note on test naming: i try to name tests after the function being tested, i cant quite tell which function this is testing - i think it's testing getSkippedTasks? if so maybe you could call it `TestGetSkippedTasks` ?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -2004,6 +2006,230 @@ func ensurePVCCreated(t *testing.T, clients test.Clients, name, namespace string
...
+	if err != nil {
+		t.Fatalf("Failure to list TaskRun's %s", err)
+	}
+	if len(actual.Items) != 1 {
+		t.Fatalf("Expected 1 TaskRuns got %d", len(actual.Items))
+	}
+	actualTaskRun := actual.Items[0]
+	if d := cmp.Diff(&actualTaskRun, expectedTaskRun, ignoreResourceVersion); d != "" {
+		t.Errorf("expected to see TaskRun %v created. Diff %s", expectedTaskRunName, diff.PrintWantGot(d))
+	}
```
bobcatfish, line 2172: is there any potential to combine this test and the one before into one test that's table driven? just ignore me if the logic is more substantially different than ive realized!

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -65,25 +65,21 @@ func ApplyContexts(spec *v1beta1.PipelineSpec, pipelineName string, pr *v1beta1.
...
-// ApplyTaskResults applies the ResolvedResultRef to each PipelineTask.Params in targets
+// ApplyTaskResults applies the ResolvedResultRef to each PipelineTask.Params and Pipeline.WhenExpressions in targets
 func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResultRefs) {
-	stringReplacements := map[string]string{}
-
-	for _, resolvedResultRef := range resolvedResultRefs {
-		replaceTarget := fmt.Sprintf("%s.%s.%s.%s", v1beta1.ResultTaskPart, resolvedResultRef.ResultReference.PipelineTask, v1beta1.ResultResultPart, resolvedResultRef.ResultReference.Result)
-		stringReplacements[replaceTarget] = resolvedResultRef.Value.StringVal
-	}
+	replacements := resolvedResultRefs.getReplacements()
```
bobcatfish, line 70: nice!

(if you wanted, this improvement would probably be a good candidate for a separate PR as well? but np if you dont want to bother esp. since this PR is getting so close to being merged)

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -141,12 +141,12 @@ func (t ResolvedPipelineRunTask) IsStarted() bool {
...
-// (1) its Condition Checks failed or
-// (2) one of the parent task's conditions failed or
+// Skip returns true if a PipelineTask will not be run because
+// (1) its Condition Checks failed or its When Expressions evaluated to false
+// (2) one of the parent task's conditions failed or When Expressions evaluated to false
 // (3) Pipeline is in stopping state (one of the PipelineTasks failed)
-// Note that this means IsSkipped returns false if a conditionCheck is in progress
-func (t ResolvedPipelineRunTask) IsSkipped(state PipelineRunState, d *dag.Graph) bool {
+// Note that this means Skip returns false if a conditionCheck is in progress
+func (t *ResolvedPipelineRunTask) Skip(state PipelineRunState, d *dag.Graph) bool {
```
bobcatfish, line 149: nice improvement!

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -2660,3 +2701,79 @@ func TestPipelineRunState_GetFinalTasks(t *testing.T) {
...
+					return &trs[0], nil
+				case "pipelinerun-mytask1-9l9zj-always-true-1-mssqb":
+					return t2, nil
+				}
+				return nil, fmt.Errorf("getTaskRun called with unexpected name %s", name)
+			},
+		},
+	}
+
+	for _, tc := range tcs {
```
bobcatfish, line 2771: i think there is only one test case here? i usually avoid putting in table driven tests with just one case unless im SURE we want more than one

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -2660,3 +2701,79 @@ func TestPipelineRunState_GetFinalTasks(t *testing.T) {
...
+		{
+			name: "When Expressions exist",
+			getTaskRun: func(name string) (*v1beta1.TaskRun, error) {
+				switch name {
+				case "pipelinerun-mytask1-9l9zj-always-true-0-mz4c7":
+					return t1, nil
+				case "pipelinerun-mytask1-9l9zj":
+					return &trs[0], nil
+				case "pipelinerun-mytask1-9l9zj-always-true-1-mssqb":
+					return t2, nil
```
bobcatfish, line 2764: i wonder if this test case needs to be so complicated - e.g. can we narrow it down to only needing to get ONE taskrun without losing anything important?

bobcatfish (commented):
Just some minor feedback from me at this point, looking great!

/lgtm

pkg/reconciler/pipelinerun/resources/apply.go
jerop: moved this refactor to a separate PR: https://github.com/tektoncd/pipeline/pull/3169

docs/pipelines.md
```diff
@@ -316,8 +317,56 @@ tasks:
...
+      - input: "$(tasks.check-file.results.status)"
+        operator: in
+        values: ["exists"]
+    taskRef:
+        name: echo-file-exists
+```
+
+For an end-to-end example, see [PipelineRun with WhenExpressions](../examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml).
+
+When `WhenExpressions` are specified in a `Task`, [`Conditions`](#guard-task-execution-using-conditions) should not be speficied in the same `Task`. The `Pipeline` will be rejected as invalid if both `WhenExpressions` and `Conditions` are included.
```
pritidesai, line 357: Minor NIT: typo - s/speficied/specified/

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -331,6 +331,18 @@ type PipelineRunStatusFields struct {
 
 	// PipelineRunSpec contains the exact spec used to instantiate the run
 	PipelineSpec *PipelineSpec `json:"pipelineSpec,omitempty"`
+
+	// list of tasks that were skipped due to when expressions evaluating to false
+	// +optional
+	SkippedTasks []SkippedTask `json:"skippedTasks,omitempty"`
```
pritidesai, line 337: No changes needed here. Looking at `json:...` here made me think what if someone tries to mock with the system by adding a bunch of `SkippedTasks` in the status. Looking at the overall `struct`, it seems possible to specify `Status` similar to `Spec` since most of the fields are marked as `optional` with `JSON` schema supported 🤔  is it something by design?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -2004,6 +2006,230 @@ func ensurePVCCreated(t *testing.T, clients test.Clients, name, namespace string
...
+	pr.Status.InitializeConditions()
+	taskrunstatus := getTaskRunsStatus(pr, pipelineRunState)
+	if d := cmp.Diff(expectedPipelineRunStatus.TaskRuns, taskrunstatus, ignoreLastTransitionTime); d != "" {
+		t.Fatalf("Expected PipelineRun status to match TaskRun(s) status, but got a mismatch: %s", diff.PrintWantGot(d))
+	}
+	dag, err := dagFromState(pipelineRunState)
+	if err != nil {
+		t.Fatalf("Could not get a dag from the state %#v: %v", pipelineRunState, err)
+	}
+	skippedtasks := getSkippedTasks(pr, pipelineRunState, dag)
```
pritidesai, line 2219: NIT: I would recommend checking list of skipped tasks in the pipeline run status rather than directly calling `getSkippedTasks` function here. That way, we are making sure that the status has a list of expected skipped tasks.

pritidesai:
Thanks a bunch @jerop for this wonderful feature and working so patiently with all of us. This is great. You will make many users happy 😸  with this feature 🎉  🎉 🎉 

/approve

jerop:
@tektoncd/core-maintainers we need one more sign off for this PR if you have some time -- so far we have:
- @pritidesai - https://github.com/tektoncd/pipeline/pull/3135#issuecomment-686968456
- @bobcatfish - https://github.com/tektoncd/pipeline/pull/3135#pullrequestreview-482267992
- @dibyom - https://github.com/tektoncd/pipeline/pull/3135#issuecomment-685934762
- @sbwsg - https://github.com/tektoncd/pipeline/pull/3135#issuecomment-683963287

Thank you for the reviews!

afrittoli:
I'm reviewing the PR, not done yet.

jerop:
thank you @afrittoli 👍

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -613,6 +614,16 @@ func getTaskRunsStatus(pr *v1beta1.PipelineRun, state []*resources.ResolvedPipel
 	return status
 }
 
+func getSkippedTasks(pr *v1beta1.PipelineRun, state []*resources.ResolvedPipelineRunTask, d *dag.Graph) []v1beta1.SkippedTask {
+	skipped := []v1beta1.SkippedTask{}
+	for _, rprt := range state {
+		if rprt.Skip(state, d) {
```
afrittoli, line 620: Nice :)

The `Skip` method will be true in a number of cases:
- when expression is failed
- condition is failed
- a task is not reachable via the DAG (because of a previous skip)
- a task failed so no new task is scheduled

It might we worth adding to the docs this information. It can be in a PR on top.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -2004,6 +2005,188 @@ func ensurePVCCreated(t *testing.T, clients test.Clients, name, namespace string
...
+	))}
+	prs := []*v1beta1.PipelineRun{tb.PipelineRun("test-pipeline-run-different-service-accs", tb.PipelineRunNamespace("foo"),
+		tb.PipelineRunSpec("test-pipeline",
+			tb.PipelineRunServiceAccountName("test-sa-0"),
+		),
+	)}
+	ts := []*v1beta1.Task{
+		tb.Task("a-task", tb.TaskNamespace("foo")),
+		tb.Task("b-task", tb.TaskNamespace("foo")),
+		tb.Task("c-task", tb.TaskNamespace("foo")),
```
afrittoli, line 2106: I think it would be interesting to have a `d-task` that depends in some for from `c-task` (via result or `runAfter`) and to see that it is skipped too.

pkg/apis/pipeline/v1beta1/resultref_test.go
```diff
@@ -433,3 +434,315 @@ func TestLooksLikeResultRef(t *testing.T) {
 		})
 	}
 }
+
+func TestNewResultReferenceWhenExpressions(t *testing.T) {
```
afrittoli, line 438: Nice :)

afrittoli (commented):
Thanks for this! It's a great improvement on what we have today, I look forward to using this!

The only ask I'd have is for one extra check, a task that depends on a skipped task - I verified that it works fine so I'd be happy for that to be added in a PR on top.

/lgtm

pkg/reconciler/pipelinerun/pipelinerun_test.go
afrittoli: https://gist.github.com/afrittoli/ae5cb276b765f04b248bee0f974cd992

vdemeester (commented):
/lgtm

pkg/reconciler/pipelinerun/pipelinerun_test.go
jerop: added, thank you 😊

docs/pipelineruns.md
```diff
@@ -457,6 +457,54 @@ False|PipelineRunTimeout|Yes|The `PipelineRun` timed out.
...
+```yaml
+Completion Time:  2020-08-27T15:07:34Z
+Conditions:
+  Last Transition Time:  2020-08-27T15:07:34Z
+  Message:               Tasks Completed: 1 (Failed: 0, Cancelled 0), Skipped: 1
+  Reason:                Completed
+  Status:                True
+  Type:                  Succeeded
+Skipped Tasks:
+  Skip - This - Task
```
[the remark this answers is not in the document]
jerop: yes, that's right

jerop:
Thank you for the reviews @vdemeester, @afrittoli, @pritidesai, @bobcatfish, @dibyom and @sbwsg 🙏

PR #3136, opened 2020-08-26
Fix(git): add warning of the mismatch of git cred and url

### LINKED ISSUE #3094: If a git secret has ssh-auth type, warn about urls starting with `https:` and vice versa
### Feature request

Validate the creds-init Secrets that will be injected into a Step container. If a Git secret has a type of ssh-auth but is going to be used against a repo URL starting with `https://` then `git` won't even try to use the SSH keys. We should warn about this via an Event on the TaskRun or similar.

Similarly, if a creds-init Secret has a type of `basic-auth` but the URL being used to initialize that credential is of the form `user@domain.com` then `git` will ignore the basic auth credential, which can also be confusing. Again we should try and validate that the URL and credential line up and warn the user if they don't.

This is another fiddly edge case in our creds-init dance. See https://github.com/tektoncd/pipeline/issues/1983#issuecomment-672864395 for an example where a user hit a bump related to this.

### Use case

User supplies SSH credential for their github repo in a k8s Secret.  User accidentally puts the `https://` URL in their Git PipelineResource.  The git clone step fails because the URL is not an ssh URL and so `git` ignores the SSH creds. The user should be warned about this mismatch more explicitly.

FogDong:
# Changes

Fix(git): add warning of the mismatch of git cred and url
Fix #3094 

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

    Fix(git): add warning of the mismatch of git cred and url

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -31,6 +32,11 @@ import (
 const (
 	sshKnownHostsUserPath          = ".ssh/known_hosts"
 	sshMissingKnownHostsSSHCommand = "ssh -o StrictHostKeyChecking=accept-new"
+)
+
+var (
+	// sshURLRegexFormat matches the url of SSH git repository
+	sshURLRegexFormat = regexp.MustCompile(`(ssh://[\w\d\.]+|.+@?.+\..+:)(:[\d]+){0,1}/*(.*)`)
 )
 
 func run(logger *zap.SugaredLogger, dir string, args ...string) (string, error) {
@@ -69,6 +75,7 @@ func Fetch(logger *zap.SugaredLogger, spec FetchSpec) error {
 	if err := ensureHomeEnv(logger); err != nil {
 		return err
 	}
+	validateGitAuth(logger, spec.URL)
 
 	if spec.Path != "" {
 		if _, err := run(logger, "", "init", spec.Path); err != nil {
@@ -243,4 +250,26 @@ func userHasKnownHostsFile(logger *zap.SugaredLogger) (bool, error) {
 	}
 	f.Close()
 	return true, nil
+}
+
+func validateGitAuth(logger *zap.SugaredLogger, url string) {
+	homeenv := os.Getenv("HOME")
+	sshCred := true
+	if _, err := os.Stat(homeenv + "/.ssh"); os.IsNotExist(err) {
+		sshCred = false
+	}
+	urlSSHFormat := ValidateGitSSHURLFormat(url)
+	if sshCred && !urlSSHFormat {
+		logger.Warnf("SSH credentials have been provided but the URL(%q) is not a valid SSH URL. This warning can be safely ignored if the URL is for a public repo or you are using basic auth", url)
+	} else if !sshCred && urlSSHFormat {
+		logger.Warnf("URL(%q) appears to need SSH authentication but no SSH credentials have been provided", url)
+	}
+}
+
+// ValidateGitSSHURLFormat validates the given URL format is SSH or not
+func ValidateGitSSHURLFormat(url string) bool {
+	if sshURLRegexFormat.MatchString(url) {
+		return true
+	}
+	return false
 }
--- a/pkg/git/git_test.go
+++ b/pkg/git/git_test.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2020 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package git
+
+import "testing"
+
+func TestValidateGitSSHURLFormat(t *testing.T) {
+	tests := []struct {
+		url  string
+		want bool
+	}{
+		{
+			url:  "git@github.com:user/project.git",
+			want: true,
+		},
+		{
+			url:  "git@127.0.0.1:user/project.git",
+			want: true,
+		},
+		{
+			url:  "http://github.com/user/project.git",
+			want: false,
+		},
+		{
+			url:  "https://github.com/user/project.git",
+			want: false,
+		},
+		{
+			url:  "http://127.0.0.1/user/project.git",
+			want: false,
+		},
+		{
+			url:  "https://127.0.0.1/user/project.git",
+			want: false,
+		},
+		{
+			url:  "http://host.xz/path/to/repo.git/",
+			want: false,
+		},
+		{
+			url:  "https://host.xz/path/to/repo.git/",
+			want: false,
+		},
+		{
+			url:  "ssh://user@host.xz:port/path/to/repo.git/",
+			want: true,
+		},
+		{
+			url:  "ssh://user@host.xz/path/to/repo.git/",
+			want: true,
+		},
+		{
+			url:  "ssh://host.xz:port/path/to/repo.git/",
+			want: true,
+		},
+		{
+			url:  "ssh://host.xz/path/to/repo.git/",
+			want: true,
+		},
+		{
+			url:  "git://host.xz/path/to/repo.git/",
+			want: false,
+		},
+		{
+			url:  "/path/to/repo.git/",
+			want: false,
+		},
+		{
+			url:  "file://~/path/to/repo.git/",
+			want: false,
+		},
+		{
+			url:  "user@host.xz:/path/to/repo.git/",
+			want: true,
+		},
+		{
+			url:  "host.xz:/path/to/repo.git/",
+			want: true,
+		},
+		{
+			url:  "user@host.xz:path/to/repo.git",
+			want: true,
+		},
+	}
+
+	for _, tt := range tests {
+		got := ValidateGitSSHURLFormat(tt.url)
+		if got != tt.want {
+			t.Errorf("Validate URL(%v)'s SSH format got %v, want %v", tt.url, got, tt.want)
+		}
+	}
+}


FogDong:
PTAL 😉  @sbwsg

pkg/git/git.go
```diff
@@ -31,6 +32,7 @@ import (
 const (
 	sshKnownHostsUserPath          = ".ssh/known_hosts"
 	sshMissingKnownHostsSSHCommand = "ssh -o StrictHostKeyChecking=accept-new"
+	sshURLFormat                   = `.+@([\w\d\.]+)(:[\d]+){0,1}/*(.*)`
```
sbwsg, line 35: From https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols

> To clone a Git repository over SSH, you can specify an ssh:// URL like this:
> 
> ```
> $ git clone ssh://[user@]server/project.git
> ```
> 
> Or you can use the shorter scp-like syntax for the SSH protocol:
> 
> ```
> $ git clone [user@]server:project.git
> ```
> 
> In both cases above, if you don’t specify the optional username, Git assumes the user you’re currently logged in as.

I'm not sure this regex captures all combinations (e.g. `user@` is optional. `:` can precede port or `<project>.git` in the scp syntax. It also looks like it's possible to have a url without `/`).  It would be great to introduce some testing to make sure we're covering as many different variations of this url as we can.  So it might make sense to perform the url check in a separate function that can be exercised independently of the `.ssh` directory checking.

pkg/git/git.go
```diff
@@ -244,3 +249,35 @@ func userHasKnownHostsFile(logger *zap.SugaredLogger) (bool, error) {
...
+	if euid == 0 {
+		if _, err := os.Stat("/root/.ssh"); os.IsNotExist(err) {
+			sshCred = false
+		}
+	} else {
+		if _, err := os.Stat(homepath + "/.ssh"); os.IsNotExist(err) {
+			sshCred = false
+		}
+	}
+	sshFormatRegex := regexp.MustCompile(sshURLFormat)
```
sbwsg, line 271: Given the format remains fixed we can probably compile this once at the top of the file? e.g.

```go
// near top of file
var sshFormatRegex = regexp.MustCompile(sshURLFormat)
```

pkg/git/git.go
FogDong: https://www.debuggex.com/r/LU8VCTiHXLMN1m8X

FogDong:
Fixed. 😆  Please review @sbwsg

FogDong:
Can anybody take a look at this PR? 🤔

/cc @vdemeester

pkg/git/git.go
```diff
@@ -244,3 +253,34 @@ func userHasKnownHostsFile(logger *zap.SugaredLogger) (bool, error) {
...
+	// Special case the root user/directory
+	if euid == 0 {
+		if _, err := os.Stat("/root/.ssh"); os.IsNotExist(err) {
+			sshCred = false
+		}
+	} else {
+		if _, err := os.Stat(homepath + "/.ssh"); os.IsNotExist(err) {
+			sshCred = false
+		}
+	}
```
vdemeester, line 274: @sbwsg correct me if I am wrong but we *could* use `/tekton/creds` path in order to use a "always" working path ? (although this makes `git-init` only works inside a tekton managed container)

pkg/git/git.go
```diff
@@ -244,3 +253,34 @@ func userHasKnownHostsFile(logger *zap.SugaredLogger) (bool, error) {
 	f.Close()
 	return true, nil
 }
+
+func validateGitAuth(logger *zap.SugaredLogger, url string) error {
```
vdemeester, line 257: nit: We could split the `.ssh` check and the regex :angel:

pkg/git/git.go
FogDong: The `ensureHomeEnv func` will create ssh symlink in `/tekton/home` so I change the stat path here.

pkg/git/git.go
```diff
@@ -244,3 +253,26 @@ func userHasKnownHostsFile(logger *zap.SugaredLogger) (bool, error) {
...
+	if sshCred && !urlSSHFormat {
+		logger.Warnf("The given URL does not match the ssh credential")
+	} else if !sshCred && urlSSHFormat {
+		logger.Warnf("The given ssh URL does not match the credential")
+	}
+
+	return nil
+}
+
+func validateGitSSHURLFormat(url string) bool {
```
sbwsg, line 273: It would be really useful to include tests for this function, even though it's private, to make sure that we don't accidentally break any accepted SSH URL formats in future.

pkg/git/git.go
```diff
@@ -244,3 +253,26 @@ func userHasKnownHostsFile(logger *zap.SugaredLogger) (bool, error) {
 	f.Close()
 	return true, nil
 }
+
+func validateGitAuth(logger *zap.SugaredLogger, url string) error {
+	homeenv := os.Getenv("HOME")
+	sshCred := true
+	if _, err := os.Stat(homeenv + "/.ssh"); os.IsNotExist(err) {
```
sbwsg, line 260: Thinking about this some more I'm not sure this is enough to correctly validate SSH credentials.  A user can mix SSH and non-SSH git credentials in a single Step. For example, with two Secrets:

```
apiVersion: v1
kind: Secret
metadata:
  name: example-gitlab-creds
  annotations:
    tekton.dev/git-0: https://gitlab.com
type: kubernetes.io/basic-auth
stringData:
  username: myuser
  password: hunter2
---
apiVersion: v1
kind: Secret
metadata:
  name: example-github-creds
  annotations:
    tekton.dev/git-0: github.com
type: kubernetes.io/ssh-auth
stringData:
  ssh-privatekey: blahblahblah
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: example-service-account
secrets:
  - name: example-gitlab-creds
  - name: example-github-creds
---
kind: TaskRun
metadata:
  name: foobar
spec:
  serviceAccountName: example-service-account
  taskSpec:
    steps:
    - name: wait
      image: alpine:latest
      script: |
        echo "foo"
```

In this example YAML both a `~/.ssh` directory and a `~/.git-credentials` file will exist side by side.  `~/.git-credentials` is created for the gitlab Secret.  So this code will print a warning because the gitlab basic auth URL is not SSH-formatted.  But the user never intends to use SSH auth with a gitlab URL.

The full solution to this problem is probably quite a lot more complicated - it would need to read the `~/.ssh/config` file and figure out whether the given `url` is intended to be used with SSH authentication.

We could work around this somewhat by updating the warning messages to be more relaxed.  Something like:

```
logger.Warnf("SSH credentials have been provided but the URL %q is not a valid SSH URL", url)
```

But this feels like it would lead to more user confusion and I don't think this is a great workaround either.

pkg/git/git.go
```diff
@@ -244,3 +253,26 @@ func userHasKnownHostsFile(logger *zap.SugaredLogger) (bool, error) {
...
+		sshCred = false
+	}
+	urlSSHFormat := validateGitSSHURLFormat(url)
+	if sshCred && !urlSSHFormat {
+		logger.Warnf("The given URL does not match the ssh credential")
+	} else if !sshCred && urlSSHFormat {
+		logger.Warnf("The given ssh URL does not match the credential")
+	}
+
+	return nil
```
sbwsg, line 270: We don't seem to ever return an error from this func, only print warnings.  Not sure if `error` is needed in function signature.

pkg/git/git.go
```diff
@@ -244,3 +253,26 @@ func userHasKnownHostsFile(logger *zap.SugaredLogger) (bool, error) {
...
+
+func validateGitAuth(logger *zap.SugaredLogger, url string) error {
+	homeenv := os.Getenv("HOME")
+	sshCred := true
+	if _, err := os.Stat(homeenv + "/.ssh"); os.IsNotExist(err) {
+		sshCred = false
+	}
+	urlSSHFormat := validateGitSSHURLFormat(url)
+	if sshCred && !urlSSHFormat {
+		logger.Warnf("The given URL does not match the ssh credential")
```
sbwsg, line 265: It would be great to include the URL that did not match expected format.  `The given URL %q does not match...`

pkg/git/git.go
FogDong: I met the similar problem when I tested the code: if the given URL is a public repo and the cred is SSH type, the warning log will be shown and the pod can still pull the code successfully.
But it will be more too complex if we check the URL is public or not. So my point is we can just let the warning show since it's only a warning and people will not care about the warning if the job success. 🤔 WDYT?

pkg/git/git.go
```diff
@@ -244,3 +251,25 @@ func userHasKnownHostsFile(logger *zap.SugaredLogger) (bool, error) {
...
+	homeenv := os.Getenv("HOME")
+	sshCred := true
+	if _, err := os.Stat(homeenv + "/.ssh"); os.IsNotExist(err) {
+		sshCred = false
+	}
+	urlSSHFormat := ValidateGitSSHURLFormat(url)
+	if sshCred && !urlSSHFormat {
+		logger.Warnf("SSH credentials have been provided but the URL(%q) is not a valid SSH URL", url)
+	} else if !sshCred && urlSSHFormat {
+		logger.Warnf("The given SSH URL(%q) does not match the credential", url)
```
sbwsg, line 265: Suggest slightly different log message to make it clearer what is happening:

```
URL(%q) appears to need SSH authentication but no SSH credentials have been provided
```

pkg/git/git_test.go
```diff
@@ -0,0 +1,145 @@
...
+		{
+			url:  "host.xz:~user/path/to/repo.git/",
+			want: false,
+		},
+		{
+			url:  "user@host.xz:path/to/repo.git",
+			want: true,
+		},
+		{
+			url:  "host.xz:path/to/repo.git",
```
sbwsg, line 134: I think this is a valid SSH url for git in the "scp-like syntax"?  `git` assumes the `user@` portion is the currently logged-in user.  So if current user in container is `git` then this will be interpreted as `git@host.xz:path/to/repo.git`

pkg/git/git.go
sbwsg: Hrm, tricky problem!  OK, I think we can make this work with an even longer log message:

```
SSH credentials have been provided but the URL %q is not a valid SSH URL. This warning can be safely ignored if the URL is for a public repo or you are using basic auth
```

@vdemeester what do you think of this?  Is it too vague?  The other thing we could do is drop this message and only print a warning if a URL is _definitely_ an SSH url but the user has 0 SSH credentials available in the Step.  That is 100% of the time a legitimate error I think?

sbwsg (commented):
Sorry, I've left more comments.  I'm not totally sure if we want the warning message when SSH credentials are provided but the URL is non-SSH.  Might be too vague?  Curious what others think about this!

pkg/git/git_test.go
FogDong: The url with `user@` should be a valid SSH url like the example above, but without `user@`, it's not a valid one.

pkg/git/git.go
FogDong: I'll change the message first but I'm also not quite sure that if we need that message when SSH credentials are provided but the URL is non-SSH. Let me know if there is better solutions!

pkg/git/git_test.go
vdemeester: @FogDong that's @sbwsg's point. From a git perspective, `git clone host.xz:path/to/repo.git` is valid (it will use the current logged in user, or the user specified in the `.ssh/config`)

sbwsg: Indeed, sorry, my explanation was not clear.  From https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols :

> $ git clone [user@]server:project.git
> [...] **if you don’t specify the optional username, Git assumes the user you’re currently logged in as.**

sbwsg (approved):
Thanks @FogDong !

FogDong:
Can I get a lgtm? /cc @vdemeester 😉

withlin (approved):
/lgtm

bobcatfish:
@FogDong @sbwsg can you update the release note with more information about what this change is doing?

sbwsg:
@bobcatfish sure thing; updated!

PR #3138, opened 2020-08-26
Introducing InternalTektonResultType as a ResultType

Peaorl:
# Changes

This PR introduces a new TektonResultType namely 'InternalTektonResultType'.
Currently `TaskRunResultType`, `PipelineResourceResultType` and `UnknownResultType` exist as a ResultType for a `PipelineResourceResult`.
All results that conform to one of these `ResultTypes` are normally exposed as either as a task or resource result.

In light of #3087 the need for a `ResultType` that is not exposed arises.
In #3087, a `Step` can emit a result indicating a `Step` timeout has occurred.
This is a result that should not be exposed hence the need for a new `ResultType` named `InternalTektonResultType`.
This PR ensures results of this type are filtered out.

Introducing an `InternalTektonResultType` ensures a future proof solution to internal results that should not be exposed as a TaskRun or PipelineResource result.
Aside from the example in #3087, a present candidate is the result written out by a Step containing a "StartedAt" key.
Up until now this result is filtered out by a specific function. 
Marking it as an `InternalTektonResultType` now allows for this result to automatically be filtered out.

Up until this PR `Sidecars` can write results.
Since this is not explicitly documented it's unclear whether this is intended behavior.
In this PR it's ensured only `Steps` can write results.

Additionally this commit brings about refactoring/moving (and sometimes renaming) of functions related to converting `Pod` statuses to `TaskRun` statuses from `pkg/reconciler/taskrun/taskrun.go` to `pkg/pod/status/status.go`. 
This is accompanied with moving unit tests from `taskrun_test.go` to `status_test.go`.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```release-note
NONE
```


pkg/pod/status.go
```diff
@@ -97,27 +96,51 @@ func SidecarsReady(podStatus corev1.PodStatus) bool {
...
+				results, err = termination.ParseMessage(s.State.Terminated.Message)
 				if err != nil {
-					logger.Errorf("error setting the start time of step %q in taskrun %q: %w", s.Name, tr.Name, err)
-				}
-				if time != nil {
-					s.State.Terminated.StartedAt = *time
-					s.State.Terminated.Message = message
+					logger.Errorf("termination message could not be parsed as JSON: %v", err)
+				} else {
+					//Further processing if the termination message is JSON formatted
```
sbwsg, line 130: nit: I think the code here is readable and doesn't need an additional comment, but it's only a nit - leave if you still think this is useful.

pkg/pod/status.go
```diff
@@ -135,51 +158,70 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev
...
-	} else {
-		updateIncompleteTaskRun(trs, pod)
-	}
+	trs.TaskRunResults = removeDuplicateResults(trs.TaskRunResults)
 
 	// Sort step states according to the order specified in the TaskRun spec's steps.
 	trs.Steps = sortTaskRunStepOrder(trs.Steps, taskSpec.Steps)
 
-	return *trs
+	return *trs, err
```
sbwsg, line 166: Hm, it's ambiguous to me what the final returned error will be - it seems like it will be whatever the last call to `extractStartedAtTimeFromResults(results)` or `termination.ParseMessage(s.State.Terminated.Message)` returned.  In some cases this could be `nil` (since the last Step could Parse just fine) even if there were errors in earlier Steps?

It might be worth using the multierror package to aggregate together the complete list of errors?  See [this line from the taskrun reconciler](https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/taskrun/taskrun.go#L117) that shows usage of multierror.

Peaorl: Good catch, I implemented multierrors now!

pkg/pod/status.go
```diff
@@ -135,51 +166,86 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev
...
 
 	// Sort step states according to the order specified in the TaskRun spec's steps.
 	trs.Steps = sortTaskRunStepOrder(trs.Steps, taskSpec.Steps)
 
-	return *trs
+	return *trs, merr.ErrorOrNil()
 }
 
-// removeStartInfoFromTerminationMessage searches for a result called "StartedAt" in the JSON-formatted
+// createMessageFromResults searches for a result called "StartedAt" in the JSON-formatted
```
sbwsg, line 177: it looks like this comment is out of date now.  the func doesn't do anything special for `StartedAt` anymore.

pkg/pod/status.go
```diff
@@ -97,27 +98,57 @@ func SidecarsReady(podStatus corev1.PodStatus) bool {
...
+	}
+
 	trs.PodName = pod.Name
 	trs.Steps = []v1beta1.StepState{}
 	trs.Sidecars = []v1beta1.SidecarState{}
 
+	var merr *multierror.Error
+
 	for _, s := range pod.Status.ContainerStatuses {
 		if IsContainerStep(s.Name) {
```
sbwsg, line 124: we might be able to reduce the nesting and improve readability a bit here if we split the list of statuses into step / sidecar first and then run a for loop over each:

```go
var stepStatuses []corev1.ContainerStatus
var sidecarStatuses []corev1.ContainerStatus
for _, s := range pod.Status.ContainerStatuses {
  if IsContainerStep(s.Name) {
    stepStatuses = append(stepStatuses, s)
  } else if isContainerSidecar(s.Name) {
    sidecarStatuses = append(sidecarStatuses, s)
  }
}

for _, s := range stepStatuses {
  // do step specific stuff
}

for _, s := range sidecarStatuses {
  // do sidecar specific stuff
}
```

Could even have a step-specific and sidecar-specific func?  up to you.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -185,7 +184,11 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, tr *v1beta1.TaskRun) pkg
 	// Reconcile this copy of the task run and then write back any status
 	// updates regardless of whether the reconciliation errored out.
 	if err = c.reconcile(ctx, tr, taskSpec, rtr); err != nil {
-		logger.Errorf("Reconcile error: %v", err.Error())
+		if merr, ok := err.(*multierror.Error); ok {
```
sbwsg, line 187: I think you can just print the `merr.Error()` and it will print them all in a list for you?  Is that not happening?

Peaorl: That works!

pkg/pod/status.go
Peaorl: Thanks! Separated them out into specific functions.

Peaorl:
While migrating unit tests from `pkg/reconciler/taskrun/taskrun_test.go` to `pkg/pod/status/status.go` I encountered one [test case](https://github.com/tektoncd/pipeline/blob/9bc3fd8097e13eecc28f6afe165bfb91bf55f42c/pkg/reconciler/taskrun/taskrun_test.go#L2522) I don't fully understand. 
I believe this test case aims to check whether a taskrun status is set to false if the corresponding task fails. 
Nevertheless, this test case does not actually execute a function and the [function](https://github.com/tektoncd/pipeline/blob/9bc3fd8097e13eecc28f6afe165bfb91bf55f42c/pkg/reconciler/taskrun/taskrun.go#L630) it presumably would call is not responsible for setting the taskrun status. The described `podStatus` is also not set to fail.
In this PR I changed the [test](https://github.com/tektoncd/pipeline/blob/c6f1b5caa002dd9b7d5994d27b7a93fde045a48e/pkg/pod/status_test.go#L783) such that if a task (pod) fails it's checked whether the taskrun status is set accordingly. 

Another thing I noticed in the same test is that the description has a [termination message](https://github.com/tektoncd/pipeline/blob/9bc3fd8097e13eecc28f6afe165bfb91bf55f42c/pkg/reconciler/taskrun/taskrun_test.go#L2527) with a result that does not follow the struct format of a [`PipelineResourceResult`](https://github.com/tektoncd/pipeline/blob/9bc3fd8097e13eecc28f6afe165bfb91bf55f42c/pkg/apis/pipeline/v1beta1/resource_types.go#L122). For example, the key `name` does not exist as `PipelineResourceResult` field.
On master and also in this PR, such "results" get dropped from the taskrun termination message.
Once the termination message is [unmarshalled](https://github.com/tektoncd/pipeline/blob/9bc3fd8097e13eecc28f6afe165bfb91bf55f42c/pkg/pod/status.go#L114) into [`PipelineResourceResults`](https://github.com/tektoncd/pipeline/blob/9bc3fd8097e13eecc28f6afe165bfb91bf55f42c/pkg/pod/status.go#L114), JSON objects that don't adhere to the `PipelineResourceResult` structure are dropped without error by the [unmarshal](https://golang.org/pkg/encoding/json/#Unmarshal) function.

I am wondering, is this behavior intentional? On master, no test case catches this behavior.
[This](https://github.com/tektoncd/pipeline/blob/c6f1b5caa002dd9b7d5994d27b7a93fde045a48e/pkg/pod/status_test.go#L813) test now records the behavior of dropping incorrect results from the taskrun termination message.
Additionally, [ParseMessage](https://github.com/tektoncd/pipeline/blob/c6f1b5caa002dd9b7d5994d27b7a93fde045a48e/pkg/termination/parse.go#L46) now logs incorrect results.
Alternatively,  we could decide that incorrect results are kept "as is" for the taskrun termination message.
Note that it's normally impossible for incorrect results to show up in the taskrun termination message if the Tekton code only writes `PipelineResourceResults` to its termination message.

Lastly, [ResourceRef](https://github.com/tektoncd/pipeline/blob/c6f1b5caa002dd9b7d5994d27b7a93fde045a48e/pkg/apis/pipeline/v1beta1/resource_types.go#L113) is now a pointer. This ensures that an empty `resourceRef` field is not marshalled  back into the termination message (zero values like a nil pointer are not marshalled) as is the [case](https://github.com/tektoncd/pipeline/blob/9bc3fd8097e13eecc28f6afe165bfb91bf55f42c/pkg/termination/write_test.go#L65) on master currently.

pkg/apis/pipeline/v1beta1/resource_types.go
```diff
@@ -123,10 +123,10 @@ type PipelineResourceResult struct {
 	Key          string `json:"key"`
 	Value        string `json:"value"`
 	ResourceName string `json:"resourceName,omitempty"`
-	// This field should be deprecated and removed in the next API version.
+	// The field ResourceName should be deprecated and removed in the next API version.
```
sbwsg, line 126: Is this right?  I have to confess I'm a bit confused by these fields and #2694 doesn't make it totally clear.  @dlorenc is it ResourceName or ResourceRef that's supposed to be deprecated?

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -185,7 +184,9 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, tr *v1beta1.TaskRun) pkg
 	// Reconcile this copy of the task run and then write back any status
 	// updates regardless of whether the reconciliation errored out.
 	if err = c.reconcile(ctx, tr, taskSpec, rtr); err != nil {
-		logger.Errorf("Reconcile error: %v", err.Error())
```
sbwsg, line 188: Looking at this again, I don't totally understand why this changed.  I think we want to print any non-nil error regardless of whether it's a multierror or not?

sbwsg (commented):
> While migrating unit tests from pkg/reconciler/taskrun/taskrun_test.go to pkg/pod/status/status.go I encountered one test case I don't fully understand.
> I believe this test case aims to check whether a taskrun status is set to false if the corresponding task fails.
> Nevertheless, this test case does not actually execute a function and the function it presumably would call is not responsible for setting the taskrun status. The described podStatus is also not set to fail.
> In this PR I changed the test such that if a task (pod) fails it's checked whether the taskrun status is set accordingly.

Good catch.  I think the idea of the original test was to confirm that a failed Task does not generate any Results.  But as you say this test didn't actually call any functions and didn't test anything 😨 

> Another thing I noticed in the same test is that the description has a termination message with a result that does not follow the struct format of a PipelineResourceResult. For example, the key name does not exist as PipelineResourceResult field.
> On master and also in this PR, such "results" get dropped from the taskrun termination message.
> Once the termination message is unmarshalled into PipelineResourceResults, JSON objects that don't adhere to the PipelineResourceResult structure are dropped without error by the unmarshal function.

`digest` and `name` were fields used in the alpha API that have since been removed. Have a look through the git history of `pkg/apis/pipeline/v1alpha1/resource_types.go`, for example, and you'll spot those field names. I think the best course of action is what you've done - removing the old test and codifying the behavior in a new test.

> Lastly, ResourceRef is now a pointer. This ensures that an empty resourceRef field is not marshalled back into the termination message (zero values like a nil pointer are not marshalled) as is the case on master currently.

👍 I think this is totally fine.

sbwsg:
I left some small nits but no blockers from my perspective!

/approve

pkg/reconciler/taskrun/taskrun.go
Peaorl: Correct, changed it!

pkg/pod/status.go
```diff
@@ -97,89 +98,177 @@ func SidecarsReady(podStatus corev1.PodStatus) bool {
...
+	var merr *multierror.Error
+
+	for _, s := range stepStatuses {
+		if s.State.Terminated != nil && len(s.State.Terminated.Message) != 0 {
+			msg := s.State.Terminated.Message
+
+			results, err := termination.ParseMessage(logger, msg)
+			if err != nil {
+				logger.Errorf("termination message could not be parsed as JSON: %v", err)
+				merr = multierror.Append(merr, err)
```
ImJasonH, line 156: style nit: If you `continue` after this `multierror.Append`, the following `else` block can be outside the `if` and unindented one tab-width.

```go
if err != nil {
  merr = append(...)
  continue
}
time, err := ...
```

pkg/pod/status_test.go
```diff
@@ -694,47 +1023,37 @@ func TestMakeTaskRunStatus(t *testing.T) {
...
-				Status: v1beta1.TaskRunStatus{
-					TaskRunStatusFields: v1beta1.TaskRunStatusFields{
-						StartTime: &metav1.Time{Time: startTime},
-					},
-				},
 			}
 
 			logger, _ := logging.NewLogger("", "status")
-			got := MakeTaskRunStatus(logger, tr, pod, c.taskSpec)
+			got, err := MakeTaskRunStatus(logger, tr, pod, c.taskSpec)
```
ImJasonH, line 1041: Can we check this error closer to where it's initialized?

ImJasonH (commented):
Mostly lgtm, just a few style/readability comments. 👍

pkg/pod/status.go
Peaorl: I think in this case we need the `else` statement since this code:
```go
		trs.Steps = append(trs.Steps, v1beta1.StepState{
			ContainerState: *s.State.DeepCopy(),
			Name:           trimStepPrefix(s.Name),
			ContainerName:  s.Name,
			ImageID:        s.ImageID,
		})
```
right after the `if`/`else` block needs to happen regardless, but only after the `if`/`else` block.

ImJasonH (commented):
/lgtm

PR #3141, opened 2020-08-26
use status helper functions in pod status unit tests

adshmh:
# Changes

use status helper functions in pod status unit tests. Addresses https://github.com/tektoncd/pipeline/pull/2804#issuecomment-654527839

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

    ```release-note
    NONE
    ```

### DIFF
--- a/pkg/pod/status_test.go
+++ b/pkg/pod/status_test.go
@@ -33,19 +33,6 @@ import (
 
 var ignoreVolatileTime = cmp.Comparer(func(_, _ apis.VolatileTime) bool { return true })
 
-var conditionRunning apis.Condition = apis.Condition{
-	Type:    apis.ConditionSucceeded,
-	Status:  corev1.ConditionUnknown,
-	Reason:  v1beta1.TaskRunReasonRunning.String(),
-	Message: "Not all Steps in the Task have finished executing",
-}
-var conditionSucceeded apis.Condition = apis.Condition{
-	Type:    apis.ConditionSucceeded,
-	Status:  corev1.ConditionTrue,
-	Reason:  v1beta1.TaskRunReasonSuccessful.String(),
-	Message: "All Steps have completed executing",
-}
-
 func TestMakeTaskRunStatus(t *testing.T) {
 	for _, c := range []struct {
 		desc      string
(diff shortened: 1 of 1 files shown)


pkg/pod/status_test.go
```diff
@@ -971,3 +867,32 @@ func TestMarkStatusSuccess(t *testing.T) {
 		t.Errorf("Unexpected status: %s", diff.PrintWantGot(d))
 	}
 }
+
+func statusRunning() duckv1beta1.Status {
```
pritidesai, line 871: thank you @adshmh for this refactoring 🙏 would like to see using [t.Helper()](https://blog.golang.org/go1.9) here so that it reports accurate line number in case of failure. And similarly, rest of the three functions statusFailure, statusSuccess, and statusPending.

adshmh: Thank you for the review. Fixed.

pkg/pod/status_test.go
```diff
@@ -1290,3 +1167,36 @@ func TestMarkStatusSuccess(t *testing.T) {
...
+	t.Helper()
+	return duckv1beta1.Status{
+		Conditions: []apis.Condition{{
+			Type:    apis.ConditionSucceeded,
+			Status:  corev1.ConditionUnknown,
+			Reason:  reason,
+			Message: message,
+		}},
+	}
+}
```
vdemeester, line 1202: All those methods do not need `t *testing.T` :angel:

vdemeester (changes requested):

pkg/pod/status_test.go
pritidesai: yes `t.Helper()` needs access to `*t testing.T` variable

vdemeester: @pritidesai no need to call `t.Helper` if the function itself doesn't *use* `*testing.T`. `t.Helper()` is only useful (and needed) if you want the *report* (`t.Log`, `t.Error`, …) done in a function to be reported as the *caller* of this function. Here, we don't use `t *testing.T` at all in these functions, so we don't need to call `t.Helper()`, and so we don't need to pass `*testing.T`

sbwsg:
@adshmh just checking in, do you want to carry on with this PR?  Looks like there is still some open feedback to address.  Cheers!

adshmh:
> @adshmh just checking in, do you want to carry on with this PR? Looks like there is still some open feedback to address. Cheers!

Thank you for the reminder. I have updated the PR to address all the reviews.

sbwsg:
Thanks for refactoring!

/approve

vdemeester (commented):
/lgtm

PR #3156, opened 2020-09-01
Make DefaultThreadsPerController, QPS and Burst configurable via flags

pierretasci:
# Changes

Make DefaultThreadsPerController, QPS, and Burst configurable via flags. This helps tune a deployment depending on expected concurrency in production.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Allows DefaultThreadsPerController, QPS, and Burst to be configured via flags
```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -25,6 +25,8 @@ import (
 	"github.com/tektoncd/pipeline/pkg/reconciler/taskrun"
 	"github.com/tektoncd/pipeline/pkg/version"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/rest"
+	"knative.dev/pkg/controller"
 	"knative.dev/pkg/injection"
 	"knative.dev/pkg/injection/sharedmain"
 	"knative.dev/pkg/signals"
@@ -48,6 +50,12 @@ var (
 	imageDigestExporterImage = flag.String("imagedigest-exporter-image", "", "The container image containing our image digest exporter binary.")
 	namespace                = flag.String("namespace", corev1.NamespaceAll, "Namespace to restrict informer to. Optional, defaults to all namespaces.")
 	versionGiven             = flag.String("version", "devel", "Version of Tekton running")
+	qps                      = flag.Int("kube-api-qps", int(rest.DefaultQPS), "Maximum QPS to the master from this client")
+	burst                    = flag.Int("kube-api-burst", rest.DefaultBurst, "Maximum burst for throttle")
+	threadsPerController     = flag.Int("threads-per-controller", controller.DefaultThreadsPerController, "Threads (goroutines) to create per controller")
+	disableHighAvailability  = flag.Bool("disable-ha", false, "Whether to disable high-availability functionality for this component.  This flag will be deprecated "+
+		"and removed when we have promoted this feature to stable, so do not pass it without filing an "+
+		"issue upstream!")
 )
 
 func main() {
@@ -68,7 +76,18 @@ func main() {
 	if err := images.Validate(); err != nil {
 		log.Fatal(err)
 	}
-	sharedmain.MainWithContext(injection.WithNamespaceScope(signals.NewContext(), *namespace), ControllerLogKey,
+	controller.DefaultThreadsPerController = *threadsPerController
+
+	cfg := sharedmain.ParseAndGetConfigOrDie()
+	// multiply by 2, no of controllers being created
+	cfg.QPS = 2 * float32(*qps)
+	cfg.Burst = 2 * *burst
+
+	ctx := injection.WithNamespaceScope(signals.NewContext(), *namespace)
+	if !*disableHighAvailability {
+		ctx = sharedmain.WithHADisabled(ctx)
+	}
+	sharedmain.MainWithConfig(ctx, ControllerLogKey, cfg,
 		taskrun.NewController(*namespace, images),
 		pipelinerun.NewController(*namespace, images),
 	)
--- a/pkg/pod/entrypoint.go
+++ b/pkg/pod/entrypoint.go
@@ -209,7 +209,7 @@ func StopSidecars(nopImage string, kubeclient kubernetes.Interface, pod corev1.P
 	}
 	if updated {
 		if _, err := kubeclient.CoreV1().Pods(newPod.Namespace).Update(newPod); err != nil {
-			return fmt.Errorf("error adding ready annotation to Pod %q: %w", pod.Name, err)
+			return fmt.Errorf("error stopping sidecars of Pod %q: %w", pod.Name, err)
 		}
 	}
 	return nil
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -122,6 +122,7 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, tr *v1beta1.TaskRun) pkg
 		c.timeoutHandler.Release(tr.GetNamespacedName())
 		pod, err := c.KubeClientSet.CoreV1().Pods(tr.Namespace).Get(tr.Status.PodName, metav1.GetOptions{})
 		if err == nil {
+			logger.Debugf("Stopping sidecars for TaskRun %q of Pod %q", tr.Name, tr.Status.PodName)
 			err = podconvert.StopSidecars(c.Images.NopImage, c.KubeClientSet, *pod)
 			if err == nil {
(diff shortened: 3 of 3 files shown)


cmd/controller/main.go
```diff
@@ -65,7 +72,18 @@ func main() {
...
 	}
-	sharedmain.MainWithContext(injection.WithNamespaceScope(signals.NewContext(), *namespace), ControllerLogKey,
+	controller.DefaultThreadsPerController = *threadsPerController
+
+	cfg := sharedmain.ParseAndGetConfigOrDie()
+	cfg.QPS = float32(*qps)
+	cfg.Burst = *burst
+
+	ctx := injection.WithNamespaceScope(signals.NewContext(), *namespace)
+	if !*enableLeaderElection {
```
dlorenc, line 82: The leader election part worries me a bit. I think we'd want to setup some end to end testing of this before making it available. 

Do you think it's worth trying to include in this change, or separating out to another PR where we can figure out the necessary testing plan?

dlorenc:
nit: i think you have a few extra "```" characters in the release notes section as well :)

cmd/controller/main.go
nishpa214: To my understanding, I _think_ if we remove the option of enabling leader election, it might be a breaking change, as the current release seems to support it via https://github.com/tektoncd/pipeline/blob/master/config/config-leader-election.yaml. I am not too super familiar with knative but looking through PR https://github.com/knative/pkg/pull/1476, looks like one could enable it by setting `EnabledComponents`.

I could break that out in a separate commit though, if we want to add some sort end to end test for it. I might need some guidance with that though, not sure what we might test in addition to what knative is already testing for :)

Either way, I guess we will want to add something in the release notes, if folks are using leader election now and how to use going forward.

vdemeester:
/hold
This bumps some k8s and knative dependencies. I guess the knative dependencies bump are required right ? (and the k8s ones might be a result of that update). We may want to split thoses (either in separate PR and/or commit), and we need to look at the knative changes (like the min support k8s version, …) before merging this.

cmd/controller/main.go
dlorenc: Ah, maybe my mistake then. It's not actually clear to me if that works. The comment is "an inactive but valid configuration follows": https://github.com/tektoncd/pipeline/blob/master/config/config-leader-election.yaml#L24

But either way - I'm not quite sure how not including this flag would be a breaking change. It didn't exist before, why do we need it here?

nishpa214:
> This bumps some k8s and knative dependencies. I guess the knative dependencies bump are required right ? (and the k8s ones might be a result of that update).

@vdemeester yes, knative dependencies bump was required to pick up the change to make QPS and Burst configurable. 

> We may want to split thoses (either in separate PR and/or commit), and we need to look at the knative changes (like the min support k8s version, …) before merging this.

Cool, will create another PR for bumping knative changes.

vdemeester:
For some reason, the tests error seems legit. Looking at the code, the default behavior shouldn't change, but it seems something definitely changes breaking the timeout behavior.

nishpa214:
Thanks @vdemeester for helping me out w/ this. As discussed in Slack, we are going to try and add one change at a time to see if we can find the root cause.

nishpa214:
Looks like docker failure

cmd/controller/main.go
```diff
@@ -48,6 +50,12 @@ var (
 	imageDigestExporterImage = flag.String("imagedigest-exporter-image", "", "The container image containing our image digest exporter binary.")
 	namespace                = flag.String("namespace", corev1.NamespaceAll, "Namespace to restrict informer to. Optional, defaults to all namespaces.")
 	versionGiven             = flag.String("version", "devel", "Version of Tekton running")
+	qps                      = flag.Int("kube-api-qps", int(rest.DefaultQPS), "Maximum QPS to the master from this client")
```
nishpa214, line 53: Using same flags as being proposed in knative https://github.com/tektoncd/pipeline/pull/3156

nishpa214, line 53: Using same flags as being proposed in knnative: https://github.com/knative/pkg/issues/1700

nishpa214:
another docker issue :( 

cmd/controller/main.go
```diff
@@ -48,6 +50,12 @@ var (
...
+	burst                    = flag.Int("kube-api-burst", rest.DefaultBurst, "Maximum burst for throttle")
+	threadsPerController     = flag.Int("threads-per-controller", controller.DefaultThreadsPerController, "Threads (goroutines) to create per controller")
+	disableHighAvailability  = flag.Bool("disable-ha", false, "Whether to disable high-availability functionality for this component.  This flag will be deprecated "+
```
nishpa214, line 56: Any guidance on this? I choose to keep the same flag as what `MainWithConfig ` would have. See: https://github.com/knative/pkg/blob/1e373a9e5dea1697474130224289d0ee1ec96261/injection/sharedmain/main.go#L182

nishpa214:
@dlorenc tests are all good. Please take a look when you can! :)

sbwsg:
Code looks good to me.  Minor bookkeeping thing: could you update the commit message to include a slightly more in-depth description of the change.  Something like:

> Prior to this commit QPS, threads-per-controller and maximum burst were not configurable when launching the Tekton controller.  This meant that Tekton deployments could not be tuned to the environments it was deployed in.
> 
> This commit makes DefaultThreadsPerController, QPS, and Burst configurable via flags. This helps tune a deployment depending on expected concurrency in production.

sbwsg:
Since the most recent push is a commit message update I'm going to re-add @vdemeester 's lgtm and add my own approve:

/lgtm
/approve

cmd/controller/main.go
```diff
@@ -68,7 +76,18 @@ func main() {
...
-	sharedmain.MainWithContext(injection.WithNamespaceScope(signals.NewContext(), *namespace), ControllerLogKey,
+	controller.DefaultThreadsPerController = *threadsPerController
+
+	cfg := sharedmain.ParseAndGetConfigOrDie()
+	// multiply by 2, no of controllers being created
+	cfg.QPS = 2 * float32(*qps)
+	cfg.Burst = 2 * *burst
+
+	ctx := injection.WithNamespaceScope(signals.NewContext(), *namespace)
+	if !*disableHighAvailability {
```
afrittoli, line 87: This seems to be backwards? If disable is false, we should enable, but we're getting the context with HA disabled.

vdemeester: Oh right, I think the flag used to be `enableLeaderElection` (or something). cc @nishpa214

xiujuan95:
@pierretasci I very like this PR. But could you pls tell me how can I configure the QPS, thread and burst? By a configMap, deployment or other kubernetes resources? And How can I check the change works? For example, the default QPS is 20, I want to modify it to 30.  Then how can know I have changed it successfully? Is one kubernetes resource will show it?

xiujuan95:
@pierretasci Sorry bother you again! 

I want to change the default Thread/QPS/Burst to be 32/100/100, not sure if below setting is correct or not?
```
containers:
      - args:
        - -kube-api-qps=100
        - -kube-api-burst=100
        - -threads-per-controller=32
```
Because I don't know how to confirm this change takes effect, I need to get a sure answer about how to set the flag from you. Thanks in advance!

pierretasci:
> @pierretasci Sorry bother you again!
> 
> I want to change the default Thread/QPS/Burst to be 32/100/100, not sure if below setting is correct or not?
> 
> ```
> containers:
>       - args:
>         - -kube-api-qps=100
>         - -kube-api-burst=100
>         - -threads-per-controller=32
> ```
> 
> Because I don't know how to confirm this change takes effect, I need to get a sure answer about how to set the flag from you. Thanks in advance!

@xiujuan95: @nishpa214 should be able to help with that

nishpa214:
@xiujuan95 You want to set those parameters for `tekton-pipelines-controller` container in deployment spec. Here: https://github.com/tektoncd/pipeline/blob/master/config/controller.yaml#L76


You want to set them as following:
```sh
...
"-shell-image", "gcr.io/distroless/base:debug@sha256:72a0093a0214e414527a97d359313992534f94a689449615875d922097f0ba62",
"-kube-api-qps", "100",
"-kube-api-burst", "100",
"-threads-per-controller": "32"
```

xiujuan95:
@nishpa214 Thanks! 
But for the last one `-threads-per-controller`, I think it should be `"-threads-per-controller": "32"` rather than `"-threads-per-controller", "32"`, right?

PR #3160, opened 2020-09-01
fixing nightly unit test failure

### LINKED ISSUE #2815: Nightly unit test failure due to race condition when initializing config stores
# Expected Behavior

Nightly build unit test passes

# Actual Behavior

Nightly build unit test failed:

TaskRun is: `pipeline-release-nightly-gkd24-unit-tests-l2pdb `

Failure: 
```
  --- FAIL: TestReconcile_ExplicitDefaultSA/success (0.08s)
```

# Additional Info

- Tekton Pipeline version: v0.13.2

- I tried to reproduce this locally and failed: 
```
 go test -race -cover -count=10 -run="TestReconcile_ExplicitDefaultSA" ./pkg/reconciler/taskrun
```

pritidesai:
# Changes

Unless service account is explicitly set in taskrun spec, by default, service account is set to "default".

Scott's error [capture](https://gist.github.com/sbwsg/ce24e70e389e205dcb9d0890afc4193c#file-testreconcile_explicitdefaultsa_success_failed-log-L17) says:

```
* failed to create task run pod "test-taskrun-run-success": translating TaskSpec to Pod: serviceaccounts "default" not found. Maybe missing or invalid Task foo/test-task
```

`test-taskrun-run-success` taskrun has no service account set. Later in the config map, `default-service-account` is set to `pipelines`. But if service account is not specified in the `taskrun` spec, task run controller sets the service account to `default` which is clear from the error message where controller is looking for service account `default` instead of `pipelines`.

It can also mean that the config map settings/updates done in the unit test are not visible to taskrun controller which needs fix. 

This error is intermittent and time plays role here since the other unit tests in `taskrun_test.go` are creating `default` service account (and never deleted). If this unit test `TestReconcile_ExplicitDefaultSA` is the first one to start, we are missing `default` service account and fails with this kind of error.

Fixes #2815 

/cc @bobcatfish @sbwsg 

/kind bug

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Fixed nightly build failure due to unit test failure
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -359,7 +359,7 @@ func TestReconcile_ExplicitDefaultSA(t *testing.T) {
 		tb.TaskRunServiceAccountName("test-sa"),
 	))
 	taskruns := []*v1beta1.TaskRun{taskRunSuccess, taskRunWithSaSuccess}
-	defaultSAName := "pipelines"
+	defaultSAName := "default"
 	d := test.Data{
 		TaskRuns: taskruns,
 		Tasks:    []*v1beta1.Task{simpleTask, saTask},


pritidesai:
@afrittoli we could also look into `configMap` settings to see why `default-service-account` is not getting set to `pipelines` instead its set to `default`.

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -359,7 +359,7 @@ func TestReconcile_ExplicitDefaultSA(t *testing.T) {
 		tb.TaskRunServiceAccountName("test-sa"),
 	))
 	taskruns := []*v1beta1.TaskRun{taskRunSuccess, taskRunWithSaSuccess}
-	defaultSAName := "pipelines"
+	defaultSAName := "default"
```
bobcatfish, line 362: looking at the PR that added this change (https://github.com/tektoncd/pipeline/pull/1227) and trying to understand the intent, i think the goal was to explicitly set the serviceaccount to something other than default, is that your understanding as well? (or maybe it just needs to be something other than "tekton" https://github.com/tektoncd/pipeline/blob/master/pkg/apis/config/testdata/config-defaults.yaml#L22 ?)

it looks like the default value is "default" so i think by changing this to "default" we might be removing the thing this test is meant to cover

bobcatfish (commented):
> This error is intermittent and time plays role here since the other unit tests in taskrun_test.go are creating default service account (and never deleted). If this unit test TestReconcile_ExplicitDefaultSA is the first one to start, we are missing default service account and fails with this kind of error.

Great catch @pritidesai !!!

>  serviceaccounts "default" not found

I want to dig into this a little bit:

* I don't quite understand why "default" wouldn't exist - afaik there is always a "default" serviceaccount in every namespace (i think?)
* The error says "default" was not found, but the change in the PR changes the test to use "default" instead of "pipelines" 

I think this makes sense if the problem was that that "pipelines" service account didnt existing, but the error above looks like it's the other way around - could you explain a bit more?

I also don't totally understand how this could have ever worked since I don't see anything creating the "pipelines" service account either 🤔

pritidesai:
> I want to dig into this a little bit:
> 
> * I don't quite understand why "default" wouldn't exist - afaik there is always a "default" serviceaccount in every namespace (i think?)
> * The error says "default" was not found, but the change in the PR changes the test to use "default" instead of "pipelines"
> 
> I think this makes sense if the problem was that that "pipelines" service account didnt existing, but the error above looks like it's the other way around - could you explain a bit more?

`default` service account would exist in a running instance 😜  Here, the validation or check is being done against the test asset (specifically, fake clients) and that's reason this unit test has to [create/register](https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/taskrun/taskrun_test.go#L480) the service account in that client.

This unit test creates `pipelines` service account and sets `pipelines` as default service account through config map but taskrun controller fails to read that change done in config map. Without any config map changes, for taskrun controller, default service account is `default` and thats why test is complaining that `default` does not exist in the fake client (my brain is fried at this point 😆 )

> I also don't totally understand how this could have ever worked since I don't see anything creating the "pipelines" service account either 🤔

The unit test is creating `pipelines` service account. The test is checking if spec has set any service account which is not true in this case and results in `saName` to be empty. When `saName` is empty, its set to `defaultSAName` which is `pipelines` in this particular case.

pkg/reconciler/taskrun/taskrun_test.go
pritidesai: yup its my understanding as well

This particular unit test is testing two things:

(1) taskrun without having specified service account in the spec unlike the [next test case](https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/taskrun/taskrun_test.go#L359) in the same function (2) changing default service account through config map

Trying to think if it's possible to separate these two.

And I am not sure if we could be removing the thing this test is meant to cover because default value is `tekton` according to `testdata` and irrespective of what it is set to (other than `tekton`), the same value is set in config-map and sent to controller. 🤔 

I am still looking into this and would appreciate input from config map experts 🙏

pritidesai: I am working on getting this hard coded default out of `creds_init.go` 👩‍💻

pritidesai: I have created PR https://github.com/tektoncd/pipeline/pull/3168 to remove this hard coded default and also add it to the pipelinerun/taskrun `spec`.

bobcatfish:
@pritidesai i am still confused but I think it's just because I need to take some time to understand this and it's the evening before a long weekend and I just can't do it! Apologies for my confusion and weekend brain! I'll try to summarize it again but you've probably already addressed it:

* The error in the PR description says the default service account doesnt exist, but the change in the PR is changing a test to USE that very service account that apparently didnt exist in the test failure

I think it makes sense to me that the default service account would exist on a running instance, but not in this mock environment, and i can see how the order of test execution could make that sometimes not get created, i just cant understand how the change in the PR goes along with that 😦  (and you've probably explained it already!!)

Happy to defer to you in the meantime and I appreciate your thorough analysis! I might be back with more questions later but in the meantime:

/approve

pritidesai:
No worries @bobcatfish yup I can understand it being so tricky and appreciate you following the entire analysis. This screen shot from Scott's error capture might be useful understanding and feel free to hold this unless you are sure.

Even when the `config-defaults` ConfigMap has default set to `pipelines` and no service account specified in the taskrun spec, taskrun controller is looking for `default`:  

Screen Shot 2020-09-03 at 10 00 23 PM

Since I am not able to reproduce this, I am feeding off of this error capture and changing it to `default`. I have opened another PR # 3168 related to the findings from this effort. That PR is implementing changes to
set service account in taskrun spec if neither specified in the spec nor in `config-defaults` ConfigMap compared to what we have today setting service account in taskrun spec if not specified in the spec but exists in `config-defaults` ConfigMap.

pkg/reconciler/taskrun/taskrun_test.go
vdemeester: I don't think the problem is the *hardcoded* `default` serviceaccount (although, it would be nice if it wasn't harcoded :stuck_out_tongue: )

vdemeester (commented):
So I think this is not the right fix :upside_down_face: 

As far as I understand, we do create the `serviceAccount` during the tests. One thing that we do not do (and that we do when we setup tasks, …) is to make sure the informers are up-to-date too. 

I take a quick look at it, and this is the "fix" I came up with (@bobcatfish @pritidesai)

pritidesai:
Thanks @vdemeester I am closing this for now assuming this is fixed with PR #3189 🙏

Issue #3097, opened 2020-08-12
Support use-case of building sources with re-usable cache


quintesse:
### Feature request

Be able to build sources using a writable volume shared between tasks in a pipeline, while at the same time having a different writable volume for caching downloaded artifacts (Maven artifacts, NPMs, etc).

### Use case

Right now with the Affinity Assistant it's not possible anymore to have multiple writable volumes and the way the docs put it it seems that it is considered to be a somewhat exotic requirement. But wanting to keep downloaded artifacts around so future builds will complete more quickly seems a very common thing to want to do.

So ideally we'd want two workspaces, one to build the sources and one to hold the downloaded artifacts.

Now, from different sources we've seen it mentioned that you can use different sub-paths within the same volume, and while that's definitely a solution it seems that this forces the Task to do clean-up of the build environment itself. So instead of getting a clean setup each time we basically get a possibly dirty environment which might affect the build process. That seems less than ideal.

Being able to strictly separate the (throw-away) build environment and the download cache makes everything much more controllable and less complicated.

But honestly, for the build environment we really aren't even interested in using PVCs, but we're forced to do so because `emptyDir` can't be used between Tasks in Pipelines. And yes we can use VolumeClaimTemplates but they still result in the volume staying around until the PipelineRun gets removed, which might be much longer than necessary.

quintesse:
NB: related pull-request with information about issues related to the Affinity Assistant and multiple writable volumes: https://github.com/tektoncd/pipeline/pull/2885

jlpettersson:
> > Being able to strictly separate the (throw-away) build environment and the download cache makes everything much more controllable and less complicated.
> 
> +💯

+💯

> This is something multiple customers did bring back to us on the affinity-assistant. I think we should have done the feature flag the opposite (aka enabling it) but that's too late for changing that now.

My opinion is that Tekton should not leave users with problem A Or problem B, but instead solve both problems, or at least solve and guide the user on how to solve it the Tekton way.


> That said, I think we should be able to give the opportunity for the user to disable the affinity assistant per `*Run` (or `Pipeline` I think ?). The user should be able to decide to disable the affinity assistant, by using an annotation for example. That way, it's enabled by default and can be disable by the user _ad-hoc_ (on specific runs) or on the whole installation.

I don't see how that makes sense? Only some runs can be deadlocked because the pods ended up in different AZs?

vdemeester:
[...]

Issue #3052, opened 2020-08-03
Investigate to create a Custom Scheduler to schedule TaskRun pods


jlpettersson:
### Feature request

Investigate if it would be doable to create a [Custom Scheduler](https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/) for scheduling TaskRun pods, e.g. co-scheduling pods that share workspace PVC volume.

### Use case

When [the affinity assistant was introduced](https://github.com/tektoncd/pipeline/pull/2630) it solved problems with concurrent access to workspace volumes and deadlock if pods were scheduled to different AZ.

Using pod-affinity to achieve Node Affinity for TaskRun pods was the least complex solution that was evaluated.

The current solution works for common cases, but it is not a perfect solution. E.g. there may be problems when TaskRun require different amount of resources and the Nodes need to be autoscaled up as in https://github.com/tektoncd/pipeline/issues/3049

Adding a custom scheduler will probably introduce more complexity and code. But it probably solve the problem in a more generic way than using the Affinity Assistant.

dlorenc:
It seems like writing a custom scheduler is pretty straightforward: https://github.com/kelseyhightower/scheduler

but dealing with edge cases would probably be a lot of effort. I wonder if it would be possible to write a best-effort scheduler that runs first, but bails out to the real one in complex situations.

denkensk:
I suggest that we can evaluate that if Affinity Assistant can be implement by the Scheduling Framework.
FYI：https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/20180409-scheduling-framework.md

And we maybe can enhance the coscheduling to support the requirement.  https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/pkg/coscheduling
@jlpettersson @dlorenc

vincent-pli:
@denkensk 
That's very interesting, I would like to take a try.

ImJasonH:
I think a good concrete next step here would be for someone to experiment/prototype with the scheduler framework and report back to the community with any findings/demos, and help us concretely understand what the code would look like to, for instance, replace AA with custom scheduling.

Based on those findings we could start a design doc to more concretely outline requirements and next steps, or maybe determine that delving into scheduling really isn't worth the effort and shouldn't be pursued at this time.

@vincent-pli is that something you'd be interested in exploring and driving?

vincent-pli:
@ImJasonH 
[...]

PR #3161, opened 2020-09-02
Disable Affinity Assistant by default

ImJasonH:
There's been some confusion among users about the benefits/limitations of AA, in particular that it limits TaskRuns from mounting two persistent workspaces.

See:
- https://github.com/tektoncd/pipeline/issues/3097
- https://github.com/tektoncd/pipeline/issues/2829
- https://github.com/tektoncd/pipeline/pull/2885

We've heard that some Tekton installation providers disable this feature by default to avoid this confusion, and this PR is intended to start a conversation about disabling it in the official Tekton release config until the limitations/confusion can be reduced or eliminated.

/cc @jlpettersson 
/cc @bobcatfish 
/cc @vdemeester 

**NB:** If this is included in the next Tekton release, users who upgrade to that release with `kubectl apply` will see AA disabled, perhaps unexpectedly.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Disable Affinity Assistant in the official Tekton release config.
```

### DIFF
--- a/config/config-feature-flags.yaml
+++ b/config/config-feature-flags.yaml
@@ -1,4 +1,4 @@
-# Copyright 2019 The Tekton Authors
+#f Copyright 2019 The Tekton Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -24,12 +24,12 @@ data:
   # Setting this flag to "true" will prevent Tekton to create an
   # Affinity Assistant for every TaskRun sharing a PVC workspace
   #
-  # The default behaviour is for Tekton to create Affinity Assistants
+  # The default behaviour is for Tekton to not create Affinity Assistants
   #
   # See more in the workspace documentation about Affinity Assistant
   # https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md#affinity-assistant-and-specifying-workspace-order-in-a-pipeline
   # or https://github.com/tektoncd/pipeline/pull/2630 for more info.
-  disable-affinity-assistant: "false"
+  disable-affinity-assistant: "true"
   # Setting this flag to "true" will prevent Tekton overriding your
   # Task container's $HOME environment variable.
   #


vdemeester (commented):
@ImJasonH thanks for opening this for discussion. Indeed, we did disable this by default on `openshift-pipelines` because it gets more in the way to our current users that it helps them at least (given their current use of pipeline).

One thing we should try to keep in mind in the long run is, when we push a new feature that can have this impact, we may want to start by having an `opt-in` behavior at first (and later switch to `opt-out`). Doing the opposite (as this PR does), feels a bit backward :stuck_out_tongue: 

/approve
/hold

ImJasonH:
> One thing we should try to keep in mind in the long run is, when we push a new feature that can have this impact, we may want to start by having an opt-in behavior at first (and later switch to opt-out). Doing the opposite (as this PR does), feels a bit backward 😛

Yeah I completely agree, it's not ideal. I was thinking it might be worth adding to the TEP template some prompt to discuss how we'd release the newly proposed feature, and use that to encourage people to think about opt-in vs opt-out (and just having feature flags for everything in general!)

jlpettersson:
My opinion is probably known :)

Without the AA:
* Some pipelines in a regional cluster will deadlock - example https://github.com/tektoncd/pipeline/pull/2546
* Tasks are enforced to run sequentially (may timeout) - https://github.com/tektoncd/pipeline/issues/2586 and [removed "warning" in PR](https://github.com/tektoncd/pipeline/pull/2630/commits/77db01421f445db8f39a57115b4f427d3c79bcd7#diff-9f267bf6d3ac2686750c18b691c47066L224-L226)
* Pipelines executes slowly if the PV needs to be mounted on different Nodes

With the AA:
* A constraint is introduced - or more upfront (since the above problems exists): At most one PVC per Task. Other than that, the above problems are solved.
* Pipeline executes 20-30 seconds faster per task that use the PV, except the first.

bobcatfish:
> At most one PVC per Task.

This seems like a pretty big constraint to me :S We designed Tasks to be able to handle more than one workspace intentionally.

vdemeester:
>     * A constraint is introduced - or more upfront (since the above problems exists): At most one PVC per Task. Other than that, the above problems are solved.

This is the core of the problem for me. I do see value on this but I think in an ideal world
- this would have been `opt-in` initially to gather feedback on this without disallowing some behavior.
- having a way to disable the behavior "on-demand" for a run (as a "do at your own risk" thing) would be more than nice. If a user need to go around the constraint for 2% of it's use case, but feel safer having the assistant in place for the 98% rest, he should be able to do it

> Some pipelines in a regional cluster will deadlock

Not all cluster are regional. Depending on the users, for some, they don't have this problem, at all.

pierretasci:
In my experience, not all k8s clusters I have encountered work well with the Affinity Assistant because of non-standard configurations. Not all installations allow access to the K8s nodes or they have non-standard labels on the nodes. Since this feature configures the core experience rather than extends it, I personally tend to prefer opt-in semantics. Said another way, the default should work with the majority of cases possible but that is just my 2 cents.

ImJasonH:
Yeah I think the central issue is that the tradeoff is not clearly laid out in documentation (that comment is a _great_ distillation though!), and it introduces a serious limitation for many users. And it was enabled by default in the last release. Ideally we'd have made it opt-in then, and now that it's opt-out, rolling it back to opt-in is not necessarily a safe change. 😞

jlpettersson:
>Since this feature configures the core experience rather than extends it

It was introduced to avoid the problems with deadlocks and parallelism-problems rather than add a feature.

>Said another way, the default should work with the majority of cases possible but that is just my 2 cents.

+1. I had the same reasoning when this was added - to solve the assumed most common cases: https://github.com/tektoncd/pipeline/pull/2630#discussion_r428910545

>Using PVC with access mode ReadWriteOnce (the most common access mode) can only be mounted on one Node at a time. In addition, they most commonly lives in a single datacenter/Availability Zone. So if a pipeline starts e.g. with two parallel tasks - they may be scheduled to two different Availability Zones - this goes fine - until a Task try to mount both volumes - located in different AZ, and the pipeline is deadlocked - but this only happens when scheduled to different zones. 

>And it was enabled by default in the last release.

The Affinity Assistant was introduced in [v0.13.0](https://github.com/tektoncd/pipeline/releases/tag/v0.13.0)

>Not all cluster are regional.

That is true. But it was also motivated to allow for parallel task execution. But not all clusters have more than one Node either. But it happened that "Node Affinity" solved both origin problems.

>We designed Tasks to be able to handle more than one workspace intentionally.

The constraint is for PVCs, not strictly for workspaces since a workspace can be backed by other things than a PVC, and a Task can use multiple workspaces backed by the same PVC (using subPath).

[A Custom Scheduler](https://github.com/tektoncd/pipeline/issues/3052) is a better solution than the Affinity Assistant for this problem - it should schedule the Pods based on the same "constraints" as the Affinity Assistant, but it does not mount the PVC itself, as the AA had to do - so it has a bit friendlier behavior.

sbwsg:
There's been some discussion around this in WGs now and I think the consensus was:

1. we shouldn't disable this by default now because we'd be messing with folks' existing installations.
2. we will exercise more caution about enabling new features by default in future.

@tektoncd/core-maintainers or anyone else on this thread: does anyone have anything further to add?  Is there a purpose in keeping this PR open any longer?

ImJasonH:
> There's been some discussion around this in WGs now and I think the consensus was:
> 
> 1. we shouldn't disable this by default now because we'd be messing with folks' existing installations.
> 2. we will exercise more caution about enabling new features by default in future.
> 
> @tektoncd/core-maintainers or anyone else on this thread: does anyone have anything further to add? Is there a purpose in keeping this PR open any longer?

This is an excellent summary. I don't think there's anything else for this PR.

PR #3168, opened 2020-09-03
default service account

pritidesai:
# Changes

`Pipelinerun` can specify service account name in the specification. If its not specified, controller reads `default-service-account` from `config-defaults` ConfigMap. And if `config-defaults` is also missing `default-service-account`, service account is set to `default`. Before this PR, this `default` value was set in [creds-init.go](https://github.com/tektoncd/pipeline/blob/master/pkg/pod/creds_init.go#L47) just before creating the Pod. So the pod is created with `default` service account but pipelinerun/taskrun spec is not made aware of this change. Instead of delaying this decision/set until a pod is created, this PR is setting the service account as part of pipelinerun/taskrun spec. This brings consistency from the controller perspective while choosing the default service account when not explicitly specified in ConfigMap. 

To summarize, controller reads service account from `config-defaults` ConfigMap and sets it in the pipelinerun/taskrun spec if spec is missing service account. And now (with this PR) if service account is also missing from `config-defaults`, controller sets the service account to `default` in the spec.  

This was discovered while working on issue #2815 and PR #3160. 

/kind bug

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Service account when missing from pipelinerun/taskrun spec and ConfigMap, controller sets it to default in the spec.
```

### DIFF
--- a/pkg/pod/creds_init.go
+++ b/pkg/pod/creds_init.go
@@ -19,6 +19,7 @@ package pod
 import (
 	"fmt"
 
+	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	"github.com/tektoncd/pipeline/pkg/credentials"
 	"github.com/tektoncd/pipeline/pkg/credentials/dockercreds"
@@ -43,8 +44,10 @@ const credsInitHomeMountPrefix = "tekton-creds-init-home"
 // caller. If no matching annotated secrets are found, nil lists with a
 // nil error are returned.
 func credsInit(serviceAccountName, namespace string, kubeclient kubernetes.Interface) ([]string, []corev1.Volume, []corev1.VolumeMount, error) {
+	// service account if not specified in pipeline/task spec, read it from the ConfigMap
+	// and defaults to `default` if its missing from the ConfigMap as well
 	if serviceAccountName == "" {
-		serviceAccountName = "default"
+		serviceAccountName = config.DefaultServiceAccountValue
 	}
 
 	sa, err := kubeclient.CoreV1().ServiceAccounts(namespace).Get(serviceAccountName, metav1.GetOptions{})
--- a/pkg/apis/config/default.go
+++ b/pkg/apis/config/default.go
@@ -32,6 +32,7 @@ const (
 	NoTimeoutDuration              = 0 * time.Minute
 	defaultTimeoutMinutesKey       = "default-timeout-minutes"
 	defaultServiceAccountKey       = "default-service-account"
+	DefaultServiceAccountValue     = "default"
 	defaultManagedByLabelValueKey  = "default-managed-by-label-value"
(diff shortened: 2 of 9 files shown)


vdemeester (commented):
/lgtm

pkg/pod/creds_init.go
```diff
@@ -43,8 +44,10 @@ const credsInitHomeMountPrefix = "tekton-creds-init-home"
 // caller. If no matching annotated secrets are found, nil lists with a
 // nil error are returned.
 func credsInit(serviceAccountName, namespace string, kubeclient kubernetes.Interface) ([]string, []corev1.Volume, []corev1.VolumeMount, error) {
+	// service account if not specified in pipeline/task spec, read it from the ConfigMap
+	// and defaults to `default` if its missing from the ConfigMap as well
 	if serviceAccountName == "" {
-		serviceAccountName = "default"
+		serviceAccountName = config.DefaultServiceAccountValue
```
afrittoli, line 50: It's interesting that we force a service account to be specified, so we do not allow taskruns and pipelineruns to rely on the default for the namespace...  
With the changes from this PR in place, the only case where serviceAccountName was empty here would be if it was sent to "" in the configmap or in the pipelinerun/taskrun - so I wonder if should instead fail or at least log a warning?

afrittoli (commented):
Thanks for this!
I agree with this change, but I wonder if in creds init we should silently ovverride an empty string now.
Before it might have happened because of config not set, but now it may only happen in service account is deliberately set to an empty string, which I think we should at least raise a warning about?

pkg/pod/creds_init.go
pritidesai: nope, it's perfectly fine to have service account name set to "". This change is avoiding hard coded `default` string here and instead reading from constant.

afrittoli: What I meant is that we do not allow serviceAccount to be an empty string - we did not before and we still not do. My concern is that if the service account is set to empty string in the config map, we do not honour that, we override it instead with `config.DefaultServiceAccountValue` (which is the string "default"). This is not a change introduced in this PR, it's something that existed before, but I think it would be worth (perhaps in a different PR) to add a warning or something so the user is aware that the service account name has been overwritten compared to config.

pritidesai: I see what you mean, warning should work. Will create a separate PR. Warn just the service account or include other defaults as well happening in [default.go](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/config/default.go)?

vdemeester:
(temporarly removing them from the merge queue to unlock some)

vdemeester:
/lgtm cancel
@pritidesai can you rebase and push to unblock this ? I can't force-unblock it as it jobs for this do not appear anywhere anymore (see [here](https://prow.tekton.dev/?repo=tektoncd%2Fpipeline&author=pritidesai))

vdemeester:
:tada:

Issue #2127, opened 2020-03-01
Simple if/conditional support for skipping


skaegi:
A common pattern used in some of the pipelines we're migrating to Tekton, is to check which branch is being built, and if the branch name matches a particular string like "dev", perform some extra logic.

I'd like to propose adding support to Pipeline with either a simple `if` or augmenting the `conditions` to allow in-lined boolean logic to decide whether to skip a Task. What's important for this use-case is that this check is **not** an "assertion" and just a skip guard so that the Task succeeds with a `reason` of "skipped". Also see #1023

For the syntax of the expression, I'd suggest we re-use JSONPath filter expressions, where the boolean logic is based on whether or not the JSONPath result list is empty. For example, given a parameter named "branch" a JSONPath filter expression for "dev" would look something like this...
```
$.params.branch == 'dev'
```

In the context of a Pipeline this might look something like...
```
apiVersion: tekton.dev/v1alpha1
kind: Pipeline
metadata:
  name: two-tasks-one-if
spec:
  params:
    - name: branch
  tasks:
  - name: skippable-task
    if: $.params.branch == 'dev'
    taskRef:
      name: special-dev-only-task
  - name: run-regardless-of-first-task-skipped
    taskRef:
      name: always-run-this
    runAfter:
      - skippable-task
```

skaegi:
Also see https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.9.2.1 -- for some additional boolean logic operators that we could support above and beyond a single expression string

skaegi:
I uses a JSONPath Expression in the above example but an alternative worth considering is to use CEL.

pritidesai:
Yup, love both ❤️ more or less both serves similar purpose, can be concise and descriptive at the same time, sample [here](https://github.com/tektoncd/triggers/blob/28e992239402e0030f60e8ad5c25c2777af278fb/examples/eventlisteners/cel-eventlistener-interceptor.yaml#L21).

```
- cel:
            filter: "header.canonical('X-GitHub-Event') == 'push'"
```

This example can be addressed using `conditions` and `runOn` 😄  but `condition` would be heavy weight and complex just to validate params 😞 Also `runOn` is designed to have `always` equivalent to ["success", "failure", "skip"] (not implemented in the existing PR yet).

```
apiVersion: tekton.dev/v1alpha1
kind: Condition
metadata:
  name: compare-branch
spec:
  params:
    - name: "branch"
  check:
    image: alpine
[...]

Issue #3109, opened 2020-08-17
Allow Tasks in Pipelines to express a dependency on workspaces 'from' previous Tasks


bobcatfish:
### Feature request

In features like conditions we distinguish between different kinds of dependencies:

1) ordering (runAfter)
2) resource dependencies (when a Task relies on a result or a PipelineResource from another Task).

Depending on a workspace that has been acted on by a previous Task is a case of (2) but since workspaces do not have a "from" or similar syntax we have no way of expressing (2) and it will always look like (1).

The syntax for this could look like this:

```yaml
  tasks:
    - name: task1
      taskRef:
        name: write-to-workspace
      workspaces:
        - name: output
          workspace: pipeline-ws1
    - name: task2
      taskRef:
        name: read-from-workspace
      workspaces:
        - name: src
          workspace: $(tasks.task1.workspaces.output)
```

Or we could use a [from syntax like we do with PipelineResources](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-the-from-parameter):

```yaml
  tasks:
    - name: task1
      taskRef:
        name: write-to-workspace
      workspaces:
        - name: output
          workspace: pipeline-ws1
    - name: task2
      taskRef:
        name: read-from-workspace
      workspaces:
        - name: src
           from: $(tasks.task1.workspaces.output)
```

Probably we'd want to use the first syntax to be consistent with [the syntax for results](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#passing-one-tasks-results-into-the-parameters-of-another). I prefer `from` because it's slightly more explicit but that's probably a discussion for another issue!

### Use case

Any Pipeline where you want one Task to act on a workspace and another to do something with it. Some examples:

1. TaskA clones from git to a workspace, TaskB builds the cloned code
2. TaskA builds a binary - but conditionally, only if the files the binary builds from have changed. TaskB runs a test with that binary, but should not run if TaskA was skipped

sbwsg:
I agree `from` might make more sense. In the first syntax snippet the user could reasonably assume that only the name of the workspace will be interpolated. But they'd be wrong - the name of the workspace will be injected _and_ there's an implicit runAfter added to the pipelinetask. `from` gives a clearer signal, at least to me, that there's something else going on over just the string interpolation.

[...]

PR #3176, opened 2020-09-08
Support skipping single task only

jerop:
# Changes

When `WhenExpressions` evaluate to False, the guarded `Task` and
its branch (dependent `Tasks`) are skipped

A `Task` is dependent on and in the branch of another `Task` as specified
by ordering using `runAfter` or by resources using `Results`, `Workspaces`
 and `Resources`

In some use cases, when `WhenExpressions` evaluate to `False`,
users need to skip the guarded `Task` only and allow ordering-dependent
`Tasks` to execute

When  `WhenExpressions` evaluate to `False`, it is possible to allow for
execution of ordering-dependent `Tasks` as specified by `runAfter` using
the `whenSkipped` field by setting it to `runBranch`.

However, setting `whenSkipped` in `Tasks` without `WhenExpressions`
or `Tasks` with resource dependencies is invalid, and will cause
`Pipeline` validation errors. This ensures consistency in executing or
skipping subsequent tasks.

Further details in [TEP](https://github.com/tektoncd/community/blob/main/teps/0007-conditions-beta.md#skipping-1)

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```release-note
When  `WhenExpressions` evaluate to `False`, it is possible to allow
execution of ordering-dependent `Tasks` as specified by `runAfter`
using the `whenSkipped` field by setting it to `runBranch`
```

### DIFF
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -437,6 +437,37 @@ tasks:
       name: lint-source
 ```
 
+When  `WhenExpressions` evaluate to `False`, the `Task` and its branch (of dependent `Tasks`) will be skipped by default while the rest of the `Pipeline` will execute. 
+
+A `Task` branch is made up of dependent `Tasks`, where there are two types of dependencies:
+- _Resource dependency_: based on resources needed from parent `Task`, which includes [`Results`](#using-results) and [`Resources`](#specifying-resources).
+- _Ordering dependency_: based on [`runAfter`](#using-the-runafter-parameter) which provides sequencing of `Tasks` when there may not be resource dependencies.
+
+In `Tasks` guarded using `WhenExpressions` and that have ordering dependencies only, use `whenSkipped` to specify what happens when the guarded `Task` is skipped.
+The `whenSkipped` take either `SkipBranch` or `RunBranch` policies. To allow execution of ordering-dependent `Tasks`, set `whenSkipped` to `RunBranch` in the guarded parent task.
+If neither `SkipBranch` nor `RunBranch` is specified, the default one is `SkipBranch`. 
+
+However, setting `whenSkipped` in `Tasks` without `WhenExpressions` or `Tasks` with resource dependencies is invalid and will cause `Pipeline` validation errors.
+
+In this example, `task-should-be-skipped` will be skipped and `task-should-be-executed` will be executed: 
+
+```yaml
+tasks:
+  - name: task-should-be-skipped
+    when:
+      - input: "foo"
+        operator: in
+        values: ["bar"]
+    whenSkipped: RunBranch
+    taskRef:
+        name: exit-1
+  - name: task-should-be-executed
+    runAfter:
+    - task-should-be-skipped
+    taskRef:
+        name: exit-0  
+```
+
 For an end-to-end example, see [PipelineRun with WhenExpressions](../examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml).
 
 When `WhenExpressions` are specified in a `Task`, [`Conditions`](#guard-task-execution-using-conditions) should not be specified in the same `Task`. The `Pipeline` will be rejected as invalid if both `WhenExpressions` and `Conditions` are included.
--- a/examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml
+++ b/examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml
@@ -82,6 +82,7 @@ spec:
           - input: "$(params.path)"
             operator: notin
             values: ["README.md"]
+        whenSkipped: RunBranch
         taskSpec:
           steps:
             - name: echo
@@ -99,6 +100,18 @@ spec:
             - name: echo
               image: ubuntu
               script: exit 1
+      - name: task-should-be-executed-after-skipped-parent-task # whenSkipped set to runBranch in parent task
+        runAfter:
+          - task-should-be-skipped-2
+        when:
+          - input: "$(params.path)"
+            operator: in
+            values: ["README.md"]
+        taskSpec:
+          steps:
+            - name: echo
+              image: ubuntu
+              script: 'echo created README.md'
     finally:
       - name: finally-task-should-be-skipped-1 # when expression using execution status, evaluates to false
         when:
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation.go
@@ -71,6 +71,7 @@ func (ps *PipelineSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	errs = errs.Also(validateTasksAndFinallySection(ps))
 	errs = errs.Also(validateFinalTasks(ps.Tasks, ps.Finally))
 	errs = errs.Also(validateWhenExpressions(ps.Tasks, ps.Finally))
+	errs = errs.Also(validateWhenSkipped(ps.Tasks))
 	return errs
 }
 
@@ -617,4 +618,52 @@ func validateGraph(tasks []PipelineTask) *apis.FieldError {
 		return apis.ErrInvalidValue(err.Error(), "tasks")
 	}
 	return nil
+}
+
+func validateWhenSkipped(tasks []PipelineTask) (errs *apis.FieldError) {
+	d, err := dag.Build(PipelineTaskList(tasks), PipelineTaskList(tasks).Deps())
+	if err != nil {
+		return apis.ErrInvalidValue(err.Error(), "tasks")
+	}
+	for i, t := range tasks {
+		if t.WhenSkipped != "" {
+			if t.WhenExpressions == nil || hasResourceDependencies(t, toMap(tasks), d) {
+				errs = errs.Also(apis.ErrDisallowedFields("whenSkipped").ViaFieldIndex("tasks", i))
+			}
+			if t.WhenSkipped != RunBranch && t.WhenSkipped != SkipBranch {
+				errs = errs.Also(apis.ErrInvalidValue("RunBranch and SkipBranch only allowed in", "whenSkipped").ViaFieldIndex("tasks", i))
+			}
+		}
+	}
+	return errs
+}
+
+func toMap(tasks []PipelineTask) map[string]PipelineTask {
+	taskMap := make(map[string]PipelineTask)
+	for _, task := range tasks {
+		taskMap[task.Name] = task
+	}
+	return taskMap
+}
+
+func hasResourceDependencies(parentTask PipelineTask, taskMap map[string]PipelineTask, d *dag.Graph) bool {
+	if node, ok := d.Nodes[parentTask.Name]; ok {
+		for _, childNode := range node.Next {
+			childTask := taskMap[childNode.Task.HashKey()]
+			if isResourceDependent(parentTask, childTask) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func isResourceDependent(parentTask PipelineTask, childTask PipelineTask) bool {
+	resourceDeps := childTask.resourceDeps()
+	for _, resourceParent := range resourceDeps {
+		if resourceParent == parentTask.Name {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/apis/pipeline/v1beta1/when_types.go
+++ b/pkg/apis/pipeline/v1beta1/when_types.go
@@ -159,3 +159,13 @@ func (wes WhenExpressions) ReplaceWhenExpressionsVariables(replacements map[stri
 	}
 	return replaced
 }
+
+// WhenSkippedPolicy describes what should happen when a Task is skipped because its WhenExpressions evaluated to false.
+// WhenSkippedPolicy can be specified only in Tasks that are guarded with WhenExpressions and do not have resource dependencies.
+// If none of the following policies is specified, the default one is SkipBranch.
+type WhenSkippedPolicy string
+
+const (
+	SkipBranch WhenSkippedPolicy = "SkipBranch"
+	RunBranch  WhenSkippedPolicy = "RunBranch"
+)
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
@@ -124,6 +124,24 @@ var pts = []v1beta1.PipelineTask{{
 	Name:    "mytask15",
 	TaskRef: &v1beta1.TaskRef{Name: "taskWithReferenceToTaskResult"},
 	Params:  []v1beta1.Param{{Name: "param1", Value: *v1beta1.NewArrayOrString("$(tasks.mytask1.results.result1)")}},
+}, {
+	Name:    "mytask16",
+	TaskRef: &v1beta1.TaskRef{Name: "taskWithWhenExpressions"},
+	WhenExpressions: []v1beta1.WhenExpression{{
+		Input:    "foo",
+		Operator: selection.NotIn,
+		Values:   []string{"foo", "bar"},
+	}},
+	WhenSkipped: v1beta1.RunBranch,
+}, {
+	Name:    "mytask17",
+	TaskRef: &v1beta1.TaskRef{Name: "taskWithWhenExpressions"},
+	WhenExpressions: []v1beta1.WhenExpression{{
+		Input:    "foo",
+		Operator: selection.NotIn,
+		Values:   []string{"foo", "bar"},
+	}},
+	WhenSkipped: v1beta1.SkipBranch,
 }}
 
 var p = &v1beta1.Pipeline{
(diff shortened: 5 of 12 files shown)


jerop:
`runAfter` is meant to be used to specify ordering when there are no resource dependencies - https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-the-runafter-parameter

however, it seems that `Workspaces` don't have a way to specify resource dependencies so they use `runAfter`, further detail in https://github.com/tektoncd/pipeline/issues/3109

need to look into what this means for skipping ordering-dependent tasks as specified by `runAfter`

/hold

docs/pipelines.md
```diff
@@ -352,6 +352,25 @@ tasks:
         name: echo-file-exists
 ```
 
+When  `WhenExpressions` evaluate to `False`, it is possible to allow for execution of ordering-dependent `Tasks` as specified by [`runAfter`](#using-the-runafter-parameter) using the `continueAfterSkip` field by setting it to `true`. In this example, `task-should-be-skipped` will be skipped and `task-should-be-executed` will be executed.
```
bobcatfish, line 355: could you mention what the default value is as well? i assume false?

bobcatfish:
@jerop i think for now it's fine to treat runAfter as the same for all circumstances - however if you wanted to tackle #3109 before this that'd be cool too. basically i dont think either needs to wait on the other

this would only be a problem for a fairly specific use case:

1. Someone has a workspace coming "from" another Task, e.g. task B needs workspace X from task A, and this is expressed using runAfter
2. That person uses a when expression on task A, and sets continueAfterSkip: true

imo (2) is actually an error - and i think if you did the same thing with a result, we'd fail it on validation, is that right? So not having #3109 before doing this would mean that there is an error case we aren't able to catch with validation. Definitely it would be better to be able to catch this, but I don't feel like that needs to block this, what do you think?

jerop:
> @jerop i think for now it's fine to treat runAfter as the same for all circumstances - however if you wanted to tackle #3109 before this that'd be cool too. basically i dont think either needs to wait on the other
> 
> this would only be a problem for a fairly specific use case:
> 
> 1. Someone has a workspace coming "from" another Task, e.g. task B needs workspace X from task A, and this is expressed using runAfter
> 2. That person uses a when expression on task A, and sets continueAfterSkip: true
> 
> imo (2) is actually an error - and i think if you did the same thing with a result, we'd fail it on validation, is that right? So not having #3109 before doing this would mean that there is an error case we aren't able to catch with validation. Definitely it would be better to be able to catch this, but I don't feel like that needs to block this, what do you think?

@bobcatfish thank you for clarifying the use case for `workspace` + `runAfter` described in that issue -- agreed that's something we can tackle separately (assigned myself)

i think of `continueAfterSkip` as a flag that allows ordering-dependent child tasks only to execute and it doesn't affect resource-dependent tasks (so they would continue to be skipped if parent tasks were skipped) -- https://github.com/tektoncd/pipeline/pull/3176/files#diff-4e301064fef70f3c0db810267c0959f2R192-R196

so if there's a task with `WhenExpressions` that evaluate to false and `continueAfterSkip` is set to true, and it has both ordering-dependent tasks and resource-dependent tasks, the ordering-dependent tasks will be executed while resource-dependent tasks will be skipped

pritidesai:
> so if there's a task with WhenExpressions that evaluate to false and continueAfterSkip is set to true, and it has both ordering-dependent tasks and resource-dependent tasks, the ordering-dependent tasks will be executed while resource-dependent tasks will be skipped -- shown in this test case: https://github.com/tektoncd/pipeline/pull/3176/files#diff-f7c8532cfe79319c0fff711a4979a6d6R2204-R2210

Will have to be little more careful here. It sounds like inconsistent behavior for resource dependent task vs ordering dependent task. Also a task could have both runAfter and task result and would be skipped (with these changes) due to resource dependency. 

The way finally task with resource dependency is being handled is that final task is at least attempted but would fail at the param validation since the task result does not exist. But when we add param level defaults for task results, that final task would be executed with the default value.

Unfortunately I dont have any ideas how to fix this inconsistency other than allowing to specify `default` along with task result:

```
spec:
 tasks:
  - name: release-env
     taskRef:
       Name: boskos-release
     params:
       - name: proj-name
         value: $(tasks.get-env.results.project)
         default: best-default-ever
```

The long term solution is to support [schema validation for params](https://github.com/tektoncd/pipeline/issues/1393#issuecomment-558833526).

jerop:
> Will have to be little more careful here. It sounds like inconsistent behavior for resource dependent task vs ordering dependent task. 

when we distinguish `ordering-dependent tasks` from `resource-dependent tasks`, and handle them differently then i think it might not necessarily be inconsistent 🤔 

> Also a task could have both runAfter and task result and would be skipped (with these changes) due to resource dependency.

in this case, the child task would `runAfter` a parent task that is different from the parent task from which it needs a `result` (because `runAfter` [is used when there are no resource dependencies](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#using-the-runafter-parameter)), and:
- if `WhenExpressions` in the parent task it `ranAfter` evaluated to `false` and was skipped, then:
  - if `continueAfterSkip` in the parent task is `false`, then that parent task being skipped will block the execution of the child task
  - if `continueAfterSkip` in the parent task is `true`, then that parent task being skipped will not block the execution of the child task
- if `WhenExpressions` in the parent task it needed a `result` from evaluate to `false` and was skipped, the child task will be skipped whether or not `continueAfterSkip` is set in the parent task (because this is a `resource dependency`)

> The way finally task with resource dependency is being handled is that final task is at least attempted but would fail at the param validation since the task result does not exist. But when we add param level defaults for task results, that final task would be executed with the default value.

> Unfortunately I dont have any ideas how to fix this inconsistency other than allowing to specify `default` along with task result:
> 
> ```
> spec:
>  tasks:
>   - name: release-env
>      taskRef:
>        Name: boskos-release
>      params:
>        - name: proj-name
>          value: $(tasks.get-env.results.project)
>          default: best-default-ever
> ```
> 

when expressions use params from the pipeline level and use task results directly (not through params in the task level) -- does this mean we'd have to support default task results in when expressions as well (in addition to params)? (an alternative could be default results in the result-producing task)

if we implement similarly to finally, that means the resource validation would fail in the `resource-dependent` child task, the child task would fail and the pipeline would fail

however, i think it's a valid option to continue with skipping `resource-dependent` tasks while possibly executing `ordering-dependent` tasks

jerop:
@pritidesai @bobcatfish please review again when you can :)

pritidesai:
Hey @jerop what is the expected behavior in case of the following example?

Here, Task A through F are resulting in this kind of sequence using `runAfter`. Even when task A is skipped, the entire branch is executed as expected.

```
A (guarded) (continueAfterSkip=true)--- B --- C --- D --- E --- F
```

Now if we tweak this example a little:

```
A (guarded) (continueAfterSkip=true)--- B --- C(guarded) --- D --- E --- F
```

I was expecting the whole branch B through E getting executed with skipping A along with C being skipped. But here, since C is skipped without `(continueAfterSkip=true)`, D through F are skipped as well. So the `continueAfterSkip=true` on A doesn't apply to the whole branch in presence of guarded task (task C). 


We dont have such execution logic possible with `conditions` today so whatever we set as a design principle with `continueAfterSkip` wouldn't break any backward compatibility.

jerop:
> Hey @jerop what is the expected behavior in case of the following example?
> 
> Here, Task A through F are resulting in this kind of sequence using `runAfter`. Even when task A is skipped, the entire branch is executed as expected.
> 
> ```
> A (guarded) (continueAfterSkip=true)--- B --- C --- D --- E --- F
> ```
> 
> Now if we tweak this example a little:
> 
> ```
> A (guarded) (continueAfterSkip=true)--- B --- C(guarded) --- D --- E --- F
> ```
> 
> I was expecting the whole branch B through E getting executed with skipping A along with C being skipped. But with this PR, since C is skipped without `(continueAfterSkip=true)`, D through F are skipped as well. So the `continueAfterSkip=true` on A doesn't apply to the whole branch in presence of guarded task (task C).
> 
> We dont have such execution logic possible with `conditions` today so whatever we set as a design principle with `continueAfterSkip` wouldn't break any backward compatibility.

Currently, we check if any of the parent tasks is skipped and skip the child task if any of the parent tasks is skipped. 

Setting continueAfterSkip to true in a parent task allows for execution (doesn't cause skipping) of its child tasks if that specific parent task is skipped. However, if the child task has another parent that was skipped (and doesn't have `continueAfterSkip` set to true) the child task would be skipped as well. 

So in this example you provided where tasks A and C are skipped because their when expressions evaluated to false: 

A (guarded) (continueAfterSkip=true)--- B --- C(guarded) --- D --- E --- F

- task B: its parent task A was skipped but had continueAfter skip set to true, so it's executed 
- task D, task E and task F: their parent task C was skipped, so they're skipped as well (so they're skipped because C was skipped, not because A was skipped) 

this change in checking the parent tasks is here: https://github.com/tektoncd/pipeline/pull/3176/files#diff-39d9baa708e379638f742c8b9fd36d9d15c64c3386de958b5a3f309ca8fa037aR178-R183

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -819,6 +828,52 @@ func TestIsSkipped(t *testing.T) {
 			},
 		}},
 		expected: false,
+	}, {
```
pritidesai, line 831: Can we add more tests here for examples such as:

1.

```
                                           B 
                                         / 
A (guarded) (CAS=true)
                                        \ 
                                          C  
```

2. 

```
                                           B 
                                         / 
A (guarded) (CAS=true)
                                        \ 
                                          C (guarded) (CAS=true) - D 
```

3. 

```
A (guarded) (CAS=true)
                                           \
                                                C
                                           /
B                                      
```

jerop: sounds good, will add them :)

pritidesai:
> task D, task E and task F: their parent task C was skipped, so they're skipped as well (so they're skipped because C was skipped, not because A was skipped)

right its because of C, so basically the nearest parent influences scheduling and overwrites what A is configured to.
A and C both are not direct parents to E and F but are skipped because of the nearest parent C. Let's please document this.

jerop:
> > task D, task E and task F: their parent task C was skipped, so they're skipped as well (so they're skipped because C was skipped, not because A was skipped)
> 
> right its because of C, so basically the nearest parent influences scheduling and overwrites what A is configured to.
> A and C both are not direct parents to E and F but are skipped because of the nearest parent C. Let's please document this.

it's not really the nearest parent influencing scheduling, moreso that `continueAfterSkip` makes a parent task not block the execution of its child task (but other parent tasks could block the execution of the same child task)

taking this example, where task C is runAfter tasks A and B (so both parents are equally near):

  ```
         A (guarded)(CAS = true) --> skipped
             \
              C --> skipped
             / 
         B (guarded) --> skipped
 ```
task A was did **not** block the execution of task C, but task B blocked the execution of task C -- so task C was skipped because it's parent B was skipped

can add clarifications and a more detailed example to the documentation

pritidesai:
thanks @jerop yup I understand, but in this example:

```
A (guarded) (continueAfterSkip=true)--- B --- C(guarded) --- D --- E --- F
```

A is guarded with `continueAfterSkip` set to true which allows B, C, D, E and F (rest of the dependencies) to continue executing even if A is skipped. So now when C is executed and skipped (with `continueAfterSkip`) changes from executing to not executing for D, E, and F.  So from root node (`A`) point of view, all the dependencies are getting equal opportunity to be scheduled but `C` is overriding that when its skipped. Its perfectly valid I guess but results in a little confusing precedence order.

pritidesai:
A task is guarded with `when` expression, if that guarded task is skipped, an option to let it's dependencies run is named `continueAfterSkip`, how about calling it something similar to `runDependencies` or `runDeps` ? 🤔

bobcatfish:
@jerop pointed out to me that it's probably a better practice to use a string value here instead of a bool b/c it's the most flexible in the future.

I'm not sure what to call it but maybe it helps to try to list what we're expressing:

1. Behavior for "order dependent" Tasks only, i.e. "runAfter" Tasks
2. NOT behavior for "resource dependent" Tasks; these are always skipped

For (1), we're trying to offer 2 options:
a. If the when expression is false and the Task is skipped, run the runAfter Tasks anyway
b. If the when expression is false and the Task is skipped, do NOT fun the runAfter Tasks

A couple of different ways of looking at it:

1. Instead of thinking about the dependencies, think about the branch: I think one reason why _to me_ defaulting to stopping after a skipped Tasks makes sense is that I'm thinking about traversing a branch of a tree
2. Maybe this is expressing scoping on the when expression itself

General thoughts:
* include "branch" somewhere in the name?
* refer to the "when" vs "skip"?

Couple ideas:

```yaml
whenSkipped: runNext/stopBranch
```

```yaml
skip: taskOnly/branch
```

pritidesai:
> it's probably a better practice to use a string value here instead of a bool b/c it's the most flexible in the future.

I am little afraid to over utilize the same keyword, better to keep them as mush focused or limited as possible.

> For (1), we're trying to offer 2 options:
a. If the when expression is false and the Task is skipped, run the runAfter Tasks anyway
b. If the when expression is false and the Task is skipped, do NOT run the runAfter Tasks

```
tasks:
  - name: TaskA
     when: <resulting in failure>
     continueAfterSkip: true
  - name: TaskB
     runAfter: [ TaskA ]
```

`TaskA` with a when expression does not know that `TaskB` is dependent on `TaskA` or `TaskB` is part of the same branch as `TaskA`. `runAfter` in `TaskB` defines the relationship between `TaskA` and `TaskB`.

> Maybe this is expressing scoping on the when expression itself

Yup agree, scoping along with precedence. 

```
A (guarded) (continueAfterSkip=true)--- B --- C(guarded) --- D --- E --- F
```

When expression in TaskA has a scope of the entire branch (B through F) but when expression in TaskC takes higher precedence for tasks D through F.

> refer to the "when" vs "skip"?

I like using `when` instead of `skip`.

jerop:
> (christie) @jerop pointed out to me that it's probably a better practice to use a string value here instead of a bool b/c it's the most flexible in the future.
> (priti) I am little afraid to over utilize the same keyword, better to keep them as much focused or limited as possible.

@pritidesai for more context on why I pointed out to @bobcatfish that we should use a string alias instead of a bool, I read the [k8s API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md) that mentions: “_Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. `TerminationMessagePolicy`)_” -- so while I see the value in keep it as limited as possible, I think there's value in allowing this kind of flexibility in options beyond the two allowed by a bool...

@pritidesai @bobcatfish thank you for the naming suggestions! I like the `whenSkipped` but think we should modify the string aliases to be `runBranch` and `skipBranch`, what do you think?

```yaml
whenSkipped: runBranch/skipBranch
```

bobcatfish:
Fixes #2127

jerop:
Fixes https://github.com/tektoncd/pipeline/issues/1023

jerop:
@pritidesai @bobcatfish ready for review, ptal 🙏

sbwsg (commented):
I left another question re: naming (sorry! I know this has already been discussed a lot).  But I think this PR is looking really solid and otherwise is ready to go from my pov.

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -158,3 +158,13 @@ func (wes WhenExpressions) ReplaceWhenExpressionsVariables(replacements map[stri
...
 }
+
+// WhenSkippedPolicy describes what should happen when a Task is skipped
+// because its WhenExpressions evaluated to false. WhenSkippedPolicy can
+// be specified only in Tasks that are guarded with WhenExpressions and do not
+// have resource dependencies.
+type WhenSkippedPolicy string
+
+const (
+	RunBranch WhenSkippedPolicy = "runBranch"
```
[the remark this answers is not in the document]
jerop: Thanks for the review @sbwsg! 

I'd been conflicted about whether to include `skipBranch` or not, given it's the default behavior in all other cases, so added `runBranch` only to keep it simple because that's the only new behavior we're adding -- though the example you've provided makes me lean towards adding the `skipBranch` as well

Will capitalize the aliases to `RunBranch` and `SkipBranch`

examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml
```diff
@@ -99,6 +100,18 @@ spec:
             - name: echo
               image: ubuntu
               script: exit 1
+      - name: task-should-be-executed-after-skipped-parent-task # whenSkipped set to runBranch in parent task
```
pritidesai, line 103: Unfortunately, we do not have a way to verify that this task was executed. This looks good for now, but we might want to create an `e2e` test in future to verify such scenarios.

pkg/apis/pipeline/v1beta1/when_types.go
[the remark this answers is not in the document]
pritidesai: yup I agree with @sbwsg we need a default state set to `SkipBranch` but I am leaning towards not adding default as well  🙃  

defaults makes more sense in case of `boolean` but with string types, if its not needed, we can avoid setting defaults.

`default` will make code even more complicated. We can may be change the `TEP` to not include `skipBranch`, thoughts?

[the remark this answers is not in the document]
pritidesai: when `when` expression evaluates to false, by default, the entire branch is skipped, if we want to enable it, use `runBranch` policy in `whenSkipped`

examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml
jerop: yes, that'd be a great next step to this: https://github.com/tektoncd/pipeline/issues/3076

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -471,3 +472,51 @@ func validateGraph(tasks []PipelineTask) *apis.FieldError {
...
 }
+
+func validateWhenSkipped(tasks []PipelineTask) (errs *apis.FieldError) {
+	d, err := dag.Build(PipelineTaskList(tasks))
+	if err != nil {
+		return apis.ErrInvalidValue(err.Error(), "tasks")
+	}
+	for i, t := range tasks {
+		if t.WhenSkipped != "" {
+			if t.WhenExpressions == nil || hasResourceDependencies(t, toMap(tasks), d) {
```
sbwsg, line 483: I wonder if it might be worth having a separate error message for the `hasResourceDependencies` check?

```go
Task %q defines whenSkipped but child Task %q has a resource dependency on it
```

The reason being that I think it might be a bit confusing for the error to only say that the `whenSkipped` field is disallowed in this case without additional info.

It would be even more amazing if we could break it down further using information computed similarly to `resourceDeps()` like, `Task %q has resource dependency on Task %q (param %q references result %q)`, but that's totally not necessary for this changeset I don't think!

sbwsg: Oh, also a small optimization nit: we are calling `toMap` each iteration of the for loop but I think we only need to do so once beforehand.

sbwsg (approved):

michaelsauter:
What's the status on this? Is this supposed to land in the next release?

FWIW, I ran into this PR because from reading the docs, I assumed that `when` expressions would not actually skip the following tasks ... I like how the documentation is worded in this PR, much clearer now.

jerop:
new proposal accepted: https://github.com/tektoncd/community/blob/main/teps/0059-skipping-strategies.md

PR #3177, opened 2020-09-08
Remove pkg/system.

dlorenc:
# Changes

This is almost identical to the code in knative/pkg/system which we already
import and use in a few cases.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
The controller now requires that the SYSTEM_NAMESPACE environment variable is set. This was set by default before, but is now required.
```


sbwsg (approved):

pkg/system/names.go
```diff
@@ -1,33 +0,0 @@
...
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
-
-package system
-
-import "os"
-
-const (
-	DefaultNamespace      = "tekton-pipelines"
```
vdemeester, line 22: the `knative/pkg` system package doesn't have this `DefaultNamespace`, so this *might* break users if the are running the controller without `SYSTEM_NAMESPACE`.

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -58,6 +57,8 @@ import (
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/logging"
 	pkgreconciler "knative.dev/pkg/reconciler"
+	"knative.dev/pkg/system"
+	_ "knative.dev/pkg/system/testing"
```
vdemeester, line 61: I think the linter usually asks for a comment on top of `_` import (so that we know why it's there)

vdemeester (commented):
One comment — it shouldn't affect user that much but… it might for some.
/cc @skaegi @chhsia0

pkg/reconciler/taskrun/taskrun_test.go
dlorenc: Ah yeah it doesn't appear to require it for test files. I'll leave a comment anyway.

pkg/system/names.go
bobcatfish: can we set SYSTEM_NAMESPACE to a default in the config yamls maybe?

bobcatfish (commented):
Nice cleanup!

pkg/system/names.go
dlorenc: It is set here:
```yaml
        env:
        - name: SYSTEM_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
```

dlorenc: https://github.com/tektoncd/pipeline/blob/master/config/controller.yaml#L80

So this would really only present an issue if people ran with modified YAMLs.

vdemeester: @dlorenc yeah that was my thought. It's a really *niche* problem as I think the number of user / integrator is really small. Most likely having a big warning in the release note should be enough.

dlorenc: Added a section in the release notes.

vdemeester:
```
pkg/reconciler/pipelinerun/affinity_assistant_test.go:171:86: undefined: system.GetNamespace
```

@dlorenc I think there is some legit failure in there (due to a rebase).

/lgtm cancel

Issue #2185, opened 2020-03-07
Support for standalone Pipelines using standalone TaskRuns


mustafaakin:
Since we can now create TaskRun's without creating the respective Tasks, by specifiying taskSpec instead of taskRef, it would be great if we could use them in PipelineRuns.

It would help when you create pipelines out of the Tekton, dynamically, with another configuration language such as Jsonnet. What do you think?

vdemeester:
@mustafaakin thanks for the issue. You should be able to do that with `PipelineRun`. You can use `PipelineSpec` to [embedded a `Pipeline` spec](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md#specifying-a-pipeline), and in `Pipeline` Spec you can use `TaskSpec` (instead of `TaskRef`) to embedded a `Task` spec (looks like something is missing in the doc).

/kind question

mustafaakin:
@vdemeester oh, good to know, I missed it! Really glad to hear that. Thanks.

PR #3189, opened 2020-09-10
TestReconcile_ExplicitDefaultSA flakiness fix 🍂

vdemeester:
# Changes

This tries to fix the test flakiness by making sure the informers for
serviceaccounts are up-to-date when we execute the tests. Otherwise we
may run into failure because the serviceaccount is not found.

/cc @pritidesai @bobcatfish 

# Submitter Checklist

review them:

## Reviewer Notes

/kind misc
/area testing
/release-note-none


test/controller.go
```diff
@@ -90,6 +92,7 @@ type Informers struct {
 	Condition        informersv1alpha1.ConditionInformer
 	Pod              coreinformers.PodInformer
 	ConfigMap        coreinformers.ConfigMapInformer
+	ServiceAccount   coreinformers.ServiceAccountInformer
```
pritidesai, line 95: @vdemeester I am trying to understand this, why making this as part of informer? and creating fake service account informer?

test/controller.go
```diff
@@ -241,6 +245,13 @@ func SeedTestData(t *testing.T, ctx context.Context, d Data) (Clients, Informers
 			t.Fatal(err)
 		}
 	}
+	c.Kube.PrependReactor("*", "serviceaccounts", AddToInformer(t, i.ServiceAccount.Informer().GetIndexer()))
```
pritidesai, line 248: This is similar to the other resources we have pipeline, pipelinerun, etc. and I am guessing this would be needed for all service accounts that a task/pipeline is expecting including `default` 🤔

vdemeester: As of today, we don't have the reactor, so the informer are/might be out-of-sync. This can cause errors during tests where the serviceaccount is not found. Having this fixes that problem.

vdemeester: but indeed, we should update the unit tests that manually create the ServiceAccounts to use this instead

pritidesai:
sounds good, thanks @vdemeester lets go with this. Next we will be updating rest of the unit tests in `taskrun_test.go` and add the same informer update in `pipelinerun_test.go`
/approve

bobcatfish:
Part of #2185 - might need to make same change for other tests as well?

Issue #3188, opened 2020-09-10
using WhenExpressions, skipped task runs


r0bj:
# Expected Behavior
According to documentation https://tekton.dev/docs/pipelines/pipelines/#guard-task-execution-using-whenexpressions
"If any of the WhenExpressions evaluate to False, the Task is not run and the Task is listed in the Skipped Tasks section of the PipelineRunStatus."

# Actual Behavior
WhenExpressions evaluate to False but Task runs.

# Steps to Reproduce the Problem

Example from https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml with one modification in line 63 ("exists" -> "some-value"):
```
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: guarded-pipeline
  namespace: tekton-execution
spec:
  params:
    - name: path
      type: string
      description: The path of the file to be created
  workspaces:
    - name: source
      description: |
        This workspace will receive the cloned git repo and be passed
        to the next Task to create a file
  tasks:
    - name: create-file
      when:
        - input: "$(params.path)"
          operator: in
          values: ["README.md"]
      workspaces:
        - name: source
          workspace: source
      taskSpec:
        workspaces:
          - name: source
            description: The workspace to create the readme file in
        steps:
          - name: write-new-stuff
            image: ubuntu
            script: 'touch $(workspaces.source.path)/README.md'
    - name: check-file
      params:
        - name: path
          value: "$(params.path)"
      workspaces:
        - name: source
          workspace: source
      runAfter:
        - create-file
      taskSpec:
        params:
          - name: path
        workspaces:
          - name: source
            description: The workspace to check for the file
        results:
          - name: status
            description: indicates whether the file exists or is missing
        steps:
          - name: check-file
            image: alpine
            script: |
              if test -f $(workspaces.source.path)/$(params.path); then
                printf exists | tee /tekton/results/status
              else
                printf missing | tee /tekton/results/status
              fi
    - name: echo-file-exists
      when:
        - input: "$(tasks.check-file.results.status)"
          operator: in
          values: ["some-value"]
      taskSpec:
        steps:
          - name: echo
            image: ubuntu
            script: 'echo file exists'
    - name: task-should-be-skipped
      when:
        - input: "$(params.path)"
          operator: notin
          values: ["README.md"]
      taskSpec:
        steps:
          - name: echo
            image: ubuntu
            script: exit 1
---
[...]

Issue #3196, opened 2020-09-10
WhenExpressions are neglected when finally tasks are present


VeereshAradhya:
# Expected Behavior
WhenExpressions should work when finally tasks are present

# Actual Behavior
WhenExpressions will not work when finally tasks are present

# Steps to Reproduce the Problem

1. Create a Pipeline with tasks, when expressions and finally tasks 
2. Start the pipeline

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  Client Version: version.Info{Major:"1", Minor:"10+", GitVersion:"v1.10.0+b3b92b2", GitCommit:"b3b92b2", GitTreeState:"clean", BuildDate:"2020-07-15T09:27:21Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"linux/amd64"}
  Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.3+8b0a82f", GitCommit:"8b0a82f", GitTreeState:"clean", BuildDate:"2020-07-10T05:34:00Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version:
  ```
  $ tkn version
  Client version: 0.12.0
  Pipeline version: v0.16.0
  Triggers version: unknown
  ```

**Command Execution Logs**:
Execution logs contain the output of both `pipeline with finally task` and `pipeline without finally task`
```
[varadhya@localhost workspace-testing]$ tkn t ls
NAME          DESCRIPTION          AGE
final-task    This is first task   8 minutes ago
first-task    This is first task   17 minutes ago
second-task   This is first task   17 minutes ago
[varadhya@localhost workspace-testing]$ 
[varadhya@localhost workspace-testing]$ tkn p ls
No Pipelines found
[varadhya@localhost workspace-testing]$ 
[varadhya@localhost workspace-testing]$ cat pipeline.yaml 
apiVersion: tekton.dev/v1beta1 
kind: Pipeline 
metadata:
  name: test-when-expression-pipeline
spec:
  params:
    - name: status
  tasks:
    - name: first-task
      taskRef:
        name: first-task
        kind: Task
    - name: second-task
      taskRef:
        name: second-task
        kind: Task
      when:
        - input: "$(params.status)"
          operator: in
          values: ["Success", "success"]
[varadhya@localhost workspace-testing]$ 
[...]

PR #3194, opened 2020-09-10
Add more details on using Results in WhenExpressions

jerop:
# Changes

Add more information to the docs about using `Results` in `WhenExpressions`.

Specifically:
- add details and examples on how `Results` are used in `WhenExpressions` in addition to `Parameters`
- change the name of the result used in the `WhenExpressions` examples, from `status` to `exists` to clarify that it's not the task status

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [x] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

Double check this list of stuff that's easy to miss:

- If you are adding [a new binary/image to the `cmd` dir](../cmd), please update
  [the release Task](../tekton/publish.yaml) to build and release this image.

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

# Release Notes
```release-note
    NONE
```

PR #3197, opened 2020-09-10
Add check for skipping in When Expressions with Result variable replacement

jerop:
# Changes

tasks with when expressions using variables that evaluated to false were mistakenly executed because it didn't have an additional check to validate that the task should be skipped after task result variable replacement

the point where we excluded the tasks with when expressions containing static inputs or parameter variables (which are replaced earlier as read from the pipeline) is here: https://github.com/tektoncd/pipeline/blob/079a6c8bce6eb5663689fad7f50d0fce7591e6bc/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L260-L272

at that point, when expressions containing results don't have the variables replaced so they are not excluded -- that function is used here: 
https://github.com/tektoncd/pipeline/blob/906414229d999cd9c7e6a5401f937c0bb3d97e7f/pkg/reconciler/pipelinerun/pipelinerun.go#L499-L511

the task results are applied later, right after which we add this additional check to exclude the when expressions containing results: https://github.com/tektoncd/pipeline/blob/906414229d999cd9c7e6a5401f937c0bb3d97e7f/pkg/reconciler/pipelinerun/pipelinerun.go#L522-L527

Fixes https://github.com/tektoncd/pipeline/issues/3188

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
`Tasks` with `WhenExpressions` using variable replacements are not executed when the `WhenExpressions` evaluate to false
```

### DIFF
--- a/examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml
+++ b/examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml
@@ -66,6 +66,16 @@ spec:
           - name: echo
             image: ubuntu
             script: 'echo file exists'
+    - name: echo-file-missing
+      when:
+        - input: "$(tasks.check-file.results.exists)"
+          operator: in
+          values: ["missing"]
+      taskSpec:
+        steps:
+          - name: echo
+            image: ubuntu
+            script: 'echo file missing'
     - name: task-should-be-skipped
       when:
         - input: "$(params.path)"
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -522,7 +522,7 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
 	resources.ApplyTaskResults(nextRprts, resolvedResultRefs)
 
 	for _, rprt := range nextRprts {
-		if rprt == nil {
+		if rprt == nil || rprt.Skip(pipelineState, d) {
 			continue
 		}
 		if rprt.ResolvedConditionChecks == nil || rprt.ResolvedConditionChecks.IsSuccess() {
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -2081,6 +2081,21 @@ func TestReconcileWithWhenExpressionsWithParameters(t *testing.T) {
 	if d := cmp.Diff(actualSkippedTasks, expectedSkippedTasks); d != "" {
 		t.Errorf("expected to find Skipped Tasks %v. Diff %s", expectedSkippedTasks, diff.PrintWantGot(d))
 	}
+
+	skippedTasks := []string{"hello-world-2"}
+	for _, skippedTask := range skippedTasks {
+		labelSelector := fmt.Sprintf("tekton.dev/pipelineTask=%s,tekton.dev/pipelineRun=test-pipeline-run-different-service-accs", skippedTask)
+		actualSkippedTask, err := clients.Pipeline.TektonV1beta1().TaskRuns("foo").List(metav1.ListOptions{
+			LabelSelector: labelSelector,
+			Limit:         1,
+		})
+		if err != nil {
+			t.Fatalf("Failure to list TaskRun's %s", err)
+		}
+		if len(actualSkippedTask.Items) != 0 {
+			t.Fatalf("Expected 0 TaskRuns got %d", len(actualSkippedTask.Items))
+		}
+	}
 }
 
 func TestReconcileWithWhenExpressionsWithTaskResults(t *testing.T) {
@@ -2188,6 +2203,21 @@ func TestReconcileWithWhenExpressionsWithTaskResults(t *testing.T) {
 	}}
 	if d := cmp.Diff(actualSkippedTasks, expectedSkippedTasks); d != "" {
 		t.Errorf("expected to find Skipped Tasks %v. Diff %s", expectedSkippedTasks, diff.PrintWantGot(d))
+	}
+
+	skippedTasks := []string{"c-task", "d-task"}
+	for _, skippedTask := range skippedTasks {
+		labelSelector := fmt.Sprintf("tekton.dev/pipelineTask=%s,tekton.dev/pipelineRun=test-pipeline-run-different-service-accs", skippedTask)
+		actualSkippedTask, err := clients.Pipeline.TektonV1beta1().TaskRuns("foo").List(metav1.ListOptions{
+			LabelSelector: labelSelector,
+			Limit:         1,
+		})
+		if err != nil {
+			t.Fatalf("Failure to list TaskRun's %s", err)
+		}
+		if len(actualSkippedTask.Items) != 0 {
+			t.Fatalf("Expected 0 TaskRuns got %d", len(actualSkippedTask.Items))
+		}
 	}
 }
 


examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml
```diff
@@ -66,6 +66,16 @@ spec:
           - name: echo
             image: ubuntu
             script: 'echo file exists'
+    - name: echo-file-missing
+      when:
+        - input: "$(tasks.check-file.results.status)"
```
bobcatfish, line 71: i wonder if we should call this something other than "status" to reduce confusion around being able to execute if a previous task fails

jerop: made the update on the naming in this open pr: https://github.com/tektoncd/pipeline/pull/3194

should i move the naming change here?

bobcatfish:
Can you confirm that the example you updated fails without your change?

I also want to check, #3196 implies this is finally related but it sounds like that's a red herring and it's actually related to variable replacement?

(I'm wondering also if makes sense to have an example/test that verifies that finally works properly with this as well)

pritidesai:
@jerop, please add a unit test in `pipelinerun_test.go`. I just modified an existing unit test to not match the task result value (resulting when expression to `false`) and that task was skipped.

I am trying to reproduce it through the integration test from examples.

jerop:
@pritidesai thank you for the review! 

added a task to the when expressions example that reproduced the issue, and also added checks in the unit tests to ensure the taskruns are not created when variable replacement is used -- please have a look when you have a chance

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -522,7 +522,7 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
 	resources.ApplyTaskResults(nextRprts, resolvedResultRefs)
 
 	for _, rprt := range nextRprts {
-		if rprt == nil {
+		if rprt == nil || rprt.Skip(pipelineState, d) {
```
pritidesai, line 525: This looks fine for now but would like to see this being evaluated as an attribute to `WhenExpressions` similar to what we have for the parameter. That way, all the checks are in one place irrespective of param mismatch or task result mismatch, a task and all its dependencies are guaranteed to be skipped.

jerop: the point where we excluded the tasks with when expressions containing static inputs or parameter variables (which are replaced earlier as read from the pipeline) is here: https://github.com/tektoncd/pipeline/blob/079a6c8bce6eb5663689fad7f50d0fce7591e6bc/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L260-L272


at that point, when expressions containing results don't have the variables replaced so they are not excluded -- that function is used here: 
https://github.com/tektoncd/pipeline/blob/906414229d999cd9c7e6a5401f937c0bb3d97e7f/pkg/reconciler/pipelinerun/pipelinerun.go#L499-L511


the task results are applied later right before this additional check to exclude the when expressions containing results: https://github.com/tektoncd/pipeline/blob/906414229d999cd9c7e6a5401f937c0bb3d97e7f/pkg/reconciler/pipelinerun/pipelinerun.go#L522-L527

the options I see are:

1. removing the earlier check, so that we have only one check after task results have been applied -- but this means will be processing more tasks than necessary which could have been excluded earlier
2. as we have it now, do two checks with the first one reducing the list next tasks and reducing unnecessary work -- I'm leaning towards this but can see the benefit of doing the check once

please let me know what you think

pritidesai: @jerop I am with you - option 2 is fine.

But I was thinking more towards, adding call to `ApplyTaskResults` in `when_types.go` and update `AllowsExecution` to deal with constants, params, and task results. `AllowsExecution` in the end would be responsible for checking if constants match, resolving params/task results and check if they match. Similar to what we do with `applyReplacements` and `GetVarSubsutitionExpressions` as attributes to `WhenExpression`.

These changes does not have to be part of this PR. It's more about refactoring which can be done in a separate PR.

jerop: thanks for the clarification @pritidesai -- will do that refactor soon

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -2081,6 +2081,21 @@ func TestReconcileWithWhenExpressionsWithParameters(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("Failure to list TaskRun's %s", err)
+		}
+		if len(actualSkippedTask.Items) != 0 {
+			t.Fatalf("Expected 0 TaskRuns got %d", len(actualSkippedTask.Items))
+		}
+	}
 }
 
 func TestReconcileWithWhenExpressionsWithTaskResults(t *testing.T) {
```
pritidesai, line 2101: @jerop I am expecting a pipeline task here which has reference to task result but the result does not match with what is expected i.e.

```
		tb.PipelineTask("task-result-mismatch", "task",
			tb.PipelineTaskWhenExpression("$(tasks.a-task.results.aResult)", selection.In, []string{"random"}),
		),
```

pritidesai: And also please add a test, checking the pipeline skips the entire branch in case of a task result mismatch.

jerop: @pritidesai the example with result reference doesn't match what is expected is already here (c-task): https://github.com/tektoncd/pipeline/pull/3197/files/906414229d999cd9c7e6a5401f937c0bb3d97e7f#diff-f7c8532cfe79319c0fff711a4979a6d6R2107

and it has a dependent (d-task) that's also skipped https://github.com/tektoncd/pipeline/pull/3197/files/906414229d999cd9c7e6a5401f937c0bb3d97e7f#diff-f7c8532cfe79319c0fff711a4979a6d6R2112

examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml
```diff
@@ -66,6 +66,16 @@ spec:
           - name: echo
             image: ubuntu
             script: 'echo file exists'
+    - name: echo-file-missing
```
bobcatfish, line 69: i think it would be worth adding a couple comments or changing the names of the pipelinetasks to reflect what we are trying to cover here, which i think is:

1. using a task result which causes the when expression to fail (echo-file-missing)
2. using a pipeline param which causes the when expression to fail (task-should-be-skipped)

This example doesnt seem to include any cases where the when expression succeeds 🤔  cant remember if there was another exmaple? i think we should add that as well if not

but we dont have to do that in this pr!

bobcatfish (commented):
I agree with @pritidesai about refactoring the code a bit - and @jerop with your analysis of trying to find a way to only have to check once so we dont have to worry about checking for skipped tasks correctly in two different places!

but let's tackle that after we get this fixed :D

/lgtm
/approve

examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml
jerop: sounds good, will update the names/add comments 

`echo-file-exists` is an example using when expressions with results and succeeds -- it uses the same result `$(tasks.check-file.results.exists)` and evaluates to true:
https://github.com/tektoncd/pipeline/blob/18dd0af6abf304c18b1484c52cf3b75c7e985160/examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml#L59-L78

`create-file` is the example that uses when expressions with params and succeeds:
https://github.com/tektoncd/pipeline/blob/18dd0af6abf304c18b1484c52cf3b75c7e985160/examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml#L16-L31

bobcatfish: whoops, sorry i missed that @jerop !

PR #3201, opened 2020-09-10
Disable keep-alive for cloud-event connections

### LINKED ISSUE #3190: Cloud Events delivery leaks open connections
# Expected Behavior
The pipeline controller does not leak open connection.
Possible fixes:
- re-use the HTTP client across reconciles
- do not keep idle connections open
- a combination of the two

# Actual Behavior
As described in https://github.com/tektoncd/triggers/issues/687, the pipeline controller seems to create a new HTTP client on every reconcile. The transport associated to it keeps the connections open.

1. This is reproducible in the upstream CI/CD based Tekton (dogfooding)
2. Configure pipeline to send cloud events to an event listener sink
3. Run many taskruns and pipelineruns

# Additional Info
See https://github.com/tektoncd/triggers/issues/687 for more details

- Kubernetes version:

```
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.13-gke.1", GitCommit:"688c6543aa4b285355723f100302d80431e411cc", GitTreeState:"clean", BuildDate:"2020-07-21T02:37:26Z", GoVersion:"go1.13.9b4", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

```
Pipeline version: v0.15.2
Triggers version: v0.7.0
```

afrittoli:
# Changes

Disable keep alive forces the HTTP client to drop the connection
once a response is received. This avoids building up large numbers
of idle connections and it fixes the immediate issue.

After this we may want to see how to ensure we can re-use connection
and also set and idle-connection timeout.

Fixes #3190

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```release-note
This release fixes an issue with the cloud event delivery that lead to an increasingly high number of open (idle) connections towards the configured target URL. The fix is achieved by disabling keep-alive, which means that - at least for now - we will be setting up a new connection for every cloud event. This is the current behaviour too, only now old connections will be closed immediatly.
```

/kind bug

### DIFF
--- a/pkg/reconciler/events/cloudevent/cloudeventclient.go
+++ b/pkg/reconciler/events/cloudevent/cloudeventclient.go
@@ -18,6 +18,7 @@ package cloudevent
 
 import (
 	"context"
+	"net/http"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


pkg/reconciler/events/cloudevent/cloudeventclient.go
```diff
@@ -35,11 +36,13 @@ type CECKey struct{}
 func withCloudEventClient(ctx context.Context, cfg *rest.Config) context.Context {
 	logger := logging.FromContext(ctx)
 
-	p, err := cloudevents.NewHTTP()
+	var useOnceTransport http.RoundTripper = &http.Transport{
```
ImJasonH, line 39: Can you add a comment describing why we're doing this, with a link to the bug and a TODO to remove it when some upstream bug is closed? I don't want to forget we're doing this.

dibyom:
This seems fine for now. We were already opening new connections for each cloud event delivery. This ensures that we close those connections instead of keeping them idle. In the future, would be nice to actually reuse the connections.
/lgtm

vdemeester (commented):
/lgtm

pkg/reconciler/events/cloudevent/cloudeventclient.go
afrittoli: Good point, I will do.
I don't have an upstream bug created yet, but I'll link the pipeline bug for context.

afrittoli:
Follow up feature: https://github.com/tektoncd/pipeline/issues/3204

bobcatfish:
Cherry picking into 0.16.x in https://github.com/tektoncd/pipeline/pull/3215

PR #3207, opened 2020-09-11
Update TaskRun status docs

qu1queee:
# Changes

When a TaskRun is cancelled on purpose, the completionTime will be set, this pr update the docs to reflect that. The current documentation states that the taskRun will not get a CompletionTime, which I think is not correct.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

### DIFF
--- a/docs/taskruns.md
+++ b/docs/taskruns.md
@@ -323,7 +323,7 @@ The following tables shows how to read the overall status of a `TaskRun`:
 Unknown|Started|No|The TaskRun has just been picked up by the controller.
 Unknown|Pending|No|The TaskRun is waiting on a Pod in status Pending.
 Unknown|Running|No|The TaskRun has been validate and started to perform its work.
-Unknown|TaskRunCancelled|No|The user requested the TaskRun to be cancelled. Cancellation has not be done yet.
+Unknown|TaskRunCancelled|Yes|The user requested the TaskRun to be cancelled. Cancellation has not be done yet.
 True|Succeeded|Yes|The TaskRun completed successfully.
 False|Failed|Yes|The TaskRun failed because one of the steps failed.
 False|\[Error message\]|No|The TaskRun encountered a non-permanent error, and it's still running. It may ultimately succeed.


dibyom:
hi @qu1queee thanks for opening the PR, could you sign the CLA?

qu1queee:
@dibyom done

docs/taskruns.md
```diff
@@ -323,7 +323,7 @@ The following tables shows how to read the overall status of a `TaskRun`:
 Unknown|Started|No|The TaskRun has just been picked up by the controller.
 Unknown|Pending|No|The TaskRun is waiting on a Pod in status Pending.
 Unknown|Running|No|The TaskRun has been validate and started to perform its work.
-Unknown|TaskRunCancelled|No|The user requested the TaskRun to be cancelled. Cancellation has not be done yet.
+Unknown|TaskRunCancelled|Yes|The user requested the TaskRun to be cancelled. Cancellation has not be done yet.
```
afrittoli, line 326: Uhm, this would be surprising. Did you find a case where `status` was `Unknown` and `completionTime` not set? 

I'm asking because the controller always sets the `completionTime` and the `status` at the same time - or at least that's what's supposed to do. 

Specifically for cancellation, when the user requests cancellation, it may be that it is successful right away - in which case the `status` is set to `False` and `completionTime` is set - see L331 below.
It maybe that cancellation is not possible right away, in which case the reason will still be set to `TaskRunCancelled`, but `status` will stay to `Unknown` and `completionTime` will not be set.
I don't have a reproducer handy for this condition, the most common case will be that in L331 - but this is still a possibility.

afrittoli (changes requested):
Thank you for your contribution!

I believe the table is correct how it is, but if not there may be an issue in the controller.
In any case we should not be setting the `completionTime` for a `TaskRun` in an `Unknown` state.

docs/taskruns.md
qu1queee: you are right, this was a mistake, the COMPLETIONTIME is never set as long as the SUCCEEDED column is `Unknown`, see:

```
$ k get taskruns --all-namespaces -w
NAMESPACE        NAME                            SUCCEEDED   REASON   STARTTIME   COMPLETIONTIME
test-build-201   buildrun-test-build-201-pvzh8
test-build-201   buildrun-test-build-201-pvzh8
test-build-201   buildrun-test-build-201-pvzh8   Unknown     Pending   0s
test-build-201   buildrun-test-build-201-pvzh8   Unknown     Pending   0s
test-build-201   buildrun-test-build-201-pvzh8   Unknown     Pending   1s
test-build-201   buildrun-test-build-201-pvzh8   Unknown     Pending   2s
test-build-201   buildrun-test-build-201-pvzh8   Unknown     Pending   3s
test-build-201   buildrun-test-build-201-pvzh8   False       TaskRunCancelled   3s          0s
```

@afrittoli pls close the pr.

Issue #3178, opened 2020-09-08
Deprecate and remove test builders


imjasonh:
https://github.com/tektoncd/pipeline/tree/master/internal/builder

New uses of and additions to these packages should be discouraged, and eventually these packages should be phased out entirely.

Hand-maintaining these test builder methods is toilsome and error-prone. They claim to help readability, but instead they introduce a Tekton-specific dialect of Go when users could just write Go-standard struct declarations themselves, which have the advantage of named fields over positional function args.

See #3124 for an example of migrating off of test builders.

/assign @ImJasonH

pritidesai:
It will be great to break this down one test file or one Go struct (Task/PipelineTask/Pipeline) at a time and would make a good `good-first-issue`, see the changes [here](https://github.com/tektoncd/pipeline/pull/3124/files#diff-6d08de825ac4bd80c1560c62a3bc63fdR28) for reference.

ImJasonH:
Here's a little oneliner to find the top 10 files by occurrences of `tb.`, which is generally what we alias test builder imports to:

```
$ git grep -c "tb\." | sed -e 's/:/ /g' | sort -k2 -n -r | head -n 10
pkg/reconciler/pipelinerun/pipelinerun_test.go 749
pkg/reconciler/taskrun/taskrun_test.go 737
pkg/reconciler/pipelinerun/resources/apply_test.go 170
pkg/apis/pipeline/v1alpha1/pipeline_validation_test.go 128
internal/builder/v1beta1/task_test.go 83
internal/builder/v1beta1/pipeline_test.go 80
test/v1alpha1/pipelinerun_test.go 78
pkg/reconciler/events/cloudevent/cloud_event_controller_test.go 78
pkg/timeout/handler_test.go 74
pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go 64
```

And a count of total `tb\.` instances:

```
$ git grep "tb\." | wc -l
    3141
```

I'd probably start with just replacing references to `tb.Pipeline` and `tb.PipelineRun`, and, if you feel like it in the process, cleaning up instances where a test creates a Pipeline then a PipelineRun instead of just creating a PipelineRun with `PipelineSpec` defined inline directly (same with Tasks). That'd knock out a lot of instances pretty quickly. Checkpoint every so often and send a PR.

With enough of those changes you'd be able to delete some unused test builder methods, until eventually you have all of them and you can delete whole files/packages 🎉 

[...]

Issue #3139, opened 2020-08-26
When Expressions Status


jerop:
Tracking discussions about When Expressions Status, particularly:
- how to use the current knative's `ConditionSucceeded` in When Expressions
- introducing a `ConditionsSkipped` type to use in place of `ConditionSucceeded` 

@pritidesai's [comment](https://github.com/tektoncd/pipeline/pull/3135#issuecomment-680681770):
> 
> Ran a simple PipelineRun with one task with `when` expression evaluating to `false` which resulted in a TaskRun with status `failed` similar to what happens when the condition fails for a task (status `failed` with `ConditionCheckFailed`).
> 
> ```
> apiVersion: tekton.dev/v1beta1
> kind: PipelineRun
> metadata:
>   name: pipelinerun-to-skip-task
> spec:
>   pipelineSpec:
>     tasks:
>       - name: skip-this-task
>         when:
>           - input: "foo"
>             operator: notin
>             values: ["foo"]
>         taskSpec:
>           steps:
>             - name: echo
>               image: ubuntu
>               script: |
>                 echo "Good Night!"
> ```
> 
> ```
> $ tkn pr describe pipelinerun-to-skip-task
> Name:        pipelinerun-to-skip-task
> Labels:
>  tekton.dev/pipeline=pipelinerun-to-skip-task
> 
> 🌡️  Status
> 
> STARTED        DURATION    STATUS
> 1 minute ago   0 seconds   Succeeded(Completed)
> 
> 🗂  Taskruns
> 
>  NAME                                              TASK NAME        STARTED   DURATION   STATUS
>  ∙ pipelinerun-to-skip-task-skip-this-task-25tgt   skip-this-task   ---       ---        Failed(WhenExpressionsEvaluatedToFalse)
> ```
> 
> But the way users might start using `when` expressions would be to skip or ignore a task based on the param which was not possible before without writing conditions. For an ignore use case, the status `failed` might not make sense. Also, the `taskrun` which is listed here is just the `taskrun` object created as part of the `pipelinerun` status not actual `taskrun` which can not be described anyways:
> 
> ```
> tkn tr describe pipelinerun-to-skip-task-skip-this-task-25tgt
> Error: failed to get TaskRun pipelinerun-to-skip-task-skip-this-task-25tgt: taskruns.tekton.dev "pipelinerun-to-skip-task-skip-this-task-25tgt" not found
> ```
> 
[...]

PR #3242, opened 2020-09-16
consuming task results in finally

### LINKED ISSUE #2557: Design and implement task results in final tasks
finally at the pipeline level is on its way:

[design doc](https://docs.google.com/document/d/1lxpYQHppiWOxsn4arqbwAFDo4T0-LCqpNa6p-TJdHrw/edit) 
[PR](https://github.com/tektoncd/pipeline/pull/2437) - which will be split into multiple PRs

First iteration of final tasks are implemented without task results and `from` clause for Pipeline Resources. In next iteration, please finalize the [design](https://docs.google.com/document/d/10iEJqVstY6k3KNvAXgffIJLcHRbPQ-GIAfQk5Dlrf3c/edit#heading=h.rya2cci9aa5x) for adding support for task results.

Reference:
* https://github.com/tektoncd/pipeline/pull/2437#discussion_r420298647 

* https://github.com/tektoncd/pipeline/issues/2446 (Implement finally at the pipeline level)

* https://github.com/tektoncd/pipeline/issues/1684 (part of designing failure strategy for taskrun in a pipeline)

pritidesai:
# Changes

Final tasks can be configured to consume `results` of `PipelineTasks` from `tasks` section.

Based on the TEP, a final task failing to resolve task results is added to the list of `skippedTasks` and warning is recorded in controller logs.

```
unable to find result referenced by param "param1" in "final-task" ...
```

Implements TEP [#0004](https://github.com/tektoncd/community/blob/master/teps/0004-task-results-in-final-tasks.md)
Closes #2557 

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Final tasks can be configured to consume `results` of `PipelineTasks` from `tasks` section. 
```

### DIFF
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -788,6 +788,32 @@ spec:
           value: "someURL"
 ```
 
+#### Consuming `Task` execution results in `finally`
+
+Final tasks can be configured to consume `Results` of `PipelineTask` from the `tasks` section:
+
+```yaml
+spec:
+  tasks:
+    - name: clone-app-repo
+      taskRef:
+        name: git-clone
+  finally:
+    - name: discover-git-commit
+      params:
+        - name: commit
+          value: $(tasks.clone-app-repo.results.commit)
+```
+**Note:** The scheduling of such final task does not change, it will still be executed in parallel with other
+final tasks after all non-final tasks are done.
+
+The controller resolves task results before executing the finally task `discover-git-commit`. If the task
+`clone-app-repo` failed or skipped with [when expression](#guard-task-execution-using-whenexpressions) resulting in
+uninitialized task result `commit`, the finally Task `discover-git-commit` will be included in the list of
+`skippedTasks` and continues executing rest of the final tasks. The pipeline exits with `completion` instead of
+`success` if a finally task is added to the list of `skippedTasks`.
+
+
 ### `PipelineRun` Status with `finally`
 
 With `finally`, `PipelineRun` status is calculated based on `PipelineTasks` under `tasks` section and final tasks.
@@ -899,35 +925,6 @@ no `runAfter` can be specified in final tasks.
 `Tasks` in a `Pipeline` can be configured to run only if some conditions are satisfied using `conditions`. But the
 final tasks are guaranteed to be executed after all `PipelineTasks` therefore no `conditions` can be specified in
 final tasks.
-
-#### Cannot configure `Task` execution results with `finally`
-
-Final tasks can not be configured to consume `Results` of `PipelineTask` from `tasks` section i.e. the following
-example is not supported right now but we are working on adding support for the same (tracked in issue
-[#2557](https://github.com/tektoncd/pipeline/issues/2557)).
-
-```yaml
-spec:
-  tasks:
-    - name: count-comments-before
-      taskRef:
-        Name: count-comments
-    - name: add-comment
-      taskRef:
-        Name: add-comment
-    - name: count-comments-after
-      taskRef:
-        Name: count-comments
-  finally:
-    - name: check-count
-      taskRef:
-        Name: check-count
-      params:
-        - name: before-count
-          value: $(tasks.count-comments-before.results.count) #invalid
-        - name: after-count
-          value: $(tasks.count-comments-after.results.count) #invalid
-```
 
 #### Cannot configure `Pipeline` result with `finally`
 
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -581,7 +581,18 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
 	if len(fnextRprts) != 0 {
 		// apply the runtime context just before creating taskRuns for final tasks in queue
 		resources.ApplyPipelineTaskContext(fnextRprts, pipelineRunFacts.GetPipelineTaskStatus(ctx))
-		nextRprts = append(nextRprts, fnextRprts...)
+
+		// Before creating TaskRun for scheduled final task, check if it's consuming a task result
+		// Resolve and apply task result wherever applicable, report warning in case resolution fails
+		for _, rprt := range fnextRprts {
+			resolvedResultRefs, err := resources.ResolveResultRef(pipelineRunFacts.State, rprt)
+			if err != nil {
+				logger.Infof("Final task %q is not executed as it could not resolve task params for %q: %v", rprt.PipelineTask.Name, pr.Name, err)
+				continue
+			}
+			resources.ApplyTaskResults(resources.PipelineRunState{rprt}, resolvedResultRefs)
+			nextRprts = append(nextRprts, rprt)
+		}
 	}
 
 	for _, rprt := range nextRprts {
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -4644,6 +4644,187 @@ func TestReconciler_ReconcileKind_PipelineTaskContext(t *testing.T) {
 	}
 }
 
+func TestReconcileWithTaskResultsInFinalTasks(t *testing.T) {
+	names.TestingSeed()
+
+	ps := []*v1beta1.Pipeline{{
+		ObjectMeta: metav1.ObjectMeta{Name: "test-pipeline", Namespace: "foo"},
+		Spec: v1beta1.PipelineSpec{
+			Tasks: []v1beta1.PipelineTask{{
+				Name:    "dag-task-1",
+				TaskRef: &v1beta1.TaskRef{Name: "dag-task"},
+			}, {
+				Name:    "dag-task-2",
+				TaskRef: &v1beta1.TaskRef{Name: "dag-task"},
+			}},
+			Finally: []v1beta1.PipelineTask{{
+				Name:    "final-task-1",
+				TaskRef: &v1beta1.TaskRef{Name: "final-task"},
+				Params: []v1beta1.Param{{
+					Name: "finalParam",
+					Value: v1beta1.ArrayOrString{
+						Type:      "string",
+						StringVal: "$(tasks.dag-task-1.results.aResult)",
+					}},
+				},
+			}, {
+				Name:    "final-task-2",
+				TaskRef: &v1beta1.TaskRef{Name: "final-task"},
+				Params: []v1beta1.Param{{
+					Name: "finalParam",
+					Value: v1beta1.ArrayOrString{
+						Type:      "string",
+						StringVal: "$(tasks.dag-task-2.results.aResult)",
+					}},
+				},
+			}},
+		},
+	}}
+
+	prs := []*v1beta1.PipelineRun{{
+		ObjectMeta: metav1.ObjectMeta{Name: "test-pipeline-run-final-task-results", Namespace: "foo"},
+		Spec: v1beta1.PipelineRunSpec{
+			PipelineRef:        &v1beta1.PipelineRef{Name: "test-pipeline"},
+			ServiceAccountName: "test-sa-0",
+		},
+	}}
+
+	ts := []*v1beta1.Task{{
+		ObjectMeta: metav1.ObjectMeta{Name: "dag-task", Namespace: "foo"},
+	}, {
+		ObjectMeta: metav1.ObjectMeta{Name: "final-task", Namespace: "foo"},
+		Spec: v1beta1.TaskSpec{
+			Params: []v1beta1.ParamSpec{{
+				Name: "finalParam",
+				Type: v1beta1.ParamTypeString,
+			}},
+		},
+	}}
+
+	trs := []*v1beta1.TaskRun{{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-pipeline-run-final-task-results-dag-task-1-xxyyy",
+			Namespace: "foo",
+			OwnerReferences: []metav1.OwnerReference{{
+				Kind:       "PipelineRun",
+				Name:       "test-pipeline-run-final-task-results",
+				APIVersion: "tekton.dev/v1beta1",
+			}},
+			Labels: map[string]string{
+				"tekton.dev/pipeline":     "test-pipeline",
+				"tekton.dev/pipelineRun":  "test-pipeline-run-final-task-results",
+				"tekton.dev/pipelineTask": "dag-task-1",
+			},
+		},
+		Spec: v1beta1.TaskRunSpec{
+			ServiceAccountName: "test-sa",
+			TaskRef:            &v1beta1.TaskRef{Name: "hello-world"},
+		},
+		Status: v1beta1.TaskRunStatus{
+			Status: duckv1beta1.Status{
+				Conditions: duckv1beta1.Conditions{apis.Condition{
+					Type:   apis.ConditionSucceeded,
+					Status: corev1.ConditionTrue,
+					Reason: v1beta1.PipelineRunReasonSuccessful.String(),
+				}},
+			},
+			TaskRunStatusFields: v1beta1.TaskRunStatusFields{
+				TaskRunResults: []v1beta1.TaskRunResult{{Name: "aResult", Value: "aResultValue"}},
+			},
+		},
+	}, {
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-pipeline-run-final-task-results-dag-task-2-xxyyy",
+			Namespace: "foo",
+			OwnerReferences: []metav1.OwnerReference{{
+				Kind:       "PipelineRun",
+				Name:       "test-pipeline-run-final-task-results",
+				APIVersion: "tekton.dev/v1beta1",
+			}},
+			Labels: map[string]string{
+				"tekton.dev/pipeline":     "test-pipeline",
+				"tekton.dev/pipelineRun":  "test-pipeline-run-final-task-results",
+				"tekton.dev/pipelineTask": "dag-task-2",
+			},
+		},
+		Spec: v1beta1.TaskRunSpec{
+			ServiceAccountName: "test-sa",
+			TaskRef:            &v1beta1.TaskRef{Name: "hello-world"},
+		},
+		Status: v1beta1.TaskRunStatus{
+			Status: duckv1beta1.Status{
+				Conditions: duckv1beta1.Conditions{apis.Condition{
+					Type:   apis.ConditionSucceeded,
+					Status: corev1.ConditionFalse,
+				}},
+			},
+		},
+	}}
+
+	d := test.Data{
+		PipelineRuns: prs,
+		Pipelines:    ps,
+		Tasks:        ts,
+		TaskRuns:     trs,
+	}
+	prt := NewPipelineRunTest(d, t)
+	defer prt.Cancel()
+
+	reconciledRun, clients := prt.reconcileRun("foo", "test-pipeline-run-final-task-results", []string{}, false)
+
+	expectedTaskRunName := "test-pipeline-run-final-task-results-final-task-1-9l9zj"
+	expectedTaskRun := v1beta1.TaskRun{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      expectedTaskRunName,
+			Namespace: "foo",
+			OwnerReferences: []metav1.OwnerReference{{
+				Kind:               "PipelineRun",
+				Name:               "test-pipeline-run-final-task-results",
+				APIVersion:         "tekton.dev/v1beta1",
+				Controller:         &trueb,
+				BlockOwnerDeletion: &trueb,
+			}},
+			Labels: map[string]string{
+				"tekton.dev/pipeline":     "test-pipeline",
+				"tekton.dev/pipelineRun":  "test-pipeline-run-final-task-results",
+				"tekton.dev/pipelineTask": "final-task-1",
+			},
+			Annotations: map[string]string{},
+		},
+		Spec: v1beta1.TaskRunSpec{
+			ServiceAccountName: "test-sa-0",
+			TaskRef:            &v1beta1.TaskRef{Name: "final-task"},
+			Params:             []v1beta1.Param{{Name: "finalParam", Value: v1beta1.ArrayOrString{Type: "string", StringVal: "aResultValue"}}},
+			Resources:          &v1beta1.TaskRunResources{},
+			Timeout:            &metav1.Duration{Duration: config.DefaultTimeoutMinutes * time.Minute},
+		},
+	}
+
+	// Check that the expected TaskRun was created
+	actual, err := clients.Pipeline.TektonV1beta1().TaskRuns("foo").List(prt.TestAssets.Ctx, metav1.ListOptions{
+		LabelSelector: "tekton.dev/pipelineTask=final-task-1,tekton.dev/pipelineRun=test-pipeline-run-final-task-results",
+		Limit:         1,
+	})
+
+	if err != nil {
+		t.Fatalf("Failure to list TaskRun's %s", err)
+	}
+	if len(actual.Items) != 1 {
+		t.Fatalf("Expected 1 TaskRuns got %d", len(actual.Items))
+	}
+	actualTaskRun := actual.Items[0]
+	if d := cmp.Diff(expectedTaskRun, actualTaskRun, ignoreResourceVersion); d != "" {
+		t.Errorf("expected to see TaskRun %v created. Diff %s", expectedTaskRunName, diff.PrintWantGot(d))
+	}
+	expectedSkippedTasks := []v1beta1.SkippedTask{{
+		Name: "final-task-2",
+	}}
+
+	if d := cmp.Diff(expectedSkippedTasks, reconciledRun.Status.SkippedTasks); d != "" {
+		t.Fatalf("Expected to see only one final task (final-task-2) in the list of skipped tasks. Diff: %s", diff.PrintWantGot(d))
+	}
+}
+
 // NewPipelineRunTest returns PipelineRunTest with a new PipelineRun controller created with specified state through data
 // This PipelineRunTest can be reused for multiple PipelineRuns by calling reconcileRun for each pipelineRun
 func NewPipelineRunTest(data test.Data, t *testing.T) *PipelineRunTest {
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -216,6 +216,19 @@ func (t *ResolvedPipelineRunTask) parentTasksSkip(facts *PipelineRunFacts) bool 
 	return false
 }
 
+// IsFinallySkipped returns true if a finally task is not executed and skipped due to task result validation failure
+func (t *ResolvedPipelineRunTask) IsFinallySkipped(facts *PipelineRunFacts) bool {
+	if t.IsStarted() {
+		return false
+	}
+	if facts.checkDAGTasksDone() && facts.isFinalTask(t.PipelineTask.Name) {
+		if _, err := ResolveResultRef(facts.State, t); err != nil {
+			return true
+		}
+	}
+	return false
+}
+
 // GetRun is a function that will retrieve a Run by name.
 type GetRun func(name string) (*v1alpha1.Run, error)
 
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
@@ -376,6 +376,12 @@ func (facts *PipelineRunFacts) GetSkippedTasks() []v1beta1.SkippedTask {
 			}
 			skipped = append(skipped, skippedTask)
 		}
+		if rprt.IsFinallySkipped(facts) {
+			skippedTask := v1beta1.SkippedTask{
+				Name: rprt.PipelineTask.Name,
+			}
+			skipped = append(skipped, skippedTask)
+		}
 	}
 	return skipped
 }
@@ -464,6 +470,9 @@ func (facts *PipelineRunFacts) getPipelineTasksCount() pipelineRunStatusCount {
 			s.Failed++
 		// increment skip counter since the task is skipped
 		case t.Skip(facts):
+			s.Skipped++
+		// checking if any finally tasks were referring to invalid/missing task results
+		case t.IsFinallySkipped(facts):
 			s.Skipped++
 		// increment incomplete counter since the task is pending and not executed yet
 		default:
--- a/test/pipelinefinally_test.go
+++ b/test/pipelinefinally_test.go
@@ -17,7 +17,11 @@ package test
 
 import (
 	"context"
+	"strings"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/pipeline/test/diff"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 
@@ -32,7 +36,7 @@ import (
 	"knative.dev/pkg/test/helpers"
 )
 
-func TestPipelineLevelFinally_OneDAGTaskFailed_Failure(t *testing.T) {
+func TestPipelineLevelFinally_OneDAGTaskFailed_InvalidTaskResult_Failure(t *testing.T) {
 	ctx := context.Background()
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
@@ -55,8 +59,8 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_Failure(t *testing.T) {
 		t.Fatalf("Failed to create dag Task: %s", err)
 	}
 
-	finalTask := getSuccessTask(t, namespace)
-	if _, err := c.TaskClient.Create(ctx, finalTask, metav1.CreateOptions{}); err != nil {
+	successTask := getSuccessTask(t, namespace)
+	if _, err := c.TaskClient.Create(ctx, successTask, metav1.CreateOptions{}); err != nil {
 		t.Fatalf("Failed to create final Task: %s", err)
 	}
 
@@ -65,22 +69,49 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_Failure(t *testing.T) {
 		t.Fatalf("Failed to create final Task checking executing status: %s", err)
 	}
 
-	pipeline := getPipeline(t,
-		namespace,
-		map[string]string{
-			"dagtask1": task.Name,
-			"dagtask2": delayedTask.Name,
-			"dagtask3": finalTask.Name,
+	taskProducingResult := getSuccessTaskProducingResults(t, namespace)
+	if _, err := c.TaskClient.Create(ctx, taskProducingResult, metav1.CreateOptions{}); err != nil {
+		t.Fatalf("Failed to create Task producing task results: %s", err)
+	}
+
+	taskConsumingResult := getSuccessTaskConsumingResults(t, namespace)
+	if _, err := c.TaskClient.Create(ctx, taskConsumingResult, metav1.CreateOptions{}); err != nil {
+		t.Fatalf("Failed to create Task consuming task results: %s", err)
+	}
+
+	we := v1beta1.WhenExpressions{{
+		Input:    "foo",
+		Operator: "notin",
+		Values:   []string{"foo"},
+	}}
+
+	dag := map[string]pipelineTask{
+		"dagtask1": {
+			TaskName: task.Name,
 		},
-		map[string]string{
-			"dagtask3": cond.Name,
+		"dagtask2": {
+			TaskName: delayedTask.Name,
 		},
-		map[string]string{
-			"finaltask1": finalTask.Name,
-			"finaltask2": finalTaskWithStatus.Name,
+		"dagtask3": {
+			TaskName:  successTask.Name,
+			Condition: cond.Name,
 		},
-		map[string][]v1beta1.Param{
-			"finaltask2": {{
+		"dagtask4": {
+			TaskName: successTask.Name,
+			When:     we,
+		},
+		"dagtask5": {
+			TaskName: taskProducingResult.Name,
+		},
+	}
+
+	f := map[string]pipelineTask{
+		"finaltask1": {
+			TaskName: successTask.Name,
+		},
+		"finaltask2": {
+			TaskName: finalTaskWithStatus.Name,
+			Param: []v1beta1.Param{{
 				Name: "dagtask1-status",
 				Value: v1beta1.ArrayOrString{
 					Type:      "string",
@@ -100,7 +131,42 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_Failure(t *testing.T) {
 				},
 			}},
 		},
-	)
+		// final task consuming result from a failed dag task
+		"finaltaskconsumingdagtask1": {
+			TaskName: taskConsumingResult.Name,
+			Param: []v1beta1.Param{{
+				Name: "dagtask-result",
+				Value: v1beta1.ArrayOrString{
+					Type:      "string",
+					StringVal: "$(tasks.dagtask1.results.result)",
+				},
+			}},
+		},
+		// final task consuming result from a skipped dag task due to when expression
+		"finaltaskconsumingdagtask4": {
+			TaskName: taskConsumingResult.Name,
+			Param: []v1beta1.Param{{
+				Name: "dagtask-result",
+				Value: v1beta1.ArrayOrString{
+					Type:      "string",
+					StringVal: "$(tasks.dagtask4.results.result)",
+				},
+			}},
+		},
+		"finaltaskconsumingdagtask5": {
+			TaskName: taskConsumingResult.Name,
+			Param: []v1beta1.Param{{
+				Name: "dagtask-result",
+				Value: v1beta1.ArrayOrString{
+					Type:      "string",
+					StringVal: "$(tasks.dagtask5.results.result)",
+				},
+			}},
+		},
+	}
+
+	pipeline := getPipeline(t, namespace, dag, f)
+
 	if _, err := c.PipelineClient.Create(ctx, pipeline, metav1.CreateOptions{}); err != nil {
 		t.Fatalf("Failed to create Pipeline: %s", err)
 	}
@@ -114,9 +180,24 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_Failure(t *testing.T) {
 		t.Fatalf("Waiting for PipelineRun %s to fail: %v", pipelineRun.Name, err)
 	}
 
+	// Get the status of the PipelineRun.
+	pr, err := c.PipelineRunClient.Get(ctx, pipelineRun.Name, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("Failed to get PipelineRun %q: %v", pipelineRun.Name, err)
+	}
+
 	taskrunList, err := c.TaskRunClient.List(ctx, metav1.ListOptions{LabelSelector: "tekton.dev/pipelineRun=" + pipelineRun.Name})
 	if err != nil {
 		t.Fatalf("Error listing TaskRuns for PipelineRun %s: %s", pipelineRun.Name, err)
+	}
+
+	// expecting taskRuns for dagtask1, dagtask2, dagtask3 (with condition failure), dagtask5, finaltask1, finaltask2, and finaltaskconsumingdagtask5
+	if len(taskrunList.Items) != 7 {
+		var s []string
+		for _, n := range taskrunList.Items {
+			s = append(s, n.Labels["tekton.dev/pipelineTask"])
+		}
+		t.Fatalf("Error retrieving TaskRuns for PipelineRun %s. Expected 5 taskRuns and found taskRuns for: %s", pipelineRun.Name, strings.Join(s, ", "))
 	}
 
 	var dagTask1EndTime, dagTask2EndTime, finalTaskStartTime *metav1.Time
@@ -136,6 +217,12 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_Failure(t *testing.T) {
 		case n == "dagtask3":
 			if !isSkipped(t, n, taskrunItem.Status.Conditions) {
 				t.Fatalf("dag task %s should have skipped due to condition failure", n)
+			}
+		case n == "dagtask4":
+			t.Fatalf("task %s should have skipped due to when expression", n)
+		case n == "dagtask5":
+			if err := WaitForTaskRunState(ctx, c, taskrunItem.Name, TaskRunSucceed(taskrunItem.Name), "TaskRunSuccess"); err != nil {
+				t.Errorf("Error waiting for TaskRun to succeed: %v", err)
 			}
 		case n == "finaltask1":
 			if err := WaitForTaskRunState(ctx, c, taskrunItem.Name, TaskRunSucceed(taskrunItem.Name), "TaskRunSuccess"); err != nil {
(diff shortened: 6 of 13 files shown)


docs/pipelines.md
```diff
@@ -688,6 +688,36 @@ spec:
...
+        Name: count-comments
+  finally:
+    - name: check-count
+      taskRef:
+        Name: check-count
+      params:
+        - name: before-count
+          value: $(tasks.count-comments-before.results.count) #invalid
+        - name: after-count
+          value: $(tasks.count-comments-after.results.count) #invalid
```
jerop, line 715: i think the `#invalid` comments should be removed

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -305,6 +311,18 @@ func (state PipelineRunState) GetTaskRunsStatus(pr *v1beta1.PipelineRun) map[str
 				})
 			}
 		}
+		// Maintain a TaskRun Object in pr.Status for a finally task which could not resolve task results
```
jerop, line 314: saw that this was already agreed upon in the TEP, but I think it might be worthwhile to explore adding these final tasks to skipped tasks (or another separate section for final tasks with invalid results) -- instead of creating a TaskRun object for a final task that wasn't executed 

related discussion: https://github.com/tektoncd/pipeline/issues/3139

jerop (commented):
excited to see results supported in finally tasks! 

we can look into extending it to support guarding execution of final tasks based on task results using when expressions

docs/pipelines.md
pritidesai: great catch @jerop, done 👍

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
pritidesai: yes thats an alternative, another alternative could be: Tekton controller replacing such unresolved param with type default which is empty string for string type params. Let that final task execute expecting that the task author has implemented sanitization on such input params. @bobcatfish thoughts?

vdemeester (commented):
/lgtm

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -69,6 +69,9 @@ type ResolvedPipelineRunTask struct {
 	ResolvedTaskResources *resources.ResolvedTaskResources
 	// ConditionChecks ~~TaskRuns but for evaling conditions
 	ResolvedConditionChecks TaskConditionCheckState // Could also be a TaskRun or maybe just a Pod?
+	// this flag is set for a task when task result resolution fails for that task
+	// and mainly used to create a task run object with failure for that task
+	InvalidTaskResultsInFinally bool
```
jerop, line 74: wondering if there's a way to avoid saving execution state in resolvedpipelineruntask because it no longer means that it's a PipelineTask with its references resolved (looking into refactoring so that we don't put the taskrun in this struct as well)....we'd initially considered saving the skipped state here when working on when expressions, but decided with calling the method twice (when skipping and when updating status)

cc @bobcatfish

pritidesai: PipelineTask being skipped due to when expression failure is a different scenario then final task skip/failing due to task result resolution failure.

When expression is part of `ResolvedPipelineRunTask` and can be evaluated from `GetPipelineConditionStatus` i.e. we can call `Skip` again and will receive accurate result (once a PipelineTask is declared skipped, `Skip` would return true for that PipelineTask based on WhenExpression evaluation inside of `Skip`).

With final task being skipped/failed due to task results resolution failure is determined while retrieving execution queue and not scheduled if task result resolution fails. There is no means of storing such information in `ResolvedPipelineRunTask` except this flag `InvalidTaskResultsInFinally`. 

The key difference is a source of truth available in `ResolvedPipelineRunTask` to all the reconciling cycles v/s reconciler making decision in one cycle and making it available all subsequent cycles.

I hope this make sense 😜 

I will be happy to hear if we can still avoid such flags (not in favor of it either).

jerop: Yes, that makes sense

Maybe in a future PR after we do some refactoring, we can avoid having that flag there

anderoo:
Happy to see this PR. Moves us closer to being able to conditionally execute finally tasks.

pritidesai:
Please review whenever you can @tektoncd/core-maintainers, would like to get this in 0.17 🙏

sbwsg:
Small suggestion, doesn't have to be a blocker on this PR, but it would be great to see an example yaml added to examples/v1beta1/pipelineruns for this too if possible.

/approve

pritidesai:
thanks @sbwsg, thats a great suggestion. Putting it on a brief hold until I add and end-to-end example.
/hold

pritidesai:
/hold cancel

@sbwsg @jerop I have updated this PR with following changes:

1. Include an [example](https://github.com/tektoncd/pipeline/pull/3242/files#diff-56c052a6b622ca906e10799a94023df4R167) with finally task utilizing result.

2. Updated the [flag](https://github.com/tektoncd/pipeline/pull/3242/files#diff-4e301064fef70f3c0db810267c0959f2R75) `InvalidTaskResultsInFinally` from `bool` to `string` to include the error message in the taskRun object in pipelineRun Status instead of just getting added to controller logs.

3. Populate `pipelineRunState` with such [taskRun](https://github.com/tektoncd/pipeline/pull/3242/files#diff-4e301064fef70f3c0db810267c0959f2R367) object from pipelineRun Status along with [the unit test](https://github.com/tektoncd/pipeline/pull/3242/files#diff-4514daff4848977b060192ad3a3694e4R1855). Discovered an issue with this PR while adding end-to-end example 😢 . Without these changes, `GetPipelineRunConditionStatus` iterating over `pipelineRunState` has no knowledge of a taskRun being declared as failure when such taskRun does not exist in `state`.

sbwsg:
Awesome, thanks Priti, looks good to me!

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3925,6 +3925,132 @@ func TestReconcilePipeline_TaskSpecMetadata(t *testing.T) {
 	}
 }
 
+func TestReconcileWithTaskResultsInFinalTasks(t *testing.T) {
```
popcor255, line 3928: I believe we are trying to deprecate test builders from our tests. :sweat_smile: https://github.com/tektoncd/pipeline/issues/3178

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3925,6 +3925,132 @@ func TestReconcilePipeline_TaskSpecMetadata(t *testing.T) {
...
+		),
+		tb.TaskRunLabel("tekton.dev/pipeline", "test-pipeline"),
+		tb.TaskRunLabel("tekton.dev/pipelineRun", "test-pipeline-run-final-task-results"),
+		tb.TaskRunLabel("tekton.dev/pipelineTask", "final-task-1"),
+		tb.TaskRunSpec(
+			tb.TaskRunTaskRef("final-task-1"),
+			tb.TaskRunServiceAccountName("test-sa-0"),
+			tb.TaskRunParam("finalParam", "aResultValue"),
+		),
+	)
```
popcor255, line 4031: I believe we are trying to deprecate test builders from our tests. :sweat_smile: #3178 
It's not a blocker. I would like to lgtm this. I am glad to make this change later, if you want to merge this now.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3925,6 +3925,132 @@ func TestReconcilePipeline_TaskSpecMetadata(t *testing.T) {
...
+	ps := []*v1beta1.Pipeline{tb.Pipeline("test-pipeline", tb.PipelineNamespace("foo"), tb.PipelineSpec(
+		tb.PipelineTask("dag-task-1", "dag-task"),
+		tb.PipelineTask("dag-task-2", "dag-task"),
+		tb.FinalPipelineTask("final-task-1", "final-task-1",
+			tb.PipelineTaskParam("finalParam", "$(tasks.dag-task-1.results.aResult)"),
+		),
+		tb.FinalPipelineTask("final-task-2", "final-task-2",
+			tb.PipelineTaskParam("finalParam", "$(tasks.dag-task-2.results.aResult)"),
+		),
+	))}
```
popcor255, line 3939: I believe we are trying to deprecate test builders from our tests. :sweat_smile: #3178 
It's not a blocker. I would like to lgtm this. I am glad to make this change later, if you want to merge this now.

pritidesai: thanks @popcor255 yes, will change that later along with the other tests.

vdemeester (commented):
/lgtm

vdemeester:
```
2020-10-07T09:32:59.866Z	DEBUG	TestReconcileTaskResourceResolutionAndValidation/Fail_ValidateResolvedTaskResources	taskrun/reconciler.go:322	Updating status with:   v1beta1.TaskRunStatus{
  	Status: v1beta1.Status{
  		ObservedGeneration: 0,
- 		Conditions:         nil,
+ 		Conditions: v1beta1.Conditions{
+ 			{
+ 				Type:               "Succeeded",
+ 				Status:             "False",
+ 				LastTransitionTime: apis.VolatileTime{Inner: s"2020-10-07 09:32:59.864401656 +0"...},
+ 				Reason:             "TaskRunValidationFailed",
+ 				Message:            "invalid input resources for task test-task-missing-resource: Tas"...,
+ 			},
+ 		},
  		Annotations: nil,
  	},
  	TaskRunStatusFields: v1beta1.TaskRunStatusFields{
  		PodName:        "",
- 		StartTime:      nil,
+ 		StartTime:      s"2020-10-07 09:32:59.863482481 +0000 UTC m=+23.220666862",
  		CompletionTime: nil,
  		Steps:          nil,
  		... // 3 identical fields
  		TaskRunResults: nil,
  		Sidecars:       nil,
- 		TaskSpec:       nil,
+ 		TaskSpec:       &v1beta1.TaskSpec{Resources: &v1beta1.TaskResources{Inputs: []v1beta1.TaskResource{{...}}}},
  	},
  }

2020-10-07T09:32:59.879Z	DEBUG	TestReconcileTaskResourceResolutionAndValidation/Fail_ValidateResolvedTaskResources	taskrun/reconciler.go:322	Updating status with:   v1beta1.TaskRunStatus{
  	Status: v1beta1.Status{
  		ObservedGeneration: 0,
- 		Conditions:         nil,
+ 		Conditions: v1beta1.Conditions{
+ 			{
+ 				Type:               "Succeeded",
+ 				Status:             "False",
+ 				LastTransitionTime: apis.VolatileTime{Inner: s"2020-10-07 09:32:59.864401656 +0"...},
+ 				Reason:             "TaskRunValidationFailed",
+ 				Message:            "invalid input resources for task test-task-missing-resource: Tas"...,
+ 			},
+ 		},
  		Annotations: nil,
  	},
  	TaskRunStatusFields: v1beta1.TaskRunStatusFields{
  		PodName:        "",
- 		StartTime:      nil,
+ 		StartTime:      s"2020-10-07 09:32:59.863482481 +0000 UTC m=+23.220666862",
  		CompletionTime: nil,
  		Steps:          nil,
  		... // 3 identical fields
  		TaskRunResults: nil,
  		Sidecars:       nil,
- 		TaskSpec:       nil,
+ 		TaskSpec:       &v1beta1.TaskSpec{Resources: &v1beta1.TaskResources{Inputs: []v1beta1.TaskResource{{...}}}},
  	},
  }
```
Some unit errors *seem* legit ? :thinking: 
/lgtm cancel

pritidesai:
```
fatal: unable to access 'https://github.com/tektoncd/pipeline.git/': Could not resolve host: github.com
```
/retest

pritidesai:
> Some unit errors seem legit ? 🤔
/lgtm cancel

😭  Yup, recent commits caused the failures, fixed now 🙏

popcor255:
:cry: I check the issues to see if this is a flake. This looks legit error. Im not sure why? :confused: /lgtm cancel

pritidesai:
@popcor255 I have been seeing this kind of intermittent failure quite recently on other [PRs](https://github.com/tektoncd/pipeline/pull/3353#issuecomment-705080791) as well. 

```
Expected 2 number of successful events from pipelinerun and taskrun but got 3; list of received events
```

pritidesai:
@popcor255 tests are all green now 💃  PTAL

docs/pipelines.md
```diff
@@ -699,6 +699,36 @@ spec:
...
+    - name: check-count
+      taskRef:
+        Name: check-count
+      params:
+        - name: before-count
+          value: $(tasks.count-comments-before.results.count)
+        - name: after-count
+          value: $(tasks.count-comments-after.results.count)
+```
+**Note** The scheduling of such final task does not change, it will still be executed in parallel with other
```
afrittoli, line 728: NIT: Missing ':' after **Note**

docs/pipelines.md
```diff
@@ -699,6 +699,36 @@ spec:
           value: "someURL"
 ```
 
+#### Consuming `Task` execution results in `finally`
+
+Final tasks can be configured to consume `Results` of `PipelineTask` from `tasks` section:
```
afrittoli, line 704: NIT: from *the* `tasks` section

afrittoli:
Thanks you for this! I really look forward to using this in our `dogfooding` cluster.

I'm a bit concerned about the final task failure on missing results. 
It matches what specified in the TEP, so unfortunately I'm a bit late to the discussion:
```
Q. What happens to Pipeline when the dependent task boskos-acquire either failed or not executed and the task result leased-resource is not initialized?

A. The finally Task boskos-release is attempted but declared failure since the task result leased-resource is not initialized. Pipeline controller logs the reason of this failure as invalid result reference InvalidTaskResultReference. Pipeline continues executing rest of the final tasks and exits with failure.
```

I have two concerns:
- From a pipeline run status point of view, we're now hosting in the pipeline run status a reference to a taskrun that was never created (at least that's my understanding). We're already doing that for `Conditions` and it's really confusing and hard to parse for both humans and UIs (dashboard / tkn). I'd hope we would do less of that and not more.
- From a workflow pov, I'd rather skip a finally task that cannot be executed because of missing results. The outcome of the pipeline will then match the reason for the results to be missing:
  - if the results are missing because the producer task failed, the pipeline will fail
  - if the results are missing because the producer was skipped:
     - if the producer was skipped without failures, the pipeline may still pass
     - if the producer was skipped because of some failure, the pipeline will fail

If we allowed using `when` expressions on finally tasks, a way to avoid failing a pipeline because of missing results could be to specify all required when expressions on the task, however that may not be trivial:
```
[Task A](guarded by X) -> [Task B] -> [Task C](produce result Z), [Finally Task D](consumes result Z)
```

With the current implementation, if A is skipped, B and C will not run, D will fail and the whole pipeline will fail.
The alternative for the pipeline author would be to guard D with X, but that does not feel to me like a good user experience.

If there is general agreement on the current approach, I'm ok with it.
In any case we should clearly document the behaviour in our docs, so that pipeline authors may know what to expect.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -521,7 +521,20 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
...
+	// Before creating TaskRun for scheduled final task, check if it's consuming a task result
+	// Resolve and apply task result wherever applicable, report failure in case resolution fails
+	for _, rprt := range finallyRprts {
+		resolvedResultRefs, err := resources.ResolveResultRefs(pipelineRunFacts.State, resources.PipelineRunState{rprt})
+		if err != nil {
+			rprt.InvalidTaskResultsInFinally = err.Error()
+			logger.Warnf("Declaring final task %q failed as it failed to resolve task params for %q with error %v", rprt.PipelineTask.Name, pr.Name, err)
+			continue
+		}
+		resources.ApplyTaskResults(resources.PipelineRunState{rprt}, resolvedResultRefs)
```
afrittoli, line 535: It feels a bit un-natural to use two functions that work on slices, passing them elements of a slice one by one embedded in 1-element slices.
I wonder if it would make sense to make a version of those functions that work on single elements, and use that both here as well as for the implementation of the current functions that work on slices.

afrittoli: Not something that needs to be fixed in the PR, it's also fine to change it as a follow-up if you agree.

pritidesai:
> From a pipeline run status point of view, we're now hosting in the pipeline run status a reference to a taskrun that was never created (at least that's my understanding). We're already doing that for Conditions and it's really confusing and hard to parse for both humans and UIs (dashboard / tkn). I'd hope we would do less of that and not more.

Without creating such object in `PipelineRun` status, how could we inform the pipeline author what happened to that final task without accessing controller logs? Why that final was not executed?

bobcatfish:
Thanks for raising this @afrittoli - I had agreed with the initial proposal but I'm not sure I fully realized this meant there would be a failed TaskRun created as well and I also hope we can find some different way of doing this.

> Without creating such object in PipelineRun status, how could we inform the pipeline author what happened to that final task without accessing controller logs? Why that final was not executed?

Some ideas: 
1. In slack @jerop mentioned maybe considering the Task "skipped" and including it in "skipped tasks".
2. Another possibility might be to have some new section specifically for finally tasks that don't execute.
3. Another possible approach would be go the opposite direction from what Andrea is suggested, e.g. embrace having a 'placeholder' taskrun in the status, e.g. when a pipelinerun starts, start building up 'placeholder' taskruns in the status for _all_ task
4. We could revisit the shape of the PipelineRun status entirely and how TaskRuns are represented - imo we haven't paid as much attention to status design as we have to spec and this has all kind of built up ad-hoc

I'm kind of inclined toward (4) (pre-GA), with maybe something like (1) in the meantime to unblock this work?

pritidesai:
> I'm kind of inclined toward (4) (pre-GA), with maybe something like (1) in the meantime to unblock this work?

`finally` task must execute and can not be skipped based on the original contract. The best way to handle this would be continue execution as is with default type value i.e. empty string in case of a string type and empty array in case of array type. Leave it up to pipeline author to decide how he wants to deal with such zero values. Make pipeline author aware of this kind of functionality by documenting it.

Going with (1) in meanwhile but we will get into backward compatibility cycle. And lots of use cases are eyeing on this feature so would like to get in ASAP.

bobcatfish:
> finally task must execute and can not be skipped based on the original contract.

Apologies for the frustration but I think it might be worth revisiting this contract :(

> The best way to handle this would be continue execution as is with default type value i.e. empty string in case of a string type and empty array in case of array type. Leave it up to pipeline author to decide how he wants to deal with such zero values. Make pipeline author aware of this kind of functionality by documenting it.

We discussed this option a fair amount iirc - doing this would mean changing a different "contract" i.e. what it means to express that a Task needs the result of another Task: at the moment the contract is that if the Task runs, it will be given that result from the previously executed Task.

Either way we are going to be changing a contract.

Maybe we can discuss more in the API working group on monday?

bobcatfish:
>  We're already doing that for Conditions and it's really confusing and hard to parse for both humans and UIs (dashboard / tkn).

one counter point @afrittoli : if the UI did show this finally task as "failed" when what happened was it really 'failed to execute b/c the result wasn't available' that's not quite as bad as what we see with conditions b/c in this case, we already have a failed pipelinerun. the conditions case was especially confusing b/c the condition task "failed" but the pipelinerun did not fail

pritidesai:
> Apologies for the frustration but I think it might be worth revisiting this contract :(

@bobcatfish Please don't apologize, no frustration at all, its part of the development cycle 😉  Its better to find the solution with everyone's agreement rather than building something which no one would like to use 😆 

> Either way we are going to be changing a contract.

🤣  🤣  🤣  

Yup, API WG it is.

pritidesai:
The way its implemented in this PR, a sample pipelinerun with two final tasks, `final-task-1` and `final-task-2` where `final-task-2` has a reference to task result of a `dag` task and that `dag` task failed:

```
tkn pr describe
Name:              write-and-cleanup-workspace
Namespace:         default
Pipeline Ref:      clone-cleanup-workspace
Service Account:   default
Timeout:           1h0m0s
Labels:
 tekton.dev/pipeline=clone-cleanup-workspace

🌡️  Status

STARTED         DURATION     STATUS
6 minutes ago   17 seconds   Failed

💌 Message

Tasks Completed: 3 (Failed: 2, Cancelled 0), Skipped: 0 (Failed to resolve task params for finally Task final-task-2 in PipelineRun write-and-cleanup-workspace: unable to find result referenced by param "commit" in "final-task-2": could not find successful taskrun for task "clone-app-repo")

📦 Resources

 No resources

⚓ Params

 No params

📝 Results

 No results

📂 Workspaces

 NAME           SUB PATH   WORKSPACE BINDING
 ∙ git-source   ---        VolumeClaimTemplate

🗂  Taskruns

 NAME                                                 TASK NAME        STARTED         DURATION     STATUS
 ∙ write-and-cleanup-workspace-final-task-2-fswll     final-task-2     ---             ---          Failed(InvalidTaskResultsInFinallyTask)
 ∙ write-and-cleanup-workspace-final-task-1-79mlt     final-task-1     6 minutes ago   11 seconds   Succeeded
 ∙ write-and-cleanup-workspace-clone-app-repo-qp2w5   clone-app-repo   6 minutes ago   5 seconds    Failed
```

Screen Shot 2020-10-08 at 1 46 05 PM

pritidesai:
@bobcatfish @afrittoli @jerop @sbwsg 

As I am thinking more and more about the use cases we are trying support with this feature, I am not able to think beyond `cleanup` (the one and only use case specified in the [TEP](https://github.com/tektoncd/community/blob/master/teps/0004-task-results-in-final-tasks.md#motivation)):

UC1:

The resources are acquired by the setup `Task` which has the names of resources. The clean up `Tasks`
must clean up only those resources. Also, cleanup must be done only if the setup was successful but regardless of the failure of the any other `Tasks` in the `Pipeline`.

```none
setup
  |
  | e.g. Project Name
  v
cleanup
```

Similar [use case](https://github.com/tektoncd/pipeline/issues/2557#issuecomment-692078619) from @spikeburton.

For this use case, if resource acquisition fail, ideally cleanup should be skipped. Cleanup can not be declared failure since the cleanup itself didn't fail. To address this use case, the best alternative here would be to `skip` that `cleanup` task with a reason. Thoughts?

The notification use case we have been discussing does not apply to this feature in practice because when a task fails, task result of that failed task is not available to the consumer. And to access the execution status of build task, we have another feature [coming soon](https://github.com/tektoncd/pipeline/pull/3390).

@chhsia0 what is the expectation with notification use case in terms of communicating the failure logs?

sbwsg:
Hm, resource cleanup and notifications / signalling are the two use cases I'm familiar with.

jerop:
@pritidesai i've also seen cleanup and notifications use cases only, maybe we can do a survey (similar to what @sbwsg did for credentials [here](https://groups.google.com/g/tekton-dev/c/n_-NSPRUcm0/m/-IRKv0ZmBgAJ)) to ask users about their use cases for finally tasks (with results and/or when expressions)

vincent-pli:
@pritidesai 
I think i can provide another use case, but not definitely.

It's about `Global exception handler` (global in one `pipeline`).
Our user want to handle the exception raised in `tasks` in a `pipeline`, the `task` will not failed when exception raised, we plan to emit the exception to `result`. and in the `final` task, we want a summary of the running situation to do some next job. the reason adopt `final` task rather than normal task stand on the end is because there maybe some unexpected error will make the `task` failed, `final` one will always run. 

So we expect the `result` could be inject to `final` task,  is that a case?

pritidesai:
thanks @vincent-pli yes I think this use case fits well with the requirements i.e. as long as the `exception` is communicated through `results` and the `task` producing `exception` does not fail, `finally` task will always run.

The [proposal](https://docs.google.com/document/d/1tV1LgPOINnmlDV-oSNdLB39IlLcQRGaYAxYZjVwVWcs/edit?ts=5fa3127a#heading=h.h3uuz4svzvgs) is almost final now and I am in process of updating [TEP](https://github.com/tektoncd/community/blob/master/teps/0004-task-results-in-final-tasks.md) followed by this PR 🏃‍♀️ 

Thank you everyone for your patience 🙏

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: I have introduced a new function to handle single task for now, `taskResults` is due for refactoring. After @sbwsg's pipelineRun results PR #3472 is merged, I will take up refactoring `taskResults`. Changes in PR #3472 have simplified resolving pipelineRun results and cleaned up `resultrefresolution.go`.

pritidesai:
I have again updated this PR based on the discussion in the [TEP](https://github.com/tektoncd/community/pull/254#issuecomment-728332142), PTAL 🙏

pritidesai:
This is ready for review, PTAL 🙏

vdemeester (approved):
/meow

pritidesai:
thanks @sbwsg  and @vdemeester for approvals 🎉 

looking for one `lgtm`, @afrittoli, @bobcatfish, @jerop, if you have bandwidth? 🙏

pkg/reconciler/pipelinerun/pipelinerun_test.go
bobcatfish: pinging this again

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3999,6 +3999,138 @@ func TestReconcilePipeline_TaskSpecMetadata(t *testing.T) {
...
+	if len(actual.Items) != 1 {
+		t.Fatalf("Expected 1 TaskRuns got %d", len(actual.Items))
+	}
+	actualTaskRun := actual.Items[0]
+	if d := cmp.Diff(&actualTaskRun, expectedTaskRun, ignoreResourceVersion); d != "" {
+		t.Errorf("expected to see TaskRun %v created. Diff %s", expectedTaskRunName, diff.PrintWantGot(d))
+	}
+
+	finalTaskSkipped := false
+	for _, s := range reconciledRun.Status.SkippedTasks {
```
bobcatfish, line 4124: would it make sense to also make sure the overall status of the reconciled run is failure in this case? unless that's something you want to cover in an e2e test instead

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3999,6 +3999,138 @@ func TestReconcilePipeline_TaskSpecMetadata(t *testing.T) {
...
+		if s.Name == "final-task-2" {
+			finalTaskSkipped = true
+		}
+	}
```
bobcatfish, line 4128: should we verify that this is also the ONLY task that is skipped?

pkg/reconciler/pipeline/dag/dag.go
```diff
@@ -66,16 +66,29 @@ func (g *Graph) addPipelineTask(t Task) (*Node, error) {
 	return newNode, nil
 }
 
-// Build returns a valid pipeline Graph. Returns error if the pipeline is invalid
-func Build(tasks Tasks) (*Graph, error) {
+// BuildWithoutLinks builds a Pipeline Graph with the specified tasks as nodes (without any links).
+// Returns error if the pipeline is invalid
```
bobcatfish, line 70: I don't understand what BuildWithoutLinks is doing, could you add more detail, e.g. explain what "links" means in this context? I think it means that the graph is created without any edges, in which case it's not a graph (i mean technically it is but not in a useful way) - is the idea to get a list of nodes? would it make sense to create a different data structure or maybe just return a list of []Node?

bobcatfish (commented):
Hey @pritidesai ! I've looked at the end to end and reconciler tests and I started to look at the code (just BuildWithoutLinks so far).

Here is my feedback so far:

* what do you think about adding some use of results to the existing e2e test(s) as well? https://github.com/tektoncd/pipeline/blob/master/test/pipelinefinally_test.go
* I'm also thinking that based on some of the bugs we ran into with when expressions, it might make sense to create a more complex example, e.g. maybe including a when expression somewhere in the finally example as well (any other features we might want to include?)
* Similar to the previous point, it seems like most of the tests are using a case where a Task producing a result fails - can we include some where it is skipped as well? In this case, the overall PipelineRun should still succeed (if im understanding correctly!) but the skipped task and the final task that needs the result should both be skipped

docs/pipelines.md
```diff
@@ -759,6 +759,36 @@ spec:
...
+  finally:
+    - name: check-count
+      taskRef:
+        Name: check-count
+      params:
+        - name: before-count
+          value: $(tasks.count-comments-before.results.count)
+        - name: after-count
+          value: $(tasks.count-comments-after.results.count)
+```
```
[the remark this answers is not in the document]
pritidesai: I have replaced an example with more common task `git-clone` so that its easier to follow.

docs/pipelines.md
```diff
@@ -759,6 +759,36 @@ spec:
...
+```yaml
+spec:
+  tasks:
+    - name: count-comments-before
+      taskRef:
+        Name: count-comments
+    - name: add-comment
+      taskRef:
+        Name: add-comment
+    - name: count-comments-after
```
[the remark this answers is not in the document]
pritidesai: I have replaced an example with more common task `git-clone` so that its easier to follow.

pkg/reconciler/pipelinerun/pipelinerun_test.go
pritidesai: Reconciled run does not result in failure here, controller logs an error and continues creating taskRun for other final tasks. I have covered the overall pipelineRun status in `e2e` test.

pkg/reconciler/pipelinerun/pipelinerun_test.go
pritidesai: yup, good point 👍 done

pkg/reconciler/pipelinerun/pipelinerun_test.go
pritidesai: I have added two additional final tasks in `e2e` test [consuming](https://github.com/tektoncd/pipeline/pull/3242/files#diff-7815ee756d1326fe39c09f4deb5c7c36aa4a8f14b4319d5bf8ac73a1929a3d10R94) task result from a failed task and a skipped task (due to when expression). Overall status of this test `pipelineRun` is failure as it has at least one `dag` failed. Let me know if you would like to see one more test pipeline with just dag with when expression and a final task consuming result from that dag which would result in overall pipelineRun status to `completed`.

pkg/reconciler/pipeline/dag/dag.go
pritidesai: `BuildWithoutLinks` is creating graph without linking any dependencies. 

Task results introduces dependency between two tasks i.e. consumer task is dependent on producer task and from the `dag` perspective, both tasks are part of the same graph. But in case of `finally` consuming results from `dag`, both are part of a separate sections. And as per the `finally` contract, it can not depend on any `dag` task.

You are right practically its just a list of `Node`s, all finally tasks being root nodes of a graph.

We could convert it into a list instead of graph. This was originally designed to be a graph to support finally tasks being dependent on (order and resource dependency) other finally tasks.

I have renamed `BuildWithoutLinks` to `BuildWithoutDeps` if its easier to understand.

pritidesai: Another alternative is to have a build function taking both `dag` and `finally` and return two graphs:

```
func Build(dag Tasks, finally Tasks) (*Graph, *Graph, error) {
```

Build graph as is for `dag` tasks but while building graph for `finally`, avoid any dependencies if its a `dag` task.

Or

Build graph with all the `tasks` but avoid adding dependencies if it exists in a list of `noDeps`:

```
func Build(tasks Tasks, noDeps Tasks) (*Graph, error) {
```

Call it like:

```
d := Build(dagTasks, []PipelineTasks)
dfinally := Build(finallyTasks, dagTasks)
```

bobcatfish: Thanks for explaining!

I think it's important to use the right data structure for whatever we're doing - it makes the code much clearer and easier to maintain in the long run.

In this case, if a list of nodes is what we need, do we need to involve a graph object at all? Can we create a function that is not related to Graph at all and produces a list of Nodes from a list of Tasks?

e.g. 

```
func GetNodes(tasks Tasks) []Node
```

Alternatively: maybe the problem is that determining the deps shouldnt be part of building the graph? what if `deps` was passed into the build function?

pritidesai: > I think it's important to use the right data structure for whatever we're doing - it makes the code much clearer and easier to maintain in the long run.

yup, building `Graph` for `finally` section is essential to support `runAfter` and `workspaces` in future.

pritidesai: > Alternatively: maybe the problem is that determining the deps shouldnt be part of building the graph? what if deps was passed into the build function?

yup, working on this alternative.

pritidesai: Taking `deps` out of `build` does simplify `build`, see changes [here](https://github.com/tektoncd/pipeline/compare/master...pritidesai:build-deps?expand=1) 

And, irrespective of `finally` section being built as a `Graph` or a list of `Node[]`, it has to fetch `deps` to validate validate result reference belongs to a valid `dag` task. 

PTAL 🙏

bobcatfish: Nice! I wonder if it would help if there was a function you could call to get all the deps as a map?

e.g. something like

```

type PipelineTasks []PipelineTask
func (tasks PipelineTasks) Deps() map[string][] {
        deps := map[string][]string{}
	for _, pt := range tasks {
		deps[pt.HashKey()] = pt.Deps()
	}
        return deps
}
```

then the original line:

```
if _, err := dag.Build(PipelineTaskList(tasks)); err != nil {
```

would just become something like :

```
if _, err := dag.Build(PipelineTaskList(tasks),  tasks.Deps()); err != nil {
```

bobcatfish: (oh! `PipelineTaskList` already IS the type I defined above as `PipelineTasks` :D)

pkg/reconciler/pipelinerun/pipelinerun_test.go
pritidesai: 🤦‍♀️  shying away in this PR

pkg/reconciler/pipeline/dag/dag.go
pritidesai: Great suggestion, it looks much simpler now, PR # https://github.com/tektoncd/pipeline/pull/3583

pritidesai: > validate the result reference belonging to a valid dag task which is missing from the changes in this PR 😢

I have added validation [here](https://github.com/tektoncd/pipeline/pull/3242/files#diff-78ff878240189b20e0ef82f2912c93c0b3c7b445865e04fa9b98aab851be6130R343).

pritidesai:
This is rebased and ready for review 🙏

pritidesai:
> pinging this again
Gave it a one more push, have dropped using test builder in `pipelinerun_test.go`.

docs/pipelines.md
```diff
@@ -760,6 +760,33 @@ spec:
...
+          value: $(tasks.clone-app-repo.results.commit)
+```
+**Note:** The scheduling of such final task does not change, it will still be executed in parallel with other
+final tasks after all non-final tasks are done.
+
+The controller resolves task results before executing the finally task `discover-git-commit`. If the task
+`clone-app-repo` failed or skipped with [when expression](#guard-task-execution-using-whenexpressions) resulting in
+uninitialized task result `commit`, the finally Task `discover-git-commit` will be included in the list of
+`skippedTasks`. The controller logs the validation failure `unable to find result referenced by param "commit" in
+"discover-git-commit"` and continues executing rest of the final tasks. The pipeline exits with `completion` instead of
```
bobcatfish, line 786: note: i really dont think we should include anything about what the controller will likely log in these docs. if we start documenting that kind of thing, then users might come to rely on it.

imo the docs should cover the expected functionality that would be more or less protected by our backwards compatibility policy - if we wanted to change what the controller logs, i dont think we should need to change our docs.

i also think it's important that we distinguish between:
- how the features are implemented
- what the features do

users need to know about what the features do, and as much as possible shouldn't have to think about how the features are implemented (which also leaves us maintainers free to change the implementation around without impacting users)

test/yamls/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
```diff
@@ -164,6 +164,18 @@ spec:
...
+        - name: commit
+          value: $(tasks.clone-app-repo.results.commit)
+      taskSpec:
+        params:
+          - name: commit
+        steps:
+          - name: display-commit
+            image: alpine
+            script: |
+              echo $(params.commit)
```
bobcatfish, line 178: im confused, why does a yaml file with this name exist in 2 different places?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -4350,6 +4350,188 @@ func TestReconcilePipeline_TaskSpecMetadata(t *testing.T) {
...
+		t.Fatalf("Failure to list TaskRun's %s", err)
+	}
+	if len(actual.Items) != 1 {
+		t.Fatalf("Expected 1 TaskRuns got %d", len(actual.Items))
+	}
+	actualTaskRun := actual.Items[0]
+	if d := cmp.Diff(expectedTaskRun, actualTaskRun, ignoreResourceVersion); d != "" {
+		t.Errorf("expected to see TaskRun %v created. Diff %s", expectedTaskRunName, diff.PrintWantGot(d))
+	}
+
```
bobcatfish, line 4525: continuing on this conversation:

> me: would it make sense to also make sure the overall status of the reconciled run is failure in this case? unless that's something you want to cover in an e2e test instead

> @pritidesai: Reconciled run does not result in failure here, controller logs an error and continues creating taskRun for other final tasks. I have covered the overall pipelineRun status in e2e test.

I think I almost understand, but I think technically the situation is that the controller is waiting for the finally tasks to complete before updating the status to failed, is that right?

> controller logs an error and continues creating taskRun for other final tasks

this reconcile should result in all finally tasks "running" (since they run in parallel) and then the it would the the last reconcile after the finally tasks have finished that would set the overall status to failed, is that right?

test/pipelinefinally_test.go
```diff
@@ -55,25 +58,69 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_Failure(t *testing.T) {
...
+		},
+		"dagtask4": {
+			TaskName: successTask.Name,
+			When:     we,
+		},
+	}
+
+	f := map[string]pipelineTask{
+		"finaltask1": {
+			TaskName: successTask.Name,
```
bobcatfish, line 96: would it make sense to have a final task also successfully consuming a result, for completeness?

test/pipelinefinally_test.go
```diff
@@ -110,11 +163,15 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_Failure(t *testing.T) {
 			if !isSkipped(t, n, taskrunItem.Status.Conditions) {
 				t.Fatalf("dag task %s should have skipped due to condition failure", n)
 			}
+		case n == "dagtask4":
```
bobcatfish, line 166: nitpick but maybe it would make sense to also assert on the length of taskrunList? apologies since this test was already like this, but it seems like it would currently pass if extra taskruns were returned

test/pipelinefinally_test.go
```diff
@@ -123,6 +180,26 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_Failure(t *testing.T) {
...
+		Name: "finaltaskconsumingdagtask1",
+	}, {
+		Name: "finaltaskconsumingdagtask4",
+	}}
+
+	if d := cmp.Diff(pr.Status.SkippedTasks, expectedSkippedTasks); d != "" {
+		t.Fatalf("Expected four skipped tasks, dag task with condition failure dagtask3, dag task with when expression,"+
+			"two final tasks with missing result reference finaltaskconsumingdagtask1 and finaltaskconsumingdagtask4 in SkippedTasks."+
+			" Diff: %s", diff.PrintWantGot(d))
+	}
```
bobcatfish, line 202: thanks for updating this test!!

test/pipelinefinally_test.go
```diff
@@ -123,6 +180,26 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_Failure(t *testing.T) {
...
 }
 
 func TestPipelineLevelFinally_OneFinalTaskFailed_Failure(t *testing.T) {
```
bobcatfish, line 205: do we need to update more than one end to end test? i feel like the previous one was pretty comprehensive, especially if it also has a finally task successfully execute while consuming a result

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -589,7 +589,19 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
 	pipelineRunFacts.ResetSkippedCache()
 
 	// GetFinalTasks only returns tasks when a DAG is complete
-	nextRprts = append(nextRprts, pipelineRunFacts.GetFinalTasks()...)
```
bobcatfish, line 592: this is outside of the scope of this PR but i wonder if this function (`runNextSchedulableTask`) has too much responsibility. it seems to be in charge of:
- deciding what to execute next
- resolving references
- actually executing

those feel like at least 3 distinct responsibilities. i think if we get a chance it would be great to detangle this a bit.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -589,7 +589,19 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
...
+	finallyRprts := pipelineRunFacts.GetFinalTasks()
+
+	// Before creating TaskRun for scheduled final task, check if it's consuming a task result
+	// Resolve and apply task result wherever applicable, report warning in case resolution fails
+	for _, rprt := range finallyRprts {
+		resolvedResultRefs, err := resources.ResolveResultRef(pipelineRunFacts.State, rprt)
+		if err != nil {
+			logger.Warnf("Final task %q is not executed as it failed to resolve task params for %q with error %v", rprt.PipelineTask.Name, pr.Name, err)
```
bobcatfish, line 599: should this be a warning or an error? we only need to apply results when we're ready to run final tasks and we're only ready once all of the non-final tasks have finished.

pkg/reconciler/pipelinerun/pipelinerun.go
bobcatfish: not to mention that it doesnt have unit tests :(

pkg/reconciler/pipelinerun/pipelinerun.go
bobcatfish: im thinking that what is happening here is that we are logging this warning in the case where a finally task needs a result from a task that was skipped.

1. i dont think that should be a warning. logs that are warn, error or fatal should be indicating to the operator of the system that some kind of action needs to be taken. i.e. the _operator_ not the _user_. the user gets informed of actions they need to take via the status (e.g. error messages). the operator is the person admin-ing the tekton installation and needs to know if there is something wrong with the setup. so maybe info here? (https://dzone.com/articles/logging-levels-what-they-are-and-how-they-help-you seems like an okay article on log levels)
2. it would be great if one day we could get to a point where the decision of whether or not to skip was only done in one place - is there any way to do that in this PR? i.e. before we try to resolve the references? (i.e. why are we trying to resolve references on something that needs to be skipped) it seems like we should be able to look at a task's dependencies and see that it cant be run?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -204,6 +204,19 @@ func (t *ResolvedPipelineRunTask) parentTasksSkip(facts *PipelineRunFacts) bool
 	return false
 }
 
+// IsFinallySkipped returns true if a finally task is not executed and skipped due to task result validation failure
+func (t *ResolvedPipelineRunTask) IsFinallySkipped(facts *PipelineRunFacts) bool {
```
bobcatfish, line 208: im surprised we need special "is skipped" logic for a finally task - wouldnt the logic for whether or not a task is skipped be the same as for non-finally tasks? (i.e. it depends on a task that failed or was skipped?)

bobcatfish (commented):
I was able to get into the code this time :D very excited to see this progressing!

docs/pipelines.md
pritidesai: I can remove this log from here for now but without this log, if I am running into a finally task being part of `skippedTasks`, I have no way to look for `why` from doc perspective (without looking at the code).

test/yamls/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
pritidesai: Please refer to https://github.com/tektoncd/pipeline/pull/3564#issuecomment-736252906

pkg/reconciler/pipelinerun/pipelinerun_test.go
pritidesai: > I think I almost understand, but I think technically the situation is that the controller is waiting for the finally tasks to complete before updating the status to failed, is that right?

Yes the controller is waiting for the finally tasks to complete before update the status. The status condition need not be necessarily failure. If finally task has a reference to a skipped dag task, the controller sets the status to `completed` after all finally tasks are finished. The controller sets the status to `failed` if any of the finally or a dag task fails. Let me know if this is not clear.

test/pipelinefinally_test.go
pritidesai: haha I had added it and dropped it, sure will add it back.

test/pipelinefinally_test.go
pritidesai: sure I will add it explicitly, it is covered by the default:

https://github.com/tektoncd/pipeline/blob/705e97146fea0e6a7f55e3b06f46e317a9c12659/test/pipelinefinally_test.go#L118-L119

`default` would raise error if it finds any missing or extra taskRuns, I will add explicit count assertion.

pritidesai: will update this error message as well since its not clear.

test/pipelinefinally_test.go
pritidesai: nope, I think updating one is enough, I will add a finally task with a successful task result consumption in the first test itself.

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: 😭 it does

I will work on refactoring it after this is merged, created issue #3610

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: > should this be a warning or an error?

I think it should be warning, its indicating that the finally task was skipped for some reason. Skipping a dag does not warn users but here users are warned when a finally task is skipped. Finally being finally task must be executed and warn if they are not for some reason.

But based on your description, there is no action needs to be taken if a finally task is skipped in case of a cleanup/notification use case. may be convert it to info then? I am not convinced but at the same time I dont see a use case.


> im thinking that what is happening here is that we are logging this warning in the case where a finally task needs a result from a task that was skipped.

yup skipped or failed.

> we only need to apply results when we're ready to run final tasks and we're only ready once all of the non-final tasks have finished.

Until we are ready to run finally tasks, `finallyRprts` is empty.

> it would be great if one day we could get to a point where the decision of whether or not to skip was only done in one place - is there any way to do that in this PR?

The only way we could do it to maintain a list of [skippedTasks](https://github.com/tektoncd/pipeline/compare/master...pritidesai:pipelinerunstatus#diff-188e8bbe2f5c7cb0b9de03be02a957537e46bb71add977e4992533375ef08998R351) in `pipelineRunFacts`. Once we determine that a finally task is being skipped, add it into the list. And check this list while updating the status condition and status.SkippedTasks. The same logic would work for dag tasks as well. We could merge this as is and refactor after this is merged.

(@afrittoli's attempt on simplifying [skipped](https://github.com/tektoncd/pipeline/commit/72455ba0a81c9d806b26809842282c045ca43d41#diff-2d6877ada24d6e8ae6d232d24799f21c520e590e5b432f9adba5e2e62eaa40b7R53) tasks)

> why are we trying to resolve references on something that needs to be skipped

At this point, we are discovering task result resolution failure and therefore not executing finally task. Finally task consuming task results from a dag task but not dependent on a dag task. Both belong to separate sections, no relationships established from a finally to dag.

pritidesai: This could be further simplified for both finally and dag, instead of trying to resolve task results, just check if the referenced task was failed or skipped before trying to resolve results.

pritidesai:
Hey @bobcatfish I think I have responded to most of the comments. I have tried to explain both of these:

(1) Need for a separate `isFinallySkipped` function and possible refactoring. Can be avoided by introducing a list of `skippedTasks` as part of `pipelineRunFacts`. 
(2) Checking if a finally task must be skipped in one single i.e. instead of resolving a task result, check the reference task status, if failed or skipped, skip finally

Let me know if you would want to see any more changes in this PR including any refactoring. I will be happy to change this PR or create separate PR if possible.

test/pipelinefinally_test.go
pritidesai: done

docs/pipelines.md
pritidesai: deleted log from doc.

test/pipelinefinally_test.go
pritidesai: done

test/pipelinefinally_test.go
pritidesai: https://github.com/tektoncd/pipeline/pull/3242/files#diff-7815ee756d1326fe39c09f4deb5c7c36aa4a8f14b4319d5bf8ac73a1929a3d10R168

test/pipelinefinally_test.go
pritidesai: verifying the param value in a finally task (referring to result of a dag task) https://github.com/tektoncd/pipeline/pull/3242/files#diff-7815ee756d1326fe39c09f4deb5c7c36aa4a8f14b4319d5bf8ac73a1929a3d10R210

pritidesai:
one more ping 🛎
🙏🏻 Would like to get this in if possible 🙏🏻

pritidesai:
> /lgtm cancel
> /hold

Hey @vdemeester we discussed this in the planning meeting this morning. I think we put this PR on hold so that it can be redesigned (skip instead of creating an object reference for a finally task referring to missing task results). Can we cancel the hold  since the design is updated?

We have moved this PR to next milestone 0.21 for now.

vdemeester:
> > /lgtm cancel
> > /hold
> 
> Hey @vdemeester we discussed this in the planning meeting this morning. I think we put this PR on hold so that it can be redesigned (skip instead of creating an object reference for a finally task referring to missing task results). Can we cancel the hold since the design is updated?
> 
> We have moved this PR to next milestone 0.21 for now.

👍🏼 

/hold cancel

vdemeester:
@pritidesai needs a rebase 🙏🏼

pritidesai:
> @pritidesai needs a rebase 🙏🏼

sorry @vdemeester for delay, done 👍

PR #3217, opened 2020-09-12
fix for when expressions issues

pritidesai:
# Changes

1. Instead of creating a new empty array and appending resolved when expression to it, creating an array based on the existing one and replacing in place.

2. With when expressions having reference to task results, evaluating and applying task results before checking if it's time to run finally tasks. This change is necessary to make sure all the leaf nodes of DAG have resolved task results in their when expressions to check whether DAG is done or not. This change of order works today since finally tasks does not need to resolve task results yet. Once we add task results support in finally, we will have to make one more call to apply task results after adding finally tasks to the execution queue.

3. Do not evaluate when expressions until all parents are done, wait for parent tasks to be completed before evaluating when expressions. This check will make sure that tasks resulting in when expressions failures, are not declared skipped in advance i.e. before their parents (declared with `runAfter`) are started/done.

Fixes #3196
Fixes #3203

The following pipelines are fixed with these changes:

[when-expression-params-with-finally.yaml](https://gist.github.com/pritidesai/605f45336997d5a9ad37586d6558b1cb)
[when-expression-task-results-with-finally.yaml](https://gist.github.com/pritidesai/5d496730976d30e4f4f523d60aba7d5e)
[when-expression-evaluating-to-false-with-runafter.yaml](https://gist.github.com/pritidesai/401e420ff93746adb0e3147c614c45e9)

/kind bug 

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [x] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

Double check this list of stuff that's easy to miss:

- If you are adding [a new binary/image to the `cmd` dir](../cmd), please update
  [the release Task](../tekton/publish.yaml) to build and release this image.

## Reviewer Notes

[...]

PR #3244, opened 2020-09-16
Apply replacements in tasks and finally tasks separately

jerop:
# Changes

Previously, we created a slice containing both tasks and finally tasks then applied replacements across all of them. However, the variable replacements we make in conditions and when expressions in tasks don't apply to finally tasks. Moreover, this joined slice caused strange memory allocation side effects when finally tasks were present vs when they were missing.

We explored reverting the mutation of when expressions when applying replacements, as discussed in https://github.com/tektoncd/pipeline/pull/3217#discussion_r488123753 but decided to remove that because it introduces inconsistencies as discussed in https://github.com/tektoncd/pipeline/pull/3244#discussion_r489538714 and https://github.com/tektoncd/pipeline/pull/3244#issuecomment-693698922.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```release-note
    NONE
```

/cc @pritidesai @bobcatfish
/kind cleanup

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/apply.go
+++ b/pkg/reconciler/pipelinerun/resources/apply.go
@@ -88,16 +88,17 @@ func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResul
 func ApplyReplacements(p *v1beta1.PipelineSpec, replacements map[string]string, arrayReplacements map[string][]string) *v1beta1.PipelineSpec {
 	p = p.DeepCopy()
 
-	// replace param values for both DAG and final tasks
-	tasks := append(p.Tasks, p.Finally...)
-
-	for i := range tasks {
-		tasks[i].Params = replaceParamValues(tasks[i].Params, replacements, arrayReplacements)
-		for j := range tasks[i].Conditions {
-			c := tasks[i].Conditions[j]
+	for i := range p.Tasks {
+		p.Tasks[i].Params = replaceParamValues(p.Tasks[i].Params, replacements, arrayReplacements)
+		for j := range p.Tasks[i].Conditions {
+			c := p.Tasks[i].Conditions[j]
 			c.Params = replaceParamValues(c.Params, replacements, arrayReplacements)
 		}
-		tasks[i].WhenExpressions = tasks[i].WhenExpressions.ReplaceWhenExpressionsVariables(replacements)
+		p.Tasks[i].WhenExpressions = p.Tasks[i].WhenExpressions.ReplaceWhenExpressionsVariables(replacements)
+	}
+
+	for i := range p.Finally {
+		p.Finally[i].Params = replaceParamValues(p.Finally[i].Params, replacements, arrayReplacements)
 	}
 
 	return p
--- a/pkg/reconciler/pipelinerun/resources/apply_test.go
+++ b/pkg/reconciler/pipelinerun/resources/apply_test.go
@@ -174,6 +174,35 @@ func TestApplyParameters(t *testing.T) {
 					tb.PipelineTaskParam("final-task-first-param", "default-value"),
 					tb.PipelineTaskParam("final-task-second-param", "second-value"),
 				))),
+	}, {
+		name: "parameter evaluation with both tasks and final tasks",
+		original: tb.Pipeline("test-pipeline",
+			tb.PipelineSpec(
+				tb.PipelineParamSpec("first-param", v1beta1.ParamTypeString, tb.ParamSpecDefault("default-value")),
+				tb.PipelineParamSpec("second-param", v1beta1.ParamTypeString),
+				tb.PipelineTask("first-task-1", "first-task",
+					tb.PipelineTaskParam("first-task-first-param", "$(params.first-param)"),
+				),
+				tb.PipelineTask("first-task-2", "first-task",
+					tb.PipelineTaskWhenExpression("$(params.first-param)", selection.In, []string{"$(params.second-param)"})),
+				tb.FinalPipelineTask("final-task-1", "final-task",
+					tb.PipelineTaskParam("final-task-second-param", "$(params.second-param)"),
+				))),
+		run: tb.PipelineRun("test-pipeline-run",
+			tb.PipelineRunSpec("test-pipeline",
+				tb.PipelineRunParam("second-param", "second-value"))),
+		expected: tb.Pipeline("test-pipeline",
+			tb.PipelineSpec(
+				tb.PipelineParamSpec("first-param", v1beta1.ParamTypeString, tb.ParamSpecDefault("default-value")),
+				tb.PipelineParamSpec("second-param", v1beta1.ParamTypeString),
+				tb.PipelineTask("first-task-1", "first-task",
+					tb.PipelineTaskParam("first-task-first-param", "default-value"),
+				),
+				tb.PipelineTask("first-task-2", "first-task",
+					tb.PipelineTaskWhenExpression("default-value", selection.In, []string{"second-value"})),
+				tb.FinalPipelineTask("final-task-1", "final-task",
+					tb.PipelineTaskParam("final-task-second-param", "second-value"),
+				))),
 	}}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {


pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -105,11 +105,11 @@ func (wes WhenExpressions) HaveVariables() bool {
 }
 
 // ReplaceWhenExpressionsVariables interpolates variables, such as Parameters and Results, in
-// the Input and Values.
+// the Input and Values. Returns a new instance of WhenExpressions.
 func (wes WhenExpressions) ReplaceWhenExpressionsVariables(replacements map[string]string) WhenExpressions {
-	replaced := wes
-	for i := range wes {
-		replaced[i] = wes[i].applyReplacements(replacements)
+	var replaced []WhenExpression
```
pritidesai, line 110: @jerop we might run into that kind of memory issue again in future if we decide to introduce `when` expressions in finally tasks.

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -88,16 +88,17 @@ func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResul
...
+		p.Tasks[i].Params = replaceParamValues(p.Tasks[i].Params, replacements, arrayReplacements)
+		for j := range p.Tasks[i].Conditions {
+			c := p.Tasks[i].Conditions[j]
 			c.Params = replaceParamValues(c.Params, replacements, arrayReplacements)
 		}
-		tasks[i].WhenExpressions = tasks[i].WhenExpressions.ReplaceWhenExpressionsVariables(replacements)
+		p.Tasks[i].WhenExpressions = p.Tasks[i].WhenExpressions.ReplaceWhenExpressionsVariables(replacements)
+	}
+
+	for i := range p.Finally {
```
pritidesai, line 100: Great to separate this out 👍

pkg/apis/pipeline/v1beta1/when_types.go
pritidesai: @bobcatfish This does not fix the underlying issue. Notice, the parameter resolution is successful. `$(params.first-param)` and `$(params.second-param)` are replaced with their respective values in `tasks` but its not reflected in `p`. By separating it out (which is great, I do not object) we are just avoiding calling `ReplaceWhenExpressionsVariables` when `finally` exists. 

```
go test ./pkg/reconciler/pipelinerun/resources/... -tags=unit -run=TestApplyParameters
(string) (len=6) "tasks:"
([]v1beta1.PipelineTask) (len=2 cap=2) {
 (v1beta1.PipelineTask) {
  Name: (string) (len=12) "first-task-1",
  TaskRef: (*v1beta1.TaskRef)(0xc0003ca8d0)({
   Name: (string) (len=10) "first-task",
   Kind: (v1beta1.TaskKind) "",
   APIVersion: (string) ""
  }),
  TaskSpec: (*v1beta1.EmbeddedTask)(<nil>),
  Conditions: ([]v1beta1.PipelineTaskCondition) <nil>,
  WhenExpressions: (v1beta1.WhenExpressions) (len=1 cap=1) {
   (v1beta1.WhenExpression) {
    Input: (string) (len=13) "default-value",
    Operator: (selection.Operator) (len=2) "in",
    Values: ([]string) (len=1 cap=1) {
     (string) (len=12) "second-value"
    }
   }
  },
  Retries: (int) 0,
  RunAfter: ([]string) <nil>,
  Resources: (*v1beta1.PipelineTaskResources)(<nil>),
  Params: ([]v1beta1.Param) <nil>,
  Workspaces: ([]v1beta1.WorkspacePipelineTaskBinding) <nil>,
  Timeout: (*v1.Duration)(<nil>)
 },
 (v1beta1.PipelineTask) {
  Name: (string) (len=12) "final-task-1",
  TaskRef: (*v1beta1.TaskRef)(0xc0003ca900)({
   Name: (string) (len=10) "first-task",
   Kind: (v1beta1.TaskKind) "",
   APIVersion: (string) ""
  }),
  TaskSpec: (*v1beta1.EmbeddedTask)(<nil>),
  Conditions: ([]v1beta1.PipelineTaskCondition) <nil>,
  WhenExpressions: (v1beta1.WhenExpressions) <nil>,
  Retries: (int) 0,
  RunAfter: ([]string) <nil>,
  Resources: (*v1beta1.PipelineTaskResources)(<nil>),
  Params: ([]v1beta1.Param) <nil>,
  Workspaces: ([]v1beta1.WorkspacePipelineTaskBinding) <nil>,
  Timeout: (*v1.Duration)(<nil>)
 }
}
(string) (len=2) "p:"
(*v1beta1.PipelineSpec)(0xc0003b8780)({
 Description: (string) "",
 Resources: ([]v1beta1.PipelineDeclaredResource) <nil>,
 Tasks: ([]v1beta1.PipelineTask) (len=1 cap=1) {
  (v1beta1.PipelineTask) {
   Name: (string) (len=12) "first-task-1",
   TaskRef: (*v1beta1.TaskRef)(0xc0003ca8d0)({
    Name: (string) (len=10) "first-task",
    Kind: (v1beta1.TaskKind) "",
    APIVersion: (string) ""
   }),
   TaskSpec: (*v1beta1.EmbeddedTask)(<nil>),
   Conditions: ([]v1beta1.PipelineTaskCondition) <nil>,
   WhenExpressions: (v1beta1.WhenExpressions) (len=1 cap=1) {
    (v1beta1.WhenExpression) {
     Input: (string) (len=21) "$(params.first-param)",
     Operator: (selection.Operator) (len=2) "in",
     Values: ([]string) (len=1 cap=1) {
      (string) (len=22) "$(params.second-param)"
     }
    }
   },
   Retries: (int) 0,
   RunAfter: ([]string) <nil>,
   Resources: (*v1beta1.PipelineTaskResources)(<nil>),
   Params: ([]v1beta1.Param) <nil>,
   Workspaces: ([]v1beta1.WorkspacePipelineTaskBinding) <nil>,
   Timeout: (*v1.Duration)(<nil>)
  }
 },
 Params: ([]v1beta1.ParamSpec) (len=2 cap=2) {
  (v1beta1.ParamSpec) {
   Name: (string) (len=11) "first-param",
   Type: (v1beta1.ParamType) (len=6) "string",
   Description: (string) "",
   Default: (*v1beta1.ArrayOrString)(0xc0003e8140)({
    Type: (v1beta1.ParamType) (len=6) "string",
    StringVal: (string) (len=13) "default-value",
    ArrayVal: ([]string) <nil>
   })
  },
  (v1beta1.ParamSpec) {
   Name: (string) (len=12) "second-param",
   Type: (v1beta1.ParamType) (len=6) "string",
   Description: (string) "",
   Default: (*v1beta1.ArrayOrString)(<nil>)
  }
 },
 Workspaces: ([]v1beta1.PipelineWorkspaceDeclaration) <nil>,
 Results: ([]v1beta1.PipelineResult) <nil>,
 Finally: ([]v1beta1.PipelineTask) (len=1 cap=1) {
  (v1beta1.PipelineTask) {
   Name: (string) (len=12) "final-task-1",
   TaskRef: (*v1beta1.TaskRef)(0xc0003ca900)({
    Name: (string) (len=10) "first-task",
    Kind: (v1beta1.TaskKind) "",
    APIVersion: (string) ""
   }),
   TaskSpec: (*v1beta1.EmbeddedTask)(<nil>),
   Conditions: ([]v1beta1.PipelineTaskCondition) <nil>,
   WhenExpressions: (v1beta1.WhenExpressions) <nil>,
   Retries: (int) 0,
   RunAfter: ([]string) <nil>,
   Resources: (*v1beta1.PipelineTaskResources)(<nil>),
   Params: ([]v1beta1.Param) <nil>,
   Workspaces: ([]v1beta1.WorkspacePipelineTaskBinding) <nil>,
   Timeout: (*v1.Duration)(<nil>)
  }
 }
})
--- FAIL: TestApplyParameters (0.00s)
    --- FAIL: TestApplyParameters/single_parameter_with_when_expression (0.00s)
        apply_test.go:182: ApplyParameters() got diff (-want, +got):   &v1beta1.PipelineSpec{
              	Description: "",
              	Resources:   nil,
              	Tasks: []v1beta1.PipelineTask{
              		{
              			... // 2 identical fields
              			TaskSpec:   nil,
              			Conditions: nil,
              			WhenExpressions: v1beta1.WhenExpressions{
              				{
            - 					Input:    "default-value",
            + 					Input:    "$(params.first-param)",
              					Operator: "in",
            - 					Values:   []string{"second-value"},
            + 					Values:   []string{"$(params.second-param)"},
              				},
              			},
              			Retries:  0,
              			RunAfter: nil,
              			... // 4 identical fields
              		},
              	},
              	Params:     {{Name: "first-param", Type: "string", Default: &{Type: "string", StringVal: "default-value"}}, {Name: "second-param", Type: "string"}},
              	Workspaces: nil,
              	... // 2 identical fields
              }
FAIL
FAIL	github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun/resources	1.599s
FAIL
```

pritidesai: Introducing `append` here might take us into inconsistent behavior in future 😞. 

We have been replacing `params` in place (may be change that too 🤔): 

https://github.com/tektoncd/pipeline/blob/81802739f5356f927c4f976a7a3f000fee95b773/pkg/reconciler/pipelinerun/resources/apply.go#L106-L110

bobcatfish: @pritidesai you were right, using append like this had some unexpected side effects!

I spent a bunch of time trying to understand why the mutations were consistently succeeding, but returning a new when expressions list wasn't and I think I finally understand it.

I created this example to illustrate it: https://play.golang.org/p/Fdnf84zT4L-

I'm not 100% what the best solution is to support when expressions in finally tasks, but i also think we can cover that when we get there.

I think what ended up biting us here is that ApplyReplacements is trying to mix mutating and avoiding mutation (which was my recommendation! apologies @jerop !)

```
p.Tasks[i].WhenExpressions = p.Tasks[i].WhenExpressions.ReplaceWhenExpressionsVariables(replacements)
```

The above statement is subtly trying to do both, b/c it's trying to mutate `p.Tasks[i].WhenExpressions` but unfortunately it's in some cases mutating a copy and in some cases mutating the original

bobcatfish:
I am still mostly convinced that a "functional" side effect free approach is the long term way to go, however trying to mix the approaches here has definitely caused some problems.

pkg/apis/pipeline/v1beta1/when_types.go
bobcatfish: I was thinking of this more overnight and I think the other kinda "root cause" here is that when we were using append(), we were then iterating over the resulting slice and mutating it, as if it would mutate the copy "p", but it was almost a coincidence that it did, because we're using structs by value. it just so happened we could mutate the values in "p" because it happened that the copy semantics for slices copied the pointer and still pointed at the slices in "p"

/me gets off soapbox

bobcatfish:
In the long run I think we should decide whether to always mutate or always return brand new instances and be consistent about it; in the short term I think it's good to remove the append() from this function since the differences in behavior between when it makes a new slice and when it doesn't are pretty extreme so I think we should go ahead with this PR:

/lgtm

Want to hear @pritidesai 's thoughts too tho before we merge!

pkg/apis/pipeline/v1beta1/when_types.go
jerop: thank you @pritidesai and @bobcatfish!

undid the revert of the mutation and will leave that to be something to be explored in the future

so this pr only separates the replacements in tasks and in finally tasks

pritidesai:
this looks great, yes, consistency is must, let's evaluate this further in a separate PR while refactoring `params`. 
/approve

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -88,16 +88,17 @@ func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResul
...
-		for j := range tasks[i].Conditions {
-			c := tasks[i].Conditions[j]
+	for i := range p.Tasks {
+		p.Tasks[i].Params = replaceParamValues(p.Tasks[i].Params, replacements, arrayReplacements)
+		for j := range p.Tasks[i].Conditions {
+			c := p.Tasks[i].Conditions[j]
 			c.Params = replaceParamValues(c.Params, replacements, arrayReplacements)
 		}
-		tasks[i].WhenExpressions = tasks[i].WhenExpressions.ReplaceWhenExpressionsVariables(replacements)
+		p.Tasks[i].WhenExpressions = p.Tasks[i].WhenExpressions.ReplaceWhenExpressionsVariables(replacements)
```
bobcatfish, line 97: i think this is misleading b/c since we're mutating now, we don't need to assign anything back: however it's consistent with what we're doing with params so i agree with your choice to go this route for now 👍

bobcatfish (commented):
/lgtm

bobcatfish:
TestTaskRunTimeout is flakey maybe? 🤔 

PR #3256, opened 2020-09-18
Sort pod container statuses based on Step order in taskSpec

Peaorl:
# Changes

This commit closes tektoncd#3239

Tekton determines the `TaskRun` status message of a failed `TaskRun` based on the results of the first terminated `Step` (pod container). Until now, Tekton sorted pod container statuses based on the _FinishedAt_ and _StartedAt_ timestamps set by Kubernetes. 
Occasionally, a `Step` terminated in response to the first terminated `Step` could have the same timestamps as the first terminated Step. 
Therefore, Tekton was not always able to correctly determine what the first terminated `Step` was, and as a result, Tekton may set an incorrect `TaskRun` status message.

In this commit, pod container statuses are sorted based on the container order as specified by Tekton in the podSpec. 
Tekton bases this order on the user provided taskSpec and `Steps` added internally by Tekton. Therefore, Tekton accounts for internally added `Steps` when sorting pod container statuses.
# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```release-note
Steps in the TaskRun status field are now sorted according to the Step order specified in the taskSpec
```


vdemeester:
/kind bug
/ok-to-test
@Peaorl this might need a release note (other than `NONE`) :stuck_out_tongue:

GregDritschler:
> In this commit, pod container statuses are sorted based on the Step order set in the taskSpec. This order ought to be correct as Tektonenforces Steps to be scheduled in this order. In case Tekton adds extra Steps (such as for pipelineresources), Tekton already updates the taskSpec with these Steps. Therefore, Tekton accounts for these internally added Steps when sorting.

The taskSpec passed to MakeTaskRunStatus does not have the internal steps.  It is the user's taskSpec.

Perhaps you are looking at the steps in the TaskRun's status?  Those do have the internal steps because the code fabricates steps for them from the container statuses.  You'll notice those steps are not in order relative to the user's steps.

I suggest testing a TaskRun with a PipelineResource that has an error (e.g. a git resource with a repo that doesn't exist), to confirm the behavior.

Also this change needs to be separated from your other PR #3138 (or, if it depends on it, you need to hold this one until that one is approved and merged).

Peaorl:
Thanks Greg! Good catch, I made sure the updated taskSpec is passed on to the appropriate function (MakeTaskRunStatus).

/hold

GregDritschler:
> Thanks @GregDritschler ! Good catch, I made sure the updated taskSpec is passed on to the appropriate function (MakeTaskRunStatus).
> 

So now the updated taskSpec is passed to MakeTaskRunStatus when the pod is created, but what happens on subsequent reconcile calls when the pod exists?

Peaorl:
I changed it such that the container order in the `podSpec` is used. Tekton sets this container order when the pod is created and bases the order on the `taskSpec` and the internally added `Steps`.

sbwsg (approved):

PR #3258, opened 2020-09-18
Add more details on bumping catalog deps

bobcatfish:
# Changes

Looking at https://github.com/tektoncd/catalog/pull/508 I realized there
is more to update than just the e2e scripts.

I'm actually wondering if we want to bump all of these - I think if
we're doing to do that, we should bump the version of the Task as well
(doesn't seem like it makes sense to have different copies of the same
Task at the same version behave differently). So I'm wondering if we
want to be more selective about when we bump these?

Would like you hear your thoughts @vdemeester @vinamra28 

/hold

/kind documentation

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
NONE
```

### DIFF
--- a/tekton/release-cheat-sheet.md
+++ b/tekton/release-cheat-sheet.md
@@ -130,6 +130,13 @@ the pipelines repo, a terminal window and a text editor.
 14. Announce the release in Slack channels #general and #pipelines.
 
 15. Update [the catalog repo](https://github.com/tektoncd/catalog) test infrastructure
-to use the new release by updating the `RELEASE_YAML` link in [e2e-tests.sh](https://github.com/tektoncd/catalog/blob/v1beta1/test/e2e-tests.sh).
+to use the new release by updating:
+* The `RELEASE_YAML` link in [e2e-tests.sh](https://github.com/tektoncd/catalog/blob/master/test/e2e-tests.sh).
+* The images used in catalog Tasks that are built and published with Pipelines:
+  * [pull-request](https://github.com/tektoncd/catalog/blob/master/task/pull-request/0.1/pull-request.yaml)
+  * [git-batch-merge](https://github.com/tektoncd/catalog/blob/master/task/git-batch-merge/0.2/git-batch-merge.yaml)
+  * [git-clone](https://github.com/tektoncd/catalog/blob/master/task/git-clone/0.2/README.md)
+  * [kaniko](https://github.com/tektoncd/catalog/blob/master/task/kaniko/0.1/kaniko.yaml)
+  * [kubeconfig-creator](https://github.com/tektoncd/catalog/blob/master/task/kubeconfig-creator/0.1/kubeconfig-creator.yaml)
 
 Congratulations, you're done!


vdemeester:
So… what would we do, if those images where not "tied" to the `tektoncd/pipeline` releases ?
Those images are there because they are still used as part of the `PipelineResource`. Depending on the future of the resources, they could be maintained elsewhere, at their own pace.

That said, I tend to agree that we may not want to bump those each time **but** this require us to go through the changelog of release and see if the behaviour changed or not ;and sometimes, it changes for some reason that do not appear evident just by looking at the changelog. Bumping *blindly* is the "easiest" path in that matter.

vdemeester:
Note, that from the [TEP-003](https://github.com/tektoncd/community/blob/master/teps/0003-tekton-catalog-organization.md#versioning-resources), an update is required if we bump the image. If we follow it blindly, we should be those images at each new release of tektoncd/pipeline (not for bugfix release). One less "constraining" rules would be bump the version *if* the behavior changed between two version — but this is more involved because we would need to ensure no behavior changed between two tektoncd/pipeline release.

tekton/release-cheat-sheet.md
```diff
@@ -130,6 +130,13 @@ the pipelines repo, a terminal window and a text editor.
 14. Announce the release in Slack channels #general and #pipelines.
 
 15. Update [the catalog repo](https://github.com/tektoncd/catalog) test infrastructure
-to use the new release by updating the `RELEASE_YAML` link in [e2e-tests.sh](https://github.com/tektoncd/catalog/blob/v1beta1/test/e2e-tests.sh).
+to use the new release by updating:
+* The `RELEASE_YAML` link in [e2e-tests.sh](https://github.com/tektoncd/catalog/blob/master/test/e2e-tests.sh).
```
pritidesai, line 134: Changing `RELEASE_YAML` to the new release would actually have all the tasks run against the new release, isn't it?

https://github.com/tektoncd/catalog/blob/dc4ced0942f0f9a3ad21e1dbcb8bc669e3f0d81b/test/e2e-tests.sh#L19

tekton/release-cheat-sheet.md
```diff
@@ -130,6 +130,13 @@ the pipelines repo, a terminal window and a text editor.
...
+* The images used in catalog Tasks that are built and published with Pipelines:
```
pritidesai, line 135: bump images but no bumping the version, for example, `task/git-clone/0.2` stays as is?

```
labels:
    app.kubernetes.io/version: "0.2"
```

Sorry for asking these questions, I am trying to understand the process. I understand we are still trying to find the right balance. 🙃

tekton/release-cheat-sheet.md
vinamra28: yes, pointing the `RELEASE_YAML` to the new release would allow tasks/pipelines in the catalog to run against the new pipeline version.

tekton/release-cheat-sheet.md
vinamra28: yeah, as per [TEP-003](https://github.com/tektoncd/community/blob/master/teps/0003-tekton-catalog-organization.md#versioning-resources) we should have bumped the version of the resource but having different versions of the resource with a change in just the image version might not make sense until and unless there is a breaking change. Also in the current pipeline release there was some fix in `git-init` image which might have affected the users using the current `git-clone` task `v0.2` so that's why we decided to bump the image version of that task and not adding a new task version. Having said that, we will try to make sure that we bump the version of the task only if there is a breaking change and if there is a bug fix then we should update the current task version

pritidesai:
Should we merge this and have release managers update the dependencies in catalog? Release manager can a PR and the PR can be updated based on the feedback from catalog expert, similar to https://github.com/tektoncd/catalog/pull/576.

vinamra28:
Sounds good to me 🙂
LGTM

bobcatfish:
> Should we merge this and have release managers update the dependencies in catalog? Release manager can create a PR and the PR can be updated based on the feedback from catalog expert, similar to tektoncd/catalog#576.

There's still the question of whether or not to bump the Task version: https://github.com/tektoncd/catalog/pull/597/files#r539321005

bobcatfish:
I've created https://github.com/tektoncd/catalog/issues/644 to continue this discussion.

PR #3185, opened 2020-09-09
Enhance v1beta1 validation code for task 🍸

vdemeester:
# Changes

Opening this to gather some feedback.

This make a better use of knative.dev/pkg apis errors package in order
to make the code simpler, *and* the report better.

- This allows reporting multiple validation error at once — so, one
  failure message (webhook error) would be more useful to the user.
- This enhance the report as it displays more information, like
  `resources.inputs[0].invalidsource.type` instead of
  `resources.inputs.source.type` (the index appear, it's easier for the
  user to know where the error is).
- This reduce `if err != nil` path, simplifying the code.

/hold

- This needs more work (at least on this package)
- <del>Tests that depends on some of this may fail, I need to fix it</del>
- It's gonna be really useful on function like [`validateVariables`](https://github.com/tektoncd/pipeline/blob/master/pkg/apis/pipeline/v1beta1/task_validation.go#L333) — it's now used there :wink: 
- This PR would only do the `Task` part (without moving too much code away), follow-up includes
  - applying this to the rest of the `v1beta1` package
  - extract some code into more files (those are big files, lot's of stuff could be extracted) and more functions
  - make a pass on what is exported, and why

/cc @afrittoli @bobcatfish @sbwsg @nikhil-thomas 

Signed-off-by: Vincent Demeester <vdemeest@redhat.com>

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [x] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

Double check this list of stuff that's easy to miss:

- If you are adding [a new binary/image to the `cmd` dir](../cmd), please update
  [the release Task](../tekton/publish.yaml) to build and release this image.

## Reviewer Notes

[...]

PR #3271, opened 2020-09-21
WIP: helper method to detect key collisions

ImJasonH:
Today, we have a number of cases where repeated key/value pairs are used to simulate a map, in keeping with KRM style. We validate that key names (usually named `name:`) are unique. For example:

```
params:
- name: repo
  value: github.com/foo
- name: revision
  value: master
- name: repo
  value: github.com/bar
```

This is invalid because two items provide the value for the parameter named `foo`. Note that it would _still_ be invalid even if both params provided the same value.

In a lot of places in Tekton we _incorrectly_ return [`apis.ErrMultipleOneOf`](https://godoc.org/knative.dev/pkg/apis#ErrMultipleOneOf) in such cases -- this is incorrect because `MultipleOneOf` is intended to signify that only one of fields A or B should be set, but not both. For example, specifying `taskRef` and `taskSpec` is an `ErrMultipleOneOf`.

This PR adds a helper method that uses reflection to inspect a provided slice of structs, pull out the values of a field (e.g., `Name`), and make sure that slice doesn't specify two items with the same value for,  e.g., `name:`.

This PR includes unit tests for the new method, and converts an existing incorrect usage of `ErrMultipleOneOf` to demonstrate its usage.

/area testing
/kind cleanup
Spawned from https://github.com/knative/pkg/issues/1726

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Introducing a reflect-based helper method to validate key conflicts
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/taskrun_validation.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_validation.go
@@ -19,11 +19,9 @@ package v1beta1
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	"github.com/tektoncd/pipeline/pkg/apis/validate"
 	"k8s.io/apimachinery/pkg/api/equality"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/apis"
 )
 
@@ -60,10 +58,13 @@ func (ts *TaskRunSpec) Validate(ctx context.Context) *apis.FieldError {
 		}
 	}
 
-	if err := validateParameters(ts.Params); err != nil {
+	if err := validate.KeyConflict(ts.Params, "Name", "spec.params.name"); err != nil {
 		return err
 	}
 
+	if err := validate.KeyConflict(ts.Workspaces, "Name", "spec.workspaces.name"); err != nil {
+		return err
+	}
 	if err := validateWorkspaceBindings(ctx, ts.Workspaces); err != nil {
 		return err
 	}
@@ -91,29 +92,11 @@ func (ts *TaskRunSpec) Validate(ctx context.Context) *apis.FieldError {
 
 // validateWorkspaceBindings makes sure the volumes provided for the Task's declared workspaces make sense.
 func validateWorkspaceBindings(ctx context.Context, wb []WorkspaceBinding) *apis.FieldError {
-	seen := sets.NewString()
 	for _, w := range wb {
-		if seen.Has(w.Name) {
-			return apis.ErrMultipleOneOf("spec.workspaces.name")
-		}
-		seen.Insert(w.Name)
-
 		if err := w.Validate(ctx).ViaField("workspace"); err != nil {
 			return err
 		}
 	}
 
-	return nil
-}
-
-func validateParameters(params []Param) *apis.FieldError {
-	// Template must not duplicate parameter names.
-	seen := sets.NewString()
-	for _, p := range params {
-		if seen.Has(strings.ToLower(p.Name)) {
-			return apis.ErrMultipleOneOf("spec.params.name")
-		}
-		seen.Insert(p.Name)
-	}
 	return nil
 }
--- a/pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
@@ -108,7 +108,10 @@ func TestTaskRun_Workspaces_Invalid(t *testing.T) {
 				}},
 			},
 		},
-		wantErr: apis.ErrMultipleOneOf("spec.workspaces.name"),
+		wantErr: &apis.FieldError{
+			Message: `item 1 value for field "Name" conflicts with value for item 0`,
+			Paths:   []string{"spec.workspaces.name"},
+		},
 	}}
 	for _, ts := range tests {
 		t.Run(ts.name, func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


bobcatfish:
🙏  apologies jason just making sure /retest is a legit command

/retest

bobcatfish:
... it ... isn't?

vdemeester (commented):
I really am not a fan of using `reflect` for this (or for anything really) ; the code looks more complicated than it should. 

In the end, we want to get the duplicate `string` in an array of string (or a `sets.NewString()` *set*) right ? (and get the index). We could write a methods that just does that without using `reflect`. The method resp. would be "for a given set of string, give me the index of the one that are duplicated and the value" — and from that point we build the error.

ImJasonH:
> I really am not a fan of using `reflect` for this (or for anything really) ; the code looks more complicated than it should.
> 
> In the end, we want to get the duplicate `string` in an array of string (or a `sets.NewString()` _set_) right ? (and get the index). We could write a methods that just does that without using `reflect`. The method resp. would be "for a given set of string, give me the index of the one that are duplicated and the value" — and from that point we build the error.

Yeah, I agree it's kind of gross. The reason it's WIP is so we can discuss whether it's _too_ gross to be allowed to exist. And it might be! 😄 

Checking duplicates isn't the hard part of what it's doing though. The "hard part", and the reason it uses `reflect`, is because it wants to take a slice-of-`Thing`s, and check that all the `things` have unique `Name` values. But it can't know ahead of time that `Name` is the field name that must be unique. Sometimes it might be `Key` or `SomethingElse` that has to be unqiue. Even if it _did_ hard-code `Name` as the must-be-unique field name (it's usually `Name` in practice), that wouldn't actually remove any `reflect` usage. But that would simplify the method signature, so maybe it's worth it.

All the `reflect` is, because of Go's type system and current lack of generics, we can't just take a `[]struct{ Name string }`, and it can't even take a `[]interface{ Name() string  }` ([example](https://play.golang.org/p/UZsAmxm19Fy)), so it takes an `interface{}` that it enforces at runtime is a slice...of structs...that have the exported field. Maybe in the future generics will make this much simpler?

Another alternative would be to (ab)use `encoding/json` to handle `reflect` for us. In that case we could take a `interface{}`, and `json.Marshal` it, then `json.Unmarshal` into a `[]struct { Name string }`, and from there check uniqueness. In that world `json.Marshal`/`json.Unmarshal` is doing all the `reflect` we'd be doing ourselves, and error checking, and much more, and also generating `[]byte`s we'd just throw out.

It's possible this just isn't worth generalizing, which would be a fine outcome of this discussion honestly. If nothing else I got [a tweet](https://twitter.com/ImJasonH/status/1308075833014263808) out of it. 😆

vdemeester:
> Yeah, I agree it's kind of gross. The reason it's WIP is so we can discuss whether it's _too_ gross to be allowed to exist. And it might be! smile
> 
> Checking duplicates isn't the hard part of what it's doing though. The "hard part", and the reason it uses `reflect`, is because it wants to take a slice-of-`Thing`s, and check that all the `things` have unique `Name` values. But it can't know ahead of time that `Name` is the field name that must be unique. Sometimes it might be `Key` or `SomethingElse` that has to be unqiue. Even if it _did_ hard-code `Name` as the must-be-unique field name (it's usually `Name` in practice), that wouldn't actually remove any `reflect` usage. But that would simplify the method signature, so maybe it's worth it.
> 
> All the `reflect` is, because of Go's type system and current lack of generics, we can't just take a `[]struct{ Name string }`, and it can't even take a `[]interface{ Name() string }` ([example](https://play.golang.org/p/UZsAmxm19Fy)), so it takes an `interface{}` that it enforces at runtime is a slice...of structs...that have the exported field. Maybe in the future generics will make this much simpler?

Right, that's why I don't think we should try to be smart. Looking at the current usage (in the PR), we **always** work with strings. And from my point of view, the function that does the "duplicate check" doesn't have to know anything about field. It just "needs" to tell the "field" for the path error, but that can be passed as a string **and** the rest of the path can be build using `apis.FieldError` api). 

We can build a temporary struct (an array of string), work on it, get the duplicate "index" and let a higher method do the field part (using `apis.FieldError` function `ViaFieldIndex` or `ViaFieldKey` makes it trivial even, see [here](https://github.com/tektoncd/pipeline/pull/3185/files#diff-08b601d2ab6b5717a0decae25c9056e9R255) for example).  All this can be achieved without `reflec` not generics. I don't think generics would help that much here ; making less smart and more "focused" methods does.

What I would do, in gross *go* pseudo code. I didn't really try to implement this so it might be completely wrong but pretty sure we can make something work without `reflect`.

```go
paramNames := extractParamNames(ts.Params) 
// we could use go types smartly here, define a type Params []Param
// and have a Names() method on it, but in the end, it's just a matter of building an array of string
// `apis.FieldError` api has this really nice way to build the error path
err := validateDuplicateString(paramNames, "name").ViaField("params")

func validateDuplicateString(s []string, path string) {
// […]
// use apis.ErrGeneric("the error message", path)
}
```

As someone (or *many ones*) said : "reflection is hard and is to be avoided." :upside_down_face: 

> It's possible this just isn't worth generalizing, which would be a fine outcome of this discussion honestly. If nothing else I got [a tweet](https://twitter.com/ImJasonH/status/1308075833014263808) out of it. laughing

I don't think it is worth generalizing indeed, or at least not to the point of using `reflect`.

ImJasonH:
In your pseudocode example above, `paramNames := extractParamNames(ts.Params)` is the part I was (mistakenly!) trying to generalize, so it could take `extractNames(ts.Workspaces)` or `extractNames(step.Env)`. Once that's done, duplicate checking is straightforward and easy to generalize.

Your feedback is completely spot-on though, generalizing this doesn't save us enough to justify the risk of waking the spirits of the undead by invoking `reflect`. I'll close the PR. 💀

PR #3274, opened 2020-09-22
Introduce Optional Workspaces

sbwsg:
# Changes

Initial design approved in [TEP 10](https://github.com/tektoncd/community/blob/master/teps/0010-optional-workspaces.md).

Prior to this commit any Workspaces declared by a Task are required to be provided by a TaskRun or Pipeline/Run.

This commit introduces the concept of Optional Workspaces: a Task may declare a Workspace as optional and a TaskRun may then omit that Workspace at runtime. Pipelines and PipelineRuns may do the same.

A new variable is introduced for workspaces, "bound", which interpolates to either "true" or "false" depending on whether a workspace binding is included in a TaskRun or not.

Default TaskRun Workspaces are _not_ injected in place of an omitted optional Workspace.

# Submitter Checklist

# Release Notes

```release-note
Introduce optional workspaces: A Task or Pipeline may declare a workspace optional and conditionally change their behaviour based on its presence. A TaskRun or PipelineRun may omit that workspace and thereby modify the Task or Pipeline behaviour.
```

### DIFF
--- a/docs/workspaces.md
+++ b/docs/workspaces.md
@@ -72,6 +72,18 @@ specific `Volume` information to use for the `Workspaces` used by each `Pipeline
 `PipelineRuns` have the added responsibility of ensuring that whatever `Volume` type they
 provide can be safely and correctly shared across multiple `Tasks`.
 
+### Optional `Workspaces`
+
+Both Tasks and Pipelines can declare a Workspace "optional". When an optional Workspace
+is declared the TaskRun or PipelineRun may omit a Workspace Binding for that Workspace.
+The Task or Pipeline behaviour may change when the Binding is omitted. This feature has
+many uses:
+
+- A Task may optionally accept credentials to run authenticated commands.
+- A Pipeline may accept optional configuration that changes the linting or compilation
+parameters used.
+- An optional build cache may be provided to speed up compile times.
+
 ## Configuring `Workspaces`
 
 This section describes how to configure one or more `Workspaces` in a `TaskRun`.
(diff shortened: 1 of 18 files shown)


sbwsg:
oops, need to include an update to workspaces.md with some new info

pkg/reconciler/taskrun/resources/apply.go
```diff
@@ -108,24 +109,37 @@ func ApplyContexts(spec *v1beta1.TaskSpec, rtr *ResolvedTaskResources, tr *v1bet
 	return ApplyReplacements(spec, replacements, map[string][]string{})
 }
 
-// ApplyWorkspaces applies the substitution from paths that the workspaces in w are mounted to, the
-// volumes that wb are realized with in the task spec ts and the PersistentVolumeClaim names for the
+// ApplyWorkspaces applies the substitution from paths that the workspaces in decls mounted to, the
+// volumes that binds are realized with in the task spec and the PersistentVolumeClaim names for the
 // workspaces.
-func ApplyWorkspaces(spec *v1beta1.TaskSpec, w []v1beta1.WorkspaceDeclaration, wb []v1beta1.WorkspaceBinding) *v1beta1.TaskSpec {
+func ApplyWorkspaces(spec *v1beta1.TaskSpec, decls []v1beta1.WorkspaceDeclaration, binds []v1beta1.WorkspaceBinding) *v1beta1.TaskSpec {
```
vdemeester, line 115: `decls` for declared workspaces ?

vdemeester (commented):
looking good, one comment/question :upside_down_face:

pkg/reconciler/taskrun/resources/apply.go
sbwsg: yeah, exactly right!  Maybe `declarations` to make it a bit clearer?

vdemeester: @sbwsg maybe :stuck_out_tongue_closed_eyes: it's a nit, but I would indeed prefer `declarations` :stuck_out_tongue_closed_eyes:

docs/workspaces.md
```diff
@@ -72,6 +72,18 @@ specific `Volume` information to use for the `Workspaces` used by each `Pipeline
...
+Both Tasks and Pipelines can declare a Workspace "optional". When an optional Workspace
+is declared the TaskRun or PipelineRun may omit a Workspace Binding for that Workspace.
+The Task or Pipeline behaviour may change when the Binding is omitted. This feature has
+many uses:
+
+- A Task may optionally accept credentials to run authenticated commands.
+- A Pipeline may accept optional configuration that changes the linting or compilation
+parameters used.
+- An optional build cache may be provided to speed up compile times.
+
```
jerop, line 86: I suggest that we add documentation about what happens when a workspace is optional but it has a default specified, saw that that case was covered in the tests but I'd imagine a user would wonder about that when reading the docs

https://github.com/tektoncd/pipeline/blob/dd548e6709be41db82e2e1f6e0574551530bd480/pkg/reconciler/taskrun/taskrun_test.go#L2851-L2855

jerop (commented):
looks great! 

just a comment on optional + default workspaces

docs/workspaces.md
sbwsg: Great idea, I've updated the workspaces doc to add a new section which describes the default config in a bit more detail and includes this info as a note.

pkg/reconciler/taskrun/resources/apply.go
sbwsg: updated!

sbwsg:
Does anyone have any other suggestions or changes they'd like to see?  If not, 🙏 lgtm?  😄

sbwsg:
This also fixes https://github.com/tektoncd/pipeline/issues/3275 😅

popcor255 (approved):
/lgtm

PR #3278, opened 2020-09-23
Provide configuration option to disallow omitting known_hosts

yaoxiaoqi:
# Changes

This commit closes tektoncd#2981

Prior to this commit when a Git SSH secret does not include a known_hosts field, Tekton will accept any public key that an SSH/Git server returns. This could be a security concern and may not be an organization's desired behavior.

This commit introduces a feature flag `require-git-ssh-secret-known-hosts`: the default value is false. When the flag is true, `known_host` field must be included in Git SSH Secret.

`checkGitSSHSecret()` will check the type of every secrets. If the type of the secret is `kubernetes.io/ssh-auth` and `require-git-ssh-secret-known-hosts` is true, `checkGitSSHSecret` will validate whether `known_hosts` field is in the secret data. If `known_hosts` is not found, it will emit an error which ceases the creation of the pod.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```release-note
Provide configuration option to disallow omitting known_hosts in Git SSH Secret.
```

### DIFF
--- a/pkg/apis/pipeline/v1alpha1/pipeline_validation_test.go
+++ b/pkg/apis/pipeline/v1alpha1/pipeline_validation_test.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 	"testing"
 
-	tb "github.com/tektoncd/pipeline/internal/builder/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	v1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	corev1 "k8s.io/api/core/v1"
@@ -29,7 +28,7 @@ import (
 )
 
 func TestPipeline_Validate(t *testing.T) {
-	tests := []struct {
+	for _, tt := range []struct {
 		name            string
 		p               *v1alpha1.Pipeline
 		failureExpected bool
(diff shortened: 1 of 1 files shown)


pkg/pod/creds_init.go
```diff
@@ -63,6 +68,11 @@ func credsInit(serviceAccountName, namespace string, kubeclient kubernetes.Inter
 			return nil, nil, nil, err
 		}
 
+		err = checkGitSSHSecret(ctx, secret)
```
ImJasonH, line 71: Go style nit: If we can, generally we try to write cases like this as:

```go
if err := checkGitSSHSecret(ctx, secret); err != nil {
  return nil, nil, nil, err
}
```

This keeps the returned error variable in the smallest possible scope, and doesn't change the value of the previous `err` value.

Over time this form becomes easy to read at a glance as, "try this; if it returns an error, return it"

pkg/pod/creds_init_test.go
```diff
@@ -169,3 +174,99 @@ func TestCredsInit(t *testing.T) {
 		})
 	}
 }
+
+func TestCheckGitSSHSecret(t *testing.T) {
+	for _, tc := range []struct {
+		desc         string
+		configMap    *corev1.ConfigMap
+		secret       *corev1.Secret
+		wantError    bool
```
ImJasonH, line 183: I think we can simplify these test cases by just saying that, if `wantErrorMsg` is not `""`, then we expect an error, and we expect it to have that message.

Basically, `wantError` can be derived as `wantError := tc.wantErrorMsg != ""` down near line 257, and can be removed from these test cases.

ImJasonH (commented):
The change looks great, thanks for writing such good tests! Just a couple small comments from me, but I think Scott should also take a look since he has more context on this feature.

sbwsg:
I've just fetched the code and run it.  One issue I've encountered is that when `known_hosts` is missing the TaskRun will fail with the following `status.condition`:

```yaml
status:
  conditions:
  - lastTransitionTime: "2020-09-23T15:21:17Z"
    message: 'failed to create task run pod "authenticating-git-commands": translating
      TaskSpec to Pod: Git SSH Secret must have "known_hosts" included when feature
      flag "require-git-ssh-secret-known-hosts" is set to true. Maybe invalid TaskSpec'
    reason: CouldntGetTask
    status: "False"
    type: Succeeded
```

The `reason` here, `CouldntGetTask`, is I think the wrong reason to use. Ideally the reason would be [`TaskRunValidationFailed`](https://github.com/tektoncd/pipeline/blob/49d4b5aedf5510a84a92a4b2faf96c844a21f8c5/pkg/pod/status.go#L45-L47) instead.

Unfortunately the `reason` is set at a higher level in the code.  We set it [here in the reconciler](https://github.com/tektoncd/pipeline/blob/49d4b5aedf5510a84a92a4b2faf96c844a21f8c5/pkg/reconciler/taskrun/taskrun.go#L491) which is quite far from the validation that you're performing in this PR.  In that same func we _do_ return a different `reason` if the [node exceeded a resource quota](https://github.com/tektoncd/pipeline/blob/49d4b5aedf5510a84a92a4b2faf96c844a21f8c5/pkg/reconciler/taskrun/taskrun.go#L466-L480) so I wonder if we could follow a similar pattern for the `Git SSH Secret must have "known_hosts"` error?

ImJasonH:
Ugh, this failure is really annoying:

```
pkg/apis/config/feature_flags.go:32:2: G101: Potential hardcoded credentials (gosec)
	requireGitSSHSecretKnownHostsKey        = "require-git-ssh-secret-known-hosts"
```

It looks like some linter tool is trying to helpfully block us from committing a sensitive string to the repo, but this isn't a credential at all, so we just need to tell it to ignore it.

It looks like you can add `//nolint : gosec` on the line above.

Could you also squash your commits into one commit before we approve and merge?

yaoxiaoqi:
> Ugh, this failure is really annoying:
> 
> ```
> pkg/apis/config/feature_flags.go:32:2: G101: Potential hardcoded credentials (gosec)
> 	requireGitSSHSecretKnownHostsKey        = "require-git-ssh-secret-known-hosts"
> ```
> 
> It looks like some linter tool is trying to helpfully block us from committing a sensitive string to the repo, but this isn't a credential at all, so we just need to tell it to ignore it.
> 
> It looks like you can add `//nolint : gosec` on the line above.
> 
> Could you also squash your commits into one commit before we approve and merge?

Sure, I will make the linter ignore this line and squash the commits. And I'm going to correct the error reason as Scott said. This will add some new code which needs to be reviewed by Scott. So we can't merge this pr immediately.

PR #3281, opened 2020-09-24
Remove test builders from validate_params_test.go

popcor255:
# Changes

* Using minimal internal types are simpler than full pipeline objects. 
* Removed duplicate tests
# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

/area testing
/kind cleanup

#3178

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/validate_params_test.go
+++ b/pkg/reconciler/pipelinerun/resources/validate_params_test.go
@@ -19,103 +19,92 @@ package resources
 import (
 	"testing"
 
-	tb "github.com/tektoncd/pipeline/internal/builder/v1beta1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestValidateParamTypesMatching_Valid(t *testing.T) {
-	tcs := []struct {
-		name          string
-		p             *v1beta1.Pipeline
-		pr            *v1beta1.PipelineRun
-		errorExpected bool
+
+	stringValue := *v1beta1.NewArrayOrString("stringValue")
+	arrayValue := *v1beta1.NewArrayOrString("arrayValue", "arrayValue")
+
+	for _, tc := range []struct {
+		name        string
+		description string
+		pp          []v1beta1.ParamSpec
+		prp         []v1beta1.Param
 	}{{
 		name: "proper param types",
-		p: tb.Pipeline("a-pipeline", tb.PipelineSpec(
-			tb.PipelineParamSpec("correct-type-1", v1beta1.ParamTypeString),
-			tb.PipelineParamSpec("mismatching-type", v1beta1.ParamTypeString),
-			tb.PipelineParamSpec("correct-type-2", v1beta1.ParamTypeArray))),
-		pr: tb.PipelineRun("a-pipelinerun", tb.PipelineRunSpec(
-			"test-pipeline",
-			tb.PipelineRunParam("correct-type-1", "somestring"),
-			tb.PipelineRunParam("mismatching-type", "astring"),
-			tb.PipelineRunParam("correct-type-2", "another", "array"))),
-		errorExpected: false,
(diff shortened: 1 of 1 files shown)


pkg/reconciler/pipelinerun/resources/validate_params_test.go
```diff
@@ -86,92 +76,106 @@ func TestValidateParamTypesMatching_Valid(t *testing.T) {
...
-		p    *v1beta1.Pipeline
-		pr   *v1beta1.PipelineRun
+
+	mockStringValue := *v1beta1.NewArrayOrString("mockStringValue")
+	mockArrayValue  := *v1beta1.NewArrayOrString("mockArrayValue", "mockArrayValue")
+
+	for _, tc := range []struct {
+		name          string
+		description	  string
+		pp            []v1beta1.ParamSpec
```
popcor255, line 86: just why...
:crying_cat_face:

pkg/reconciler/pipelinerun/resources/validate_params_test.go
```diff
@@ -86,92 +76,106 @@ func TestValidateParamTypesMatching_Valid(t *testing.T) {
...
+		prp			  []v1beta1.Param
+		errorExpected bool
```
popcor255, line 88: just why...
:crying_cat_face:

vdemeester (changes requested):
> Test names use underscores instead of blank spaces. Golang test convert blanks to underscores into the terminal anyway. Changing in the code base ahead of time makes it easier to find failed test in editors.

I am not really a fan of doing that change. The fact that `go` replace converts blank to something is an implementation detail.

> A description was added to each test case.

Not a fan of that one either, the `name` field should be sufficient *and* it's gonna be extra burden to maintain. Just like most comments, it might become incorrect with what the test actually does.

ImJasonH:
> > Test names use underscores instead of blank spaces. Golang test convert blanks to underscores into the terminal anyway. Changing in the code base ahead of time makes it easier to find failed test in editors.
> 
> I am not really a fan of doing that change. The fact that `go` replace converts blank to something is an implementation detail.

+1, I've even been changing underscores to spaces in other changes 😆 

> Not a fan of that one either, the `name` field should be sufficient _and_ it's gonna be extra burden to maintain. Just like most comments, it might become incorrect with what the test actually does.

Agree here too, test names should be short and descriptive enough to not need a longer description. If you do need to describe something in more detail, that's fine, but I'd do it in a `// comment` and not in code that gets printed with the test. Having `name` and `description` is also inconsistent with other tests throughout the codebase, and I don't think we should update them all to match.

ImJasonH:
/kind cleanup

Poking the kind label check 🤞

popcor255:
@vdemeester @ImJasonH Thanks for the feedback! :smile:
I made the changes. You can take another

pkg/reconciler/pipelinerun/resources/validate_params_test.go
```diff
@@ -19,159 +19,133 @@ package resources
...
 
 func TestValidateParamTypesMatching_Valid(t *testing.T) {
-	tcs := []struct {
-		name          string
-		p             *v1beta1.Pipeline
-		pr            *v1beta1.PipelineRun
-		errorExpected bool
+
+	mockStringValue := *v1beta1.NewArrayOrString("mockStringValue")
+	mockArrayValue := *v1beta1.NewArrayOrString("mockArrayValue", "mockArrayValue")
```
vdemeester, line 29: nit: technically, this is not a "mock" :stuck_out_tongue_closed_eyes: we could name it `stringValue` or anything.
But it's a nit, no need to fix it now :wink:

vdemeester (approved):
/meow
/hold cancel

pkg/reconciler/pipelinerun/resources/validate_params_test.go
popcor255:
```bash
%s/mockStringValue/stringValue/g
%s/mockArrayValue/arrayValue/g
```

popcor255:
poke @dlorenc

jerop (commented):
/lgtm

PR #3286, opened 2020-09-26
Base entrypoint image on distroless

ImJasonH:
Reopening #2562 which for some reason GitHub/Prow won't let me reopen 🤷‍♂️ 

With [`ko` gaining multi-arch support](https://github.com/google/ko/pull/38), this change is becoming a bit more relevant, since it makes every TaskRun require a `cp` binary in the base image, which makes it harder to support multi-arch support in Tekton.

-----

The only reason it was based on busybox was to have access to `cp`,
which it used to copy its binary to /tekton/tools for use in later
steps.

Instead of relying on `cp`, this adds a "cp mode" to the entrypoint
binary itself. When invoked with the positional args `cp <src> <dst>`,
it copies src to dst using Go's os and io packages.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Base entrypoint image on distroless/static:nonroot and give it a mode to copy itself to the destination, instead of relying on cp being present in the base image.
```

cc @mattmoor @afrittoli @vdemeester

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -4,11 +4,9 @@ baseImageOverrides:
   # They are produced from ./images/Dockerfile
   github.com/tektoncd/pipeline/cmd/creds-init: gcr.io/tekton-nightly/github.com/tektoncd/pipeline/build-base:latest
   github.com/tektoncd/pipeline/cmd/git-init: gcr.io/tekton-nightly/github.com/tektoncd/pipeline/build-base:latest
+
   # GCS fetcher needs root due to workspace permissions
   github.com/tektoncd/pipeline/vendor/github.com/GoogleCloudPlatform/cloud-builders/gcs-fetcher/cmd/gcs-fetcher: gcr.io/distroless/static:latest
   # PullRequest resource needs root because in output mode it needs to access pr.json
   # which might have been copied or written with any level of permissions.
   github.com/tektoncd/pipeline/cmd/pullrequest-init: gcr.io/distroless/static:latest
-
-  # Our entrypoint image does not need root, it simply needs to be able to 'cp' the binary into a shared location.
-  github.com/tektoncd/pipeline/cmd/entrypoint: gcr.io/distroless/base:debug-nonroot
--- a/cmd/entrypoint/main.go
+++ b/cmd/entrypoint/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"flag"
+	"io"
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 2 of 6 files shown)


mattmoor:
/lgtm

FWIW :D

vdemeester:
> With [`ko` gaining multi-arch support](https://github.com/google/ko/pull/38), this change is becoming a bit more relevant, since it makes every TaskRun require a `cp` binary in the base image, which makes it harder to support multi-arch support in Tekton.

I am not sure I follow entirely this, before this change, we *just* need to ensure that entrypoint image is multi-arch (just like `busybox` is I think), right ? Now it's based on a non-os image so we don't really need to care about multi-arch because there is only binaries in there ?

This just bakes `cp` in entrypoint as an very lightweight alternative of `cp`. One thing though, is this be considered as a breaking change as we are removing a some `debug` (busybox) tooling from the entrypoint image ?

ImJasonH:
> I am not sure I follow entirely this, before this change, we _just_ need to ensure that entrypoint image is multi-arch (just like `busybox` is I think), right ? Now it's based on a non-os image so we don't really need to care about multi-arch because there is only binaries in there ?

Yep, exactly. It was based on `distroless/base:debug` before as an optimal middle-ground between "as minimal as possible (distroless)" and "contains `cp`", but now that we're trying to also make it multi-arch, I think it makes sense to base it on `distroless/static` instead. And we can do that just by moving `cp` into the binary itself.

Basing on `busybox` would also be an option, since it's multi-arch, but that moves us a bit further from the "as minimal as possible" goal.

> This just bakes `cp` in entrypoint as an very lightweight alternative of `cp`. One thing though, is this be considered as a breaking change as we are removing a some `debug` (busybox) tooling from the entrypoint image ?

I don't think we should consider it a breaking change. The "debug stuff" was only present in the `entrypoint` image when it runs as an `initContainer` to copy the `entrypoint` binary out to the shared volume. I'm not even sure how a user would come to depend on that behavior, but if they did they're definitely doing something weird. 🙃

vdemeester:
> Yep, exactly. It was based on `distroless/base:debug` before as an optimal middle-ground between "as minimal as possible (distroless)" and "contains `cp`", but now that we're trying to also make it multi-arch, I think it makes sense to base it on `distroless/static` instead. And we can do that just by moving `cp` into the binary itself.

Well, we are not copying `cp` into the binary itself, we are "coding" `cp` :stuck_out_tongue_closed_eyes:.

> Basing on `busybox` would also be an option, since it's multi-arch, but that moves us a bit further from the "as minimal as possible" goal.

Isn't `distroless/base:debug` using `busybox` ? (I thought so, maybe I am wrong – or maybe it's alpine based).

> I don't think we should consider it a breaking change. The "debug stuff" was only present in the `entrypoint` image when it runs as an `initContainer` to copy the `entrypoint` binary out to the shared volume. I'm not even sure how a user would come to depend on that behavior, but if they did they're definitely doing something weird. upside_down_face

We do publish the entrypoint image, so any user could use it as its own, independently of what Tekton does with it. I do agree it would be weird to use it (instead of an `alpine` or `busybox` image), but still, some might :sweat_smile:

ImJasonH:
> Well, we are not copying `cp` into the binary itself, we are "coding" `cp` 😝.

Right right. Copying it's functionality, is what I meant.

> Isn't `distroless/base:debug` using `busybox` ? (I thought so, maybe I am wrong – or maybe it's alpine based).

I don't think so:
Screen Shot 2020-09-28 at 9 16 49 AM

(from https://viz.kontain.me)

> We do publish the entrypoint image, so any user could use it as its own, independently of what Tekton does with it. I do agree it would be weird to use it (instead of an `alpine` or `busybox` image), but still, some might 😅

I think I still consider that far enough outside intended use that if it breaks anybody it's because they were doing something we don't expect to support. ([Hyrum's Law](https://www.hyrumslaw.com/), of course, aside).

vdemeester:
https://github.com/GoogleContainerTools/distroless#base-operating-system Interesting, it seems to track debian :upside_down_face: :angel:

Issue #2992, opened 2020-07-23
TestReconcile_CloudEvents Unit test failed in nightly job


savitaashture:
# Expected Behavior

Pipeline nightly unit test passes

# Actual Behavior

Pipeline nightly unit test failed:

Taskrun : pipeline-release-nightly-hkptk-unit-tests-2h6tv

Failure:
```
=== RUN   TestReconcile_CloudEvents
2020-07-23T03:25:50.769Z	ERROR	TestReconcile_CloudEvents	taskrun/controller.go:58	Failed to create taskrun metrics recorder running_taskruns_count: cannot register view "running_taskruns_count"; a different view with the same name is already registered
taskruns_pod_latency: cannot register view "taskruns_pod_latency"; a different view with the same name is already registered
github.com/tektoncd/pipeline/pkg/reconciler/taskrun.NewController.func1
	/workspace/src/github.com/tektoncd/pipeline/pkg/reconciler/taskrun/controller.go:58
github.com/tektoncd/pipeline/pkg/reconciler/taskrun.getTaskRunController
	/workspace/src/github.com/tektoncd/pipeline/pkg/reconciler/taskrun/taskrun_test.go:271
github.com/tektoncd/pipeline/pkg/reconciler/taskrun.TestReconcile_CloudEvents
	/workspace/src/github.com/tektoncd/pipeline/pkg/reconciler/taskrun/taskrun_test.go:726
testing.tRunner
	/usr/local/go/src/testing/testing.go:1039
2020-07-23T03:25:50.770Z	INFO	TestReconcile_CloudEvents	taskrun/controller.go:94	Setting up event handlers
2020-07-23T03:25:50.774Z	INFO	TestReconcile_CloudEvents.config-store	configmap/store.go:154	defaults/features config "config-defaults" config was added or updated: &config.Defaults{DefaultTimeoutMinutes:60, DefaultServiceAccount:"", DefaultManagedByLabelValue:"tekton-pipelines", DefaultPodTemplate:(*pod.Template)(nil), DefaultCloudEventsSink:"http://synk:8080"}
2020-07-23T03:25:50.775Z	INFO	TestReconcile_CloudEvents.config-store	configmap/store.go:154	defaults/features config "feature-flags" config was added or updated: &config.FeatureFlags{DisableHomeEnvOverwrite:false, DisableWorkingDirOverwrite:false, DisableAffinityAssistant:false, RunningInEnvWithInjectedSidecars:true}
[...]

PR #3292, opened 2020-09-28
Accept CloudEvents in any order for CE reconcile test

sbwsg:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/2992

Prior to this commit test cloud events received during reconcile were
required to be emitted in a strict order. We've been seeing some flakes in
our unit tests related to this and in
https://github.com/tektoncd/pipeline/issues/2992 we posited that this might
be happening because the cloud events controller emits each event in a
separate go routine. We added some extra logging to try and catch the
details of the flakes and it does appear that the events are being sent
out-of-order.

This commit adds a new helper func that allows the expected cloudevents
to appear in any order while still failing the test if an unexpected event
(or not enough events) are received.

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -311,9 +311,12 @@ func checkEvents(t *testing.T, fr *record.FakeRecorder, testName string, wantEve
 
 func checkCloudEvents(t *testing.T, fce *cloudevent.FakeClient, testName string, wantEvents []string) error {
 	t.Helper()
-	return eventFromChannel(fce.Events, testName, wantEvents)
+	return eventFromChannelUnordered(fce.Events, wantEvents)
 }
 
+// eventFromChannel takes a chan of string, a test name, and a list of events that a test
+// expects to receive. The events must be received in the same order they appear in the
+// wantEvents list. Any extra or too few received events are considered errors.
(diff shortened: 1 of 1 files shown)


pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -350,6 +353,47 @@ func eventFromChannel(c chan string, testName string, wantEvents []string) error
...
+					expected = expected[:len(expected)-1]
+					break
+				}
+			}
+			if !found {
+				return fmt.Errorf("unexpected event received: %q", event)
+			}
+		case <-timer.C:
+			if len(expected) != 0 {
+				return fmt.Errorf("timed out waiting for %d more events", len(expected))
```
GregDritschler, line 389: maybe include the remaining expected events too

fmt.Errorf("timed out waiting for %d more events: %#v", len(expected), expected)

sbwsg: Good idea; done!

vdemeester (commented):
/lgtm
/cc @afrittoli

afrittoli (commented):
Thanks for this!
I wonder if we should do the same thing in the pipeline run controller tests, but in any case this PR looks.

afrittoli (commented):
/approve

PR #3299, opened 2020-09-29
Add ability to replace image names for examples tests based on arch

barthy1:
# Changes

Some images which are used to run the examples tests are amd64 only.
To run the examples tests for other architectures it makes sense to replace the
amd64 image names to arch-specific ones at runtime.

- add code to search and replace image names based on architecture in the
examples yaml files at runtime.
- provide mapping between amd64 and s390x image names.
- remove several tests from excluded list for s390x.

# Submitter Checklist

review them:

# Release Notes
```release-note
NONE
```

### DIFF
--- a/test/examples_test.go
+++ b/test/examples_test.go
@@ -66,8 +66,9 @@ func waitValidateTaskRunDone(t *testing.T, c *clients, taskRunName string) {
 }
 
 // substituteEnv substitutes docker repos and bucket paths from the system
-// environment for input to allow tests on local clusters. It also unsets the
-// namespace for ServiceAccounts so that they work under test.
+// environment for input to allow tests on local clusters. It unsets the
+// namespace for ServiceAccounts so that they work under test. It also
+// replaces image names to arch specific ones, based on provided mapping.
 func substituteEnv(input []byte, namespace string) ([]byte, error) {
 	// Replace the placeholder image repo with the value of the
 	// KO_DOCKER_REPO env var.
@@ -81,6 +82,10 @@ func substituteEnv(input []byte, namespace string) ([]byte, error) {
 	// the test namespace using `ko create -n`
 	output = defaultNamespaceRE.ReplaceAll(output, []byte("namespace: "+namespace))
 
+	// Replace image names to arch specific ones, where it's necessary
+	for existingImage, archSpecificImage := range imagesMappingRE {
+		output = existingImage.ReplaceAll(output, archSpecificImage)
+	}
 	return output, nil
 }
 
--- a/test/multiarch_utils.go
+++ b/test/multiarch_utils.go
@@ -18,6 +18,7 @@ package test
 
 import (
 	"os"
+	"regexp"
 	"runtime"
 	"testing"
 
@@ -25,8 +26,9 @@ import (
 )
 
 var (
-	imageNames    = initImageNames()
-	excludedTests = initExcludedTests()
+	imageNames      = initImageNames()
+	excludedTests   = initExcludedTests()
+	imagesMappingRE map[*regexp.Regexp][]byte
 )
 
 const (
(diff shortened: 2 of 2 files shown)


test/multiarch_utils.go
```diff
@@ -59,6 +61,33 @@ func initImageNames() map[int]string {
 	}
 }
 
+// generate the map ready to search and replace image names with regexp for examples files.
+// search is done using "image: <name>" pattern.
+func GetImagesMappingRE() map[*regexp.Regexp][]byte {
```
ImJasonH, line 66: Does this method need to be exported?

test/multiarch_utils.go
```diff
@@ -46,6 +47,7 @@ func getTestArch() string {
 	return runtime.GOARCH
 }
 
+// return the map with arch dependent image names for e2e tests
```
ImJasonH, line 50: Go style nit: Doc comments should be in the form of `// methodName returns the map ...` -- in other words, the first word in the comment should be the method name.

sbwsg (approved):

test/examples_test.go
```diff
@@ -81,6 +82,10 @@ func substituteEnv(input []byte, namespace string) ([]byte, error) {
 	// the test namespace using `ko create -n`
 	output = defaultNamespaceRE.ReplaceAll(output, []byte("namespace: "+namespace))
 
+	// Replace image names to arch specific ones, where it's necessary
```
ImJasonH, line 85: Can you update the method's doc comment to mention this new replacement?

test/examples_test.go
```diff
@@ -81,6 +82,10 @@ func substituteEnv(input []byte, namespace string) ([]byte, error) {
...
+	for existingImage, archSpecificImage := range imagesMappingRE {
+		output = existingImage.ReplaceAll(output, archSpecificImage)
```
ImJasonH, line 87: I'm getting a little nervous about proliferating regexp-based replacements in our YAML files, it seems like it might end up making the tests brittle. Should we investigate using a purpose-built tool like kustomize to apply targetted replacements instead? (doesn't have to block this PR, just curious about your thoughts)

sbwsg:
Thanks @barthy1 ! You'll need to also add a release-note section at the bottom of your PR description that looks like this:

``````
```release-note
NONE
```
``````

File: test/multiarch_utils.go (hunk 1 above)
Comment (barthy1) [reply test/multiarch_utils.go:66]: otherwise golangci-lint fails with
```
# make golangci-lint
🐱 running golangci-lint…
test/multiarch_utils.go:66:6: `getImagesMappingRE` is unused (deadcode)
func getImagesMappingRE() map[*regexp.Regexp][]byte {
     ^
Makefile:171: recipe for target 'golangci-lint' failed
make: *** [golangci-lint] Error 1
```
Will be happy to get workaround here

File: test/examples_test.go (hunk 4 above)
Comment (barthy1) [reply test/examples_test.go:88]: I agree that massive regexp replacements might be dangerous. To support given regexp usage, it's for non amd64 only (so won't cover most of the test runs), it's for very specific case ("image: <name>") and regexp is already used in the same func.
Anyway, I'd like to spend some time to investigate kustomize usage option for examples tests(outside of this PR if you don't mind), because in general I agree with you that every work should be done by designed tool.

File: test/examples_test.go (hunk 3 above)
Comment (barthy1) [reply test/examples_test.go:86]: done

File: test/multiarch_utils.go (hunk 2 above)
Comment (barthy1) [reply test/multiarch_utils.go]: done

File: test/examples_test.go (hunk 4 above)
Comment (ImJasonH) [reply test/examples_test.go:88]: sgtm, no need to do it in this PR but I'd like to consider it in future changes is all.

File: test/multiarch_utils.go (hunk 1 above)
Comment (ImJasonH) [reply test/multiarch_utils.go:66]: Ughhhh, the tool is just wrong in this case, it's used by a var.

As a workaround, we could try:

```go
var imagesMappingRE map[*regexp.Regexp][]byte

func init() {
  imagesMappingRE =  getImagesMappingRE()
}
```

vdemeester: we can also use the `// nolint` comments :wink:

barthy1: thanks @ImJasonH , with this modification golangci-lint checks passed

barthy1:
@ImJasonH do you have other suggestions for this PR or you are ok with the current state?

ImJasonH (commented):
/lgtm

PR #3302, opened 2020-09-29
perf: create (pipeline|task)run timeout checks in background

eddie4941:
Both the pipelinerun and taskrun controllers start timeout checks at
startup. They do this by iterating though each namespace and spawning a
go routine for each pipelinerun/taskrun to run the check in the
background. Although each timeout check is done in a go routine, the
iteration through namespaces and creation of the timeout checks is done
in a blocking manner. This adds significant latency at startup when the
number of namespaces is large. Ultimately, this causes a delay in how
fast each controllers can actually start reconciling resources. To
speed up the startup time, this changes the logic so that the iteration
through namespaces is done in the background. The timeout checks were
already carried out in separate go routines and were therefore safe to
use in a concurrent context, so no extra logic was needed to make this
change work in a concurrent context.

# Changes

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
controller and startup time is improved when lots of namespaces are being managed
```

### DIFF
--- a/pkg/timeout/handler_test.go
+++ b/pkg/timeout/handler_test.go
@@ -111,7 +111,7 @@ func TestTaskRunCheckTimeouts(t *testing.T) {
 	}
 
 	th.SetCallbackFunc(f)
-	th.CheckTimeouts(context.Background(), testNs, c.Kube, c.Pipeline)
+	go th.CheckTimeouts(context.Background(), testNs, c.Kube, c.Pipeline)
 
 	for _, tc := range []struct {
 		name           string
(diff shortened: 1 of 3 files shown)


pkg/timeout/handler_test.go
```diff
@@ -111,7 +111,7 @@ func TestTaskRunCheckTimeouts(t *testing.T) {
 	}
 
 	th.SetCallbackFunc(f)
-	th.CheckTimeouts(context.Background(), testNs, c.Kube, c.Pipeline)
+	go th.CheckTimeouts(context.Background(), testNs, c.Kube, c.Pipeline)
```
eddie4941, line 114: The test already use a polling approach below via `wait.PollImmediate` so Its okay not to block here. However, if leaving this here non blocking is a concern, im happy to block until the call is done to preserve old behavior.

sbwsg (approved):

Issue #2559, opened 2020-05-06
Enforce TaskRun timeout using entrypoint binary


imjasonh:
# Problem

Today, TaskRun timeouts are enforced by the Tekton controller scheduling a goroutine to check in the future that a TaskRun has completed -- if it hasn't, it stops execution by deleting the TaskRun's underlying Pod, and updates the TaskRun's status to indicate timeout.

This leads to Pod logs being GCed more quickly than in the case of successful or failed TaskRuns, where the Pod stays around on the cluster, potentially making debugging timeouts harder than necessary.

# Proposal

Instead, we could use the entrypoint binary we inject to order steps, and have the entrypoint binary also take a flag describing the time after which execution should fail with timeout. This would be passed in by the controller, and the entrypoint binary would execute the underlying command, with a context that times out after that time (`time.Until(deadline)`) and reports timeout.

# Considerations

This moves timeout enforcement into each executing Pod, so timeouts wouldn't be enforced until the Pod actually starts executing -- if a TaskRun's Pod is never scheduled, then its timeout wouldn't be enforced, unlike today. We could keep controller-side timeout enforcement for scheduling timeouts, maybe?

Related to #1690 which proposes using the entrypoint binary to enforce per-step timeouts.

ImJasonH:
/reopen
/remove-lifecycle-rotten

I think this is still a worthwhile change, and might even make a good-first-issue.

I think it would still be worth enforcing timeouts in the reconciler, to catch cases where a pod sat unschedulable until its timeout elapsed, and we could just not create the pod in that case.

ywluogg:
I'm interested in taking this if no one is working on it.

ImJasonH:
> I'm interested in taking this if no one is working on it.

Sounds good! There's going to be some overlap with @Peaorl 's work in #3087, which you should just be aware of, but otherwise this seems fairly straightforward.

One thing we should make sure to do is guard this behavior change with a [feature flag](https://github.com/tektoncd/pipeline/blob/master/config/config-feature-flags.yaml) defaulting to false, in case the change in behavior takes anybody by surprise. In general I don't expect it to cause any problems though.

PR #3304, opened 2020-09-30
Add enforcement of TaskRun's timeout to step level

ywluogg:
# Changes

Enforces the TaskRun's timeout to step level and guard the feature with a feature flag. By default the new enforcement is disabled.

Added unit tests and integration tests along with the feature.

**NOTE: PR is based on another PR [#3087](https://github.com/tektoncd/pipeline/pull/3087). Will rebase once that is merged.**

Closes [#2559](https://github.com/tektoncd/pipeline/issues/2559)

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

Enforces the TaskRun's timeout to step level and guard the feature with a feature flag. By default the new enforcement is disabled.

### DIFF
--- a/cmd/entrypoint/runner.go
+++ b/cmd/entrypoint/runner.go
@@ -21,7 +21,7 @@ type realRunner struct {
 
 var _ entrypoint.Runner = (*realRunner)(nil)
 
-func (rr *realRunner) Run(timeout string, args ...string) error {
+func (rr *realRunner) Run(timeout string, taskRunDeadline string, args ...string) error {
 	if len(args) == 0 {
 		return nil
 	}
@@ -38,9 +38,10 @@ func (rr *realRunner) Run(timeout string, args ...string) error {
 	// Add timeout to context if a non-zero timeout is specified for a step
 	ctx := context.Background()
 	var cancel context.CancelFunc
-	timeoutContext, _ := time.ParseDuration(timeout)
-	if timeoutContext != time.Duration(0) {
-		ctx, cancel = context.WithTimeout(ctx, timeoutContext)
+
+	if timeout != "" || taskRunDeadline != "" {
+		timeoutDuration := soonestTimeout(timeout, taskRunDeadline)
+		ctx, cancel = context.WithTimeout(ctx, timeoutDuration)
 		defer cancel()
 	}
 
@@ -78,4 +79,19 @@ func (rr *realRunner) Run(timeout string, args ...string) error {
 	}
 
 	return nil
+}
+
+// soonestTimeout accepts a step timeout duration and taskrun timeout deadline in string format
+// and returns whichever one is non-zero (if any) or that which will occur soonest.
+func soonestTimeout(stepTimeout, taskRunDeadline string) time.Duration {
+	taskRunEnd, _ := time.Parse(time.UnixDate, taskRunDeadline)
+	taskRunDuration := time.Until(taskRunEnd)
+
+	stepDuration, _ := time.ParseDuration(stepTimeout)
+
+	if stepTimeout == "" || taskRunEnd.Before(time.Now().Add(stepDuration)) {
+		return taskRunDuration
+	}
+
+	return stepDuration
 }
--- a/cmd/entrypoint/runner.go
+++ b/cmd/entrypoint/runner.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"context"
 	"os"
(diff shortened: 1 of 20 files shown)


cmd/entrypoint/runner.go
```diff
@@ -33,7 +35,25 @@ func (rr *realRunner) Run(args ...string) error {
...
+	ctx := context.Background()
+	var cancel context.CancelFunc
+
+	deadline, _ := time.Parse(time.UnixDate, taskRunDeadline)
+	taskRunDuration := time.Until(deadline)
+
+	stepDuration, _ := time.ParseDuration(timeout)
+	timeoutDuration := stepDuration
+
+	if timeoutDuration != time.Duration(0) && taskRunDeadline != "" && deadline.Before(time.Now().Add(timeoutDuration)) {
```
sbwsg, line 48: Does this not mean that the `taskRunDeadline` will only be applied if a Step timeout has also been set?  What happens if `timeout` is "0s" or ""?

A helper func might be useful here:

```go
// soonestTimeout accepts a step and taskrun duration and returns whichever one is non-zero (if any)
// or that which will occur soonest.
func soonestTimeout(step, taskRun time.Duration) time.Duration {
	if step == time.Duration(0) {
		return taskRun
	}
	if taskRun == time.Duration(0) {
		return step
	}
	if step < taskRun {
		return step
	}
	return taskRun
}
```

cmd/entrypoint/runner_test.go
```diff
@@ -14,9 +16,37 @@ func TestRealRunnerSignalForwarding(t *testing.T) {
...
+	} else {
+		t.Fatalf("step didn't timeout")
+	}
+}
+
+// TestRealRunnerDeadline tests whether cmd is killed after nanosecond * 1000, set by TaskRun deadline, even though it's supposed to sleep for 10 milliseconds.
+func TestRealRunnerDeadline(t *testing.T) {
+	rr := realRunner{}
+	timeout := "1s"
+	if err := rr.Run(timeout, time.Now().Add(time.Nanosecond * 100).Format(time.UnixDate), "sleep", "0.01"); err != nil {
```
sbwsg, line 45: It might be worth adding a couple of tests to codify how the function deals with edge cases like: invalid deadline string, negative deadline string, empty string.

config/config-feature-flags.yaml
```diff
@@ -40,14 +40,23 @@ data:
...
-  # working directory if not set by the user but this will change
-  # in an upcoming release.
+  # The default behaviour currently is for the TaskRun's timeout
+  # check is done at TaskRun's reconciler level insteaf of at
+  # Step level.
+  #
+  # See https://github.com/tektoncd/pipeline/issues/2559 for more
+  # info.
+  disable-taskrun-step-timeout-check: "true"
+  # Setting this flag to "true" will check Step's timeout against
```
sbwsg, line 53: I think we have lost the original comment for the `disable-working-directory-overwrite` flag here.

config/config-feature-flags.yaml
```diff
@@ -40,14 +40,23 @@ data:
...
-  # The default behaviour currently is for Tekton to override the
-  # working directory if not set by the user but this will change
-  # in an upcoming release.
+  # The default behaviour currently is for the TaskRun's timeout
+  # check is done at TaskRun's reconciler level insteaf of at
+  # Step level.
+  #
+  # See https://github.com/tektoncd/pipeline/issues/2559 for more
+  # info.
+  disable-taskrun-step-timeout-check: "true"
```
sbwsg, line 52: Suggest making this `enforce-taskrun-timeouts-in-step-entrypoint: "false"` to avoid the double-negative that occurs when `disable` and `false` together.

We've used `disable-` for many of the other feature flags mostly because they turn off features which existed before the feature flags were introduced.  Here though I think we're adding a new implementation and offering users the opportunity to test it out before we switch over to it permanently.

cmd/entrypoint/runner.go
ywluogg: Oops messed up the commit version. Moved them to a separate function

cmd/entrypoint/runner_test.go
ywluogg: Added a few more tests for Step timeout as well

config/config-feature-flags.yaml
ywluogg: Oops

PR #3308, opened 2020-09-30
Add a short example of using optional workspaces in when expressions

sbwsg:
We've just added support for optional workspaces but I forgot to include
an example showing how to use them as part of a When Expression.

This commit adds a PipelineRun example YAML showing use of a
workspaces.\<name>.bound variable in a when expression.

I've also added a line to the pipelines.md file mentioning that you
can use a when expression to evaluate whether an optional workspace was
bound or not.

# Changes

# Submitter Checklist

review them:

# Release Notes

```release-note
Added a short example of using optional workspaces in when expressions.
```

### DIFF
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -115,6 +115,7 @@ spec:
 For more information, see:
 - [Using `Workspaces` in `Pipelines`](workspaces.md#using-workspaces-in-pipelines)
 - The [`Workspaces` in a `PipelineRun`](../examples/v1beta1/pipelineruns/workspaces.yaml) code example
+- The [variables available in a `PipelineRun`](variables.md#variables-available-in-a-pipeline), including `workspaces.<name>.bound`.
 
 ## Specifying `Parameters`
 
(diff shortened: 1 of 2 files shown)


docs/pipelines.md
```diff
@@ -362,6 +362,7 @@ There are a lot of scenarios where `WhenExpressions` can be really useful. Some
 - Checking if a git file has changed in the previous commits
 - Checking if an image exists in the registry
 - Checking if the name of a CI job matches
+- Checking if an optional Workspace has been provided
```
jerop, line 365: I think we should add more detail that the workspace variable is available, in addition to Parameters and Results, in this part of the docs:
https://github.com/tektoncd/pipeline/blob/0b85cea1ff071f4769771a65b97d2684f73f2014/docs/pipelines.md#L324-L331

And include a snippet and reference the end-to-end example here:
https://github.com/tektoncd/pipeline/blob/0b85cea1ff071f4769771a65b97d2684f73f2014/docs/pipelines.md#L333-L355

jerop (commented):
the example looks great, just some suggestions on the docs

docs/pipelines.md
sbwsg: Nice thank you, I've updated the doc per your suggestions.

pritidesai (approved):
Looks great, thanks @sbwsg

examples/v1beta1/pipelineruns/using-optional-workspaces-in-when-expressions.yaml
```diff
@@ -0,0 +1,66 @@
...
+# in a when expression to selectively run different portions of a Pipeline
+# based on the presence of an optional workspace.
+#
+# In the PipelineRun below an optional message-of-the-day workspace is accepted
+# by the Pipeline. If that workspace is provided then the print-motd task is
+# executed. If that workspace is not provided then a print-default-motd task
+# is run instead. We supply a ConfigMap for the workspace and so the print-motd
+# task ends up running and printing the contents of each entry in the ConfigMap.
+apiVersion: v1
+kind: ConfigMap
```
pritidesai, line 11: Not sure how `configmap` is handled in our CI. 

`kubectl apply` followed by `kubectl delete` can not be used with `generateName`. With `kubectl create` and `generateName`, pipelineRun is assigned random name which can be deleted using `kubectl delete pr`  or `tkn pr delete` but leaves the `configmap` as is. Does our cleanup cron job, deletes such config maps? Does this concern even apply in our CI?

sbwsg: hm, it looks like we deploy them using `ko apply` all into the same namespace and cluster.

`ko apply` in library.sh: https://github.com/tektoncd/plumbing/blob/ef09ae0cb20c1b320b5e3f7a8707030a50d8be17/scripts/library.sh#L472-L480

i believe we run yaml suite in a brand new cluster though.  so the configmap is created and then quickly deleted when the cluster is cleaned up.

Issue #3310, opened 2020-10-01
TaskRun is failing to run/complete in some cases (on different namespace, …)


vdemeester:
# Expected Behavior

OpenShift Pipelines 1.0.1 installed on OpenShift 4.4.17 is failing to run the below `TaskRun` in some specific namespaces. While it's working on some namespaces it's failing in specific with the following error reported in `tekton-pipelines-controller`:

```
{
  "level": "error",
  "logger": "tekton.taskrun-controller",
  "caller": "taskrun/taskrun.go:163",
[...]

PR #3316, opened 2020-10-01
Derive cancel patch bytes once at controller startup

ImJasonH:
This avoids JSON marshaling the same object every time a PipelineRun is
cancelled, and marshals the object once at startup. If for some reason
marshaling that object fails (which it should never do), the controller
will exit and restart.

Failure to marshal this object should only be caused by a bug in our
code, since the object isn't derived from anything related to user
requests. If we have a bug in our cancel patch generation, we'd rather
find out by having the controller crashloop than find out by having
PipelineRuns be uncancellable, which might be harder to detect.

# Submitter Checklist

review them:

- [n] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)

## Reviewer Notes

# Release Notes

```release-note
Fail more loudly at controller startup when we fail to marshal the JSON Patch request to cancel TaskRuns owned by cancelled PipelineRuns.
```

cc @ywluogg since I think we should also generate patch bytes like this for updating annotations (#3310)

### DIFF
--- a/pkg/reconciler/pipelinerun/cancel.go
+++ b/pkg/reconciler/pipelinerun/cancel.go
@@ -19,37 +19,44 @@ package pipelinerun
 import (
 	"encoding/json"
 	"fmt"
+	"log"
 	"strings"
 	"time"
 
-	"go.uber.org/zap"
-	jsonpatch "gomodules.xyz/jsonpatch/v2"
-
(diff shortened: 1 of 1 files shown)


pritidesai (approved):
thanks @ImJasonH 👍

sbwsg (approved):

PR #3331, opened 2020-10-05
Replace Pod Update() with Pod Patch() to update Pod in place

ywluogg:
# Changes

Replace Pod Update() with Pod Patch() to update Pod in place to solve the issue that when Pod gets updated when the entrypoint tries to update the Pod's annotations.

Closes [#3310](https://github.com/tektoncd/pipeline/issues/3310)

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

### DIFF
--- a/pkg/pod/entrypoint.go
+++ b/pkg/pod/entrypoint.go
@@ -26,6 +26,9 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/client-go/kubernetes"
 )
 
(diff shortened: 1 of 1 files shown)


ImJasonH:
I think we can simplify this change. The patch we're trying to make is, effectively, "add the annotation `tekton.dev/ready: READY`" -- this value isn't dependent on any aspect of the Pod's current state, we're just saying "whatever the current state is, add this annotation key+value".

In that case, the patch can be expressed as:

```
[{
  "op": "add",
  "path": "/metadata/annotations/tekton.dev~1ready",
  "value":"READY"
}]'
```

(I found [this helpful SO question](https://stackoverflow.com/questions/55573724/create-a-patch-to-add-a-kubernetes-annotation) to explain the `~1` part)

Like we do with [generating the patch to cancel a TaskRun](https://github.com/tektoncd/pipeline/blob/025f57a04754902df665cec6d51c388a070a4c6c/pkg/reconciler/pipelinerun/cancel.go#L37), these patch bytes can be generated once at init, and [reused for all future patches](https://github.com/tektoncd/pipeline/blob/025f57a04754902df665cec6d51c388a070a4c6c/pkg/reconciler/pipelinerun/cancel.go#L60), along the lines of:

```go
if _, err := kubeclient.CoreV1().Pods(pod.Namespace).Patch(ctx, pod.Name, types.JSONPatchType, updateReadyPatchBytes, metav1.PatchOptions{}, ""); err != nil {
  return err
}
```

This avoids having to consider any aspect of the current state of the Pod -- you don't have to copy over existing annotations, or any Pod fields, you're just telling k8s to blindly apply this patch to the object, whatever its current state is.

vdemeester:
@ywluogg thanks for the PR. Looks like it has been fixed by @ImJasonH in #3494, so I'll close this one.

PR #3346, opened 2020-10-06
Avoid dangling symlinks in git-init

sbwsg:
fixes https://github.com/tektoncd/catalog/issues/531

# Changes

When the following conditions are met:

1. the feature flag disable-home-env-overwrite is "true"
2. the container user is root
3. no git / ssh secret is attached to a taskrun service account
4. user is running new-ish version of catalog git-clone task with git-init v0.15.2+

git-init will error out in the git-clone task because we create a circular symlink
from /root/.ssh to itself and then try to look up /root/.ssh/known_hosts.

This commit adds a check to avoid this from happening:

If the user's `$HOME/.ssh` directory doesn't exist or if they aren't able to access it
for any reason, then we don't try to create a symlink to it at all since we can trust that
the user is incapable of utilizing the credential.

This commit also expands an existing check to see if the `$HOME/.ssh` directory is
the same as the user's home directory + '.ssh'.  This was originally only checked if
the user was nonroot, but now this is also checked if the user is root too.

# Submitter Checklist

review them:

# Release Notes

```release-note
Fixed a bug in `git-init` that allowed a circular symlink to be created from `/root/.ssh` to itself if no SSH credentials are present in the service account and the `disable-home-env-overwrite` flag is set to `"true"`.
```

/kind bug

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -206,9 +206,11 @@ func SubmoduleFetch(logger *zap.SugaredLogger, spec FetchSpec) error {
 	return nil
 }
 
+// ensureHomeEnv works around an issue where ssh doesn't respect the HOME env variable. If HOME is set and
+// different from the user's detected home directory then symlink .ssh from the home directory to the HOME env
+// var. This way ssh will see the .ssh directory in the user's home directory even though it ignores
+// the HOME env var.
 func ensureHomeEnv(logger *zap.SugaredLogger) error {
-	// HACK: This is to get git+ssh to work since ssh doesn't respect the HOME
-	// env variable.
 	homepath, err := homedir.Dir()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


vdemeester (commented):
/lgtm
/cc @chmouel

sbwsg:
Updated to use `filepath` package when joining path strings together, and to `filepath.Clean` directory names when comparing them to each other.

vdemeester (approved):
@tektoncd/core-maintainers should we consider this for 0.17.1 ?

sbwsg:
@vdemeester I think so, but curious to hear what others think.

sbwsg:
OK, I think that's enough time to solicit feedback from anyone else.  Let's indeed cherry-pick this into 0.17.

/label needs-cherry-pick

Issue #3303, opened 2020-09-29
Webhook log has repeated TLS handshake error EOF messages


GregDritschler:
# Expected Behavior

The webhook log does not contain errors during normal operation.

# Actual Behavior

The webhook log contains the following error repeated every few seconds:

```
2020/09/29 19:14:53 http: TLS handshake error from 10.104.2.1:57866: EOF
2020/09/29 19:15:01 http: TLS handshake error from 10.104.2.1:57884: EOF
2020/09/29 19:15:03 http: TLS handshake error from 10.104.2.1:57888: EOF
```

# Steps to Reproduce the Problem

* Find the `tekton-pipelines-webhook` pod and display its log.

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
    Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.0", 
    GitCommit:"ddf47ac13c1a9483ea035a79cd7c10005ff21a6d", GitTreeState:"clean", BuildDate:"2018-12-03T21:04:45Z", 
    GoVersion:"go1.11.2", Compiler:"gc", Platform:"windows/amd64"}
    Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.15+IKS", 
    GitCommit:"a7ace239946d71c9305593eacf74b9a9efa25e3b", GitTreeState:"clean", BuildDate:"2020-09-09T00:11:53Z", 
    GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version:

    ```
    Pipeline version: v0.16.3
    ```

The error is caused by the liveness and readiness probes that were added by PR #3162.  Manually removing those probes from the webhook deployment and restarting the webhook pod stops the errors.

From the client perspective, the probe is working.  It is a tcp socket probe so it is connecting to the socket successfully and then disconnecting.  From the server perspective, it appears that the golang tls library is complaining that a client connected and then did not perform the tls handshake, hence the "EOF" message.

Although the probes are working, the logs are filling up with messages that make it appear like something is wrong with the webhook.

GregDritschler:
Hi  @ywluogg.   As the original contributor of the liveness and readiness probes can you please investigate and resolve this issue?  Thank you.

ywluogg:
Taking a look now

ywluogg:
[...]

Issue #3111, opened 2020-08-18
As a tekton pipeline user, I want to use liveness/readiness probes to check controller pod health


xiujuan95:
### Feature request

I want to use liveness and readiness probes to detect if my tekton controller pod is healthy or not. However, seem like liveness and readiness field don't be included in controller deployment:https://github.com/tektoncd/pipeline/blob/master/config/controller.yaml

About this request, actually, I have done some experiments. I configure liveness and readiness like below:
```
livenessProbe:
          failureThreshold: 3
          httpGet:
            path: /metrics
            port: 9090
            scheme: HTTP
          initialDelaySeconds: 5
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 1
readinessProbe:
          failureThreshold: 3
          httpGet:
            path: /metrics
            port: 9090
            scheme: HTTP
          initialDelaySeconds: 5
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 1
```

And I check the pod event, it tells me the probes failed:
```
Events:
  Type     Reason     Age                From                  Message
  ----     ------     ----               ----                  -------
  Normal   Scheduled  35m                default-scheduler     Successfully assigned tekton-pipelines/tekton-pipelines-controller-7cd74569b7-mm96v to 10.242.0.19
  Normal   Pulled     35m                kubelet, 10.242.0.19  Container image "icr.io/obs/codeengine/tekton-pipeline/controller-10a3e32792f33651396d02b6855a6e36:v0.14.2-rc2@sha256:845358c3bb68e6900b421545ee40352391baccb751ef63915758016c4745bdbe" already present on machine
  Normal   Created    35m                kubelet, 10.242.0.19  Created container tekton-pipelines-controller
  Normal   Started    35m                kubelet, 10.242.0.19  Started container tekton-pipelines-controller
  Warning  Unhealthy  35m (x2 over 35m)  kubelet, 10.242.0.19  Readiness probe failed: Get http://172.30.18.233:9090/metrics: dial tcp 172.30.18.233:9090: connect: connection refused
  Warning  Unhealthy  35m (x2 over 35m)  kubelet, 10.242.0.19  Liveness probe failed: Get http://172.30.18.233:9090/metrics: dial tcp 172.30.18.233:9090: connect: connection refused
```
However, my pod is still running normally:
```
kubectl get pod -n tekton-pipelines -o wide
[...]

PR #3349, opened 2020-10-07
Switch webhook liveness/readiness probes to use http ports

ywluogg:
# Changes

Switch webhook liveness/readiness probes to http ports and use a separate port from actual webhook HTTPS ports.
This fixes misleading errors mentioned in [#3303](https://github.com/tektoncd/pipeline/issues/3303)

Webhook logs:
```
kubectl -n tekton-pipelines logs $(kubectl -n tekton-pipelines get pods -l app=tekton-pipelines-webhook -o name)
2020/10/06 23:57:31 maxprocs: Leaving GOMAXPROCS=2: CPU quota undefined
2020/10/06 23:57:31 Server listening on port 8080
```

This solves the issue because it doesn't send simple TCP request to a HTTPS port, which will cause TLS handshake to fail.

Other considerations will be 1. continue using the same HTTPS port but establish HTTPS handshake. This is not done because using the same port as actual webhook port for probes is not ideal for traffic and responsibility separation. 

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -18,6 +18,8 @@ package main
 
 import (
 	"context"
+	"log"
+	"net/http"
 	"os"
 
 	defaultconfig "github.com/tektoncd/pipeline/pkg/apis/config"
@@ -213,6 +215,23 @@ func main() {
 		SecretName:  secretName,
 	})
 
+	mux := http.NewServeMux()
+
+	mux.HandleFunc("/", handler)
+	mux.HandleFunc("/health", handler)
+	mux.HandleFunc("/readiness", handler)
+
+	port := os.Getenv("PROBES_PORT")
+	if port == "" {
+		port = "8080"
+	}
+
+	go func() {
+		// start the web server on port and accept requests
+		log.Printf("Readiness and health check server listening on port %s", port)
+		log.Fatal(http.ListenAndServe(":"+port, mux))
+	}()
+
 	sharedmain.WebhookMainWithConfig(ctx, "webhook",
 		sharedmain.ParseAndGetConfigOrDie(),
 		certificates.NewController,
@@ -221,4 +240,8 @@ func main() {
 		newConfigValidationController,
 		newConversionController,
 	)
+}
+
+func handler(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(http.StatusOK)
 }
--- a/config/webhook.yaml
+++ b/config/webhook.yaml
@@ -91,17 +91,24 @@ spec:
           containerPort: 8008
         - name: https-webhook
           containerPort: 8443
+        - name: probes
+          containerPort: 8080
         livenessProbe:
-          tcpSocket:
-            port: https-webhook
+          httpGet:
+            path: /health
+            port: probes
+            scheme: HTTP
           initialDelaySeconds: 5
           periodSeconds: 10
           timeoutSeconds: 5
         readinessProbe:
-          tcpSocket:
-            port: https-webhook
+          httpGet:
+            path: /readiness
+            port: probes
+            scheme: HTTP
           initialDelaySeconds: 5
           periodSeconds: 10
+          timeoutSeconds: 5
 ---
 apiVersion: v1
 kind: Service
@@ -131,6 +138,8 @@ spec:
   - name: https-webhook
     port: 443
     targetPort: 8443
+  - name: probes
+    port: 8080
   selector:
     app.kubernetes.io/name: webhook
     app.kubernetes.io/component: webhook


jerop (commented):
/kind bug 

please consider adding more detail in the commit message as described in https://github.com/tektoncd/community/blob/master/standards.md#commit-messages

cmd/webhook/main.go
```diff
@@ -213,6 +215,22 @@ func main() {
...
+	mux.HandleFunc("/readiness", readinessHandler)
+
+	// use PORT environment variable, or default to 8080
+	port := os.Getenv("PORT")
+	if port == "" {
+		port = "8080"
+	}
+
+	// start the web server on port and accept requests
+	log.Printf("Server listening on port %s", port)
```
sbwsg, line 231: Nit: suggest `Readiness and health check server listening on port %s`

cmd/webhook/main.go
```diff
@@ -213,6 +215,22 @@ func main() {
...
 	})
 
+	mux := http.NewServeMux()
+
+	mux.HandleFunc("/", handler)
+	mux.HandleFunc("/health", healthHandler)
+	mux.HandleFunc("/readiness", readinessHandler)
+
+	// use PORT environment variable, or default to 8080
+	port := os.Getenv("PORT")
```
sbwsg, line 225: What do you think about choosing a more descriptive env var name such as `PROBES_PORT` or similar?

cmd/webhook/main.go
```diff
@@ -213,6 +215,22 @@ func main() {
 		SecretName:  secretName,
 	})
 
+	mux := http.NewServeMux()
+
+	mux.HandleFunc("/", handler)
+	mux.HandleFunc("/health", healthHandler)
+	mux.HandleFunc("/readiness", readinessHandler)
+
+	// use PORT environment variable, or default to 8080
```
sbwsg, line 224: The code below this line is self-explanatory.  I don't think the comment is necessary.

cmd/webhook/main.go
```diff
@@ -213,6 +215,22 @@ func main() {
...
+	port := os.Getenv("PORT")
+	if port == "" {
+		port = "8080"
+	}
+
+	// start the web server on port and accept requests
+	log.Printf("Server listening on port %s", port)
+	log.Fatal(http.ListenAndServe(":"+port, mux))
```
sbwsg, line 232: I think this will block and `sharedmain.WebhookMainWithConfig` won't run until `ListenAndServe` returns.  Though I might be missing something?

cmd/webhook/main.go
```diff
@@ -222,3 +240,15 @@ func main() {
 		newConversionController,
 	)
 }
+
+func handler(w http.ResponseWriter, r *http.Request) {
```
sbwsg, line 244: Given that all three of these funcs are the same I suggest only having one until such time as we need to provide different functionality.  What do you think?

ywluogg: I was struggling deciding whether to separate them to make the code clearer or just use the same one for different paths. Yeah now I'm more agreed with yours.

cmd/webhook/main.go
ywluogg: Ah oops forgot to do it concurrently.

config/webhook.yaml
```diff
@@ -91,17 +91,24 @@ spec:
...
         readinessProbe:
-          tcpSocket:
-            port: https-webhook
+          httpGet:
+            path: /readiness
+            port: probes
+            scheme: HTTP
           initialDelaySeconds: 5
           periodSeconds: 10
+          timeoutSeconds: 5
```
afrittoli, line 111: Given that the check is a `no-op`, having a 5s timeout sounds good.

afrittoli (commented):
Thanks for this!
Eventually it would be nice to have probes that actually verify that the service is running - however this is definitely a step forward!
It would be great to do the same for the controller (in a different PR) if you'd like.
/lgtm

ywluogg:
Sorry I was distracted by something else for a bit and now was able to come back to this. @afrittoli I think @ImJasonH is still not convinced to add the probes in controller and I also think it will be reasonable to follow the comment you added in #3111. I will keep track of the health checks they've added to knative package and take a look at that to make probes available in controller.

> @mattmoor @pmorie I found knative/pkg#1048 but it's not clear to me then whether it is available yet or not. If not we could perhaps add a "shallow" /healthz for now, that always reports "ok" (as @ImJasonH suggested) and switch to the knative/pkg one once it becomes available.

ImJasonH:
/lgtm

@sbwsg do you have any other comments?

sbwsg:
Thanks @ywluogg !

/approve

qu1queee:
@ywluogg quick question, would be ok to have this same implementation for the missing probes in https://github.com/tektoncd/pipeline/blob/master/config/controller.yaml ?

ywluogg:
> @ywluogg quick question, would be ok to have this same implementation for the missing probes in https://github.com/tektoncd/pipeline/blob/master/config/controller.yaml ?

The issue was addressed in [#3111](https://github.com/tektoncd/pipeline/issues/3111). I will add the probes in controller once the health checks mentioned in #3111 is available.

afrittoli:
The approach in knative for the webhook is actually to use the existing HTTPs endpoint like used to do, but with HTTPs instead of TCP, and passing a special `k-kubelet-probe` header:

https://github.com/knative/serving/blob/1d259fb640b5ba9748a430a0cfa90d451d5826d2/config/core/deployments/webhook.yaml#L95-L102 

I guess we could do the same for Tekton and avoid the need for the extra server?
For the controller there is no solution yet, so one way forward might be to have a dedicated small HTTP server until we have a solution from `knative/pkg`.

mattmoor:
You don't need the `k-kubelet-probe` header, this is sort of a vestige of a time when Istio incorrectly rewrote K8s probes to drop the header *it* sends.  Maybe it's a special `User-Agent`?  I think we have an `IsKubeletProbe` function somewhere that checks for these things.

ywluogg:
> The approach in knative for the webhook is actually to use the existing HTTPs endpoint like used to do, but with HTTPs instead of TCP, and passing a special `k-kubelet-probe` header:
> 
> https://github.com/knative/serving/blob/1d259fb640b5ba9748a430a0cfa90d451d5826d2/config/core/deployments/webhook.yaml#L95-L102
> 
> I guess we could do the same for Tekton and avoid the need for the extra server, but it would require doing TLS handshakes, so perhaps its good to stick with this solution :)
> 
> For the controller there is no solution yet, so one way forward might be to have a dedicated small HTTP server until we have a solution from `knative/pkg`. [knative/pkg#1048](https://github.com/knative/pkg/issues/1048) is back open

Pinging @ImJasonH again for any suggestion for adding a temporary http server in controller. I think currently we may still need for info from other users about why they need this probe in the controller.

afrittoli:
The probe needed is mostly the liveness one.
When running multiple replicas through leader election, the probe helps detecting a faulty controller and releasing its bucket of resources to other controllers.

ywluogg:
It sounds necessary to add at least the liveness probe in controller based on this situation. I don't think there is another simpler way to check on liveness under this situation. But I'm waiting for @ImJasonH con re-assurance.

ImJasonH:
I agree, it sounds like HA controller will benefit from having a liveness probe, which we can implement now with a simple dummy HTTP server, and later improve with real health checks.

ywluogg:
Sounds good will add it as soon as possible.

xiujuan95:
@ywluogg Go through above discussions, are you only enable `liveness` probe for controller, now? Could you pls also add readiness probe for controller? It's necessary for me. Thanks in advance!

ywluogg:
@xiujuan95 do you mind also provide some details about the situation you plan to use it? Thanks

xiujuan95:
@ywluogg Sure, I can. Actually, It's a requirement from my manager team. It requires all pods need to enable liveness and readiness check. Besides, I think `readiness` can tell us if all containers within this pod are ready or not. It also makes sense! So pls also help add it! Thanks in advance!

ywluogg:
Sounds good I will add them too.

xiujuan95:
@ywluogg Thanks a lot!

PR #3350, opened 2020-10-07
[Automated] Upgrade to latest dependencies

mattmoor:
Produced via:
```shell
./hack/update-deps.sh --upgrade
./hack/update-codegen.sh
```
/cc vdemeester ImJasonH
/assign vdemeester ImJasonH


mattmoor:
🤔  @vdemeester more `third_party/` churn.  I wonder if my local box has some skew with HEAD, but I'm surprised this is getting through CI both ways...

PR #3355, opened 2020-10-09
[master] Upgrade to latest dependencies

mattmoor:
Produced via:
```shell
./hack/update-deps.sh --upgrade
./hack/update-codegen.sh
```
/cc vdemeester ImJasonH
/assign vdemeester ImJasonH


ImJasonH:
This deleted [third_party/LICENSE](https://github.com/tektoncd/pipeline/blob/master/third_party/LICENSE), presumably because it also removed our dependency on some golang.org/x/ packages. Just making sure to call this out in case this was unexpected.

mattmoor:
@ImJasonH I think this is a sign that someone needs to update:
```
go get github.com/google/go-licenses
```

The actions do a fresh `go get` each run, so they should reflect the latest.

mattmoor:
switching bots

PR #3379, opened 2020-10-13
Add feature flag to disable creds-init

sbwsg:
Fixes https://github.com/tektoncd/pipeline/issues/2791

Prior to this commit Tekton scans service accounts attached to taskruns
and mounts secrets matching a specific format into all the Steps of a
Task. Then the entrypoint goes through those secrets and copies them
to the user's home directory so they can be used by `git`, `docker`, etc

This commit adds a feature-flag, `disable-creds-init`, that lets users
turn off the service account scanning behaviour. The default is "false"
so the creds-init behaviour remains enabled. Disabling creds-init can be
desirable for a lot of reasons:

1. The process is opaque and difficult to debug
2. It's not particularly well suited for non-root, multi-UID tasks.
3. Only a limited set of credential types are supported

Warning: Disabling creds-init will probably break PipelineResources.

# Changes

# Submitter Checklist

review them:

- [x] Includes tests

# Release Notes

```release-note
Tekton's built-in credential mechanism can now be disabled by setting the `disable-creds-init` feature-flag to "true".
```

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -169,10 +169,14 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec 
 	for i, s := range stepContainers {
 		// Mount /tekton/creds with a fresh volume for each Step. It needs to
 		// be world-writeable and empty so creds can be initialized in there. Cant
-		// guarantee what UID container runs with.
-		v, vm := getCredsInitVolume()
-		volumes = append(volumes, v)
-		s.VolumeMounts = append(s.VolumeMounts, vm)
+		// guarantee what UID container runs with. If creds-init is disabled via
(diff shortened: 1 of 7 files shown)


pkg/pod/pod.go
```diff
@@ -170,9 +170,11 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec
 		// Mount /tekton/creds with a fresh volume for each Step. It needs to
 		// be world-writeable and empty so creds can be initialized in there. Cant
 		// guarantee what UID container runs with.
```
jerop, line 172: may be useful to update this to mention that it mounts fresh volume for each step _except when creds-init is disabled_

sbwsg:
@jerop thanks a lot for the feedback! I've made the changes you requested.

vdemeester (approved):

Issue #3152, opened 2020-09-01
Add variable expansion in the workspaces.mountPath field of Tasks


coryrc:
# Expected Behavior

If a Task has `spec.workspaces[].mountPath: "$(params.dest-dir)"` then said workspace should be mounted to the expanded value; i.e. if dest-dir is `/tmp`.

I believe it is supposed to be because I see examples of it used [in the catalog](https://github.com/tektoncd/catalog/blob/a9600e084a52cf54e578a316aff2f11f5731a98c/task/golang-test/0.1/golang-test.yaml#L38).

# Actual Behavior

It gets mounted to the literal directory `/$(params.dest-dir)`, as in this line from `mount`:
`[unit-tests : execute] /dev/sdb on /$(params.dest-dir) type ext4 (rw,relatime)`

# Steps to Reproduce the Problem

I believe it should fail to even be `kubectl create`'d if it is interpreting these right, so `kubectl create -f` on this file should fail if it is working correctly:

```yaml
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: no-mountpath-expansion
spec:
  workspaces:
    - name: source
      mountPath: "$(params.invalid-parameter)"
  params:
    - name: valid-parameter
  steps:
    - image: debian:buster
      workingDir: $(params.valid-parameter)
      script: |
        echo hello world                                                                                                                                         
```

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-13T16:12:48Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.13-gke.1", GitCommit:"688c6543aa4b285355723f100302d80431e411cc", GitTreeState:"clean", BuildDate:"2020-07-21T02:37:26Z", GoVersion:"go1.13.9b4", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

Client version: 0.12.0
Pipeline version: v0.15.2
Triggers version: unknown

sbwsg:
[...]

PR #3380, opened 2020-10-13
Add variable expansion in the workspaces.mountPath field of Tasks #3152

khalkie:
# Changes

* included Workspace in ApplyReplacements logic so any variables will be expanded

* added another Task Validation operation in createPod, which runs after the variable expansion - it may be possible that the Workspaces are no longer valid (i.e. have duplicate names) after the variable expansion

Fixes: #3786

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Users are able to use variable substitution in the mountPath field of a Workspace within a Task
```


sbwsg:
Commits should be squashed into 1 using `git rebase -i HEAD^^^^^` or something similar.

Suggest also updating [`docs/variables.md`](https://github.com/tektoncd/pipeline/blob/master/docs/variables.md#fields-that-accept-variable-substitutions) to include `workspaces[].mountPath` in the list of places variables can be used.

I have to confess I'm still a tad uneasy allowing a field to accept variables when the field itself can also be used as a variable elsewhere.  However I pulled down the branch and couldn't get anything to break except when doing something really silly like introducing a cycle by using a workspace path variable in a param that I then also used inside the workspace's mountPath field so I am ok with this going in 👍.

sbwsg:
Thinking about this some more yesterday evening:

- it might also be worth including a description of this variable support to our [`workspaces.md`](https://github.com/tektoncd/pipeline/blob/master/docs/workspaces.md) doc.
- We should probably add a code comment above the new call to Validate with something like, "perform another validation pass because...".

I also wonder if there's some kind of test we can construct that codifies the accepted/supported ordering of variable applications. This might be a little out of scope for this specific PR but I wonder if we could build a table of variables with various dependencies on each other and check that they replace in the right way.  This would be try and catch errors introduced if another dev ever comes along and reorders the variable Apply sequencing.

vdemeester:
> I have to confess I'm still a tad uneasy allowing a field to accept variables when the field itself can also be used as a variable elsewhere. However I pulled down the branch and couldn't get anything to break except when doing something really silly like introducing a cycle by using a workspace path variable in a param that I then also used inside the workspace's mountPath field so I am ok with this going in +1.

I am not sure I follow, the variable for workspaces is `workspaces.<workspaceName>.path`, which is not _really_ `mountPath`. It "derives" from `mountPath` but is not directly the same variable. That said, I do agree that we need to make sure this doesn't break the variable interpolation "work".

bobcatfish:
i think we'd want to update https://github.com/tektoncd/pipeline/blob/master/docs/variables.md#fields-that-accept-variable-substitutions as well

afrittoli:
@khalkie are you still working on this? 
I think it would be a really useful feature to have, some `go` tasks in the catalog might benefit from it.

khalkie:
Sorry so much for the delay!  I recreated this fix in another PR - https://github.com/tektoncd/pipeline/pull/3934

Issue #3378, opened 2020-10-13
PipelineRun fails too eagerly on missing resources


mattmoor:
# Expected Behavior

PipelineRun only fails when referenced resources do not exist.

# Actual Behavior

PipelineRun fails when referenced resources do not YET exist in it's informer cache at the time of reconciliation.

You can see the bug here as the `lister` is passed to fetch the resource, and on error it manifests as a permanent failure of the pipeline resource: https://github.com/tektoncd/pipeline/blob/1710b688c0238d21ebf43cd637b79e3cb627f054/pkg/reconciler/pipelinerun/pipelinerun.go#L334-L341

# Steps to Reproduce the Problem

This is an intermittent flake in the e2e testing, which manifests with a message like:
```
PipelineRun arendelle-8mqgj/git-check-pipeline-run can''t be Run; it tries to bind Resources that don''t exist: error following resource reference for git-repo: pipelineresource.tekton.dev "git-source-resource" not found
```

# Additional Info

This is roughly at HEAD (I observed this downstream, but the issue is clearly upstream)

mattmoor:
Slack thread: https://tektoncd.slack.com/archives/CLCCEBUMU/p1602598837131500

GregDritschler:
This is very closely related to #2740.  The pipelinerun and taskrun reconcilers use the client to get the Pipeline or Task respectively to avoid lister cache issues.  There was a lot of good discussion in that issue.  I'm not sure why it was closed.

mattmoor:
There is a lot of nuance, and honestly there are facets of the Tekton resource model that make this challenging (as I highlighted in that thread).

The conflict here is between:
1. fast-feedback when a user fat-fingers a bad reference
2. working in the presence of slow informer syncing

Using the client could solve both, but then puts things into conflict with scaling the system.

I think an appropriate solution here would be to give the informer caches a grace period before declaring missing references fatal.  This could key off of `metadata.creationTimestamp`.

mattmoor:
I have a change, which will hopefully build in a grace period for this here: https://github.com/tektoncd/pipeline/pull/3385

Let's see how the e2e tests respond to this 🤞

LiYa-Xu:
Hi  all ,I am not sure I met a same issue. The e2e test case fails on case `TestTaskRunPipelineRunCancel`, I got message like
```
CONT  TestTaskRunPipelineRunCancel/retries=1
[...]

PR #3385, opened 2020-10-14
Don't fail immediately on missing resources.

mattmoor:
# Changes

As the PipelineRun reconciler executes, it resolves resources using the informer's lister cache.  Currently, when that cache is behind the pipeline run will immediately fail.  This change builds in a buffer of `resources.MinimumAge` and a helper `resources.IsYoung` that elide this check, returning the error to the controller framework to requeue the key for later processing (with backoff).

Fixes: https://github.com/tektoncd/pipeline/issues/3378

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Add a grace period for resources to appear before failing *Runs
```

### DIFF
--- a/pkg/reconciler/resources.go
+++ b/pkg/reconciler/resources.go
@@ -23,11 +23,11 @@ import (
 )
 
 const (
-	// MinimumResourceAge is the age at which resources stop being IsYoungResource.
-	MinimumResourceAge = 30 * time.Second
+	// minimumResourceAge is the age at which resources stop being IsYoungResource.
+	minimumResourceAge = 30 * time.Second
 )
 
-// IsYoungResource checks whether the resource is younger than MinimumResourceAge, based on its creation timestamp.
+// IsYoungResource checks whether the resource is younger than minimumResourceAge, based on its creation timestamp.
 func IsYoungResource(obj kmeta.Accessor) bool {
-	return time.Now().Sub(obj.GetCreationTimestamp().Time) < MinimumResourceAge
+	return time.Now().Sub(obj.GetCreationTimestamp().Time) < minimumResourceAge
 }
--- a/pkg/reconciler/resources.go
+++ b/pkg/reconciler/resources.go
@@ -1,0 +1,33 @@
+/*
+Copyright 2019 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reconciler
+
+import (
+	"time"
+
+	"knative.dev/pkg/kmeta"
+)
+
+const (
+	// MinimumResourceAge is the age at which resources stop being IsYoungResource.
+	MinimumResourceAge = 30 * time.Second
+)
+
+// IsYoungResource checks whether the resource is younger than MinimumResourceAge, based on its creation timestamp.
+func IsYoungResource(obj kmeta.Accessor) bool {
+	return time.Now().Sub(obj.GetCreationTimestamp().Time) < MinimumResourceAge
+}
--- a/pkg/reconciler/resources.go
+++ b/pkg/reconciler/resources.go
@@ -24,7 +24,7 @@ import (
 
 const (
 	// minimumResourceAge is the age at which resources stop being IsYoungResource.
-	minimumResourceAge = 30 * time.Second
+	minimumResourceAge = 5 * time.Second
 )
 
 // IsYoungResource checks whether the resource is younger than minimumResourceAge, based on its creation timestamp.
--- a/pkg/apis/pipeline/v1beta1/taskrun_types.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_types.go
@@ -148,6 +148,17 @@ func (trs *TaskRunStatus) MarkResourceNotConvertible(err *CannotConvertError) {
 	})
 }
 
+// MarkResourceOngoing sets the ConditionSucceeded condition to ConditionUnknown
+// with the reason and message.
+func (trs *TaskRunStatus) MarkResourceOngoing(reason TaskRunReason, message string) {
+	taskRunCondSet.Manage(trs).SetCondition(apis.Condition{
+		Type:    apis.ConditionSucceeded,
+		Status:  corev1.ConditionUnknown,
+		Reason:  reason.String(),
+		Message: message,
+	})
+}
+
 // MarkResourceFailed sets the ConditionSucceeded condition to ConditionFalse
 // based on an error that occurred and a reason
 func (trs *TaskRunStatus) MarkResourceFailed(reason TaskRunReason, err error) {
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -38,6 +38,7 @@ import (
 	listers "github.com/tektoncd/pipeline/pkg/client/listers/pipeline/v1beta1"
 	resourcelisters "github.com/tektoncd/pipeline/pkg/client/resource/listers/resource/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/contexts"
+	tknreconciler "github.com/tektoncd/pipeline/pkg/reconciler"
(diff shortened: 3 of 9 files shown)


pkg/reconciler/resources.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"time"
+
+	"knative.dev/pkg/kmeta"
+)
+
+const (
+	// MinimumResourceAge is the age at which resources stop being IsYoungResource.
+	MinimumResourceAge = 30 * time.Second
```
ImJasonH, line 27: Does this need to be exported?

ImJasonH (commented):
/lgtm

pkg/reconciler/resources.go
mattmoor: technically no, but I thought it might be useful if folks needed to use it for unit testing.T

mattmoor:
cc @jlpettersson @afrittoli @vdemeester

pkg/reconciler/resources.go
mattmoor: Wow, the `T` was a typo, but it just worked out so well I'm not even going to fix it 😂

afrittoli (commented):
Thanks for this, it sounds like a good idea.

This benefits the ability to run for uses cases where task / taskrun (as well as pipeline / pipelinerun) are created at the same time and adds a bit of pain when authoring pipelines as typos and other similar issues will take 30s longer to emerge; it's probably a fair trade.

Eventually we might want to expose the definition of "young" in a config or start-up flag, so that it can be set to zero in development environments.

I wonder how this will interact with tekton bundles? @pierretasci @vdemeester

pkg/reconciler/resources.go
ImJasonH: If it seems like it will be useful in the future we can export it then, in general I don't like to export things unless we know we want to make it part of the public Go API we expose to users.

mattmoor: done

vdemeester:
> I wonder how this will interact with tekton bundles? @pierretasci @vdemeester

This might in a good way, as we could re-use that "young" state when pulling the bundle takes time. Other than this, I think it's ok :wink: 

/lgtm

bobcatfish:
> This benefits the ability to run for uses cases where task / taskrun (as well as pipeline / pipelinerun) are created at the same time and adds a bit of pain when authoring pipelines as typos and other similar issues will take 30s longer to emerge; it's probably a fair trade.

Can we explore this a bit more? I think 30 seconds might be a bit generous here - maybe we can make it significantly shorter, e.g. 1 second?

Otherwise this is a pretty big shift in expectations imo; before this change, making a mistake in a Run would fail immediately, now you need to know to check back in 30 seconds. Additionally, I don't think the status will reflect what is going on, is that right? Would you be able to tell the difference b/w this being re-queued and a reconciler being down?

I'm also wondering: when will  IsYoungResource realistically return false for a PipelineRun? Since PipelineRuns are always executed when created, it seems like they will always be "young"? And in fact, what happens if a Pipeline is executing, and a resource it needs is deleted within the first 30 seconds of execution? This would result in the already running PipelineRun being "suspended", i.e. would be re-queued, which is probably not what we want?

I think there is a bigger discussion to have here about what behavior we want when referenced resources are missing.

/hold

mattmoor:
@bobcatfish my expectation is sort of that *working* pipelines do not fail, which is a bigger violation of the principle of least surprise IMO 😉 

Ultimately, I think any non-zero value here is an improvement, so I'm happy to adjust it.  There isn't going to be a "right" answer here, it'll be trade-offs based on SLAs, and I don't know if/what SLA exists for when K8s responds to watches (nor am I sure that even that's a perfect approximation).

mattmoor:
Outside of very narrow use cases, using the client for gets is basically guaranteed to become a problem at scale.

`resources.GetResourcesFromBindings` is called on the core path through `reconcile` which means if you had 1000 resources, a global resync (e.g. rolling out a new controller, failover) then you need to eat 1000 API calls before your workqueue is free to do any real work.  The default client-side rate limit is 5 QPS, which means it'd take roughly 200s to do nothing.

I'm not sure how stressful the Tekton e2e tests are, but running the "chaos duck" during e2e tests in Knative uncovered a bunch of these.  It kills controllers every 30s, so if you choke on failovers you'll never survive.

bobcatfish:
> Outside of very narrow use cases, using the client for gets is basically guaranteed to become a problem at scale.

That's a good point however we are already using gets for Tasks and Pipelines - if we're seeing a performance problem there, i suggest we tackle that performance problem directly.

In the meantime it seems reasonable to me that we update the PipelineResource fetching logic to match how we're retrieving Pipelines and Tasks?

mattmoor:
> we are already using gets for Tasks and Pipelines

Those gets are done O(once in the lifecycle of a *Run) to snapshot the Task/Pipeline, where this is O(per reconcile).

"Just do a Get through the client" is literally asking me to introduce what I know is a scale bug, so I'm not going to do it, sorry.


If you want to make a more invasive change to bring the order of API calls more in line with Task/Pipeline, then be my guest, but it's a bigger change than I signed up for, and I still think that all of the API traffic for reads is worrisome.  Feel free to close if you want something else entirely, as I don't think that change would benefit from this.

bobcatfish:
> Those gets are done O(once in the lifecycle of a *Run) to snapshot the Task/Pipeline, where this is O(per reconcile).

Ah, because since we started storing the Task / Pipeline spec in the Run itself so we can grab it from there afterward.

I wonder if that means that the most consistent approach here would be to do the same with the Resource spec? (retrieve with Get, store in the Run) I'm surprised we didn't do that already, I'm guessing it's because of the uncertain state of PipelineResources.

In that case, if we treat this as more of a stop gap until we have decided the fate of PipelineResources that seems reasonable to me - I also misunderstood the PR initially and thought we were doing this for all "resource" retrieval, not just PipelineResources. Maybe we can lower the tolerance significantly? Talking with @ImJasonH he had suggested 5 seconds, i wonder if we could get away with something even shorter but 5 seconds seems okay.

Unless we think it's reasonable to start dereferencing the PipelineResource spec into the status now ( @vdemeester @afrittoli )

bobcatfish:
Oh and one more question: would it make sense to update the Run status in this case? Esp. if we go with the 30 second window, if I'm understanding correctly, a user wouldn't be able to tell the difference between a Run that is being retried b/c its PipelineResources don't exist, and a Run that isn't being reconciled at all?

vdemeester:
> > Those gets are done O(once in the lifecycle of a *Run) to snapshot the Task/Pipeline, where this is O(per reconcile).
> 
> Ah, because since we started storing the Task / Pipeline spec in the Run itself so we can grab it from there afterward.
> 
> I wonder if that means that the most consistent approach here would be to do the same with the Resource spec? (retrieve with Get, store in the Run) I'm surprised we didn't do that already, I'm guessing it's because of the uncertain state of PipelineResources.
> 
> In that case, if we treat this as more of a stop gap until we have decided the fate of PipelineResources that seems reasonable to me - I also misunderstood the PR initially and thought we were doing this for all "resource" retrieval, not just PipelineResources. Maybe we can lower the tolerance significantly? Talking with @ImJasonH he had suggested 5 seconds, i wonder if we could get away with something even shorter but 5 seconds seems okay.

I am fine starting with 5s and see if we need to refine this later on (and have a configuration for it so that users/integration can customize that period on time).
 
> Unless we think it's reasonable to start dereferencing the PipelineResource spec into the status now ( @vdemeester @afrittoli )

I would rather not do it, at least not until we are sure of what we are doing with `PipelineResource`.

mattmoor:
Alright, back from long weekend 🏝️  😎 

> would it make sense to update the Run status in this case

Sure, I added calls that toggle the reason/message on `Succeeded: Unknown` for this case.

> I am fine starting with 5s and see if we need to refine this later on

Changed to 5s, hopefully this covers the majority of cases 🤞 


I rebased, but put all the meaningful changes in a separate commit.

bobcatfish:
Welcome back!

> Sure, I added calls that toggle the reason/message on Succeeded: Unknown for this case.

Great, I think that helps!!

> Changed to 5s, hopefully this covers the majority of cases 🤞

Sounds reasonable to me! There's the last question of making it configurable as @vdemeester suggested - I am slightly inclined to go-ahead as is since this _only_ applies to PipelineResources and I think it's likely we'd be removing any config option we added later on (either b/c we don't have PipelineResources or b/c we'd be embedding the spec like we do with the other types). But I wouldn't block adding it if you feel strongly @vdemeester 

/hold cancel

mattmoor:
Yeah, +1 to delaying the addition of a knob until the future is clearer.

mattmoor:
This is the flake @afrittoli has been chasing:
```
Expected 2 number of successful events from pipelinerun and taskrun but got 3; list of received events
```

/retest

Issue #3382, opened 2020-10-13
v0.17.0 Pipeline can't be Run; it has an invalid spec: missing field(s): tasks[1].when[0]


r0bj:
# Expected Behavior

Pipeline is executed.

# Actual Behavior

PipelineValidationFailed:
```
  Warning  Failed         6m50s (x2 over 6m50s)  PipelineRun  Pipeline tekton-execution/build-images can't be Run; it has an invalid spec: missing field(s): tasks[1].when[0], tasks[2].when[0], tasks[4].when[0]
  Warning  InternalError  6m50s (x2 over 6m50s)  PipelineRun  1 error occurred:
           * missing field(s): tasks[1].when[0], tasks[2].when[0], tasks[4].when[0]
```

```
Status:
  Completion Time:  2020-10-13T23:30:43Z
  Conditions:
    Last Transition Time:  2020-10-13T23:30:43Z
    Message:               Pipeline tekton-execution/build-images can't be Run; it has an invalid spec: missing field(s): tasks[1].when[0], tasks[2].when[0], tasks[4].when[0]
    Reason:                PipelineValidationFailed
    Status:                False
    Type:                  Succeeded
  Pipeline Spec:
    Params:
      Name:  git-repo-url
      Type:  string
      Name:  git-revision
      Type:  string
      Name:  git-ref
      Type:  string
    Tasks:
      Name:  validate-git-ref
      Params:
        Name:   git-ref
        Value:  $(params.git-ref)
      Task Ref:
        Kind:  Task
        Name:  validate-git-ref
      Name:    fetch-repository
      Params:
        Name:   url
        Value:  $(params.git-repo-url)
        Name:   revision
        Value:  $(params.git-revision)
        Name:   deleteExisting
        Value:  true
      Task Ref:
        Kind:  Task
        Name:  git-clone
      When:
        Input:
        Operator:
        Values:    <nil>
      Workspaces:
        Name:       output
        Workspace:  build-images-workspace
```

It seems that "When" fields are empty.

# Steps to Reproduce the Problem

PipelineRun definition:

```
---
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: build-images
  namespace: tekton-execution
spec:
  params:
  - name: git-repo-url
  - name: git-revision
  - name: git-ref
  workspaces:
  - name: build-images-workspace
  tasks:

  - name: validate-git-ref
    taskRef:
      name: validate-git-ref
    params:
    - name: git-ref
[...]

PR #3389, opened 2020-10-14
Fix json annotations in WhenExpressions

jerop:
# Changes

When a pipeline with when expressions is created in v0.16.* then it is run in a pipelinerun in v0.17.0 or v0.17.1, a pipeline validation error about missing fields would be thrown -- as reported in https://github.com/tektoncd/pipeline/issues/3382

That happened because json annotations were added to when expressions type in v0.17.0 so that the fields would have lowercase, such as in the code completion in tekton intellij plugin as described in redhat-developer/intellij-tekton#223

Without the json annotations, the fields were stored with first letters capitalized, that is `Input`, `Operator` and `Values`. With the json annotations, the fields were expected to be lowercase, that is `input`, `operator` and `values`,  causing the missing fields error in pipeline validation

As such, users would have to reapply pipelines definitions created in previous versions to make them work in v0.17.0 or v0.17.1  -- to remove this requirement, we need to support both the uppercase and lowercase first letters for the annotations

Split up [the when expressions example](https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml) to separate pipelinerun and pipeline to make it easier to test a pipeline created in v0.16.3 and run with these changes:
- pipeline: https://gist.github.com/jerop/dd2d64fbc6de5b353ea449c02a81b59e
- pipelinerun: https://gist.github.com/jerop/13a58279cb5553c9b5d27f6735c23bbb

Fixes https://github.com/tektoncd/pipeline/issues/3382

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```release-note
pipelines with when expressions created in v0.16.3 can be run
```

/kind cleanup
/cc @pritidesai @bobcatfish

### DIFF
--- a/pkg/apis/pipeline/v1beta1/when_types.go
+++ b/pkg/apis/pipeline/v1beta1/when_types.go
@@ -25,16 +25,61 @@ import (
 type WhenExpression struct {
 	// Input is the string for guard checking which can be a static input or an output from a parent Task
 	Input string `json:"input"`
+
+	// DeprecatedInput for backwards compatibility with <v0.17
+	// it is the string for guard checking which can be a static input or an output from a parent Task
+	// +optional
+	DeprecatedInput string `json:"Input,omitempty"`
+
 	// Operator that represents an Input's relationship to the values
 	Operator selection.Operator `json:"operator"`
+
+	// DeprecatedOperator for backwards compatibility with <v0.17
+	// it represents a DeprecatedInput's relationship to the DeprecatedValues
+	// +optional
+	DeprecatedOperator selection.Operator `json:"Operator,omitempty"`
+
 	// Values is an array of strings, which is compared against the input, for guard checking
 	// It must be non-empty
 	Values []string `json:"values"`
+
+	// DeprecatedValues for backwards compatibility with <v0.17
+	// it represents a DeprecatedInput's relationship to the DeprecatedValues
+	// +optional
+	DeprecatedValues []string `json:"Values,omitempty"`
+}
+
+// GetInput returns the input string for guard checking
+// based on DeprecatedInput (<v0.17) and Input
+func (we *WhenExpression) GetInput() string {
+	if we.Input == "" {
+		return we.DeprecatedInput
+	}
+	return we.Input
+}
+
+// GetOperator returns the relationship between input and values
+// based on DeprecatedOperator (<v0.17) and Operator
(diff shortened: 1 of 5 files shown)


pritidesai (approved):
thanks @jerop

vdemeester (commented):
So I wonder if we should "just" revert (and cherry-pick real quick to do a 0.17.2) or if we should go for a cleaner fix.

The cleaner fix would be to have `Input` (`DeprecatedInput` field in go) and `input` (`Input` field in go) and have some code to convert the `Input` into `input`. 

I am ok with both, either we get that fix for "conversion" fix in 0.17.2, or in 0.18.0 (with a revert)

bobcatfish:
@vdemeester I'm torn too and am actually really interested if you have a preference.

I like that the revert will "correct" the situation to some extent without the risk of us rushing in more extensive changes (also side note i suggested this to @jerop ) - but if anyone has used when expressions with 0.17.0 or 0.17.1, i think this will be a backwards incompatible change for them. i'm not certain, but i think going from storing the fields with lowercase (0.17.0 or 0.17.1) back to uppercase (this change) would be backwards incompatible? since probably less ppl are in this situation than the other, i _think_ it's worth it?

I agree with you re. the cleaner fix, i think that's definitely the way to go, I guess it's a question of whether it's worth an interim release that "corrects" the backwards incompatible change in the meantime 🤔 🤔 🤔

jerop:
split up [the when expressions example](https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml) to separate pipelinerun and pipeline to make it easier to test pipeline created in v0.16.3 and running with these changes:
- pipeline: https://gist.github.com/jerop/dd2d64fbc6de5b353ea449c02a81b59e
- pipelinerun: https://gist.github.com/jerop/13a58279cb5553c9b5d27f6735c23bbb

sbwsg (approved):

pritidesai:
@bobcatfish @vdemeester folks have started migrating their pipelines to use `when` expressions with 0.16, migrating to 0.17 without this fix would break them. Thanks @jerop for implementing the cleaner fix.

vdemeester (commented):
/lgtm

Issue #28, opened 2018-09-10
Create comparison of existing CI systems


bobcatfish:
As requested in [the build CRD working group](https://github.com/knative/docs/blob/master/community/WORKING-GROUPS.md#build), we should compare the features we are looking for in the Pipeline CRD to features provided by existing solutions to explain why we feel there is a need for this project.

bobcatfish:
[Pipeline CRD CI/CD Comparison](https://docs.google.com/spreadsheets/d/1ejHvL-GEC0CTXi4WpQ6mA2D4fAChGrlQkLOfqS9cKBk/), shared with members of [knative-dev](https://github.com/knative/docs/blob/master/community/WORKING-GROUPS.md#build).

The document compares the features we'd like the Pipeline CRD to Concourse, Jenkins, Argo, Prow, Bamboo and TeamCity. (Note however that the goal is that any of these solutions could add the Pipeline CRD to their tools, in which case they would also be supporting the features the Pipeline CRD will support!)

PR #3390, opened 2020-10-15
access execution status of any task in finally

### LINKED ISSUE #1020: Expose pipeline run metadata to Tasks and Conditions
Condition containers need access to metadata about the pipeline run. The metadata will be in the form of  a `pr-metadata.json` file that is present in the `workspace` directory.

Part of implementation of #27

ref: https://github.com/tektoncd/pipeline/issues/1137

pritidesai:
# Changes

Introducing a variable which can be used to access the execution status of any `pipelineTask` in a `pipeline`. This can be used in any `finally` task (and limited to `finally` tasks for the first iteration).

Use `$(tasks.<pipelineTask>.status)` as param value which resolves to the status, one of, `Succeeded`, `Failed`, and `None`.

E.g.:

```yaml
    finally:
    - name: finaltask
      params:
        - name: task1Status
          value: "$(tasks.task1.status)"
      taskSpec:
        params:
          - name: task1Status
        steps:
          - image: ubuntu
            name: print-task-status
            script: |
              if [ $(params.task1Status) == "Failed" ]
              then
                echo "Task1 has failed, now process this failure"
              fi
```

Partially Closes #1020 
Implements TEP [#0028](https://github.com/tektoncd/community/blob/master/teps/0028-task-execution-status-at-runtime.md)

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Introducing a variable $(tasks.<taskName>.status) to access execution status of any non finally pipelineTask in finally.
```

/kind feature

### DIFF
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -820,6 +820,44 @@ Overall, `PipelineRun` state transitioning is explained below for respective sce
 Please refer to the [table](pipelineruns.md#monitoring-execution-status) under Monitoring Execution Status to learn about
 what kind of events are triggered based on the `Pipelinerun` status.
 
+
+### Using Execution `Status` of `pipelineTask`
+
+Finally Task can utilize execution status of any of the `pipelineTasks` under `tasks` section using param:
+
+```yaml
+    finally:
+    - name: finaltask
+      params:
+        - name: task1Status
+          value: "$(tasks.task1.status)"
+      taskSpec:
+        params:
+          - name: task1Status
+        steps:
+          - image: ubuntu
+            name: print-task-status
+            script: |
+              if [ $(params.task1Status) == "Failed" ]
+              then
+                echo "Task1 has failed, continue processing the failure"
+              fi
+```
+
+This kind of variable can have any one of the values from the following table:
+
+| Status | Description |
+| ------- | -----------|
+| Succeeded | `taskRun` for the `pipelineTask` completed successfully |
+| Failed | `taskRun` for the `pipelineTask` completed with a failure or cancelled by the user |
+| None | the `pipelineTask` has been skipped or no execution information available for the `pipelineTask` |
+
+For an end-to-end example, see [`status` in a `PipelineRun`](../examples/v1beta1/pipelineruns/pipelinerun-task-execution-status.yaml).
+
+**Note:** `$(tasks.<pipelineTask>.status)` is instantiated and available at runtime and must be used as a param value
+as is without concatenating it with any other param or string, for example, this kind of usage is not validated/supported
+`task status is $(tasks.<pipelineTask>.status)`.
+
 ### Known Limitations
 
 ### Specifying `Resources` in Final Tasks
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -23,6 +23,7 @@ For instructions on using variable substitutions see the relevant section of [th
 | `context.pipelineRun.namespace` | The namespace of the `PipelineRun` that this `Pipeline` is running in. |
 | `context.pipelineRun.uid` | The uid of the `PipelineRun` that this `Pipeline` is running in. |
 | `context.pipeline.name` | The name of this `Pipeline` . |
+| `tasks.<pipelineTaskName>.status` | The execution status of the specified `pipelineTask`, only available in `finally` tasks. |
 
 
 ## Variables available in a `Task`
--- a/examples/v1beta1/pipelineruns/pipelinerun-task-execution-status.yaml
+++ b/examples/v1beta1/pipelineruns/pipelinerun-task-execution-status.yaml
@@ -1,0 +1,44 @@
+kind: PipelineRun
+apiVersion: tekton.dev/v1beta1
+metadata:
+  generateName: pr-execution-status-
+spec:
+  serviceAccountName: 'default'
+  pipelineSpec:
+    tasks:
+    - name: task1 # successful task
+      taskSpec:
+        steps:
+        - image: ubuntu
+          name: hello
+          script: |
+            echo "Hello World!"
+    - name: task2 # skipped task
+      when:
+        - input: "true"
+          operator: "notin"
+          values: ["true"]
+      taskSpec:
+        steps:
+          - image: ubuntu
+            name: success
+            script: |
+              exit 0
+    finally:
+    - name: task3 # this task verifies the status of dag tasks, it fails if verification fails
+      params:
+        - name: task1Status
+          value: "$(tasks.task1.status)"
+        - name: task2Status
+          value: "$(tasks.task2.status)"
+      taskSpec:
+        params:
+          - name: task1Status
+          - name: task2Status
+        steps:
+          - image: alpine
+            name: verify-dag-task-status
+            script: |
+              if [[ $(params.task1Status) != "Succeeded" ||  $(params.task2Status) != "None" ]]; then
+                exit 1;
+              fi
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation.go
@@ -62,6 +62,7 @@ func (ps *PipelineSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	errs = errs.Also(validatePipelineParameterVariables(ps.Tasks, ps.Params).ViaField("tasks"))
 	errs = errs.Also(validatePipelineParameterVariables(ps.Finally, ps.Params).ViaField("finally"))
 	errs = errs.Also(validatePipelineContextVariables(ps.Tasks))
+	errs = errs.Also(validateExecutionStatusVariables(ps.Tasks, ps.Finally))
 	// Validate the pipeline's workspaces.
 	errs = errs.Also(validatePipelineWorkspaces(ps.Workspaces, ps.Tasks, ps.Finally))
 	// Validate the pipeline's results
@@ -288,6 +289,40 @@ func validatePipelineContextVariables(tasks []PipelineTask) *apis.FieldError {
 	}
 	errs := validatePipelineContextVariablesInParamValues(paramValues, "context\\.pipelineRun", pipelineRunContextNames)
 	return errs.Also(validatePipelineContextVariablesInParamValues(paramValues, "context\\.pipeline", pipelineContextNames))
+}
+
+func validateExecutionStatusVariables(tasks []PipelineTask, finallyTasks []PipelineTask) (errs *apis.FieldError) {
+	// creating a list of pipelineTask names to validate tasks.<name>.status
+	pipelineRunTasksContextNames := sets.String{}
+	for idx, t := range tasks {
+		for _, param := range t.Params {
+			// validate dag pipeline tasks not accessing execution status of other pipeline task
+			if ps, ok := GetVarSubstitutionExpressionsForParam(param); ok {
+				for _, p := range ps {
+					if strings.HasPrefix(p, "tasks.") && strings.HasSuffix(p, ".status") {
+						errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("pipeline tasks can not refer to execution status of any other pipeline task"),
+							"value").ViaFieldKey("params", param.Name).ViaFieldIndex("tasks", idx))
+					}
+				}
+			}
+		}
+		pipelineRunTasksContextNames.Insert(t.Name)
+	}
+
+	// validate finally tasks accessing execution status of a dag task specified in the pipeline
+	var paramValues []string
+	for _, t := range finallyTasks {
+		for _, param := range t.Params {
+			paramValues = append(paramValues, param.Value.StringVal)
+			paramValues = append(paramValues, param.Value.ArrayVal...)
+		}
+	}
+	for _, paramValue := range paramValues {
+		if strings.HasPrefix(stripVarSubExpression(paramValue), "tasks.") && strings.HasSuffix(stripVarSubExpression(paramValue), ".status") {
+			errs = errs.Also(substitution.ValidateVariablePS(paramValue, "tasks", "status", pipelineRunTasksContextNames).ViaField("value"))
+		}
+	}
+	return errs
 }
 
 func validatePipelineContextVariablesInParamValues(paramValues []string, prefix string, contextNames sets.String) (errs *apis.FieldError) {
--- a/pkg/reconciler/pipelinerun/resources/apply.go
+++ b/pkg/reconciler/pipelinerun/resources/apply.go
@@ -87,6 +87,17 @@ func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResul
 	}
 }
 
+//ApplyPipelineTaskContext replaces context variables referring to execution status with the specified status
+func ApplyPipelineTaskContext(state PipelineRunState, replacements map[string]string) {
+	for _, resolvedPipelineRunTask := range state {
+		if resolvedPipelineRunTask.PipelineTask != nil {
+			pipelineTask := resolvedPipelineRunTask.PipelineTask.DeepCopy()
+			pipelineTask.Params = replaceParamValues(pipelineTask.Params, replacements, nil)
+			resolvedPipelineRunTask.PipelineTask = pipelineTask
+		}
+	}
+}
+
 // ApplyWorkspaces replaces workspace variables in the given pipeline spec with their
 // concrete values.
 func ApplyWorkspaces(p *v1beta1.PipelineSpec, pr *v1beta1.PipelineRun) *v1beta1.PipelineSpec {
--- a/pkg/reconciler/pipelinerun/resources/apply_test.go
+++ b/pkg/reconciler/pipelinerun/resources/apply_test.go
@@ -882,4 +882,42 @@ func TestApplyTaskResultsToPipelineResults(t *testing.T) {
 			}
 		})
 	}
+
+}
+
+func TestApplyTaskRunContext(t *testing.T) {
+	r := map[string]string{
+		"tasks.task1.status": "succeeded",
+		"tasks.task3.status": "none",
+	}
+	state := PipelineRunState{{
+		PipelineTask: &v1beta1.PipelineTask{
+			Name:    "task4",
+			TaskRef: &v1beta1.TaskRef{Name: "task"},
+			Params: []v1beta1.Param{{
+				Name:  "task1",
+				Value: *v1beta1.NewArrayOrString("$(tasks.task1.status)"),
+			}, {
+				Name:  "task3",
+				Value: *v1beta1.NewArrayOrString("$(tasks.task3.status)"),
+			}},
+		},
+	}}
+	expectedState := PipelineRunState{{
+		PipelineTask: &v1beta1.PipelineTask{
+			Name:    "task4",
+			TaskRef: &v1beta1.TaskRef{Name: "task"},
+			Params: []v1beta1.Param{{
+				Name:  "task1",
+				Value: *v1beta1.NewArrayOrString("succeeded"),
+			}, {
+				Name:  "task3",
+				Value: *v1beta1.NewArrayOrString("none"),
+			}},
+		},
+	}}
+	ApplyPipelineTaskContext(state, r)
+	if d := cmp.Diff(expectedState, state); d != "" {
+		t.Fatalf("ApplyTaskRunContext() %s", diff.PrintWantGot(d))
+	}
 }
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -132,6 +132,18 @@ func (t ResolvedPipelineRunTask) IsStarted() bool {
 	return t.TaskRun != nil && t.TaskRun.Status.GetCondition(apis.ConditionSucceeded) != nil
 }
 
+// IsConditionStatusFalse returns true when a task has succeeded condition with status set to false
+// it includes task failed after retries are exhausted, cancelled tasks, and time outs
+func (t ResolvedPipelineRunTask) IsConditionStatusFalse() bool {
+	if t.IsStarted() {
+		if t.IsCustomTask() {
+			return t.Run.Status.GetCondition(apis.ConditionSucceeded).IsFalse()
+		}
+		return t.TaskRun.Status.GetCondition(apis.ConditionSucceeded).IsFalse()
+	}
+	return false
+}
+
 func (t *ResolvedPipelineRunTask) checkParentsDone(facts *PipelineRunFacts) bool {
 	stateMap := facts.State.ToMap()
 	node := facts.TasksGraph.Nodes[t.PipelineTask.Name]
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
@@ -17,6 +17,7 @@ limitations under the License.
 package resources
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
@@ -26,6 +27,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/apis"
+)
+
+const (
+	// ReasonUnknown indicates that the execution status of a pipelineTask is unknown
+	PipelineTaskStateNone = "None"
+	// PipelineTaskStatusPrefix is a prefix of the param representing execution state of pipelineTask
+	PipelineTaskStatusPrefix = "tasks."
+	// PipelineTaskStatusSuffix is a suffix of the param representing execution state of pipelineTask
+	PipelineTaskStatusSuffix = ".status"
 )
 
 // PipelineRunState is a slice of ResolvedPipelineRunTasks the represents the current execution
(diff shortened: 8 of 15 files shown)


examples/v1beta1/pipelineruns/pipelinerun-task-execution-status.yaml
```diff
@@ -0,0 +1,56 @@
...
+      taskSpec:
+        params:
+          - name: pipelineRun-pipelineTask-task1
+          - name: pipelineRun-pipelineTask-task2
+          - name: pipelineRun-pipelineTask-task3
+        steps:
+          - image: ubuntu
+            name: print-failed-task-status
+            script: |
+              echo "Pipeline Task Status: Task1:  $(params.pipelineRun-pipelineTask-task1)"
```
ImJasonH, line 53: Can we make this example fail if the reported status isn't what we expect? That might catch a future bug that would otherwise go undetected.

pritidesai: yup definitely, any pointer on how to add failed run under `/examples`? I think we added support for it at some point 🤔 will it have to live under `/test`?

pritidesai: I couldn't find anywhere 😭 I will move this test under `/test` but would like to keep this example file as well. Do we run examples from [no-ci](https://github.com/tektoncd/pipeline/tree/master/examples/v1beta1/pipelineruns/no-ci), may be move there?

bobcatfish:
hey @pritidesai !! I'm very excited to see this moving forward :D

Do you think this might be worth a TEP to explore the problem and the alternatives a bit?

For example the mapping between the "execution states" and "status" you're proposing (https://github.com/tektoncd/pipeline/issues/1020#issuecomment-706396281) is interesting - it's also different from what you currently see in the status field of an executing TaskRun/PipelineRun so I think it might be worth contrasting this with some different approaches, e.g.:

1. Exposing the entire status of a Run as metadata, using something like json path to extract values (i think we've looked at this before and it's pretty verbose but i think it's worth considering)
2. Exposing just the execution status information, but mirroring the structure in the actual status of the run (e.g. instead of `context.pipelineRun.Tasks.Foo.Succeeded`, `context.pipelineRun.Tasks.status.conditions.Succeeded` <-- we could return either `Unknown`, `True`, `False`, `` or make them slightly more readable `Running`, `Succeeded`, `Failed`, `None`

Looking at [the docs on the execution status](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md#monitoring-execution-status) it feels like this proposal is making the trying to make the "reason" field available to users, and treating that as the overall status (and mixing 'skipped' in as well)

> Unknown: no information available on that pipelineTask execution

This one in particular is confusing: "unknown" as a condition status currently means "running" but we are using "unknown" to mean "not set" (i think?)

bobcatfish:
One more thing that might be worth exploring: some of these statuses won't have meaning or be available. For example:

- Failed - only available in Finally Tasks (all other Tasks will stop executing)
- TaskRunCancelled - i _think_ that even finally tasks are not executed if a task is cancelled, so i dont think realistically that this value will ever be present?

For all of these, the values will only be available to tasks that execute during or after the task that is trying to use the values

pritidesai:
> hey @pritidesai !! I'm very excited to see this moving forward :D

🎉 🎉 🎉 

> Do you think this might be worth a TEP to explore the problem and the alternatives a bit?

Yup, definitely 👍 

> For example the mapping between the "execution states" and "status" you're proposing ([#1020 (comment)](https://github.com/tektoncd/pipeline/issues/1020#issuecomment-706396281)) is interesting - it's also different from what you currently see in the status field of an executing TaskRun/PipelineRun so I think it might be worth contrasting this with some different approaches, e.g.:

The use cases we have here are:

* UC1: `finally` task must be able to check the execution state of any `DAG` task.
* UC2: A `DAG` task must be able to check if other `DAG` task was skipped. The reason to include only `skipped` here, like you mentioned in the comment below ⬇️ ,  if any of the `DAG` tasks is failed or cancelled, `pipeline` starts executing `finally` tasks and does not schedule rest of the `DAG` tasks.

The proposed list of states here are derived based on `ResolvedPipelineTask.TaskRun.Status.GetCondition(apis.ConditionSucceeded)` i.e. `Succeeded` `ConditionType` in addition to `Status` and `Reason`. The same methods that we use to get the execution queue in `pipelinerun.go`.  The state is also derived based on `pipelineRun.Status.SkippedTasks` for tasks which does not have any `taskRun` associated with them and to distinguish them from not yet scheduled tasks.

* Succeeded: a pipelineTask was successful, a respective pod was created and completed successfully. Condition `status` is set to `true`.

* Failed: a pipelineTask failed, a respective pod was created but exited with error. Condition `status` is set to `false` and have exhausted all the retries.

* TaskRunCancelled: taskRun for that pipelineTask is cancelled by the user. Condition `status` is set to `false` and the reason is set to `TaskRunCancelled`.

* Skipped: a pipelineTask is skipped because of the when expression or its parent was skipped.

* Unknown: this is very crucial and ambiguous, this is more like none of the above. pipelineTask was not skipped but did not have any taskRun associated with it. No information on whether this task is going to be scheduled next or will be left unscheduled due to a `DAG` failure.

> 1. Exposing the entire status of a Run as metadata, using something like json path to extract values (i think we've looked at this before and it's pretty verbose but i think it's worth considering)

👍  This can be next in line and can enhance the existing proposal.

The existing proposal brings quick access to the state without investing into json path.

> 2. Exposing just the execution status information, but mirroring the structure in the actual status of the run (e.g. instead of `context.pipelineRun.Tasks.Foo.Succeeded`, `context.pipelineRun.Tasks.status.conditions.Succeeded` <-- we could return either `Unknown`, `True`, `False`, `` 

`context.pipelineRun.Tasks.status.conditions.Succeeded` with `Unknown`, `True`, `False` would not bring much value since `False` can mean failed or cancelled which is still fine but will be difficult to capture `skipped`.

> or make them slightly more readable `Running`, `Succeeded`, `Failed`, `None`

Definitely we can rename the states to `Running`, `Succeeded`, `Failed`, `None` making them more readable. 

> Looking at [the docs on the execution status](https://github.com/tektoncd/pipeline/blob/master/docs/pipelineruns.md#monitoring-execution-status) it feels like this proposal is making the trying to make the "reason" field available to users, and treating that as the overall status (and mixing 'skipped' in as well)

Yup, so that its not hidden in `None` or `Unknown`.

> > Unknown: no information available on that pipelineTask execution
> 
> This one in particular is confusing: "unknown" as a condition status currently means "running" but we are using "unknown" to mean "not set" (i think?)

It means there is absolutely no way to determine what is the status at that point in time. We could introduce `Running` state which is available but unknown could mean none of the proposed state. This state should not be returned unless a state of a child is requested even before a parent is executed or finished executing. When a parent is executing, we have no idea whats going to happen to that parent and if that child be ever scheduled/executed.

So worth a TEP 😉

pritidesai:
>  I need to confirm the cancelled scenario.

When any `DAG` task is cancelled, pipeline triggers finally tasks.

pritidesai:
> Do you think this might be worth a TEP to explore the problem and the alternatives a bit?

https://github.com/tektoncd/community/pull/234

bobcatfish:
> The proposed list of states here are derived based on `ResolvedPipelineTask.TaskRun.Status.GetCondition(apis.ConditionSucceeded)` i.e. `Succeeded` `ConditionType` in addition to `Status` and `Reason`. The same methods that we use to get the execution queue in `pipelinerun.go`.  The state is also derived based on `pipelineRun.Status.SkippedTasks` for tasks which does not have any `taskRun` associated with them and to distinguish them from not yet scheduled tasks.

Thanks for the detailed explanation! I'm going to take my comments mostly into the TEP; in general I'm wondering if there is a simpler subset of these values we could start with initially, and maybe to keep things simpler, start with just the expected values of the Succeeded ConditionType, then we can be deliberate about adding more if needed.

> context.pipelineRun.Tasks.status.conditions.Succeeded with Unknown, True, False would not bring much value since False can mean failed or cancelled which is still fine but will be difficult to capture skipped.

At the moment I don't think there's a scenario where a task is cancelled and other tasks will start to execute after that. If there is later on, we could add additional information to the variable replacement to cover it.

> A pipeline does not have skipped state.

Oh yeah great point! imo this is even more reason not to include this as one of the possible values of the new field (more in my TEP comments!)

> When any DAG task is cancelled, pipeline triggers finally tasks.

Oh, do you mean when a taskrun that is part of a pipelinerun is cancelled? I was thinking of when the entire PipelineRun is cancelled.

I guess I could see someone doing that but it seems like a strange scenario to support - but in the TEP it looks like "cancelled" isnt being proposed initially so maybe we can cross that bridge when we come to it :D

pritidesai:
hey @afrittoli I noticed you are releasing 0.20 (from owners doc 😄 ). I have added this as part of milestone for now based on the [ask](https://github.com/tektoncd/pipeline/issues/1020#issuecomment-747790476).

docs/variables.md
```diff
@@ -39,6 +39,7 @@ This page documents the variable substitutions supported by `Tasks` and `Pipelin
 | `context.taskRun.namespace` | The namespace of the `TaskRun` that this `Task` is running in. |
 | `context.taskRun.uid` | The uid of the `TaskRun` that this `Task` is running in. |
 | `context.task.name` | The name of this `Task`. |
+| `tasks.<pipelineTaskName>.status` | The execution status of the specified `pipelineTask`, only available in `finally` tasks. |
```
GregDritschler, line 42: Shouldn't this be under "Variables available in a `Pipeline`" ?

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -253,6 +254,23 @@ func validatePipelineContextVariables(tasks []PipelineTask) *apis.FieldError {
 	return errs.Also(validatePipelineContextVariablesInParamValues(paramValues, "context\\.pipeline", pipelineContextNames))
 }
 
+func validateExecutionStatusContextVariables(tasks []PipelineTask) (errs *apis.FieldError) {
+	// creating a list of pipelineTask names to validate context.pipelineRun.Tasks.<name>
+	pipelineRunTasksContextNames := sets.String{}
+	var paramValues []string
+	for _, t := range tasks {
+		pipelineRunTasksContextNames.Insert(t.Name)
+		for _, param := range t.Params {
```
GregDritschler, line 263: I'm a little confused; maybe I'm misreading the code.  I think we want to validate the following:

1. Normal (non-final) tasks cannot use the $(tasks.<pipelinetask>.status) syntax.
2. Final tasks can use this syntax but must refer to a defined pipeline task.

Reading the code, it seems to be validating the parameter references in normal tasks refer to other existing pipeline tasks?

pkg/substitution/substitution.go
```diff
@@ -60,6 +60,22 @@ func ValidateVariableP(value, prefix string, vars sets.String) *apis.FieldError
 	return nil
 }
 
+func ValidateVariablePS(value, prefix string, suffix string, vars sets.String) *apis.FieldError {
+	if vs, present := extractVariablesFromString(value, prefix); present {
```
GregDritschler, line 64: I think there may be an exposure here.  The function is validating references starting with the given prefix, so in your case `$(tasks.`.   It is not considering the suffix (results) when gathering these.  There are other supported references that start with `$(tasks.` (e.g. result references) so I'm concerned it could extract those too and try to validate them which is not your intention, right?

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -324,6 +334,32 @@ func (facts *PipelineRunFacts) GetSkippedTasks() []v1beta1.SkippedTask {
 	return skipped
 }
 
+// GetPipelineTaskStatus returns the status of a PipelineTask depending on its taskRun
+func (facts *PipelineRunFacts) GetPipelineTaskStatus(ctx context.Context) map[string]string {
+	// construct a map of context.pipelineRun.tasks.<pipelineTask>.status and its state
```
GregDritschler, line 339: please update comment to use current syntax

pkg/reconciler/pipelinerun/resources/pipelinerunstate_test.go
```diff
@@ -1196,3 +1197,117 @@ func TestAdjustStartTime(t *testing.T) {
...
+			ResolvedTaskResources: &resources.ResolvedTaskResources{
+				TaskSpec: &task.Spec,
+			},
+		}},
+		dagTasks: []v1beta1.PipelineTask{pts[9]},
+		expectedStatus: map[string]string{
+			PipelineTaskStatusPrefix + pts[9].Name + PipelineTaskStatusSuffix: PipelineTaskStateNone,
+		},
+	}, {
+		name: "tasks-when-expression-failed-and-task-succeeded",
```
GregDritschler, line 1262: I'm a bit puzzled by this one.  If I'm looking at the right one, pts[10] does have a when condition that would fail.  So how do this pipeline task have a TaskRun?  I don't know what your intentions are with this test.

test/pipelinefinally_test.go
```diff
@@ -115,6 +142,27 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_Failure(t *testing.T) {
 				t.Errorf("Error waiting for TaskRun to succeed: %v", err)
 			}
 			finalTaskStartTime = taskrunItem.Status.StartTime
+		case n == "finaltask2":
+			if err := WaitForTaskRunState(ctx, c, taskrunItem.Name, TaskRunSucceed(taskrunItem.Name), "TaskRunSuccess"); err != nil {
```
GregDritschler, line 146: I'm curious about the WaitForTaskRunState call here and in some of the other case blocks.  The test waits for the PipelineRun to fail (WaitForPipelineRunState) prior to this.  Can any TaskRuns still be running?  I know there have been problems in the past putting the PipelineRun into a done state before the TaskRuns were done but I thought that was resolved...

docs/variables.md
pritidesai: I don't know, I was thinking a task here means a pipelineTask and that's why added it here.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
pritidesai: Great catch @GregDritschler your are absolutely correct.

This validation checks a pipelineTask specified in `$(tasks.pipelinetask.status)` exists in the pipeline. This function is old and needs to be updated. Will fix it.

docs/variables.md
afrittoli: The two sections describe variables available when authoring pipelines and tasks respectively. 
All the examples (including in the TEP) show that the variable is used within a Pipeline, which is correct because a task cannot know about `pipelineTask` names, unless perhaps it's an embedded one. 
So this variable should be moved to the pipeline section.

docs/pipelines.md
```diff
@@ -791,6 +791,40 @@ Overall, `PipelineRun` state transitioning is explained below for respective sce
 Please refer to the [table](pipelineruns.md#monitoring-execution-status) under Monitoring Execution Status to learn about
 what kind of events are triggered based on the `Pipelinerun` status.
 
+
+### Using Execution `Status` of `pipelineTask`
+
+Finally Task can utilize execution status of any of the `pipelineTasks` under `tasks` section using param:
+
+```yaml
```
afrittoli, line 799: I forgot if we support `when` expressions on finally pipeline tasks yet? 
That would be a nice example to use here, but probably it's not supported yet?

pkg/reconciler/pipelinerun/resources/pipelinerunstate_test.go
pritidesai: This test is meant to validate a task with when expression evaluating to failure is skipped. I have fixed this test case. Thanks!

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
pritidesai: Fixed 🙏

docs/variables.md
pritidesai: thanks @afrittoli I have moved this to pipeline section 🙏

docs/pipelines.md
pritidesai: yup that would be great, we dont have when expression support in finally yet. We have a feature request open here https://github.com/tektoncd/pipeline/issues/3438 😄

pkg/apis/pipeline/v1beta1/pipeline_validation.go
pritidesai: Fixed 🙏

pkg/substitution/substitution.go
pritidesai: right, `status` is specified while calling this function in `suffix`.  After this extraction, suffix is trimmed before comparing against `vars`.

`ValidateVariablePS` is extracting prefix `tasks` and trimming suffix `status` and validates the left over value is one of the pipelineTasks (specified in `vars`).

you are right, this function would extract other variations (result reference) as well like you mentioned but result references are validated with `validateParamResults`. Any suggestion on how to better handle this?

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -290,6 +291,38 @@ func validatePipelineContextVariables(tasks []PipelineTask) *apis.FieldError {
...
 
+func validateExecutionStatusVariables(tasks []PipelineTask, finallyTasks []PipelineTask) (errs *apis.FieldError) {
+	// creating a list of pipelineTask names to validate tasks.<name>.status
+	pipelineRunTasksContextNames := sets.String{}
+	for idx, t := range tasks {
+		for _, param := range t.Params {
+			// validate dag pipeline tasks not accessing execution status of other pipeline task
+			if ps, ok := GetVarSubstitutionExpressionsForParam(param); ok {
+				for _, p := range ps {
+					if strings.HasPrefix(p, "tasks") && strings.HasSuffix(p, "status") {
```
GregDritschler, line 302: I suggest tightening this check up a bit by including the periods.

`if strings.HasPrefix(p, "tasks.") && strings.HasSuffix(p, ".status") {`

pkg/substitution/substitution.go
GregDritschler: > Any suggestion on how to better handle this? If not, I would leave it as is since result references does go through validation separately.

I'm concerned that when finally tasks are allowed to use result references in the future, this code will break because it will reject those references.

The "quick and dirty" way to fix is would be to check if the reference at least has the suffix before trimming it.  If it doesn't have that suffix then it isn't the type of reference your are trying to check.  A more involved fix would be to have code specific to this type of reference that uses a regex for exact matching.  For example, resultref.go has code specific to parsing result references.  I understand though that you are trying to leverage what's already there.

pritidesai: Thanks @GregDritschler, I have added an explicit check on the prefix and suffix before trimming it. 

I was actually looking at how task results are validated and to my surprise, the pipeline task names are not validated while validating the task results syntax i.e. `validateParamResults` validates the syntax $(tasks.mytask.results.total) but does not validate if `mytask` is actually defined as a pipeline task. Its not worrisome as such since Its captured in building `dag` function which fails with: 

```
validation failed: invalid value: couldn't add link between your-task and my-task: task your-task depends on my-task but my-task wasn't present in Pipeline: spec.tasks
```

pkg/apis/pipeline/v1beta1/pipeline_validation.go
pritidesai: Done 👍

test/pipelinefinally_test.go
pritidesai: Right, its waiting for the pipelineRun to fail using this function `WaitForPipelineRunState` right after creating a run. It's more of a polling for that state i.e. wait until pipeline is done. The state right away might not be set as expected. How can we make sure that the run finishes without waiting for it?

GregDritschler: I was saying that it's odd that after waiting for the pipelinerun to finish, the test also then waits for taskruns to finish.  The pipelinerun shouldn't be finished if there are taskruns still running.  I know there are some corner cases where this isn't true -- cancel comes to mind -- but I'm uncertain whether there is something going on here that makes it happen.

pkg/reconciler/pipelinerun/resources/apply_test.go
```diff
@@ -610,3 +610,41 @@ func TestApplyWorkspaces(t *testing.T) {
...
+				Value: *v1beta1.NewArrayOrString("succeeded"),
+			}, {
+				Name:  "task3",
+				Value: *v1beta1.NewArrayOrString("none"),
+			}},
+		},
+	}}
+	ApplyPipelineTaskContext(state, r)
+	if d := cmp.Diff(expectedState, state); d != "" {
+		t.Fatalf("AppluTaskRunContext() %s", diff.PrintWantGot(d))
```
sbwsg, line 647: nit: `Applu` -> `Apply`

sbwsg (approved):

examples/v1beta1/pipelineruns/pipelinerun-task-execution-status.yaml
```diff
@@ -0,0 +1,44 @@
...
+          value: "$(tasks.task1.status)"
+        - name: task2Status
+          value: "$(tasks.task2.status)"
+      taskSpec:
+        params:
+          - name: task1Status
+          - name: task2Status
+        steps:
+          - image: alpine
+            name: print-failed-task-status
```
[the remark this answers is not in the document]
pritidesai: thanks @sbwsg, fixed the step name and added comment

pkg/reconciler/pipelinerun/resources/apply_test.go
pritidesai: Fixed 🤦‍♀️

test/pipelinefinally_test.go
pritidesai: Aah, I see what you mean. I am not sure either, pipelinerun completion mean all taskruns completed within the test `context`, ideally it should.

But, I see it as more than just waiting for the taskrun to finish, `WaitForTaskRunState` also verifies the state and returns error if state does not match. Do you suggest any alternative? I can follow up in a separate PR, may be write a separate routine to just validate the state?

pkg/substitution/substitution.go
afrittoli: Thanks @pritidesai - I do not see the check in this function - perhaps you refer to https://github.com/tektoncd/pipeline/pull/3390/files#diff-78ff878240189b20e0ef82f2912c93c0b3c7b445865e04fa9b98aab851be6130R321 (L321)? However that check will only work for parameters in the "$(var)" format, but not if extra text or more variables are included.

afrittoli (commented):
Thank you @pritidesai !

This looks good, the only problem left that I see is the parameter variable validation, which makes the assumption that the parameter is in the format "$(var)" only.

I would prefer not checking explicitly for cancel / timeout, since that code will never be reached anyways and we have no tests for it. It should be a quick fix if you want to update the PR it can still be in v0.20.0

pritidesai:
thanks @afrittoli, I have created issue on validation as per our slack conversation and updated the checks to check for successful condition type with status `true` and `false` instead of explicitly checking for cancelled or time out. Are there any changes needed for this to make it to 0.20?

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -87,6 +87,17 @@ func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResul
 	}
 }
 
+//ApplyPipelineTaskContext replaces context variables referring to execution status with the specified status
```
afrittoli, line 90: NIT:  missing space `// Apply...`

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -132,6 +132,18 @@ func (t ResolvedPipelineRunTask) IsStarted() bool {
 	return t.TaskRun != nil && t.TaskRun.Status.GetCondition(apis.ConditionSucceeded) != nil
 }
 
+// IsConditionStatusFalse returns true when a task has succeeded condition with status set to false
+// it includes task failed after retries are exhausted, cancelled tasks, and time outs
+func (t ResolvedPipelineRunTask) IsConditionStatusFalse() bool {
```
afrittoli, line 137: Nice, thank you!

pkg/substitution/substitution.go
afrittoli: I think I reasonable fix here could be to have a variation of `extractVariablesFromString` that returns the stripped `suffix` so that it could be verified.
We should follow up on this PR to had such fix to make sure the code does not break when we add support for results like @GregDritschler mentioned.

afrittoli (commented):
Thanks for all the updates and the patience!
It's great to have this new feature in Tekton :)
/lgtm

riceluxs1t:
@pritidesai 

$(tasks.<pipelineTask>.status) How would a PipelineRun that has its Pipeline/Task specs fully inlined access these parameters? I believe for a fully inlined PipelineRun, tasks get random suffixes attached to their names.

pritidesai:
@riceluxs1t please provide an example YAML if possible.

`$(tasks.<pipelineTask>.status)` represents execution status of a pipelineTask under `tasks` section and is only accessible in `finally` tasks. For pipelineRun with inlined pipeline/task spec, these parameters stays the same as the pipelineTask names are specified by the users and are not updated.

Here is an example of pipelineRun with inlined specification: https://github.com/tektoncd/pipeline/blob/master/examples/v1beta1/pipelineruns/pipelinerun-task-execution-status.yaml

> I believe for a fully inlined PipelineRun, tasks get random suffixes attached to their names.

Are you seeing suffixes getting attached to pipelineTask names?

PR #3395, opened 2020-10-15
Mark failure to copy cred to user home as warning

sbwsg:
Contributes to https://github.com/tektoncd/pipeline/issues/3399

# Changes

Prior to this commit a message could be seen in step logs
when tekton failed to copy a credential from /tekton/creds
to /tekton/home or $HOME (if disable-home-override feature
flag is "true"). That message can be confusing because it
appears to be an error when in actual fact it's a warning.

The message appears when Steps inside a Task run with
varying UIDs - one UID may copy the credentials first
and then subsequent steps will attempt to copy over them.
If a Step running as root, for examples, copies creds
into /tekton/home, other non-root Steps will not be able
to utilize those credentials. Unfortunately this can manifest
in our own PipelineResources where the Steps they inject
use a mix of root & non-root base images.

The message can also appear if a user has explicltly
mounted credentials (via workspace or volumemount) in
/tekton/home and has _also_ attached a service account
to the task with creds-init credentials on it.

The message serves as an indicator of a potential problem but
not 100% guarantee that any issues in a Step are related to it.

This commit updates the message to include a "warning:"
prefix to indicate that it might (though not always)
be a potential source of issues in your Step.

This commit also adds documentation to docs/auth.md to help
users diagnose the impact of this message (including when
they can safely ignore it).

# Submitter Checklist

review them:

# Release Notes

```release-note
A message related to copying credentials has been more clearly marked as a warning to reduce confusion when it appears in Step logs.
```

/kind cleanup

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -27,6 +27,12 @@ refers to `TaskRuns` and `PipelineRuns` as `Runs` for the sake of brevity.
   - [`basic-auth` for Git](#basic-auth-for-git)
   - [`ssh-auth` for Git](#ssh-auth-for-git)
   - [`basic-auth` for Docker](#basic-auth-for-docker)
+  - [Errors and their meaning](#errors-and-their-meaning)
+    - ["unsuccessful cred copy" Warning](#unsuccessful-cred-copy-warning)
+      - [Multiple Steps with varying UIDs](#multiple-steps-with-varying-uids)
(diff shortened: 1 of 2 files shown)


sbwsg:
/hold

This needs some additional documentation added in auth.md to help users searching for more info on the error message.

sbwsg:
/hold cancel

I've added a section to our ever-growing auth.md to try and capture the reasons this warning appears and how to diagnose it.

vdemeester (approved):
/meow

PR #3437, opened 2020-10-22
distinguish task dependencies: ordering vs resource

jerop:
# Changes

there are two types of dependencies between tasks:
- _resource-dependency_: based on resources needed from parent task,
which includes results, workspaces and resources
- _ordering-dependency_: based on runAfter which provides sequencing of
tasks when there may not be resource dependencies

however, it is currently difficult to distinguish between the two

this PR provides a way to know the dependency type and will be useful
for future work such as allowing execution of ordering-dependent tasks
when continueAfterSkip is set to true

/kind misc

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_types.go
@@ -170,23 +170,31 @@ func (pt PipelineTask) HashKey() string {
 
 func (pt PipelineTask) Deps() []string {
 	deps := []string{}
-	deps = append(deps, pt.RunAfter...)
+
+	deps = append(deps, pt.resourceDeps()...)
+	deps = append(deps, pt.orderingDeps()...)
+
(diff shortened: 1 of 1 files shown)


sbwsg (approved):
This looks good to me.  I do wonder if these need to be public funcs though?  If we keep them private I think we still get the benefit of their naming to help distinguish purpose but we can also consider them an "implementation detail" of `Deps()` and avoid testing them right now.  What do you think?

jerop:
> This looks good to me. I do wonder if these need to be public funcs though? If we keep them private I think we still get the benefit of their naming to help distinguish purpose but we can also consider them an "implementation detail" of `Deps()` and avoid testing them right now. What do you think?

agreed, i need to use the functions in `pipeline_validation.go` only so that should work

PR #3454, opened 2020-10-26
Add an unsuccessful TaskRun testcase in conformance tests

yaoxiaoqi:
# Changes

Add an unsuccessful TaskRun testcase in conformance tests

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes
```release-note
NONE
```

### DIFF
--- a/test/conformance_test.go
+++ b/test/conformance_test.go
@@ -20,14 +20,19 @@ package test
 
 import (
 	"context"
+	"strings"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/pkg/apis"
 	knativetest "knative.dev/pkg/test"
 )
+
+type conditionFn func(name string) ConditionAccessorFn
 
 func TestTaskRun(t *testing.T) {
 	ctx := context.Background()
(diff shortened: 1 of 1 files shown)


test/conformance_test.go
```diff
@@ -38,33 +42,119 @@ func TestTaskRun(t *testing.T) {
...
+		fn:                      TaskRunSucceed,
+		expectedConditionStatus: corev1.ConditionTrue,
+		expectedStepState: []v1beta1.StepState{{
+			ContainerState: corev1.ContainerState{
+				Terminated: &corev1.ContainerStateTerminated{
+					ExitCode: 0,
+					Reason:   "Completed",
+				},
+			},
+			Name:          "unnamed-0",
```
ImJasonH, line 75: So the [API spec](https://github.com/tektoncd/pipeline/blob/master/docs/api-spec.md#stepstate) doesn't specify whether implementations have to populate `Name` or `ContainerName`, or how should populate them when the user's requested step wasn't given a name, so we shouldn't require them in this test.

That being said, these seem like useful things to include in future changes to the spec, and cover in conformance tests. But for now, we should ignore them.

test/conformance_test.go
```diff
@@ -38,33 +42,119 @@ func TestTaskRun(t *testing.T) {
...
 
-	if err := WaitForTaskRunState(ctx, c, trName, TaskRunSucceed(trName), "WaitTaskRunDone"); err != nil {
-		t.Errorf("Error waiting for TaskRun to finish: %s", err)
-		return
-	}
-	tr, err := c.TaskRunClient.Get(ctx, trName, metav1.GetOptions{})
-	if err != nil {
-		t.Fatalf("Failed to get TaskRun `%s`: %s", trName, err)
+			ignoreTerminatedFields := cmpopts.IgnoreFields(corev1.ContainerStateTerminated{}, "StartedAt", "FinishedAt", "ContainerID")
+			ignoreStepFields := cmpopts.IgnoreFields(v1beta1.StepState{}, "ImageID")
```
ImJasonH, line 149: ImageID _is_ REQUIRED by the spec. It's fine to ignore it for this change, but we should have a TODO to fix it before considering these tests done.

To make it easier to test, we can specify each step image by digest (`busybox@sha256:a9286defaba7b3a519d585ba0e37d0b2cbee74ebfe590960b0b1d6a5e97d1e1d`) and the step's imageID should be the digest portion (`sha256:a9286defaba7b3a519d585ba0e37d0b2cbee74ebfe590960b0b1d6a5e97d1e1d`)

That can happen in another change, but we should have a TODO for now.

test/conformance_test.go
```diff
@@ -38,33 +42,119 @@ func TestTaskRun(t *testing.T) {
...
+			if d := cmp.Diff(tr.Status.Steps, tc.expectedStepState, ignoreTerminatedFields, ignoreStepFields); d != "" {
+				t.Fatalf("-got, +want: %v", d)
+			}
+		})
 	}
+}
 
+func checkRequiredFieldsInTaskRun(t *testing.T, tr *v1beta1.TaskRun, cs corev1.ConditionStatus) {
```
ImJasonH, line 157: I'm not sure this method adds much, since it's only called in one place. It could just be inlined above with a comment, like `// Check required fields.`

test/conformance_test.go
yaoxiaoqi: I see. But I'm going to reserve the StepState struct although we only check one field of it for now. It might be used in the future.

test/conformance_test.go
yaoxiaoqi: Sure. I added the check for ImageID in this change.

ImJasonH (commented):
/lgtm

Issue #3460, opened 2020-10-26
TestPipelineRunTimeout is flaky


mattmoor:
# Expected Behavior

TestPipelineRunTimeout consistently passes with the properly observed failure mode.

# Actual Behavior

TestPipelineRunTimeout often observes a "Failed" status on the PipelineRun(!!!).

# Steps to Reproduce the Problem

Run the test a lot.

I will post my analysis here shortly, as I believe I know WHY this is happening.

mattmoor:
Every so often this test will fail with:
```
timeout_test.go:102: Error waiting for PipelineRun pipeline-run-timeout-qaakhkfe to finish: "pipeline-run-timeout-qaakhkfe" completed with the wrong reason: Failed
```

The setup is a task that sleeps for 10s and a pipeline that has a 5s timeout:
https://github.com/tektoncd/pipeline/blob/adc4cfa733af86874017d07459cc5ad62bf82bd3/test/v1alpha1/timeout_test.go#L49-L61

This check expects the "Reason" to be "PipelineRunTimeout":
https://github.com/tektoncd/pipeline/blob/adc4cfa733af86874017d07459cc5ad62bf82bd3/test/v1alpha1/timeout_test.go#L100-L103

However, when the test intermittently fails the status shows the TaskRun timed out, but the PipelineRun Failed:
```
        spec:
          pipelineRef:
            name: pipeline-run-timeout-kjlunngp
          serviceAccountName: default
          timeout: 5s     <-- Expected to timeout (the task sleeps for 10s)
        status:
          completionTime: "2020-10-24T12:37:13Z"
          conditions:
          - lastTransitionTime: "2020-10-24T12:37:13Z"
            message: 'Tasks Completed: 1 (Failed: 1, Cancelled 0), Skipped: 0'
            reason: Failed    <-- We get Failed instead of PipelineRunTimeout
            status: "False"
            type: Succeeded
          pipelineSpec:
            tasks:
            - name: foo
              taskRef:
                kind: Task
                name: pipeline-run-timeout-lfsmutbs
          startTime: "2020-10-24T12:37:09Z"
          taskRuns:
            pipeline-run-timeout-mncrmugc-foo-74kf8:
              pipelineTaskName: foo
              status:
                completionTime: "2020-10-24T12:37:13Z"
                conditions:
                - lastTransitionTime: "2020-10-24T12:37:13Z"
[...]

PR #3461, opened 2020-10-26
Adjust PipelineRun's StartTime based on TaskRun state.

mattmoor:
# Changes

Occasionally, it is possible for us to be reconciling a PipelineRun and have the
status we intend to report reflect an inaccurate StartTime (see issue for
details).  This corrects for those circumstances by ensuring that the StartTime
we report for a PipelineRun is never later than the smallest CreationTimestamp
of a child TaskRun.

Fixes: https://github.com/tektoncd/pipeline/issues/3460

/kind bug

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Fixes a bug where PipelineRun may report as Failed when it really timed out.
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
@@ -73,6 +73,10 @@ func (state PipelineRunState) IsBeforeFirstTaskRun() bool {
 }
 
 // AdjustStartTime adjusts potential drift in the PipelineRun's start time.
+//
+// The StartTime will only adjust earlier, so that the PipelineRun's StartTime
+// is no later than any of its constituent TaskRuns.
+//
 // This drift could be due to us either failing to record the Run's start time
 // previously, or our own failure to observe a prior update before reconciling
 // the resource again.
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
@@ -23,6 +23,7 @@ import (
 	"github.com/tektoncd/pipeline/pkg/reconciler/pipeline/dag"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/apis"
 )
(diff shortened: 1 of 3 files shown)


pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -71,6 +72,23 @@ func (state PipelineRunState) IsBeforeFirstTaskRun() bool {
 	return true
 }
 
+// AdjustStartTime adjusts potential drift in the PipelineRun's start time.
```
ImJasonH, line 75: In addition to stating why we do this, can you add a sentence describing how we adjust, to make this easier to read?

ImJasonH (commented):
Does this need a test?

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
mattmoor: tweaked wording, PTAL 🙏

pritidesai:
> Does this need a test?

yes please, a test would be great. Does the same bug apply to `taskRun` start time?

mattmoor:
I pushed unit tests for AdjustStartTime.

> Does the same bug apply to taskRun start time

@pritidesai I haven't looked at the TaskRun StartTime logic.  In principle it could similarly benefit from rationalizing its StartTime with the `metadata.creationTimestamp` of its child resources (e.g. Pods), but AFAIK it shouldn't manifest in failures the way it does in the linked bug for PipelineRun -> TaskRun because it is a "leaf" of the timeout evaluation chain.  So it's at least more tolerant of minor drift.

pritidesai:
wow that was quick @mattmoor 🙏 

So `taskRun` has `initializeConditions()` if it hasn't started and updating startTime in case of any such gap:

https://github.com/tektoncd/pipeline/blob/adc4cfa733af86874017d07459cc5ad62bf82bd3/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L256-L273

https://github.com/tektoncd/pipeline/blob/adc4cfa733af86874017d07459cc5ad62bf82bd3/pkg/reconciler/taskrun/taskrun.go#L92-L99

which means, we might not see this bug in `taskRun` 🤔

mattmoor:
@pritidesai that same logic is in PipelineRun (the initialization and clamping to its own creation timestamp), but the question is what outside resources we'd want to clamp the StartTime on.  Maybe the Pod?

pritidesai:
> @pritidesai that same logic is in PipelineRun (the initialization and clamping to its own creation timestamp), but the question is what outside resources we'd want to clamp the StartTime on. Maybe the Pod?

yup I noticed `pipelineRun` having the same logic 😜  yup podCreation time sounds right.

pritidesai (approved):
thanks @mattmoor 

I was thinking of moving updating `pr.Status` to a separate function for readability, we have many different `status` fields being updated here, will create a refactoring PR.

vdemeester (commented):
/lgtm

Issue #3133, opened 2020-08-25
Implement Custom Tasks


ImJasonH:
See [TEP-0002](https://github.com/tektoncd/community/blob/master/teps/0002-custom-tasks.md)

- Allow `TaskRef`s in `PipelineTask`s that specify an `apiVersion` and  `kind`
- Create `Run`s for those `Ref`s owned by the `PipelineRun`
- Collect `RunStatus`es in the `status` of the owning `PipelineRun`
- Simple end-to-end test demonstrating a `Run` being created from a `PipelineRun`
- Document Custom Task controller protocol and best practices
- Provide a sample Custom Task controller (based on knative/pkg? kubebuilder?)

/assign @ImJasonH

sbwsg:
Moving to 0.18 as there's still some work being made towards this goal.

dibyom:
Ref: https://github.com/tektoncd/pipeline/pull/3463

vdemeester:
#3463 has been merged, should we close this ?

afrittoli:
Closing this as it has been implemented. Follow up work can be treated separately.

PR #757, opened 2019-04-13
Update TaskRun status even if PipelineRun is done

dicarlo2:
# Changes

Due to the DAG execution model, it's possible for the PipelineRun to error while other TaskRuns are still executing. This change resolves a bug where the status for the other TaskRuns would not propagate to the PipelineRun if it has failed.

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md#commit-messages)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md)
for more details._

# Release Notes

```
Fix propagation of concurrent TaskRun status to failed PipelineRuns
```

PR #3463, opened 2020-10-27
Integrate custom tasks into Pipelines

GregDritschler:
# Changes

A Pipeline Task now can reference a custom task.  This causes the PipelineRun reconciler to create a Run instead of a TaskRun.
Pipeline parameters can be passed to the custom task.  The custom task's results can be used by other Pipeline tasks and by the Pipeline results.

/kind feature

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
In package pkg/apis/pipeline/v1alpha1, the type of field RunStatusFields.Results has changed from
v1beta1.TaskRunResult to v1alpha1.RunResult.
```


GregDritschler:
@ImJasonH @vdemeester This PR supercedes Jason's original PR #3115.  I made the following changes:

* Rebased to latest master.
* Resolved most of the TODOs.
* Resolved the import cycle problem getting the v1beta1 PipelineRun status to include the v1alpha1 Run status.
* Added code to propagate PR cancel request to Runs.
* Added code to handle result references to results produced by custom tasks.
* Improved unit test coverage for Runs.
* Updated docs.

The main TODO which I did not address was to handle Conditions in pipeline tasks that use custom tasks. At the time of Jason's PR, I believe that "when" support was not yet merged. Now that "when" support is merged and Conditions are being deprecated, I would like to propose that Conditions NOT be supported in pipeline tasks that use custom tasks. If everyone agrees then I can update this PR to validate that restriction or I can do that in a follow-on PR.

Thanks in advance for your review.

afrittoli (commented):
Thanks for this, I was really looking forward to it!
I'll be digging into the PR in the next days!

Quick question, without having looked at the code: would it make sense to split the part that solves the import cycle in a dedicated initial PR to reduce the size of this one?

pkg/apis/run/v1alpha1/runstatus_types.go
```diff
@@ -0,0 +1,144 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"knative.dev/pkg/apis"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+)
+
+// This package exists to avoid an import cycle between v1alpha1 and v1beta1.
+// It contains common definitions needed by v1alpha1.Run and v1beta1.PipelineRun.
```
vdemeester, line 30: I don't think this is required **and** I think we should move **all** types related to `Run` types, not just the one that did pause problem for import cycle.

vdemeester: This means `Run`, `RunSpec`, `RunList`, … types

vdemeester (changes requested):
We should move all `Run` related types to `pkg/apis/run/v1alpha1` instead of just the one that were a problem for import cycle. This would allow us to *still* not import anything from `"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"` but `"github.com/tektoncd/pipeline/pkg/apis/run/v1alpha1"` instead — similar to what we do with `pkg/apis/resource/v1alpha1`.

pkg/apis/run/v1alpha1/runstatus_types.go
GregDritschler: > We should move all Run related types to pkg/apis/run/v1alpha1 instead of just the one that were a problem for import cycle. This would allow us to still not import anything from "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1" but "github.com/tektoncd/pipeline/pkg/apis/run/v1alpha1" instead — similar to what we do with pkg/apis/resource/v1alpha1

Okay, so I should not even leave behind any aliases in "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"?  In other words it's a completely incompatible change of the package.  It doesn't matter to me, I'll do whatever everyone wants here.

ImJasonH:
Thanks so much for doing this Greg! 🎉 

> Now that "when" support is merged and Conditions are being deprecated, I would like to propose that Conditions NOT be supported in pipeline tasks that use custom tasks.

This seems reasonable to me, so long as it's documented and produces a descriptive error message. I'd prefer it to happen in a followup PR though to keep this one relatively narrowly scoped.

pkg/apis/run/v1alpha1/runstatus_types.go
ImJasonH: Our policy is that we don't care about breaking changes that only affect Go clients. This is especially true for anything related to Runs, which are alpha and haven't had much real-world usage outside of, honestly, you or me. 😆

vdemeester: @GregDritschler yeah I think it's fine. The alias can stay there if we feel there is a need for them there… but I am fine if they are not there any more, as @ImJasonH said, we don't care too much about breaking changes that only affect Go clients, **especially** for `v1alpha` :upside_down_face:

pkg/apis/run/v1alpha1/runstatus_types.go
```diff
@@ -0,0 +1,144 @@
...
+// MarkRunSucceeded changes the Succeeded condition to True with the provided reason and message.
+func (r *RunStatus) MarkRunSucceeded(reason, messageFormat string, messageA ...interface{}) {
+	runCondSet.Manage(r).MarkTrueWithReason(apis.ConditionSucceeded, reason, messageFormat, messageA...)
+	succeeded := r.GetCondition(apis.ConditionSucceeded)
+	r.CompletionTime = &succeeded.LastTransitionTime.Inner
+}
+
+// MarkRunFailed changes the Succeeded condition to False with the provided reason and message.
+func (r *RunStatus) MarkRunFailed(reason, messageFormat string, messageA ...interface{}) {
+	runCondSet.Manage(r).MarkFalse(apis.ConditionSucceeded, reason, messageFormat, messageA...)
```
ImJasonH, line 116: Should this be MarkFalseWithReason to match MarkTrueWithReason above? 🤔

GregDritschler: Yeah I had noticed that but this is how the functions are defined in the knative package.

```
	// MarkTrue sets the status of t to true, and then marks the happy condition to
	// true if all dependents are true.
	MarkTrue(t ConditionType)

	// MarkTrueWithReason sets the status of t to true with the reason, and then marks the happy
	// condition to true if all dependents are true.
	MarkTrueWithReason(t ConditionType, reason, messageFormat string, messageA ...interface{})

	// MarkUnknown sets the status of t to Unknown and also sets the happy condition
	// to Unknown if no other dependent condition is in an error state.
	MarkUnknown(t ConditionType, reason, messageFormat string, messageA ...interface{})

	// MarkFalse sets the status of t and the happy condition to False.
	MarkFalse(t ConditionType, reason, messageFormat string, messageA ...interface{})
```

pkg/apis/run/v1alpha1/runstatus_types.go
GregDritschler: Moving the RunSpec is harder because it uses v1beta1.Params and v1alpha1.TaskRef.  This means a lot of refactoring which must be done carefully to avoid breaking these widely used types.  I discussed offline with Vincent and he agreed this could be considered for a follow on to this PR.

bobcatfish:
@GregDritschler could you fill in the checklist:

image

Discussed in OWNERS meeting today, sounds like merging this fixes #3133 (correct me if I'm wrong @ImJasonH @GregDritschler !)

GregDritschler:
> Discussed in OWNERS meeting today, sounds like merging this fixes #3133 (correct me if I'm wrong @ImJasonH @GregDritschler !)

There are a few things left to do (as separate PRs since this one is already very big):
* Disallow use of deprecated Conditions with custom tasks
* Enable use of service account name and workspaces with custom tasks
* Provide a sample custom task controller (I have started one)

vincent-pli:
What's the process of this implements? expect to try the `customer task`.

ImJasonH (commented):
/lgtm

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -284,69 +305,84 @@ func ValidateServiceaccountMapping(p *v1beta1.PipelineSpec, pr *v1beta1.Pipeline
 // the spec. If it is unable to retrieve an instance of a referenced Task, it  will return
 // an error, otherwise it returns a list of all of the Tasks retrieved.  It will retrieve
 // the Resources needed for the TaskRun using the mapping of providedResources.
+//
+// TODO: The proliferation of Get* methods here is a smell. This method should
+// take a versioned clientset instead which includes methods to retrieve these
+// resources, with generated fakes for testing.
```
vdemeester, line 311: I agree with the start of the comment (it's a smell), not with the rest though (the *solution*). "versioned clientset" assumes the `get` methods only use k8s clients, which is not the case (with Tekton OCI bundles, we do fetch through a very different method :upside_down_face:)

vdemeester (commented):
@tektoncd/core-collaborators in the light of discussion around ["feature-flags"](https://github.com/tektoncd/pipeline/pull/3492#issuecomment-721880114), should this feature be under a feature flag or not ? :upside_down_face:

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
GregDritschler: This is Jason's comment.  @ImJasonH should I change or remove the comment?

ImJasonH: I'm fine to remove it. I agree with Vincent that with the addition of OCI bundles the prescribed remedy for the smell might not be correct.

GregDritschler:
> @tektoncd/core-collaborators in the light of discussion around ["feature-flags"](https://github.com/tektoncd/pipeline/pull/3492#issuecomment-721880114), should this feature be under a feature flag or not ? 🙃

I have added the feature flag.

animeshsingh:
@vdemeester @afrittoli this is urgent for us - any blocker now to merge?

pritidesai:
Couple of questions:

1. When a `pipelineRun` is created with a custom task but the custom controller does not update `pr.Status` for some reason, does `pipelineRun` eventually time out? or does it keep waiting indefinitely?

2. I wrote a simple `pipeline` with one custom task and created a pipelineRun with that pipeline. Tekton controller logs says `Creating a new Run object pipelinerun-with-custom-task-run-custom-task-g4dcc` and `creationTimestamp` of that run object is set:

```
kubectl get run pipelinerun-with-custom-task-run-custom-task-g4dcc -o json
{
    "apiVersion": "tekton.dev/v1alpha1",
    "kind": "Run",
    "metadata": {
        "annotations": {
            "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"tekton.dev/v1beta1\",\"kind\":\"Pipeline\",\"metadata\":{\"annotations\":{},\"name\":\"pipeline-with-custom-task\",\"namespace\":\"default\"},\"spec\":{\"tasks\":[{\"name\":\"run-custom-task\",\"taskRef\":{\"apiVersion\":\"example.dev/v1alpha1\",\"kind\":\"Example\"}}]}}\n"
        },
        "creationTimestamp": "2020-11-19T06:04:00Z",
```

but no start time:

```
kubectl get run
NAME                                                                                     SUCCEEDED   REASON   STARTTIME   COMPLETIONTIME
pipelinerun-with-custom-task-run-custom-task-g4dcc
```

Is it again custom controller updating the status of this `run` along with `pipelineRun`?

pritidesai:
> When a pipelineRun is created with a custom task but the custom controller does not update pr.Status for some reason, does pipelineRun eventually time out? or does it keep waiting indefinitely?

Alright, so, pipelineRun eventually timed out, updating the `pr.Status` with ` "reason": "PipelineRunTimeout"` and `completionTime` along with the run. But listing the run still shows blank for `REASON`, `COMPLETIONTIME` etc

ImJasonH:
>  When a pipelineRun is created with a custom task but the custom controller does not update pr.Status for some reason, does pipelineRun eventually time out? or does it keep waiting indefinitely?

We decided in the TEP not to have the PipelineRun controller be responsible for custom task Runs beyond creating them. So if there's no custom task controller handling and updating that Run object, it'll look like the Run just never starts until the PipelineRun itself times out. That's also why there's no `startTime`, there's nothing watching to update it.

We can revisit that decision and have the PipelineRun controller enforce some kind of "are you there?" check on custom task controllers, such that if a custom task Run isn't started within 30s or something (configurable) the PipelineRun controller can kill it and fail the PipelineRun as a safety mechanism.

I definintely don't think that needs to happen in this PR, and should probably only be decided on after users have played with it to see how big of a deal it is to not have the safety mechanism.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -74,52 +79,69 @@ func (t ResolvedPipelineRunTask) IsDone(facts *PipelineRunFacts) bool {
...
+// IsCustomTask returns true if the PipelineTask references a Custom Task.
+func (t ResolvedPipelineRunTask) IsCustomTask() bool {
+	return t.CustomTask
+}
+
 // IsSuccessful returns true only if the taskrun itself has completed successfully
 func (t ResolvedPipelineRunTask) IsSuccessful() bool {
-	if t.TaskRun == nil {
-		return false
+	if !t.IsCustomTask() {
```
pritidesai, line 89: Can we switch the condition from `!t.IsCustomTask()` to `t.IsCustomTask()` wherever possible, for example:

```	
if t.IsCustomTask() {
		return t.Run != nil && t.Run.IsSuccessful()
	}
	if t.TaskRun == nil {
		return false
	}
	return t.TaskRun.IsSuccessful()
```

config/config-feature-flags.yaml
```diff
@@ -82,3 +82,8 @@ data:
   # This is an experimental feature and thus should still be considered
   # an alpha feature.
   enable-tekton-oci-bundles: "false"
+  # Setting this flag to "true" enables the use of custom tasks from
+  # within pipelines.
+  # This is an experimental feature and thus should still be considered
+  # an alpha feature.
+  enable-custom-tasks: "false"
```
pritidesai, line 89: Should we call it `enable-custom-run` since it can include `task` and `pipeline` in future?

pkg/apis/pipeline/v1alpha1/run_types.go
```diff
@@ -18,14 +18,12 @@ package v1alpha1
 
 import (
 	"fmt"
-	"time"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	v1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	runv1alpha1 "github.com/tektoncd/pipeline/pkg/apis/run/v1alpha1"
```
afrittoli, line 24: I'm trying to review this change, but I don't see this package in the PR?
I wonder if it was missed in the commit...

afrittoli (commented):
Thanks @GregDritschler - I don't see the pkg/apis/run package in the PR, is it perhaps missing from the commit?

pkg/apis/pipeline/v1alpha1/run_types.go
afrittoli: it's there and I don't know how I missed it? please ignore me 🤦

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -102,27 +102,61 @@ func validatePipelineTaskName(name string) *apis.FieldError {
 func validatePipelineTask(ctx context.Context, t PipelineTask, taskNames sets.String) *apis.FieldError {
```
pritidesai, line 102: This function is used to validate `tasks` and tasks under `finally` section. Is custom task supported as part of `finally`?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -580,10 +591,18 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
 			continue
 		}
 		if rprt.ResolvedConditionChecks == nil || rprt.ResolvedConditionChecks.IsSuccess() {
-			rprt.TaskRun, err = c.createTaskRun(ctx, rprt, pr, as.StorageBasePath(pr))
-			if err != nil {
-				recorder.Eventf(pr, corev1.EventTypeWarning, "TaskRunCreationFailed", "Failed to create TaskRun %q: %v", rprt.TaskRunName, err)
-				return fmt.Errorf("error creating TaskRun called %s for PipelineTask %s from PipelineRun %s: %w", rprt.TaskRunName, rprt.PipelineTask.Name, pr.Name, err)
+			if !rprt.IsCustomTask() {
```
pritidesai, line 594: Similar to the comment above, can we please switch the condition check from `!customTask` to `customTask`?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -929,28 +983,42 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
 		}
 	} else {
-		prStatus.TaskRuns = make(map[string]*v1beta1.PipelineRunTaskRunStatus)
+		pr.Status.TaskRuns = make(map[string]*v1beta1.PipelineRunTaskRunStatus)
 	}
 	// Loop over all the TaskRuns associated to Tasks
 	for _, taskrun := range trs {
+		// Only process TaskRuns that are owned by this PipelineRun.
+		// This skips TaskRuns that are indirectly created by the PipelineRun (e.g. by custom tasks).
+		if len(taskrun.OwnerReferences) < 1 || taskrun.OwnerReferences[0].UID != pr.ObjectMeta.UID {
```
pritidesai, line 1018: What is the purpose of this check here? Runs for custom tasks are created under runs not taskruns so they should not be part of this list here 🤔

pkg/apis/run/v1alpha1/runstatus_types.go
```diff
@@ -0,0 +1,144 @@
...
+var runCondSet = apis.NewBatchConditionSet()
+
+// GetCondition returns the Condition matching the given type.
+func (r *RunStatus) GetCondition(t apis.ConditionType) *apis.Condition {
+	return runCondSet.Manage(r).GetCondition(t)
+}
+
+// InitializeConditions will set all conditions in runCondSet to unknown for the PipelineRun
+// and set the started time to the current time
+func (r *RunStatus) InitializeConditions() {
```
pritidesai, line 83: these functions are not used anywhere, do we need them? 

* InitializeConditions
* SetCondition
* MarkRunSucceeded
* MarkRunFailed
* MarkRunRunning

pkg/apis/pipeline/v1beta1/pipeline_validation.go
GregDritschler: Yes it is supported in finally.

pkg/reconciler/pipelinerun/pipelinerun.go
GregDritschler: If a PipelineRun includes a Run that creates TaskRuns (such as the looping task), the PipelineRun's labels will be propagated to the TaskRuns created by the Run.  We don't want the PipelineRun adopting those TaskRuns.  It should only adopt the ones it directly created as part of the pipeline.

pkg/apis/run/v1alpha1/runstatus_types.go
GregDritschler: They are used by controllers implementing a custom task.

pritidesai: thanks, can we add such controller (can be separate PR) or a reference to such controller (if it exists somewhere may be in experimental repo) in doc?

GregDritschler: There is the task-loops controller in experimental already and I have created a simpler sample controller which I will be submitting separately.

config/config-feature-flags.yaml
vdemeester: @GregDritschler @ImJasonH ok with `enable-custom-run` as feature-flag name ?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
vdemeester: Agreed.

config/config-feature-flags.yaml
GregDritschler: "Custom run" sounds funny to me.  The feature is called "custom tasks" not "custom run".  Also the flag does not affect whether Run can be used directly.  It only controls whether pipelines can create runs.

afrittoli: I don't have a strong feeling about this. I think that even though the CRD is called `Run`, "custom-task" is more readable as config option name.  And in future we might have the `PipelineTask` custom task :D

pkg/apis/pipeline/v1alpha1/run_types.go
```diff
@@ -75,85 +73,16 @@ func (rs RunSpec) GetParam(name string) *v1beta1.Param {
...
+const (
+	// RunReasonCancelled must be used in the Condition Reason to indicate that a Run was cancelled.
+	RunReasonCancelled = "RunCancelled"
+)
 
-	// RunStatusFields inlines the status fields.
-	RunStatusFields `json:",inline"`
-}
+// RunStatus defines the observed state of Run.
+type RunStatus = runv1alpha1.RunStatus
```
afrittoli, line 82: Any special reason for moving only the status part to its own package?
I think it would make sense to have them all in the same package outside of pipeline?

GregDritschler: This was discussed earlier with @vdemeester:

> Moving the RunSpec is harder because it uses v1beta1.Params and v1alpha1.TaskRef. This means a lot of refactoring which must be done carefully to avoid breaking these widely used types. I discussed offline with Vincent and he agreed this could be considered for a follow on to this PR.

Personally I feel there is not much value in moving the rest of it.  There's bigger fish to fry.

afrittoli: Ah, I missed that discussion, I agree.
I small request though, as reading the code without this context might have the same question. Could you file an issue about the follow-up refactor, and add a comment into the code to refer to it? (A follow up PR also works)

vincent-pli:
@GregDritschler @ImJasonH I'm not check the details, just a quick question: does `task` in `finally` section also could point to a `customer task`?   since `finally` can not support sequence run, so I guess `customer task` could help on this part.

afrittoli (commented):
> @GregDritschler @ImJasonH I'm not check the details, just a quick question: does `task` in `finally` section also could point to a `customer task`? since `finally` can not support sequence run, so I guess `customer task` could help on this part.

Yes, there is no restriction against using custom tasks in `finally`

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -337,6 +339,10 @@ type PipelineRunStatusFields struct {
 	// +optional
 	TaskRuns map[string]*PipelineRunTaskRunStatus `json:"taskRuns,omitempty"`
 
+	// map of PipelineRunRunStatus with the run name as the key
+	// +optional
+	Runs map[string]*PipelineRunRunStatus `json:"runs,omitempty"`
```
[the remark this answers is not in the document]
ImJasonH: TaskRuns is a map, so it should be JSON/YAML-encoded alphabetized by keys, not in insertion order. In any case, we shouldn't guarantee any ordering so we can change it in the future, and rely on clients to do smart things if they want to.

+1 about combinding these maps though, it'd be nice but it's also a bit clunky given Go's type system. 😕

config/config-feature-flags.yaml
chmouel: `custom-tasks` sounds more explicit to me, `custom-runs` feels like a feature flag for some custom *execution* of how task are run and not necessary about Pipeline,Task/Runs

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -188,6 +189,10 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 			logger.Errorf("Failed to update TaskRun status for PipelineRun %s: %v", pr.Name, err)
 			return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
 		}
+		if err := c.updateRunsStatusDirectly(pr); err != nil {
+			logger.Errorf("Failed to update Run status for PipelineRun %s: %v", pr.Name, err)
+			return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
+		}
```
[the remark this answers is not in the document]
GregDritschler: Yes I wrestled with this question and decided it was okay to run it without the feature flag check for the reasons you mentioned.

afrittoli (commented):
Thanks for this. 
I mostly finished going through the code.

One small thing about the tests is that it seems that most cases deal with either taskruns or runs. It might be good to have some tests with both.

I would like to know your thoughts about using an interface instead of adding extra fields to the resolved pipeline run task, but otherwise it looks good. It will need to be rebased.

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
afrittoli: ok - it might be worth adding a comment in the code for this

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -188,6 +189,10 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 			logger.Errorf("Failed to update TaskRun status for PipelineRun %s: %v", pr.Name, err)
 			return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
 		}
+		if err := c.updateRunsStatusDirectly(pr); err != nil {
```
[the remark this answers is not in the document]
GregDritschler: Yes I had the same thoughts about this code.  I had tracked it down to PR https://github.com/tektoncd/pipeline/pull/757.
In theory this code is no longer needed with the "stopping" mode in PipelineRuns but there is at least one case I know of which doesn't make use of the stopping mode, which is canceling the PR.  That doesn't wait for the TRs to finish.  I guess it's debatable whether the PR status needs to have the canceled TR status in it or just the status at time of cancel.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -61,8 +62,12 @@ func (e *ConditionNotFoundError) Error() string {
 // ResolvedPipelineRunTask contains a Task and its associated TaskRun, if it
 // exists. TaskRun can be nil to represent there being no TaskRun.
 type ResolvedPipelineRunTask struct {
```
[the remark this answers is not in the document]
GregDritschler: No I hadn't considered doing that. This is the direction Jason went and I didn't look into redoing it.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -61,8 +62,12 @@ func (e *ConditionNotFoundError) Error() string {
...
-	TaskRunName           string
-	TaskRun               *v1beta1.TaskRun
+	TaskRunName string
+	TaskRun     *v1beta1.TaskRun
+	// If the PipelineTask is a Custom Task, RunName and Run will be set.
+	CustomTask            bool
+	RunName               string
```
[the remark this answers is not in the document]
GregDritschler: Do you mean the comment line or something else?

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -161,21 +171,57 @@ func (state PipelineRunState) GetTaskRunsStatus(pr *v1beta1.PipelineRun) map[str
...
-		if _, ok := candidateTasks[t.PipelineTask.Name]; ok && t.TaskRun != nil {
-			status := t.TaskRun.Status.GetCondition(apis.ConditionSucceeded)
-			if status != nil && status.IsFalse() {
-				if !(t.TaskRun.IsCancelled() || status.Reason == v1beta1.TaskRunReasonCancelled.String() || status.Reason == ReasonConditionCheckFailed) {
-					if len(t.TaskRun.Status.RetriesStatus) < t.PipelineTask.Retries {
-						tasks = append(tasks, t)
+		if _, ok := candidateTasks[t.PipelineTask.Name]; ok {
+			if t.TaskRun == nil && t.Run == nil {
+				tasks = append(tasks, t)
+			} else if t.TaskRun != nil { // only TaskRun currently supports retry
```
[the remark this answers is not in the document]
GregDritschler: This test is a bit different. It's testing if the TaskRun/Run exist, not whether it's intended to be one or the other.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -375,12 +421,12 @@ func (facts *PipelineRunFacts) getPipelineTasksCount() pipelineRunStatusCount {
...
 			s.Succeeded++
-		// increment failure counter since the task has failed
-		case t.IsFailure():
-			s.Failed++
 		// increment cancelled counter since the task is cancelled
 		case t.IsCancelled():
 			s.Cancelled++
+		// increment failure counter since the task has failed
+		case t.IsFailure():
+			s.Failed++
```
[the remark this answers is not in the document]
GregDritschler: The original order checked isFailure() before isCancelled().  This had a couple of issues.
1. Cancelled Runs are also failed so it was impossible to count them as cancelled.  
2. TaskRuns have a quirk where isFailure() would say false if the TaskRun is eligible for another retry, so they would be counted as cancelled only if they were cancelled with no retries remaining.

I reversed the order of the checks which seemed to make sense -- cancel is more specialized than failure so it should be checked first.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
GregDritschler: Done.

pkg/reconciler/pipelinerun/pipelinerun.go
GregDritschler: Done.

pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -57,11 +66,20 @@ func cancelPipelineRun(ctx context.Context, logger *zap.SugaredLogger, pr *v1bet
...
 	}
+	// Loop over the Runs in the PipelineRun status.
+	for runName := range pr.Status.Runs {
+		logger.Infof("cancelling Run %s", runName)
+
+		if _, err := clientSet.TektonV1alpha1().Runs(pr.Namespace).Patch(ctx, runName, types.JSONPatchType, cancelRunPatchBytes, metav1.PatchOptions{}, ""); err != nil {
+			errs = append(errs, fmt.Errorf("Failed to patch Run `%s` with cancellation: %s", runName, err).Error())
+			continue
+		}
+	}
```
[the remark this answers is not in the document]
GregDritschler: Done.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -929,28 +983,42 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
 	} else {
-		prStatus.TaskRuns = make(map[string]*v1beta1.PipelineRunTaskRunStatus)
+		pr.Status.TaskRuns = make(map[string]*v1beta1.PipelineRunTaskRunStatus)
 	}
 	// Loop over all the TaskRuns associated to Tasks
 	for _, taskrun := range trs {
+		// Only process TaskRuns that are owned by this PipelineRun.
+		// This skips TaskRuns that are indirectly created by the PipelineRun (e.g. by custom tasks).
+		if len(taskrun.OwnerReferences) < 1 || taskrun.OwnerReferences[0].UID != pr.ObjectMeta.UID {
+			logger.Infof("Found a TaskRun %s that is not owned by this PipelineRun", taskrun.Name)
```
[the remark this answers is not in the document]
GregDritschler: Done

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1012,7 +1080,34 @@ func updatePipelineRunStatusFromTaskRuns(logger *zap.SugaredLogger, prName strin
...
+	}
+	if pr.Status.Runs == nil {
+		pr.Status.Runs = make(map[string]*v1beta1.PipelineRunRunStatus)
+	}
+	// Loop over all the Runs associated to Tasks
+	for _, run := range runs {
+		// Only process Runs that are owned by this PipelineRun.
+		// This skips Runs that are indirectly created by the PipelineRun (e.g. by custom tasks).
+		if len(run.OwnerReferences) < 1 && run.OwnerReferences[0].UID != pr.ObjectMeta.UID {
+			logger.Infof("Found a Run %s that is not owned by this PipelineRun", run.Name)
```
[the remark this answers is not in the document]
GregDritschler: Done.

GregDritschler:
I've opened issues for the follow-on work to this PR.

Custom Tasks: Disallow use of Conditions #3592
Custom Tasks: Allow use of workspaces #3593
Custom Tasks: Allow use of service accounts #3594
Custom Tasks: Support Pipeline Results #3595
Custom Tasks: Move all Run-related Go types to pkg/apis/run #3596

@vdemeester @afrittoli @pritidesai FYI.  I believe I've resolved all the review issues. If not let me know.  I'd like to get this one closed out.

vdemeester (approved):
/meow
Let's get this in for 0.19 :stuck_out_tongue_closed_eyes:

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
afrittoli: Good point about the cancel use case.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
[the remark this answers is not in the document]
afrittoli: I meant `RunName`. Since TaskRun and Run are mutually exclusive here, we just need one field for the name.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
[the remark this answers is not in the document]
afrittoli: OK, maybe something for future PRs. We might run loop tasks in future or pipelines as tasks, so it would be good to have a model that can work with an interface here.

afrittoli (commented):
Thanks for the updates and for the great work on this!

There are a few more comments that we could add to into the code to track design decisions for posterity, but nothing blocking.

/lgtm

vdemeester:
arf, @GregDritschler needs yet another rebase, sorry :sweat: :bow:

GregDritschler:
@afrittoli @vdemeester Rebase completed.

afrittoli (commented):
Thanks!
/lgtm

Issue #3465, opened 2020-10-28
PipelineRun emitting task results that were never produced


Fabian-K:
Hi,

we have a task that emits a result and a pipeline that propagates the task result using (https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#emitting-results-from-a-pipeline). In case the task failed, the PipelineRun contains the pipelineResult with the template string as a value, in this example `$(tasks.print-date.results.current-date-unix-timestamp)`

```yaml
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  name: test-run
  namespace: default
spec:
  pipelineRef:
    name: simple-pipeline
  serviceAccountName: default
  timeout: 1h0m0s
status:
  completionTime: "2020-10-28T14:48:57Z"
  conditions:
    - lastTransitionTime: "2020-10-28T14:48:57Z"
      message: 'Tasks Completed: 1 (Failed: 1, Cancelled 0), Skipped: 0'
      reason: Failed
      status: "False"
      type: Succeeded
  pipelineResults:
    - name: current-date-unix-timestamp
      value: $(tasks.print-date.results.current-date-unix-timestamp)
  pipelineSpec:
    <...>
```

IMO this is unexpected as this could be interpreted as the actual value (but i guess this is open for discussion). It however is for sure not consistent with the behavior of the task run which for this case does not include the result at all.

```yaml
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  name: test-run-print-date-t55td
  namespace: default
spec:
  resources: {}
  serviceAccountName: default
  taskRef:
    kind: Task
    name: print-date
  timeout: 1h0m0s
status:
  completionTime: "2020-10-28T14:48:57Z"
  conditions:
    - lastTransitionTime: "2020-10-28T14:48:57Z"
      message: |
        "step-print-date-unix-timestamp" exited with code 1 (image: "docker.io/library/bash@sha256:01fad26fa8ba21bce6e8c47222acfdb54649957f1e86d53a0c8e03360271abf6"); for logs run: kubectl -n default logs test-run-print-date-t55td-pod-8xwbp -c step-print-date-unix-timestamp
      reason: Failed
      status: "False"
      type: Succeeded
  podName: test-run-print-date-t55td-pod-8xwbp
  startTime: "2020-10-28T14:48:51Z"
  steps:
    - container: step-print-date-unix-timestamp
[...]

PR #3472, opened 2020-10-29
Fix bug where PipelineRun emits task results that were never produced

sbwsg:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/3465

Prior to this commit a Pipeline Result that references a Task Result of a failed TaskRun would end up with a value of the literal variable. For example: if Task "foo" failed, a Pipeline Result with value of "$(tasks.foo.results.bar)" would end up in the PipelineRun.Status.Results list with the literal value, "$(tasks.foo.results.bar)". It was therefore difficult to assess programatically whether results were populated correctly or were the result of some invalid TaskRun condition.

This commit fixes the bug by filtering out PipelineRun Results that reference failed TaskRuns. It was quite difficult to follow the flow of execution wrt PipelineRun Results and so ultimately I had to refactor the whole lot to figure out where the bug was. The final code is quite a bit shorter than the original and has improved test coverage to more robustly exercise the behaviour of PipelineResults in various failure scenarios.

# Submitter Checklist

review them:

```release-note
Fix bug where PipelineRun emits task results that were never produced. Previously a Pipeline Result that contained an invalid variable would be added to the PipelineRun with that unreplaced variable intact. Now a Pipeline Result that contains an invalid variable will not be emitted by the PipelineRun at all.
```

### DIFF
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -604,6 +604,20 @@ references the `outputValue` `Result` emitted by the `calculate-sum` `Task`.
 
 For an end-to-end example, see [`Results` in a `PipelineRun`](../examples/v1beta1/pipelineruns/pipelinerun-results.yaml).
 
+A `Pipeline Result` is not emitted if any of the following are true:
+- A `PipelineTask` referenced by the `Pipeline Result` failed. The `PipelineRun` will also
+have failed.
+- A `PipelineTask` referenced by the `Pipeline Result` was skipped.
+- A `PipelineTask` referenced by the `Pipeline Result` didn't emit the referenced `Task Result`. This
+should be considered a bug in the `Task` and [may fail a `PipelineTask` in future](https://github.com/tektoncd/pipeline/issues/3497).
+- The `Pipeline Result` uses a variable that doesn't point to an actual `PipelineTask`. This will
+result in an `InvalidTaskResultReference` validation error during `PipelineRun` execution.
+- The `Pipeline Result` uses a variable that doesn't point to an actual result in a `PipelineTask`.
+This will cause an `InvalidTaskResultReference` validation error during `PipelineRun` execution.
+
+**Note:** Since a `Pipeline Result` can contain references to multiple `Task Results`, if any of those
+`Task Result` references are invalid the entire `Pipeline Result` is not emitted.
+
 ## Configuring the `Task` execution order
 
 You can connect `Tasks` in a `Pipeline` so that they execute in a Directed Acyclic Graph (DAG).
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -22,7 +22,6 @@ import (
 	"path/filepath"
 	"reflect"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/hashicorp/go-multierror"
@@ -176,7 +175,15 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 		// and may not have had all of the assumed default specified.
 		pr.SetDefaults(contexts.WithUpgradeViaDefaulting(ctx))
 
-		c.updatePipelineResults(ctx, pr, getPipelineFunc)
+		if _, pipelineSpec, err := resources.GetPipelineData(ctx, pr, getPipelineFunc); err != nil {
+			msg := fmt.Sprintf("Failed to get Pipeline Spec to process Pipeline Results for PipelineRun %s/%s: %v", pr.Namespace, pr.Name, err)
+			logger.Error(msg)
+			logger.Warnf("An error processing Pipeline Results overwrites existing Succeeded Condition for PipelineRun %s/%s: %v", pr.Namespace, pr.Name, pr.Status.GetCondition(apis.ConditionSucceeded))
+			pr.Status.MarkFailed(ReasonCouldntGetPipeline, msg)
+		} else {
+			pr.Status.PipelineResults = resources.ApplyTaskResultsToPipelineResults(pipelineSpec.Results, pr.Status.TaskRuns)
+		}
+
 		if err := artifacts.CleanupArtifactStorage(ctx, pr, c.KubeClientSet); err != nil {
 			logger.Errorf("Failed to delete PVC for PipelineRun %s: %v", pr.Name, err)
 			return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
@@ -315,21 +322,6 @@ func (c *Reconciler) resolvePipelineState(
 		pst = append(pst, resolvedTask)
 	}
 	return pst, nil
-}
-
-func (c *Reconciler) updatePipelineResults(ctx context.Context, pr *v1beta1.PipelineRun, getPipelineFunc resources.GetPipeline) {
-	logger := logging.FromContext(ctx)
-
-	_, pipelineSpec, err := resources.GetPipelineData(ctx, pr, getPipelineFunc)
-	if err != nil {
-		logger.Errorf("Failed to determine Pipeline spec to use for pipelinerun %s: %v", pr.Name, err)
-		pr.Status.MarkFailed(ReasonCouldntGetPipeline,
-			"Error retrieving pipeline for pipelinerun %s/%s: %s",
-			pr.Namespace, pr.Name, err)
-		return
-	}
-	resolvedResultRefs := resources.ResolvePipelineResultRefs(pr.Status, pipelineSpec.Results)
-	pr.Status.PipelineResults = getPipelineRunResults(pipelineSpec, resolvedResultRefs)
 }
 
 func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, getPipelineFunc resources.GetPipeline) error {
@@ -620,27 +612,6 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
 		}
 	}
 	return nil
-}
-
-func getPipelineRunResults(pipelineSpec *v1beta1.PipelineSpec, resolvedResultRefs resources.ResolvedResultRefs) []v1beta1.PipelineRunResult {
-	var results []v1beta1.PipelineRunResult
-	stringReplacements := map[string]string{}
-
-	for _, resolvedResultRef := range resolvedResultRefs {
-		replaceTarget := fmt.Sprintf("%s.%s.%s.%s", v1beta1.ResultTaskPart, resolvedResultRef.ResultReference.PipelineTask, v1beta1.ResultResultPart, resolvedResultRef.ResultReference.Result)
-		stringReplacements[replaceTarget] = resolvedResultRef.Value.StringVal
-	}
-	for _, result := range pipelineSpec.Results {
-		in := result.Value
-		for k, v := range stringReplacements {
-			in = strings.Replace(in, fmt.Sprintf("$(%s)", k), v, -1)
-		}
-		results = append(results, v1beta1.PipelineRunResult{
-			Name:  result.Name,
-			Value: in,
-		})
-	}
-	return results
 }
 
 func (c *Reconciler) updateTaskRunsStatusDirectly(pr *v1beta1.PipelineRun) error {
--- a/pkg/reconciler/pipelinerun/resources/apply.go
+++ b/pkg/reconciler/pipelinerun/resources/apply.go
@@ -18,8 +18,11 @@ package resources
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	"knative.dev/pkg/apis"
 )
 
 // ApplyParameters applies the params from a PipelineRun.Params to a PipelineSpec.
@@ -84,6 +87,8 @@ func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResul
 	}
 }
 
+// ApplyWorkspaces replaces workspace variables in the given pipeline spec with their
+// concrete values.
 func ApplyWorkspaces(p *v1beta1.PipelineSpec, pr *v1beta1.PipelineRun) *v1beta1.PipelineSpec {
 	p = p.DeepCopy()
 	replacements := map[string]string{}
@@ -123,4 +128,74 @@ func replaceParamValues(params []v1beta1.Param, stringReplacements map[string]st
 		params[i].Value.ApplyReplacements(stringReplacements, arrayReplacements)
 	}
 	return params
+}
+
+// ApplyTaskResultsToPipelineResults applies the results of completed TasksRuns to a Pipeline's
+// list of PipelineResults, returning the computed set of PipelineRunResults. References to
+// non-existent TaskResults or failed TaskRuns result in a PipelineResult being considered invalid
+// and omitted from the returned slice. A nil slice is returned if no results are passed in or all
+// results are invalid.
+func ApplyTaskResultsToPipelineResults(results []v1beta1.PipelineResult, taskRunStatuses map[string]*v1beta1.PipelineRunTaskRunStatus) []v1beta1.PipelineRunResult {
+	taskStatuses := map[string]*v1beta1.PipelineRunTaskRunStatus{}
+	for _, trStatus := range taskRunStatuses {
+		taskStatuses[trStatus.PipelineTaskName] = trStatus
+	}
+
+	var runResults []v1beta1.PipelineRunResult = nil
+	stringReplacements := map[string]string{}
+	for _, pipelineResult := range results {
+		variablesInPipelineResult, _ := v1beta1.GetVarSubstitutionExpressionsForPipelineResult(pipelineResult)
+		validPipelineResult := true
+		for _, variable := range variablesInPipelineResult {
+			if _, isMemoized := stringReplacements[variable]; isMemoized {
+				continue
+			}
+			if resultValue := taskResultValue(variable, taskStatuses); resultValue != nil {
+				stringReplacements[variable] = *resultValue
+			} else {
+				validPipelineResult = false
+			}
+		}
+		if validPipelineResult {
+			finalValue := pipelineResult.Value
+			for variable, value := range stringReplacements {
+				v := fmt.Sprintf("$(%s)", variable)
+				finalValue = strings.Replace(finalValue, v, value, -1)
+			}
+			runResults = append(runResults, v1beta1.PipelineRunResult{
+				Name:  pipelineResult.Name,
+				Value: finalValue,
+			})
+		}
+	}
+
+	return runResults
+}
+
+// taskResultValue returns a pointer to the result value for a given task result variable. A nil
+// pointer is returned if the variable is invalid for any reason.
+func taskResultValue(variable string, taskStatuses map[string]*v1beta1.PipelineRunTaskRunStatus) *string {
+	variableParts := strings.Split(variable, ".")
+	if len(variableParts) != 4 || variableParts[0] != "tasks" || variableParts[2] != "results" {
+		return nil
+	}
+
+	taskName, resultName := variableParts[1], variableParts[3]
+
+	status, taskExists := taskStatuses[taskName]
+	if !taskExists || status.Status == nil {
+		return nil
+	}
+
+	cond := status.Status.GetCondition(apis.ConditionSucceeded)
+	if cond == nil || cond.Status != corev1.ConditionTrue {
+		return nil
+	}
+
+	for _, trResult := range status.Status.TaskRunResults {
+		if trResult.Name == resultName {
+			return &trResult.Value
+		}
+	}
+	return nil
 }
--- a/pkg/reconciler/pipelinerun/resources/apply_test.go
+++ b/pkg/reconciler/pipelinerun/resources/apply_test.go
@@ -25,8 +25,11 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	resourcev1alpha1 "github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1"
 	"github.com/tektoncd/pipeline/test/diff"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/selection"
+	"knative.dev/pkg/apis"
+	duckv1beta1 "knative.dev/pkg/apis/duck/v1beta1"
 )
 
 func TestApplyParameters(t *testing.T) {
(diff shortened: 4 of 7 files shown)


docs/pipelines.md
```diff
@@ -574,6 +574,9 @@ After a `Pipeline` has executed the `PipelineRun` will be populated with the `Re
 emitted by the `Pipeline`. These will be written to the `PipelineRun's`
 `status.pipelineResults` field.
 
+If a `Pipeline Result` references a failed `TaskRun` then the `Pipeline Result` is
+considered invalid and will not be emitted by the completed `PipelineRun`.
+
```
jerop, line 579: the result could also be missing because the task was skipped as well...not only when it failed...

pritidesai:
@sbwsg I haven't looked at the code yet but we might get into a scenario where task could succeed without producing any result and the same literal might end up in pipelineRun results.

docs/pipelines.md
pritidesai: also successful taskrun without producing task result 🤔

sbwsg:
> @sbwsg I haven't looked at the code yet but we might get into a scenario where task could succeed without producing any result and the same literal might end up in pipelineRun results.

Thanks for the feedback @pritidesai ! I believe this is accounted for - [here's the code that decides on the validity of a PipelineResult's variable reference](https://github.com/tektoncd/pipeline/pull/3472/files#diff-97e88b63f962b155c9f85caa16ac893b604eead06063cb0a4618d88f8f5c1fb0R173-R199) and [a new test case](https://github.com/tektoncd/pipeline/pull/3472/files#diff-44c74890ca26c70492312151417116129a59f80678b4bacb4ffdf7a03a8628feR672-R693) that exercises the behaviour of a PipelineResult referencing a Task Result that is not produced by the TaskRun.  The result reference will be considered invalid and the Pipeline Result won't be emitted.  LMK if you see edge cases I'm not covering here - I want to account for as many of these rules as possible with unit tests so that we can better catch regressions.

docs/pipelines.md
sbwsg: > the result could also be missing because the task was skipped as well...not only when it failed...

Ah, good catch, thanks @jerop, updated docs to expand the list of reasons that a pipeline result would be invalid.

Regarding the successful taskrun w/o producing task result, I think that's accounted for - see [my other comment](https://github.com/tektoncd/pipeline/pull/3472#issuecomment-719506324).

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -176,7 +175,15 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 		// and may not have had all of the assumed default specified.
 		pr.SetDefaults(contexts.WithUpgradeViaDefaulting(ctx))
 
-		c.updatePipelineResults(ctx, pr, getPipelineFunc)
+		if _, pipelineSpec, err := resources.GetPipelineData(ctx, pr, getPipelineFunc); err != nil {
+			logger.Errorf("Failed to get Pipeline Spec to process Pipeline Results: %v", err)
+			pr.Status.MarkFailed(ReasonCouldntGetPipeline,
```
sbwsg, line 180: Question for @pritidesai or @afrittoli - given that we are in a block guarded by `pr.IsDone()`, is it too late to call `pr.Status.MarkFailed()`?  I hadn't noticed this prior to making these changes but the old code did this too (see `updatePipelineResults` in the diff.

pritidesai: Good point, `pr.IsDone` declares pipeline either `success` or `failure`.  `markFailed` can switch the `success` to `failure` with setting its `completionTime`. This can definitely cause issue since the `pipeline` would be marked as finished with the `completionTime` but it's still running.

Its missing call to `finishReconcileUpdateEmitEvents` which will make sure the `pipeline` actually finishes and doesnt continue execution after `MarkFailed`.

```
		pr.Status.MarkFailed(ReasonCouldntGetPipeline, "Failed to get Pipeline Spec to process Pipeline Results %s/%s: %s", pr.Namespace, pr.Name, err)
		return c.finishReconcileUpdateEmitEvents(ctx, pr, before, nil)
```

pritidesai: this brings us back to similar discussion with task results, all the tasks executed successfully but a particular pipeline result which was being referenced doesn't exist so declare pipeline failed 🤔

sbwsg: I'm going to remove the call to `MarkFailed` here I think.

sbwsg: I've removed the call to `MarkFailed` and now instead am using the controller's Recorder (if available) to record a warning event (in addition to logging an error from the controller).

GregDritschler: FYI, as part of the Run integration with pipelines I tried to have the e2e test verify the pipeline results.  I had to take that out because there was a race condition between the test case waking up (seeing the PipelineRun was done) and the controller updating the PipelineRun with the results.  I think the results should be added to the PR at the same time it is marked done.  I'm just bringing this up for your awareness; it's up to you whether you want to tackle it in this PR.

docs/pipelines.md
```diff
@@ -574,6 +574,13 @@ After a `Pipeline` has executed the `PipelineRun` will be populated with the `Re
 emitted by the `Pipeline`. These will be written to the `PipelineRun's`
 `status.pipelineResults` field.
 
+A `Pipeline Result` is not emitted if any of the following are true:
+- a `TaskRun` referenced by the `Pipeline Result` failed
+- a `TaskRun` referenced by the `Pipeline Result` was skipped
+- a `TaskRun` referenced by the `Pipeline Result` didn't emit the referenced `Task Result`
+- the `Pipeline Result` uses a variable that doesn't point to a task in the Pipeline
+- the `Pipeline Result` uses a variable that doesn't point to a result of a task in the Pipeline
```
bobcatfish, line 582: @sbwsg continuation of #3497 - these all seem like errors to me

- the `Pipeline Result` uses a variable that doesn't point to a task in the Pipeline
- the `Pipeline Result` uses a variable that doesn't point to a result of a task in the Pipeline

If I'm understanding correctly, these both mean that the Pipeline was created incorrectly - we could catch these at validation time before we start actually executing the Pipeline

1. a `TaskRun` referenced by the `Pipeline Result` failed <-- this would fail the pipelinerun anyway
2. a `TaskRun` referenced by the `Pipeline Result` was skipped
3. a `TaskRun` referenced by the `Pipeline Result` didn't emit the referenced `Task Result`

These would all be at runtime - I could imagine that someone might be annoyed if their entire Pipeline ran successfully but one of the above caused a pipeline level result not to be set - however I think that could be addressed by a combo of:

1. https://github.com/tektoncd/pipeline/issues/3497 <-- case (3) would fail after the Task ran
2. https://github.com/tektoncd/community/pull/240 <-- we could include this use case if we want to be able to support defaults in the case of (2) (and maybe 3 if we let a Task specify a default result?)

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: that's a good point @GregDritschler 

yup, we dont have to tackle everything in this PR, sorry @sbwsg for polluting this PR.

If I am not mistaken, taskRun has the similar behavior where a taskRun is marked as successful along with its completion time before populating task results:

https://github.com/tektoncd/pipeline/blob/05d67e427c722a2a57e58328d7097e21429b7524/pkg/pod/status.go#L272-L282

https://github.com/tektoncd/pipeline/blob/05d67e427c722a2a57e58328d7097e21429b7524/pkg/pod/status.go#L109-L113

After a taskRun is marked as successful, taskRun status is updated with task results (through [setTaskRunStatusBasedOnStepStatus](https://github.com/tektoncd/pipeline/blob/05d67e427c722a2a57e58328d7097e21429b7524/pkg/pod/status.go#L132)):

https://github.com/tektoncd/pipeline/blob/05d67e427c722a2a57e58328d7097e21429b7524/pkg/pod/status.go#L162-L164

may be worth a discussion in API WG? 😜

sbwsg: Thanks for the feedback, I'm going to back out the change and re-introduce `MarkFailed` since it's the existing behaviour and we can discuss further.

docs/pipelines.md
sbwsg: Agreed on all points.  Validation should catch incorrect variable usage and Tasks that don't emit any described results should fail, causing the Pipeline to also fail.

Regarding the validation, we do have this - it runs during DAG scheduling, so when we decide what to run next we find an invalid result reference and error the PipelineRun with `InvalidTaskResultReference`.

> we could catch these at validation time before we start actually executing the Pipeline

We could push this earlier and validate all Task result references before execution begins but we would need to fetch and inspect every `Task` referenced by a result variable before starting the Pipeline.  Otherwise we don't know what the correct result names are.  So it's feasible and do-able.  I think we need to decide if that's preferable vs whatever time it costs to fetch all the referenced Tasks and inspect their results before running the Pipeline.

Reading back I am realizing this morning that I was wrong in [my comment on #3497](https://github.com/tektoncd/pipeline/issues/3497#issuecomment-721850638). A PipelineRun was never put into a failed state just because a Task Result reference is invalid.  A PipelineRun is marked failed if it's unable to fetch the PipelineSpec needed to process Pipeline Results.  Otherwise the way it's worked before and after this change is that missing or invalid results don't update the final Success/Fail state of the PipelineRun.

pkg/reconciler/pipelinerun/pipelinerun.go
sbwsg: Done - MarkFailed is re-introduced.

docs/pipelines.md
sbwsg: I've created #3499 to explore the possibility and costs of fetching and validating task result references before PipelineRun execution starts.

bobcatfish: I think I'm still a bit confused about both what the current behavior is and what the behavior will be after this PR - would it make sense to maybe make this section of the docs also indicate what behavior we expect in each of these casess?

pritidesai: I agree and open to merge this since there is no functionality change compared to the existing behavior. And continue this discussion with #3499.

sbwsg: Updated the doc with more detail about the accompanying PipelineRun state when these conditions are hit.

pkg/reconciler/pipelinerun/resources/apply_test.go
```diff
@@ -609,3 +612,273 @@ func TestApplyWorkspaces(t *testing.T) {
...
+			Value: "rae",
+		}},
+	}, {
+		description: "multiple-results-multiple-successful-tasks",
+		results: []v1beta1.PipelineResult{{
+			Name:  "pipeline-result-1",
+			Value: "$(tasks.pt1.results.foo)",
+		}, {
+			Name:  "pipeline-result-2",
+			Value: "$(tasks.pt1.results.foo), $(tasks.pt2.results.baz), $(tasks.pt1.results.bar), $(tasks.pt2.results.baz), $(tasks.pt1.results.foo)",
```
pritidesai, line 828: @sbwsg while your are updating the doc, might be worth mentioning this particular scenario.

I think without these changes (this PR):

```
"$(tasks.pt1.results.foo), $(tasks.pt2.results.baz), $(tasks.pt1.results.bar), $(tasks.pt2.results.baz), $(tasks.pt1.results.foo)"
```

if `$(tasks.pt2.results.baz)` was not initialized, this was resulting in:

```
"do, $(tasks.pt2.results.baz), mi, $(tasks.pt2.results.baz), do"
```

(I haven't confirmed yet)

But now with these changes, its not set at all and not part of the results any more.

pritidesai: yup just confirmed, on master, its producing "do, $(tasks.pt2.results.baz), mi, $(tasks.pt2.results.baz), do" 😢

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -124,3 +127,73 @@ func replaceParamValues(params []v1beta1.Param, stringReplacements map[string]st
...
+	}
+
+	return runResults
+}
+
+// taskResultValue returns a pointer to the result value for a given task result variable. A nil
+// pointer is returned if the variable is invalid for any reason.
+func taskResultValue(variable string, taskStatuses map[string]*v1beta1.PipelineRunTaskRunStatus) *string {
+	variableParts := strings.Split(variable, ".")
+	if len(variableParts) != 4 || variableParts[0] != "tasks" || variableParts[2] != "results" {
```
pritidesai, line 177: NIT: `tasks` and `results` could be constants and shared with other result related logic (not part of this PR), can be done seperately

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
```diff
@@ -250,30 +187,6 @@ func getReferencedTaskRun(pipelineState PipelineRunState, reference *v1beta1.Res
...
-}
-
-func findTaskResultForPipelineResult(taskStatus *v1beta1.TaskRunStatus, reference *v1beta1.ResultRef) (*v1beta1.TaskRunResult, error) {
-	results := taskStatus.TaskRunStatusFields.TaskRunResults
-	for _, result := range results {
-		if result.Name == reference.Result {
-			return &result, nil
-		}
-	}
-	return nil, fmt.Errorf("Could not find result with name %s for task run %s", reference.Result, reference.PipelineTask)
```
pritidesai, line 274: This `error` is getting lost and not surfaced to user or logs,  here is the chain of function calls: `findTaskResultForPipelineResult`  <- `resolveResultRefForPipelineResult`  <- `extractResultRefsForPipelineResults` <- `convertPipelineResultToResultRefs`

 `convertPipelineResultToResultRefs` returns nil if `error` is not `nil` without logging it or returning the error along but not sure how the result reference as is makes it to the `pipelineResults`. 

With, #3499, will able to pick up such errors and report them.

pritidesai (approved):
thanks a bunch @sbwsg for fixing this issue and appreciate all the cleanup along. It's much easier to understand now.

pkg/reconciler/pipelinerun/resources/apply_test.go
sbwsg: Thanks @pritidesai 👍 right, that's the bad behaviour we're aiming to fix here.

sbwsg: I've added a Note after the list of reasons that a `Pipeline Result` will not be emitted.  The note talks about a Pipeline Result that references multiple Taskrun Results and makes clear that any invalid reference from the Pipeline Result will mean the entire result is not emitted.

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
sbwsg: Good catch, this is the reason that I ultimately decided the new `taskResultValue()` func shouldn't bother returning an error and just a pointer to the string value if there is one.

docs/pipelines.md
```diff
@@ -586,6 +586,20 @@ references the `outputValue` `Result` emitted by the `calculate-sum` `Task`.
...
+- A `TaskRun` referenced by the `Pipeline Result` was skipped.
+- A `TaskRun` referenced by the `Pipeline Result` didn't emit the referenced `Task Result`. This
+should be considered a bug in the `Task` and [may fail a `TaskRun` in future](https://github.com/tektoncd/pipeline/issues/3497).
+- The `Pipeline Result` uses a variable that doesn't point to a task in the Pipeline. This will
+result in an `InvalidTaskResultReference` validation error during execution of the `Pipeline`.
+- The `Pipeline Result` uses a variable that doesn't point to a result of a task in the Pipeline.
+This will result in an `InvalidTaskResultReference` validation error during execution of the `Pipeline`.
+
+**Note:** Since a `Pipeline Result` can contain references to multiple `Task Results`, if any of those
+`Task Result` references are invalid the entire `Pipeline Result` is not emitted.
```
bobcatfish, line 601: i think as a user i might find it hard to reason about what will happen based on this description, maybe it would be clearer to flip it around a bit, and i think it might be more clear to a user to say something like "task" or "pipeline task" instead of "task run":

PipelineRun execution will fail if a pipeline result requires any of the following:
* A result from a Task that is skipped
* A result from a Task that it declares but when executed it doesn't actually emit it. This should be considered a bug in the `Task` and [may fail a `TaskRun` in future](https://github.com/tektoncd/pipeline/issues/3497).
* A result from a Task that it does not declare
* A result from a Task that does not exit in the Pipeline


I didn't include these ones b/c i'm not sure if they WILL have that same error:
* A `TaskRun` referenced by the `Pipeline Result` failed. The `PipelineRun` will also have failed. <-- since execution will stop in this case, i dont think there would be an error related to the result?

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -124,3 +127,73 @@ func replaceParamValues(params []v1beta1.Param, stringReplacements map[string]st
 	}
 	return params
 }
+
+// ApplyTaskResultsToPipelineResults applies the results of completed TasksRuns to a Pipeline's
+// list of PipelineResults, returning the computed set of PipelineRunResults. References to
+// non-existent TaskResults or failed TaskRuns result in a PipelineResult being considered invalid
+// and omitted from the returned slice. A nil slice is returned if no results are passed in or all
+// results are invalid.
```
bobcatfish, line 135: i apologize for probably having already asked this, but it seems weird to me that we just ignore the invalid values, feels like we should be returning an error?

bobcatfish (commented):
@sbwsg is it possible to update the release note to explain what the new expected behavior is? "Fix bug where PipelineRun emits task results that were never produced" <-- not clear to me what will happen now (from some discussions in the PR it sounds like this is not actually changing the functionality? but i think the update is that now task results that cannot be satisfied with be ""? Or won't be present at all?)

anyway in general it seems like this is better than the current state and it seems like a nice refactoring of the existing code - hopefully we can keep improving the error handling around results!

/approve

pkg/reconciler/pipelinerun/resources/apply.go
sbwsg: I'm not sure it's quite right to say that we ignore the invalid values.  We return a `nil` pointer if the value is invalid - in other words, no valid value was found for the result reference.

I don't totally agree that this is an error state but I don't feel strongly enough to dispute it.  What should we do with that error when it's returned?  In the old code it was silently ignored.

sbwsg:
Updated release note to specify the old and new behaviour.

sbwsg:
> but i think the update is that now task results that cannot be satisfied with be ""? Or won't be present at all?)

Yup, that's exactly right.  Before the result would be emitted by the Pipeline with `"$(task.foo.results.bar)"` left in it unreplaced.

sbwsg:
Just to make things a tad more concrete, here's the yaml from a PipelineRun of a pipeline result that references a skipped task result:

```yaml
  status:
    completionTime: "2020-11-30T15:55:17Z"
    conditions:
    - lastTransitionTime: "2020-11-30T15:55:17Z"
      message: 'Tasks Completed: 3 (Failed: 0, Cancelled 0), Skipped: 3'
      reason: Completed
      status: "True"
      type: Succeeded
    pipelineResults:
    - name: bar
      value: $(tasks.task-should-be-skipped-1.results.bar)
```

Notice that: 1) the PipelineRun is in a Succeeded/True condition and 2) the `pipelineResults` have the variable syntax printed verbatim because the referenced task was skipped.  (2) is the thing we're actually attempting to fix here.  We don't want variable syntax showing up in the pipeline results.  Instead we don't want the pipeline result to be emitted at all.

docs/pipelines.md
sbwsg: OK, I've updated the section to reference "Pipeline Task" throughout instead of "TaskRun".

Inre: your comments on PipelineRun execution behaviour, I think there are at least two discrete questions we're trying to answer in this secion of the doc:

1. What is the final "shape" of a pipeline result if it contains an invalid reference?
2. What happens to pipelinerun execution when pipeline results contain these invalid references?

I think that (1) is answered by the existing PR.  Personally I think that (2) is independent of the results and shouldn't be documented in a section titled `Emitting Results from a Pipeline` since it's orthogonal.  The Pipeline Results don't have any bearing on the execution status of the PipelineRun.  That doesn't mean we shouldn't document it but it does seem to me a little out-of-scope to try and document PipelineRun execution state in a section dedicated to result data.

> PipelineRun execution will fail if a pipeline result requires any of the following:
> 
>* A result from a Task that is skipped

PipelineRun execution doesn't fail when a Task is skipped and a result references it. This is kinda an example of why I'm a tad hesitant to include documentation about pipelinerun execution state in a section dedicated to result data. I think this will come back to bite us if we try and document it here.  One day we'll change the behaviour of PipelineRuns and we'll have to remember that we documented pipelinerun execution in multiple places that aren't `pipelineruns.md` and then comb back through to make sure all those places reflect the changed behaviour.  Personally I'd prefer that we keep the documentation in `pipelines.md` as tightly focused as possible on how Pipeline Results behave.

pkg/reconciler/pipelinerun/resources/apply.go
pritidesai: > feels like we should be returning an error?

Returning an error might break `pipelineRuns`. There is no error reported in logs or to users at all. If we introduce error now, `pipelineRuns` which used to be successful will start failing with this error. And at the same time, the least we could do is log the error if needed.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -176,7 +175,14 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 		// and may not have had all of the assumed default specified.
 		pr.SetDefaults(contexts.WithUpgradeViaDefaulting(ctx))
 
-		c.updatePipelineResults(ctx, pr, getPipelineFunc)
+		if _, pipelineSpec, err := resources.GetPipelineData(ctx, pr, getPipelineFunc); err != nil {
+			msg := fmt.Sprintf("Failed to get Pipeline Spec to process Pipeline Results for PipelineRun %s/%s: %v", pr.Namespace, pr.Name, err)
+			logger.Error(msg)
+			pr.Status.MarkFailed(ReasonCouldntGetPipeline, msg)
```
afrittoli, line 181: I know this has been discussed already, but I think we should not change the final condition of a pipeline once it has been already set since it would break the meaningfulness of the condition. Consumers would not have a way anymore to know if a pipeline status is final or not.
We could to move this result logic so that we evaluate it before the condition is set; if you prefer not to do that in this PR, we could for now log / emit a warning.

sbwsg: I'll add a log line / warning temporarily as this PR is already large enough that I'd prefer not to bloat it further (and given this is the current state of reconciler in `master` I feel it's safer to make the change separately).  Created https://github.com/tektoncd/pipeline/issues/3667 to tackle this immediately afterwards.

sbwsg: Thanks for reviewing @afrittoli .  I'm now logging a warning that the existing PipelineRun Succeeded condition is going to be overwritten.  As part of that message I've recorded which PipelineRun is affected and the existing Succeeded condition before this new failure is introduced.

afrittoli: Thanks for the update @sbwsg. It seems to me that updating a pipeline success status after completion would be introducing an issue, so this patch would fix an issue in favour of introducing another one. 
I understand and agree that you don't want to bloat this PR, but I think a better alternative here would be to send an event warning the user that while the pipeline succeeded the results are not reliable as there was a failure.
Alternatively I would recommend waiting for the next release, so that this PR and the follow-up may be released together.

sbwsg: > It seems to me that updating a pipeline success status after completion would be introducing an issue, so this patch would fix an issue in favour of introducing another one.

@afrittoli I just want to clarify here.  Our existing code in master already overwrites the pipeline status if it fails to get the pipeline data while processing Pipeline Results.  You can see this happening here: https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/pipelinerun/pipelinerun.go#L320-L333

That function, `updatePipelineResults`, is called from https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/pipelinerun/pipelinerun.go#L179 which is inside a block guarded by `if pr.IsDone()`.  This definitely isn't a new issue I'm introducing in order to fix my other issue I promise 😅

afrittoli: Thanks for the clarification, I didn't realise that 🙏

pkg/reconciler/pipelinerun/resources/apply.go
afrittoli: We should eventually move the execution of this logic to before the pipelinerun state is set.
That way could still fail the pipeline if results are broken or invalid in some way.
Maybe something for next release.

afrittoli (commented):
Thank you for this, it is really good step forward.
And thanks for moving a lot of the results code out of pipelinerun.go!!
/lgtm

sbwsg:
Looks like integration failed due to timeouts when communicating with the api server

sbwsg:
The integration job was marked as failed while the build log said it was still running.  Not entirely sure what's up so tried again.

sbwsg:
```
2021/01/12 17:45:35 process.go:153: Running: gcloud compute networks create tpipeline-e2e-net1349049644675502080 --project=tekton-prow-14 --subnet-mode=auto
ERROR: (gcloud.compute.networks.create) Could not fetch resource:
 - Internal error. Please try again or contact Google Support. (Code: '5B8B7971BCA3B.AD38666.49063965')
```

Hmm.  Might just need to sit and wait a bit for things to settle before rerunning.

PR #3491, opened 2020-11-04
Add http and https proxy support for git-init

chmouel:
# Changes

Let's set http and https proxy if we find the HTTP_PROXY and HTTPS_PROXY
environment variables or if the -httpProxy or -httpsProxy are passed.

Tested manually against a  [mitmproxy](http://mitm.it/) install which seems to works well : 

image

Will be the tested functionally in a subsequent update to the git-clone catalog task.

/kind feature

# Submitter Checklist

review them:

- [😿] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [👍🏼] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [👍🏼] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [👍🏼] Release notes block has been filled in or deleted (only if no user facing changes)

# Release Notes

```release-note
Add http and https proxy support for git-init
```

### DIFF
--- a/cmd/git-init/main.go
+++ b/cmd/git-init/main.go
@@ -36,6 +36,9 @@ func init() {
 	flag.StringVar(&fetchSpec.Refspec, "refspec", "", "The Git refspec to fetch the revision from (optional)")
 	flag.StringVar(&fetchSpec.Path, "path", "", "Path of directory under which Git repository will be copied")
 	flag.BoolVar(&fetchSpec.SSLVerify, "sslVerify", true, "Enable/Disable SSL verification in the git config")
+	flag.StringVar(&fetchSpec.HTTPProxy, "httpProxy", "", "Set an http proxy to use")
+	flag.StringVar(&fetchSpec.HTTPSProxy, "httpsProxy", "", "Set an https proxy to use")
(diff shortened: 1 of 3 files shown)


sbwsg:
Changes look good to me - just wondering if we need both mechanisms for setting the proxies?  Would just using the env vars be enough?

chmouel:
> Changes look good to me - just wondering if we need both mechanisms for setting the proxies? Would just using the env vars be enough?

This is a good point and i actually used only the env var, but this is to make it easy to pass it thru from git-parameters via the git-clone task.

There is no env support for podTemplates currently.

chmouel:
sorry about the multiple amend, I am bit rusty on tektoncd/pipelines dev 😅

sbwsg:
I just went and had a look at the `git-clone` task and it already exports these environment variables based on the `httpProxy` and `httpsProxy` params: https://github.com/tektoncd/catalog/blob/dc4ced0942f0f9a3ad21e1dbcb8bc669e3f0d81b/task/git-clone/0.2/git-clone.yaml#L112-L114

I think I am confused - isn't the proxy already supported in this case?

chmouel:
@sbwsg you are right! I have mis read a bug report, i think we may want the CA path think tho. But i can add this directly to the git-init task with a optional namespace thing

thanks for the review!

/close

Issue #3459, opened 2020-10-26
Support Kubernetes style feature-gates for new API fields


skaegi:
Some of the Tekton objects are now "beta" and hopeful "stable" in the not too distant future. When doing work in the API WG we should not be immediately graduating any new fields added to "beta" or whatever degree of stability the CRD currently is. Instead we should provide a mechanism for them to go through some sort of Alpha state until we're sure they work well. 

The alternative is that we'll be forced to retain backwards compatibility with mistakes we might be able to catch with real world use before the feature is promoted to beta or stable.

--
What I'm hoping for is a little pre-discussion before I create a concrete TEP. What I'm looking at is the use of techniques from Kubernetes sig-architure like -- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#adding-unstable-features-to-stable-versions . The idea being that (when possible) you add new feature fields as optional fields that use feature-gates to prevent their use based on the associated feature flag setting.

This issue is specifically targeting the addition of new fields as that currently is the critical problem, however we should also consider applying feature-flags in Tekton to new CRDs.

vdemeester:
Interesting extract of the doc
---
An API change is considered compatible if it:

   * adds new functionality that is not required for correct behavior (e.g.,
does not add a new required field)
   * does not change existing semantics, including:
     * the semantic meaning of default values *and behavior*
     * interpretation of existing API types, fields, and values
     * which fields are required and which are not
     * mutable fields do not become immutable
     * valid values do not become invalid
     * explicitly invalid values do not become valid

Put another way:

1. Any API call (e.g. a structure POSTed to a REST endpoint) that succeeded
before your change must succeed after your change.
2. Any API call that does not use your change must behave the same as it did
before your change.
3. Any API call that uses your change must not cause problems (e.g. crash or
degrade behavior) when issued against an API servers that do not include your
change.
4. It must be possible to round-trip your change (convert to different API
versions and back) with no loss of information.
5. Existing clients need not be aware of your change in order for them to
continue to function as they did previously, even when your change is in use.
6. It must be possible to rollback to a previous version of API server that
does not include your change and have no impact on API objects which do not use
your change.  API objects that use your change will be impacted in case of a
rollback.

---
* Any field with a default value in one API version must have a *non-nil* default
  value in all API versions.  This can be split into 2 cases:
  * Adding a new API version with a default value for an existing non-defaulted
    field: it is required to add a default value semantically equivalent to
[...]

PR #3142, opened 2020-08-26
Introduces Tekton bundles: take 2

pierretasci:
# Changes

Introduces Tekton Bundles with docs, apis, examples, and updates to the reconcilers to allow Task and Pipeline references from Tekton Bundles

*Note, I had to re-submit this PR to get the CLA to pass*

Summary of changes:
- adds docs for how to use Tekton Bundles as refs including a tekton bundle contract
- adds a field to task and pipeline refs to specify a tekton bundle url
- adds validation and tests to the api
- Wires up the OCI remote fetcher to the pipeline ref and task ref handlers
- adds tests into the pipelinerun and taskrun reconcilers to test for remote refs
- requires a change to the way we resolve tasks in the pipelinerun resource handler because we want to decide how to fetch each task ref on a case by case basis instead of using one uniform task fetcher

This is a large change but there are significant side-effects to any one part of this change and it felt half-hearted to only provide the API change to either tasksruns or pipelines and not both at the same time.

# Submitter Checklist

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [x] Release notes block has been filled in or deleted (only if no user facing changes)

# Release Notes

```release-note
Task and Pipeline refs now support remote references via Tekton Bundles
```

PR #3492, opened 2020-11-04
Put Tekton OCI bundles behind a feature flags 🎏

vdemeester:
# Changes

In order to expose Tekton OCI bundle less and mark them as "alpha"
still, let's put their usage under a feature-flag. This will allow us
to experiment, refactor and enhance those without having to support
them nor expose them too much to the end-user.

Enabling the feature gates is a explicit choice for the user and is
documented to make sure users understand this is subject to changes.

This adds a new feature-flags field called "enable-tekton-oci-bundles"
that defaults to false.
If the feature-flag is off, Tekton OCI bundle are not usable. The
admission controller will disallow its usage, and the controller will
not take the bundle field into account.

Note: the e2e tests will be skip on the CI temporarly because we do
not have the *framework* in place to switch feature-flags during
tests. This will be worked out in parallel.

To add a bit more context, I am experiencing pipelinerun completely stuck at resolving tasks when using Tekton bundles, that's why I feel we should put this under a feature flag. This also takes *inspiration* from "Support Kubernetes style feature-gates for new API fields" #3459.

> 

/kind misc

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Introduce a feature-flag `enable-tekton-oci-bundle` to be able to use Tekton Bundle through the pipeline API
```


docs/taskruns.md
```diff
@@ -95,10 +95,13 @@ spec:
           - /kaniko/executor
         args:
           - --destination=gcr.io/my-project/gohelloworld
-```
+``**
```
chmouel, line 98: nit?

docs/pipelineruns.md
```diff
@@ -156,6 +140,27 @@ spec:
...
+   pipelineRef:
+     name: mypipeline
+     bundle: docker.io/myrepo/mycatalog:v1.0
+ ```
+
+The syntax and caveats are similar to using `Tekton Bundles` for  `Task` references
+in [Pipelines](pipelines.md#tekton-bundles) or [TaskRuns](taskruns.md#tekton-bundles).
+
+`Tekton Bundles` may be constructed with any toolsets that produce valid OCI image artifacts
+so long as the artifact adheres to the [contract](tekton-bundle-contracts.md).
```
chmouel, line 161: Maybe udpate the tekton-bundle-contracts file to add it needs to have the feature flag enabled?

chmouel (approved):

bobcatfish:
Quick question: since using the OCI bundles is optional, I want to understand the advantage of feature flagging them? I mean if someone doesn't want to use this feature, they can simply not use it 🤔 

It sounds like you mostly want to use this to indicate that the feature is alpha? I'm not sure I agree with this way of doing it - however also if we don't add the feature flag before our next release we'll be stuck, so I don't want to block it.

Feature flagging won't stop the fields from showing up in the API, it'll just make them not work - just not really sure how that helps users.

vdemeester:
> It sounds like you mostly want to use this to indicate that the feature is alpha? I'm not sure I agree with this way of doing it - however also if we don't add the feature flag before our next release we'll be stuck, so I don't want to block it.

Yes, that's exactly it, indicate the feature is alpha.

> Feature flagging won't stop the fields from showing up in the API, it'll just make them not work - just not really sure how that helps users.

Nothing can really *stop* the field to show up in the API *but* we can mark those fields as "alpha" (document them, etc..) See #3459 for more context about this (and how it's done in kubernetes too)

vdemeester:
From #3459 :

> Some of the Tekton objects are now "beta" and hopeful "stable" in the not too distant future. When doing work in the API WG we should not be immediately graduating any new fields added to "beta" or whatever degree of stability the CRD currently is. Instead we should provide a mechanism for them to go through some sort of Alpha state until we're sure they work well.

(just to put some context here too)

bobcatfish:
@vdemeester do you think it might make sense to have one flag for all of these "alpha" features?

Basically I'm worried that this feature flag will create an even worse user experience:

* folks cant try alpha features without enabling them
* since the features are still IN the API, they will just not work

I think I'm not totally sold on the user benefit of making this part of our API do nothing by default

bobcatfish:
Maybe we can discuss in the working group today! I think there are kinda 2 different discussions:

1. What to do about "alpha" features in beta/v1 APIs in general (which I totally agree is something we do not handle well and I also don't want to rush into a solution)
2. Whether or not this particular feature needs to be behind a feature flag specifically (cuz it's not the first feature we've added directly to our beta API - @vdemeester i think you ran into some performance problems with the feature? i think it's a slightly different story if our reason for adding this flag is "there is a known problem with this feature we are trying to fix" vs "generally b/c it is alpha")

vdemeester:
> @vdemeester do you think it might make sense to have one flag for all of these "alpha" features?

I am not sure nor decided, but I feel, we shouldn't hide them all under one flag, but instead have them all use their own flag (or group them by flag that make sense instead of their "status").

> Basically I'm worried that this feature flag will create an even worse user experience:
>     * folks cant try alpha features without enabling them
>     * since the features are still IN the API, they will just not work
> I think I'm not totally sold on the user benefit of making this part of our API do nothing by default

Well this is debatable, what is a worse use experience between having to explicitly enable a feature to be able to use a field or having a field that could be changed, removed later on. By forcing the user to explicitly enable it, we ensure that the user has the knowledge on what it implies to enable and use this feature (aka alpha, not supported, can be removed later on, …) — it allows us gather feedback on features without having to commit to it too much or too long. It allows the user to use feature in *preview* and have an impact on what the future holds. All of this while working on the stability of the API :upside_down_face:. 

I am open to alternative though. One of it would be to have a `vXalphaX` API living along with the beta/stable API. In our case (and at the time of this comment), this would be having an *active* `v1alpha2` API that *replicates* `v1beta1` with additionnal fields (note that this is harder to implement this cleanly with only CRDs, easier to implement with a full apiserver).

On "*folks cant try alpha features without enabling them*", this is indeed *by design*. To test an alpha feature in a time where we ship a beta or stable api, the user should have to do something explicitly, whether it is by enable a feature-flag, use a `vXalphaX` API, …

Also, note that this is not proposing anything new, it's *just* following what k8s does (as explained in #3459 and [here at length](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#adding-unstable-features-to-stable-versions)) – although in the case of k8s, setting `feature-flags` is harder (because it is directly on the `apiserver` command line arguments).

> Maybe we can discuss in the working group today! I think there are kinda 2 different discussions:
>     1. What to do about "alpha" features in beta/v1 APIs in general (which I totally agree is something we do not handle well and I also don't want to rush into a solution)
>     2. Whether or not this particular feature needs to be behind a feature flag specifically (cuz it's not the first feature we've added directly to our beta API - @vdemeester i think you ran into some performance problems with the feature? i think it's a slightly different story if our reason for adding this flag is "there is a known problem with this feature we are trying to fix" vs "generally b/c it is alpha")

On 1., we need to decide this rather sooner than later :sweat_smile: Otherwise, anything we merge in without a feature-flag (or something to "hide it") is considered as beta (and stable once in `v1` api) with the support that comes with it (deprecation, etc…).
On 2., I am not sure I agree nor disagree. I'm having issues (more than performance, some pipelinerun are just stuck forever and will only end with "timeout" or cancel) while using this feature and in general, it should also be considered as an *alpha* feature.

pritidesai:
> On 2., I am not sure I agree nor disagree. I'm having issues (more than performance, some pipelinerun are just stuck forever and will only end with "timeout" or cancel) while using this feature and in general, it should also be considered as an alpha feature.

🙋‍♀️  I have ran into similar issues (stuck pipelinerun) while reviewing the bundle PR, havent got chance to play with it after.

bobcatfish:
From working group today: thanks for discussing this! I'm happy to add this feature flag for this particular feature as part of the 0.18 release and then continue to discuss our overall strategy separately ( @vdemeester I'm thinking maybe we can try to aim to agree on a policy before 0.19? )

vdemeester:
> From working group today: thanks for discussing this! I'm happy to add this feature flag for this particular feature as part of the 0.18 release and then continue to discuss our overall strategy separately ( @vdemeester I'm thinking maybe we can try to aim to agree on a policy before 0.19? )

Yes definitely. @afrittoli and I are gonna work on a TEP for this :wink:

docs/pipelineruns.md
```diff
@@ -156,6 +140,27 @@ spec:
        ...
 ```
 
+#### Tekton Bundles
```
afrittoli, line 143: I feel like we might want to design a page dedicated to alpha features eventually, or at least one that contains a link to all alpha feature docs, but not necessarily in this patch.

docs/taskruns.md
afrittoli: yeah, this needs to be fixed

test/tektonbundles_test.go
```diff
@@ -515,3 +383,148 @@ func TestTektonBundlesUsingImproperFormat(t *testing.T) {
 		t.Fatalf("Error waiting for PipelineRun to finish with expected error: %s", err)
 	}
 }
+
+func tarImageInOCIFormat(namespace string, img v1.Image) ([]byte, error) {
```
afrittoli, line 387: Are just these just moved around? or modified?

vdemeester: yeah those are just moved at the end of the file (to have the Test at first).. cosmetic change :sweat_smile:

pkg/apis/config/feature_flags.go
```diff
@@ -31,12 +31,14 @@ const (
 	disableCredsInitKey                     = "disable-creds-init"
 	runningInEnvWithInjectedSidecarsKey     = "running-in-environment-with-injected-sidecars"
 	requireGitSSHSecretKnownHostsKey        = "require-git-ssh-secret-known-hosts" // nolint: gosec
+	enableTektonOCIBundles                  = "enable-tekton-oci-bundles"
```
afrittoli, line 34: It would be nice to update the feature_flags_test.go and related test files too

afrittoli (commented):
I tried running the example on a cluster that runs this PR and I get:

```
status:
  completionTime: "2020-11-05T18:06:17Z"
  conditions:
  - lastTransitionTime: "2020-11-05T18:06:17Z"
    message: 'error when listing tasks for taskRun remote-task-reference: tasks.tekton.dev
      "hello-world" not found'
    reason: TaskRunResolutionFailed
    status: "False"
    type: Succeeded
  podName: ""
  startTime: "2020-11-05T18:06:17Z"
```

This is the taskrun:
```
kind: TaskRun
metadata:
  name: remote-task-reference
spec:
  taskRef:
    name: hello-world
    bundle: docker.io/ptasci67/example-oci@sha256:053a6cb9f3711d4527dd0d37ac610e8727ec0288a898d5dfbd79b25bcaa29828
```

I was expecting a validation error, but I think we normally ignore unknown fields, so I guess this should be ok?

afrittoli (commented):
Uhm, actually I think that the changes to `taskrun_validation.go` are missing?

vdemeester:
> Uhm, actually I think that the changes to `taskrun_validation.go` are missing?

OMG, I forgot those, thanks, I'll update tomorrow morning.

docs/pipelineruns.md
bobcatfish, line 143: can we add a link to this new section in the TOC at the top?

bobcatfish (commented):
@pritidesai we'll need to make sure to call this out clearly in the 0.18 release notes: that you have to explicitly enable this feature to use it

pritidesai:
> @pritidesai we'll need to make sure to call this out clearly in the 0.18 release notes: that you have to explicitly enable this feature to use it

I have updated the release notes in the PR #3142  which introduced this feature. Also one of the highlights of 0.18: "bundles behind feature flags"

examples/v1beta1/taskruns/tekton-bundles.yaml
```diff
@@ -1,9 +0,0 @@
-# TODO: Move the example image to a tekton owned repo.
```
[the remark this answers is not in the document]
vdemeester: Yeah I moved it to `no-ci`, forgot to commit the file :sweat: 
Should be fixed now.

pkg/apis/config/feature_flags.go
vdemeester: fixed

docs/pipelineruns.md
vdemeester: done :wink:

afrittoli (commented):
Thanks for putting this together!
The code looks good, it works fine on my kind cluster, so
/approve

pritidesai:
Disabled by default:

```
-> kubectl apply -f pipeline-bundle.yaml
Error from server (BadRequest): error when creating "troubleshoot/bundle.yaml": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: must not set the field(s): spec.pipelinespec.tasks[0].taskref.bundle

-> kubectl apply -f examples/v1beta1/taskruns/no-ci/tekton-bundles.yaml
Error from server (BadRequest): error when creating "examples/v1beta1/taskruns/no-ci/tekton-bundles.yaml": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: must not set the field(s): spec.taskref.bundle
```

pritidesai:
Enabled with `feature-flag` but stuck which will be fixed in 0.19:

```
kubectl get tr remote-task-reference
NAME                    SUCCEEDED   REASON   STARTTIME   COMPLETIONTIME
remote-task-reference
```

docs/pipelineruns.md
```diff
@@ -9,6 +9,7 @@ weight: 4
 - [Overview](#overview)
 - [Configuring a `PipelineRun`](#configuring-a-pipelinerun)
   - [Specifying the target `Pipeline`](#specifying-the-target-pipeline)
+  - [Tekton Bundles](#tekton-bundles)
```
pritidesai, line 12: NIT: "Specifying Tekton Bundles" 😆

pritidesai:
thanks a bunch @vdemeester for putting this together 😻 
/lgtm

PR #3500, opened 2020-11-05
Refactor the way timeouts are handled

mattmoor:
/kind cleanup

Fixes: https://github.com/tektoncd/pipeline/issues/2905

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Fixes an issue where TaskRuns and PipelineRuns may not properly timeout.
```


mattmoor:
Staging this to let Prow test the first chunk as I look at doing the same elsewhere in the codebase.  Will split each round into separate commits, if folks want to take a look and give feedback incrementally.

cc @ImJasonH @vdemeester @afrittoli

ImJasonH:
cc @yaoxiaoqi

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -1748,8 +1748,8 @@ func TestReconcileInvalidTaskRuns(t *testing.T) {
 
 			// Check actions and events
 			actions := clients.Kube.Actions()
-			if len(actions) != 3 || actions[0].Matches("namespaces", "list") {
```
mattmoor, line 1751: Note that all of these checks were previously incorrect because they were missing `!` and had their arguments transposed. 🙃

mattmoor:
Alright, I have two more commits staged that:
1. Make the same change to PipelineRun
2. `rm -rf pkg/timeout` (this resulted in some dep changes)

I kinda want to see the integration test results before pushing that, but heads up that it's coming (again separate commits).

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -497,16 +505,13 @@ func (c *Reconciler) updateLabelsAndAnnotations(ctx context.Context, tr *v1beta1
 func (c *Reconciler) handlePodCreationError(ctx context.Context, tr *v1beta1.TaskRun, err error) error {
 	switch {
 	case isExceededResourceQuotaError(err):
-		backoff, currentlyBackingOff := c.timeoutHandler.GetBackoff(tr.GetNamespacedName(), *tr.Status.StartTime, *tr.Spec.Timeout)
-		if !currentlyBackingOff {
-			go c.timeoutHandler.SetTimer(tr.GetNamespacedName(), time.Until(backoff.NextAttempt))
-		}
-		msg := fmt.Sprintf("TaskRun Pod exceeded available resources, reattempted %d times", backoff.NumAttempts)
```
mattmoor, line 504: This is another change worth calling out.  Previously this bounded the number of attempts, but this state was stored in-memory in the controller process, so it would be lost across failovers and rollouts.

We could perhaps add logic here to `snooze` this, but I went with the simplest change as a starting point for discussion.

mattmoor:
ok it passed, I'm going to push the other two commits now and remove the WIP.

mattmoor:
Apparently I didn't delete enough code 🤣

mattmoor:
Alright, e2e tests have passed twice (once with pipelinerun changes).

Hopefully this time everything is green, but I'd appreciate any feedback, so we can hopefully squash the timeout_test flake (or start chasing what's left).

bobcatfish:
Thanks for working on this @mattmoor ! Looking forward to taking a look - tomorrow our team is having an "offsite" so they'll be a delay for me personally at least

> so we can hopefully squash the timeout_test flake (or start chasing what's left).

Are you seeing these flakes locally, or in recent PRs, or somewhere else?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -217,6 +215,15 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
...
 		return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
 	}
+	defer func() {
+		if pr.Status.StartTime == nil || pr.Spec.Timeout == nil {
+			return
+		}
+		// Compute the time since the task started.
+		elapsed := time.Since(pr.Status.StartTime.Time)
+		// Snooze this resource until the timeout has elapsed.
+		c.snooze(pr, pr.Spec.Timeout.Duration-elapsed)
```
ImJasonH, line 225: getPipelineRunTimeout previously checked that pr.Spec.Timeout wasn't nil, and consulted the configmap for a default timeout, how does this code handle the default timeout?

pkg/reconciler/taskrun/taskrun.go
ImJasonH: Will exceeding resources be retried later until pod creation succeeds? It seems like a regression to just immediately fail in that case.

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -3234,7 +3231,7 @@ func TestFailTaskRun(t *testing.T) {
 				taskLister:        testAssets.Informers.Task.Lister(),
 				clusterTaskLister: testAssets.Informers.ClusterTask.Lister(),
 				resourceLister:    testAssets.Informers.PipelineResource.Lister(),
-				timeoutHandler:    nil, // Not used
+				snooze:            func(acc kmeta.Accessor, amnt time.Duration) {},
```
ImJasonH, line 3234: Can you make this call t.Error so we're sure snooze isn't called in this case?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -2183,16 +2183,13 @@ func TestHandlePodCreationError(t *testing.T) {
 		taskLister:        testAssets.Informers.Task.Lister(),
 		clusterTaskLister: testAssets.Informers.ClusterTask.Lister(),
 		resourceLister:    testAssets.Informers.PipelineResource.Lister(),
-		timeoutHandler:    timeout.NewHandler(ctx.Done(), testAssets.Logger),
+		snooze:            func(acc kmeta.Accessor, amnt time.Duration) {},
```
ImJasonH, line 2186: We should implement snooze here either to prove it is called, or prove it isn't, to aid future improvements in this code.

ImJasonH (commented):
Thanks for this change, I think it's going to be a useful simplification, I just have some concerns about resource-exhaustion handling.

AIUI, for my own understanding, the old timeout handling code ended up calling Enqueue anyway, so we were always reliant on the workqueue to schedule the timeout check, and if the workqueue backed up we'd start to fall behind on timeout checks, so this shouldn't introduce any new behavior except that we rely on EnqueueAfter's internal delay mechanism instead of our own. Is that all correct?

pkg/reconciler/pipelinerun/pipelinerun.go
mattmoor: Brought back `getPipelineRunTimeout`, but changed its return type.

pkg/reconciler/taskrun/taskrun.go
mattmoor: This returns the error so it will be requeued without delay, and the workqueue itself should backoff.

pkg/reconciler/taskrun/taskrun_test.go
mattmoor: good idea, done.

pkg/reconciler/taskrun/taskrun_test.go
mattmoor: also done.

mattmoor (commented):
> AIUI, for my own understanding, the old timeout handling code ended up calling Enqueue anyway, 

Yes

> so we were always reliant on the workqueue to schedule the timeout check, and if the workqueue backed up we'd start to fall behind on timeout checks, 

The problem I saw was that we could actually end up calling the callback and processing the key *before* the resource was actually timed out[1], and if we miss that one shot the entire timeout handling is shot because we had very edge triggered logic to kick off the timeout handling logic.

[1] - I suspect this is due to jitter in the `.status.StartTime` due to the stale informer issues we saw previously here: https://github.com/tektoncd/pipeline/issues/3460 (with pipelinerun, where we clamped the StartTime to the child TaskRun time), so there is *likely* more we can do here (cc @pritidesai called this out), but this is certainly worth doing anyways as it's much more resilient than what's there now.

> so this shouldn't introduce any new behavior except that we rely on EnqueueAfter's internal delay mechanism instead of our own. Is that all correct?

Essentially yes.  In theory we could have used the old method here, but each invocation consumes a NEW go routine (not idempotent), which I suspect would tax the system under load.  AFAIK the workqueue doesn't used go routines for `EnqueueAfter` so I suspect this is more efficient even than what's there, and idempotent to boot so we can blindly `EnqueueAfter` and let it deduplicate internally.

pkg/reconciler/pipelinerun/pipelinerun.go
ImJasonH: I don't see `getPipelineRunTimeout` in the latest commit (unless I'm just missing it).

I see the nil timeout check above, but we're not falling back to the ConfigMap default with `config.FromContextOrDefaults(ctx).Defaults.DefaultTimeoutMinutes`. In theory the defaulting webhook should set this for us, but just in case something slips through without a timeout set we should still enforce the default.

ImJasonH: Nope, I was just missing it 🤦‍♂️ 

Please squash commits before merging 😄

mattmoor: Squashed!

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -797,12 +804,12 @@ func combineTaskRunAndTaskSpecAnnotations(pr *v1beta1.PipelineRun, pipelineTask
 	return annotations
 }
 
-func getPipelineRunTimeout(ctx context.Context, pr *v1beta1.PipelineRun) metav1.Duration {
+func getPipelineRunTimeout(ctx context.Context, pr *v1beta1.PipelineRun) time.Duration {
```
ImJasonH, line 807: We need one of these for TaskRun timeouts too, the same config is used for both (which seems sort of odd to me 🤔 ), so maybe we can just write one `GetTimeout(context.Context, *metav1.Duration) time.Duration` and share it in both places.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -217,6 +215,15 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 		logger.Errorf("Error while syncing the pipelinerun status: %v", err.Error())
 		return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
 	}
+	defer func() {
+		if pr.Status.StartTime == nil || pr.Spec.Timeout == nil {
```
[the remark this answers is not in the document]
mattmoor: Done

pkg/reconciler/pipelinerun/pipelinerun.go
mattmoor: I noticed there is a `GetTimeout()` above the lines in the taskrun reconciler, so I'm going so use that to handle defaulting, but it just uses the static default, not the configurable default.  This should probably be fixed as well?

mattmoor: Alright fixed in a separate commit.

vdemeester (approved):
/meow

dlorenc:
/lgtm

nice!

PR #3223, opened 2020-09-14
Feat: Add pause condition in pipeline run

FogDong:
# Changes

Feat: Add pause condition in pipeline run.

The pause condition will be valid in runAfter pipelines.
When users change the pipeline run's status to paused, after the
completion of the current pipeline, the next pipeline that runs after
the previous one will be paused.

The way we handle `pause` is similar to the way we handle `cancel`.

Reference #2217

/cc @withlin 

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [x] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [x] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

Double check this list of stuff that's easy to miss:

- If you are adding [a new binary/image to the `cmd` dir](../cmd), please update
  [the release Task](../tekton/publish.yaml) to build and release this image.

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

# Release Notes
    
      Feat: Add pause condition in pipeline run.

      The pause condition will be valid in runAfter pipelines.
      When users change the pipeline run's status to paused, after the
      completion of the current pipeline, the next pipeline that runs after
      the previous one will be paused.

PR #3522, opened 2020-11-13
Implement Pending PipelineRun status (TEP-0015)

jbarrick-mesosphere:
# Changes

Implements the Pending PipelineRun status as described in [TEP-0015](https://github.com/tektoncd/community/blob/master/teps/0015-pending-pipeline.md). (based on https://github.com/tektoncd/pipeline/pull/3223)

/kind feature

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Added PipelineRunPending setting to PipelineRun Spec Status to allow creating PipelineRuns in a Pending state.
```

### DIFF
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -21,6 +21,7 @@ weight: 4
   - [Configuring a failure timeout](#configuring-a-failure-timeout)
 - [Monitoring execution status](#monitoring-execution-status)
 - [Cancelling a `PipelineRun`](#cancelling-a-pipelinerun)
+- [Pending `PipelineRuns`](#pending-pipelineruns)
 - [Events](events.md#pipelineruns)
 
 
@@ -527,6 +528,26 @@ spec:
   # […]
   status: "PipelineRunCancelled"
 ```
+
+## Pending `PipelineRuns`
+
+A `PipelineRun` can be created as a "pending" `PipelineRun` meaning that it will not actually be started until the pending status is cleared.
+
+Note that a `PipelineRun` can only be marked "pending" before it has started, this setting is invalid after the `PipelineRun` has been started.
+
+To mark a `PipelineRun` as pending, set `.spec.status` to `PipelineRunPending` when creating it:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: go-example-git
+spec:
+  # […]
+  status: "PipelineRunPending"
+```
+
+To start the PipelineRun, clear the `.spec.status` field. Alternatively, update the value to `PipelineRunCancelled` to cancel it.
 
 ---
 
--- a/internal/builder/v1beta1/pipeline.go
+++ b/internal/builder/v1beta1/pipeline.go
@@ -114,9 +114,14 @@ func PipelineDescription(desc string) PipelineSpecOp {
 	}
 }
 
-// PipelineRunCancelled sets the status to cancel to the TaskRunSpec.
+// PipelineRunCancelled sets the status to cancel the PipelineRunSpec.
 func PipelineRunCancelled(spec *v1beta1.PipelineRunSpec) {
 	spec.Status = v1beta1.PipelineRunSpecStatusCancelled
+}
+
+// PipelineRunPending sets the status to pending to the PipelineRunSpec.
+func PipelineRunPending(spec *v1beta1.PipelineRunSpec) {
+	spec.Status = v1beta1.PipelineRunSpecStatusPending
 }
 
 // PipelineDeclaredResource adds a resource declaration to the Pipeline Spec,
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
@@ -108,6 +108,11 @@ func (pr *PipelineRun) GetTimeout(ctx context.Context) time.Duration {
 	return pr.Spec.Timeout.Duration
 }
 
+// IsPending returns true if the PipelineRun's spec status is set to Pending state
+func (pr *PipelineRun) IsPending() bool {
+	return pr.Spec.Status == PipelineRunSpecStatusPending
+}
+
 // GetNamespacedName returns a k8s namespaced name that identifies this PipelineRun
 func (pr *PipelineRun) GetNamespacedName() types.NamespacedName {
 	return types.NamespacedName{Namespace: pr.Namespace, Name: pr.Name}
@@ -202,6 +207,8 @@ const (
 	// PipelineRunSpecStatusCancelled indicates that the user wants to cancel the task,
 	// if not already cancelled or terminated
 	PipelineRunSpecStatusCancelled = "PipelineRunCancelled"
+
+	PipelineRunSpecStatusPending = "PipelineRunPending"
 )
 
 // PipelineRef can be used to refer to a specific instance of a Pipeline.
@@ -243,6 +250,8 @@ const (
 	// This reason may be found with a corev1.ConditionFalse status, if the cancellation was processed successfully
 	// This reason may be found with a corev1.ConditionUnknown status, if the cancellation is being processed or failed
 	PipelineRunReasonCancelled PipelineRunReason = "Cancelled"
+	// PipelineRunReasonPending is the reason set when the PipelineRun is in the pending state
+	PipelineRunReasonPending PipelineRunReason = "PipelineRunPending"
 	// PipelineRunReasonTimedOut is the reason set when the PipelineRun has timed out
 	PipelineRunReasonTimedOut PipelineRunReason = "PipelineRunTimeout"
 	// PipelineRunReasonStopping indicates that no new Tasks will be scheduled by the controller, and the
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
@@ -92,7 +92,7 @@ func TestPipelineRun_Invalidate(t *testing.T) {
 					Status: "PipelineRunCancell",
 				},
 			},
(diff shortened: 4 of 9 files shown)


pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -1161,6 +1161,50 @@ func TestReconcileOnCancelledPipelineRun(t *testing.T) {
...
+				tb.TaskRunServiceAccountName("test-sa"),
+			),
+		),
+	}
+
+	d := test.Data{
+		PipelineRuns: prs,
+		Pipelines:    ps,
+		Tasks:        ts,
+		TaskRuns:     trs,
```
bobcatfish, line 1194: im a bit confused by why the task run exists? if the pipelinerun is created "pending" then i think no taskruns should exist? i.e. there should never be a time when a pending pipelinerun exists and taskruns exist for the same pipelinerun?

bobcatfish (commented):
Thanks @jbarrick-mesosphere !!

What do you think about maybe including an end to end test as well? https://github.com/tektoncd/pipeline/tree/master/test#end-to-end-tests

that would give you some flexibility to do something like: create a pipelinerun with a quick timeout, but pending, make sure it doesn't time out, then remove pending and verify it succeeds or times out or something (i.e. will make sure more pieces work together)

bobcatfish:
btw this is a relatively small PR so what im suggesting might be overkill but probably the fastest way to get this merged might be to break it up a little, e.g. if you had a PR with just the types and the validation, that could probably get merged a lot faster than this entire PR

pkg/reconciler/pipelinerun/pipelinerun_test.go
jbarrick-mesosphere: Yeah, actually I brought this over from the pause PR that I based this on and just trusted it, but now I'm realizing I don't understand it at all. I'll work on it :)

jbarrick-mesosphere:
> btw this is a relatively small PR so what im suggesting might be overkill but probably the fastest way to get this merged might be to break it up a little, e.g. if you had a PR with just the types and the validation, that could probably get merged a lot faster than this entire PR

Thanks, if this makes it significantly easier for you as a reviewer I'm happy to do this, but I'm in no rush to get the API changes merged in without the implementation.

pkg/reconciler/pipelinerun/pipelinerun_test.go
jbarrick-mesosphere: Okay, it should be more correct now.

pritidesai:
/ok-to-test

Logistics 😄  - please squash the commits

FogDong:
Thanks! I'll close my original PR after this is merged😸

docs/pipelineruns.md
```diff
@@ -528,6 +529,26 @@ spec:
...
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: go-example-git
+spec:
+  # […]
+  status: "PipelineRunPending"
+```
+
+To start the PipelineRun, delete the `.spec.status` field.
```
ImJasonH, line 550: I'd say "unset the `.spec.status` field, or update its value cancel it"

internal/builder/v1beta1/pipeline.go
```diff
@@ -114,11 +114,16 @@ func PipelineDescription(desc string) PipelineSpecOp {
 	}
 }
 
-// PipelineRunCancelled sets the status to cancel to the TaskRunSpec.
+// PipelineRunCancelled sets the status to cancel to the PielineRunSpec.
```
ImJasonH, line 117: typo here: PipelineRunSpec

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
```diff
@@ -92,7 +92,7 @@ func TestPipelineRun_Invalidate(t *testing.T) {
 					Status: "PipelineRunCancell",
 				},
 			},
-			want: apis.ErrInvalidValue("PipelineRunCancell should be PipelineRunCancelled", "spec.status"),
+			want: apis.ErrInvalidValue("PipelineRunCancell should be PipelineRunCancelled or PipelineRunPending", "spec.status"),
```
ImJasonH, line 95: typo: PipelineRunCancel (one l)

ImJasonH:
I notice that the [TEP says it's a non-goal](https://github.com/tektoncd/community/blob/master/teps/0015-pending-pipeline.md#non-goals) to support pending TaskRuns, "as that would requiring pausing the Kubernetes pods themselves" -- I think that's actually a bit misstated -- pending TaskRuns could just not create Pods, the same way that pending PipelineRuns just don't create TaskRuns. I think it's worth reconsidering implementing pending TaskRuns for consistency, or at least coming up with a better rationale for not doing so.

bobcatfish:
> Thanks, if this makes it significantly easier for you as a reviewer I'm happy to do this, but I'm in no rush to get the API changes merged in without the implementation.

I think I'm just over-applying this principle and this PR is pretty easy to review as it is 👍

jbarrick-mesosphere:
> I notice that the [TEP says it's a non-goal](https://github.com/tektoncd/community/blob/master/teps/0015-pending-pipeline.md#non-goals) to support pending TaskRuns, "as that would requiring pausing the Kubernetes pods themselves" -- I think that's actually a bit misstated -- pending TaskRuns could just not create Pods, the same way that pending PipelineRuns just don't create TaskRuns. I think it's worth reconsidering implementing pending TaskRuns for consistency, or at least coming up with a better rationale for not doing so.

Nice catch, yeah this is left over from when the proposal was to actually pause a running pipeline.

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
jbarrick-mesosphere: Yes, this is a previously existing test for an incorrect status, thus the typo.

jbarrick-mesosphere:
> Thanks @jbarrick-mesosphere !!
> 
> What do you think about maybe including an end to end test as well? https://github.com/tektoncd/pipeline/tree/master/test#end-to-end-tests

Added an e2e test. :+1:

ImJasonH:
> Nice catch, yeah this is left over from when the proposal was to actually pause a running pipeline.

I see. Is the intention to support creating pending TaskRuns as well then? Either way, can you update the TEP with this?

jbarrick-mesosphere:
> I see. Is the intention to support creating pending TaskRuns as well then?

I don't currently have a use-case for this.

> Either way, can you update the TEP with this?

I can update the TEP, but that is outside of the scope of this PR.

ImJasonH:
> I don't currently have a use-case for this.

The initial TEP proposal PR says it will cover TaskRuns: https://github.com/tektoncd/community/pull/203

Another recent TEP, https://github.com/tektoncd/community/pull/228, assumes pending TaskRuns to limit concurrency during PipelineRun execution.

jbarrick-mesosphere:
> The initial TEP proposal PR says it will cover TaskRuns: [tektoncd/community#203](https://github.com/tektoncd/community/pull/203)

I wrote the TEP, it doesn't say that.
 
> Another recent TEP, [tektoncd/community#228](https://github.com/tektoncd/community/pull/228), assumes pending TaskRuns to limit concurrency during PipelineRun execution.

That would be a discussion to have on that TEP's PR.

ImJasonH:
> I wrote the TEP, it doesn't say that.

Hey, sorry, I think I spoke imprecisely. I just meant the PR's title as merged was "Add a pending setting to Tekton PipelineRun and TaskRuns", so it seemed like we had intended to support it at some point. It's fine that the TEP's scope eventually narrowed -- better that than the opposite! -- and that the actually-merged TEP didn't cover pending TaskRuns at all.

Your previous comment had been about lacking a use case, and I found the pipeline-concurrency TEP as a use case which was depending on pending TaskRuns, seemingly assuming they'd be added as part of your TEP. You're absolutely right that the pipeline-concurrency TEP is the place to discuss the addition of pending TaskRuns as a requirement for concurrency-limited pipelines.

(To be clear, none of this discussion should block this PR, which otherwise looks great to me.)

If you felt so inclined to extend this work in another PR to include pending TaskRuns, being as you have recent experience implementing such a change, I'm sure the authors of the pipeline-concurrency TEP would appreciate it. If that's outside your needs and you'd prefer the author of that TEP or the community at large to support pending TaskRuns, that's fine too. Thanks for your contributions of all kinds throughout this process. :)

jbarrick-mesosphere:
This should be ready for review.

FogDong:
Do we have any plans to move forward this PR? 🤔

vincent-pli:
Any process of the PR, here we have a case and expect to leverage this feature:

There is a specific `task` named `terminator` and when arrive the `task`, the `pipeline` should be pause, and before `cancel` need to do some cleanup work... 
@ImJasonH @bobcatfish

ImJasonH (commented):
Sorry I didn't realize this PR was still pending (pun intended), it looks great, and thanks for adding tests!

/lgtm

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -202,6 +207,8 @@ const (
 	// PipelineRunSpecStatusCancelled indicates that the user wants to cancel the task,
 	// if not already cancelled or terminated
 	PipelineRunSpecStatusCancelled = "PipelineRunCancelled"
+
+	PipelineRunSpecStatusPending = "PipelineRunPending"
```
sbwsg, line 211: nit: I think it's worth adding a comment here similar to PipelineRunSpecStatusCancelled describing its purpose.  Something like:

```go
// PipelineRunSpecStatusPending indicates that the user wants to postpone starting a PipelineRun
// until some condition is met
```

sbwsg:
Given how long this PR has been open and that I only have a nit about a comment, I'm happy to add that comment in a follow-up PR.

/approve

PR #3524, opened 2020-11-15
Fix recursion issue on Skip

### LINKED ISSUE #3521: Pinning CPU with a pipeline with 50 tasks and use of params/results chain
We are hitting a scalability problem that pins the controller in our clusters when we have a complex pipeline with close to 50 tasks that makes heavy use of params and results. I'll enclose a synthetic test case but this is a real world case that we are seeing in our clusters.

What we see is that the controller is CPU pinned before creating the first Task pod. In my cluster what I'm seeing is a very sudden degradation in start time for the synthetic test case when I run a pipeline that contains:
1) 40 Tasks -- 10 seconds to start
2) 50 Tasks -- 1 minute to start
3) 60 Tasks -- 41 minutes to start
... that feels like we have an algorithm that is out of whack.

The following synthetic tests are standalone and don't use a workspace or anything fancy and might be good in our tests later on...

[pipe40p.yaml.txt](https://github.com/tektoncd/pipeline/files/5539176/pipe40p.yaml.txt)
[pipe50p.yaml.txt](https://github.com/tektoncd/pipeline/files/5539177/pipe50p.yaml.txt)
[pipe60p.yaml.txt](https://github.com/tektoncd/pipeline/files/5539179/pipe60p.yaml.txt)

afrittoli:
# Changes

Fix recursion issue on Skip

The pipelinerun state is made of resolved pipelinerun tasks (rprt),
which are build from the actual status of the associated taskruns.

It is computationaly easy to know if a taskrun started, or completed
successfully or unsuccessfully; however determining whether a taskrun
has been skipped or will be skipped in the pipeline run execution,
requires evaluating the entire pipeline run status and associated dag.

The Skip method used to apply to a single rprt, evaluate the entire
pipeline run status and dag, return whether the specific rprt was
going to be skipped, and throw away the rest.

We used to invoke the Skip method on every rprt in the pipeline state
to calculate candidate tasks for execution. To make things worse,
we also invoked the "Skip" method as part of the "isDone", defined as
a logical OR between "isSuccessful", "isFailed" and "Skip".

With this change we compute the list of tasks to be skipped once,
incrementally, by caching the results of each invocation of 'Skip'.
We store the result in a map to the pipelinerun facts, along with
pipelinerun state and associated dags. We introdce a new method on the
pipelinerun facts called "IsTaskSkipped".

This solution manages to hide some of the details of the skip logic from
the core reconciler logic, bit it still requires the cache to be
manually reset in a couple of places. I believe further refactor could
help, but I wanted to keep this PR as little as possible.
I will further pursue this work by revining #2821

This changes adds a unit test that reproduces the issue in #3521, which
used to fail (with timeout 30s) and now succeedes for pipelines roughly
up to 120 tasks / 120 links. On my laptop, going beyond 120 tasks/links
takes longer than 30s, so I left the unit test at 80 to avoid
introducing a flaky test in CI. There is still work to do to improve
this further, some profiling / tracing work might help.

Breaking large pipelines in logical groups (branches or pipelines in
pipelines) would help reduce the complexity and computational cost for
very large pipelines.

Fixes #3521

Co-authored-by: Scott <sbws@google.com>

/kind bug

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Fix and issue in the pipeline state resolution which lead to very long or failed start times and high controller CPU in case of pipelines with a large number of dependencies between tasks (40+).
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_types.go
@@ -19,6 +19,7 @@ package v1beta1
 import (
 	"github.com/tektoncd/pipeline/pkg/reconciler/pipeline/dag"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 // +genclient
@@ -174,7 +175,16 @@ func (pt PipelineTask) Deps() []string {
 	deps = append(deps, pt.resourceDeps()...)
 	deps = append(deps, pt.orderingDeps()...)
 
-	return deps
+	uniqueDeps := sets.NewString()
+	for _, w := range deps {
+		if uniqueDeps.Has(w) {
+			continue
+		}
+		uniqueDeps.Insert(w)
+
+	}
+
+	return uniqueDeps.List()
 }
 
 func (pt PipelineTask) resourceDeps() []string {
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -133,13 +133,7 @@ func (t *ResolvedPipelineRunTask) checkParentsDone(facts *PipelineRunFacts) bool
 	return true
 }
 
-// Skip returns true if a PipelineTask will not be run because
-// (1) its When Expressions evaluated to false
-// (2) its Condition Checks failed
-// (3) its parent task was skipped
-// (4) Pipeline is in stopping state (one of the PipelineTasks failed)
-// Note that this means Skip returns false if a conditionCheck is in progress
-func (t *ResolvedPipelineRunTask) Skip(facts *PipelineRunFacts) bool {
+func (t *ResolvedPipelineRunTask) skip(facts *PipelineRunFacts) bool {
 	if facts.isFinalTask(t.PipelineTask.Name) || t.IsStarted() {
 		return false
 	}
@@ -149,6 +143,22 @@ func (t *ResolvedPipelineRunTask) Skip(facts *PipelineRunFacts) bool {
 	}
 
 	return false
+}
+
+// Skip returns true if a PipelineTask will not be run because
+// (1) its When Expressions evaluated to false
+// (2) its Condition Checks failed
+// (3) its parent task was skipped
+// (4) Pipeline is in stopping state (one of the PipelineTasks failed)
+// Note that this means Skip returns false if a conditionCheck is in progress
+func (t *ResolvedPipelineRunTask) Skip(facts *PipelineRunFacts) bool {
+	if facts.SkipCache == nil {
+		facts.SkipCache = make(map[string]bool)
+	}
+	if _, cached := facts.SkipCache[t.PipelineTask.Name]; !cached {
+		facts.SkipCache[t.PipelineTask.Name] = t.skip(facts) // t.skip() is same as our existing t.Skip()
+	}
+	return facts.SkipCache[t.PipelineTask.Name]
 }
 
 func (t *ResolvedPipelineRunTask) conditionsSkip() bool {
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
@@ -37,6 +37,16 @@ type PipelineRunFacts struct {
 	State           PipelineRunState
 	TasksGraph      *dag.Graph
 	FinalTasksGraph *dag.Graph
+
+	// SkipCache is a hash of PipelineTask names that stores whether a task will be
+	// executed or not, because it's either not reachable via the DAG due to the pipeline
+	// state, or because it has failed conditions.
+	// We cache this data along the state, because it's expensive to compute, it requires
+	// traversing potentially the whole graph; this way it can built incrementally, when
+	// needed, via the `Skip` method in pipelinerunresolution.go
+	// The skip data is sensitive to changes in the state. The ResetSkippedCache method
+	// can be used to clean the cache and force re-computation when needed.
+	SkipCache map[string]bool
 }
 
 // pipelineRunStatusCount holds the count of successful, failed, cancelled, skipped, and incomplete tasks
@@ -51,6 +61,11 @@ type pipelineRunStatusCount struct {
 	Cancelled int
 	// number of tasks which are still pending, have not executed
 	Incomplete int
+}
+
+// ResetSkippedCache resets the skipped cache in the facts map
+func (facts *PipelineRunFacts) ResetSkippedCache() {
+	facts.SkipCache = make(map[string]bool)
 }
 
 // ToMap returns a map that maps pipeline task name to the resolved pipeline run task
--- a/pkg/reconciler/pipeline/dag/dag.go
+++ b/pkg/reconciler/pipeline/dag/dag.go
@@ -137,12 +137,16 @@ func linkPipelineTasks(prev *Node, next *Node) error {
 }
 
 func visit(currentName string, nodes []*Node, path []string, visited map[string]bool) error {
+	var sb strings.Builder
 	for _, n := range nodes {
 		path = append(path, n.Task.HashKey())
 		if _, ok := visited[n.Task.HashKey()]; ok {
 			return errors.New(getVisitedPath(path))
 		}
-		visited[currentName+"."+n.Task.HashKey()] = true
+		sb.WriteString(currentName)
+		sb.WriteByte('.')
+		sb.WriteString(n.Task.HashKey())
+		visited[sb.String()] = true
 		if err := visit(n.Task.HashKey(), n.Prev, path, visited); err != nil {
 			return err
 		}
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -524,6 +524,9 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
 		return err
 	}
 
+	// Reset the skipped status to trigger recalculation
+	pipelineRunFacts.ResetSkippedCache()
+
 	after := pipelineRunFacts.GetPipelineConditionStatus(pr, logger)
 	switch after.Status {
 	case corev1.ConditionTrue:
@@ -565,6 +568,9 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
 	}
 
 	resources.ApplyTaskResults(nextRprts, resolvedResultRefs)
+	// After we apply Task Results, we may be able to evaluate more
+	// when expressions, so reset the skipped cache
+	pipelineRunFacts.ResetSkippedCache()
 
 	// GetFinalTasks only returns tasks when a DAG is complete
 	nextRprts = append(nextRprts, pipelineRunFacts.GetFinalTasks()...)
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate_test.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package resources
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
@@ -408,6 +409,7 @@ func TestGetNextTaskWithRetries(t *testing.T) {
 }
 
 func TestPipelineRunState_SuccessfulOrSkippedDAGTasks(t *testing.T) {
+	largePipelineState := buildPipelineStateWithLargeDepencyGraph(t)
 	tcs := []struct {
 		name          string
 		state         PipelineRunState
@@ -454,6 +456,10 @@ func TestPipelineRunState_SuccessfulOrSkippedDAGTasks(t *testing.T) {
 			"not skipped since it failed",
 		state:         conditionCheckFailedWithOthersFailedState,
 		expectedNames: []string{pts[5].Name},
+	}, {
+		name:          "large deps, not started",
+		state:         largePipelineState,
+		expectedNames: []string{},
 	}}
 	for _, tc := range tcs {
 		t.Run(tc.name, func(t *testing.T) {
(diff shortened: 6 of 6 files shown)


pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -160,30 +131,17 @@ func (t *ResolvedPipelineRunTask) conditionsSkip() bool {
...
 
-func (t *ResolvedPipelineRunTask) whenExpressionsSkip(facts *PipelineRunFacts) bool {
-	if t.checkParentsDone(facts) {
-		if len(t.PipelineTask.WhenExpressions) > 0 {
-			if !t.PipelineTask.WhenExpressions.HaveVariables() {
-				if !t.PipelineTask.WhenExpressions.AllowsExecution() {
-					return true
-				}
+func (t *ResolvedPipelineRunTask) whenExpressionsSkip() bool {
+	if len(t.PipelineTask.WhenExpressions) > 0 {
```
ImJasonH, line 135: Not blocking: Was there an explicit style decision involved in making this nested `if`s instead of `&&`?

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -185,6 +199,76 @@ func (facts *PipelineRunFacts) IsStopping() bool {
 	return false
 }
 
+// IsTaskSkipped returns true if the PipelineRun won't be scheduling the name task
+func (facts *PipelineRunFacts) IsTaskSkipped(pipelineTaskName string) bool {
+	if facts.WontRun == nil {
+		// We do lazy initalisation of this map
+		facts.WontRun = facts.SkipMap()
+	}
+	wontRun, _ := facts.WontRun[pipelineTaskName]
```
ImJasonH, line 208: Nit: this can just `return` the bool directly

ImJasonH (commented):
/lgtm

Just a couple style questions, in case they help debugging this code in the future. Feel free to ignore them.

Thanks for tracking this down!

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: Not really, here I only removed the outer `if`, so it wasn't really a decision :)
I tried to minimise changes to the code in case we want to use this for minor releases.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
afrittoli: good catch, thanks

afrittoli:
/cc @vdemeester @ImJasonH @skaegi @jerop @pritidesai 

I added this to the v0.18 milestone, I think it's probably worth a minor release.
I'm not sure we need this on v0.17.0.

sbwsg:
Performance much improved, thanks @afrittoli ! 👍   I do have one issue which is that this is quite a large rewrite of logic for a point release.  It seems the entire skip logic is refactored?

As I was reading the PR I was trying to reason about the specific area that is causing the slow down. Am I right that it is the recursive calls `Skip()` -> `parentTasksSkip()` -> `Skip()`?

Is it possible that we could get a similar speedup with a **much** smaller code change just by memoizing the return value of `Skip`?  Something like:

```go
func (t *ResolvedPipelineRunTask) Skip(facts *PipelineRunFacts) bool {
       if _, cached := facts.skipCache[t.PipelineTask.Name]; !cached {
               facts.skipCache[t.PipelineTask.Name] = t.skip(facts) // t.skip() is same as our existing t.Skip()
       }
       return facts.skipCache[t.PipelineTask.Name]
}
```

You'd still need to reset the skip cache as you do now but I wonder if this would provide much of the benefit with quite a lot less code changes?

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -34,9 +34,23 @@ type PipelineRunState []*ResolvedPipelineRunTask
...
 	FinalTasksGraph *dag.Graph
+
+	// WontRun is a hash of all PipelineTask names that stores whether a task will be
+	// executed or not, because it's either not reachable via the DAG due to the pipeline
+	// state, or because it has failed conditions.
+	// We store this list in along the state, because it requires traversing the whole
+	// graph; this way so we can build it once, when the reconcile starts, and update
+	// at the end of reconcile, instead of having to traverse the whole graph every time
+	// we want to know if a specific task will be skipped.
+	WontRun map[string]bool
```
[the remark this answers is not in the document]
afrittoli: good idea, I will fix that, thanks

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -844,19 +860,15 @@ func TestIsSkipped(t *testing.T) {
...
-			for taskName, isSkipped := range tc.expected {
-				rprt := stateMap[taskName]
-				if rprt == nil {
-					t.Fatalf("Could not get task %s from the state: %v", taskName, tc.state)
-				}
-				if d := cmp.Diff(isSkipped, rprt.Skip(&facts)); d != "" {
-					t.Errorf("Didn't get expected isSkipped %s", diff.PrintWantGot(d))
+			actual := facts.SkipMap()
+			for taskname, expected := range tc.expected {
+				if d := cmp.Diff(actual[taskname], expected); d != "" {
```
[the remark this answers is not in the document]
afrittoli: ack, good catch

afrittoli:
> Performance much improved, thanks @afrittoli ! 👍 I do have one issue which is that this is quite a large rewrite of logic for a point release. It seems the entire skip logic is refactored?
> 
> As I was reading the PR I was trying to reason about the specific area that is causing the slow down. Am I right that it is the recursive calls `Skip()` -> `parentTasksSkip()` -> `Skip()`?
> 
> Is it possible that we could get a similar speedup with a **much** smaller code change just by memoizing the return value of `Skip`? Something like:
> 
> ```go
> func (t *ResolvedPipelineRunTask) Skip(facts *PipelineRunFacts) bool {
>        if _, cached := facts.skipCache[t.PipelineTask.Name]; !cached {
>                facts.skipCache[t.PipelineTask.Name] = t.skip(facts) // t.skip() is same as our existing t.Skip()
>        }
>        return facts.skipCache[t.PipelineTask.Name]
> }
> ```
> 
> You'd still need to reset the skip cache as you do now but I wonder if this would provide much of the benefit with quite a lot less code changes?

That might work indeed, I'll need to pass it though some testing to see if it fixes the issue.

pritidesai:
> Is it possible that we could get a similar speedup with a much smaller code change just by memoizing the return value of Skip? 

Yup, I agree with @sbwsg

`skip` has been a long pending beast which needs fix. Thanks @afrittoli to create such state which helps improving performance. I will appreciate minimal code change in the patch since `skip` involves many different combinations of in a pipeline `when` expressions, `conditions` and `finally`. Would like to see `e2e` test involving all these artifacts with these changes.

afrittoli:
> > Is it possible that we could get a similar speedup with a much smaller code change just by memoizing the return value of Skip?
> 
> Yup, I agree with @sbwsg
> 
> `skip` has been a long pending beast which needs fix. Thanks @afrittoli to create such state which helps improving performance. I will appreciate minimal code change in the patch since `skip` involves many different combinations in a pipeline `when` expressions, `conditions` and `finally`. Would like to see `e2e` test involving all these artifacts with these changes.

I tried my original fix and @sbwsg proposed fix both via unit test as well as on a cluster with the pipeline60 and they both worked fine.

With my original patch I managed to go up to 120 tasks in the unit tests, with the new version I've not managed to go beyond the 80 tasks, but that fluctuation is likely to be related to how busy my laptop CPU is.

I agree that a smaller patch is better for a point release. @sbwsg is it ok if I update this PR as you suggested?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -151,6 +145,22 @@ func (t *ResolvedPipelineRunTask) Skip(facts *PipelineRunFacts) bool {
...
+// (2) its Condition Checks failed
+// (3) its parent task was skipped
+// (4) Pipeline is in stopping state (one of the PipelineTasks failed)
+// Note that this means Skip returns false if a conditionCheck is in progress
+func (t *ResolvedPipelineRunTask) Skip(facts *PipelineRunFacts) bool {
+	if facts.SkipCache == nil {
+		facts.SkipCache = make(map[string]bool)
+	}
+	if _, cached := facts.SkipCache[t.PipelineTask.Name]; !cached {
+		facts.SkipCache[t.PipelineTask.Name] = t.skip(facts) // t.skip() is same as our existing t.Skip()
```
sbwsg, line 159: Suggest removing the comment now that the original `t.Skip` is renamed `t.skip`

sbwsg (approved):
Awesome, thanks for tackling this Andrea!

sbwsg:
Oops, one more thing: `IsTaskSkipped` is mentioned in the commit message but no longer appears in the diff.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -37,6 +37,16 @@ type PipelineRunFacts struct {
...
+
+	// SkipCache is a hash of PipelineTask names that stores whether a task will be
+	// executed or not, because it's either not reachable via the DAG due to the pipeline
+	// state, or because it has failed conditions.
+	// We cache this data along the state, because it's expensive to compute, it requires
+	// traversing potentially the whole graph; this way it can built incrementally, when
+	// needed, via the `Skip` method in pipelinerunresolution.go
+	// The skip data is sensitive to changes in the state. The ResetSkippedCache method
+	// can be used to clean the cache and force re-computation when needed.
+	SkipCache map[string]bool
```
pritidesai, line 49: we might be able to instantiate this cache based on `pr.Status.SkippedTasks` which holds the list of `skippedTasks`.

I will not block this PR for `skippedTasks`, we can revisit this in next iteration. @jerop?

afrittoli:
It looks like CI is slower than my laptop:
```
=== RUN   TestPipelineRunState_SuccessfulOrSkippedDAGTasks/large_deps,_not_started
panic: test timed out after 10m0s

goroutine 274 [running]:
testing.(*M).startAlarm.func1()
	/usr/local/go/src/testing/testing.go:1628 +0x11f
created by time.goFunc
	/usr/local/go/src/time/sleep.go:167 +0x52
```

I'll reduce the size of the pipeline.

sbwsg:
(Copying my comment thread from slack for posterity):

I've checked out the branch and I am noticing some slightly odd behaviour.  Every test run varies dramatically in amount of time large_deps,_not_started takes.  Some runs are 7 seconds, some are blowing through the 30s timeout cap.

I'm poking around with pprof to see if I can figure out where all the time is being spent when the test run gets slow.  When i hit the 30s timeout  I am seeing the following top from pprof:

```
     3.83s 11.92% 11.92%     23.86s 74.28%  github.com/tektoncd/pipeline/pkg/reconciler/pipeline/dag.visit
     2.97s  9.25% 21.17%      7.73s 24.07%  runtime.concatstrings
     2.80s  8.72% 29.89%      2.80s  8.72%  runtime.duffcopy
     2.62s  8.16% 38.04%      2.62s  8.16%  aeshashbody
     2.02s  6.29% 44.33%      3.75s 11.67%  runtime.mapassign_faststr
     1.90s  5.92% 50.25%      3.93s 12.24%  runtime.mapaccess2_faststr
     1.73s  5.39% 55.64%      1.73s  5.39%  runtime.kevent
     1.68s  5.23% 60.87%      3.92s 12.20%  runtime.mallocgc
     1.61s  5.01% 65.88%      1.61s  5.01%  runtime.memmove
     1.60s  4.98% 70.86%      1.60s  4.98%  runtime.pthread_kill
```

So all the time is being spent in `dag.visit`.  Digging a bit further I notice that it times out during the `DagFromState` function in the test, which happens before the `successfulOrSkippedDAGTasks` function is exercised.  So somewhere in building the graph for the test we are slowing waaaaaay down sporadically.

pritidesai:
@sbwsg I ran the commands you pasted 🆙  and its the same `dag.visit` showing on the top (I have checked out this branch and working with it):

```
      flat  flat%   sum%        cum   cum%
     3.89s 10.92% 10.92%      3.89s 10.92%  runtime.duffcopy
     3.67s 10.31% 21.23%     25.30s 71.05%  github.com/tektoncd/pipeline/pkg/reconciler/pipeline/dag.visit
     2.80s  7.86% 29.09%      8.11s 22.77%  runtime.concatstrings
     2.29s  6.43% 35.52%      5.12s 14.38%  runtime.mallocgc
     2.23s  6.26% 41.79%      2.23s  6.26%  aeshashbody
     2.13s  5.98% 47.77%      2.13s  5.98%  runtime.pthread_kill
     1.79s  5.03% 52.79%      3.87s 10.87%  runtime.mapassign_faststr
     1.77s  4.97% 57.76%      1.77s  4.97%  runtime.pthread_cond_wait
     1.69s  4.75% 62.51%      3.10s  8.71%  runtime.mapaccess2_faststr
     1.68s  4.72% 67.23%      1.68s  4.72%  runtime.memmove
```

pritidesai:
@afrittoli I ran the unit test you have added here on `master` and profiling it, results in:

```
     5.90s 11.31% 11.31%      5.90s 11.31%  runtime.pthread_cond_wait
     4.99s  9.57% 20.88%      4.99s  9.57%  runtime.kevent
     4.82s  9.24% 30.12%      4.82s  9.24%  runtime.pthread_kill
     4.62s  8.86% 38.98%     21.95s 42.09%  runtime.mapassign_faststr
     3.90s  7.48% 46.46%      8.19s 15.70%  runtime.evacuate_faststr
     2.88s  5.52% 51.98%      2.88s  5.52%  aeshashbody
        2s  3.84% 55.82%      3.33s  6.39%  runtime.scanobject
     1.90s  3.64% 59.46%      1.90s  3.64%  runtime.memclrNoHeapPointers
     1.61s  3.09% 62.55%      1.85s  3.55%  runtime.heapBitsSetType
     1.24s  2.38% 64.93%      1.24s  2.38%  runtime.memmove
```

I am wondering why these changes have so much time spent on `dag.visit`

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: fixed

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -174,7 +175,16 @@ func (pt PipelineTask) Deps() []string {
...
+	uniqueDeps := sets.NewString()
+	for _, w := range deps {
+		if uniqueDeps.Has(w) {
+			continue
+		}
+		uniqueDeps.Insert(w)
+
+	}
+
+	return uniqueDeps.List()
```
pritidesai, line 187: this significantly reduces the call to `dag.visit` and makes it possible for the unit test to complete within five seconds. The mysterious dot notation in `dag.visit` is still unsolved:

https://github.com/tektoncd/pipeline/blob/747f4baa014e44a9143eafb7d87a8f9a6c5a3e60/pkg/reconciler/pipeline/dag/dag.go#L145

Its creating a link between two nodes which indirectly connected (`C.A` in `A -> B -> C`) but where/how is this data being utilized is a mystery.

pritidesai:
/lgtm

@skaegi let us know if you still run into scaling issues, the examples you have specified must be fixed with these changes.

Thanks @afrittoli  and @sbwsg for fixes and help troubleshoot this 🙏

pritidesai:
Cheery picked with PR #3534

PR #3539, opened 2020-11-18
fix cycle detection logic in dag

pritidesai:
# Changes

`visited` map holds name of the nodes being visited as keys and `true` as value. This map is being updated with `currentName.HashKey` on every iteration which results in keys such as `a.d`, `b.d` where a, b, and d are nodes of a `dag`. But such keys are not utilized anywhere in the same `visit` function. `Visit` function checks existence of the node just
by the name without any string concatenation.

This extra addition in the map is causing severe delay for a graph with more than 60 nodes with distinct dependencies.

`visit` function is called on line 131 with all the parents of the existing `pipelineTask` to check if any of the parents exist in the `visited` map. `visit` function is then recursively called to check each parent's parent to see if any of them exist in the `visited` map, so on and so forth to detect a cycle.  

/kind bug

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipeline/dag/dag.go
+++ b/pkg/reconciler/pipeline/dag/dag.go
@@ -126,9 +126,8 @@ func linkPipelineTasks(prev *Node, next *Node) error {
 		return fmt.Errorf("cycle detected; task %q depends on itself", next.Task.HashKey())
 	}
 	// Check if we are adding cycles.
-	visited := map[string]bool{prev.Task.HashKey(): true, next.Task.HashKey(): true}
 	path := []string{next.Task.HashKey(), prev.Task.HashKey()}
-	if err := visit(next.Task.HashKey(), prev.Prev, path, visited); err != nil {
+	if err := lookForNode(prev.Prev, path, next.Task.HashKey()); err != nil {
 		return fmt.Errorf("cycle detected: %w", err)
 	}
 	next.Prev = append(next.Prev, prev)
@@ -136,18 +135,13 @@ func linkPipelineTasks(prev *Node, next *Node) error {
 	return nil
 }
 
-func visit(currentName string, nodes []*Node, path []string, visited map[string]bool) error {
-	var sb strings.Builder
+func lookForNode(nodes []*Node, path []string, next string) error {
 	for _, n := range nodes {
 		path = append(path, n.Task.HashKey())
-		if _, ok := visited[n.Task.HashKey()]; ok {
+		if n.Task.HashKey() == next {
 			return errors.New(getVisitedPath(path))
 		}
-		sb.WriteString(currentName)
-		sb.WriteByte('.')
-		sb.WriteString(n.Task.HashKey())
-		visited[sb.String()] = true
-		if err := visit(n.Task.HashKey(), n.Prev, path, visited); err != nil {
+		if err := lookForNode(n.Prev, path, next); err != nil {
 			return err
 		}
 	}
--- a/pkg/reconciler/pipeline/dag/dag_test.go
+++ b/pkg/reconciler/pipeline/dag/dag_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package dag_test
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -479,7 +480,7 @@ func TestBuild_ConditionsParamsFromTaskResults(t *testing.T) {
 	assertSameDAG(t, expectedDAG, g)
 }
 
-func TestBuild_Invalid(t *testing.T) {
+func TestBuild_InvalidDAG(t *testing.T) {
 	a := v1beta1.PipelineTask{Name: "a"}
 	xDependsOnA := v1beta1.PipelineTask{
 		Name: "x",
@@ -539,37 +540,114 @@ func TestBuild_Invalid(t *testing.T) {
 			Resources:    []v1beta1.PipelineTaskInputResource{{From: []string{"none"}}},
 		}},
 	}
+	aRunsAfterE := v1beta1.PipelineTask{Name: "a", RunAfter: []string{"e"}}
+	bDependsOnA := v1beta1.PipelineTask{
+		Name: "b",
+		Resources: &v1beta1.PipelineTaskResources{
+			Inputs: []v1beta1.PipelineTaskInputResource{{From: []string{"a"}}},
+		},
+	}
+	cRunsAfterA := v1beta1.PipelineTask{
+		Name:     "c",
+		RunAfter: []string{"a"},
+	}
+	dDependsOnBAndC := v1beta1.PipelineTask{
+		Name: "d",
+		Resources: &v1beta1.PipelineTaskResources{
+			Inputs: []v1beta1.PipelineTaskInputResource{{From: []string{"b", "c"}}},
+		},
+	}
+	eRunsAfterD := v1beta1.PipelineTask{
+		Name:     "e",
+		RunAfter: []string{"d"},
+	}
+	fRunsAfterD := v1beta1.PipelineTask{
+		Name:     "f",
+		RunAfter: []string{"d"},
+	}
+	gDependsOnF := v1beta1.PipelineTask{
+		Name: "g",
+		Resources: &v1beta1.PipelineTaskResources{
+			Inputs: []v1beta1.PipelineTaskInputResource{{From: []string{"f"}}},
+		},
+	}
 
 	tcs := []struct {
 		name string
 		spec v1beta1.PipelineSpec
+		err  string
 	}{{
+		// a
+		// |
+		// a ("a" depends on resource from "a")
 		name: "self-link-from",
 		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{selfLinkFrom}},
+		err:  "cycle detected",
 	}, {
+		// a
+		// |
+		// a ("a" runAfter "a")
 		name: "self-link-after",
 		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{selfLinkAfter}},
+		err:  "cycle detected",
 	}, {
+		// a (also "a" depends on resource from "z")
+		// |
+		// x ("x" depends on resource from "a")
+		// |
+		// z ("z" depends on resource from "x")
 		name: "cycle-from",
 		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{xDependsOnA, zDependsOnX, aDependsOnZ}},
+		err:  "cycle detected",
 	}, {
+		// a (also "a" runAfter "z")
+		// |
+		// x ("x" runAfter "a")
+		// |
+		// z ("z" runAfter "x")
 		name: "cycle-runAfter",
 		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{xAfterA, zAfterX, aAfterZ}},
+		err:  "cycle detected",
 	}, {
+		// a (also "a" depends on resource from "z")
+		// |
+		// x ("x" depends on resource from "a")
+		// |
+		// z ("z" runAfter "x")
 		name: "cycle-both",
 		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{xDependsOnA, zAfterX, aDependsOnZ}},
+		err:  "cycle detected",
+	}, {
+		// This test make sure we detect a cyclic branch in a DAG with multiple branches.
+		// The following DAG is having a cyclic branch with an additional dependency (a runAfter e)
+		//   a
+		//  / \
+		// b   c
+		//  \ /
+		//   d
+		//  / \
+		// e   f
+		//     |
+		//     g
+		name: "multiple-branches-with-one-cyclic-branch",
+		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{aRunsAfterE, bDependsOnA, cRunsAfterA, dDependsOnBAndC, eRunsAfterD, fRunsAfterD, gDependsOnF}},
+		err:  "cycle detected",
 	}, {
 		name: "duplicate-tasks",
 		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{a, a}},
+		err:  "duplicate pipeline task",
 	}, {
 		name: "invalid-task-name-from",
 		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{invalidTaskFrom}},
+		err:  "wasn't present in Pipeline",
 	}, {
 		name: "invalid-task-name-after",
 		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{invalidTaskAfter}},
+		err:  "wasn't present in Pipeline",
 	}, {
 		name: "invalid-task-name-from-conditional",
 		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{invalidConditionalTask}},
+		err:  "wasn't present in Pipeline",
 	},
 	}
 	for _, tc := range tcs {
@@ -578,7 +656,8 @@ func TestBuild_Invalid(t *testing.T) {
 				ObjectMeta: metav1.ObjectMeta{Name: tc.name},
 				Spec:       tc.spec,
 			}
-			if _, err := dag.Build(v1beta1.PipelineTaskList(p.Spec.Tasks)); err == nil {
+			_, err := dag.Build(v1beta1.PipelineTaskList(p.Spec.Tasks))
+			if err == nil || !strings.Contains(err.Error(), tc.err) {
 				t.Errorf("expected to see an error for invalid DAG in pipeline %v but had none", tc.spec)
 			}
 		})


pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -669,3 +669,59 @@ func TestBuild_ConditionsParamsFromTaskResults(t *testing.T) {
...
+//   a
+//  / \
+// b   c
+//  \ /
+//   d
+//  / \
+// e   f
+//     |
+//     g
+func TestBuild_InvalidDAGWithCycle(t *testing.T) {
```
sbwsg, line 685: I'm not 100% sure, but should we duplicate this test in `pkg/reconciler/pipeline/dag/dagv1beta1_test.go` as well?

pkg/reconciler/pipeline/dag/dag.go
```diff
@@ -128,26 +128,21 @@ func linkPipelineTasks(prev *Node, next *Node) error {
...
+func visit(nodes []*Node, path []string, visited map[string]bool) error {
 	for _, n := range nodes {
 		path = append(path, n.Task.HashKey())
 		if _, ok := visited[n.Task.HashKey()]; ok {
 			return errors.New(getVisitedPath(path))
 		}
-		sb.WriteString(currentName)
-		sb.WriteByte('.')
-		sb.WriteString(n.Task.HashKey())
-		visited[sb.String()] = true
```
sbwsg, line 149: I'm a bit confused :S We no longer add anything to `visited` except for the two linked nodes (line 129).  Is that right?

sbwsg (commented):
Some comments to address but I think the change is good!

pkg/reconciler/pipeline/dag/dag.go
sbwsg: If that's right, do we need the `visisted` map at all?

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -669,3 +669,59 @@ func TestBuild_ConditionsParamsFromTaskResults(t *testing.T) {
...
+		Resources: &v1beta1.PipelineTaskResources{
+			Inputs: []v1beta1.PipelineTaskInputResource{{From: []string{"a"}}},
+		},
+	}
+	cRunsAfterA := v1beta1.PipelineTask{
+		Name:     "c",
+		RunAfter: []string{"a"},
+	}
+	dDependsOnBAndC := v1beta1.PipelineTask{
+		Name: "f",
```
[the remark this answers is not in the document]
pritidesai: thanks @sbwsg, great catch, fixed it.

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -669,3 +669,59 @@ func TestBuild_ConditionsParamsFromTaskResults(t *testing.T) {
...
+	}
+
+	p := &v1beta1.Pipeline{
+		ObjectMeta: metav1.ObjectMeta{Name: "invalid-dag-with-cycle"},
+		Spec: v1beta1.PipelineSpec{
+			Tasks: v1beta1.PipelineTaskList{a, bDependsOnA, cRunsAfterA, dDependsOnBAndC, eRunsAfterD, fRunsAfterD, gDependsOnF},
+		},
+	}
+	if _, err := dag.Build(v1beta1.PipelineTaskList(p.Spec.Tasks)); err == nil {
+		t.Errorf("expected to see an error for invalid DAG in pipeline %v but had none", p.Spec)
```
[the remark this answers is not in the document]
pritidesai: yup agree, I dropped printing pipelineSpec. We don't need it, `Build` works with just a list of tasks, updated the test.

pkg/reconciler/pipeline/dag/dag.go
pritidesai: > I'm a bit confused :S We no longer add anything to visited except for the two linked nodes (line 129). Is that right?

we no longer add anything to visited. `visited` map does not need any update within the `visit` function. `linkPipelineTasks` function adds two pipelineTasks in the visited map before creating a link between the two.

The overall logic is:

`Build` function adds all the pipelineTasks from the `tasks` section in the `Graph`.  And creates a map of a `pipelineTask -> a list of parents (resource and order dependencies)`.

For the unit test I have added, the dependency [map](https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/pipeline/dag/dag.go#L79) looks like, e.g. node `g` has a parent node `f`, etc:

```
(map[string][]string) (len=7) {
 (string) (len=1) "g": ([]string) (len=1 cap=1) {
  (string) (len=1) "f"
 },
 (string) (len=1) "a": ([]string) (len=1 cap=1) {
  (string) (len=1) "e"
 },
 (string) (len=1) "b": ([]string) (len=1 cap=1) {
  (string) (len=1) "a"
 },
 (string) (len=1) "c": ([]string) (len=1 cap=1) {
  (string) (len=1) "a"
 },
 (string) (len=1) "d": ([]string) (len=2 cap=2) {
  (string) (len=1) "b",
  (string) (len=1) "c"
 },
 (string) (len=1) "e": ([]string) (len=1 cap=1) {
  (string) (len=1) "d"
 },
 (string) (len=1) "f": ([]string) (len=1 cap=1) {
  (string) (len=1) "d"
 }
}
``` 

Now, its iterating over this map to add a link between a `pipelineTask` and its parent node (one parent at a time). Before adding a link, it checks for any cycles.

Here, pt is the pipelineTask `b` and `previousTask` is actually parent `a`.

https://github.com/tektoncd/pipeline/blob/473e3f3cc74b215f8c22cb007fcd7413f93f3917/pkg/reconciler/pipeline/dag/dag.go#L84

Now, `addLink` with `prev` (parent) set to `a` and `next` set to `b`  calls `linkPipelineTasks`.  `linkPipelineTasks` created `visited` map with `a` and `b` and call `visit` with `b` and parents of `a` (`prev.Prev`), in the unit test, parent of `a` is `e`. `visit` checks if `e` exists in the `visited`, which is not `true` so now recursively call with `e` parents. `e` does not have any parents at this point of time and creates path `e -> a -> b` which is valid, no cycles. Update the `graph` setting `b.Prev=a` and `a.Next=b`.

Cycle is detected by iteratively checking parent and its parent until there are no parents in the `visited` map.

> If that's right, do we need the visisted map at all?

`visited` map is needed to check if any of the parent match with what we are visiting right now. 

https://github.com/tektoncd/pipeline/blob/473e3f3cc74b215f8c22cb007fcd7413f93f3917/pkg/reconciler/pipeline/dag/dag.go#L143

pkg/reconciler/pipeline/dag/dag_test.go
pritidesai: sure I added in `v1beta1` as well

afrittoli:
Thanks for this! I didn't manage to review yet, I'll do tomorrow 🙏 

I wonder if on long term we could consider using something like https://godoc.org/github.com/twmb/algoimpl/go/graph for the DAG logic? It already detects cycles, it allows identifying  strongly connected components which could be handy for the pipeline workflow logic. 
That at the cost of an extra dependency of course :)

bobcatfish:
Hey @pritidesai !! Nice catch, it looks like this one has been around for a while :D It took me a while to wrap my head around the changes here and I had to do some investigation to understand them, so I want to try to explain these changes and can you confirm if this is how you are seeing it as well?

To try to understand what's happening I had to look back at https://github.com/tektoncd/pipeline/commit/a1f27336b15b94caaf97e4843dd2d9f99b901ef3 when the visited list was updated to do that concatenation.

Here is my understanding of the problem that was being fixed at that time:

```
             42
            /  \
         100    200
           \   / 
            101 
              | 
             102
```

1. We'd try to add a link between 101 + 102 via addPrevPipelineTask
2. addPrevPipelineTask would try to "visit" all of 101's previous Tasks
3. visit would recursively "visit" all of the previous Tasks of those Tasks

While (2) and (3) were happening, a "visited" list was being built up
When 101 was "visited" the first time, it would visit it's previous tasks, etc. adding 100 and then 42 to the "visited" list
Then 200 would be "visited" - and 42 is a previous task of 200, so it would be "visited" as well
At that point, 42 would already be in the "visited" list and would incorrectly detect a cycle

So https://github.com/tektoncd/pipeline/commit/a1f27336b15b94caaf97e4843dd2d9f99b901ef3  attempted to fix this by changing the keys in visited.

BUT!

As @pritidesai pointed out, we may ADD the concatenated keys but WE DONT USE THEM WHEN CHECKING KEYS

I think what this means is that https://github.com/tektoncd/pipeline/commit/a1f27336b15b94caaf97e4843dd2d9f99b901ef3 effectively made us _ignore_ all these entries we were adding to the list.

And it works! Which is why in this PR @pritidesai can _remove_ these additions into the visited list b/c we didn't need them in the first place.

And the REASON we don't need them is because the actual question we are trying to ask when we call `visit` from `linkPipelineTasks` is a bit obscured: we are trying to ask "is this node its own parent" (which means there is a cycle). And we ask this for every Task, so we don't need to be tracking all the Tasks all the time.

So in fact I think we could simplify `linkPipelineTasks` a bit to make this more clear, instead of:

```
       // i have a strong hunch we don't even need both of these nodes in "visited" for cycle detection to work, i think we just need the somewhat confusingly named "next" (maybe a name like "curr" would be more clear?)
        visited := map[string]bool{prev.Task.HashKey(): true, next.Task.HashKey(): true}
	if err := visit(prev.Prev, path, visited); err != nil {
```

Something like:

```
        // Look through all of prev's ancestors and see if the "next" node is present
	if err := lookForNode(prev.Prev, path, next); err != nil {
```

> TestBuild_InvalidDAGWithCycle

I don't think we actually need this new test (maybe not the other one either) - The test TestBuild_InvalidDAGWithCycle passes even without the changes (the error message is different but a cycle is still detected) and it looks like TestBuild_Invalid is testing this case (there are 3 different versions of cycles (a has to run after z but z runs after x which runs after a i think?) (it could definitely use a comment depicting the graph - and maybe a better name)

Because I think what this PR is doing is not actually changing any functionality, but is optimizing it (we're no longer storing the concatenated keys which were being ignored anyway)

> I wonder if on long term we could consider using something like https://godoc.org/github.com/twmb/algoimpl/go/graph for the DAG logic? It already detects cycles via the "strongly connected components". That at the cost of an extra dependency of course :)

I have mixed feelings! On the one hand I feel like this DAG logic is at the heart of our Pipeline execution so it feels weird to entrust it to an external lib, on the other hand we had this inefficiency for nearly 2 years and didn't notice (on the other other hand, this didn't actually cause any bugs). Might be interesting to try swapping it out and see what it's like?

pritidesai:
> can you confirm if this is how you are seeing it as well?

yes @bobcatfish you absolutely nailed it. 


> And the REASON we don't need them is because the actual question we are trying to ask when we call visit from linkPipelineTasks is a bit obscured: we are trying to ask "is this node its own parent" (which means there is a cycle). And we ask this for every Task, so we don't need to be tracking all the Tasks all the time.

I agree, it is bit obscured, before adding any new link, its actually checking if this node (`next`) is its own parent by traversing through the `prev.Prev` until `prev.Prev` is `empty` or finds match (`next` == `prev.Prev`).

>  `if err := lookForNode(prev.Prev, path, next); err != nil {`

`visited` can be replaced with `next` like you pointed out. I will update the PR.

> The test TestBuild_InvalidDAGWithCycle passes even without the changes 

yup, since there is no functionality change 😄 This test validates a little bit more complex graph compared to existing invalid test:

```
    a
   / \
 b   c
   \ /
   d
  / \
e    f (a runAfter e)
     |
     g
``` 

TestBuild_Invalid is validating one single cycle in three different forms (combination of runAfter and from) but from the `Build` unit test perspective its the same graph:

```
		 a
		 |
		 x
		 |
		 z (also a depends on/runAfter z)
```

> Might be interesting to try swapping it out and see what it's like?

I generally avoid third party dependencies for the same reasons you have mentioned, specially for the heart of the pipeline execution.

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -669,3 +669,59 @@ func TestBuild_ConditionsParamsFromTaskResults(t *testing.T) {
...
+		},
+	}
+
+	p := &v1beta1.Pipeline{
+		ObjectMeta: metav1.ObjectMeta{Name: "invalid-dag-with-cycle"},
+		Spec: v1beta1.PipelineSpec{
+			Tasks: v1beta1.PipelineTaskList{a, bDependsOnA, cRunsAfterA, dDependsOnBAndC, eRunsAfterD, fRunsAfterD, gDependsOnF},
+		},
+	}
+	if _, err := dag.Build(v1beta1.PipelineTaskList(p.Spec.Tasks)); err == nil {
```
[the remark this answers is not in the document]
pritidesai: Hey @sbwsg, there are multiple paths possible leading to a cycle. It all depends on the order in which dependency map is processed. 

This test can fail with one of these paths:

```
	if err == nil || !(strings.Contains(err.Error(), "e -> a -> b -> d -> e") ||
		strings.Contains(err.Error(), "a -> c -> b -> d -> e -> a") ||
		strings.Contains(err.Error(), "d -> e -> a -> b -> d")) {
		t.Errorf("expected cycle error but got %v", err)
	}
```

I can replace the existing error checking with the above check if we decide to keep this test.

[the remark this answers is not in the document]
pritidesai: Running the unit test multiple times with `count=1` leads to many different combination of paths including:

```
		"a -> c -> b -> d -> e -> a",
		"a -> b -> c -> d -> e -> a",
		"a -> b -> d -> e -> a",
		"a -> c -> d -> e -> a",
		"b -> d -> e -> a -> b",
		"c -> d -> e -> a -> c",
```

I walked through the graph one more time and there are much more combinations possible. I think its fine to leave it as just `cycle detected` instead of looking for a specific path in the unit test.

vdemeester:
> I have mixed feelings! On the one hand I feel like this DAG logic is at the heart of our Pipeline execution so it feels weird to entrust it to an external lib, on the other hand we had this inefficiency for nearly 2 years and didn't notice (on the other other hand, this didn't actually cause any bugs). Might be interesting to try swapping it out and see what it's like?

That's an interesting though, because I would have gone the other way. The DAG is not something specific to Tekton, the way we build it and us it is, so I would think using an external library that does the "dag" well would be good enough for us and would allow us to add value to where we need to. But I don't have strong opinion on this :upside_down_face:

pkg/reconciler/pipeline/dag/dag_test.go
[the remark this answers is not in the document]
sbwsg: Great point, the order of traversal isn't stable across runs because the nodes are stored in a hash map.  `cycle detected` makes sense 👍

sbwsg (approved):

bobcatfish:
> This test validates a little bit more complex graph compared to existing invalid test

in order to make sure it's clear which tests are covering what to future maintainers, what do you think about changing the names of these tests to make it more clear what the line is between them? and/or maybe even combining them into one table driven test?

bobcatfish:
Speaking of tests, this is a bit of scope creep, but why does dagv1beta1_test.go  duplicate all the tests is in dag_test.go? Is there some way we can avoid having to maintain this duplication?

pritidesai:
> what do you think about changing the names of these tests to make it more clear what the line is between them?

I have renamed the test.

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -669,3 +691,56 @@ func TestBuild_ConditionsParamsFromTaskResults(t *testing.T) {
 	}
 	assertSameDAG(t, expectedDAG, g)
 }
+
+// This test make sure we detect cycle when it exists.
+// This means we "visit" "a" twice, from two different paths.
```
bobcatfish, line 696: I think this sentence " This means we "visit" "a" twice, from two different paths." is making this more confusing than it needs to be - the problem is that a needs to run after e but e needs to run after a (details of the implementation around "visiting" i dont think really help that much)

bobcatfish (commented):
Thanks @pritidesai 

can we also rename TestBuild_Invalid so it's clear what relationship this test has to TestBuild_InvalidDAG_OneCyclicBranch?

e.g. something like
TestBuild_InvalidDAG_OneCyclicBranch and TestBuild_InvalidDAG_OneCycle
or 
TestBuild_InvalidDAG_CycleInBranch and TestBuild_InvalidDAG_MultipleCycleTypes

Bonus: putting the tests near each other in the file makes it clear they are related

pkg/reconciler/pipeline/dag/dagv1beta1_test.go
```diff
@@ -391,18 +392,39 @@ func TestBuild_Invalid_v1beta1(t *testing.T) {
...
+		// |
+		// z ("z" runAfter "x")
 		name: "cycle-runAfter",
 		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{xAfterA, zAfterX, aAfterZ}},
 	}, {
+		// a (also "a" depends on resource from "z")
+		// |
+		// x ("x" depends on resource from "a")
+		// |
+		// z ("z" runAfter "x")
```
bobcatfish, line 427: Can we include these cycles in these comments too (and the one on the new test)?

e.g. something like this (or better! haha it's hard to show the direction):

```
 → a
|  ↓   
|  x
 \ ↓ 
   z
```

pritidesai:
> Speaking of tests, this is a bit of scope creep, but why does dagv1beta1_test.go duplicate all the tests is in dag_test.go? Is there some way we can avoid having to maintain this duplication?

Warranted some investigation 😉 

`dag_test.go` was referring to `v1alpha1` API whereas `dagv1beta1_test.go` is referring to `v1beta1` and was dependent on the issue #2410 before deprecating/deleting tests under `dag_test.go`.

And we ended up moving `dag_test.go` to `v1beta1` about two months back
https://github.com/tektoncd/pipeline/commit/e356ca54305695dd22aa4f2042ceb8598fa8da71#diff-eb55d5a93fabcf61a0a897c11c28ffe92fc96ec02f8c7a6bdbcaa97fd3af8cc1

I guess its safe to delete one `_test.go` at this point @vdemeester? I can create a separate PR to cleanup, if needed.

pkg/reconciler/pipeline/dag/dag_test.go
pritidesai: I have rephrased the comments.

pkg/reconciler/pipeline/dag/dagv1beta1_test.go
pritidesai: Yeah I tried adding them but looked very weird. Wish I could apply my artistic skills here 🤣

pritidesai:
> can we also rename TestBuild_Invalid so it's clear what relationship this test has to TestBuild_InvalidDAG_OneCyclicBranch?

`TestBuild_Invalid` tests DAG for other failure as well e.g. duplicate task names, invalid task name in dependency, etc. Have renamed the test to `TestBuild_Invalid`. 

While creating a separate cleanup PR, I will rearrange test functions further e.g. local functions towards the end of the file, rearrange `TestBuild` and `TestBuild_Invalid`, etc.

pkg/reconciler/pipeline/dag/dagv1beta1_test.go
pritidesai:
```
		// -> a
		// |  |
		// |  v
		// -- a ("a" depends on resource from "a")
```

🤔

pritidesai:
@bobcatfish if build passes, its ready for review 🙏

vdemeester:
> And we ended up moving `dag_test.go` to `v1beta1` about two months back
> [e356ca5#diff-eb55d5a93fabcf61a0a897c11c28ffe92fc96ec02f8c7a6bdbcaa97fd3af8cc1](https://github.com/tektoncd/pipeline/commit/e356ca54305695dd22aa4f2042ceb8598fa8da71#diff-eb55d5a93fabcf61a0a897c11c28ffe92fc96ec02f8c7a6bdbcaa97fd3af8cc1)
> 
> I guess its safe to delete one `_test.go` at this point @vdemeester? I can create a separate PR to cleanup, if needed.

Yeah I think so :wink:

pkg/reconciler/pipeline/dag/dagv1beta1_test.go
bobcatfish: i think anything is fine, even if it doesnt look great, just something to make it a complete representation of the graph in question

bobcatfish (commented):
i think we are almost there! thanks for your patience!

> TestBuild_Invalid tests DAG for other failures as well e.g. duplicate task names, invalid task name in dependency, etc. Have renamed the test to TestBuild_Invalid.

I'd like to give one last plug for having one test for all of these cases, or for removing the invalid DAG cases from the existing function - i.e. one place to find all the invalid graphs, it doesn't seem to me like we're gaining much from the 2 separate tests at the moment, since the cases they are covering overlap. so id much prefer to see one test that covers all the cases, or a clear line between which cases each of the 2 tests cover

pritidesai:
> i think we are almost there! thanks for your patience!

Definitely, have merged them all in one.

Also, if cleanup PR #3556 is merged before this, will update this PR to drop changes to `dagv1beta1_test.go`.

afrittoli:
> > I have mixed feelings! On the one hand I feel like this DAG logic is at the heart of our Pipeline execution so it feels weird to entrust it to an external lib, on the other hand we had this inefficiency for nearly 2 years and didn't notice (on the other other hand, this didn't actually cause any bugs). Might be interesting to try swapping it out and see what it's like?
> 
> That's an interesting though, because I would have gone the other way. The DAG is not something specific to Tekton, the way we build it and us it is, so I would think using an external library that does the "dag" well would be good enough for us and would allow us to add value to where we need to. But I don't have strong opinion on this :upside_down_face: 

Hey, yeah, that was my thinking exactly, but I don't have a strong feeling about this either

bobcatfish:
Thanks @pritidesai ! looks good to me :D

even tho you'll need it again for the conflict, a symbolic lgtm:

/lgtm

pritidesai:
`unable to access 'https://github.com/GoogleContainerTools/skaffold/': Could not resolve host:` 🤔

PR #3550, opened 2020-11-20
tests: environement setup and other refactorings

vdemeester:
# Changes

This update the e2e tests suite to use `TestMain` in order to set up a
environment object for the whole test. This object is meant to store
anything that could be useful for the test to gather information from
it and act based on it (skipping, getting the right image, …)

In addition, this extract the clients into their own package, as well
as the log helpers.

It also refactor a bit how multi-arch is handled, by porting most of
the feature into the environment.

There is a more improvements to do based on this, but this is a first step.

/kind misc
/area testing

/cc @barthy1 @ImJasonH @mattmoor 

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
NONE
```


barthy1:
@vdemeester just ping me, when it's ready to run on s390x env :)

vdemeester:
Note on possible follow-up:
- Run more tests in the same namespace. We create a namespace for almost each and every test (even with table-driven ones). It costs a little bit of testing time.
- Add some "requirement" or "skip" *framework* (not necessarly an external one)
- Expose this to other projects (triggers, cli, …)

vdemeester:
```
no child with platform linux/linux
```

Oups :sweat_smile:

vdemeester:
It is ready for review @barthy1 
/cc @afrittoli

test/helm_task_test.go
```diff
@@ -42,13 +45,17 @@ var (
 // TestHelmDeployPipelineRun is an integration test that will verify a pipeline build an image
 // and then using helm to deploy it
 func TestHelmDeployPipelineRun(t *testing.T) {
+	if testEnv.Platform == "linux/s390x" {
+		t.Skipf("skip for s390x architecture")
+	}
```
barthy1, line 50: May be it's better to have 1 entry point for different e2e tests to skip, which can be extended for other architectures? For instance in `setupWithCleanup`?

test/internal/environment/environment.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, err
+	}
+	v, err := c.KubeClient.Kube.Discovery().ServerVersion()
+	if err != nil {
+		return nil, err
+	}
+	return &Execution{
+		Clients:     c,
+		Info:        v,
+		Platform:    v.Platform,
```
barthy1, line 61: I really like this option to get `platform` from actual target k8s cluster, so no need any more to use `TEST_RUNTIME_ARCH`, in case when tests are initiated on machine of one arch, but will be executed on k8s cluster with another arch 👍 
Can you also remove mention about this env variable from [doc](https://github.com/tektoncd/pipeline/tree/master/test#running) ?

test/examples_test.go
```diff
@@ -53,47 +49,23 @@ func getCreatedTektonCRD(input []byte, kind string) (string, error) {
...
-		output = existingImage.ReplaceAll(output, archSpecificImage)
-	}
-	return output, nil
-}
-
 // koCreate wraps the ko binary and invokes `ko create` for input within
 // namespace
 func koCreate(input []byte, namespace string) ([]byte, error) {
-	cmd := exec.Command("ko", "create", "--platform", "linux/"+getTestArch(), "-n", namespace, "-f", "-")
+	cmd := exec.Command("ko", "create", "--platform", testEnv.Platform, "-n", namespace, "-f", "-")
```
barthy1, line 68: After test case "initiate tests on amd64, execute on s390x", learnt that `--kubeconfig` should be added here. However it probably can be separate PR, as this functionality is not available in the initial code.

barthy1:
the e2e and examples tests passed for s390x. Provided just a few suggestions from my side.

test/helm_task_test.go
vdemeester: So I want to make the `skip` very explicit *per* tests. Aka, if a test needs to be skip for whatever reason, it needs to be known when reading the test. In `docker/docker` we used the concept of `requirements` (`requirements.IsNetworkAvailable`, …)

test/internal/environment/environment.go
vdemeester: we can yes :upside_down_face:

test/examples_test.go
vdemeester: @barthy1 I am not sure I understand :thinking:

barthy1: If `go test` is running with `--kubeconfig`, `ko create` won't have knowledge about the special kubeconfig and will be targeted to default cluster

vdemeester: @barthy1 ahhh very good point ! Yeah this can be done in a follow-up or now, I'll see if I've got some time to fix

test/helm_task_test.go
barthy1: ok, I have to say I like to keep related things in one place, just not to spam the code with `if` statements for different cases (and many archs in the future), however I'd like to believe that all tests will work for all archs 🤞 and there won't be many exclusions required..

vdemeester: example in `docker/docker`: https://github.com/moby/moby/blob/master/integration/container/create_test.go#L74

sbwsg (approved):

vdemeester:
/close
I'll rework this at some point.

PR #3560, opened 2020-11-24
Fix PipelineRun serviceAccountNames for finally tasks

ljupchokotev:
# Changes

Mapping a service account to a finally task in a PipelineRun's
serviceAccountNames field or taskRunSpecs would result in an error saying that the task doesn't
exist in the Pipeline.

This commit fixes the ValidateServiceaccountMapping and ValidateTasskRunSpecs functions by iterating over
the finally tasks in addition to the normal tasks of a Pipeline.

Also, added test cases for the bug.

/kind bug

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Fixes a bug where PipelineRun's serviceAccountNames and taskPodSpecs couldn't be applied on finally tasks and resulted in an error.
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -267,6 +267,10 @@ func ValidateTaskRunSpecs(p *v1beta1.PipelineSpec, pr *v1beta1.PipelineRun) erro
 		pipelineTasks[task.Name] = task.Name
 	}
 
+	for _, task := range p.Finally {
+		pipelineTasks[task.Name] = task.Name
+	}
+
 	for _, taskrunSpec := range pr.Spec.TaskRunSpecs {
 		if _, ok := pipelineTasks[taskrunSpec.PipelineTaskName]; !ok {
 			return fmt.Errorf("PipelineRun's taskrunSpecs defined wrong taskName: %q, does not exist in Pipeline", taskrunSpec.PipelineTaskName)
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -282,6 +282,10 @@ func ValidateServiceaccountMapping(p *v1beta1.PipelineSpec, pr *v1beta1.Pipeline
 		pipelineTasks[task.Name] = task.Name
 	}
 
+	for _, task := range p.Finally {
+		pipelineTasks[task.Name] = task.Name
+	}
+
 	for _, name := range pr.Spec.ServiceAccountNames {
 		if _, ok := pipelineTasks[name.TaskName]; !ok {
 			return fmt.Errorf("PipelineRun's ServiceAccountNames defined wrong taskName: %q, does not exist in Pipeline", name.TaskName)
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
@@ -1796,6 +1796,76 @@ func TestValidateWorkspaceBindingsWithInvalidWorkspaces(t *testing.T) {
 	}
 }
 
+func TestValidateTaskRunSpecs(t *testing.T) {
+	for _, tc := range []struct {
+		name    string
+		p       *v1beta1.Pipeline
+		run     *v1beta1.PipelineRun
+		wantErr bool
+	}{{
+		name: "valid task mapping",
+		p: tb.Pipeline("pipelines", tb.PipelineSpec(
+			tb.PipelineTask("mytask1", "task",
+				tb.PipelineTaskInputResource("input1", "git-resource")),
+		)),
+		run: tb.PipelineRun("pipelinerun", tb.PipelineRunSpec("pipeline",
+			tb.PipelineTaskRunSpecs(
+				[]v1beta1.PipelineTaskRunSpec{{
+					PipelineTaskName:       "mytask1",
+					TaskServiceAccountName: "default",
+				}},
+			),
+		)),
+		wantErr: false,
+	}, {
+		name: "valid finally task mapping",
+		p: tb.Pipeline("pipelines", tb.PipelineSpec(
+			tb.PipelineTask("mytask1", "task",
+				tb.PipelineTaskInputResource("input1", "git-resource")),
+			tb.FinalPipelineTask("myfinaltask1", "finaltask"),
+		)),
+		run: tb.PipelineRun("pipelinerun", tb.PipelineRunSpec("pipeline",
+			tb.PipelineTaskRunSpecs(
+				[]v1beta1.PipelineTaskRunSpec{{
+					PipelineTaskName:       "myfinaltask1",
+					TaskServiceAccountName: "default",
+				}},
+			),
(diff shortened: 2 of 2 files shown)


pritidesai:
thanks @ljupchokotev for the fix. 

/approve

Please note that the `pipelineRun.Spec.ServiceAccountNames` are [deprecated](https://github.com/tektoncd/pipeline/blob/master/docs/deprecations.md) and will be removed as early as May, 2021.

@vdemeester these changes are fixing finally task to use `pipelineRun.Spec.ServiceAccountNames`. It was not restricted at the time finally tasks were implemented but now since its [deprecated](https://github.com/tektoncd/pipeline/commit/28d13472156fcf2ceef4b0ae409fe368b73f96dc#diff-188e8bbe2f5c7cb0b9de03be02a957537e46bb71add977e4992533375ef08998), what is the procedure? Merge this now and drop support while deprecating in May?

pritidesai:
or avoid fixing it? putting it on hold until we clarify

/hold

ljupchokotev:
@pritidesai just looked at the code and it seems that even by using `taskRunSpecs` you would have the same issue. See [here](https://github.com/tektoncd/pipeline/blob/373081364f4bf2f1776e3ea7ed1edda2f71a53f0/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L266), it doesn't iterate over the finally tasks which means you won't be able to set a service account for a finally task and fail with the same issue as what this PR tries to fix.

Should I fix that part and include it in this PR?

vdemeester:
> @pritidesai just looked at the code and it seems that even by using `taskRunSpecs` you would have the same issue. See [here](https://github.com/tektoncd/pipeline/blob/373081364f4bf2f1776e3ea7ed1edda2f71a53f0/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L266), it doesn't iterate over the finally tasks which means you won't be able to set a service account for a finally task and fail with the same issue as what this PR tries to fix.
> 
> Should I fix that part and include it in this PR?

Definitely yes :wink:

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -282,6 +282,10 @@ func ValidateServiceaccountMapping(p *v1beta1.PipelineSpec, pr *v1beta1.Pipeline
 		pipelineTasks[task.Name] = task.Name
 	}
 
+	for _, task := range p.Finally {
+		pipelineTasks[task.Name] = task.Name
+	}
+
 	for _, name := range pr.Spec.ServiceAccountNames {
```
vdemeester, line 289: Interesting, this is a miss when we deprecated `ServiceAccountNames` :thinking: it shouldn't be looking here (or at least it should be sharing the same code as what the reconciler does)

vdemeester (changes requested):

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
ljupchokotev: > > @pritidesai just looked at the code and it seems that even by using `taskRunSpecs` you would have the same issue. See [here](https://github.com/tektoncd/pipeline/blob/373081364f4bf2f1776e3ea7ed1edda2f71a53f0/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L266), it doesn't iterate over the finally tasks which means you won't be able to set a service account for a finally task and fail with the same issue as what this PR tries to fix.
> > Should I fix that part and include it in this PR?
> 
> Definitely yes wink

Done, also edited the PR description to match the current state.

ljupchokotev:
> > @pritidesai just looked at the code and it seems that even by using `taskRunSpecs` you would have the same issue. See [here](https://github.com/tektoncd/pipeline/blob/373081364f4bf2f1776e3ea7ed1edda2f71a53f0/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L266), it doesn't iterate over the finally tasks which means you won't be able to set a service account for a finally task and fail with the same issue as what this PR tries to fix.
> > Should I fix that part and include it in this PR?
> 
> Definitely yes wink

Done, also edited the PR description to match the current state

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: I am still trying to understand this but it looks a bit weird so far 😕 

Reconciler runs validation on `pipelineRun.Spec.ServiceAccountNames` and `pipelineRun.Spec.taskRunSpecs`, making sure the tasks specified in these sections exist under `tasks` and `finally`:

https://github.com/tektoncd/pipeline/blob/4fdd018d4b54b5cbd44cf03fd1450ef727942e3e/pkg/reconciler/pipelinerun/pipelinerun.go#L443

and

https://github.com/tektoncd/pipeline/blob/4fdd018d4b54b5cbd44cf03fd1450ef727942e3e/pkg/reconciler/pipelinerun/pipelinerun.go#L451

But, `resolvePipelineState` resolves task with `pr.Spec.ServiceAccountName`:

https://github.com/tektoncd/pipeline/blob/4fdd018d4b54b5cbd44cf03fd1450ef727942e3e/pkg/reconciler/pipelinerun/pipelinerun.go#L471

https://github.com/tektoncd/pipeline/blob/4fdd018d4b54b5cbd44cf03fd1450ef727942e3e/pkg/reconciler/pipelinerun/pipelinerun.go#L266

Only applies to `bundle` at this point, i.e. bundles does not honor `pr.Spec.ServiceAccountNames` or `pr.Spec.TaskRunSpecs` 🤔  

https://github.com/tektoncd/pipeline/blob/4fdd018d4b54b5cbd44cf03fd1450ef727942e3e/pkg/reconciler/taskrun/resources/taskref.go#L53

Looking further 🔍

pritidesai: alright, `createTaskRun` resolves service account name in expected order: 

https://github.com/tektoncd/pipeline/blob/4fdd018d4b54b5cbd44cf03fd1450ef727942e3e/pkg/reconciler/pipelinerun/pipelinerun.go#L654

First, read serviceAccountName specified in `pr.Spec.ServiceAccountNames`:

https://github.com/tektoncd/pipeline/blob/4fdd018d4b54b5cbd44cf03fd1450ef727942e3e/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L141

which get replaced with `pr.Spec.TaskRunSpecs` if specified:

https://github.com/tektoncd/pipeline/blob/4fdd018d4b54b5cbd44cf03fd1450ef727942e3e/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L438-L439

pritidesai: > it shouldn't be looking here `pr.Spec.ServiceAccountNames` (or at least it should be sharing the same code as what the reconciler does)

It's looking here to validate which is fine from the validation perspective, reconciler reads `serviceAccountName`, `serviceAccountNames`, and `taskRunSpecs`, and sets the service account name in expected order.

pritidesai:
@vdemeester any more changes needed here?

vdemeester (approved):

vdemeester (commented):
/lgtm

PR #3582, opened 2020-12-02
Add a "dropNetworking" function and unit tests to the runner package.

dlorenc:
# Changes

This will be used for TEP-0025 [Hermekton](https://github.com/tektoncd/community/blob/master/teps/0025-hermekton.md)
to drop network access for the process run by the runner.

The unit tests here aren't great, but are better than nothing IMO. They can only run on Linux, and test that
network access is not available. To do this, they try to make a network request. If the tests were to run
in a sandboxed environment, we would miss this.

They will be reinforced by e2e tests that properly verify network access exists without this call, and then doesn't
exist with this call.

/kind feature

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
NONE
```

### DIFF
--- a/cmd/entrypoint/namespaces_linux.go
+++ b/cmd/entrypoint/namespaces_linux.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"os/exec"
+	"syscall"
+)
+
+// dropNetworking modifies the supplied exec.Cmd to execute in a net set of namespaces that do not
+// have network access
+func dropNetworking(cmd *exec.Cmd) {
+	// These flags control the behavior of the new process.
+	// Documentation for these is available here: https://man7.org/linux/man-pages/man2/clone.2.html
+	// We mostly want to just create a new network namespace, unattached to any networking devices.
+	// The other flags are necessary for that to work.
+
+	if cmd.SysProcAttr == nil {
+		// We build this up piecemeal in case it was already set, to avoid overwriting anything.
+		cmd.SysProcAttr = &syscall.SysProcAttr{}
+	}
+	cmd.SysProcAttr.Cloneflags = syscall.CLONE_NEWNS |
+		syscall.CLONE_NEWPID | // NEWPID creates a new process namespace
+		syscall.CLONE_NEWNET | // NEWNET creates a new network namespace (this is the one we really care about)
+		syscall.CLONE_NEWUSER // NEWUSER creates a new user namespace
+
+	// We need to map the existing user IDs into the new namespace.
+	// Just map everything.
+	cmd.SysProcAttr.UidMappings = []syscall.SysProcIDMap{
+		{
+			ContainerID: 0,
+			HostID:      0,
+			// Map all users
+			Size: 4294967295,
+		},
+	}
+
+	// This is needed to allow programs to call setgroups when in a new Gid namespace.
+	// Things like apt-get install require this to work.
+	cmd.SysProcAttr.GidMappingsEnableSetgroups = true
+	// We need to map the existing group IDs into the new namespace.
+	// Just map everything.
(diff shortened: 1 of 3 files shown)


cmd/entrypoint/namespaces_test.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"os/exec"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+// This isn't a great unit test, but it's the best I can think of.
+// It can only run on Linux, and tries to make a network connection.
+func TestDropNetworking(t *testing.T) {
```
sbwsg, line 14: What's the expected behaviour wrt Sidecars alongside Steps?  These share a `localhost` under normal circumstances in a Pod I think?

My assumption reading this is that `localhost` of one process is not accessible to another process in new networking namespace, is that right?  Worth unit testing that behaviour as well with a local http server running in a separate process?

cmd/entrypoint/namespaces_test.go
```diff
@@ -0,0 +1,38 @@
...
+package main
+
+import (
+	"os/exec"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+// This isn't a great unit test, but it's the best I can think of.
```
sbwsg, line 12: Maybe document the reasons that it's not great here too in case an enterprising dev comes along with ideas to fix it?  Having said that, if the issues are just that it's linux-only or that it calls out to an external website then I suggest not including this comment.  A linux-only test is fine (imo) for a linux-only feature (until we're proactively supporting bsd / jails?) and we have other tests that expect access to external resources e.g. external git repositories.

sbwsg (approved):

cmd/entrypoint/namespaces_test.go
dlorenc: Good question, I think for that we'll need an actual e2e test to really simulate the behavior of the containers in a Pod. I'll make sure we include that case.

bobcatfish:
Loving these small PRs!!!! 😍

cmd/entrypoint/namespaces_test.go
dlorenc: Good call! Explained the main issue in the comment.

sbwsg:
This lgtm but i already approved it.  @tektoncd/core-maintainers any objections to merging this?

cmd/entrypoint/namespaces_linux.go
```diff
@@ -0,0 +1,50 @@
...
+	cmd.SysProcAttr.GidMappings = []syscall.SysProcIDMap{
+		{
+			ContainerID: 0,
+			HostID:      0,
+
+			//  Map all groups
+			Size: 4294967295,
+		},
+	}
+}
```
bobcatfish, line 50: this function is beautiful 🤩

cmd/entrypoint/namespaces_test.go
```diff
@@ -0,0 +1,41 @@
...
```
bobcatfish, line 12: this is such a nitpick but i think if you start docstrings with anything other than the name of the function it's confusing for tools

cmd/entrypoint/namespaces_test.go
```diff
@@ -0,0 +1,41 @@
...
+// It attempts to verify there is no network access by making a network
+// request. If the test were to run in an offline environment, or an already
+// sandboxed environment, the test could pass even if the dropNetworking
+// function did nothing.
```
bobcatfish, line 16: i guess the other thing you could do is assert on the exact values you set on the cmd object but that doesn't feel super useful

bobcatfish (commented):
/lgtm

PR #3583, opened 2020-12-02
retrieving deps outside of dag.build

pritidesai:
# Changes

Adding an attribute function to `PipelineTaskList` to retrieve `deps` (thanks @bobcatfish for great suggestion), passing this `deps` to `dag.Build` function instead of retrieving deps from within that function.

This change was motivated while introducing task result consumption in finally tasks. Finally tasks can consume results from dag tasks but dag tasks are not part of the same graph/section and must not be added as part of deps of any finally tasks. This change allows reconciler (with the knowledge of dag tasks vs finally tasks) to manage deps while building a graph.

/kind misc

# Submitter Checklist

review them:

- [] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)

## Reviewer Notes

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipeline/dag/dag.go
+++ b/pkg/reconciler/pipeline/dag/dag.go
@@ -67,17 +67,16 @@ func (g *Graph) addPipelineTask(t Task) (*Node, error) {
 }
 
 // Build returns a valid pipeline Graph. Returns error if the pipeline is invalid
-func Build(tasks Tasks) (*Graph, error) {
+func Build(tasks Tasks, deps map[string][]string) (*Graph, error) {
 	d := newGraph()
 
-	deps := map[string][]string{}
 	// Add all Tasks mentioned in the `PipelineSpec`
 	for _, pt := range tasks.Items() {
 		if _, err := d.addPipelineTask(pt); err != nil {
 			return nil, fmt.Errorf("task %s is already present in Graph, can't add it again: %w", pt.HashKey(), err)
 		}
-		deps[pt.HashKey()] = pt.Deps()
 	}
+
 	// Process all from and runAfter constraints to add task dependency
 	for pt, taskDeps := range deps {
 		for _, previousTask := range taskDeps {
--- a/pkg/apis/pipeline/v1alpha1/pipeline_types.go
+++ b/pkg/apis/pipeline/v1alpha1/pipeline_types.go
@@ -201,6 +201,14 @@ func (l PipelineTaskList) Items() []dag.Task {
 	return tasks
 }
 
+func (l PipelineTaskList) Deps() map[string][]string {
+	deps := map[string][]string{}
+	for _, pt := range l {
+		deps[pt.HashKey()] = pt.Deps()
+	}
+	return deps
+}
+
 // PipelineTaskParam is used to provide arbitrary string parameters to a Task.
 type PipelineTaskParam = v1beta1.PipelineTaskParam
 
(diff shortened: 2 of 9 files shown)


pkg/reconciler/pipeline/dag/dag.go
```diff
@@ -67,17 +67,16 @@ func (g *Graph) addPipelineTask(t Task) (*Node, error) {
 }
 
 // Build returns a valid pipeline Graph. Returns error if the pipeline is invalid
-func Build(tasks Tasks) (*Graph, error) {
+func Build(tasks Tasks, deps map[string][]string) (*Graph, error) {
```
vdemeester, line 70: Any reason to pass `deps` there as it comes from `tasks` ? we could just do

```go
func Build(tasks Tasks) (*Graph, error) {
	deps = tasks.Deps()
}
```

vdemeester (commented):
This changes makes us do 2 loops instead of one right ?

pkg/reconciler/pipeline/dag/dag.go
bobcatfish: @vdemeester that's what it currently does - maybe @pritidesai can add more info in the commit message to explain but the reason for this change is to be able to call Build with an empty dep list

in the case of finally tasks, we create 2 DAGs. when finally tasks depend on results from non-finally tasks, Deps() will return the non-finally Tasks (or will it error? im not sure). but in that case we don't want to build connections between the finally tasks and the non-finally tasks

making this change enables us to build a graph in both cases

another option might be to stop treating finally tasks as a graph but i have a strong feeling we will be adding more graph like features in future so im not sure we should do that now

bobcatfish:
nice!! one request, @pritidesai could you include some unit tests for the new function?

/approve

pkg/reconciler/pipeline/dag/dag.go
bobcatfish: oh actually @pritidesai 's commit message does already explain this pretty well i think: https://github.com/tektoncd/pipeline/pull/3583/commits/9716d32b77d5e4a63f7be8c624406dfc28b8e88f

vdemeester: oh I did not read the commit :sweat_smile:

pritidesai:
> nice!! one request, @pritidesai could you include some unit tests for the new function?

yup, creating a separate PR.

pritidesai:
>> nice!! one request, @pritidesai could you include some unit tests for the new function?

> yup, creating a separate PR.

Please see PR #3597

bobcatfish:
hey @vdemeester just a comment that it would have been better to include the unit tests in the same PR (I think including tests and docs in PRs has been a great strategy for us; i want to make sure we don't end up having to play catch up with changes) - maybe it would be been clearer if i'd included a `/block` when i added the `/approve`?

vdemeester:
a `/hold`, yes :wink:

pritidesai:
> hey @vdemeester just a comment that it would have been better to include the unit tests in the same PR (I think including tests and docs in PRs has been a great strategy for us; i want to make sure we don't end up having to play catch up with changes) - maybe it would be been clearer if i'd included a `/block` when i added the `/approve`?

I think its my mistake, should have added unit tests in this PR instead of leaving it up to the reviewer. I was thinking the graphs we are testing in `dag_test.go` are sufficient which is indirectly testing `Deps()` but its always better to add unit tests for newly added functions. My apologies 🙏

bobcatfish:
No worries @pritidesai !! :) ❤️

PR #3603, opened 2020-12-05
Implement a Runnable duck type.

mattmoor:
# Changes

This sketches out a duck type for the shared fields between Task and Pipeline, which may be used to fetch either resource and reason about their common properties.

This type may also be used with the `duck.InformerFactory` to stamp out new Informer/Lister pairs for GVRs that adhere to this contract.

This also adds a boilerplate `implements_test.go` to verify this and existing duck contracts are satisfied by the Tekton resource types.

/kind cleanup

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
NONE
```


mattmoor:
My motivation for this was that I was recently authoring some client code for the Tekton API and I wanted a way to generically reason about these properties across Tasks and Pipelines, as well as forthcoming custom-tasks.  I ended up needing to copy/pasta a bunch of logic that would ideally be shared across these, where if this type had existed I could have used it with the dynamic client to have a single source of truth.

I spent about 5 minutes on this, so I'm happy to change any of it, but thought I'd open a PR to also highlight some of the utilities we have for ensuring consistent APIs across duck types.

cc @ImJasonH @vdemeester @afrittoli

ImJasonH (commented):
Is this interface something you can define in your own codebase? I'm not opposed to having it defined here, but there's (as yet) nothing defined that uses it. So if you can define and use it in your code to unblock yourself that might be good.

Not all Custom Task CRDs will implement this so we might want to change it in the future, which means deprecation burden.

mattmoor:
I could define it downstream for my own consumption, but I think the key thing for ensuring folks ability to reason about the "category" of custom tasks will be providing ways (like this) to reason about the pieces that are expected to be common.

Given that this is a go-level thing, I'm a bit less worried about the deprecation period aspect of it, since it only applies to folks vendoring things and it can be broken as-needed to alert folks of the changing definition, but IDK what position you folks have taken on go-level deprecation/breakages? 🤷

ImJasonH:
> I could define it downstream for my own consumption, but I think the key thing for ensuring folks ability to reason about the "category" of custom tasks will be providing ways (like this) to reason about the pieces that are expected to be common.
> 
> Given that this is a go-level thing, I'm a bit less worried about the deprecation period aspect of it, since it only applies to folks vendoring things and it can be broken as-needed to alert folks of the changing definition, but IDK what position you folks have taken on go-level deprecation/breakages? 🤷

It's true, we've taken the stance that Go-level breakages aren't covered by the deprecation policy. At the same time, we shouldn't break things unnecessarily and cause churn either.

Another concern is documentation burden. Custom tasks are just now being implemented and supported in Pipelines. There's very little documentation yet, and even fewer samples. I'd prefer to commit to these shapes, how they compose, how they're supported by Tekton-provided helper methods, after we have a bit more experience using them in the real world. In particular, some custom task implementations might be Parameterizeable without being Resultable, or might be Workspaceable but nothing else.

I think the best way forward for now would be to define these in your own code, and wait for custom tasks to settle down a bit  more and get better documentation and samples. I fully expect something like this will be useful in the future, but there's also no urgent need to define and officially support these right now.

mattmoor:
👍 LMK when you have something I can use 😇

PR #3611, opened 2020-12-08
Run controller and webhook also as non-root group. Drop capabilities.

benhxy:
# Changes

/kind misc
Containers running as root group or with excessive capabilities
have larger security impact if compromised. It is recommended
to run containers as non-root and drop unneeded capabilities.
Controller and Webhook are regular servers and don't need special
capabilities. This change is tested with master version on a GKE cluster.

# Submitter Checklist

review them:

## Reviewer Notes

```release-note
NONE
```

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -111,8 +111,12 @@ spec:
           value: tekton.dev/pipeline
         securityContext:
           allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - all
           # User 65532 is the distroless nonroot user ID
           runAsUser: 65532
+          runAsGroup: 65532
         ports:
         - name: probes
           containerPort: 8080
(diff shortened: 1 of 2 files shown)


benhxy:
@vinayakankugoyal FYI

dibyom:
/lgtm

/cc @vdemeester 

Thanks! We should probably do this for the other deployments in Tekton too (triggers, operator, results)

vdemeester:
> /lgtm
> 
> /cc @vdemeester
> 
> Thanks! We should probably do this for the other deployments in Tekton too (triggers, operator, results)

Agreed 😉 . This is another thing we will need to "mangle" for OpenShift Pipelines (because OpenShift does a lot of this on its own) but it is a really good thing to do !

/lgtm

PR #3614, opened 2020-12-08
Add some more release process details 🤓

bobcatfish:
# Changes

Worked with @jerop to create the v0.19.0 release today and there were a
few parts of the cheat sheet that felt like they could use a bit more detail:

- You can configure kubectl to connect to multiple clusters if you want
- Some info on how to choose the commit to release
- The step that applies the PipelineResource was applying a file in the
  tekton dir that probably wasnt the one you meant to apply (the old
  instructions directed ppl to edit that file)
- ~Applying the env vars doesn't work with out `export`~ (thanks for explaining export vs. set @sbwsg !)
- Changed the numbering in markdown to all be `1.` so you can add more
  steps without having to update all the numbers
- Added some guidance on naming the release.
- Mentioned that you have to uncheck the pre-release box (which I didn't
  realize last time XD)
- Don't use patch versions in the example

/kind documentation

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
NONE
```

### DIFF
--- a/tekton/release-cheat-sheet.md
+++ b/tekton/release-cheat-sheet.md
@@ -6,13 +6,20 @@ the pipelines repo, a terminal window and a text editor.
 
 1. `cd` to root of Pipelines git checkout.
 
-2. Point `kubectl` at dogfooding cluster.
+1. Configure `kubectl` to use the  dogfooding cluster
+   [as the default context](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/)
+   (**in the final step we will configure the default context to point to something else; until then be aware
+    that any kubctl commands you run will act on the dogfooding cluster**)
 
     ```bash
     gcloud container clusters get-credentials dogfooding --zone us-central1-a --project tekton-releases
     ```
 
-3. Create PipelineResource for new version. e.g.
+1. Select the commit you would like to build the release from, most likely the
+   most recent commit at https://github.com/tektoncd/pipeline/commits/master
+   and note the commit's hash.
+
+1. Create a `.yaml` file containing PipelineResource for new version. e.g.
 
     ```yaml
     apiVersion: tekton.dev/v1alpha1
@@ -25,26 +32,26 @@ the pipelines repo, a terminal window and a text editor.
       - name: url
         value: https://github.com/tektoncd/pipeline
       - name: revision
-        value: # Example: 33e0847e67fc9804689e50371746c3cdad4b0a9d
+        value: # The commmit you selected in the last step, e.g. 33e0847e67fc9804689e50371746c3cdad4b0a9d
     ```
 
-4. `kubectl apply -f tekton/resources.yaml`
+1. Apply file you just made to the dogfooding cluster: `kubectl apply -f your-pipeline-resource-file.yaml`
 
-5. Create environment variables for bash scripts in later steps.
+1. Create environment variables for bash scripts in later steps.
 
     ```bash
-    TEKTON_VERSION=# Example: v0.11.2
-    TEKTON_RELEASE_GIT_RESOURCE=# Example: tekton-pipelines-v0-11-2
-    TEKTON_IMAGE_REGISTRY=gcr.io/tekton-releases
+    TEKTON_VERSION=# Example: v0.11.0
+    TEKTON_RELEASE_GIT_RESOURCE=# Name of the resource you created, e.g.: tekton-pipelines-v0-11-0
+    TEKTON_IMAGE_REGISTRY=gcr.io/tekton-releases # only change if you want to publish to a different registry
     ```
 
-6. Confirm commit SHA matches what you want to release.
+1. Confirm commit SHA matches what you want to release.
 
     ```bash
     kubectl get pipelineresource "$TEKTON_RELEASE_GIT_RESOURCE" -o=jsonpath="{'Target Revision: '}{.spec.params[?(@.name == 'revision')].value}{'\n'}"
     ```
 
-7. Execute the release pipeline.
+1. Execute the release pipeline.
 
     **If you are backporting include this flag: `--param=releaseAsLatest="false"`**
 
@@ -70,19 +77,22 @@ the pipelines repo, a terminal window and a text editor.
     pipeline-release
     ```
 
-8. Watch logs of pipeline-release.
+1. Watch logs of pipeline-release.
 
-9. The YAMLs are now released! Anyone installing Tekton Pipelines will now get the new version. Time to create a new GitHub release announcement:
+1. The YAMLs are now released! Anyone installing Tekton Pipelines will now get the new version. Time to create a new GitHub release announcement:
+
+    1. Choose a name for the new release! The usual pattern is "< cat breed > < famous robot >" e.g. "Ragdoll Norby".
+       (Check the previous releases to avoid repetition: https://github.com/tektoncd/pipeline/releases.)
 
     1. Create additional environment variables
 
     ```bash
     TEKTON_OLD_VERSION=# Example: v0.11.1
-    TEKTON_RELEASE_NAME=# Example: "Ragdoll Norby"
+    TEKTON_RELEASE_NAME=# The release name you just chose, e.g.: "Ragdoll Norby"
     TEKTON_PACKAGE=tektoncd/pipeline
     ```
 
-    2. Execute the Draft Release task.
+    1. Execute the Draft Release task.
 
     ```bash
     tkn task start \
@@ -95,26 +105,30 @@ the pipelines repo, a terminal window and a text editor.
       create-draft-release
     ```
 
-    4. Watch logs of create-draft-release
+    1. Watch logs of create-draft-release
 
-    5. On successful completion, a URL will be logged. Visit that URL and sort the
-    release notes. **Double-check that the list of commits here matches your expectations
-    for the release.** You might need to remove incorrect commits or copy/paste commits
-    from the release branch. Refer to previous releases to confirm the expected format.
+    1. On successful completion, a URL will be logged. Visit that URL and sort the
+    release notes.
+      1. Manually add upgrade and deprecation notices based on the generated release notes
+      1. Double-check that the list of commits here matches your expectations
+         for the release. You might need to remove incorrect commits or copy/paste commits
+         from the release branch. Refer to previous releases to confirm the expected format.
 
-    6. Publish the GitHub release once all notes are correct and in order.
+    1. Un-check the "This is a pre-release" checkbox since you're making a legit for-reals release!
 
-10. Edit `README.md` on `master` branch, add entry to docs table with latest release links.
+    1. Publish the GitHub release once all notes are correct and in order.
 
-11. Push & make PR for updated `README.md`
+1. Edit `README.md` on `master` branch, add entry to docs table with latest release links.
 
-12. **Important: Stop pointing `kubectl` at dogfooding cluster.**
+1. Push & make PR for updated `README.md`
+
+1. **Important: Stop pointing `kubectl` at dogfooding cluster.**
 
     ```bash
     kubectl config use-context my-dev-cluster
     ```
 
-13. Test release that you just made.
+1. Test release that you just made.
 
     ```bash
     # Test latest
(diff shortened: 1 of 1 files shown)


tekton/release-cheat-sheet.md
```diff
@@ -6,13 +6,20 @@ the pipelines repo, a terminal window and a text editor.
 
 1. `cd` to root of Pipelines git checkout.
 
-2. Point `kubectl` at dogfooding cluster.
+1. Configure `kubectl` to use the  dogfooding cluster.
 
     ```bash
+    # This will update your kubectl to automatically connect to the dogfood cluster.
+    # Remember to point it back to your own cluster afterward, or configure acesss
+    # to multiple clusters: https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/
```
dlorenc, line 14: I don't think the steps you linked to are really an "or". The gcloud command does exactly that - it fetches the auth information, creates a new cluster entry in your kubectl, and then also swaps the current cluster over to that one. You need to do that the first time you use the dogfood cluster, but you can use normal kubectl config commands every time after that.

dlorenc (approved):

tekton/release-cheat-sheet.md
```diff
@@ -25,26 +32,26 @@ the pipelines repo, a terminal window and a text editor.
...
+1. Apply file you just made to the dogfooding cluster: `kubectl apply -f your-pipeline-resource-file.yaml`
 
-5. Create environment variables for bash scripts in later steps.
+1. Create environment variables for bash scripts in later steps.
 
     ```bash
-    TEKTON_VERSION=# Example: v0.11.2
-    TEKTON_RELEASE_GIT_RESOURCE=# Example: tekton-pipelines-v0-11-2
-    TEKTON_IMAGE_REGISTRY=gcr.io/tekton-releases
+    export TEKTON_VERSION=# Example: v0.11.2
```
pritidesai, line 43: since you are updating it, how about pointing it to a new release instead of patch i.e. replacing `.2` with `.0` in `v0.11.2`

tekton/release-cheat-sheet.md
```diff
@@ -70,19 +77,22 @@ the pipelines repo, a terminal window and a text editor.
...
     ```
 
-8. Watch logs of pipeline-release.
+1. Watch logs of pipeline-release.
+
+1. The YAMLs are now released! Anyone installing Tekton Pipelines will now get the new version. Time to create a new GitHub release announcement:
 
-9. The YAMLs are now released! Anyone installing Tekton Pipelines will now get the new version. Time to create a new GitHub release announcement:
+    1. Choose a name for the new release! The usual pattern is "< cat breed > < famous robot >" e.g. "Ragdoll Norby".
+       (Check the previous releases to avoid repetition: https://github.com/tektoncd/pipeline/releases.)
```
pritidesai, line 85: I have listed all in a gist [here](https://gist.github.com/pritidesai/4fc02c1adb34e85adb1e88a8392f0340#file-tekton-release-titles-csv) 😱 I think it will be very helpful to make it part of the repo/website somewhere, just the table instead of going through more than dozen different release pages.

tekton/release-cheat-sheet.md
```diff
@@ -95,26 +105,28 @@ the pipelines repo, a terminal window and a text editor.
       create-draft-release
     ```
 
-    4. Watch logs of create-draft-release
+    1. Watch logs of create-draft-release
 
-    5. On successful completion, a URL will be logged. Visit that URL and sort the
+    1. On successful completion, a URL will be logged. Visit that URL and sort the
```
pritidesai, line 110: I think we do have to manually add Upgrade and deprecation notices, I am not 100% sure though.

jerop: that's true, I manually added those notices today

vdemeester (approved):

tekton/release-cheat-sheet.md
```diff
@@ -6,13 +6,20 @@ the pipelines repo, a terminal window and a text editor.
 
 1. `cd` to root of Pipelines git checkout.
 
-2. Point `kubectl` at dogfooding cluster.
+1. Configure `kubectl` to use the  dogfooding cluster.
 
     ```bash
+    # This will update your kubectl to automatically connect to the dogfood cluster.
+    # Remember to point it back to your own cluster afterward, or configure acesss
```
sbwsg, line 13: I'd argue this comment isn't necessary and actually hinders the "cheat-sheet"-ness here.  The line before this comment says "Configure `kubectl` to use the dogfooding cluster" which is pretty much the same thing as "This will update your kubectl to connect to the dogfood cluster".  And then the comment about changing back to your own cluster is a dedicated listed action at the end of this cheat-sheet which tells the user to "Stop pointing `kubectl` at dogfooding cluster." with the kubectl command to use.

Generally speaking in a "cheat-sheet" this kind of commentary tends to slow me down rather than help me, and in addition it's asking me to hold something in my working memory while I go through a bunch of other steps, which is a little antithetical to the doc's purpose.

If the action at the end which describes how to switch back to your own cluster isn't enough I suggest updating it to be more helpful rather than add exposition in earlier sections.

tekton/release-cheat-sheet.md
```diff
@@ -25,26 +32,26 @@ the pipelines repo, a terminal window and a text editor.
...
-5. Create environment variables for bash scripts in later steps.
+1. Create environment variables for bash scripts in later steps.
 
     ```bash
-    TEKTON_VERSION=# Example: v0.11.2
-    TEKTON_RELEASE_GIT_RESOURCE=# Example: tekton-pipelines-v0-11-2
-    TEKTON_IMAGE_REGISTRY=gcr.io/tekton-releases
+    export TEKTON_VERSION=# Example: v0.11.2
+    export TEKTON_RELEASE_GIT_RESOURCE=# Name of the resource you created, e.g.: tekton-pipelines-v0-11-2
+    export TEKTON_IMAGE_REGISTRY=gcr.io/tekton-releases # only change if you want to publish to a different registry
```
sbwsg, line 45: Which step further down failed when you didn't include the `export`?  FWIW I don't think I've ever had to use them and it shouldn't be necessary unless (for example) I opened a tmux after setting these envs, or otherwise somehow created a subprocess shell before using them.

sbwsg (commented):
I'd like to push back a bit on any PRs that are introducing multiline comments without associated listed actions to the cheat-sheet.  If we really want a doc with exposition / prose then I'd like to suggest we bring back the long-form deployment guide we used to have.  Totally my own POV but I much prefer to keep these kinds of docs as focused, concise and accurate as possible.

tekton/release-cheat-sheet.md
```diff
@@ -126,9 +138,9 @@ the pipelines repo, a terminal window and a text editor.
     kubectl apply --filename https://storage.googleapis.com/tekton-releases/pipeline/previous/v0.11.2/release.yaml
     ```
 
-14. Announce the release in Slack channels #general and #pipelines.
+1. Announce the release in Slack channels #general and #pipelines.
 
-15. Update [the catalog repo](https://github.com/tektoncd/catalog) test infrastructure
+1. Update [the catalog repo](https://github.com/tektoncd/catalog) test infrastructure
 to use the new release by updating the `RELEASE_YAML` link in [e2e-tests.sh](https://github.com/tektoncd/catalog/blob/v1beta1/test/e2e-tests.sh).
```
bobcatfish, line 144: @jerop just noticed this link is using the wrong branch

tekton/release-cheat-sheet.md
bobcatfish: oh wow that would be really handy to refer to XD

tekton/release-cheat-sheet.md
bobcatfish: I'll add this to the list!

tekton/release-cheat-sheet.md
bobcatfish: >  If the action at the end which describes how to switch back to your own cluster isn't enough I suggest updating it to be more helpful rather than add exposition in earlier sections.

Personally I find that my strict adherence to the steps in the cheat sheet wanes the further I get down the list - and maybe more importantly, if something interrupts me halfway through, this would leave me in a dangerous state where hopefully I get to the last item in the cheat sheet list but if I don't, I might modify the CI cluster

> it's asking me to hold something in my working memory while I go through a bunch of other steps

I feel like its worse to ask you to definitely complete all the steps and hold in your working memory that you are by default connected to a "production" cluster 😬 

@sbwsg what if we default to using the `--context` arg in the instructions? maybe we could assume you have a `--context dogfooding` context configured or similar, and link to some docs on how to set that up?

tekton/release-cheat-sheet.md
bobcatfish: In my environment (zsh with iterm on a mac) a line like `TEKTON_RELEASE_GIT_RESOURCE="foo"` is only applicable for the running command (process?), e.g. something like `TEKTON_RELEASE_GIT_RESOURCE="foo" mycommand` but isn't retained for future commands unless exported, e.g.:

1. I run the command to set the var:
```
TEKTON_RELEASE_GIT_RESOURCE="foo"
```

2. I run `env` to see what environment variables are set, it's not listed

3. I export it instead:
```
 export TEKTON_RELEASE_GIT_RESOURCE="foo"
```

4. Now when I run `env` I see it set:
```
env
...
TEKTON_RELEASE_GIT_RESOURCE=foo
_=/usr/bin/env
```

I think this explains it pretty well: https://stackoverflow.com/questions/7328223/unix-export-command 

> If you set a variable at the command-line like
> $ FOO="bar"
> That variable will not be visible in child processes

I kinda wonder how it's working with your setup!

tekton/release-cheat-sheet.md
bobcatfish: kk ill try to make this more accurate

tekton/release-cheat-sheet.md
sbwsg: I'm running the same shell and OS.  It's true that it won't show up with `env` because `env` only shows variables marked for export.  Run `set` and you'll see it listed.

Here's a tiny example to test it out:

```bash
$ FOO=bar
$ echo $FOO
```

This prints "bar" for me.

tekton/release-cheat-sheet.md
sbwsg: > @sbwsg what if we default to using the --context arg in the instructions? maybe we could assume you have a --context dogfooding context configured or similar, and link to some docs on how to set that up?

Seems like a reasonable compromise!  👍

tekton/release-cheat-sheet.md
bobcatfish: ahhh okay i think im starting to understand what's happening - i think im explaining what you already understand, but for my own understanding: so if you say `FOO=bar`, it will be set for the current process (TIL about `set`) - if you wanted FOO to be set for any child processes,  you'd need to use `export`

```bash
# set BAR in the current environment/process
✗ BAR=baz
✗ echo $BAR
baz

# we have a script that tries to use the same variable
 ✗ cat foo.sh
set -x

echo $BAR

# the variable isn't exported, so when this is run, a child process starts and is not provided with that variable
✗ ./foo.sh
+ echo

# if we export the variable, it will be provided to the child process
✗ export BAR=hello
✗ ./foo.sh
+ echo hello
hello
```

And in the cheatsheet, we're resolving the variables in the shell that runs the commands, so no need to pass them to any child processes - so I think I'm on board now with not including `export` - lemme know if my understanding isn't correct @sbwsg !

bobcatfish: I think I've made this same change in other docs 🤦‍♀️  I'll try to reverse it when I see it. Thanks again @sbwsg , I feel like my overall understanding is deeper now :D

tekton/release-cheat-sheet.md
```diff
@@ -95,40 +98,59 @@ the pipelines repo, a terminal window and a text editor.
...
     ```bash
     # Test latest
-    kubectl apply --filename https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml
+    kubectl --context dogfooding apply --filename https://storage.googleapis.com/tekton-releases/pipeline/latest/release.yaml
     ```
 
     ```bash
     # Test backport
-    kubectl apply --filename https://storage.googleapis.com/tekton-releases/pipeline/previous/v0.11.2/release.yaml
+    kubectl --context dogfooding apply --filename https://storage.googleapis.com/tekton-releases/pipeline/previous/v0.11.2/release.yaml
```
bobcatfish, line 126: @sbwsg @afrittoli is the idea that you'd run these tests directly against the dogfood cluster, or against your own cluster? (i have to admit ive always skipped this step... 😅 )

tekton/release-cheat-sheet.md
bobcatfish: okay updated @sbwsg lemme know what you think!

tekton/release-cheat-sheet.md
sbwsg: I feel like "test" is, in hindsight, maybe too strong a word here - all I do is apply the new release to my own cluster and watch the deployments in the tekton namespace to make sure that the controller comes up successfully with the new version.

sbwsg (approved):

sbwsg:
Looks great, thanks for making these changes!

tekton/release-cheat-sheet.md
sbwsg: (but yeah, sorry, to answer the question - apply them to your own cluster, not to dogfooding)

bobcatfish: ahhh kk, I'll change this to use the my-dev-cluster context then :D

sbwsg:
Given that this has three approves I'm going to switch mine to an lgtm. 🚀 

/lgtm

Issue #3608, opened 2020-12-08
Metrics `podLatency` always seems to be 0


yaoxiaoqi:
# Expected Behavior
podLatency is supposed to report the pod creation->start latency, which should not be 0 almost every time.
# Actual Behavior
Currently, podLatency calculates the duration between pod creation time and pod scheduled time, which means if your k8s cluster is not busy enough it would be zero most times. I started 25 taskrun at the same time and none of them has non-zero podLatency.
# Steps to Reproduce the Problem

1. open http://localhost:9090/metrics and find the `tekton_taskruns_pod_latency` metrics
2. it is 0 every time

```log
tekton_taskruns_pod_latency{namespace="arendelle-nsfqw",pod="duplicate-pod-task-run-bidobdij-pod-77wd9",task="anonymous",taskrun="duplicate-pod-task-run-bidobdij"} 0
tekton_taskruns_pod_latency{namespace="arendelle-nsfqw",pod="duplicate-pod-task-run-blpdlefs-pod-rgsdq",task="anonymous",taskrun="duplicate-pod-task-run-blpdlefs"} 0
tekton_taskruns_pod_latency{namespace="arendelle-nsfqw",pod="duplicate-pod-task-run-etmqrfnn-pod-plv8h",task="anonymous",taskrun="duplicate-pod-task-run-etmqrfnn"} 0
tekton_taskruns_pod_latency{namespace="arendelle-nsfqw",pod="duplicate-pod-task-run-ghrrcovo-pod-mltbw",task="anonymous",taskrun="duplicate-pod-task-run-ghrrcovo"} 0
tekton_taskruns_pod_latency{namespace="arendelle-nsfqw",pod="duplicate-pod-task-run-gjjdplpb-pod-pl2ck",task="anonymous",taskrun="duplicate-pod-task-run-gjjdplpb"} 0
tekton_taskruns_pod_latency{namespace="arendelle-nsfqw",pod="duplicate-pod-task-run-gwkmpygl-pod-jxgnr",task="anonymous",taskrun="duplicate-pod-task-run-gwkmpygl"} 0
tekton_taskruns_pod_latency{namespace="arendelle-nsfqw",pod="duplicate-pod-task-run-hndhvbnr-pod-tn4th",task="anonymous",taskrun="duplicate-pod-task-run-hndhvbnr"} 0
[...]

PR #3624, opened 2020-12-10
Change the implementation of the metrics pod_latency

yaoxiaoqi:
# Changes

This PR closes https://github.com/tektoncd/pipeline/issues/3608 .

Currently, podLatency reports the duration between pod creation time and pod scheduled
time, which means if the k8s cluster is not busy enough it would be 0 at most times.

In this PR, podLatency is changed to report the duration between pod creation time
and the time at which pod phase turns to `Running`. The startRunningTime is not stored
in the pod status so the time is collected in the reconciliation loop. Due to the latency
of reconciler, the minimum value of podLatency is 2s.

Compared to old podLatency, the new podLatency can report the duration that a pod takes 
to get the primary containers ready and started more accurately. For example, some pods
needs to download the image before running. The old podLatency is still 0 in this situation. 
But the new podLatency will report the downloading time with 2 second margin.

The unit of podLatency is changed to seconds from nanoseconds for better understanding.

/kind feature

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Change podLatency to report the duration between between pod creation time and the time at which pod phase turns to `Running`
```

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -634,6 +634,7 @@ github.com/google/gofuzz v1.1.0 h1:Hsa8mG0dQ46ij8Sl2AYJDUv1oA9/d6Vk+3LG99Oe02g=
 github.com/google/gofuzz v1.1.0/go.mod h1:dBl0BpW6vV/+mYPU4Po3pmUjxk6FQPldtuIdl/M65Eg=
 github.com/google/licenseclassifier v0.0.0-20190926221455-842c0d70d702/go.mod h1:qsqn2hxC+vURpyBRygGUuinTO42MFRLcsmQ/P8v94+M=
 github.com/google/licenseclassifier v0.0.0-20200402202327-879cb1424de0/go.mod h1:qsqn2hxC+vURpyBRygGUuinTO42MFRLcsmQ/P8v94+M=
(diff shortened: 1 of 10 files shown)


ImJasonH:
Hey Yumeng, thanks for the PR! 🎉 

Two things:

1.) Since this is a change to the user-facing API, we'll need more approvers to sign off. If there are significant questions we might want to answer them in a brief TEP which we share with the community for feedback before proceeding.

2.) Can you squash your commits into a single commit, per https://github.com/tektoncd/community/blob/master/standards.md#commits ?

yaoxiaoqi:
> 1.) Since this is a change to the user-facing API, we'll need more approvers to sign off. If there are significant questions we might want to answer them in a brief TEP which we share with the community for feedback before proceeding.
> 
Sure, I will write a short TEP on this. I also concerned whether it's ok to add a new variable `runAt` to implement this, so I also sent another PR https://github.com/tektoncd/pipeline/pull/3635 without `runAt`. It would be great if you could look at that either.

> 2.) Can you squash your commits into a single commit, per https://github.com/tektoncd/community/blob/master/standards.md#commits ?
>
Definitely. I thought there was no need to squash them because the PR is far away from being merged😂. I also need to add more tests before merging if this implementation is approved.

jerop:
this change requires a TEP, and any prior discussions can be done in issue https://github.com/tektoncd/pipeline/issues/3608, so closing this PR for now (@yaoxiaoqi feel free to reopen when the TEP is implementable)

PR #3635, opened 2020-12-14
Change the implementation of the metrics podLatency with a map

yaoxiaoqi:
# Changes

This PR fixs #3608 .

Currently, podLatency reports the duration between pod creation time and pod scheduled
time, which means if the k8s cluster is not busy enough it would be 0 at most times.

In this PR, podLatency is changed to report the duration between pod creation time
and the time at which pod phase turns to `Running`. The startRunningTime is not stored
in the pod status so the time is collected in the reconciliation loop. Due to the latency
of reconciler, the minimum value of podLatency is 2s.

The unit of podLatency is changed to seconds from nanoseconds for better understanding.

Compared to another alternative PR https://github.com/tektoncd/pipeline/pull/3624. Instead
of adding a new taskrun status variable, this PR adds a map to metrics structure which
is used to save the startRunning time of every taskrun pods. The key of the map is
pod UID. This has some caveat that the delay of podLatency is longer than
https://github.com/tektoncd/pipeline/pull/3624.

/kind feature

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Change podLatency to report the duration between between pod creation time and the time at which pod phase turns to `Running`
```

### DIFF
--- a/pkg/reconciler/taskrun/metrics.go
+++ b/pkg/reconciler/taskrun/metrics.go
@@ -30,6 +30,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/metrics"
 )
@@ -57,7 +58,7 @@ var (
 
 	podLatency = stats.Float64("taskruns_pod_latency",
 		"scheduling latency for the taskruns pods",
-		stats.UnitMilliseconds)
+		stats.UnitSeconds)
 
 	cloudEvents = stats.Int64("cloudevent_count",
 		"number of cloud events sent including retries",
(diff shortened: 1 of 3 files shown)


pkg/reconciler/taskrun/metrics.go
```diff
@@ -314,11 +318,21 @@ func (r *Recorder) RecordPodLatency(pod *corev1.Pod, tr *v1beta1.TaskRun) error
 		return err
 	}
 
-	metrics.Record(ctx, podLatency.M(float64(latency)))
+	metrics.Record(ctx, podLatency.M(float64(latency/time.Second)))
 
 	return nil
 }
 
+func (r *Recorder) MarkRunningTime(pod *corev1.Pod, runningTime metav1.Time) error {
```
ImJasonH, line 326: So with this implementation, the runAt time is effectively only stored in this map, in memory. If the controller restarts (e.g., a panic, a node failure, an upgrade), that value is lost, and metrics will either report no data, or incorrect data.

To mitigate this, we can rely on the global resync that happens when a controller restarts, to re-populate the value, but that might be tricky too, and tricky to test since it relies on causing unusual state to exercise the code path.

PR #3636, opened 2020-12-14
Improve documentation around variable behaviour and gotchas

### LINKED ISSUE #3039: Variable Substitution documentation does not explain how it works
The [main doc page](https://github.com/tektoncd/pipeline/tree/master/docs) has a title "Understanding Tekton Pipelines" and under that, there is a link to [Variable Substitutions](https://github.com/tektoncd/pipeline/blob/master/docs/variables.md).

I expect that link to _explain how variable substitutions work_.

From [discussions on Slack](https://tektoncd.slack.com/archives/CLCCEBUMU/p1596130404055400), I now know that there is also:
* arrays
* star-operator to expand arrays (explained in [specifying params](https://github.com/tektoncd/pipeline/blob/master/docs/tasks.md#specifying-parameters))

Also 
* how it works within the `script:` field is not documented. It should be noted that variable substitution can be used within the middle of a _script_ and is executed before (?)
* that it can be expanded in `steps.args` list is not documented

* It would also be good to describe how I can _debug_ variable substitution? e.g. how I quickly can see to what values was inserted. (or maybe this is obvious?)

I expect those things to be explained on the docs, when I follow the doc-flow described initially.

### LINKED ISSUE #3458: Document current best practice for avoiding script injection with variable substitution
## Change Request

In [TEP 17](https://github.com/tektoncd/community/pull/208) @coryrc documented a way to prevent script injection when substituting variables in, for example, `script` fields of Steps.

The approach is to do something like this:

```yaml
   steps:
    - env:
        - name: SCRIPT_CONTENTS
          value: $(params.script)
      script: |
        printf '%s' "${SCRIPT_CONTENTS}" > input-script.sh
        chmod +x input-script.sh
        cd "$(params.package)"
        runner.sh /input-script
```

### What To Do

Make a PR to Tekton's official docs that shows this approach along with a description of how/why this mitigates the risk of script injection.

sbwsg:
# Changes

We used to link from the docs directory's README.md to our variables.md doc
but variables.md doesn't really explain the way variables work - it just
describes what values they can be used for and where they can be used in
our CRDs.

This commit changes the docs directory's README.md to point at the variable
documentation in our task doc, and adds a few lines there to talk a bit
more about how the substitution happens.

This commit also describes how to safely inject variables into `bash`
`script` blocks in Tasks without accidentally eval'ing the variable
contents.

Closes #3039

Closes #3458

/kind documentation

# Submitter Checklist

review them:

# Release Notes

```release-note
Add a doc describing how to safely use variables in script blocks of tasks without eval'ing their contents
```

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -26,6 +26,7 @@ weight: 1
     - [Substituting `Array` parameters](#substituting-array-parameters)
     - [Substituting `Workspace` paths](#substituting-workspace-paths)
     - [Substituting `Volume` names and types](#substituting-volume-names-and-types)
+    - [Substituting in `Script` blocks](#substituting-in-script-blocks)
 - [Code examples](#code-examples)
   - [Building and pushing a Docker image](#building-and-pushing-a-docker-image)
   - [Mounting multiple `Volumes`](#mounting-multiple-volumes)
@@ -581,6 +582,14 @@ The `description` field is an optional field that allows you to add an informati
 
 ### Using variable substitution
 
+Tekton provides variables to inject values into the contents of certain fields.
+The values you can inject come from a range of sources including other fields
+in the Task, context-sensitive information that Tekton provides, and runtime
+information received from a TaskRun.
+
+The mechanism of variable substitution is quite simple - string replacement is
+performed by the Tekton Controller when a TaskRun is executed.
+
 `Tasks` allow you to substitute variable names for the following entities:
 
 - [Parameters and resources](#substituting-parameters-and-resources)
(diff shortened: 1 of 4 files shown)


sbwsg:
Closes https://github.com/tektoncd/pipeline/issues/3039

sbwsg:
Closes https://github.com/tektoncd/pipeline/issues/3458

jlpettersson (commented):
Thanks,
this looks good.

/lgtm

docs/tasks.md
```diff
@@ -581,14 +582,22 @@ The `description` field is an optional field that allows you to add an informati
...
 ### Using variable substitution
 
+Tekton provides variables to inject values into the contents of certain fields.
+The values you can inject come from a range of sources including other fields
+in the Task or Pipeline, context-sensitive information that Tekton provides,
+and runtime information received from a TaskRun or PipelineRun.
+
+The mechanism of variable substitution is quite simple - string replacement is
+performed by the Tekton Controller when a TaskRun or PipelineRun is executed.
+
```
jerop, line 592: it would be helpful to have this in [pipelines docs](https://github.com/tektoncd/community/pull/293) because it's also just referencing the `variables.md` as well, what do you think?

sbwsg: Yeah, that makes sense, I'll update this PR to add details to the pipelines doc too.

jerop (approved):
thanks @sbwsg! 
/meow

Issue #3590, opened 2020-12-03
Support for params with `.`s


mattmoor:
# Expected Behavior

A way to use params with `.`s

# Actual Behavior

No way to reference params with `.`s

# Steps to Reproduce the Problem

```yaml
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: hello
spec:
  params:
    - name: dev.mattmoor.foo
  steps:
    - name: echo
      image: ubuntu
      command: ["/bin/bash", "-c"]
      args:
        - |
          echo $(params."dev.mattmoor.foo")
```

# Additional Info

I tested this against `github.com/mattmoor/mink`, so it may be a latent downstream bug, but I've got the validation wired up, so I was surprised this was accepted if it doesn't work.

mattmoor:
_Another take on this is "validation hole!", but let me explain **why** I want domain-style parameter names..._

more to come.

mattmoor:
Another examples of this, which seems especially relevant to syntax precedence, is containerd plugins:
```toml
  [plugins."io.containerd.grpc.v1.cri".containerd]
    disable_snapshot_annotations = true
```

If we were to follow this precedent, then it would suggest allowing: `$(params."this.is.one.name")` syntax.

However, this precedent does double-duty, since it uses reverse-DNS (like Java imports, and CloudEvent types).

I'll take a crack as allowing the `param."foo"` syntax to see how far I can get.

/assign

mattmoor:
I noticed this in my triage of `params.%s` and don't see much context for *why*:
```go
// convertParamTemplates replaces all instances of $(params.x) in the container to $(inputs.params.x) for each param name.
func convertParamTemplates(step *v1beta1.Step, params []v1beta1.ParamSpec) {
	replacements := make(map[string]string)
	for _, p := range params {
		replacements[fmt.Sprintf("params.%s", p.Name)] = fmt.Sprintf("$(inputs.params.%s)", p.Name)
		v1beta1.ApplyStepReplacements(step, replacements, map[string][]string{})
	}

	v1beta1.ApplyStepReplacements(step, replacements, map[string][]string{})
}
```

@vdemeester based on your `FIXME` comments, this seems like it's moving things the wrong direction.  Is this still needed?

I'm going to ignore it for now.

sbwsg:
[...]

PR #3637, opened 2020-12-14
Support quoting parameter names

mattmoor:
# Changes

/kind feature

This change plumbs through support for quoting user-provided names during variable resolution as proposed [here](https://github.com/tektoncd/pipeline/issues/3590#issuecomment-744576896).

In addition to `$(params.foo)` you can now use `$(params."foo")`, which is useful for disambiguation when the name contains `.` characters (e.g. `$(params."dev.tekton.foo")`).

I have updated unit tests, and added a minor update to docs, but if there are more tests or docs that should be updated, please send me pointers.

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Support quoting user-provided names during variable substitution, e.g. $(params."foo") or $(results."bar".path)
```

### DIFF
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -7,7 +7,11 @@ weight: 15
 # Variable Substitutions Supported by `Tasks` and `Pipelines`
 
 This page documents the variable substitutions supported by `Tasks` and `Pipelines`.
+
 **Note:** Tekton does not escape the contents of variables. Task authors are responsible for properly escaping a variable's value according to the shell, image or scripting language that the variable will be used in.
+
+**Note:** Variables of the form `<foo>` may be quoted to eliminate ambiguity when the name contains `.` characters, e.g. `$(params."dev.tekton.foo")`.
+
 
 ## Variables available in a `Pipeline`
 
--- a/pkg/reconciler/pipelinerun/resources/apply.go
+++ b/pkg/reconciler/pipelinerun/resources/apply.go
@@ -33,6 +33,7 @@ func ApplyParameters(p *v1beta1.PipelineSpec, pr *v1beta1.PipelineRun) *v1beta1.
 
 	patterns := []string{
 		"params.%s",
+		"params.%q",
 	}
 
 	// Set all the default stringReplacements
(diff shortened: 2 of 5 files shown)


mattmoor:
cc @vdemeester @afrittoli @ImJasonH

mattmoor:
Something strange is going on, both of those passed.  I think in the past this correlated with Prow being overloaded?

sbwsg:
This feature addition feels a tad under-specified to me.  Is this intended as a proposal or do you want us to proceed on merging this as-is?

Here's what I'm gathering from reading the code: This change affects params, workspaces, results and resource variables.  From an initial scan it doesn't look like we've added this support to `context` variables, for example.  Only certain portions of variables are quotable (e.g. `"resources.%s.%q.%s"` is supported but not `"resources.%q.%s.%s"`; `"workspaces.%q.%s"` is supported but not `"workspaces.%q.%q"`).  The doc atm doesn't capture these ambiguities or explain the reasoning.  Some `apply` tests have been updated to remove the `inputs.` prefix, which might mean we're dropping a bit of coverage of backwards-compatible behaviour - unclear to me atm.

I'd like to throw another design in to the ring: support bracket notation instead, e.g. `params["my.param.with.periods"]` which could in turn be taken a step further to array param subscripts ala `params.myparam[1]` or `params["my.array.param.with.periods"]["1"]` or something similar.  This isn't so much to dissuade you from running with your design, but I reckon there's enough conceptual room here to at least describe some alternatives in a PR, say why we discount them, and get some conversation going around it.

TL;DR I'm wary of lgtm'ing this without a bit of discussion first.  It might even be TEP-worthy?  Curious what other folks think @tektoncd/core-maintainers

mattmoor:
> "resources.%s.%q.%s" is supported but not "resources.%q.%s.%s"

The intention was to support quote all user-sourced names, there are a few of these (some I'm adding) where the `%s` is used for built-in strings.  In the example above, my inference was that the first `%s` was generally used for "inputs" or "outputs", which is perhaps incorrect.

I am happy to discuss this further, but nobody was discussing anything in the linked issue.  I'm glad this triggered some 😉 

---

On TEPs, perhaps it is worth framing this as an incidental aspect of the broader idea I think we should pursue, which I have roughly sketched here: https://github.com/tektoncd/pipeline/issues/3590#issuecomment-738051808

Ultimately, without that, I have effectively no energy to pursue this further.

PR #3646, opened 2020-12-23
add sparse checkout to git

a-rothwell:
# Changes
Hi, 

I have added functionality to support sparse-checkout for git

This works by enabling the feature in the git config and then updating the sparse-checkout file before fetch.

The feature request is [here](https://github.com/tektoncd/catalog/issues/488)

I have also added a test to check that .git/info/sparse-checkout is updated as expected

Thanks!

/kind feature

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

### DIFF
--- a/cmd/git-init/main.go
+++ b/cmd/git-init/main.go
@@ -39,6 +39,7 @@ func init() {
 	flag.BoolVar(&fetchSpec.Submodules, "submodules", true, "Initialize and fetch Git submodules")
 	flag.UintVar(&fetchSpec.Depth, "depth", 1, "Perform a shallow clone to this depth")
 	flag.StringVar(&terminationMessagePath, "terminationMessagePath", "/tekton/termination", "Location of file containing termination message")
+	flag.StringVar(&fetchSpec.SparseCheckoutDirectories, "sparseCheckoutDirectories", "", "String of directory patterns separated by a comma")
 }
 
 func main() {
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -59,16 +59,17 @@ func run(logger *zap.SugaredLogger, dir string, args ...string) (string, error) 
 
 // FetchSpec describes how to initialize and fetch from a Git repository.
 type FetchSpec struct {
-	URL        string
-	Revision   string
-	Refspec    string
-	Path       string
-	Depth      uint
-	Submodules bool
-	SSLVerify  bool
-	HTTPProxy  string
-	HTTPSProxy string
-	NOProxy    string
+	URL                       string
+	Revision                  string
+	Refspec                   string
+	Path                      string
+	Depth                     uint
+	Submodules                bool
+	SSLVerify                 bool
+	HTTPProxy                 string
+	HTTPSProxy                string
+	NOProxy                   string
+	SparseCheckoutDirectories string
 }
 
 // Fetch fetches the specified git repository at the revision into path, using the refspec to fetch if provided.
@@ -86,6 +87,9 @@ func Fetch(logger *zap.SugaredLogger, spec FetchSpec) error {
 			return fmt.Errorf("failed to change directory with path %s; err: %w", spec.Path, err)
 		}
 	} else if _, err := run(logger, "", "init"); err != nil {
+		return err
+	}
+	if err := configSparseCheckout(logger, spec); err != nil {
 		return err
 	}
 	trimmedURL := strings.TrimSpace(spec.URL)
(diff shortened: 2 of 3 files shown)


pkg/git/git.go
```diff
@@ -281,3 +286,35 @@ func ValidateGitSSHURLFormat(url string) bool {
...
+
+		dirPatterns := strings.Split(spec.SparseCheckoutDirectories, spec.SparseCheckoutDelimiter)
+
+		cwd, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		file, err := os.OpenFile(filepath.Join(cwd, ".git/info/sparse-checkout"), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			logger.Errorf("failed to open sparse-checkout file; err: %v", err)
```
sbwsg, line 304: nit: no need for `err` in the logged message here.  Suggest `"failed to open sparse-checkout file: %v"`.  Similar comment for the other messages passed to `logger.Errorf` in this func.

pkg/git/git.go
```diff
@@ -281,3 +286,35 @@ func ValidateGitSSHURLFormat(url string) bool {
...
+	if spec.SparseCheckoutDirectories != "" {
+		if _, err := run(logger, "", "config", "core.sparsecheckout", "true"); err != nil {
+			return err
+		}
+
+		dirPatterns := strings.Split(spec.SparseCheckoutDirectories, spec.SparseCheckoutDelimiter)
+
+		cwd, err := os.Getwd()
+		if err != nil {
+			return err
```
sbwsg, line 300: Any reason to use `logger.Errorf` for the other errors in this func except for this one?  Would it be worth logging this in a similar way?

cmd/git-init/main.go
```diff
@@ -39,6 +39,8 @@ func init() {
 	flag.BoolVar(&fetchSpec.Submodules, "submodules", true, "Initialize and fetch Git submodules")
 	flag.UintVar(&fetchSpec.Depth, "depth", 1, "Perform a shallow clone to this depth")
 	flag.StringVar(&terminationMessagePath, "terminationMessagePath", "/tekton/termination", "Location of file containing termination message")
+	flag.StringVar(&fetchSpec.SparseCheckoutDirectories, "sparseCheckoutDirectories", "", "String of directory patterns separated by a delimiter")
+	flag.StringVar(&fetchSpec.SparseCheckoutDelimiter, "sparseCheckoutDelimiter", ";", "Delimiter used to separate sparse checkout directories")
```
chmouel, line 43: I don't think we need that level of granularity as a flag

Moreover the delimiter would cause some issues if the user doesn't quote the ";", it will be interepreted as another command.

What about remove the flag and setting the delimiter as a standard comma "," ?

chmouel (changes requested):

pkg/git/git_test.go
```diff
@@ -206,6 +228,27 @@ func TestFetch(t *testing.T) {
...
+				defer sparseFile.Close()
+
+				var sparsePatterns []string
+
+				scanner := bufio.NewScanner(sparseFile)
+				for scanner.Scan() {
+					sparsePatterns = append(sparsePatterns, scanner.Text())
+				}
+
+				if !reflect.DeepEqual(dirPatterns, sparsePatterns) {
```
chmouel, line 247: what about using cmp.Diff, which is a library used thought-out all the tests in pipeline ? correct me if I am wrong but it does the same thing ?

a-rothwell: cmp.Diff will work here. I will make that change!

cmd/git-init/main.go
a-rothwell: That is true, I understand how semicolons could cause issues for users.
Removing the delimiter flag and using a comma will make it simpler.

pkg/git/git.go
a-rothwell: I think it would be worth adding to maintain consistency.

sbwsg (approved):

sbwsg:
@a-rothwell thanks for making those changes.  You'll need to squash your commits into 1 prior to merge.

chmouel:
/lgtm

Thanks 🙇🏽

chmouel:
it woudl be great if we can the git-clone task updated to support this feature too 🙏🏽 @a-rothwell

a-rothwell:
Thank you both for helping me with this!
I will update the task in the catalog

PR #3653, opened 2021-01-05
Skip the dropNetworking test if it doesn't have the correct privileges.

dlorenc:
# Changes

I decided to check for the required privileges directly, but just running it
on something that should always work and looking for an error. I considered
directly testing the capabilities required, but that couples the test to the
implementation and is probably not going to be perfect anyway.

/kind failing-test

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
NONE
```

### DIFF
--- a/cmd/entrypoint/namespaces_test.go
+++ b/cmd/entrypoint/namespaces_test.go
@@ -15,6 +15,18 @@ import (
 // sandboxed environment, the test could pass even if the dropNetworking
 // function did nothing.
 func TestDropNetworking(t *testing.T) {
+	// First make sure we can run the dropNetworking command.
+	// Some older kernels require special configurations to run this.
+	// I haven't been able to come up with an exhaustive list of what is needed,
+	// but it includes things like CAP_SYS_ADMIN, kernel.unprivileged_userns_clone=1
+	// and maybe others.
+	// For the sake of this test just check it first.
+	testCmd := exec.Command("true")
+	dropNetworking(testCmd)
+	if _, err := testCmd.CombinedOutput(); err != nil {
+		t.Skipf("skipping test as required namespace features are not available: %v", err)
+	}
+
 	cmd := exec.Command("curl", "google.com")
 	dropNetworking(cmd)
 	b, err := cmd.CombinedOutput()


dlorenc:
cc @gabemontero

gabemontero:
the change works in my env @dlorenc thanks

vdemeester (approved):

cmd/entrypoint/namespaces_test.go
```diff
@@ -15,6 +15,18 @@ import (
...
 // function did nothing.
 func TestDropNetworking(t *testing.T) {
+	// First make sure we can run the dropNetworking command.
+	// Some older kernels require special configurations to run this.
+	// I haven't been able to come up with an exhaustive list of what is needed,
+	// but it includes things like CAP_SYS_ADMIN, kernel.unprivileged_userns_clone=1
+	// and maybe others.
+	// For the sake of this test just check it first.
+	testCmd := exec.Command("true")
+	dropNetworking(testCmd)
```
chmouel, line 25: If there is an issue with dropNetworking() would this always be Skip then  and make the test not working ? Maybe filtering on the type of error ?

cmd/entrypoint/namespaces_test.go
```diff
@@ -15,6 +15,18 @@ import (
...
+	if _, err := testCmd.CombinedOutput(); err != nil {
+		t.Skip()
```
bobcatfish, line 27: could you include a message indicating why the test is being skipped (in addition to the comment which is great btw)? (looks like skip can take a string https://golang.org/pkg/testing/#hdr-Skipping)

dlorenc: Done! Thanks, I forgot you could put a string in there.

cmd/entrypoint/namespaces_test.go
```diff
@@ -15,6 +15,18 @@ import (
...
+	// First make sure we can run the dropNetworking command.
+	// Some older kernels require special configurations to run this.
+	// I haven't been able to come up with an exhaustive list of what is needed,
+	// but it includes things like CAP_SYS_ADMIN, kernel.unprivileged_userns_clone=1
+	// and maybe others.
+	// For the sake of this test just check it first.
+	testCmd := exec.Command("true")
+	dropNetworking(testCmd)
+	if _, err := testCmd.CombinedOutput(); err != nil {
+		t.Skipf("skipping test as required namespace features are not available: %v", err)
```
ImJasonH, line 27: Would it make sense to also include the combined stdout/stderr here too?

dlorenc: Nothing shows up in the output when I get an error. I'm not sure if that's always the case, though:

```
    namespaces_test.go:27: skipping test as required namespace features are not available: fork/exec /bin/true: operation not permitted
```

cmd/entrypoint/namespaces_test.go
vdemeester: I think we could even filter on the capabilities available and/or on the user uid that executes the test. As is, it's a bit weird because if the function fails for another reason than not having enough capabilities to remove capabilities (oh the irony 😛 ), it's gonna skip instead of fail.

PR #3679, opened 2021-01-12
Fix bug where step status ordering did not match step container ordering

### LINKED ISSUE #3677: image-digest-exporter step appears out of order with its position in pod
# Expected Behavior

The `image-digest-exporter` Step should appear after all of the user's Steps in a TaskRun's `status.steps` list.

# Actual Behavior

In https://dashboard.dogfooding.tekton.dev/#/namespaces/default/pipelineruns/pipeline-release-nightly-zf2lz?pipelineTask=publish-images&step=image-digest-exporter-vcqhc we see the `image-digest-exporter` sorted out of order, appearing before user-created Steps `copy-to-latest-bucket`, `tag-images` and `run-ko`.

In that example the `tag-images` Step was the first to fail but due to misordering of the steps' statuses it appears that the `image-digest-exporter` was the original failure.  This confuses several subsequent things:

1. The TaskRun gets the wrong Succeeded condition, claiming that the `image-digest-exporter` was the original failure.
2. The dashboard similarly shows `image-digest-exporter` as the source of failure:

    Screen Shot 2021-01-12 at 9 49 31 AM

3. The dashboard also shows that the actual failing Step (`tag-images` in this case) was "Not Run", which is incorrect.

# Additional Info

- Resources Involved

I've captured both the TaskRun's status field from the dashboard as well as the full Pod yaml for that TaskRun [in this gist](https://gist.github.com/sbwsg/1c41dc6d0145b39357c16e570b32a274).

- Kubernetes version:

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.14-gke.1600", GitCommit:"7c407f5cc8632f9af5a2657f220963aa7f1c46e7", GitTreeState:"clean", BuildDate:"2020-12-07T09:22:27Z", GoVersion:"go1.13.15b4", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

```
v0.18.1
```

sbwsg:
# Changes

Fixes #3677 

Pipelines sorts container statuses so they're in the same order as steps. The sorting is not currently working right. See #3677 for an example of this where `image-digest-exporter` was sorted before some user steps. This confuses a lot of things - the `Succeeded` status records the wrong step as original source of failure, the Dashboard shows the wrong step as the source of failure in a TaskRun, and also shows some Steps as "Not Run" even if they did.

This commit replaces the custom sort code with a much simpler version: just walk the containers and place the statuses in the same order.

# Submitter Checklist

review them:

# Release Notes

```release-note
Fixed bug where sorted order of step statuses did not match order of step containers.
```

### DIFF
--- a/pkg/pod/status.go
+++ b/pkg/pod/status.go
@@ -19,7 +19,6 @@ package pod
 import (
 	"encoding/json"
 	"fmt"
-	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


sbwsg:
Running unit tests locally I don't see any errors.  I'm going to try rerunning that.

sbwsg:
The build tests passed and the integration tests are "still running" according to their prow pages.  There's something really funky going on with the github status reporting here.  Re-running one-at-a-time to see what happens.

vdemeester (approved):

pritidesai:
wow a huge simplification, thanks @sbwsg 👍 

Just by creating a map of container names and its status, then iterating over the desired order of step containers and changing the order of container statuses based on that order. 

This might leave us with a list of container statuses with duplicate containers and/or some missing if the for loop is interrupted which can only be possible when a process is killed. And should not be concern here.

/lgtm

Issue #3667, opened 2021-01-08
PipelineRun's completed status can be overwritten by failed results lookup


sbwsg:
# Expected Behavior

When a PipelineRun is completed the final status it records should be considered immutable and not change after that point.

# Actual Behavior

When a Pipeline includes Pipeline Results there is a slim chance that its PipelineRuns' final status can be overwritten.  This is because the PipelineRun reconciler does not process Pipeline Results until after the Run is in a completed state.  The code for this is here: https://github.com/tektoncd/pipeline/blob/f121db996dbc1867d8f6616c887c1374a5770db2/pkg/reconciler/pipelinerun/pipelinerun.go#L326-L328

That linked function, `updatePipelineResults()` is called from https://github.com/tektoncd/pipeline/blob/f121db996dbc1867d8f6616c887c1374a5770db2/pkg/reconciler/pipelinerun/pipelinerun.go#L179 which is inside a block guarded by `if pr.IsDone()`. Despite this the `updatePipelineResults` func can change the state of the PipelineRun to indicate a failure. This therefore has the opportunity to modify an existing completed status and overwrite information pertinent to the user.

This issue was created in response to feedback on PR 3472: https://github.com/tektoncd/pipeline/pull/3472/files#r552632569

sbwsg:
The code has moved around a bit since #3472 has been merged but the issue still affects the reconciler here: https://github.com/tektoncd/pipeline/blob/6e6b6c8b8fb80195177f3f955ff84ab2ee320bf1/pkg/reconciler/pipelinerun/pipelinerun.go#L179-L182

PR #3684, opened 2021-01-13
Move resolution of Pipeline Results to end of completed/successful reconcile

sbwsg:
Fixes https://github.com/tektoncd/pipeline/issues/3667

# Changes

We currently resolve Pipeline Results on the first reconcile _after_ a PipelineRun is completed.
This is bad because the resolution can fail. When that happens we overwrite the Succeeded condition
of the PipelineRun which means we're losing the final status info. It also means there's a
bit of time where the PipelineRun is in one completed state only to have it change to a different
completed state, potentially confusing any services watching PipelineRun conditions.

This commit moves the resolution of Pipeline Results to the end of a reconcile when the PipelineRun
is put into a Succeeded/True condition. The benefit of this is that we don't have to look up
the Pipeline data to compute the results since it's already available to us. This removes the
part of result resolution that could fail and means that we don't have to clobber the PipelineRun's
final state.

/kind bug

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -175,15 +175,6 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 		// and may not have had all of the assumed default specified.
 		pr.SetDefaults(contexts.WithUpgradeViaDefaulting(ctx))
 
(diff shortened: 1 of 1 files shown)


sbwsg:
Unit tests have succeeded but looks like it hasn't been picked up by prow.  Trying again.

vdemeester (approved):

pritidesai:
thanks @sbwsg for the fix 👍 
/lgtm

PR #3685, opened 2021-01-13
Ignore generated openapi code when running golangci-lint

sbwsg:
# Changes

Currently a dev using `make golangci-lint` will get a bunch of errors in openapi_generated.go.
This file isn't something we're supposed to make updates in, it's generated, so I don't think
reporting those errors makes a lot of sense.

This commit skips the openapi_generated.go file when running `make golangci-lint`.

/kind misc

# Submitter Checklist

review them:

# Release Notes

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -28,6 +28,7 @@ run:
   - e2e
   skip-files:
   - .*/zz_generated.deepcopy.go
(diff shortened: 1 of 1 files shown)


vdemeester (approved):
/meow

pritidesai:
thanks @sbwsg 
/lgtm

Issue #2343, opened 2020-04-07
Improve UX of getting credentials into Tasks


sbwsg:
# Expected Behavior

Reading the docs a user can quickly and easily discover how to:
1. declare credentials their Tasks accept
2. attach credentials to TaskRuns or PipelineRuns

Once a user has read the docs, there's a low hurdle to attaching credentials to their CI workloads.

# Actual Behavior

### Problems with Documentation and User Flow

Our auth.md doc describes a system for getting credentials out of Secrets and into Steps.

The process for users is documented as follows:
1. add the credential information to a Secret in k8s
2. add some annotations to that Secret to ensure it's recognized by Tekton
3. add the Secret to a Service Account
4. add the Service Account's name to your TaskRun.

The number of steps isn't too bad but there's a lot of room for confusion:
* the annotations and type information on Secrets has to be entered pretty precisely
* errors don't show up until a Task is running
* validation of the credentials is non-existent
* there's a disconnect between the Secret and the Run: the user adds the _service account name_ to the TaskRun and then needs to hope the Secrets are picked up and parsed correctly with no other feedback from Tekton.

### Problems with Implementation

On the controller-side Pod startup looks as follows with credentials:
1. a creds-init Image is injected as an init-container of the TaskRun Pod
2. creds-init receives Secrets as Volumes. Secrets must include specific type and annotations.
3. creds-init copies the credentials out of those Volumes into /tekton/home or /tekton/creds
4. entrypoint, before each Step, copies the credentials out of /tekton/creds into the Step Container's $HOME

There are a couple of challenges with this:
- Secrets get mounted by Kubernetes as in-memory filesystems but creds-init copies those credentials to disk (/tekton/home and /tekton/creds are `emptyDir` volumes). This diminishes a security feature built into the platform.

- Our guidance on credentials for Task Authors and Users is not very strong, despite how long auth.md is. It's not easy to figure out precisely what you should implement or expect, regardless if you're a Task Author or a Task User.

- Every Step gets access to all credentials. Auth.md states at the end: "For sensitive credentials that should not be made available to some steps, do not use the mechanisms outlined here. Instead, the user should declare an explicit Volume from the Secret and manually VolumeMount it into the Step."  So we're kind of acknowledging here that there are problems with the existing flow and Tekton's treatment of creds.

# Requirements and Constraints

### Requirements
* Improve ease-of-use over the existing creds-init system.
* Using authenticated resources like private repos should be practically as simple as using public.
[...]

PR #3700, opened 2021-01-20
Add Step and Sidecar Workspaces feature

sbwsg:
# Changes

TEP: https://github.com/tektoncd/community/blob/master/teps/0029-step-workspaces.md
Fixes https://github.com/tektoncd/pipeline/issues/2343

Access to Workspaces can now be limited to specific Steps and Sidecars in a Task.
This allows Task authors to isolate sensitive data to specific images, reducing
exposure of assets like credentials.

As a side-effect of this change Workspaces are now mounted to Sidecars by
default, just as they are to Steps.

# Submitter Checklist

review them:

# Release Notes

```release-note
- New alpha feature added: "Step and Sidecar Workspaces"
    - Requires "enable-api-fields: alpha" feature gate.
    - Sidecars in a Task now receive Workspaces, just like Steps do.
    - Steps and Sidecars can now specify which Workspaces they want access to. This isolates the workspace to only those Steps or Sidecars that need them, allowing Task authors to limit exposure of sensitive data like credentials.
    - The default behaviour for workspaces in Steps remains the same - all Steps get access to them.
```


sbwsg:
Build tests are failing with following message:

```
/home/prow/go/src/github.com/tektoncd/pipeline is out of date. Please run hack/update-codegen.sh
```

But `update-codegen.sh` doesn't create any new code when I run it.  This error is actually appearing because `update-openapigen.sh` needs to be run.  I'm going to update that error message as part of this PR.

bobcatfish (commented):
Started with some initial comments on docs and examples, all super minor so far! Will return to review more ASAP

docs/workspaces.md
```diff
@@ -138,38 +145,47 @@ spec:
...
+  - name: build-source
+    image: third-party-source-builder:latest # This image doesn't get access to ssh-credentials.
+  - name: lint-source
+    image: third-party-source-linter:latest # This image doesn't get access to ssh-credentials.
 ```
 
-**Note:** `Sidecars` _must_ explicitly opt-in to receiving the `Workspace` volume. Injected `Sidecars` from
-non-Tekton sources will not receive access to `Workspaces`.
+It can potentially be useful to mount `Workspaces` to different locations on a per-`Step` or
+per-`Sidecar` basis and this is also supported:
```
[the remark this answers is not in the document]
sbwsg: Yeah, exactly.  The moment a `workspace` entry is present on a step/sidecar it's considered isolated.

docs/workspaces.md
```diff
@@ -88,6 +88,13 @@ many uses:
 parameters used.
 - An optional build cache may be provided to speed up compile times.
 
+### Isolated `Workspaces`
+
+Certain kinds of data are more sensitive than others. To reduce exposure of sensitive data Task
+authors can isolate `Workspaces` to only those `Steps` and `Sidecars` that require access to
+them. The primary use-case for this is credentials but it can apply to any data that should have
+its access strictly limited to only specific container images.
```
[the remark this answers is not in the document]
sbwsg: Thanks for reviewing, added the link.

examples/v1beta1/taskruns/workspace-isolation.yaml
```diff
@@ -0,0 +1,50 @@
...
+#
+# The Step writes a file to the Workspace and then waits for the Sidecar
+# to respond. The Sidecar sees the Step's file and writes its response.
+# The Step sees the response and exits.
+kind: TaskRun
+apiVersion: tekton.dev/v1beta1
+metadata:
+  generateName: workspace-isolation-
+spec:
+  timeout: 2m0s
```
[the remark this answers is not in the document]
sbwsg: Updated with comments in both loops mentioning the use of `timeout` to prevent looping forever.

hack/verify-codegen.sh
```diff
@@ -52,6 +52,6 @@ if [[ $ret -eq 0 ]]
 then
   echo "${REPO_ROOT_DIR} up to date."
 else
-  echo "${REPO_ROOT_DIR} is out of date. Please run hack/update-codegen.sh"
+  echo "${REPO_ROOT_DIR} is out of date. Please run hack/update-codegen.sh and hack/update-openapigen.sh"
```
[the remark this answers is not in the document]
sbwsg: Good catch, I've updated DEVELOPMENT.md.

pkg/workspace/apply_test.go
```diff
@@ -522,3 +722,23 @@ func TestApply(t *testing.T) {
...
+	}
+	bindings := []v1beta1.WorkspaceBinding{{
+		Name: "source",
+		PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+			ClaimName: "testpvc",
+		},
+	}}
+	vols := workspace.CreateVolumes(bindings)
+	_, err := workspace.Apply(ts, bindings, vols)
+	if err == nil || !strings.Contains(err.Error(), "did not exist in declared workspaces") {
```
bobcatfish, line 741: nit: i find that asserting against error messages (or log messages) is a bit brittle to maintain over time - if it's important to assert that the correct error is returned, i usually go to one of 2 alternatives:

* return different types of errors  (https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully is a very interesting post on error handling in Go and iirc doenst like that approach either)
* treat the different error types as a sign that one function is doing too much, i.e. if it can fail in 2 different ways and its important to distinguish between both, maybe you're better off with two separate functions (it does SEEM in this case like maybe Apply is also doing validation)

pkg/workspace/apply_test.go
```diff
@@ -509,6 +510,205 @@ func TestApply(t *testing.T) {
...
+					VolumeMounts: []corev1.VolumeMount{{
+						Name:      "ws-sn8zf",
+						MountPath: "/workspace/source",
+					}},
+				},
+			}},
+			Workspaces: []v1beta1.WorkspaceDeclaration{{
+				Name: "source",
+			}},
+		},
```
bobcatfish, line 711: nice, clear test cases!

pkg/workspace/apply.go
```diff
@@ -121,3 +140,45 @@ func Apply(ts v1beta1.TaskSpec, wb []v1beta1.WorkspaceBinding, v map[string]core
...
+			}
+		}
+	}
+	for i := range ts.Sidecars {
+		sidecar := &ts.Sidecars[i]
+		for _, workspaceUsage := range sidecar.Workspaces {
+			if workspaceUsage.Name == workspaceName {
+				vm := volumeMount
+				if workspaceUsage.MountPath != "" {
+					vm.MountPath = workspaceUsage.MountPath
```
bobcatfish, line 177: it seems like this is the code that handles the mountPath overriding logic, but i dont see it being explicitly included in the tests cases below? (apologies if i just missed it) i could see there being an extra case or two for this?

bobcatfish: i think the coverage report might be backing this up as well:

image

pkg/reconciler/taskrun/resources/apply_test.go
```diff
@@ -786,6 +786,56 @@ func TestApplyWorkspaces(t *testing.T) {
...
+		}},
+		want: &v1beta1.TaskSpec{Steps: []v1beta1.Step{{
+			Script: `echo "/workspace/ws"`,
+		}}, Sidecars: []v1beta1.Sidecar{{
+			Script: `echo "/bar"`,
+			Workspaces: []v1beta1.WorkspaceUsage{{
+				Name:      "ws",
+				MountPath: "/bar",
+			}},
+		}}},
```
bobcatfish, line 838: nice!!! 😎

pkg/reconciler/taskrun/resources/apply.go
```diff
@@ -144,6 +144,40 @@ func ApplyWorkspaces(spec *v1beta1.TaskSpec, declarations []v1beta1.WorkspaceDec
...
+	// Replace instances of the workspace path variable that are overridden per-Sidecar
+	for i := range spec.Sidecars {
+		sidecar := &spec.Sidecars[i]
+		for _, usage := range sidecar.Workspaces {
+			if usage.Name == declaration.Name && usage.MountPath != "" {
+				stringReplacements[variable] = usage.MountPath
+				v1beta1.ApplySidecarReplacements(sidecar, stringReplacements, emptyArrayReplacements)
+			}
+		}
+	}
```
bobcatfish, line 174: just nitpicking and i dont have a clear alternative suggestion, but it feels like some of this logic belongs to how "workspaces" work and i'd almost expect to see it in the workspaces package

(i.e. which path to use feels like it's more about how workspaces work, and less about how string substitution works)

bobcatfish (commented):
I think pretty much all my comments are just idle thoughts or unimportant nitpicks! The only at all "significant" comment I have is a think there might be a bit of missing test coverage on the mount path overriding.

Looks great!

/approve

i wonder if we want to explicitly do anything about how this impacts pipelineresources? since this was implemented before using steptemplates, removing the volumes from the step template would also mean removing it from pipelineresources, but right now we don't cover it with any tests or mention it in docs - i think its probably better this way; its hard to imagine any pipelineresources that would be impacted, probably no one even knew that was happening, and lastly we probably want to discourage using pipelineresources anyway, but i wanted to bring it up and see what you think

pkg/apis/pipeline/v1beta1/workspace_types.go
```diff
@@ -111,3 +111,12 @@ type WorkspacePipelineTaskBinding struct {
 	// +optional
 	SubPath string `json:"subPath,omitempty"`
 }
+
+// WorkspaceUsage is used by a Step or Sidecar to declare that it wants isolated access
+// to a Workspace defined in a Task.
+type WorkspaceUsage struct {
+	// Name is the name of the workspace this Step or Sidecar wants access to.
+	Name string `json:"name"`
+	// MountPath is the path that the workspace should be mounted to inside the Step or Sidecar.
```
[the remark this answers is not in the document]
sbwsg: Updated comment.

docs/workspaces.md
```diff
@@ -137,38 +150,47 @@ spec:
...
+It can potentially be useful to mount `Workspaces` to different locations on a per-`Step` or
+per-`Sidecar` basis and this is also supported:
+
+```yaml
+  workspaces:
+  - name: source-code
+  steps:
+  - workspaces:
+    - name: source-code
+      mountPath: /source
```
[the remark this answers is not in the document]
sbwsg: Updated to the more complex example.

bobcatfish:
image

yaaaasssss

image

pkg/workspace/apply_test.go
sbwsg: Removed string error message check.

pkg/workspace/apply.go
sbwsg: Unit test added to cover custom mountPath in sidecar and step.

pkg/reconciler/taskrun/resources/apply.go
sbwsg: I'm not sure there's a good way to disentangle the choice of mountpath for each case (step / sidecar / shared) from the string substitution logic.  Particularly since the variables being replaced are order-dependent (we have to replace the isolated workspace mountpaths before those that are shared). I can't see an approach that doesn't either leak some workspace logic into the string substitution package or leak some string substitution logic into the workspace package.

We can always revisit?  There's nothing user-facing here that will prevent a refactor if we figure out a better separation.

sbwsg:
Thanks for reviewing @bobcatfish !

docs/workspaces.md
```diff
@@ -137,38 +150,53 @@ spec:
...
+      mountPath: /foo # overrides mountPath
+  - name: edit-files-2
+    workspaces:
+    - name: foo # no mountPath specified so will use /workspaces/ws
+  sidecars:
+  - name: watch-files-on-workspace
+    workspaces:
+    - name: foo
+      mountPath: /files # overrides mountPath
+```
```
pritidesai, line 199: the workspace `foo` in each step is referring to the same workspace `ws` specified under `workspaces`? should they be named same as defined under `workspaces`?

examples/v1beta1/taskruns/workspace-isolation.yaml
```diff
@@ -0,0 +1,52 @@
...
+  workspaces:
+  - name: signals
+    emptyDir: {}
+  taskSpec:
+    workspaces:
+    - name: signals
+    steps:
+    - image: alpine:3.12.0
```
pritidesai, line 20: NIT: step names are missing

docs/workspaces.md
pritidesai: What does "override" mean here? the step `edit-files-1` is executed with the mount path `/foo` and `edit-files-2` executed with the default mountPath `/ws`. But the files created in `edit-files-1` under `/foo` Will be visible in `edit-files-2` under `/ws`. Its just the path name which differs. Am I getting it right?

pritidesai: I ran a pipeline with creating files (`/foo/sample1` and `/foo/sample2`) in `edit-files-1`. Both files were accessible as `/ws/sample1` and `/ws/sample2` in `edit-files-2`.

I don't see any issues with this behavior but may be a one liner explanation might help in the doc. Not a blocker though. 🙏

afrittoli:
Thank you for this, it's a great feature to add. I skimmed through the code and it looks good (thanks for nice docs 🙏 ) - the only concern I have is the lack of a feature flag. Since this adds to a beta API, adding it without a flag means that the new API field becomes beta the moment it's released.

I'm aware that we don't have a clear policy right now, I'll try to progress on https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md so that we may have better guidance for the future. 

I agree with @bobcatfish proposal to have a single API flag for (most) API changes would make sense, and I know @vdemeester expressed positive feedback to that idea, so perhaps we should add the flag along with documentation so that people may start using it. 

@sbwsg @bobcatfish @vdemeester we're almost at release time, I see two alternatives:
1. complete the work on the TEP, implement the changes associated and then merge this PR, hopefully in the next release
2. implement an API feature flag now, rebase this one on top and they might both be merged in time for this release

docs/workspaces.md
sbwsg: They should be `name`d "ws".  That's a bug, well spotted! will fix.  Also I will make this a validation error if it isn't already.  "foo" isn't a valid workspace name here.

sbwsg:
The feature can wait until a single shared feature flag or gate is implemented.  I will rebase on top of that work when it's in and put this behind it.

vdemeester (approved):
/meow

sbwsg:
I've moved this PR back to 0.24.  It's unlikely I'll be able to get this set up behind a feature gate prior to the 0.23 release early next week but no biggie if it only waits another month.

sbwsg:
TODO:

- [ ] Put behind alpha feature gate
- [ ] Add testing to ensure that the feature gate does indeed turn this on / off

sbwsg:
@bobcatfish @vdemeester @afrittoli 

I've implemented the `enable-api-fields` feature flag as part of this PR and put the new workspace feature behind the `alpha` flag.

I've moved the example YAMLs for this feature into their own directory and added some code to our examples test runner so that they only execute if the `enable-api-fields` feature flag is `alpha`.

sbwsg:
I'm not sure how to conditionally allow validations based on the `enable-api-fields` flag though.  I'm not sure if the validation webhook has access to the feature flag configmap.

sbwsg:
@vdemeester I've updated this PR to add "enable-api-fields" checks to the Validating webhook and disallow use of the new feature unless the gate is "alpha".

I'll port over the shared unit test code for checking feature gate to the Enable API Fields PR.

vdemeester:
> I'm not sure how to conditionally allow validations based on the `enable-api-fields` flag though. I'm not sure if the validation webhook has access to the feature flag configmap.

It should have it (we use the mechanism for default, custom-task, …).

afrittoli:
@sbwsg now that the feature gates PR is merged, this will need a rebase 🙏

sbwsg:
> now that the feature gates PR is merged, this will need a rebase 🙏

Rebased on top of new feature gates code.  I think this is ready for another look.

sbwsg:
I'm going to cancel the hold.  I can't find the comment where it was added and everythoing's up to date now with the feature gate work.

/hold cancel

examples/v1beta1/taskruns/workspace-in-sidecar.yaml
```diff
@@ -19,6 +19,10 @@ spec:
     - name: signals
     steps:
     - image: alpine:3.12.0
+      resources:
```
pritidesai, line 22: @sbwsg I am wondering why these changes are showing up here 🤔  These were already a part of previous [PR](https://github.com/tektoncd/pipeline/commit/565c55b30e422fc7fca42bc122561aa231106204#diff-d3a6963f568541460e24acddf2bcc5d3337bbbc8632e91a63ee59c6084891a2eR22-R25) you created.

pritidesai: and which was merged ...

pkg/apis/config/feature_flags.go
```diff
@@ -118,7 +118,7 @@ func NewFeatureFlagsFromMap(cfgMap map[string]string) (*FeatureFlags, error) {
 }
 
 // setEnabledAPIFields sets the "enable-api-fields" flag based on the content of a given map.
-// If the feature gate is invalid or missing then an error is returned.
+// If the feature gate is invalid or missing then the flag is set to its default.
```
pritidesai, line 121: NIT: this is not true, it does return error if the value is anything outside of `stable` and `alpha`, some git weirdness?

docs/workspaces.md
```diff
@@ -135,6 +158,9 @@ spec:
 
 #### Sharing `Workspaces` with `Sidecars`
```
pritidesai, line 159: NIT: reading for the first time looks like, the two notes in the same section is contradicting each other. The first note is saying with `alpha` API, the workspaces are automatically available to sidecars  and second note is about `stable` API. With `stable` API, sidecars have to opt-in. This is fine for now, no need to change anything here. But we might have to repeat ourselves and be explicit about the behavior with `stable` compared to `alpha` 😞  May be keep the `stable` API behavior as is and add changes being introduced with `alpha` in the end of the section 🤔

pkg/apis/config/feature_flags.go
sbwsg: Ah 😅 result of a bad merge. Good catch, thank you!

sbwsg (commented):
I rebased off of latest `main`; the weird artifacts are gone now.

docs/workspaces.md
sbwsg: Yeah you're right it was confusing in this context.  I removed mention of the alpha api from here.

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -110,6 +112,46 @@ func validateDeclaredWorkspaces(workspaces []WorkspaceDeclaration, steps []Step,
...
+
+	wsNames := sets.NewString()
+	for _, w := range workspaces {
+		wsNames.Insert(w.Name)
+	}
+
+	for stepIdx, step := range steps {
+		if len(step.Workspaces) != 0 {
+			errs = errs.Also(ValidateEnabledAPIFields(ctx, "step workspaces", config.AlphaAPIFields).ViaIndex(stepIdx).ViaField("steps"))
+		}
```
pritidesai, line 133: Should we return an error here or continue validating workspaces even when a feature gate is not set to `alpha`? 🤔 Generally we accumulate all the validation errors before exiting.

pritidesai:
Is this feature supported for a task within a pipeline? if so, will be great to have an example for the same 🙏

pkg/apis/pipeline/v1beta1/task_validation.go
sbwsg: Yeah, I've decided to leave them accumulating. We can always return early in future if we see reports that it's confusing this way.

sbwsg:
> Is this feature supported for a task within a pipeline? if so, will be great to have an example for the same 🙏

Added a pipelinerun example.

pritidesai:
> Added a pipelinerun example.

Thanks a bunch @sbwsg, one more question (sorry for nagging on this, trying to make sure I understand it since this is the first alpha feature 😜 ). I am not seeing the examples under `v1beta1/taskruns/alpha` and `v1beta1/pipelineruns/alpha` ran as part of the [integration](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/3700/pull-tekton-pipeline-integration-tests/1389891147064676355/) tests. Is that expected? The examples are more of a reference rather than validating the functionality?

afrittoli:
> # Release Notes
> ```
> - Sidecars in a Task now receive Workspaces, just like Steps do.
> - Steps and Sidecars can now specify which Workspaces they want access to. This isolates the workspace to only those Steps or Sidecars that need them, allowing Task authors to limit exposure of sensitive data like credentials.
> - The default behaviour for workspaces in Steps remains the same - all Steps get access to them.
> ```

Could you mention in the release notes that this change in behaviour is alpha / gated by a feature flag - so there will be no change in default behaviour when upgrading to v0.24? 🙏

sbwsg:
> > Added a pipelinerun example.
> 
> Thanks a bunch @sbwsg, one more question (sorry for nagging on this, trying to make sure I understand it since this is the first alpha feature 😜 ). I am not seeing the examples under `v1beta1/taskruns/alpha` and `v1beta1/pipelineruns/alpha` ran as part of the [integration](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/3700/pull-tekton-pipeline-integration-tests/1389891147064676355/) tests. Is that expected? The examples are more of a reference rather than validating the functionality?

I created https://github.com/tektoncd/pipeline/issues/3896 to tackle it and will pick it up asap.

> Could you mention in the release notes that this change in behaviour is alpha / gated by a feature flag - so there will be no change in default behaviour when upgrading to v0.24? 🙏

Sure thing!  Done.

examples/v1beta1/taskruns/alpha/workspace-isolation.yaml
```diff
@@ -0,0 +1,55 @@
...
+        #!/usr/bin/env ash
+        echo "hello world" > "$(workspaces.signals.path)"/start
+        echo "Signalled start"
+        # We protect from looping indefinitely by the TaskRun's timeout.
+        while [ ! -f "$(workspaces.signals.path)"/ready ] ; do
+          echo "Waiting for $(workspaces.signals.path)/ready"
+          sleep 1
+        done
+        echo "Saw ready file"
+    - name: check-signals-acess
```
afrittoli, line 34: NIT: s/acess/access

pkg/workspace/apply.go
```diff
@@ -98,6 +100,23 @@ func Apply(ts v1beta1.TaskSpec, wb []v1beta1.WorkspaceBinding, v map[string]core
 		ts.StepTemplate = &corev1.Container{}
 	}
 
+	isolatedWorkspaces := sets.NewString()
+
+	alphaAPIEnabled := config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields == "alpha"
```
pritidesai, line 105: NIT: can use `config.AlphaAPIFields` instead of string literal `"alpha"`

afrittoli (commented):
Review in progress 🙏 I will finish tomorrow if no-one gets to it first

pkg/workspace/apply.go
```diff
@@ -106,12 +125,23 @@ func Apply(ts v1beta1.TaskSpec, wb []v1beta1.WorkspaceBinding, v map[string]core
...
 			SubPath:   wb[i].SubPath,
 			ReadOnly:  w.ReadOnly,
-		})
+		}
+
+		if alphaAPIEnabled {
+			if isolatedWorkspaces.Has(w.Name) {
+				mountAsIsolatedWorkspace(ts, w.Name, volumeMount)
+			} else {
+				mountAsSharedWorkspace(ts, volumeMount)
```
pritidesai, line 139: I am guessing this will go away once this feature is promoted to stable 🤔

pritidesai: or rather the else ⬇️

pkg/workspace/apply_test.go
```diff
@@ -522,3 +524,310 @@ func TestApply(t *testing.T) {
 		})
 	}
 }
+
+func TestApply_IsolatedWorkspaces(t *testing.T) {
```
pritidesai, line 528: NIT: The tests included here are pretty comprehensive (go coverage is very obvious) 😍  but will be great to have a test with two different workspaces out of which one is isolated to a step and one is isolated to a sidecar.

pritidesai:
thanks a bunch @sbwsg for your patience on this 🙏 

It looks great. I am a little nervous without having `alpha` examples running as part of the CI process. But not a blocker here. I have validated all the examples on my local cluster and works smooth. We can wait until https://github.com/tektoncd/pipeline/issues/3896 is implemented (hopefully soon 🤞). I have also validated [workspaces.yaml](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/workspaces.yaml) `pipelineRun` with `alpha` API and works as expected without any issues.

/lgtm

PR #3708, opened 2021-01-22
Fix duplicate Pods when TaskRun or Pod labels are changed

### LINKED ISSUE #3656: Duplicated pods created for single Taskrun
# Expected Behavior
One pod for each `Taskrun`
# Actual Behavior
Multiple ones

# Additional Info
Sometimes the issue happen.

In my case, I register a `mutating admission webhook` for `Pod`, I guess it will delay the pod creation, and there must be race condition in reconcile of `taskrun`, see the code snippet:
https://github.com/tektoncd/pipeline/blob/55fdb95816a3b2c47ea6b8ee8789761615de60f1/pkg/reconciler/taskrun/taskrun.go#L380-L384

The first reconcile after a `taskrun` created should be:
`Check tr.Status.PodName  -> Create pod --> update TR`
and in second reconcile, we expect the pod is already created, but seems not.

so that's because `webhook` delay the pod creation,
If my guess is correct I think the issue will be found in many scenario, for example: the `scheduler` is slow or network is not good enough or `kubelet` on some node slow .etc.

Any idea, thanks. 
- Kubernetes version:

  ```
  (paste your output here)
  ```

- Tekton Pipeline version:

GregDritschler:
Fixes #3656 

The TaskRun reconciler may be given a TaskRun that does not have a previous status update containing the name of the pod it created.  The TaskRun reconciler deals with this by listing pods associated with the TaskRun.  It uses the TaskRun's labels, which were propagated to the pod, to find the pod.  If the user changes the TaskRun's labels, the Task's labels which are propagated to the TaskRun, or the pod's labels, while the TaskRun is running, it may cause the code to not find the existing pod and as a result it may create a duplicate.

# Changes

Change the TaskRun reconciler to only use the Tekton-controlled label `tekton.dev/taskRun` when calling LIST to find the pod.

/kind bug

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Fix duplicate Pods when TaskRun or Pod labels are changed
```

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -40,7 +40,8 @@ const (
 	// ResultsDir is the folder used by default to create the results file
 	ResultsDir = "/tekton/results"
 
-	taskRunLabelKey = pipeline.GroupName + pipeline.TaskRunLabelKey
+	// TaskRunLabelKey is the name of the label added to the Pod to identify the TaskRun
+	TaskRunLabelKey = pipeline.GroupName + pipeline.TaskRunLabelKey
 )
 
 // These are effectively const, but Go doesn't have such an annotation.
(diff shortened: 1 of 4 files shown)


vdemeester (approved):
Thanks for the fix @GregDritschler 😻 
/meow

Issue #3499, opened 2020-11-05
Add earlier validation of pipeline result references


sbwsg:
### Feature request

At the moment we fail a PipelineRun with `InvalidTaskResultReference` when we're scheduling the next TaskRuns to execute.  This can happen after some tasks in the pipeline have already executed, which takes time.  It should be possible to inspect all referenced Tasks before a PipelineRun starts and confirm that any references to Task Results from variables in the Pipeline are not misspelled or otherwise invalid.  This would result in a shorter time to failure and a shorter time to the user realizing their mistake.

### Use case

When a user submits a PipelineRun they could receive an earlier error message if any variables in the Pipeline refer to results that simply don't exist on the tasks being referenced.

### More Details

~~We'll need to decide if we're happy to pay whatever cost there is holding back the PipelineRun's start of execution until the reconciler has fetched and checked every referenced Task and its results.~~

Turns out we already fetch all the tasks as part of each PipelineRun reconcile.  So it should just be a case of validating the relationships between them on the first reconcile.

bobcatfish:
Was discussing with @jerop today and I think it's totally worth it! We could probably add it in such away that it only happens once per Run

pritidesai:
Just to be clear, this validation is also needed for a task result reference in a PipelineTask in addition to pipeline results. 

We have `tasks` section validation making sure referenced task exist but not validating the result reference is defined in the task's result section.

sbwsg:
Another use-case for reading all the Tasks in a Pipeline at startup time: It would be useful to check that a workspace marked "optional" at the pipeline level is indeed also "optional" in all the Tasks that want to consume that workspace.  Because if not that should probably be a validation error.

souleb:
can work on that

sbwsg:
Hey @souleb, thanks for jumping in. I'm halfway through implementing but will unassign myself and ping you if I'm not able to finish.

PR #3711, opened 2021-01-22
Validate dependencies between resolved resources in a PipelineRun

sbwsg:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/3499

When a PipelineRun is started there are several error conditions that can be
hit due to invalid Pipeline configuration. These errors have previously
only surfaced halfway through execution of the PipelineRun because validating
them requires a number of different resources to be resolved.

This commit performs validation on resolved dependencies of a PipelineRun before
it is allowed to start:
- All result variables used in the Pipeline are checked to be pointing at valid
Tasks and TaskResults. This requires looking at resolved TaskSpecs.
-  Workspaces marked optional by the Pipeline are confirmed to also be Optional
in the Tasks they're passed to. This also requires looking at resolved TaskSpecs.

This validation required searching for result variables in a PipelineTask in a very
similar way we do in several other places in our codebase. I've refactored all of these
to use a common func called `PipelineTaskResultRefs()`.

# Submitter Checklist

review them:

# Release Notes

```release-note
Added extra validations before PipelineRun can start: all result variables in the Pipeline must be valid and optional workspaces from a pipeline can only be passed to tasks expecting optional workspaces.
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/resultref.go
+++ b/pkg/apis/pipeline/v1beta1/resultref.go
@@ -129,3 +129,27 @@ func parseExpression(substitutionExpression string) (string, string, error) {
 	}
 	return subExpressions[1], subExpressions[3], nil
 }
+
+// PipelineTaskResultRefs walks all the places a result reference can be used
+// in a PipelineTask and returns a list of any references that are found.
+func PipelineTaskResultRefs(pt *PipelineTask) []*ResultRef {
+	refs := []*ResultRef{}
+	for _, condition := range pt.Conditions {
+		for _, p := range condition.Params {
+			expressions, _ := GetVarSubstitutionExpressionsForParam(p)
+			refs = append(refs, NewResultRefs(expressions)...)
+		}
+	}
+
+	for _, p := range pt.Params {
+		expressions, _ := GetVarSubstitutionExpressionsForParam(p)
+		refs = append(refs, NewResultRefs(expressions)...)
+	}
+
+	for _, whenExpression := range pt.WhenExpressions {
+		expressions, _ := whenExpression.GetVarSubstitutionExpressions()
+		refs = append(refs, NewResultRefs(expressions)...)
+	}
+
+	return refs
+}
--- a/pkg/reconciler/pipelinerun/resources/validate_dependencies.go
+++ b/pkg/reconciler/pipelinerun/resources/validate_dependencies.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2021 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
+import (
+	"fmt"
+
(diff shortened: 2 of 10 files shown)


sbwsg:
/hold

There's quite a bit of test coverage I need to add before this goes in.

pkg/reconciler/pipelinerun/resources/validate_dependencies.go
```diff
@@ -0,0 +1,99 @@
+package resources
```
pritidesai, line 1: sorry @sbwsg for early review, missing copyright license here and in the test file 😉

sbwsg: Ah, great point, I completely forgot, thanks @pritidesai

sbwsg:
/hold cancel

Improved coverage.

pkg/reconciler/pipelinerun/resources/validate_dependencies.go
```diff
@@ -0,0 +1,115 @@
...
+						if !tws.Optional {
+							return fmt.Errorf("pipeline workspace %q is marked optional but pipeline task %q requires it be provided", pws.Workspace, rprt.PipelineTask.Name)
+						}
+					}
+				}
+			}
+		}
+	}
+	return nil
+}
```
jerop, line 115: this makes me think that it may be clearer to have the reason above as `ReasonRequiredWorkspaceMarkedOptional` instead of `ReasonOptionalWorkspaceNotSupportedByTask`

https://github.com/tektoncd/pipeline/blob/2cf44c90bad4d3321fd118cdb08d559164a4fa47/pkg/reconciler/pipelinerun/pipelinerun.go#L105-L107

sbwsg: Thanks for reviewing, I've made this change.

jerop (approved):
thanks @sbwsg for the added validation and making using result references cleaner!

pkg/apis/pipeline/v1beta1/resultref.go
```diff
@@ -129,3 +129,27 @@ func parseExpression(substitutionExpression string) (string, string, error) {
 	}
 	return subExpressions[1], subExpressions[3], nil
 }
+
+// PipelineTaskResultRefs walks all the places a result reference can be used
+// in a PipelineTask and returns a list of any references that are found.
+func PipelineTaskResultRefs(pt *PipelineTask) []*ResultRef {
```
pritidesai, line 135: thanks @sbwsg this is a huge simplification. But I am trying to understand this, any specific reason moving this logic here from `pipeline_types.go`? collecting all resultrefs under `resultref.go`?

sbwsg: Are you referring here to the fact that I originally made this a method on `*PipelineTask`?

In a follow-up PR I'd like to move all the result-related code to its own package. Just like we have `pkg/workspace` I think we should have `pkg/results`.  It's always been a little strange to me that all of the results processing lives in `v1beta1`.

So what I'd like to avoid is attaching the result processing logic to our v1beta1 types.  I initially added this as a method on `*PipelineTask` because it was "neat" to be able to call `pt.ResultReferences()`.  But I ultimately wasn't convinced that the "neatness" was worth coupling that logic to the beta type.

pritidesai: nevermind got it 😜  its a pure refactoring to minimize iterating over resultRefs in three separate for loops.

pritidesai: yup, sounds good @sbwsg thanks for explaining it.

pritidesai:
Thanks @sbwsg, this PR has lot of simplifications around results and I appreciate these changes 👍 

/lgtm

Issue #3712, opened 2021-01-25
Add nonroot user to base image


TakiTake:
### Feature request

Could you add nonroot user to base image?
Like this
```
FROM gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.18.1
RUN addgroup -S -g 65532 nonroot && adduser -S -u 65532 nonroot -G nonroot
ENV HOME=/home/nonroot
```

I'd like to run Task as nonroot user as much as possible. But there is no prefer user in the base image. For example, not in gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init

### Use case

Run git-clone task as nonroot user.
https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2

I tested with following TaskRun.
And I have disabled home env overwrite to use different home from /home/tekton.

```
$ kubectl get cm feature-flags -n tekton-pipelines -o yaml
apiVersion: v1
data:
  disable-affinity-assistant: "false"
  disable-creds-init: "false"
  disable-home-env-overwrite: "true"
  disable-working-directory-overwrite: "false"
  enable-custom-tasks: "false"
  enable-tekton-oci-bundles: "false"
  require-git-ssh-secret-known-hosts: "false"
  running-in-environment-with-injected-sidecars: "true"
kind: ConfigMap
...
```

#### Without nonroot user

```
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  generateName: git-clone-
spec:
  params:
  - name: revision
    value: main
  - name: url
    value: git@github.com:TakiTake/private.git
  serviceAccountName: default
  taskRef:
    kind: Task
    name: git-clone
  timeout: 1h0m0s
  workspaces:
  - emptyDir: {}
    name: output
  podTemplate:
    securityContext:
      runAsNonRoot: true
      runAsUser: 65532
```

```
$ tkn tr logs
? Select taskrun: git-clone-tpqf7 started 23 seconds ago
task git-clone has failed: "step-clone" exited with code 1 (image: "gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init@sha256:406e9a4f9d10dbc59e141c1584d1d4dcff995012071eb425a3c117ec0dfe16af"); for logs run: kubectl -n default logs git-clone-tpqf7-pod-mpf4c -c step-clone

[...]

PR #3727, opened 2021-01-28
Add nonroot user to pipeline's build-base image

sbwsg:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/3712

Prior to this commit the git-init image only provided a root
user, meaning git clones had to be performed as root.

After this commit the git-init image provides a nonroot user
with UID and GID of 65532 and a home directory of /home/nonroot.
Utilizing this UID results in creds-init credentials being placed
in /home/nonroot/.ssh when disable-home-env-overwrite is "true",
and results in the cloned files being owned by UID 65532.

/kind misc

# Submitter Checklist

review them:

# Release Notes

```release-note
Added non-root user 65532 to pipelines' build-base image. Git-init can now be used to clone repositories as a non-root user.
```

### DIFF
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -1,5 +1,7 @@
 FROM alpine:3.11
-  
+
+RUN addgroup -S -g 65532 nonroot && adduser -S -u 65532 nonroot -G nonroot
+
 RUN apk add --update git git-lfs openssh-client \
     && apk update \
     && apk upgrade


zhangtbj:
Hi @sbwsg 

A quick question, After this PR is merged, does it mean we can use `non-root` base image for `git-init` in future?

Then how about `gcs-fetcher` and `pullrequest-init` in https://github.com/tektoncd/pipeline/blob/master/.ko.yaml#L7-L11
Thanks!

chmouel:
I think we are missing a `"USER nonroot"` to make nonroot as the default of the image, isnt it ?

sbwsg:
> I think we are missing a "USER nonroot" to make nonroot as the default of the image, isnt it ?

I didn't want to switch the default in the initial pull request because I'm actually not sure who or what might get broken by that.  Initially I just wanted to provide the non root option then figure out what might break over time.

But do you think it's a good idea to just change the default?  I could be convinced if you think so!

chmouel:
I am not so sure to be honest :) 

O vanilla k8 it definitively could be a problem.

As you probably know and just synthesizing the problem to make sure I understand it, on Tekton we cannot have a mix of root containers and non root containers, if we switch one of our init containers as non root we need to make sure that every other init containers we are running are non root as well.

For example if we have a Pipeline with : 

- a workspace
- a task doing something in the workspace with a workingDir directive.
- a following task using git_clone which use the workingDir

If the workingDir init container run as root and the git_clone task that runs after it is set as non root it will fails.

The solution would be i think is to make sure all our images as shipped from tekton run as non root,

I am not so sure because it should work because we have been doing this since a long time on OpenShift Pipelines, running all container as non root : 

https://github.com/openshift/tektoncd-pipeline/blob/master/openshift/ci-operator/uidwrapper#L1-L8

But openshfit does some magic too  with regards to volume binding to the random user that vanilla k8 doesn't do, and we cannot "enforce" users to always run everything as non root in their pipelines.

So in retrospect I think you are right we should not do this by default, if we want to move to a non root world, we may have to switch everything to non root at the same time and test the hell out of all the permutations to see how we can handle the PVC issue 😅

sbwsg:
> A quick question, After this PR is merged, does it mean we can use non-root base image for git-init in future?

This PR adds `nonroot` user to our `build-base` image, which [git-init already uses](https://github.com/tektoncd/pipeline/blob/master/.ko.yaml#L5). Once this goes in `git-init` should then have support for `nonroot`.  I've opened [an on-hold PR to the catalog](https://github.com/tektoncd/catalog/pull/637) that will receive this `git-init` version once this change is published in an official release.

> Then how about gcs-fetcher and pullrequest-init in https://github.com/tektoncd/pipeline/blob/master/.ko.yaml#L7-L11
Thanks!

Yeah good call those could also be updated to use this image as a base so that `nonroot` becomes available for them too.

ImJasonH:
> > Then how about gcs-fetcher and pullrequest-init in https://github.com/tektoncd/pipeline/blob/master/.ko.yaml#L7-L11
> > Thanks!
> 
> Yeah good call those could also be updated to use this image as a base so that `nonroot` becomes available for them too.

`pullrequest-init` doesn't need any of the packages available in build-base, so I'd prefer not to base it on build-base if we can avoid it. (`gcs-fetcher` is going away soon so we don't need to care about it)

zhangtbj:
For the production usage. It is best we can use `non-root` base image for those. We need to follow some security rules on production environment (For Tekton, we applied some security exception now :( )

ImJasonH:
> For the production usage. It is best we can use `non-root` base image for those. We need to follow some security rules on production environment (For Tekton, we applied some security exceptions now :( )

I agree we should strive to get pullrequest-init to use a non-root user, I just don't think basing it on build-base is the best solution to that. Either way it doesn't block this PR, let's take that discussion to another issue so we can focus on that there.

zhangtbj:
> Either way it doesn't block this PR, let's take that discussion to another issue so we can focus on that there.

Sure thing. Just interest in this PR and ask an additional question here. The question is not related to this PR change.

Thanks! :)

sbwsg:
Created issue to capture request for pullrequest nonroot user: https://github.com/tektoncd/pipeline/issues/3746

vdemeester (approved):
/meow

chmouel:
This should help I think https://github.com/kubernetes/enhancements/pull/2101

Issue #3438, opened 2020-10-22
Be able to use when expressions in finally Tasks


bobcatfish:
### Feature request

To be able to combine [when expressions](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#guard-task-execution-using-whenexpressions) with [task execution status](https://github.com/tektoncd/community/pull/234) in finally tasks to be able to express when you want actions to happen in the case of the failure of another Task (without having to modify the Task to be aware of the success/failure case, e.g. use Tasks straight from the catalog).

### Use case

From Andrew Johnson on slack:

> I'd like to conditionally run a set of steps (can be tasks as well) if a task in the main pipeline fails - So an OnFailure (of a specific task) as opposed to Finally.

For example, let's say we want to post to slack (using [this slack catalog task](https://github.com/tektoncd/catalog/tree/master/task/send-to-channel-slack/0.1#post-a-message-to-slack) when a certain Task in the Pipeline fails:

```
  tasks:
    - name: do-things
      taskRef:
        name: do-things
  finally:
    - name: alert-slack
       when:
          - input: "$(context.pipelineRun.Tasks.do-things.status)"
            operator: in
            values: ["failed"]
      taskRef:
        name: send-to-channel-slack
      params:
        - name: channel-id
          value: foo-bar
        ...
```

Andrew also describes how this has to be worked around when we do not have this feature:

> The way I'm currently orchestrating this is by writing a failed status to a workspace file before the task and then removing the failed status after the task. If, when we get to finally, the status is still failed then the steps (or tasks) should run. This currently requires custom tasks that check for the file first.

This means that in order to get this functionality:

1. The executing Task has to be aware of the need to track its status in a workspace
2. The finally Task has to be aware of the meaning of the values written to the workspace

(1) and (2) would prevent a user from being able to use catalog Tasks as is for this use case

bobcatfish:
@jerop @pritidesai  might be relevant to your interests! feel free to close if there is already an issue tracking this

jerop:
> @jerop @pritidesai might be relevant to your interests! feel free to close if there is already an issue tracking this

this is something we've been discussing as next steps after execution status and task results are supported in finally tasks but don't have an issue to track afaik, thanks for opening this...will assign myself 

/assign

jerop:
[...]

PR #3738, opened 2021-02-01
WhenExpressions in Finally Tasks

jerop:
# Changes

Users can guard execution of `Tasks` using `WhenExpressions`, but that
is currently not supported in `Finally Tasks`.

This change adds support for `WhenExpressions` in `Finally Tasks` not
only to provide efficient guarded execution but also to improve the
reusability of `Tasks` in `Finally`. The proposal is described further in
the [WhenExpressions in Finally Tasks TEP](https://github.com/tektoncd/community/blob/master/teps/0045-whenexpressions-in-finally-tasks.md).

Given we've recently added support for `Results` and `Status` in
`Finally Tasks`, this is an opportune time to enable `WhenExpressions`
in `Finally Tasks`.

/kind feature

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
WhenExpressions are supported in Finally Tasks not only to provide efficient guarded execution but also to improve the
reusability of `Tasks` in `Finally`
```

### DIFF
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -31,6 +31,10 @@ weight: 3
     - [Consuming `Task` execution results in `finally`](#consuming-task-execution-results-in-finally)
     - [`PipelineRun` Status with `finally`](#pipelinerun-status-with-finally)
     - [Using Execution `Status` of `pipelineTask`](#using-execution-status-of-pipelinetask)
+    - [Guard `Finally Task` execution using `WhenExpressions`](#guard-finally-task-execution-using-whenexpressions)
+      - [`WhenExpressions` using `Parameters` in `Finally Tasks`](#whenexpressions-using-parameters-in-finally-tasks)
+      - [`WhenExpressions` using `Results` in `Finally Tasks`](#whenexpressions-using-results-in-finally-tasks)
+      - [`WhenExpressions` using `Execution Status` of `PipelineTask` in `Finally Tasks`](#whenexpressions-using-execution-status-of-pipelinetask-in-finally-tasks)  
     - [Known Limitations](#known-limitations)
       - [Specifying `Resources` in Final Tasks](#specifying-resources-in-final-tasks)
       - [Cannot configure the Final Task execution order](#cannot-configure-the-final-task-execution-order)
@@ -895,6 +899,118 @@ This kind of variable can have any one of the values from the following table:
 
 For an end-to-end example, see [`status` in a `PipelineRun`](../examples/v1beta1/pipelineruns/pipelinerun-task-execution-status.yaml).
 
+### Guard `Finally Task` execution using `WhenExpressions`
+
+Similar to `Tasks`, `Finally Tasks` can be guarded using [`WhenExpressions`](#guard-task-execution-using-whenexpressions)
+that operate on static inputs or variables. Like in `Tasks`, `WhenExpressions` in `Finally Tasks` can operate on
+`Parameters` and `Results`. Unlike in `Tasks`, `WhenExpressions` in `Finally Tasks` can also operate on the [`Execution
+Status`](#using-execution-status-of-pipelinetask) of `Tasks`.
+
+#### `WhenExpressions` using `Parameters` in `Finally Tasks`
+
+`WhenExpressions` in `Finally Tasks` can utilize `Parameters` as demonstrated using [`golang-build`](https://github.com/tektoncd/catalog/tree/master/task/golang-build/0.1)
+and [`send-to-channel-slack`](https://github.com/tektoncd/catalog/tree/master/task/send-to-channel-slack/0.1) Catalog
+`Tasks`:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: pipelinerun-
+spec:
+  pipelineSpec:
+    params:
+    - name: enable-notifications
+      type: string
+      description: a boolean indicating whether the notifications should be sent
+    tasks:
+      - name: golang-build
+        taskRef:
+          name: golang-build
+      # […]
+    finally:
+      - name: notify-build-failure # executed only when build task fails and notifications are enabled
+        when:
+          - input: $(tasks.golang-build.status)
+            operator: in
+            values: ["Failed"]
+          - input: $(params.enable-notifications)
+            operator: in
+            values: ["true"]
+        taskRef:
+          name: send-to-slack-channel
+      # […]
+  params:
+    - name: enable-notifications
+      value: true
+```
+
+#### `WhenExpressions` using `Results` in `Finally Tasks`
+
+`WhenExpressions` in `Finally Tasks` can utilize `Results`, as demonstrated using [`git-clone`](https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.2)
+and [`github-add-comment`](https://github.com/tektoncd/catalog/tree/master/task/github-add-comment/0.2) Catalog `Tasks`:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: pipelinerun-
+spec:
+  pipelineSpec:
+    tasks:
+      - name: git-clone
+        taskRef:
+          name: git-clone
+      - name: go-build
+      # […]
+    finally:
+      - name: notify-commit-sha # executed only when commit sha is not the expected sha
+        when:
+          - input: $(tasks.git-clone.results.commit)
+            operator: notin
+            values: [$(params.expected-sha)]
+        taskRef:
+          name: github-add-comment
+      # […]
+  params:
+    - name: expected-sha
+      value: 54dd3984affab47f3018852e61a1a6f9946ecfa
+```
+
+If the `WhenExpressions` in a `Finally Task` use `Results` from a skipped or failed non-finally `Tasks`, then the
+`Finally Task` would also be skipped and be included in the list of `Skipped Tasks` in the `Status`, [similarly to when using
+`Results` in other parts of the `Finally Task`](#consuming-task-execution-results-in-finally).
+
+#### `WhenExpressions` using `Execution Status` of `PipelineTask` in `Finally Tasks`
+
+`WhenExpressions` in `Finally Tasks` can utilize [`Execution Status` of `PipelineTasks`](#using-execution-status-of-pipelinetask), 
+as as demonstrated using [`golang-build`](https://github.com/tektoncd/catalog/tree/master/task/golang-build/0.1) and
+[`send-to-channel-slack`](https://github.com/tektoncd/catalog/tree/master/task/send-to-channel-slack/0.1) Catalog `Tasks`:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: pipelinerun-
+spec:
+  pipelineSpec:
+    tasks:
+      - name: golang-build
+        taskRef:
+          name: golang-build
+      # […]
+    finally:
+      - name: notify-build-failure # executed only when build task fails
+        when:
+          - input: $(tasks.golang-build.status)
+            operator: in
+            values: ["Failed"]
+        taskRef:
+          name: send-to-slack-channel
+      # […]
+```
+
+For an end-to-end example, see [PipelineRun with WhenExpressions](../examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml).
 
 ### Known Limitations
 
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation.go
@@ -70,7 +70,7 @@ func (ps *PipelineSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	errs = errs.Also(validatePipelineResults(ps.Results))
 	errs = errs.Also(validateTasksAndFinallySection(ps))
 	errs = errs.Also(validateFinalTasks(ps.Tasks, ps.Finally))
-	errs = errs.Also(validateWhenExpressions(ps.Tasks))
+	errs = errs.Also(validateWhenExpressions(ps.Tasks, ps.Finally))
 	return errs
 }
 
@@ -329,22 +329,32 @@ func validateExecutionStatusVariablesInFinally(tasks []PipelineTask, finally []P
 	ptNames := PipelineTaskList(tasks).Names()
 	for idx, t := range finally {
 		for _, param := range t.Params {
-			// retrieve a list of substitution expression from a param
-			if ps, ok := GetVarSubstitutionExpressionsForParam(param); ok {
-				// validate tasks.pipelineTask.status if this expression is not a result reference
-				if !LooksLikeContainsResultRefs(ps) {
-					for _, p := range ps {
-						// check if it contains context variable accessing execution status - $(tasks.taskname.status)
-						if containsExecutionStatusRef(p) {
-							// strip tasks. and .status from tasks.taskname.status to further verify task name
-							pt := strings.TrimSuffix(strings.TrimPrefix(p, "tasks."), ".status")
-							// report an error if the task name does not exist in the list of dag tasks
-							if !ptNames.Has(pt) {
-								errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("pipeline task %s is not defined in the pipeline", pt),
-									"value").ViaFieldKey("params", param.Name).ViaFieldIndex("finally", idx))
-							}
-						}
-					}
+			if expressions, ok := GetVarSubstitutionExpressionsForParam(param); ok {
+				errs = errs.Also(validateExecutionStatusVariablesExpressions(expressions, ptNames, "value").ViaFieldKey(
+					"params", param.Name).ViaFieldIndex("finally", idx))
+			}
+		}
+		for i, we := range t.WhenExpressions {
+			if expressions, ok := we.GetVarSubstitutionExpressions(); ok {
+				errs = errs.Also(validateExecutionStatusVariablesExpressions(expressions, ptNames, "").ViaFieldIndex(
+					"when", i).ViaFieldIndex("finally", idx))
+			}
+		}
+	}
+	return errs
+}
+
+func validateExecutionStatusVariablesExpressions(expressions []string, ptNames sets.String, fieldPath string) (errs *apis.FieldError) {
+	// validate tasks.pipelineTask.status if this expression is not a result reference
+	if !LooksLikeContainsResultRefs(expressions) {
+		for _, expression := range expressions {
+			// check if it contains context variable accessing execution status - $(tasks.taskname.status)
+			if containsExecutionStatusRef(expression) {
+				// strip tasks. and .status from tasks.taskname.status to further verify task name
+				pt := strings.TrimSuffix(strings.TrimPrefix(expression, "tasks."), ".status")
+				// report an error if the task name does not exist in the list of dag tasks
+				if !ptNames.Has(pt) {
+					errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("pipeline task %s is not defined in the pipeline", pt), fieldPath))
 				}
 			}
 		}
@@ -429,9 +439,6 @@ func validateFinalTasks(tasks []PipelineTask, finalTasks []PipelineTask) *apis.F
 		if len(f.Conditions) != 0 {
 			return apis.ErrInvalidValue(fmt.Sprintf("no conditions allowed under spec.finally, final task %s has conditions specified", f.Name), "").ViaFieldIndex("finally", idx)
 		}
-		if len(f.WhenExpressions) != 0 {
-			return apis.ErrInvalidValue(fmt.Sprintf("no when expressions allowed under spec.finally, final task %s has when expressions specified", f.Name), "").ViaFieldIndex("finally", idx)
-		}
 	}
 
 	ts := PipelineTaskList(tasks).Names()
@@ -487,10 +494,13 @@ func validateTasksInputFrom(tasks []PipelineTask) (errs *apis.FieldError) {
 	return errs
 }
 
-func validateWhenExpressions(tasks []PipelineTask) (errs *apis.FieldError) {
+func validateWhenExpressions(tasks []PipelineTask, finalTasks []PipelineTask) (errs *apis.FieldError) {
 	for i, t := range tasks {
 		errs = errs.Also(validateOneOfWhenExpressionsOrConditions(t).ViaFieldIndex("tasks", i))
 		errs = errs.Also(t.WhenExpressions.validate().ViaFieldIndex("tasks", i))
+	}
+	for i, t := range finalTasks {
+		errs = errs.Also(t.WhenExpressions.validate().ViaFieldIndex("finally", i))
 	}
 	return errs
 }
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
@@ -277,6 +277,55 @@ func TestPipelineSpec_Validate_Failure(t *testing.T) {
 			Paths:   []string{"tasks[0].when[0]"},
 		},
 	}, {
+		name: "invalid pipeline with final task having when expression with invalid operator (not In/NotIn)",
+		ps: &PipelineSpec{
+			Description: "this is an invalid pipeline with invalid pipeline task",
+			Tasks: []PipelineTask{{
+				Name:    "invalid-pipeline-task",
+				TaskRef: &TaskRef{Name: "bar-task"},
+			}},
+			Finally: []PipelineTask{{
+				Name:    "invalid-pipeline-task-finally",
+				TaskRef: &TaskRef{Name: "bar-task"},
+				WhenExpressions: []WhenExpression{{
+					Input:    "foo",
+					Operator: selection.Exists,
+					Values:   []string{"foo"},
+				}},
+			}},
+		},
+		expectedError: apis.FieldError{
+			Message: `invalid value: operator "exists" is not recognized. valid operators: in,notin`,
+			Paths:   []string{"finally[0].when[0]"},
+		},
+	}, {
+		name: "invalid pipeline with dag task and final task having when expression with invalid operator (not In/NotIn)",
+		ps: &PipelineSpec{
+			Description: "this is an invalid pipeline with invalid pipeline task",
+			Tasks: []PipelineTask{{
+				Name:    "invalid-pipeline-task",
+				TaskRef: &TaskRef{Name: "bar-task"},
+				WhenExpressions: []WhenExpression{{
+					Input:    "foo",
+					Operator: selection.Exists,
+					Values:   []string{"foo"},
+				}},
+			}},
+			Finally: []PipelineTask{{
+				Name:    "invalid-pipeline-task-finally",
+				TaskRef: &TaskRef{Name: "bar-task"},
+				WhenExpressions: []WhenExpression{{
+					Input:    "foo",
+					Operator: selection.Exists,
+					Values:   []string{"foo"},
+				}},
+			}},
+		},
+		expectedError: apis.FieldError{
+			Message: `invalid value: operator "exists" is not recognized. valid operators: in,notin`,
+			Paths:   []string{"tasks[0].when[0]", "finally[0].when[0]"},
+		},
+	}, {
 		name: "invalid pipeline with one pipeline task having when expression with invalid values (empty)",
 		ps: &PipelineSpec{
 			Description: "this is an invalid pipeline with invalid pipeline task",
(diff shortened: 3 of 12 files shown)


jerop:
closes https://github.com/tektoncd/pipeline/issues/3438

pritidesai:
thanks @jerop, I haven't reviewed the PR yet but I would like to see `e2e` tests in [pipelinefinally_test.go](https://github.com/tektoncd/pipeline/blob/master/test/pipelinefinally_test.go)

jerop:
> thanks @jerop, I haven't reviewed the PR yet but I would like to see `e2e` tests in [pipelinefinally_test.go](https://github.com/tektoncd/pipeline/blob/master/test/pipelinefinally_test.go)

i'd added them in [pipelinerun_test.go](https://github.com/tektoncd/pipeline/blob/master/pkg/reconciler/pipelinerun/pipelinerun_test.go), will add in [pipelinefinally_test.go](https://github.com/tektoncd/pipeline/blob/master/test/pipelinefinally_test.go) as well, thanks @pritidesai! 

/hold

jerop:
@pritidesai I've added the tests in [pipelinefinally_test.go](https://github.com/tektoncd/pipeline/blob/master/test/pipelinefinally_test.go), please take a look 🙏🏾 

/hold cancel

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -4855,10 +4875,12 @@ func TestReconcileWithTaskResultsInFinalTasks(t *testing.T) {
 	}
 	expectedSkippedTasks := []v1beta1.SkippedTask{{
 		Name: "final-task-2",
+	}, {
+		Name: "final-task-3",
 	}}
 
 	if d := cmp.Diff(expectedSkippedTasks, reconciledRun.Status.SkippedTasks); d != "" {
-		t.Fatalf("Expected to see only one final task (final-task-2) in the list of skipped tasks. Diff: %s", diff.PrintWantGot(d))
+		t.Fatalf("Didn't get the expected the list of skipped tasks. Diff: %s", diff.PrintWantGot(d))
```
t-persson, line 4883:
```suggestion
		t.Fatalf("Didn't get the expected list of skipped tasks. Diff: %s", diff.PrintWantGot(d))
```

docs/pipelines.md
```diff
@@ -880,10 +900,119 @@ This kind of variable can have any one of the values from the following table:
...
+      description: a boolean indicating whether the notifications should be sent
+    tasks:
+      - name: golang-build
+        taskRef:
+          name: golang-build
+      # […]
+    finally:
+      - name: notify-build-failure # executed only when build task fails and notifications are enabled
+        when:
+          - input: $(tasks.golang-build.status)
```
chmouel, line 935: I see most of the examples or tests refer to task status, is there a way to guard again pipeline run status? I think the user in the send to slack example would typically would want to grab the status of the execution of the whole pipeline if it has failed or succeeded, what do you think having this as the example?

jerop: pipeline run status is calculated based on both tasks and finally tasks, as described [here](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#pipelinerun-status-with-finally), so I don't see how we can guard against that because it won't be available yet when executing finally tasks

if you're referring to the overall status of the tasks only, not finally tasks, I can imagine that a user could pass in all the statuses of the tasks to the values array, as shown below 

```yaml
finally:
      - name: notify-failure # executed only when tasks failed
        when:
          - input: "Failed"
 	    operator: in
            values: [$(tasks.unit-tests.status), $(tasks.golang-build.status), $(tasks.deploy.status)]         
```

jerop: /cc @pritidesai on execution status variables available in finally

sbwsg (approved):

docs/pipelines.md
pritidesai: @chmouel I have a WIP proposal to support [aggregate](https://github.com/tektoncd/pipeline/issues/1020#issuecomment-747156100) status of `tasks` section, would that work for your use case? 

The proposal includes introducing a new context variable `$(tasks.status)` (feedback on syntax is welcome) 🙏. `$(tasks.status)` would be set to `Failed` if any on of the `dag` task fails, and `Succeeded` if all `dag` tasks succeed.

jerop: the aggregate status of `tasks` would be a very useful in `whenexpressions` in `finally tasks`, thanks for the update @pritidesai!

docs/pipelines.md
```diff
@@ -6,27 +6,47 @@ weight: 3
 -->
 # Pipelines
 
-- [Overview](#pipelines)
-- [Configuring a `Pipeline`](#configuring-a-pipeline)
+- [Pipelines](#pipelines)
```
bobcatfish, line 9: i think something odd happened here - not sure why we have a top level link back to the top of the page? (maybe you're using a tool that's doing that?)

maybe these updates to the TOC would make sense in a separate PR?  i think i like having the more verbose TOC but it will be harder to keep up to date 🤔

docs/pipelines.md
```diff
@@ -880,10 +900,119 @@ This kind of variable can have any one of the values from the following table:
...
+          - input: $(params.enable-notifications)
+            operator: in
+            values: ["true"]
+        taskRef:
+          name: send-to-slack-channel
+      # […]
+  params:
+    - name: enable-notifications
+      value: true
+```
```
bobcatfish, line 947: cool example :D

docs/pipelines.md
```diff
@@ -880,10 +900,119 @@ This kind of variable can have any one of the values from the following table:
...
+        taskRef:
+          name: boskos-acquire
+      - name: use-resource
+      # […]
+    finally:
+      - name: boskos-release # executed only when leased resource is phonetic-project
+        when:
+          - input: $(tasks.boskos-acquire.results.leased-resource)
+            operator: in
+            values: ["phonetic-project"]
```
bobcatfish, line 972: this example feels a bit weird - not sure why we are releasing only in this one case, feels more likely that this would be checking the status of the boskos-acquire task 🤔   np if you want to leave it - alternative idea, git-clone task maybe?

alternatively: maybe we don't need to have an example of results + finally? knowing how to use variable replacement in general could be enough?

docs/pipelines.md
```diff
@@ -880,10 +900,119 @@ This kind of variable can have any one of the values from the following table:
...
+        taskRef:
+          name: boskos-release
+      # […]
+```
+
+If the `WhenExpressions` in a `Finally Task` use `Results` from a skipped or failed non-finally `Tasks`, then the
+`Finally Task` would also be skipped and be included in the list of `Skipped Tasks` in the `Status`, [similarly to when using
+`Results` in other parts of the `Finally Task`](https://github.com/tektoncd/pipeline/blob/master/docs/pipelines.md#consuming-task-execution-results-in-finally).
```
bobcatfish, line 980: you can also link to this with `pipeline.md#consuming-task...` (i..e you dont need the whole URL, and using the short form means github can preview the link in PRs etc)

docs/pipelines.md
```diff
@@ -880,10 +900,119 @@ This kind of variable can have any one of the values from the following table:
...
+          name: golang-build
+      # […]
+    finally:
+      - name: notify-build-failure # executed only when build task fails
+        when:
+          - input: $(tasks.golang-build.status)
+            operator: in
+            values: ["Failed"]
+        taskRef:
+          name: send-to-slack-channel
```
bobcatfish, line 1007: woooot

docs/pipelines.md
bobcatfish: (feel free to ignore me)

test/pipelinefinally_test.go
```diff
@@ -84,6 +85,11 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_InvalidTaskResult_Failure(t *test
 		t.Fatalf("Failed to create Task producing task results: %s", err)
 	}
 
+	taskConsumingResultInParam := getSuccessTaskConsumingResultsInParam(t, namespace)
+	if _, err := c.TaskClient.Create(ctx, taskConsumingResultInParam, metav1.CreateOptions{}); err != nil {
+		t.Fatalf("Failed to create Task consuming task results in param: %s", err)
+	}
+
 	taskConsumingResult := getSuccessTaskConsumingResults(t, namespace)
```
bobcatfish, line 93: im finding this a bit hard to follow, i think mostly b/c the details of the tasks are hidden in the functions (i.e. it was confusing before this change!), im wondering if some comments could help explain, e.g. what is taskConsumingResultInParam and how is it different from taskConsumingResult and why do we need both

test/pipelinefinally_test.go
```diff
@@ -201,13 +247,16 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_InvalidTaskResult_Failure(t *test
 		t.Fatalf("Error listing TaskRuns for PipelineRun %s: %s", pipelineRun.Name, err)
 	}
 
-	// expecting taskRuns for dagtask1, dagtask2, dagtask3 (with condition failure), dagtask5, finaltask1, finaltask2, and finaltaskconsumingdagtask5
-	if len(taskrunList.Items) != 7 {
+	// expecting taskRuns for dagtask1, dagtask2, dagtask3 (with condition failure), dagtask5, finaltask1, finaltask2,
+	// finaltaskconsumingdagtask5, finaltaskconsumingdagtask5inwhenexpressionstrue, finaltaskconsumingdagtask5executionstatusinwhenexpressionstrue
```
bobcatfish, line 251: i feel like these names are getting a bit long - i wonder if there is some other way we can make this clearer 🤔

test/pipelinefinally_test.go
```diff
@@ -201,13 +247,16 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_InvalidTaskResult_Failure(t *test
...
+	expectedTaskRunsCount := 9
+	if len(taskrunList.Items) != expectedTaskRunsCount {
 		var s []string
 		for _, n := range taskrunList.Items {
 			s = append(s, n.Labels["tekton.dev/pipelineTask"])
 		}
-		t.Fatalf("Error retrieving TaskRuns for PipelineRun %s. Expected 5 taskRuns and found taskRuns for: %s", pipelineRun.Name, strings.Join(s, ", "))
+		t.Fatalf("Error retrieving TaskRuns for PipelineRun %s. Expected %s taskRuns and found %s taskRuns for: %s",
+			pipelineRun.Name, strconv.Itoa(expectedTaskRunsCount), strconv.Itoa(len(taskrunList.Items)), strings.Join(s, ", "))
```
bobcatfish, line 259: instead of calling ` strconv.Itoa` we could use the `%d` format operator (%s is forcing us to use a string: https://golang.org/pkg/fmt/#hdr-Printing)

test/pipelinefinally_test.go
```diff
@@ -269,7 +318,17 @@ func TestPipelineLevelFinally_OneDAGTaskFailed_InvalidTaskResult_Failure(t *test
 					t.Errorf("Error resolving task result reference in a finally task %s", n)
 				}
 			}
-		case n == "finaltaskconsumingdagtask1" || n == "finaltaskconsumingdagtask4":
+		case n == "finaltaskconsumingdagtask5inwhenexpressionstrue":
+			if err := WaitForTaskRunState(ctx, c, taskrunItem.Name, TaskRunSucceed(taskrunItem.Name), "TaskRunSuccess"); err != nil {
```
bobcatfish, line 322: do we want to be waiting here? I think we already waited for the pipelinerun as a whole to get into the state we expected (or maybe im wrong?) - if the taskruns aren't in the expected state after the pipelinerun has completed, i think something has gone wrong?

test/pipelinefinally_test.go
```diff
@@ -477,12 +556,17 @@ func getPipeline(t *testing.T, namespace string, dag map[string]pipelineTask, f
...
-			Params:  v.Param,
-		})
+		}
+		if len(v.Param) != 0 {
+			finalTask.Params = v.Param
+		}
+		if len(v.When) != 0 {
+			finalTask.WhenExpressions = v.When
+		}
+		fpt = append(fpt, finalTask)
```
bobcatfish, line 569: i wonder how useful this function actually is vs declaring a pipeline inline

bobcatfish (commented):
So far I've reviewed docs, examples and the end to end test - planning to come back and review the rest but dont want to lose this so far :)

docs/pipelines.md
jerop: moved the TOC updates to a separate PR: https://github.com/tektoncd/pipeline/pull/3743

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -4855,10 +4875,12 @@ func TestReconcileWithTaskResultsInFinalTasks(t *testing.T) {
...
 		Name: "final-task-2",
+	}, {
+		Name: "final-task-3",
 	}}
 
 	if d := cmp.Diff(expectedSkippedTasks, reconciledRun.Status.SkippedTasks); d != "" {
-		t.Fatalf("Expected to see only one final task (final-task-2) in the list of skipped tasks. Diff: %s", diff.PrintWantGot(d))
+		t.Fatalf("Didn't get the expected list of skipped tasks. Diff: %s", diff.PrintWantGot(d))
 	}
 }
```
bobcatfish, line 4885: nice update!

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -361,10 +404,15 @@ func TestPipelineSpec_Validate_Failure(t *testing.T) {
...
+			Finally: []PipelineTask{{
+				Name:            "invalid-pipeline-task-finally",
+				TaskRef:         &TaskRef{Name: "foo-task"},
+				WhenExpressions: []WhenExpression{{}},
+			}},
 		},
 		expectedError: apis.FieldError{
 			Message: `missing field(s)`,
-			Paths:   []string{"tasks[1].when[0]"},
+			Paths:   []string{"tasks[1].when[0]", "finally[0].when[0]"},
```
bobcatfish, line 415: im a little nervous (which maybe is a sign that im overthinking this!) about all the test here combining the two scenarios, i.e. validation failure in the pipelinetask AND in the finally task.

at least since the path contains both we can know both were caught, but im wondering if it'd be good to also cover a case where only one is set.

since the new cases we want to add are so similar to the existing ones, we have a few choices:

1. Combine the cases together like this PR does
2. Duplicate the cases
3. To be extra sure, do (2) AND do (1) (test them all separately AND together)

I think (3) is overkill, so I'm happy to go with this - I feel like if we do (1) we should do (2) for _at least_ one scenario, or do (2) and have one test that combines the cases like in (1)

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -486,11 +493,14 @@ func validateTasksInputFrom(tasks []PipelineTask) (errs *apis.FieldError) {
 	return errs
 }
 
-func validateWhenExpressions(tasks []PipelineTask) (errs *apis.FieldError) {
+func validateWhenExpressions(tasks []PipelineTask, finalTasks []PipelineTask) (errs *apis.FieldError) {
 	for i, t := range tasks {
 		errs = errs.Also(validateOneOfWhenExpressionsOrConditions(t).ViaFieldIndex("tasks", i))
```
bobcatfish, line 498: maybe this is overkill, but now that we're allowing when expressions in finally, does it make sense to add a test that ensures that conditions aren't allows in when expressions? (just ignore me if this already exists)

pkg/reconciler/pipelinerun/resources/apply_test.go
```diff
@@ -1062,6 +1082,11 @@ func TestApplyTaskRunContext(t *testing.T) {
 				Name:  "task3",
 				Value: *v1beta1.NewArrayOrString("$(tasks.task3.status)"),
 			}},
+			WhenExpressions: v1beta1.WhenExpressions{{
+				Input:    "$(tasks.task1.status)",
+				Operator: selection.In,
+				Values:   []string{"$(tasks.task3.status)"},
```
bobcatfish, line 1088: can we add coverage for results replacements in when expressions via the tests in this file as well? (i only see status and param here, and only see results replacement tested in reconciler, end to end and examples)

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -93,6 +93,7 @@ func ApplyPipelineTaskContext(state PipelineRunState, replacements map[string]st
 		if resolvedPipelineRunTask.PipelineTask != nil {
 			pipelineTask := resolvedPipelineRunTask.PipelineTask.DeepCopy()
 			pipelineTask.Params = replaceParamValues(pipelineTask.Params, replacements, nil)
+			pipelineTask.WhenExpressions = pipelineTask.WhenExpressions.ReplaceWhenExpressionsVariables(replacements)
```
bobcatfish, line 96: I'm a bit confused by this line - this doesn't seem specific to finally tasks, im wondering why pipelinetasks with when expressions currently work without this line? (was there a bug? if so, might make sense to separate that out into a separate commit and/or PR if possible)

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -156,12 +156,20 @@ func (t *ResolvedPipelineRunTask) checkParentsDone(facts *PipelineRunFacts) bool
...
+	if facts.isFinalTask(t.PipelineTask.Name) {
+		if t.whenExpressionsSkip(facts) {
+			return true
+		}
+	}
+
+	if facts.isDAGTask(t.PipelineTask.Name) {
+		if t.conditionsSkip() || t.whenExpressionsSkip(facts) || t.parentTasksSkip(facts) || facts.IsStopping() {
+			return true
+		}
```
bobcatfish, line 172: nit: is it possible to word these as a switch statement? feels weird that facts.isFinalTask can be true and yet facts.isDAGTask(t.PipelineTask.Name) still gets evaluated

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -972,32 +1007,38 @@ func TestIsSkipped(t *testing.T) {
...
 		expected: map[string]bool{
 			"mytask13": false,
 		},
 	}, {
 		name:  "run-cancelled",
-		state: runCancelled,
+		tasks: runCancelled,
 		expected: map[string]bool{
 			"mytask13": false,
 		},
```
bobcatfish, line 1019: similar to above comments about combining cases together, i wonder if it would make sense to add a few new ones here too?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -972,32 +1007,38 @@ func TestIsSkipped(t *testing.T) {
...
 	}} {
 		t.Run(tc.name, func(t *testing.T) {
-			d, err := dagFromState(tc.state)
+			d, err := dagFromState(tc.tasks)
+			if err != nil {
+				t.Fatalf("Could not get a dag from the TC state %#v: %v", tc.tasks, err)
+			}
+			f, err := dagFromState(tc.finally)
```
bobcatfish, line 1026: there's something very odd about getting a "dag" from the finally tasks - but i think thats a whole other story XD

bobcatfish (commented):
Overall looks great!!! All of my comments are pretty much nitpicks or just wondering if we want (EVEN MORE) unit tests

Very exciting to see this coming together of the features you and @pritidesai have been working on (finally, when expressions, status)!!!! 💃

pkg/apis/pipeline/v1beta1/pipeline_validation.go
jerop: > that conditions aren't allows in when expressions 

did you mean "that conditions aren't allowed in finally"? if so, that's already tested in:

https://github.com/tektoncd/pipeline/blob/075910e6a0e7e40fb01530aab05d4ec8ac5433a7/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go#L2001-L2014

docs/pipelines.md
jerop: 🎉

pkg/reconciler/pipelinerun/resources/apply.go
jerop: The `ApplyPipelineTaskContext` function replaces the context variables for execution status. It is written more generally, I guess so it could be easily reused later, but it is currently used on finally tasks only in:

https://github.com/tektoncd/pipeline/blob/08177fab0e2f2d2cc8e30e67f018a552338fca26/pkg/reconciler/pipelinerun/pipelinerun.go#L614-L617

so the line added in this PR is to substitute execution status from pipeline tasks to the when expressions in the finally tasks, so there's no bug

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
jerop: simplified the logic and removed the double checks

test/pipelinefinally_test.go
jerop: shortened them and added comments to describe them

test/pipelinefinally_test.go
jerop: replaced it, thanks!

test/pipelinefinally_test.go
jerop: that's right, replaced with checking the taskrun status

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
jerop: we can't have `finally tasks` without `dag tasks`, so we can't do option 2 with `finally tasks` only

left the `dag tasks` as they are, and added cases with combined `dag tasks` and `finally tasks`

pkg/reconciler/pipelinerun/resources/apply_test.go
jerop: result replacements in when expressions are already tested in this file at: https://github.com/tektoncd/pipeline/blob/33c3a6972691924aa011d15bc7307b5a19775bcb/pkg/reconciler/pipelinerun/resources/apply_test.go#L291-L447

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
jerop: will separate them, waiting for this to get in https://github.com/tektoncd/pipeline/pull/3744

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
jerop: created a PR for the refactor so that we consider them separately: https://github.com/tektoncd/pipeline/pull/3744

jerop:
moved out the testing refactors to a separate PR that we need to get merged first: https://github.com/tektoncd/pipeline/pull/3744

/hold

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
jerop: added new ones instead of combining

test/pipelinefinally_test.go
jerop: @pritidesai what do you think? maybe we can refactor this in a future PR?

test/pipelinefinally_test.go
jerop: refactored to reuse the same function, which tasks the params list as an argument

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: This could be simplified by utilizing `IsFinallySkipped` instead of changing this function.

https://github.com/tektoncd/pipeline/blob/01286cf978750e85609fa0a0e9a1aa177fc724f8/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L220

This skip function is primarily used to build cache of skipped `dag` tasks. This caching was introduced in PR -https://github.com/tektoncd/pipeline/pull/3524

https://github.com/tektoncd/pipeline/blob/01286cf978750e85609fa0a0e9a1aa177fc724f8/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go#L59

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -193,7 +199,7 @@ func (t *ResolvedPipelineRunTask) conditionsSkip() bool {
 }
 
 func (t *ResolvedPipelineRunTask) whenExpressionsSkip(facts *PipelineRunFacts) bool {
-	if t.checkParentsDone(facts) {
+	if facts.isFinalTask(t.PipelineTask.Name) || t.checkParentsDone(facts) {
```
pritidesai, line 202: this change is not needed if we are utilizing `IsFinallySkipped`

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: Just introduce one `if` statement after [line 227](https://github.com/tektoncd/pipeline/blob/01286cf978750e85609fa0a0e9a1aa177fc724f8/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L227) 

```
		if t.whenExpressionsSkip(facts) {
			return true
		}

```

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -939,9 +940,39 @@ func TestIsSkipped(t *testing.T) {
 		expected: map[string]bool{
 			"mytask10": false,
 		},
+	}, {
+		name: "tasks-when-expressions-in-finally-passed",
```
pritidesai, line 944: these tests can be moved to `TestResolvedPipelineRunTask_IsFinallySkipped`

pritidesai: PR - https://github.com/tektoncd/pipeline/pull/3744 will not be needed if we change `IsFinallySkipped`.

jerop:
thanks for the review @pritidesai 🙏🏾 

changed to using `IsFinallySkipped` and closed https://github.com/tektoncd/pipeline/pull/3744

/hold cancel

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
jerop: actually ended up doing (3), but the ones with finally tasks have dag tasks as well (but the dag tasks in some of those cases don't have when expressions)

pkg/reconciler/pipelinerun/resources/apply.go
bobcatfish: ahh i see: since it was only used for finally tasks (and since we dont support status in non finally tasks), got it! thanks for explaining :D

pritidesai:
@jerop `GetSkippedTasks` adds the finally tasks into the list of skippedTasks:

https://github.com/tektoncd/pipeline/blob/98507859fa961f2152b65ad911defc2f0a2ef804/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go#L369-L386

Can we please set WhenExpressions for finally tasks when they are specified?

```
		if rprt.IsFinallySkipped(facts) {
			skippedTask := v1beta1.SkippedTask{
				Name: rprt.PipelineTask.Name,
			}
			if rprt.PipelineTask.WhenExpressions != nil || len(rprt.PipelineTask.WhenExpressions) != 0 {
				skippedTask.WhenExpressions = rprt.PipelineTask.WhenExpressions
			}
			skipped = append(skipped, skippedTask)
		}
```

such that finally tasks being in the list of skipped tasks have the when expressions initialized:

```
    {
        "name": "finally-task-should-be-skipped-1"
      },
      {
        "name": "finally-task-should-be-skipped-2"
      },
      {
        "name": "finally-task-should-be-skipped-3"
      }
```

jerop:
thanks for the review @pritidesai! 🙏🏾 

> Can we please set `WhenExpressions` for finally tasks when one or more expressions are specified?
> 
> ```
> 		if rprt.IsFinallySkipped(facts) {
> 			skippedTask := v1beta1.SkippedTask{
> 				Name: rprt.PipelineTask.Name,
> 			}
> 			if rprt.PipelineTask.WhenExpressions != nil || len(rprt.PipelineTask.WhenExpressions) != 0 {
> 				skippedTask.WhenExpressions = rprt.PipelineTask.WhenExpressions
> 			}
> 			skipped = append(skipped, skippedTask)
> 		}
> ```

makes sense to do it similarly to `Tasks`, but did it a bit differently to ensure we don't include `WhenExpressions` in `Finally Tasks` that were skipped because they referenced missing `Results` (not because their `WhenExpressions` evaluated to `False`):

https://github.com/tektoncd/pipeline/blob/c719b6a12fec2e4d0d5d147b451d7fc991b9638b/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go#L379-L389

didn't check for nil because it doesn't change anything, since it defaults to nil anyway, like in the `Tasks` part of the function:
https://github.com/tektoncd/pipeline/blob/c719b6a12fec2e4d0d5d147b451d7fc991b9638b/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go#L372-L378

pritidesai:
thanks a bunch @jerop, great point 🙇‍♀️

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -4714,6 +4714,71 @@ func TestReconcileWithTaskResultsInFinalTasks(t *testing.T) {
...
+						StringVal: "param",
+					}},
+				},
+				WhenExpressions: v1beta1.WhenExpressions{{
+					Input:    "$(tasks.dag-task-1.results.aResult)",
+					Operator: selection.NotIn,
+					Values:   []string{"aResultValue"},
+				}},
+			}, {
+				// final task executed because when expressions evaluated to false
```
pritidesai, line 4734: NIT: s/executed/skipped/

test/pipelinefinally_test.go
pritidesai: sure we can in a separate PR. This function is useful since its shared between two separate tests. We had the tests implemented without this common function and found a lot of repetition. We can declare a pipeline inline in favor of the readability instead of using this function.

pritidesai:
Thanks a bunch @jerop for this feature.

I am having hard time understanding validation unit tests. I feel we have many overlapping tests. Ideally, would like to see the separate unit tests validating when expression syntax e.g. input and values are initialized, a valid operator is specified, etc. Now once we have validated the syntax, I don't see a need to have separate tests to validate the syntax again as part of either `tasks` or `finally`. Rather validate `when` expressions references (e.g. referencing a task result or defined parameter) with the `tasks` and `finally` sections. This can be refactored and simplified later and will not block this PR.

/lgtm

pkg/reconciler/pipelinerun/pipelinerun_test.go
jerop: good catch, fixed it

jerop:
@pritidesai I agree, we need to refactor the validation testing -- can open an issue to track that

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -630,7 +630,7 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
 	}
 
 	for _, rprt := range nextRprts {
-		if rprt == nil || rprt.Skip(pipelineRunFacts) {
+		if rprt == nil || rprt.Skip(pipelineRunFacts) || rprt.IsFinallySkipped(pipelineRunFacts) {
```
pritidesai, line 633: hey @jerop, based on my understanding, `rprt.Skip(pipelineRunFacts)` returns `false` for any finally task but `rprt.IsFinallySkipped(pipelineRunFacts)` evaluates when expression and returns `true` since the when expression evaluates to `false`.

pritidesai: but this would work, provided `nextRprts` contains the finally tasks 🤣

Issue #3206, opened 2020-09-11
Pipelines with finally clause in them makes tekton-pipelines-webhook spam conversion errors between v1alpha1 and v1beta1


benke33:
# Expected Behavior
Pipelines with `finally:` in them should not make the `tekton-piplines-webhook` spam convert errors.

# Actual Behavior
When applying any Pipeline that includes `finally:` in it, the `tekton-pipelines-webhook` starts spamming these conversion errors:

```json
{
  "level": "info",
  "logger": "webhook",
  "caller": "webhook/conversion.go:42",
  "msg": "Webhook ServeHTTP request=&http.Request{Method:\"POST\", URL:(*url.URL)(0xc000714280), Proto:\"HTTP/1.1\", ProtoMajor:1, ProtoMinor:1, Header:http.Header{\"Accept\":[]string{\"application/json, */*\"}, \"Accept-Encoding\":[]string{\"gzip\"}, \"Content-Length\":[]string{\"2019\"}, \"Content-Type\":[]string{\"application/json\"}, \"User-Agent\":[]string{\"kube-apiserver-admission\"}}, Body:(*http.body)(0xc00084d7c0), GetBody:(func() (io.ReadCloser, error))(nil), ContentLength:2019, TransferEncoding:[]string(nil), Close:false, Host:\"tekton-pipelines-webhook.tekton-pipelines.svc:443\", Form:url.Values(nil), PostForm:url.Values(nil), MultipartForm:(*multipart.Form)(nil), Trailer:http.Header(nil), RemoteAddr:\"100.109.0.3:60412\", RequestURI:\"/resource-conversion?timeout=30s\", TLS:(*tls.ConnectionState)(0xc000788370), Cancel:(<-chan struct {})(nil), Response:(*http.Response)(nil), ctx:(*context.cancelCtx)(0xc00084d800)}",
  "commit": "a162a1d"
}
{
  "level": "info",
  "logger": "webhook",
  "caller": "conversion/conversion.go:133",
  "msg": "Converting [kind=Pipeline group=tekton.dev version=v1beta1] to version tekton.dev/v1alpha1",
  "commit": "a162a1d",
  "uid": "b2b3d9a8-e22b-4724-8105-61ccabaa9bf5",
  "desiredAPIVersion": "tekton.dev/v1alpha1",
  "inputType": "[kind=Pipeline group=tekton.dev version=v1beta1]",
[...]

PR #3757, opened 2021-02-11
Remove error when converting Finally field to alpha1

sbwsg:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/3206

/hold

I'm posting this PR because I don't know what the right thing to do here is.  Looking for feedback from anyone who knows the conversion webhook stuff better than me.

Prior to this commit pipelines returned an error when converting from
a v1beta1 pipeline to a v1alpha1 pipeline if that pipeline had a Finally
section in it. As a result of this error the kubeapi would repeatedly
request v1alpha1 pipelines every second, filling up our webhook logs
with these errors.

[This is the place that kubernetes is repeatedly hitting us with conversion attempts.](https://github.com/kubernetes/apiserver/blob/ab31715e135fe7a7fbcd6f2d91ee25c2c7e6eaf9/pkg/storage/cacher/cacher.go#L383-L387)

This commit removes the error when downgrading a beta1 resource to alpha1.
I actually have no idea if this is the correct thing to do and I can't find
any docs explaining the dos and donts of conversion. This stops the spam
though.

/kind bug

# Submitter Checklist

review them:

# Release Notes

```release-note
Fix issue where webhook logs would rapidly fill up with conversion errors when a pipeline with Finally was applied to the cluster.
```

### DIFF
--- a/pkg/apis/pipeline/v1alpha1/pipeline_conversion.go
+++ b/pkg/apis/pipeline/v1alpha1/pipeline_conversion.go
@@ -100,10 +100,9 @@ func (sink *PipelineSpec) ConvertFrom(ctx context.Context, source v1beta1.Pipeli
 			}
 		}
 	}
-	// finally clause was introduced in v1beta1 and not available in v1alpha1
-	if len(source.Finally) > 0 {
-		return ConvertErrorf(FinallyFieldName, ConversionErrorFieldNotAvailableMsg)
-	}
+	// source.Finally is dropped on the floor because otherwise kube api spams pipelines
+	// webhook repeatedly. See https://github.com/tektoncd/pipeline/issues/3206 for more
+	// info.
 	return nil
 }
 


afrittoli:
I'm happy to be convinced otherwise, but as far as I understand the problem, I think this approach is not an option.
If we returned an incomplete resource to an alpha client and not raise an error, the client would have no way of knowing that the resource is incomplete. It could alter it an push it back to the cluster, effectively deleting parts of it.
Even if the client was aware of the error, it would not have any use of an incomplete resource.

sbwsg:
Capturing all the ideas that have been generated during a lengthy Slack discussion on this topic:

1. Could we disable the kube api's caching of v1alpha1 resources so that we can continue to return errors to v1alpha1 users?  This would be ideal because then those users are aware they've "lost" information in the downgrade.
2. Could we backport Finally to v1alpha1?
3. Could we convert down from beta's Finally to alpha Tasks while preserving the "finally"-ness somehow?
    * Use a `runAfter` in the ex-Finally Tasks so they're scheduled at the end of the v1alpha1 Pipeline
    * Add an annotation on the Pipeline that records the mapping of v1lpha1 Task names to v1beta1 Finally Task names
    * Re-hydrate Finally if `apply`ing a v1alpha1 Pipeline w/ the annotations to the server
4. Could we "poison" the v1alpha1 Pipeline we return so that it's cached without error by the api server but fails validation if the user attempts to re-`apply` the v1alpha1 version?
    * E.g. inject an additional "fake" PipelineTask with an invalid name that will trigger validation errors on submission

sbwsg:
Closing this in favor of #3779

Issue #2725, opened 2020-06-01
Expose retries to the the Task/TaskRun


IdanAdar:
# Expected Behavior
For the value of `retries` to be available at runtime so I could track it and potentially run different code flows depending in the current retry counter.

# Actual Behavior
Currently not exposed.

wlynch:
/kind feature

Hi! Thanks for the feature request!

Can you describe in a little more detail what your use case is? e.g. are you looking for retry information to be exposed to running step containers, or the Tekton controller itself to be used with [Conditions](https://tekton.dev/docs/pipelines/pipelines/#guard-task-execution-using-conditions)?

Thanks! :D

IdanAdar:
I would like to use it in steps, yes. I did not think about exposing it in Conditions, but I don't see why not...

We have some tasks (with a single step) that will be retried on failure. In this case though, depending on the number of attempts done I may want to dispatch a different notification, so knowing the current attempt # will help me accomplish that.

IdanAdar:
@wlynch why?

ImJasonH:
Just to clarify, is the end goal to enable something like this Pipeline?

```yaml
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: retry-param
spec:
  tasks:
  - name: retry-me
    retries: 5
    taskSpec:
      steps:
      - image: busybox
        script: echo 'I am attempt number $(context.retry-count) of $(context.retries)' && exit 1
```

In this example, when the Pipeline is run, it should run 5 TaskRuns, the first one printing `I am attempt 1 of 5`, the last one printing `I am attempt 5 of 5` -- is that roughly the feature you're looking for @IdanAdar ?

(In the example above, the names of the placeholders are made up, we'd probably find a better name for both of them)

bobcatfish:
>  potentially run different code flows depending in the current retry counter.

The idea of changing the behavior of a Task based on the retries seems like a bad idea but it's hard to say without knowing the specific use case.

However making this information available seems pretty reasonable either way? Might be worth bringing up in an API working group at least ( @tektoncd/core-maintainers )

IdanAdar:
For example, we have an e2e tests task that runs continuously.

If it fails, it runs again and again thanks to the number of retries. 

Whenever it fails, it sends a Slack notification. I want it to send the notification only if it fail the defined number of retires. Exposing the number will enable me so.

vdemeester:
[...]

PR #3770, opened 2021-02-18
Add variables context.pipelineTask.retries and context.task.retry-count

yaoxiaoqi:
# Changes

`context.pipelineTask.retries` exposes the total retries of a pipelineTask.

`context.task.retry-count` exposes the current number of retry. The reason why the
variable isn't named `context.pipelineTask.retry-count` is that every task has a retry
count, only pipelineTask has retries.

Rename function `ApplyPipelineTaskContext` to `ApplyPipelineTaskStateContext` in
"pipelinerun/apply.go" to distinguish from `ApplyPipelineTaskContexts`.

related issue: https://github.com/tektoncd/pipeline/issues/2725
TEP: https://github.com/tektoncd/community/blob/main/teps/0039-add-variable-retries-and-retrycount.md

/kind feature

# Submitter Checklist

review them:

## Reviewer Notes

# Release Notes

```release-note
Add variables context.pipelineTask.retries and context.task.retry-count
```

### DIFF
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -20,6 +20,7 @@ weight: 3
     - [Guard `Task` execution using `Conditions`](#guard-task-execution-using-conditions)
     - [Configuring the failure timeout](#configuring-the-failure-timeout)
   - [Using variable substitution](#using-variable-substitution)
+    - [Using the `retries` and `retry-count` variable substitutions](#using-the-retries-and-retry-count-variable-substitutions)
   - [Using `Results`](#using-results)
     - [Passing one Task's `Results` into the `Parameters` or `WhenExpressions` of another](#passing-one-tasks-results-into-the-parameters-or-whenexpressions-of-another)
     - [Emitting `Results` from a `Pipeline`](#emitting-results-from-a-pipeline)
@@ -549,6 +550,35 @@ performed by the Tekton Controller when a PipelineRun is executed.
 
 See the [complete list of variable substitutions for Pipelines](./variables.md#variables-available-in-a-pipeline)
 and the [list of fields that accept substitutions](./variables.md#fields-that-accept-variable-substitutions).
+
+### Using the `retries` and `retry-count` variable substitutions
+
+Tekton supports variable substitution for the [`retries`](#using-the-retries-parameter)
+parameter of `PipelineTask`. Variables like `context.pipelineTask.retries` and
+`context.task.retry-count` can be added to the parameters of a `PipelineTask`.
+`context.pipelineTask.retries` will be replaced by `retries` of the `PipelineTask`, while
+`context.task.retry-count` will be replaced by current retry number of the `PipelineTask`.
+
+```yaml
+params:
+- name: pipelineTask-retries
+  value: "$(context.pipelineTask.retries)"
+taskSpec:
+  params:
+  - name: pipelineTask-retries
+  steps:
+  - image: ubuntu
+    name: print-if-retries-exhausted
+    script: |
+      if [ "$(context.task.retry-count)" == "$(params.pipelineTask-retries)" ]
+      then
+        echo "This is the last retry."
+      fi
+      exit 1
+```
+
+**Note:** Every `PipelineTask` can only access its own `retries` and `retry-count`. These
+values aren't accessible for other `PipelineTask`s.
 
 ## Using `Results`
 
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -25,6 +25,7 @@ For instructions on using variable substitutions see the relevant section of [th
 | `context.pipeline.name` | The name of this `Pipeline` . |
 | `tasks.<pipelineTaskName>.status` | The execution status of the specified `pipelineTask`, only available in `finally` tasks. The execution status can be set to any one of the values (`Succeeded`, `Failed`, or `None`) described [here](pipelines.md#using-execution-status-of-pipelinetask)|
 | `tasks.status` | An aggregate status of all the `pipelineTasks` under the `tasks` section (excluding the `finally` section). This variable is only available in the `finally` tasks and can have any one of the values (`Succeeded`, `Failed`, `Completed`, or `None`) described [here](pipelines.md#using-aggregate-execution-status-of-all-tasks).  |
+| `context.pipelineTask.retries` | The retries of this `PipelineTask`. |
 
 ## Variables available in a `Task`
 
@@ -43,6 +44,7 @@ For instructions on using variable substitutions see the relevant section of [th
 | `context.taskRun.namespace` | The namespace of the `TaskRun` that this `Task` is running in. |
 | `context.taskRun.uid` | The uid of the `TaskRun` that this `Task` is running in. |
 | `context.task.name` | The name of this `Task`. |
+| `context.task.retry-count` | The current retry number of this `Task`. |
 
 ### `PipelineResource` variables available in a `Task`
 
--- a/examples/v1beta1/pipelineruns/using-retries-and-retry-count-variables.yaml
+++ b/examples/v1beta1/pipelineruns/using-retries-and-retry-count-variables.yaml
@@ -1,0 +1,37 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: retry-example-
+spec:
+  serviceAccountName: 'default'
+  pipelineSpec:
+    tasks:
+    - name: retry-me
+      retries: 5
+      params:
+      - name: pipelineTask-retries
+        value: "$(context.pipelineTask.retries)"
+      - name: pipelineTask-retry-count
+        value: "$(context.task.retry-count)"
+      taskSpec:
+        params:
+        - name: pipelineTask-retries
+        - name: pipelineTask-retry-count
+        steps:
+        - image: alpine:3.12.0
+          script: |
+            #!/usr/bin/env sh
+            if [ "$(params.pipelineTask-retry-count)" == "$(params.pipelineTask-retries)" ]; then
+              echo "This is the last retry."
+              exit 0;
+            fi
+            echo "The PipelineTask has retried $(context.task.retry-count) times."
+            exit 1
+    - name: hello-world
+      runAfter:
+      - retry-me
+      taskSpec:
+        steps:
+        - image: busybox
+          script: |
+            echo "hello world"
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -614,7 +614,7 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
 	fnextRprts := pipelineRunFacts.GetFinalTasks()
 	if len(fnextRprts) != 0 {
 		// apply the runtime context just before creating taskRuns for final tasks in queue
-		resources.ApplyPipelineTaskContext(fnextRprts, pipelineRunFacts.GetPipelineTaskStatus())
+		resources.ApplyPipelineTaskStateContext(fnextRprts, pipelineRunFacts.GetPipelineTaskStatus())
 
 		// Before creating TaskRun for scheduled final task, check if it's consuming a task result
 		// Resolve and apply task result wherever applicable, report warning in case resolution fails
@@ -707,6 +707,7 @@ func (c *Reconciler) createTaskRun(ctx context.Context, rprt *resources.Resolved
 		return c.PipelineClientSet.TektonV1beta1().TaskRuns(pr.Namespace).UpdateStatus(ctx, tr, metav1.UpdateOptions{})
 	}
 
+	rprt.PipelineTask = resources.ApplyPipelineTaskContexts(rprt.PipelineTask)
 	taskRunSpec := pr.GetTaskRunSpec(rprt.PipelineTask.Name)
 	tr = &v1beta1.TaskRun{
 		ObjectMeta: metav1.ObjectMeta{
--- a/pkg/reconciler/pipelinerun/resources/apply.go
+++ b/pkg/reconciler/pipelinerun/resources/apply.go
@@ -18,6 +18,7 @@ package resources
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
@@ -68,6 +69,17 @@ func ApplyContexts(spec *v1beta1.PipelineSpec, pipelineName string, pr *v1beta1.
 	return ApplyReplacements(spec, replacements, map[string][]string{})
 }
 
+// ApplyPipelineTaskContexts applies the substitution from $(context.pipelineTask.*) with the specified values.
+// Uses "0" as a default if a value is not available.
+func ApplyPipelineTaskContexts(pt *v1beta1.PipelineTask) *v1beta1.PipelineTask {
+	pt = pt.DeepCopy()
+	replacements := map[string]string{
+		"context.pipelineTask.retries": strconv.Itoa(pt.Retries),
+	}
+	pt.Params = replaceParamValues(pt.Params, replacements, map[string][]string{})
+	return pt
+}
+
 // ApplyTaskResults applies the ResolvedResultRef to each PipelineTask.Params and Pipeline.WhenExpressions in targets
 func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResultRefs) {
 	stringReplacements := resolvedResultRefs.getStringReplacements()
@@ -87,8 +99,8 @@ func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResul
 	}
 }
 
-// ApplyPipelineTaskContext replaces context variables referring to execution status with the specified status
-func ApplyPipelineTaskContext(state PipelineRunState, replacements map[string]string) {
+// ApplyPipelineTaskStateContext replaces context variables referring to execution status with the specified status
+func ApplyPipelineTaskStateContext(state PipelineRunState, replacements map[string]string) {
 	for _, resolvedPipelineRunTask := range state {
 		if resolvedPipelineRunTask.PipelineTask != nil {
 			pipelineTask := resolvedPipelineRunTask.PipelineTask.DeepCopy()
--- a/pkg/reconciler/taskrun/resources/apply.go
+++ b/pkg/reconciler/taskrun/resources/apply.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -107,6 +108,7 @@ func ApplyContexts(spec *v1beta1.TaskSpec, rtr *ResolvedTaskResources, tr *v1bet
 		"context.task.name":         rtr.TaskName,
 		"context.taskRun.namespace": tr.Namespace,
 		"context.taskRun.uid":       string(tr.ObjectMeta.UID),
+		"context.task.retry-count":  strconv.Itoa(len(tr.Status.RetriesStatus)),
 	}
 	return ApplyReplacements(spec, replacements, map[string][]string{})
 }
--- a/examples/v1beta1/pipelineruns/using_context_variables.yaml
+++ b/examples/v1beta1/pipelineruns/using_context_variables.yaml
@@ -7,6 +7,7 @@ spec:
   pipelineSpec:
     tasks:
     - name: task1
+      retries: 7
       params:
       - name: pipeline-uid
         value: "$(context.pipelineRun.uid)"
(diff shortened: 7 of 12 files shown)


sbwsg:
This looks pretty ready to me.  I have two suggested changes:

1. Could you add a section to `docs/pipelineruns.md` that describes usage of the variables and explains that they're only available in the single pipelineTask they refer to?
2. Could you also add an example PipelineRun that uses retries under `examples/v1beta1/pipelineruns`?

yaoxiaoqi:
> 1. Could you add a section to `docs/pipelineruns.md` that describes usage of the variables and explains that they're only available in the single pipelineTask they refer to?

I added a section in `docs/pipelines.md`, since I felt hard to find a right place to insert the section in `docs/pipelineruns.md`. And there are more contexts in `docs/pipelines.md` like what `retries` is and how to use variable substitution. Feel free to let me know if you still want to add a section in `pipelineruns.md`.

pkg/reconciler/taskrun/resources/apply.go
```diff
@@ -97,6 +98,15 @@ func ApplyResources(spec *v1beta1.TaskSpec, resolvedResources map[string]v1beta1
 	return ApplyReplacements(spec, replacements, map[string][]string{})
 }
 
+// ApplyPipelineTaskContexts applies the substitution from $(context.pipelineTask.*) with the specified values.
+// Uses "" as a default if a value is not available.
```
sbwsg, line 102: It looks like `pt.Retries` will be 0 if a value is not provided.  so this will set the variable to `"0"`.  Is that right?

sbwsg:
> @sbwsg I added a section in docs/pipelines.md, since I felt hard to find a right place to insert the section in docs/pipelineruns.md. And there are more contexts in docs/pipelines.md like what retries is and how to use variable substitution. Feel free to let me know if you still want to add a section in pipelineruns.md.

No that's ok - the docs you've added look good to me!

pkg/reconciler/taskrun/resources/apply.go
yaoxiaoqi: Yes, I think zero is the expected value for us. If `retries` of a PipelineTask is set to 5, this means it will run 6 times at most (first run + 5 retries). If a PipelineTask doesn't specify `retries`, it will only run one time. The retry is supposed to be 0(first run + 0 retry) under this circumstance.

examples/v1beta1/pipelineruns/using-retries-and-retry-count-variables.yaml
```diff
@@ -0,0 +1,35 @@
...
+            echo "This is the last retry. \n"
+            exit 0;
+          fi
+          exit 1
+  - name: hello-world
+    runAfter:
+      - retry-me 
+    taskSpec:
+      steps:
+        - image: busybox::
```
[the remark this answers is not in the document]
yaoxiaoqi: oh sorry. mistyped them when i was testing.

pkg/reconciler/taskrun/resources/apply.go
sbwsg: So I think this code comment might not be quite right - it says `""` is the default if a value is not available but it sounds like `"0"` is the default if a value is not available.

sbwsg (approved):

sbwsg:
Thanks for this @yaoxiaoqi ! I left one more small comment but not a blocker.

sbwsg:
It looks like the cluster is so busy it's taking forever for the workspace-in-sidecar sidecar to start up, resulting in the examples test timing out.

docs/pipelines.md
```diff
@@ -548,6 +549,29 @@ performed by the Tekton Controller when a PipelineRun is executed.
 See the [complete list of variable substitutions for Pipelines](./variables.md#variables-available-in-a-pipeline)
 and the [list of fields that accept substitutions](./variables.md#fields-that-accept-variable-substitutions).
 
+### Using the `retries` and `retry-count` variable substitutions
+
+Tekton supports variable substitution for the [`retries`](#using-the-runafter-parameter)
```
pritidesai, line 554: s/using-the-runafter-parameter/using-the-retries-parameter/

sorry, its very similar 😉

docs/pipelines.md
```diff
@@ -548,6 +549,29 @@ performed by the Tekton Controller when a PipelineRun is executed.
...
+parameter of `PipelineTask`. Varibles like `context.pipelineTask.retries` and
```
pritidesai, line 555: s/Varibles/Variables/

pkg/reconciler/taskrun/resources/apply.go
```diff
@@ -97,6 +98,15 @@ func ApplyResources(spec *v1beta1.TaskSpec, resolvedResources map[string]v1beta1
 	return ApplyReplacements(spec, replacements, map[string][]string{})
 }
 
+// ApplyPipelineTaskContexts applies the substitution from $(context.pipelineTask.*) with the specified values.
+// Uses "0" as a default if a value is not available.
+func ApplyPipelineTaskContexts(spec *v1beta1.TaskSpec, pt *v1beta1.PipelineTask) *v1beta1.TaskSpec {
+	replacements := map[string]string{
+		"context.pipelineTask.retries": strconv.Itoa(pt.Retries),
```
pritidesai, line 105: This variable is only limited to a `pipelineTask` with an inlined `taskSpec`. Is that by design? `pipelineTask` does imply that the task has to be `pipelineTask` but `pipelineTask` can either have `taskRef` or `taskSpec` and `retries` are supported with both possibilities.

bobcatfish: my assumption these would be available to use as param values - and using them in an inline spec feels like a bonus

(i dont think Tasks should be able to refer to these values since they need to work outside of the context of pipelines)

yaoxiaoqi: I'm unsure if `pipelineTask` can be specified with `taskRef` either. It's a little bit weird that a variable called `context.pipelineTask.*` in the spec of an individual `Task`.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -727,7 +727,7 @@ func (c *Reconciler) createTaskRun(ctx context.Context, rprt *resources.Resolved
 		// We pass the entire, original task ref because it may contain additional references like a Bundle url.
 		tr.Spec.TaskRef = rprt.PipelineTask.TaskRef
 	} else if rprt.ResolvedTaskResources.TaskSpec != nil {
-		tr.Spec.TaskSpec = rprt.ResolvedTaskResources.TaskSpec
+		tr.Spec.TaskSpec = tresources.ApplyPipelineTaskContexts(rprt.ResolvedTaskResources.TaskSpec, rprt.PipelineTask)
```
pritidesai, line 730: This is throwing me off 😢  Why limit `ApplyPipelineTaskContexts` when a pipelineTask has inlined specification 🤔 

It works great as long as its available as param. Can we please add an example similar to [using_context_variables.yaml
](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/using_context_variables.yaml)


https://github.com/tektoncd/pipeline/blob/01a3388b1d71ff94b421324c93b31d3b55b62a59/examples/v1beta1/pipelineruns/using_context_variables.yaml#L10-L12

I tried using it as a param which fails with:

```
admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: non-existent variable in "echo $(params.pipelineTask-retries)\
```

pkg/reconciler/taskrun/resources/apply.go
pritidesai: please see my [comment](https://github.com/tektoncd/pipeline/pull/3770/files#r601923749) 🆙

sbwsg:
Hey @yaoxiaoqi !  We're currently planning to include this in the 0.23 release going out next early next week (5th or 6th April).  Do you think you'll be able to take a look and address the remaining feedback before then?  If not, no problem, I will move this to the 0.24 milestone instead; let me know.

sbwsg:
Moving this one back to 0.24 to give time to address the remaining feedback.

yaoxiaoqi:
Hi @pritidesai I need your opinion on my previous comment. Thanks!

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -136,6 +146,17 @@ func ApplyReplacements(p *v1beta1.PipelineSpec, replacements map[string]string,
 	return p
 }
 
+func ApplyPipelineTaskReplacement(pt *v1beta1.PipelineTask, replacements map[string]string, arrayReplacements map[string][]string) *v1beta1.PipelineTask {
+	pt = pt.DeepCopy()
+	pt.Params = replaceParamValues(pt.Params, replacements, arrayReplacements)
+	for j := range pt.Conditions {
```
sbwsg, line 152: Do you have use-cases for `retry-count` and `retries` in when expressions and conditions?  I am trying to think of some but am not coming up with anything right now.

If we don't have a strong idea of how they would be used for these fields I suggest only supporting them in `params` for now.

sbwsg (commented):
Hey @yaoxiaoqi sorry I am coming to this review late.

I don't think we should add support for the variable to be used directly in `taskSpec` or `taskRef`.  Instead the var should always be passed down via params.

The reason I think this way is because we don't support any other variables being directly replaced in `taskSpec` like this (as far as I know/remember).  From my pov it's better to match the existing behaviour of variables. In future we could decide to allow direct substitution of variables in `taskSpec` and support it for all vars.

pkg/reconciler/pipelinerun/resources/apply.go
yaoxiaoqi: Sorry, I can't either. Got it. I'll only support the replacement in `params`.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -727,7 +728,7 @@ func (c *Reconciler) createTaskRun(ctx context.Context, rprt *resources.Resolved
...
```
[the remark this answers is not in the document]
yaoxiaoqi: Ok, but I notice some variables like $(context.taskRun.uid) can be directly used in the taskSpec. There is an example to do so [examples/v1beta1/taskruns/using_context_variables.yaml](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/taskruns/using_context_variables.yaml)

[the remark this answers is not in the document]
sbwsg: You're absolutely right about `TaskRuns` with embedded `taskSpecs`.

Here we deal with `Pipeline` -> embedded `PipelineTask` -> embedded `taskSpec`. No other pipeline context variable in this situation gets replaced. Let's look at another example:

```yaml
kind: Pipeline
metadata:
  name: foo
spec:
  tasks:
  - name: t
    taskSpec:
      steps:
      - image: alpine
        script: |
          echo $(context.pipelineRun.uid)
---
kind: PipelineRun
metadata:
  name: foo-run
spec:
  pipelineRef:
    name: foo
```

Trying to access `$(context.pipelineRun.uid)` directly in this way doesn't work. There's an error because the variable did not get replaced in the `taskSpec`:

```
[t : unnamed-0] + context.pipelineRun.uid
[t : unnamed-0] /tekton/scripts/script-0-8tpb7: line 3: context.pipelineRun.uid: not found
```

I do think we should improve this - the pipeline context variables should be usable in the `taskSpec`.  But I think that should happen for all the context variables at once, so possibly in another PR to discuss?

yaoxiaoqi:
Hi @sbwsg , I updated the code. Pls review the PR. Thanks!

docs/variables.md
```diff
@@ -43,6 +43,8 @@ For instructions on using variable substitutions see the relevant section of [th
 | `context.taskRun.namespace` | The namespace of the `TaskRun` that this `Task` is running in. |
 | `context.taskRun.uid` | The uid of the `TaskRun` that this `Task` is running in. |
 | `context.task.name` | The name of this `Task`. |
+| `context.task.retry-count` | The current retry number of this `Task`. |
+| `context.pipelineTask.retries` | The retries of this `PipelineTask`. |
```
sbwsg, line 47: nit: Move this one up to the "Variables available in a Pipeline" section

sbwsg (approved):
~~This looks about ready to merge to me.  I already approved but doing so again.

Thanks @yaoxiaoqi for your patience and for tackling this feature!

@pritidesai now that the variable replacement is removed from the nested PipelineTask `taskSpec` do you agree we can merge?~~

Edit: Sorry, see my reply below - I think we should move the `retry-count` into a param too.

sbwsg:
Oh wait, the `context.task.retry-count` is still available in tasks even though it's a pipeline feature.  I thought the taskrun controller is responsible for performing the retries but it's not - the pipelines controller performs the retry in `pkg/reconciler/pipelinerun/pipelinerun.go` with `createTaskRun()`.

@yaoxiaoqi how do you feel about also moving this in to a param-only variable?

sbwsg:
I went back to the TEP and I can see a bit clearer how confusing this all is.  In [the proposal](https://github.com/tektoncd/community/blob/main/teps/0039-add-variable-retries-and-retrycount.md#proposal) we said two slightly different things:

1. "these 2 variables should only exist in the context of PipelineRun."
    * To me this says that the variables should not be available to individual tasks.
2. "For context.pipelineTask.retry-count, we add a replacement to ApplyContexts function for TaskRun to update the retry-count correctly. Since this happens after the TaskRun status is appended to the retriesStatus."
    * This is clearly where the confusion starts to crop up.  Above we said the variables should only exist in the context of a PipelineRun.  But here we say the taskrun reconciler will replace this variable.

I don't think there's any way to reconcile these two concepts.  They're at odds with each other.  I think we should only supply both variables via params in the PipelineTask.  What does everyone else think? @yaoxiaoqi @pritidesai @bobcatfish

pritidesai:
> I went back to the TEP and I can see a bit clearer how confusing this all is. In [the proposal](https://github.com/tektoncd/community/blob/main/teps/0039-add-variable-retries-and-retrycount.md#proposal) we said two slightly different things:
> 
> 1. "these 2 variables should only exist in the context of PipelineRun."
>    
>    * To me this says that the variables should not be available to individual tasks.
> 2. "For context.pipelineTask.retry-count, we add a replacement to ApplyContexts function for TaskRun to update the retry-count correctly. Since this happens after the TaskRun status is appended to the retriesStatus."
>    
>    * This is clearly where the confusion starts to crop up.  Above we said the variables should only exist in the context of a PipelineRun.  But here we say the taskrun reconciler will replace this variable.

Yup, I agree, it is confusing. So `context.task.retry-count` is also available to someone just running `taskRun` but with it will be set to 0. Like you pointed out above and had brought it up during the TEP [discussion](https://github.com/tektoncd/community/pull/239#discussion_r567147056),  the pipeline controller performs the retry in `pkg/reconciler/pipelinerun/pipelinerun.go` with `createTaskRun()`. I think we have a few options:

1. Have the taskrun controller initialize and populate the retry-count like its implemented in this PR but check if the `taskRun` belongs any `pipelineRun` by checking the labels since the `pipelineRun` labels are propagated to `taskRun`. 

2. Replace `retry-count` with the number of `attempt`s. Retry means to try executing again and does not make sense to initialize the counter while the task is executed the first time. After it fails once, the task is retried and hence the counter is set to 1 while getting executed second time. By replacing `retry-count` with `attempt`, we can have it implemented as part of the pipeline controller instead of taskrun controller. 

any other option?

sbwsg:
Would another option be to have `createTaskRun` perform the retry-count substitution?

```go
func (c *Reconciler) createTaskRun(ctx, rprt, pr, storageBase) {
    tr, _ := c.taskRunLister.TaskRuns(pr.Namespace).Get(rprt.TaskRunName)
    if tr == nil {
        rprt.PipelineTask = resources.ApplyRetryCount(tr, rprt.PipelineTask, 0)
    } else {
        rprt.PipelineTask = resources.ApplyRetryCount(rprt.PipelineTask, len(tr.Status.RetriesStatus))
    }
    // ...
```

pritidesai:
> Would another option be to have `createTaskRun` perform the retry-count substitution?
> 
> ```go
> func (c *Reconciler) createTaskRun(ctx, rprt, pr, storageBase) {
>     logger := logging.FromContext(ctx)
>     tr, _ := c.taskRunLister.TaskRuns(pr.Namespace).Get(rprt.TaskRunName)
>     if tr == nil {
>         rprt.PipelineTask = resources.ApplyRetryCount(rprt.PipelineTask, 0)
>     } else {
>         rprt.PipelineTask = resources.ApplyRetryCount(rprt.PipelineTask, len(tr.Status.RetriesStatus))
>     }
>     // ...
> ```
> 
> Edit: I might be misunderstanding something fundamental that means this wouldn't work?

Yeah I don't this so it would work for this reason (I haven't tried so cannot say for sure 😉 ):

In case of a task being retried, `createTaskRun` is updating the condition in the status, the specification or parameters are not updated.

```go
	tr, _ := c.taskRunLister.TaskRuns(pr.Namespace).Get(rprt.TaskRunName)
	if tr != nil {
		//is a retry
		addRetryHistory(tr)
		clearStatus(tr)
		tr.Status.SetCondition(&apis.Condition{
			Type:   apis.ConditionSucceeded,
			Status: corev1.ConditionUnknown,
		})
		return c.PipelineClientSet.TektonV1beta1().TaskRuns(pr.Namespace).UpdateStatus(ctx, tr, metav1.UpdateOptions{})
	}
```

sbwsg:
@yaoxiaoqi what do you think of taking this approach so that all of the retry-related changes remain in the pipelinerun reconciler?

yaoxiaoqi:
> Would another option be to have `createTaskRun` perform the retry-count substitution?

Hi @sbwsg , things would definitely be much cleaner if we align the replacement behavior of `retries` and `retry-count`. But `retry-count` is a little bit tricky to deal with because it changes (plus 1 ) every time a `taskRun` is retried. As I mentioned in the TEP, I choose to use `$(context.task.retry-count)` instead of `$(context.pipelineTask.retry-count)` because it can replace the variable in the right timing and it works well.

Actually, I considered your option when designing. But the problem is that even though you updated the parameters in pipelineTask, you can't update the params in the taskRun, since we only replace the params once in [this line](https://github.com/tektoncd/pipeline/blob/18f337d027e74659b35f9a84f560162438f0957d/pkg/reconciler/pipelinerun/pipelinerun.go#L720). When the same taskRun retries, we won't replace the params again. 

This means we can't just call `updateStatus` when we find a taskRun needs to retry. We need to update the whole taskRun to ensure the `retry-count` value is correct. I simply tried to update the whole taskRun, but interestingly the `pipelineTask` stopped retrying and the whole `pipelineRun` hung there forever. I'd like to know your opinion. And if you think `retries` and `retry-count` should be aligned and I will try to dig into the problem and try to figure out a solution.

yaoxiaoqi:
> 2. Replace `retry-count` with the number of `attempt`s. Retry means to try executing again and does not make sense to initialize the counter while the task is executed the first time. After it fails once, the task is retried and hence the counter is set to 1 while getting executed second time. By replacing `retry-count` with `attempt`, we can have it implemented as part of the pipeline controller instead of taskrun controller.
>

Hi @pritidesai , I would like to know where I can get the `attempt` of a TaskRun? I didn't find it in the taskRun structure.

sbwsg:
> I simply tried to update the whole taskRun, but interestingly the retry-count sometimes is less than the true value, sometimes is equal to the true value.

Ahhhh ok, I didn't understand that this was happening.  It's really interesting!   I hadn't seen this behaviour before but I can confirm this happens in my testing as well: I added a little bit of code to do the replacement in the PipelineRun reconciler and update TaskRun's `Spec.Params`. I sometimes get the desired number of retries but I also sometimes just get a constant "1".  This is really strange.

sbwsg:
OK, I am happy to go ahead with the `context.task.retry-count` being replaced in the taskrun reconciler unless there are any objections from @bobcatfish or @pritidesai

There's one remaining nit I added to move the `pipelineTask` variable up to the pipelines section of `docs/variables.md` but otherwise this looks good to me.

/approve

yaoxiaoqi:
> There's one remaining nit I added to move the `pipelineTask` variable up to the pipelines section of `docs/variables.md` but otherwise this looks good to me.


Done

pritidesai:
sorry @yaoxiaoqi haven't been able to review this since the latest updates, will review it later in the week 🙏

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -707,6 +707,7 @@ func (c *Reconciler) createTaskRun(ctx context.Context, rprt *resources.Resolved
 		return c.PipelineClientSet.TektonV1beta1().TaskRuns(pr.Namespace).UpdateStatus(ctx, tr, metav1.UpdateOptions{})
 	}
 
+	rprt.PipelineTask = resources.ApplyPipelineTaskContexts(rprt.PipelineTask)
```
pritidesai, line 710: I am happy to merge this as is, but setting the context variable here will only make it available to pipelineTask and will not be able to custom task. Since the `retries` is constant and does not change with the pipeline execution, we can set this while resolving pipeline task.

pritidesai:
/lgtm

thanks @yaoxiaoqi for your patience and all the work 🙏  The changes look much cleaner now 🎉

PR #3779, opened 2021-02-22
Losslessly roundtrip Pipelines with Finally from beta to alpha and back

### LINKED ISSUE #3206: Pipelines with finally clause in them makes tekton-pipelines-webhook spam conversion errors between v1alpha1 and v1beta1

sbwsg:
# Changes

Fixes #3206 

Prior to this commit pipelines returned an error when converting from
a v1beta1 pipeline to a v1alpha1 pipeline if that pipeline had a Finally
section in it. As a result of this error the kubeapi would repeatedly
request v1alpha1 pipelines every second, filling up our webhook logs
with these errors.

This commit allows Finally to be serialized into alpha Pipelines, keeping
the Finally Tasks as an annotation on the resource. If the alpha Pipeline
is then re-applied to the cluster the Finally annotation is rehydrated into
the Finally section of the stored v1beta1 resource.

This is a follow-up to https://github.com/tektoncd/pipeline/pull/3757.  In that PR we simply squashed the error message, but in this PR we squash the error while also retaining the Finally information during conversion roundtrips.

# Submitter Checklist

review them:

# Release Notes

```release-note
v1beta1 Pipelines can now be requested with v1alpha1 version without losing Finally tasks. Applying the returned v1alpha1 version will store the resource as v1beta1 with the Finally section restored to its original state.
```

### DIFF
--- a/pkg/apis/pipeline/v1alpha1/conversion_error.go
+++ b/pkg/apis/pipeline/v1alpha1/conversion_error.go
@@ -34,6 +34,3 @@ const (
 type CannotConvertError = v1beta1.CannotConvertError
 
 var _ error = (*CannotConvertError)(nil)
-
-// convertErrorf creates a CannotConvertError from the field name and format string.
-var convertErrorf = v1beta1.ConvertErrorf
--- a/pkg/apis/pipeline/v1alpha1/pipeline_conversion.go
+++ b/pkg/apis/pipeline/v1alpha1/pipeline_conversion.go
@@ -19,6 +19,7 @@ package v1alpha1
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
@@ -26,6 +27,7 @@ import (
 )
 
 const FinallyFieldName = "finally"
+const finallyAnnotationKey = "tekton.dev/v1beta1Finally"
 
 var _ apis.Convertible = (*Pipeline)(nil)
 
@@ -34,10 +36,29 @@ func (source *Pipeline) ConvertTo(ctx context.Context, obj apis.Convertible) err
 	switch sink := obj.(type) {
 	case *v1beta1.Pipeline:
 		sink.ObjectMeta = source.ObjectMeta
-		return source.Spec.ConvertTo(ctx, &sink.Spec)
+		if err := source.Spec.ConvertTo(ctx, &sink.Spec); err != nil {
+			return err
+		}
+		if source.Annotations != nil {
+			if _, ok := source.Annotations[finallyAnnotationKey]; ok {
+				finally := []v1beta1.PipelineTask{}
+				if err := json.Unmarshal([]byte(source.Annotations[finallyAnnotationKey]), &finally); err != nil {
+					return fmt.Errorf("error converting finally annotation into beta field: %w", err)
+				}
+				if err := v1beta1.ValidatePipelineTasks(ctx, sink.Spec.Tasks, finally); err != nil {
+					return fmt.Errorf("error converting finally annotation into beta field: %w", err)
+				}
+				delete(sink.ObjectMeta.Annotations, finallyAnnotationKey)
+				if len(sink.ObjectMeta.Annotations) == 0 {
+					sink.ObjectMeta.Annotations = nil
+				}
+				sink.Spec.Finally = finally
+			}
+		}
 	default:
 		return fmt.Errorf("unknown version, got: %T", sink)
 	}
+	return nil
 }
 
 func (source *PipelineSpec) ConvertTo(ctx context.Context, sink *v1beta1.PipelineSpec) error {
@@ -81,6 +102,16 @@ func (sink *Pipeline) ConvertFrom(ctx context.Context, obj apis.Convertible) err
 	switch source := obj.(type) {
 	case *v1beta1.Pipeline:
 		sink.ObjectMeta = source.ObjectMeta
+		if len(source.Spec.Finally) != 0 {
+			b, err := json.Marshal(source.Spec.Finally)
+			if err != nil {
+				return err
+			}
+			if sink.ObjectMeta.Annotations == nil {
+				sink.ObjectMeta.Annotations = make(map[string]string)
+			}
+			sink.ObjectMeta.Annotations[finallyAnnotationKey] = string(b)
+		}
 		return sink.Spec.ConvertFrom(ctx, source.Spec)
 	default:
 		return fmt.Errorf("unknown version, got: %T", sink)
@@ -99,10 +130,6 @@ func (sink *PipelineSpec) ConvertFrom(ctx context.Context, source v1beta1.Pipeli
 				return err
 			}
 		}
-	}
-	// finally clause was introduced in v1beta1 and not available in v1alpha1
-	if len(source.Finally) > 0 {
-		return convertErrorf(FinallyFieldName, ConversionErrorFieldNotAvailableMsg)
 	}
 	return nil
 }
--- a/pkg/apis/pipeline/v1alpha1/pipeline_conversion_test.go
+++ b/pkg/apis/pipeline/v1alpha1/pipeline_conversion_test.go
@@ -215,7 +215,6 @@ func TestPipelineConversionFromWithFinally(t *testing.T) {
 			if err := p.ConvertTo(context.Background(), ver); err != nil {
 				t.Errorf("ConvertTo() = %v", err)
 			}
-			// modify ver to introduce new field which causes failure to convert v1beta1 to v1alpha1
 			source := ver
 			source.(*v1beta1.Pipeline).Spec.Finally = []v1beta1.PipelineTask{{Name: "finaltask", TaskRef: &TaskRef{Name: "task"}}}
 			got := &Pipeline{}
@@ -231,7 +230,7 @@ func TestPipelineConversionFromWithFinally(t *testing.T) {
 	}
 }
 
-func TestPipelineConversionFromBetaToAlphaWithFinally_Failure(t *testing.T) {
+func TestPipelineConversionFromBetaToAlphaWithFinally(t *testing.T) {
 	p := &v1beta1.Pipeline{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:       "foo",
@@ -240,18 +239,20 @@ func TestPipelineConversionFromBetaToAlphaWithFinally_Failure(t *testing.T) {
 		},
 		Spec: v1beta1.PipelineSpec{
 			Tasks:   []v1beta1.PipelineTask{{Name: "mytask", TaskRef: &TaskRef{Name: "task"}}},
-			Finally: []v1beta1.PipelineTask{{Name: "mytask", TaskRef: &TaskRef{Name: "task"}}},
+			Finally: []v1beta1.PipelineTask{{Name: "myfinallytask", TaskRef: &TaskRef{Name: "task"}}},
 		},
 	}
-	t.Run("finally not available in v1alpha1", func(t *testing.T) {
-		got := &Pipeline{}
-		if err := got.ConvertFrom(context.Background(), p); err != nil {
-			cce, ok := err.(*CannotConvertError)
-			// conversion error (cce) contains the field name which resulted in the failure and should be equal to "finally" here
-			if ok && cce.Field == FinallyFieldName {
-				return
-			}
-			t.Errorf("ConvertFrom() should have failed")
+	t.Run("finally stored by v1alpha1 and rehydrated for v1beta1", func(t *testing.T) {
+		downgrade := &Pipeline{}
+		if err := downgrade.ConvertFrom(context.Background(), p); err != nil {
+			t.Errorf("error converting from v1beta1 with finally field to v1alpha1 with finally annotation: %v", err)
+		}
+		upgrade := &v1beta1.Pipeline{}
+		if err := downgrade.ConvertTo(context.Background(), upgrade); err != nil {
+			t.Errorf("error converting from v1alpha1 with finally annotation to v1beta1 with finally field: %v", err)
+		}
+		if d := cmp.Diff(p, upgrade); d != "" {
+			t.Errorf("unexpected difference between v1beta1 with finally field and round-tripped v1beta1 with finally field: %s", diff.PrintWantGot(d))
 		}
 	})
 }
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation.go
@@ -49,7 +49,7 @@ func (ps *PipelineSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 		errs = errs.Also(apis.ErrGeneric("expected at least one, got none", "description", "params", "resources", "tasks", "workspaces"))
 	}
 	// PipelineTask must have a valid unique label and at least one of taskRef or taskSpec should be specified
-	errs = errs.Also(validatePipelineTasks(ctx, ps.Tasks, ps.Finally))
+	errs = errs.Also(ValidatePipelineTasks(ctx, ps.Tasks, ps.Finally))
 	// All declared resources should be used, and the Pipeline shouldn't try to use any resources
 	// that aren't declared
 	errs = errs.Also(validateDeclaredResources(ps.Resources, ps.Tasks, ps.Finally))
@@ -76,7 +76,7 @@ func (ps *PipelineSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 
 // validatePipelineTasks ensures that pipeline tasks has unique label, pipeline tasks has specified one of
 // taskRef or taskSpec, and in case of a pipeline task with taskRef, it has a reference to a valid task (task name)
-func validatePipelineTasks(ctx context.Context, tasks []PipelineTask, finalTasks []PipelineTask) *apis.FieldError {
+func ValidatePipelineTasks(ctx context.Context, tasks []PipelineTask, finalTasks []PipelineTask) *apis.FieldError {
 	// Names cannot be duplicated
 	taskNames := sets.NewString()
 	var errs *apis.FieldError
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
@@ -767,7 +767,7 @@ func TestValidatePipelineTasks_Success(t *testing.T) {
 	}}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			err := validatePipelineTasks(context.Background(), tt.tasks, []PipelineTask{})
+			err := ValidatePipelineTasks(context.Background(), tt.tasks, []PipelineTask{})
 			if err != nil {
 				t.Errorf("Pipeline.validatePipelineTasks() returned error for valid pipeline tasks: %v", err)
 			}
@@ -927,7 +927,7 @@ func TestValidatePipelineTasks_Failure(t *testing.T) {
 			if tt.wc != nil {
 				ctx = tt.wc(ctx)
 			}
-			err := validatePipelineTasks(ctx, tt.tasks, []PipelineTask{})
+			err := ValidatePipelineTasks(ctx, tt.tasks, []PipelineTask{})
 			if err == nil {
 				t.Error("Pipeline.validatePipelineTasks() did not return error for invalid pipeline tasks")
 			}


sbwsg:
/hold

This is a proof-of-concept that we can pass Finally back to v1alpha1 without losing information when storing it back to v1beta1.  Open to feedback and questions here, still tidying up and editing tests.

sbwsg:
Also, some more context, Kubernetes' [guidance for CRD versioning](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions) links out to their [API changes guide](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md) wherein they state the following rule:

> It must be possible to round-trip your change (convert to different API versions and back) with no loss of information.

pritidesai:
Thanks @sbwsg for initiating this PR.

I tried creating a sample pipeline in `v1beta1` with `finally` in it but it failed with this error:

```
kubectl apply -f 01-pipeline-with-finally.yaml
Error from server (BadRequest): error when creating "01-pipeline-with-finally.yaml": admission webhook "webhook.pipeline.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "finally"
```

Webhook logs:

```
{"level":"info","ts":"2021-02-22T19:03:10.335Z","logger":"tekton-pipelines-webhook","caller":"webhook/admission.go:131","msg":"remote admission controller audit annotations=map[string]string(nil)","commit":"9731b96","knative.dev/kind":"tekton.dev/v1alpha1, Kind=Pipeline","knative.dev/namespace":"default","knative.dev/name":"01-pipeline-with-final-task","knative.dev/operation":"CREATE","knative.dev/resource":"tekton.dev/v1alpha1, Resource=pipelines","knative.dev/subresource":"","knative.dev/userinfo":"{minikube-user  [system:masters system:authenticated] map[]}","admissionreview/uid":"614e9634-79ce-4e22-84aa-7f7ee90a7316","admissionreview/allowed":false,"admissionreview/result":"&Status{ListMeta:ListMeta{SelfLink:,ResourceVersion:,Continue:,RemainingItemCount:nil,},Status:Failure,Message:mutation failed: cannot decode incoming new object: json: unknown field \"finally\",Reason:BadRequest,Details:nil,Code:400,}"}
```

I am looking into your changes next.

sbwsg:
Here's an example manual test to try with this PR's branch:

```bash
# Apply a v1beta1 pipeline with finally section
$ k apply -f ./examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml

# Get the v1alpha1 rendering of that pipeline
$ kubectl get pipelines.v1alpha1.tekton.dev clone-cleanup-workspace -o yaml  > test.yaml

# Open test.yaml in your text editor of choice.
# Edit to have a new name.
# Edit to remove the UUID from the metadata.

# Re-apply the v1alpha1 yaml
k apply -f ./test.yaml
```

Now when you `kubectl get` the pipeline stored from your v1alpha1 yaml you'll see the `finally` list is still available.

pritidesai:
> I tried creating a sample pipeline in v1beta1 with finally in it but it failed with this error:

Sorry, it was my bad 😞 , the sample pipeline I was using was invalid (`finally` with `v1alpha1` 😜 )

This works great with annotations having the entire finally section in it:

```
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"tekton.dev/v1beta1","kind":"Pipeline","metadata":{"annotations":{},"name":"clone-cleanup-workspace","namespace":"default"},"spec":{"finally":[{"name":"cleanup","taskRef":{"name":"cleanup-workspace"},"workspaces":[{"name":"source","workspace":"git-source"}]},{"name":"check-git-commit","params":[{"name":"commit","value":"$(tasks.clone-app-repo.results.commit)"}],"taskSpec":{"params":[{"name":"commit"}],"steps":[{"image":"alpine","name":"check-commit-initialized","script":"if [[ ! $(params.commit) ]]; then\n  exit 1\nfi\n"}]}}],"tasks":[{"name":"clone-app-repo","params":[{"name":"url","value":"https://github.com/tektoncd/community.git"},{"name":"subdirectory","value":"application"}],"taskRef":{"name":"git-clone-from-catalog"},"workspaces":[{"name":"output","workspace":"git-source"}]}],"workspaces":[{"name":"git-source"}]}}
    tekton.dev/v1beta1Finally: '[{"name":"cleanup","taskRef":{"name":"cleanup-workspace","kind":"Task"},"workspaces":[{"name":"source","workspace":"git-source"}]},{"name":"check-git-commit","taskSpec":{"metadata":{},"params":[{"name":"commit","type":"string"}],"steps":[{"name":"check-commit-initialized","image":"alpine","resources":{},"script":"if
      [[ ! $(params.commit) ]]; then\n  exit 1\nfi\n"}]},"params":[{"name":"commit","value":"$(tasks.clone-app-repo.results.commit)"}]}]'
```

And verified its possible to round-trip.

But I am not sure if this is the right solution though (I can approve these since at least the log spamming is stopped with this PR 😄 ).

My concerns:

1. We are getting into this kind of annotations with every new field we add in `beta`(at least until we have `v1`).
2. I just tried creating `beta` pipeline with when expressions and listing it in `pipelines.v1alpha1.tekton.dev`. The conversion silently (without any error) ignores `when` expressions and the converted pipeline does not include any of those expressions.

**beta**:

```
kind: Pipeline
metadata:
  name: 01-pipeline-with-when
spec:
  tasks:
    - name: dag-task
      when:
        - input: "foo"
          operator: in
          values: [ "foo" ]
      taskSpec:
        steps:
          - name: success
            image: ubuntu
            script: |
              exit 0
```

**converted to alpha**:

```
spec:
  tasks:
  - name: dag-task
    taskSpec:
      steps:
      - image: ubuntu
        name: success
        resources: {}
        script: |
          exit 0
```

And, we haven't heard any complains on these missing when expressions which kind of tells us that the users are not dependent on this conversion.

We will have to fix when expressions if we go this route.

3. How are we going to handle conversions from `v1` to `v1beta1`? Will there be expectation of having such annotations in `v1` resources?

afrittoli:
> > I tried creating a sample pipeline in v1beta1 with finally in it but it failed with this error:
> 
> Sorry, it was my bad 😞 , the sample pipeline I was using was invalid (`finally` with `v1alpha1` 😜 )
> 
> This works great with annotations having the entire finally section in it:
> 
> ```
> metadata:
>   annotations:
>     kubectl.kubernetes.io/last-applied-configuration: |
>       {"apiVersion":"tekton.dev/v1beta1","kind":"Pipeline","metadata":{"annotations":{},"name":"clone-cleanup-workspace","namespace":"default"},"spec":{"finally":[{"name":"cleanup","taskRef":{"name":"cleanup-workspace"},"workspaces":[{"name":"source","workspace":"git-source"}]},{"name":"check-git-commit","params":[{"name":"commit","value":"$(tasks.clone-app-repo.results.commit)"}],"taskSpec":{"params":[{"name":"commit"}],"steps":[{"image":"alpine","name":"check-commit-initialized","script":"if [[ ! $(params.commit) ]]; then\n  exit 1\nfi\n"}]}}],"tasks":[{"name":"clone-app-repo","params":[{"name":"url","value":"https://github.com/tektoncd/community.git"},{"name":"subdirectory","value":"application"}],"taskRef":{"name":"git-clone-from-catalog"},"workspaces":[{"name":"output","workspace":"git-source"}]}],"workspaces":[{"name":"git-source"}]}}
>     tekton.dev/v1beta1Finally: '[{"name":"cleanup","taskRef":{"name":"cleanup-workspace","kind":"Task"},"workspaces":[{"name":"source","workspace":"git-source"}]},{"name":"check-git-commit","taskSpec":{"metadata":{},"params":[{"name":"commit","type":"string"}],"steps":[{"name":"check-commit-initialized","image":"alpine","resources":{},"script":"if
>       [[ ! $(params.commit) ]]; then\n  exit 1\nfi\n"}]},"params":[{"name":"commit","value":"$(tasks.clone-app-repo.results.commit)"}]}]'
> ```
> 
> And verified its possible to round-trip.
> 
> But I am not sure if this is the right solution though (I can approve these changes since at least the log spamming is stopped with this PR 😄 ).
> 
> My concerns:
> 
> 1. We are getting into this kind of annotations with every new field we add in `beta`(at least until we have `v1`).
> 2. I just tried creating `beta` pipeline with when expressions and listing it in `pipelines.v1alpha1.tekton.dev`. The conversion silently (without any error) ignores `when` expressions and the converted pipeline does not include any of those expressions.
> 
> **beta**:
> 
> ```
> kind: Pipeline
> metadata:
>   name: 01-pipeline-with-when
> spec:
>   tasks:
>     - name: dag-task
>       when:
>         - input: "foo"
>           operator: in
>           values: [ "foo" ]
>       taskSpec:
>         steps:
>           - name: success
>             image: ubuntu
>             script: |
>               exit 0
> ```
> 
> **converted to alpha**:
> 
> ```
> spec:
>   tasks:
>   - name: dag-task
>     taskSpec:
>       steps:
>       - image: ubuntu
>         name: success
>         resources: {}
>         script: |
>           exit 0
> ```
> 
> And, we haven't heard any complains on these missing when expressions which kind of tells us that the users are not dependent on this conversion.
> 
> We will have to fix when expressions if we go this route.

Yes, will have to add annotations for every new field we *add* (or that we added in the past).

To be honest the alternatives we've considered until now do not seem better:
- stop serving `v1alpha1`: we could perhaps do that, but that won't fix the issue for feature API versions, so we'd still need to find a solution there
- back-port new fields to old API versions: in my opinion this goes against the whole idea of API versions, but I'd be happy to be proven wrong

I'm not entirely sure what the format of these annotation should be, we could discuss whether we want to have one per field, one per api version, or perhaps event store the complete resource as it's stored in an annotation. Things might get complicated if we add a new field with a new field for instance...

> 
> 1. How are we going to handle conversions from `v1` to `v1beta1`? Will there be expectation of having such annotations in `v1` resources?

I think so, unless we find a better solution.

afrittoli:
Question: would one be able to create resources with these annotations? So effectively create `v1beta1` resources through the `v1alpha1` API? I think the answer is yes, so we would need to run validation on the part of the resource which is converted from the annotation. That might be already the case with this PR, I'm not entirely sure.

afrittoli:
Another idea on how to adding fields to API versions.
Let's say we serve `v1beta1` and `v1`. To add a new field to `v1` we would:

- add the new field to `v1`
- create `v1beta2` with the new field back-ported
- stop serving `v1beta1`

Issues with this approach:

- Creating new versions is probably more work
- If we create a new version for every new field, we may end up creating too much api version churn
- We probably cannot just stop serving `v1beta1`
- Using these revamped old versions (`v1beta2`) would require users to change their resources... if they do so, they might as well change to the most recent API version instead. 

Possible solutions:
- we could use annotations, and once we reach a critical mass of new fields make a new version
- instead of stopping to serve old versions right away, we could keep them around for a while with annotations
- we could provide a `v1beta1` -> `v1beta2` migration script: migrating resources from `v1beta1` to `v1beta2` would be very easy to automate, simply replace the api version on the top

afrittoli:
> Another idea on how to adding fields to API versions.
> Let's say we serve `v1beta1` and `v1`. To add a new field to `v1` we would:
> 
> * add the new field to `v1`
> * create `v1beta2` with the new field back-ported
> * stop serving `v1beta1`
> 
> Issues with this approach:
> 
> * Creating new versions is probably more work
> * If we create a new version for every new field, we may end up creating too much api version churn
> * We probably cannot just stop serving `v1beta1`
> * Using these revamped old versions (`v1beta2`) would require users to change their resources... if they do so, they might as well change to the most recent API version instead.
> 
> Possible solutions:
> 
> * we could use annotations, and once we reach a critical mass of new fields make a new version
> * instead of stopping to serve old versions right away, we could keep them around for a while with annotations
> * we could provide a `v1beta1` -> `v1beta2` migration script: migrating resources from `v1beta1` to `v1beta2` would be very easy to automate, simply replace the api version on the top

@vdemeester I think these processes we should define and document as part of the feature gates TEP

sbwsg:
> Question: would one be able to create resources with these annotations? So effectively create v1beta1 resources through the v1alpha1 API? I think the answer is yes, so we would need to run validation on the part of the resource which is converted from the annotation. That might be already the case with this PR, I'm not entirely sure.

Good point, I was surprised by this - validation is not performed on the converted resource.  So upgrading from v1alpha1 w/ invalid finally annotation did not result in error.

I've pushed a change so that the Finally is validated after unmarshalling from json annotation.

sbwsg:
For reference I want looking for any other examples where an annotation is used to store upgrade/downgrade information. The kubernetes Deployment type stores `deprecated.deployment.rollback.to` in an annotation going from `apps` to `v1`: https://github.com/kubernetes/kubernetes/blob/6cacfe9e3a54ca9d532f9f085adfb6455d3b8c82/pkg/apis/apps/v1/conversion.go#L69-L78
    * The deprecated value type is an int64
    * The upgraded value type is a struct with the int64 in a nested field.

dibyom:
@sbwsg what do you think about storing the entire v1beta1 spec instead of just just the finally field in a special annotation? That way we don't have to keep adding new annotations as we keep adding new fields (e.g. when expressions).

sbwsg:
> @sbwsg what do you think about storing the entire v1beta1 spec instead of just just the finally field in a special annotation? That way we don't have to keep adding new annotations as we keep adding new fields (e.g. when expressions).

I think that makes a lot of sense!

If a user makes changes to the v1alpha1 yaml, do we accept those changes or ignore them when re-hydrating the v1beta1 version from the annotation?

souleb:
A question concerning the Finally section issue. 

From my understanding, if new feature is introduced it is in the following order v1alpha -> v1beta -> V1. If correct then new features would always be introduced first in the alpha version of v1 until we freeze it for v2?

dibyom:
> Edit: I'd probably lean toward copying the v1alpha1 fields directly over the top of the rehydrated v1beta1 fields, so keeping the changes.

Yeah, that sounds good to me!

afrittoli:
> > Edit: I'd probably lean toward copying the v1alpha1 fields directly over the top of the rehydrated v1beta1 fields, so keeping the changes.
> 
> Yeah, that sounds good to me!

Yeah, I think we'd have to, otherwise we'd expose a readonly v1alpha1.
I guess that means that one might make changes to the annotation too, and effectively submit changes to in the `v1beta1` part too, via the annotations.

pritidesai:
I am trying to understand why webhook conversions are necessary. 

Kuberenetes [doc](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions) on specifying multiple version explains the conversion can be set to `None` when the schema is identical for all versions and just need the `apiVersion` set to appropriate value :

```
conversion:
    # None conversion assumes the same schema for all versions and only sets the apiVersion
    # field of custom resources to the proper value
    strategy: None
``` 

We do not have identical schemas and that's the reason, we implemented conversion.

Kubernetes [doc](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion) on webhooks conversion lists three reasons and one of them is:

```
custom resource PUT request is in a different version than storage version.
```

We always store the resources in `v1beta1` which implies that even resources created in `v1alpha1` are converted and stored in `v1beta1`. When we migrate to `v1`, all resources created in `v1beta1` will be converted and stored in `v1`. So even though, users are requesting to create resources in `v1alpha1`, they are always converted up and stored in the highest version. 

https://github.com/tektoncd/pipeline/blob/7a204f0e463747e9fff26c97d01a5b0e217eb809/config/300-pipeline.yaml#L49

Now, when it comes to utilizing these resources in a run referenced by `taskRef` or `pipelineRef`, what version is the controller using? `v1alpha1` or `v1beta1`? I didnt find usage or enforcing `APIVersion` while retrieving pipeline/task definition. This field (`APIVersion`) is supported in `PipelineRef` struct but not enforced by the reconciler:

https://github.com/tektoncd/pipeline/blob/0078cfe72cd202bb6cc244615339a0b747e8708f/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L212

Screen Shot 2021-02-24 at 10 04 04 AM

Screen Shot 2021-02-24 at 10 03 28 AM

Does this imply that `v1alpha1` run is pulling `task` and/or `pipeline` references in the latest stored version which is `v1beta1` in our case?

afrittoli:
> I am trying to understand why webhook conversions are necessary.
> 
> Kuberenetes [doc](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions) on specifying multiple version explains the conversion can be set to `None` when the schema is identical for all versions and just need the `apiVersion` set to appropriate value :
> 
> ```
> conversion:
>     # None conversion assumes the same schema for all versions and only sets the apiVersion
>     # field of custom resources to the proper value
>     strategy: None
> ```
> 
> We do not have identical schemas and that's the reason, we implemented conversion.
> 
> Kubernetes [doc](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#webhook-conversion) on webhooks conversion lists three reasons and one of them is:
> 
> ```
> custom resource PUT request is in a different version than storage version.
> ```
> 
> We always store the resources in `v1beta1` which implies that even resources created in `v1alpha1` are converted and stored in `v1beta1`. When we migrate to `v1`, all resources created in `v1beta1` will be converted and stored in `v1`. So even though, users are requesting to create resources in `v1alpha1`, they are always converted up and stored in the highest version.
> 
> https://github.com/tektoncd/pipeline/blob/7a204f0e463747e9fff26c97d01a5b0e217eb809/config/300-pipeline.yaml#L49
> 
> Now, when it comes to utilizing these resources in a run referenced by `taskRef` or `pipelineRef`, what version is the controller using? `v1alpha1` or `v1beta1`? I didnt find usage or enforcing `APIVersion` while retrieving pipeline/task definition. This field (`APIVersion`) is supported in `PipelineRef` struct but not enforced by the reconciler:
> 
> https://github.com/tektoncd/pipeline/blob/0078cfe72cd202bb6cc244615339a0b747e8708f/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L212
> 
> Screen Shot 2021-02-24 at 10 04 04 AM
> 
> Screen Shot 2021-02-24 at 10 03 28 AM
> 
> Does this imply that `v1alpha1` run is pulling `task` and/or `pipeline` references in the latest stored version which is `v1beta1` in our case?

The reason conversions are required is that we can only store one version while we serve multiple ones (which versions we store / serve may be different for each CRD). That is true for both read and writes.

If someone is requesting `v1alpha1`, when the stored version is `v1beta1`, we need to convert the `v1beta1` to `v1alpha1` to do that - or return an error if not possible, like we do today.
If someone it trying to submit a `v1alpha1` when the stored version is `v1beta1`, we need to convert it first to `v1beta1` first.

From tekton side we only expose a mutating webhook; we create a `MutatingWebhookConfiguration` to tell the k8s API server to invoke our webhook for convertions:

```
 k get mutatingwebhookconfiguration
NAME                          WEBHOOKS   AGE
webhook.pipeline.tekton.dev   1          18h
webhook.triggers.tekton.dev   1          18h
```

pritidesai:
Thanks @afrittoli 

> If someone is requesting v1alpha1, when the stored version is v1beta1, we need to convert the v1beta1 to v1alpha1 to do that - or return an error if not possible, like we do today.

I am trying to understand who is this `someone` here? I don't see Tekton controller enforcing specific versions while retrieving definitions. Feel free to ignore me if these questions are annoying 🤣  🤦‍♀️

afrittoli:
> Thanks @afrittoli
> 
> > If someone is requesting v1alpha1, when the stored version is v1beta1, we need to convert the v1beta1 to v1alpha1 to do that - or return an error if not possible, like we do today.
> 
> I am trying to understand who is this `someone` here? I don't see Tekton controller enforcing specific versions while retrieving definitions. Feel free to ignore me if these questions are annoying 🤣 🤦‍♀️

It's a very valid question. A client can request a specific API version - if I'm not mistake the API version is part of the URL. 
An old client (for instance an old version of `tkn` or someone using an old version of the client library in pipeline) could try to list all Tasks in the v1alpha1 API for instance. We discovered this issue because the cacher module in k8s tries to list all resources for all served API versions, thus causing the spamming. I hope this answers your question?

pritidesai:
thanks @afrittoli yup, I understand a client (`kubectl`, `tkn`, `kubernetes/client-go`, or something else) can request a specific API version. Also, I have documented the k8s cacher module problem [here](https://github.com/tektoncd/pipeline/issues/3206#issuecomment-777995928).

My concern is the reconciler is fetching `pipeline/task` definition using `Tektonclient.TektonV1beta1()`

https://github.com/tektoncd/pipeline/blob/93e368dbcb014e8650a6ba8ad01ca3c5f6eb6554/pkg/reconciler/taskrun/resources/taskref.go#L121

https://github.com/tektoncd/pipeline/blob/93e368dbcb014e8650a6ba8ad01ca3c5f6eb6554/pkg/reconciler/pipelinerun/resources/pipelineref.go#L94

Is this valid and by design? Would this pull resources in the version they are created? or version has no impact at this point?

vdemeester:
> thanks @afrittoli yup, I understand a client (`kubectl`, `tkn`, `kubernetes/client-go`, or something else) can request a specific API version. Also, I have documented the k8s cacher module problem [here](https://github.com/tektoncd/pipeline/issues/3206#issuecomment-777995928).
> 
> My concern is the reconciler is fetching `pipeline/task` definition using `Tektonclient.TektonV1beta1()`
> 
> https://github.com/tektoncd/pipeline/blob/93e368dbcb014e8650a6ba8ad01ca3c5f6eb6554/pkg/reconciler/taskrun/resources/taskref.go#L121
> 
> https://github.com/tektoncd/pipeline/blob/93e368dbcb014e8650a6ba8ad01ca3c5f6eb6554/pkg/reconciler/pipelinerun/resources/pipelineref.go#L94
> 
> Is this valid and by design? Would this pull resources in the version they are created? or version has no impact at this point?

The controller works with what is stored, and thus works with v1beta1 for `Pipeline`/`Task` object yes, it is by design and valid. We don't really, at the controller level, need to care about "what version they are created", we really *only* work with the version we store. Querying `v1beta1` there is to not have to go through the conversion as we know there is a version of these resource in v1beta1 because however they were created from, they were converted to v1beta1 and stored as such. So yeah, *version doesn't have an impact at this point* (in the controller).

sbwsg:
> A question concerning the Finally section issue.
> 
> From my understanding, if new feature is introduced it is in the following order v1alpha -> v1beta -> V1. If correct then new features would always be introduced first in the alpha version of v1 until we freeze it for v2?

We haven't done a very good job of following this kind of promotion for our features.  There are some which we've put behind feature flags and some we haven't, but even then it seems to me a user could still hit problems.  A flagged feature that doesn't successfully "roundtrip" through older versions could still be a source of error or information loss.

Solidifying this process as part of the Feature Gate work (as @afrittoli suggested above) makes sense to me.

pritidesai:
Not sure if we have discussed using `last-applied-configuration` instead of adding tekton specific annotation (`tekton.dev/v1beta1finally`):

```
k get pipeline.v1alpha1.tekton.dev 01-pipeline-with-final-task -o yaml
kind: Pipeline
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"tekton.dev/v1beta1","kind":"Pipeline","metadata":{"annotations":{},"name":"01-pipeline-with-final-task","namespace":"default"},"spec":{"finally":[{"name":"final-task","taskSpec":{"steps":[{"image":"ubuntu","name":"success","script":"exit 0\n"}]}}],"tasks":[{"name":"dag-task","taskSpec":{"steps":[{"image":"ubuntu","name":"success","script":"exit 0\n"}]}}]}}
    tekton.dev/v1beta1Finally: '[{"name":"final-task","taskSpec":{"metadata":{},"steps":[{"name":"success","image":"ubuntu","resources":{},"script":"exit 0\n"}]}}]
```

If we can not go that route, would it be possible to have a different versions of specs in `last-applied-configuration` and `tekton.dev/v1beta1Finally`? Should we be concern about the differences at all?

pritidesai:
With every new API version we introduce, the lower API versions are required to implement conversion routines i.e. `ConvertTo` the stored version/s and `ConvertFrom` the served versions. While introducing `v1beta1`, `v1alpha1` implemented a function converting the alpha resources to `v1beta1` and a function converting from beta resources to `v1alpha1`.

Just to get an overall idea with conversion process if we add `v1beta2` and `v1` (stored by default) (while also serving `v1alpha1` and `v1beta1`): 

| API version | ConvertTo | ConvertFrom | served | stored |
| ---- | --- | --- | --- | --- |
| v1alpha1 |  v1 | v1beta1, v1beta2, v1 | true | false |
| v1beta1 |  v1 |  v1beta2, v1 | true | false |
| v1beta2 |  v1 |  v1 | true | false |
| v1 |  None |  None | true | true |

Compared to what we have today (serving `v1alpha1` and `v1beta1` while storing `v1beta1`):

| API version | ConvertTo | ConvertFrom | served | stored |
| ---- | --- | --- | --- | --- |
| v1alpha1 |  v1beta1 | v1beta1 | true | false |
| v1beta1 |  None |  None | true | true |

sbwsg:
> Just to get an overall idea with conversion process if we add v1beta2 and v1 (stored by default) (while also serving v1alpha1 and v1beta1):

I had no idea the lower versions had compounding `ConvertFrom` responsibilities like that.  I assumed k8s would call `ConvertFrom` for each step in the chain v1 -> v1beta2 -> v1beta1 -> v1alpha1.  Interesting!

sbwsg:
Update from yesterday's API WG call: it sounded like most folks agreed that putting the whole v1beta1 object into an annotation on the v1alpha1 yaml is likely the best route to go down.  We're going to look at this again in next week's API WG as the discussion spilled over time this week.

In the meantime I'm going to update this PR to implement that approach and see if there are any immediate gotchas.  I'll add a test for this with a very large document to see if anything explodes along the way.

pritidesai:
sorry, couldn't resist sharing, this tutorial has a great example of round tripping two separate API versions and explanation on conversion: https://youtu.be/AAxuEPIzHUQ?t=2994

vdemeester (approved):

pritidesai:
/lgtm

Thanks a bunch @sbwsg for adding this fix. As per the discussion on slack, we are limiting the fix to a pipeline with `finally` for now. We will revisit this in future to address all the new fields added in `beta` and API conversions in general.

(there are conflicts and will need to resolve them 😉  :)

munnerz:
I'm very late to the party on this, but happy to see it fixed :)

> back-port new fields to old API versions: in my opinion this goes against the whole idea of API versions, but I'd be happy to be proven wrong

I see this was mentioned as an option - this is in fact the *correct* solution to this problem, and what is recommended upstream.

The purpose of API versioning/creating a new API version is to permit authors of APIs to make *breaking* changes (i.e., renaming a field) to the schema of an API version. It is *not* to denote what functionality is supported in a given API.

When adding a net-new field, it is perfectly valid to add this within the same API version (without creating anything new). This is the approach upstream APIs taken regularly, and [cert-manager](https://github.com/jetstack/cert-manager) has been taking as well.

Doing it this way will also avoid huge amounts of annotations being added in future for other fields as you add them to the latest API version (and also stop you needing to cut v2, v3, v4 etc whenever you add new functionality 😄)

PR #3816, opened 2021-03-08
Serialize and deserialize Finally in PipelineRuns too

sbwsg:
# Changes

In commit 3ea5981 we added a way for Pipeline.Spec.Finally to be serialized into
alpha Pipelines when downgrading from v1beta1. This was done to fix an
error occurring during conversion and to avoid potential data loss.

This commit adds a similar feature to PipelineRuns. Specifically, if a
PipelineRun nests a PipelineSpec with Finally then that too needs to be
serialized and stored on the alpha version to be deserialized later.

/kind bug

# Submitter Checklist

review them:

# Release Notes

```release-note
Fix issue where PipelineRun with finally in nested pipeline spec would lose those finally tasks when converted down to v1alpha1 and back up to v1beta1 again.
```

### DIFF
--- a/pkg/apis/pipeline/v1alpha1/pipeline_conversion.go
+++ b/pkg/apis/pipeline/v1alpha1/pipeline_conversion.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/pkg/apis"
 )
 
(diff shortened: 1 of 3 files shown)


pritidesai (approved):
The changes looks fine but I am trying to understand the usage here. Since the specifications are inlined, it can not be used by any other `run`. These changes discloses the `finally` section when listed the `run` in `v1alpha1`.

vdemeester (approved):
/lgtm

sbwsg:
> The changes looks fine but I am trying to understand the usage here. Since the specifications are inlined, it can not be used by any other run. These changes discloses the finally section when listed the run in v1alpha1.

This PR was just about matching the behaviour of `PipelineRun.PipelineSpec.spec` with `Pipeline.spec` because I felt like that would be an odd sharp edge users could hit.

PR #3881, opened 2021-04-14
Implement TEP-0033: Add enable-api-fields feature-flag

sbwsg:
# Changes

Implements TEP-0033:
https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md

This commit introduces a new feature flag, "enable-api-fields", that
can be set to "alpha" or "stable". The default is "stable".

This commit provides:

- dev documentation describing how to check the feature gate and use the provided tools during testing
- a helper for performing feature gate validation in our validating webhook
- a test helper, `requireGate` that allows an integration test to be easily skipped if "enable-api-fields" is not "alpha".
- a way to run YAML examples specific to each feature-gate:

New examples can be added under a directory called "alpha". When
the "enable-api-fields" flag is set to "alpha" the test runner will
_only_ run those tests in "alpha" directories. Similarly if the flag is
"stable" then the examples under "alpha" will be prevented from running.
See the Step and Sidecar Workspaces branch for examples that use this
new mechanism:

- https://github.com/sbwsg/pipeline/commit/ba4ca8d18e5a0c6a3ffc3e71424b95385314a5cf#diff-b33f1e031894c58e1c3f02fd3291bc14bdae76333d9c19a3b672b88736821a6b
- https://github.com/sbwsg/pipeline/commit/ba4ca8d18e5a0c6a3ffc3e71424b95385314a5cf#diff-71bc2dcaacd303b560c738c3f1fb2e4eb35ebe628e8cf53038f7b85c63ccfb5d
- https://github.com/sbwsg/pipeline/commit/ba4ca8d18e5a0c6a3ffc3e71424b95385314a5cf#diff-568a92e9e9f73030ce0dfbb15adf76f4e65e7e28aef0474421ba5b415911f9c7

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
A new feature-flag, "enable-api-fields" has been added. Valid values are "alpha" and "stable". It defaults to "stable".

Setting the "enable-api-fields" flag tells Tekton what level of API stability you require in the cluster. As new features are added to Pipelines we'll first place those features under the "alpha" flag. When the feature is ready we'll promote it to "stable". This process is described in TEP-0033[1].

Opting in to "alpha" gives you early exposure to exciting new features as they're added to Pipelines but those features are still undergoing development and could be subject to backwards-incompatible changes.

[1] https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md
```

### DIFF
--- a/pkg/apis/config/feature_flags.go
+++ b/pkg/apis/config/feature_flags.go
@@ -20,11 +20,14 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 )
 
 const (
+	StableAPIFields                         = "stable"
+	AlphaAPIFields                          = "alpha"
 	disableHomeEnvOverwriteKey              = "disable-home-env-overwrite"
 	disableWorkingDirOverwriteKey           = "disable-working-directory-overwrite"
 	disableAffinityAssistantKey             = "disable-affinity-assistant"
@@ -33,6 +36,7 @@ const (
 	requireGitSSHSecretKnownHostsKey        = "require-git-ssh-secret-known-hosts" // nolint: gosec
 	enableTektonOCIBundles                  = "enable-tekton-oci-bundles"
 	enableCustomTasks                       = "enable-custom-tasks"
+	enableAPIFields                         = "enable-api-fields"
 	DefaultDisableHomeEnvOverwrite          = false
 	DefaultDisableWorkingDirOverwrite       = false
 	DefaultDisableAffinityAssistant         = false
@@ -41,6 +45,7 @@ const (
 	DefaultRequireGitSSHSecretKnownHosts    = false
 	DefaultEnableTektonOciBundles           = false
 	DefaultEnableCustomTasks                = false
+	DefaultEnableAPIFields                  = StableAPIFields
 )
 
 // FeatureFlags holds the features configurations
@@ -54,6 +59,7 @@ type FeatureFlags struct {
 	RequireGitSSHSecretKnownHosts    bool
 	EnableTektonOCIBundles           bool
 	EnableCustomTasks                bool
+	EnableAPIFields                  string
 }
 
 // GetFeatureFlagsConfigName returns the name of the configmap containing all
@@ -105,7 +111,26 @@ func NewFeatureFlagsFromMap(cfgMap map[string]string) (*FeatureFlags, error) {
 	if err := setFeature(enableCustomTasks, DefaultEnableCustomTasks, &tc.EnableCustomTasks); err != nil {
 		return nil, err
 	}
+	if err := setEnabledAPIFields(cfgMap, DefaultEnableAPIFields, &tc.EnableAPIFields); err != nil {
+		return nil, err
+	}
 	return &tc, nil
+}
+
+// setEnabledAPIFields sets the "enable-api-fields" flag based on the content of a given map.
+// If the feature gate is invalid or missing then the flag is set to its default.
+func setEnabledAPIFields(cfgMap map[string]string, defaultValue string, feature *string) error {
+	value := defaultValue
+	if cfg, ok := cfgMap[enableAPIFields]; ok {
+		value = strings.ToLower(cfg)
+	}
+	switch value {
+	case AlphaAPIFields, StableAPIFields:
+		*feature = value
+	default:
+		return fmt.Errorf("invalid value for feature flag %q: %q", enableAPIFields, value)
+	}
+	return nil
 }
 
 // NewFeatureFlagsFromConfigMap returns a Config for the given configmap
--- a/pkg/apis/config/testdata/feature-flags-invalid-boolean.yaml
+++ b/pkg/apis/config/testdata/feature-flags-invalid-boolean.yaml
@@ -1,4 +1,4 @@
-# Copyright 2020 The Tekton Authors
+# Copyright 2021 The Tekton Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -18,11 +18,4 @@ metadata:
   name: feature-flags
   namespace: tekton-pipelines
 data:
-  disable-home-env-overwrite: "true"
-  disable-working-directory-overwrite: "true"
-  disable-affinity-assistant: "true"
-  running-in-environment-with-injected-sidecars: "false"
-  require-git-ssh-secret-known-hosts: "true"
-  enable-tekton-oci-bundles: "true"
-  enable-custom-tasks: "true"
-  enable-api-fields: "alpha"
+  "disable-creds-init": "im-really-not-a-valid-boolean"
--- a/pkg/apis/pipeline/v1beta1/version_validation.go
+++ b/pkg/apis/pipeline/v1beta1/version_validation.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2021 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+	"knative.dev/pkg/apis"
+)
+
+// ValidateEnabledAPIFields checks that the enable-api-fields feature gate is set
+// to the wantVersion value and, if not, returns an error stating which feature
+// is dependent on the version and what the current version actually is.
+func ValidateEnabledAPIFields(ctx context.Context, featureName, wantVersion string) *apis.FieldError {
+	currentVersion := config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields
+	if currentVersion != wantVersion {
+		var errs *apis.FieldError
+		message := fmt.Sprintf(`%s requires "enable-api-fields" feature gate to be %q but it is %q`, featureName, wantVersion, currentVersion)
+		return errs.Also(apis.ErrGeneric(message, "workspaces"))
+	}
+	var errs *apis.FieldError = nil
+	return errs
+}
--- a/test/gate.go
+++ b/test/gate.go
@@ -1,0 +1,27 @@
+package test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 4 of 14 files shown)


test/gate.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/system"
+)
+
+// requireGate returns a setup func that will skip the current
+// test if the feature-flag with given name does not equal
+// given value. It will fatally fail the test if it cannot get
+// the feature-flag configmap.
+func requireGate(name, value string) func(context.Context, *testing.T, *clients, string) {
```
vdemeester, line 16: 😍

vdemeester (commented):
Looks very good to me 🙃

sbwsg:
TODO:
- [ ] Add words to DEVELOPMENT.md to tell devs how to use these things when writing new features
- [ ] Figure out how to conditionally Validate alpha features / deny submission when an alpha feature is used in "stable" environment.

sbwsg:
/hold

Still work to be done on this.

sbwsg:
Updated with a validation function that can be used by alpha API fields to guard them from being used when `enable-api-fields` is `stable`.

sbwsg:
And now also updated with developer docs explaining how to use these new tools while writing feature-guarded code.  I think this is ready for review!

vdemeester (commented):
/lgtm

pkg/apis/config/feature_flags.go
```diff
@@ -105,9 +111,25 @@ func NewFeatureFlagsFromMap(cfgMap map[string]string) (*FeatureFlags, error) {
...
+func setEnabledAPIFields(cfgMap map[string]string, defaultValue string, feature *string) {
+	value := defaultValue
+	if cfg, ok := cfgMap[enableAPIFields]; ok {
+		value = strings.ToLower(cfg)
+	}
+	switch value {
+	case AlphaAPIFields, StableAPIFields:
+		*feature = value
+	default:
+		*feature = defaultValue
```
pritidesai, line 129: I don't have a strong opinion here (I like the way its implemented) but we are ignoring any value other than `stable` and `alpha` and defaulting to `stable` which is different from `bool` feature flags.  I think we should at least log this somewhere that its being defaulted to `stable`. On the other side, we could be ignoring a spelling mistake when someone tried to set it to `alpha` and by default we are operating `stable` APIs.

vdemeester: Right, I think it makes sense to at least log something in case of the value not being one we support, *but* the user will have to check the logs of the controller or the webhook (and know what to look for) to see it. But agreed we should log it anyway.

sbwsg: Indeed, I didn't want our machinery to end up in a broken state because the user mis-typed a value in the configmap.  But adding a log line is a great idea; I'll do that.

sbwsg: Hm.  We don't have access to `context` in a lot of places that leverage this code.  This might be a bigger change than I thought.

sbwsg: @pritidesai @vdemeester would you be ok with me using the go `log` package to print this message? The reason is that `NewFeatureFlagsFromConfigMap` is [passed into knative and called by the framework](https://github.com/tektoncd/pipeline/blob/879ba4f65732808a0614d76d5993af0bac736009/pkg/apis/config/store.go#L83) without a `ctx` variable. I can't think of a way to shim support for the logger into that function without completely rewriting how the config package is structured.  Since that function then [calls `NewFeatureFlagsFromMap`](https://github.com/tektoncd/pipeline/blob/879ba4f65732808a0614d76d5993af0bac736009/pkg/apis/config/feature_flags.go#L113), which in turn calls `setEnabledAPIFields` (from this PR), I don't see a great way to handle this except for leaning on the go `log` package.

Any other ideas on ways we could tackle this?

test/gate.go
```diff
@@ -0,0 +1,27 @@
...
+	return func(ctx context.Context, t *testing.T, c *clients, namespace string) {
```
souleb, line 17: namespace parameter seems not used. Is it here because required in the Setup() declaration?

pkg/apis/config/feature_flags.go
sbwsg: From the owners meeting just now:

1. We could return an error in a similar fashion as we do when boolean flags aren't correctly set to "true" / "false"
2. Knative pkg may be able to help us out here - they may have some configmap validation webhook tooling we can leverage to validate these on submission

test/gate.go
sbwsg: Yes you're exactly right - [the setup func expects functions with signature](https://github.com/tektoncd/pipeline/blob/879ba4f65732808a0614d76d5993af0bac736009/test/init_test.go#L52) matching:

```go
func(context.Context, *testing.T, *clients, string) // the string at the end here is namespace
```

pkg/apis/config/feature_flags.go
sbwsg: I've gone ahead and implemented (1) - returning an error if the value is mis-spelled.  Created https://github.com/tektoncd/pipeline/issues/3895 to investigate validating the configmap contents somehow.

pkg/apis/config/testdata/feature-flags-invalid-boolean.yaml
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: feature-flags
+  namespace: tekton-pipelines
+data:
+  "disable-creds-init": "im-really-not-a-valid-boolean"
```
pritidesai, line 21: 🤣  👍

pkg/apis/pipeline/v1beta1/version_validation.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+	"knative.dev/pkg/apis"
+)
+
+// ValidateEnabledAPIFields checks that the enable-api-fields feature gate is set
+// to the wantVersion value and, if not, returns an error stating which feature
+// is dependent on the version and what the current version actually is.
+func ValidateEnabledAPIFields(ctx context.Context, featureName, wantVersion string) *apis.FieldError {
```
pritidesai, line 30: @sbwsg I am trying to understand the usage of this function, its only referenced in the unit tests for now. How do we envision this being utilized in future?

test/path_filtering.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+type pathFilter func(string) bool
+
+// getPathFilter returns a pathFilter that filters out examples
+// unsuitable for the current feature-gate. For example,
+// if the enable-api-fields feature flag is currently set
+// to "alpha" then all "stable" and "beta" examples would be
+// excluded. Similarly when the flag is set to "stable",
+// all "alpha" and "beta" examples are filtered out.
```
pritidesai, line 36: I am trying to understand this, please bear with me.

If I am reading this right, the feature flag value is considered as absolute and only allows examples of that particular version to be tested.

But based on the TEP, setting the feature gate to `alpha` means the controller must support `alpha` and `stable` both:

| Feature Versions ->  | beta | alpha |
| ---  | --- | --- |
| stable | x | |
| alpha | x | x |

x == "**enabled**"

Should we run stable examples with alpha feature gate?  Am I mixing things up here? 🙃  Is this meant to work with something else? 🤔

pkg/apis/pipeline/v1beta1/version_validation.go
sbwsg: Ah, good catch, there's an example of usage over in the Step & Sidecar Workspaces PR: https://github.com/tektoncd/pipeline/pull/3700/files#diff-988b00fa4f188a96ff5901c0bbc9f65d4a8355417185d4fe0ae97d8c14e04417R131-R133

The below snippet ensures that the new field is only valid to use when `enable-api-fields` is `alpha`:

```go
	for stepIdx, step := range steps {
		if len(step.Workspaces) != 0 {
			errs = errs.Also(ValidateEnabledAPIFields(ctx, "step workspaces", config.AlphaAPIFields).ViaIndex(stepIdx).ViaField("steps"))
		}
```

test/path_filtering.go
sbwsg: I think you're absolutely right here.  I've updated the func to run both alpha and stable examples when enable-api-fields is "alpha".

test/path_filtering.go
```diff
@@ -0,0 +1,87 @@
...
+	if enableAPIFields == "" {
+		cmd := exec.Command("kubectl", "get", "configmap", "feature-flags", "-n", "tekton-pipelines", "-o", `jsonpath="{.data['enable-api-fields']}"`)
+		output, err := cmd.Output()
+		if err != nil {
+			return "", fmt.Errorf("error getting feature-flags configmap: %v", err)
+		}
+		output = bytes.TrimSpace(output)
+		output = bytes.Trim(output, "\"")
+		if len(output) == 0 {
+			enableAPIFields = "stable"
```
pritidesai, line 70: I think we are missing `return` here. The `enableAPIFields` will be overwritten by the next statement even if output is empty.

pritidesai: continue... instead of returning `"stable"`

sbwsg: Thank you. I need test code for my test code >.<  This should be fixed now, I've changed it to:

```go
if len(output) == 0 {
  output = []byte("stable")
}
```

pritidesai:
thanks a bunch @sbwsg for all your hard work in implementing this, great work 👍 

/approve

test/gate.go
waveywaves: Neat !

waveywaves:
This looks great @sbwsg :smile: 
Can't wait to use it to move https://github.com/tektoncd/pipeline/pull/3857 behind a feature gate !
Also, is there a concensus/guidelines around when/how a feature 'graduates' from alpha to stable ? These guidelines might be a little bit obvious with respect to testing rigour and community acceptability, but it would be nice to have these covered in a doc somewhere probably.

sbwsg:
lgtm :pray

sbwsg:
> Also, are there any guidelines around when/how a feature 'graduates' from alpha to stable ? These guidelines might be a little bit obvious with respect to testing rigour and community acceptability, but it would be nice to have these covered in a doc somewhere probably.

Great question!  At the moment the [TEP leaves it to the project OWNERS discretion](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#promotion-to-beta-and-beyond):

> This will likely require the feature being available for at least one release if not several releases (and maybe use in dogfooding) to be able to collect and handle user feedback.

sbwsg:
@waveywaves would you be able to switch your approve to an lgtm?  This would put the PR into a mergeable state since Priti already added the approve.  Thanks a lot for reviewing!

waveywaves:
I dunno if it will work, but I'll give it a go.

/lgtm

test/path_filtering.go
```diff
@@ -0,0 +1,87 @@
+// +build examples
```
afrittoli, line 1: We don't have unit tests for this module

PR #3883, opened 2021-04-14
Bump knative to 0.22 🧙

### LINKED ISSUE #3720: Race condition on startup when using configuring controller via configmaps
# Expected Behavior

When  a user [configures the Pipeline controller's behavior via configmaps](https://github.com/tektoncd/pipeline/blob/master/docs/install.md#customizing-basic-execution-parameters), e.g. disabling the affinity assistant or enabling custom tasks, the controller should always respect these configured values.

When the controller tries to use these configured values, it should always reflect the settings in the config maps if they exist.

# Actual Behavior

We are using [an object provided by knative/pkg called configStore](https://github.com/knative/pkg/blob/a2394e88959df2ca1900855474ae43d7da0a9046/codegen/cmd/injection-gen/generators/reconciler_reconciler.go#L278-L280) to access feature flags (configuration provided by users via config maps). It used to be that we loaded these values once on startup but in https://github.com/tektoncd/pipeline/pull/2637 we switched over to using these config store objects so we could hot reload changes made to the config maps the while the controller is running.

1. Controller is started via [main](https://github.com/tektoncd/pipeline/blob/0fddf2ede69a76045117ecfaf55e1977f41e7e08/cmd/controller/main.go#L60)
2. main calls knative's [sharedmain.MainWithConfig](https://github.com/tektoncd/pipeline/blob/0fddf2ede69a76045117ecfaf55e1977f41e7e08/cmd/controller/main.go#L112)
3. [MainWithConfig](https://github.com/tektoncd/pipeline/blob/0fddf2ede69a76045117ecfaf55e1977f41e7e08/vendor/knative.dev/pkg/injection/sharedmain/main.go#L159) [starts the config watchers](https://github.com/tektoncd/pipeline/blob/0fddf2ede69a76045117ecfaf55e1977f41e7e08/vendor/knative.dev/pkg/injection/sharedmain/main.go#L206-L212)
4. `cmw.Start` is [InformedWatcher.Start](https://github.com/tektoncd/pipeline/blob/0fddf2ede69a76045117ecfaf55e1977f41e7e08/vendor/knative.dev/pkg/configmap/informer/informed_watcher.go#L130)
5. Start calls [registerCallbackAndStartInformer](https://github.com/tektoncd/pipeline/blob/0fddf2ede69a76045117ecfaf55e1977f41e7e08/vendor/knative.dev/pkg/configmap/informer/informed_watcher.go#L141)
6. That calls [sharedInformerFactory.Start](https://github.com/tektoncd/pipeline/blob/0fddf2ede69a76045117ecfaf55e1977f41e7e08/pkg/client/resource/informers/externalversions/factory.go#L111-L121)
7. Ultimately this is where the interesting bit happens, all the informers are run in separate go routines:
https://github.com/tektoncd/pipeline/blob/0fddf2ede69a76045117ecfaf55e1977f41e7e08/pkg/client/resource/informers/externalversions/factory.go#L115-L120

The gorutine started in (7) ([client-go cache controller's Run I think?](https://github.com/tektoncd/pipeline/blob/0fddf2ede69a76045117ecfaf55e1977f41e7e08/vendor/k8s.io/client-go/tools/cache/controller.go#L121)) is what ([waves hands](https://github.com/knative/pkg/issues/1907#issuecomment-733254238)) ultimately leads to reading the configured values from the config maps and storing them such that when the tekton controller tries to read these values (via [FromContextOrDefaults](https://github.com/tektoncd/pipeline/blob/0fddf2ede69a76045117ecfaf55e1977f41e7e08/pkg/apis/config/store.go#L45-L61)) the idea is that it will see the values populated by the goroutine.

The problem is that there is no synchronization so sometimes the controller tries to read those values before the goroutines have populated them.

This causes tests of this functionality to sometimes fail, but more importantly, there is a chance (a very small chance) this could happen to users.

0. Configure a feature via config map
1. Start the controller
2. Immediately try to start a taskrun that uses the feature you have configured
3. Some tiny eensy teensy percentage of the time, you might see that the configured value isn't used and the controller uses its hard coded default instead

(3) would happen SO rarely that you might never be able to reproduce it - see https://github.com/tektoncd/pipeline/issues/2815#issuecomment-733207368 on how to reproduce this via the integration tests

# Additional Info

This is the bug in knative/pkg describing the problem and requesting it to be fixed or a way to work around it: https://github.com/knative/pkg/issues/1960

I think this is the route cause of flakey failures of:
- TestReconcile_ExplicitDefaultSA (https://github.com/tektoncd/pipeline/issues/2815)
- TestReconcile_FeatureFlags (https://github.com/tektoncd/pipeline/issues/3587)
- TestReconcile_FeatureFlags/disable-working-dir-overwrite (https://github.com/tektoncd/pipeline/issues/3351)

## What to do??

While we definitely want to fix this issue, I think that letting our tests continue to flake because of this issue is reducing the quality of the signal we're getting from our tests. Now that we know this is a problem, we don't need the tests to keep failing the way they do.

But because the details of this are hidden in unexported knative/pkg code I'm not sure how to work around it. Some ideas, all terrible:

1. Disable the impacted tests <-- not great b/c then if something ELSE around this functionality breaks, we won't know
2. Retry the impacted tests <-- probably better than (1) but still not a guarantee they won't flake
3. Add an arbitrary sleep into the impacted tests between when they instantiate the controller and when they use the value <-- not _guaranteed_ to fix the problem but since chances of it happening are so small, it might be enough
4. Do the same as (3) but do it in the actual controller <-- pretty awful, and would make controller start up take longer, but _might_ "fix" the actual problem for users too

vdemeester:
# Changes

Keeping knative up-to-date.
Fixes #3720

/kind misc

/cc @afrittoli @sbwsg 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Update knative dependency to 0.22, and default minimum kubernetes version supported, v1.18.0
```


ImJasonH (commented):
/approve

afrittoli:
This will require upgrading our infra clusters :)

vdemeester:
> This will require upgrading our infra clusters :)

@afrittoli are we still running 1.17 ?

afrittoli:
> > This will require upgrading our infra clusters :)
> 
> @afrittoli are we still running 1.17 ?

Yes, I updated them to `1.17` in December, so it's now time to a `1.18` update

bobcatfish:
Fixes #3720 🎉  :D

sbwsg:
You'll need to rebase on top of `main` to get the integration tests passing.

1. `git fetch upstream`
2. `git rebase upstream/main`
3. `git push --force <your remote> <branch name>`

hack/update-deps.sh
```diff
@@ -22,7 +22,7 @@ source $(git rev-parse --show-toplevel)/vendor/github.com/tektoncd/plumbing/scri
 
 cd ${REPO_ROOT_DIR}
 
-VERSION="master"
+VERSION="release-0.22"
```
pritidesai, line 25: 🎉

pritidesai:
thank you @vdemeester 

/lgtm

PR #3893, opened 2021-04-19
Allow dot character in resource names

### LINKED ISSUE #3891: Why does a Tekton resource cannot contain a dot?
I see this validation being enforced on Tekton resources: https://github.com/tektoncd/pipeline/blob/879ba4f65732808a0614d76d5993af0bac736009/pkg/apis/validate/metadata.go#L31-L36 A dot is a a perfectly valid character within the resource name as per [Kubernetes docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/). Why Tekton does not allow for dots in resource names?

Dots are generally very useful, for example in cases where you would like to have versions as part of the name.

psschwei:
# Changes
Fixes #3891 to allow dot (`.`) characters in resource names.

The basic approach here was to update `pkg/apis/validate/metadata.go` so that it checks if Tekton resource names meet the criteria of [DNS Subdomain names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names) (`.` is a valid character in a DNS subdomain). It does this by using the `IsDNS1123Subdomain()` function from `k8s.io/apimachinery/pkg/util/validation`. 

There had previously been a 63 character limit on resource names that, to my understanding, was put in place due to the fact that Tekton pod names are composed based on `pipelineRun + pipeline + pipelineTask`. For reference, a DNS subdomain can be up to 253 characters in length.

Some updates were made to the tests as well:
* tests that checked for failure if a `.` was found in the resource name were updated to instead check for a comma
* several tests used camelCase in the `metadata.name` field -- capital letters are not valid for `metadata.name`, and the new validation method will throw errors on any resource name in camelCase. I've updated the tests to use all lowercase letters for resource names. As this PR has been steadily increasing in size, I'll add new tests for camelCase in a separate PR
* `examples/v1beta1/pipelineruns/pipelinerun.yaml` was updated to include resource names with a `.` in it, in order to catch any future issues that may arise with resource names containing a `.`

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Allow resource names to contain the dot character (".") 
Resource names now validated using the common `validation.IsDNS1123Subdomain()` function
```

### DIFF
--- a/pkg/apis/pipeline/v1alpha1/condition_validation_test.go
+++ b/pkg/apis/pipeline/v1alpha1/condition_validation_test.go
@@ -56,10 +56,10 @@ func TestCondition_Invalid(t *testing.T) {
 	}{{
 		name: "invalid meta",
 		cond: &v1alpha1.Condition{
-			ObjectMeta: metav1.ObjectMeta{Name: "invalid.,name"},
+			ObjectMeta: metav1.ObjectMeta{Name: "invalid,name"},
 		},
 		expectedError: apis.FieldError{
-			Message: "Invalid resource name: special character . must not be present",
+			Message: `invalid resource name "invalid,name": must be a valid DNS label`,
 			Paths:   []string{"metadata.name"},
 		},
 	}, {
--- a/pkg/apis/pipeline/v1alpha1/run_validation_test.go
+++ b/pkg/apis/pipeline/v1alpha1/run_validation_test.go
@@ -36,20 +36,18 @@ func TestRun_Invalid(t *testing.T) {
 		want *apis.FieldError
 	}{{
 		name: "missing spec",
-		run:  &v1alpha1.Run{},
-		want: apis.ErrMissingField("spec"),
-	}, {
-		name: "invalid metadata",
 		run: &v1alpha1.Run{
-			ObjectMeta: metav1.ObjectMeta{Name: "run.name"},
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "temp",
+			},
 		},
-		want: &apis.FieldError{
-			Message: "Invalid resource name: special character . must not be present",
-			Paths:   []string{"metadata.name"},
-		},
+		want: apis.ErrMissingField("spec"),
 	}, {
 		name: "missing ref",
 		run: &v1alpha1.Run{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "temp",
+			},
 			Spec: v1alpha1.RunSpec{
 				Ref: nil,
 			},
@@ -58,6 +56,9 @@ func TestRun_Invalid(t *testing.T) {
 	}, {
 		name: "missing apiVersion",
 		run: &v1alpha1.Run{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "temp",
+			},
 			Spec: v1alpha1.RunSpec{
 				Ref: &v1alpha1.TaskRef{
 					APIVersion: "",
@@ -68,6 +69,9 @@ func TestRun_Invalid(t *testing.T) {
 	}, {
 		name: "missing kind",
 		run: &v1alpha1.Run{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "temp",
+			},
 			Spec: v1alpha1.RunSpec{
 				Ref: &v1alpha1.TaskRef{
 					APIVersion: "blah",
@@ -79,6 +83,9 @@ func TestRun_Invalid(t *testing.T) {
 	}, {
 		name: "non-unique params",
 		run: &v1alpha1.Run{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "temp",
+			},
 			Spec: v1alpha1.RunSpec{
 				Ref: &v1alpha1.TaskRef{
 					APIVersion: "blah",
@@ -111,6 +118,9 @@ func TestRun_Valid(t *testing.T) {
 	}{{
 		name: "no params",
 		run: &v1alpha1.Run{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "temp",
+			},
 			Spec: v1alpha1.RunSpec{
 				Ref: &v1alpha1.TaskRef{
 					APIVersion: "blah",
@@ -122,6 +132,9 @@ func TestRun_Valid(t *testing.T) {
 	}, {
 		name: "unnamed",
 		run: &v1alpha1.Run{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "temp",
+			},
 			Spec: v1alpha1.RunSpec{
 				Ref: &v1alpha1.TaskRef{
 					APIVersion: "blah",
@@ -132,6 +145,9 @@ func TestRun_Valid(t *testing.T) {
 	}, {
 		name: "unique params",
 		run: &v1alpha1.Run{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "temp",
+			},
 			Spec: v1alpha1.RunSpec{
 				Ref: &v1alpha1.TaskRef{
 					APIVersion: "blah",
@@ -149,6 +165,9 @@ func TestRun_Valid(t *testing.T) {
 	}, {
 		name: "valid workspace",
 		run: &v1alpha1.Run{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "temp",
+			},
 			Spec: v1alpha1.RunSpec{
 				Ref: &v1alpha1.TaskRef{
 					APIVersion: "blah",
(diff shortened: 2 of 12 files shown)


pkg/apis/pipeline/v1alpha1/condition_validation_test.go
```diff
@@ -54,15 +54,6 @@ func TestCondition_Invalid(t *testing.T) {
 		cond          *v1alpha1.Condition
 		expectedError apis.FieldError
 	}{{
-		name: "invalid meta",
-		cond: &v1alpha1.Condition{
-			ObjectMeta: metav1.ObjectMeta{Name: "invalid.,name"},
```
ImJasonH, line 59: I would have expected `,` to still be invalid...

psschwei: Good catch -- I didn't notice the comma when I was removing tests. Since that's not a Tekton restriction but rather a [Kubernetes one](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names), do you think we should keep this test? Or potentially update the other tests to check for a comma instead of a period (using say `bad,name` instead of `bad.name`)?

pritidesai:
@ImJasonH after dropping the restriction for a `dot` character, the `metadata_validation.go` would have a single check validating the length of the resource name. Is there any more validation needed here or should the length validation be part of `metadata_validation.go`?

ImJasonH:
> @ImJasonH after dropping the restriction for a `dot` character, the `metadata_validation.go` would have a single check validating the length of the resource name. Is there any more validation needed here or should the length validation be part of `metadata_validation.go`?

Keeping it in a shared function seems useful, just to document what validation means right now. I don't feel that strongly though.

sbwsg:
Could I request we also add an example yaml in `examples/v1beta1/pipelineruns` that includes a task with the new name format and a pipeline with the new name format, referenced by `taskRef` and `pipelineRef`?  This way we're exercising the new naming format end-to-end and if we accidentally break `taskRef` pointing to names with "." in future we'd catch it early.

afrittoli:
> Could I request we also add an example yaml in `examples/v1beta1/pipelineruns` that includes a task with the new name format and a pipeline with the new name format, referenced by `taskRef` and `pipelineRef`? This way we're exercising the new naming format end-to-end and if we accidentally break `taskRef` pointing to names with "." in future we'd catch it early.

Good idea!
Would it be acceptable to use the dot in the name of some existing example, to avoid adding an entire new E2E tests for testing this alone?

psschwei:
I can do either option (add a new example or edit an existing one)... if we go with the latter, would the main [pipelinerun.yaml](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/pipelinerun.yaml) one be a suitable candidate? Thinking I would update to use `unit.tests` task and `demo.pipeline` pipeline...

sbwsg:
Absolutely, updating an existing one to use the format sounds good to me too.

sbwsg:
The integration tests are failing due to a version change in the cluster that gets spun up for the tests to run in.  Folks in slack are exploring the problem atm.

pritidesai:
> Keeping it in a shared function seems useful, just to document what validation means right now. I don't feel that strongly though.

Thanks @ImJasonH, yup definitely. Sorry for confusion, I meant do we need the validation checking the length of the name or rely on Kubernetes API to validate the name just like the other three criteria (1) contain only lowercase alphanumeric characters or '-', and (2 & 3) start/end with an alphanumeric character)?

I feel the validation is half baked here (if we want to keep it) and missing imposing the other restrictions. Also, the way `pipelineTask` names are validated using `validation.IsDNS1123Label()`, we could use the same function instead of hard coding the length 🙃 .

https://github.com/tektoncd/pipeline/blob/879ba4f65732808a0614d76d5993af0bac736009/pkg/apis/pipeline/v1beta1/pipeline_types.go#L256-L264

ImJasonH:
Ah my mistake, yeah, let's standardize on `IsDNS1123Label` if that's what K8s will enforce anyway.

sbwsg:
You'll need to rebase on top of `main` to get the integration tests passing.

1. `git fetch upstream`
2. `git rebase upstream/main`
3. `git push --force <your remote> <branch name>`

psschwei:
Rebased, but still need to update to use `IsDNS1123Label`
/hold

psschwei:
So using `IsDNS1123Label` seems to break a lot of the unit tests...

pkg/apis/pipeline/v1alpha1/run_validation_test.go
```diff
@@ -36,20 +36,18 @@ func TestRun_Invalid(t *testing.T) {
...
-		name: "invalid metadata",
-		run: &v1alpha1.Run{
-			ObjectMeta: metav1.ObjectMeta{Name: "run.name"},
-		},
-		want: &apis.FieldError{
-			Message: "Invalid resource name: special character . must not be present",
-			Paths:   []string{"metadata.name"},
+		run:  &v1alpha1.Run{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "temp",
```
psschwei, line 41: This seems to be required if we want to use `IsDNS1123Label` for validation

psschwei: Will tackle the other 38 failing tests tomorrow

pritidesai: sounds good, thanks @psschwei

psschwei:
So the integration tests are failing with the following error:
```
   Error from server (BadRequest): error when creating "STDIN": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: invalid resource name "unit.tests": must be a valid DNS label: metadata.name
        Error from server (BadRequest): error when creating "STDIN": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: invalid resource name "demo.pipeline": must be a valid DNS label: metadata.name
```

Looking closer at the Kubernetes docs, while a `.` is valid for [DNS Subdomain names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names), it is not a valid character for [DNS Label names](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names). 

Assuming that the `metadata.name` for Pipelines and Tasks are DNS Labels rather than Subdomains (based on the fact that existing code was limiting names to <= 63 characters), it turns out that `.` may not be a valid character after all. 

/meow

pkg/apis/pipeline/v1alpha1/condition_validation_test.go
psschwei: I've added back in tests to check that commas in names are invalid

psschwei:
PR description has been updated with more detail. This should be ready to go now.
/hold cancel
/meow

pkg/apis/validate/metadata.go
```diff
@@ -28,9 +29,9 @@ const MaxLength = 63
 func ObjectMetadata(meta metav1.Object) *apis.FieldError {
 	name := meta.GetName()
 
-	if strings.Contains(name, ".") {
+	if err := validation.IsDNS1123Subdomain(name); len(err) > 0 {
```
pritidesai, line 32: thanks a bunch @psschwei for thorough analysis on this.

After validating the resource name to be a valid `DNSSubdomain`, I would like to add the length check back to make sure the name does not exceed 63 characters (magical number 🤣) just like before. But this time instead of hard coding the number, we can rely on a constant defined in API machinery `DNS1123LabelMaxLength`

psschwei: API machinery constant added to #L27

vdemeester: Does this mean that we *allowed* name like `fooBar` and now we don't ? Or was it already not allowed by the kubernetes api server ?

vdemeester (commented):
One comment on the DNS1123 validation.

psschwei:
> One comment on the DNS1123 validation.
> Does this mean that we allowed name like fooBar and now we don't ? Or was it already not allowed by the kubernetes api server ?

It appears the api server blocked it:
```
$ cat echo-hello-world.yaml 
kind: Task
metadata:
  name: echo-Hello-world
spec:
  steps:
    - name: echo
      image: ubuntu
      command:
        - echo
      args:
        - "Hello World"
$ k apply -f echo-hello-world.yaml -n paul
The Task "echo-Hello-world" is invalid: metadata.name: Invalid value: "echo-Hello-world": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

vdemeester:
@psschwei indeed, it seems it is either the api-server or the generic validation knative.dev/pkg provides us.
/lgtm

pkg/apis/validate/metadata.go
psschwei: > Does this mean that we _allowed_ name like `fooBar` and now we don't ? Or was it already not allowed by the kubernetes api server ?

Answered this below in main comments, but camelCase names were not previously allowed (blocked by either the api server of knative validation package, so this involves no change in functionality

pritidesai:
thanks @psschwei for all the work which started as a simple change 😜 

/approve

PR #3901, opened 2021-04-28
TEP-0061, Allow custom task to be embedded.

### LINKED ISSUE #3682: Custom Tasks: question with taskSpec
related #3463

The current custom task can run in a pipeline using `pipeline.taskRef`. Will there be any support for `pipeline.taskSpec` in the roadmap?

/cc @GregDritschler @fenglixa

ScrapCodes:
# Changes

## TEP-0061, Allow custom task to be embedded
    
    
 1. API changes, This PR adds new APIs i.e. adds a field `Spec *EmbeddedRunSpec` to `RunSpec`
 2. An embedded task will accept new field  `Spec` with type `runtime.RawExtension` in addition to `ApiVersion` and `Kind` fields of type string (as part of `runtime.TypeMeta`) :
 3. Validation changes, in addition to adding support for Run.RunSpec.Spec the validations will be changed to support "One of `Run.RunSpec.Spec` or `Run.RunSpec.Ref`" only and not both as part of a single API request to kubernetes.
 4. Unit tests + e2e test to cover the newly added fields and above mentioned validations.
 5. Documentation changes associated with TEP-0061.
References: 
1. https://github.com/tektoncd/community/pull/415
2. https://github.com/tektoncd/pipeline/issues/3682
3. https://github.com/tektoncd/community/blob/main/teps/0061-allow-custom-task-to-be-embedded-in-pipeline.md

@Tomcli
@pritidesai 
@afrittoli 
@animeshsingh 

/cc @sbwsg @imjasonh

/kind feature

Closes #3682 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note

# It is now possible to embed the spec of a custom task in a Run resource, whether stand-alone or embedded in a Pipeline.

e.g.

          kind: Run
          metadata:
            name: simpletasklooprun
          spec:
            params:
              - name: word
                value:
                  - jump
                  - land
              - name: suffix
                value: ing
            spec:
              apiVersion: custom.tekton.dev/v1alpha1
              kind: TaskLoop
              spec:
                # This is a new field, which comprises of custom-task spec.
                taskRef:
                  name: simpletask
                iterateParam: word
                timeout: 60s
                retries: 2


1. API changes, This PR adds new APIs i.e. adds a field `Spec *EmbeddedRunSpec` to `RunSpec`
2. An embedded task will accepts new field `Spec` with type `runtime.RawExtension` in addition to
    `ApiVersion` and `Kind` fields of type string (as part of `runtime.TypeMeta`) 
3. Validation changes, in addition to adding support for `Run.RunSpec.Spec` the validations will be changed
    to support "One of `Run.RunSpec.Spec` or `Run.RunSpec.Ref` " only and not both as part of a single
    API request to kubernetes.

action required: Developers of custom controllers (existing and new), who would like to support
 embedded spec for their custom task, need to implement the logic required to extract, validate
 and use the custom task spec from the new RunSpec.Spec field. Please review the documentation
 on upgrading, for more details and some examples.

```


Tomcli:
@ScrapCodes looks like you haven't run `hack/update-codegen.sh` to regenerate the api code

afrittoli:
> # Changes
> ##TEP-0061 (1/3), Allow custom task to be embedded.
> 
> ```
>     This is first in the three part series of PRs for implementing the TEP-0061.
> ```
> 
> 1. API changes, This PR adds new APIs i.e. adds a field `Spec *v1beta1.EmbeddedTask` to `RunSpec`
> 2. An embedded task will accept new field  `Spec` with type `runtime.RawExtension` in addition to `ApiVersion` and `Kind` fields of type string (as part of `runtime.TypeMeta`) :
> 3. Validation changes, in addition to adding support for Run.RunSpec.Spec the validations will be changed to support "One of `Run.RunSpec.Spec` or `Run.RunSpec.Ref`" only and not both as part of a single API request to kubernetes.
> 4. Unit tests to cover the newly added fields and above mentioned validations.
> 
> (2/3) Will consist of an e2e test.

To be honest, I would feel more comfortable having the E2E test part of this PR.
The PR is not too large, so it would be ok to have all in one go in fact.

> (3/3) Documentation changes associated with TEP-0061.

Again, I would prefer to have this in the same PR, but I'm ok with having it as a follow-up, as long as we merge both PR as part of the same release.

> 
> /kind feature
> 
> # Submitter Checklist
> 
> * [ ]  [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
> * [x]  [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
> * [x]  Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
> * [x]  Release notes block below has been filled in or deleted (only if no user facing changes)
> 
> # Release Notes
> ```
>  1. API changes, This PR adds new APIs i.e. adds a field `Spec *v1beta1.EmbeddedTask` to `RunSpec`
>  2. An embedded task will accepts new field `Spec` with type `runtime.RawExtension` in addition to
>      `ApiVersion` and `Kind` fields of type string (as part of `runtime.TypeMeta`) 
>  3. Validation changes, in addition to adding support for `Run.RunSpec.Spec` the validations will be changed
>      to support "One of `Run.RunSpec.Spec` or `Run.RunSpec.Ref` " only and not both as part of a single
>     API request to kubernetes.
> ```

NIT: Could you add a more high level description on top or the release notes, something along the lines of 
"It is now possible to embed the spec of a custom task in a `Run` resource, whether stand-alone or embedded in a Pipeline. 
[include a yaml example]. Embedding the spec of a custom task depends on this being supported in each custom task controller that wants to support this feature."
 
> 
> ```
> action required: For developers of custom controllers (existing and new), now a `RunSpec`
>  can have a `Ref` or `Spec`, Scenario, where custom controller supports either or both `Ref`
>  and `Spec`. Please update the validation logic so that it responds with proper validation errors
>  or response to the user for missing a `Ref` and or `Spec`.
> ```

Could you keep this block ^^^ together with the release notes? Having two blocks might break the release notes automation.

I think it's a good idea to make a call for action, but we could say that "Developers of custom controllers (existing and new), who would like to support embedded spec for their custom task, need to implement the logic required to extract, validate and use the custom task spec from the new `RunSpec.Spec` field. "

As discussed during the API WG and noted by @imjasonh - having validation of the spec via a validating webhook would mean that all validating webhook of all custom tasks would be invoked for each `Run` - which might be too much. Perhaps we should recommend async validation during the reconcile cycle instead for now?

> 
> References:
> 
> 1. [tektoncd/community#415](https://github.com/tektoncd/community/pull/415)
> 2. #3682
> 3. https://github.com/tektoncd/community/blob/main/teps/0061-allow-custom-task-to-be-embedded-in-pipeline.md
> 
> @Tomcli
> @pritidesai
> @afrittoli
> @animeshsingh
> 
> /cc @sbwsg @imjasonh

afrittoli:
Thanks a lot @ScrapCodes for this, I look forward to this feature!  I will review the code today.
We will have release v0.24 next week (it's planned for the 5th), do you think we include this in the milestone for the release?

ScrapCodes:
> @ScrapCodes: Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our release note process to remove it.

Looks like this robot does not interpret nested backticks. So, removed them.

ScrapCodes:
Hi @afrittoli, thanks for taking a look !

Let us aim for 0.24 release, at the moment, only thing pending is review + e2e test. Do we have a code freeze process?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -759,12 +760,30 @@ func (c *Reconciler) createRun(ctx context.Context, rprt *resources.ResolvedPipe
...
+			Metadata: rprt.PipelineTask.TaskSpec.Metadata,
+			TaskSpec: rprt.PipelineTask.TaskSpec.TaskSpec,
+		}
+		j, err := json.Marshal(rprt.PipelineTask.TaskSpec.Spec)
+		if err != nil {
+			return nil, err
+		}
+		r.Spec.Spec.Spec = runtime.RawExtension{
+			Raw: j,
+		}
```
afrittoli, line 785: I'm not sure I understand this. We set the `Run.Spec` in L763.
Why do we need to rebuild and `EmbeddedTask` from the `PipelineTask.TaskSpec` (which is an `EmbeddedTask` already)?

docs/runs.md
```diff
@@ -64,6 +66,62 @@ A `Run` definition supports the following fields:
...
+
+This initiates the execution of a `Run` of a custom task of type `Example`, in
+the `example.dev` API group,  with the version `v1alpha1`.
+
+#### Developer guide for custom controllers supporting `spec`.
+
+1. A custom controller may support either of `Ref` or `Spec` or both. In case, it does not
+   support one of them, custom controller should respond with proper validation error.
+   
+2. It is highly recommended that the custom controller has appropriate validations
```
afrittoli, line 101: Isn't this already validated by the pipeline controller?

docs/runs.md
```diff
@@ -64,6 +66,62 @@ A `Run` definition supports the following fields:
...
+   to handle `Run` which misses a `Ref` or a `Spec`.
+
+3. In case of an embedded `spec`, a custom task controller may not actually create
```
afrittoli, line 104: I'm not sure we need this?  It's up to the custom controller to decide what they need.

In Tekton we never create a resource for embedded specs - it does not add any value - or worst, it creates confusion, since it's not clear anymore which version of the resource is being executed, the one on the cluster or the one embedded.

So if we say something I think it should be that we discourage custom controllers from storing the embedded custom resource in `etcd`.

docs/runs.md
```diff
@@ -64,6 +66,62 @@ A `Run` definition supports the following fields:
...
+    NAME              SUCCEEDED   REASON      STARTTIME   COMPLETIONTIME
+    embedded-run      True        Succeeded   1m          1m
+    ```
+    
+    But,
+    ```shell
+    kubectl get example
+    No resources found ...
+    ```
+4. Validation of the fields of the custom task is delegated to the custom task controller.
```
afrittoli, line 118: NIT: missing blank line on top of L118

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -357,10 +357,12 @@ func ResolvePipelineRunTask(
 	rprt := ResolvedPipelineRunTask{
 		PipelineTask: &task,
 	}
-
```
afrittoli, line 360: removed the empty line?

afrittoli (commented):
This is definitely in the right direction, thank you!
I left a couple of comments, mostly minor. 
I'm not sure about the `Run` creation logic in the pipeline run controller, but otherwise it looks good!

pkg/reconciler/pipelinerun/pipelinerun.go
ScrapCodes: My understanding is, we need to embed a `json.Marshal` of the `TaspSpec.Spec`. Please feel free to correct me.

docs/runs.md
ScrapCodes: You are right, it can be confusing and it is up to the custom controller, which way they want to implement.

ScrapCodes:
Thank you @afrittoli , your feedback was very helpful, esp. the adding of `//+optional` to the API field and removing the redundant code from `pipelinerun.go` and your comments on the doc have been addressed hopefully. Please take a look again, when you can.

pkg/apis/pipeline/v1alpha1/run_validation_test.go
```diff
@@ -43,18 +44,19 @@ func TestRun_Invalid(t *testing.T) {
...
-		name: "missing ref",
+		name: "Empty spec",
 		run: &v1alpha1.Run{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: "temp",
 			},
 			Spec: v1alpha1.RunSpec{
-				Ref: nil,
+				Ref:  nil,
+				Spec: nil,
```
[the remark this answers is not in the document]
ScrapCodes: done !

Tomcli:
@afrittoli @pritidesai Thanks for the reviews. Do we need to also ask someone else to review? or we have to merge our TEP first?
https://github.com/tektoncd/community/pull/415

test/custom_task_test.go
```diff
@@ -61,30 +62,45 @@ func TestCustomTask(t *testing.T) {
...
 					Tasks: []v1beta1.PipelineTask{{
-						Name: "custom-task",
+						Name: "custom-task-ref",
 						TaskRef: &v1beta1.TaskRef{
 							APIVersion: apiVersion,
 							Kind:       kind,
 						},
+					}, {
+						Name: "custom-task-spec",
+						TaskSpec: &v1beta1.EmbeddedTask{
```
sbwsg, line 72: Suggest adding a `Spec` nested here and validating that it's rendered to `Raw` as expected

pkg/apis/pipeline/v1alpha1/run_types.go
```diff
@@ -41,6 +41,10 @@ type RunSpec struct {
 	// +optional
 	Ref *TaskRef `json:"ref,omitempty"`
 
+	// Spec is a specification of a custom task
+	// +optional
+	Spec *v1beta1.EmbeddedTask `json:"spec,omitempty"`
```
sbwsg, line 46: Ah, something I had not realised during TEP review is that this also means we're nesting `TaskSpec` fields (like pipeline resources etc,) and `PipelineTaskMetadata` fields (annotations, labels) into the `RunSpec`.

Is there a reason to re-use `EmbeddedTask` here?  Or could we introduce a new type that only includes the fields we want for `RunSpec`?  E.g.

```go
type EmbeddedRunSpec struct {
	// +optional
	runtime.TypeMeta `json:",inline,omitempty"`

	// Spec is a specification of a custom task
	// +optional
	Spec runtime.RawExtension `json:"spec,omitempty"`
}
```

sbwsg: Sorry, another question after reviewing this with the pipelines owners group - what is the purpose of `runtime.TypeMeta` in this location?  Would this potentially differ from the `run.apiVersion` and `run.kind`?

ScrapCodes: >Sorry, another question after reviewing this with the pipelines owners group - what is the purpose of runtime.TypeMeta in this location? Would this potentially differ from the run.apiVersion and run.kind? 

It will not differ from `run.apiVersion` and `run.kind`, my guess is for brevity we prefer this.

>Is there a reason to re-use EmbeddedTask here? Or could we introduce a new type that only includes the fields we want for RunSpec? E.g.

Intention is, custom tasks will become consistent with regular tasks at some point. e.g. we need `PipelineTaskMetadata` to provide label to embedded task. 
If you still feel otherwise, let us know.

sbwsg: ## TypeMeta

Ah, ok.  Going back to the docs in this example helped clear things up for me wrt TypeMeta:

```
kind: Run
metadata:
  name: embedded-run
spec:
  spec:
    apiVersion: example.dev/v1alpha1
    kind: Example
    spec:
      field1: value1
      field2: value2
```

The run's API version and the spec's API version can differ - one describes the tekton resource "tekton.dev/v1alpha1.Run" and the other describes the Custom Task's own apiVersion and kind.  I find this distinction _endlessly_ confusing but that's just the way it is with Custom Tasks. Apologies for muddying things up here.  TypeMeta is fine as-is.

---

## EmbeddedTask

I do want to avoid using `EmbeddedTask` if we can for the following reasons:
- It brings baggage with it that we don't have a concrete use case for yet in Runs (e.g. pipeline resources).
- All of the fields from `TaskSpec` are already representable by virtue of using `runtime.RawExtension` I think?
- We can make `EmbeddedRunSpec` a subset of fields from `EmbeddedTask` and eventually converge them if that's a goal we have.  Although I would also say here that I wasn't aware of this being a Goal - is it mentioned in the TEP or a comment on it?  I probably missed it.

So to support metadata fields as well it sounds like the shape would become:

```go
type RunSpecMetadata struct {
	// +optional
	Labels map[string]string `json:"labels,omitempty"`

	// +optional
	Annotations map[string]string `json:"annotations,omitempty"`
}

type EmbeddedRunSpec struct {
	// +optional
	runtime.TypeMeta `json:",inline,omitempty"`

	// +optional
	Metadata RunSpecMetadata `json:"metadata,omitempty"`

	// Spec is a specification of a custom task
	// +optional
	Spec runtime.RawExtension `json:"spec,omitempty"`
}
```

The `createRun` func becomes a little more complicated because it has to copy values from the `EmbeddedTask` to the `EmbeddedRunSpec` but I much prefer we start with a constrained set of fields and expand it when needed.  WDYT?

ScrapCodes: Alright, agree with you, do you think we would need to update the TEP with above suggestion for types?

ScrapCodes: And get it approved?

ScrapCodes: @sbwsg At the moment a `PipelineTask` has a nested `EmbeddedTask` and as a result, the custom task can be expressed under the field `taskSpec`. After this change, how will that change? 

Few options:
1. We introduce a new optional field `spec` to store the `EmbeddedRunSpec` inside a `PipelineTask`. Downside, two different field with similar sounding field names and purposes.
2. Leave current `EmbeddedTask` as is, and introduce a new type `EmbeddedRunSpec` for `Run`. Downside, same fields copied twice.
3. Any other suggestions?

sbwsg: Once this PR is merged the TEP can be moved to `"implemented"` state.  My suggestion would be to include these changes as part of that PR.

Also pinging @tektoncd/core-maintainers for any other opinions / feedback on the desired process here.

sbwsg: > Leave current EmbeddedTask as is, and introduce a new type EmbeddedRunSpec for Run. Downside, same fields copied twice.

This would definitely be my preference.

ScrapCodes: Yes, indeed.

ScrapCodes: @sbwsg One question, now that `Metadata` field is shifted from `EmbeddedTask` to `EmbeddedSpec` which is inlined into `EmbeddedTask`, is it a matter of concern?

internal/builder/v1beta1/pipeline_test.go
```diff
@@ -152,14 +152,15 @@ func TestPipeline(t *testing.T) {
 			}, {
 				Name: "task-with-taskSpec",
 				TaskSpec: &v1beta1.EmbeddedTask{
-					Metadata: v1beta1.PipelineTaskMetadata{
-						Labels:      map[string]string{"label": "labelvalue"},
-						Annotations: map[string]string{"annotation": "annotationvalue"},
+					EmbeddedSpec: v1beta1.EmbeddedSpec{
+						Metadata: v1beta1.PipelineTaskMetadata{
+							Labels:      map[string]string{"label": "labelvalue"},
```
ScrapCodes, line 157: Please note, those already using the PipelineTaskMetadata field in the EmbeddedTask, will have to make this change. 
//cc @sbwsg

ScrapCodes:
Hello @sbwsg , hopefully this should address all of the comments. Can you please take a look again ! 😄

internal/builder/v1beta1/pipeline_test.go
sbwsg: Ah I see, the problem becomes for users who are interacting with Tekton by importing its types - they now have to update their code to populate the embedded struct.  Is that right?

That's a really interesting point.  I think this is an argument in favor of keeping the two structs (`EmbeddedRunSpec` and `EmbeddedTask`) as totally separate copies in that case.

ScrapCodes: ok, will update the code accordingly.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -765,6 +767,22 @@ func (c *Reconciler) createRun(ctx context.Context, rprt *resources.ResolvedPipe
 		},
 	}
 
+	if rprt.PipelineTask.TaskSpec != nil {
+		j, err := json.Marshal(rprt.PipelineTask.TaskSpec.Spec)
+		if err != nil {
+			return nil, err
+		}
+		r.Spec.Spec = &v1alpha1.EmbeddedRunSpec{
```
sbwsg, line 775: 👍

sbwsg (approved):

sbwsg:
Thanks a lot @ScrapCodes !

animeshsingh:
Thanks @ScrapCodes! Critical feature for our use-cases

docs/pipelines.md
```diff
@@ -1179,6 +1179,28 @@ If the `taskRef` specifies a name, the custom task controller should look up the
...
+```
+
+If the custom task controller supports the in-line or embedded task spec, this will create a `Run` of a custom task of
+type `Example` in the `example.dev` API group with the version `v1alpha1`.
+
+If the `taskSpec` is not supported, the custom task controller should produce proper validation errors.
+
+Please take a look at the
+[developer guide for custom controllers supporting `taskSpec`.](runs.md#developer-guide-for-custom-controllers-supporting-spec)
+
```
pritidesai, line 1203: Please add a reference to the TEP here so that its easier to relate to the proposal 🙏

docs/runs.md
```diff
@@ -47,8 +47,10 @@ A `Run` definition supports the following fields:
   - [`metadata`][kubernetes-overview] - Specifies the metadata that uniquely identifies the
     `Run`, such as a `name`.
   - [`spec`][kubernetes-overview] - Specifies the configuration for the `Run`.
-    - [`ref`](#specifying-the-target-custom-task) - Specifies the type and
+    - [`ref`](#specifying-the-target-custom-task-with-ref) - Specifies the type and
```
pritidesai, line 50: the reference is invalid here, please replace it with #1-specifying-the-target-custom-task-with-ref

docs/runs.md
```diff
@@ -47,8 +47,10 @@ A `Run` definition supports the following fields:
...
       (optionally) name of the custom task type to execute.
+    - [`spec`](#specifying-the-target-custom-task-by-embedding-its-spec) -
```
pritidesai, line 52: same here, please replace the reference with `#2-specifying-the-target-custom-task-by-embedding-its-spec`

docs/runs.md
```diff
@@ -64,6 +66,19 @@ A `Run` definition supports the following fields:
 
 ### Specifying the target Custom Task
 
+A custom task resource's `Spec` may be directly embedded in the `Run` or it may
+be referred to by a `Ref`. But, not both at the same time.
+
+1. [Specifying the target Custom Task with ref](#specifying-the-target-custom-task-with-ref)
```
pritidesai, line 72: same as above, the valid reference is `#1-specifying-the-target-custom-task-with-ref`

docs/runs.md
```diff
@@ -64,6 +66,19 @@ A `Run` definition supports the following fields:
...
+   Referring a custom task (i.e. `Ref` ) promotes reuse of custom task definitions.
+
+2. [Specifying the target Custom Task by embedding its spec](#specifying-the-target-custom-task-by-embedding-its-spec)
```
pritidesai, line 75: ditto 🙃 valid reference here is ` #2-specifying-the-target-custom-task-by-embedding-its-spec`

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -765,6 +767,22 @@ func (c *Reconciler) createRun(ctx context.Context, rprt *resources.ResolvedPipe
...
+			TypeMeta: runtime.TypeMeta{
+				APIVersion: rprt.PipelineTask.TaskSpec.APIVersion,
+				Kind:       rprt.PipelineTask.TaskSpec.Kind,
+			},
+			Metadata: rprt.PipelineTask.TaskSpec.Metadata,
+			Spec: runtime.RawExtension{
+				Raw: j,
```
pritidesai, line 782: 🎉

pkg/apis/pipeline/v1alpha1/run_types.go
```diff
@@ -36,11 +37,27 @@ var (
 	}
 )
 
+// EmbeddedRunSpec allows custom task definitions to be embedded
+type EmbeddedRunSpec struct {
+	runtime.TypeMeta `json:",inline,omitempty"`
```
pritidesai, line 42: We consider this as a required field so do we need to add `omitempty` clause here?

pkg/apis/pipeline/v1alpha1/run_validation_test.go
```diff
@@ -43,18 +44,39 @@ func TestRun_Invalid(t *testing.T) {
 		},
 		want: apis.ErrMissingField("spec"),
 	}, {
-		name: "missing ref",
+		name: "Empty spec",
```
pritidesai, line 47: => empty ref and spec

pkg/apis/pipeline/v1alpha1/run_validation.go
```diff
@@ -36,20 +36,30 @@ func (r *Run) Validate(ctx context.Context) *apis.FieldError {
...
 // Validate Run spec
 func (rs *RunSpec) Validate(ctx context.Context) *apis.FieldError {
+	// this covers the case rs.Ref == nil && rs.Spec == nil
 	if equality.Semantic.DeepEqual(rs, &RunSpec{}) {
 		return apis.ErrMissingField("spec")
 	}
 
-	if rs.Ref == nil {
-		return apis.ErrMissingField("spec.ref")
+	if rs.Ref != nil && rs.Spec != nil {
```
pritidesai, line 44: need an additional check here, if both are `nil`

pkg/apis/pipeline/v1alpha1/run_validation_test.go
pritidesai: the test name does not match with ref and spec both being `nil`, please add one more check in the validation routine and update this test.

pkg/controller/filter.go
```diff
@@ -44,13 +44,18 @@ func FilterRunRef(apiVersion, kind string) func(interface{}) bool {
...
 			// created somehow, don't panic.
 			return false
 		}
-
-		return r.Spec.Ref.APIVersion == apiVersion && r.Spec.Ref.Kind == v1alpha1.TaskKind(kind)
+		result := false
+		if r.Spec.Ref != nil {
+			result = r.Spec.Ref.APIVersion == apiVersion && r.Spec.Ref.Kind == v1alpha1.TaskKind(kind)
+		} else if r.Spec.Spec != nil {
+			result = r.Spec.Spec.APIVersion == apiVersion && r.Spec.Spec.Kind == kind
```
pritidesai, line 56: Nothing wrong with your changes here, just an observation:

`r.Spec.Ref.Kind` is of type `TaskKind` and that's why needs the conversion v/s `r.Spec.Spec.Kind` is of type `string`. `TaskKind` is also originally used to signify whether a task is `task` or `clustertask` but with custom task, it can be set to anything.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -358,9 +358,12 @@ func ResolvePipelineRunTask(
...
 	}
 
+	isTaskRefCustomTask := rprt.PipelineTask.TaskRef != nil && rprt.PipelineTask.TaskRef.APIVersion != "" &&
+		rprt.PipelineTask.TaskRef.Kind != ""
+	isTaskSpecCustomTask := rprt.PipelineTask.TaskSpec != nil && rprt.PipelineTask.TaskSpec.APIVersion != "" &&
+		rprt.PipelineTask.TaskSpec.Kind != ""
 	cfg := config.FromContextOrDefaults(ctx)
-	rprt.CustomTask = cfg.FeatureFlags.EnableCustomTasks && rprt.PipelineTask.TaskRef != nil &&
-		rprt.PipelineTask.TaskRef.APIVersion != "" && rprt.PipelineTask.TaskRef.Kind != ""
+	rprt.CustomTask = cfg.FeatureFlags.EnableCustomTasks && (isTaskRefCustomTask || isTaskSpecCustomTask)
```
pritidesai, line 366: Please separate this entire logic into a separate function to check whether a task is a custom task and assign `rprt.CustomTask` appropriate value here based on that new function. `ResolvePipelineRunTask` will be simplified instead of making everything part of it.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -765,6 +767,22 @@ func (c *Reconciler) createRun(ctx context.Context, rprt *resources.ResolvedPipe
 		},
 	}
 
+	if rprt.PipelineTask.TaskSpec != nil {
+		j, err := json.Marshal(rprt.PipelineTask.TaskSpec.Spec)
```
pritidesai, line 771: what happens when `spec` is empty here? do we need a check here `rprt.PipelineTask.TaskSpec.Spec`?

pritidesai:
the changes looks great @ScrapCodes. I am done reviewing the PR and have requested couple of changes. 🙏 

One more thing I am not sure about is all the third party changes, without changing the `mod` file, how come we are seeing the dependency change 🤔

pkg/reconciler/pipelinerun/pipelinerun.go
ScrapCodes: there is no error for empty spec, in fact a non empty byte sequence is generated. Do you think we should be more defensive here?

pkg/apis/pipeline/v1alpha1/run_validation_test.go
ScrapCodes: The reason is, both nil and `Empty spec` are same in terms of what happens underneath. 
Actual check says, 
`if equality.Semantic.DeepEqual(rs, &amp;RunSpec{}) {		return apis.ErrMissingField("spec")` i.e. empty spec and so is the name. Please see the reply to your comment above as well, and let me know, if this is still concerning.

ScrapCodes:
> the changes looks great @ScrapCodes. I am done reviewing the PR and have requested couple of changes. 🙏
> 
> One more thing I am not sure about is all the third party changes, without changing the `mod` file, how come we are seeing the dependency change 🤔

I am not sure either, they are generated as part of running, `hack/update-codegen.sh`.

ScrapCodes:
Looks like it is a problem with my local environment. Even after removing those `third_party` additions, build passes i.e. build passes regardless of those `third_party` additions are present or not. Is this concerning from the point of view, how build is setup.

ScrapCodes:
Thank you @pritidesai for taking a look. Please take a look again.

Thank you @sbwsg !

Thank you @animeshsingh and @Tomcli for all the support 😄

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: I am trying to think if there is any use case where the `spec` could be empty, could this ever be used in any way? 

Since this is not producing any error, its safe to go with the way its implemented now and revisit if needed in future.

Tomcli:
thanks @pritidesai for your review

pkg/apis/pipeline/v1alpha1/run_validation.go
ScrapCodes: Thanks, now I have understood ! Will open a PR with the fix soon.

ScrapCodes:
@pritidesai I have opened a follow-up PR with the validation: https://github.com/tektoncd/pipeline/pull/3977

PR #3915, opened 2021-05-05
Graceful Pipeline Run Termination

rafalbigaj:
The implementation of TEP-0058: Graceful Pipeline Run Termination.

The new `spec.Status` values:

- `StoppedRunFinally` - To "stop" (i.e. let the tasks complete, then execute finally tasks) a Pipeline
- `CancelledRunFinally` - To "cancel" (i.e. interrupt any executing non finally tasks, then execute finally tasks)
- `Cancelled` - Same as today's `PipelineRunCancelled` - i.e. interrupt any executing tasks without running finally tasks

`PipelineRunCancelled` is deprecated (replaced by `Cancelled`)

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
# The new PipelineRun spec statuses have been added to control the way who a PipelineRun is being canceled or stopped.

- "StoppedRunFinally" - To stop (i.e. let the tasks complete, then execute finally tasks) a PipelineRun
- "CancelledRunFinally" - To cancel (i.e. interrupt any executing non finally tasks, then execute finally tasks)
- "Cancelled" - replaces today's "PipelineRunCancelled" - i.e. interrupt any executing tasks without running finally tasks

Support for existing statuses has been left unchanged.

The status "PipelineRunCancelled" is deprecated and replaced by "Cancelled" (it would be removed in v1).

The new states are released as alpha API features.

[1] https://github.com/tektoncd/community/blob/main/teps/0058-graceful-pipeline-run-termination.md
```

### DIFF
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -561,8 +561,9 @@ Task Runs:
 ## Cancelling a `PipelineRun`
 
 To cancel a `PipelineRun` that's currently executing, update its definition
-to mark it as cancelled. When you do so, the spawned `TaskRuns` are also marked
-as cancelled and all associated `Pods` are deleted. For example:
+to mark it as "PipelineRunCancelled". When you do so, the spawned `TaskRuns` are also marked
+as cancelled and all associated `Pods` are deleted. Pending final tasks are not scheduled. 
+For example:
 
 ```yaml
 apiVersion: tekton.dev/v1beta1
@@ -572,6 +573,46 @@ metadata:
 spec:
   # […]
   status: "PipelineRunCancelled"
+```
+
+Warning: "PipelineRunCancelled" status is deprecated and would be removed in V1, please use "Cancelled" instead.  
+
+## Gracefully cancelling a `PipelineRun`
+
+[Graceful pipeline run termination](https://github.com/tektoncd/community/blob/main/teps/0058-graceful-pipeline-run-termination.md)
+is currently an **_alpha feature_**.
+
+To gracefully cancel a `PipelineRun` that's currently executing, update its definition
+to mark it as "CancelledRunFinally". When you do so, the spawned `TaskRuns` are also marked
+as cancelled and all associated `Pods` are deleted. Final tasks are scheduled normally. 
+For example:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: go-example-git
+spec:
+  # […]
+  status: "CancelledRunFinally"
+```
+
+
+## Gracefully stopping a `PipelineRun`
+
+To gracefully stop a `PipelineRun` that's currently executing, update its definition
+to mark it as "StoppedRunFinally". When you do so, the spawned `TaskRuns` are completed normally,
+but no new non-final task is scheduled. Final tasks are executed afterwards.
+For example:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: go-example-git
+spec:
+  # […]
+  status: "StoppedRunFinally"
 ```
 
 ## Pending `PipelineRuns`
--- a/internal/builder/v1beta1/pipeline.go
+++ b/internal/builder/v1beta1/pipeline.go
@@ -115,12 +115,27 @@ func PipelineDescription(desc string) PipelineSpecOp {
 	}
 }
 
-// PipelineRunCancelled sets the status to cancel the PipelineRunSpec.
+// PipelineRunCancelled sets the status to Cancelled in the PipelineRunSpec.
 func PipelineRunCancelled(spec *v1beta1.PipelineRunSpec) {
 	spec.Status = v1beta1.PipelineRunSpecStatusCancelled
 }
 
-// PipelineRunPending sets the status to pending to the PipelineRunSpec.
+// PipelineRunCancelledDeprecated sets the status to PipelineRunCancelled in the PipelineRunSpec.
+func PipelineRunCancelledDeprecated(spec *v1beta1.PipelineRunSpec) {
+	spec.Status = v1beta1.PipelineRunSpecStatusCancelledDeprecated
+}
+
+// PipelineRunCancelledRunFinally sets the status to cancel and run finally in the PipelineRunSpec.
+func PipelineRunCancelledRunFinally(spec *v1beta1.PipelineRunSpec) {
+	spec.Status = v1beta1.PipelineRunSpecStatusCancelledRunFinally
+}
+
+// PipelineRunStoppedRunFinally sets the status to stop and run finally in the PipelineRunSpec.
+func PipelineRunStoppedRunFinally(spec *v1beta1.PipelineRunSpec) {
+	spec.Status = v1beta1.PipelineRunSpecStatusStoppedRunFinally
+}
+
+// PipelineRunPending sets the status to pending in the PipelineRunSpec.
 func PipelineRunPending(spec *v1beta1.PipelineRunSpec) {
 	spec.Status = v1beta1.PipelineRunSpecStatusPending
 }
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
@@ -85,7 +85,17 @@ func (pr *PipelineRun) HasStarted() bool {
 
 // IsCancelled returns true if the PipelineRun's spec status is set to Cancelled state
 func (pr *PipelineRun) IsCancelled() bool {
-	return pr.Spec.Status == PipelineRunSpecStatusCancelled
+	return pr.Spec.Status == PipelineRunSpecStatusCancelled || pr.Spec.Status == PipelineRunSpecStatusCancelledDeprecated
+}
+
+// IsGracefullyCancelled returns true if the PipelineRun's spec status is set to CancelledRunFinally state
+func (pr *PipelineRun) IsGracefullyCancelled() bool {
+	return pr.Spec.Status == PipelineRunSpecStatusCancelledRunFinally
+}
+
+// IsGracefullyStopped returns true if the PipelineRun's spec status is set to StoppedRunFinally state
+func (pr *PipelineRun) IsGracefullyStopped() bool {
+	return pr.Spec.Status == PipelineRunSpecStatusStoppedRunFinally
 }
 
 func (pr *PipelineRun) GetTimeout(ctx context.Context) time.Duration {
@@ -216,9 +226,22 @@ type TimeoutFields struct {
 type PipelineRunSpecStatus string
 
 const (
+	// Deprecated: "PipelineRunCancelled" indicates that the user wants to cancel the task,
+	// if not already cancelled or terminated (replaced by "Cancelled")
+	PipelineRunSpecStatusCancelledDeprecated = "PipelineRunCancelled"
+
 	// PipelineRunSpecStatusCancelled indicates that the user wants to cancel the task,
 	// if not already cancelled or terminated
-	PipelineRunSpecStatusCancelled = "PipelineRunCancelled"
+	PipelineRunSpecStatusCancelled = "Cancelled"
+
+	// PipelineRunSpecStatusCancelledRunFinally indicates that the user wants to cancel the pipeline run,
+	// if not already cancelled or terminated, but ensure finally is run normally
+	PipelineRunSpecStatusCancelledRunFinally = "CancelledRunFinally"
+
+	// PipelineRunSpecStatusStoppedRunFinally indicates that the user wants to stop the pipeline run,
+	// wait for already running tasks to be completed and run finally
+	// if not already cancelled or terminated
+	PipelineRunSpecStatusStoppedRunFinally = "StoppedRunFinally"
 
 	// PipelineRunSpecStatusPending indicates that the user wants to postpone starting a PipelineRun
 	// until some condition is met
@@ -271,6 +294,12 @@ const (
 	// PipelineRunReasonStopping indicates that no new Tasks will be scheduled by the controller, and the
 	// pipeline will stop once all running tasks complete their work
 	PipelineRunReasonStopping PipelineRunReason = "PipelineRunStopping"
+	// PipelineRunReasonCancelledRunningFinally indicates that pipeline has been gracefully cancelled
+	// and no new Tasks will be scheduled by the controller, but final tasks are now running
+	PipelineRunReasonCancelledRunningFinally PipelineRunReason = "CancelledRunningFinally"
+	// PipelineRunReasonStoppedRunningFinally indicates that pipeline has been gracefully stopped
+	// and no new Tasks will be scheduled by the controller, but final tasks are now running
+	PipelineRunReasonStoppedRunningFinally PipelineRunReason = "StoppedRunningFinally"
 )
 
 func (t PipelineRunReason) String() string {
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
@@ -115,11 +115,7 @@ func (ps *PipelineRunSpec) Validate(ctx context.Context) (errs *apis.FieldError)
 		}
 	}
 
-	if ps.Status != "" {
-		if ps.Status != PipelineRunSpecStatusCancelled && ps.Status != PipelineRunSpecStatusPending {
-			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("%s should be %s or %s", ps.Status, PipelineRunSpecStatusCancelled, PipelineRunSpecStatusPending), "status"))
-		}
-	}
+	errs = errs.Also(validateSpecStatus(ctx, ps.Status))
 
 	if ps.Workspaces != nil {
 		wsNames := make(map[string]int)
@@ -133,6 +129,32 @@ func (ps *PipelineRunSpec) Validate(ctx context.Context) (errs *apis.FieldError)
 	}
 
 	return errs
+}
+
+func validateSpecStatus(ctx context.Context, status PipelineRunSpecStatus) *apis.FieldError {
+	switch status {
+	case "":
+		return nil
+	case PipelineRunSpecStatusPending,
+		PipelineRunSpecStatusCancelledDeprecated:
+		return nil
+	case PipelineRunSpecStatusCancelled,
+		PipelineRunSpecStatusCancelledRunFinally,
+		PipelineRunSpecStatusStoppedRunFinally:
+		return ValidateEnabledAPIFields(ctx, "graceful termination", "alpha")
+	}
+
+	cfg := config.FromContextOrDefaults(ctx)
+	if cfg.FeatureFlags.EnableAPIFields == config.AlphaAPIFields {
+		return apis.ErrInvalidValue(fmt.Sprintf("%s should be %s, %s, %s or %s", status,
+			PipelineRunSpecStatusCancelled,
+			PipelineRunSpecStatusCancelledRunFinally,
+			PipelineRunSpecStatusStoppedRunFinally,
+			PipelineRunSpecStatusPending), "status")
+	}
+	return apis.ErrInvalidValue(fmt.Sprintf("%s should be %s or %s", status,
+		PipelineRunSpecStatusCancelledDeprecated,
+		PipelineRunSpecStatusPending), "status")
 }
 
 func validateTimeoutDuration(field string, d *metav1.Duration) (errs *apis.FieldError) {
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
@@ -94,6 +94,35 @@ func TestPipelineRun_Invalid(t *testing.T) {
 		},
 		want: apis.ErrInvalidValue("PipelineRunCancell should be PipelineRunCancelled or PipelineRunPending", "spec.status"),
 	}, {
+		name: "wrong pipelinerun cancel with alpha features",
+		pr: v1beta1.PipelineRun{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "pipelinelinename",
+			},
+			Spec: v1beta1.PipelineRunSpec{
+				PipelineRef: &v1beta1.PipelineRef{
+					Name: "prname",
+				},
+				Status: "PipelineRunCancell",
+			},
+		},
+		want: apis.ErrInvalidValue("PipelineRunCancell should be Cancelled, CancelledRunFinally, StoppedRunFinally or PipelineRunPending", "spec.status"),
+		wc:   enableAlphaAPIFields,
+	}, {
+		name: "alpha pipelinerun graceful cancel",
+		pr: v1beta1.PipelineRun{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "pipelinelinename",
+			},
+			Spec: v1beta1.PipelineRunSpec{
+				PipelineRef: &v1beta1.PipelineRef{
+					Name: "prname",
+				},
+				Status: v1beta1.PipelineRunSpecStatusCancelled,
+			},
+		},
+		want: apis.ErrGeneric("graceful termination requires \"enable-api-fields\" feature gate to be \"alpha\" but it is \"stable\""),
+	}, {
 		name: "use of bundle without the feature flag set",
 		pr: v1beta1.PipelineRun{
 			ObjectMeta: metav1.ObjectMeta{
@@ -270,6 +299,61 @@ func TestPipelineRun_Validate(t *testing.T) {
 			},
 		},
 		wc: apis.WithinDelete,
+	}, {
+		name: "pipelinerun cancelled",
+		pr: v1beta1.PipelineRun{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "pipelinerunname",
+			},
+			Spec: v1beta1.PipelineRunSpec{
+				Status: v1beta1.PipelineRunSpecStatusCancelled,
+				PipelineRef: &v1beta1.PipelineRef{
+					Name: "prname",
+				},
+			},
+		},
+		wc: enableAlphaAPIFields,
+	}, {
+		name: "pipelinerun cancelled deprecated",
+		pr: v1beta1.PipelineRun{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "pipelinerunname",
+			},
+			Spec: v1beta1.PipelineRunSpec{
+				Status: v1beta1.PipelineRunSpecStatusCancelledDeprecated,
+				PipelineRef: &v1beta1.PipelineRef{
+					Name: "prname",
+				},
+			},
+		},
+	}, {
+		name: "pipelinerun gracefully cancelled",
+		pr: v1beta1.PipelineRun{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "pipelinerunname",
+			},
+			Spec: v1beta1.PipelineRunSpec{
+				Status: v1beta1.PipelineRunSpecStatusCancelledRunFinally,
+				PipelineRef: &v1beta1.PipelineRef{
+					Name: "prname",
+				},
+			},
+		},
+		wc: enableAlphaAPIFields,
+	}, {
+		name: "pipelinerun gracefully stopped",
+		pr: v1beta1.PipelineRun{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "pipelinerunname",
+			},
+			Spec: v1beta1.PipelineRunSpec{
+				Status: v1beta1.PipelineRunSpecStatusStoppedRunFinally,
+				PipelineRef: &v1beta1.PipelineRef{
+					Name: "prname",
+				},
+			},
+		},
+		wc: enableAlphaAPIFields,
 	}}
 
 	for _, ts := range tests {
@@ -408,7 +492,7 @@ func TestPipelineRunWithAlphaFields_Invalid(t *testing.T) {
 			},
 		},
 		want: apis.ErrInvalidValue("-48h0m0s should be >= 0", "spec.timeouts.pipeline"),
-		wc:   enableTektonTimeoutFields(),
+		wc:   enableAlphaAPIFields,
 	}, {
 		name: "negative pipeline tasks Timeout",
 		pr: v1beta1.PipelineRun{
@@ -425,7 +509,7 @@ func TestPipelineRunWithAlphaFields_Invalid(t *testing.T) {
 			},
 		},
 		want: apis.ErrInvalidValue("-48h0m0s should be >= 0", "spec.timeouts.tasks"),
-		wc:   enableTektonTimeoutFields(),
+		wc:   enableAlphaAPIFields,
 	}, {
 		name: "negative pipeline finally Timeout",
 		pr: v1beta1.PipelineRun{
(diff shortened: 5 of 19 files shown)


pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -59,11 +60,51 @@ func init() {
...
+	}
+	return nil
+}
+
+// cancelPipelineTaskRuns patches `TaskRun` and `Run` with canceled status
+func cancelPipelineTaskRuns(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, omitTaskGraph *dag.Graph, clientSet clientset.Interface) []string {
 	errs := []string{}
 
+	isTaskOmittedFromCancel := func(taskName string) bool {
+		if omitTaskGraph == nil {
```
souleb, line 94: what is the use case here? When would we need to omit task when cancelling a pipelinerun tasks?

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -95,6 +95,14 @@ func TestCancelPipelineRun(t *testing.T) {
 			{ObjectMeta: metav1.ObjectMeta{Name: "t1"}},
 			{ObjectMeta: metav1.ObjectMeta{Name: "t2"}},
 		},
+	}, {
+		name: "deprecated-state",
+		pipelineRun: &v1beta1.PipelineRun{
+			ObjectMeta: metav1.ObjectMeta{Name: "test-pipeline-run-cancelled"},
+			Spec: v1beta1.PipelineRunSpec{
+				Status: v1beta1.PipelineRunSpecStatusCancelledDeprecated,
```
souleb, line 103: can we have a test with the non deprecated stopping and cancelling fields?

internal/builder/v1beta1/pipeline.go
```diff
@@ -120,6 +120,16 @@ func PipelineRunCancelled(spec *v1beta1.PipelineRunSpec) {
 	spec.Status = v1beta1.PipelineRunSpecStatusCancelled
 }
 
+// PipelineRunCancelledRunFinally sets the status to cancel and run finally the PipelineRunSpec.
+func PipelineRunCancelledRunFinally(spec *v1beta1.PipelineRunSpec) {
```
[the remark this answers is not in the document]
rafalbigaj: Thanks for pointing!
This pkg is used widely in `pkg/reconciler/pipelinerun/pipelinerun_test.go`. I added this function for consistency with other tests in that test suite.
Could we stay with this approach and refactor the whole suite afterwards?

pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
```diff
@@ -101,3 +97,18 @@ func (ps *PipelineRunSpec) Validate(ctx context.Context) (errs *apis.FieldError)
 
 	return errs
 }
+
+func validateSpecStatus(status PipelineRunSpecStatus) *apis.FieldError {
```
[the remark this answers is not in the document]
rafalbigaj: Thanks! Done.

internal/builder/v1beta1/pipeline.go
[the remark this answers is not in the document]
souleb: Yeah having a specific PR to refactor that part sounds good.

pkg/reconciler/pipelinerun/cancel_test.go
rafalbigaj: This test was added to cover deprecated status name (PipelineRunCancelled). The already existing ones cover the new status name (Cancelled).

pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
```diff
@@ -101,3 +97,30 @@ func (ps *PipelineRunSpec) Validate(ctx context.Context) (errs *apis.FieldError)
...
+		PipelineRunSpecStatusCancelledDeprecated:
+		return nil
+	case PipelineRunSpecStatusCancelled,
+		PipelineRunSpecStatusCancelledRunFinally,
+		PipelineRunSpecStatusStoppedRunFinally:
+		if alphaAPIFieldsEnabled {
+			return nil
+		}
+	}
+	if alphaAPIFieldsEnabled {
```
souleb, line 116: Sorry to come back to this again. Here when the `alphaFields` is not enabled, we might have an error without any information given to the user. Would be nice to have a specific error in that case stating that it fails because the `alphaFields` are not enabled.

There is a helper function you may use:
https://github.com/tektoncd/pipeline/blob/b86a9a2204d988fbd2b187f220d64b5a102c1ca3/pkg/apis/pipeline/v1beta1/version_validation.go#L30

rafalbigaj:
@pritidesai @vdemeester would you have any comments on this one (assuming TEP is finally merged 😄  )?

docs/pipelineruns.md
```diff
@@ -535,7 +571,7 @@ metadata:
   name: go-example-git
 spec:
   # […]
-  status: "PipelineRunCancelled"
```
pritidesai, line 538: Hey @rafalbigaj lets keep this example as is with a note saying this `status` will be deprecated in V1, please use `Cancelled` instead. Also add a reference to the TEP in the doc. The rest of the changes to the doc looks good.

rafalbigaj: Thanks @pritidesai for this suggestion!
I updated docs and added link to TEP.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -1426,6 +1490,496 @@ func TestReconcileOnCancelledPipelineRun(t *testing.T) {
 	}
 }
 
+func TestReconcileOnCancelledRunFinallyPipelineRun(t *testing.T) {
```
pritidesai, line 1493: thanks @rafalbigaj for adding comprehensive set of tests 🙏 

Can we change this test to have two pipelineTasks instead of one and making sure no new taskruns are created?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -1426,6 +1490,496 @@ func TestReconcileOnCancelledPipelineRun(t *testing.T) {
...
+	prs := []*v1beta1.PipelineRun{tb.PipelineRun("test-pipeline-run-cancelled-run-finally",
+		tb.PipelineRunNamespace("foo"),
+		tb.PipelineRunSpec("test-pipeline", tb.PipelineRunServiceAccountName("test-sa"),
+			tb.PipelineRunCancelledRunFinally,
+		),
+		tb.PipelineRunStatus(tb.PipelineRunStartTime(time.Now())),
+	)}
+	ps := []*v1beta1.Pipeline{tb.Pipeline("test-pipeline", tb.PipelineNamespace("foo"), tb.PipelineSpec(
+		tb.PipelineTask("hello-world-1", "hello-world"),
+		tb.FinalPipelineTask("final-task-1", "some-task"),
```
pritidesai, line 1545: can we please add one more `dag` task here just to cover that task is not scheduled instead `taskRun` for finally is created.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -1426,6 +1490,496 @@ func TestReconcileOnCancelledPipelineRun(t *testing.T) {
...
+		t.Errorf("Expected a CompletionTime to be nil on incomplete PipelineRun but was %v", reconciledRun.Status.CompletionTime)
+	}
+
+	// This PipelineRun should still be complete and unknown, and the status should reflect that
+	if !reconciledRun.Status.GetCondition(apis.ConditionSucceeded).IsUnknown() {
+		t.Errorf("Expected PipelineRun status to be complete and unknown, but was %v", reconciledRun.Status.GetCondition(apis.ConditionSucceeded))
+	}
+
+	// There should be exactly one task run triggered for the "final-task-1" final task
+	if len(reconciledRun.Status.TaskRuns) != 2 {
```
pritidesai, line 1666: The comment and the check does not match.

Also, wre we expecting two `taskRuns` here? The pipeline has one final task which has associated taskRun already running. And I think we are testing the final `taskRun` is not impacted and continues running, right?. There can not be one more taskRun. The `dag` can not be running after finally starts executing.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -1426,6 +1490,496 @@ func TestReconcileOnCancelledPipelineRun(t *testing.T) {
...
+	}
+
+	testAssets, cancel := getPipelineRunController(t, d)
+	defer cancel()
+	c := testAssets.Controller
+	clients := testAssets.Clients
+
+	// Make the patch call fail, i.e. make it so that the controller fails to cancel the TaskRun
+	clients.Pipeline.PrependReactor("patch", "taskruns", func(action ktesting.Action) (bool, runtime.Object, error) {
+		return true, nil, fmt.Errorf("i'm sorry Dave, i'm afraid i can't do that")
```
pritidesai, line 1741: 😆

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -240,8 +241,12 @@ func (state PipelineRunState) getNextTasks(candidateTasks sets.String) []*Resolv
 }
 
 // IsStopping returns true if the PipelineRun won't be scheduling any new Task because
-// at least one task already failed or was cancelled in the specified dag
+// it was cancelled or stopped gracefully or at least one task already failed or was cancelled in the specified dag
 func (facts *PipelineRunFacts) IsStopping() bool {
+	if facts.SpecStatus == v1beta1.PipelineRunSpecStatusCancelledRunFinally ||
+		facts.SpecStatus == v1beta1.PipelineRunSpecStatusStoppedRunFinally {
```
pritidesai, line 247: Instead of leveraging `stopping` for `CancelledRunFinally`, is it possible to not utilize `stopping` mode? How about introduce a `cancelledRunFinally` mode similar to `stopping`?

pkg/reconciler/pipelinerun/pipelinerun_test.go
rafalbigaj: Thanks for suggestion. The second task is in place.

pkg/reconciler/pipelinerun/pipelinerun_test.go
rafalbigaj: Done. Thanks!

pkg/reconciler/pipelinerun/pipelinerun_test.go
rafalbigaj: Thanks for pointing that. The comment now explains that one task run represents already completes task and the second one the running finally task.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
rafalbigaj: Sure, makes sense.

rafalbigaj:
@pritidesai thanks a lot for all your suggestions! 🙏 
Additionally,  I delivered a few more units for better coverage.

pritidesai:
> Additionally, I delivered a few more units for better coverage.

thanks a bunch @rafalbigaj, the coverage looks great now 👍  I am reviewing your changes as we speak 🙏

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -255,12 +256,22 @@ func (facts *PipelineRunFacts) IsStopping() bool {
...
+}
+
 // DAGExecutionQueue returns a list of DAG tasks which needs to be scheduled next
 func (facts *PipelineRunFacts) DAGExecutionQueue() (PipelineRunState, error) {
 	tasks := PipelineRunState{}
-	// when pipeline run is stopping, do not schedule any new task and only
+	// when pipeline run is stopping, gracefully cancelled or stopped, do not schedule any new task and only
 	// wait for all running tasks to complete and report their status
-	if !facts.IsStopping() {
+	if !facts.IsStopping() && !facts.IsGracefullyCancelled() && !facts.IsGracefullyStopped() {
```
pritidesai, line 274: 🎉

pritidesai:
Alright, I am almost done reviewing this PR, changes looks great, thank you so much for your patience @rafalbigaj 🙏 .

 One last concern:

`Cancelled` and `PipelineRunCancelled` results in the following state transitioning:

```
kubectl get pr --watch
NAME                       SUCCEEDED   REASON   STARTTIME   COMPLETIONTIME
01-terminate-pipelinerun
01-terminate-pipelinerun   Unknown     Running                0s
01-terminate-pipelinerun   Unknown     Running                8s
01-terminate-pipelinerun   False       PipelineRunCancelled   8s          0s
```

i.e. `Running` -> `PipelineRunCancelled`

`PipelineRunCancelled` will be deprecated at some point (probably in V1). I think we should transition to the reason `Cancelled` instead of `PipelineRunCancelled` when the status is set to `Cancelled`. @rafalbigaj wdyt?

The other two states (`CancelledRunFinally` and `StoppedRunFinally`) are transitioning as expected:

```
kubectl get pr --watch
NAME                       SUCCEEDED   REASON   STARTTIME   COMPLETIONTIME
01-terminate-pipelinerun
01-terminate-pipelinerun   Unknown     Running   0s
01-terminate-pipelinerun   Unknown     Running   15s
01-terminate-pipelinerun   Unknown     PipelineRunStopping   15s
01-terminate-pipelinerun   Unknown     PipelineRunStopping   39s
01-terminate-pipelinerun   False       Cancelled             40s         1s
01-terminate-pipelinerun   False       Cancelled             102s        63s
```

On the other side, to counter this transition to `PipelineRunStopping`, we could transition to their respective states before `cancelled`, such as:

* CancelledRunFinally:

```
kubectl get pr --watch
NAME                       SUCCEEDED   REASON   STARTTIME   COMPLETIONTIME
01-terminate-pipelinerun
01-terminate-pipelinerun   Unknown     Running   0s
01-terminate-pipelinerun   Unknown     Running   15s
01-terminate-pipelinerun   Unknown     CancelledRunningFinally   15s
01-terminate-pipelinerun   Unknown     CancelledRunningFinally   39s
01-terminate-pipelinerun   False       Cancelled             40s         1s
01-terminate-pipelinerun   False       Cancelled             102s        63s
```

* StoppedRunFinally:

```
kubectl get pr --watch
NAME                       SUCCEEDED   REASON   STARTTIME   COMPLETIONTIME
01-terminate-pipelinerun
01-terminate-pipelinerun   Unknown     Running   0s
01-terminate-pipelinerun   Unknown     Running   15s
01-terminate-pipelinerun   Unknown     PipelineRunStopping   15s
01-terminate-pipelinerun   Unknown     StoppedRunningFinally   39s
01-terminate-pipelinerun   False       Cancelled             40s         1s
01-terminate-pipelinerun   False       Cancelled             102s        63s
```

These reasons can be set based on the `dag` taskRuns and `finally` taskRuns as part of the same function:

https://github.com/tektoncd/pipeline/blob/63d0b6ed572bb06eaea437270caa37ec0165721f/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go#L353-L355

afrittoli (commented):
Thank you @rafalbigaj - this looks good.
I'm half way through the review - sharing my comments so far.
You will need a rebase 🙏

docs/pipelineruns.md
```diff
@@ -538,6 +539,46 @@ spec:
   status: "PipelineRunCancelled"
 ```
 
+Warning: "PipelineRunCancelled" status is deprecated and would be removed in V1, please use "Cancelled" instead.  
+
+## Gracefully cancelling a `PipelineRun`
+
+[Graceful pipeline run termination](https://github.com/tektoncd/community/blob/main/teps/0058-graceful-pipeline-run-termination.md)
+is currently an **_experimental alpha feature_**.
```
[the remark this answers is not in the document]
rafalbigaj: 👍

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
```diff
@@ -397,3 +481,16 @@ func enableTektonOCIBundles(t *testing.T) func(context.Context) context.Context
...
+		s := config.NewStore(logtesting.TestLogger(t))
+		s.OnConfigChanged(&corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{Name: config.GetFeatureFlagsConfigName()},
+			Data: map[string]string{
+				"enable-api-fields": config.AlphaAPIFields,
+			},
+		})
+		return s.ToContext(ctx)
+	}
+}
```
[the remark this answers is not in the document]
rafalbigaj: Thanks for pointing!

pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -59,11 +60,51 @@ func init() {
...
+			return false
+		}
+		_, ok := omitTaskGraph.Nodes[taskName]
+		return ok
+	}
```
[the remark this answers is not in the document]
rafalbigaj: That's a very good point 💯 .

I introduced `IsRunning` in `PipelineRunFacts`, which returns true if the `PipelineRun` is still running (non-final) tasks. Code seems to be much cleaner now.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -392,6 +393,15 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
 		return controller.NewPermanentError(err)
 	}
 
+	// once the list of final tasks is ready, check if pipeline is not stopped
+	if cfg.FeatureFlags.EnableAPIFields == apisconfig.AlphaAPIFields && pr.IsGracefullyCancelled() {
+		// If the pipelinerun is stopped, cancel tasks, but run finally
+		err := gracefullyCancelPipelineRun(ctx, logger, pr, dfinally, c.PipelineClientSet)
+		if err != nil {
+			return controller.NewPermanentError(err)
```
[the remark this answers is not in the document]
rafalbigaj: Thanks for catching up! 👍 

I added test for retry scenario.

rafalbigaj:
@pritidesai I like your suggestions very much, thanks! 🙏 

The reason "PipelineRunCancelled" is used only when status is set to "PipelineRunCancelled" now, "Cancelled" is set otherwise.

I added new `PipelineRun` condition reasons as well:

- `CancelledRunningFinally` - indicates that pipeline has been gracefully cancelled, but final tasks are still running
- `StoppedRunningFinally` - indicates that pipeline has been gracefully stopped, but final tasks are still running

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -271,6 +294,12 @@ const (
 	// PipelineRunReasonStopping indicates that no new Tasks will be scheduled by the controller, and the
 	// pipeline will stop once all running tasks complete their work
 	PipelineRunReasonStopping PipelineRunReason = "PipelineRunStopping"
+	// PipelineRunReasonCancelledRunningFinally indicates that pipeline has been gracefully cancelled
+	// and no new Tasks will be scheduled by the controller, but final tasks are now running
+	PipelineRunReasonCancelledRunningFinally PipelineRunReason = "CancelledRunningFinally"
+	// PipelineRunReasonStoppedRunningFinally indicates that pipeline has been gracefully stopped
+	// and no new Tasks will be scheduled by the controller, but final tasks are now running
+	PipelineRunReasonStoppedRunningFinally PipelineRunReason = "StoppedRunningFinally"
```
afrittoli, line 302: Nice! :)

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
```diff
@@ -583,17 +667,15 @@ func enableTektonOCIBundles(t *testing.T) func(context.Context) context.Context
...
-		featureFlags, _ := config.NewFeatureFlagsFromMap(map[string]string{
-			"enable-api-fields": "alpha",
-		})
-		cfg := &config.Config{
-			Defaults: &config.Defaults{
-				DefaultTimeoutMinutes: 60,
-			},
-			FeatureFlags: featureFlags,
-		}
-		return config.ToContext(ctx, cfg)
```
afrittoli, line 597: Thanks for the cleanup

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
rafalbigaj: Yes, @pritidesai suggested that 😄

pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -59,6 +60,39 @@ func init() {
...
 // cancelPipelineRun marks the PipelineRun as cancelled and any resolved TaskRun(s) too.
 func cancelPipelineRun(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, clientSet clientset.Interface) error {
+	errs := cancelPipelineTaskRuns(ctx, logger, pr, clientSet)
+
+	// If we successfully cancelled all the TaskRuns and Runs, we can consider the PipelineRun cancelled.
+	if len(errs) == 0 {
+		reason := ReasonCancelled
+		if pr.Spec.Status == v1beta1.PipelineRunSpecStatusCancelledDeprecated {
+			reason = ReasonCancelledDeprecated
+		}
```
pritidesai, line 70: thanks @rafalbigaj for adding this check 🙏 and setting it to deprecated reason when the deprecated status is set, it will be easier to cleanup and remove it from the project while upgarding to V1.

pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -80,17 +114,21 @@ func cancelPipelineRun(ctx context.Context, logger *zap.SugaredLogger, pr *v1bet
...
-		// update pr completed time
-		pr.Status.CompletionTime = &metav1.Time{Time: time.Now()}
-	} else {
+
+	return errs
+}
+
+// gracefullyCancelPipelineRun marks any non-final resolved TaskRun(s) as cancelled and runs finally.
+func gracefullyCancelPipelineRun(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, pipelineRunFacts *resources.PipelineRunFacts, clientSet clientset.Interface) error {
+	if !pipelineRunFacts.IsRunning() {
```
pritidesai, line 123: NIT: this check can be outside of this function, mainly do not call `gracefullyCancelPipelineRun` if `!pipelineRunFacts.IsRunning()` along with two other conditions in reconciler:

```
if cfg.FeatureFlags.EnableAPIFields == apisconfig.AlphaAPIFields && pr.IsGracefullyCancelled() {
```

The reason being, `pipelineRunFacts` can be avoided being in the list of params since its not used anywhere else besides this `if` condition in `gracefullyCancelPipelineRun`.

pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -80,17 +114,21 @@ func cancelPipelineRun(ctx context.Context, logger *zap.SugaredLogger, pr *v1bet
...
+		// There is no running DAG task to be cancelled. Skipping this action if already running finally.
+		return nil
+	}
+
+	errs := cancelPipelineTaskRuns(ctx, logger, pr, clientSet)
```
pritidesai, line 128: Also, wouldn't this now cancel the running finally tasks? 🤔 earlier we had a check here to see if a taskRun belongs to `finally` dag and if so, we did not cancel that particular taskRun.

pritidesai: sorry for confusion, I see how you changed this now. Do not attempt to cancel if any of the `dag` is not running, i.e. any of the finally is running. I think it might be better to check explicitly if any of the finally is running i.e `return nil if isFinallyRunning()` because `isRunning` returns `false` if there is no `taskRun` is associated with a `dag`.  It could be possible that nothing is scheduled yet and pipeline is just starting.

Please feel free to explain/counter it if I am not understanding it right.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -323,34 +346,49 @@ func (facts *PipelineRunFacts) GetPipelineConditionStatus(pr *v1beta1.PipelineRu
...
+			Message: message,
 		}
 	}
 
 	// Hasn't timed out; not all tasks have finished.... Must keep running then....
-	// transition pipeline into stopping state when one of the tasks(dag/final) cancelled or one of the dag tasks failed
-	// for a pipeline with final tasks, single dag task failure does not transition to interim stopping state
-	// pipeline stays in running state until all final tasks are done before transitioning to failed state
-	if s.Cancelled > 0 || (s.Failed > 0 && facts.checkFinalTasksDone()) {
+	switch {
```
pritidesai, line 381: this looks awesome, thank you @rafalbigaj for adding this 🙏

test/pipelinefinally_test.go
```diff
@@ -473,6 +480,246 @@ func TestPipelineLevelFinally_OneFinalTaskFailed_Failure(t *testing.T) {
...
+	}
+
+	// verify dag task succeeded and final task failed
+	for _, taskrunItem := range taskrunList.Items {
+		switch n := taskrunItem.Labels["tekton.dev/pipelineTask"]; n {
+		case "dagtask1":
+			if !isCancelled(t, n, taskrunItem.Status.Conditions) {
+				t.Fatalf("dag task %s should have been cancelled", n)
+			}
+		case "finaltask1":
```
pritidesai, line 591: we could add similar check on `finaltask2`, making sure that final task was skipped since its consuming a task result produced by cancelled dag task, please feel free to delay adding this check to a followup PR.

test/pipelinefinally_test.go
```diff
@@ -473,6 +480,246 @@ func TestPipelineLevelFinally_OneFinalTaskFailed_Failure(t *testing.T) {
...
+		switch n := taskrunItem.Labels["tekton.dev/pipelineTask"]; n {
+		case "dagtask1":
+			if !isSuccessful(t, n, taskrunItem.Status.Conditions) {
+				t.Fatalf("dag task %s should have succeeded", n)
+			}
+		case "finaltask1":
+			if !isSuccessful(t, n, taskrunItem.Status.Conditions) {
+				t.Fatalf("first final task %s should have succeeded", n)
+			}
+		case "finaltask2":
```
pritidesai, line 713: 🎉

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -1494,6 +1563,538 @@ func TestReconcileOnCancelledPipelineRun(t *testing.T) {
...
+	prs := []*v1beta1.PipelineRun{tb.PipelineRun("test-pipeline-run-cancelled-run-finally",
+		tb.PipelineRunNamespace("foo"),
+		tb.PipelineRunSpec("test-pipeline", tb.PipelineRunServiceAccountName("test-sa"),
+			tb.PipelineRunCancelledRunFinally,
+		),
+		tb.PipelineRunStatus(tb.PipelineRunStartTime(time.Now())),
+	)}
+	ps := []*v1beta1.Pipeline{tb.Pipeline("test-pipeline", tb.PipelineNamespace("foo"), tb.PipelineSpec(
+		tb.PipelineTask("hello-world-1", "hello-world"),
+		tb.PipelineTask("hello-world-2", "hello-world"),
```
pritidesai, line 1578: please make them into a `dag`, `hello-world-2` running after `hello-world-1` 🙏

pkg/reconciler/pipelinerun/pipelinerun_test.go
pritidesai: similar to the comment above, please make `hello-world-2` running after `hello-world-1` 🙏

pritidesai:
We are almost there, thanks a bunch once again for all the effort @rafalbigaj 

The changes look great and the pipeline state transitioning is much more meaningful now:

```
01-terminate-pipelinerun   Unknown     Running     4s
01-terminate-pipelinerun   False       PipelineRunCancelled   49s         43s

01-terminate-pipelinerun   Unknown     Running                   4s
01-terminate-pipelinerun   Unknown     StoppedRunningFinally     37s
01-terminate-pipelinerun   False       Cancelled                 39s

01-terminate-pipelinerun   Unknown     Running                8s
01-terminate-pipelinerun   Unknown     CancelledRunningFinally   18s
01-terminate-pipelinerun   False       Cancelled                 19s         21s

01-terminate-pipelinerun   Unknown     Running   4s
01-terminate-pipelinerun   False       Cancelled   31s         21s
```

I would like to hear your thoughts on `isRunning` before approving 🎉

pkg/reconciler/pipelinerun/cancel.go
rafalbigaj: Yes, that works well. Thanks!

pkg/reconciler/pipelinerun/pipelinerun_test.go
rafalbigaj: 👍

pkg/reconciler/pipelinerun/cancel.go
afrittoli: In case the has been no taskrun created yet `isRunning` will return false, so `gracefullyCancelPipelineRun` will not be executed, and that will continue to happen once final tasks are running, so unless I'm missing something, I think this specific part of the code is ok. @pritidesai wdyt?
@pritidesai comment made me thinking at the case where there is no taskrun in the state, and there are three possible cases that I can think can get us to that situation:

1. the pipeline is just starting - no taskrun created yet. In this case the graceful cancel method will not cancel anything, because there is nothing to cancel, which is ok. No new task from the DAG will be scheduled as this is checked for in the DAGExecutionQueue method, which is ok. However, no new task from finally will be scheduled either because GetFinalTask will only check whether all Tasks are done or at least one is failed, which won’t match is no task was cancelled. So I think this looks like an issue.
2. the pipeline is in pending mode: there is no check for a graceful cancel before the pending check, which means that a graceful cancel will be ignored on a pipeline in pending mode. I think this might be considered ok if there is a finally section to run, because the pipeline shouldn’t run anything (even finally) until it’s pending mode, but if there is no finally section it could make sense to cancel the pipeline. This is not a big deal, but it might be worth documenting that a pipeline in pending mode will need a forced cancel to be cancelled.
3. the pipeline was cancelled before any taskrun could be started, and it’s now running the finally section (this may only happen once the first issue is solved). For this one, once the first issue is solved, it should behave correctly

afrittoli: As @rafalbigaj correctly pointed out, `checkTasksDone` will return true is the DAG state is empty, so (1) - and thus (3) will work correctly today. I think we should document (2), but I'd be happy for it to be a follow up docs PR, as long as it happens in the same release.

afrittoli: One thing that we should fix in the future though is the behaviour of `checkDAGTasksDone`  and `checkFinalTasksDone`.  The current code reads:

```
// check if all DAG tasks done executing (succeeded, failed, or skipped)
func (facts *PipelineRunFacts) checkDAGTasksDone() bool {
	return facts.checkTasksDone(facts.TasksGraph)
}
```

We assume that if tasks are not in the state they were skipped, but I don't think that's correct?
This works todays anyways because of the way those functions are used, but I think we should fix the logic or at least the comments to match the actual behaviour.

afrittoli: If I'm not mistaken, the behaviour with no TaskRun is verified by `TestReconcileOnCancelledRunFinallyPipelineRunWithFinalTask`

rafalbigaj: @afrittoli yes, that's correct

afrittoli (commented):
Thanks for yet another update and for your patience!
I went through all the various tests and coverage again, and everything looks good to me.
I would be nice to have another quick docs PR on top to document that a pending pipeline cannot be cancelled via graceful cancel 🙏 
/approve

pkg/reconciler/pipelinerun/cancel.go
rafalbigaj: @pritidesai My understanding is that there is no need to cancel dag tasks if none is running. Therefore, `IsRunning` check seems to me to be quite natural.

pritidesai:
thank you once again @rafalbigaj for all the work and your patience, the changes look great. Very exciting for this 🎉 

/lgtm

PR #3922, opened 2021-05-07
Disable gocritic in pipelines

sbwsg:
# Changes

The test-runner docker image from plumbing now includes a newer version
of golangci-lint (1.39.0) which apparently switched on or changed some gocritic
configuration *somewhere*.

There are a _lot_ of syntax issues in pipelines that gocritic does not like,
including many in the generated code of `pkg/client`. Rather than hand-edit
everything or figure out all the flags required to disable the features this
commit simply disables gocritic so that PRs can get moving again.

We can re-enable it once we figure out what the necessary flags are to get
it working (or we fix all the syntax issues).

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,7 +9,6 @@ linters:
   - gofmt
   - goimports
   - gosec
-  - gocritic
   - golint
   - misspell
 output:


pritidesai:
/approve

thanks Scott 🎉

afrittoli:
Uhm, this is really strange, we should not run linters on `pkg/client` https://github.com/tektoncd/pipeline/blob/1e00760623cc4e12c2cab308b0d39e600b1877e9/.golangci.yml#L34 
Do you have an example link?

sbwsg:
> @sbwsg any link to a build failure ? as @afrittoli commented, `pkg/client` is skip in pipeline (in triggers, it's a bit [different](https://github.com/tektoncd/triggers/blob/main/.golangci.yml)).

No, none. I saw it while running `make golangci-lint` and trying to remove all the lint errors manually.

PR #3930, opened 2021-05-10
Change the way we run e2e tests with feature gates in pipelines

sbwsg:
# Changes

Prior to this commit pipelines ran e2e tests for both "stable" and
"alpha" feature gates in a single prow job. This makes the tests
take twice as long as running running either individually.

In a [parallel commit to Plumbing](https://github.com/tektoncd/plumbing/pull/833) we are introducing a new prow job
that is specifically for running alpha e2e tests. This means both
the alpha and stable tests can be run concurrently and hopefully
should reduce the iteration time on PRs - failures should show up
sooner.

The alpha feature gate prow job will set the PIPELINE_FEATURE_GATE
env var when it is initiated. Our e2e script recognizes this var
and sets the feature-flags configmap accordingly. If the env var
is not set then we default to running under "stable" feature gate.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -52,11 +52,14 @@ function run_e2e() {
   go_test_e2e -tags=examples -timeout=20m ./test/ || failed=1
 }
 
-set_feature_gate "stable"
-run_e2e
+if [ "$PIPELINE_FEATURE_GATE" == "" ]; then
+  set_feature_gate "stable"
+  run_e2e
+else
+  set_feature_gate "$PIPELINE_FEATURE_GATE"
+  run_e2e
+fi
 
-set_feature_gate "alpha"
-run_e2e
 
 (( failed )) && fail_test
 success


sbwsg:
Let's see if this works so soon after the plumbing merge:

sbwsg:
Re-running integration too to see if that now skips alpha:

sbwsg:
It looks like a timeout test failed because of cluster slowness. Let's rerun and see what happens.

sbwsg:
nice.  everything's passed and the integration runs each set their own feature gates as expected.

vdemeester (approved):

test/e2e-tests.sh
```diff
@@ -52,11 +52,14 @@ function run_e2e() {
   go_test_e2e -tags=examples -timeout=20m ./test/ || failed=1
 }
 
-set_feature_gate "stable"
-run_e2e
+if [ "$PIPELINE_FEATURE_GATE" == "" ]; then
```
afrittoli, line 55: NIT: We could also do: 
```shell
set_feature_gate "${PIPELINE_FEATURE_GATE:-stable}"
run_e2e
```

But this works too.

afrittoli (commented):
Thank you!!
/lgtm

PR #3937, opened 2021-05-11
Skip *heavy* validation on deletion 🙃

vdemeester:
# Changes

When deleting an object, we don't need to pursue all the validation
that we do at creation. It reduces the work to be done as part of the
validation *and* allows invalid version of the resource (from previous
versions for example) to be deleted safely.

The main reasoning for this change is to not face issues to reduce the situation where an upgrade makes object not deletable (because of *newly* invalid field). One example that happen in triggers is upgrading from 0.10.x to 0.12.x make `EventListener` not deletable at all.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Skip *heavy* validation on deletion in the webhook
```

### DIFF
--- a/pkg/apis/pipeline/v1alpha1/cluster_task_validation.go
+++ b/pkg/apis/pipeline/v1alpha1/cluster_task_validation.go
@@ -29,5 +29,8 @@ func (t *ClusterTask) Validate(ctx context.Context) *apis.FieldError {
 	if err := validate.ObjectMetadata(t.GetObjectMeta()); err != nil {
 		return err.ViaField("metadata")
 	}
+	if apis.IsInDelete(ctx) {
+		return nil
+	}
 	return t.Spec.Validate(ctx)
 }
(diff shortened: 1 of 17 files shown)


sbwsg:
> The main reasoning for this change is to not face issues to reduce the situation where an upgrade makes object not deletable (because of newly invalid field). One example that happen in triggers is upgrading from 0.10.x to 0.12.x make EventListener not deletable at all.

Nice.  This has also hit me when I'm developing a new feature.  Going back to the current HEAD of pipelines/main means that the taskruns with the new feature syntax cannot be deleted.

sbwsg (approved):
Cool. I don't think I've seen `IsInDelete` and [related funcs from knative](https://pkg.go.dev/knative.dev/pkg/apis#IsInDelete) before.

pritidesai:
this is interesting, `context` tells you whether a resource is being deleted 😲 

/lgtm

nikhil-thomas:
> this is interesting, `context` tells you whether a resource is being deleted 😲
> 
> /lgtm

Thanks for pointing it out. 🧑‍💻 👍 
very useful in tektoncd/operator work. 🙂

pritidesai:
would this work? 🤔 

vdemeester:
/retest
Somehow the jobs got killed..

PR #3941, opened 2021-05-12
Only fetch the definitions once 🧙

### LINKED ISSUE #3916: Run vs Definition "race" and "source of truth": `PipelineRun` errors out after `Pipeline` changed
# Expected Behavior

Once a `PipelineRun` starts, it does continue to the end (if success) no matter what happens to the referenced `Pipeline` (a change, …).

# Actual Behavior

If something happens to the `Pipeline` that is referred by the `PipelineRun`, the controller might fail the execution because incompatibilities, missing resource/parameter/… or even missing `Pipeline` (or `Task`).

1. Create a `PipelineRun` referencing a `Pipeline` with several `Task` (that take enough time for you to mutate the `Pipeline`)
2. Update the `Pipeline` with an additionnal param and an additonal `Task` that uses that param
3. See it fail on that new `Task` because of missing param

An even easier case to reproduce is
1. Create a `PipelineRun` referencing a `Pipeline` with several `Task` (that take enough time for you to mutate the `Pipeline`)
2. Delete the Pipeline
3. See the `PipelineRun` fail with `CouldntGetPipeline` error

I suspect the problem to be true even when using Tekton Bundle, if the tag changed between 2 *reconcile*, we may fetch the definition again (which is a performance problem, but this is kinda tracked in #3305.

# Additional Info

- Kubernetes version:

Any

- Tekton Pipeline version:

Any

# Possible solution

We are dereferencing the definitions in the `status` part of `PipelineRun` (and `TaskRun`). This could be used as the "source-of-truth" after the first reconciling "loop" and we wouldn't fetch the `Pipeline` definition anymore. 

This means:
- Making sure we do dereference definition from Tekton Bundles
- Change the behavior of fetching `Pipeline` definition to see if it's available in the `status`. If not, fetch the `Pipeline` definition, if not, use the definition in the `status` as the source of truth.

/assign

vdemeester:
# Changes

Since we "dereference" fetched definition and store them into the
`status`, we can use this as a source of truth. This allows us to
fetch only once the definitions (Task, Pipeline) *and* has the benefit
to make the runs (PipelineRun, TaskRun) immune to change to what they
refer, while the are executing.

Without this change, if you run a PipelineRun that reference a
Pipeline that is deleted during execution, your PipelineRun will fail
at some point because it cannot fetch the definition anymore, even if
it stored them in its status. This is the case for TaskRun too.

Fixes #3916

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Only fetch the definition once, and then used the spec stored in the status as source of truth. 
This reduce the probable race condition when a `PipelineRun` or a `TaskRun` refers to a `Pipeline` or `Task` that changes during its execution.
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelineref.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelineref.go
@@ -37,6 +37,18 @@ func GetPipelineFunc(ctx context.Context, k8s kubernetes.Interface, tekton clien
 	cfg := config.FromContextOrDefaults(ctx)
 	pr := pipelineRun.Spec.PipelineRef
 	namespace := pipelineRun.Namespace
+	// if the spec is already in the status, do not try to fetch it again, just use it as source of truth
+	if pipelineRun.Status.PipelineSpec != nil {
+		return func(_ context.Context, name string) (v1beta1.PipelineObject, error) {
+			return &v1beta1.Pipeline{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      name,
+					Namespace: namespace,
+				},
+				Spec: *pipelineRun.Status.PipelineSpec,
+			}, nil
+		}, nil
+	}
 	switch {
 	case cfg.FeatureFlags.EnableTektonOCIBundles && pr != nil && pr.Bundle != "":
 		// Return an inline function that implements GetTask by calling Resolver.Get with the specified task type and
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -207,8 +207,8 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 		return c.finishReconcileUpdateEmitEvents(ctx, pr, before, nil)
 	}
 
+	// If the pipelinerun is cancelled, cancel tasks and update status
 	if pr.IsCancelled() {
-		// If the pipelinerun is cancelled, cancel tasks and update status
 		err := cancelPipelineRun(ctx, logger, pr, c.PipelineClientSet)
 		return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
 	}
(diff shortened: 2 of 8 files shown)


vdemeester:
This probably needs some more discussions (in the API WG or somewhere) and some e2e tests.

vdemeester:
One thing to discuss here too is related to [this comment](https://github.com/tektoncd/pipeline/issues/3916#issuecomment-834322012) : should this be behind a configuration ? (and if yes, what should be the default behavior)

sbwsg:
This looks good to me!  Agree it needs testing but otherwise I'm happy to see this move forward.

I don't think this specific PR should be behind a flag.  From my pov this behaviour should simply be the way it works: if the specs are cached then we use them. A configuration option to turn off spec caching is then a separate feature request.

sbwsg (approved):

pkg/reconciler/pipelinerun/resources/pipelineref.go
```diff
@@ -37,6 +37,18 @@ func GetPipelineFunc(ctx context.Context, k8s kubernetes.Interface, tekton clien
 	cfg := config.FromContextOrDefaults(ctx)
 	pr := pipelineRun.Spec.PipelineRef
 	namespace := pipelineRun.Namespace
+	// if the spec is already in the status, do not try to fetch it again, just use it as source of truth
+	if pipelineRun.Status.PipelineSpec != nil {
```
pritidesai, line 41: sorry for naive question here, haven't looked at all the changes yet and until I recollect 😜 is it possible to have `nil`  status? do we need an additional check on `status != nil`?

vdemeester: `Status PipelineRunStatus `json:"status,omitempty"`` so if it is empty, the struct will be there (no way it can be nil 🙃)

pkg/reconciler/pipelinerun/resources/pipelineref.go
```diff
@@ -37,6 +37,18 @@ func GetPipelineFunc(ctx context.Context, k8s kubernetes.Interface, tekton clien
...
+		return func(_ context.Context, name string) (v1beta1.PipelineObject, error) {
+			return &v1beta1.Pipeline{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      name,
+					Namespace: namespace,
+				},
+				Spec: *pipelineRun.Status.PipelineSpec,
+			}, nil
+		}, nil
+	}
```
pritidesai, line 51: can we please add a unit test for this? 🥺

pritidesai: So this function is implemented to fetch the definition from the cluster or bundle reference but we are checking if the pipeline is already running i.e. status is not nil, return the stored specifications irrespective of bundle or no bundle.

pritidesai:
thanks @vdemeester the changes look great and if I can recollect the specifications are stored in the `status` after resolving `params` so they are ready to use as is after fetching. One request, it will be great to see a unit test for changes in `pipelineRef.go`

Also, we might have to apply similar changes to `run` in addition to `taskRun`, not necessarily in this PR, can be a separate effort.

pkg/reconciler/pipelinerun/resources/pipelineref.go
vdemeester: done 😉 
Yes, the pipeline doesn't really have to run though, just the pipelineSpec in status being populated

pritidesai:
thanks @vdemeester 

/lgtm

PR #3954, opened 2021-05-17
Add "format" support for mustParse test helpers 🧙

vdemeester:
# Changes

Allow using `mastParse` in a similar fashion to `fmt.Sprintf`, adding
support for formating and replacing variables in there.

/cc @tektoncd/core-maintainers @imjasonh 

The alternative (which is also really viable) is to not change it and rely on users doing `fmt.Sprintf` prior to passing the yaml.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/yaml.go
+++ b/test/yaml.go
@@ -17,6 +17,7 @@ limitations under the License.
 package test
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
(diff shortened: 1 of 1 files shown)


imjasonh (commented):
I wonder if we should jump straight to [`text/template`](https://golang.org/pkg/text/template/) instead. I worry a bit about the readability of something like:

```
tr := mustParseTaskRun(t, `
metadata:
  name: %s
  labels:
    my-label: %s
spec:
  taskRef:
    name: %s
`, name, label, taskName)
```

vs

```
tr := mustParseTaskRun(t, `
metadata:
  name: {{.name}}
  labels:
    my-label: {{.label}}
spec:
  taskRef:
    name: {{.taskName}}
`, map[string]string{
  "name": name,
  "label": label,
  "taskName": taskName,
})
```

`text/template` also opens the door to loops/conditionals/etc, but I'm not sure if that's a point for against introducing it. 🤔 

Once we choose fmt or template, we're a little bit stuck with it unless we split into `mustParseTaskFmt` and `mustParseTaskTemplate`.

In either case, I guess they're both possible by just fmting/templating before passing to `mustParseX`

sbwsg:
The longer the YAML gets the more frustrating ordered format specifiers are going to be wrt update and maintenance.

Also agree strongly with:

> The alternative (which is also really viable) is to not change it and rely on users doing fmt.Sprintf prior to passing the yaml.

vdemeester:
@imjasonh very good point, didn't really realized that (I had small yaml 😅).  Then I think we shouldn't touch these function but we could provide some based off of `text/template` indeed.

vdemeester:
Let's close this one for now.
/close

Issue #1569, opened 2019-11-14
Handle sidecar execution signalling with entrypoint injection


ImJasonH:
# Background

Today, Tekton injects an `entrypoint` binary into running steps, to control ordering of execution of those steps. Broadly, this is done by having the binary wait for files passed via the `--wait_file` flag, and having the binary write a file when it completes, specified by the `--post_file` flag.

In order to run steps A, B, C, the TaskRun controller creates a Pod with entrypoint binary configured like:

```
- image: entrypoint
  args: ['--post_file=a-done', 'stepA', 'stepA-args...'] # start immediately
- image: entrypoint
  args: ['--wait_file=a-done', '--post_file=b-done', 'stepB', 'stepB-args...'] # wait for A to finish
- image: entrypoint
  args: ['--wait_file=b-done', '--post_file=c-done', 'stepC', 'stepC-args...'] # wait for B to finish
```

Also: The entrypoint binary has been capable of waiting for multiple files since #1430 -- all specified files must exist before the actual step execution begins.

# Problem

Supporting Sidecars introduced a couple problems, which required additional ~~terrible hacks~~ creative workarounds.

First, stepA shouldn't start until all sidecar containers have started. This was accounted for by having `stepA`'s container mount a Downward-API-backed volume containing the value of a READY annotation on the Pod. When the controller sees that sidecar containers are up, it applies this annotation, triggering `stepA` to start (and so on).

Second, sidecars should be gracefully stopped once all TaskRun steps have completed successfully. This was accounted for by having the TaskRun controller update sidecar containers' `image:` field to a container image that exits successfully immediately, regardless of what args are passed (since K8s doesn't allow us to update `args` or `command`, etc., only `image`).

These "creative workarounds" work fine, but they require the TaskRun controller to have more control over and awareness of the underlying Pod than it should -- it would be nice to just have the controller create and watch Pods, and never have to update them (to add annotations, or update container `image` fields).

# Proposal

When converting a TaskRun spec to a Pod, the TaskRun controller should also inject the entrypoint binary into specified sidecar images, just as it does for steps. This allows it to have more control over sidecar execution ordering and signalling, and runs entirely within the Pod, not requiring any updates from outside.

[...]

PR #3956, opened 2021-05-18
Add support for experiemental hermetic execution mode to TaskRuns

priyawadhwa:
# Changes

This PR adds support for an experimental hermetic execution mode. If users specify this on their TaskRun, then all user step containers are run without network access. Any containers created or injected by tekton are not affected, and sidecar containers are not affected (since they aren't controller by the entrypointer, we can't make them hermetic as of now).

This PR also adds an integration test to make sure that network access isn't available when hermetic mode is enabled!

TEP: https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

Add support for experimental hermetic execution mode to TaskRuns

```release-note
Add support for experimental hermetic execution mode to TaskRuns
```

cc @dlorenc

### DIFF
--- a/cmd/entrypoint/runner.go
+++ b/cmd/entrypoint/runner.go
@@ -25,6 +25,7 @@ import (
 	"syscall"
 
 	"github.com/tektoncd/pipeline/pkg/entrypoint"
+	"github.com/tektoncd/pipeline/pkg/pod"
 )
 
 // TODO(jasonhall): Test that original exit code is propagated and that
@@ -57,6 +58,10 @@ func (rr *realRunner) Run(ctx context.Context, args ...string) error {
 	// dedicated PID group used to forward signals to
 	// main process and all children
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+
+	if os.Getenv("TEKTON_RESOURCE_NAME") == "" && os.Getenv(pod.TektonHermeticEnvVar) == "1" {
+		dropNetworking(cmd)
+	}
 
 	// Start defined command
 	if err := cmd.Start(); err != nil {
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -42,6 +42,13 @@ const (
 
 	// TaskRunLabelKey is the name of the label added to the Pod to identify the TaskRun
 	TaskRunLabelKey = pipeline.GroupName + pipeline.TaskRunLabelKey
+
+	// TektonHermeticEnvVar is the env var we set in containers to indicate they should be run hermetically
+	TektonHermeticEnvVar = "TEKTON_HERMETIC"
+	// ExecutionModeAnnotation is an experimental optional annotation to set the execution mode on a TaskRun
+	ExecutionModeAnnotation = "experimental.tekton.dev/execution-mode"
+	// ExecutionModeHermetic indicates hermetic execution mode
+	ExecutionModeHermetic = "hermetic"
 )
 
 // These are effectively const, but Go doesn't have such an annotation.
@@ -164,6 +171,16 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec 
 	if len(implicitEnvVars) > 0 {
 		for i, s := range stepContainers {
 			env := append(implicitEnvVars, s.Env...)
+			stepContainers[i].Env = env
+		}
+	}
+
+	// Add env var if hermetic execution was requested & if the alpha API is enabled
+	alphaAPIEnabled := config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields == config.AlphaAPIFields
+	if taskRun.Annotations[ExecutionModeAnnotation] == ExecutionModeHermetic && alphaAPIEnabled {
+		for i, s := range stepContainers {
+			// Add it at the end so it overrides
+			env := append(s.Env, corev1.EnvVar{Name: TektonHermeticEnvVar, Value: "1"})
 			stepContainers[i].Env = env
 		}
 	}
--- a/test/README.md
+++ b/test/README.md
@@ -12,7 +12,7 @@ go test ./...
 # Integration tests (against your current kube cluster)
 go test -v -count=1 -tags=e2e -timeout=20m ./test
 
-#conformance tests  (against your current kube cluster)
+# Conformance tests  (against your current kube cluster)
 go test -v -count=1 -tags=conformance -timeout=10m ./test
 ```
 
--- a/test/hermetic_taskrun_test.go
+++ b/test/hermetic_taskrun_test.go
@@ -1,0 +1,93 @@
+// +build e2e
+
+/*
+Copyright 2021 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TestHermeticTaskRun make sure that the hermetic execution mode actually drops network from a TaskRun step
+// it does this by first running the TaskRun normally to make sure it passes
+// Then, it enables hermetic mode and makes sure the same TaskRun fails because it no longer has access to a network.
+func TestHermeticTaskRun(t *testing.T) {
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	c, namespace := setup(ctx, t, requireAnyGate(map[string]string{"enable-api-fields": "alpha"}))
+	t.Parallel()
+	defer tearDown(ctx, t, c, namespace)
+
+	// first, run the task run with hermetic=false to prove that it succeeds
+	regularTaskRunName := "not-hermetic"
+	regularTaskRun := taskRun(regularTaskRunName, namespace, "")
+	t.Logf("Creating TaskRun %s, hermetic=false", regularTaskRunName)
+	if _, err := c.TaskRunClient.Create(ctx, regularTaskRun, metav1.CreateOptions{}); err != nil {
+		t.Fatalf("Failed to create TaskRun `%s`: %s", regularTaskRunName, err)
+	}
+	if err := WaitForTaskRunState(ctx, c, regularTaskRunName, Succeed(regularTaskRunName), "TaskRunCompleted"); err != nil {
+		t.Fatalf("Error waiting for TaskRun %s to finish: %s", regularTaskRunName, err)
+	}
+
+	// now, run the task mode with hermetic mode
+	// it should fail, since it shouldn't be able to access any network
+	hermeticTaskRunName := "hermetic-should-fail"
+	hermeticTaskRun := taskRun(hermeticTaskRunName, namespace, "hermetic")
+	t.Logf("Creating TaskRun %s, hermetic=true", hermeticTaskRunName)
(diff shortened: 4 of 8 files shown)


cmd/entrypoint/runner.go
```diff
@@ -58,6 +59,31 @@ func (rr *realRunner) Run(ctx context.Context, args ...string) error {
 	// main process and all children
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 
+	if os.Getenv("TEKTON_RESOURCE_NAME") == "" && os.Getenv(pod.HermeticEnvVar) == "1" {
+		cmd.SysProcAttr.Cloneflags = syscall.CLONE_NEWNS |
```
dlorenc, line 63: I think you can call the "dropNetworking" function instead - I actually did commit that part!

https://github.com/tektoncd/pipeline/blob/1f5980f8c8a05b106687cfa3e5b3193c213cb66e/cmd/entrypoint/namespaces_linux.go#L25

pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -66,6 +66,10 @@ type TaskRunSpec struct {
 	// Workspaces is a list of WorkspaceBindings from volumes to workspaces.
 	// +optional
 	Workspaces []WorkspaceBinding `json:"workspaces,omitempty"`
+	// ExecutionMode specifies the execution mode to run user-specified containers in
+	// (tekton injected containers will not run with this policy)
+	// +optional
+	ExecutionMode ExecutionMode `json:"executionMode,omitempty"`
```
dlorenc, line 72: We had discussed including this as an annotation (or maybe only in the v1alpha1 API only)? first, to indicate that it's experimental: https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md#implementation-plan

I'm not quite sure what the best approach would be here. The API docs don't really say either: https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md
Probably a good topic for the API working group.

pkg/pod/pod.go
```diff
@@ -168,6 +171,15 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec
 		}
 	}
 
+	// Add env var if hermetic execution was requested
+	if taskRun.Spec.ExecutionMode.Hermetic {
+		for i, s := range stepContainers {
+			// Add it at the end so it overrides
+			env := append(s.Env, corev1.EnvVar{Name: HermeticEnvVar, Value: "1"})
```
dlorenc, line 178: The order here is very important - it has to come at the end. A few unit tests would be useful to make sure people can't override it with their own env vars. We might also want to make the variable "TEKTON_HERMETIC" to indicate that it's a reserved word.

pkg/apis/pipeline/v1beta1/taskrun_types.go
afrittoli: We now have an feature gated API fields with related testing helpers that can be used to introduce "alpha" level features into the beta API: https://github.com/tektoncd/pipeline/tree/main/docs/developers#adding-feature-gated-api-fields

afrittoli:
In terms of documentation, it feels to me like hermetic execution mode would deserve a dedicated file in the `docs` folder.
We could then link to it from the various resource docs, `taskrun.md` for this PR.

pkg/apis/pipeline/v1beta1/taskrun_types.go
vdemeester: Agreed on discussing this in the API working group. We should try to not add a new field at first (I tend to like the annotation approach) ; and if we do, we should *hide* it behind the alpha api feature flag (cc @sbwsg )

dlorenc: +1, so something like:

* annotation behind alpha feature flag
* beta field behind alpha feature flag
* beta field without feature flag ?

test/hermetic_taskrun_test.go
```diff
@@ -0,0 +1,88 @@
...
+func taskRun(name, namespace string, hermetic bool) *v1beta1.TaskRun {
+	return &v1beta1.TaskRun{
+		ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace},
+		Spec: v1beta1.TaskRunSpec{
+			ExecutionMode: v1beta1.ExecutionMode{
+				Hermetic: hermetic,
+			},
+			Timeout: &metav1.Duration{Duration: time.Minute},
+			TaskSpec: &v1beta1.TaskSpec{
+				Steps: []v1beta1.Step{
```
sbwsg, line 73: Suggest adding another test to check that `Sidecars` are prevented from accessing the network too.

One slight wrinkle will be that a failing Sidecar doesn't result in a failed TaskRun. Checking the container status might be the way to go there?

test/hermetic_taskrun_test.go
```diff
@@ -0,0 +1,88 @@
...
+	regularTaskRun := taskRun(regularTaskRunName, namespace, false)
+	t.Logf("Creating TaskRun %s, hermetic=false", regularTaskRunName)
+	if _, err := c.TaskRunClient.Create(ctx, regularTaskRun, metav1.CreateOptions{}); err != nil {
+		t.Fatalf("Failed to create TaskRun `%s`: %s", regularTaskRunName, err)
+	}
+	if err := WaitForTaskRunState(ctx, c, regularTaskRunName, Succeed(regularTaskRunName), "TaskRunCompleted"); err != nil {
+		t.Fatalf("Error waiting for TaskRun %s to finish: %s", regularTaskRunName, err)
+	}
+
+	// now, run the task mode with hermetic mode
```
sbwsg, line 51: Would be great to migrate these comments into a docstring on the test func explaining what it does.

test/README.md
```diff
@@ -12,7 +12,7 @@ go test ./...
 # Integration tests (against your current kube cluster)
 go test -v -count=1 -tags=e2e -timeout=20m ./test
 
-#conformance tests  (against your current kube cluster)
+# Conformance tests  (against your current kube cluster)
```
sbwsg, line 15: 👍  thank you!

cmd/entrypoint/runner.go
priyawadhwa: oooh ty!

test/hermetic_taskrun_test.go
priyawadhwa: done!

pkg/pod/pod.go
priyawadhwa: SGTM

pkg/apis/pipeline/v1beta1/taskrun_types.go
priyawadhwa: sgtm, I switched to an annotation and put it behind the alpha feature flag

pkg/pod/pod.go
```diff
@@ -168,6 +171,15 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec
 		}
 	}
 
+	// Add env var if hermetic execution was requested
+	if taskRun.Spec.ExecutionMode.Hermetic {
+		for i, s := range stepContainers {
```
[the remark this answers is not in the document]
priyawadhwa: I think for now this alpha feature can only cover step containers, since the sidecar containers aren't controlled by the entrypointer (https://github.com/tektoncd/pipeline/issues/1569) -- without the entrypointer I'm not sure how to enforce a hermetic environment on the sidecar containers. 

If anyone has any suggestions pls let me know!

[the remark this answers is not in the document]
sbwsg: Oh I see!  Good point.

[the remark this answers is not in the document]
dlorenc: Ugh, I was hoping the sidecar KEP would be started by now but it looks like it missed the 1.22 train as well. A couple suggestions in the meantime:

* do nothing, document that sidecars are not hermetic
* disallow sidecars when hermetic mode is enabled

I can see pros and cons of each, let's start with 1. for alpha and get some feedback?

pkg/pod/pod.go
```diff
@@ -42,6 +42,13 @@ const (
 
 	// TaskRunLabelKey is the name of the label added to the Pod to identify the TaskRun
 	TaskRunLabelKey = pipeline.GroupName + pipeline.TaskRunLabelKey
+
+	// TektonHermeticEnvVar is the env var we set in containers to indicate they should be run hermetically
+	TektonHermeticEnvVar = "HERMETIC"
```
sbwsg, line 47: I might have missed a separate convo elsewhere; do we still want this to be `"TEKTON_HERMETIC"` to indicate it's a reserved name?

sbwsg (commented):
One more suggestion but otherwise this lgtm: We have [the "Alpha Features" section in `docs/install.md`](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features) - we could mention Hermetic builds there, and link to `docs/hermetic.md`?

sbwsg:
Awesome, cheers!

/lgtm

sbwsg:
Integration tests should be back in working order once https://github.com/tektoncd/pipeline/pull/3960 is merged.  You'll need to rebase once that's in.

priyawadhwa:
Thanks @sbwsg!

test/hermetic_taskrun_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TestHermeticTaskRun make sure that the hermetic execution mode actually drops network from a TaskRun step
+// it does this by first running the TaskRun normally to make sure it passes
+// Then, it enables hermetic mode and makes sure the same TaskRun fails because it no longer has access to a network.
+func TestHermeticTaskRun(t *testing.T) {
+	ctx := context.Background()
```
chmouel, line 35: I think we had that issue before for other test calling dropNetworking() 

if this test is running on a dev laptop environment that is not Linux, the test will fail. 

Since calling to dropNetworking which is only available on Linux https://github.com/tektoncd/pipeline/blob/567dce3cc/cmd/entrypoint/namespaces.go#L10 would panic()

chmouel: ah nm this is the e2e tests, i didn't realize so forget my comment :)

vdemeester: Right @chmouel this is not gonna be a problem. One problem though might be running this test against OpenShift or a cluster that enforce some lower privileges (default to user, drop some privileges) that might make the calls in `dropNetworking` to fail 🤔 I am not entirely sure though.

vdemeester (commented):
@priyawadhwa @dlorenc any reason for what is happening in `dropNetworking` to be in the entrypoint and not outside (on the spec level — enforce by the controller through an annotation or something) ? Does using this feature requires that the user that runs the container (`USER` in Dockerfile, …) to be root, or something else ?

test/hermetic_taskrun_test.go
chmouel: Not just OpenShift but i guess dropNetworking would be a privileged operation in general.

chmouel:
If we open the door to netns manipulation from entrypoint or via api perhaps we can get userns too 😍

dlorenc:
> @priyawadhwa @dlorenc any reason for what is happening in `dropNetworking` to be in the entrypoint and not outside (on the spec level — enforce by the controller through an annotation or something) ?

Not sure I understand this part - there's some discussion of other ways to do this in the TEP. The short answer is, there aren't any, or weren't when I last tried: https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md

> Does using this feature requires that the user that runs the container (`USER` in Dockerfile, …) to be root, or something else ?

There's some discussion on what is required here: https://github.com/tektoncd/pipeline/commit/b90a0efc4d7cfb4631a7ebb74de5f88b4df9dc2d#diff-e991ea79e58bc08e494f6f5f3bfbfeaa7d55b1bfbb68f90641500c1a7c890d21

In short - it's not exactly clear but it does work in most kernels. I think we can make it work without root ( if it doesn't already ) using some combination of user namespaces and setuid/dropping privleges on the binary itself.

dlorenc:
/lgtm

nice!!!

sbwsg:
The PR isn't merging and this might be either because we typically squash commits down to 1 prior to merge or it might mean a rebase is needed.

Suggest doing both of these and force pushing once more.  Hopefully once that's done tide should be content to merge.

priyawadhwa:
Thanks for the tip @sbwsg, I've rebased & squashed!

sbwsg:
re-adding the lgtm  🤞 

/lgtm

priyawadhwa:
Yay! Thanks everyone!

Issue #3871, opened 2021-04-07
Tekton converts double dollar symbols to one


MartinKanters:
# Expected Behavior

When the script element in a Tekton Task definition has double dollar symbols, it should remain as two dollar symbols.

# Actual Behavior

When the script element in a Tekton Task definition has double dollar symbols, the two symbols are reduced to one dollar symbol.

# Steps to Reproduce the Problem

1. Apply the Sample TaskRun manifest defined at the bottom of this description
2. See the output log of Tekton. (also at the bottom) 
3. I have verified that it is not a bug in the output log itself.

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", 
  GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-22T19:05:30Z", 
  GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
  Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", 
  GitCommit:"14f897abdc7b57f0850da68bd5959c9ee14ce2fe", GitTreeState:"clean", BuildDate:"2021-01-22T17:29:38Z", 
  GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

  `v0.22.0`

---

Sample TaskRun:

```
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  generateName: dollar-dollar-taskrun-
spec:
  taskSpec:
    steps:
      - name: dollar-dollar-demo
        image: alpine:latest
        script: |
          echo 'two dollar signs: $$'
          echo 'four dollar signs: $$$$'
          echo 'two dollar signs from an environment variable: ' $DOLLAR_DOLLAR_ENV
        env:
          - name: DOLLAR_DOLLAR_ENV
            value: '$$'
```
Generated output (The lines prefixed with `+` are the result of the `set -x` which is added implicitly by Tekton):
```
+ echo 'two dollar signs: $'
two dollar signs: $
+ echo 'four dollar signs: $$'
four dollar signs: $$
[...]

PR #3963, opened 2021-05-20
Encode scripts as base 64 to avoid k8s mangling "$$"

sbwsg:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/3871

Kubernetes replaces instances of "$$" in container args fields with "$". This can muck up the contents of script fields because scripts are passed into a TaskRun Pod as an arg to an init container. In particular, `$$` is used for the current process id in bash scripts.

Prior to this commit we [tried to prevent the replacement](https://github.com/tektoncd/pipeline/pull/3888) from happening by:

1. putting scripts into annotations on a pod and projecting them using downward
API
    - con: the max size of the annotations map is capped to ~250kB. The aggregate size of all scripts in a single Task therefore becomes constrained by this. Any other systems using annotations will reduce the available headroom. Backwards incompatible.

2. replacing instances of "$$" in scripts with "$$$$" for k8s to then process
back to "$$"
    - con: k8s doesn't actually process _all_ instances of "$$". For example, if you write an arg with format "echo $(eval \$$foo)" then k8s will see the first "$(", assume it's a variable reference, and pass it through verbatim. So user's scripts with bash variable become broken by tekton's new replacement. Backwards incompatible.

This commit takes a third approach, proposed by @MartinKanters, encoding scripts as base64 in the controller and then having them decoded in the init container. This bypasses Kubernetes' args processing completely because dollar
signs aren't used in the default base64 alphabet.

This approach avoids introducing a backwards-incompatible limit to the aggregate script size and doesn't mangle existing bash scripts with variable replacements.

The most noticeable trade-offs we now make are:

1. Tiny scripts can be up to 300% bigger, but as scripts get longer the max increase gets closer to 133%.
2. Also the TaskRun's `initContainer` YAML is a bit less human readable:

```
    initContainers:
    - args:
      - -c
      - |
        tmpfile="/tekton/scripts/script-0-f8fmf"
        touch ${tmpfile} && chmod +x ${tmpfile}
        cat > ${tmpfile} << '_EOF_'
        IyEvYmluL3NoCnNldCAteGUKZWNobyAibm8gc2hlYmFuZyI=
        _EOF_
        /tekton/tools/entrypoint decode-script "${tmpfile}"
```

The entrypoint is extended to decode base64 files so that the `shellImage`
(which is used to write scripts to disk for Step containers) is not required
to package a `base64` binary.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Scripts in Tasks are now written into the Task's pod using base64 to avoid kubernetes' built-in arg processing. This means they're a little larger than they were prior to this release but otherwise should continue working as expected.
```

### DIFF
--- a/cmd/entrypoint/subcommands/cp.go
+++ b/cmd/entrypoint/subcommands/cp.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2020 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package subcommands
+
+import (
+	"io"
+	"os"
+)
+
+const CopyCommand = "cp"
+
+// Owner has permission to write and execute, and anybody has
+// permission to execute.
+const dstPermissions = 0311
+
+// cp copies a files from src to dst.
+func cp(src, dst string) error {
+	s, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer s.Close()
+
+	d, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE, dstPermissions)
+	if err != nil {
+		return err
+	}
+	defer d.Close()
+
+	_, err = io.Copy(d, s)
+	return err
+}
--- a/cmd/entrypoint/subcommands/decode_script.go
+++ b/cmd/entrypoint/subcommands/decode_script.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2020 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package subcommands
+
+import (
+	"bytes"
+	"encoding/base64"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+)
+
+const DecodeScriptCommand = "decode-script"
+
+// decodeScript rewrites a script file from base64 back into its original content from
+// the Step definition.
+func decodeScript(scriptPath string) error {
+	decodedBytes, permissions, err := decodeScriptFromFile(scriptPath)
+	if err != nil {
+		return fmt.Errorf("error decoding script file %q: %w", scriptPath, err)
+	}
+	err = ioutil.WriteFile(scriptPath, decodedBytes, permissions)
+	if err != nil {
+		return fmt.Errorf("error writing decoded script file %q: %w", scriptPath, err)
+	}
+	return nil
+}
+
+// decodeScriptFromFile reads the script at scriptPath, decodes it from
+// base64, and returns the decoded bytes w/ the permissions to use when re-writing
+// or an error.
+func decodeScriptFromFile(scriptPath string) ([]byte, os.FileMode, error) {
+	scriptFile, err := os.Open(scriptPath)
+	if err != nil {
+		return nil, 0, fmt.Errorf("error reading from script file %q: %w", scriptPath, err)
+	}
+	defer scriptFile.Close()
+
+	encoded := bytes.NewBuffer(nil)
+	if _, err = io.Copy(encoded, scriptFile); err != nil {
+		return nil, 0, fmt.Errorf("error reading from script file %q: %w", scriptPath, err)
+	}
+
+	fileInfo, err := scriptFile.Stat()
+	if err != nil {
+		return nil, 0, fmt.Errorf("error statting script file %q: %w", scriptPath, err)
+	}
+	perms := fileInfo.Mode().Perm()
+
+	decoded := make([]byte, base64.StdEncoding.DecodedLen(encoded.Len()))
+	n, err := base64.StdEncoding.Decode(decoded, encoded.Bytes())
+	if err != nil {
+		return nil, 0, fmt.Errorf("error decoding script file %q: %w", scriptPath, err)
+	}
+	return decoded[0:n], perms, nil
+}
--- a/cmd/entrypoint/subcommands/subcommands.go
+++ b/cmd/entrypoint/subcommands/subcommands.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2020 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package subcommands
+
+import (
+	"fmt"
+)
+
+type SubcommandSuccessful struct {
+	message string
+}
+
+func (err SubcommandSuccessful) Error() string {
+	return err.message
+}
+
+type SubcommandError struct {
+	subcommand string
+	message    string
+}
+
+func (err SubcommandError) Error() string {
+	return fmt.Sprintf("%s error: %s", err.subcommand, err.message)
+}
+
+// Process takes the set of arguments passed to entrypoint and executes any
+// subcommand that the args call for. An error is returned to the caller to
+// indicate that a subcommand was matched and to pass back its success/fail
+// state. The returned error will be nil if no subcommand was matched to the
+// passed args, SubcommandSuccessful if args matched and the subcommand
+// succeeded, or any other error if the args matched but the subcommand failed.
+func Process(args []string) error {
+	if len(args) == 0 {
+		return nil
+	}
+	switch args[0] {
+	case CopyCommand:
+		// If invoked in "cp mode" (`entrypoint cp <src> <dst>`), simply copy
+		// the src path to the dst path. This is used to place the entrypoint
+		// binary in the tools directory, without requiring the cp command to
+		// exist in the base image.
+		if len(args) == 3 {
+			src, dst := args[1], args[2]
+			if err := cp(src, dst); err != nil {
+				return SubcommandError{subcommand: CopyCommand, message: err.Error()}
+			}
+			return SubcommandSuccessful{message: fmt.Sprintf("Copied %s to %s", src, dst)}
+		}
+	case DecodeScriptCommand:
+		// If invoked in "decode-script" mode (`entrypoint decode-script <src>`),
+		// read the script at <src> and overwrite it with its decoded content.
+		if len(args) == 2 {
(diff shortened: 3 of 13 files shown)


sbwsg:
TODO:
- [ ] Split `decodeScript` into a few pieces: decodeFromFile, writeFile
- [ ] Add unit tests for decodeFromFile (maybe using new `fs` package?)

examples/v1alpha1/taskruns/step-script.yaml
```diff
@@ -97,3 +97,38 @@ spec:
...
+          exit(3)
+        if '$$$$' != '\u0024\u0024\u0024\u0024':
+          print('four dollar signs ($$$$) are not passed through as expected :(')
+          exit(3)
+        print('dollar signs appear to be handled correctly! :)')
+
+    # Test that bash scripts with variable evaluations work as expected
+    - name: bash-variable-evaluations
+      image: bash:5.1.8
+      workingDir: /workspace
```
MartinKanters, line 123: Is this workspace needed here? (also for the v1beta1 version)

sbwsg: Ah good catch, it isn't needed when running as root.  I think I was playing around with different `runAsUser` settings and added this for non-root.  Removed from both examples, cheers!

cmd/entrypoint/subcommands/subcommands.go
```diff
@@ -0,0 +1,60 @@
...
+		// If invoked in "decode-script" mode (`entrypoint decode-script <src>`),
+		// read the script at <src> and overwrite it with its decoded content.
+		if len(args) == 2 {
+			src := args[1]
+			if err := decodeScript(src); err != nil {
+				return SubcommandError{subcommand: DecodeScriptCommand, message: err.Error()}
+			}
+			return SubcommandSuccessful{message: fmt.Sprintf("Decoded script %s", src)}
+		}
+	default:
```
bobcatfish, line 57: does this mean invalid args are ignored? (or does this just stop it from processing something that doesnt look like a subcommand?)

bobcatfish:
ugh this was a surprisingly tricky one!!! ps. thanks for the FABULOUS commit message 😍 !!!

> k8s doesn't actually process all instances of "$$"

🙃 


image

feels like these could potentially benefit from some test coverage, what do you think? im thinking especially [subcommands](https://00f74ba44ba6d8aadc4a5c8ab9c8b84204d85aecf2-apidata.googleusercontent.com/download/storage/v1/b/tekton-prow/o/pr-logs%2Fpull%2Ftektoncd_pipeline%2F3963%2Fpull-tekton-pipeline-go-coverage%2F1396061799480037377%2Fartifacts%2Fline-cov.html?jk=AFshE3V7TRPwB6KAptcP7Vb9N6J2R4BNhDJZUj6gsbReKEOdo6hf0P-f5qqk0V9iIxTHW1GMIeXrkYv1pxLABovGw5k41cBF35FYIt-uxu670Lxkfuc-sOT4QdYvUgKv_bDozzAl6RDW13HIo2d9NWjjyNNqTJ2jlSU_8Eik79aIDcRlXwJqV4GHUyfJYnyzsgoL1p-tMK-ga3nW5eh444-8hq1cQHxT0iogQ-GYItifbJJ70BGaP1FXVkVRX9_AsTgyGb9lOq6DJZ7PmwqgU3YBdJ7G4UaxjJ8tDGf6Pi-FD6WtP4rI_w7nwc2T0k0hMB3GFL-zWtTERVquesQfUGOQVHkADQ0xxOKSxuDmnnPWOoQA2D8SYRBmH_nROGZ3H4PHSfx5d1DJLodoOd3Vj-AOOcyhaxqaGKb4YsV6PlGt_-a06ZG0IRu3JK2F4horPeSjPV4IS0D8ZmUWFm-utZ6TReWqDCOeXte03CTJzahegAR5Yyq827CD4FpAOI9MOXbNO92IPYWoEaVYz9GgHS8-icLx91iZuhtXRodtrD1Tg440LXgsdcJKykUKhhNTL2R_ehOxo2dmW7hn-Y-fRF6fCqjdUiWfVwUV2Sr-8QEeL6jVxx33yqo9uGwuFZ2v9NuJ9XYnxx2RHUxIn2h5kz4XO2WTDCcIefpgb7RVDwqQVxSc7TExdgDTFuVhSfloMvIth_gvvyTGcDWiAWHhO0T4RSVLtUrr6AyIwvygPdr64KO_TOJ3k-EmQDdYyHFIQuQn-zfFg2IDGuW0ltSIOwI2RZfo2iLjA-qgXtrakzFRXWJlf59TdFcP7V7V-tuDbc2sGfapG2vhHu5cZX-uFdNqdi06PNMBNwoZ6tPQuWuRr78ArBseBORDBYtQmJiwAwVGDrJ954QQnnJZ7EbjL20lFoKSpZJ9kEH1C2tKaOngyjjZF0hjL0pLKL3UE0Q4RGfst79gnkVUZIJkoNolRPkMP6m5ANiKxiE_gvHY8mkZTkglH9l2PDZWjie2heHDP5xu7o_pi9l3Px2Z6DyDRYzWq_KArDFYZ_VtnCqE52PSQ_p7RgDg32iHO4G8XCgtnKbBXrYrwcUp8uqiYQbeFDQHCw4&isca=1#file3)

just lemme know if you feel its not worth it, i noticed you included unit tests for decode script so maybe you felt it wasnt worth it to maintain unit tests for these other tiny files - i think it's the indexing into the `args` object in `Process` that has me really feeling like "maybe some unit tests"

cmd/entrypoint/subcommands/subcommands.go
sbwsg: Yeah, just trying to capture the fact that the default operation is to "do nothing, return nil".

sbwsg (commented):
Ah yeah good point, I added tests for `decodeScript` because I wanted to try out the new `fstest` built-in package, unfortunately it only lets you mock a read-only filesystem so I didn't bother writing additional new tests for `cp` and `subcommands` as they both write. I'll add some for these using a `TempDir`.

sbwsg:
"Build still running" for both failed jobs.

sbwsg:
Tests have been updated to cover `subcommands.go` and `cp.go` in addition to `decode_script.go`.

sbwsg:
The build results say the alpha tests succeeded. Trying again.

cmd/entrypoint/subcommands/decode_script.go
```diff
@@ -0,0 +1,56 @@
+package subcommands
+
+import (
+	"bytes"
+	"encoding/base64"
+	"fmt"
+	"io"
+	"io/fs"
```
vdemeester, line 8: small note: this forces us to us Go 1.16 to build tekton 📦 
cc @nikhil-thomas @chmouel @piyush-garg

vdemeester (commented):
SGTM 👍🏼

chmouel:
Pretty neat @sbwsg  and thanks for the great commit message, did base-122 was investigated for size?

This blog post details it, https://blog.kevinalbs.com/base122 

I was wondering if the argument had a size limit, but it seems that k8/container runtime is limited by what the linux kernel is providing, which is pretty much unlimited as the system provide it,  from the manpage https://linux.die.net/man/2/execve : 

```
On kernel 2.6.23 and later, most architectures support a size limit derived from the soft RLIMIT_STACK resource limit (see getrlimit(2)) that is in force at the time of the execve() call. (Architectures with no memory management unit are excepted: they maintain the limit that was in effect before kernel 2.6.23.) This change allows programs to have a much larger argument and/or environment list. For these architectures, the total size is limited to 1/4 of the allowed stack size. (Imposing the 1/4-limit ensures that the new program always has some stack space.) Since Linux 2.6.25, the kernel places a floor of 32 pages on this size limit, so that, even when RLIMIT_STACK is set very low, applications are guaranteed to have at least as much argument and environment space as was provided by Linux 2.6.23 and earlier. (This guarantee was not provided in Linux 2.6.23 and 2.6.24.) Additionally, the limit per string is 32 pages (the kernel constant MAX_ARG_STRLEN), and the maximum number of strings is 0x7FFFFFFF.
```

cmd/entrypoint/subcommands/decode_script.go
chmouel: I don't think we would be able to upgrade to go1.16 for productized RHEL for quite some time,

sbwsg: ahhh ok good to know! i will tackle this without the fs mock.

sbwsg:
Cheers @chmouel I hadn't considered using a different encoding like Base 122! My only hesitation is that the design and code are controlled by a single person it looks like, and it sounds from the readme that the author still considers the design a work-in-progress.

Given that we're not exposing the encoding as part of our API we could always iterate on it and try out better / more compact representations over time if users complain about size or other issues.

cmd/entrypoint/subcommands/decode_script.go
sbwsg: Removed the features that required go 1.16 and have tested with go1.15.13 to confirm everything's passing as expected.

afrittoli (commented):
Thanks for this, it looks good - and better than all previous options.
Great PR description!!
/approve

chmouel:
/lgtm

thanks @sbwsg , let's hope this fix our variables encoding issues 🤞🏻

sbwsg:
oops.  I'm missing a file truncate somewhere in the switch back to the `os` package.

vdemeester (approved):
/lgtm

sbwsg:
OK, decoded script files should be correctly truncated now.  With my previous change the script file was ending up as the decoded script + the tail of the base64 string.

cmd/entrypoint/subcommands/cp.go
```diff
@@ -0,0 +1,30 @@
+package subcommands
```
pritidesai, line 1: I love this refactoring ❤️  awesome work @sbwsg 👏

pritidesai, line 1: Missing copyright license headers 😜

sbwsg: 👍 thanks @pritidesai ; added the license headers.

pkg/pod/script.go
```diff
@@ -107,22 +108,18 @@ func convertListOfSteps(steps []v1beta1.Step, initContainer *corev1.Container, p
...
-		// randomly generate a string to (hopefully) prevent
-		// collisions.
-		heredoc := names.SimpleNameGenerator.RestrictLengthWithRandomSuffix(fmt.Sprintf("%s-heredoc-randomly-generated", namePrefix))
+		heredoc := "_EOF_" // underscores because base64 doesnt include them in its alphabet
 		initContainer.Args[1] += fmt.Sprintf(`tmpfile="%s"
 touch ${tmpfile} && chmod +x ${tmpfile}
 cat > ${tmpfile} << '%s'
 %s
 %s
+/tekton/tools/entrypoint decode-script "${tmpfile}"
```
pritidesai, line 122: NIT: can we use something more meaningful here instead of `tmpfile`, may be a script file or a encoded script file? looks odd to see a `tmpfile` in pod description:

```
kubectl get pod/dollar-dollar-taskrun-lztnb-pod-tvs2h -o json |  jq .spec.initContainers
...
{
    "args": [
      "-c",
      "tmpfile=\"/tekton/scripts/script-0-ck986\"\ntouch ${tmpfile} && chmod +x ${tmpfile}\ncat > ${tmpfile} << '_EOF_'\nIyEvYmluL3NoCnNldCAteGUKZWNobyAndHdvIGRvbGxhciBzaWduczogJCQnCmVjaG8gJ2ZvdXIgZG9sbGFyIHNpZ25zOiAkJCQkJwplY2hvICd0d28gZG9sbGFyIHNpZ25zIGZyb20gYW4gZW52aXJvbm1lbnQgdmFyaWFibGU6ICcgJERPTExBUl9ET0xMQVJfRU5WCg==\n_EOF_\n/tekton/tools/entrypoint decode-script \"${tmpfile}\"\n"
...
```

pritidesai:
> Also the TaskRun's initContainer YAML is a bit less human readable:

It's a very strong disadvantage but there is no way around to deal with such special characters. 

thanks @sbwsg

/lgtm

pkg/pod/script.go
sbwsg: good call, created https://github.com/tektoncd/pipeline/pull/4041

Issue #3965, opened 2021-05-20
Hermekton Beta Tracking Issue


priyawadhwa:
This is an overarching issue we can use to track the different things we'll need to do for hermekton beta:

- [ ] Figure out how we want to handle sidecar containers. We can either just document that they aren't supported or actually fail if hermetic execution mode is applied on a Task with sidecar containers
- [ ] Add sample working tasks & pipelines
- [ ] Make sure hermekton works if the user isn't running as root (more details in https://github.com/tektoncd/pipeline/pull/3956#issuecomment-844298352)
- [ ] Check if this will work across most major kernel versions
- [ ] Documentation on limitations/generally make sure docs are up to date
- [ ] Remove the alpha feature flag requirement for hermekton

Relevant Links:
TEP: https://github.com/tektoncd/community/blob/main/teps/0025-hermekton.md
Hermekton alpha PR: #3956

imjasonh:
>  Figure out how we want to handle sidecar containers. We can either just document that they aren't supported or actually fail if hermetic execution mode is applied on a Task with sidecar containers

One option, at least for declared sidecars, would be to inject the entrypoint even if we don't otherwise use it to start the container. That wouldn't help with injected sidecars, but then we could just detect/block injected sidecars, instead of blocking all sidecars.

priyawadhwa:
@imjasonh wouldn't we have to use the entrypointer to start the container as well? Right now hermekton basically takes the command the entrypointer would run and runs it in a namespace without network access, so I think we'd need the entrypointer to start sidecars as well for this to work-- 

https://github.com/tektoncd/pipeline/blob/145043aece13ed928425defc7ac2b36e30f5f631/cmd/entrypoint/runner.go#L63

PR #3973, opened 2021-05-25
Add hermetic test running as non-root user

priyawadhwa:
Add a hermekton test running as a non-root user. ref https://github.com/tektoncd/pipeline/issues/3965

# Changes

/kind misc
/release-note-none

# Submitter Checklist

  functionality, content, code)

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,7 +40,7 @@ require (
 	k8s.io/gengo v0.0.0-20201214224949-b6c5ce23f027 // indirect
 	k8s.io/klog v1.0.0
 	k8s.io/kube-openapi v0.0.0-20210113233702-8566a335510f
-	k8s.io/utils v0.0.0-20210111153108-fddb29f9d009 // indirect
+	k8s.io/utils v0.0.0-20210111153108-fddb29f9d009
 	knative.dev/pkg v0.0.0-20210331065221-952fdd90dbb0
 )
 
--- a/test/hermetic_taskrun_test.go
+++ b/test/hermetic_taskrun_test.go
@@ -20,12 +20,14 @@ package test
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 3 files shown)


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/tektoncd/pipeline
 go 1.13
 
 require (
+	github.com/AlekSi/pointer v1.1.0 // indirect
```
sbwsg, line 6: It looks like the test is using [`k8s.io/utils/pointer`](https://github.com/kubernetes/utils/blob/da695404a2bc/pointer/pointer.go#L114).  I'm not 100% sure but it doesn't look like this module is used.

sbwsg (approved):

go.mod
sbwsg: The code for the test file [is importing `k8s.io/utils/pointer`](https://github.com/tektoncd/pipeline/pull/3973/files#diff-2b91646885f5ea26b953eaab2bc73293cb29ba32a0de26c46d3f2521f558d8a5R30) for those funcs I think, isn't it?  I might be totally misunderstanding what's happening.

priyawadhwa: Oh I didn't even see that ! Thanks for pointing it out, fixed 😄

test/hermetic_taskrun_test.go
```diff
@@ -91,3 +110,35 @@ apt-get install -y curl`,
...
+		},
+		Spec: v1beta1.TaskRunSpec{
+			Timeout: &metav1.Duration{Duration: time.Minute},
+			TaskSpec: &v1beta1.TaskSpec{
+				Steps: []v1beta1.Step{
+					{
+						Container: corev1.Container{
+							Name:  "curl",
+							Image: "gcr.io/cloud-builders/curl",
+							SecurityContext: &corev1.SecurityContext{
```
pritidesai, line 130: thanks @priyawadhwa for the changes, can this `SecurityContext` be configured for a pod?

pritidesai:
very exciting, thanks @priyawadhwa 👍 

/lgtm

test/hermetic_taskrun_test.go
pritidesai: we can include an additional test with security context at the pod level ...

Issue #3978, opened 2021-05-26
Permission error with nonroot user in examples test


barthy1:
Several days ago I noticed that for s390x test env `run-steps-as-non-root` test started failing. Looks like permission problem for nonroot user. 
The error is ` chdir to cwd ("/home/nonroot") set in config.json failed: permission denied: unknown`. 

# Expected Behavior
`go test -v -count=1 -tags=examples -timeout=20m ./test -run ^TestExamples/v1beta1/taskruns/run-steps-as-non-root` should be executed successfully
# Actual Behavior
Pod is not initialized

```
NAMESPACE                NAME                                           READY   STATUS                    RESTARTS   AGE
arendelle-4jjgj          show-non-root-steps-run-qr8tj-pod-qjwjn        0/2     Init:ContainerCannotRun   0          6m25s
```

```
 Warning  Failed     20s   kubelet            Error: failed to start container "place-tools": Error response from daemon: OCI runtime create failed: container_linux.go:367: starting container process caused: chdir to cwd ("/home/nonroot") set in config.json failed: permission denied: unknown
```

Test passes if I use `65532` user in the [config](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/taskruns/run-steps-as-non-root.yaml#L36). Can the reason be in distroless nonroot base image? So far found the [issue](https://github.com/tektoncd/pipeline/issues/3273) which described similar behaviour.

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

```
# kubectl version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-13T13:28:09Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/s390x"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.5", GitCommit:"e6503f8d8f769ace2f338794c914a96fc335df0f", GitTreeState:"clean", BuildDate:"2020-06-26T03:39:24Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/s390x"}
```

- Tekton Pipeline version:

```
# kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'
v20210525-4ced3cb269
```

[...]

PR #3979, opened 2021-05-26
Setup WorkingDir for place-tools init container to avoid permission error

barthy1:
# Changes

With 1001 user `run-steps-as-non-root` test can fail because of missing access permissions, see https://github.com/tektoncd/pipeline/issues/3978

Possible solutions are to use 65532 user, which is configured one for gcr.io/distroless/static:nonroot base image or to change working directory to "/" from the default "/home/nonroot" 

"/" working directory allows to use not only 65532 user, so this path is chosen.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


sbwsg:
The way this change currently works it appears we're telling users that they may not use any UID in their TaskRun's `securityContext` which is not provided by the distroless image.  But that's an image Tekton is "forcing" on users by having it as the base of the `place-tools` init container.

One of the steps in this example uses uid `2000` but the ubuntu image does not seem to provide this user.  I wonder - does the same error appear when running this step?  I guess maybe not because the ubuntu image has the WORKDIR set to `/`, which is allowed for any UID even if they don't exist in `/etc/passwd`?

I don't think there's anything specific in `place-tools` that requires the workingDir to be `/home/nonroot`.  We could override this in Tekton to be be `/` instead?  That way hopefully a TaskRun's `securityContext` wouldn't trigger this error in our init containers.

pkg/pod/entrypoint.go
```diff
@@ -95,6 +95,10 @@ func orderContainers(entrypointImage string, commonExtraEntrypointArgs []string,
 	initContainer := corev1.Container{
 		Name:  "place-tools",
 		Image: entrypointImage,
+		// Rewrite default WorkingDir from "/home/nonroot" to "/"
+		// as suggested at https://github.com/GoogleContainerTools/distroless/issues/718
```
sbwsg, line 99: 👍 nice reference.

sbwsg (approved):

vdemeester (approved):
/lgtm

PR #4018, opened 2021-06-09
Fix Windows image entrypoint paths for parity with Linux

aiden-deloryn:
# Changes

Tekton expects to find the entrypoint binary in `/ko-app/entrypoint`
which is equivalent to `C:\ko-app\entrypoint` on Windows. See
`pkg/pod/entrypoint.go:orderContainers()`. Currently entrypoint is
stored in `C:\ProgramData\tektoncd\pipeline\entrypoint.exe` causing
this to break on Windows. To avoid the need for platform-specific
logic in the controller, and for parity with Linux, this change
updates 'entrypoint' and 'nop' images for Windows to store binaries
in `C:\ko-app\`.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/images/windows/entrypoint/Dockerfile
+++ b/images/windows/entrypoint/Dockerfile
@@ -6,11 +6,11 @@ COPY . c:/gopath/src/github.com/tektoncd/pipeline
 
 WORKDIR c:/gopath/src/github.com/tektoncd/pipeline
 
-RUN go build -o c:/gopath/bin/entrypoint.exe ./cmd/entrypoint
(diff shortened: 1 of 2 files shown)


sbwsg (approved):

imjasonh:
/lgtm

FWIW https://github.com/google/ko/pull/374 builds images with entrypoint configured to `c:\ko-app\entrypoint`, so this should be compatible when we start actually building Windows images with ko 👍

PR #4057, opened 2021-06-23
Merge *Run's PodTemplate with Default 🧙

### LINKED ISSUE #3569: Default Podtemplate getting overriden with the pipelinerun and taskrun level podtemplate
Right, now when you specify a default pod template and also specify pod template at pipelinerun/taskrun level. The default pod template is getting overridden by the one at the pipelinerun/taskrun level.

Is there any specific reason for it?

Is not doing the merge of both templates with giving precedence to one at taskrun/pipelinerun level be better so the user can reuse the default with some changes that can be specified at pipelinerun/taskrun level?

cc @vdemeester 

/kind question

vdemeester:
# Changes

Right, now when you specify a default pod template and also specify
pod template at pipelinerun/taskrun level. The default pod template is
getting overridden by the one at the pipelinerun/taskrun level.

This changes this behavior by only overriding the `PodTemplate` field
that are defined by the `TaskRun` and not the whole
`PodTemplate`. This means it is possible to define a `NodeSelector`
widely (through the default) without it being overriden when the user
sets a `SecurityContext` on his `TaskRun`.

This is also done on the `PipelineRun` PodTemplate.

Closes #3569

/cc @sbwsg @afrittoli @pritidesai @skaegi @bobcatfish 

*This probably need a bit more unit tests, but I think it's worth opening to discuss if the approach is correct (for each fields)*

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
TaskRun and PipelineRun's PodTemplate are now merge with the default PodTemplate. This means any field that is not specified in a TaskRun or PipelineRun's PodTemplate will come from the configured default PodTemplate (if defined).
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/taskrun_defaults.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_defaults.go
@@ -60,12 +60,66 @@ func (trs *TaskRunSpec) SetDefaults(ctx context.Context) {
 	}
 
 	defaultPodTemplate := cfg.Defaults.DefaultPodTemplate
-	if trs.PodTemplate == nil {
-		trs.PodTemplate = defaultPodTemplate
-	}
+	trs.PodTemplate = mergePodTemplateWithDefault(trs.PodTemplate, defaultPodTemplate)
 
 	// If this taskrun has an embedded task, apply the usual task defaults
 	if trs.TaskSpec != nil {
 		trs.TaskSpec.SetDefaults(ctx)
+	}
+}
+
+func mergePodTemplateWithDefault(tpl, defaultTpl *PodTemplate) *PodTemplate {
+	switch {
+	case defaultTpl == nil:
+		// No configured default, just return the template
+		return tpl
+	case tpl == nil:
+		// No template, just return the default template
+		return defaultTpl
+	default:
+		// Otherwise, merge fields
+		if tpl.NodeSelector == nil {
+			tpl.NodeSelector = defaultTpl.NodeSelector
+		}
+		if tpl.Tolerations == nil {
+			tpl.Tolerations = defaultTpl.Tolerations
+		}
+		if tpl.SecurityContext == nil {
+			tpl.SecurityContext = defaultTpl.SecurityContext
+		}
+		if tpl.Volumes == nil {
+			tpl.Volumes = defaultTpl.Volumes
+		}
+		if tpl.RuntimeClassName == nil {
+			tpl.RuntimeClassName = defaultTpl.RuntimeClassName
+		}
+		if tpl.AutomountServiceAccountToken == nil {
+			tpl.AutomountServiceAccountToken = defaultTpl.AutomountServiceAccountToken
+		}
+		if tpl.DNSPolicy == nil {
+			tpl.DNSPolicy = defaultTpl.DNSPolicy
+		}
+		if tpl.DNSConfig == nil {
+			tpl.DNSConfig = defaultTpl.DNSConfig
+		}
+		if tpl.EnableServiceLinks == nil {
+			tpl.EnableServiceLinks = defaultTpl.EnableServiceLinks
+		}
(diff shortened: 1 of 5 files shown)


vdemeester:
This also probably needs some docs 🙃 
/hold

sbwsg:
This looks good from my perspective.  👍  /approve

chmouel (commented):
/lgtm

bobcatfish:
@vdemeester you mentioned adding some docs, is this something we can do before we merge? (maybe at https://github.com/tektoncd/pipeline/blob/main/docs/podtemplates.md) 

Also looks like the coverage went down slightly - in both cases it looks like it's just one line per file not getting touched, maybe it's easy to add those to the test cases?

🙏

sbwsg:
Re-adding my approve.

/approve

pkg/apis/pipeline/v1beta1/taskrun_defaults.go
```diff
@@ -60,12 +60,66 @@ func (trs *TaskRunSpec) SetDefaults(ctx context.Context) {
...
 	if trs.TaskSpec != nil {
 		trs.TaskSpec.SetDefaults(ctx)
 	}
 }
+
+func mergePodTemplateWithDefault(tpl, defaultTpl *PodTemplate) *PodTemplate {
+	switch {
+	case defaultTpl == nil:
+		// No configured default, just return the template
+		return tpl
```
pritidesai, line 75: Could `tpl` be `nil` here? or not possible to have both `defaultTpl` and `tpl` to be `nil` at this point?

pritidesai: or don't bother to merge, just return `tpl` here irrespective of it being `nil`

pkg/apis/pipeline/v1beta1/taskrun_defaults.go
```diff
@@ -60,12 +60,66 @@ func (trs *TaskRunSpec) SetDefaults(ctx context.Context) {
...
+	case tpl == nil:
+		// No template, just return the default template
+		return defaultTpl
+	default:
+		// Otherwise, merge fields
+		if tpl.NodeSelector == nil {
+			tpl.NodeSelector = defaultTpl.NodeSelector
```
pritidesai, line 82: do we need a check here if `defaultTpl.NodeSelector` is not `nil` and likewise for the rest of the fields?

pritidesai: or again just return default irrespective of it being `nil`

pritidesai:
/lgtm

thanks @vdemeester

pritidesai:
tests seems flaky

/test tekton-pipeline-build-tests
/test tekton-pipeline-alpha-integration-tests

pkg/apis/pipeline/v1beta1/taskrun_defaults.go
vdemeester: Yeah if there is no default, we just return the taskrun template, be it nil or not 😛

pkg/apis/pipeline/v1beta1/taskrun_defaults.go
```diff
@@ -60,12 +60,66 @@ func (trs *TaskRunSpec) SetDefaults(ctx context.Context) {
...
+		}
+		if tpl.ImagePullSecrets == nil {
+			tpl.ImagePullSecrets = defaultTpl.ImagePullSecrets
+		}
+		if tpl.HostAliases == nil {
+			tpl.HostAliases = defaultTpl.HostAliases
+		}
+		if tpl.HostNetwork == false && defaultTpl.HostNetwork == true {
+			tpl.HostNetwork = true
+		}
```
Letty5411, line 122: maybe sometimes the taskrun don't want to use hostnetwork, and set it in tasktemplate explicitly, however in this code, if the global template's value is true, it will be covered.

Issue #3627, opened 2020-12-10
e2e TestGitPipelineRun test fails on newly installed pipeline


barthy1:
# Expected Behavior
`TestGitPipelineRun` test should pass

# Actual Behavior
`TestGitPipelineRun` test fails because of timeout. Corresponding taskrun is created but no pod is started.
Usually second attempt is successful.

```
# oc get tr -A
NAMESPACE         NAME                                     SUCCEEDED   REASON   STARTTIME   COMPLETIONTIME
arendelle-8nxtw   git-check-pipeline-run-git-check-72mth

# oc get pod -n arendelle-8nxtw
No resources found in arendelle-8nxtw namespace.
```

# Steps to Reproduce the Problem

1. Install k8s cluster using `kind`
2. Install nightly pipeline build
```
kubectl apply --filename https://storage.googleapis.com/tekton-releases-nightly/pipeline/latest/release.yaml
```
3. Run the test just after pipeline is installed
```
# go test -v -tags=e2e -count=1 ./test -run ^TestGitPipelineRun/tekton_@_master_with_PR_refspec
```
4. Wait 10mins and get the error 
```
# go test -v -tags=e2e -count=1 ./test -run ^TestGitPipelineRun/tekton_@_master_with_PR_refspec
Using kubeconfig at `/root/.kube/config` with cluster ``
Cluster information
- Version: v1.19.1
- BuildDate: 2020-09-14T07:30:52Z
- GoVersion: go1.15
- Platform: linux/amd64
=== RUN   TestGitPipelineRun
=== RUN   TestGitPipelineRun/tekton_@_master_with_PR_refspec
=== PAUSE TestGitPipelineRun/tekton_@_master_with_PR_refspec
=== CONT  TestGitPipelineRun/tekton_@_master_with_PR_refspec
    init_test.go:147: Create namespace arendelle-8nxtw to deploy to
    init_test.go:163: Verify SA "default" is created in namespace "arendelle-8nxtw"
    git_checkout_test.go:103: Creating Git PipelineResource git-source-resource
    git_checkout_test.go:119: Creating PipelineRun git-check-pipeline-run
panic: test timed out after 10m0s
...
FAIL	github.com/tektoncd/pipeline/test	600.088s
FAIL
```
# Additional Info

- Kubernetes version:
```
# kubectl version
[...]

PR #4073, opened 2021-07-02
Set fake AWS credentials on controller to workaround aws-sdk bug

sbwsg:
Workaround for issue https://github.com/tektoncd/pipeline/issues/4087 while we look into better solutions.

# Changes

Several issues have now reared their head which are directly caused
by an update to the aws-sdk. The update results in extremely long
delays in the execution of tasks after the Pipelines controller is
first deployed in a cluster. The aws-sdk is initialized through
a transitive dependency that pipelines pulls in via k8schain or go-containerregistry.

Here are the recent issues directly related to this bug:

- https://github.com/tektoncd/pipeline/issues/3627 (Since December!)
- https://github.com/tektoncd/pipeline/issues/4084

One quick way to work around this problem is to set fake AWS
credentials in the environment of the deployed controller. This
apparently causes the aws-sdk to skip whatever process it has
introduced that causes massive delays in execution. So this commit
does exactly that - set fake aws creds in the deployments env vars.

This is an unfortunate hack to try and mitigate the problem until
a better solution presents itself. Ideally go-containerregistry or
k8s-pkg-credentialprovider would provide a way for us to disable
AWS SDK initialization to avoid this misbehaviour.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Work around a bug in the AWS go SDK that causes extremely long delays in task startup times.
```

### DIFF
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -93,6 +93,16 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        # These phony AWS credentials are here to work around a bug in the aws go sdk
+        # that causes extremely long delays in the execution of tasks after the initial
+        # deployment of the Tekton Pipelines controller.
+        - name: AWS_ACCESS_KEY_ID
+          value: foobarbaz
+        - name: AWS_SECRET_ACCESS_KEY
+          value: foobarbaz
+        - name: AWS_DEFAULT_REGION
+          value: foobarbaz
+
         # If you are changing these names, you will also need to update
         # the controller's Role in 200-role.yaml to include the new
         # values in the "configmaps" "get" rule.


config/controller.yaml
```diff
@@ -93,6 +93,16 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        # These phony AWS credentials are here to work around a bug in the aws go sdk
```
bobcatfish, line 96: could we add an issue to track this and mention it in the comment? (either here or in go-containerregistry) 🙏 

basically just trying to make sure when someone stumbles on this years from now cuz they need to change it that they know where to go to see what the latest state is (i mean they can always track it through the blame but since this is kind of a hack workaround, it feels like we'd like to maybe pursue a long term solution at some point)

bobcatfish (commented):
Thanks for figuring this out @sbwsg ! This seems fine as a short term fix but I think we should at least track this somewhere so we can come back to it (i think you might be working on this already but id like to track it anyway just in case)

im also wondering what (if any) implication this might have for folks who are actually  using AWS credentials - might be worth asking if someone using AWS is willing to try this out and make sure it doesnt cause any issues for them

/approve

config/controller.yaml
imjasonh: +1, we should track this in its own parent issue and probably open or identify an aws-sdk-go issue that's at the bottom of this turtle stack.

imjasonh:
+1 on making this change as a short-term stopgap. Another route to fixing this that we should explore:

If aws-sdk-go has fixed the bug since [v1.31.12](https://github.com/aws/aws-sdk-go/releases/tag/v1.31.12) (which is the version we depend on, a year+ old), we should try to see if this bug goes away just by upgrading dependencies.

We currently depend on k8schain from January; we could upgrade to the latest release, v0.5.1, which depends on k8s-pkg-credentialprovider v1.21.0-1, which depends on aws-sdk-go v1.35.24 (November 2020)

The latest aws-sdk-go is v1.39.4 (two hours ago), we could see if [upgrading that dependency](https://github.com/vdemeester/k8s-pkg-credentialprovider/pull/13) then upgrading k8schain then upgrading tektoncd/pipeline fixes the bug.

sbwsg:
@bobcatfish @imjasonh I've created https://github.com/tektoncd/pipeline/issues/4087 to capture the problem with all the notes I've been piecing together.

sbwsg:
Updating the SDK may well help indeed, have just seen that https://github.com/aws/aws-sdk-go/pull/3066 from back in January reduces a  timeout in requests to the ec2 metadata service.  I am super confused how we are seeing timeout times of ~75 seconds per connection attempt while their timeout configuration only had it at 5 seconds prior to that PR though.  I might not be fully grokking what's actually going wrong here.

afrittoli:
@imjasonh @sbwsg I'd be happy to merge that asap even as a temporary workaround - so that it might still get into the next release or minor release - wdyt? @pritidesai

pritidesai:
this is really phony but as long as its a temporary workaround, I am fine with this, thanks a bunch @sbwsg 🙏 

/lgtm

PR #4075, opened 2021-07-06
support referencing array params in when expression values

### LINKED ISSUE #4061: Cannot pass array to spec.tasks[0].when[0].values
I am using Tekton in openshift 4.7 and want to the when clause in pipeline to have a dynamic list to switch off/on tasks. In my test both task should not run, but only the one with the hard coded list ["blue","yellow","green"] skips. Can someone guide to what to do. Thank you very much in advance.

 Pipeline
```yaml
---
kind: Pipeline
metadata:
  name: test
spec:
  workspaces:
    - name: psb-deployment-checkout
  params:
    - name: deployments
      type: array
      description: "array of deployment to do"
  tasks:
    - name: getpodno1
      taskRef:
        name: openshift-client
      params:
        - name: SCRIPT
          value: "oc get pods"
      when:
        - input: "red"
          operator: in
          values: ["blue","yellow","green"]
      workspaces:
        - name: output
          workspace: psb-deployment-checkout
    - name: getpodno2
      taskRef:
        name: openshift-client
      params:
        - name: SCRIPT
          value: "oc get pods"
      when:
        - input: "red"
          operator: in
          values: ["$(spec.params.deployments)"]
      workspaces:
        - name: output
          workspace: psb-deployment-checkout
```

PipelineRun

```yaml
---
kind: PipelineRun
metadata:
  name: testrun
spec:
  pipelineRef:
    name: test
  serviceAccountName: default
  params:
    - name: deployments
      value:
        - "green"
        - "yellow"
        - "blue"
  workspaces:
    - name: psb-deployment-checkout
      persistentVolumeClaim:
        claimName: psb-deployment-pvc
      subPath: git-checkout
```

pritidesai:
# Changes

Array params are referenced using $(params.arrayParam[*]). When expressions `values` is a list of `string` and was designed to reference and resolve an array parameter. But the implementation was limiting the array param usage in the when expression `values`. Adding support for the array params in when expressions.

/kind bug

Closes #4061 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
When expression values can have array parameter reference now, such as, values: [$(params.arrayParam[*])]
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
@@ -1262,6 +1262,22 @@ func TestValidatePipelineParameterVariables_Success(t *testing.T) {
 			}},
 		}},
 	}, {
+		name: "valid string parameter variables in input, array reference in values in when expression",
+		params: []ParamSpec{{
+			Name: "baz", Type: ParamTypeString,
+		}, {
+			Name: "foo", Type: ParamTypeArray, Default: &ArrayOrString{Type: ParamTypeArray, ArrayVal: []string{"anarray", "elements"}},
+		}},
+		tasks: []PipelineTask{{
+			Name:    "bar",
+			TaskRef: &TaskRef{Name: "bar-task"},
+			WhenExpressions: []WhenExpression{{
+				Input:    "$(params.baz)",
+				Operator: selection.In,
+				Values:   []string{"$(params.foo[*])"},
+			}},
+		}},
+	}, {
 		name: "valid array parameter variables",
 		params: []ParamSpec{{
 			Name: "baz", Type: ParamTypeArray, Default: &ArrayOrString{Type: ParamTypeArray, ArrayVal: []string{"some", "default"}},
@@ -1380,7 +1396,7 @@ func TestValidatePipelineParameterVariables_Failure(t *testing.T) {
 			Paths:   []string{"[0].when[0].input"},
 		},
 	}, {
-		name: "invalid string parameter variables in when expression, array reference in values",
+		name: "Invalid array parameter variable in when expression, array reference in input with array notation [*]",
 		params: []ParamSpec{{
 			Name: "foo", Type: ParamTypeArray, Default: &ArrayOrString{Type: ParamTypeArray, ArrayVal: []string{"anarray", "elements"}},
 		}},
@@ -1388,14 +1404,14 @@ func TestValidatePipelineParameterVariables_Failure(t *testing.T) {
 			Name:    "bar",
 			TaskRef: &TaskRef{Name: "bar-task"},
 			WhenExpressions: []WhenExpression{{
-				Input:    "bax",
+				Input:    "$(params.foo)[*]",
 				Operator: selection.In,
-				Values:   []string{"$(params.foo)"},
+				Values:   []string{"$(params.foo[*])"},
 			}},
 		}},
 		expectedError: apis.FieldError{
-			Message: `variable type invalid in "$(params.foo)"`,
-			Paths:   []string{"[0].when[0].values"},
+			Message: `variable type invalid in "$(params.foo)[*]"`,
+			Paths:   []string{"[0].when[0].input"},
 		},
 	}, {
 		name: "invalid pipeline task with a parameter combined with missing param from the param declarations",
--- a/pkg/apis/pipeline/v1beta1/when_types_test.go
+++ b/pkg/apis/pipeline/v1beta1/when_types_test.go
@@ -261,7 +261,7 @@ func TestReplaceWhenExpressionsVariables(t *testing.T) {
 	}}
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			got := tc.whenExpressions.ReplaceWhenExpressionsVariables(tc.replacements)
+			got := tc.whenExpressions.ReplaceWhenExpressionsVariables(tc.replacements, nil)
 			if d := cmp.Diff(tc.expected, got); d != "" {
 				t.Errorf("Error evaluating When Expressions in test case %s", diff.PrintWantGot(d))
 			}
@@ -271,10 +271,11 @@ func TestReplaceWhenExpressionsVariables(t *testing.T) {
 
 func TestApplyReplacements(t *testing.T) {
 	tests := []struct {
-		name         string
-		original     *WhenExpression
-		replacements map[string]string
-		expected     *WhenExpression
+		name              string
+		original          *WhenExpression
+		replacements      map[string]string
+		arrayReplacements map[string][]string
+		expected          *WhenExpression
 	}{{
 		name: "replace parameters variables",
 		original: &WhenExpression{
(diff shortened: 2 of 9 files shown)


jerop (commented):
thanks for this fix @pritidesai 😁 

could we please add a task with array params to the example pipeline with when expressions? also would be great to have an test case invalidating specifying array parameters in the input field

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -1379,24 +1395,6 @@ func TestValidatePipelineParameterVariables_Failure(t *testing.T) {
 			Message: `variable type invalid in "$(params.foo)"`,
 			Paths:   []string{"[0].when[0].input"},
 		},
-	}, {
-		name: "invalid string parameter variables in when expression, array reference in values",
```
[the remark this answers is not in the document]
pritidesai: array reference in input is covered by a separate test case above https://github.com/tektoncd/pipeline/pull/4075/files#diff-f66bfbe2808c385605ad5831c843abc71ec6c87b778e7ce72859f0e561bfe726L1365

I will add an additional test with array notation in input.

pkg/apis/pipeline/v1beta1/when_validation_test.go
```diff
@@ -51,6 +51,13 @@ func TestWhenExpressions_Valid(t *testing.T) {
 			Operator: selection.In,
 			Values:   []string{""},
 		}},
+	}, {
+		name: "valid input with string variable and values with array variables",
```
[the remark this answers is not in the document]
pritidesai: the set of tests here running positive tests. We could add it as one of the failure tests but nothing changed in terms of validating the `input`. 

These tests are validating `WhenExpressions.validate()` defined [here](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/when_validation.go#L35). The purpose of `WhenExpressions.validate()` is to validate each field for non-empty values and validating task result pattern.

This test is also not necessary here and can be removed.

pritidesai:
looks like a flake 😞 
Screen Shot 2021-07-07 at 5 00 42 PM

vdemeester (commented):
If we support it for when expression, should we also support this on variable interpolation ? (otherwise might feel a bit inconsistent ).

pritidesai:
> If we support it for when expression, should we also support this on variable interpolation ? (otherwise might feel a bit inconsistent ).

I think we do support this on variable [interpolations](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#specifying-parameters) using the same syntax `$(params.arrayParam[*])`, we are missing support for the array indexing i.e. `$(params.arrayParam[2])`

vdemeester:
> > If we support it for when expression, should we also support this on variable interpolation ? (otherwise might feel a bit inconsistent ).
> 
> I think we do support this on variable [interpolations](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#specifying-parameters) using the same syntax `$(params.arrayParam[*])`, we are missing support for the array indexing i.e. `$(params.arrayParam[2])` in general.

Ah I see 👌🏼

pkg/apis/pipeline/v1beta1/when_validation.go
```diff
@@ -79,12 +78,15 @@ func (wes WhenExpressions) validatePipelineParametersVariables(prefix string, pa
 	for idx, we := range wes {
 		errs = errs.Also(validateStringVariable(we.Input, prefix, paramNames, arrayParamNames).ViaField("input").ViaFieldIndex("when", idx))
 		for _, val := range we.Values {
-			errs = errs.Also(validateStringVariable(val, prefix, paramNames, arrayParamNames).ViaField("values").ViaFieldIndex("when", idx))
+			// one of the values could be a reference to an array param, such as, $(params.foo[*])
+			// extract the variable name from the pattern $(params.foo[*]), if the variable name matches with one of the array params
+			// validate the param as an array variable otherwise, validate it as a string variable
+			if arrayParamNames.Has(strings.TrimSuffix(strings.TrimPrefix(val, "$("+prefix+"."), "[*])")) {
```
jerop, line 84: we're checking if the variable is an array param in two places is slightly different ways, wondering if we can create a common function to check for array param that can be used in both places

https://github.com/tektoncd/pipeline/blob/232ec0e2f9c554f6901951761c73f350207d7e55/pkg/apis/pipeline/v1beta1/when_types.go#L72

https://github.com/tektoncd/pipeline/blob/232ec0e2f9c554f6901951761c73f350207d7e55/pkg/apis/pipeline/v1beta1/when_validation.go#L84

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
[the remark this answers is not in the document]
jerop: thank you!

pkg/apis/pipeline/v1beta1/when_validation.go
pritidesai: Yup, it's very similar but the first expression is returning `params.arrayParam1` from `$(params.arrayParam1[*])` while the second expression is returning `arrayParam1` from `$(params.arrayParam1[*])`.

The `arrayReplacements` in the first expression hold a list of array parameters with `params` prefix while the `arrayReplacements` hold the list of parameters without such prefix.

I have created a common function to return just the parameter name which works for the second expression and added `prefix` back for the first expression.

Validation and applying replacements works with a different patterns because the `apply.go` adds the `prefix` before creating a list of parameter names:

https://github.com/tektoncd/pipeline/blob/2e5780d80a1dd68b0bdd7a494cbb0edf2529a929/pkg/reconciler/pipelinerun/resources/apply.go#L44

We can refactor this further (in a separate PR) to have a consistent validation and replacements routines.

jerop: thanks for the refactor, looks great!

jerop (approved):
thanks @pritidesai!

pkg/apis/pipeline/v1beta1/when_types_test.go
```diff
@@ -307,10 +308,47 @@ func TestApplyReplacements(t *testing.T) {
...
+			Input:    "readme.md",
+			Operator: selection.In,
+			Values:   []string{"dev", "stage"},
+		},
+	}, {
+		name: "replace string and array params",
+		original: &WhenExpression{
+			Input:    "$(params.path)",
+			Operator: selection.In,
+			Values:   []string{"$(params.branches[*])", "$(params.files[*])"},
```
sbwsg, line 334: One additional test-case here for completeness would be `Values` with `string` and `array` vars mixed together:

```go
original: &WhenExpression{
  Input: "$(params.path)",
  Operator: selection.In,
  Values: []string{"$(params.branches[*])", "$(params.matchPath)", "$(params.files[*])"},
},
replacements: map[string]string{
  "params.path": "readme.md",
  "params.matchPath": "foo.txt",
},
arrayReplacements: map[string][]string{
  "params.branches": {"dev", "stage"},
  "params.files": {"readme.md", "test.go"},
},
expected: &WhenExpression{
  Input: "readme.md",
  Operator: selection.In,
  Values: []string{"dev", "stage", "foo.txt", "readme.md", "test.go"},
},
```

sbwsg (commented):
Requesting one additional unit test case but otherwise this lgtm!

pkg/apis/pipeline/v1beta1/when_types_test.go
pritidesai: thanks @sbwsg for an additional unit test, done 👍

PR #4085, opened 2021-07-09
Skipping Strategies

jerop:
# Changes

This change implements skipping strategies to give users the flexibility
to skip a single guarded Task only and unblock execution of its
dependent Tasks.

Today, WhenExpressions are specified within Tasks but they guard the
Task and its dependent Tasks. To provide flexible skipping strategies,
we want to change the scope of WhenExpressions from guarding a Task and
its dependent Tasks to guarding the Task only. If a user wants to guard
a Task and its dependent Tasks, they can:
- cascade the WhenExpressions to the dependent Tasks
- compose the Task and its dependent Tasks as a sub-Pipeline that's
guarded and executed together using Pipelines in Pipelines (but this is
still an experimental feature)

Changing the scope of WhenExpressions to guard the Task only is
backwards-incompatible, so to make the transition smooth:
- we'll provide a feature flag, scope-when-expressions-to-task, which:
  - will default to false to guard a Task and its dependent Tasks
  - can be set to true to guard a Task only
- after migration, we'll change the global default for the feature flag
to true to guard a Task only by default
- eventually, we'll remove the feature flag and guard a Task only going
forward

Implements [TEP-0059: Skipping Strategies](https://github.com/tektoncd/community/blob/main/teps/0059-skipping-strategies.md)
Closes https://github.com/tektoncd/pipeline/issues/2127

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
to guard a `Task` only and unblock execution of its dependent `Tasks`, set the `scope-when-expressions-to-task` feature flag to "true"

action required: `when` expressions guarding a `Task` and its dependent `Tasks` is deprecated and will be removed - migrate by setting `scope-when-expressions-to-task` feature flag to "true" and cascade `when` expressions to subsequent `Tasks` if needed
```


docs/pipelines.md
```diff
@@ -465,6 +465,22 @@ There are a lot of scenarios where `WhenExpressions` can be really useful. Some
 - Checking if the name of a CI job matches
 - Checking if an optional Workspace has been provided
 
+#### Guarding a `Task` and its dependent `Tasks`
+
+When  `WhenExpressions` evaluate to `False`, the `Task` and its branch (of dependent `Tasks`) will be skipped by
+default while the rest of the `Pipeline` will execute. The global default scope of `WhenExpressions` is set to `Branch`;
+`scope-when-expressions-to-task` field in [`config/config-feature-flags.yaml`](./../config/config-feature-flags.yaml)
+defaults to `False`.
```
sbwsg, line 473: Small nit: suggest `"false"` here to match case with what's already in the configmap & to indicate it's a string value rather than a yaml bool.

docs/pipelines.md
```diff
@@ -465,6 +465,22 @@ There are a lot of scenarios where `WhenExpressions` can be really useful. Some
...
+
+**Note:** Scoping `WhenExpressions` to a `Task` and its dependent `Tasks` is deprecated. To guard a `Task` and its
+dependent `Tasks`, cascade `WhenExpressions` to the specific dependent `Tasks` to be guarded as well.
+
+#### Guarding a `Task` only
+
+To guard a `Task` only and unblock execution of its dependent `Tasks`, set the global default scope of `WhenExpressions`
+to `Task` using the `scope-when-expressions-to-task` field in [`config/config-feature-flags.yaml`](./../config/config-feature-flags.yaml)
+by changing it to `True`.
```
sbwsg, line 482: Similar nit here: suggest `"true"` to match case & indicate string value type.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -493,10 +493,11 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
...
 	pipelineRunFacts := &resources.PipelineRunFacts{
-		State:           pipelineRunState,
-		SpecStatus:      pr.Spec.Status,
-		TasksGraph:      d,
-		FinalTasksGraph: dfinally,
+		State:                      pipelineRunState,
+		SpecStatus:                 pr.Spec.Status,
+		TasksGraph:                 d,
+		FinalTasksGraph:            dfinally,
+		ScopeWhenExpressionsToTask: config.FromContextOrDefaults(ctx).FeatureFlags.ScopeWhenExpressionsToTask,
```
sbwsg, line 500: nice 👍 I like this approach of adding the scope state into the facts object rather than passing the ctx all the way down into the dag processing.

sbwsg (approved):
Looks great! Suggest also updating [`docs/install.md`'s feature flags section](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#customizing-the-pipelines-controller-behavior) with info about this new flag.

docs/pipelines.md
bobcatfish: another nit from me: i think it'd be better to link to https://github.com/tektoncd/pipeline/blob/main/docs/install.md#customizing-the-pipelines-controller-behavior i.e. the doc about the config map vs the config map itself

(ps i think that doc might need an update also)

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3487,6 +3487,156 @@ func TestReconcileWithWhenExpressionsWithTaskResults(t *testing.T) {
...
+		tb.PipelineTask("a-task", "a-task"),
+		tb.PipelineTask("b-task", "b-task",
+			tb.PipelineTaskWhenExpression("$(tasks.a-task.results.aResult)", selection.In, []string{"aResultValue"}),
+			tb.PipelineTaskWhenExpression("aResultValue", selection.In, []string{"$(tasks.a-task.results.aResult)"}),
+		),
+		tb.PipelineTask("c-task", "c-task",
+			tb.PipelineTaskWhenExpression("$(tasks.a-task.results.aResult)", selection.In, []string{"missing"}),
+		),
+		tb.PipelineTask("d-task", "d-task", tb.RunAfter("c-task")),
+	))}
```
bobcatfish, line 3502: just a nit - the yaml test libs (https://github.com/tektoncd/pipeline/blob/main/test/yaml.go)  imo make this way more readable! (not to mention easier to write XD)

on the other hand there's definitely something to be said for being consistent with the rest of the file, often more important XD

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3487,6 +3487,156 @@ func TestReconcileWithWhenExpressionsWithTaskResults(t *testing.T) {
...
+			Input:    "aResultValue",
+			Operator: "in",
+			Values:   []string{"missing"},
+		}},
+	}}
+	if d := cmp.Diff(actualSkippedTasks, expectedSkippedTasks); d != "" {
+		t.Errorf("expected to find Skipped Tasks %v. Diff %s", expectedSkippedTasks, diff.PrintWantGot(d))
+	}
+
+	skippedTasks := []string{"c-task", "d-task"}
```
bobcatfish, line 3624: im a bit confused why we expect d-task to be skipped here, but we don't expect it to be in the list of expected skipped tasks?

docs/pipelines.md
```diff
@@ -465,6 +465,22 @@ There are a lot of scenarios where `WhenExpressions` can be really useful. Some
 - Checking if the name of a CI job matches
 - Checking if an optional Workspace has been provided
 
+#### Guarding a `Task` and its dependent `Tasks`
+
+When  `WhenExpressions` evaluate to `False`, the `Task` and its branch (of dependent `Tasks`) will be skipped by
```
bobcatfish, line 470: i think it would be good to define (or link to a definition elsewhere) of what "dependent tasks" means

There are also some docs on "conditions" that look like they might be trying to explain this - might be worth borrowing and/or updating those as well: https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#guard-task-execution-using-conditions

docs/pipelines.md
```diff
@@ -465,6 +465,22 @@ There are a lot of scenarios where `WhenExpressions` can be really useful. Some
...
+default while the rest of the `Pipeline` will execute. The global default scope of `WhenExpressions` is set to `Branch`;
```
bobcatfish, line 471: i think that this deserves a bit more explanation as well - i.e. some explanation of "branch scope"

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3487,6 +3487,156 @@ func TestReconcileWithWhenExpressionsWithTaskResults(t *testing.T) {
...
+	ps := []*v1beta1.Pipeline{tb.Pipeline("test-pipeline", tb.PipelineNamespace("foo"), tb.PipelineSpec(
+		tb.PipelineTask("a-task", "a-task"),
+		tb.PipelineTask("b-task", "b-task",
+			tb.PipelineTaskWhenExpression("$(tasks.a-task.results.aResult)", selection.In, []string{"aResultValue"}),
+			tb.PipelineTaskWhenExpression("aResultValue", selection.In, []string{"$(tasks.a-task.results.aResult)"}),
+		),
+		tb.PipelineTask("c-task", "c-task",
+			tb.PipelineTaskWhenExpression("$(tasks.a-task.results.aResult)", selection.In, []string{"missing"}),
+		),
+		tb.PipelineTask("d-task", "d-task", tb.RunAfter("c-task")),
```
bobcatfish, line 3501: i'm a bit confused by the setup of this test (maybe some comments might help?)

it seems like we're making a graph of:

a-task runs on its own
b-task depends on a-task, is guarded with a when expression based on a-task which will always be true
c-task depends on a-task, is guarded with a when expression based on a-task which will always be falst
d-task runs after c-task

we're setting this up so that a-task has already run, therefore b-task and c-task will get evaluated next - b-task should run, and c-task should be skipped

Since this test turns on `scope-when-expressions-to-task`, i would expect that skipping c-task would have no impact d-task, i.e. it would start running at the same time as b-task

(with `scope-when-expressions-to-task` off, the default, id expect d-task to get skipped also) 

so after that wall of text haha, im wondering:
1. am i understanding the setup correctly? (and maybe can we add some comments to the test to make it more clear)
2. why isn't d-task also running after the reconcile?

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -56,7 +56,8 @@ type PipelineRunFacts struct {
 	// needed, via the `Skip` method in pipelinerunresolution.go
 	// The skip data is sensitive to changes in the state. The ResetSkippedCache method
 	// can be used to clean the cache and force re-computation when needed.
-	SkipCache map[string]bool
+	SkipCache                  map[string]bool
+	ScopeWhenExpressionsToTask bool
```
bobcatfish, line 60: the docstring for PipelineRunFacts could use an update i think as well since it seems to be trying to list everything in this object

might be better to replace it with a description of what this object seems to be for instead of a list of its contents, e.g. something like "PipelineRunFacts holds the state of all of the components that make up the Pipeline graph and can be used to <something something christie got lazy 😅 >"

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -227,7 +227,11 @@ func (t *ResolvedPipelineRunTask) parentTasksSkip(facts *PipelineRunFacts) bool
 	stateMap := facts.State.ToMap()
 	node := facts.TasksGraph.Nodes[t.PipelineTask.Name]
 	for _, p := range node.Prev {
-		if stateMap[p.Task.HashKey()].Skip(facts) {
+		parentTask := stateMap[p.Task.HashKey()]
+		if parentTask.Skip(facts) {
+			if facts.ScopeWhenExpressionsToTask && parentTask.PipelineTask.WhenExpressions != nil {
```
bobcatfish, line 232: are there any cases we should think about where, parent task is skipped, it HAS a when expression, but the when expression isnt the reason it was skipped?

(im wondering if we should be checking more than just `parentTask.PipelineTask.WhenExpressions != nil`, i.e. checking that they are present AND that they evaluated to skip)

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -227,7 +227,11 @@ func (t *ResolvedPipelineRunTask) parentTasksSkip(facts *PipelineRunFacts) bool
...
+				continue
+			}
```
bobcatfish, line 234: im confused what the purpose of this task is based on the name `parentTasksSkip` - what is this function's job? if there was any way you'd be willing to take on a bit of scope creep and at least add a docstring if not also rename this function id be very appreciative 🙏 

(i cant tell if this functions job is to say: should this task be skipped since it's parent task was skipped, or maybe: was the parent task skipped, or something else)

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -227,7 +227,11 @@ func (t *ResolvedPipelineRunTask) parentTasksSkip(facts *PipelineRunFacts) bool
 	stateMap := facts.State.ToMap()
 	node := facts.TasksGraph.Nodes[t.PipelineTask.Name]
 	for _, p := range node.Prev {
-		if stateMap[p.Task.HashKey()].Skip(facts) {
+		parentTask := stateMap[p.Task.HashKey()]
+		if parentTask.Skip(facts) {
+			if facts.ScopeWhenExpressionsToTask && parentTask.PipelineTask.WhenExpressions != nil {
+				continue
```
bobcatfish, line 233: this feels like the opposite of what i would expect - but i might be misunderstanding the purpose of this function! id think that if ScopeWhenExpressionsToTask is true, then it doesnt matter that the parent was skipped?

that for the parent being skipped to matter, ScopeWhenExpressionsToTask would need to be false - or there would need to be a resource dependency?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
bobcatfish: i think it would also help to have some unit tests for just this function - often a sign that this unexported function deserves to be in its own package somewhere and exported :D

when i look at pipelinerunresolution_test, it seems like the behavior is what i would expect, but when i look at this logic in this function i dont understand how that's being accomplished - some unit test directed at just this function might help

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -983,6 +984,196 @@ func TestIsSkipped(t *testing.T) {
...
+		scopeWhenExpressionsToTask: false,
+		expected: map[string]bool{
+			"mytask11": true,
+			"mytask18": true,
+			"mytask19": true,
+		},
+	}, {
+		name: "tasks-when-expressions-scoped-to-task-run-multiple-dependent-tasks",
+		state: PipelineRunState{{
+			// skipped because when expressions evaluate to false
```
bobcatfish, line 1087: thanks for adding these comments, really helps me understand the tests! :D

bobcatfish (commented):
so exciting to see this change happening!!! hope you don't mind a barrage of feedback from me - mostly about docs + some code/test clarity

also i didnt see much in the docs or tests handling the behavioral differences around resource vs. ordering dependencies - my understanding is that even if the when expression is scoped to just the task, that if other tasks depend on results (and eventually workspaces) from that task, they will still be skipped (because we have no way to provide a valid result to them). is that right? (and either way i think the docs + tests maybe should represent these cases also)

p.s. great commit message!! :D

> after migration, we'll change the global default for the feature flag to true to guard a Task only by default

i'm wondering how to minimize the problems this could cause 🤔  hopefully no one is relying on it and if they are the see our warnings 😟  

maybe we should send some announcements to tekton-dev and tekton-users about this

we could even have something like the CLI notice when ppl are using when expressions to guard branches and warn them - probably too far but we could even have the CLI 'phone home' and tell us when folks are doing that; phoning home should probably be discussed on its own tho, maybe its enough to loudly warn ppl

at the very least i think we should make the release note that accompanies this (and subsequent ones) stand out a bit more - we need to keep announcing this as loudly as we can

pkg/reconciler/pipelinerun/pipelinerun_test.go
jerop: > i'm a bit confused by the setup of this test (maybe some comments might help?)
> 
> it seems like we're making a graph of:
> 
> a-task runs on its own
> b-task depends on a-task, is guarded with a when expression based on a-task which will always be true
> c-task depends on a-task, is guarded with a when expression based on a-task which will always be false
> d-task runs after c-task
> 

yes, that's the correct set up of the test

> we're setting this up so that a-task has already run, therefore b-task and c-task will get evaluated next - b-task should run, and c-task should be skipped
> 

that's right, then after b-task and c-task are done, d-task will be evaluated in the next reconcile loop

> Since this test turns on `scope-when-expressions-to-task`, i would expect that skipping c-task would have no impact d-task, i.e. it would start running at the same time as b-task
> 

d-task can't run at the same time as b-task -- it has to run in the next loop (after b-task and c-task are done) because of how we traverse the pipeline graph

regardless of the when expressions being scoped to task, the execution order of tasks is still enforced and we won't look at d-task until the next time getNextTasks is called in the next reconcile loop 
https://github.com/tektoncd/pipeline/blob/c8dc797cf5a6f11f90cb742d014470a444fcdc60/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go#L219-L240

> (with `scope-when-expressions-to-task` off, the default, id expect d-task to get skipped also)
> 

yes, that's right

> so after that wall of text haha, im wondering:
> 
> 1. am i understanding the setup correctly? (and maybe can we add some comments to the test to make it more clear)

good idea, added some comments to the test and also simplified it to need only one reconcile loop to make it clearer

> 2. why isn't d-task also running after the reconcile?

in this reconcile, only b-task and c-task will run -- d-task is supposed to run in the next

pkg/reconciler/pipelinerun/pipelinerun_test.go
jerop: rewrote the test, it may be readable enough now -- but happy to look into the test yamls later if it's still confusing

pkg/reconciler/pipelinerun/pipelinerun_test.go
jerop: that was a typo, should not have included d-task there

but still, we're using this list to check that there are no taskruns for these pipeline tasks -- which we expect there to be no d-task because it's not in the current tasks for this reconcile loop (should be in the next)

improved the test such that it shows how scoping impacts ordering and resource dependent tasks

pkg/reconciler/pipelinerun/pipelinerun_test.go
jerop: updated the test such that we have:
- a-task runs on its own, is guarded with when expressions that evaluate to false so is skipped
- b-task depends on a-task based on runafter, is executed because it runs after a-task and when expressions are scoped to task
- c-task depends on a-task based on result, is attempted because when expressions are scoped to task but then get skipped because of missing result references from a-task

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
jerop: glad to hear it 😁

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
jerop: updated the docstring to explain its purpose

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
jerop: > when i look at pipelinerunresolution_test, it seems like the behavior is what i would expect, but when i look at this logic in this function i dont understand how that's being accomplished - some unit test directed at just this function might help

sounds good, added unit tests for that function (checking whether to skip because of parent tasks)

jerop:
> Looks great! Suggest also updating [`docs/install.md`'s feature flags section](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#customizing-the-pipelines-controller-behavior) with info about this new flag.

updated that doc, thank you @sbwsg!

jerop:
thanks for the review @bobcatfish!

> also i didnt see much in the docs or tests handling the behavioral differences around resource vs. ordering dependencies - my understanding is that even if the when expression is scoped to just the task, that if other tasks depend on results (and eventually workspaces) from that task, they will still be skipped (because we have no way to provide a valid result to them). is that right? (and either way i think the docs + tests maybe should represent these cases also)
> 

yes, that's correct -- added more details to the docs and comments to the tests to describe how they handle those cases

> p.s. great commit message!! :D

thanks 😁 

> 
> > after migration, we'll change the global default for the feature flag to true to guard a Task only by default
> 
> i'm wondering how to minimize the problems this could cause 🤔 hopefully no one is relying on it and if they are the see our warnings 😟
> 
> maybe we should send some announcements to tekton-dev and tekton-users about this
> 

yes, we included in the [upgrade & migration strategy](https://github.com/tektoncd/community/blob/main/teps/0059-skipping-strategies.md#upgrade--migration-strategy) that we'll "over-communicate during the migration in Slack, email and working group meetings" -- will send the first announcements when this is released 

> we could even have something like the CLI notice when ppl are using when expressions to guard branches and warn them - probably too far but we could even have the CLI 'phone home' and tell us when folks are doing that; phoning home should probably be discussed on its own tho, maybe its enough to loudly warn ppl
> 

do we do CLI warnings for anything else? wondering if it'll be easy to add that warning to tkn 

> at the very least i think we should make the release note that accompanies this (and subsequent ones) stand out a bit more - we need to keep announcing this as loudly as we can

yes yes, added more details and 'action-required' to the release notes

docs/deprecations.md
```diff
@@ -25,3 +25,4 @@ being deprecated.
...
 | [The `disable-home-env-overwrite` flag will be removed](https://github.com/tektoncd/pipeline/issues/2013)                                     | [v0.24.0](https://github.com/tektoncd/pipeline/releases/tag/v0.24.0)         | Beta                                                                                                     | February 10 2022                    |
 | [The `disable-working-dir-overwrite` flag will be removed](https://github.com/tektoncd/pipeline/issues/1836)                                  | [v0.24.0](https://github.com/tektoncd/pipeline/releases/tag/v0.24.0)         | Beta                                                                                                     | February 10 2022                    |
+| [The `scope-when-expressions-to-task` flag will be removed](https://github.com/tektoncd/pipeline/issues/1836)                                 | [v0.26.0](https://github.com/tektoncd/pipeline/releases/tag/v0.26.0)         | Beta                                                                                                     | February 10 2022                    |
```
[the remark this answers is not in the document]
jerop: updated the migration plan in the tep as well - https://github.com/tektoncd/community/pull/481

pkg/reconciler/pipelinerun/pipelinerun_test.go
bobcatfish: > d-task can't run at the same time as b-task -- it has to run in the next loop (after b-task and c-task are done) because of how we traverse the pipeline graph

that doesnt sound right to me - is this the way we evaluate other concurrent tasks? e.g. if a pipeline has 2 tasks, `a` and `b`, and they have no dependencies or `runAfter`, they both run simultaneously - do they not get both created in the next reconcile loop, or do we do them one at a time? 

this line makes me think we start all the tasks that are available to run:

https://github.com/tektoncd/pipeline/blob/8907e408c73480482c618d08eeaaefe0bd4aac68/pkg/reconciler/pipelinerun/pipelinerun.go#L637

i dont think that when expressions should make that behavior any different, and i would count that as a bug in when expressions if they do (how can we even guarantee in that case that the pipelinrun would immediately be reconciled again?)

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: sorry for coming late to the discussion here, I agree with @bobcatfish, if the flag is set to `true`, no need to check if a parent was skipped or not. 

I do not understand this statement:

> if no, skip the current task because this parent task was skipped

A task can be skipped because of the `when` expressions or a task result resolution fails. From the `when` expression scoping perspective, irrespective of one parent skipped vs a couple of parents skipped, the `when` expressions scoped to that parent/s.

pkg/reconciler/pipelinerun/pipelinerun_test.go
bobcatfish: @jerop and i spent some time talking this through and i take back what i said about the behavior being a bug XD sorry for jumping to that conclusion @jerop ! 🙏 

Sounds like there are 2 different and valid approaches we could use with different trade offs. Imagine a Pipeline with 100 tasks: they run sequentially (using runAfter) but the first 99 have when expressions that evaluate to false and they should be skipped:

```
    1 (when false)
    |
    V
    2 (when false)
    |
    V
    3 (when false)
    |
    V
    4...99 (when false)
    |
    V
    100
```

The question is: what happens in the first reconcile loop?

1. I assumed that the entire graph would be evaluated and at the end of the first reconcile loop, 100 would be run
2. The implementation here would look at the next task to run, which on the first reconcile would be 1, evaluate it, determine it should be skipped, and complete. The next reconcile would look at 2, and so on.

(1) assumes the job of the reconcile loop is to determine which task should _run_ next
(2) assumes the job of the reconcile loop is to determine which task should be _evaluated_ next

I think (2) is simpler to implement (esp given our existing code) b/c going with (1) would mean we have to do some kind of "evaluate until done" looping within the reconcile loop that we've never had to do before.

But I think (2) is less efficient b/c running the above pipeline would take 100 separate reconciles (including calls to the API server, to etcd, and the being notified by the API server that we need to evaluate again). (BUT (1) might potentially block reconciling of other pipelines by doing a loop over 99 when expressions in one reconcile loop?)

@jerop  and I think it makes sense to go with (2) for now and address possible efficiency issues later if needed (🤞  hopefully no one is running a Pipeline like the above XD)

requests from me:
- Verify that the reconcile loop will be triggered again after a Task is evaluated as skipped (i.e. that the change we are making in the reconcile loop will trigger another reconcile) - if not, we'll need to use the approach where we traverse all the skipped tasks in one loop
- Add something to our [developer docs](https://github.com/tektoncd/pipeline/tree/main/docs/developers) explaining the tradeoff made here (e.g. the simplicity in using multiple reconciles vs getting everything done in one reconcile)
- I liked the test case with d-task and think it's worth adding back XD it stimulated a great conversation

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
bobcatfish: thanks @jerop, makes a lot more sense to me now!! 🙏

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
bobcatfish: When i look at the function now, it makes sense to me - not sure if I was seeing something else before that was fixed or if changing the name of the function just cleared it all up for me but all my concerns are addressed!

docs/pipelines.md
```diff
@@ -465,6 +465,141 @@ There are a lot of scenarios where `WhenExpressions` can be really useful. Some
...
+#### Guarding a `Task` and its dependent `Tasks`
+
+When  `when` expressions evaluate to `False`, the `Task` and its dependent `Tasks` will be skipped by default while the 
+rest of the `Pipeline` will execute. Dependencies between `Tasks` can be either ordering ([`runAfter`](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-the-runafter-parameter))
+or resource (e.g. [`Results`](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-results)) 
+dependencies, as further described in [configuring execution order](#configuring-the-task-execution-order). The global 
+default scope of `when` expressions is set to a `Task` and its dependent`Tasks`; `scope-when-expressions-to-task` field 
+in [`config/config-feature-flags.yaml`](install.md#customizing-the-pipelines-controller-behavior) defaults to "false".
+
+**Note:** Scoping `when` expressions to a `Task` and its dependent `Tasks` is deprecated. To guard a `Task` and its
```
bobcatfish, line 477: i think there's something I'm still missing here, and apologies if I'm just being picky about the wording, but my understanding is:

* Scoping when expressions to a task and the tasks are explicitly ordered to run after it (via runAfter) is deprecated
* Even with the new behavior, if a task is skipped and a subsequent task has a resource dependency on the skipped task, the subsequent task _will still be skipped_, that is: the when expression is still scoped to the task _and its (resource dependent) dependents_

At the moment I think this is the only option until we have [TEP-0048](https://github.com/tektoncd/community/pull/240) - and even then users would need to explicitly configure defaults to use that functionality

So in summary I'm not sure if it's just that I'm expecting to see this called out more explicitly in the docs (i.e. just a doc update - or point me at something in the docs im missing), or if this has been implemented with different assumptions (and there's a difference in functionality b/w this and how i thought it would work)?

docs/pipelines.md
```diff
@@ -465,6 +465,141 @@ There are a lot of scenarios where `WhenExpressions` can be really useful. Some
...
+      operator: in
+      values:
+        - merge
+  taskRef:
+    name: manual-approval
+
+- name: slack-msg
+  params:
+    - name: approver
+      value: $(tasks.manual-approval.results.approver)
```
bobcatfish, line 583: hmmm similar to my comment above, I'm not sure how this would be possible? even tho skipping is scoped to only `manual-approval`, my understanding is that this should still get skipped b/c otherwise what value will ` $(tasks.manual-approval.results.approver)` have?

if this IS how this is working right now, I'm guessing this is an empty string? [This design doc has some info about why this might not be what we want](https://docs.google.com/document/d/1tV1LgPOINnmlDV-oSNdLB39IlLcQRGaYAxYZjVwVWcs/edit), particularly [Option b (results default)](https://docs.google.com/document/d/1tV1LgPOINnmlDV-oSNdLB39IlLcQRGaYAxYZjVwVWcs/edit#heading=h.phrhie8akizi), which also links to [a previous discussion about this](https://docs.google.com/document/d/10iEJqVstY6k3KNvAXgffIJLcHRbPQ-GIAfQk5Dlrf3c/edit#heading=h.19ybq5czkwi6)

The summary section of https://github.com/tektoncd/community/pull/240 uses this same Approver example, pointing out that maybe using a string like "None" might be the desired behavior. For example maybe the slack message is something like "Pull request 4555 was approved by <approver>" - Something like "Pull request 4555 was approved by None" or maybe ("Pull request 4555 was approved implicitly; no manual approval needed") would be better than "Pull request 4555 was approved by "

bobcatfish (commented):
Thanks for all the back and forth on this @jerop ! The last thing for me here is that I think resource dependencies aren't quite being handled the way the TEP describes (just let me know if im misunderstanding)

[TEP-0059 use cases](https://github.com/tektoncd/community/blob/main/teps/0059-skipping-strategies.md#use-cases):

> slack-msg should be skipped because of the missing Result reference to the approver name

ps apologies for dropping this comment on you and then disappearing for 2 weeks 😅  🙏

docs/pipelines.md
pritidesai: > hmmm similar to my comment above, I'm not sure how this would be possible? even tho skipping is scoped to only `manual-approval`, my understanding is that this should still get skipped b/c otherwise what value will ` $(tasks.manual-approval.results.approver)` have?
> 

$(tasks.manual-approval.results.approver) has NO value i.e. no empty string if manual-approval is skipped. The task result resolution indicates the value is missing and 
slack-msg has no sufficient data to run and will be skipped. But at least the slack-msg was scheduled and attempted without creating any taskRun.  I think this aligns with the proposal.

> if this IS how this is working right now, I'm guessing this is an empty string? [This design doc has some info about why this might not be what we want](https://docs.google.com/document/d/1tV1LgPOINnmlDV-oSNdLB39IlLcQRGaYAxYZjVwVWcs/edit), particularly [Option b (results default)](https://docs.google.com/document/d/1tV1LgPOINnmlDV-oSNdLB39IlLcQRGaYAxYZjVwVWcs/edit#heading=h.phrhie8akizi), which also links to [a previous discussion about this](https://docs.google.com/document/d/10iEJqVstY6k3KNvAXgffIJLcHRbPQ-GIAfQk5Dlrf3c/edit#heading=h.19ybq5czkwi6)
> 
> The summary section of [tektoncd/community#240](https://github.com/tektoncd/community/pull/240) uses this same Approver example, pointing out that maybe using a string like "None" might be the desired behavior. For example maybe the slack message is something like "Pull request 4555 was approved by " - Something like "Pull request 4555 was approved by None" or maybe ("Pull request 4555 was approved implicitly; no manual approval needed") would be better than "Pull request 4555 was approved by "

Once we add support for the task results without results, the message will be sent saying PR was approved implicitly or by no one.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -39,6 +39,24 @@ const (
...
+const (
+	WhenExpressionsSkip         SkippingReason = "WhenExpressionsSkip"
+	ConditionsSkip              SkippingReason = "ConditionsSkip"
+	ParentTasksSkip             SkippingReason = "ParentTasksSkip"
+	IsStoppingSkip              SkippingReason = "IsStoppingSkip"
+	IsGracefullyCancelledSkip   SkippingReason = "IsGracefullyCancelledSkip"
+	IsGracefullyStoppedSkip     SkippingReason = "IsGracefullyStoppedSkip"
+	MissingResultsReferenceSkip SkippingReason = "MissingResultsReferenceSkip"
+	None                        SkippingReason = "None"
+)
```
pritidesai, line 53: I am trying to understand the list of reasons here and how they are utilized. 

These are great and I really appreciate the thorough explanation to the users why a task was skipped but at the same time, I think this might cause false positives or incorrect responses if communicated to the user.

How is reason ParentTasksSkip applicable with scoping the when expression to the task? No need to introduce a new reason for the current functionality which will be deprecated in 9 months.

For a task with multiple parents, there could be multiple reasons for skipping that task. Can we keep it simple for the first iteration and introduce these if they are visible in the status?

pkg/reconciler/pipelinerun/pipelinerun_test.go
jerop: @bobcatfish you were actually right - turns out that what we actually do is (1) where the reconcile loop determines which task would run next

followed up with @pritidesai about the `successfulOrSkippedDAGTasks` function which enables (1) to happen: https://github.com/tektoncd/pipeline/blob/a728ce3c3d717fe52ff0b1eb71d357460498115c/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go#L480-L490

so what happened in the initial test case? 
- `a-task` runs on its own
- `b-task` depends on `a-task` (based on `runAfter`), is guarded with a `when` expression which will always be true
    - is executed successfully
- `c-task` depends on `a-task` (based on a `Result`), is guarded with a `when` expression with a missing `Result`
    - is skipped because of a missing `Result`
    - test case was misconstructed because `a-task` ran successfully but somehow didn't produce that `Result`
- `d-task` runs after `c-task`
    - is skipped because its parent task `c-task` was skipped because of a missing `Result`

I'd thought what we did was (2) where the reconcile loop determines which task would be evaluated next, and that that was the reason `d-task` was not executed but it was actually because of the above reason

I've added expanded the test case to cover the multiple scenarios with both ordering and resource dependent tasks

docs/pipelines.md
jerop: yes, that is correct that the ordering-dependent tasks should execute successfully and the resource dependencies will need default results -- added more documentation to explain this in further detail

docs/pipelines.md
jerop: yes, and by skipping it right now because of missing results (instead of failing) sets it up for continued execution once default results are available and the user enables them

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
jerop: had a conversation with @pritidesai about this, but following up here in case someone has the same questions

while there are multiple reasons for skipping, the precedence of the reasons is important and factored in here

https://github.com/tektoncd/pipeline/blob/1174247beb3c5146787899b8a5e2214c1d4867a1/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L193-L221

from pipeline state --> parent --> when expressions --> results references --> conditions (deprecated) 

either way, we need to know the exact reason a task was skipped so that we can handle the exception when when expressions are scoped to task in (specifically L#279):

https://github.com/tektoncd/pipeline/blob/1174247beb3c5146787899b8a5e2214c1d4867a1/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L266-L286

sbwsg:
@bobcatfish @pritidesai do you think you'll have time to re-review prior to 0.27?  I'm planning to start the release process Monday next week but happy to delay a day or two if you'll have time early next week instead.

Also ping @tektoncd/core-maintainers in case anyone else has time to review as well.

pritidesai:
> @bobcatfish @pritidesai do you think you'll have time to re-review prior to 0.27? I'm planning to start the release process Monday next week but happy to delay a day or two if you'll have time early next week instead.

Hey @sbwsg yes please, I am starting to review the changes this week and might go into early next week. 🙏

sbwsg:
No worries, I'll aim to start the release process tues or wed next week.

docs/deprecations.md
```diff
@@ -25,3 +25,5 @@ being deprecated.
...
 | [The `disable-home-env-overwrite` flag will be removed](https://github.com/tektoncd/pipeline/issues/2013)                                     | [v0.24.0](https://github.com/tektoncd/pipeline/releases/tag/v0.24.0)         | Beta                                                                                                     | February 10 2022                    |
 | [The `disable-working-dir-overwrite` flag will be removed](https://github.com/tektoncd/pipeline/issues/1836)                                  | [v0.24.0](https://github.com/tektoncd/pipeline/releases/tag/v0.24.0)         | Beta                                                                                                     | February 10 2022                    |
+| [The `scope-when-expressions-to-task` flag will be flipped from `false` to `true`](https://github.com/tektoncd/pipeline/issues/1836)          | [v0.26.0](https://github.com/tektoncd/pipeline/releases/tag/v0.26.0)         | Beta                                                                                                     | February 10 2022                    |
```
pritidesai, line 28: we might have to change this to 0.27, do we? 🤔

docs/install.md
```diff
@@ -366,6 +366,10 @@ use of custom tasks in pipelines.
 most stable features to be used. Set it to "alpha" to allow alpha
 features to be used.
 
+- `scope-when-expressions-to-task`: set this flag to "true" to scope `when` expressions to guard a `Task` only. Set it
+  to "false" to guard a `Task` and its dependent `Tasks`. It defaults to to "false". For more information, see [guarding
```
pritidesai, line 370: nit: extra to in - It defaults to to "false".

docs/deprecations.md
jerop: that's right, fixed it

docs/install.md
jerop: fixed it, thanks :)

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3488,6 +3488,167 @@ func TestReconcileWithWhenExpressionsWithTaskResults(t *testing.T) {
 	}
 }
 
+func TestReconcileWithWhenExpressionsScopedToTask(t *testing.T) {
```
pritidesai, line 3491: nit: adding a graph in the comment above would be helpful to understand

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3488,6 +3488,167 @@ func TestReconcileWithWhenExpressionsWithTaskResults(t *testing.T) {
...
+	// initialize the pipelinerun with the skipped a-task
+	prs[0].Status.SkippedTasks = append(prs[0].Status.SkippedTasks, v1beta1.SkippedTask{
+		Name: "a-task",
+		WhenExpressions: v1beta1.WhenExpressions{{
+			Input:    "foo",
+			Operator: selection.In,
+			Values:   []string{"bar"},
+		}},
+	})
+	// initialize a-task, b-task and c-task
```
pritidesai, line 3531: nit: comment does not match here, missing `d-task` and `e-task`

pkg/reconciler/pipelinerun/pipelinerun_test.go
jerop: added a graph, so hard to draw them though, hope it helps!

pritidesai:
Hey @jerop I ran a simple [pipeline](https://gist.github.com/pritidesai/602e49813bb9220c8c4d07bf15c92fbe) with these changes but the pipelineRun is stuck after executing the first task. The `pipelineRun` has declared the second task skipped but not able to schedule the third task.

```
 kubectl get pr when-expression-lvmm9 -o json | jq .status.conditions
[
  {
    "lastTransitionTime": "2021-08-06T20:56:53Z",
    "message": "Tasks Completed: 1 (Failed: 0, Cancelled 0), Incomplete: 2, Skipped: 1",
    "reason": "Running",
    "status": "Unknown",
    "type": "Succeeded"
  }
]
```

jerop:
> Hey @jerop I ran a simple [pipeline](https://gist.github.com/pritidesai/602e49813bb9220c8c4d07bf15c92fbe) with these changes but the pipelineRun is stuck after executing the first task. The `pipelineRun` has declared the second task skipped but not able to schedule the third task.
> 
> ```
>  kubectl get pr when-expression-lvmm9 -o json | jq .status.conditions
> [
>   {
>     "lastTransitionTime": "2021-08-06T20:56:53Z",
>     "message": "Tasks Completed: 1 (Failed: 0, Cancelled 0), Incomplete: 2, Skipped: 1",
>     "reason": "Running",
>     "status": "Unknown",
>     "type": "Succeeded"
>   }
> ]
> ```

the problem here was that we needed to resolve task results and evaluate whether to skip in order to continue down the graph to get the next tasks to skip - the same way that we do when there's no result references

fixed it and the example below executes as expected:
Screen Shot 2021-08-10 at 5 54 45 PM

also represented this example in an end to end test here: https://github.com/tektoncd/pipeline/blob/a3d83403c15b31d33ec0fb5be77aa23602b7d6fa/pkg/reconciler/pipelinerun/pipelinerun_test.go#L3973-L4091

thank you @pritidesai 😸

pritidesai:
thanks a bunch @jerop for spending so much effort on this 🙏 
/lgtm

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -210,49 +248,83 @@ func (t *ResolvedPipelineRunTask) conditionsSkip() bool {
...
 			return true
 		}
 	}
 	return false
 }
 
+// skipBecauseResultReferencesAreMissing checks if the task references results that cannot be resolved, which is a
+// reason for skipping the task, and applies result references if found
+func (t *ResolvedPipelineRunTask) skipBecauseResultReferencesAreMissing(facts *PipelineRunFacts) bool {
+	if t.checkParentsDone(facts) && t.hasResultReferences() {
```
pritidesai, line 287: hey @jerop this check needs to be revisited. Its causing a task without any when expressions or its parent without having any when expression to skip in case a result reference is missing. The alpha test failure is legit causing a task to be skipped:

```
               "skippedTasks": [
                    {
                        "name": "task2"
                    }
                ],
```

Task2 should fail here since it has a result reference but the results were not initialized by the produce task.

pritidesai: The [alpha](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4085/pull-tekton-pipeline-alpha-integration-tests/1425292326049157120/) failure: 

```
Error waiting for PipelineRun to fail: "pipelinerun-with-failing-step" completed successfully, should have been failed with reason "InvalidTaskResultReference"
```

I am putting this PR on hold until we fix this check, hopefully for not too long 🤞 

/hold

jerop: thanks @pritidesai! added a check that it applies to finally tasks and if parent was skipped due to when expression only -- so won't apply to the onError alpha feature

pritidesai: Awesome thanks a bunch @jerop 🙏👍

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -210,26 +248,50 @@ func (t *ResolvedPipelineRunTask) conditionsSkip() bool {
...
+	return false
+}
+
+// skipBecauseResultReferencesAreMissing checks if the task references results that cannot be resolved, which is a
+// reason for skipping the task, and applies result references if found
+func (t *ResolvedPipelineRunTask) skipBecauseResultReferencesAreMissing(facts *PipelineRunFacts) bool {
+	if t.checkParentsDone(facts) && t.hasResultReferences() {
+		resolvedResultRefs, pt, err := ResolveResultRefs(facts.State, PipelineRunState{t})
+		rprt := facts.State.ToMap()[pt]
+		if err != nil && (t.IsFinalTask(facts) || rprt.Skip(facts).SkippingReason == WhenExpressionsSkip) {
```
pritidesai, line 290: Hey @jerop i don’t think we need to limit this check to finally 🤔, in case of the DAG tasks, a task is skipped without any when expression or without having when expression in any of the parents.

the other alpha test is failing with:

‘’’
examples_test.go:65: Failed waiting for task run done: timed out waiting for the condition
‘’’

or is it a flake 🤔

jerop: that was a flake, i tested it locally and it passed

jerop: it just passed here too: https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4085/pull-tekton-pipeline-alpha-integration-tests/1425449864539410432/

jerop: it's not limited to finally only, we check whether:
- current task is a finally task has missing result references
- current task has missing result references from a parent task that was skipped due to its whenexpressions

pritidesai: Awesome 👏 thanks

pritidesai: /hold cancel

Issue #4086, opened 2021-07-12
Rename the 'finally' key to 'final' in the specification and while still allowing 'finally' to help the users who already have this key as part of their specification


Srivaralakshmi:
**Expected behaviour:** 

- The name used to represent the final set of tasks at the end of the pipeline must be represented as `final` in the specification: https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#adding-finally-to-the-pipeline
- while the key is renamed from `finally` to `final`, continue to allow the `finally` key to accommodate the users who already have the `finally` key as part of their specification.  

**Actual behaviour:** At present the name used to represent the final set of tasks at the end of the pipeline is represented as `finally` in the specification: https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#adding-finally-to-the-pipeline

**Reason for this proposal:**

- The word 'final' is an adjective that is describing the noun 'task', just like the term 'conditional' task (which is deprecated now and is replaced with 'when' expression).
- 'finally' is an adverb and when we use it to describe 'task', it sounds incorrect and weird. 
- The docs mention both terms and switches between: 'final' task and 'finally' task, in multiple places (please see: https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#adding-finally-to-the-pipeline). This is very confusing even for a new developer. 
- We can have `final` tasks and not `finally` tasks. We need to use one terminology consistently. Also the terminology must be correct both technically and from the language perspective as well.

Srivaralakshmi:
@nikhil-thomas PTAL at this ^ issue, comment on the issue, and kindly tag others. This will help drive some discussion. Thank you!

bobcatfish:
hey @Srivaralakshmi - thanks for opening this issue.

The motivation for the term "finally" was inspiration from languages like python which have ["try/catch/finally" syntax](https://docs.python.org/3/reference/compound_stmts.html#finally). [This design doc has some more context around the design](https://docs.google.com/document/d/1lxpYQHppiWOxsn4arqbwAFDo4T0-LCqpNa6p-TJdHrw/edit#). What our current syntax is trying to say is:

1. First, run all these tasks (`tasks` section - which actually is a DAG of tasks, you could argue we could have a better name for this as well)
2. Finally, after that's done, whether it succeed or not, run these other tasks (`finally` section)

It sounds like one thing we could certainly do in the short term is to consistently use "finally tasks" in our docs instead of sometimes saying "final tasks".

[...]

PR #4090, opened 2021-07-12
Always refer to "finally" tasks 📖

bobcatfish:
# Changes

This commit changes all references to "final tasks" to "finally tasks".
We were using a mix of "final tasks" and "finally tasks" to refer to
these tasks, so this commit is making us consistently use one version.
However "finally tasks" is less grammatically correct that "final tasks"
because finally is an adverb - "finally tasks" only makes sense if
"finally tasks" is considered short for "finally tasks in a pipeline".
Another option sould be to go the other way and always use "final tasks"
(and never say "finally tasks"). However if we do this, @Srivaralakshmi
pointed out in https://github.com/tektoncd/pipeline/issues/4086 that
this may make it confusing in that reading something like 'the final tasks'
may make a user expect to see a seciton in the pipeline called `final`
instead of the actual section `finally`.

We have another option which is to consider `finally`- an adverb - a
bad name for that section of the Pipeline spec and change it to `final`.
My understanding of the syntax is a bit different, e.g.:
```yaml
spec:
  tasks:
    - name: tests
      taskRef:
        name: integration-test
  final:
    - name: cleanup-test
      taskRef:
        name: cleanup
```
However that would be a big change (tho pre-v1 would be
the time to do it!) and `finally` is consistent with how this kind of
functionality is expressed in languages like Python.

(Also I understand the finally section to be saying: once all of the above is
done, finally run clean-test, and not to say "cleanup-test is a finally
kind of task". )

The last option would be to support both final and finally but that
feels to me like it would be even more confusing.

/kind documentation

Question for the reviewers: which approach do you think we should take:

1. This one
2. Rename uses of "finally tasks" to "final tasks" (i.e. dont try to force using an adverb as an adjective)
3. Rename the "finally" section to "final"
4. Support both "finally" and "final" in the pipeline spec
5. Something else?

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -414,7 +414,7 @@ If you set the timeout to 0, the `PipelineRun` fails immediately upon encounteri
 
 > :warning: ** `timeout`will be deprecated in future versions. Consider using `timeouts` instead.
 
-You can use the `timeouts` field to set the `PipelineRun's` desired timeout value in minutes.  There are three sub-fields than can be used to specify failures timeout for the entire pipeline, for tasks, and for finally tasks.  
+You can use the `timeouts` field to set the `PipelineRun's` desired timeout value in minutes.  There are three sub-fields than can be used to specify failures timeout for the entire pipeline, for tasks, and for `finally` tasks.
 
 ```yaml
 timeouts:
@@ -563,7 +563,7 @@ Task Runs:
 To cancel a `PipelineRun` that's currently executing, update its definition
 to mark it as "PipelineRunCancelled". When you do so, the spawned `TaskRuns` are also marked
 as cancelled, all associated `Pods` are deleted, and their `Retries` are not executed.
-Pending final tasks are not scheduled.
+Pending `finally` tasks are not scheduled.
 
 For example:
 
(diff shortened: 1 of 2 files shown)


bobcatfish:
After writing this out im kinda leaning toward (2) instead - always saying "final" tasks, i.e. "final tasks" are tasks in the "finally section", but it seems like that might increase the confusion that @Srivaralakshmi was describing 🤔 

/hold

bobcatfish:
I've made one more tweak which might help: instead of saying "finally tasks" i've added code block markup: `finally` tasks - which might help make it a bit more clear that `finally` here is referring to a symbol in the pipeline spec vs trying to act as an adverb

jerop (commented):
there are several spots with ``finally` task` instead of `finally` tasks

that is, there's double code markup -- around "finally" and around "finally tasks"

otherwise looks great 👍🏾

vdemeester (approved):
Agreeing on (1) too 👼🏼

pritidesai:
thanks a bunch @bobcatfish for making it read better 😸

PR #4100, opened 2021-07-15
Loosen permission checks on TestCp to handle umasks.

### LINKED ISSUE #4099: cmd/entrypoint/subcommands/cp test does not respect umask
# Expected Behavior

[TestCp](https://github.com/tektoncd/pipeline/blob/584b5278138aac455d47e25eb9930b85a29db387/cmd/entrypoint/subcommands/cp_test.go#L27) passes on main branch.

# Actual Behavior

```sh
$ go test .
--- FAIL: TestCp (0.00s)
    cp_test.go:54: expected permissions 0311 for destination file but found 0310
FAIL
FAIL    github.com/tektoncd/pipeline/cmd/entrypoint/subcommands 0.017s
FAIL
```

This appears to be caused by an assumption that the system [umask](https://man7.org/linux/man-pages/man2/umask.2.html) would be fully permissive, and the [cp func](https://github.com/tektoncd/pipeline/blob/efaa416f02f8e7a8809346ddcc62fa50e1328398/cmd/entrypoint/subcommands/cp.go#L38) would fully preserve permissions. On my system, the default umask is:

```sh
$ umask
027
```

This strips out the all-read permissions on the file, since [os.OpenFile](https://pkg.go.dev/os#OpenFile) will set permissions post-mask -

> If the file does not exist, and the O_CREATE flag is passed, it is created with mode perm (before umask).

Some potential fixes-
1. Make a manual `os.Chmod` after the file creation, though this would not respect any existing permissions on the file. 
2. Set the umask via [unix.Umask](https://pkg.go.dev/golang.org/x/sys@v0.0.0-20210630005230-0f9fa26af87c/unix#Umask) but this is also problematic since IIUC this will set it for system and would only work for unix environments.
3. Loosen the test requirements to be less sensitive to permissions.

@sbwsg Any preferences between these options?

# Additional Info

- Tekton Pipeline version: efaa416f02f8e7a8809346ddcc62fa50e1328398

/assign @wlynch

wlynch:
# Changes

System umasks can futher restrict permissions passed to os.OpenFile,
resulting in a file that's more restricted than the permission value we
supplied. Instead of trying to match the permissions exactly, this
change modifies the check to make sure that the new file does not exceed
the expected permissions, but allows anything more restrictive.

Fixes #4099

/kind bug

/cc @sbwsg 

# Submitter Checklist

  functionality, content, code)
- [n/a] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/cmd/entrypoint/subcommands/cp_test.go
+++ b/cmd/entrypoint/subcommands/cp_test.go
@@ -46,8 +46,17 @@ func TestCp(t *testing.T) {
 		t.Fatalf("error statting destination file: %v", err)
 	}
 
-	if info.Mode().Perm() != dstPermissions {
-		t.Errorf("expected permissions %#o for destination file but found %#o", dstPermissions, info.Mode().Perm())
+	// os.OpenFile is subject to umasks, so the created permissions of the
+	// created dst file might be more restrictive than dstPermissions.
+	// excludePerm represents the value of permissions we do not want in the
+	// resulting file - e.g. if dstPermissions is 0311, excludePerm should be
+	// 0466.
+	// This is done instead of trying to look up the system umask, since this
+	// relies on syscalls that we are not sure will be portable across
+	// environments.
+	excludePerm := os.ModePerm ^ dstPermissions
+	if p := info.Mode().Perm(); p&excludePerm != 0 {
+		t.Errorf("expected permissions <= %#o for destination file but found %#o", dstPermissions, p)
 	}
 }
 


sbwsg (approved):

sbwsg:
👍  nice

pritidesai:
thank you @wlynch sounds good 👍 
/lgtm

PR #4111, opened 2021-07-21
Drop the tracker.

mattmoor:
# Changes

Tekton's use of the `tracker` is redundant with the informer events it already has configured, and the tracker lease the pipeline run controller is using is probably too small to actually be effective for long-running pipelines.

The tracker is built around the premise that resyncs happen to "refresh" things, and so to be safe the tracker lease should be a multiple of the resync period.  The pipeline controller is currently hardcoding `30m`, but the default resync period is `10h`, so anything taking longer than `30m` via the tracker will actually drop events.  The controller also seems to track a `TaskRun` with the name of the `PipelineRun`, which also feels wrong...  🤔 

Fortunately, I believe that both usages of the tracker are wholly redundant with informer events on owned resources anyhow, so this change should just remove unneeded complexity.

For additional background, the purpose of the tracked isn't to track owned resources, but rather to track resources referenced via `corev1.ObjectReference` (or equivalent).  We use this in Knative to track changes to Addressable resources that some resource might be delivering events to (e.g. maybe the address changed, or it is no longer Ready).

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

cc @dlorenc @imjasonh @vdemeester @afrittoli

### DIFF
--- a/pkg/reconciler/pipelinerun/controller.go
+++ b/pkg/reconciler/pipelinerun/controller.go
@@ -42,7 +42,6 @@ import (
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/kmeta"
 	"knative.dev/pkg/logging"
-	"knative.dev/pkg/tracker"
 )
 
 // NewController instantiates a new controller.Impl from knative.dev/pkg/controller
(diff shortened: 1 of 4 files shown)


mattmoor:
Rebased this on #4110 which just landed.

vdemeester (approved):
Seems good to me 😉

afrittoli (commented):
Thanks!
Going back all the way to https://github.com/tektoncd/pipeline/pull/757, this is the first reference of the tracker that I see, but it was in a different place in the code, and it does not seem to be applicable anymore - today a Pipeline is not marked as done until all associated TaskRun have finished execution in some way.

/lgtm

PR #4121, opened 2021-07-28
adding label `memberOf` to identify a taskRun in the pipeline

### LINKED ISSUE #3715: Is there any way to tell whether any of the finally tasks failed?
I have a pipeline that has
task1 -> task 2 -> ... -> taskN | singleFinallyTask.

When `singleFinallyTask` fails for whatever reasons, is there an easy way to use `kubectl` or `tkn` to tell?

### LINKED ISSUE #3721: Feature request to detect whether executed TaskRun is part of finally section
### Feature request

Ability to fetch/differentiate TaskRuns executed as part of Finally section in the PipelineRun

### Use case

Currently, there is no way to detect whether executed TaskRun is part of Finally section directly. Adding this feature can help external systems via SDK to directly query the TaskRuns which are part of Finally without applying complex logic.

pritidesai:
# Changes

A label named *memberOf* is added to all the taskRuns/Runs created while executing the pipeline based on their respective membership. A  `taskRun` will have a label `tekton.dev/memberOf=tasks` for the `task` defined under "tasks" section and a `taskRun` will have a label `tekton.dev/memberOf=finally` for the `task` defined under "finally" section. 

**Implementation Iterations**

Iteration1:

A label is added to a `taskRun` for a `task` which is part of the `finally` section. The label added with a pattern `tekton.dev/finallyTask` and set to `true` for a `finally` task.

Closes #3721 
Closes #3715 

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
A taskRun/pod created for a finally task can be identified with a label tekton.dev/finallyTask which is set to true.
```

### DIFF
--- a/docs/labels.md
+++ b/docs/labels.md
@@ -81,6 +81,12 @@ Tekton automatically adds labels to Tekton entities as described in the followin
 			<td>Name of the <code>TaskRun</code> that created the <code>Pod</code>.</td>
 		</tr>
 		<tr>
+			<td><code>tekton.dev/memberOf</code></td>
+			<td><code>TaskRuns</code> that are created automatically during the execution of a <code>PipelineRun</code>.</td>
+			<td><code>TaskRuns, Pods</code></td>
+			<td><code>tasks</code> or <code>finally</code> depending on the <code>PipelineTask</code>'s membership in the <code>Pipeline</code>.</td>
+		</tr>
+		<tr>
 			<td><code>app.kubernetes.io/instance</code>, <code>app.kubernetes.io/component</code></td>
 			<td><code>Pods</code>, <code>StatefulSets</code> (Affinity Assistant)</td>
 			<td>No propagation.</td>
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -907,6 +907,22 @@ func getTaskrunLabels(pr *v1beta1.PipelineRun, pipelineTaskName string, includeP
 	if pipelineTaskName != "" {
 		labels[pipeline.GroupName+pipeline.PipelineTaskLabelKey] = pipelineTaskName
 	}
+	if pr.Status.PipelineSpec != nil {
+		// check if a task is part of the "tasks" section, add a label to identify it during the runtime
+		for _, f := range pr.Status.PipelineSpec.Tasks {
+			if pipelineTaskName == f.Name {
+				labels[pipeline.GroupName+pipeline.MemberOfLabelKey] = v1beta1.PipelineTasks
+				break
+			}
+		}
+		// check if a task is part of the "finally" section, add a label to identify it during the runtime
+		for _, f := range pr.Status.PipelineSpec.Finally {
+			if pipelineTaskName == f.Name {
+				labels[pipeline.GroupName+pipeline.MemberOfLabelKey] = v1beta1.PipelineFinallyTasks
+				break
+			}
+		}
+	}
 	return labels
 }
 
--- a/pkg/apis/pipeline/register.go
+++ b/pkg/apis/pipeline/register.go
@@ -48,6 +48,10 @@ const (
 
 	// RunKey is used as the label identifier for a Run
 	RunKey = "/run"
+
+	// MemberOfLabelKey is used as the label identifier for a PipelineTask
+	// Set to Tasks/Finally depending on the position of the PipelineTask
+	MemberOfLabelKey = "/memberOf"
 )
 
 var (
--- a/pkg/apis/pipeline/v1beta1/pipeline_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_types.go
@@ -35,6 +35,10 @@ import (
 const (
 	// PipelineTasksAggregateStatus is a param representing aggregate status of all dag pipelineTasks
 	PipelineTasksAggregateStatus = "tasks.status"
+	// PipelineTasks is a value representing a task is a member of "tasks" section of the pipeline
+	PipelineTasks = "tasks"
+	// PipelineFinallyTasks is a value representing a task is a member of "finally" section of the pipeline
+	PipelineFinallyTasks = "finally"
 )
 
 // +genclient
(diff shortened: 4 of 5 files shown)


vdemeester (approved):

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -929,6 +929,16 @@ func combineTaskRunAndTaskSpecLabels(pr *v1beta1.PipelineRun, pipelineTask *v1be
...
+
+	// check if a task is part of the finally section, add a label to identify it during the runtime
+	if pr.Status.PipelineSpec != nil {
+		for _, f := range pr.Status.PipelineSpec.Finally {
+			if pipelineTask.Name == f.Name {
+				labels[pipeline.GroupName+pipeline.FinallyLabelKey] = "true"
+				break
+			}
+		}
+	}
```
afrittoli, line 941: This doesn't feel like 100% right place for this, this function combines `Task` spec and `TaskRun` labels, which is why the function does not have access to the pipeline spec, and you have to go and get it from the status. Whether a `PipelineTask` belongs to the `finally` section is a `Pipeline` level concern, so perhaps we could pass that info down to the `createTaskRun` function, adding an extra boolean param:

```
				if rprt.IsFinalTask(pipelineRunFacts) {
					rprt.TaskRun, err = c.createTaskRun(ctx, rprt, pr, as.StorageBasePath(pr), getFinallyTaskRunTimeout, True)
				} else {
					rprt.TaskRun, err = c.createTaskRun(ctx, rprt, pr, as.StorageBasePath(pr), getTaskRunTimeout, False)
				}
```

Alternatively you could use the [`GetPipelineFunc`](https://github.com/tektoncd/pipeline/blob/a728ce3c3d717fe52ff0b1eb71d357460498115c/pkg/reconciler/pipelinerun/resources/pipelineref.go#L36) function, which returns the spec from the status if available, or it will fetch it otherwise.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -929,6 +929,16 @@ func combineTaskRunAndTaskSpecLabels(pr *v1beta1.PipelineRun, pipelineTask *v1be
 			labels[key] = value
 		}
 	}
+
+	// check if a task is part of the finally section, add a label to identify it during the runtime
+	if pr.Status.PipelineSpec != nil {
```
afrittoli, line 934: If `nil` we just ignore it and go on?

afrittoli (commented):
Thanks for this, it sounds like a good idea!!
I would alter the implementation slightly though - see comments inline.

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: thanks @afrittoli I think this was a reasonable place since its combining labels from the `task` and `taskRun`. Thanks for suggestion, I have moved it to `getTaskrunLabels` function where we are adding other `tekton.dev` labels such as `tekton.dev/pipelineRun` and `tekton.dev/pipelineTask`.

Also, its safe and reasonable to get the specification from the `status` after we added it as a source of truth in https://github.com/tektoncd/pipeline/pull/3941

`createTaskRun` is growing complex with every new functionality we add. We really need a `factory` style interface for the `taskRun` and `Run` instead of operating on the parameters.

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: This `if` statement is just avoiding `nil` pointer reference and this logic is mainly for the `finally` task. By the time, a `finally` taskRun is getting created, the `status` is guaranteed to have the specification (which should be initialized before the first `taskRun` is created).

sbwsg:
I am curious why `"tekton.dev/finallyTask":"true"` was chosen over the `"memberOf"` approach?  I had a quick scan through the linked issues and couldn't find any pros/cons one way or the other.

I have a hunch that "memberOf" is better for the following reason: Imagine you have a cluster full of completed taskruns from your current Tekton install. You update to the new Tekton version which adds the "finallyTask" label. You query for "tekton.dev/finallyTask" of "true" and this returns all the finally tasks created after the label was introduced. Then you want the set of dag tasks so you query for "tekton.dev/finallyTask" of `<blank>` or `<nil>` or whatever your tooling supports. This returns both all of the dag taskruns _and_ all of the old taskruns created before the new label was introduced (regardless of finally/dag). With "memberOf" you can query directly for all the finally taskruns created after the label was introduced, all of the dag taskruns created after the label was introduced, and all of the "other" taskruns created before the label was introduced. 

No problem if we do decide it's preferable to stick with a label of "true", just wanted to make sure this choice is discussed first. I'd much prefer we don't come back and add another label like "memberOf" later if we can simply introduce it now.

WDYT?

pritidesai:
> I am curious why "tekton.dev/finallyTask":"true" was chosen over the "memberOf" approach? I had a quick scan through the linked issues and couldn't find any pros/cons one way or the other.

The only reason to choose one over the other: `bool` might be easier to consume compared to a pipeline specific field `finally`/`tasks` 😺 At the same time, you have captured very compelling significance of using `memberOf`, changing it to have `memberOf` label 👩‍💻

docs/labels.md
```diff
@@ -80,6 +80,12 @@ Tekton automatically adds labels to Tekton entities as described in the followin
 			<td>No propagation.</td>
 			<td>Name of the <code>TaskRun</code> that created the <code>Pod</code>.</td>
 		</tr>
+		<tr>
+			<td><code>tekton.dev/memberOf</code></td>
+			<td><code>TaskRuns</code> that are created automatically during the execution of a <code>PipelineRun</code>.</td>
+			<td><code>TaskRuns, Pods</code></td>
+			<td><code>tasks</code> or <code>finally</code> depending on the <code>PipelineTask</code>'s membership in the <code>Pipeline</code>.</td>
```
sbwsg, line 87: nice 👍  maps to the field names in the pipeline.

sbwsg:
Thanks @pritidesai !

/lgtm

PR #4126, opened 2021-07-29
Use `knative.dev/pkg/changeset`

mattmoor:
# Changes

This replaces the need for `./pkg/version` and a flag with the automatic inference of the version of the deployed code from the git metadata symlinked into `kodata/`

This package is already used to automatically decorate the zap loggers, and a few other places (iirc).

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


mattmoor:
cc @vdemeester @afrittoli @imjasonh @dlorenc

imjasonh:
Is the change to controller.yaml going to run afoul of the modification done in

https://github.com/tektoncd/pipeline/blob/a728ce3c3d717fe52ff0b1eb71d357460498115c/tekton/publish.yaml#L131-L132

mattmoor:
Thanks for making my eyes bleed @imjasonh 😉 

There's probably an expression in there we no longer need, but I hate touching bash that presubmits don't check.  Lemme look closer.

mattmoor:
@imjasonh @vdemeester this should be RFAL.

mattmoor:
post-weekend bump 😁

vdemeester:
@mattmoor does it support only a commit hash, or can it show a tag ?

mattmoor:
> does it support only a commit hash, or can it show a tag

As it is today, I believe it resolves the tag to a commit, but all the info is in kodata, so we could adapt the changeset package to do more.

vdemeester (approved):

vdemeester:
So @mattmoor one *downside* of this approach is, it fails if the images are not build with `ko` 😛 (which is the case for OpenShift Pipelines for example, we rebuild them differently). https://github.com/knative/pkg/blob/main/changeset/commit.go#L60 is *less* forgiving than what I initially though 😛

pkg/pod/pod.go
```diff
@@ -290,7 +290,11 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec
 	}
 
 	podAnnotations := taskRun.Annotations
-	podAnnotations[ReleaseAnnotation] = version.PipelineVersion
+	version, err := changeset.Get()
+	if err != nil {
+		return nil, err
+	}
+	podAnnotations[ReleaseAnnotation] = version
```
vdemeester, line 297: Maybe we should log something and print `unknown` here instead of erroring out, wdyt @mattmoor ?

Issue #3140, opened 2020-08-26
PipelineRun Status bloat and excessive updates


skaegi:
In our production systems we are seeing large PR objects frequently over 1MB in size when serialized as JSON with nearly all the information in status pipelineSpec and taskruns. This might seem big now but I would bet this is really just the beginning and as we see more Tekton usage things are going to get a lot larger. In addition as a pipeline runs the PRs get updated along with changes to status.taskruns. This is triggering a huge amount of traffic for listening systems like the dashboard and our other systems that react to events in cluster. We're already hitting internal cluster limitations and have had to trim the PRs significantly to avoid overwhelming our messaging systems. 

Looking at how the core Kubernetes API use status, it seems limited to very basic information and avoiding redundant information already captured in child or related objects. How painful would it be to re-work status to be lightweight?

ImJasonH:
To me this seems to be closely related to https://github.com/tektoncd/pipeline/issues/454 -- if we had an off-cluster system to store *Run statuses, we wouldn't be bound by etcd's storage and performance issues around l arge objects. (That being said, we _would_ be bound by whatever _that system's_ limitations are 😄 )

As an alternative, we could phase out reporting all of each sub-*Runs' statuses in the PipelineRun, and just require clients to look up all the constituent *Runs' statuses in order to display all that information. Or we could only report sub-Runs' `.status.conditions` to signal success/failure/ongoing, and stop reporting per-step information, timing, etc. These would need to be carefully communicated and staged slowly to avoid breaking clients that may rely on this information today.

bobcatfish:
i believe the chains folks will have some thoughts here as well @dlorenc

bobcatfish:
Also more info in https://github.com/tektoncd/pipeline/issues/279

The PRs that made these changes are https://github.com/tektoncd/pipeline/pull/2444 and https://github.com/tektoncd/pipeline/pull/2489

> In our production systems we are seeing large PR objects frequently over 1MB in size when serialized as JSON with nearly all the information in status pipelineSpec and taskruns.

one thought: maybe dereferencing all the taskruns into the pipelinerun is too far?

ImJasonH:
> one thought: maybe dereferencing all the taskruns into the pipelinerun is too far?

[...]

PR #4127, opened 2021-07-29
Implement implicit parameter resolution.

wlynch:
# Changes

This is the implementation of
[TEP-0023](https://github.com/tektoncd/community/blob/main/teps/0023-implicit-mapping.md).

This adds information to the defaulting context to allow parameters to
be passed down the stack during evaluation. This functionality is gated
behind the alpha feature flag.

Additionally, Tasks are now allowed to accept more params than are
actually used. This should generally be safe, since extra params that
are provided should not affect the behavior of the Task itself.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
- Tasks can now accept more parameters than are actually used.
- Parameters can now be implicitly propagated to inlined specs (alpha feature flag must be enabled) -
  e.g. an inlined Task can access parameters of its parent PipelineRun without
  needing to explicitly define each param.
```

### DIFF
--- a/config/200-role.yaml
+++ b/config/200-role.yaml
@@ -52,7 +52,7 @@ rules:
   - apiGroups: [""]
     resources: ["configmaps"]
     verbs: ["get"]
-    resourceNames: ["config-logging", "config-observability", "config-leader-election"]
+    resourceNames: ["config-logging", "config-observability", "config-leader-election", "feature-flags"]
   - apiGroups: [""]
     resources: ["secrets"]
     verbs: ["list", "watch"]
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -246,6 +246,75 @@ case is when your CI system autogenerates `PipelineRuns` and it has `Parameters`
 provide to all `PipelineRuns`. Because you can pass in extra `Parameters`, you don't have to
 go through the complexity of checking each `Pipeline` and providing only the required params.
 
+#### Implicit Parameters
+
+**([alpha only](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features))**
+
+When using an inlined spec, parameters from the parent `PipelineRun` will be
+available to any inlined specs without needing to be explicitly defined. This
+allows authors to simplify specs by automatically propagating top-level
+parameters down to other inlined resources.
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: echo-
+spec:
+  params:
+    - name: MESSAGE
+      value: "Good Morning!"
+  pipelineSpec:
+    tasks:
+      - name: echo-message
+        taskSpec:
+          steps:
+            - name: echo
+              image: ubuntu
+              script: |
+                #!/usr/bin/env bash
+                echo "$(params.MESSAGE)"
+```
+
+On creation, this will resolve to a fully-formed spec and will be returned back
+to clients to avoid ambiguity:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: echo-
+spec:
+  params:
+  - name: MESSAGE
+    value: Good Morning!
+  pipelineSpec:
+    params:
+    - name: MESSAGE
+      type: string
+    tasks:
+    - name: echo-message
+      params:
+      - name: MESSAGE
+        value: $(params.MESSAGE)
+      taskSpec:
+        params:
+        - name: MESSAGE
+          type: string
+        spec: null
+        steps:
+        - name: echo
+          image: ubuntu
+          script: |
+            #!/usr/bin/env bash
+            echo "$(params.MESSAGE)"
+```
+
+Note that all implicit Parameters will be passed through to inlined resources
+(i.e. PipelineRun -> Pipeline -> Tasks) even if they are not used.
+Extra parameters passed this way should generally be safe (since they aren't
+actually used), but may result in more verbose specs being returned by the API.
+
 ### Specifying custom `ServiceAccount` credentials
 
 You can execute the `Pipeline` in your `PipelineRun` with a specific set of credentials by
--- a/docs/taskruns.md
+++ b/docs/taskruns.md
@@ -163,6 +163,69 @@ spec:
 
 **Note:** If a parameter does not have an implicit default value, you must explicitly set its value.
 
+#### Implicit Parameters
+
+**([alpha only](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features))**
+
+When using an inlined `taskSpec`, parameters from the parent `TaskRun` will be
+available to the `Task` without needing to be explicitly defined.
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  generateName: hello-
+spec:
+  params:
+    - name: message
+      value: "hello world!"
+  taskSpec:
+    # There are no explicit params defined here.
+    # They are derived from the TaskRun params above.
+    steps:
+    - name: default
+      image: ubuntu
+      script: |
+        echo $(params.message)
+```
+
+On creation, this will resolve to a fully-formed spec and will be returned back
+to clients to avoid ambiguity:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  generateName: hello-
+spec:
+  params:
+    - name: message
+      value: "hello world!"
+  taskSpec:
+    params:
+    - name: message
+      type: string
+    steps:
+    - name: default
+      image: ubuntu
+      script: |
+        echo $(params.message)
+```
+
+Note that all implicit Parameters will be passed through to inlined resource,
+even if they are not used. Extra parameters passed this way should generally
+be safe (since they aren't actually used), but may result in more verbose specs
+being returned by the API.
+
+#### Extra Parameters
+
+**([alpha only](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features))**
+
+You can pass in extra `Parameters` if needed depending on your use cases. An example use
+case is when your CI system autogenerates `TaskRuns` and it has `Parameters` it wants to
+provide to all `TaskRuns`. Because you can pass in extra `Parameters`, you don't have to
+go through the complexity of checking each `Task` and providing only the required params.
+
 ### Specifying `Resources`
 
 If a `Task` requires [`Resources`](tasks.md#specifying-resources) (that is, `inputs` and `outputs`) you must
--- a/pkg/apis/pipeline/v1beta1/param_context.go
+++ b/pkg/apis/pipeline/v1beta1/param_context.go
@@ -1,0 +1,172 @@
+// Copyright 2021 The Tekton Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1beta1
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+)
+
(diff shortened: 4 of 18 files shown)


wlynch:
/hold

Looks like I missed some tests

docs/pipelineruns.md
```diff
@@ -246,6 +246,34 @@ case is when your CI system autogenerates `PipelineRuns` and it has `Parameters`
...
 
+#### Implicit Parameters
+
+**(v1alpha1 only)**
+
+When using an inlined spec, parameters from the parent `PipelineRun` will be
+available to any inlined specs without needing to be explicitly defined.
+
+```yaml
+apiVersion: tekton.dev/v1alpha1
```
bobcatfish, line 257: it looks like this functionality is being marked alpha by adding it only to the alpha type - thats a different approach than we've taken via [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md) - the approach we're favoring is to us [enable-api-fields: alpha](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#customizing-the-pipelines-controller-behavior) to gate functionality like this - this is a _controller_ setting though so if this is implemented in the webhook, maybe we need to either port that flag over to the webhook, or let the webhook read that value from the same configmap as the controller? probably doesnt make sense to need to configure it twice... TEP-0033 could use an update to include the webhook, though it does say:

> Since the API fields will exist in the CRD with or without the flags, this will be enforced at runtime by the webhook admission controller: if the flag is not enabled and a user tries to use an alpha field, the webhook will not allow the CRD to be stored and will return an actionable error. (Same for beta.)

which feels like it implies that the webhook should be aware of this configuration

wlynch:
/hold 

Talked a bit off thread about this - seems like this should have been added to v1beta1 instead, relying on the alpha flag to enable the functionality. Will update.

wlynch:
/hold cancel

Good to review! 🎉

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -88,15 +90,27 @@ func validateParams(paramSpecs []v1beta1.ParamSpec, params []v1beta1.Param) erro
...
+		}
 	}
 
 	// Now that we have checked against missing/extra params, make sure each param's actual type matches
 	// the user-specified type.
 	var wrongTypeParamNames []string
 	for _, param := range params {
+		if _, ok := paramTypes[param.Name]; !ok {
+			// Ignore any missing params - this happens when extra params were
+			// passed to the task that aren't being used.
```
vdemeester, line 111: Aren't missing params the opposite, aka when a param is defined by a `Task` and not on the `TaskRun` ?

pkg/apis/pipeline/v1beta1/param_context.go
```diff
@@ -0,0 +1,158 @@
+package v1beta1
```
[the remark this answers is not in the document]
wlynch: Done.

pkg/reconciler/taskrun/validate_resources.go
wlynch: Reincluded - I realized after why this is needed.

For cases where we use taskRefs, we can't propagate the implicit param down to the Task to create the 1:1 mapping this validation wants. In these cases, we may have excess params provided by the parameter, a subset of which are actually used / type checked against the resolved task spec - everything else that doesn't match we can just pass through since there's no pairing.

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -88,15 +90,27 @@ func validateParams(paramSpecs []v1beta1.ParamSpec, params []v1beta1.Param) erro
...
+	// If alpha features are enabled, disable checking for extra params.
+	// Extra params are needed to support
+	// https://github.com/tektoncd/community/blob/main/teps/0023-implicit-mapping.md
+	// So that parent params can be passed down to subresources (even if they
+	// are not explicitly used).
+	if config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields != "alpha" {
+		extraParams := list.DiffLeft(providedParams, neededParams)
+		if len(extraParams) != 0 {
+			return fmt.Errorf("didn't need these params but they were provided anyway: %s", extraParams)
+		}
```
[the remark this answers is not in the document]
wlynch: Restored (see comment below). Added a comment that mirrors the PipelineRun wording, since that seems to have the clearest user benefit rather than just and implementation detail of this feature.

sbwsg (approved):

docs/taskruns.md
```diff
@@ -163,6 +163,64 @@ spec:
 
 **Note:** If a parameter does not have an implicit default value, you must explicitly set its value.
 
+#### Implicit Parameters
+
+**(alpha only)**
```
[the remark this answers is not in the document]
wlynch: done.

docs/pipelineruns.md
```diff
@@ -246,6 +246,70 @@ case is when your CI system autogenerates `PipelineRuns` and it has `Parameters`
 provide to all `PipelineRuns`. Because you can pass in extra `Parameters`, you don't have to
 go through the complexity of checking each `Pipeline` and providing only the required params.
 
+#### Implicit Parameters
+
+**(alpha only)**
```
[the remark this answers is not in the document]
wlynch: done.

wlynch:
ping for review!

pritidesai:
thanks @wlynch for this, looking forward to this feature 😍  This will simplify the resource definitions a whole lot.

I have started to review this PR. In meanwhile, can you please update the PR description and release notes to exclude `v1alpha1` API and include the feature flag `enable-api-alpha` set to `alpha`. Also will be great if you squash all the commits 🙏

config/200-role.yaml
```diff
@@ -52,7 +52,7 @@ rules:
   - apiGroups: [""]
     resources: ["configmaps"]
     verbs: ["get"]
-    resourceNames: ["config-logging", "config-observability", "config-leader-election"]
+    resourceNames: ["config-logging", "config-observability", "config-leader-election", "feature-flags"]
```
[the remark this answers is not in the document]
pritidesai: That's interesting, then how come the webhook is throwing an error when an `alpha` feature is used without enabling featuring flag and allow with the feature flag set. May be I am missing something, will find out as I review the rest of the changes.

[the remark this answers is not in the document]
wlynch: Although the controller was trying to read the config, we weren't ever trying to use it anywhere in the webhook before this change.

A side affect of https://github.com/tektoncd/pipeline/blob/78d6d75473ea00118fa1e07cbf6b8e97439689ad/cmd/webhook/main.go#L68 is [it tries to load all of the Pipeline controller configs](https://github.com/tektoncd/pipeline/blob/78d6d75473ea00118fa1e07cbf6b8e97439689ad/pkg/apis/config/store.go#L85-L89), regardless of whether they are actually used or not.

pritidesai:
I ran a `pipeline` with one parameter `GREETINGS`. And the pipeline has one `pipelineTask` with one parameter `MESSAGE`. The parameter in the `pipelineTask` is explicitly mapped to the parameter in the `pipeline`:

```yaml
       params:
          - name: MESSAGE
            value: $(params.GREETINGS)
```

Without these changes, the resolved `pipelineSpec` includes this mapping in the `pipelineTask`:

```yaml
spec:
  pipelineSpec:
    params:
    - default: Hello World!
      description: Greetings, default is Hello World!
      name: GREETINGS
      type: string
    tasks:
    - name: echo-greetings
      params:
      - name: MESSAGE
        value: $(params.GREETINGS)
      taskSpec:
```

With these changes, in addition to the explicit mapping, the resolved spec includes an additional implicit mapping along with the `pipeline` parameter listed as `pipelineTask` and `task` parameter. `GREETINGS` was not listed as the parameter of any `task`.

```yaml
spec:
  pipelineSpec:
    params:
    - default: Hello World!
      description: Greetings, default is Hello World!
      name: GREETINGS
      type: string
    tasks:
    - name: echo-greetings
      params:
      - name: MESSAGE
        value: $(params.GREETINGS)
      - name: GREETINGS
        value: $(params.GREETINGS)
      taskSpec:
``` 

I am not sure if this can cause any conflicts but I think we can atleast document this behavior. Wdyt? I have the entire `pipelineRun` and resolved specification documented [here](https://gist.github.com/pritidesai/11d1d7a8af624f21d59b5e57907bd8c8) for your reference..

pkg/apis/pipeline/v1beta1/param_context.go
```diff
@@ -0,0 +1,158 @@
...
+
```
pritidesai, line 2: missing license header 😜

pritidesai: and the test file below 😸

wlynch: Done.

wlynch:
> I ran a `pipeline` with one parameter `GREETINGS`. And the pipeline has one `pipelineTask` with one parameter `MESSAGE`. The parameter in the `pipelineTask` is explicitly mapped to the parameter in the `pipeline`:
>
> I am not sure if this can cause any conflicts but I think we can atleast document this behavior. Wdyt? I have the entire `pipelineRun` and resolved specification documented [here](https://gist.github.com/pritidesai/11d1d7a8af624f21d59b5e57907bd8c8) for your reference..
> 
> EDIT: This could potentially explode the list of parameters in each `pipelineTask` and `task`. I am sure there must be a way to avoid this, looking into it.

Copying my respond from Slack here (with minor edits) - I think this is WAI (at least for now). you would need it to cover weird edge cases like this:

```yaml
kind: PipelineRun
metadata:
  generateName: echo-
spec:
  params:
    - name: MESSAGE1
      value: "Good Morning!" 
  pipelineSpec:
    params:
    - name: MESSAGE2
      default: Hello World!
    tasks:
      - name: echo-message
        params:
          - name: MESSAGE3
            value: $(params.MESSAGE2)
        taskSpec:
          steps:
            - name: echo
              image: ubuntu
              script: |
                #!/usr/bin/env bash
                echo "$(params.MESSAGE1)" "$(params.MESSAGE2)" "$(params.MESSAGE3)"
```

This behavior seems to be most consistent with what we're trying to go for. e.g. if I was to write something similar in Go it would look like -

```go
a := "Good Morning!"
func() {
  b := "Hello World!"
  func(c string) {
    fmt.Println(a, b, c)
  }(b)
}()
```

The params available always needs to be a union of the other parameters in the context, since we don't know which will be used. The verboseness of the output config is a consequence of us returning the fully resolved parameter context so it's clear which params reached the Task. It shouldn't cause conflicts - params defined lower in the stack should always take precedence, and unused parameters shouldn't modify behavior since they aren't used. I suppose there is a small concern around complex pipelines hitting etcd limits, but the work around there would be to stop inlining everything and break things into discrete Tasks.

Alternatives I can think of:
- do some more complex matching (e.g. only set things that are used) - but this is tricky because we don’t actually know the params that are used until we evaluate them (which happens post-webhook on reconcile). This would require either more complex param validation upfront, or we accept the implicit config as-is and handle the resolution later.
- we remove the mutating defaulting behavior and store the implicit configs as-is. The configs will look more concise to users, and the parameter context mucking would only be visible to the reconciler. I wasn’t sure what impact this would have on current webhook validation, so I avoided this for the time being.

---

The alternatives make me think a reasonable next step would be to look into accepting the implicit configs as-is, and see what impact this would have on the current parameter validation (and also get feedback if not including the fully resolved spec would make debugging harder). My preference would be to tackle this in another change/PR.

Let me know what you think!

docs/pipelineruns.md
```diff
@@ -246,6 +246,75 @@ case is when your CI system autogenerates `PipelineRuns` and it has `Parameters`
...
+          image: ubuntu
+          script: |
+            #!/usr/bin/env bash
+            echo "$(params.MESSAGE)"
+```
+
+Note that all implicit Parameters will be passed through to inlined resources
+(i.e. PipelineRun -> Pipeline -> Tasks) even if they are not used.
+Extra parameters passed this way should generally be safe (since they aren't
+actually used), but may result in more verbose specs being returned by the API.
```
pritidesai, line 316: thanks a bunch @wlynch for adding this note 🙏

pritidesai:
thanks a bunch @wlynch 

> The alternatives make me think a reasonable next step would be to look into accepting the implicit configs as-is, and see what impact this would have on the current parameter validation (and also get feedback if not including the fully resolved spec would make debugging harder). My preference would be to tackle this in another change/PR.

> Let me know what you think!

Yup, we can certainly tackle this in a follow up PR. My major [concern](https://github.com/tektoncd/pipeline/issues/3140) is exploding the size of the `pipelineRun` status, especially for complex pipelines with a huge number of tasks and each task having huge number of parameters.

/lgtm

🤞 for the integration tests

Issue #1826, opened 2020-01-08
Support Windows


ImJasonH:
This issue tracks work needed to make Tekton run on a Kubernetes cluster with Windows nodes. Please contribute your use cases, ideas and experience.

First, some assumptions I've been operating under:

1. Tekton controller components themselves don't need to run on Windows nodes. The cluster could have N Linux nodes and M Windows nodes, run the controller on one of the Linux nodes, and run requested workloads on Windows nodes. If we decide full-Windows clusters are the goal, then we'll need to expand this work to cover getting controller components built and running for Windows.
1. The user has to explicitly specify they want a TaskRun to execute on a Windows node, using `nodeSelector` -- Tekton isn't responsible for somehow detecting that a TaskRun should run on a Windows node. We might change how a TaskRun's Pod is generated based on the user's specification, but ideally we wouldn't have to, for simplicity.

If either of these are contentious, or you have more to add, please comment.

Next, some things I'm fairly sure will break without trying:

1. **Building Images**: Entrypoint binary injection is done today by prepending a step containing the entrypoint binary, which copies that binary to `/tekton/entrypoint` -- to support Windows, that binary will have to be built to run on Windows, and will have to be based on some minimal Windows image (i.e., not `distroless` as it is today). This is doable with manifest lists, which `ko` has some support for. Likewise, internal support images used to implement resources (`git-init`, `creds-init`, `google/cloud-sdk`, etc., will need to be able to be built for Windows)

1. **File Paths**: Step ordering works by looking for files to exist at some predetermined path, and those paths are specified using `/`-separated paths (`/tekton/tools`, `/tekton/downward/ready`, etc.) -- this is likewise true for `/workspace`, `/tekton/home`, etc. If we decide in the future to use fsnotify to watch files instead of polling, we'll have to make sure that works for Windows too (or fallback to polling).

1. **Script Mode**: Script mode is implemented today by writing an executable file to `/tekton/scripts/something`, then invoking it as the `Command` in the step container. Writing this file and making it executable, as well as where it exists, will require work to support Windows.

Sources:
* https://cloud.google.com/blog/products/containers-kubernetes/how-to-deploy-a-windows-container-on-google-kubernetes-engine

afrittoli:
@ImJasonH any special reason for this to be on the priority list? I will remove it for now, feel free to bring it back to the API WG if needed

lukehb:
[...]

PR #4128, opened 2021-07-30
Add script mode support to windows tasks

DrWadsy:
# Changes

Steps and sidecars can contain a script field. In Linux tasks these
scripts are copied into files which are made executable and then
steps are added to the task to execute those files. This commit adds
comparable functionality if scripts are used in a task which will run
on a Windows node.

On a Windows node the mechanics are different, due to how Windows
handles executable files. The key difference is that Tekton needs to
know that a script will run on Windows, and how to run the file
(which interpreter to use). This is done through a ‘windows shebang’
line at the start of the script.

The line must begin with `#!win`. After that the user needs to provide
the interpreter to use, as well as any necessary arguments. The line
must be written such that the name of the file containing the script to
execute can be appended to the end.

For example, to run the script in the file `test.ps1` with powershell,
the command would usually be `powershell -File test.ps1` and so the
shebang line must be `#!win powershell -File`.

If no interpreter is provided (i.e. the shebang line is only ‘#!win’) then
the script contents will be stored in a .cmd file and executed.

Finally, since a pod cannot contain a mix of Windows and Linux
containers a windows shell image has been added to the Images
structure, which will be used in the place-scripts step when needed
on a windows node.

Related:
Issue [#1826](https://github.com/tektoncd/pipeline/issues/1826)
[TEP-0057](https://github.com/tektoncd/community/blob/main/teps/0057-windows-support.md)

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes
```release-note
Add script mode support for windows Tasks/Pipelines
```


pkg/pod/script.go
```diff
@@ -205,3 +230,57 @@ cat > ${scriptfile} << '%s'
...
 	return base64.StdEncoding.EncodeToString([]byte(script))
 }
+
+func checkWindowsRequirement(steps []v1beta1.Step, sidecars []v1beta1.Sidecar) bool {
+	requiresWindows := false
+	// Detect windows shebangs
+	for _, step := range steps {
+		cleaned := strings.TrimSpace(step.Script)
+		if strings.HasPrefix(cleaned, "#!win") {
+			requiresWindows = true
```
imjasonh, line 240: Could this just be `return true`?

Same below with the `requiresWindows = true; break` which just does `return requiresWIndows`.

imjasonh (commented):
First of all, this is great. 🤩 

I'd love to have an e2e test covering this, which probably means engaging with the plumbing repo to ensure the test cluster has some Windows nodes. That'd be great to have even without this script support, to ensure we can run even non-script Windows tasks.

I don't think we necessarily _need_ to block this PR on e2e tests, but I want to make sure we have those in place before we promise much more support.

pkg/pod/script_test.go
```diff
@@ -365,7 +365,7 @@ func TestConvertScripts_WithSidecar(t *testing.T) {
 		MountPath: "/another/one",
 	}}
 
-	gotInit, gotSteps, gotSidecars := convertScripts(images.ShellImage, []v1beta1.Step{{
+	gotInit, gotSteps, gotSidecars := convertScripts(images.ShellImage, images.ShellImageWin, []v1beta1.Step{{
 		Script: `#!/bin/sh
 script-1`,
 		Container: corev1.Container{Image: "step-1"},
```
imjasonh, line 371: Can we also add unit test coverage for this new behavior in this file?

pkg/pod/pod.go
```diff
@@ -133,9 +133,9 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec
 	// Convert any steps with Script to command+args.
 	// If any are found, append an init container to initialize scripts.
 	if alphaAPIEnabled {
-		scriptsInit, stepContainers, sidecarContainers = convertScripts(b.Images.ShellImage, steps, taskSpec.Sidecars, taskRun.Spec.Debug)
+		scriptsInit, stepContainers, sidecarContainers = convertScripts(b.Images.ShellImage, b.Images.ShellImageWin, steps, taskSpec.Sidecars, taskRun.Spec.Debug)
```
imjasonh, line 136: Should windows support go into the alpha feature gate, alongside debug support? 🤔

pkg/pod/script.go
DrWadsy: Yes, this should just be return true. The variable is an artifact from before I refactored this into a helper method, I'll change that :)

pkg/pod/pod.go
DrWadsy: Good point.
I changed this call because I changed the signature of the convertScripts method, but we could pass nil and check for that before we do any windows checks?

DrWadsy:
Is there anything else I need to do here to get this moving forward?

imjasonh:
After https://github.com/tektoncd/pipeline/pull/4139 is merged it would be great to have some e2e tests defined for this as well. We're still working on adding automation to those tests, though.

pkg/entrypoint/termination
```diff
@@ -0,0 +1 @@
+[{"key":"StartedAt","value":"2021-08-19T11:40:06.469+10:00","type":"InternalTektonResult"},{"key":"ExitCode","value":"2","type":"InternalTektonResult"},{"key":"StartedAt","value":"2021-08-19T11:40:06.504+10:00","type":"InternalTektonResult"},{"key":"StartedAt","value":"2021-08-19T11:40:06.519+10:00","type":"InternalTektonResult"},{"key":"StartedAt","value":"2021-08-19T11:40:06.531+10:00","type":"InternalTektonResult"}]
```
sbwsg, line 1: This is an artifact of a unit test in `pkg/entrypoint` and should be removed I think.  If you rebase on `main` and re-run `go test ./...` it shouldn't appear again.

pkg/pod/pod.go
sbwsg: Passing nil or an empty string for the windows shell image when `alphaAPIEnabled` is `false` would work ok I think. An alternative would be to write an entirely windows-focused `convertScripts` implementation and only invoke it if `alphaAPIEnabled` is `true` and `requiresWindows` would be true. Might involve a bit more duplication or end up being more lines but it might reduce the number of windows-specific branches? Could also push a change like that back to a refactor pr in future as well though.

pkg/pod/script.go
```diff
@@ -74,16 +74,28 @@ var (
 // It does this by prepending a container that writes specified Script bodies
 // to executable files in a shared volumeMount, then produces Containers that
 // simply run those executable files.
-func convertScripts(shellImage string, steps []v1beta1.Step, sidecars []v1beta1.Sidecar, debugConfig *v1beta1.TaskRunDebug) (*corev1.Container, []corev1.Container, []corev1.Container) {
+func convertScripts(shellImageLinux string, shellImageWin string, steps []v1beta1.Step, sidecars []v1beta1.Sidecar, debugConfig *v1beta1.TaskRunDebug) (*corev1.Container, []corev1.Container, []corev1.Container) {
 	placeScripts := false
 	// Place scripts is an init container used for creating scripts in the
```
sbwsg, line 79: This comment goes with the `corev1.Container` struct declared on line 94 I think.  Suggest moving it down or removing it.

pkg/pod/script.go
```diff
@@ -211,3 +238,55 @@ cat > ${scriptfile} << '%s'
...
+	requiresWindows := false
+	// Detect windows shebangs
+	for _, step := range steps {
+		cleaned := strings.TrimSpace(step.Script)
+		if strings.HasPrefix(cleaned, "#!win") {
+			return true
+		}
+	}
+	// If no step needs windows, then check sidecars to be sure
+	if !requiresWindows {
```
sbwsg, line 252: Given `requiresWindows` is never changed I don't think it's needed.  Simply running the loop over steps and then sidecars should be equivalent I think?

sbwsg (commented):
I left a bunch of comments but it looks like it's almost ready to go to me.  Biggest question to my mind is whether we branch in our existing funcs (like `convertScripts`) or write windows-specific versions of those.

docs/tasks.md
```diff
@@ -265,6 +265,51 @@ steps:
...
       /bin/my-binary
 ```
+
+**Windows scripts**
+
+Scripts in tasks that will eventually run on windows nodes need a custom shebang line, so that Tekton knows how to run the script. The format of the shebang line is:
+
+`#!win <interpreter command> <args>`
+
+Unlike linux, we need to specify how to interpret the script file which is generated by Tekton. The example below shows how to execute a powershell script:
```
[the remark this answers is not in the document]
DrWadsy: We have to be explicit that this script will run under windows, and we also have to tell tekton _how_ this script is to be run (which is what a linux shebang does for us). Windows needs to know which executable (e.g. powershell.exe, or pwsh.exe) it needs to use in order to run the script

pkg/entrypoint/termination
DrWadsy: Good catch :) I'm pretty sure I've rebased on main recently, so I'll remove that file...

sbwsg:
Thanks a lot for tackling those changes @DrWadsy . One more suggestion here - to maintain parity with other alpha features I think we should probably reject submitted TaskRuns / Tasks that use a windows script if the `enable-api-fields` feature flag is not `alpha`.  This gives users immediate feedback when they submit rather than waiting for a runtime error in the TaskRun's description.

Adding validation is super easy - we can do this in `pkg/apis/pipeline/v1beta1/task_validation.go`; here's existing code that performs a very similar check for the `onError` field of each Step: https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/task_validation.go#L224-L233

I think the additional validation would look something like:

```go
if s.Script != "" {
  cleaned := strings.TrimSpace(s.Script)
  if strings.HasPrefix(cleaned, "#!win") {
    errs = errs.Also(ValidateEnabledAPIFields(ctx, "windows script support", config.AlphaAPIFields).ViaField("script"))
  }
}
```

Adding a unit test to confirm this behaviour should then be as simple as adding an entry to the table in https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/task_validation_test.go#L1200-L1202

sbwsg:
Nice one, cheers!

/approve

sbwsg:
Also just noticed - could you squash the commits down too? We typically squash down into 1 before merge to help keep the history a little more easily scannable. I normally do that with `git rebase -i HEAD^^^^^^^^^^^^^` (one caret for each of the 13 commits) then you'd need to `git push --force` the branch. Doing so won't affect the approve I just added.

DrWadsy:
@imjasonh I made some code changes and squashed this all down into a single commit, and so the PR lost the lgtm you added. Could you please check this over one more time? :)

dlorenc:
@imjasonh is out for another week I think, so I took a look.

/lgtm

PR #4145, opened 2021-08-11
Add debug with breakpoint configuration to PipelineRun

jstrachan:
Based on TEP-0042, this commit will allow users to enable the `debug.breakpoints` property on any `TaskRun` resources generated by a `PipelineRun`

the reconciler will pass any `debug.breakpoints` properties from the PipelineRun's taskRunSpecs over to the generated resolved `TaskRun` instances so that per Task debugging can be enabled

Signed-off-by: James Strachan <james.strachan@gmail.com>

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

### DIFF
--- a/docs/debug.md
+++ b/docs/debug.md
@@ -21,7 +21,12 @@ weight: 11
 ## Overview
 
 `Debug` spec is used for troubleshooting and breakpointing runtime resources. This doc helps understand the inner 
-workings of debug in Tekton. Currently only the `TaskRun` resource is supported. 
+workings of debug in Tekton. Debugging is supported in the `TaskRun`. 
+
+
+## Debugging PipelineRuns
+
+To see how to configure a `PipelineRun` to enable debugging of a `TaskRun` follow the [PipelineRun Debugging Documentation](pipelineruns.md#debugging-pipelineruns). The rest of the documentation assumes debugging of a `TaskRun`.
 
 ## Debugging TaskRuns
 
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -23,6 +23,7 @@ weight: 500
   - [Monitoring execution status](#monitoring-execution-status)
   - [Cancelling a `PipelineRun`](#cancelling-a-pipelinerun)
   - [Pending `PipelineRuns`](#pending-pipelineruns)
+  - [Debugging `PipelineRuns`](#debugging-pipelineruns)
 
 
 
@@ -638,6 +639,28 @@ spec:
 ```
 
 To start the PipelineRun, clear the `.spec.status` field. Alternatively, update the value to `PipelineRunCancelled` to cancel it.
+                      
+
+### Debugging `PipelineRuns`
+
+#### Breakpoint on Failure
+
+You can configure a TaskRun created by a `PipelineRun` to enable debug as follows:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: test-case-run
+spec:
+  taskRunSpecs:
+  - pipelineTaskName: my-task 
+    debug:
+      breakpoint: ["onFailure"]
+...
+```
+
+Then refer to the [TaskRun Debugging Documentation](taskruns.md#debugging-a-taskrun) for details of how to debug the `TaskRun`
 
 ---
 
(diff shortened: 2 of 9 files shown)


pritidesai:
Trying to understand how this would work? debug every resource in a sequence or the order its created in? would this pause the execution of the entire pipeline? how would you resume? do you really need debugging every resource a pipeline creates? or just a step level debug is sufficient, for example, pausing a particular [step](https://github.com/tektoncd/pipeline/pull/4106#discussion_r683546445) irrespective of whether its part of a `taskRun` or a `pipelineRun`? Of course, a step level debug wouldn't work with pipeline resources.

jstrachan:
@pritidesai this PR let’s you enable debugging on a PipelineRun for a named Task, so that when the controller creates the TaskRun, the `debug` setting is applied. From that point onwards the TaskRun the breakpoint works exactly the same as if you created a TaskRun with the same `debug` configuration. 

If a TaskRun pauses on a breakpoint there is no effect on any other Task in the same pipeline. 

Resume works the same as with a TaskRun - running scripts inside the pod.

Basically debug/breakpoints only affect a TaskRun. The reason for this PR is that the tekton controller creates the TaskRun resources if you use a PipelineRun

jstrachan:
@vdemeester you mean in the validate code: https://github.com/tektoncd/pipeline/pull/3857/files#diff-1153e019c943846e52e599290b137d535b134d52a5d922e32ed42c5cdefe6f8aR81 or is there another way to mark the `debug` field as alpha?

jstrachan:
@vdemeester I just added `alpha` validation to the new `debug` field on PipelineRun.spec.taskRunSpecs https://github.com/tektoncd/pipeline/pull/4145/commits/322a861bd387d9a3056e7fdb7b395642e7dd34f1

jstrachan:
here's a little [blog and demo of how to use this PR with Jenkins X and Lens to debug pipelines](https://jenkins-x.io/blog/2021/08/18/debug-tekton/)

waveywaves:
Thanks for the PR on debug !

jstrachan:
@waveywaves how about we allow a global debug setting applied to all Tasks and a per task debug option & a way to merge the 2 structs? So that folks can use one or the other?

waveywaves:
> how about we allow a global debug setting applied to all Tasks and a per task debug option & a way to merge the 2 structs? So that folks can use one or the other?

@jstrachan 
that sounds good :100: 
and also I am guessing in this case the individual task debug spec get precedence over the global PipelineRun debug spec ? 

@bobcatfish @vdemeester I am not sure, but do we need a discussion around this(maybe in the next working group call) and update to the existing TEP to hash out the details before proceeding ? This level of change where taskRun debug is being enabled through pipelineRuns should be alright initially to let users enable taskRun debug through pipelineRuns. wdyt ?

jstrachan:
@waveywaves OK I've added a PipelineRun.Spec.Debug flag in addition to TaskRun specific structs; so you can debug all Tasks with a top level setting or add specific settings on a TaskRun instead

jstrachan:
@waveywaves the initial implementation of merge combines the breakpoint names from the `PipelineRun.Spec.Debug` and the `TaskRunSpec[i].Debug`. Would  you refer we just use the `TaskRunSpec[i].Debug` if its specified otherwise use the `PipelineRun.Spec.Debug`?

vdemeester:
> @bobcatfish @vdemeester I am not sure, but do we need a discussion around this(maybe in the next working group call) and update to the existing TEP to hash out the details before proceeding ? This level of change where taskRun debug is being enabled through pipelineRuns should be alright initially to let users enable taskRun debug through pipelineRuns. wdyt ?

Indeed, I think we need to update the TEP or do a follow-up TEP on PipelineRun support.

yelhouti:
@jerop as you might know, James is no longer working the project. Could anyone on the Teton team try to reimplemented this on top the TEP which now marked as implémentés ? This a great future to have.

wuhuizuo:
@jstrachan @jerop @waveywaves I think it can be reopened and then make some updates:

- The [TEP](https://github.com/tektoncd/community/blob/main/teps/0097-breakpoints-for-taskruns-and-pipelineruns.md#pipelinerun-proposal) is accepted.
- There a developing issue: https://github.com/tektoncd/pipeline/issues/5184, but it's closed by bot without completion.

🌹 🌹 🌹

PR #4146, opened 2021-08-11
`when` expressions in skipped `tasks` status

jerop:
# Changes

today, `when` expressions are included in skipped `tasks` regardless of
reason for skipping. for dag `tasks`, they are included even when they
are unresolved, while for finally tasks, they are included only when
they are resolved.

now that we have skipping reason, we want to check that the reason for
skipping is that `when` expressions evaluated to false before including
the resolved `when` expressions to the skipped `tasks` status.

we may explore adding the skipping reason to the skipped `tasks` status
in the future, but for now this change ensures that the status is
providing useful information to the users. it also brings consistency in
how we add `when` expressions to skipped `tasks` status in both
dag `tasks` and finally `tasks`.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -3941,11 +3941,6 @@ func TestReconcileWithWhenExpressionsScopedToTask(t *testing.T) {
 	}, {
 		// was attempted, but has missing results references
 		Name: "e-task",
-		WhenExpressions: v1beta1.WhenExpressions{{
-			Input:    "$(tasks.a-task.results.aResult)",
-			Operator: "in",
-			Values:   []string{"aResultValue"},
-		}},
 	}, {
 		Name: "f-task",
 	}}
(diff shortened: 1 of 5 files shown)


pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -3941,11 +3941,6 @@ func TestReconcileWithWhenExpressionsScopedToTask(t *testing.T) {
 	}, {
 		// was attempted, but has missing results references
 		Name: "e-task",
-		WhenExpressions: v1beta1.WhenExpressions{{
-			Input:    "$(tasks.a-task.results.aResult)",
-			Operator: "in",
-			Values:   []string{"aResultValue"},
-		}},
```
pritidesai, line 3948: Hey @jerop I think this was a nice feature to have it, `e-task` was skipped because the result reference `$(tasks.a-task.results.aResult)` is missing. After we add a reason to the skippedTask like you mentioned, it will be very obvious that we were not able to resolve a result reference. WDYT?

jerop: that's right, people could deduce from the unresolved reference that that was the reason for skipping

jerop: yea, maybe it makes sense to wait until we add the reason field before we make this cleanup

jerop:
closing this for now, will reopen it as we add the skipping reason to the status

PR #4147, opened 2021-08-11
(cleanup) use skipping reason instead of existence of variables

jerop:
# Changes

today, we use existence of variables to determine whether to add
`when` expressions to the skipped `tasks` status for finally tasks
(we only add them when they are resolved, have no variables)

we recently added skipping reason in https://github.com/tektoncd/pipeline/pull/4085

now that we have skipping reason, we can use it to check that the reason
for skipping is that `when` expressions evaluated to false before
including the resolved `when` expressions to the skipped `tasks` status

we plan to explore adding skipping reason to the skipped `tasks` status
in the future, but for now this change reuses this new functionality

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

/cc @pritidesai

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
@@ -423,7 +423,7 @@ func (facts *PipelineRunFacts) GetSkippedTasks() []v1beta1.SkippedTask {
 			}
 			// include the when expressions only when the finally task was skipped because
 			// its when expressions evaluated to false (not because results variables were missing)
-			if !rprt.PipelineTask.WhenExpressions.HaveVariables() {
+			if rprt.IsFinallySkipped(facts).SkippingReason == WhenExpressionsSkip {
 				skippedTask.WhenExpressions = rprt.PipelineTask.WhenExpressions
 			}
 			skipped = append(skipped, skippedTask)
--- a/pkg/apis/pipeline/v1beta1/when_types.go
+++ b/pkg/apis/pipeline/v1beta1/when_types.go
@@ -54,13 +54,6 @@ func (we *WhenExpression) isTrue() bool {
 	return !we.isInputInValues()
 }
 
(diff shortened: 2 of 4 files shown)


pritidesai (approved):

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -423,7 +423,7 @@ func (facts *PipelineRunFacts) GetSkippedTasks() []v1beta1.SkippedTask {
 			}
 			// include the when expressions only when the finally task was skipped because
 			// its when expressions evaluated to false (not because results variables were missing)
-			if !rprt.PipelineTask.WhenExpressions.HaveVariables() {
+			if rprt.IsFinallySkipped(facts).SkippingReason == WhenExpressionsSkip {
```
pritidesai, line 426: NIT: will be great if you can add a couple of unit tests for this change 🙏

vdemeester (approved):

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
jerop: added, thanks 😁

bobcatfish, line 426: whoa nice update, much clearer!!

(im a bit confused about why "has variables" is enough - e.g. what if hte when expression had variables but it was skipped for a different reason? anyway that doesnt matter now b/c you've fixed it XD)

bobcatfish (commented):
wow nice cleanup!! :D

/lgtm

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
jerop: it wasn't enough but a good proxy for many cases

PR #4151, opened 2021-08-12
Change "fail" to "stopAndFail" 🧐

bobcatfish:
# Changes

This commit proposes changes the wording of the `fail` option for
`onError` from `fail` to make it more explicitly clear that this option
(which is the default behavior when `onError` is not specified) will
both cause the Task to fail AND stop execution of any subsequent steps.

This change doesn't propose changing any of the described functionality,
just the value of the field.

Change to the TEP should be approved before this is merged
(https://github.com/tektoncd/community/pull/497) and if we go ahead with
it we should do a patch release ASAP to minimize user impact - that
being said, this change impacts the default value of the field which
users are unlikely to specify explicitly - users are much more likely to
start using `onError: continue` and for them there would be no impact.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
⚠️ Backwards incompatible change to new onError feature ⚠️ the value `fail` has been renamed to `stopAndFail` - if you are not using the `onError` field or you are using `onError: continue` you will not be impacted.
```

### DIFF
--- a/cmd/entrypoint/main.go
+++ b/cmd/entrypoint/main.go
@@ -44,7 +44,7 @@ var (
 	timeout             = flag.Duration("timeout", time.Duration(0), "If specified, sets timeout for step")
 	breakpointOnFailure = flag.Bool("breakpoint_on_failure", false, "If specified, expect steps to not skip on failure")
 	onError             = flag.String("on_error", "", "Set to \"continue\" to ignore an error and continue when a container terminates with a non-zero exit code."+
-		" Set to \"fail\" to declare a failure with a step error and stop executing the rest of the steps.")
+		" Set to \"stopAndFail\" to declare a failure with a step error and stop executing the rest of the steps.")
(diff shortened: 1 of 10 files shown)


bobcatfish:
Since this is technically a backwards incompatible change, [we need approval from > 50% of owners](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#backwards-incompatible-changes). PTAL @tektoncd/core-maintainers

sbwsg (approved):

dibyom (approved):
This definitely sounds clearer!

pritidesai (approved):
sounds good, thanks @bobcatfish

bobcatfish:
alternatively we could always call it:

image

😇

imjasonh (approved):

bobcatfish:
Thanks all!!

/hold cancel

PR #4154, opened 2021-08-13
cleaning up label keys 🧹

pritidesai:
# Changes

We have defined `GroupName` as a constant which is set to `tekon.dev` in `register.go`. We have other label keys defined in the same go file which are used by the controller to set the appropriate labels. These keys are `TaskLabelKey`, `PipelineLabelKey`, etc. These keys are set to `"/task"`, `"/pipeline"`, etc. Now while adding such labels in the Tekton
resources, we join `GroupName` to the constant label key. Instead, this commit changes the way we have defined a label key to include `GroupName` as part of the label key and use just the label key instead of concatenating `GroupName` everywhere. This cleans up the code and makes it simple to add tekton internal labels and at the same time retrieve them.

Most of the changes here are done using these two commands:

```
find . -type f -name "*.go" -print0 | xargs -0 sed -i '' -e 's/pipeline.GroupName+//g'
find . -type f -name "*.go" -print0 | xargs -0 sed -i '' -e 's/pipeline.GroupName +//g'
```

And I ran a sample pipeline to make sure the labels still have `tekton.dev` 🤣 

```
kubectl get tr  pipelinerun-print-hello-print-hello-ppbfh -o json | jq .metadata.labels
{
  "app.kubernetes.io/managed-by": "tekton-pipelines",
  "tekton.dev/memberOf": "tasks",
  "tekton.dev/pipeline": "pipeline-print-hello",
  "tekton.dev/pipelineRun": "pipelinerun-print-hello",
  "tekton.dev/pipelineTask": "print-hello",
  "tekton.dev/task": "print-hello"
}
```

This cleanup was motivated while resuming work on https://github.com/tektoncd/pipeline/pull/2891 😜 

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
The constants defined for the Tekton specific labels now include GroupName(tekton.dev) in the definition, for example:

Before this change:
const TaskLabelKey = "/task"

After this change:
const TaskLabelKey =  GroupName + "/task"

If you are importing these constants in your application from the pipeline source, you don't need to add GroupName anymore. 
```


sbwsg:
I think it might be worth including a release note for this. If anyone's maintaining their own code that imports and uses these constants from the pipelines codebase then they'll need to be aware of the change. 

/approve

Looks like the integration failure was just a flake.

vdemeester (approved):
/lgtm

Issue #4153, opened 2021-08-12
Update `publish-release` task to support release signing


sbwsg:
### Feature request

Sign the images that we publish during a release.

### More Info

Over in the Plumbing repo we now have [a documented guide to signing Tekton releases](https://github.com/tektoncd/plumbing/blob/main/docs/signing.md) in the dogfooding cluster.

In order to support signing we need to update our image publishing machinery to emit a result with name `IMAGES` and a given format. I believe this is the task we'll need to update: https://github.com/tektoncd/pipeline/blob/main/tekton/publish.yaml

priyawadhwa:
Awesome! Yup, `IMAGES` will work -- basically it expects a comma-separated list of images built in that TaskRun, qualified by digest.

The [chains publish.yaml](https://github.com/tektoncd/chains/blob/main/release/publish.yaml#L51) might be a helpful example.

PR #4156, opened 2021-08-13
Add metadata to our publish task for Tekton Chains to observe & sign

sbwsg:
Fixes https://github.com/tektoncd/pipeline/issues/4153

# Changes

This commit adds an annotation to indicate that build provenance should be
generated and an `IMAGES` result composed of a comma-separated list of
imageNames+digest to be signed.

This change is based on
https://github.com/tektoncd/chains/blob/main/release/publish.yaml and
https://github.com/tektoncd/plumbing/blob/main/docs/signing.md

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Tekton Pipelines releases are now signed by Tekton Chains.
```

### DIFF
--- a/tekton/publish.yaml
+++ b/tekton/publish.yaml
@@ -2,6 +2,8 @@ apiVersion: tekton.dev/v1beta1
 kind: Task
 metadata:
   name: publish-release
+  annotations:
+    chains.tekton.dev/transparency-upload: "true"
 spec:
   params:
     - name: package
@@ -52,6 +54,10 @@ spec:
         value: "$(params.imageRegistryRegions)"
       - name: OUTPUT_RELEASE_DIR
         value: "$(workspaces.output.path)/$(params.versionTag)"
+  results:
+  # IMAGES result is picked up by Tekton Chains to sign the release.
+  # See https://github.com/tektoncd/plumbing/blob/main/docs/signing.md for more info.
+  - name: IMAGES
   steps:
 
   - name: create-ko-yaml
@@ -177,6 +183,8 @@ spec:
         IMAGE_WITHOUT_SHA_AND_TAG=${IMAGE_WITHOUT_SHA%%:*}
         IMAGE_WITH_SHA=${IMAGE_WITHOUT_SHA_AND_TAG}@${IMAGE##*@}
 
+        echo $IMAGE_WITH_SHA, >> $(results.IMAGES.path)
+
         if [[ "$(params.releaseAsLatest)" == "true" ]]
         then
           crane cp ${IMAGE_WITH_SHA} ${IMAGE_WITHOUT_SHA_AND_TAG}:latest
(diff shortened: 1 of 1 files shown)


tekton/publish.yaml
```diff
@@ -177,6 +183,8 @@ spec:
         IMAGE_WITHOUT_SHA_AND_TAG=${IMAGE_WITHOUT_SHA%%:*}
         IMAGE_WITH_SHA=${IMAGE_WITHOUT_SHA_AND_TAG}@${IMAGE##*@}
 
+        echo $IMAGE_WITH_SHA, >> $(results.IMAGES.path)
```
bobcatfish, line 186: this is probably a silly question but will a trailing comma be okay? im guessing yes and that the effort to remove it isn't worth it XD

bobcatfish: ps do you know if the crane cp commands below will also produce more unique shas?

bobcatfish:
VERY EXCITING!!!!

/approve

tekton/publish.yaml
sbwsg: Yeah I think so - this is taken verbatim from chains own publish script: https://github.com/tektoncd/chains/blob/main/release/publish.yaml

sbwsg: /cc @priyawadhwa

tekton/publish.yaml
```diff
@@ -1,7 +1,9 @@
-apiVersion: tekton.dev/v1beta1
+gpiVersion: tekton.dev/v1beta1
```
pritidesai, line 1: typo? 🤦‍♀️ 🤣

sbwsg: Oh.  Wow.  Vim.

tekton/publish.yaml
sbwsg: > ps do you know if the crane cp commands below will also produce more unique shas?

Oo, this I don't know much about unfortunately.

tekton/publish.yaml
sbwsg: Good catch, fixed, thanks!

tekton/publish.yaml
priyawadhwa: trailing comma should be fine!

& the`crane cp` command should result in the same digest!

tekton/publish.yaml
```diff
@@ -177,6 +183,8 @@ spec:
         IMAGE_WITHOUT_SHA_AND_TAG=${IMAGE_WITHOUT_SHA%%:*}
         IMAGE_WITH_SHA=${IMAGE_WITHOUT_SHA_AND_TAG}@${IMAGE##*@}
 
+        echo $IMAGE_WITH_SHA, >> $(results.IMAGES.path)
+
         if [[ "$(params.releaseAsLatest)" == "true" ]]
         then
           crane cp ${IMAGE_WITH_SHA} ${IMAGE_WITHOUT_SHA_AND_TAG}:latest
```
pritidesai, line 190: Comparing this to Chains publish task, we are missing REGION specific sha in the result here

https://github.com/tektoncd/chains/blob/main/release/publish.yaml#L184

Do we need to add one more `echo` command here?

priyawadhwa: I'd recommend adding that in so that all images are signed

sbwsg: Amazing, thanks again - I didn't copy verbatim enough 😆 

Added the additional region `echo`

pritidesai:
thanks @sbwsg 
/lgtm

PR #4157, opened 2021-08-13
Cherry pick onError:fail rename and release signing metadata in to 0.27.x branch

sbwsg:
# Changes

This PR brings in two commits for the release of 0.27.1:

- [Change "fail" to "stopAndFail"](https://github.com/tektoncd/pipeline/pull/4151)
- [Add metadata to our publish task for Tekton Chains to observe and sign](https://github.com/tektoncd/pipeline/pull/4156)

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/cmd/entrypoint/main.go
+++ b/cmd/entrypoint/main.go
@@ -44,7 +44,7 @@ var (
 	timeout             = flag.Duration("timeout", time.Duration(0), "If specified, sets timeout for step")
 	breakpointOnFailure = flag.Bool("breakpoint_on_failure", false, "If specified, expect steps to not skip on failure")
 	onError             = flag.String("on_error", "", "Set to \"continue\" to ignore an error and continue when a container terminates with a non-zero exit code."+
-		" Set to \"fail\" to declare a failure with a step error and stop executing the rest of the steps.")
+		" Set to \"stopAndFail\" to declare a failure with a step error and stop executing the rest of the steps.")
 	stepMetadataDir     = flag.String("step_metadata_dir", "", "If specified, create directory to store the step metadata e.g. /tekton/steps/<step-name>/")
 	stepMetadataDirLink = flag.String("step_metadata_dir_link", "", "creates a symbolic link to the specified step_metadata_dir e.g. /tekton/steps/<step-index>/")
 )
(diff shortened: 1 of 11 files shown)


pritidesai:
@sbwsg I am wondering if introducing signed releases in a patch would cause any conflict with a `.0` (point) release? 🤔  As in, when the release is applies using:

```
kubectl apply -f https://storage.googleapis.com/tekton-releases/pipeline/previous/v0.27.1/release.yaml
```

vs

```
kubectl apply -f https://storage.googleapis.com/tekton-releases/pipeline/previous/v0.27.0/release.yaml
```

pritidesai:
I understand we are just singing the images but still wanted to bring it up ...

sbwsg:
> @sbwsg I am wondering if introducing signed releases in a patch would cause any conflict with a .0 (point) release? 🤔

I don't think the signing process modifies the deployed resources in any way.  It allows users to verify the images that they're installing with the Pipelines YAML but doesn't require that YAML to be any different to that of previous releases.

sbwsg:
I can test this, though, once I've made the release. If something is terribly wrong then I can immediately roll back the publish yaml in the cluster and perform another release. wdyt?

pritidesai:
I was more concerned about the image these files are pointing to:

```
 gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/controller:v0.27.0@sha256:015e8f9cf38416652545ce5e16776379a979af64373250ec489861886a468fcf
```

Not signed vs Signed?  0.27.1 will generate a new image which will be signed where 0.27.0 is pointing to an image which is not. How do the users consume the new releases? There is a very small window between a point release and a patch in this case so should be fine.

/approve

bobcatfish:
> Not signed vs Signed? 0.27.1 will generate a new image which will be signed where 0.27.0 is pointing to an image which is not. How do the users consume the new releases? 

my understanding is that users consuming this signed release won't actually notice any difference, unless they want to start verifying the signature (which is something i think in the long run we'll want to start recommending and/or requiring as part of installation - maybe the operator can enforce it?)

but maybe you're imagining a different scenario @pritidesai ? in the meantime it looks like you're okay with going ahead with this so adding my sign off as well:

/lgtm

Issue #3903, opened 2021-04-28
Starting PipelineRun in Pending Or Cancelled State Doesn't Add Pipeline Label


andrewballantyne:
# Expected Behavior

When I create a PipelineRun against a `pipelineRef.name` to an existing Pipeline I have on my OpenShift cluster, a label is added for me `tekton.dev/pipeline = <pipelineName>`. I'd expect this to happen for `spec.status` of `PipelineRunPending` and `PipelineRunCancelled` as well, when they are included.

# Actual Behavior

However, when I create the same PipelineRun with the same configuration (pipelineRef, name, labels, etc) but with an added `spec.status` to either `PipelineRunCancelled` or `PipelineRunPending`, it starts and omits the label I used to find all related PipelineRuns to my Pipeline. I'm no longer able to find related PipelineRuns, especially those in pending state to start them.

# Steps to Reproduce the Problem

1. Create any PipelineRun (Pipeline spec or ref) that has no labels and see the controller/operator adds them

2. Repeat the process but add `status: PipelineRunCancelled` or `status: PipelineRunPending` to the spec and push that PipelineRun

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

```
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.0-rc.0+6143dea", GitCommit:"6143dea8e6a5046b467a67f7d1fda8e63833a2e7", GitTreeState:"clean", BuildDate:"2021-04-25T16:18:28Z", GoVersion:"go1.16.1", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

```
OpenShift Pipelines Operator 1.4.0
Pipeline version: v0.22.0
Triggers version: v0.12.1
```

jerop:
Thank you for the issue report @andrewballantyne!

Fixing it in https://github.com/tektoncd/pipeline/pull/4163

PR #4163, opened 2021-08-16
Propagate Pipeline name label to PipelineRun

jerop:
# Changes

Currently, a `PipelineRun` started in a `Cancelled` or `Pending` state doesn't receive the `Pipeline` name label. 

As such, a user cannot find all the `PipelineRuns` related to a given `Pipeline`. This is a challenge for a user searching for associated `PipelineRuns` in `Pending` state in order to start them.

In this change, we ensure that the `Pipeline` name label is propagated from the `Pipeline` to the `PipelineRun`.  

Fixes https://github.com/tektoncd/pipeline/issues/3903.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
labels and annotations are propagated from pipeline to pipelinerun, even when the pipelinerun is pending or cancelled
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -206,6 +206,11 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 		return c.finishReconcileUpdateEmitEvents(ctx, pr, before, nil)
 	}
 
+	if err := propagatePipelineNameLabelToPipelineRun(pr); err != nil {
+		logger.Errorf("Failed to propagate pipeline name label to pipelinerun %s: %v", pr.Name, err)
+		return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
+	}
+
 	// If the pipelinerun is cancelled, cancel tasks and update status
 	if pr.IsCancelled() {
 		err := cancelPipelineRun(ctx, logger, pr, c.PipelineClientSet)
@@ -929,6 +934,21 @@ func getTaskrunAnnotations(pr *v1beta1.PipelineRun) map[string]string {
 		annotations[key] = val
 	}
 	return annotations
+}
+
+func propagatePipelineNameLabelToPipelineRun(pr *v1beta1.PipelineRun) error {
+	if pr.ObjectMeta.Labels == nil {
+		pr.ObjectMeta.Labels = make(map[string]string)
+	}
+	switch {
+	case pr.Spec.PipelineRef != nil && pr.Spec.PipelineRef.Name != "":
+		pr.ObjectMeta.Labels[pipeline.PipelineLabelKey] = pr.Spec.PipelineRef.Name
+	case pr.Spec.PipelineSpec != nil:
+		pr.ObjectMeta.Labels[pipeline.PipelineLabelKey] = pr.Name
+	default:
+		return fmt.Errorf("pipelineRun %s not providing PipelineRef or PipelineSpec", pr.Name)
+	}
+	return nil
 }
 
 func getTaskrunLabels(pr *v1beta1.PipelineRun, pipelineTaskName string, includePipelineLabels bool) map[string]string {
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -2557,9 +2557,14 @@ func TestReconcileCancelledFailsTaskRunCancellation(t *testing.T) {
 			tb.PipelineRunStartTime(time.Now()),
 		),
 	)}
+	ps := []*v1beta1.Pipeline{tb.Pipeline("test-pipeline", tb.PipelineNamespace("foo"), tb.PipelineSpec(
+		tb.PipelineTask("hello-world-1", "hello-world"),
+		tb.PipelineTask("hello-world-2", "hello-world"),
+	))}
 
 	d := test.Data{
 		PipelineRuns: prs,
+		Pipelines:    ps,
 	}
 
 	testAssets, cancel := getPipelineRunController(t, d)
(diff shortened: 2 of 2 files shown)


pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -320,6 +314,28 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
...
+
+	// If the pipelinerun is cancelled, cancel tasks and update status
+	if pr.IsCancelled() {
+		err := cancelPipelineRun(ctx, logger, pr, c.PipelineClientSet)
+		if err != nil {
+			logger.Errorf("Failed to cancel pipelinerun %s: %v", pr.Name, err)
+			return controller.NewPermanentError(err)
+		}
+		return nil
+	}
```
vdemeester, line 337: It seems a bit weird to move this that *late* in the reconciler loop just for being able to put labels on it. Any reason for not getting only the pipeline meta before reconcile and do the "propagateLabels" earlier instead ?

jerop: makes sense, moved the propagation of labels and annotations earlier in the reconcile loop instead of moving cancellation later

vdemeester (approved):
/cc @sbwsg

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -330,37 +342,11 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
...
-	if err != nil {
-		logger.Errorf("Failed to determine Pipeline spec to use for pipelinerun %s: %v", pr.Name, err)
-		pr.Status.MarkFailed(ReasonCouldntGetPipeline,
-			"Error retrieving pipeline for pipelinerun %s/%s: %s",
-			pr.Namespace, pr.Name, err)
-		return controller.NewPermanentError(err)
-	}
-
 	// Store the fetched PipelineSpec on the PipelineRun for auditing
 	if err := storePipelineSpec(ctx, pr, pipelineSpec); err != nil {
```
sbwsg, line 346: I think it might be a good idea to move this up along with `GetPipelineData`. Otherwise a Pending PipelineRun will end up fetching the data multiple times from the cluster / bundle: first while it's in Pending state and then again when it's moved out from Pending and reaches here and stores the spec in the status. wdyt?

sbwsg:
Thinking about this some I'm a bit torn about this change. The pending state of PipelineRuns was [originally introduced](https://github.com/tektoncd/community/pull/203/files) to limit burden on busy clusters and allow (per the TEP) "a custom controller to implement any `PipelineRun` scheduling policy that is transparent to the `PipelineRun` controller and `PipelineRun` creator".

So I'm wondering two things:

1. Are we maybe going to end up re-introducing burden on the cluster here because even in Pending state we are now hitting the cluster to resolve a `pipelineRef`?
2. Do we consider this a breaking change to a non-alpha feature?  Previously Pending meant that resolution wasn't going to happen, but now it will.  Should we be concerned about that?

I might be overthinking this though and it's fine.  WDYT @jerop?

Pinging @tektoncd/core-maintainers here to get others' input too.

jerop:
@sbwsg thanks for the additional context from the TEP!

> 1. Are we maybe going to end up re-introducing burden on the cluster here because even in Pending state we are now hitting the cluster to resolve a `pipelineRef`?

if users are having trouble starting the pending pipelineruns for a given pipeline because of the missing pipeline label, maybe the burden to resolve that may be necessary? it would not be as much load as actually starting up the pipelinerun, so I wonder if that hit is acceptable?

it makes me also wonder how common this operation is, i.e. do many users need to start pending pipelineruns for a specific pipeline? or do they start pending pipelineruns based on other criteria? 

cc @jbarrick-mesosphere @andrewballantyne

> 2. Do we consider this a breaking change to a non-alpha feature?  Previously Pending meant that resolution wasn't going to happen, but now it will.  Should we be concerned about that?

good point, i'm not sure about this 🤔 

if we consider it a breaking change and still want to make this change, does this mean this will be behind its own feature flag (not the general feature flag), right?

pkg/reconciler/pipelinerun/pipelinerun.go
jerop: makes sense, moved it up, thanks!

sbwsg:
Yeah I can definitely get on board with the idea that this isn't _so much_ additional overhead that it's problematic. 👍 

I think you're right about the flagging - if we want to make a backwards-incompatible change to behaviour it needs its own feature flag. 

Given that we don't have usage data it's tricky to determine whether anyone's actually relying on this specific order of operations. That fact alone makes me lean towards treating it as a backwards-incompatible change - in the absence of data I tend towards assuming we'd break someone / anyone, just to be cautious - but I don't feel super strongly about it, esp since we're in beta.

This might be a good topic to raise in the API WG to get a general consensus on whether folks agree or not that it's backwards-incomatible?

andrewballantyne:
Just to add my two cents in here as I logged #3903. 

Inside OpenShift Console using the OpenShift Pipelines Operator (which packages Tekton for us), we have a details/list page that presents a list to the user of all Pipeline Runs that are associated with the current selected Pipeline. The problem here is we use the pipeline label to do that lookup. So the Pending Pipelines currently do not display in that list despite having a reference to the Pipeline within' them. It would be nice to have that label in order to allow for the list to present for Pending PipelineRuns; alongside the existing currently running and past ran ones.

The only coding alternative I'm aware of is to fetch all Pipeline Runs in the namespace and then to filter them down based on the pipelineRef. This has an obvious undesirable weight.

The reason this was logged was because we added a feature to start the paused Pipelines, but finding them is not as easy due to the aforementioned issues. Cancelled Pipelines was lumped into this because it is also missing that label but there isn't any real action for the user.

The how, cost and implementation details I'll obviously leave up to you guys. There are a few work arounds the user can do to find their Pipeline that is Paused or Cancelled, it'll just not be straight forward.

sbwsg:
Ahhh I think we can set that one specific label without performing any resolution / changing order of operations here. We know the pipeline's name as soon as we have the `PipelineRun`.  It's either:

1. The `spec.pipelineRef.name`
2. The name of the PipelineRun ([if `spec.pipelineSpec` is used instead of `pipelineRef`](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/resources/pipelinespec.go#L45-L46))

jerop:
> Ahhh I think we can set that one specific label without performing any resolution / changing order of operations here. We know the pipeline's name as soon as we have the `PipelineRun`. It's either:
> 
> 1. The `spec.pipelineRef.name`
> 2. The name of the PipelineRun ([if `spec.pipelineSpec` is used instead of `pipelineRef`](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/resources/pipelinespec.go#L45-L46))

great idea @sbwsg - updated the PR to only add the pipeline name label - thanks!

sbwsg:
Perfect! Possibly a good candidate to bring in to v0.27.2 as well? I will do that tomorrow.

/lgtm
/add-label needs-cherry-pick

jerop:
> Perfect! Possibly a good candidate to bring in to v0.27.2 as well? I will do that tomorrow.

yes yes, thank you @sbwsg 😁

andrewballantyne:
I appreciate all the help and thanks for getting this done @jerop

PR #4164, opened 2021-08-17
Patch vendor/ apimachinery to work on 1.22

### LINKED ISSUE #4158: Pipelinerun doesn't show status on k8s 1.22
# Expected Behavior
pipeline runs successfully

# Actual Behavior
pipelinrun doesn't show any status 

1. Install pipeline on k8s 1.22
2. install any pipeline and create a pipelinerun
3. check the status of pipelinerun

# Additional Info

Controller shows following error
***
{"level":"error","ts":"2021-08-16T10:19:47.673Z","logger":"tekton-pipelines-controller","caller":"controller/controller.go:564","msg":"Reconcile error","commit":"42a632f","knative.dev/controller":"github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.Reconciler","knative.dev/kind":"tekton.dev.PipelineRun","duration":0.017707999,"error":"admission webhook \"webhook.pipeline.tekton.dev\" denied the request: mutation failed: cannot decode incoming new object: json: unknown field \"subresource\"","stacktrace":"github.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller.(*Impl).handleErr\n\tgithub.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller/controller.go:564\ngithub.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller.(*Impl).processNextWorkItem\n\tgithub.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller/controller.go:541\ngithub.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller.(*Impl).RunContext.func3\n\tgithub.com/tektoncd/pipeline/vendor/knative.dev/pkg/controller/controller.go:477"}
***
This is probably an issue from knative/pkg, a similar issue was created https://github.com/knative/serving/issues/11448

- Kubernetes version:

```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.5", GitCommit:"6b1d87acf3c8253c123756b9e61dac642678305f", GitTreeState:"archive", BuildDate:"2021-03-30T00:00:00Z", GoVersion:"go1.16", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.0", GitCommit:"c2b5237ccd9c0f1d600d3072634ca66cefdf272f", GitTreeState:"clean", BuildDate:"2021-08-04T20:01:24Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"linux/amd64"}
```
- Tekton Pipeline version:

```
Pipeline version: v0.27.0
```

vdemeester:
# Changes

A new field is there in k8s 1.22, name subresource in the
managedfields. This proves to make knative based types to not be valid
on 1.22 and above, making tektoncd/pipeline as well as any other
component using knative/pkg to be broken on 1.22 and above.

Fixes #4158

/kind bug

Opening this PR to discuss what is the way forward, taking the following into account:
- Any version of tektoncd/pipeline (or any other component using knative/pkg) won't work on kubernetes 1.22 and above (without this patch)
- Adding this field will probably be backported to 1.21 and maybe even 1.22 apimachinery branches — so updating those deps would fix it, but the question is the timeframe
- What should we do for previous releases ? (0.27, 0.26, …) A bigger question is, what is our support *matrix* for pipeline (should release X support the latest k8s released at that time or not ? …) ?
- Assuming we would want this patch in (here or on release branches), what would be our "patch" management ?

# Submitter Checklist

  functionality, content, code)

# Release Notes

TBD


sbwsg:
I'm tempted to suggest that we only patch 0.27 and 0.26 and _maybe_ 0.25, declaring anything less as incompatible with Kubernetes 1.22+. Whatever we decide I'm happy to perform the patches on the older release branches since I'll be doing it for 0.27 anyway.

What are you thinking with "patch management"? A shell script to apply the change with `awk` / `sed`?  Or something totally different? I would've thought just having the change in our `vendor` tree would be enough but I might be misunderstanding our options here.

vdemeester:
> I'm tempted to suggest that we only patch 0.27 and 0.26 and _maybe_ 0.25, declaring anything less as incompatible with Kubernetes 1.22+. Whatever we decide I'm happy to perform the patches on the older release branches since I'll be doing it for 0.27 anyway. We might also want to apply docs-only patches to the older releases so that users can be informed of the platform incompatibility?

Right, I would have gone for 0.27 and 0.26 only initially, and document a larger set of others that there is an incompat. But there is one thing we need to figure out / document : "support" lifecycle, aka what is the lifecycle of a given release, when should we patch it, … I don't think we have this documented anywhere yet.

> What are you thinking with "patch management"? A shell script to apply the change with `awk` / `sed`? Or something totally different? I would've thought just having the change in our `vendor` tree would be enough but I might be misunderstanding our options here.

Downstream (OpenShift Pipelines), we are using a folder (`openshift/patches`) and `git format-patch` and `git am` 🙃. Having the change only on the `vendor` folder makes it non-reproducible (or at least harder). Imaging we need to another fix in 0.27 with bumping a new dependency (fixing a CVE or something) ; if we don't have a patch to apply or something, there is no guarantee that our `vendor` patch that we did by hand well still be there.
The job `pull-tekton-pipeline-build-tests` fails for this exact reason today, I just made the change without making it *reproducible* when running `./hack/update*` scripts.

vdemeester:
An alternative would be to do what knative did [here](https://github.com/knative/serving/pull/11848), allowing unknown fields and get the validation through openapi schemas. But this needs https://github.com/tektoncd/pipeline/compare/main...vdemeester:schema-gen and as far as I tried, this only works with one version of CRD (trying this with multiple version and... it panics)

sbwsg:
Using `git` for the patching makes total sense to me given it's a tool already in our workflow.

/approve

sbwsg:
ping @tektoncd/core-maintainers it would be good to get input from one other person before we move ahead with this. We'd like to release as part of 0.27.2 so PTAL.

vdemeester:
I'll update the `./hack/update-deps.sh` with this patch to apply and a warning that this is temporary, linking to an issue I'll create to track the resolution of this in k8s or knative/pkg (or naturally with dep bumps).

vdemeester:
> please add some release notes

Oh indeed..

PR #4108, opened 2021-07-20
WIP Refactor ref resolution

sbwsg:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/3305

This PR refactors the way task and pipelinerun resolution works. The intention with this PR is to be entirely backwards-compatible with the existing way resolution works. In other words, I'm not trying to introduce new behaviours with these commits, just set the stage for future possible improvements.

I'm pushing this up now to start getting feedback from interested parties. I've still got a ways to go in terms of getting new unit tests updated and introduced. I'm hopeful that integration tests shouldn't be too badly affected. 🤞 

There are two commits as part of this PR to aid reviewers:

1. Remove task and pipeline spec resolution from the existing reconcilers
2. Move ref resolution to new reconcilers

I'll squash these once we're happy to move ahead.

Prior to this commit the taskrun and pipelinerun reconcilers were responsible for resolving specs from taskRef / taskSpec / pipelineRef / pipelineSpec / bundles.

The first commit in this PR updates the existing taskrun/pipelinerun reconcilers to ignore any taskrun or pipelinerun that doesn't have a spec cached in their status. This means all taskruns and pipelineruns are ignored - the logic for populating status from a spec has been removed entirely from the controller.

The second commit in this PR introduces 2 new controllers whose sole responsibility is to resolve specs from taskruns and pipelineruns. Resolved specs then get written to the status field of the runs and the "old" reconcilers are able to pick them up and begin actuating them as usual.

The new resolver reconcilers are also currently responsible for propagating labels and annotations from the task / pipeline to the taskrun / pipelinerun. This is because our existing taskrun/pipelinerun reconcilers expect only the _spec_ to be embedded in their status fields, not the entire task/pipeline objects. The knock-on effect of this is that the metadata from the resolved task/pipeline is lost by the time the taskrun/pipelinerun reconciler receive them.

One final thing this commit introduces is a library, `internal/resolution` that may eventually be available for downstream projects to implement their own task and pipeline resolvers. At the moment all it does is expose the default resolution behaviour of Tekton Pipelines as a func + a couple of useful helper methods for propagating labels/annotations. This would allow someone implementing their own resolver reconciler to generally adhere to the Tekton Pipelines "resolution contract" with only 100 lines or so of Go. It's not yet intended for external consumption but I think it's an interesting direction we can take future work on tekton's built-in resolution code and provides a nice enough separation of concerns in our own codebase.

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

[...]

PR #4168, opened 2021-08-17
Add controller flag to turn off built-in resolution

sbwsg:
# Changes

We're currently trying to work out what the best way forward for [TEP-0060](https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md)
(remote resolution) is for Tekton Pipelines. As part of that we're
creating a controller in the experimental repo that we can use to test out
a bunch of the Alternatives from the TEP and figure out what works best
in terms of implementation and DX.

Right now the experimental controller would race with the Pipelines controllers
since they're all going to try and "resolve" any taskRef in Taskruns and
PipelineRuns.

This commit adds a flag to the pipelines controller that explicitly switches
off resolution performed by the taskrun and pipelinerun reconcilers. This gives
us just enough room to perform resolution out-of-band and try out some of the
alternatives.

When the `-experimental-disable-in-tree-resolution` flag is set the taskrun
and pipelinerun controllers will ignore taskruns and pipelineruns that don't
have `status.taskSpec` or `status.pipelineSpec` populated.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Introduce experimental flag to disable built-in resolution behaviour of the taskrun and pipelinerun reconcilers.
```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -55,6 +55,10 @@ var (
 	disableHighAvailability  = flag.Bool("disable-ha", false, "Whether to disable high-availability functionality for this component.  This flag will be deprecated "+
 		"and removed when we have promoted this feature to stable, so do not pass it without filing an "+
 		"issue upstream!")
+	experimentalDisableInTreeResolution = flag.Bool(disableInTreeResolutionFlag, false,
+		"Disable resolution of taskrun and pipelinerun refs by the taskrun and pipelinerun reconcilers.")
+
+	disableInTreeResolutionFlag = "experimental-disable-in-tree-resolution"
 )
 
 func main() {
@@ -108,10 +112,20 @@ func main() {
 		log.Fatal(http.ListenAndServe(":"+port, mux))
 	}()
 
+	taskrunControllerConfig := taskrun.ControllerConfiguration{
+		Images:                   images,
+		DisableTaskRefResolution: *experimentalDisableInTreeResolution,
+	}
+
+	pipelinerunControllerConfig := pipelinerun.ControllerConfiguration{
+		Images:                       images,
+		DisablePipelineRefResolution: *experimentalDisableInTreeResolution,
+	}
+
 	ctx = filteredinformerfactory.WithSelectors(ctx, v1beta1.ManagedByLabelKey)
 	sharedmain.MainWithConfig(ctx, ControllerLogKey, cfg,
-		taskrun.NewController(*namespace, images),
-		pipelinerun.NewController(*namespace, images),
+		taskrun.NewController(*namespace, taskrunControllerConfig),
+		pipelinerun.NewController(*namespace, pipelinerunControllerConfig),
 	)
 }
 
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -132,11 +132,21 @@ type Reconciler struct {
 	cloudEventClient  cloudevent.CEClient
 	metrics           *pipelinerunmetrics.Recorder
 	pvcHandler        volumeclaim.PvcHandler
+
+	// disableResolution is a flag to the reconciler that it should
+	// not be performing resolution of pipelineRefs.
+	// TODO(sbwsg): Once we've agreed on a way forward for TEP-0060
+	// this can be removed in favor of whatever that chosen solution
+	// is.
+	disableResolution bool
 }
 
 var (
 	// Check that our Reconciler implements pipelinerunreconciler.Interface
 	_ pipelinerunreconciler.Interface = (*Reconciler)(nil)
+
+	// Indicates pipelinerun resolution hasn't occurred yet.
+	errResourceNotResolved = fmt.Errorf("pipeline ref has not been resolved")
 )
 
 // ReconcileKind compares the actual state with the desired, and attempts to
@@ -230,6 +240,13 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 	// updates regardless of whether the reconciliation errored out.
 	if err = c.reconcile(ctx, pr, getPipelineFunc); err != nil {
 		logger.Errorf("Reconcile error: %v", err.Error())
+	}
+
+	if c.disableResolution && err == errResourceNotResolved {
+		// This is not an error: an out-of-band process can
+		// still resolve the PipelineRun, at which point
+		// reconciliation can continue as normal.
+		err = nil
 	}
 
 	if err = c.finishReconcileUpdateEmitEvents(ctx, pr, before, err); err != nil {
@@ -334,6 +351,10 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
 			Message: fmt.Sprintf("PipelineRun %q is pending", pr.Name),
 		})
 		return nil
+	}
+
+	if c.disableResolution && pr.Status.PipelineSpec == nil {
+		return errResourceNotResolved
 	}
 
 	pipelineMeta, pipelineSpec, err := resources.GetPipelineData(ctx, pr, getPipelineFunc)
--- a/config/controller.yaml
+++ b/config/controller.yaml
@@ -80,7 +80,10 @@ spec:
           "-shell-image", "gcr.io/distroless/base@sha256:aa4fd987555ea10e1a4ec8765da8158b5ffdfef1e72da512c7ede509bc9966c4",
           # for script mode to work with windows we need a powershell image
           # pinning to nanoserver tag as of July 15 2021
-          "-shell-image-win", "mcr.microsoft.com/powershell:nanoserver@sha256:b6d5ff841b78bdf2dfed7550000fd4f3437385b8fa686ec0f010be24777654d6"
+          "-shell-image-win", "mcr.microsoft.com/powershell:nanoserver@sha256:b6d5ff841b78bdf2dfed7550000fd4f3437385b8fa686ec0f010be24777654d6",
+          # Experimental. Uncomment below to disable TaskRun and PipelineRun
+          # reconcilers' built-in taskRef and pipelineRef resolution procedures.
+          # "-experimental-disable-in-tree-resolution",
         ]
         volumeMounts:
         - name: config-logging
--- a/pkg/reconciler/pipelinerun/controller.go
+++ b/pkg/reconciler/pipelinerun/controller.go
@@ -42,8 +42,18 @@ import (
 	"knative.dev/pkg/logging"
 )
 
+// ControllerConfiguration holds fields used to configure the
+// PipelineRun controller.
+type ControllerConfiguration struct {
+	// Images are the image references used across Tekton Pipelines.
+	Images pipeline.Images
+	// DisablePipelineRefResolution tells the controller not to perform
+	// resolution of pipeline refs from the cluster or bundles.
+	DisablePipelineRefResolution bool
+}
+
 // NewController instantiates a new controller.Impl from knative.dev/pkg/controller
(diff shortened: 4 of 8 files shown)


sbwsg:
I originally tried to refactor the existing reconcilers to depend entirely on these new resolvers in #4108 but doing so was far too invasive and the PR was going to be just ridiculously huge. The main mistake I made was trying to give the resolver reconcilers total ownership of the resources before the taskrun/pipelinerun reconcilers saw them. I did this by using a filtering handler to effectively "hide" unresolved TaskRuns/PipelineRuns from the taskrun/pipelinerun reconcilers. But taking that approach meant duplicating massive chunks of the existing TaskRun and PipelineRun reconcilers in order to account for every single feature that occurs before or during resolution. For example: metrics, event emitting, cancellation handling, pending logic, timeouts, error messaging etc etc etc. All can interact with or be influenced by the resolution process in one way or another, so had to be kept in-tact when the resolver reconcilers were in control.

So instead I'm taking a different tack:

First I introduce the new resolution reconcilers to Pipelines behind an experimental controller arg and add a feature flag which disables the taskrun/pipelinerun resolution process. This means users who aren't interested in trying this stuff out won't be burdened by the changes.

Second is a different architectural approach: now the taskrun and pipelinerun reconcilers don't ignore unresolved resources - they're processed as normal.  Instead they simply stop reconciling them if they recognize that resolution hasn't happened yet. This means they still do stuff like setting and handling timeouts / cancellation / pending states. And the error messaging and events emitted remain entirely the same as before. The only difference is that resolution can now happen "out of band" and they'll wait for this to occur before they try and build DAGs or spin up pods. Seems to work pretty well 👍

sbwsg:
/hold

Let's figure out if there's agreement on https://github.com/tektoncd/community/pull/493 before this gets reviewed or merged.

sbwsg:
That darn `workspace-in-sidecar` example.  I need to revisit that one, it's super flakey.

docs/developers/README.md
```diff
@@ -627,4 +627,37 @@ cat /tekton/steps/step-unnamed-<step-index>/exitCode
...
+
+To disable remote resolution in the TaskRun and PipelineRun reconcilers, set
+the
+[experimental-disable-ref-resolution](../../config/config-feature-flags.yaml)
+feature flag to `"true"`.
+
+To enable the experimental resolution reconcilers add
+`-experimental-enable-resolution-reconcilers` as an additional `arg` passed to
+the `tekton-pipelines-controller` in its [Deployment
+YAML](../../config/controller.yaml).
```
bobcatfish, line 654: this is probably a silly question but im wondering where the line is b/w command line args and configuration we configure via config maps - could it make sense that both of these options are configurable via config map?

(it seems like [we're mostly just using args for images](https://github.com/tektoncd/pipeline/blob/8e1006971158a23750388220cfe6e16aa69f1e21/config/controller.yaml#L63-L80) - not really 100% sure why those wouldn't be configurable via config map also)

bobcatfish: oh the code is right above - also threads-per-controller, disable-ha, namespace - maybe the idea is these CAN'T change at runtime? (unless they can)

vdemeester: I would say yes, can't change at runtime, and need a very very explicit change from the user

sbwsg: Two considerations:

1. There's only one spot in our codebase where we can "spin up" reconcilers. That's in [`cmd/controller/main.go`](https://github.com/tektoncd/pipeline/blob/8e1006971158a23750388220cfe6e16aa69f1e21/cmd/controller/main.go#L110-L113). At this early stage in the program's startup we don't have access to configmap values yet.
2. The reconcile loops (`ReconcileKind()`) of our reconcilers don't have direct access to the command line flags passed to the application. We have no existing avenue to pass those flags from the controller's `main.go` down to the reconcilers.

I'm happy to create a way for passing the values from CLI args down to reconcilers - is that what we're discussing here?  I don't think we can spin up / tear down the resolution reconcilers via runtime configmap changes tho.

sbwsg: I realise I didn't really address the original point - where the line is b/w command line args and configuration we configure via config maps.  I didn't approach this with a clear line in mind. It was just easiest to plumb the configuration through from both of these sides to their respective points in the application. Open to changes.

sbwsg:
Added to milestone 0.28 🤞

docs/developers/README.md
bobcatfish: no problem at all, thanks for the detailed explanation @sbwsg  - it sounds like the line (at least for now) is configuration that impacts how the reconcile loops are started maybe? config for that needs to be passed via command line args, and the rest can be config map

so long story short sounds like you made the right choice to configure this via command line arg

my only remaining question is if we want both options (both the command line arg and the config map param) - i think im not focusing on the most important parts of this proposal so feel free to just make a choice and move on, but my thought is we could initially just have one way of configuring this (maybe the command line option) which both enables the experimental reconciler and disables the current resolution (esp since if this becomes our long term solution im guessing we'd rename or remove `experimental-disable-ref-resolution` anyway) - but also ignore me if that's just more work to make happen than its worth

bobcatfish: nevermind, my bad, i just thought this through and realized of course you want 2 flags because you want to be able to:

- Turn off resolution and use the new reconcilers to resolve
- Turn off resolution but use some different reconciler to resolve (e.g. git)

👍 👍 👍  to your approach

pkg/reconciler/resolver/taskrun_resolver_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+	pipelineClientSet clientset.Interface
+	taskrunLister     listers.TaskRunLister
+	configStore       reconciler.ConfigStore
+}
+
+const defaultTaskRunResolutionErrorReason = resolution.ReasonTaskRunResolutionFailed
+
+var _ controller.Reconciler = &TaskRunResolverReconciler{}
+
+func (r *TaskRunResolverReconciler) Reconcile(ctx context.Context, key string) error {
```
bobcatfish, line 48: maybe this is happening somewhere else but im wondering if these new reconcilers should check if `experimentalDisableRefResolution` is set before they do anything?

im definitely over thinking it but im wondering what will happen if the controller is configured to start these but `experimentalDisableRefResolution` is false - i.e. both the existing and the new reconcilers try to resolve the refs at the same time?

bobcatfish (commented):
I think this is a great incremental step forward!!

/approve

my last 2 cents is wondering at what point we want to switch to this being the default; right now if there are 2 paths that do the resolution this might present some challenges if folks try to maintain this and change one path and not the other

(maybe we can do something like: 1) after the next release, switch the end to end tests to use this second reconciler by default, 2) by the following release make this the default behavior and then 3) by the release after completely remove the old way of doing resolution?)

pkg/reconciler/resolver/taskrun_resolver_reconciler.go
sbwsg: This is a great point.  They'd collide spectacularly. I think for this reason it would be better to have two CLI flags and eliminate the ConfigMap entry.  That way we can error out early during controller startup if the user incorrectly sets `-experimental-enable-resolution-reconcilers` _without_ switching off the existing resolution machinery.

sbwsg:
I've updated the PR to use two CLI arguments instead of a ConfigMap / CLI split.

/hold cancel

mattmoor:
Generally multiple controllers updating the same resource's status is a "code smell", and it looks like this is doing a bunch of gymnastics to work around the reconciler framework, which I'm guessing is at least somewhat related.

I have only paged in a little bit of the context on this, but would be happy to chat about what you are trying to do and brain storm paths forward for this that don't deviate from best practices.

vdemeester:
@sbwsg @mattmoor given the 0.28 schedule (this week or next week), should we hold off this one and move it to the next milestone ?

sbwsg:
Hi @vdemeester @mattmoor I'd like to get this in for 0.28. I'm happy to drop the second reconciler from this PR completely and just leave the `-experimental-disable-in-tree-resolution` flag.

This gives us enough room to start trying out the alternatives from the [remote resolution TEP](https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md) in an experimental controller.

sbwsg:
Here's the additional context for this change (taken from [TEP 0060](https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md#proposal)):

1. Add a new controller in the Experimental Repo implementing some of the Alternatives from TEP 0060 (incl. separate child resource)
2. To support (1) add an experimental flag to disable Pipelines' built-in resolution of Tasks and Pipelines. This way Pipelines and the experimental controller won't race to resolve.

sbwsg:
I've reduced this PR to just the controller flag which disables built-in TaskRun and PipelineRun resolution. 

This gives an experimental controller just enough wiggle room to implement some of the alternatives described in TEP-0060 without racing against the PipelineRun and TaskRun reconcilers.

cmd/controller/main.go
```diff
@@ -110,8 +114,8 @@ func main() {
 
 	ctx = filteredinformerfactory.WithSelectors(ctx, v1beta1.ManagedByLabelKey)
 	sharedmain.MainWithConfig(ctx, ControllerLogKey, cfg,
-		taskrun.NewController(*namespace, images),
-		pipelinerun.NewController(*namespace, images),
+		taskrun.NewController(*namespace, images, *experimentalDisableInTreeResolution),
```
mattmoor, line 117: I know it's sort of a misnomer, but it feels like it'd be nice if this was passed through `images` 🤔 

Since this is a breaking signature change anyways, I wonder whether it's worth renaming the struct and passing this field through it instead.

sbwsg: I think I'm missing a bit here - why's that preferable?

sbwsg: I've added a struct for each controller to pass the images and disableresolution flag through.

cmd/controller/main.go
```diff
@@ -108,10 +112,20 @@ func main() {
...
 
+	taskrunControllerConfig := taskrun.ControllerConfiguration{
+		Images:                   images,
+		DisableTaskRefResolution: *experimentalDisableInTreeResolution,
+	}
+
+	pipelinerunControllerConfig := pipelinerun.ControllerConfiguration{
+		Images:                       images,
+		DisablePipelineRefResolution: *experimentalDisableInTreeResolution,
+	}
```
mattmoor, line 123: Seems inconsistent with `pipeline.Images` for these to specialize between the two controllers, since the pipeline run controller only uses a subset of the `pipeline.Images` struct, but the same "options" struct is passed to both controllers.

My personal preference would be to rename/generalize the `Images` struct to reflect a more complete set of options, but this seems nitpicky and I have some other changes I've been thinking about to some of this handling, so perhaps we land this and I can try to find something we're happy with.

thanks for the change. 👍

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -132,11 +132,21 @@ type Reconciler struct {
...
+	// is.
+	disableResolution bool
 }
 
 var (
 	// Check that our Reconciler implements pipelinerunreconciler.Interface
 	_ pipelinerunreconciler.Interface = (*Reconciler)(nil)
+
+	// Indicates pipelinerun resolution hasn't occurred yet.
+	errResourceNotResolved = fmt.Errorf("pipeline ref has not been resolved")
```
mattmoor, line 149: nit: `errors.New` when there's no formatting.

🤔  I thought this was something golangci-lint would complain about, sorry I didn't notice before.

Also cc @n3wscott in case there's something useful in the controller framework we can use here to simplify any of the logic.  None of the things that come to mind for me seem quite right.

mattmoor (commented):
/lgtm

Issue #4179, opened 2021-08-22
PipelineRun controller mutates the TaskRun lister cache


mattmoor:
# Expected Behavior

All taskruns fetched via the lister are copied prior to any mutation.

# Actual Behavior

This section of code fetches a TaskRun from the lister, and mutates it:
https://github.com/tektoncd/pipeline/blob/44fb5362874cfbae3587d868b732a9410a41ccc5/pkg/reconciler/pipelinerun/pipelinerun.go#L749-L758

Note that while `addRetryHistory` copies the status field, it writes the result back to the original object:
https://github.com/tektoncd/pipeline/blob/44fb5362874cfbae3587d868b732a9410a41ccc5/pkg/reconciler/pipelinerun/pipelinerun.go#L918-L922

# Steps to Reproduce the Problem

Read the code...?  I'm not sure whether this has any related fallout yet, but I'll attempt a fix and update here with any additional landmines I might uncover.

# Additional Info

PR #4180, opened 2021-08-22
Don't modify the lister cache's copy.

mattmoor:
/kind bug

Fixes: https://github.com/tektoncd/pipeline/issues/4179

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -748,6 +748,8 @@ func (c *Reconciler) createTaskRun(ctx context.Context, rprt *resources.Resolved
 
 	tr, _ := c.taskRunLister.TaskRuns(pr.Namespace).Get(rprt.TaskRunName)
 	if tr != nil {
+		// Don't modify the lister cache's copy.
+		tr = tr.DeepCopy()
 		// is a retry
 		addRetryHistory(tr)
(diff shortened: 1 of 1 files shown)


mattmoor:
I'm not thrilled at the idea of discarding the error here either, but I figured it was worth leaving that alone for this change...

mattmoor:
Ok, if I'm being really honest, I'm not a fan of any aspect of how the pipeline is handling retries by messing with another controller's resource's status to recycle it instead of just creating new taskruns 😂

imjasonh (approved):
I'd love some more systematic way to avoid this fwiw, identifying and fixing these by hand isn't going to scale. 🤔

mattmoor:
@imjasonh There is one! 😉  The knative table testing framework checks for this, which I think has effectively eliminated this from our controllers, but Tekton never adopted that because it landed around/after the split 😞 

It'd be a pretty big undertaking, which I probably can't sign up for, but I'd be happy to consult, and I'm sure the Knative folks would be willing to field questions, since most of Knative is using it.

mattmoor:
Here's where we do it, fwiw: https://github.com/knative/pkg/blob/b558677ab03404ed118ba3e179a7438fe2d8509a/reconciler/testing/table.go#L390-L394

PR #4186, opened 2021-08-23
Converts ResultType from string to int enum

### LINKED ISSUE #4150: Convert `"InternalTektonResult"` from string to int enum
Every Step that Pipelines runs emits a JSON object with metadata about its execution: the content of task results, a StartedAt timestamp, step exit codes etc etc.

The JSON object is passed back to the Tekton Controller by way of the "Termination Message", a Kubernetes feature that allows a Container to write a short amount of content that ends up back in the Pod's YAML. The emphasis here is on "short" - the termination message is limited to only a few thousand bytes, which in turn means that the JSON object Steps return needs to fit within this.

Several fields in the emitted JSON object are "InternalTektonResults", meaning that they are metadata which the Tekton Controller needs but that doesn't need to be surfaced in typically user-facing fields like Task Results. Each of these "internal-only" fields is given a "type" member in the Termination Message that looks like this:

```json
{"type": "InternalTektonResult"}
```

Each entry in the JSON object that uses this `type` consumes, at minimum, 28 bytes of the termination message allotment (6 for "type" and 22 for "InternalTektonResult"). As we add more metadata which needs to be passed around this overhead will grow.

For the longer term we've discussed methods of avoiding the Termination Message altogether. In the short term we could easily curtail this wastage by changing `"type"` from a string to an int. e.g. `{"type": 0}`. The int would then map to an enum of possible types (of which there are few). This would drop the overhead of each internal result from 28 bytes to 7.

sm43:
This converts the ResultType in PipelineResourceResult from string
to int enum to curtail the size of json object emittted by steps.
ResultTypes are converted because they made JSON messages bigger, which in
turn limited the amount of space in termination messages for task results. String
support is maintained for backwards compatibility.
    
 Closes #4150
    
 Signed-off-by: Shivam Mukhade <smukhade@redhat.com>

# Submitter Checklist

  functionality, content, code)

# Release Notes

ResultType in JSON blob of termination message is changed from String to Int. So, for ex previouly we used to have `"type": 
"TaskRunResultType"` in the terminaton message which now would appear as `"type": 1` to curtail the size of json.

### DIFF
--- a/pkg/apis/pipeline/v1beta1/task_types.go
+++ b/pkg/apis/pipeline/v1beta1/task_types.go
@@ -23,13 +23,13 @@ import (
 
 const (
 	// TaskRunResultType default task run result value
-	TaskRunResultType ResultType = "TaskRunResult"
+	TaskRunResultType ResultType = 1
 	// PipelineResourceResultType default pipeline result value
-	PipelineResourceResultType ResultType = "PipelineResourceResult"
+	PipelineResourceResultType = 2
 	// InternalTektonResultType default internal tekton result value
-	InternalTektonResultType ResultType = "InternalTektonResult"
+	InternalTektonResultType = 3
 	// UnknownResultType default unknown result type value
-	UnknownResultType ResultType = ""
+	UnknownResultType = 10
 )
 
 // +genclient
--- a/pkg/apis/pipeline/v1beta1/openapi_generated.go
+++ b/pkg/apis/pipeline/v1beta1/openapi_generated.go
@@ -1211,8 +1211,8 @@ func schema_pkg_apis_pipeline_v1beta1_PipelineResourceResult(ref common.Referenc
 					},
 					"type": {
 						SchemaProps: spec.SchemaProps{
-							Type:   []string{"string"},
-							Format: "",
+							Type:   []string{"integer"},
+							Format: "int32",
 						},
 					},
 				},
--- a/pkg/apis/pipeline/v1beta1/resource_types.go
+++ b/pkg/apis/pipeline/v1beta1/resource_types.go
@@ -17,9 +17,11 @@ limitations under the License.
 package v1beta1
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/go-multierror"
 	resource "github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1"
 	v1 "k8s.io/api/core/v1"
 )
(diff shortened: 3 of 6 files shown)


sm43:
@sbwsg Could you take a look and let me know if I am right track? 😅

sbwsg:
One potential risk here is if a controller is upgraded while a TaskRun is mid-execution. If that would happen then the taskrun's pod could return `"TaskRunResult"` but the controller could be expecting `1`.  Is there a way we could work around this by implementing [the `Unmarshaller` interface](https://pkg.go.dev/encoding/json#Unmarshaler) for `ResultType`? So if our custom `UnmarshalJSON()` detects a string then we convert it to its equivalent int?

sbwsg (commented):
Looks good to me so far!

sm43:
@sbwsg yeah that make sense to not change `type` to`t`. Thank you adding the code snippet.
Updated the pr with changes. 
@ScrapCodes Updated `UnknownResultType` to have value `10` let me know if that is good enough
ptal :)

pkg/apis/pipeline/v1beta1/task_types.go
```diff
@@ -23,13 +23,13 @@ import (
...
+	TaskRunResultType ResultType = 1
 	// PipelineResourceResultType default pipeline result value
-	PipelineResourceResultType ResultType = "PipelineResourceResult"
+	PipelineResourceResultType = 2
 	// InternalTektonResultType default internal tekton result value
-	InternalTektonResultType ResultType = "InternalTektonResult"
+	InternalTektonResultType = 3
 	// UnknownResultType default unknown result type value
-	UnknownResultType ResultType = ""
+	UnknownResultType = 10
```
sbwsg, line 32: 👍

sbwsg (approved):
Fantastic, thanks a lot @sm43 !

vdemeester (commented):
A very naive question, isn't this kind-of breaking the API ? Even though pipeline doesn't consume it, there might be tools that are consuming this and used to see strings (like `TaskRunResult`) and will see an int now, isn't it ?

sbwsg:
@vdemeester the way we pass messages from Tasks to the controller is an internal implementation detail isn't it?

vdemeester:
> @vdemeester the way we pass messages from Tasks to the controller is an internal implementation detail isn't it?

Well, isn't this "exposed" as part of the `status` field of a `TaskRun` or `PipelineRun` ? If yes, then this means a user that was looking at those (for whatever reason) will have a "breakage" as the type changed (using to be a string, and is a number now). This is, I think, also affecting the `tkn` cli 🙃 (cc @piyush-garg) as we display results.

sbwsg:
It's exposed at the following location in taskruns:

```yaml
status:
  steps:
  - container: step-print-date-human-readable
    terminated:
      message: '[{"key":"current-date-human-readable","value":"Mon Sep 13 14:28:59
        UTC 2021\n","type":"TaskRunResult"},{"key":"current-date-unix-timestamp","value":"1631543338\n","type":"TaskRunResult"}]'
```

To be super explicit: we're talking about the stringified JSON blob in the termination message field. The "type" field in that stringified JSON blob will change from a string ("TaskRunResult") to an int.

You're saying that consumers of our API are parsing the stringified JSON blob for this data rather than looking in the `status.taskResults` list for the value?

vdemeester:
> You're saying that consumers of our API are parsing the stringified JSON blob for this data rather than looking in the `status.taskResults` list for the value?

They might but yes, if it's only exposed there, then I think we are safe 😛

pritidesai:
I kind of agree with @vdemeester, whoever is looking at the `terminated` JSON blob will have a message with int instead of `TaskRunResult`. But I don't think it was designed to consume it in that way anyways so I am all good with these changes.

I would like to see the changes in the release notes to be on the safer side. @sm43 will you be able to please add the release notes with the status JSON blob change.

afrittoli (commented):
Thanks for this!
I agree with @pritidesai, we need to call this out in the release notes of this PR and of next release. We might want to document that the termination message is not meant as a stable API, even if it's exposed within the status.
Once the release notes are updated this lgtm.

sbwsg:
Nice idea @afrittoli I'll make a PR to document the lack of guarantees around the termination message.

sbwsg:
Created https://github.com/tektoncd/pipeline/pull/4226 to capture termination message note.

sm43:
@sbwsg @pritidesai added release notes, ptal if it looks okay.

pritidesai:
thanks @sm43 

/lgtm

PR #4188, opened 2021-08-24
Attempt to lookup cached entrypoint by digest, if possible

### LINKED ISSUE #4187: ResolveEntrypoints looks up manifest breaking Dockhub Registry Limit
# Expected Behavior
The (global) cache would be used for the image command across all TaskRuns.

# Actual Behavior
Every new TaskRun generated (with no command set) seems to make its own call out to Dockerhub to retrieve the manifest, causing a count against the Dockerhub limits.

1. Create TaskRuns directly (not using a PipelineRun)
2. Don't specify a command
3. Run enough times to break the Dockerhub limit

# Version Info

- Kubernetes version:

  ```
  Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.0+bafe72f", GitCommit:"bafe72fb05eddc8246040b9945ec242b9f805935", GitTreeState:"clean", BuildDate:"2021-03-14T16:01:39Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version: v0.24.1

# Additional Information

- [Pod Build resolveEntrypoints](https://github.com/tektoncd/pipeline/blob/2e5780d80a1dd68b0bdd7a494cbb0edf2529a929/pkg/pod/pod.go#L157)
- [Entrypoint Lookup](https://github.com/tektoncd/pipeline/blob/2e5780d80a1dd68b0bdd7a494cbb0edf2529a929/pkg/pod/entrypoint_lookup.go#L45)
- Looking at [img, err = cache.Get(ctx, origRef, namespace, serviceAccountName)`](https://github.com/tektoncd/pipeline/blob/2e5780d80a1dd68b0bdd7a494cbb0edf2529a929/pkg/pod/entrypoint_lookup.go#L67) I am unsure where the function its calling is. I see that it comes from the builder.
  - Does this mean that it only applies to each instance of a TaskRun or PipelineRun and hence across TaskRuns its always going to go out to DockerHub?

imjasonh:
Today, when the step's image is not specified by digest (i.e., most of the
time), we issue an image pull to get the image's entrypoint, then cache
it. This means that when the step isn't specified by digest (i.e., most
of the time) we hit the remote registry, and can hit DockerHub's rate
limits.

With this change, we'll issue a HEAD request to the registry to first
lookup the digest of the image, then see if we already have the
entrypoint cached by that digest. If the HEAD request fails or the
entrypoint isn't found by image digest, we proceed as normal with the
image pull.

/kind feature

Fixes #4187

# Submitter Checklist

- [n] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [n] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [y] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [y] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [y] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

```release-note
Attempt to avoid incurring a rate limited image pull to lookup cached image entrypoints, if we can avoid it.
```

### DIFF
--- a/pkg/pod/entrypoint_lookup_impl.go
+++ b/pkg/pod/entrypoint_lookup_impl.go
@@ -77,6 +77,18 @@ func (e *entrypointCache) Get(ctx context.Context, ref name.Reference, namespace
 		Architecture: runtime.GOARCH,
 		OS:           runtime.GOOS,
 	}
+
+	// Attempt to lookup the image's digest.
+	// - if HEAD request fails, proceed to GET with remote.Image below --
+	//   some registries don't support HEAD requests, but those with the
+	//   strictest rate limiting (DockerHub) do.
+	desc, err := remote.Head(ref, remote.WithAuthFromKeychain(mkc))
+	if err == nil {
+		if img, ok := e.lru.Get(desc.Digest.String()); ok {
+			return img.(v1.Image), nil
+		}
+	}
+
 	img, err := remote.Image(ref, remote.WithAuthFromKeychain(mkc), remote.WithPlatform(pf))
 	if err != nil {
 		return nil, fmt.Errorf("error getting image manifest: %v", err)


vdemeester (approved):
/meow

imjasonh:
Worth noting: if the image is a multi-arch image (like `ubuntu`, `alpine`, etc.), then `remote.Head` will return the _index's_ digest, which will never be populated in the cache. 😢  This is because the `remote.Image` we call later uses `WithPlatform` of the worker node's OS+arch, and returns the platform-specific image (not index), which gets returned and set in the cache in `entrypoint_lookup.go`.

So this is likely not a full solution to the problem, but should help in the case of single-platform images. We can't seem to resolve index->image (by platform) without a remote `GET`, which incurs the rate limit hit.

PR #4207, opened 2021-09-01
Remove the gosec Makefile target.

dlorenc:
# Changes

This is not managed by golangci-lint, which does it's own verisoning. The
version in golangci-lint likely won't match this version, so they're most likely
mismatched anyway.

Given that this Makefile target isn't used, we should drop it in favor of `make golangci-lint`
for consistency.

Signed-off-by: Dan Lorenc <lorenc.d@gmail.com>

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


mattmoor (approved):
/lgtm

vdemeester:
> This is not managed by golangci-lint, which does it's own verisoning. The
> version in golangci-lint likely won't match this version, so they're most likely
> mismatched anyway.
> 
> Given that this Makefile target isn't used, we should drop it in favor of `make golangci-lint`
> for consistency.

So quick note, `Makefile` is not used anywhere in the CI (yet ? 😛). This would be true for other *checks* target that are in the `Makefile` (notably `errcheck`).

Those were added at a time where they weren't enabled in the golangci-lint configuration, so that a developer could run a one-sot on it's own, or in the case of `gosec`, to run them on `_test.go` even if they are disabled by teh [`golangci-lint` configuration](https://github.com/tektoncd/pipeline/blob/main/.golangci.yml#L20).

For me, `golangci-lint` is made to make sure we can run the same set of checks on the CI and locally. Even though this is not even guaranteed to be true as, `golangci-lint` version here and on the CI might differ 🙃. But anyway, in my opinion, nothing should prevent us to add more checks on the `Makefile` (or as scripts, …) for the developers.

dlorenc:
> So quick note, `Makefile` is not used anywhere in the CI (yet ? 😛). This would be true for other _checks_ target that are in the `Makefile` (notably `errcheck`).

Yep, I was using this to test out golangci-lint updates locally before they make their way through plumbing and into CI. It's nice to be able to change this somewhat independently. The gosec one confused me because it was already out of sync with the one in golangci-lint, so I was seeing different output.

dlorenc:
Rebased, needs another LGTM.

afrittoli (commented):
/lgtm

Issue #1493, opened 2019-10-29
Should evicted pods be retried?


abayer:
This is as much a question as it is a bug report, so... =)

When a `TaskRun` pod gets deleted while it's running, it'll get automatically recreated, but `Evicted` `TaskRun` pods don't get recreated, so far as I can tell. Should we be handling eviction the same way we do deletion? In practice, they're analogous cases, so it feels like they should be handled in the same way, but I'm not 100% sure.

kav:
I'm seeing this in the wild on a shared cluster and am definitely a yes

vincent-pli:
I think it's definitely needed, but behavior for `Evicted` pod in kubernates seems not stable:
https://github.com/kubernetes/kubernetes/issues/88304
https://github.com/kubernetes/kubernetes/issues/55051

I have tried the case, but the `Evicted` pod has no `.Status.Reason` with `Evicted` value, but the official said it is.

Anybody hit the same issue?

afrittoli:
@abayer @vincent-pli @kav any update on this? I would put this into the backlog for now, unless someone is hit hard by this and wants to step up and fix it?

kav:
I've worked around it for the time being so it's not urgent for me

PR #4208, opened 2021-09-01
Remove the pkg/jsonpath library.

dlorenc:
# Changes

This was originally added as part of #1951 and #1493 in January 2020,
but the work was never finished and the package isn't used anywhere.

It's a non-trivial amount of code and maintenance to keep up to date as we
bump linters and scanners, so we should delete it until there's a clear use
case.

I'm not aware of anyone using this now, so another reason to remove it now is to
prevent someone from taking a dependency on it.

Signed-off-by: Dan Lorenc <dlorenc@google.com>

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
The package `pkg/jsonpath` has been removed. This does not effect anyone using tektoncd/pipeline directly, but may break users importing this code as a library during an update. We're unaware of any usage of this package right now.
```


skaegi:
Thanks Dan. This is the right call and can always be resurrected later.

afrittoli (commented):
/lgtm

PR #4209, opened 2021-09-01
Don't rely on condition ordering.

mattmoor:
I noticed a few places in the metrics code that was relying on `Conditions[0]` to access the `Succeeded` condition.  Generally we try to sort the conditions to ensure determinism across reconciles, and our sorting generally puts the summary condition (Ready, Succeeded) first, but this is mostly for UX and not something the consuming code should rely upon.

While I think it's extremely unlikely that this behavior will change, I believe that the code is much more readable this way because it no longer assumes the reader knows about the sorting or has the expectation that the `Succeeded` condition is always first.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/pipelinerunmetrics/metrics.go
+++ b/pkg/pipelinerunmetrics/metrics.go
@@ -29,6 +29,7 @@ import (
 	"go.opencensus.io/tag"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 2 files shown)


mattmoor:
Yay, at least closing/reopening fixes EasyCLA 😁

vdemeester (approved):

PR #4217, opened 2021-09-06
Support activeDeadlineSeconds for Tekton pods 🦌

vdemeester:
# Changes

Kubernetes (and OpenShift) mark `Pod` as either Terminating — has a
"relatively" short life and will terminate at some point — and
NonTerminating — is supposed to run for ever. Kubernetes does the
difference between the two using the `ActiveDeadlineSeconds` field. A
`Pod` with `activeDeadlineSeconds` set will be considered as
Terminating. For example, `Job`'s `Pod` have this field set and are
considered as Terminated.

Currently the pods created by tekton fall under the NonTerminating
quota limits of Kubernetes and OpenShift. This can create issues as
generally builds should fall under the separate terminating quota
limits.

This sets the `activeDeadlineSeconds` field or TaskRun's `Pod` so that
they are considered Terminating. It also sets the value of this field
to a higher value than the specified (or default) Timeout set on the
TaskRun so that Kubernetes won't try to terminate the `Pod` before
Pipeline does.

/kind feature
/cc @sbwsg @bobcatfish @mattmoor @pritidesai @jerop 

Tentatively adding the 0.28 milestone 🙃 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Set `activeDeadlineSeconds` on Tekton's Pod so that they are considered Terminating for Kubernetes. 
This helps supporting ResourceQuota a bit better as now Tekton Pipeline's pod are considered terminating and thus can be using a specific scoped ResourceQuota for those.
```

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -43,6 +43,10 @@ const (
 
 	// ExecutionModeHermetic indicates hermetic execution mode
 	ExecutionModeHermetic = "hermetic"
+
+	// deadlineFactor is the factor we multiply the taskrun timeout with to determine the activeDeadlineSeconds of the Pod.
+	// It has to be higher than the timeout (to not be killed before)
+	deadlineFactor = 1.5
 )
 
 // These are effectively const, but Go doesn't have such an annotation.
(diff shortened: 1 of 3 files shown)


imjasonh:
Nice. This also has the effect that pods won't run forever if the controller happens to be unavailable or late to process timeouts.

vdemeester:
> Nice. This also has the effect that pods won't run forever if the controller happens to be unavailable or late to process timeouts.

Indeed 😛

afrittoli:
I imagine the active deadline seconds are counted from when the pod is scheduled?

It might be nice to add a note in the docs about this, but it's ok as a follow-up pr if you prefer.

/approve

PR #4228, opened 2021-09-15
Refactor the way the Pipelines controllers receive flags.

mattmoor:
This change has a few parts, but the overarching goal is to make it easier for downstreams to repackage Tekton Pipelines across updates to the flags that are parsed and passed into the controllers.
1. Instead of defining `var`s for the flags, and then extracting the flags into the `pipeline.Images` struct, this change takes advantage of `flag.NewFooVar` to pass in the address of the variable to update.  This lets us have flag parsing process things directly into its final variable.
2. Wrap `Images` in `FlagOptions`, which is passed to controllers.  I'm sort of ambivalent about the nested struct vs. renaming Images, but opted for the former because so many files reference `pipeline.Images` and I wanted to reduce the scope of the change.
3. For flags only used in `main()` use a local flag definition (vs. `FlagOptions`).  It turns out the `*namespace` we were passing to the `NewController` methods was unused!
4. For globals we've turned into flags, use the `flag.NewFooVar` form to update the global directly.
5. Colocate the `flag` setup with the validation setup.  This allows us to get coverage of the flag processing code, and it turns out there were some bugs in the names of flags emitted from validation!

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Downstream consumers of `taskrun.NewController` and `pipelinerun.NewController` will have to fix a breaking signature change; however, they can now take advantage of `pipeline.NewFlagOptions(flag.CommandLine)` to reduce churn due to flag changes.
```

### DIFF
--- a/pkg/apis/pipeline/images.go
+++ b/pkg/apis/pipeline/images.go
@@ -17,9 +17,33 @@ limitations under the License.
 package pipeline
 
 import (
+	"flag"
 	"fmt"
 	"sort"
 )
+
+// FlagOptions holds options passed to the Tekton Pipeline controllers via command-line flags.
+type FlagOptions struct {
+	Images                        Images
+	ExperimentalDisableResolution bool
+}
+
+func NewFlagOptions(fs *flag.FlagSet) *FlagOptions {
+	fo := &FlagOptions{}
+	fs.StringVar(&fo.Images.EntrypointImage, "entrypoint-image", "", "The container image containing our entrypoint binary.")
+	fs.StringVar(&fo.Images.NopImage, "nop-image", "", "The container image used to stop sidecars")
+	fs.StringVar(&fo.Images.GitImage, "git-image", "", "The container image containing our Git binary.")
+	fs.StringVar(&fo.Images.KubeconfigWriterImage, "kubeconfig-writer-image", "", "The container image containing our kubeconfig writer binary.")
+	fs.StringVar(&fo.Images.ShellImage, "shell-image", "", "The container image containing a shell")
+	fs.StringVar(&fo.Images.ShellImageWin, "shell-image-win", "", "The container image containing a windows shell")
+	fs.StringVar(&fo.Images.GsutilImage, "gsutil-image", "", "The container image containing gsutil")
+	fs.StringVar(&fo.Images.PRImage, "pr-image", "", "The container image containing our PR binary.")
+	fs.StringVar(&fo.Images.ImageDigestExporterImage, "imagedigest-exporter-image", "", "The container image containing our image digest exporter binary.")
+	fs.BoolVar(&fo.ExperimentalDisableResolution, "experimental-disable-in-tree-resolution", false,
+		"Disable resolution of taskrun and pipelinerun refs by the taskrun and pipelinerun reconcilers.")
+
+	return fo
+}
 
 // Images holds the images reference for a number of container images used
 // across tektoncd pipelines.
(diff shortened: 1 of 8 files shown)


mattmoor:
@sbwsg this is the refactor I had in mind, which is mostly centered around leveraging `flag.NewFooVar` to avoid defining and copying every flag, as well as making it easier for downstream consumers of `NewController` methods to avoid (or reduce) flag churn.

Didn't mean to be a jerk and rewrite the code you just changed, but I didn't want to hold up your change.  I think the pieces you want landed, so 🤞

mattmoor:
blah dead test globals

vdemeester (commented):
I am not a huge fan of the refactor as this bring `flags` notion (and package) to api or reconciler package. Not a big deal because it's the built-in library but still.

Having `FlagOptions` named `Options` instead, and move `NewFlagOptions` to the main would look better (for me 😅 ). In a nutshell, it would mean, renaming `pipeline.Image` to `pipeline.Option` 😝 and passe it to the controller.

mattmoor:
@vdemeester Sure, I can take a swing at that.  I was mostly trying to throw something together, as I'd been mulling a change like this for a while.  This was really just my first stab at it 🗡️

mattmoor:
I dropped `NewFlagOptions` entirely and inlined it into `main()`, which avoids the copies.  Hopefully this looks better 😁

mattmoor:
Hmm, @imjasonh hit this a bunch today too.  I wonder why they are especially flaky today...

/retest

pkg/apis/pipeline/images.go
```diff
@@ -21,6 +21,13 @@ import (
 	"sort"
 )
 
+// Options holds options passed to the Tekton Pipeline controllers
+// typically via command-line flags.
+type Options struct {
```
sbwsg, line 26: `pkg/apis/pipeline` is a bit of a "util" / "bag of random stuff" package (storage options, path constants, images struct, gvk definitions, etc). Given that, I suggest putting the `Options` struct into its own `controller_options.go`, or renaming `images.go`. Ideally `images.go` would only contain code related to `Images`. Also suggest bringing the field comments over from the `ControllerConfiguration` type we're replacing for that same reason - additional context in an otherwise context-less package.

pkg/reconciler/taskrun/controller.go
```diff
@@ -40,18 +40,8 @@ import (
...
-	// Images are the image references used across Tekton Pipelines.
-	Images pipeline.Images
-	// DisableTaskRefResolution tells the controller not to perform
-	// resolution of task refs from the cluster or bundles.
-	DisableTaskRefResolution bool
-}
-
 // NewController instantiates a new controller.Impl from knative.dev/pkg/controller
-func NewController(namespace string, conf ControllerConfiguration) func(context.Context, configmap.Watcher) *controller.Impl {
+func NewController(fo *pipeline.Options) func(context.Context, configmap.Watcher) *controller.Impl {
```
sbwsg, line 44: Ah ok after reading back over comments I think `fo` at one point stood for `FlagOptions`, but now is quite puzzling as a new reader. Suggest `opts` or `options` here just to help the reader trace the path when they're reading the function body. Similar comment for use of `fo` throughout.

vdemeester (approved):

pkg/reconciler/taskrun/controller.go
mattmoor: For sure, will put this in the next push.

pkg/apis/pipeline/images.go
mattmoor: Good call, done.

PR #4234, opened 2021-09-16
Use an Informer to list LimitRanges 🥼

vdemeester:
# Changes

Using a LimitRange lister here instead, so this doesn't end up hitting
the real API server on each call.

Taking into account a review :
https://github.com/tektoncd/pipeline/pull/4176#issuecomment-920366660.

/kind cleanup
/cc @imjasonh @bobcatfish @pritidesai @sbwsg @afrittoli 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


vdemeester:
I am missing something very obvious in the unit tests.. but.. for some reason I don't see it..  😅

pkg/internal/limitrange/transformer_test.go
```diff
@@ -20,12 +20,17 @@ import (
...
 	"github.com/google/go-cmp/cmp"
+	ttesting "github.com/tektoncd/pipeline/pkg/reconciler/testing"
+	"github.com/tektoncd/pipeline/test"
 	"github.com/tektoncd/pipeline/test/diff"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	fakek8s "k8s.io/client-go/kubernetes/fake"
+	// "k8s.io/apimachinery/pkg/runtime"
```
imjasonh, line 29: Can remove this

imjasonh (commented):
/approve

internal/builder/v1beta1/pod.go
```diff
@@ -124,9 +123,6 @@ func PodContainer(name, image string, ops ...ContainerOp) PodSpecOp {
 		c := &corev1.Container{
 			Name:  name,
 			Image: image,
-			Resources: corev1.ResourceRequirements{
-				Requests: map[corev1.ResourceName]resource.Quantity{},
-			},
```
vdemeester, line 129: I am not sure why this was needed to be changed to fix the issue with some tests but... meh.

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -1814,7 +1814,7 @@ func TestReconcile_SetsStartTime(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if err := testAssets.Controller.Reconciler.Reconcile(context.Background(), getRunName(taskRun)); err == nil {
+	if err := testAssets.Controller.Reconciler.Reconcile(testAssets.Ctx, getRunName(taskRun)); err == nil {
```
vdemeester, line 1817: A general rule of thumbs in theses tests : *never use `context.Background`* as `context.Context` for the `Reconcile` loop, as it bypass **all* the init magic `knative/pkg` does 😅

sbwsg (approved):

afrittoli (commented):
Thanks for this!
/lgtm

pkg/reconciler/taskrun/taskrun_test.go
mattmoor: So there is a little bit of magic this does, but the clients and informers are only injected on the `ctx` passed to `NewController` methods.  The `.Get()` methods aren't free (they are a list walk vs. field lookup), so we want folks to stash these during initialization and access them via fields.

So I'd say `context.Background()` is actually probably fine here, just not in calls to `NewController` :sweat`

pkg/internal/limitrange/transformer_test.go
```diff
@@ -985,3 +990,27 @@ func cmpRequestsAndLimits(t *testing.T, want, got corev1.PodSpec) {
...
+	// ServiceAccount
+	serviceAccountInformer := fakeserviceaccountinformer.Get(ctx)
+	kubeclient.PrependReactor("*", "serviceaccounts", test.AddToInformer(t, serviceAccountInformer.Informer().GetIndexer()))
+	for _, ts := range serviceaccounts {
+		if _, err := kubeclient.CoreV1().ServiceAccounts(ts.Namespace).Create(ctx, &ts, metav1.CreateOptions{}); err != nil {
+			t.Fatal(err)
+		}
+	}
+	kubeclient.ClearActions()
+	return ctx, limitRangeInformer, cancel
```
mattmoor, line 1015: Could consumers just use `fakelimitrangeinformer.Get(ctx)` instead of returning it here?

mattmoor (commented):
/lgtm

This LGTM, just a nit about `setup` if you want to preserve the signature.

mattmoor:
/hold

In case you want to do the nit, otherwise feel free to remove

pkg/internal/limitrange/transformer_test.go
vdemeester: c'est pas faux 😝

mattmoor (commented):
/lgtm
/hold cancel

pkg/reconciler/taskrun/taskrun_test.go
vdemeester: @mattmoor yeah, it's a bit tricky. The whole file would benefit from a small refactoring.

pkg/internal/limitrange/transformer_test.go
vdemeester: > c'est pas faux stuck_out_tongue_closed_eyes

Man I am tired.. "c'est pas faux" means "it's not wrong".. speaking in french while commenting.. makes this.. 😅

Issue #4227, opened 2021-09-14
Restrict step volumes as read-only


wlynch:
# Expected Behavior

Internal `/tekton` folders should only be manipulated by trusted Tekton components (i.e. controller / entrypoint).

# Actual Behavior

Tekton mounts the following volumes for each step:

```
# Read-write
- mountPath: /tekton/tools
  name: tekton-internal-tools
- mountPath: /tekton/downward
  name: tekton-internal-downward
- mountPath: /tekton/creds
  name: tekton-creds-init-home-0
- mountPath: /workspace
  name: tekton-internal-workspace
- mountPath: /tekton/home
  name: tekton-internal-home
- mountPath: /tekton/results
  name: tekton-internal-results
- mountPath: /tekton/steps
  name: tekton-internal-steps

# Read-only
- mountPath: /tekton/scripts
  name: tekton-internal-scripts
  readOnly: true
- mountPath: /var/run/secrets/kubernetes.io/serviceaccount
  name: kube-api-access-9fk95
  readOnly: true
```

Since many of these directories are mounted read-write and are shared by all steps in the task, these can be manipulated to cause unexpected behavior for steps.

Directory | Description | Consequence | Recommendation
----------|------------|---------------|-------------------
/tekton/tools | Contains entrypoint, post_files for coordinating step starts | Bad step could manipulate it's own entrypoint (requires container restart), or write post_files to prematurely cause steps to start. | Move entrypoint into its own directory <br><br> Only let step write its own post_files
/tekton/steps | [Step exit codes](https://github.com/tektoncd/pipeline/blob/78d6d75473ea00118fa1e07cbf6b8e97439689ad/docs/developers/README.md#how-to-access-the-exit-code-of-a-step-from-any-subsequent-step-in-a-task) | Bad step can mutate the perceived exit code for other steps. This does **not** affect container exit / Task status. | Only let step write its own exit codes.
/tekton/results | [Results](https://github.com/tektoncd/pipeline/blob/edcaa970f62e1031c4f7243ebc827e453ead6cb5/docs/tasks.md#emitting-results) output | None - Results are defined at the Task level. | Nothing - expected behavior
[...]

PR #4236, opened 2021-09-16
Mark Downward API VolumeMount as readonly.

wlynch:
# Changes

This is effectively a no-op change, since the Downwards API Volume is
already readonly. This simply updates the VolumeMount to also mark the
volume as read-only for completeness.

Tests are repetitive, and would probably benefit from some refactoring to
simplify the common volumes, but considering that out of scope for this change.

This is part of #4227 to lock down step volumes as much as possible.

# Submitter Checklist

  functionality, content, code)


dlorenc:
/lgtm

Nice!

PR #4257, opened 2021-09-24
document how to verify tekton pipelines release

pritidesai:
# Changes

Adding detailed documentation on how to verify Tekton Pipelines release. Tekton pipelines started signing the release since v0.27.1. A step-by-step process to verify the release includes:

* collecting the digest of the release
* searching the transparency logs for that digest
* requesting provenance of the release
* using cosign to verify each artifact

/kind documentation

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Document step-by-step process of how to verify Tekton Pipelines release.
```


docs/install.md
```diff
@@ -9,16 +9,24 @@ weight: 100
...
+- [Configuring CloudEvents notifications](#configuring-cloudevents-notifications)
+- [Configuring self-signed cert for private registry](#configuring-self-signed-cert-for-private-registry)
+- [Customizing basic execution parameters](#customizing-basic-execution-parameters)
+    - [Customizing the Pipelines Controller behavior](#customizing-the-pipelines-controller-behavior)
+    - [Alpha Features](#alpha-features)
+- [Configuring High Availability](#configuring-high-availability)
+- [Configuring tekton pipeline controller performance](#configuring-tekton-pipeline-controller-performance)
+- [Creating a custom release of Tekton Pipelines](#creating-a-custom-release-of-tekton-pipelines)
+- [Verify Tekton Pipelines release](#verify-tekton-pipelines-release)
+- [Next steps](#next-steps)
```
afrittoli, line 29: Nice! :)

docs/install.md
```diff
@@ -424,6 +432,145 @@ Out-of-the-box, Tekton Pipelines Controller is configured for relatively small-s
...
+Tekton Pipeline's images are being signed by [Tekton Chains](https://github.com/tektoncd/chains) since [0.27.1](https://github.com/tektoncd/pipeline/releases/tag/v0.27.1). You can verify the images with
+`cosign` using the [Tekton's public key](https://raw.githubusercontent.com/tektoncd/chains/main/tekton.pub).
+
+With Go 1.16+, you can install `cosign` by running:
+
+```shell
+go install github.com/sigstore/cosign/cmd/cosign@latest
+```
+
+Before, you can verify Tekton Pipeline images, collect the list of images created as part of the release. You can search
```
afrittoli, line 449: NIT: extra comma after "before" ?

docs/install.md
```diff
@@ -424,6 +432,145 @@ Out-of-the-box, Tekton Pipelines Controller is configured for relatively small-s
...
+
+which results in:
+
+```shell
+Found matching entries (listed by UUID):
+eb69098b520a93ce140f4facfa596f4947ae7158910531d4882e466c24faec12
+257d788c3e8f3937e7085de5571ffd78d0156461eb7ca67baa9e9e0cdc7b4dc1
+```
+
+Tekton Chains generates provenance based the custom format in which the `subject` holds the list of artifacts which were
```
afrittoli, line 484: based -> based on

afrittoli: Maybe include a link to the provenance spec https://github.com/tektoncd/chains/blob/main/PROVENANCE_SPEC.md

docs/install.md
```diff
@@ -424,6 +432,145 @@ Out-of-the-box, Tekton Pipelines Controller is configured for relatively small-s
...
+      "name": "gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/webhook",
+      "digest": {
+        "sha256": "f19dd16303ff8a8d55d706db5c8f8db593ba597684888bb15e70420fc4824103"
+      }
+    }
+  ],
+  ...
+```
+
+Now, you can verify each of these images using the Tekton public key:
```
afrittoli, line 551: I don't fully understand the process.

- We start from the sha of the controller image taken from the release file.
- From there we search a matching provenance record. 
- We pick one of the two (does it matter which one? we do we get two?) 
- We look at the provenance record and get a list of the images in the release, with their shas
- We use `cosign` with the each of the images from the provenance record and match the output of cosign to the sha in the provenance record. This is done by adding the release tag to the image name

This way we check the images based on the tag, but we don't check the images in the `release.yaml` - apart perhaps from the controller one. Is that the intention? 

If Tekton users want to very a Tekton release before installing it from the `release.yaml` we should perhaps use the images from that file for the process?

Since the `release.yaml` itself is not signed we cannot guarantee that it has not been tampered with, but at least we can give users a way to check that the signature for all the images included in there is valid.

afrittoli (commented):
Thank you for doing this!! 
I left a couple of comments - let me know what you think.

docs/install.md
pritidesai: > We start from the sha of the controller image taken from the release file.

Sha in the `release.yaml` is used to retrieve the provenance. Provenance collection can only be done using the digest (based on my understanding 😄 ).

> We pick one of the two (does it matter which one? we do we get two?)

Yup, we get two. Doesn't matter which one we choose, both results in the same record.

> We look at the provenance record and get a list of the images in the release, with their shas

> This way we check the images based on the tag, but we don't check the images in the release.yaml - apart perhaps from the controller one. Is that the intention?

That's a good point. I have added a step here to verify the deployment specifications in the `release.yaml` with the provenance collected before using the images.

Let me know how does it read now 🙏

pritidesai: > We pick one of the two (does it matter which one? we do we get two?)

> Yup, we get two. Doesn't matter which one we choose, both results in the same record.

With the latest chains on dogfooding, the issue with multiple `UUID` is resolved. The provenance collection now returns a single `UUID`.  This is for the latest `v0.28.0`:

```shell
rekor-cli search --sha sha256:0c320bc09e91e22ce7f01e47c9f3cb3449749a5f72d5eaecb96e710d999c28e8
Found matching entries (listed by UUID):
68a53d0e75463d805dc9437dda5815171502475dd704459a5ce3078edba96226
```

pritidesai: thanks @priyawadhwa, the issue with multiple `UUID` is resolved 🙏 

@afrittoli let me know what do you think about the step-by-step process now?

PR #4278, opened 2021-10-05
Split /tekton/run directories into separate volumes.

wlynch:
# Changes

This changes splits each `/tekton/run` volume into separate volumes
so that steps can only mutate their own runtime information. This
prevents steps from unexpectedly interfering with other step execution.

To do this, we repurpose the `/tekton/run/#` path into a
step-specific directory. Since this was previously used by the
entrypoint for the post/wait files, we now use `/tekton/run/#/out` as
the post/wait filepath instead.

This does not change behavior of the directory, it enforces expected
behavior of steps.

`/tekton/run` is considered an internal implementation detail and is not
covered by the API compatibility policy, so it is safe to make changes
to the behavior of these files/paths.

This does not stop user execution from writing the step's own `/tekton/run/#`
folder. This needs more discussion/design - additional changes (if
needed) will be made in another commit.

This change is only focused on `/tekton/run` to reduce PR complexity. We will
likely want to make a similar change to /tekton/steps in another commit.
We may also look to consolidate all per-step volumes into a single
source (i.e. creds-init does something similar as well).

AFAICT, Ephemeral Volumes (i.e. EmptyDir) are exempt from Node Volume
limits (https://kubernetes.io/docs/concepts/storage/storage-limits/) - 
spot checked this with a TaskRun with 100+ steps on both kind and GKE.

Part of #4227 

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

Internal only changes.


wlynch:
> Would it be worthwhile to add either an e2e test or an example yaml that exercises the RO behaviour of the volumes?
> 
> E.g. something like:
> 
> ```yaml
> kind: TaskRun
> apiVersion: tekton.dev/v1beta1
> spec:
>   taskSpec:
>     steps:
>     - image: # image with a shell
>       script: |
>         set +e # dont fail the script on error
>         printf "hello world" > /tekton/steps/run/1/foo.txt
>         if [ $? -eq 0 ] ; then
>           echo "able to write to run directory of non-current step"
>           exit 1
>         fi
>       - image: # this step exists only to have the run/1 volumemount in the first step
>         script: echo "step 2"
> ```

Done. (I wish there was a separate test-only directory, but added to examples for now)

Issue #4282, opened 2021-10-06
Publish task fails, IMAGES result too large


afrittoli:
# Expected Behavior

It is possible to release Tekton Pipelines

# Actual Behavior

The publish task fails because the `IMAGES` result it too large:

```
{"level":"fatal","ts":1633523339.3185616,"caller":"entrypoint/entrypointer.go:203","msg":"Error while handling results: Termination message is above max allowed size 4096, caused by large task result.","stacktrace":"github.com/tektoncd/pipeline/pkg/entrypoint.Entrypointer.Go\n\tgithub.com/tektoncd/pipeline/pkg/entrypoint/entrypointer.go:203\nmain.main\n\tgithub.com/tektoncd/pipeline/cmd/entrypoint/main.go:126\nruntime.main\n\truntime/proc.go:225"}
```

# Steps to Reproduce the Problem

1. Trigger a nightly release

# Additional Info

The `IMAGES` result is used by Tekton Chais to sign the container images.
The result includes all the container images produced by `ko` plus all their copies to the various regional registries.

- Kubernetes version:

```
Server Version: version.Info{Major:"1", Minor:"19+", GitVersion:"v1.19.13-gke.1900", GitCommit:"ee714a7b695ca42b9bd0c8fe2c0159024cdcba5e", GitTreeState:"clean", BuildDate:"2021-08-11T09:19:42Z", GoVersion:"go1.15.13b5", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

```
Client version: 0.19.0
Pipeline version: v0.27.3
Triggers version: v0.16.0
Dashboard version: v0.19.0
```

afrittoli:
The result looks like this:
```
gcr.io/tekton-nightly/github.com/tektoncd/pipeline/cmd/controller@sha256:8e749dc794d6c26b54842599eaa61b6ecbc1161d4c8207f6227089a74272d838,
us.gcr.io/tekton-nightly/github.com/tektoncd/pipeline/cmd/controller@sha256:8e749dc794d6c26b54842599eaa61b6ecbc1161d4c8207f6227089a74272d838,
eu.gcr.io/tekton-nightly/github.com/tektoncd/pipeline/cmd/controller@sha256:8e749dc794d6c26b54842599eaa61b6ecbc1161d4c8207f6227089a74272d838,
[...]

PR #4283, opened 2021-10-06
Only add gcr.io images to publish results

afrittoli:
# Changes

The release process generates and number of container images and
publishes them to gcr.io as well as three regional versions of the
registry. Today all those images are added to the IMAGES result,
for signing by chains, however that causes the nightly build to
fail as we hit the termination message size limit.

Until we have a way to store larger results, we shall only sign
images on gcr.io as a workaround.

Workaround to #4282

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Due to a technical limitation, only images published to `gcr.io`, which are included in the `release.yaml`, are signed.
Images on `eu.gcr.io`, `us.gcr.io` and `asia.gcr.io` are not signed yet.
```

### DIFF
--- a/tekton/publish.yaml
+++ b/tekton/publish.yaml
@@ -202,6 +202,8 @@ spec:
             TAG="$(params.versionTag)"
             crane cp ${IMAGE_WITH_SHA} ${REGION}.${IMAGE_WITHOUT_SHA_AND_TAG}:$TAG
           fi
-          echo ${REGION}.$IMAGE_WITH_SHA, >> $(results.IMAGES.path)
+          # Until we are able to store larger results, we cannot include the
+          # regional copies of the images in the result - see https://github.com/tektoncd/pipeline/issues/4282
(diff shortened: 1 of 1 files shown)


vdemeester:
Could we "split" the result into *multiple* bucket (that we know of) ? Like the main one, asia, europe, … ?

vdemeester (approved):

afrittoli:
> Could we "split" the result into _multiple_ bucket (that we know of) ? Like the main one, asia, europe, … ?

We could, however in my understanding we would need to run one `Task` per bucket then.
AFAIU the size limit applies the collection of results, so event having more results in one `Task` would not help.
Chains only supports one `IMAGES` list result and many individual ones. We could add support having multiple lists, but again that would not help with the size limitation.

pritidesai:
/lgtm

would it be overkill to run `publish` task one per region? 😲

Issue #4276, opened 2021-10-05
start using `YAMLParsers` in `/test/*_test.go`


pritidesai:
@imjasonh added `YAML` helpers in https://github.com/tektoncd/pipeline/blob/main/test/yaml.go which takes a `YAML` in a string and returns respective desearlized objects including:

* Task
* TaskRun
* Pipeline
* PipelineRun

These helpers can help simplify the way we have implemented the integration tests. Most of the integration tests under `/test` initialize objects directly or have implemented functions to create such objects. The integration tests also serves as a great reference after `/examples` for the users to learn how to implement task/pipeline specifications. It will be beneficial to our users and contributors, if we can simplify the tests i.e instead of initializing these objects, add `YAML` specifications in the tests.

A huge thank you to @abayer for undertaking this cleanup 🙏 

/kind cleanup

PR #4284, opened 2021-10-06
Convert test/cluster_resource_test.go's Tekton structs to YAML

abayer:
# Changes

Part of #4276

This is just one file from `test/...`, but it seemed like a good "let's start here" case to make sure the approaches I've gone with are acceptable:
* To deal with cases where we're parameterizing the struct currently, this keeps the parameterizing, using `fmt.Sprintf(...)`, since I decided that the raw multiline string literal was the way to go even if we need to do some `%s` replacement in it.
* For the moment, at least, I'm only worrying about Tekton structs - so the `Secret` and `ConfigMap` used in this test I just leave as is, rather than changing them to YAML. If y'all feel we should move those too, well, I'll do that. =)

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/cluster_resource_test.go
+++ b/test/cluster_resource_test.go
@@ -20,11 +20,12 @@ package test
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
-	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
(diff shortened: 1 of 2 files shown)


abayer:
#4286 fixes the flaky `pkg/reconciler/taskrun/taskrun_test.go` `TestReconcile/params` failure, fwiw.

Issue #4297, opened 2021-10-11
PipelineRun reconciler panics when Pending PipelineRun is immediately placed into failed state.


sbwsg:
# Expected Behavior

A PipelineRun whose `spec.status` is `PipelineRunPending` should not trigger a panic when placed into a failed state before execution begins.

# Actual Behavior

When a PipelineRun is marked as failed when its status is `PipelineRunPending` the PipelineRun reconciler panics with the following error:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x21b730d]

goroutine 75 [running]:
github.com/tektoncd/pipeline/pkg/pipelinerunmetrics.(*Recorder).DurationAndCount(0xc0003fdaa0, 0xc00064c500)
        /Users/scott/dev/go/src/github.com/tektoncd/pipeline/pkg/pipelinerunmetrics/metrics.go:219 +0xcd
github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun.(*Reconciler).ReconcileKind.func1(0x2ae29c0)
        /Users/scott/dev/go/src/github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun/pipelinerun.go:209 +0x26
created by github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun.(*Reconciler).ReconcileKind
        /Users/scott/dev/go/src/github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun/pipelinerun.go:208 +0x106f
```

Once this happens the pipelines controller pod enters into a `CrashLoopBackoff` until the offending PipelineRun is deleted.

# Steps to Reproduce the Problem

1. Create a PipelineRun in `PipelineRunPending` state
2. Update the PipelineRun's Status to mark it as failed
3. Watch the controller logs & pod status

# Additional Info

- Kubernetes version:

**Output of `kubectl version`:**

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #4298, opened 2021-10-11
Fix panic when pending pipelinerun is failed

ghost:
# Changes

Fixes https://github.com/tektoncd/pipeline/issues/4297

PipelineRuns that are created with status PipelineRunPending
can be placed into a failed state before their execution begins.
For example: a third-party controller may be watching for pending
PipelineRuns to perform some checks on them prior to execution
beginning. If those checks fail the controller might choose to
set the PipelineRun status to failed with a relevant error message
indicating which check failed and why.

Prior to this commit when a pending PR failed our metrics code
could panic because the PR's StartTime is nil.

This commit adds a guard to the metrics code to ensure that StartTime
is not nil before computing the PR's duration. If it is nil then
we assume the duration is 0. A unit test confirming this behaviour
has been added as well.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fixed an issue where the PipelineRun reconciler could panic if a PipelineRun with spec.status set to PipelineRunPending was placed into a failed state before execution was able to begin.
```

### DIFF
--- a/pkg/pipelinerunmetrics/metrics.go
+++ b/pkg/pipelinerunmetrics/metrics.go
@@ -216,9 +216,12 @@ func (r *Recorder) DurationAndCount(pr *v1beta1.PipelineRun) error {
 		return fmt.Errorf("ignoring the metrics recording for %s , failed to initialize the metrics recorder", pr.Name)
 	}
 
-	duration := time.Since(pr.Status.StartTime.Time)
-	if pr.Status.CompletionTime != nil {
-		duration = pr.Status.CompletionTime.Sub(pr.Status.StartTime.Time)
+	duration := time.Duration(0)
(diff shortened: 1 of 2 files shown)


bobcatfish:
@pritidesai i wonder if it would be helpful if we made a template for creating flake bugs specifically to make it easier to record them 🤔

pritidesai:
> @pritidesai i wonder if it would be helpful if we made a template for creating flake bugs specifically to make it easier to record them 🤔

sorry @bobcatfish for delayed response, just saw this. 

Definitely, I think it will be useful to create such template. I am not sure if its possible but when prow reports integration test failure, if we can attach such template so that the PR author can create a new issue.

PR #4328, opened 2021-10-21
Use cmpopts.IgnoreFields instead of IgnoreType.

wlynch:
# Changes

From
https://pkg.go.dev/github.com/google/go-cmp/cmp/cmpopts#IgnoreTypes:

> IgnoreTypes returns an Option that ignores all values assignable to certain types,
> which are specified by passing in a value of each type.

This means when we pass in `apis.Condition{}.LastTransitionTime.Inner.Time`,
we are actually ignoring all `time.Time`, not just that particular field.

This change uses `IgnoreFields` to correct the behavior. Luckily, this doesn't
impact the tests in a meaningful way.

/kind bug

/cc @dibyom 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/pod/status_test.go
+++ b/pkg/pod/status_test.go
@@ -1198,7 +1198,7 @@ func TestMarkStatusRunning(t *testing.T) {
 		Message: "Not all Steps in the Task have finished executing",
 	}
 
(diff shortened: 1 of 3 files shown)


wlynch:
Root issue where we discovered this: https://github.com/tektoncd/triggers/issues/1245

Issue #4339, opened 2021-10-28
ImagePullSecrets for "system" images


chenbh:
### Feature request

Add a controller flag or configuration option to specify which secret(s) to use for pulling "system" images like entrypoint, nop, shell-image, etc.

### Use case

In a firewalled or air gapped environment where the cluster can't reach gcr.io, images can be mirrored into a private registry behind the firewall. While it's trivial to update the deployment yaml for the controller and webhook, there's no way to specify `imagePullSecrets` for the [images passed to the controller via args](https://github.com/tektoncd/pipeline/blob/6ac0cbc81d7df69cf163e3c980c76efc559afc96/pkg/apis/pipeline/images.go#L24-L47). This breaks tekton as pods like `affinity-assistant` and init containers are unable to start.

I'm willing to work on this feature.

imjasonh:
Is there some reason you can make the mirrored Tekton images available for unauthenticated users within the internal registry? That might be easier than plumbing through `imagePullSecrets` to access otherwise publicly-available, non-sensitive Tekton images that happen to have been mirrored into the internal registry.

If that's not possible (e.g., due to limitations in the internal registry, or broad policies against unauthenticated images access), we can probably solve this a few different ways, but I wanted to make sure we considered the easiest one first 😄

imjasonh:
> 2. Extend the podTemplate functionality to affinity-assistant pods. default-pod-template can be used to propagate imagePullSecret to all TaskRuns and get the init containers working, but only the tolerations and node selector from the PipelineRun pod template are propagated to the affinity-assistant pod.

I think it might be confusing to conflate the `default-pod-template` that gets applied to workload pods and the configuration of Affinity Assistant pods. I might prefer to have a separate `affinity-assistant-pod-template` config option, to separate these configs.

This also lets us be more diligent about which PodTemplate fields we support in `default-pod-template` vs for Affinity Assistant. AFAIK nobody has expressed a need to set the `runtimeClassName` on AA, and I'd like to be able to consider adding it explicitly, as well as future fields we support for `default-pod-template`.

chenbh:
> broad policies against unauthenticated images access
Unfortunately this is the case. Technically this can be solved by attaching the imagePullSecret to the ServiceAccount, but we're operating in an environment with a clear distinction between users (i.e. Pipeline/Task authors) and cluster operators (i.e. ability to configure/deploy Tekton) and I would prefer this functionality to be the responsibilities of the operators.

[...]

PR #4348, opened 2021-11-02
Use fields from PodTemplate when creating affinity-assistant pod

chenbh:
This is a less intrusive way of solving #4339 using the existing pod templating mechanic. I'm not sure if this is the ideal solution since the users will still have to manually create the secrets in every namespace that has a PipelineRun.

# Changes

Add `ImagePullSecrets`, `RuntimeClassName`, `PriorityClassName`, and `SchedulerName` fields from the pod template to the affinity-assistant StatefulSet. Also use the default pod template instead of just the one from PipelineRun spec.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
The affinity-assistant pod now supports templating `ImagePullSecrets`, `RuntimeClassName`, `PriorityClassName`, and `SchedulerName` from the PipelineRun spec or defaults.
```


vdemeester (approved):

docs/podtemplates.md
```diff
@@ -7,7 +7,7 @@ weight: 1400
 # Pod templates
 
 A Pod template defines a portion of a [`PodSpec`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#pod-v1-core)
-configuration that Tekton can use as "boilerplate" for a Pod that runs your `Tasks` and `Pipelines`.
+configuration that Tekton can use as "boilerplate" for a Pod that runs your `Tasks`, `Pipelines` and [Affinity Assistants](./workspaces.md#specifying-workspace-order-in-a-pipeline-and-affinity-assistants).
```
afrittoli, line 10: Thanks for this.

I think it would be good to document how to configure the `Pod` template for the affinity assistant.
Looking at the code, it looks like it would be taken from the global `Pod` template config, so you could at least updates L15-19 to reflect that.

pkg/reconciler/pipelinerun/affinity_assistant_test.go
```diff
@@ -115,6 +156,56 @@ func TestThatCustomTolerationsAndNodeSelectorArePropagatedToAffinityAssistant(t
 	if len(stsWithTolerationsAndNodeSelector.Spec.Template.Spec.NodeSelector) != 1 {
 		t.Errorf("expected a NodeSelector in the StatefulSet")
 	}
+
+	if len(stsWithTolerationsAndNodeSelector.Spec.Template.Spec.ImagePullSecrets) != 1 {
+		t.Errorf("expected ImagePullSecrets in the StatefulSet")
+	}
+}
+
+func TestMergedPodTemplatesArePropagatedToAffinityAssistant(t *testing.T) {
```
afrittoli, line 165: 👍

afrittoli (commented):
Thank you for this PR!!

The code looks good.
I think we should have just a bit docs before this is merged, so that it's clearer to users how to configure the template for the affinity assistant.

The affinity assistant already uses parts of the pod template, so this consistently extends the current approach. However, for people with an existing pod template in config this may trigger a sudden change in behaviour, so I wonder if this should require a dedicated feature flag.

I tend to see this more like a bug fix - so no feature flag required - but I wanted to ask the opinion of the test of the @tektoncd/core-maintainers .

vdemeester:
> I tend to see this more like a bug fix - so no feature flag required - but I wanted to ask the opinion of the test of the @tektoncd/core-maintainers .

I agree there 👍🏼

vdemeester (approved):

imjasonh:
/hold

See comment in https://github.com/tektoncd/pipeline/issues/4339#issuecomment-984653838 -- I think we might want to consider a separate pod template config just for Affinity Assistant.

tomonight:
cool !!! @chenbh  , you do what i wanna do  , i dont know why they choose some field to assemble pod template, i just now create issue #4431

sbwsg:
I also agree with @imjasonh here - the template for the affinity assistant should be a separate bit of config from the default pod template applied to TaskRun pods (while respecting backwards-compatibility by copying over the tolerations and nodeselector from the default template as we do now). This would keep the affinity assistant an operator concern and not bleed new responsibilities into that default template.

If I'm understanding @imjasonh 's suggestion correctly then this would require:

1. Updating the [config-defaults.yaml](https://github.com/tektoncd/pipeline/blob/e6327c869317f83d52f59932542d06269993b075/config/config-defaults.yaml#L54-L57) with a new field specific to the affinity assistant (`affinity-assistant-pod-template`).
1. Update our config package to bring in the new template (see [default.go](https://github.com/tektoncd/pipeline/blob/e6327c869317f83d52f59932542d06269993b075/pkg/apis/config/default.go#L113-L119) where we parse the pod template)
1. Update the [affinity assistant code](https://github.com/tektoncd/pipeline/blob/e6327c869317f83d52f59932542d06269993b075/pkg/reconciler/pipelinerun/affinity_assistant.go#L131-L141) to use this new template (respecting the use of pr.PodTemplate for overriding the fields that it currently does now).

Then test + document that new field.

@imjasonh does this work match what you were describing in https://github.com/tektoncd/pipeline/issues/4339#issuecomment-984653838 ?
@chenbh wdyt of this suggestion, do you still have bandwidth to proceed with this PR?

imjasonh:
> @imjasonh does this work match what you were describing in [#4339 (comment)](https://github.com/tektoncd/pipeline/issues/4339#issuecomment-984653838) ?

Exactly right! 👍

chenbh:
> @chenbh wdyt of this suggestion, do you still have bandwidth to proceed with this PR?

👍 I currently don't have bandwidth, but I should be more free in about 2 weeks

docs/podtemplates.md
chenbh: Ended up adding a dedicated section for it: https://github.com/tektoncd/pipeline/pull/4348/files#diff-d8925b360ac9c8031196d7a047fa9b04b848577c34988d08e1396d54e48419fbR25-R38

chenbh:
@imjasonh I've updated the PR, lmk if you have any suggestions

imjasonh (approved):
Thanks!

chenbh:
@imjasonh what should I do to get this PR moving forward? I believe it needs a lgtm label (and maybe the `do-not-merge/hold` label removed?) but I'm not sure who to ping for that.

imjasonh:
Yeah it'll need an lgtm from someone other than me.

@afrittoli @pritidesai @bobcatfish

pkg/apis/pipeline/v1beta1/pod.go
```diff
@@ -20,3 +20,93 @@ import "github.com/tektoncd/pipeline/pkg/apis/pipeline/pod"
...
+			tpl.HostNetwork = true
+		}
+		return tpl
+	}
+}
+
+// AAPodTemplate holds pod specific configuration for the affinity-assistant
+type AAPodTemplate = pod.AffinityAssistantTemplate
+
+// MergeAAPodTemplateWithDefault is the same as MergePodTemplateWithDefault but
```
sbwsg, line 89: Despite them being extremely similar I suggest documenting MergeAAPodTemplateWithDefault as if MergePodTemplateWithDefault didn't exist. The reason being that MergePodTemplateWithDefault might get renamed, refactored, or removed and this comment would likely not be updated in lock-step.

sbwsg (commented):
One small nit but otherwise this looks good to me, cheers!

sbwsg:
I can follow up on my own comment in a new PR.  No need to block on it.

/lgtm

sbwsg:
Might be that I can't remove the hold, or could be related to github's recent pains.  @imjasonh could you try?

Issue #4169, opened 2021-08-17
`workspace-in-sidecar` example test is very flakey


ghost:
# Expected Behavior

Examples should run reasonably consistently from PR to PR.

# Actual Behavior

I estimate that the `workspace-in-sidecar` example fails about half the time I initiate an integration run in our CI. It just doesn't seem to be able to complete successfully on a regular basis. It's super unclear why this is but here's the error log from the [most recent run](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4168/pull-tekton-pipeline-alpha-integration-tests/1427773298288103424/) that failed for me:

```
    build_logs.go:35: Could not get logs for pod workspace-in-sidecar-phrks-pod-x95zf: container "step-unnamed-0" in pod "workspace-in-sidecar-phrks-pod-x95zf" is waiting to start: PodInitializing
```

~~Hunch: maybe setting explicit resource requirements for the containers in the example would help?~~
^ This unfortunately did not help after all.

# Steps to Reproduce the Problem

1. Open a PR and let the integration suites run
2. Check if `workspace-in-sidecar` example failed
3. Re-run the test suites if the example passed

sbwsg:
Here's a snippet of the yaml from the most recent failure I saw:

```yaml
        status:
          conditions:
          - lastProbeTime: null
            lastTransitionTime: "2021-08-17T23:41:18Z"
            status: "True"
            type: Initialized
          - lastProbeTime: null
            lastTransitionTime: "2021-08-17T23:41:12Z"
            message: 'containers with unready status: [step-unnamed-0 sidecar]'
            reason: ContainersNotReady
            status: "False"
            type: Ready
          - lastProbeTime: null
            lastTransitionTime: "2021-08-17T23:41:12Z"
            message: 'containers with unready status: [step-unnamed-0 sidecar]'
            reason: ContainersNotReady
            status: "False"
            type: ContainersReady
          - lastProbeTime: null
            lastTransitionTime: "2021-08-17T23:41:12Z"
            status: "True"
            type: PodScheduled
          containerStatuses:
          - image: alpine:3.12.0
            imageID: ""
            lastState: {}
            name: sidecar
            ready: false
            restartCount: 0
            started: false
            state:
              waiting:
                reason: PodInitializing
          - image: alpine:3.12.0
            imageID: ""
            lastState: {}
            name: step-unnamed-0
            ready: false
            restartCount: 0
            started: false
            state:
[...]

PR #4349, opened 2021-11-02
Rewrite flakey workspace-in-sidecar example

sbwsg:
# Changes

Issue: https://github.com/tektoncd/pipeline/issues/4169

The `workspace-in-sidecar` example taskruns are the source of regular
flakes in Pipelines' CI. The examples work by using files in a shared
emptyDir volume to synchronize the behaviour of two containers.

This commit introduces a named pipe for synchronizing behaviour between
the two containers, removing one of the file polling loops. The size of the
shared volume has been set extremely low (just in case the problem is related
to disk pressure on the kubelet) and extra log lines are also included to help
narrow down where a freeze might be occurring.

Adding a hold to see if we can get the CI to fail on workspace-in-sidecar
examples for debugging.

/hold

/kind flake

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/examples/v1beta1/taskruns/alpha/workspace-in-sidecar.yaml
+++ b/examples/v1beta1/taskruns/alpha/workspace-in-sidecar.yaml
@@ -12,28 +12,41 @@ spec:
   timeout: 60s
   workspaces:
   - name: signals
-    emptyDir: {}
+    emptyDir:
+      sizeLimit: 1M
   taskSpec:
     workspaces:
     - name: signals
     steps:
     - image: alpine:3.12.0
+      resources:
+        requests:
+          memory: "16Mi"
+          cpu: "100m"
       script: |
         #!/usr/bin/env ash
-        echo "foo" > "$(workspaces.signals.path)"/bar
-        echo "Wrote bar file"
-        while [ ! -f "$(workspaces.signals.path)"/ready ] ; do
-          echo "Waiting for $(workspaces.signals.path)/ready"
-          sleep 1
-        done
-        echo "Saw ready file"
+        mkfifo "$(workspaces.signals.path)/channel"
+        echo "Greeting sidecar..."
+        echo "hello there" > "$(workspaces.signals.path)/channel"
+        echo "Awaiting response..."
+        read response < "$(workspaces.signals.path)/channel"
+        echo "Sidecar responded: ${response}"
+        echo "Step Done."
     sidecars:
     - image: alpine:3.12.0
+      resources:
+        requests:
+          memory: "16Mi"
+          cpu: "100m"
       script: |
         #!/usr/bin/env ash
-        while [ ! -f "$(workspaces.signals.path)"/bar ] ; do
-          echo "Waiting for $(workspaces.signals.path)/bar"
-          sleep 1
+        while [[ ! -p "$(workspaces.signals.path)/channel" ]] ; do
+          sleep 4
+          echo "Polling channel..."
         done
-        touch "$(workspaces.signals.path)"/ready
-        echo "Wrote ready file"
+        echo "Awaiting greeting..."
+        read message < "$(workspaces.signals.path)"/channel
+        echo "Step sent message: ${message}"
+        echo "Responding..."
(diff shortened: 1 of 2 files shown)


bobcatfish:
i havent seen a named pipe in so long 🤩 

(if you want to try to get the CI to trigger this, one terribly hacky thing you could do is update the test logic to run this test a bunch of times - like 100 times)

bobcatfish:
@sbwsg i wonder if it's possible this is some kind of interaction between the volumeMounts specification and workspaces - probably a red herring but i was surprised to see this example wasn't using the sidecar workspaces feature (i guess it pre-dates it?)

sbwsg:
@bobcatfish great idea about running the test lots of times, I'll give that a shot!

The `alpha` example uses Sidecar Workspaces - it's automatic so the sidecar gets access to the workspace without the explicit `volumeMount` that appears in the non-alpha copy of the taskrun.

sbwsg:
I updated the test runner to repeat the workspace-in-sidecar example 20 times per run, and then re-ran the suite 3 times. So after 60 executions the `workspace-in-sidecar` example didn't hit a timeout or error.

I don't think that this PR necessarily solves the problem with the example but hopefully the extra log lines I've added will help surface the real underlying problem when it rears its head again.

bobcatfish:
> So after 60 executions the workspace-in-sidecar example didn't hit a timeout or error.

i guess you jinxed it @sbwsg  X'D

that's one way to reproduce an error i suppose XD

sbwsg:

Frustratingly the tests that failed during this alpha integration run weren't the `workspace-in-sidecar` examples :(

sbwsg:
The isolated workspaces example failed because it took longer than 1 minute to complete.

```
--- FAIL: TestExamples/v1beta1/pipelineruns/alpha/isolated-workspaces (62.23s)
```

The `TestDuplicatePodTaskRun` integration test also failed but with no clear explanation. I did notice that it looks like 50 taskrun pods were spun up as part of that test.

jerop (approved):
excited to see this fix as i just hit this issue in another pr! thanks @sbwsg 🎉

pritidesai:
thank you @sbwsg for detailed explanation 🙏  hoping not to see this flake again 🤣 

/lgtm

pritidesai:
```
    examples_test.go:62: Failed waiting for task run done: "workspace-in-sidecar-h74lx" failed
    build_logs.go:35: Could not get logs for pod workspace-in-sidecar-h74lx-pod-swrd6: container "step-unnamed-0" in pod "workspace-in-sidecar-h74lx-pod-swrd6" is waiting to start: PodInitializing
```

flake while fixing flake 🤔 

sbwsg:
Sigh, ok clearly my changes don't fix the flake :D

Based on timestamps from the taskrun and pod here's the order of operations:

```
(T=0s)  TaskRun created                   2021-11-10T21:00:58Z
(T=1s)  Pod scheduled, created            2021-11-10T21:00:59Z
(T=18s) Pod reaches Initialized condition 2021-11-10T21:01:16Z
(T=60s) Step times out, marked finished   2021-11-10T21:01:58Z
(T=90s) Pod deleted                       2021-11-10T21:02:28Z
```

Couple observations here:

1. When the timeout was hit the test runner recorded the Pod's state. Both the step and sidecar containers were marked as `started: false` with `reason: PodInitializing`. This contradicts the Pod's `condition` which suggests that it was fully initialized at T=18s. I can't explain this difference.
2. In tandem with (1), no logs were captured because `container "step-unnamed-0" in pod "workspace-in-sidecar-h74lx-pod-swrd6" is waiting to start: PodInitializing`. Super weird that the Pod was in an Initialized state after 18 seconds but the containers were still in a holding pattern after 60.
3. The Step really only had 42 seconds to complete, not 1 minute, since initialization took 18s.
4. Duration from timeout to pod deletion is exactly 30s, which matches the pod's termination grace period. Trapping SIGINT in the sidecar and exiting immediately might speed this up and release resources back to the cluster sooner.

PR #4352, opened 2021-11-03
Lockdown /tekton/step folders to their own steps.

wlynch:
# Changes

This change symlinks /tekton/step folders to a step's corresponding /tekton/run folder.
This is an incremental change to lock down the /tekton folder to prevent tampering
of exitCode files from other steps.
Note: this does not completely protect against a step from tampering from its own
output - more work will needed in a future PR to fully lock this down, but this
is a step in the right direction (and a complete fix will likely require a more
involved design).

- Moves `exitCode` output to `/tekton/run/<step #>/status`
- Symlinks `/tekton/steps/<step #>` and `/tekton/steps/<step name>` to
`/tekton/run/<step #>/status`.
- Creates new `tekton-init` entrypoint subcommand to initialize the
Tekton step directory.
- Removes `-step_metadata_dir_link` flag from the main entrypoint binary
(this behavior is now handled by the initcontainer).

Co-authored-by: Lee Bernick <leebernick@google.com>

/kind cleanup

## Things to bikeshed:

1. Thoughts on `/tekton/run/0/status` as the run directory name for user facing data?

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
- /tekton/steps is now read-only.
- Directories in /tekton/steps are now symlinks. The content for the resolved paths remains the same.
```


docs/developers/README.md
```diff
@@ -536,14 +537,44 @@ integration tests run. When the flag in your cluster is `alpha` _all_
...
+for storing the step specific information/metadata. Steps can only write
+metadata to their own `/run` directory - all other step volumes are mounted as
+`readonly`.
 
-Let's take an example of a task with three steps, each exiting with non-zero
-exit code:
+### `/tekton/steps`
+
+`/tekton/steps` contains Step information that is intended to be used by user
+steps (and is therefore covered by the Tekton API compatibility policy).
```
bobcatfish, line 550: I'm wondering:
1. intended to be used by user steps in what way?
2. this is currently not included in our compat policy, maybe we need to update it - our compatibility policy currently mentions "the structure of the directories created in executing containers by Tekton" (https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#what-is-the-api) and links to https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#reserved-directories which indicates everything other than /workspace and /tekton/results is considered an implementation detail

bobcatfish: i think im generally not understanding why we want both steps and run to exist - i think i understand that "steps" is intended to be used by the contents of the steps themselves (the user specified steps) but i dont understand why - and also if run is readonly anyway (except for the current step) im not sure why we need the 2 dirs

bobcatfish (commented):
Some initial drive by feedback from me, mostly I think I'm missing some details around why we need to maintain the 2 folders.

Also looks like a lot of coverage went down, maybe some missing unit test cases?

docs/developers/README.md
wlynch: We have conflicting documentation - https://github.com/tektoncd/pipeline/blob/5dcd43431fe1fdc62a4ed59fde3f01ca43536d72/docs/developers/README.md?plain=1#L195-L200 says it is in scope.

If it's not in scope, I'm more than happy to remove the `/steps` directory altogether. :D The symlink shenanigans are purely to preserve the /tekton/steps file structure for backwards compatibility.

wlynch: /cc @pritidesai who wrote the original documentation in https://github.com/tektoncd/pipeline/commit/92746a2362117c729ad865717dd0dab5ec3d326b

wlynch:
Part of #4227

docs/developers/README.md
```diff
@@ -536,14 +537,44 @@ integration tests run. When the flag in your cluster is `alpha` _all_
...
 
-## What and Why of `/tekton/steps`
+## What and Why of `/tekton/run`
 
-`/tekton/steps/` is an implicit volume mounted on a pod and created for storing
-the step specific information/metadata. There is one more subdirectory created
-under `/tekton/steps/` for each step in a task.
+`/tekton/run` is a collection of implicit volumes mounted on a pod and created
+for storing the step specific information/metadata. Steps can only write
+metadata to their own `/run` directory - all other step volumes are mounted as
```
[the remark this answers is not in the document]
wlynch: Added a note!

docs/developers/README.md
bobcatfish: ah yikes! some scope creep but if we could fix this so that we are only attempting to list the folders in the compat policy in one place that would be ideal 🙏  (pretty good chance adding 2 sources was my doing...)

Looking at TEP-0040 and the origin of the change, it's not clear to me if that folder needs to be part of the API spec or not (https://github.com/tektoncd/community/blob/main/teps/0040-ignore-step-errors.md#proposal), since we provide `$(steps.<step-name>.exitCode.path)` that does give us some freedom to potentially change the location of the directory. The TEP even says:

> If you would like to use the tekton internal path, you can access the exit code by reading the file (it is not recommended though)

i.e. the path is referred to as "internal" several times in the TEP

so basically I have mixed feelings - if @pritidesai and/or others feel strongly the path should remain part of our API, no objections from me - but if possible I'd rather consider that an implementation detail and reply on variable replacement insteaed

pritidesai: thanks @wlynch and @bobcatfish for bringing my attention to this!

TEP-0040 proposed using the path variable `$(steps.<step-name>.exitCode.path)` and avoid using the tekton internal path. 

At the same time, we had a vision of having the `debug` feature utilizing the `/steps` directory structure (not yet implemented) and the symlinks will make it easier to invoke the debug scripts.

https://github.com/tektoncd/community/pull/463#issuecomment-873225465 

Just based on TEP-0040, I would avoid specifying it as covered by the Tekton API compatibility policy.

vdemeester (approved):

docs/developers/README.md
wlynch: SGTM! I'll make these changes in another PR so it's clear why we're making the change.

cmd/entrypoint/main.go
```diff
@@ -45,8 +45,7 @@ var (
 	breakpointOnFailure = flag.Bool("breakpoint_on_failure", false, "If specified, expect steps to not skip on failure")
 	onError             = flag.String("on_error", "", "Set to \"continue\" to ignore an error and continue when a container terminates with a non-zero exit code."+
 		" Set to \"stopAndFail\" to declare a failure with a step error and stop executing the rest of the steps.")
-	stepMetadataDir     = flag.String("step_metadata_dir", "", "If specified, create directory to store the step metadata e.g. /tekton/steps/<step-name>/")
-	stepMetadataDirLink = flag.String("step_metadata_dir_link", "", "creates a symbolic link to the specified step_metadata_dir e.g. /tekton/steps/<step-index>/")
+	stepMetadataDir = flag.String("step_metadata_dir", "", "If specified, create directory to store the step metadata e.g. /tekton/steps/<step-name>/")
```
afrittoli, line 48: NIT: it looks like the "=" lost its alignment?

cmd/entrypoint/post_writer.go
```diff
@@ -13,11 +14,16 @@ type realPostWriter struct{}
...
 func (*realPostWriter) Write(file string, content string) {
 	if file == "" {
 		return
 	}
+
+	// Create directory if it doesn't already exist
+	if err := os.MkdirAll(filepath.Dir(file), os.ModePerm); err != nil {
+		log.Fatalf("Error creating parent directory of %q: %v", file, err)
+	}
+
```
afrittoli, line 26: Could you add tests for this new behaviour?
I tried removing this and unit tests in `cmd/entrypoint/post_writer_test.go` still work fine.

examples/v1beta1/taskruns/readonly-internal-dir.yaml
```diff
@@ -19,3 +19,13 @@ spec:
           echo "able to write to run directory of non-current step"
           exit 1
         fi
+    - image: ubuntu
```
afrittoli, line 22: 👍

afrittoli (commented):
Thank you for this! It's great to see security oriented patches to the core bits of Tekton!
This will need a rebase (because of the a document change 😅 ) - so perhaps you could add one more unit tests to `step_init` to cover the new functionality added there to create the root folder.

Otherwise this looks good to me

cmd/entrypoint/main.go
wlynch: 🤷  gofmt is doing this, so I'm not going to fight it.

docs/developers/README.md
wlynch: Done in #4395. To limit the scope of this PR, I'm keeping the /tekton/steps folder (since I'm not sure what else is depending on it and how intrusive of a change it would be to remove). We can look into removing this in another PR.

docs/developers/README.md
```diff
@@ -172,15 +172,16 @@ Here is an example of a directory layout for a simple Task with 2 script steps:
...
 |   `-- script-1-4pjwp
 |-- steps
-|   |-- 0 -> /tekton/steps/step-unnamed-0
-|   |-- 1 -> /tekton/steps/step-foo
-|   |-- step-foo
-|   `-- step-unnamed-0
-|       `-- exitCode
+|   |-- 0 -> /tekton/run/0/status
+|   |-- 1 -> /tekton/run/1/status
+|   |-- step-foo -> /tekton/run/1/status
```
waveywaves, line 183: In https://github.com/tektoncd/pipeline/pull/4416 I have updated tekton debug scripts to write the exitCode to the $(postfile).breakpointexit file. Created https://github.com/tektoncd/pipeline/issues/4422 to track the necessary changes I'll have to make after this PR gets merged.

cmd/entrypoint/subcommands/subcommands_test.go
```diff
@@ -38,6 +38,21 @@ func TestProcessSuccessfulSubcommands(t *testing.T) {
 	if _, ok := returnValue.(SubcommandSuccessful); !ok {
 		t.Errorf("unexpected return value from decode-script command: %v", returnValue)
 	}
+
+	t.Run(StepInitCommand, func(t *testing.T) {
```
sbwsg, line 42: Nice use of `t.Run` here to segment the test. Possible to reproduce with the two `returnValue` checks above as well to keep them consistent? Something like

```go
t.Run(CopyCommand, func(t *testing.T) {
  // ...
})
t.Run(DecodeScriptCommand, func(t *testing.T) {
  // ...
})
t.Run(StepInitCommand, func(t *testing.T) {
  // ...
})
```

We might be getting into table test territory too.

docs/developers/README.md
sbwsg: Tiny nit to close the loop here: remove the remaining parenthetical in this commit describing /tekton/steps as covered ("and is therefore covered by the Tekton API compatibility policy").

docs/developers/README.md
```diff
@@ -540,14 +541,46 @@ integration tests run. When the flag in your cluster is `alpha` _all_
...
+### Entrypoint configuration
+
+The entrypoint is modified to include an additional flag representing the step
+specific directory where step metadata should be written:
+
+```
+step_metadata_dir - the dir specified in this flag is created to hold a step specific metadata
+```
+
+`step_metadata_dir` is set to `/tekton/run/<task #>/status` for the entrypoint
```
sbwsg, line 577: s/task #/step #/ ?

docs/developers/README.md
```diff
@@ -727,3 +696,19 @@ from the message is internal to Tekton Pipelines, used for book-keeping, and
 some is distributed across a number of fields of the `TaskRun's` `status`. For
 example, a `TaskRun's` `status.taskResults` is populated from the termination
 message.
+
+## Experimentally Disabling PipelineRef and TaskRef Resolution
```
sbwsg, line 700: This chunk is a rebase artifact I think, should be removed to match latest main.

sbwsg (commented):
Some small comments but otherwise I think this looks ready to merge.

cmd/entrypoint/subcommands/subcommands_test.go
wlynch: Sure thing! I'll do this in another PR though

docs/developers/README.md
wlynch: Done.

docs/developers/README.md
wlynch: Good catch! Fixed.

docs/developers/README.md
wlynch: Done.

PR #4353, opened 2021-11-03
Reduce integration test run times

sbwsg:
# Changes

The `TestDuplicatePodTaskRun` test creates 50 pods every integration run (25 for v1beta1, 25 for v1alpha1).
On a successful test run in Pipelines' CI cluster this takes around 49 seconds. Locally
in a kind cluster a successful run takes 40 seconds. This commit reduces the number of pods created to 10 per integration run (5 for v1beta1, 5 for v1alpha1). This drops the run time to anything from 10 seconds locally to 21 seconds for the first run in our CI (v1alpha1) and 36 seconds for the second run (v1beta1).

`TestStartTime` takes 73 seconds in our CI and includes multiple 10 second `sleep`s. Reducing these to 2 seconds drops the test run time to 21 seconds.

`TestSidecarTaskSupport` is sped up by trapping the exit signals sent to the container when a sidecar is stopped and by running the sidecar subtests in parallel. This looks to drop the test time from about 60 seconds to around 15 seconds. Log spam from this test is also reduced by sleeping inside an otherwise tight while loop.

~~Total time for an integration run drops from ~ 25 minutes to ~ 22 minutes.~~ Correction: The total runtime of the integration tests is too variable to assign a specific improvement.

Using this integration log for my "before" numbers: https://storage.googleapis.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/4351/pull-tekton-pipeline-integration-tests/1455940433015214081/build-log.txt

Using this integration log for my "after" numbers: https://storage.googleapis.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/4353/pull-tekton-pipeline-integration-tests/1456052547604189184/build-log.txt

And this one for the sidecar tests "after" numbers: https://prow.tekton.dev/log?job=pull-tekton-pipeline-integration-tests&id=1456370392624009217

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/duplicate_test.go
+++ b/test/duplicate_test.go
@@ -32,7 +32,7 @@ import (
 	"knative.dev/pkg/test/helpers"
 )
 
-// TestDuplicatePodTaskRun creates 10 builds and checks that each of them has only one build pod.
+// TestDuplicatePodTaskRun creates multiple builds and checks that each of them has only one build pod.
 func TestDuplicatePodTaskRun(t *testing.T) {
 	t.Parallel()
 
@@ -44,7 +44,10 @@ func TestDuplicatePodTaskRun(t *testing.T) {
 	defer tearDown(ctx, t, c, namespace)
 
 	var wg sync.WaitGroup
-	for i := 0; i < 25; i++ {
+	// The number of builds generated has a direct impact on test
+	// runtime and is traded off against proving the taskrun
+	// reconciler's efficacy at not duplicating pods.
+	for i := 0; i < 5; i++ {
 		wg.Add(1)
 		taskrunName := helpers.ObjectNameForTest(t)
 		t.Logf("Creating taskrun %q.", taskrunName)
--- a/test/sidecar_test.go
+++ b/test/sidecar_test.go
@@ -47,9 +47,11 @@ func TestSidecarTaskSupport(t *testing.T) {
 		stepCommand    []string
 		sidecarCommand []string
 	}{{
-		desc:           "A sidecar that runs forever is terminated when Steps complete",
-		stepCommand:    []string{"echo", "\"hello world\""},
-		sidecarCommand: []string{"sh", "-c", "while [[ true ]] ; do echo \"hello from sidecar\" ; done"},
+		desc:        "A sidecar that runs forever is terminated when Steps complete",
+		stepCommand: []string{"echo", "\"hello world\""},
+		// trapping the exit signals lets us stop the sidecar more
+		// rapidly than waiting for a force stop.
+		sidecarCommand: []string{"sh", "-c", "trap exit SIGTERM SIGINT; while [[ true ]] ; do echo \"hello from sidecar\"; sleep 3 ; done"},
 	}, {
 		desc:           "A sidecar that terminates early does not cause problems running Steps",
 		stepCommand:    []string{"echo", "\"hello world\""},
@@ -57,16 +59,20 @@ func TestSidecarTaskSupport(t *testing.T) {
 	}}
 
 	ctx := context.Background()
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
-	clients, namespace := setup(ctx, t)
 	t.Parallel()
 
-	knativetest.CleanupOnInterrupt(func() { tearDown(ctx, t, clients, namespace) }, t.Logf)
-	defer tearDown(ctx, t, clients, namespace)
-
 	for i, test := range tests {
+		i := i
+		test := test
 		t.Run(test.desc, func(t *testing.T) {
+			t.Parallel()
+
+			ctx, cancel := context.WithCancel(ctx)
+			defer cancel()
+			clients, namespace := setup(ctx, t)
+			knativetest.CleanupOnInterrupt(func() { tearDown(ctx, t, clients, namespace) }, t.Logf)
+			defer tearDown(ctx, t, clients, namespace)
+
 			sidecarTaskName := fmt.Sprintf("%s-%d", sidecarTaskName, i)
 			sidecarTaskRunName := fmt.Sprintf("%s-%d", sidecarTaskRunName, i)
 			task := parse.MustParseTask(t, fmt.Sprintf(`
--- a/test/start_time_test.go
+++ b/test/start_time_test.go
@@ -28,10 +28,15 @@ import (
 
 // TestStartTime tests that step start times are reported accurately.
 //
-// It runs a TaskRun with 5 steps that each sleep 10 seconds, then checks that
-// the reported step start times are 10+ seconds apart from each other.
-// Scheduling and reporting specifics can result in start times being reported
-// more than 10s apart, but they shouldn't be less than 10s apart.
+// It runs a TaskRun with multiple steps that each sleep for several
+// seconds, then checks that the reported step start times are several
+// seconds apart from each other.  Scheduling and reporting specifics
+// can result in start times being later than expected, but they
+// shouldn't be earlier.
+//
+// The number of seconds between each step has a big impact on the total
+// duration of the test so smaller is better (while still supporting the
+// test's intended purpose).
 func TestStartTime(t *testing.T) {
 	ctx := context.Background()
 	ctx, cancel := context.WithCancel(ctx)
@@ -48,16 +53,16 @@ metadata:
 spec:
   taskSpec:
     steps:
-    - image: ubuntu
-      script: sleep 10
-    - image: ubuntu
-      script: sleep 10
-    - image: ubuntu
-      script: sleep 10
-    - image: ubuntu
-      script: sleep 10
-    - image: ubuntu
-      script: sleep 10
+    - image: busybox
+      script: sleep 2
+    - image: busybox
+      script: sleep 2
+    - image: busybox
+      script: sleep 2
+    - image: busybox
+      script: sleep 2
+    - image: busybox
+      script: sleep 2
 `, namespace)), metav1.CreateOptions{})
 	if err != nil {
 		t.Fatalf("Error creating TaskRun: %v", err)
@@ -74,6 +79,7 @@ spec:
 	if got, want := len(tr.Status.Steps), len(tr.Spec.TaskSpec.Steps); got != want {
 		t.Errorf("Got unexpected number of step states: got %d, want %d", got, want)
 	}
+	minimumDiff := 2 * time.Second
 	var lastStart metav1.Time
 	for idx, s := range tr.Status.Steps {
 		if s.Terminated == nil {
@@ -81,8 +87,8 @@ spec:
 			continue
 		}
 		diff := s.Terminated.StartedAt.Time.Sub(lastStart.Time)
-		if diff < 10*time.Second {
-			t.Errorf("Step %d start time was %s since last start, wanted >10s", idx, diff)
+		if diff < minimumDiff {
+			t.Errorf("Step %d start time was %s since last start, wanted > %s", idx, diff, minimumDiff)
 		}
 		lastStart = s.Terminated.StartedAt
 	}
--- a/test/v1alpha1/timeout_test.go
+++ b/test/v1alpha1/timeout_test.go
@@ -174,7 +174,9 @@ spec:
 // TestTaskRunTimeout is an integration test that will verify a TaskRun can be timed out.
 func TestTaskRunTimeout(t *testing.T) {
 	t.Parallel()
-	ctx, cancel := context.WithTimeout(context.Background(), timeout+2*time.Minute)
+	taskTimeout := 1 * time.Second
+	testTimeout := taskTimeout + (2 * time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), testTimeout)
 	defer cancel()
 	c, namespace := setup(ctx, t)
 
@@ -203,7 +205,7 @@ spec:
   taskRef:
     name: %s
   timeout: %s
-`, helpers.ObjectNameForTest(t), namespace, task.Name, 30*time.Second))
+`, helpers.ObjectNameForTest(t), namespace, task.Name, taskTimeout))
 	if _, err := c.TaskRunClient.Create(ctx, taskRun, metav1.CreateOptions{}); err != nil {
 		t.Fatalf("Failed to create TaskRun `%s`: %s", taskRun.Name, err)
 	}
--- a/test/v1alpha1/duplicate_test.go
+++ b/test/v1alpha1/duplicate_test.go
@@ -32,7 +32,7 @@ import (
 	"knative.dev/pkg/test/helpers"
 )
 
-// TestDuplicatePodTaskRun creates 10 builds and checks that each of them has only one build pod.
+// TestDuplicatePodTaskRun creates multiple builds and checks that each of them has only one build pod.
 func TestDuplicatePodTaskRun(t *testing.T) {
 	t.Parallel()
 	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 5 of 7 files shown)


vdemeester (approved):

sbwsg:
/hold

waiting for results from changes to sidecar tests

sbwsg:
The individual test times are shortened but the overall test run time is highly, highly variable. I'm trying a run now that won't execute the YAML examples in parallel with the other tests to see if it helps.

test/duplicate_test.go
```diff
@@ -32,7 +32,7 @@ import (
 	"knative.dev/pkg/test/helpers"
 )
 
-// TestDuplicatePodTaskRun creates 10 builds and checks that each of them has only one build pod.
```
bobcatfish, line 35: haha i love how the comment was out of sync

total nit: might make sense to drop the # from the comment entirely, e.g. "creates multiple builds (runs??)" vs "creates 5" - nothing to keep in sync

i also think it could be worth having a comment to explain why 5

test/examples_test.go
```diff
@@ -239,7 +239,7 @@ func TestYamls(t *testing.T) {
 }
 
 func testYamls(t *testing.T, baseDir string, createFunc createFunc, filter pathFilter) {
-	t.Parallel()
+	// t.Parallel()
```
bobcatfish: (im assuming this is temporary for the experiment you're running?)

test/sidecar_test.go
```diff
@@ -47,26 +47,32 @@ func TestSidecarTaskSupport(t *testing.T) {
 		stepCommand    []string
 		sidecarCommand []string
 	}{{
-		desc:           "A sidecar that runs forever is terminated when Steps complete",
-		stepCommand:    []string{"echo", "\"hello world\""},
-		sidecarCommand: []string{"sh", "-c", "while [[ true ]] ; do echo \"hello from sidecar\" ; done"},
+		desc:        "A sidecar that runs forever is terminated when Steps complete",
+		stepCommand: []string{"echo", "\"hello world\""},
+		// trapping the exit signals lets us stop the sidecar more
+		// rapidly than waiting for a force stop.
```
bobcatfish, line 53: one detail thats missing for me is why trapping the signal is faster - i assume it's b/c whatever the default handler is for the signal, it's slower? (i.e. the stop is still being forced regardless i assume in that the controller is trying to stop the sidecar, but its a matter of how quickly the signal is handled?)

test/sidecar_test.go
```diff
@@ -47,26 +47,32 @@ func TestSidecarTaskSupport(t *testing.T) {
...
 	t.Parallel()
 
-	knativetest.CleanupOnInterrupt(func() { tearDown(ctx, t, clients, namespace) }, t.Logf)
-	defer tearDown(ctx, t, clients, namespace)
-
 	for i, test := range tests {
+		i := i
+		test := test
 		t.Run(test.desc, func(t *testing.T) {
+			t.Parallel()
```
bobcatfish, line 68: do we still want to call t.Parallel above as well?

test/start_time_test.go
```diff
@@ -28,10 +28,10 @@ import (
 
 // TestStartTime tests that step start times are reported accurately.
 //
-// It runs a TaskRun with 5 steps that each sleep 10 seconds, then checks that
-// the reported step start times are 10+ seconds apart from each other.
+// It runs a TaskRun with 5 steps that each sleep 2 seconds, then checks that
+// the reported step start times are 2+ seconds apart from each other.
 // Scheduling and reporting specifics can result in start times being reported
-// more than 10s apart, but they shouldn't be less than 10s apart.
+// more than 2s apart, but they shouldn't be less than 2s apart.
```
bobcatfish, line 34: similar nit - keeping the specific numbers out of the commit is a bit easier than having to keep it in sync - and also it can be nice to try to explain where "2" is coming from

test/start_time_test.go
```diff
@@ -81,8 +81,8 @@ spec:
 			continue
 		}
 		diff := s.Terminated.StartedAt.Time.Sub(lastStart.Time)
-		if diff < 10*time.Second {
-			t.Errorf("Step %d start time was %s since last start, wanted >10s", idx, diff)
+		if diff < 2*time.Second {
```
bobcatfish, line 84: another nit: 2 is repeated so many times, is it worth putting into a var?

test/v1alpha1/timeout_test.go
```diff
@@ -174,6 +174,7 @@ spec:
 // TestTaskRunTimeout is an integration test that will verify a TaskRun can be timed out.
 func TestTaskRunTimeout(t *testing.T) {
 	t.Parallel()
+	timeout := 1 * time.Second
```
bobcatfish, line 177: seems a bit odd to me to create a timeout var and in the next line immediately use a different timeout?

bobcatfish (commented):
just lots of nits from me!

/approve

test/examples_test.go
sbwsg: Yeah, it's uncommented again now!

test/duplicate_test.go
sbwsg: :D

I've dropped the number from the func comment and added a comment to explain the tradeoff in time vs proving efficacy.

More generally this feels like a kind of pseudo-fuzz test and i wonder if this is indeed a different class of testing we should be doing out-of-band in a nightly or something.

test/sidecar_test.go
sbwsg: I wasn't entirely sure how t.Parallel in sub-tests interacted with other parallel tests but after reading the docs on [Parallel](https://pkg.go.dev/testing#T.Parallel) and [Run](https://pkg.go.dev/testing#T.Run) it _seems_ like all parallel tests are run together regardless whether they're sub-tests or top-level tests.  removed the outer t.Parallel() 👍

test/start_time_test.go
sbwsg: One tricky bit here is that `2` is repeated a lot in the YAML and then only twice in the code. Using var substitution in the YAML looks like this:

```go
parse.MustParseTaskRun(`
  steps:
  - image: busybox
    script: sleep %d
  - image: busybox
    script: sleep %d
  - image: busybox
    script: sleep %d
  - image: busybox
    script: sleep %d
  - image: busybox
    script: sleep %d
`, stepSleepSeconds, stepSleepSeconds, stepSleepSeconds, stepSleepSeconds, stepSleepSeconds)
```

Soooo then I write a func to generate the repeated steps and the code becomes this:

```go
parse.MustParseTaskRun(fmt.Sprintf(`
  steps:
%s
`, generateSteps(stepRepeats, stepSleepSeconds))

func generateSteps(repeats, sleepSeconds) string {
  stepYAML = fmt.Sprintf(`
  - image: busybox
    script: sleep %d
`, sleepSeconds)
  return strings.Repeat(stepYAML, repeats)
}
```

And then by this point I can't even read my own code anymore and figure out what it's doing. As gross as the repeated "2"s are I think i'm going to leave them in the YAML and just add a var for the diff check and error message.

test/v1alpha1/timeout_test.go
sbwsg: Good point, I've named the two different timeouts with vars to make it a bit clearer.

test/start_time_test.go
sbwsg: Updated the comment to remove specific numbers and describe the tradeoff involved.

sbwsg:
/hold cancel

I think that's all the feedback addressed, thanks a lot @bobcatfish !

sbwsg:
`workspace-in-sidecar` again. hopefully will be resolved by https://github.com/tektoncd/pipeline/pull/4349

test/duplicate_test.go
vdemeester: > More generally this feels like a kind of pseudo-fuzz test and i wonder if this is indeed a different class of testing we should be doing out-of-band in a nightly or something.

We probably need a different (new) class of testing indeed, and this test would most likely be part of it.

test/sidecar_test.go
vdemeester: > My working theory is that kubernetes treats the termination of the container as a graceful shutdown, so first sends sigterm and then waits 30 seconds before sending sigkill.

It might rely on the underlying container runtime, but that is most likely what you describe — at least that's how the docker runtime (and most likely containerd too) works when you ask it to stop, it sends SIGTERM, and wait for a given time for the container to stop itself, before sending SIGKILL.

test/sidecar_test.go
vdemeester: @sbwsg interesting. Because the parent test is not `t.Parallel` it won't be executed in parallel of the other though right ? Making those subtests be run in parallel of each other but not in parallel with "external" tests. Not sure if that's the expected behavior your are looking for or not ?

vdemeester (commented):
few nit comments, overall looks good 👼🏼

test/sidecar_test.go
sbwsg: > Because the parent test is not t.Parallel it won't be executed in parallel of the other though right ?

This is the part i wasn't sure about. Here's the docs for `t.Run()`:

> Run runs f as a subtest of t called name. It runs f in a separate goroutine and blocks until f returns or calls t.Parallel to become a parallel test.

it doesn't describe the behaviour of parent / child tests running in parallel. So I went back and read the rest of the testing package docs and found this quote from the intro:

> Subtests can also be used to control parallelism. A parent test will only complete once all of its subtests complete.

I _think_ this matches what you said but I am still having a hard time parsing the docs.  So I put back the `t.Parallel` in the parent test.  I want everything (external tests + sidecar subtests) to run all together.

sbwsg:
The integration tests hit a 1 minute timeout while running TestSidecarTaskSupport.  Interestingly the sidecar itself ran to completion; it was the step container that hit the timeout.  Possibly busy cluster preventing the step's process from getting any cpu time?

```
    sidecar_test.go:108: Creating TaskRun "sidecar-test-task-run-1"
    sidecar_test.go:114: Error waiting for TaskRun "sidecar-test-task-run-1" to finish: "sidecar-test-task-run-1" failed
    sidecar_test.go:146: Primary container has nil Terminated state or did not complete successfully. Actual Terminated state: &ContainerStateTerminated{ExitCode:2,Signal:0,Reason:Error,Message:,StartedAt:2021-11-05 15:35:39 +0000 UTC,FinishedAt:2021-11-05 15:36:35 +0000 UTC,ContainerID:docker://54e2efd0ad664c4b49aa98b2eb6d425f7e2d15fa7610dbe7b3857c28f86a1e5e,}
    sidecar_test.go:161: Either the primary or sidecar containers did not terminate
```

sbwsg:
Alpha integration failed on the `workspace-in-sidecar` task.  It seems like any pod we spin up with multiple containers / processes seems to be at perpetual risk of failing.

sbwsg:
Hmm, odd error failed this one. `cannot use generate name with apply` and `error executing 'kubectl create': exit status 1`.

Here's the full log snippet:

```
    init_test.go:135: Create namespace arendelle-p7vb5 to deploy to
    init_test.go:163: Verify SA "default" is created in namespace "arendelle-p7vb5"
    examples_test.go:144: exit status 1 Output: 2021/11/09 14:37:23 NOTICE!
        -----------------------------------------------------------------
        Passing kubectl global flags to ko directly is deprecated.
        
        Instead of passing:
            ko create ... --namespace=arendelle-p7vb5
        
        Pass kubectl global flags separated by "--":
            ko create ... -- --namespace=arendelle-p7vb5
        
        For more information see:
            https://github.com/google/ko/issues/317
        -----------------------------------------------------------------
        2021/11/09 14:37:24 Using base gcr.io/distroless/static:nonroot for github.com/tektoncd/pipeline/cmd/imagedigestexporter
        2021/11/09 14:37:24 Using base gcr.io/tekton-nightly/github.com/tektoncd/pipeline/git-init-build-base:latest for github.com/tektoncd/pipeline/cmd/git-init
        2021/11/09 14:37:25 Building github.com/tektoncd/pipeline/cmd/imagedigestexporter for linux/amd64
        2021/11/09 14:37:26 Building github.com/tektoncd/pipeline/cmd/git-init for linux/amd64
        2021/11/09 14:37:28 Publishing gcr.io/tekton-prow-6/tpipeline-e2e-img/imagedigestexporter-6e7c518e6125f31761ebe0b96cc63971:latest
        2021/11/09 14:37:28 existing blob: sha256:d11256aebf76d725bacc1698ed2360974762597bf1dfeb453d7aff2a320f1563
        2021/11/09 14:37:28 existing blob: sha256:e8614d09b7bebabd9d8a450f44e88a8807c98a438a2ddd63146865286b132d1b
        2021/11/09 14:37:28 existing blob: sha256:ec7dd97ae256bd6ed3c1994e17a1a3c487200a39c82668c12f56af27b96616b2
        2021/11/09 14:37:28 existing blob: sha256:1b165ac694f675d6a5ffeca65f677875a0314fecbc3a76ae0fd26b51c2277666
        2021/11/09 14:37:28 gcr.io/tekton-prow-6/tpipeline-e2e-img/imagedigestexporter-6e7c518e6125f31761ebe0b96cc63971:latest: digest: sha256:6b2da850592af55f642ea28fd1c1f339f04babea3768548658b6a1d862e46fb6 size: 956
        2021/11/09 14:37:28 Published gcr.io/tekton-prow-6/tpipeline-e2e-img/imagedigestexporter-6e7c518e6125f31761ebe0b96cc63971@sha256:6b2da850592af55f642ea28fd1c1f339f04babea3768548658b6a1d862e46fb6
        2021/11/09 14:37:29 Publishing gcr.io/tekton-prow-6/tpipeline-e2e-img/git-init-4874978a9786b6625dd8b6ef2a21aa70:latest
        2021/11/09 14:37:29 existing blob: sha256:ac3e2af1fa23678b856c49a69901b68c0e5c3d32a07bcce6581a00cbf5153c54
        2021/11/09 14:37:29 existing blob: sha256:836d89fcf4030cf95691ff36167f86b3042f405d34cede8e294c9c331caf78dd
        2021/11/09 14:37:29 existing blob: sha256:d11256aebf76d725bacc1698ed2360974762597bf1dfeb453d7aff2a320f1563
        2021/11/09 14:37:29 existing blob: sha256:a0d0a0d46f8b52473982a3c466318f479767577551a53ffc9074c9fa7035982e
        2021/11/09 14:37:29 existing blob: sha256:f63801898bbdd6a2e6b0e5463e3c409f333074bdca5c992c6e4c9467bc909407
        2021/11/09 14:37:29 existing blob: sha256:6004038c978aaf37e47524110d831942eaa1b7d6345ed1360c092d5a5bcf46c5
        2021/11/09 14:37:29 gcr.io/tekton-prow-6/tpipeline-e2e-img/git-init-4874978a9786b6625dd8b6ef2a21aa70:latest: digest: sha256:b20c8f07b7929b7e20cd8da15ab114b28ddba0b213c73d0d2db61be77f42402b size: 1328
        2021/11/09 14:37:29 Published gcr.io/tekton-prow-6/tpipeline-e2e-img/git-init-4874978a9786b6625dd8b6ef2a21aa70@sha256:b20c8f07b7929b7e20cd8da15ab114b28ddba0b213c73d0d2db61be77f42402b
        task.tekton.dev/unit-tests created
        task.tekton.dev/git-clone created
        task.tekton.dev/kaniko created
        task.tekton.dev/demo-deploy-kubectl created
        pipeline.tekton.dev/demo-pipeline created
        pipelinerun.tekton.dev/demo-pipeline-run-1 created
        error: from default-cluster-admin-: cannot use generate name with apply
        Error: error executing 'kubectl create': exit status 1
        2021/11/09 14:37:30 error during command execution:error executing 'kubectl create': exit status 1
```

vdemeester (approved):

vdemeester (commented):
/lgtm

Issue #3557, opened 2020-11-23
Tekton is duplicating task runs that make up a pipeline run


alan-ghelardi:
I noticed recently that around 10% of our pipeline runs have duplicated task runs.

Initially, I thought that there was a relation between this issue and the fact that we're running a replicated Tekton controller as explained [here](https://github.com/tektoncd/pipeline/issues/3404), but after looking into logs and historical data, I've figured out that the issue was already happening before we've replicated the controller. Also, the duplicated tasks are always being created by the same controller instance. One thing that came to me attention is that the duplicated tasks are being created in a Go routine that has a different traceid from the original ones:

|msg                                                                                 |knative.dev/key                   |knative.dev/traceid                 |pod                          |
|------------------------------------------------------------------------------------|----------------------------------|------------------------------------|-----------------------------|
|Creating a new TaskRun object rewards-ledger-run-mb4pl-lint-m964j                   |pipelines/rewards-ledger-run-mb4pl|990295d7-b79d-4353-ba75-20ba7df2945d|tekton-pipelines-controller-0|
|Creating a new TaskRun object rewards-ledger-run-mb4pl-unit-tests-zzfkw             |pipelines/rewards-ledger-run-mb4pl|990295d7-b79d-4353-ba75-20ba7df2945d|tekton-pipelines-controller-0|
|Creating a new TaskRun object rewards-ledger-run-mb4pl-integration-tests-62bwb      |pipelines/rewards-ledger-run-mb4pl|990295d7-b79d-4353-ba75-20ba7df2945d|tekton-pipelines-controller-0|
|Creating a new TaskRun object rewards-ledger-run-mb4pl-sachem-cb6l6                 |pipelines/rewards-ledger-run-mb4pl|990295d7-b79d-4353-ba75-20ba7df2945d|tekton-pipelines-controller-0|
|Creating a new TaskRun object rewards-ledger-run-mb4pl-toxic-combination-check-wrdn2|pipelines/rewards-ledger-run-mb4pl|990295d7-b79d-4353-ba75-20ba7df2945d|tekton-pipelines-controller-0|
|Creating a new TaskRun object rewards-ledger-run-mb4pl-build-kj4bg                  |pipelines/rewards-ledger-run-mb4pl|990295d7-b79d-4353-ba75-20ba7df2945d|tekton-pipelines-controller-0|
[...]

PR #4361, opened 2021-11-04
Use kmeta.ChildName for child resources

### LINKED ISSUE #4358: Duplicated pod for a single TaskRun
# Expected Behavior
To have a single pod for each TaskRun

# Actual Behavior
Two or more pods for a single TaskRun

We are noticing that on our cluster, but we didn't manage to reliably reproduce it.
We noticed that happens when the task fails.

# Additional Info

- Kubernetes version:

```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:11:29Z", GoVersion:"go1.16.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.20-eks-8c579e", GitCommit:"8c579edfc914f013ff48b2a2b2c1308fdcacc53f", GitTreeState:"clean", BuildDate:"2021-07-31T01:34:13Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version: v0.22.0

- Other info 

I checked the logs from the controller, there are no errors related, but we can see two `Successfully reconciled taskrun ...` messages for each pod.

afrittoli:
# Changes

When a reconciler generates a child resource, it risks creating a
duplicate resource because of stale informer cache. This may happen
for instance when two reconciliations of the same resource happen
almost at the same time.
To avoid this risk, children resources should use a name that is uniquely
associated to that of the parent, so that any attempt of recreation would
fail. Such failure, when it happens, is considered as a transient error,
and it does not cause the TaskRun or the PipelineRun to fail.

We use kmeta.ChildName to generate the names. This applies to Pods
generated by TaskRuns and TaskRuns and Runs generated by PipelineRuns.
ChildName combines a base name with a suffix, and if the resulting name is
too long it shortens it by truncating it and appending an hash of the
original name: https://pkg.go.dev/github.com/knative/pkg/kmeta#ChildName

Resources that are not k8s resources, such as steps or scripts, are not
affected by this change.

Fixes #4358

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Name of `Pods` and  `TaskRuns` created by Tekton controllers are now generated with [`kmeta.ChildName`](https://pkg.go.dev/github.com/knative/pkg/kmeta#ChildName):

> ChildName generates a name for the resource based upon the parent resource and suffix. If the concatenated name is longer than K8s permits the name is hashed and truncated to permit construction of the resource, but still keeps it unique. If the suffix itself is longer than 31 characters, then the whole string will be hashed and `parent|hash|suffix` will be returned, where parent and suffix will be trimmed to fit (prefix of parent at most of length 31, and prefix of suffix at most length 30).

The name of the `Pods` owned by a `TaskRun`  is univocally associated to the owning resource. If a `TaskRun` resource is deleted and created with the same name, the child `Pod` will be created with the same name as before.
The base format of the name is `<taskrun-name>-pod`. The name may vary as described above if it becomes too long.
In case of retries, starting from the first retry, the base format of the name is `<taskrun-name>-retry<N>` where `N` is the retry number.

The name of the `TaskRuns` and `Runs` owned by a `PipelineRun`  are univocally associated to the owning resource. If a `PipelineRun` resource is deleted and created with the same name, the child `TaskRuns` will be created with the same name as before.
The base format of the name is `<pipelinerun-name>-<pipelinetask-name>`. The name may vary as described above if it becomes too long.
```

### DIFF
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -547,7 +547,7 @@ conditions:
     type: Succeeded
 startTime: "2020-05-04T02:00:11Z"
 taskRuns:
-  triggers-release-nightly-frwmw-build-ng2qk:
+  triggers-release-nightly-frwmw-build:
     pipelineTaskName: build
     status:
       completionTime: "2020-05-04T02:10:49Z"
@@ -557,7 +557,7 @@ taskRuns:
           reason: Succeeded
           status: "True"
           type: Succeeded
-      podName: triggers-release-nightly-frwmw-build-ng2qk-pod-8vj99
+      podName: triggers-release-nightly-frwmw-build-pod
       resourcesResult:
         - key: commit
           resourceRef:
@@ -616,7 +616,7 @@ Skipped Tasks:
     Values:
       foo
 Task Runs:
-  pipelinerun-to-skip-task-run-this-task-r2djj:
+  pipelinerun-to-skip-task-run-this-task:
     Pipeline Task Name:  run-this-task
     Status:
       ...
@@ -626,6 +626,20 @@ Task Runs:
       Values:
         foo
 ```
+
+The name of the `TaskRuns` and `Runs` owned by a `PipelineRun`  are univocally associated to the owning resource.
+If a `PipelineRun` resource is deleted and created with the same name, the child `TaskRuns` will be created with the
+same name as before. The base format of the name is `<pipelinerun-name>-<pipelinetask-name>`. The name may vary
+according the logic of [`kmeta.ChildName`](https://pkg.go.dev/github.com/knative/pkg/kmeta#ChildName).
+
+Some examples:
+
+| `PipelineRun` Name       | `PipelineTask` Name          | `TaskRun` Name     |
+|--------------------------|------------------------------|--------------------|
+| pipeline-run             | task1                        | pipeline-run-task1 |
+| pipeline-run             | task2-0123456789-0123456789-0123456789-0123456789-0123456789 | pipeline-runee4a397d6eab67777d4e6f9991cd19e6-task2-0123456789-0 |
+| pipeline-run-0123456789-0123456789-0123456789-0123456789 | task3 | pipeline-run-0123456789-0123456789-0123456789-0123456789-task3 |
+| pipeline-run-0123456789-0123456789-0123456789-0123456789 | task2-0123456789-0123456789-0123456789-0123456789-0123456789 | pipeline-run-0123456789-012345607ad8c7aac5873cdfabe472a68996b5c |
 
 ## Cancelling a `PipelineRun`
 
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -22,6 +22,8 @@ import (
 	"path/filepath"
 	"strconv"
 
+	"knative.dev/pkg/kmeta"
+
 	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/pod"
@@ -296,16 +298,19 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec 
 	}
 	activeDeadlineSeconds := int64(taskRun.GetTimeout(ctx).Seconds() * deadlineFactor)
 
-	pod := &corev1.Pod{
+	podNameSuffix := "-pod"
+	if taskRunRetries := len(taskRun.Status.RetriesStatus); taskRunRetries > 0 {
+		podNameSuffix = fmt.Sprintf("%s-retry%d", podNameSuffix, taskRunRetries)
+	}
+	newPod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			// We execute the build's pod in the same namespace as where the build was
 			// created so that it can access colocated resources.
 			Namespace: taskRun.Namespace,
 			// Generate a unique name based on the build's name.
-			// Add a unique suffix to avoid confusion when a build
-			// is deleted and re-created with the same name.
-			// We don't use RestrictLengthWithRandomSuffix here because k8s fakes don't support it.
-			Name: names.SimpleNameGenerator.RestrictLengthWithRandomSuffix(fmt.Sprintf("%s-pod", taskRun.Name)),
+			// The name is univocally generated so that in case of
+			// stale informer cache, we never create duplicate Pods
+			Name: kmeta.ChildName(taskRun.Name, podNameSuffix),
 			// If our parent TaskRun is deleted, then we should be as well.
 			OwnerReferences: []metav1.OwnerReference{
 				*metav1.NewControllerRef(taskRun, groupVersionKind),
@@ -338,13 +343,13 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec 
 	}
 
 	for _, f := range transformers {
-		pod, err = f(pod)
+		newPod, err = f(newPod)
 		if err != nil {
-			return pod, err
+			return newPod, err
 		}
 	}
 
-	return pod, nil
+	return newPod, nil
 }
 
 // makeLabels constructs the labels we will propagate from TaskRuns to Pods.
--- a/pkg/pod/pod_test.go
+++ b/pkg/pod/pod_test.go
@@ -21,9 +21,13 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 	"testing"
 	"time"
+
+	"knative.dev/pkg/apis"
+	duckv1beta1 "knative.dev/pkg/apis/duck/v1beta1"
+
+	"knative.dev/pkg/kmeta"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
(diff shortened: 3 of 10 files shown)


imjasonh:
This looks great! What's WIP about it? (aside from unit tests 😈 )

afrittoli:
There may be something to check about retries:
```
        status:
          completionTime: "2021-11-05T12:16:21Z"
          conditions:
          - lastTransitionTime: "2021-11-05T12:16:21Z"
            message: 'failed to create task run pod "retry-pipeline-retry-me": pods "retry-pipeline-retry-me-pod"
              already exists. Maybe invalid TaskSpec'
            reason: CouldntGetTask
            status: "False"
            type: Succeeded
```

Also I need to make sure that failures like this are considered transient and do not cause taskruns or pipelineruns to fail.

afrittoli:
> There may be something to check about retries:
> 
> ```
>         status:
>           completionTime: "2021-11-05T12:16:21Z"
>           conditions:
>           - lastTransitionTime: "2021-11-05T12:16:21Z"
>             message: 'failed to create task run pod "retry-pipeline-retry-me": pods "retry-pipeline-retry-me-pod"
>               already exists. Maybe invalid TaskSpec'
>             reason: CouldntGetTask
>             status: "False"
>             type: Succeeded
> ```
> 
> Also I need to make sure that failures like this are considered transient and do not cause taskruns or pipelineruns to fail.

For retries, the `PipelineRun` controller saves the original `TaskRun` status in the list of retries, and it clears the `TaskRun` status, so that the `TaskRun` controllers sees a `TaskRun` that needs to be reconciled, and triggers a new `Pod` creation. The `Pod` name must include the number of retry, at least starting from the first retry on.

afrittoli:
I think the failures on `pull-tekton-pipeline-alpha-integration-tests` may be flakes.
/cc @imjasonh @mattmoor @vdemeester

pkg/pod/pod.go
```diff
@@ -296,16 +298,19 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec
 	}
 	activeDeadlineSeconds := int64(taskRun.GetTimeout(ctx).Seconds() * deadlineFactor)
 
-	pod := &corev1.Pod{
+	podNameSuffix := "-pod"
+	if taskRunRetries := len(taskRun.Status.RetriesStatus); taskRunRetries > 0 {
+		podNameSuffix = fmt.Sprintf("%s-retry%d", podNameSuffix, taskRunRetries)
+	}
+	newPod := &corev1.Pod{
```
afrittoli, line 305: Renamed to `newPod` as `pod` obfuscated the package name `pod`.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -550,7 +551,7 @@ func getConditionCheckName(taskRunStatus map[string]*v1beta1.PipelineRunTaskRunS
 			}
 		}
 	}
-	return names.SimpleNameGenerator.RestrictLengthWithRandomSuffix(fmt.Sprintf("%s-%s", trName, conditionRegisterName))
```
mattmoor, line 553: Is the names package still used anywhere?

afrittoli: Yes, it's still used to generate step names, script filenames, internal volumes mount-point names.

I'm not sure whether it's actually needed in all of those places, but I wanted to focus on k8s resources first in this PR - which is where we have experience issues. Which has the advantage of making this PR easier to approach and review too :)

pkg/pod/pod_test.go
```diff
@@ -1640,8 +1715,9 @@ func TestPodBuildwithAlphaAPIEnabled(t *testing.T) {
 				t.Fatalf("builder.Build: %v", err)
 			}
 
-			if !strings.HasPrefix(got.Name, "taskrun-name-pod-") {
-				t.Errorf("Pod name %q should have prefix 'taskrun-name-pod-'", got.Name)
+			expectedName := kmeta.ChildName(tr.Name, "-pod")
+			if d := cmp.Diff(expectedName, got.Name); d != "" {
+				t.Errorf("Pod name does not matc: %q", d)
```
sbwsg, line 1720: tiny nit: matc -> match

sbwsg:
/lgtm

Way better without all those randomized strings in the names!

pkg/pod/pod_test.go
afrittoli: Nice catch, fixed

sbwsg:
The `run-steps-as-non-root` Example failed with:

```
        >>> Container step-show-user-1001:
        Signal 23 (URG) caught by ps (3.3.16).
        ps:ps/display.c:66: please report this bug
``` 

It looks like this has been recently [fixed in `ps` by ignoring the signal](https://gitlab.com/procps-ng/procps/-/commit/b6ccf865f83f98478a814c1d79d17a0b5212cc6e) so probably nothing for us to do here. Re-running.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -572,8 +573,7 @@ func getRunName(runsStatus map[string]*v1beta1.PipelineRunRunStatus, ptName, prN
 			return k
 		}
 	}
-
-	return names.SimpleNameGenerator.RestrictLengthWithRandomSuffix(fmt.Sprintf("%s-%s", prName, ptName))
+	return kmeta.ChildName(prName, fmt.Sprintf("-%s", ptName))
```
jerop, line 576: looks like you've made the same change to `Runs` as well, please documentation for that like you've done for `TaskRuns` and `Pods`

adding that detail to the commit message and pr description would be great too!

jerop (commented):
thanks @afrittoli 🎉 

added a comment documenting the change to `Runs` as well

pritidesai:
Thanks @afrittoli for this PR 🙏  I think this also fixes (both were closed but I just reopened them since the issues are still seen with recent releases) :

https://github.com/tektoncd/pipeline/issues/3126
https://github.com/tektoncd/pipeline/issues/3557

pritidesai:
I might be missing something but can we please add some tests around the release note if they don't exist already:

```
> ChildName generates a name for the resource based upon the parent resource and suffix. If the concatenated name is longer than K8s permits the name is hashed and truncated to permit construction of the resource, but still keeps it unique. If the suffix itself is longer than 31 characters, then the whole string will be hashed and `parent|hash|suffix` will be returned, where parent and suffix will be trimmed to fit (prefix of parent at most of length 31, and prefix of suffix at most length 30).
```

Documenting some samples would be ideal for the folks creating `pipelineRuns` 🙏 .

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: Nice catch, thank you. Fixed.

afrittoli:
> I might be missing something but can we please add some tests around the release note if they don't exist already:
> 
> ```
> > ChildName generates a name for the resource based upon the parent resource and suffix. If the concatenated name is longer than K8s permits the name is hashed and truncated to permit construction of the resource, but still keeps it unique. If the suffix itself is longer than 31 characters, then the whole string will be hashed and `parent|hash|suffix` will be returned, where parent and suffix will be trimmed to fit (prefix of parent at most of length 31, and prefix of suffix at most length 30).
> ```
> 

Thank you - the logic of the `ChildName` function is tested on knative side, which is why I did not add tests.
I added some tests for that now so that if the logic is changed on the library side we will notice it and have a chance to update docs along with the tests accordingly.

> Documenting some samples would be ideal for the folks creating `pipelineRuns` 🙏 .

Good idea, done!

jerop (commented):
`examples_test.go:144: exit status 1 Output: Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "webhook.pipeline.tekton.dev": Post "https://tekton-pipelines-webhook.tekton-pipelines.svc:443/defaulting?timeout=10s": context deadline exceeded`

docs/pipelineruns.md
```diff
@@ -627,6 +627,20 @@ Task Runs:
...
+The name of the `TaskRuns` and `Runs` owned by a `PipelineRun`  are univocally associated to the owning resource.
+If a `PipelineRun` resource is deleted and created with the same name, the child `TaskRuns` will be created with the
+same name as before. The base format of the name is `<pipelinerun-name>-<pipelinetask-name>`. The name may vary
+according the logic of [`kmeta.ChildName`](https://pkg.go.dev/github.com/knative/pkg/kmeta#ChildName).
+
+Some examples:
+
+| `PipelineRun` Name       | `PipelineTask` Name          | `TaskRun` Name    |
+|--------------------------|------------------------------|-------------------|
+| pipeline-run             | task1                        | my-pipeline-task1 |
```
pritidesai, line 639: NIT: will the `taskRun` name be `pipeline-run-task1` instead of `my-pipeline-task1`?

docs/pipelineruns.md
```diff
@@ -627,6 +627,20 @@ Task Runs:
...
+| pipeline-run             | task2-0123456789-0123456789-0123456789-0123456789-0123456789 | pipeline-runee4a397d6eab67777d4e6f9991cd19e6-task2-0123456789-0 |
```
pritidesai, line 640: how are the dashes added when one or both of the resource names are long?

No dash after the `pipelineRun` name here and in the last example.

`pipeline-runee4a397d6eab67777d4e6f9991cd19e6-task2-0123456789-0` (pipelineRun name is `pipeline-run` and pipelineTask name is long)
`pipeline-run-0123456789-012345607ad8c7aac5873cdfabe472a68996b5c` (pipelineRun name and pipelineTask name are long)

pritidesai (approved):
thank you @afrittoli 

a couple of minor nits but otherwise looks great, thank you for fixing this 🙏

docs/pipelineruns.md
afrittoli: `ChildName` does not add a dashes.
The way we're using it is to prepend a dash to the `PipelineTask` name.

```golang
     return kmeta.ChildName(prName, fmt.Sprintf("-%s", ptName))
```

When both names are too long an hash of the whole thing is used so that specific dash may not show up in the final name.

docs/pipelineruns.md
afrittoli: Nice catch, thank you!

vdemeester (commented):
/lgtm

PR #4362, opened 2021-11-05
WIP Batch yaml examples in integration test runs to see if it affects timings

sbwsg:
# Changes
Looking at test run times for the example yamls in our CI it's noticeable
that many individual tests fluctuate in their runtime. One run they'll take more than
two minutes, other runs less than 50 seconds. One possible explanation
for this is that the cluster is becoming overwhelmed with so many pipelineruns
and taskruns executing all at once.

This commit tries batching the example yamls so that smaller sets are
processed at one time in the cluster.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/examples_test.go
+++ b/test/examples_test.go
@@ -241,7 +241,7 @@ func TestYamls(t *testing.T) {
 
 func testYamls(t *testing.T, baseDir string, createFunc createFunc, filter pathFilter) {
 	t.Parallel()
-	batchSize := 20
+	batchSize := 40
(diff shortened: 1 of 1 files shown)


sbwsg:
One more thing I'm noticing is that deleting the namespaces after each test takes a really long time.

sbwsg:
/close

Conclusion: Batching the tests in to groups slows down the overall test run time.

Issue #2974, opened 2020-07-20
Improve testing around git packages


sbwsg:
### Feature request

We have a number of packages that work with git in some way.  Examples: `pkg/credentials/gitcreds`, `pkg/git`, `cmd/git-init`. The testing around these packages is not stellar.  In particular, `pkg/git` could really use some good testing since it's heavily utilized by both the Git PipelineResource and the `git-clone` Catalog Task.  We don't have any tests covering its public funcs but they are responsible for quite a lot behind the scenes: creating and moving files, setting ssh configuration options, configuring tls validation, calling out to the `git` binary etc etc.

lbernick:
I was curious why we are not using a git lib like [go git](https://pkg.go.dev/github.com/go-git/go-git/v5@v5.4.2), so I looked into this today. I believe it would support the functionality needed in pipelines except for sparse checkout (https://github.com/go-git/go-git/issues/90).

dibyom:
We have made some improvements. Coverage is up to around 70% now. 
Closing this. Can open a new issue to track further improvements.

PR #4367, opened 2021-11-09
Improve test coverage for pkg/git

lbernick:
# Changes

Adds tests to pkg/git for fetching submodules, symlinking .ssh file, and some SSH checks.
Addresses #2974, although more tests (e.g. for pkg/credentials/git, or for refspecs and revisions) might be necessary to achieve desired level of coverage.

/kind cleanup

# Submitter Checklist

- [ X ] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [ X ] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [ X ] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [ X ] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/git/git.go
+++ b/pkg/git/git.go
@@ -75,9 +75,15 @@ type FetchSpec struct {
 
 // Fetch fetches the specified git repository at the revision into path, using the refspec to fetch if provided.
 func Fetch(logger *zap.SugaredLogger, spec FetchSpec) error {
-	if err := ensureHomeEnv(logger); err != nil {
+	homepath, err := homedir.Dir()
+	if err != nil {
+		logger.Errorf("Unexpected error getting the user home directory: %v", err)
 		return err
 	}
+	if os.Geteuid() == 0 {
+		homepath = "/root"
+	}
+	ensureHomeEnv(logger, homepath)
 	validateGitAuth(logger, pipeline.CredsDir, spec.URL)
 
 	if spec.Path != "" {
@@ -98,7 +104,7 @@ func Fetch(logger *zap.SugaredLogger, spec FetchSpec) error {
 		return err
 	}
 
-	hasKnownHosts, err := userHasKnownHostsFile(logger)
+	hasKnownHosts, err := userHasKnownHostsFile(homepath)
 	if err != nil {
 		return fmt.Errorf("error checking for known_hosts file: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


pkg/git/git_test.go
```diff
@@ -147,20 +156,105 @@ func Test_validateGitAuth(t *testing.T) {
 			credsDir, cleanup := createTempDir(t)
 			defer cleanup()
 			if tt.wantSSHdir {
-				os.MkdirAll(credsDir+"/.ssh", 0755)
+				err := os.MkdirAll(filepath.Join(credsDir, "/.ssh"), fileMode)
```
sbwsg, line 159: I think the `/` in `"/.ssh"` might not be needed because `filepath.Join` will add it. (similar comment for other instances below)

sbwsg (commented):
One small nit otherwise this looks great!

pkg/git/git.go
```diff
@@ -250,15 +250,10 @@ func ensureHomeEnvSSHLinkedFromPath(logger *zap.SugaredLogger, homeenv string, h
...
-	homepath, err := homedir.Dir()
-	if err != nil {
-		logger.Errorf("Unexpected error: getting the user home directory: %v", err)
-		return false, err
-	}
+func userHasKnownHostsFile(homepath string) (bool, error) {
 	f, err := os.Open(filepath.Join(homepath, sshKnownHostsUserPath))
 	if err != nil {
-		if os.IsNotExist(err) {
+		if errors.Is(err, fs.ErrNotExist) {
```
sbwsg, line 256: I'm not sure we can use `io/fs` because downstream the OpenShift build process is not on a new enough version of `go` yet.

Pinging @vdemeester for confirmation here since it might have changed.

pkg/git/git.go
```diff
@@ -269,7 +264,7 @@ func userHasKnownHostsFile(logger *zap.SugaredLogger) (bool, error) {
 
 func validateGitAuth(logger *zap.SugaredLogger, credsDir, url string) {
 	sshCred := true
-	if _, err := os.Stat(credsDir + "/.ssh"); os.IsNotExist(err) {
+	if _, err := os.Stat(filepath.Join(credsDir, ".ssh")); errors.Is(err, fs.ErrNotExist) {
```
sbwsg, line 267: 👍

pkg/git/git.go
vdemeester: Yeah it did not changed 😅 If we merge as-is, we (openshift-pipelines team) would need to caring a patch 😅

pkg/git/git_test.go
```diff
@@ -147,20 +158,107 @@ func Test_validateGitAuth(t *testing.T) {
 			credsDir, cleanup := createTempDir(t)
 			defer cleanup()
 			if tt.wantSSHdir {
-				os.MkdirAll(credsDir+"/.ssh", 0755)
+				err := os.MkdirAll(filepath.Join(credsDir, ".ssh"), fileMode)
+				if err != nil {
+					t.Errorf("Error creating SSH dir: %v", err)
```
vdemeester, line 163: should we "fatal" here as.. if the creation failed, we may not need to continue that subtest ?

pkg/git/git_test.go
```diff
@@ -147,20 +158,107 @@ func Test_validateGitAuth(t *testing.T) {
...
+			if tt.homeenvEqualsHomedir {
+				homeenv = homedir
+			} else {
+				homeenv, cleanup = createTempDir(t)
+				defer cleanup()
+			}
+			if tt.homeenvSet {
+				cleanup := setEnv("HOME", homeenv, t)
+				defer cleanup()
+			}
```
vdemeester, line 249: nit: not sure if that would be better, but this could be in the "test" case themself (as functions) instead of having a boolean for each.

vdemeester (commented):
On the usage of `"io/fs"`, as @sbwsg said, if we go with it, we enforce Go 1.16 to build the project.. which is.. not *yet* available in our (official supported RH) toolchain.

Overall looks good.
cc @chmouel

pkg/git/git_test.go
lbernick: I think I prefer as is but if you feel strongly I'm happy to change.

chmouel:
looking good thanks!

/lgtm

pkg/git/git_test.go
vdemeester: 😁 yeah it's a nit 😉

vdemeester (approved):
/meow

Issue #4071, opened 2021-07-02
task time exceeded `timeouts.tasks` when task retried


ornew:
# Expected Behavior

If the `timeouts.tasks` time is exceeded, the task will not be retried.

finally should always be executed.

# Actual Behavior

If the task is retried, the task time is exceeded timeouts.tasks.

In addition, if this causes the pipeline execution time to exceed `timeouts.pipeline`, finally is force timeouted.

# Steps to Reproduce the Problem

```yaml
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: please-say-bye
spec:
  tasks:
    - name: hi
      retries: 2
      taskSpec:
        steps:
          - name: hi
            image: alpine:3.12
            script: |
              echo 'hi'
              sleep 10
  finally:
    - name: bye
      taskSpec:
        steps:
          - name: bye
            image: alpine:3.12
            script: |
              echo 'bye'
```

```yaml
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  generateName: please-say-bye-
spec:
  timeouts:
    pipeline: 10s
    tasks: 5s
    finally: 5s
  pipelineRef:
    name: please-say-bye
```

```
❯ tkn -n pipelines tr ls --label tekton.dev/pipeline=please-say-bye

NAME                             STARTED          DURATION    STATUS
please-say-bye-tzjdt-bye-pft4x   25 seconds ago   1 second    Failed(TaskRunTimeout)
please-say-bye-tzjdt-hi-qqlgs    30 seconds ago   5 seconds   Failed(TaskRunTimeout)

❯ tkn -n pipelines pr desc please-say-bye-tzjdt
Name:              please-say-bye-tzjdt
Namespace:         pipelines
Pipeline Ref:      please-say-bye
Service Account:   default
Labels:
 tekton.dev/pipeline=please-say-bye

🌡️  Status

STARTED        DURATION     STATUS
1 minute ago   14 seconds   Failed

💌 Message

Tasks Completed: 2 (Failed: 2, Cancelled 0), Skipped: 0 (TaskRun "please-say-bye-tzjdt-bye-pft4x" failed to finish within "1s")

📦 Resources

 No resources

⚓ Params

 No params

📝 Results

 No results

📂 Workspaces

 No workspaces

🗂  Taskruns

 NAME                               TASK NAME   STARTED        DURATION    STATUS
[...]

Issue #2989, opened 2020-07-22
Allow finally tasks to execute after pipeline timeout


gpaul:
### Feature request

Tekton recently introduced "Finally Tasks" that execute once a Pipeline succeeds or fails. The [design document](https://docs.google.com/document/d/1lxpYQHppiWOxsn4arqbwAFDo4T0-LCqpNa6p-TJdHrw/) lists the following use cases:

* Cleanup cluster resources after finishing (with success/failure) integration tests (Dogfooding Scenario)
* Update Pull Request with what happened overall in the pipeline (pipeline level)
* Report Test Results at the end of the test pipeline (Notifications Scenario)

These are all legitimate use cases that show the value of the "Finally Tasks at the Pipeline Level" feature.

However, these tasks will not execute in the event that a Pipeline's execution duration exceeds its Timeout value. Pipelines can have unpredictable or highly variable execution times for various reasons: network transfers, access to third-party services, or delayed execution due to resource contention can all lead to a Pipeline unexpectedly exceeding its configured Timeout.

For this reason, it is currently not possible to rely on Finally tasks for any of the aforementioned use cases. They are an elegant optimisation, but do not provide a contract that can be relied upon in general.

This feature request is to extend the existing Finally Task functionality to allow Finally Tasks to execute even if the enclosing PipelineRun times out. This rounds out the Finally Task feature by addressing the common "PipelineRun timeout" edge case.

### Use case

Every reason that one wants Finally Tasks, with the added confidence that they will execute regardless of PipelineRun timeout.

dlorenc:
This is an interesting one - I agree there should be a way to guarantee "finally" runs, but should there be a timeout for that as well?

pritidesai:
This is an interesting indeed and I agree, we should be able to guarantee that `finally` tasks run even the `Pipeline` (that they part of) timeout. This can be configurable and disabled by default.

Today, `pipelineRun` timeout takes the highest precedence and exits with failure if its tasks not able to finish in time.

```
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  name: pipelinerun-with-timeout
spec:
  timeout: "0h0m3s"
  pipelineSpec:
    tasks:
      - name: echo-good-morning
        timeout: "0h0m30s"
        taskSpec:
          steps:
            - name: echo
              image: ubuntu
              script: |
                #!/usr/bin/env bash
                echo "Good Morning!"
                sleep 60
---
```

PipelineRun failure:

```
[...]

PR #4368, opened 2021-11-09
Stop executing retried `Tasks` past their timeouts

jerop:
# Changes

Today, when a `Task` is retried, it executes past its timeout. Even
worse, the `Task` executing past its timeout causes the execution of
`Finally Tasks` to be timedout before executing for as long as their 
allocated timeouts. As such, the execution of `Finally Tasks` is not 
guaranteed even when a `finally timeout` has been allocated.

@ornew described the problem in the issue report:
"The problem for me is that Tekton still doesn't guarantee the execution
of `finally` at all. This is a really serious problem for operating in
production, as Tekton we operate has caused a lot of resource leaks and
a lot of damage by not running finally. It's the biggest problem with
existing `timeout`."

This problem is caused by the the interaction between retries and 
timeouts. When a `TaskRun` is retried:
- The status of its failed execution is stored in `RetriesStatus`.
- The status of the `TaskRun` is updated, including marking it as
`Running` and resetting its `StartTime`.

Resetting the `StartTime` is useful in recording the `StartTime` of
each retry in the `RetriesStatus`. However, using the reset time as the 
reset start time of the `TaskRun` when checking whether it has timedout 
is incorrect and causes the issues described above.

In this change, we use the actual start time of the `TaskRun` to check
whether the `TaskRun` has timedout. We do this by checking the start 
time of previous attempts as well, instead of the current attempt only. 

Alternative approaches considered include:
- not resetting the start time of the `TaskRun` upon retry, however
it will be challenging to know the execution times of each retry.
- keeping track of the actual start time in an extra field, but this is
information that's already available in the status

References:
- [TEP-0046: Finally tasks execution post pipelinerun timeout](https://github.com/tektoncd/community/blob/main/teps/0046-finallytask-execution-post-timeout.md)
- [Issue: task time exceeded `timeouts.tasks` when task retried](https://github.com/tektoncd/pipeline/issues/4071)
- [Issue: Allow finally tasks to execute after pipeline timeout](https://github.com/tektoncd/pipeline/issues/2989)

Co-authored-by: Lee Bernick <lbernick@google.com>

/kind bug
/closes https://github.com/tektoncd/pipeline/issues/4071

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fix an issue where retried `Tasks` executed past the `timeouts`. Thereby, guaranteeing that `Finally Tasks` can execute for as long as the allocated `timeouts`. 
```

### DIFF
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -528,6 +528,8 @@ The `timeout` value is a `duration` conforming to Go's
 values are `1h30m`, `1h`, `1m`, and `60s`. If you set the global timeout to 0, all `PipelineRuns`
 that do not have an individual timeout set will fail immediately upon encountering an error.
 
+The above `timeouts` include the time used to execute `retries`. 
+
 ## Monitoring execution status
 
 As your `PipelineRun` executes, its `status` field accumulates information on the execution of each `TaskRun`
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -763,7 +763,7 @@ tasks:
 ### Configuring the failure timeout
 
 You can use the `Timeout` field in the `Task` spec within the `Pipeline` to set the timeout
-of the `TaskRun` that executes that `Task` within the `PipelineRun` that executes your `Pipeline.`
+of the `TaskRun` (and its `retries`) that executes that `Task` within the `PipelineRun` that executes your `Pipeline.`
 The `Timeout` value is a `duration` conforming to Go's [`ParseDuration`](https://golang.org/pkg/time/#ParseDuration)
 format. For example, valid values are `1h30m`, `1h`, `1m`, and `60s`.
 
--- a/docs/taskruns.md
+++ b/docs/taskruns.md
@@ -398,6 +398,8 @@ If a `TaskRun` runs longer than its timeout value, the pod associated with the `
 means that the logs of the `TaskRun` are not preserved. The deletion of the `TaskRun` pod is necessary in order to
 stop `TaskRun` step containers from running.
 
+The above `timeout` is inclusive of the time used to execute the `TaskRun`'s `retries`.
+
 ### Specifying `ServiceAccount` credentials
 
 You can execute the `Task` in your `TaskRun` with a specific set of credentials by
--- a/pkg/apis/pipeline/v1beta1/taskrun_types.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_types.go
@@ -385,7 +385,8 @@ func (tr *TaskRun) IsCancelled() bool {
 
 // HasTimedOut returns true if the TaskRun runtime is beyond the allowed timeout
 func (tr *TaskRun) HasTimedOut(ctx context.Context) bool {
-	if tr.Status.StartTime.IsZero() {
+	startTime := tr.getStartTime()
+	if startTime.IsZero() {
 		return false
 	}
 	timeout := tr.GetTimeout(ctx)
@@ -393,8 +394,24 @@ func (tr *TaskRun) HasTimedOut(ctx context.Context) bool {
 	if timeout == apisconfig.NoTimeoutDuration {
 		return false
 	}
-	runtime := time.Since(tr.Status.StartTime.Time)
+	runtime := time.Since(startTime.Time)
 	return runtime > timeout
+}
+
+func (tr *TaskRun) getStartTime() *metav1.Time {
+	startTime := tr.Status.StartTime
+	if startTime.IsZero() {
+		if len(tr.Status.RetriesStatus) == 0 {
+			return startTime
+		}
+		startTime = &metav1.Time{Time: time.Now()}
+	}
+	for _, retry := range tr.Status.RetriesStatus {
+		if retry.StartTime.Time.Before(startTime.Time) {
+			startTime = retry.StartTime
+		}
+	}
+	return startTime
 }
 
 // GetTimeout returns the timeout for the TaskRun, or the default if not specified
--- a/pkg/apis/pipeline/v1beta1/taskrun_types_test.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_types_test.go
@@ -309,12 +309,78 @@ func TestHasTimedOut(t *testing.T) {
 			},
 		},
 		expectedStatus: true,
+	}, {
+		name: "TaskRun timed out with retries",
+		taskRun: &v1beta1.TaskRun{
+			Spec: v1beta1.TaskRunSpec{
+				Timeout: &metav1.Duration{
+					Duration: 10 * time.Second,
+				},
+			},
+			Status: v1beta1.TaskRunStatus{
+				Status: duckv1beta1.Status{
+					Conditions: []apis.Condition{{
+						Type:   apis.ConditionSucceeded,
+						Status: corev1.ConditionFalse,
+					}},
+				},
+				TaskRunStatusFields: v1beta1.TaskRunStatusFields{
+					StartTime: &metav1.Time{Time: time.Now().Add(-5 * time.Second)},
+					RetriesStatus: []v1beta1.TaskRunStatus{{
+						Status: duckv1beta1.Status{
+							Conditions: []apis.Condition{{
+								Type:   apis.ConditionSucceeded,
+								Status: corev1.ConditionFalse,
+							}},
+						},
+						TaskRunStatusFields: v1beta1.TaskRunStatusFields{
+							StartTime:      &metav1.Time{Time: time.Now().Add(-11 * time.Second)},
+							CompletionTime: &metav1.Time{Time: time.Now().Add(-6 * time.Second)},
+						},
+					}},
+				},
+			},
+		},
+		expectedStatus: true,
+	}, {
+		name: "TaskRun timed out with retries but current execution start time is nil",
+		taskRun: &v1beta1.TaskRun{
+			Spec: v1beta1.TaskRunSpec{
+				Timeout: &metav1.Duration{
+					Duration: 10 * time.Second,
+				},
+			},
+			Status: v1beta1.TaskRunStatus{
+				Status: duckv1beta1.Status{
+					Conditions: []apis.Condition{{
+						Type:   apis.ConditionSucceeded,
+						Status: corev1.ConditionFalse,
+					}},
+				},
+				TaskRunStatusFields: v1beta1.TaskRunStatusFields{
+					StartTime: nil,
+					RetriesStatus: []v1beta1.TaskRunStatus{{
+						Status: duckv1beta1.Status{
+							Conditions: []apis.Condition{{
+								Type:   apis.ConditionSucceeded,
+								Status: corev1.ConditionFalse,
+							}},
+						},
+						TaskRunStatusFields: v1beta1.TaskRunStatusFields{
+							StartTime:      &metav1.Time{Time: time.Now().Add(-11 * time.Second)},
+							CompletionTime: &metav1.Time{Time: time.Now().Add(-6 * time.Second)},
+						},
+					}},
+				},
+			},
+		},
+		expectedStatus: true,
 	}}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			result := tc.taskRun.HasTimedOut(context.Background())
-			if d := cmp.Diff(result, tc.expectedStatus); d != "" {
+			if d := cmp.Diff(tc.expectedStatus, result); d != "" {
 				t.Fatalf(diff.PrintWantGot(d))
 			}
 		})


pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -385,18 +385,35 @@ func (tr *TaskRun) IsCancelled() bool {
...
 	// If timeout is set to 0 or defaulted to 0, there is no timeout.
 	if timeout == apisconfig.NoTimeoutDuration {
 		return false
 	}
-	runtime := time.Since(tr.Status.StartTime.Time)
+	runtime := time.Since(startTime.Time)
 	return runtime > timeout
 }
 
+func (tr *TaskRun) getStartTime() *metav1.Time {
```
lbernick, line 401: what's a bit confusing for me about this function is that it relies on RetriesStatus already having been updated, and now you have to rely on the `getStartTime` method instead of `Status.StartTime`. What could make more sense is updating the `tr.Status.StartTime` at the same time as `RetriesStatus` is updated, and then tr.Status.StartTime can be used directly. WDYT?

pkg/apis/pipeline/v1beta1/taskrun_types_test.go
```diff
@@ -309,12 +309,78 @@ func TestHasTimedOut(t *testing.T) {
...
+				},
+			},
+		},
+		expectedStatus: true,
 	}}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			result := tc.taskRun.HasTimedOut(context.Background())
-			if d := cmp.Diff(result, tc.expectedStatus); d != "" {
```
sbwsg, line 317: good catch 👍

jerop:

workspace in sidecar integration test failed, but it's known to be flaky - https://github.com/tektoncd/pipeline/issues/4169

jerop:
`examples/v1beta1/taskruns/authenticating-git-commands.yaml` failed with:

```
        Cloning into './repo'...
        Permission denied, please try again.
        Permission denied, please try again.
        fatal: Could not read from remote repository.
        
        Please make sure you have the correct access rights
        and the repository exists.
```
could it be flaky 🤔 

jerop:
`examples_test.go:144: exit status 1 Output: Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "webhook.pipeline.tekton.dev": Post "https://tekton-pipelines-webhook.tekton-pipelines.svc:443/defaulting?timeout=10s": context deadline exceeded` 

connection issues? test runner timedout?

jerop:
alpha integration tests FINALLY passed 💃🏾

sbwsg:
The prow machines might have gotten overloaded here.  Looking at metrics for the prow instances we peaked at 100% cpu utilization a couple times over the past hour:

image

sbwsg:
The `authenticating-git-commands` errors were very curious. The logs from the sidecars indicated that ssh server setup succeeded so I can't figure out why the step which pushes to the ssh server failed to do so. I am thinking of rewriting the example YAMLs which rely on sidecars as Integration Tests. This way we might have more precise control on the timeouts, error conditions and logs we can capture when they fail.

pkg/apis/pipeline/v1beta1/taskrun_types.go
lbernick: Oh ok I understand now, thanks! Can you comment this method to make that a bit more clear?

pritidesai: I agree with @lbernick, instead of iterating over the `RetriesStatus`, can we rely on `Status.StartTime` such that `Status.StartTime` is not updated with every retry and kept it as original start time?

jerop: @pritidesai that's one of the alternatives considered, but we need to know the execution time of each attempt 

> not resetting the start time of the `TaskRun` upon retry, however it will be challenging to know the execution times of each retry

jerop: > What could make more sense is updating the `tr.Status.StartTime` at the same time as `RetriesStatus` is updated

thought @lbernick was suggesting updating the `startTime` instead of keeping it as the original `startTime` - which is what we already do - please correct me if I misunderstood that

jerop:
@sbwsg thanks for investigating the integration tests failures!

pkg/apis/pipeline/v1beta1/taskrun_types.go
jerop: maybe we can go with the second alternative where we record the actual start time as an implementation detail and we use that to compare against the timeout in `HasTimedOut` function, instead of checking the `RetriesStatus`

> keeping track of the actual start time in an extra field, but this is information that's already available in the status

pritidesai: my understanding is, there are multiple `StartTime` available as part of the taskRun status:

*  `tr.Status.StartTime`
*  a list of `tr.Status.RetriesStatus[].StartTime`

Is `tr.Status.StartTime` changing with each new retry?

jerop: > Is `tr.Status.StartTime` changing with each new retry?

@pritidesai yes, that's correct 

when we do a retry, we add the status to retry history in line 771 then we clear the status is line 772

https://github.com/tektoncd/pipeline/blob/28f950700e99fd22a175eab7e2c803248675cca0/pkg/reconciler/pipelinerun/pipelinerun.go#L763-L779

when you look into what clear status does, you see that it sets the `tr.Status.StartTime` to `nil` in line 945
https://github.com/tektoncd/pipeline/blob/28f950700e99fd22a175eab7e2c803248675cca0/pkg/reconciler/pipelinerun/pipelinerun.go#L944-L948

docs/pipelineruns.md
```diff
@@ -528,6 +528,8 @@ The `timeout` value is a `duration` conforming to Go's
 values are `1h30m`, `1h`, `1m`, and `60s`. If you set the global timeout to 0, all `PipelineRuns`
 that do not have an individual timeout set will fail immediately upon encountering an error.
 
+The above `timeouts` include the time used to execute `retries`.
```
bobcatfish, line 531: might be handy to link to docs on retries in a pipeline task here, so folks would understand what "retries" means (esp if this is all totally new to them)

docs/pipelines.md
```diff
@@ -763,7 +763,7 @@ tasks:
 ### Configuring the failure timeout
 
 You can use the `Timeout` field in the `Task` spec within the `Pipeline` to set the timeout
-of the `TaskRun` that executes that `Task` within the `PipelineRun` that executes your `Pipeline.`
+of the `TaskRun` (and its `retries`) that executes that `Task` within the `PipelineRun` that executes your `Pipeline.`
```
bobcatfish, line 766: reading this doc and looking at the example below, it's not clear to me if `timeout: "0h1m30s" would be for each individual TaskRun (and each individual retry) or for the time taken in total across the TaskRun + Retries?

* if individual, maybe we could say something like "the time allowed for execution of an individual TaskRun, and for each of it's subsequent retries
* if it's the totaly, maybe something like "the total amount of time allowed for the execution of an individual taskRun and all of it's retries"
* 
(i'm actually not sure this is clearer .. haha this is hard... maybe we can get some input from @geriom XD)

docs/taskruns.md
```diff
@@ -398,6 +398,8 @@ If a `TaskRun` runs longer than its timeout value, the pod associated with the `
 means that the logs of the `TaskRun` are not preserved. The deletion of the `TaskRun` pod is necessary in order to
 stop `TaskRun` step containers from running.
 
+The above `timeout` is inclusive of the time used to execute the `TaskRun`'s `retries`.
```
bobcatfish, line 401: similar question to above: still not sure if this is the total time for taskrun + retries, or each retry will reset the counter

pkg/apis/pipeline/v1beta1/taskrun_types.go
bobcatfish: I also feel like it's a bit odd that `getStartTime` will return a different value than if you accessed the start time directly - it feels like we're changing the meaning of "start time" so that it now means 2 different things

on the other hand given the way retries are implemented (reusing the same Taskrun) this does make a sort of sense, and has the advantage of containing the logic for this check in one place...

A couple bits of possible food for thought:
* Runs are going to need this same logic if since iirc [TEP-0069](https://github.com/tektoncd/community/blob/main/teps/0069-support-retries-for-custom-task-in-a-pipeline.md) is going to use the same logic for Runs - which makes me wonder if it makes sense to put this logic somewhere that can be leveraged by both vs. as a function on the TaskRun object?
* What would this logic look like if we DID have multiple TaskRuns created when retrying? We wouldn't be able to overload the start time lookup of one individual taskRun, i think we'd need to look up the previous attempts and find the earliest start time (or keep that information somewhere else as suggested in the alternatives)

The above 2 thoughts have me thinking that TaskRun might not be the best place for this logic but I don't feel strongly about it!

bobcatfish (commented):
One thing that stands out to me is i dont see any integration (my pet name for "reconciler") or end to end tests for this? (Which might be part of why it wasnt caught in the first place) https://github.com/tektoncd/pipeline/blob/main/test/timeout_test.go seems like it might be a good candidate to add something?

afrittoli:
>  Today, when a Task is retried, it executes past its timeout. Even
worse, the Task executing past its timeout causes the execution of
Finally Tasks to be timedout before executing for as long as their
allocated timeouts. As such, the execution of Finally Tasks is not
guaranteed even when a finally timeout has been allocated.

If I understand correctly, in case of retries, a `Task` may execute past
the [timeout set at pipeline level for the `tasks` section of the pipeline](https://tekton.dev/docs/pipelines/pipelineruns/#configuring-a-failure-timeout).
But it's still correct that each retry is granted [its whole own timeout](https://tekton.dev/docs/pipelines/tasks/#specifying-a-timeout) for each retry, as long as this doesn't it the pipeline level set timeout.
Is this correct?

docs/pipelines.md
afrittoli: `Tasks` can define their own timeout, so my expectation here was that this was provided to override that at `Pipeline` level. It could be interpreted as a timeout for the sum of all retries of a `Task`, even though that feels a bit strange to me. Either case I agree we should make it crystal-clear in the docs what this timeout is about :)

docs/taskruns.md
afrittoli: I disagree that this should be the case. A `Task` author has no knowledge about retries, and when they set a timeout is definitely specific to the single execution of the `Task`.
A `TaskRun` may override that value to a different one specific to the runtime environment (slower/faster execution nodes for instance). 

When the `Task` is executed within a `PipelineRun` with retries, the `PipelineRun` author may want a mechanism to control the timeout for a `TaskRun` and its retries but I don't think we provide that today and I would see that as a new feature.

We do provide today a way for the `PipelineRun` author to control the maximum amount of time that can be spent in the "non-finally" part of the `Pipeline` (i.e. `timeouts.tasks`) which can be used to ensure that `finally` will run.

pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -385,18 +385,35 @@ func (tr *TaskRun) IsCancelled() bool {
...
+	startTime := tr.Status.StartTime
+	if startTime.IsZero() {
+		if len(tr.Status.RetriesStatus) == 0 {
+			return startTime
+		}
+		startTime = &metav1.Time{Time: time.Now()}
+	}
+	for _, retry := range tr.Status.RetriesStatus {
```
afrittoli, line 409: I don't think the `TaskRun` controller should ever use any data from the `RetriesStatus`.

Retries cannot be defined in a `TaskRun`, they are a `Pipeline` feature really. 
Since this feature has been implemented in the `PipelineRun` controller as a `TaskRun` status reset + new `Pod`, the `RetriesStatus` was added to the `TaskRun` status to hold an history of past execution for the benefit of the `PipelineRun` controller as well as end users.

I don't know the motivation behind implementing retries like this - I assume there was a good reason for not creating new `TaskRuns` instead. But the fact remains that the `RetriesStatus` is written and managed by the `PipelineRun` controller, and I think they should be completely transparent from a `TaskRun` controller point of view.

afrittoli (commented):
Thanks for working on this.

All the possible timeout scenarios are a bit mind-boggling 😅 

We definitely should honour the `PipelineRun` defined `timeout.tasks` introduced in TEP-0046, `Task` retries should not be allowed to take the execution of the non-finally part of the pipeline beyond that limit.

However retries are defined at pipeline level, so I think any timeout defined at `Task` or `TaskRun` level should refer to the single execution of the `Task` and not that of a collection of retries.

jerop:
/hold 

@lbernick is working on refactoring the timeouts logic in https://github.com/tektoncd/pipeline/pull/4447

Issue #4373, opened 2021-11-15
Extend breakout logic in stopSidecars


SaschaSchwarze0:
### Feature request

This is a follow-on issue to [this slack thread](https://tektoncd.slack.com/archives/CK3HBG7CM/p1636965530045300?thread_ts=1636722092.041100&cid=CK3HBG7CM) and pull request [Breakout sidecar stopping if TaskSpec has no sidecar #4372](https://github.com/tektoncd/pipeline/pull/4372). The pull request only optimizes for TaskRuns that have no sidecar in their TaskSpec. There is potentially also an optimization possible if the TaskSpec has sidecars and the TaskRun's `status.Sidecars` indicates that all stop logic did already run.

This would further reduce the need to retrieve the pod which can potentially fail if the pod has already been deleted.

### Use case

As Tekton seems to run periodic reconciliations for all TaskRuns, any optimization that prevents pod retrievals that can potentially fail (and therefore imo involve the API server), is a reasonable improvement for systems with many TaskRuns.

imjasonh:
The logic to stop sidecars is here: https://github.com/tektoncd/pipeline/blob/26176ce931b98ea22d3260be23d07dc117112509/pkg/pod/entrypoint.go#L224

It should be fairly simple to have a check that at least one of the `tr.Status.Sidecar` elements has a `Running` element (indicating that it's still running and needs to be stopped). A subsequent problem is that `stopSidecars` returns the `pod`

https://github.com/tektoncd/pipeline/blob/26176ce931b98ea22d3260be23d07dc117112509/pkg/reconciler/taskrun/taskrun.go#L144

This is apparently used for recording metrics, which indicates that a global resync will also re-emit latency metrics, which can be very misleading.

I have a partial WIP PR here that might help: https://github.com/imjasonh/pipeline/pull/new/avoid-stop-sidecars -- I'll add more tests and send it out.

PR #4374, opened 2021-11-15
Avoid API server call to get Pod when sidecars are stopped

imjasonh:
If the TaskRun status indicates that all sidecars are already stopped,
there's no need to hit the API server to get the Pod's status to attempt
to stop the sidecars.

This can avoid a lot of API calls, especially during global resync, when
TaskRuns' sidecars are already stopped.

This builds on @SaschaSchwarze0's #4372 (🙏) and avoids another case where we don't need to hit the API server.

Fixes https://github.com/tektoncd/pipeline/issues/4373

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Avoids calls to the API server to stop sidecars when sidecars are already stopped
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -131,24 +131,15 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, tr *v1beta1.TaskRun) pkg
 			return cloudEventErr
 		}
 
-		pod, err := c.stopSidecars(ctx, tr)
-		if err != nil {
+		if err := c.stopSidecars(ctx, tr); err != nil {
 			return err
 		}
 
 		go func(metrics *taskrunmetrics.Recorder) {
-			err := metrics.DurationAndCount(tr)
-			if err != nil {
+			if err := metrics.DurationAndCount(tr); err != nil {
 				logger.Warnf("Failed to log the metrics : %v", err)
 			}
-			if pod != nil {
-				err = metrics.RecordPodLatency(pod, tr)
-				if err != nil {
-					logger.Warnf("Failed to log the metrics : %v", err)
-				}
-			}
-			err = metrics.CloudEvents(tr)
-			if err != nil {
+			if err := metrics.CloudEvents(tr); err != nil {
 				logger.Warnf("Failed to log the metrics : %v", err)
 			}
 		}(c.metrics)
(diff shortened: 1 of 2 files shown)


SaschaSchwarze0:
@imjasonh as you had noticed in the issue, this means that [the pod latency](https://github.com/tektoncd/pipeline/blob/v0.28.2/pkg/reconciler/taskrun/taskrun.go#L154) will only get recorded for a TaskRun with sidecars. The best approach might be to stop returning a pod from `stopSidecar` as that side effect is not nice. An alternative place to record that metric could be after the TaskRun left the pending status, but not familiar enough with the codebase to assess right away if that's possible.

imjasonh:
Rebased on top of https://github.com/tektoncd/pipeline/pull/4372

imjasonh:
> @imjasonh as you had noticed in the issue, this means that [the pod latency](https://github.com/tektoncd/pipeline/blob/v0.28.2/pkg/reconciler/taskrun/taskrun.go#L154) will only get recorded for a TaskRun with sidecars. The best approach might be to stop returning a pod from `stopSidecar` as that side effect is not nice. An alternative place to record that metric could be after the TaskRun left the pending status, but not familiar enough with the codebase to assess right away if that's possible.

Yeah, that's pretty gross. I've moved `RecordPodLatency` elsewhere, closer to where we start the first step after observing all sidecars to have been started. This should make this happen less, while still giving us useful information about how long it takes Pods to schedule and start.

vdemeester (approved):

imjasonh:
Can anybody else lend an lgtm? @afrittoli @pritidesai @jerop 🙏

imjasonh:
> Can anybody else lend an lgtm? @afrittoli @pritidesai @jerop 🙏

Reminded of this issue again, can anybody take a look?

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -203,26 +194,38 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, tr *v1beta1.TaskRun) pkg
...
 		if reason == v1beta1.TaskRunReasonCancelled || reason == v1beta1.TaskRunReasonTimedOut {
-			return nil, nil
+			return nil
+		}
+	}
+
+	// do not continue if there are no Running sidecars.
+	needsStop := false
+	for _, ss := range tr.Status.Sidecars {
+		if ss.Running != nil {
```
pritidesai, line 221: What is difference between ss.Running v/s ss.Terminated? There is a function `IsSidecarStatusRunning` defined in package `pod`, can we use the same function here instead?

https://github.com/tektoncd/pipeline/blob/2ee13837eaee98ec19b5ca5e53e19d9e532433b6/pkg/reconciler/taskrun/taskrun.go#L231

pritidesai:
thanks @imjasonh, one minor comment ⬆️  otherwise the changes looks good to me 👍

pkg/reconciler/taskrun/taskrun.go
imjasonh: Excellent suggestion! Much nicer, thanks.

pritidesai:
thank you @imjasonh 

/lgtm

is easyCLA broken now? 🙃

imjasonh:
> is easyCLA broken now? 🙃 

It seems like it might just be me?

afrittoli:
> 



> > is easyCLA broken now? 🙃
> 
> It seems like it might just be me?

I imagine it might be a CCLA for RedHat, could it be that there is an issue on RedHat side?
You could file an issue at LFX EasyCLA: https://jira.linuxfoundation.org/plugins/servlet/theme/portal/4/create/143

imjasonh:

This latest one seemed like an unrelated timeout in the step-script example test.

PR #3554, opened 2020-11-23
Initialize load testing

yaoxiaoqi:
# Changes

This PR initializes load testing. It starts multiple taskruns and sees if they could
complete in defined time.

It also provides a script to run load testing. Users can use flag `--taskrun-num`
and `--timeout` to specify the amount of taskruns they want to run and the time
in which taskruns should complete.

To run the load testing:

`./test/load-test.sh --taskrun-num 50 --timeout 10m`

/kind feature

# Submitter Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Includes [tests](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if functionality changed/added)
- [ ] Includes [docs](https://github.com/tektoncd/community/blob/master/standards.md#principles) (if user facing)
- [x] Commit messages follow [commit message best practices](https://github.com/tektoncd/community/blob/master/standards.md#commit-messages)
- [x] Release notes block has been filled in or deleted (only if no user facing changes)

_See [the contribution guide](https://github.com/tektoncd/pipeline/blob/master/CONTRIBUTING.md) for more details._

Double check this list of stuff that's easy to miss:

- If you are adding [a new binary/image to the `cmd` dir](../cmd), please update
  [the release Task](../tekton/publish.yaml) to build and release this image.

## Reviewer Notes

If [API changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md) are included, [additive changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#additive-changes) must be approved by at least two [OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS) and [backwards incompatible changes](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-incompatible-changes) must be approved by [more than 50% of the OWNERS](https://github.com/tektoncd/pipeline/blob/master/OWNERS), and they must first be added [in a backwards compatible way](https://github.com/tektoncd/pipeline/blob/master/api_compatibility_policy.md#backwards-compatible-changes-first).

# Release Notes

```release-note
Initialize load testing
```

PR #4378, opened 2021-11-17
RFC: Basic performance test

guillaumerose:
I am trying to write performance test for Tekton. I am looking at finding good metrics to follow and good test scenario.
Here is one: I create N taskRuns per second and observe how long it takes for the controller to create the pod related to each taskRun.

At the beginning, I thought it will be done in few seconds but even at a small rate (2-3 taskRuns per second), after 3min, the controller takes 30 seconds between taskRun creation and pod creation. It seems to be better if I tune QPS and burst of the k8s client of the controller.

I hope I can turn this into a real test that can run in the CI. Let me know what you think, what I should improve!

### DIFF
--- a/test/performance_test.go
+++ b/test/performance_test.go
@@ -1,0 +1,153 @@
+// +build performance
+
+package test
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/flowcontrol"
+	knativetest "knative.dev/pkg/test"
+)
+
+const (
+	qps               = 1
+	burst             = qps
+	injectionDuration = 1 * time.Minute
+)
+
+func TestInject(t *testing.T) {
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	t.Helper()
+
+	cfg := clientConfig(t)
+	kubeClient, cs := clientSets(t, cfg)
+
+	rateLimitedCfg := clientConfig(t)
+	rateLimitedCfg.RateLimiter = flowcontrol.NewTokenBucketRateLimiter(qps, burst)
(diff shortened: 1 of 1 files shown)


imjasonh:
@guillaumerose Thanks for picking this up and especially for your interest in formalizing and automating the tests. I've definitely done similar load tests in the past without observing the slowness you're seeing, so we must have had a regression at some point. Running these tests regularly would help prevent those in the future. 👍

vdemeester (commented):
I agree with @imjasonh, I think it would be nice to have those running alongside with the nightly build for example. We can start with this very simple test and grow it iteratively.

bobcatfish:
I have a couple of requests:

1. That we prioritize setting up the automation to run these tests on a nightly basis (and make results visible) - if we're not able to do that I wouldn't want to add tests to the codebase that we don't exercise (this is why we didnt end up merging https://github.com/tektoncd/pipeline/pull/3554 )
2. I think it would be good to identify the scenarios we want to exercise in the tests up and agree up front before merging, [TEP-0036](https://github.com/tektoncd/community/blob/main/teps/0036-start-measuring-tekton-pipelines-performance.md) tries to start the process of defining what we need, @guillaumerose if you're at all interested in picking up that TEP and making a proposal around it you are more than welcome :D

guillaumerose:
Thanks all!

@bobcatfish I will add my thoughts in a PR in this TEP, then we can come back to this 👍

vdemeester:
> 1. That we prioritize setting up the automation to run these tests on a nightly basis (and make results visible) - if we're not able to do that I wouldn't want to add tests to the codebase that we don't exercise (this is why we didnt end up merging [Initialize load testing #3554](https://github.com/tektoncd/pipeline/pull/3554) )

@bobcatfish this is a bit chicken-and-egg problem here. We need something to execute on to setup the nightly runs, hence having this in would be first step, 2nd step would be to setup a nightly in dogfooding, and the steps after would be to refine and enhance.

test/performance_test.go
```diff
@@ -0,0 +1,153 @@
...
+	t.Helper()
+
+	cfg := clientConfig(t)
+	kubeClient, cs := clientSets(t, cfg)
+
+	rateLimitedCfg := clientConfig(t)
+	rateLimitedCfg.RateLimiter = flowcontrol.NewTokenBucketRateLimiter(qps, burst)
+	rateLimitedKubeClient, rateLimitedCS := clientSets(t, rateLimitedCfg)
+
+	wg := sync.WaitGroup{}
```
lbernick, line 42: I know there's some ongoing discussion about what exactly to benchmark, but I recently discovered `testing.B` and I'm wondering if it could be helpful here? It looks like some of timing + waitgroup logic could be replaced with `b.runParallel`.

vdemeester: Ah I though @guillaumerose used `testing.B`.., but turns out not. Yes, maybe using it might make sense indeed.

PR #4387, opened 2021-11-18
Refactor CloudEvent related tests to use common set of helpers

lbernick:
Prior to this change, the event-related testing helpers were copied
and pasted in several files. The TaskRun Reconcile tests were updated
to not expect cloud events in a particular order, reducing flakiness,
but the PipelineRun reconcile tests and the Event tests did not receive
the same updates.

This commit moves event-related test helpers into a common package
and updates all CloudEvent tests to not expect events to occur in a given order.
This should address some of the causes of flakes in #2992 but not all.

Co-authored-by: Scott Seaward sbws@google.com

# Changes

/kind cleanup
/kind flake

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/events/cloudevent/testing.go
+++ b/pkg/reconciler/events/cloudevent/testing.go
@@ -1,0 +1,128 @@
+/*
+Copyright 2021 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cloudevent
+
+import (
+	"fmt"
+	"regexp"
+	"testing"
+	"time"
+
+	"k8s.io/client-go/tools/record"
+)
+
+// CheckEvents checks that the events received by the FakeRecorder are the same as wantEvents,
+// in the same order.
+func CheckEvents(t *testing.T, fr *record.FakeRecorder, testName string, wantEvents []string) error {
+	t.Helper()
+	err := eventsFromChannel(fr.Events, wantEvents)
+	if err != nil {
+		return fmt.Errorf("error in test %s: %v", testName, err)
+	}
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


afrittoli (commented):
Oh, nice, thank you for this!
/approve

pkg/reconciler/events/cloudevent/testing.go
```diff
@@ -0,0 +1,113 @@
+package cloudevent
```
pritidesai, line 1: please add the copyright license!

lbernick: done, thanks for catching this!

pkg/reconciler/events/cloudevent/testing.go
```diff
@@ -0,0 +1,128 @@
+/*
+Copyright 2020 The Tekton Authors
```
sbwsg, line 2: nit: 2021

pkg/reconciler/events/cloudevent/testing.go
```diff
@@ -0,0 +1,128 @@
...
+	"regexp"
+	"testing"
+	"time"
+
+	"k8s.io/client-go/tools/record"
+)
+
+// CheckEvents checks that the events received by the FakeRecorder are the same as wantEvents,
+// in the same order.
+func CheckEvents(t *testing.T, fr *record.FakeRecorder, testName string, wantEvents []string) error {
```
sbwsg, line 29: just realised that `cloudevents.CheckEvents` is not checking cloudevents, which seems like a bit of a misnomer given the package's name. Suggest moving this func to the `pkg/reconciler/events` package if possible.

lbernick: I agree that ideally we'd have CheckEvents in the events package and CheckCloudEvents in the cloudevents package. However it looks like the events package is already doing some [cloudevent related things](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/events/event.go#L45-L49) and relies on [checkCloudEvents](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/events/event_test.go#L274-L277) in tests, which would cause a circular import if I split up tests in this way. 
It seems like the events package may just need to be refactored but that's probably out of scope for this PR-- I'll create an issue to track.

bobcatfish:
hey @sbwsg @lbernick it looks like this is adding test code that will be publicly exported as part of the cloudevent package, am i understanding right? if so can we move this so it is either in a testing specific package, or not publicly exported?

sbwsg:
@bobcatfish Sure thing, moved under `test/`: https://github.com/tektoncd/pipeline/pull/4405

Issue #4116, opened 2021-07-23
Pipelinerun fails with CouldntGetTask (etcdserver: leader changed)


tibcoplord:
# Expected Behavior

Pipelinerun continues to run

# Actual Behavior

Pipelinerun reports CouldntGetTask, for example -

```
$ tkn pr list
NAME                     STARTED        DURATION     STATUS
file-datasink-r5rkd      15 hours ago   ---          Running
file-datasource-c7fmf    15 hours ago   ---          Running
file-datasink-xbbdb      17 hours ago   ---          Running
file-datasource-986c5    17 hours ago   ---          Running
scoringpipeline-l7zq7    22 hours ago   ---          Running
file-datasink-56tbw      22 hours ago   ---          Running
file-datasource-hmvnr    22 hours ago   40 minutes   Failed(CouldntGetTask)
```

The pipelinerun file-datasource-hmvnr consists of 3 tasks -

* install-datachannel
* wait
* uninstall-datachannel

All exist

The pipeline run reports a detailed status including -

```
🌡️  Status

STARTED        DURATION     STATUS
23 hours ago   40 minutes   Failed(CouldntGetTask)

💌 Message

Pipeline modelops/install-datachannel can't be Run; it contains Tasks that don't exist: Couldn't retrieve Task "uninstall-datachannel": etcdserver: leader changed

...
🗂  Taskruns

 NAME                                                TASK NAME             STARTED        DURATION     STATUS
 ∙ file-datasource-hmvnr-wait-5pns5                  wait                  23 hours ago   ---          Running
 ∙ file-datasource-hmvnr-install-datachannel-kcvwc   install-datachannel   23 hours ago   11 seconds   Succeeded
```

So the final task, uninstall-datachannel, cannot be executed.

# Steps to Reproduce the Problem

Unknown at this point

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:59:11Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #4392, opened 2021-11-23
Retry TaskRef resolution on etcd errors

lbernick:
# Changes

Prior to this change, querying etcd for a Task referenced by a TaskRun
could fail due to an error "etcdserver: leader changed", as reported in #4116.
I believe this error is transient.

This error doesn't appear to be wrapped in any error type exported by
the Kubernetes Go client or Knative, so I've reported this upstream to
the Go client in #kubernetes/kubernetes/106491.
To address the user issue until the upstream issue can be resolved,
this commit uses string comparison to detect and retry the error.
A similar strategy is used in Knative for retrying this error
during their integration tests (see
https://github.com/knative/pkg/blob/517ef0292b5362ec316d227a866821bc31ec99a8/reconciler/retry.go#L50-L66).

We don't currently have a strategy for reproducing this error, so it's
not known whether this commit will solve the problem.
We will have to see if this error continues to occur after this commit is released.

Co-authored-by: Dibyo Mukherjee dibyajyoti@google.com @dibyom 
Co-authored-by: Scott Seaward sbws@google.com @sbwsg 

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Retry Task resolution on transient etcd errors
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -294,6 +294,9 @@ func (c *Reconciler) resolvePipelineState(
 			task, providedResources,
 		)
 		if err != nil {
+			if tresources.IsGetTaskErrTransient(err) {
+				return nil, err
+			}
 			switch err := err.(type) {
 			case *resources.TaskNotFoundError:
 				pr.Status.MarkFailed(ReasonCouldntGetTask,
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -19,6 +19,7 @@ package pipelinerun
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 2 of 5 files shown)


sbwsg:
Something I'm not sure about: Does the taskrun or pipelinerun automatically get requeued for processing again if the returned error isn't permanent?  If not, is there a risk that the tr or pr ends up just "hanging around" waiting for further updates to trigger re-reconcile?

afrittoli (commented):
Thank for this change @lbernick.
I like the approach to add a new function - so we can add to it when we find more transient errors to be handled. 

I agree with @sbwsg that is would be nice to name it something that starts with `Is`. 

/approve

lbernick:
> Something I'm not sure about: Does the taskrun or pipelinerun automatically get requeued for processing again if the returned error isn't permanent? If not, is there a risk that the tr or pr ends up just "hanging around" waiting for further updates to trigger re-reconcile?

My understanding is that the taskrun/pipelinerun will be requeued if the error returned is not permanent (based on the sentence "If event is an error ... the error will be returned from Reconciler and key will requeue back into the reconciler key queue"  in the [knative generated reconciler docs](https://github.com/knative/pkg/tree/main/injection#generated-reconciler-responsibilities) and also [this commit](https://github.com/tektoncd/pipeline/commit/ce1b7b5a2703faa83e477910fe3aa22c30cdc638)).

sbwsg:
Awesome, thanks a lot!

/lgtm

PR #4400, opened 2021-11-29
Add some switches to the e2e script ⚙️

afrittoli:
# Changes

Add three switches (env variables) in the main e2e script, that
can be used to:
- skip setting up the test environment
- skip running the YAML based tests
- set the feature gate (this is also available as ENV var

They are useful for setting up the E2E tests in KinD clusters, as
well as generally useful for development purposes.

Add an env file which is used by the kind e2e script, see
https://github.com/tektoncd/plumbing/pull/927 for reference

Signed-off-by: Andrea Frittoli <andrea.frittoli@uk.ibm.com>

/kind misc

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

```release-note
NONE
```

PR #4401, opened 2021-12-01
Let Kaniko E2E test work with a KO_DOCKER_REPO env

afrittoli:
# Changes

If test runner set a KO_DOCKER_REPO variable, use it, so we run the tests
against an external container registry. If not set, the kaniko tests
will spin up its own local registry. Bring back the GCP secret support
so this works in the existing CI, but also supports the case where
KO_DOCKER_REPO points to a container registry where no secret required,
like in the kind based tests.

Add a "test_setup" function to the e2e script, which is picked up after
the setup of the cluster, which provisions a GCP service account to be
used by tests via the GCP_SERVICE_ACCOUNT_KEY_PATH env variable.

Since the local registry runs on HTTP (not HTTPS), the local registry
approach does not work for the helm test as the kubelet tries to use
HTTPS. If KO_DOCKER_REPO is not specified we either fail or skip the
test (no change in behaviour) depending on the value of missingKoFatal

This setup makes it easier to run E2E tests in different environments,
where a registry may or may not be available.

/kind misc

# Submitter Checklist

- [-] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -19,15 +19,23 @@
 
 source $(git rev-parse --show-toplevel)/test/e2e-common.sh
 
-# Script entry point.
+function test_setup() {
+  # Setup a service account for kaniko and gcs / storage artifact tests
+  gcloud config set project "${E2E_PROJECT_ID}"
 
-initialize $@
+  # create the service account
+  ACCOUNT_NAME=${E2E_PROJECT_ID}
+  gcloud iam service-accounts create $ACCOUNT_NAME --display-name $ACCOUNT_NAME
+  EMAIL=$(gcloud iam service-accounts list | grep $ACCOUNT_NAME | awk '{print $2}')
 
-header "Setting up environment"
+  # add the storage.admin policy to the account so it can push containers
+  gcloud projects add-iam-policy-binding ${E2E_PROJECT_ID} --member serviceAccount:$EMAIL --role roles/storage.admin
 
-install_pipeline_crd
+  # create the JSON key
+  gcloud iam service-accounts keys create config.json --iam-account $EMAIL
 
-failed=0
+  export GCP_SERVICE_ACCOUNT_KEY_PATH="$PWD/config.json"
+}
 
 function set_feature_gate() {
   local gate="$1"
@@ -52,6 +60,16 @@ function run_e2e() {
   go_test_e2e -tags=examples -timeout=20m ./test/ || failed=1
 }
 
+# Script entry point.
+
+initialize $@
+
+header "Setting up environment"
+
+install_pipeline_crd
+
+failed=0
+
 if [ "$PIPELINE_FEATURE_GATE" == "" ]; then
   set_feature_gate "stable"
   run_e2e
@@ -59,7 +77,6 @@ else
   set_feature_gate "$PIPELINE_FEATURE_GATE"
   run_e2e
 fi
-
 
 (( failed )) && fail_test
 success
--- a/test/helm_task_test.go
+++ b/test/helm_task_test.go
@@ -41,10 +41,12 @@ var (
 // TestHelmDeployPipelineRun is an integration test that will verify a pipeline build an image
 // and then using helm to deploy it
 func TestHelmDeployPipelineRun(t *testing.T) {
-	repo := ensureDockerRepo(t)
+	repo := ensureDockerRepo(t, "helmtasktest")
+
 	ctx := context.Background()
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
+
 	c, namespace := setup(ctx, t)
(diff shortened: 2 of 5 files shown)


test/ko_test.go
```diff
@@ -22,32 +22,15 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"testing"
 )
 
-var (
-	// Wether missing KO_DOCKER_REPO environment variable should be fatal or not
-	missingKoFatal = "true"
```
vdemeester, line 30: This was put there, as is, to enable us to skip those on our OpenShift CI, while making it fail on "normal" CI.

afrittoli: We could have a flag / environment variable about skipping vs failing in case of missing. 
I don't see the value of having it hardcoded in there?

vdemeester: It's not "harcoded" per say as.. it's used with `-ldflags '-X github.com/tektoncd/pipeline/test/v1alpha1.missingKoFatal=false' `. An environment variable would work to I think.

afrittoli: ok - I see. 

For the kaniko tests we do provide the option of setting up a registry in a sidecar, so a missing KO_DOCKER_REPO is not fatal. The way it is now the sidecar is always setup, but I think we should really do that only if no KO_DOCKER_REPO is provided. 

So, bottom line, I think the missing KO_DOCKER_REPO should never be globally "true", which is why I deleted it there.
Another way we could do this is to leave the two functions in place, and have the kaniko test use `GetDockerRepo` and the helm test use `EnsureDockerRepo`.

I'll make an update.

afrittoli:
Creating a service account with the right permissions to run the tests does not seem to work:
```
=================================
==== SETTING UP TEST CLUSTER ====
=================================
- Project is tekton-prow-8
- Cluster is gke_tekton-prow-8_us-central1_tpipeline-e2e-cls1466744752903819268
- User is prow-account@tekton-releases.iam.gserviceaccount.com
- Docker is gcr.io/tekton-prow-8/tpipeline-e2e-img
Updated property [core/project].
WARNING: You do not appear to have access to project [tekton-prow-8] or it does not exist.
Created service account [tekton-prow-8].
ERROR: (gcloud.projects.add-iam-policy-binding) User [prow-account@tekton-releases.iam.gserviceaccount.com] does not have permission to access projects instance [tekton-prow-8:getIamPolicy] (or it may not exist): Cloud Resource Manager API has not been used in project 574248271492 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/cloudresourcemanager.googleapis.com/overview?project=574248271492 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
- '@type': type.googleapis.com/google.rpc.Help
  links:
  - description: Google developers console API activation
    url: https://console.developers.google.com/apis/api/cloudresourcemanager.googleapis.com/overview?project=574248271492
- '@type': type.googleapis.com/google.rpc.ErrorInfo
  domain: googleapis.com
  metadata:
    consumer: projects/574248271492
    service: cloudresourcemanager.googleapis.com
  reason: SERVICE_DISABLED
ERROR: (gcloud.iam.service-accounts.keys.create) FAILED_PRECONDITION: Key creation is not allowed on this service account.
- '@type': type.googleapis.com/google.rpc.PreconditionFailure
  violations:
  - description: Key creation is not allowed on this service account.
    subject: projects/tekton-prow-8/serviceAccounts/tekton-prow-8@tekton-prow-8.iam.gserviceaccount.com?configvalue=tekton-prow-8%40tekton-prow-8.iam.gserviceaccount.com
    type: constraints/iam.disableServiceAccountKeyCreation
================================
```

nikhil-thomas:
> Sounds ok, I just need to validate that it works fine on our OpenShift CI 😅

I think this would be ok as:
1. at present, we use a separate e2e test script (but i agree, this is not ideal and we should start using this script as soon as we can).
2. we will not touch the `ìnitialize` part any way for tests on OpenShift. The switches in PR: https://github.com/tektoncd/pipeline/pull/4400 will make this easy to disable while testing on OpenShift
3. we run tests with `github.com/tektoncd/pipeline/test/v1alpha1.missingKoFatal=false` [here](https://github.com/openshift/tektoncd-pipeline/blob/5e6c929eb9935157581e98a24c680a903b80e8fb/openshift/e2e-tests-openshift.sh#L77)

cc @barthy1

barthy1:
just fyi I've tested the proposed code with s390x nightly pipeline test suite (pure k8s, KO_DOCKER_REPO is specified and remote, registry is HTTP). Everything works perfectly.

Issue #3323, opened 2020-10-02
Json tags on ArrayOrString are misleading


eddycharly:
https://github.com/tektoncd/pipeline/blob/78f54793880f4494e6d5748d7c27cb8633164148/pkg/apis/pipeline/v1beta1/param_types.go#L92-L96

Json tags here are misleading as `ArrayOrString` uses custom marshalling.
It confuses tools like kubebuilder and probably other tools potentially leading to incorrect generated code.

sbwsg:
We should take a look at kubernete's IntOrString type for comparison here.

sbwsg:
Looks like they don't have tags for json: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/intstr/intstr.go#L40

eddycharly:
Nope, k8s tooling usually has some special code for handling the IntOrString case (some specific open api tag, x-int-or-string, something like this).

From what I understand it’s not possible to represent the ArrayOrString pattern with the structural schema définition constraint that Kubernetes adds on top of open api.

eddycharly:
I feel like the only think we can do here if we don’t want to break compatibility is to ignore the value field and add the preserve unknown field tag :(

eddycharly:
For more context : the doc about structural schema definition (particularly the IntOrString struct) https://kubernetes.io/blog/2019/06/20/crd-structural-schema/#extensions.

And the way controller-tools takes care of IntOrString : https://github.com/kubernetes-sigs/controller-tools/blob/2f203e490e8af05e988fdac460151ff3a93f3ee7/pkg/crd/known_types.go#L91-L100

eddycharly:
More docs https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#intorstring

And a simple example:
```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: my-crds.my-group.io
spec:
  group: my-group.io
  names:
    kind: my-crd
    plural: my-crds
  scope: Namespaced
  versions:
    - name: v1alpha1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            test:
              oneOf:
                - type: integer
                - type: array
                  items: string

```

Will result in:
```
The CustomResourceDefinition "my-crds.my-group.io" is invalid: 
[...]

PR #4402, opened 2021-12-01
Add listType annotations

abayer:
# Changes

We need to have a `// +listType=...` annotation on all list fields in API types
to get `k8s.io/kube-openapi/cmd/openapi-gen` to not throw rules violations for
list fields without that annotation. I chose to use `atomic` because it seemed
the simplest option - it means that when updates are written, the whole list
will be updated, rather than individual elements in the list, and without any
special logic on uniqueness within the list or ordering by some field in the
structs in the list.

See https://github.com/kubernetes/kube-openapi/blob/3ee0da9b0b4211c407396d9c233b38b77ce19773/pkg/idl/doc.go#L21-L63 for more information on `listType`s.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

Not sure what to do re: release note here. =)

### DIFF
--- a/hack/update-openapigen.sh
+++ b/hack/update-openapigen.sh
@@ -17,15 +17,39 @@
 set -o errexit
 set -o nounset
 
+source $(git rev-parse --show-toplevel)/vendor/github.com/tektoncd/plumbing/scripts/library.sh
 source $(git rev-parse --show-toplevel)/hack/setup-temporary-gopath.sh
+
+readonly TMP_DIFFROOT="$(mktemp -d ${REPO_ROOT_DIR}/tmpdiffroot.XXXXXX)"
+
+cleanup() {
+  rm -rf "${TMP_DIFFROOT}"
+  shim_gopath_clean
+}
+
+cleanup
+
 shim_gopath
-trap shim_gopath_clean EXIT
+mkdir -p "${TMP_DIFFROOT}"
+
+trap cleanup EXIT
 
 echo "Generating OpenAPI specification ..."
 go run k8s.io/kube-openapi/cmd/openapi-gen \
     --input-dirs ./pkg/apis/pipeline/v1beta1,./pkg/apis/pipeline/pod,./pkg/apis/resource/v1alpha1,knative.dev/pkg/apis,knative.dev/pkg/apis/duck/v1beta1 \
     --output-package ./pkg/apis/pipeline/v1beta1 -o ./ \
-    --go-header-file hack/boilerplate/boilerplate.go.txt >> /dev/null
+    --go-header-file hack/boilerplate/boilerplate.go.txt \
+    -r "${TMP_DIFFROOT}/api-report"
+
+violations=$(diff --changed-group-format='%>' --unchanged-group-format='' "hack/ignored-openapi-violations.list" "${TMP_DIFFROOT}/api-report" || echo "")
+if [ -n "${violations}" ]; then
+  echo ""
+  echo "New API rule violations found which are not present in hack/ignored-openapi-violations.list. Please fix these violations:"
+  echo ""
+  echo "${violations}"
+  echo ""
+  exit 1
+fi
 
 echo "Generating swagger file ..."
 go run hack/spec-gen/main.go v0.17.2 > pkg/apis/pipeline/v1beta1/swagger.json
--- a/pkg/apis/pipeline/v1beta1/pipeline_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_types.go
@@ -92,22 +92,28 @@ type PipelineSpec struct {
 	Description string `json:"description,omitempty"`
 	// Resources declares the names and types of the resources given to the
 	// Pipeline's tasks as inputs and outputs.
+	// +listType=atomic
 	Resources []PipelineDeclaredResource `json:"resources,omitempty"`
 	// Tasks declares the graph of Tasks that execute when this Pipeline is run.
+	// +listType=atomic
 	Tasks []PipelineTask `json:"tasks,omitempty"`
 	// Params declares a list of input parameters that must be supplied when
 	// this Pipeline is run.
+	// +listType=atomic
 	Params []ParamSpec `json:"params,omitempty"`
 	// Workspaces declares a set of named workspaces that are expected to be
 	// provided by a PipelineRun.
 	// +optional
+	// +listType=atomic
 	Workspaces []PipelineWorkspaceDeclaration `json:"workspaces,omitempty"`
 	// Results are values that this pipeline can output once run
 	// +optional
+	// +listType=atomic
 	Results []PipelineResult `json:"results,omitempty"`
 	// Finally declares the list of Tasks that execute just before leaving the Pipeline
 	// i.e. either after all Tasks are finished executing successfully
 	// or after a failure which would result in ending the Pipeline
+	// +listType=atomic
 	Finally []PipelineTask `json:"finally,omitempty"`
 }
 
@@ -169,6 +175,7 @@ type PipelineTask struct {
 	// Conditions is a list of conditions that need to be true for the task to run
 	// Conditions are deprecated, use WhenExpressions instead
 	// +optional
+	// +listType=atomic
 	Conditions []PipelineTaskCondition `json:"conditions,omitempty"`
 
 	// WhenExpressions is a list of when expressions that need to be true for the task to run
@@ -182,6 +189,7 @@ type PipelineTask struct {
 	// RunAfter is the list of PipelineTask names that should be executed before
 	// this Task executes. (Used to force a specific ordering in graph execution.)
 	// +optional
+	// +listType=atomic
 	RunAfter []string `json:"runAfter,omitempty"`
 
 	// Resources declares the resources given to this task as inputs and
@@ -191,15 +199,18 @@ type PipelineTask struct {
 
 	// Parameters declares parameters passed to this task.
 	// +optional
+	// +listType=atomic
 	Params []Param `json:"params,omitempty"`
 
 	// Matrix declares parameters used to fan out this task.
 	// +optional
+	// +listType=atomic
 	Matrix []Param `json:"matrix,omitempty"`
 
 	// Workspaces maps workspaces from the pipeline spec to the workspaces
 	// declared in the Task.
 	// +optional
+	// +listType=atomic
 	Workspaces []WorkspacePipelineTaskBinding `json:"workspaces,omitempty"`
 
 	// Time after which the TaskRun times out. Defaults to 1 hour.
@@ -562,9 +573,11 @@ type PipelineTaskCondition struct {
 
 	// Params declare parameters passed to this Condition
 	// +optional
+	// +listType=atomic
 	Params []Param `json:"params,omitempty"`
 
 	// Resources declare the resources provided to this Condition as input
+	// +listType=atomic
 	Resources []PipelineTaskInputResource `json:"resources,omitempty"`
 }
 
@@ -590,9 +603,11 @@ type PipelineDeclaredResource struct {
 type PipelineTaskResources struct {
 	// Inputs holds the mapping from the PipelineResources declared in
 	// DeclaredPipelineResources to the input PipelineResources required by the Task.
+	// +listType=atomic
 	Inputs []PipelineTaskInputResource `json:"inputs,omitempty"`
 	// Outputs holds the mapping from the PipelineResources declared in
 	// DeclaredPipelineResources to the input PipelineResources required by the Task.
+	// +listType=atomic
 	Outputs []PipelineTaskOutputResource `json:"outputs,omitempty"`
 }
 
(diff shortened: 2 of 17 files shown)


jerop:
looks like this change would fix https://github.com/tektoncd/pipeline/issues/3323 🤔

abayer:
huh! Yeah, it does look that way. Hooray for accidentally fixing additional bugs, I guess? =)

abayer:
/assign @jerop 
/assign @vdemeester 

Mind taking a look? I think this is good to go, and it's a v1 blocker. =)

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -169,11 +175,12 @@ type PipelineTask struct {
...
 	// Conditions are deprecated, use WhenExpressions instead
 	// +optional
+	// +listType=atomic
 	Conditions []PipelineTaskCondition `json:"conditions,omitempty"`
 
-	// WhenExpressions is a list of when expressions that need to be true for the task to run
+	// When is a list of when expressions that need to be true for the task to run
 	// +optional
-	WhenExpressions WhenExpressions `json:"when,omitempty"`
+	When WhenExpressions `json:"when,omitempty"`
```
vdemeester, line 183: Any reason for this change ? 🙃

pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -179,29 +185,35 @@ type TaskRunStatusFields struct {
...
 	// +optional
+	// +listType=atomic
 	ResourcesResult []PipelineResourceResult `json:"resourcesResult,omitempty"`
 
-	// TaskRunResults are the list of results written out by the task's containers
+	// TaskResults are the list of results written out by the task's containers
 	// +optional
-	TaskRunResults []TaskRunResult `json:"taskResults,omitempty"`
+	// +listType=atomic
+	TaskResults []TaskRunResult `json:"taskResults,omitempty"`
```
vdemeester, line 212: Any reason for this change ?

vdemeester (commented):
SGTM except for the field name changes, I am not sure if they need to be on the same PR, and what is the reason behind it — especially as it might break "downstream" users of the `pkg/apis` (such as cli, …) on upgrading their dependencies.

pkg/apis/pipeline/v1beta1/pipeline_types.go
abayer: The OpenAPI rules consider the field name not matching the JSON tag to be a rule violation, so I renamed this (and the other one) to match. If we want to get to no rule violations, we will need to rename the fields eventually.

vdemeester: Lol.. really ?!? That's.. interesting.. it is in the code, it doesn't "show" to the user..

abayer: Yeah, I can only assume they want to make sure the OpenAPI field name matches the Go API field name for consistency. 🤷

hack/update-openapigen.sh
```diff
@@ -17,15 +17,39 @@
...
 
 echo "Generating OpenAPI specification ..."
 go run k8s.io/kube-openapi/cmd/openapi-gen \
     --input-dirs ./pkg/apis/pipeline/v1beta1,./pkg/apis/pipeline/pod,./pkg/apis/resource/v1alpha1,knative.dev/pkg/apis,knative.dev/pkg/apis/duck/v1beta1 \
     --output-package ./pkg/apis/pipeline/v1beta1 -o ./ \
-    --go-header-file hack/boilerplate/boilerplate.go.txt >> /dev/null
+    --go-header-file hack/boilerplate/boilerplate.go.txt \
+    -r "${TMP_DIFFROOT}/api-report"
+
+violations=$(diff --changed-group-format='%>' --unchanged-group-format='' "hack/ignored-openapi-violations.list" "${TMP_DIFFROOT}/api-report" || echo "")
```
abayer, line 44: @vdemeester (and also @dibyom, since we'd want the same thing in Triggers once it's calling `openapi-gen`):

So this will diff the list of API violations from running `openapi-gen` against a known list of ignored violations (in `hack/ignored-openapi-violations.list`) and error out if there are any lines in in the report that aren't in the ignore list. I don't love it, but it works and allows us not to rename fields that currently violate `names_match`.

abayer, line 44: @vdemeester (and also @dibyom, since we'd want the same thing in Triggers once it's calling `openapi-gen`):

So this will diff the list of API violations from running `openapi-gen` against a known list of ignored violations (in `hack/ignored-openapi-violations.list`) and error out if there are any lines in in the report that aren't in the ignore list. I don't love it, but it works and allows us not to rename fields that currently violate `names_match`.

The output will look like:

```
You appear to be running from outside of GOPATH.
This script will create a temporary GOPATH at /Users/abayer/go/src/github.com/tektoncd/pipeline/.gopath for code generation.
Moving to /Users/abayer/go/src/github.com/tektoncd/pipeline/.gopath/src/github.com/tektoncd/pipeline
Generating OpenAPI specification ...

New API rule violations found which are not present in hack/ignored-openapi-violations.list. Please fix these violations:
API rule violation: names_match,github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1,PipelineResourceSpec,SecretParams

Moving to /Users/abayer/go/src/github.com/tektoncd/pipeline
Deleting symlink to pipelines repo /Users/abayer/go/src/github.com/tektoncd/pipeline/.gopath/src/github.com/tektoncd/pipeline
When you are finished with codegen you can safely run the following:
sudo rm -rf ".gopath"
```

vdemeester (approved):

pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -263,15 +275,15 @@ func (trs *TaskRunStatus) SetCondition(newCond *apis.Condition) {
...
+	Container             string `json:"container,omitempty"`
 	ImageID               string `json:"imageID,omitempty"`
 }
 
 // SidecarState reports the results of running a sidecar in a Task.
 type SidecarState struct {
 	corev1.ContainerState `json:",inline"`
 	Name                  string `json:"name,omitempty"`
-	ContainerName         string `json:"container,omitempty"`
+	Container             string `json:"container,omitempty"`
```
jerop, line 286: I believe this should also not be renamed as discussed in https://github.com/tektoncd/pipeline/pull/4402#discussion_r787932214

pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -304,8 +316,8 @@ type CloudEventDeliveryState struct {
 	// SentAt is the time at which the last attempt to send the event was made
 	// +optional
 	SentAt *metav1.Time `json:"sentAt,omitempty"`
-	// Error is the text of error (if any)
-	Error string `json:"message"`
+	// Message is the text of error (if any)
+	Message string `json:"message"`
```
jerop, line 320: I believe this should also not be renamed as discussed in https://github.com/tektoncd/pipeline/pull/4402#discussion_r787932214

jerop (commented):
thanks @abayer!

pkg/apis/pipeline/v1beta1/resource_types.go
```diff
@@ -129,7 +134,7 @@ type PipelineResourceResult struct {
 	// The field ResourceRef should be deprecated and removed in the next API version.
 	// See https://github.com/tektoncd/pipeline/issues/2694 for more information.
 	ResourceRef *PipelineResourceRef `json:"resourceRef,omitempty"`
-	ResultType  ResultType           `json:"type,omitempty"`
+	Type        ResultType           `json:"type,omitempty"`
```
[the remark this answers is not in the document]
abayer: I think it (and the other ones you commented on) should be changed - otherwise, we have to homebrew some way to ignore these specific cases of the field name not matching the JSON tag.

abayer:
Ok, rebased, additional rules violations which had been added since I last rebased are addressed (including a number of fields which didn't have `json` tags at all), and four fields have been un-renamed and their resulting violations in the output are now ignored.

pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -263,15 +275,15 @@ func (trs *TaskRunStatus) SetCondition(newCond *apis.Condition) {
 type StepState struct {
 	corev1.ContainerState `json:",inline"`
 	Name                  string `json:"name,omitempty"`
-	ContainerName         string `json:"container,omitempty"`
+	Container             string `json:"container,omitempty"`
```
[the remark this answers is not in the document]
abayer: Done

pkg/apis/pipeline/v1beta1/taskrun_types.go
abayer: Done.

pkg/apis/pipeline/v1beta1/taskrun_types.go
abayer: Done.

pkg/apis/pipeline/v1beta1/resolver_types.go
```diff
@@ -33,15 +33,16 @@ type ResolverRef struct {
...
+	// +listType=atomic
 	Resource []ResolverParam `json:"resource,omitempty"`
 }
 
 // ResolverParam is a single parameter passed to a resolver.
 type ResolverParam struct {
 	// Name is the name of the parameter that will be passed to the
 	// resolver.
-	Name string
+	Name string `json:"name,omitempty"`
```
lbernick, line 44: for fields with "json:omitempty" added, should they have a "+optional" comment as well?

pkg/apis/pipeline/v1beta1/resultref.go
```diff
@@ -23,6 +23,7 @@ import (
 )
 
 // ResultRef is a type that represents a reference to a task run result
+// +k8s:openapi-gen=false
```
lbernick, line 26: what's the reason for this?

lbernick (commented):
Would you mind adding a bit more context to the commit message and PR description about what this annotation does and why it is added? Having detail in the commit message is useful when people are using tools like git log, to be able to understand more about why the change was made without having to context switch by reading through the linked issue.

pkg/apis/pipeline/v1beta1/resolver_types.go
abayer: Good question - in these cases, they probably shouldn't be `omitempty` in the first place. I'll clean that up.

pkg/apis/pipeline/v1beta1/resultref.go
abayer: `ResultRef` isn't actually used anywhere, so far as I can tell. Being me with my tendency to not muck around with things I don't actually have full context on, I decided that it didn't make sense to care about whether `ResultRef`'s fields were properly tagged if it's never used but didn't want to go too far and delete the type completely. So I opted to exclude `ResultRef` from the OpenAPI stuff entirely, hence the annotation.

abayer:
@lbernick - I've removed the probably-not-appropriate `omitempty`s and added more detail to the commit message. I've left `ResultRef` alone at least for now.

pkg/apis/pipeline/v1beta1/resultref.go
lbernick: it looks like it's used here https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/resources/resultrefresolution.go so I would include it in codegen

pkg/apis/pipeline/v1beta1/taskrun_types.go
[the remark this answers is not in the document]
lbernick: this still appears to be renamed?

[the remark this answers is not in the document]
abayer: Ha, I renamed the wrong field entirely when "reverting". 🤦

pkg/apis/pipeline/v1beta1/resultref.go
lbernick: It doesn't look like this comment was addressed

lbernick (commented):
Thanks Andrew! Just one last quick comment, could you please update the commit message/PR title to reflect that fields are not being renamed, and link to list type [docs](https://github.com/kubernetes/kube-openapi/blob/3ee0da9b0b4211c407396d9c233b38b77ce19773/pkg/idl/doc.go#L21-L63)?

pkg/apis/pipeline/v1beta1/resultref.go
abayer: Ah, don't know how I missed both the usage and your comment! 🤦 Fixing now!

Issue #4418, opened 2021-12-14
TestDAGPipelineRun is flaky


mattmoor:
# Expected Behavior

`TestDAGPipelineRun` passes consistently and actually verifies that the tasks that SHOULD run in parallel DO run in parallel.

# Actual Behavior

On busy clusters `TestDAGPipelineRun` will frequently fail with a message like:
```
dag_test.go:248: Expected parallel tasks to execute more or less at the same time, but they were 6s apart
```

# Steps to Reproduce the Problem

This is an intermittent flake on a fairly busy cluster.

# Additional Info

The test as written today doesn't necessarily even verify that `pipeline-task-2-parallel-1` and `pipeline-task-2-parallel-2` run in parallel.  If they run in sequence quickly (`<5s`) then the test will pass.

mattmoor:
My proposal for how to fix this would be as follows:
1. Have the task definition executed by `pipeline-task-2-parallel-*` sleep for `XX` seconds (probably `30s`)
2. Check that the tasks start within `30s` of each other.

This:
 * Gives the test a fairly significant amount of leeway to schedule the second TaskRun even on a congested cluster.
 * Verifies that the tasks actually run in parallel because the second task MUST start before the first is done (b/c of the sleep).

I think this will be both more stable as well as provide more meaningful coverage.

mattmoor:
Some pointers...

Here is the test (with a nice ASCII diagram): https://github.com/tektoncd/pipeline/blob/38b9f26af4d04491ef3ea2ec42bf000e56a554bd/test/v1alpha1/dag_test.go#L37-L47

The body of `echo-task` which is what we run in parallel: https://github.com/tektoncd/pipeline/blob/38b9f26af4d04491ef3ea2ec42bf000e56a554bd/test/v1alpha1/dag_test.go#L75-L79

Where we check that the tasks start within `5s`: https://github.com/tektoncd/pipeline/blob/38b9f26af4d04491ef3ea2ec42bf000e56a554bd/test/v1alpha1/dag_test.go#L243-L249

mattmoor:
There is also a version of this in `./test/dag_test.go` (I linked `v1alpha1` above)

PR #4419, opened 2021-12-14
Fix `TestDAGPipelineRun` flakiness.

mattmoor:
_See also the linked issue for a detailed explanation of the issue this fixes._

This change alters the DAG tests in two meaningful ways:
1. Have the tasks sleep, to actually increase the likelihood of task execution overlap,
2. Use the sleep duration for the minimum delta in start times.

These changes combine should guarantee that the tasks *actually* executed in parallel, but the second part also enables this test to be less flaky on busy clusters where `5s` may not be sufficient for the task to start.

A fun anecdote to note here is that the Kubernetes [SLO for Pod startup latency](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/pod_startup_latency.md#definition) is `5s` at `99P`, which means Tekton had effectively zero room for overhead. 😅 

Fixes: https://github.com/tektoncd/pipeline/issues/4418

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes
```release-note
NONE
```

### DIFF
--- a/test/v1alpha1/dag_test.go
+++ b/test/v1alpha1/dag_test.go
@@ -1,3 +1,4 @@
+//go:build e2e
 // +build e2e
 
 /*
@@ -33,6 +34,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	knativetest "knative.dev/pkg/test"
 )
+
+const sleepDuration = 30 * time.Second
 
 // TestDAGPipelineRun creates a graph of arbitrary Tasks, then looks at the corresponding
 // TaskRun start times to ensure they were run in the order intended, which is:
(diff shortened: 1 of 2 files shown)


mattmoor:
Paging @dlorenc on the CLA bit 😅

mattmoor:
awesome...  tabs vs. spaces 🤦

test/v1alpha1/dag_test.go
```diff
@@ -34,6 +35,8 @@ import (
 	knativetest "knative.dev/pkg/test"
 )
 
+const sleepDuration = 30 * time.Second
```
afrittoli, line 38: 30s feels like a long time to wait in a test?

mattmoor: Waiting for a rerun of the e2e tests is a lot longer 😉 

I'm happy to lower this, but not sure what you are comfortable with.  5s is the 99P scheduling latency on K8s when there's available capacity, and running these tests with `t.Parallel()` on KinD things get busy quickly.  I've seen 9s apart in recently memory, but think I've seen up to 13s.

mattmoor: (to be clear the `9s` and `13s` are *failures* with what's at HEAD)

mattmoor: I reduced this to 15s, which is larger than I can recall seeing this fail with.

yu-ya0614:
- [ ]

PR #4420, opened 2021-12-14
Select entrypoint command based on runtime platform

### LINKED ISSUE #4299: Tekton entrypoint resolution can break signature verification.
# Expected Behavior

If I submit a signed OCI Image Index, it's signature can be verified by something like `cosigned`.

# Actual Behavior

Because the index is resolved to an image here: https://github.com/tektoncd/pipeline/blob/5a774e618f600982ba9293c921f3ffefb979309b/pkg/pod/entrypoint_lookup_impl.go#L72-L79

The signature on the index is no longer accessible to verification, and the Tekton pod is rejected by `cosigned`.

Enable sigstore's `cosigned` on a namespace, submit a TaskRun with a multi-arch image that is only signed at the Index level (you will also need to sign your Tekton utility images).

# Additional Info

imjasonh:
This fixes a long-standing bug affecting heterogenous clusters, where the controller's platform would be used to lookup the image's entrypoint, instead of the platform of the node where the workload would eventually run.

With this change, the controller looks up _all_ the image's entrypoints and passes them to the entrypoint binary on the node, where it uses its current runtime platform to lookup the correct entrypoint to execute.

This has the added benefit that we can now pass the entire image@digest of the multi-platform image down to the Pod, instead of the (controller's) platform-specific image. This has benefits for scenarios where Pods may be blocked from running unsigned/untrusted images, since it might be the multi-platform image index that's signed/trusted, and not any particular platform-specific constituent image.

**NB: There is still a small bug here with Windows images specifically.** If a multi-platform image contains multiple entries for the same OS+arch+variant, but with different `osversion`, like [`golang`](https://explore.ggcr.dev/?image=golang%3A1.17) for instance, this change may not select the right command, and may instead take the last matching platform's `command` instead of the "correct" one for the runtime platform. When that happens, the "correct" image will be selected for the runtime node's platform (taking `osversion` into account), but will use the "incorrect" image's `command`. In practice, every image I could find that's in this state has the same command for all `osversion`s, so it's moot. But it's _possible_ a multi-platform image will tickle this bug. The "good" news is that this bug also existed in the previous incarnation of the code -- the code would select the `command` for the _first_ matching platform (not taking into account `variant` _or_ `osversion`), then would pass it to the runtime node which may expect a possibly-different `command`. To fix this bug, you'd need to include the `osversion` in the command map key, and (somehow) detect the node's `osversion` from the entrypoint binary at runtime, to select the correct platform (including `osversion`). In practice AFAIK this never happened, but just in case, it's worth calling out while I'm still thinking about it, in case something fishy happens on Windows clusters in the future, this comment might be useful to someone. If you're reading this and nodding in excitement that I might be describing your bug, may god have mercy on your soul.

Fixes #4299 
/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Changes the way image commands are passed to the entrypoint executor, enabling more correct behavior in heterogeneous clusters, and allowing for multi-platform image references to be passed to generated Pods.
```

cc @mattmoor


pkg/pod/entrypoint.go
```diff
@@ -147,15 +147,6 @@ func orderContainers(commonExtraEntrypointArgs []string, steps []corev1.Containe
 			argsForEntrypoint = append(argsForEntrypoint, resultArgument(steps, taskSpec.Results)...)
 		}
 
-		cmd, args := s.Command, s.Args
-		if len(cmd) == 0 {
-			return nil, fmt.Errorf("Step %d did not specify command", i)
-		}
```
mattmoor, line 153: Did we lose any meaningful checking with this?

pkg/pod/entrypoint_lookup.go
```diff
@@ -31,10 +31,18 @@ type EntrypointCache interface {
...
-	Get(ctx context.Context, ref name.Reference, namespace, serviceAccountName string) (v1.Image, error)
-	// Set updates the cache with a new digest->Image mapping. This will avoid a
-	// remote registry lookup next time Get is called.
-	Set(digest name.Digest, img v1.Image)
+	//
+	// It also returns the digest associated with the given reference. If
+	// the reference referred to an index, the returned digest will be the
+	// index's digest, not any platform-specific image contained by the
+	// index.
+	Get(ctx context.Context, ref name.Reference, namespace, serviceAccountName string) (*imageData, error)
```
mattmoor, line 39: Public method on Public type returning a Private type seems like a code smell.  Should more be public?  private?  interface?

pkg/pod/entrypoint_lookup.go
```diff
@@ -43,70 +51,57 @@ type EntrypointCache interface {
...
 
-		steps[i].Image = digest.String() // Specify image by digest, since we know it now.
-		steps[i].Command = ep            // Specify the command explicitly.
+		if len(id.commands) == 1 {
+			// Promote the single found command to the step's command.
+			for _, v := range id.commands {
+				steps[i].Command = v
+				break
+			}
+		} else {
```
mattmoor, line 94: Generally my bias is towards having a single way of passing this information.  This special case is probably 80+% of cases, so we're really limiting our coverage of the else path.  My personal bias would be to send everything through the `else` path and shake out any bugs more aggressively.

pkg/pod/entrypoint_lookup_impl.go
```diff
@@ -70,32 +58,71 @@ func (e *entrypointCache) Get(ctx context.Context, ref name.Reference, namespace
 		return nil, fmt.Errorf("error creating k8schain: %v", err)
 	}
 	mkc := authn.NewMultiKeychain(kc)
```
mattmoor, line 60: FYI a `NewMultiKeychain` of one thing is the same as just passing that one thing.

pkg/pod/entrypoint_lookup_impl.go
```diff
@@ -18,50 +18,38 @@ package pod
...
 }
 
-// Get gets the image from the cache for the given ref, namespace, and SA
-func (e *entrypointCache) Get(ctx context.Context, ref name.Reference, namespace, serviceAccountName string) (v1.Image, error) {
-	// If image is specified by digest, check the local cache.
-	if digest, ok := ref.(name.Digest); ok {
-		if img, ok := e.lru.Get(digest.String()); ok {
-			return img.(v1.Image), nil
-		}
-	}
```
mattmoor, line 60: I'm puzzled at what this "caches" now?  Should we rename? 🤔

pkg/pod/entrypoint_lookup.go
imjasonh: Yeah, you're right. This exposes another tiny opportunity for a bug due to missing information in the image config, but it's tiny, and having this exercised better would be good.

pkg/pod/entrypoint.go
imjasonh: Not really, at least not practically. I removed it because now it's a bit harder to tell whether there is no command, since that information might be in the env var, dependent on the runtime platform, which this code doesn't know about.

Practically speaking, if an image within a manifest list didn't specify a command or args, this PR would allow that to get down to a scheduled pod before blowing up, instead of that being caught quickly up here. In practice that's vanishingly rare (hopefully), so I don't think we really have to account for it. Unless you have some ideas?

pkg/pod/entrypoint_lookup_impl.go
```diff
@@ -50,52 +51,117 @@ func NewEntrypointCache(kubeclient kubernetes.Interface) (EntrypointCache, error
 	}, nil
 }
 
-// Get gets the image from the cache for the given ref, namespace, and SA
-func (e *entrypointCache) Get(ctx context.Context, ref name.Reference, namespace, serviceAccountName string) (v1.Image, error) {
+// Get gets the image from the cache for the given ref, namespace, and SA.
```
mattmoor, line 54: nit: `get`

mattmoor (approved):

mattmoor:
I guess I can't really approve, so:
/lgtm

imjasonh:
I've managed to appease EasyCLA, just needs an approval @vdemeester @afrittoli 🙏

mattmoor:
/test check-pr-has-kind-label

This also needs a kick and I'm not sure it'll listen to me, but YOLO

/lgtm

examples/v1beta1/taskruns/entrypoint-resolution.yaml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+    # Multi-arch image with no command defined. We should look up the command
+    # for each platform-specific image and pass it to the Pod, which selects
+    # the right command at runtime based on the node's runtime platform.
+    - image: gcr.io/tekton-nightly/github.com/tektoncd/pipeline/cmd/nop
+
+    # Single-platform image with no command defined, but with args. We'll look
+    # up the commands and pass it to the entrypoint binary via env var, then
+    # append the specified args.
+    - image: ubuntu
```
vdemeester, line 16: Shouldn't it be using `amd64/ubuntu` if, according to the comment, it's a "single-platform" command ?

vdemeester: (or the comment is wrong)

vdemeester (commented):
One nit comment on a comment, otherwise, looking very good 😬 
/lgtm

examples/v1beta1/taskruns/entrypoint-resolution.yaml
imjasonh: Nice catch, the comment was wrong. 👍

PR #4434, opened 2021-12-15
Reduce RBAC permissions for Tekton controller/webhook roles.

wlynch:
# Changes

These are recommendations from a recent security audit.

Controller permissions modified (`tekton-pipelines-controller-tenant-access`):

- Remove pod/logs permissions - generally users access logs with their
own permissions. We only use this in our own code for tests, where the
test user is accessing as a cluster admin anyway.
- Restricts event permissions to only create/update/patch, which I
believe is the minimum set necessary for publishing with client-go
(based on
https://pkg.go.dev/k8s.io/client-go@v0.23.0/tools/events#EventSink)

Webhook permissions modified (`tekton-pipelines-webhook-cluster-access`):

- Adds resourceName restriction on CRD permissions to only allow
modification of Tekton Pipelines CRDs.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Tekton tekton-pipelines-controller-tenant-access and tekton-pipelines-webhook-cluster-access
ClusterRole permissions are reduced to better fit least privilege.
This should have no effect on the Pipelines Controller/Webhook itself, but may impact users
if they were relying on these roles for other uses.
```

### DIFF
--- a/config/200-clusterrole.yaml
+++ b/config/200-clusterrole.yaml
@@ -47,10 +47,14 @@ metadata:
     app.kubernetes.io/instance: default
     app.kubernetes.io/part-of: tekton-pipelines
 rules:
-  # Read-write access to create Pods, K8s Events and PVCs (for Workspaces)
+  # Read-write access to create Pods and PVCs (for Workspaces)
   - apiGroups: [""]
-    resources: ["pods", "pods/log", "events", "persistentvolumeclaims"]
+    resources: ["pods", "persistentvolumeclaims"]
     verbs: ["get", "list", "create", "update", "delete", "patch", "watch"]
+  # Write permissions to publish events.
+  - apiGroups: [""]
+    resources: ["events"]
+    verbs: ["create", "update", "patch"]
   # Read-only access to these.
   - apiGroups: [""]
     resources: ["configmaps", "limitranges", "secrets", "serviceaccounts"]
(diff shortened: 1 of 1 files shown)


config/200-clusterrole.yaml
```diff
@@ -74,6 +78,15 @@ rules:
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions", "customresourcedefinitions/status"]
     verbs: ["get", "list", "update", "patch", "watch"]
```
imjasonh, line 80: Do you know of a situation where we patch or update our CRD objects?

config/200-clusterrole.yaml
```diff
@@ -49,8 +49,12 @@ metadata:
 rules:
   # Read-write access to create Pods, K8s Events and PVCs (for Workspaces)
```
imjasonh, line 50: Comment is out of date.

config/200-clusterrole.yaml
wlynch: Update is done in knative. If we're already allowing updates, patching seems reasonable as well - https://github.com/tektoncd/pipeline/blob/1f5980f8c8a05b106687cfa3e5b3193c213cb66e/vendor/knative.dev/pkg/webhook/resourcesemantics/conversion/reconciler.go#L111

list/watch are also needed by knative for the informers, and unfortunately also need to be global scope :( https://github.com/tektoncd/pipeline/blob/17439c01768f734468ba1ed69a333238615e568b/vendor/knative.dev/pkg/client/injection/apiextensions/informers/apiextensions/v1/customresourcedefinition/customresourcedefinition.go#L93

config/200-clusterrole.yaml
wlynch: Fixed!

config/200-clusterrole.yaml
imjasonh: Nice, thanks for looking into that.

dibyom (approved):

PR #4436, opened 2021-12-16
test/sidecar_test.go: replace t.Errorfs with t.Fatalfs

wlynch:
# Changes

t.Fatalf causes the test to fail and return (or in this case, move on to
the next t.Run), while t.Errorf casues the test to fail but keeps going.
There's a bunch of dependent logic in this test where if one step fails,
we try to access it and nil panic.

This change replaces a bunch of Errorfs -> Fatalfs where we rely on these
condition assumptions. Cases where we are just examining the result for
certain properties are left as Errorf so we can check multiple things
before returning out.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/sidecar_test.go
+++ b/test/sidecar_test.go
@@ -1,3 +1,4 @@
+//go:build e2e
 // +build e2e
 
 /*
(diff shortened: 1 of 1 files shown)


imjasonh (commented):
/lgtm

jerop (approved):
thanks @wlynch

lbernick:
LGTM! thanks for improving tests!

PR #4440, opened 2021-12-23
Use pr.Timeouts.Pipeline in Custom Task reconcile

lbernick:
# Changes

Prior to this change, the PipelineRun controller used only the deprecated field PipelineRun.Spec.Timeout to determine if a
PipelineRun containing a custom Task had passed its timeout. This commit uses the appropriate PipelineRun timeout value,
whether from spec.Timeout, spec.Timeouts.Pipeline, or defaults.

This commit DOES NOT address any issues with how spec.Timeouts is used to determine the timeout of a TaskRun, Run,
or finally TaskRun (e.g. #4071). That logic will be addressed in a separate commit.

/kind bug

# Submitter Checklist

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing (N/A)
  functionality, content, code)

# Release Notes

```release-note
[Bug fix]: Use pipeline.spec.timeouts.pipeline to determine pipeline timeout during custom task reconcile
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
@@ -153,12 +153,11 @@ func (pr *PipelineRun) GetNamespacedName() types.NamespacedName {
 }
 
 // HasTimedOut returns true if a pipelinerun has exceeded its spec.Timeout based on its status.Timeout
-func (pr *PipelineRun) HasTimedOut() bool {
-	pipelineTimeout := pr.Spec.Timeout
+func (pr *PipelineRun) HasTimedOut(ctx context.Context) bool {
+	timeout := pr.PipelineTimeout(ctx)
 	startTime := pr.Status.StartTime
 
-	if !startTime.IsZero() && pipelineTimeout != nil {
-		timeout := pipelineTimeout.Duration
+	if !startTime.IsZero() {
 		if timeout == config.NoTimeoutDuration {
 			return false
 		}
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_types_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_types_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1beta1_test
 
 import (
+	"context"
 	"testing"
 	"time"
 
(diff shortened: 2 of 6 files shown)


pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -153,17 +153,11 @@ func (pr *PipelineRun) GetNamespacedName() types.NamespacedName {
 }
 
 // IsTimedOut returns true if a pipelinerun has exceeded its spec.Timeout based on its status.Timeout
-func (pr *PipelineRun) IsTimedOut() bool {
-	return pr.HasTimedOut()
-}
-
-// HasTimedOut returns true if a pipelinerun has exceeded its spec.Timeout based on its status.Timeout
-func (pr *PipelineRun) HasTimedOut() bool {
```
bobcatfish, line 161: I was very confused why we had 2 functions with pretty much the same name doing what seemed like the same thing so i tracked it back to https://github.com/tektoncd/pipeline/commit/49ebeb6b63a1a957b1c6fa9078bda0c3d352d7c3 and it seems like @afrittoli  was introducing an interface that both PipelineRuns and Runs (custom tasks) would comply with - and im guessing that since Runs already had `HasTimedOut` he decided to standarize on that vs `IsTimedOut`?

Definitely makes sense to me to just use one or the other but I'm wondering if removing HasTimedOut breaks something that isnt covered by a test - it seems like this would make PipelineRuns no longer conform to the RunsToCompletion interface which might interfere with how events are emitted? (I would hope there would be a test covering this if so 🙏 )

lbernick: Thanks for looking into this! It looks like the RunsToCompletion interface was removed in a [later commit](https://github.com/tektoncd/pipeline/commit/51df0a8fd0f3c4ec05c67d4f75f4bb8f68946b41), so I think removing this function should be OK.

jerop: we could move this refactor (removing the `HasTimedOut`) into its own commit, and focus this commit on using pr.Timeouts.Pipeline - what do you think?

lbernick: I think that probably isn't necessary, but I'm happy to do it if you'd prefer

vdemeester (approved):

PR #4450, opened 2022-01-04
set activeDeadlineSeconds to max for tasks with notimeouts

### LINKED ISSUE #4435: Timeout can set to 0s but failed during taskrun
Since Tekton v0.28+, `activeDeadlineSeconds` is introduced for the pods and it's calculated based on the timeout. Since `activeDeadlineSeconds` must be between 1 and 2147483647s, setting timeout to 0s will cause the following errors
```
failed to create task run pod "pipeline-15bdb-run-b-jv5wt": Pod "pipeline-15bdb-run-b-gxq8z" is invalid: spec.activeDeadlineSeconds: Invalid value: 0: must be between 1 and 2147483647, inclusive. Maybe invalid TaskSpec
```

# Expected Behavior
The webhook should give me an error if I set the timeout to 0s.

# Actual Behavior
I can apply the taskRun without errors but failed when the controller is trying to create the pod.

1. Set any task's timeout to 0s.
2. The taskrun will have the above error.
3.

# Additional Info

- Kubernetes version:

  ```
  v1.21
  ```

- Tekton Pipeline version:
v0.30.0

pritidesai:
# Changes

PR #4217 introduced better handling of the resource quota by adding support for `activeDeadlineSeconds`. `activeDeadlineSeconds` is calculated based on this formula:

```
int64(taskRun.GetTimeout(ctx).Seconds() * 1.5)
```

In case when a `timeout` on a task is set to `0s` i.e. no timeout, the `taskrun` fails with ambiguous message `"Invalid value: 0: must be between 1 and 2147483647, inclusive."` This is happening because `activeDeadlineSeconds` is set to 0 in case of a `0s` `timeout` but it is outside of the permitted [range](https://github.com/kubernetes/kubernetes/blob/v1.23.1/pkg/apis/core/validation/validation.go#L3556)  (1 to maxint32).

This commit is changing the way activeDeadlineSeconds is set such that it is set to `MaxInt32` for a task with 0s timeout.

Closes #4435 

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Set activeDeadlineSeconds to max. permitted value (MaxInt32) for a task with 0s timeout (no timeouts).
This commit fixes the bug where a task with 0s timeout was failing with out of range error.
```

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -19,6 +19,7 @@ package pod
 import (
 	"context"
 	"fmt"
+	"math"
 	"path/filepath"
 	"strconv"
 
@@ -88,6 +89,9 @@ var (
 		Name:         "tekton-internal-steps",
 		VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
 	}}
+
+	// MaxActiveDeadlineSeconds is a maximum permitted value to be used for a task with no timeout
+	MaxActiveDeadlineSeconds = int64(math.MaxInt32)
 )
 
 // Builder exposes options to configure Pod construction from TaskSpecs/Runs.
@@ -300,7 +304,13 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec 
 	if shouldAddReadyAnnotationOnPodCreate(ctx, taskSpec.Sidecars) {
 		podAnnotations[readyAnnotation] = readyAnnotationValue
 	}
+
+	// calculate the activeDeadlineSeconds based on the specified timeout (uses default timeout if it's not specified)
 	activeDeadlineSeconds := int64(taskRun.GetTimeout(ctx).Seconds() * deadlineFactor)
+	// set activeDeadlineSeconds to the max. allowed value i.e. max int32 when timeout is explicitly set to 0
+	if taskRun.GetTimeout(ctx) == config.NoTimeoutDuration {
+		activeDeadlineSeconds = MaxActiveDeadlineSeconds
+	}
 
 	podNameSuffix := "-pod"
 	if taskRunRetries := len(taskRun.Status.RetriesStatus); taskRunRetries > 0 {
--- a/pkg/pod/pod_test.go
+++ b/pkg/pod/pod_test.go
@@ -1255,6 +1255,55 @@ _EOF_
 				ActiveDeadlineSeconds: &defaultActiveDeadlineSeconds,
 			},
 		}, {
+			desc: "step-with-no-timeout-equivalent-to-0-second-timeout",
+			ts: v1beta1.TaskSpec{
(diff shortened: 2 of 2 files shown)


pkg/pod/pod.go
```diff
@@ -300,7 +300,12 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec
 	if shouldAddReadyAnnotationOnPodCreate(ctx, taskSpec.Sidecars) {
 		podAnnotations[readyAnnotation] = readyAnnotationValue
 	}
-	activeDeadlineSeconds := int64(taskRun.GetTimeout(ctx).Seconds() * deadlineFactor)
+
+	var activeDeadlineSeconds *int64
+	if taskRun.GetTimeout(ctx) != config.NoTimeoutDuration {
+		activeDeadlineSeconds = new(int64)
```
Tomcli, line 306: Thanks @pritidesai. It looks good to me. Not sure do you need to recreate a new pointer here?

pritidesai: thanks @Tomcli, the `var` declaration on line 304 points to a `nil` and dereferencing a `nil` pointer on line 307 will result in runtime panic without using zero valued int64 on line 306. Hope this helps!

vdemeester:
> In case when a `timeout` on a task is set to `0s` i.e. no timeout, the `taskrun` fails with ambiguous message `"Invalid value: 0: must be between 1 and 2147483647, inclusive."` This is happening because `activeDeadlineSeconds` is getting set to 0 in case of a `0s` `timeout` but it is outside of the permitted range.
> 
> This commit is changing the way `activeDeadlineSeconds` is set such that it's not set at all for a task with `0s` `timeout`.

I would rather set it to the `Max` instead of not setting it. Not setting "in some cases" mean some pods from Tekton will appear as Terminating and some will appear as NonTerminating, making quota "handling" very very confusing. This also means we "may" want to disallow having a timeout of `0` ?

pritidesai:
thank you @vdemeester 👍  Setting it to the `max` for no timeout tasks sounds resonable. The max is high enough (`MaxInt32`) for such tasks. I have changed it accordingly.

Yup I agree, it could be confusing to have some of the pods as terminating vs nonterminating. I am not sure though to completely disallowing timeout of 0s. There might be use cases for such no timeout tasks 🤔 . Its up to the user to select no timeout if it applies to their pipeline. As long as `no timeout` is explicit, I think it's reasonable to allow specifying it.

sbwsg:
Thanks @pritidesai ! 

/lgtm

vdemeester:
@pritidesai lgtm, we just need to change the name of the PR (for "history")

pritidesai:
> @pritidesai lgtm, we just need to change the name of the PR (for "history")

Done, sorry for the delay 🙏

jerop (approved):
thanks @pritidesai! 😸

PR #4451, opened 2022-01-05
scripts: remove set -x by default 🧣

### LINKED ISSUE #4444: Consider removing script mode default flags
Currently, Tekton scripts set the flags `-ex` if no default is specified by users (#1735). This causes scripts to exit immediately if a command fails (`-e`), and print commands before they are executed (`-x`) ([bash docs](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html)). These defaults were chosen because `-x` helps with debugging, and it often doesn't make sense to continue executing a script when a command fails (the default behavior is not preferred by many users).

We may want to consider changing these defaults for 2 reasons:
1. The `-x` flag could print secrets to logs
2. Although it is [documented](https://tekton.dev/docs/pipelines/tasks/#running-scripts-within-steps), users may not expect Tekton to use any flags in shell scripts by default. Because the correct flags to use often depend on the user's preferences, we may prefer to default to what the user would expect (no defaults).

This is a backwards incompatible change, so it should be included in our V1 release if we choose to do it.

vdemeester:
# Changes

This remove the `-x` default flag on script that do not use a shebang.
The `-x` flag print commands before they are executed, which might
leak data by default that we might not want to, hence removing it by
default.

To "revert" to the previous behavior, one might need to add the
following to its Task's scripts.

```
script: |
  #!/bin/sh
  set -xe
  # […]
```

Fixes #4444

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Removed `-x` set flag from script that do not use a shebang, in order to reduce potential unwanted data leak from scripts execution.
```

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -213,7 +213,7 @@ line will have the following default preamble prepended:
 
 ```bash
 #!/bin/sh
-set -xe
+set -e
 ```
 
 You can override this default preamble by prepending a shebang that specifies the desired parser.
--- a/pkg/pod/script.go
+++ b/pkg/pod/script.go
@@ -34,7 +34,7 @@ const (
 	debugInfoVolumeName    = "tekton-internal-debug-info"
 	scriptsDir             = "/tekton/scripts"
 	debugScriptsDir        = "/tekton/debug/scripts"
-	defaultScriptPreamble  = "#!/bin/sh\nset -xe\n"
+	defaultScriptPreamble  = "#!/bin/sh\nset -e\n"
 	debugInfoDir           = "/tekton/debug/info"
 )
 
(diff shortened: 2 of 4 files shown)


imjasonh (approved):

imjasonh:
```
    tektonbundles_test.go:188: Expected logs to say hello but received [72 101 108 108 111 10]
```

I think this test might have been relying on the `+x` behavior, those bytes correspond to `"Hello\n"`

imjasonh:
Yep:

https://github.com/tektoncd/pipeline/blob/3e137645be61199418213950b9b4e1463ba32069/test/tektonbundles_test.go#L187-L189

vdemeester:
Damn.. 😓

PR #4459, opened 2022-01-10
Remove unused informers

guillaumerose:
# Changes

Tasks, ClusterTasks and Pipelines informers were launched but not used.
Removing them will improve the controller (memory and cpu).

Above objects are loaded when they are needed with the regular tekton
go client.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Reduce memory footprint of the pipeline controller
```


vdemeester (approved):

imjasonh:
Thanks!

I guess this means we're always making actual API calls to get pipeline/task definitions, instead of finding cached copies, is that something we might want to revisit?

Also: is there a way to prevent these from sneaking back in? Maybe a lint check for unused struct fields?

Issue #4430, opened 2021-12-15
update-openapigen.sh must be run in GOPATH


imjasonh:
# Expected Behavior

In a non-`GOPATH` directory, run `./hack/update-openapigen.sh` and have the resulting code pass CI checks.

# Actual Behavior

Example (from https://github.com/tektoncd/pipeline/pull/4420):
https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4420/pull-tekton-pipeline-build-tests/1470859112957874182/

The result is tons of CI-blocking diffs because paths don't match the expected paths, which have the full absolute import path prepended.

# Steps to Reproduce the Problem

1. `cd $(mktemp -d) && git clone https://github.com/tektoncd/pipeline && cd pipeline`
2. `go build ./cmd/... && go test ./pkg/...` (works, as it should)
2. `./hack/update-openapigen.sh`
3. Push a PR and see CI failures like above ☝️

vdemeester:
Similar but different: https://github.com/tektoncd/pipeline/issues/3884

Issue #3884, opened 2021-04-15
`update-codegen.sh` doesn't work outside of GOPATH


vdemeester:
Running `./hack/update-codegen.sh` from outside of `GOPATH` seems to be broken. It will create the files in `${GOPATH}/src/github.com/tektoncd/pipeline` which, if you are not in `GOPATH`, won't update your working directory and thus won't work.

See below

```bash
$ pwd
/home/vincent/src/tektoncd/pipeline
$ ./hack/update-codegen-.sh 
Generating deepcopy funcs for resource:v1alpha1
Generating clientset for resource:v1alpha1 at github.com/tektoncd/pipeline/pkg/client/resource/clientsetGenerating listers for resource:v1alpha1 at github.com/tektoncd/pipeline/pkg/client/resource/listers
Generating informers for resource:v1alpha1 at github.com/tektoncd/pipeline/pkg/client/resource/informersGenerating deepcopy funcs for pipeline:v1alpha1,v1beta1
Generating clientset for pipeline:v1alpha1,v1beta1 at github.com/tektoncd/pipeline/pkg/client/clientset
Generating listers for pipeline:v1alpha1,v1beta1 at github.com/tektoncd/pipeline/pkg/client/listers
Generating informers for pipeline:v1alpha1,v1beta1 at github.com/tektoncd/pipeline/pkg/client/informers
Generating injection for resource:v1alpha1 at github.com/tektoncd/pipeline/pkg/client/resource/injectionGenerating injection for pipeline:v1alpha1,v1beta1 at github.com/tektoncd/pipeline/pkg/client/injection
warning: ignoring symlink /home/vincent/src/tektoncd/pipeline/cmd/controller/kodata/refs
warning: ignoring symlink /home/vincent/src/tektoncd/pipeline/cmd/controller/kodata/third_party
warning: ignoring symlink /home/vincent/src/tektoncd/pipeline/cmd/entrypoint/kodata/refs
warning: ignoring symlink /home/vincent/src/tektoncd/pipeline/cmd/entrypoint/kodata/third_party
warning: ignoring symlink /home/vincent/src/tektoncd/pipeline/cmd/git-init/kodata/refs
warning: ignoring symlink /home/vincent/src/tektoncd/pipeline/cmd/git-init/kodata/third_party
warning: ignoring symlink /home/vincent/src/tektoncd/pipeline/cmd/imagedigestexporter/kodata/refs
[...]

PR #3878, opened 2021-04-13
Disable the default workingDir and HOME overrides

sbwsg:
# Changes

Prior to this commit Steps were given a default HOME env var of "/tekton/home" and a default workingDir of "/workspace". These defaults collide with any value set by the Step's image Dockerfile.

This commit removes the default home and workingDir overrides (except in those few cases where they're still expected, like PipelineResources).

See https://groups.google.com/g/tekton-dev/c/C-PL8VYN51E/m/el5Fca_PDAAJ for our tekton-dev announcement of this change.

See https://github.com/tektoncd/pipeline/issues/1836 for the original problem description and workingDir tracking issue.

See https://github.com/tektoncd/pipeline/issues/2013 for the HOME change tracking issue.

See https://github.com/tektoncd/pipeline/blob/main/docs/deprecations.md for our documented dates for these deprecations.

See https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#alpha-beta-and-ga
for our beta deprecation policy.

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

```release-note
Steps no longer receive a default HOME env var of "/tekton/home", nor a default workingDir of "/workspace". 

action required: if you rely on these defaults you now have a choice to make: either update your Task steps to include the workingDir and HOME env or set the "disable-home-env-overwrite" and "disable-working-directory-overwrite" feature flags to "false". We plan to continue supporting these flags for another 9 months following the release of 0.24 and then they too will be removed.
```

PR #4465, opened 2022-01-11
Update hack scripts to shim temp GOPATH as needed

sbwsg:
# Changes

When we run our codegen and openapigen scripts from a directory that is
not under GOPATH the generated output uses relative prefixes instead of
absolute prefixes.

The new hack/setup-temporary-gopath.sh script detects if the repo is not in
GOPATH and creates a temporary one under ".gopath" if so. codegen and
openapigen scripts now depend on setup-temporary-gopath.sh before
generating anything.

Fixes https://github.com/tektoncd/pipeline/issues/3884
Fixes https://github.com/tektoncd/pipeline/issues/4430

... for some definition of "fixes" ...

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/hack/setup-temporary-gopath.sh
+++ b/hack/setup-temporary-gopath.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+
+# Conditionally create a temporary GOPATH for codegen
+# and openapigen to execute in. This is only done if
+# the current repo directory is not within GOPATH.
+function shim_gopath() {
+  local REPO_DIR=$(git rev-parse --show-toplevel)
+  local TEMP_GOPATH="${REPO_DIR}/.gopath"
+  local TEMP_TEKTONCD="${TEMP_GOPATH}/src/github.com/tektoncd"
+  local TEMP_PIPELINE="${TEMP_TEKTONCD}/pipeline"
+  local NEEDS_MOVE=1
+
+  # Check if repo is in GOPATH already and return early if so.
+  # Unfortunately this doesn't respect a repo that's symlinked into
+  # GOPATH and will create a temporary anyway. I couldn't figure out
+  # a way to get the absolute path to the symlinked repo root.
+  if [ ! -z $GOPATH ] ; then
+    case $REPO_DIR/ in
+      $GOPATH/*) NEEDS_MOVE=0;;
+      *) NEEDS_MOVE=1;;
+    esac
+  fi
+
+  if [ $NEEDS_MOVE -eq 0 ]; then
+    return
+  fi
+
+  echo "You appear to be running from outside of GOPATH."
+  echo "This script will create a temporary GOPATH at $TEMP_GOPATH for code generation."
+
+  # Ensure that the temporary pipelines symlink doesn't exist
+  # before proceeding.
+  delete_pipeline_repo_symlink
+
+  mkdir -p "$TEMP_TEKTONCD"
+  # This will create a symlink from
+  # (repo-root)/.gopath/src/github.com/tektoncd/pipeline
+  # to the user's pipeline checkout.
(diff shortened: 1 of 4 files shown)


sbwsg:
This uses the workaround that @vdemeester documented here: https://github.com/tektoncd/pipeline/issues/3884#issuecomment-842316019

It is not pretty but it works around the problem.

hack/update-codegen.sh
```diff
@@ -82,3 +85,5 @@ ${REPO_ROOT_DIR}/hack/update-deps.sh
 
 # Make sure the OpenAPI specification and Swagger file are up-to-date
 ${REPO_ROOT_DIR}/hack/update-openapigen.sh
+
+shim_gopath_clean
```
imjasonh, line 89: Maybe `trap shim_gopath_clean EXIT` and move this above, so this gets done even if the script is interrupted? Otherwise lgtm, and thanks for doing this!

sbwsg: That works well, cheers!

hack/setup-temporary-gopath.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o nounset
+
+# Conditionally create a temporary GOPATH for codegen
+# and openapigen to execute in. This is only done if
+# the current repo directory is not within GOPATH.
+function shim_gopath() {
+  local TEMP_GOPATH="/tmp/tekton-pipelines-codegen-gopath"
```
vdemeester, line 10: Any reason to use than instead of `mktemp` ?
One reason to use `mktemp` would be that not all systems use `/tmp` as temporary directory, and usually `mktemp` handle this 🙃.

vdemeester (commented):
One question around the path used. For reference, in `docker`, in the early days (before vendor was supported by go, etc..) we would use a `.gopath` with links in the root of  the source (and have it in the `.gitignore` of course) for it to work no matter which system it would be on.
Seems like [it's still there](https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/hack/make.sh#L67).

hack/setup-temporary-gopath.sh
sbwsg: I initially avoided that because `mktemp` on Mac has non-standard behaviour, meaning more shell code to work around it.  I don't feel strongly about it but in the moment writing it felt like just using `/tmp` was going to be clearer if / when it breaks on a system without a `/tmp` directory.

sbwsg:
> we would use a `.gopath` with links in the root of the source

OK this makes sense I think - do I understand right that we'd put `.gopath` at the root of the `pipelines` repo?

sbwsg:
Something really strange is happening when the dependencies install to the temporary GOPATH: many packages end up created with `-r--r--r--` perms, which means that deleting the temp GOPATH requires sudo.  I don't totally understand where those perms are coming from so, for now, I will just have it print a comment to the user that they can safely delete `.gopath` after the script is done.

sbwsg:
OK the `.gopath` approach works nicely, going with that thanks @vdemeester !

vdemeester (approved):
/meow

jerop (commented):
thanks @sbwsg 🎉 

/lgtm

jerop:
```
=== RUN   TestRealRunnerTimeout
    runner_test.go:37: step didn't timeout
--- FAIL: TestRealRunnerTimeout (0.02s)
```

previously seen as a flake in https://github.com/tektoncd/pipeline/pull/3878#issuecomment-832121900

/test pull-tekton-pipeline-unit-tests

lbernick:
I think this does not work if GOPATH isn't set:
`/Users/leebernick/tekton/pipeline/hack/setup-temporary-gopath.sh: line 20: GOPATH: unbound variable`

Issue #4397, opened 2021-11-26
`tekton_pipelines_controller_pipelinerun_count` metric counter increases without having any pipeline executed


gmeghnag:
# Expected Behavior
To have the `tekton_pipelines_controller_pipelinerun_count` metric counter increased only when new pipelines are executed.
# Actual Behavior
After a few hours the counter is increased without having any pipelines executed.
# Steps to Reproduce the Problem on OpenShift

1. Run at least one pipeline in the cluster
2. Expose the metrics endpoint:
```
$ oc expose -n openshift-pipelines svc/tekton-pipelines-controller --port=9090 --path="/metrics"
```
3. Check the counter value:
```
$ METRICS_ENDPOINT=$(oc get route tekton-pipelines-controller -n openshift-pipelines -o jsonpath="{.spec.host}")/metrics
$ curl -s -k $METRICS_ENDPOINT | grep -v "#" | grep tekton_pipelines_controller_pipelinerun_count
```
4. Wait a few hours (4/5) without executing any pipelines, and after re-checking the counter value, you will see it increased

# Additional Info

- Kubernetes version:
```
Kubernetes Version: v1.20.0+9689d22
```

- Tekton Pipeline version:
```
$ tkn version
Client version: 0.13.1
Pipeline version: v0.22.0
Triggers version: v0.12.1
```

gmeghnag:
`DurationAndCount` method:
https://github.com/tektoncd/pipeline/blob/release-v0.22.x/pkg/reconciler/pipelinerun/metrics.go#L126-L164

Where is Invoked:
https://github.com/tektoncd/pipeline/blob/release-v0.22.x/pkg/reconciler/pipelinerun/pipelinerun.go#L199-L206

guillaumerose:
Yes I confirm this behaviour. 

If I install Tekton with a resyncPeriod of the controller = 10 seconds, then when I create a single pipeline, every 10s, the counter `tekton_pipelines_controller_pipelinerun_count{status="success"} ` is increased by 1.

Looking at the code, it shows also the duration histogram is incorrect. Older pipelines are counted many more times than younger pipelines.

The bug is by design and I think, the only way to remove it is to refactor the code to create this counter only using the lister and not using the controller loop.
If we use the controller loop, we need to use the beforeCondition/afterCondition check to find out if the pipeline was already counted or not. .. But it's not possible: if the controller restarts, we loose this information and the counter will start again at 0.

khrm:
[...]

PR #4468, opened 2022-01-12
Fix Metric tekton_pipelines_controller_pipelinerun_count

khrm:
Added condition check to avoid recount issue.
Fixes: #4397

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fix  `tekton_pipelines_controller_pipelinerun_count` which was increasing without any new addition of pipelinerun.
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -195,12 +195,6 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 			logger.Errorf("Failed to update Run status for PipelineRun %s: %v", pr.Name, err)
 			return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
 		}
-		go func(metrics *pipelinerunmetrics.Recorder) {
-			err := metrics.DurationAndCount(pr)
-			if err != nil {
-				logger.Warnf("Failed to log the metrics : %v", err)
-			}
-		}(c.metrics)
 		return c.finishReconcileUpdateEmitEvents(ctx, pr, before, nil)
 	}
 
@@ -240,6 +234,24 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 		return controller.NewRequeueAfter(pr.PipelineTimeout(ctx) - elapsed)
 	}
 	return nil
+}
+
+func (c *Reconciler) durationAndCountMetrics(ctx context.Context, pr *v1beta1.PipelineRun) {
+	logger := logging.FromContext(ctx)
+	if pr.IsDone() {
+		// We get latest pipelinerun cr already to avoid recount
+		newPr, err := c.pipelineRunLister.PipelineRuns(pr.Namespace).Get(pr.Name)
+		if err != nil {
+			logger.Errorf("Error getting PipelineRun %s when updating metrics: %w", pr.Name, err)
+		}
+		before := newPr.Status.GetCondition(apis.ConditionSucceeded)
+		go func(metrics *pipelinerunmetrics.Recorder) {
+			err := metrics.DurationAndCount(pr, before)
+			if err != nil {
+				logger.Warnf("Failed to log the metrics : %v", err)
+			}
+		}(c.metrics)
+	}
 }
 
 func (c *Reconciler) finishReconcileUpdateEmitEvents(ctx context.Context, pr *v1beta1.PipelineRun, beforeCondition *apis.Condition, previousError error) error {
@@ -319,6 +331,7 @@ func (c *Reconciler) resolvePipelineState(
 }
 
 func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, getPipelineFunc resources.GetPipeline) error {
+	defer c.durationAndCountMetrics(ctx, pr)
 	logger := logging.FromContext(ctx)
 	cfg := config.FromContextOrDefaults(ctx)
 	pr.SetDefaults(ctx)
--- a/pkg/pipelinerunmetrics/metrics.go
+++ b/pkg/pipelinerunmetrics/metrics.go
@@ -31,6 +31,7 @@ import (
 	"go.opencensus.io/tag"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/labels"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/logging"
@@ -209,13 +210,20 @@ func nilInsertTag(task, taskrun string) []tag.Mutator {
 // DurationAndCount logs the duration of PipelineRun execution and
 // count for number of PipelineRuns succeed or failed
 // returns an error if its failed to log the metrics
-func (r *Recorder) DurationAndCount(pr *v1beta1.PipelineRun) error {
-	r.mutex.Lock()
-	defer r.mutex.Unlock()
+func (r *Recorder) DurationAndCount(pr *v1beta1.PipelineRun, beforeCondition *apis.Condition) error {
 
 	if !r.initialized {
 		return fmt.Errorf("ignoring the metrics recording for %s , failed to initialize the metrics recorder", pr.Name)
 	}
+
+	afterCondition := pr.Status.GetCondition(apis.ConditionSucceeded)
+	// To avoid recount
+	if equality.Semantic.DeepEqual(beforeCondition, afterCondition) {
+		return nil
+	}
+
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
 
 	duration := time.Duration(0)
 	if pr.Status.StartTime != nil {
--- a/pkg/pipelinerunmetrics/metrics_test.go
+++ b/pkg/pipelinerunmetrics/metrics_test.go
@@ -58,7 +58,7 @@ func getConfigContext() context.Context {
 func TestUninitializedMetrics(t *testing.T) {
 	metrics := Recorder{}
 
-	if err := metrics.DurationAndCount(&v1beta1.PipelineRun{}); err == nil {
+	if err := metrics.DurationAndCount(&v1beta1.PipelineRun{}, nil); err == nil {
 		t.Error("DurationAndCount recording expected to return error but got nil")
 	}
 	if err := metrics.RunningPipelineRuns(nil); err == nil {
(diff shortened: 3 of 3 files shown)


wlynch (changes requested):
Mostly LGTM - would it be possible to write a test that captures the buggy behavior we were seeing before?

wlynch (commented):
We should also add documentation somewhere that this metric will only increment on completion.

chuangw6:
Hi @khrm,
Thanks for creating the pr! 
I pulled it to my local and did some tests. I found out that after applying the change, the two metrics of `tekton_pipelines_controller_pipelinerun_count` (`{status="success"}` and `{status="failed"}`) are not reported anymore.

After digging into it, it seems like the [equality check](https://github.com/tektoncd/pipeline/pull/4468/files#diff-8d68571b5897c5a1224d9285f0d51d417c6fe1334ed715c409f71460cc0dd4a6R221) of `before` and `after` conditions is always true, and the consequence of that is the `DurationAndCount` function always returns `nil`. I guess the main reason is that, the `ReconcileKind` function's `pr.IsDone` [block](https://github.com/tektoncd/pipeline/pull/4468/files#diff-fcdbb0e0ea64e13920b270bf5cb2ca8993c040d77a6173ec004d1795e9e20135R179-R199) (this is the place where `DurationAndCount` is called) are reconciling finished objects and just doing some sanity checks. This explains why before and after conditions remains equal when we call `DurationAndCount`. Instead I believe we might want to do `DurationAndCount` right after the status of the pipeline is set DONE or FAILED. This will not only report the metrics but also avoid the recount issue.

I managed to move the logic of calling `DurationAndCount` to the place where the pipelinerun is just completed (status is changed). The failed pipelinerun might be trickier as we have to deal with the fact that a pipelinerun can fail for various reasons. Fortunately, the good thing is that all the errors returned after `pr.Status.MarkFailed` will be propagated back to `ReconcileKind` function.

Same logic applies to the taskrun recount issue. 
Here are two commits on my working dev branch. Please let me know if you have any questions. Happy to clean up the existing one or make another PR. Thanks!

- [pipelinerun fix](https://github.com/chuangw6/pipeline/commit/fa91a17653dac18e400f811a747f72ddb9a5518d)
- [taskrun fix](https://github.com/chuangw6/pipeline/commit/c729f83112dab325e04d57f25944f713ff5b17cc)

khrm:
@chuangw6 I think that doesn't solve the issue of the recount.

I caught one more issue in my and your code due to which I had to use pipelineRunLister to solve the recount. Sometimes, two simultaneous reconciler calls were coming in which before and after conditions were different.


Try the updated pr, it should resolve the issue.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -229,6 +223,21 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 		logger.Errorf("Reconcile error: %v", err.Error())
 	}
 
+	if pr.IsDone() {
+		// We get latest pipelinerun cr already to avoid recount
+		newPr, err := c.pipelineRunLister.PipelineRuns(pr.Namespace).Get(pr.Name)
```
wlynch, line 228: I'm not super familiar with how/when the lister updates, but wouldn't pulling in the latest status mean that the current pr status and the "before" status always match, since the status is updated during `reconcile`?

It's would be useful to add a test to make sure we're handling the reconciler state changes properly (i.e. making sure we're avoiding instances where before always equals after) - right now we're really only testing the DurationAndCount func directly.

khrm: `reconcile` changes status but that isn't applied to the `pipelienrun` object in  `Informer` cache.

vdemeester (approved):

pkg/reconciler/pipelinerun/pipelinerun.go
pritidesai: My understanding aligns with @khrm, the status changes but the change is not visible through the lister.

pritidesai:
> @chuangw6 Good catch but I think that doesn't solve the issue of the recount. I move the `DurationAndCount` after the call of `c.reconcile` which fixed this.
> 
> I caught one more issue in my and your code due to which I had to use pipelineRunLister to solve the recount. Sometimes, two simultaneous reconciler calls were coming in which before and after conditions were different.
> 
> Try the updated pr, it should resolve the issue.

Hey @chuangw6 please provide an update if possible, whether these changes work for you.

chuangw6:
Hey @khrm @pritidesai, 
Thanks for the updates. As far as I can see when testing on my local, the count matrices are reported correctly after the PR is updated.

pkg/reconciler/pipelinerun/pipelinerun.go
wlynch: For my own knowledge - When would the lister value update? Is it because we're updating only the status that this is safe to do?

I'm a bit confused with the current impl/comments since we're getting the "latest" pipelinerun, but this is really the pipelinerun before the latest status update?

I played around with this change and this appears to be correct (I even tried adding some sleeps to try and trigger a race condition), but it feels like we're relying on non-obvious behavior of the lister. I'd have a lot more peace of mind if we could add a test to ensure that it's safe rely on this lister behavior and be able to catch if this assumption is no longer true (i.e. if we make an unrelated change that would cause the lister to start returning the true latest value).

pritidesai: sure, sounds good. @khrm please add a unit test including an update during the `reconcile` in addition to what we have otherwise looks good to me 👍  Thanks!

afrittoli (commented):
Thanks for this!

I would prefer if we didn't rely on the informer cache not being updated. 
It should be a small change to rely on the condition read at the beginning of the reconcile, existing tests will still apply.

If you think you can iterate on this today, I'd be happy to wait for v0.33.0 and have this included in the next release.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -319,6 +331,7 @@ func (c *Reconciler) resolvePipelineState(
 }
 
 func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, getPipelineFunc resources.GetPipeline) error {
+	defer c.durationAndCountMetrics(ctx, pr)
```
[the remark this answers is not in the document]
khrm: @afrittoli When I was running a test, somehow the same event came twice even when the informer cache was updated.  So I decided to use informer cache. Any idea why this happens? @wlynch 

Is it due resync controller?

khrm:
As I wrote there, sometimes we get pipelineruns when status with old status even-though if we use Lister we get correct status.  That's why we need to use Lister to avoid a recount. Is it due to controller resync? If so, why isn't it taking newer informer cache value?
I think it's OK to merge pr and I can see that count that comes out is correct.
This log is from https://github.com/khrm/pipeline/pull/2 changes

```
WARN  21:14:33 beforeCond :
 name test-finally-gct8w
 BeforeFromLister &{Type:Succeeded Status:Unknown Severity: LastTransitionTime:{Inner:2022-04-24 21:12:43 +0000 UTC} Reason:Running Message:Tasks Completed: 1 (Failed: 1, Cancelled 0), Incomplete: 2, Skipped: 0},
 BeforeAvailable &{Type:Succeeded Status:Unknown Severity: LastTransitionTime:{Inner:2022-04-24 21:12:43 +0000 UTC} Reason:Running Message:Tasks Completed: 1 (Failed: 1, Cancelled 0), Incomplete: 2, Skipped: 0},
 After &{Type:Succeeded Status:False Severity: LastTransitionTime:{Inner:2022-04-24 21:14:33.002968622 +0000 UTC m=+223.977069708} Reason:PipelineRunTimeout Message:PipelineRun "test-finally-gct8w" failed to finish within "2m0s"}
WARN  21:14:33 beforeCond :
 name test-finally-gct8w
 BeforeFromLister &{Type:Succeeded Status:False Severity: LastTransitionTime:{Inner:2022-04-24 21:14:33
+0000 UTC} Reason:PipelineRunTimeout Message:PipelineRun "test-finally-gct8w" failed to finish within "2m0s"},
 BeforeAvailable &{Type:Succeeded Status:Unknown Severity: LastTransitionTime:{Inner:2022-04-24 21:12:43 +0000 UTC} Reason:Running Message:Tasks Completed: 1 (Failed: 1, Cancelled 0), Incomplete: 2, Skipped: 0},
 After &{Type:Succeeded Status:False Severity: LastTransitionTime:{Inner:2022-04-24 21:14:33.019457645 +0000 UTC m=+223.993558729} Reason:PipelineRunTimeout Message:PipelineRun "test-finally-gct8w" failed to finish within "2m0s"}
```



https://tektoncd.slack.com/archives/CLCCEBUMU/p1650873192721789

pkg/reconciler/pipelinerun/pipelinerun.go
khrm: Please check this comment.

https://github.com/tektoncd/pipeline/pull/4468#issuecomment-1108299026

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
khrm: @afrittoli @wlynch Please check this comment:

https://github.com/tektoncd/pipeline/pull/4468#issuecomment-1108299026

khrm:
Response from Matt Moor (Copied from Slack)
```The events are based on changes to the lister cache, so if the event hasn’t fired, the lister hasn’t been updated


[mattmoor](https://app.slack.com/team/U01510NCCAV)  [6 days ago](https://tektoncd.slack.com/archives/CLCCEBUMU/p1651783879999779?thread_ts=1650873192.721789&cid=CLCCEBUMU)
however, when it comes in the event should trigger a reconciliation with the new state


mattmoor  6 days ago
The guarantee you get on lister cache consistency is that it is always at least as up to date as the original event that put the work into the workqueue, but:
multiple events could be coalesced before a key is processed, and
an event may not be received at the time of reconciliation (if an event comes in DURING key processing, it is queued for reprocessing regardless of how the active processing completes)


[mattmoor](https://app.slack.com/team/U01510NCCAV)  [6 days ago](https://tektoncd.slack.com/archives/CLCCEBUMU/p1651783990545769?thread_ts=1650873192.721789&cid=CLCCEBUMU)
If you are familiar with the concept of bounded staleness, the Lister cache’s bound on staleness is the most recent received event prior to reconciliation.


mattmoor  6 days ago
But it can be stale and reconcilers should tolerate that
```
OK. So that event which was in queue (when previous event was processing), would have older condition than the one in Lister's cache. So I think it makes sense to use Lister cache here.

pritidesai:
Based on the latest discussion, it is safe to rely on the lister cache to receive the updates if there are any to the pipeline run status. 

/lgtm

PR #4478, opened 2022-01-13
Fix Pipeline/Task to *Run label/annotation propagation 🥨

### LINKED ISSUE #4477: Annotations and Labels from `Task` are not propagated to the `Pod`
# Expected Behavior

As documented [here](https://tekton.dev/docs/pipelines/labels/#label-propagation) for labels (and it used to be the same for annotations), labels and annotations set on the `Task` should be propagated to the underlying `Pod` when run.

# Actual Behavior

As of today, annotations and labels set on the `Task` are *lost in translation* and only labels and annotations from the `TaskRun` are passed to the generated `Pod`.

1. Create a `Task` with an annotation and a label
2. Create a `TaskRun` that reference it
3. Look at the `Pod` associated with the `TaskRun`, it will not have the annotation nor the label set on the `Task`

# Additional Info

- Kubernetes version:

Any version

- Tekton Pipeline version:

Current main and most likely since this [PR](https://github.com/tektoncd/pipeline/pull/3941), meaning since 0.25 or 0.26 ? 🤔 

---

The main reason for this bug is that, because we are using the `.Status.Spec` to re-construct the `Task` definition and we are not storing annotation and labels in there, we are not using them when creating the `Pod`.

vdemeester:
# Changes

Before this fix, we are "forgetting" the Task and Pipeline
metadata (Labels/Annotations) when using them in a Run (PipelineRun or
TaskRun).

It is however documented that those metadata get propagated to the
generated Pod, and even the code after storing the Spec is written
that way.

Because we only store the spec, the second time we were reconciling
the TaskRun or PipelineRun, we wouldn't fetch the referenced Task or
Pipeline anymore, and thus, we would not propagate them.

Fixes #4477

This most likely need an additional unit test or e2e test, I will be working on this 👼🏼 

/kind bug
/cc @pritidesai @sbwsg @chmouel @bobcatfish 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fix labels and annotation propagation from Task/Pipeline to TaskRun/PipelineRun and thus the generated Pod
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -344,25 +344,8 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
 	}
 
 	// Store the fetched PipelineSpec on the PipelineRun for auditing
-	if err := storePipelineSpec(ctx, pr, pipelineSpec); err != nil {
+	if err := storePipelineSpecAndMergeMeta(pr, pipelineSpec, pipelineMeta); err != nil {
 		logger.Errorf("Failed to store PipelineSpec on PipelineRun.Status for pipelinerun %s: %v", pr.Name, err)
-	}
-
-	// Propagate labels from Pipeline to PipelineRun.
-	if pr.ObjectMeta.Labels == nil {
-		pr.ObjectMeta.Labels = make(map[string]string, len(pipelineMeta.Labels)+1)
-	}
-	for key, value := range pipelineMeta.Labels {
-		pr.ObjectMeta.Labels[key] = value
-	}
-	pr.ObjectMeta.Labels[pipeline.PipelineLabelKey] = pipelineMeta.Name
-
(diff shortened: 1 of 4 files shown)


pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1147,10 +1130,28 @@ func (c *Reconciler) makeConditionCheckContainer(ctx context.Context, rprt *reso
 	return &cc, err
 }
 
-func storePipelineSpec(ctx context.Context, pr *v1beta1.PipelineRun, ps *v1beta1.PipelineSpec) error {
+func storePipelineSpecAndMergeMeta(ctx context.Context, pr *v1beta1.PipelineRun, ps *v1beta1.PipelineSpec, meta *metav1.ObjectMeta) error {
```
pritidesai, line 1133: since we are updating this function here, please drop `ctx` since its not used in this function

pritidesai (approved):
thank you @vdemeester 👍

PR #4484, opened 2022-01-13
Implicit params: only propagate params to embedded TaskSpecs in Pipelines.

### LINKED ISSUE #4483: Implicit params being propogated to Refs
# Expected Behavior

Implicit params should only be available with the same resource - they shouldn't be considered for references.

# Actual Behavior

From @sbwsg in https://github.com/tektoncd/pipeline/issues/4388#issuecomment-1012435724:

> Just to add another anecdotal here - we experienced some surprising behaviour in relation to this change during the release of 0.32 of Tekton Pipelines. One of our Pipelines relied on a default param value in a referenced Task, which was then unexpectedly overridden by the implicit param defined in the Pipeline.
>
> 
> It's tricky to describe in words, here's a link to the PipelineTask as we had it before we noticed this behaviour and here's a link to the PipelineTask after we corrected for it.
> 
> ^ In the first link we don't pass the platforms param because we're relying on the default defined in the referenced Task. With implicit params that platforms param gets overridden with the value in the platforms param defined for the Pipeline. The reason for the difference in the values is a bit nuanced but regardless it looks like this might be accidentally backwards incompatible?
>
> edit: tried to disambiguate that the default was set in the referenced task.

# Additional Info

Likely affected code: https://github.com/tektoncd/pipeline/blob/1401ab773862e1290448171a7d3fe36f63cf9aa1/pkg/apis/pipeline/v1beta1/pipeline_defaults.go#L44-L47

wlynch:
# Changes

Previously we were appling the params to all PipelineTasks, including
refs. This is incorrect behavior since the parameters should only be
propagated to things within the same document.

Fixes #4483 

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-notes
[alpha only] Fixes bug where implicit params were erroneously passed to TaskRefs within ParamSpecs.
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_defaults.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_defaults.go
@@ -41,32 +41,35 @@ func (ps *PipelineSpec) SetDefaults(ctx context.Context) {
 	}
 	for i, pt := range ps.Tasks {
 		ctx := ctx // Ensure local scoping per Task
-		if config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields == "alpha" {
-			ctx = addContextParams(ctx, pt.Params)
-			ps.Tasks[i].Params = getContextParams(ctx, pt.Params...)
-		}
+
 		if pt.TaskRef != nil {
 			if pt.TaskRef.Kind == "" {
 				pt.TaskRef.Kind = NamespacedTaskKind
 			}
 		}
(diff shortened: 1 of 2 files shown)


wlynch:
/hold

(there's some additional discussion going on around this feature, so blocking submission for the time being)

vdemeester (approved):
Seems good to me 👼🏼

Issue #3799, opened 2021-03-02
Entrypoint Image Lookup does not use PodTemplate ImagePullSecrets


skaegi:
At this point this is just by looking at the code and not validated by me. (although reported by a user)
Current main -- https://github.com/tektoncd/pipeline/blob/e2f528b0b247efb3b52b10800aad4ffd9046a266/pkg/pod/entrypoint_lookup_impl.go#L64

We're creating the k8schain with the service account, but also should add in the image pull secrets referenced in the pod template.

wlynch:
/remove-lifecycle rotten I think this is the root cause of https://github.com/tektoncd/pipeline/issues/2316 - probably worth doing to make sure we're using consistent credentials.

wlynch:
@sbwsg I know you've been looking at related k8schain code recently. cc-ing you in case you have any interest in this issue.

concaf:
I believe that the certificates mounted via `SSL_CERT_DIR` are not respected either.

Yongxuanzhang:
I'm a member of tektoncd but why I cannot be assigned?

vdemeester:
@Yongxuanzhang you have to comment once, or be the one commenting `/assign` to get assigned 🙃

Yongxuanzhang:
> @Yongxuanzhang you have to comment once, or be the one commenting `/assign` to get assigned 🙃

okok, thanks!

imjasonh:
We'll need to make sure that users give the Tekton controller SA permissions to use those secrets too.

Yongxuanzhang:
> 

Hi @imjasonh, thank you so much for the comment! I'm quite new to this, could you elaborate more on this? Do you mean we should document it to tell users to give controller SA permissions? 
And to give the controller the permissions, we need to use Rolebinding to grant the controller access to secrets, or can we add imagePullSecrets to the controller's SA? 
Thanks!

Yongxuanzhang:
> 

Hi @imjasonh, 
Thank you so much for the detailed explanations!! ❤️ 
I have 1 question regarding this, 

I wonder how it works before, if we don't pass the `imagePullSecrets` explicitly to `k8schain.New`, it will also [search](https://github.com/google/go-containerregistry/blob/d9bfbcb99e526b2a9417160e209b816e1b1fb6bd/pkg/authn/k8schain/k8schain.go#L80) the  `imagePullSecrets` attached to the SA. 
[...]

PR #4496, opened 2022-01-19
Use podTemplate ImagePullSecrets for Entrypoint Image Lookup

Yongxuanzhang:
# Changes
WIP

# Submitter Checklist

  functionality, content, code)

# Release Notes


wlynch (commented):
Impl looks reasonable - though there aren't really tests for the remote behavior. :(

The kubeclient should be easy enough to fake out (https://pkg.go.dev/k8s.io/client-go/kubernetes/fake), but I'm not sure off the top of my head how we would fake out the remote containerregistry call. https://github.com/tektoncd/pipeline/blob/4186dc96dfeea5bc465f94cf2c9cf359902973d2/pkg/remote/oci/resolver_test.go#L263-L268 something we can copy for this?

/cc @imjasonh In case you have any other ideas.

pkg/pod/entrypoint_lookup_impl.go
```diff
@@ -56,18 +57,24 @@ func NewEntrypointCache(kubeclient kubernetes.Interface) (EntrypointCache, error
...
-func (e *entrypointCache) get(ctx context.Context, ref name.Reference, namespace, serviceAccountName string) (*imageData, error) {
+func (e *entrypointCache) get(ctx context.Context, ref name.Reference, namespace, serviceAccountName string, imagePullSecrets []corev1.LocalObjectReference) (*imageData, error) {
 	// If image is specified by digest, check the local cache.
 	if digest, ok := ref.(name.Digest); ok {
 		if id, ok := e.lru.Get(digest.String()); ok {
 			return id.(*imageData), nil
 		}
 	}
 
+	var pullSecretsNames []string
```
wlynch, line 68: nit: We already know the size of this list, so we can just initialize it with the set capacity.

```suggestion
	pullSecretsNames := make([]string, 0, len(imagePullSecrets))
```

docs/podtemplates.md
```diff
@@ -95,6 +95,40 @@ Pod templates support fields listed in the table below.
 	</tbody>
 </table>
 
+## Use <code>imagePullSecrets</code> to lookup entrypoint
```
imjasonh, line 98: nit: In Markdown, we can use the ` character to enable code mode, without the more verbose/distracting HTML (here and below).


```suggestion
## Use `imagePullSecrets` to lookup entrypoint
```

docs/podtemplates.md
```diff
@@ -95,6 +95,40 @@ Pod templates support fields listed in the table below.
...
+
+If no command is configured in <code>task</code> and <code>imagePullSecrets</code> is configured in <code>podTemplate</code>, the Tekton Controller will look up the entrypoint of image with <code>imagePullSecrets</code>. If the Tekton controller's service account is not granted the access to secrets in different namespace. You need to grant the access via RoleBinding:
```
imjasonh, line 100: Is this necessary, since the SA already has cluster-wide read access to secrets?

If this is meant to document what's necessary for folks that explicitly disable Tekton's cluster-wide secrets access, I think we can just assume that users sophisticated/paranoid enough to do that are able to figure out what else they'd need to do to re-enable it for required secrets.

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,77 @@
+package pod
```
imjasonh, line 1: We'll need a copyright header on this new file.

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/registry"
+	remotetest "github.com/tektoncd/pipeline/test"
+	"github.com/tektoncd/pipeline/test/parse"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	fakeclient "k8s.io/client-go/kubernetes/fake"
+)
+
+func TestGet(t *testing.T) {
```
imjasonh, line 20: I'm always glad to have more tests, and this should ensure the code is at least exercised, but it doesn't currently really ensure that the specified imagePullSecret is used when making requests to the registry. I think it's worth keeping, but maybe we should document its limitations so we don't get a false sense of security here.

imjasonh (commented):
Looks good, and thanks for adding tests! Just a few small comments, otherwise lgtm.

docs/podtemplates.md
Yongxuanzhang: Thank you!

docs/podtemplates.md
Yongxuanzhang: This is in case the access may be removed in the future, Billy told me that it might be removed so it would be better to document it.

pkg/pod/entrypoint_lookup_impl_test.go
Yongxuanzhang: Thank you for reminding! 👍

pkg/pod/entrypoint_lookup_impl_test.go
Yongxuanzhang: I have been working on trying to address. Please let me know if the latest change makes sense or not. 😄 
Thanks!

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,167 @@
...
+		testname: "correct auth",
+		username: "foo",
+		password: "bar",
+	},
+		{
+			testname: "empty auth",
+			username: "",
+			password: "",
+		},
+	} {
```
wlynch, line 93: Might also be worth testing a secret that exists but isn't authorized, no secret (in case empty string is different than nil), and a secret that does exist at all.

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,167 @@
...
+	}{{
+		testname: "correct auth",
+		username: "foo",
+		password: "bar",
+	},
+		{
+			testname: "empty auth",
+			username: "",
+			password: "",
+		},
```
wlynch, line 92: be consistent with indenting - no strict preference on which style, but just make sure it's consistent within the statement.

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,167 @@
...
+					corev1.DockerConfigKey: []byte(
+						fmt.Sprintf(`{%q: {"auth": %q}}`,
+							u.Host,
+							base64.StdEncoding.EncodeToString([]byte(tc.username+":"+tc.password)),
+						),
+					),
+				},
+			})
+
+			entrypointCache, err := NewEntrypointCache(client)
```
wlynch, line 128: check error

wlynch (changes requested):

pkg/pod/entrypoint_lookup_impl_test.go
Yongxuanzhang: sure!

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,167 @@
...
+		// Handle the requests with matching authorization.
+		auth := strings.Split(r.Header["Authorization"][0], " ")[1]
+		if auth == base64.StdEncoding.EncodeToString([]byte(username+":"+password)) {
+			f.reg.ServeHTTP(w, r)
+		} else {
+			w.WriteHeader(401)
+		}
+
+	}
+}
```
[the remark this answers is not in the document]
Yongxuanzhang: Yes, for sure! This is much better!

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,167 @@
...
+				},
+			}, &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "secret",
+					Namespace: "ns",
+				},
+				Type: corev1.SecretTypeDockercfg,
+				Data: map[string][]byte{
+					corev1.DockerConfigKey: []byte(
+						fmt.Sprintf(`{%q: {"auth": %q}}`,
```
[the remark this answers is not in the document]
Yongxuanzhang: I searched the `corev1.DockerConfigKey` and looks like we all use raw json in other files
https://github.com/tektoncd/pipeline/blob/05baf0801dde6f027d5ca3cc7459145752a6b45b/pkg/credentials/dockercreds/creds_test.go#L243

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+			ref, err := remotetest.CreateImageWithAnnotations(u.Host+"/task/test-create-image", remotetest.DefaultObjectAnnotationMapper, task)
+			if err != nil {
+				t.Errorf("uploading image failed unexpectedly with an error: %v", err)
+			}
+
+			imgRef, err := name.ParseReference(ref)
+			if err != nil {
+				t.Errorf("digest %s is not a valid reference: %v", ref, err)
+			}
```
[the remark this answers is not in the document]
Yongxuanzhang: the client creation need password and username, I guess it should stay in the teat table?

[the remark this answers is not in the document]
wlynch: No strong preference here, so long as we're testing the right thing.
There's some odd things with how this is currently structured - i.e. there's a test case called "non-existent secret", but we're actually inserting a secret with that username/password. This is also true for "empty secret" and "no secret" - these are all testing the same case.

Regardless of where secrets are created, we can move the registry creation + population outside of the test loop - this is the same data for every iteration.

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,180 @@
...
+				},
+			})
+
+			entrypointCache, err := NewEntrypointCache(client)
+			if err != nil {
+				t.Errorf("Creating entrypointCache with an error: %v", err)
+			}
+
+			imagePullSecrets := []corev1.LocalObjectReference{{Name: "secret"}}
+			if tc.name == "no secret" {
```
wlynch, line 154: i'd avoid switching on test case names - this is usually a good signal that the test case parameters need to be changed or you need to break up the tests.

e.g. instead of passing in username password, you can pass in secret refs.

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,180 @@
...
+				imagePullSecrets = nil
+			}
+
+			id, err := entrypointCache.get(ctx, imgRef, "ns", "", imagePullSecrets)
+
+			switch tc.name {
+			case "correct secret":
```
wlynch, line 161: Same as above, I don't think we should switch on test case names here - this really boils down to whether we expect an error or not - see https://github.com/tektoncd/pipeline/search?q=wantErr for some examples for how to approach this.

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,180 @@
...
+		if !ok || username != u || password != p {
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+	}
+	// Default to open
+	f.reg.ServeHTTP(w, r)
+}
+
+func TestGet(t *testing.T) {
```
wlynch, line 74: nit: Rename this test case to be more descriptive - This isn't fully exhaustive in testing entrypoint.get - it's really only focused on the ImagePullSecrets.

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,180 @@
...
+			id, err := entrypointCache.get(ctx, imgRef, "ns", "", imagePullSecrets)
+
+			switch tc.name {
+			case "correct secret":
+				if err != nil {
+					t.Fatalf("Get: %v", err)
+				}
+
+				if id == nil {
+					t.Fatalf("Failed to get image: %v", err)
```
wlynch, line 167: nit: At this point `err == nil`, in terms of `entrypoint.get` the operation was successful (i.e. it didn't fail). We should reword this error message.

wlynch (changes requested):

pkg/pod/entrypoint_lookup_impl_test.go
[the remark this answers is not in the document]
Yongxuanzhang: > There's some odd things with how this is currently structured - i.e. there's a test case called "non-existent secret", but we're actually inserting a secret with that username/password. This is also true for "empty secret" and "no secret" - these are all testing the same case.

I will think about how to structure them 

> Regardless of where secrets are created, we can move the registry creation + population outside of the test loop - this is the same data for every iteration.
Yes, I have moved them out of the iteration.

pkg/pod/entrypoint_lookup_impl_test.go
Yongxuanzhang: ah yes, thank you!

pkg/pod/entrypoint_lookup_impl_test.go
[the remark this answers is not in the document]
chuangw6: Not sure if this helps. I saw there is an existing function `[json.marshal](https://github.com/tektoncd/chains/blob/42a7bd75a10d3306f11fe7b6c940e2f36eea1846/pkg/chains/signing/wrap.go#L109)` from go package that has been used at multiple places in `Chains` to marshal go structs.

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,172 @@
...
+	ftp := newfakeHTTP()
+	s := httptest.NewServer(&ftp)
+	defer s.Close()
+
+	u, err := url.Parse(s.URL)
+	if err != nil {
+		t.Errorf("Parsing url with an error: %v", err)
+	}
+
+	task := parse.MustParseTask(t, `metadata:name: test-create-image`)
```
wlynch, line 106:
```suggestion
	task := &v1beta1.Task{Name: "test-create-image"}
```

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,172 @@
...
+		name:             "no basic secret",
+		basicSecret:      &corev1.Secret{},
+		imagePullSecrets: []corev1.LocalObjectReference{{Name: "secret"}},
+		wantErr:          true,
+	}, {
+		name:             "no imagePullSecrets",
+		basicSecret:      generateSecret(u.Host, username, password),
+		imagePullSecrets: []corev1.LocalObjectReference{},
+		wantErr:          true,
+	}} {
```
wlynch, line 148: Another useful test I thought of - `[]corev1.LocalObjectReference{badsecret, goodsecret}`

e.g. test that we iterate through the chain properly.

wlynch (commented):
Looking good! Few minor nits left, otherwise LGTM! 🎉

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,172 @@
...
+		wantErr          bool
+	}{{
+		name:             "correct secret",
+		basicSecret:      generateSecret(u.Host, username, password),
+		imagePullSecrets: []corev1.LocalObjectReference{{Name: "secret"}},
+		wantErr:          false,
+	}, {
+		name:             "unauthorized secret",
+		basicSecret:      generateSecret(u.Host, username, "wrong password"),
+		imagePullSecrets: []corev1.LocalObjectReference{{Name: "secret"}},
```
wlynch, line 131: We're repeating this a lot, and `secret` is a bit of a const (e.g. if this gets changed above, we also need to change it in N other places). Let's pull this into a const, or somehow derive this from another value to be safe?

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,172 @@
...
+		wantErr:          true,
+	}, {
+		name:             "empty secret",
+		basicSecret:      generateSecret(u.Host, "", ""),
+		imagePullSecrets: []corev1.LocalObjectReference{{Name: "secret"}},
+		wantErr:          true,
+	}, {
+		name:             "no basic secret",
+		basicSecret:      &corev1.Secret{},
```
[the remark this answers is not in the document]
Yongxuanzhang: I tried with nil and it will get errors, is this something we can address?
`panic: expected pointer, but got nil [recovered]`

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+			entrypointCache, err := NewEntrypointCache(client)
+			if err != nil {
+				t.Errorf("Creating entrypointCache with an error: %v", err)
+			}
+
+			_, err = entrypointCache.get(ctx, imgRef, "ns", "", tc.imagePullSecrets)
+
+			if (err != nil) != tc.wantErr {
+				t.Fatalf("get() error = %v, wantErr %v", err, tc.wantErr)
```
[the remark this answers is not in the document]
Yongxuanzhang: I see. Thanks!

pkg/pod/entrypoint_lookup_impl_test.go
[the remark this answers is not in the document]
wlynch: No worries. Not the end of the world. Fine as is!

pkg/pod/entrypoint_lookup_impl_test.go
Yongxuanzhang: This code suggestion will cause error, but I find something similar.

wlynch: Most of the time I just free hand these suggestions without running it through a compiler, so no worries if this isn't exact! 😅 

Key thing here is there's no need to parse yaml into a struct object - we can just define the struct object. This can be useful for in other tests to cut down on some of the verbosity, or to reference other example files, but for trivial cases like this there isn't that much benefit.

pkg/pod/entrypoint_lookup_impl_test.go
Yongxuanzhang: Are we expecting err or not for this case? I tried with this case, but `k8schain.New()` will return err `error creating k8schain: secrets "wrong secret" not found`. This is a little bit anti my intuition.

pkg/pod/entrypoint_lookup_impl_test.go
Yongxuanzhang: sure, this makes sense

Yongxuanzhang: yes, this is a very good point!

pkg/pod/entrypoint_lookup_impl_test.go
wlynch: This is good to know (also a good reason to add tests for these cases)!

I think it's reasonable to error if the secret doesn't exist. But if the bad secret exists but just isn't the right password, then I would assume the next secret would be tried (otherwise there isn't really a reason to provide multiple secrets).

pkg/pod/entrypoint_lookup_impl_test.go
```diff
@@ -0,0 +1,172 @@
...
+		imagePullSecrets: []corev1.LocalObjectReference{{Name: "secret"}},
+		wantErr:          false,
+	}, {
+		name:             "unauthorized secret",
+		basicSecret:      generateSecret(u.Host, username, "wrong password"),
+		imagePullSecrets: []corev1.LocalObjectReference{{Name: "secret"}},
+		wantErr:          true,
+	}, {
+		name:             "empty secret",
+		basicSecret:      generateSecret(u.Host, "", ""),
```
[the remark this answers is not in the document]
Yongxuanzhang: I thought we want to test empty string here. Could you elaborate more on `explicit secret`?

pkg/pod/entrypoint_lookup_impl_test.go
Yongxuanzhang: ah I see. Thanks!

pkg/pod/entrypoint_lookup_impl_test.go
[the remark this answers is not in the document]
wlynch: Sure! So an empty secret in my mind is something like:

```go
&corev1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "foo",
		},
}
```

i.e. it's a secret, but it might be the wrong type, have no contents, etc.

What this is doing is creating a secret where the username and password = "", but is otherwise a valid docker config. It's roughly equivalent to a wrong username/password combination, which we already check in another test case above so there's no strong need to have a duplicate case for that here.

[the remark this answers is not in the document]
Yongxuanzhang: I see, thank you for clarification!

pkg/pod/entrypoint_lookup_impl_test.go
Yongxuanzhang: - So I tested this and found out that {badsecret, goodsecret} cannot work but {goodsecret,badsecret} works. The direct reason is that only 1 auth is written into the request header and sent to the server. The others are not considered. 

After taking a look at the [go-containerregistry](https://github.com/google/go-containerregistry/blob/main/pkg/authn/kubernetes/keychain.go) latest code, I found out that the function use a map to store all the auths, where `k` is the remote host and value is encoded auth. So for this test case. The 2 secrets have the same auth, the first come auth will be stored in keychain and others will be ignored.
  
Related code see here:
https://github.com/google/go-containerregistry/blob/9c35968ef47ec2e4a14a82ea72167f8655c7c853/pkg/authn/kubernetes/keychain.go#L106-L133

- I think this probably makes sense, meanwhile I'm wondering if this test case is valid or not. The secret is created from .`docker/config.json`, the file is written only when user has successfully login. So it is of low possibility that the secret is bad or incorrect. (e.g. user manually manipulates the file later.) 

What do you think?

wlynch: Got it - makes sense once you know the background. I'm neutral if you want to keep the test in or remove it, but we should make sure to document this since it's non-obvious behavior.

Yongxuanzhang: Sure! I perfer to keep it and document it in the comments.

Yongxuanzhang: emmm, it may need more restructuring of the code, I will remove it for now

docs/podtemplates.md
```diff
@@ -95,6 +95,40 @@ Pod templates support fields listed in the table below.
 	</tbody>
 </table>
 
+## Use `imagePullSecrets` to lookup entrypoint
+
+If no command is configured in `task` and `imagePullSecrets` is configured in `podTemplate`, the Tekton Controller will look up the entrypoint of image with `imagePullSecrets`. The Tekton controller's service account is given access to secrets by default. See [this](https://github.com/tektoncd/pipeline/blob/0040bcaa9ab911b8b141699e94fee303e2134f63/config/200-clusterrole.yaml#L58-L61) for reference. If the Tekton controller's service account is not granted the access to secrets in different namespace. You need to grant the access via RoleBinding:
```
imjasonh, line 100: nit: Can we link to a recent released tag instead of a random commit? Or maybe just link to `main` so we don't have to keep updating it?

imjasonh, line 100:
```suggestion
If no command is configured in `task` and `imagePullSecrets` is configured in `podTemplate`, the Tekton Controller will look up the entrypoint of image with `imagePullSecrets`. The Tekton controller's service account is given access to secrets by default. See [this](https://github.com/tektoncd/pipeline/blob/0040bcaa9ab911b8b141699e94fee303e2134f63/config/200-clusterrole.yaml#L58-L61) for reference. If the Tekton controller's service account is not granted the access to secrets in different namespace, you need to grant the access via `RoleBinding`:
```

pkg/pod/entrypoint_lookup_impl_test.go
imjasonh: > The secret is created from .docker/config.json, the file is written only when user has successfully login.

FWIW, this isn't technically true. You can edit that file to contain any username/password, and many tools implement "login" simply be editing that file without validating that the username/password is correct at all.

If there are improvements we can make to that keychain, especially if it helps Tekton users, we should consider them. I'll happily approve :)

imjasonh (commented):
This change looks great! Thanks for working on this. 👍

docs/podtemplates.md
Yongxuanzhang: sure! I will fix it.

pkg/pod/entrypoint_lookup_impl_test.go
Yongxuanzhang: > FWIW, this isn't technically true. You can edit that file to contain any username/password, and many tools implement "login" simply be editing that file without validating that the username/password is correct at all.

Oh I see, thank you! I thought the login will do authentication to make sure the auth is valid.

> If there are improvements we can make to that keychain, especially if it helps Tekton users, we should consider them. I'll happily approve 

So this is an interesting case and I'm not sure which makes more sense (Sorry again I'm still new to this but I'm willing to learn more 😄 ). 
From the following code, Say if I have 2 secrets with the same remote registry, but one is incorrect and another is correct. If the bad secret comes first then the correct one will be ignored and not stored in map. Only bad secret will be written into header and sent to server. 

This is just what I see from the test case, what I cannot see is what is the real case, will this {bad secret, correct secret} happen in real use cases? If so, then I think this should be something done from the `go-containerregistry `.


https://github.com/google/go-containerregistry/blob/9c35968ef47ec2e4a14a82ea72167f8655c7c853/pkg/authn/kubernetes/keychain.go#L106

wlynch (commented):
/lgtm 

🎉

Yongxuanzhang:
> /lgtm
> 
> 🎉

Sorry the squash cleared the review, may you review it again. 😄

wlynch (approved):
/lgtm

pkg/pod/entrypoint_lookup_impl_test.go
imjasonh: Yeah, in that case, today, go-containerregistry will select the first matching secret and use it, not knowing whether it's bad or good. It will only be able to tell whether one is "bad" or "good" by using it to try to auth to the registry, and by that time it's already dropped the old possible alternatives, and will just fail with the "bad" credentials.

This is a bit of a known gap in this keychain logic, and we have some ideas about how to fix it, but it's not really something we see come up much in practice today, so it hasn't been a top priority.

pritidesai:
> Please kindly let me know if I need to add any release note.

@Yongxuanzhang please add the release notes. This PR is fixing an issue where entrypoint image lookup did not have access to or ignored the secrets specified in the podTemplate. The users can opt in by creating appropriate role binding to allow image lookup using these secrets.

Yongxuanzhang:
> > Please kindly let me know if I need to add any release note.
> 
> @Yongxuanzhang please add the release notes. This PR is fixing an issue where entrypoint image lookup did not have access to or ignored the secrets specified in the podTemplate. The users can opt in by creating appropriate role binding to allow image lookup using these secrets.

Hi, @pritidesai Thank you for reminding me of the release note! AFAK, right now it is enabled by default config, the doc is added in case the config is changed in the future.

imjasonh:
Thanks!

/approve

Yongxuanzhang:
Fixes https://github.com/tektoncd/pipeline/issues/3799

Issue #2840, opened 2020-06-19
TestTaskRunPipelineRunCancel is flakely


bobcatfish:
# Expected Behavior

TestTaskRunPipelineRunCancel should only fail due to legit bugs (or if there is some kind of transient test infra issue we can't avoid)

# Actual Behavior

The test failed recently on an unrelated PR and passed on a re-run: https://github.com/tektoncd/pipeline/pull/2782 - https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/2782/pull-tekton-pipeline-integration-tests/1273990227291017216/

The error is:

```
  TestTaskRunPipelineRunCancel/No_retries: cancel_test.go:123: Failed to cancel PipelineRun `pear`: Operation cannot be fulfilled on pipelineruns.tekton.dev "pear": the object has been modified; please apply your changes to the latest version and try again
```

jerop:
Saw this test flake again in https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4502/pull-tekton-pipeline-alpha-integration-tests/1491815954336714752/#testv1alpha1-testtaskrunpipelineruncancel

PR #4502, opened 2022-01-20
Add resolution types to the API

sbwsg:
# Changes

[TEP-0060](https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md) introduces remote resource resolution to Tekton Pipelines, allowing PipelineRuns and TaskRuns to reference Pipelines and Tasks in remote places like git repos.

This PR adds the fields for the remote resolution feature to the `v1beta1.TaskRef` and `v1beta1.PipelineRef` types. Also includes validations and unit test coverage. These fields can only be included when the `enable-api-fields` feature-flag is set to `alpha`. Even then, utilizing these fields only results in the reconcilers printing errors.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Syntax support for remote resolution, an alpha feature in development, has been added to pipelineRef and taskRef fields. Using the remote resolution fields is not yet possible, however, and will only result in errors.
```


pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
```diff
@@ -261,6 +261,63 @@ func TestTaskRunSpec_Invalidate(t *testing.T) {
 		},
 		wantErr: apis.ErrInvalidValue("breakito is not a valid breakpoint. Available valid breakpoints include [onFailure]", "debug.breakpoint"),
 		wc:      enableAlphaAPIFields,
+	}, {
```
lbernick, line 264: could you add some test cases with valid taskrefs for the new syntax?

pkg/apis/pipeline/v1beta1/swagger.json
```diff
@@ -1471,6 +1471,24 @@
         }
       }
     },
+    "v1beta1.ResolverRef": {
```
lbernick, line 1474: Just for my own understanding what is this file for? Are the changes generated or manual?

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
```diff
@@ -370,9 +370,10 @@ func TestPipelineRun_Validate(t *testing.T) {
 
 func TestPipelineRunSpec_Invalidate(t *testing.T) {
```
lbernick, line 371: similar comment as for TaskRun, could you add a test case where the ref is valid?

bobcatfish:
Love the incremental PR approach @sbwsg 💃 🎉 !!

pkg/apis/pipeline/v1beta1/swagger.json
sbwsg: It's generated by, I think, `./hack/openapigen.sh`

I believe `swagger.json` can be used to generate API clients for other programming languages. I don't know of any examples of folks using it, though.

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
sbwsg: Added a test for taskref with resolver field + alpha gate and a test for taskref with resolver field + resource fields + alpha gate.

pkg/apis/pipeline/v1beta1/resolver_types.go
```diff
@@ -0,0 +1,33 @@
...
+	// Resolver is the name of the resolver that should perform
+	// resolution of the referenced Tekton resource, such as "git".
+	// +optional
+	Resolver string `json:"resolver,omitempty"`
+	// Resource contains the parameters used to identify the
+	// referenced Tekton resource. Example entries might include
+	// "repo" or "path" but the set of params ultimately depends on
+	// the chosen resolver.
+	// +optional
+	Resource map[string]string `json:"resource,omitempty"`
```
jerop, line 32: I'm wondering about using`map[string]string` for `Resource` - could we use `Parameters` instead?
Especially considering that the proposal suggests using object `Parameters` in this field when it's available - maybe using a list of string `Parameters` here would help in the transition later? 

> Note that the fields under `resource:` are validated by a resolver, not by Tekton Pipelines. [TEP-0075 (Dictionary
Params)](https://github.com/tektoncd/community/pull/479) describes adding support for JSON object schema to Params. It would make sense to bring in the same schema support for the `resource:` field as well and eventually resolvers might be able to publish the schema they accept so that Pipelines can enforce it during validation of a `TaskRun` or `PipelineRun`. 
~ [TEP-0060](https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md#1-establish-a-new-syntax-in-tekton-pipelines-taskref-and-pipelineref-structs-for-remote-resources)

K8s API convention encouraging use of sub-objects instead of maps: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

pkg/apis/pipeline/v1beta1/resolver_types.go
```diff
@@ -0,0 +1,33 @@
...
+package v1beta1
+
+// ResolverRef can be used to refer to a Pipeline or Task in a remote
+// location like a git repo. This feature is in alpha and these fields
+// are only available when the alpha feature gate is enabled.
+type ResolverRef struct {
+	// Resolver is the name of the resolver that should perform
+	// resolution of the referenced Tekton resource, such as "git".
+	// +optional
+	Resolver string `json:"resolver,omitempty"`
```
jerop, line 26: The TEP [proposes](https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md#resolver-specifics) a specific set of resolvers: `git`, `bundle` and `in-cluster`  

What do you think about using aliases for the strings, instead of strings directly, and adding validation for resolvers themselves? 

K8s API convention encouraging use of string aliases when fields will have a list of allowed values (enumerations): https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants

jerop (commented):
Thank you @sbwsg for the iterative approach in implementing Remote Resolution 😀 

Left some comments about the types `ResolverRef`:
- `Resolver` - considering using string alias instead of string ([K8s convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants))
- `Resource` - considering using `Parameters` instead of map ([K8s convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps))

What do you think about adding the relevant documentation for the API change and validation so that we can review it alongside the change? (we can add a bolded disclaimer that the functionality doesn't work yet, even though users can pass in the fields)

pkg/apis/pipeline/v1beta1/resolver_types.go
sbwsg: Thanks for the feedback. I've changed this to use the following syntax:

```yaml
taskRef:
  remote: git
  resource:
  - name: repo
    value: https://github.com/tektoncd/catalog.git
  - name: path
    value: /task/golang-build/0.3/golang-build.yaml
```

pkg/apis/pipeline/v1beta1/resolver_types.go
sbwsg: This field is intentionally free-form. The resolution feature is intended to allow integration with any resolver an operator wants to install and that could include ones that Tekton hasn't explicitly named in its source code. e.g. "svn", "bzr", "gcs"

So, I guess I don't see this being in the spirit of what the Resolution TEP is aiming to achieve for the Tekton project but I also don't want to introduce freeform data if you feel strongly that you'd prefer to have the options explicitly named. Let me know.

sbwsg:
I've added initial documentation to `taskruns.md`, `pipelinerun.md`, and `install.md`

pkg/apis/pipeline/v1beta1/resolver_types.go
jerop: Thanks for the additional context!

How about we declare the `Resolver` type, specify the ones we provide, then integrators can specify their own:

```go
# in tekton pipelines
type Resolver string

const (
	git     Resolver = "git"
	bundle  Resolver = "bundle"
	cluster Resolver = "cluster"
)
```

```go
# in an integration
const (
	svn tekton.Resolver = "svn"
	bzr tekton.Resolver = "bzr"
	gcs tekton.Resolver = "gcs"
)
```

Similarly to how we use [selection.Operator](https://github.com/tektoncd/pipeline/blob/ecb251c18071deb241ed57e6826c252ce40b22bf/vendor/k8s.io/apimachinery/pkg/selection/operator.go#L21) from K8s apimachinery in [operator](https://github.com/tektoncd/pipeline/blob/adf1953542467da17e189f3fb7b7b1a3a3a2050c/pkg/apis/pipeline/v1beta1/when_types.go#L33) field in `when` expressions 

What do you think?

sbwsg: OK, I've updated the type from `string` to `v1beta1.ResolverName`.

jerop:
Reopened the issue tracking the flakey test: [TestTaskRunPipelineRunCancel](https://github.com/tektoncd/pipeline/issues/2840#)

PR #4509, opened 2022-01-24
Update the release notes to describe kustomize usage

sbwsg:
# Changes

Prior to this commit our release notes asked the user to `kubectl apply`
all the resources from a subdirectory of the plumbing repo. The plumbing
repo was updated to use kustomize a while ago, so these notes also
needed an update.

This commit adds instructions for piping kustomize output to an
invocation of kubectl apply.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/tekton/README.md
+++ b/tekton/README.md
@@ -92,7 +92,17 @@ Install Task from plumbing too:
 # Apply the Tasks we are using from the catalog
 kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/main/task/golang-build/0.3/golang-build.yaml
 kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/main/task/golang-test/0.2/golang-test.yaml
-kubectl apply -f https://raw.githubusercontent.com/tektoncd/plumbing/main/tekton/resources/release/
+
+# Apply Tasks and other resources from Plumbing.
+#
+# If you want to install everything, including tekton-nightly components,
+# run this command from the root of the plumbing repo (this requires
+# "tekton-nightly" namespace to already be created in your cluster):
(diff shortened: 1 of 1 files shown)


tekton/README.md
```diff
@@ -92,7 +92,10 @@ Install Task from plumbing too:
 # Apply the Tasks we are using from the catalog
 kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/main/task/golang-build/0.3/golang-build.yaml
 kubectl apply -f https://raw.githubusercontent.com/tektoncd/catalog/main/task/golang-test/0.2/golang-test.yaml
-kubectl apply -f https://raw.githubusercontent.com/tektoncd/plumbing/main/tekton/resources/release/
+
+# Apply Tasks and other resources from Plumbing
+# Run this command from the root of the plumbing repo
+kubectl kustomize ./tekton/resources/release | kubectl apply -f -
```
pritidesai, line 98: plumbing covers both namespaces `default` and `tekton-nightly` and without the namespace created before running this command, it fails with:

```
Error from server (NotFound): error when creating "STDIN": namespaces "tekton-nightly" not found
```

Should we include a pre-req here to create the required namespace?

sbwsg: Thanks @pritidesai ! Updated with expanded comment describing this requirement.

vdemeester (approved):

tekton/README.md
afrittoli: An alternative would be to install only the `default` overlay
```
kubectl kustomize ./tekton/resources/release/overlay/default
```

sbwsg: Updated with this alternative!

PR #4511, opened 2022-01-24
Disable implicit param behavior for Pipeline objects.

### LINKED ISSUE #4388: Params copied in all steps in saved pipeline in alpha mode on Openshift with tekton 0.29+
# Expected Behavior

Save my pipeline as it is

# Actual Behavior

It's like params are being copied in all steps.

original pipeline -> https://gist.github.com/lstocchi/f7dc7aa051489001d76456493cdc6b73
saved pipeline (look at params in each step) -> https://gist.github.com/lstocchi/e9e7090c9eac6d329fcf80405f10f09e

1. have a cluster with tekton v. 0.29 +
2. enable alpha features `enable-api-fields: alpha`
3. try to save a pipeline on cluster

# Additional Info

- Kubernetes version: Tried with two openshift clusters 

 ```
Client Version: 4.8.5
Server Version: 4.9.5
Kubernetes Version: v1.22.0-rc.0+a44d0f0
```

and 

```
Client Version: 4.8.5
Server Version: 4.10.0-0.nightly-2021-11-15-034648
Kubernetes Version: v1.22.1+f773b8b

```

- Tekton Pipeline version:

```
Client version: 0.20.0
Pipeline version: v0.30.0
Triggers version: v0.17.1
```

and 

```
Client version: 0.20.0
Pipeline version: v0.29.0
Triggers version: v0.17.1
```

It works fine with 0.28.0.

### PR DESCRIPTION
**Pull request by wlynch**
# Changes

We received feedback that the implicit param implementation has behavior that
some users found unexpected - primarily, that the implicit behavior modifying
Pipelines on admission. This change disables the implicit behavior for Pipelines,
but keeps the PipelineRun and TaskRun behavior in place.

This is implemented by introducing a new context variable, similar to
the existing Alpha context behavior, that is only enabled on
PipelineRun/TaskRun reconcile. This allows for minimally invasive
changes while giving us the flexibility to revert this behavior or
enable in other contexts such as during reconcile.

On top of this:
- Moves implicit param tests to their own file, adds test for Pipeline
to cover the behavior in question.
- Fixes bug with param context key using var struct{} instead of type
struct{} -> using var associates the value to any key of struct{},
whereas type struct{} creates a new type that can be used by only that
key. There's no indication that this was causing problems before.

/kind misc

Fixes #4388

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
[alpha only]: Implicit parameters now only affect PipelineRuns and TaskRuns (notably removes behavior for Pipelines). We are actively collecting feedback around this feature, so please let us know if this is disruptive!
```


sbwsg:
/lgtm

Thanks @wlynch !

vdemeester (approved):

PR #4528, opened 2022-01-27
[refactor] Remove ImplicitParamEnabled context.

wlynch:
# Changes

This change does not introduce any different behavior - it refactors the
code to remove the use of the ImplicitParamEnabled context in favor of
direct function calls.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE

### DIFF
--- a/pkg/apis/pipeline/v1beta1/implicit_params_test.go
+++ b/pkg/apis/pipeline/v1beta1/implicit_params_test.go
@@ -34,240 +34,223 @@ func TestImplicitParams(t *testing.T) {
 	ctx := context.Background()
 	cfg := config.FromContextOrDefaults(ctx)
 	cfg.FeatureFlags = &config.FeatureFlags{EnableAPIFields: "alpha"}
+	ctx = config.ToContext(ctx, cfg)
 
-	ctxs := []struct {
-		name string
-		ctx  context.Context
+	for _, tc := range []struct {
+		// Expect in and want to be the same type.
+		in   apis.Defaultable
+		want interface{}
 	}{
+		// This data needs to be inlined to ensure unique copies on each test.
 		{
-			name: "direct implicit context",
-			ctx:  v1beta1.WithImplicitParamsEnabled(ctx, true),
-		},
-		{
-			name: "alpha context",
-			ctx:  config.ToContext(ctx, cfg),
-		},
-	}
-
-	for _, tctx := range ctxs {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: pkg/apis/pipeline/v1beta1/implicit_params_test.go [hunk 1]
```diff
@@ -314,160 +297,53 @@ func TestImplicitParams_Pipeline(t *testing.T) {
```
bobcatfish, line 297: Im a bit confused about why the Pipeline spec has an implicit params test - it seems like in the updated version, the implicit params logic is only called from the pipelinerun and taskrun defaulting functions - i would think pipeline.SetDefaults wouldn't have any implicit param logic? (unless the purpose of this test is to ensure that the implicit params logic is not applied?)

one thing that might make this a bit clearer is to have test cases specifically for the new implicit params functions  as well - the current test cases seem to be covering the existing defaulting logic as well so it's a bit hard to tell what is being caused by the implicit params logic and what is caused by the defaulting logic

pkg/apis/pipeline/v1beta1/implicit_params_test.go
```diff
@@ -34,240 +34,223 @@ func TestImplicitParams(t *testing.T) {
...
 	cfg.FeatureFlags = &config.FeatureFlags{EnableAPIFields: "alpha"}
+	ctx = config.ToContext(ctx, cfg)
 
-	ctxs := []struct {
-		name string
-		ctx  context.Context
+	for _, tc := range []struct {
+		// Expect in and want to be the same type.
+		in   apis.Defaultable
+		want interface{}
```
bobcatfish, line 42: could you include a name to indicate the purpose of each test case? i think its one for a pipelinerun, one for a taskrun (and i think nothing more complicated than that?) but it wasnt easy to tell without reading through each test case

bobcatfish (commented):
i have a couple of comments about the tests - in particular im not clear on what the intention is of the TestImplicitParams_Pipeline test when that defaulting logic shouldn't include the implicit params logic (maybe it's just a naming or commenting thing to clear that up, but im thinking more focused tests might help here)

but in the meantime this has been open for a while and think the codebase is in a better place with this PR in (i.e. more explicit than it was when we were using flags in the context variable to change function behavior) so id like to see this get in even without those tweaks (if needed at all)

/approve
/lgtm

Issue #4460, opened 2022-01-10
PipelineRun Timeouts


jerop:
In [TEP-0046: Finally Tasks Execution Post PipelineRun Timeout](https://github.com/tektoncd/community/blob/main/teps/0046-finallytask-execution-post-timeout.md), we introduced new `timeouts` fields to replace the `timeout` field. 

```yaml

# Before
kind: PipelineRun
spec:
  timeout: "0h0m60s"

# After
kind: PipelineRun
spec:
  timeouts:
    pipeline: "0h4m0s"
    tasks: "0h1m0s"
    finally: "0h3m0s"
```

As a follow up to that work, we need to:
- Promote `PipelineRun.Timeouts` from Alpha to Beta
- Deprecate `PipelineRun.Timeout`

This work should be included in our V1 effort. 

/kind feature
/cc @lbernick @souleb @pritidesai

pritidesai:
thank you @jerop 👍 

Yes, definitely, the `pipelineRun.timeouts` was introduced in `0.25` about 6 releases back - https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features. We can work towards promoting it to beta and making it part of V1 efforts. We also updated our docs with a warning of deprecating `pipelineRun.timeout` in the same release which should buy us enough time for the deprecation.

lbernick:
Before promoting `pipelinerun.timeouts` to beta, we should clearly document and validate what happens if a user wants no timeout for `pipelinerun.timeouts.tasks` and `pipelinerun.timeouts.finally`. Right now, it's possible to set those timeouts to 0 (no timeout) without setting `pipelinerun.timeouts.pipeline` to no timeout, which I would consider an invalid configuration.

jerop:
@vinamra28 mentioned at Pipeline working group we discussed this and you're willing to take this on after #4611

lbernick:
#4071 is relevant here-- not sure if it's a blocker for moving timeouts to beta but it's a pretty gnarly bug

lbernick:
Just talked with Vinamra; I think it would be OK to promote timeouts to beta and deprecate Timeout despite this bug. Reasoning here is that this bug  doesn't affect Timeouts.Pipeline (which replaces Timeout), and we are confident we want Timeouts to be part of the API rather than Timeout.

PR #4539, opened 2022-02-01
Add validation for PR infinite timeouts

lbernick:
# Changes
Prior to this commit, it was possible to create a PipelineRun with `timeouts.tasks` or `timeouts.finally`
set to 0 (no timeout) without setting the `timeouts.pipeline` to 0. If `timeouts.pipeline` is not set,
it defaults to the global timeout value, which is not 0. This would imply that `timeouts.tasks` or `timeouts.finally`
is longer than `timeouts.pipeline`, and it's not clear how Tekton should handle this.

This commit results in an error returned to the user if they define a PipelineRun spec as described above,
and documents this behavior.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
[Bug Fix] Add validation for 0 timeout for `pipelinerun.timeouts.tasks` and `pipelinerun.timeouts.finally`
```

### DIFF
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -523,6 +523,7 @@ spec:
 
 If you do not specify the `timeout` value or `timeouts.pipeline` in the `PipelineRun`, the global default timeout value applies.
 If you set the `timeout` value or `timeouts.pipeline` to 0, the `PipelineRun` fails immediately upon encountering an error.
+If `timeouts.tasks` or `timeouts.finally` is set to 0, `timeouts.pipeline` must also be set to 0.
 
 The global default timeout is set to 60 minutes when you first install Tekton. You can set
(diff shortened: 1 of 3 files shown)


lbernick:
Related: #4460

bobcatfish:
nice!! good call, thanks @lbernick 

/approve

lbernick:
> it seems that the current behavior is that `timeouts.pipeline` will apply, right? it may be useful for future reference to document this existing behavior in the commit/pr

I wasn't sure what the existing behavior is so I tried it out with `timeouts.finally = 0` and `timeouts.pipeline = 1m`. Turns out the pipeline times out after 1m but the taskrun is still marked as running! I'll add some detail to the commit message.

Issue #4512, opened 2022-01-25
TaskRun Reason doesn't properly describe error for invalid ResourceRequest


wlynch:
# Expected Behavior

Errors generated from a bad ResourceRequest TaskRun spec generate a `TaskRunValidationFailed` reason (or some other better fitting error Reason).

# Actual Behavior

The reason returned is `CouldntGetTask`.

I was playing around with stepTemplate features (trying to intentionally cause failures) and noticed that the TaskRun status Reason didn't match the error that was happening.

e.g. for this TaskRun:

```yaml
kind: TaskRun
apiVersion: tekton.dev/v1beta1
metadata:
  generateName: tr-
spec:
  taskSpec:
    steps:
      - image: ubuntu
        script: |
          echo "hi!"
        resources:
          requests:
            memory: 300Mi
    stepTemplate:
      resources:
        limits:
          memory: 150Mi
```

This generated a reasonable error message in `kubectl describe`:

```
  Warning  Failed         8s    TaskRun  failed to create task run pod "tr-rrfbd": Pod "tr-rrfbd-pod" is invalid: spec.containers[0].resources.requests: Invalid value: "300Mi": must be less than or equal to memory limit. Maybe invalid TaskSpec
  Warning  InternalError  8s    TaskRun  1 error occurred:
           * failed to create task run pod "tr-rrfbd": Pod "tr-rrfbd-pod" is invalid: spec.containers[0].resources.requests: Invalid value: "300Mi": must be less than or equal to memory limit. Maybe invalid TaskSpec
```

But the TaskRun status described this as `CouldntGetTask`:

```sh
$ k get tr tr-rrfbd
NAME       SUCCEEDED   REASON           STARTTIME   COMPLETIONTIME
tr-rrfbd   False       CouldntGetTask   92m         92m
```

```yaml
  Conditions:
    Last Transition Time:  2022-01-25T16:46:11Z
    Message:               failed to create task run pod "tr-rrfbd": Pod "tr-rrfbd-pod" is invalid: spec.containers[0].resources.requests: Invalid value: "300Mi": must be less than or equal to memory limit. Maybe invalid TaskSpec
    Reason:                CouldntGetTask
    Status:                False
    Type:                  Succeeded
```

My guess is we're just using this Reason as a general "we failed to generate the pod for the Task" error, but this should probably use a different type of error (e.g. `TaskRunValidationFailed`)

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

[...]

PR #4547, opened 2022-02-04
Add taskSpec resources validation

Yongxuanzhang:
# Changes

This PR introduces the taskSpec resources validation, and mark the taskrun failed reason to `ReasonFailedValidation`. Before this commit the reason will be recognized as `CouldntGetTask`.

FIxes: https://github.com/tektoncd/pipeline/issues/4512

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes
```release-note
NONE
```


pkg/reconciler/taskrun/validate_resources_test.go
```diff
@@ -422,3 +423,118 @@ func TestValidateResolvedTaskResources_InvalidResources(t *testing.T) {
...
 }
+
+func TestValidateTaskSpecRequestResources_ValidResources(t *testing.T) {
+	ctx := context.Background()
+
+	tcs := []struct {
+		name string
+		task *v1beta1.TaskSpec
+	}{{
+		name: "no-requested-resources",
```
imjasonh, line 434: nit: These names can include spaces, they'll be converted to `_`s automatically if the `t.Run` fails. Spaces might make them easier to read in code.

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -142,3 +142,18 @@ func ValidateResolvedTaskResources(ctx context.Context, params []v1beta1.Param,
 
 	return nil
 }
+
+func ValidateTaskSpecRequestResources(ctx context.Context, taskSpec *v1beta1.TaskSpec) error {
```
imjasonh, line 146: Is this method only exported so that it can be tested from `package taskrun_test`?

If so, this is generally not a best practice. It both means the method will be unnecessarily exported to external users, who may depend on its behavior not changing. The reason the test is in `package taskrun_test` is to be able to test only exported behaviors, so perhaps we should change it to `package taskrun` and test unexported methods/behaviors.

pkg/reconciler/taskrun/validate_resources_test.go
```diff
@@ -422,3 +423,118 @@ func TestValidateResolvedTaskResources_InvalidResources(t *testing.T) {
...
+
+}
+
+func TestValidateTaskSpecRequestResources_InvalidResources(t *testing.T) {
+	ctx := context.Background()
+	tcs := []struct {
+		name string
+		task *v1beta1.TaskSpec
+	}{{
+		name: "limit-not-configured",
```
imjasonh, line 489: Same comment here

pkg/reconciler/taskrun/validate_resources_test.go
Yongxuanzhang: sure!

imjasonh (approved):

vdemeester (commented):
How will this act with LimitRange ? aka is the validation done pre or post applying limit range ? (I think it's pre). This means it only validates the users request – aka if there is LimitRange, then it would pass the validation but could fail later (LimitRange being mis-configured). I think that's fine, but I just want to make sure it is.

Yongxuanzhang:
> LimitRange

This is a good question. And I agree that if LimitRange is mis-configured it will pass this and fail somewhere else.  Validating `LimitRange` is out of the scope of this PR and issue. This one mainly focuses on validating the resources requested in `taskSpec`.

vdemeester:
Sounds good @Yongxuanzhang 🎉 
/lgtm

Issue #4519, opened 2022-01-26
Can the "place-tools" and "place-scripts" init-containers merge into one init-container?


Tomcli:
### Feature request
Our team is trying to optimize our Tekton pipelines to have fewer delays. We found out that many Tekton tasks need to run the "place-tools" and "place-scripts" init-containers to set all the Tekton tools and wrap the user scripts. Since some of our machines may have slow container startup time, we want to see can these init-containers merge into one so we can save 1-2s in each of our tasks. 

/cc @pritidesai @afrittoli

Tomcli:
cc @animeshsingh

pritidesai:
@rafalbigaj reported, for about 26 seconds of workload, initiating `place-tools` container is causing an average of 2 seconds of delay after the image is pulled from the container registry.

afrittoli:
https://github.com/tektoncd/pipeline/pull/4352 introduced a new init container.
If we are going to merge the various init containers into one, we might consider including that one too.

vdemeester:
/lifecycle frozen
Still planning to work on this 👼🏼

vdemeester:
I think I'll go ahead and close this one for now. It is old, lot's has changed, and it's not really a problem anymore.

PR #4552, opened 2022-02-08
[RFC] Add default k8s resources for init containers

### LINKED ISSUE #2933: Resource request not applied to init and sidecar containers
I get a `ResourceQuota` applied to my namespace, then I cannot make my `taskrun` run success, get 
```
forbidden: failed quota, must specify limits.cpu,limits.memory,requests.cpu,requests.memory. Maybe missing or invalid Task
```
Event I set the `resources` section in my each `steps`, or in `stepTemplate`.
So I guess there is no such section in `initContainer` or Tekton's `sidecar` container.

# Expected Behavior
`Taskrun` run success.
# Actual Behavior
Failed
1. Create `ResourceQuota` in certain namespace.
2. Create `Taskrun` in that namespace

# Additional Info

- Kubernetes version:

  ```
  (paste your output here)
  ```

- Tekton Pipeline version:

lbernick:
# Changes
This commit adds resource requests and limits for CPU and memory to Tekton-controlled init containers.
Prior to this commit, if a ResourceQuota was defined in a namespace, Tekton could not be used in that namespace
because not all containers in the generated pods had CPU and memory requests and limits. Closes #2933.

This commit does not affect Tekton's handling of LimitRange. Tekton will continue to transform container
resource requests and limits to fit LimitRanges.

**What do these init containers do?**
- "place-tools" runs the entrypoint binary and writes it into each step container
- "place-scripts" encodes each step's script and writes it to the container that will run it.
We do not limit the size of scripts.
- "step-init" initializes /tekton/steps with one file per step
- "working-dir-initializer" creates a working dir for each step that needs it

**Where do these resource requirements come from?**
- The latest entrypoint binary is about 80MB. Previous builds were closer to 40-50MB.
- CPU requirements, and memory requirements for non-entrypoint containers, are guesses with some simple local experimentation.
(It's challenging to get metrics for resource usage of Tekton init containers, and we'd need
some sort of performance or load tests to really get a good estimate.)

**Side effects**
- Tekton pods now have "burstable" quality of service by default instead of "best effort", meaning they will
be more likely to be scheduled and less likely to be evicted.
It's no longer possible to create a Tekton pod with "best effort" quality of service.

**Drawbacks**
- If we update the entrypoint code and the binary changes size, the resource requirements will need to be adjusted.
Because Tekton releases include building the entrypoint binary, this could break a release and we'd have to backport a fix.
- If a pod is OOMkilled due to exceeding resource requirements, the user will not be able to work around this problem.
This can be mitigated by setting the init container's resource requirements to higher than what it likely needs.
- If resource requirements are set too high, the pod may not be schedulable anywhere.

**Alternatives**
- We could apply resource requirements from a Task's StepTemplate to init containers. However, the StepTemplate
is meant to be used for Steps, which likely have higher resource requirements than Tekton's init containers,
and this would require users to understand what Tekton's init containers do.
- We could try to set resource requirements dynamically based on the number of Steps, the size of scripts, or other parameters.
- We could apply resource requirements to init containers only when ResourceQuotas are present in the namespace.

**Future work**
- Combine Tekton init containers into one init container (#4519).

**References**
- Resource quota docs: https://kubernetes.io/docs/concepts/policy/resource-quotas/
- Pod quality of service docs: https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/
- Tekton entrypoint images: https://pantheon.corp.google.com/gcr/images/tekton-releases/us/github.com/tektoncd/pipeline/cmd/entrypoint?project=tekton-releases

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
[Bug fix]: apply resource requests to init containers, allowing Tekton to be used in namespaces with resourcequotas.
[Side effect]: Pods will now have "burstable" QoS by default, instead of "best effort".
```

### DIFF
--- a/pkg/internal/limitrange/transformer_test.go
+++ b/pkg/internal/limitrange/transformer_test.go
@@ -431,6 +431,7 @@ func TestTransformerOneContainer(t *testing.T) {
 }
 
 func TestTransformerMultipleContainer(t *testing.T) {
+	// TODO: Add tests for multiple init containers
 	for _, tc := range []struct {
 		description string
 		limitranges []corev1.LimitRangeItem
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -31,6 +31,7 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/pipeline/pkg/names"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
@@ -180,6 +181,16 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec 
 		// into the correct location for later steps.
 		Command:      []string{"/ko-app/entrypoint", "cp", "/ko-app/entrypoint", entrypointBinary},
 		VolumeMounts: []corev1.VolumeMount{binMount},
+		Resources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("40Mi"),
+				corev1.ResourceCPU:    resource.MustParse("100m"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("150Mi"),
+				corev1.ResourceCPU:    resource.MustParse("300m"),
+			},
+		},
 	}
 
 	if alphaAPIEnabled {
@@ -413,6 +424,7 @@ func runVolume(i int) corev1.Volume {
 	}
 }
 
+// tektonDirInit returns an init container definition that sets up the /tekton/steps directory for the pod
 func tektonDirInit(image string, steps []v1beta1.Step) corev1.Container {
 	cmd := make([]string, 0, len(steps)+2)
 	cmd = append(cmd, "/ko-app/entrypoint", "step-init")
@@ -429,5 +441,15 @@ func tektonDirInit(image string, steps []v1beta1.Step) corev1.Container {
 			Name:      "tekton-internal-steps",
 			MountPath: pipeline.StepsDir,
 		}},
+		Resources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("10Mi"),
+				corev1.ResourceCPU:    resource.MustParse("10m"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("30Mi"),
+				corev1.ResourceCPU:    resource.MustParse("100m"),
+			},
+		},
 	}
 }
--- a/pkg/pod/script.go
+++ b/pkg/pod/script.go
@@ -25,6 +25,7 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/pipeline/pkg/names"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -93,6 +94,16 @@ func convertScripts(shellImageLinux string, shellImageWin string, steps []v1beta
 		Command:      []string{shellCommand},
 		Args:         []string{shellArg, ""},
 		VolumeMounts: []corev1.VolumeMount{writeScriptsVolumeMount, binMount},
+		Resources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("30Mi"),
+				corev1.ResourceCPU:    resource.MustParse("100m"),
+			},
+			Limits: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("300Mi"),
+				corev1.ResourceCPU:    resource.MustParse("300m"),
+			},
+		},
 	}
 
 	breakpoints := []string{}
--- a/pkg/pod/workingdir_init.go
+++ b/pkg/pod/workingdir_init.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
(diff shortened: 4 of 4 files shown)


lbernick:
FYI @pritidesai @guillaumerose This is an example of a change where performance benchmarking would be really useful!

lbernick:
@vdemeester would love to hear your thoughts here

afrittoli:
Thank you for the through PR description!

> # Changes
> This commit adds resource requests and limits for CPU and memory to Tekton-controlled init containers. Prior to this commit, if a ResourceQuota was defined in a namespace, Tekton could not be used in that namespace because not all containers in the generated pods had CPU and memory requests and limits. Closes #2933.
> 
> This commit does not affect Tekton's handling of LimitRange. Tekton will continue to transform container resource requests and limits to fit LimitRanges.
> 
> **What do these init containers do?**
> 
> * "place-tools" runs the entrypoint binary and writes it into each step container
> * "place-scripts" encodes each step's script and writes it to the container that will run it.
>   We do not limit the size of scripts.
> * "step-init" initializes /tekton/steps with one file per step
> * "working-dir-initializer" creates a working dir for each step that needs it
> 
> **Where do these resource requirements come from?**
> 
> * The latest entrypoint binary is about 80MB. Previous builds were closer to 40-50MB.
> * CPU requirements, and memory requirements for non-entrypoint containers, are guesses with some simple local experimentation.
>   (It's challenging to get metrics for resource usage of Tekton init containers, and we'd need
>   some sort of performance or load tests to really get a good estimate.)
> 
> **Side effects**
> 
> * Tekton pods now have "burstable" quality of service by default instead of "best effort", meaning they will
>   be more likely to be scheduled and less likely to be evicted.
>   It's no longer possible to create a Tekton pod with "best effort" quality of service.

This side effect seems to be a perfect candidate to be included in the release notes and probably in the docs somewhere?

> 
> **Drawbacks**
> 
> * If we update the entrypoint code and the binary changes size, the resource requirements will need to be adjusted.

This feels like something we should document (at least in a comment in the code).

>   Because Tekton releases include building the entrypoint binary, this could break a release and we'd have to backport a fix.
> * If a pod is OOMkilled due to exceeding resource requirements, the user will not be able to work around this problem.
>   This can be mitigated by setting the init container's resource requirements to higher than what it likely needs.
> 
> **Alternatives**
> 
> * We could apply resource requirements from a Task's StepTemplate to init containers. However, the StepTemplate
>   is meant to be used for Steps, which likely have higher resource requirements than Tekton's init containers,
>   and this would require users to understand what Tekton's init containers do.
> * We could try to set resource requirements dynamically based on the number of Steps, the size of scripts, or other parameters.
> * We could apply resource requirements to init containers only when ResourceQuotas are present in the namespace.
> 
> **Future work**
> 
> * Combine Tekton init containers into one init container ([Can the "place-tools" and "place-scripts" init-containers merge into one init-container? #4519](https://github.com/tektoncd/pipeline/issues/4519)).
> 
> **References**
> 
> * Resource quota docs: https://kubernetes.io/docs/concepts/policy/resource-quotas/
> * Pod quality of service docs: https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/
> * Tekton entrypoint images: https://pantheon.corp.google.com/gcr/images/tekton-releases/us/github.com/tektoncd/pipeline/cmd/entrypoint?project=tekton-releases
> 
> /kind bug
> 
> # Submitter Checklist
> 
> * [ ]  [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
> * [ ]  [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
> * [ ]  Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
> * [ ]  Release notes block below has been filled in or deleted (only if no user facing changes)
> 
> # Release Notes
> ```
> [Bug fix]: apply resource requests to init containers, allowing Tekton to be used in namespaces with resourcequotas
> ```

afrittoli:
> If a pod is OOMkilled due to exceeding resource requirements, the user will not be able to work around this problem.
This can be mitigated by setting the init container's resource requirements to higher than what it likely needs.

We could make these values configurable via a platform wide config map (possibly one of the existing ones).

vdemeester:
> **Side effects**
> 
> * Tekton pods now have "burstable" quality of service by default instead of "best effort", meaning they will
>   be more likely to be scheduled and less likely to be evicted.
>   It's no longer possible to create a Tekton pod with "best effort" quality of service.

By default they would be "burstable" and if you have LimitRange in your namespace, they would become "Guaranteed" — or if they set requests/limits themselves. I think it's close to the 2nd point of drawbacks (below), but this mean the users might/will have to take these into account to set the requests/limits on their step to be able to have Pod schedulable.

> **Drawbacks**
> 
> * If we update the entrypoint code and the binary changes size, the resource requirements will need to be adjusted.
>   Because Tekton releases include building the entrypoint binary, this could break a release and we'd have to backport a fix.
> * If a pod is OOMkilled due to exceeding resource requirements, the user will not be able to work around this problem.
>   This can be mitigated by setting the init container's resource requirements to higher than what it likely needs.

Setting it higher can also render the Pod unschedulable anywhere. I think this is one reason for making those requests/limits configurable.

Another drawback, at least as the PR is done, is that, it could create Pod that are invalid from the LimitRange standpoint. Aka if the value we decided for the init containers are too low or too high, the LimitRange validation will fail and the Pod won't be schedulable. In addition to be configurable by the user, this means we should also "adapt" those values (requests/limits on InitContainers) based on the LimitRange if present — aka making sure they are in the "allowed" range.

> **Alternatives**
> 
> * We could apply resource requirements from a Task's StepTemplate to init containers. However, the StepTemplate
>   is meant to be used for Steps, which likely have higher resource requirements than Tekton's init containers,
>   and this would require users to understand what Tekton's init containers do.
> * We could try to set resource requirements dynamically based on the number of Steps, the size of scripts, or other parameters.
> * We could apply resource requirements to init containers only when ResourceQuotas are present in the namespace.

The third point could be something indeed. I think it might be a mix of several of those. We should give the user a way to specify default value, and also, most likely, a "mode" of working (like "set only if ResourceQuotas or LimitRange", "set always", "set never", …)

> **Future work**
> 
> * Combine Tekton init containers into one init container ([Can the "place-tools" and "place-scripts" init-containers merge into one init-container? #4519](https://github.com/tektoncd/pipeline/issues/4519)).

In theory this is less important I think, as init containers are supposed to run in parallel and thus putting all the things in one init container shouldn't change too much — but this is all theory though 😛. The trick of merging them all in one init container is that it might make the value of requests/limits harder to decide.

Overall, I think we should at least give some options to the users on whether they want/need to have requests/limits on the init containers (and most likely their values, with us providing a decent default) 👼🏼 👍🏼

lbernick:
@wlynch suggested that we could pull these default values from a configmap, and that users could modify the configmap if there are any issues with the defaults provided. The idea here would be that most users don't have to interact with this configmap, and would only use it if the defaults caused a problem or if they wanted to do something more specific such as a guaranteed qos for pods. 

Going to close this PR as I'm not actively working on it but this can easily be reopened.

PR #4574, opened 2022-02-14
Patch temp GOPATH hack script to handle nounset option

sbwsg:
# Changes

Prior to this commit the setup-temporary-gopath.sh hack script referenced the GOPATH env variable without first checking that it was set. When `set -o nounset` is working (which it isn't on my machine 😖 ) this causes the script to exit with an error.

This commit adds a variable wrapping $GOPATH and setting a default if it's missing, which should work around the `nounset`.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


lbernick:
works for me locally!
/lgtm

vdemeester (approved):
/meow

Issue #4326, opened 2021-10-21
Support specifying resource requests at the TaskRun level


jsravn:
### Feature request

Currently resource requests can only be done at the Task level. However this is limiting when trying to reuse generic tasks. It would make more sense to me to have resources specified in the TaskRun, with the Task providing a set of defaults.

### Use case

An example that came up recently is using the [kaniko hub task](https://hub.tekton.dev/tekton/task/kaniko). The resources required will vary widely based on the contents of the Dockerfile. As is, I have to redefine this Task based on each build, making the shared task fairly useless.

Possibly related to https://github.com/tektoncd/pipeline/issues/2986.

lbernick:
Hi @jsravn, I think this use case would be addressed by #4080, which I'm currently working on, so I'm going to close this as a duplicate issue. Please feel free to reopen if needed.

triage/duplicate
/close

vdemeester:
> Hi @jsravn, I think this use case would be addressed by #4080, which I'm currently working on, so I'm going to close this as a duplicate issue. Please feel free to reopen if needed.
> 
> triage/duplicate /close

Although it might be addressed by #4080, I wonder if it should be the way it gets solved. For example, I want to use a Task from the Hub/Catalog, where I do not have any control, and this Task doesn't have any *parametrized* resource limits/requirements setup, then I want to be able to provide this at runtime.
Hence I would like to keep this one open for now.

jsravn:
I agree that it makes sense to do this generically at the TaskRun level. I think conceptually it makes sense to consider resources as an instance/run level attribute rather than that of the task.

lbernick:
Sorry this took some time, but [TEP-0094: Configuring Resources at Runtime](https://github.com/tektoncd/community/blob/main/teps/0094-configuring-resources-at-runtime.md) is now implementable.

PR #4575, opened 2022-02-14
Add Step and Sidecar Overrides to TaskRun API

lbernick:
# Changes

This commit adds TaskRunStepOverrides and TaskRunSidecarOverrides to TaskRun.Spec and
PipelineRun.Spec.PipelineTaskRunSpec, gated behind the "alpha" API flag.
This is part 1 of implementing [TEP-0094: Configuring Resource Requirements at Runtime](https://github.com/tektoncd/community/blob/main/teps/0094-configuring-resources-at-runtime.md).
Part 1 of #4326.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
[Feature] Adds API fields for Step and Sidecar Overrides to TaskRun. This feature is not yet implemented.
```


pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
```diff
@@ -112,6 +112,12 @@ func (ps *PipelineRunSpec) Validate(ctx context.Context) (errs *apis.FieldError)
 		}
 	}
 
+	if ps.TaskRunSpecs != nil {
+		for idx, trs := range ps.TaskRunSpecs {
```
sbwsg, line 116: the `range` keyword will handle `nil` values without panicking, no need to wrap in an `if` I don't think.

lbernick: good catch thanks!

afrittoli:
It looks like this might not make it for v0.33.0 - is that ok for you @lbernick ?

pritidesai:
@lbernick please add release notes 📝 It will be easier for folks to understand the changes since its labeled as a feature 🙏

afrittoli:
> @lbernick please add release notes 📝 It will be easier for folks to understand the changes since its labeled as a feature 🙏

In my understanding this change introduces the API change, documentation and validation, but no actual functionality.
Since the docs are in though, if we merged this we should say in the release notes that the feature is not actually available yet (as stated in the docs).

docs/taskruns.md
```diff
@@ -399,6 +401,56 @@ inside the `Pod`. Only the above command is affected. The `Pod's` description co
...
+  sidecarOverrides:
+    - name: logging
+      resources:
+        requests:
+          cpu: 100m
+        limits:
+          cpu: 500m
+```
+`StepOverrides` and `SidecarOverrides` must include the `name` field and may include `resources`.
+No other fields can be overridden.
```
afrittoli, line 452: NIT: maybe worth mentioning that it is not possible to override resources for steps that don't have a name

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -511,9 +511,11 @@ type PipelineTaskRun struct {
...
 // specs for a concrete Task
 type PipelineTaskRunSpec struct {
-	PipelineTaskName       string       `json:"pipelineTaskName,omitempty"`
-	TaskServiceAccountName string       `json:"taskServiceAccountName,omitempty"`
-	TaskPodTemplate        *PodTemplate `json:"taskPodTemplate,omitempty"`
+	PipelineTaskName       string                   `json:"pipelineTaskName,omitempty"`
+	TaskServiceAccountName string                   `json:"taskServiceAccountName,omitempty"`
+	TaskPodTemplate        *PodTemplate             `json:"taskPodTemplate,omitempty"`
+	StepOverrides          []TaskRunStepOverride    `json:"stepOverrides,omitempty"`
+	SidecarOverrides       []TaskRunSidecarOverride `json:"sidecarOverrides,omitempty"`
```
afrittoli, line 518: NIT: I assume that if a `PipelineRun` that specifies any of these was fetched via the `v1alpha1` API, it would lose them, which I think it probably fine, but I wonder if we should document somewhere clearly what are the fields in `v1beta` which are not in `v1alpha` and that would be lost when fetching/updating via `v1alpha1`. /cc @sbwsg @vdemeester 

Not something we need to fix as part of this PR, it just came to mind seeing the new fields 🙏

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -95,27 +103,63 @@ func validateDebug(db *TaskRunDebug) (errs *apis.FieldError) {
...
 func validateParameters(params []Param) (errs *apis.FieldError) {
-	// Template must not duplicate parameter names.
-	seen := sets.NewString()
+	var names []string
 	for _, p := range params {
-		if seen.Has(strings.ToLower(p.Name)) {
-			errs = errs.Also(apis.ErrMultipleOneOf("name").ViaKey(p.Name))
+		names = append(names, p.Name)
+	}
+	return validateNoDuplicateNames(names, false)
```
afrittoli, line 120: Nice :)

afrittoli (commented):
Thanks @lbernick, it looks good!

If you could update the release notes before I run the release, or I can also add them to the PR description if you'd like.

/approve

afrittoli:
I added a "hold" because of the release notes, but in fact those can be modified also after the merge, as long as it's done before the release notes are generated.

lbernick:
> > @lbernick please add release notes 📝 It will be easier for folks to understand the changes since its labeled as a feature 🙏
> 
> In my understanding this change introduces the API change, documentation and validation, but no actual functionality. Since the docs are in though, if we merged this we should say in the release notes that the feature is not actually available yet (as stated in the docs).

Done!

Issue #4582, opened 2022-02-16
Status of PipelineRun with Finally is flakey


devholic:
# Expected Behavior

When the tasks in `Finally` of the PipelineRun are running, 

- PipelineRun shouldn't be marked as `Completed`
- Running `Finally` tasks shouldn't be marked as `Skipped`

# Actual Behavior

- PipelineRun's `completionTime` is earlier than the last TaskRun's `completionTime`
- Running `Finally` tasks are marked as `Skipped`

# Steps to Reproduce the Problem

<details>
<summary>Problematic PipelineRun</summary>

> Some values are redacted

```yaml
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  creationTimestamp: "2022-02-16T08:14:35Z"
  generateName: cd-
  generation: 1
  labels:
    tekton.dev/pipeline: cd-8chcz
  name: cd-8chcz
  namespace: test
  resourceVersion: "512362041"
  uid: be1beefe-d8fb-4ea0-8840-1b6074d4c08a
spec:
  params:
  - name: git-repository-id
    value: <redacted>
  - name: git-repository-url
    value: <redacted>
  - name: git-repository-full-name
    value: <redacted>
  - name: git-ref
    value: <redacted>
  - name: git-revision
    value: <redacted>
  - name: git-clone-depth
    value: "10"
  - name: github-enterprise-host
    value: <redacted>
  - name: github-check-suite-id
    value: <redacted>
  - name: github-check-run-id
    value: <redacted>
  - name: github-check-run-html-url
    value: <redacted>
  - name: github-token
    value: <redacted>
  - name: pipe-log-url
    value: <redacted>
  - name: github-deployment-environment
    value: <redacted>
  - name: github-deployment-id
    value: <redacted>
  - name: github-deployment-payload
    value: '{}'
  - name: github-deployment-task
    value: deploy
  - name: pipe-workflow-file-path
    value: .pipe/deployment-cd.yaml
  pipelineSpec:
    finally:
    - name: set-success-deployment-status
      params:
      - name: DEPLOYMENT_ID
        value: $(params.github-deployment-id)
      - name: ENVIRONMENT
        value: $(params.github-deployment-environment)
      - name: GITHUB_TOKEN
        value: $(params.github-token)
      - name: REPO_FULL_NAME
        value: $(params.git-repository-full-name)
      - name: STATE
        value: success
      taskRef:
        kind: ClusterTask
        name: github-create-deployment-status
[...]

PR #4583, opened 2022-02-16
Assume task not skipped if the run is associated

devholic:
# Changes

Resolves https://github.com/tektoncd/pipeline/issues/4582

Currently, the status of PipelineRun with Finally is flakey. For example, PipelineRun's completionTime is earlier than the last TaskRun's completionTime, and also Running Finally tasks are marked as Skipped.

This issue occurs when TaskRun has no conditions. Because [the task state context is not applied to the running Final tasks](https://github.com/tektoncd/pipeline/blob/62c9b0d861bd2ae01263dfa67ba41fb0fa2bc9a9/pkg/reconciler/pipelinerun/pipelinerun.go#L635-L652) and [IsFinallySkipped](https://github.com/tektoncd/pipeline/blob/62c9b0d861bd2ae01263dfa67ba41fb0fa2bc9a9/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L314-L339) assumes [TaskRun with no conditions as not started](https://github.com/tektoncd/pipeline/blob/62c9b0d861bd2ae01263dfa67ba41fb0fa2bc9a9/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L167-L175).

This commit resolves this issue by assuming the finally task is not skipped if the run is associated.

/kind bug

# Submitter Checklist

# Release Notes

```release-note
Fixes controller with the high value of `ThreadsPerController` to report the correct status of PipelineRun, which contains Finally tasks.
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -209,6 +209,15 @@ func (t ResolvedPipelineRunTask) isCancelled() bool {
 	}
 }
 
+// isScheduled returns true when the PipelineRunTask itself has a TaskRun
+// or Run associated.
+func (t ResolvedPipelineRunTask) isScheduled() bool {
+	if t.IsCustomTask() {
+		return t.Run != nil
+	}
+	return t.TaskRun != nil
+}
+
 // isStarted returns true only if the PipelineRunTask itself has a TaskRun or
 // Run associated that has a Succeeded-type condition.
 func (t ResolvedPipelineRunTask) isStarted() bool {
@@ -249,7 +258,7 @@ func (t *ResolvedPipelineRunTask) skip(facts *PipelineRunFacts) TaskSkipStatus {
 	var skippingReason v1beta1.SkippingReason
 
 	switch {
-	case facts.isFinalTask(t.PipelineTask.Name) || t.isStarted():
+	case facts.isFinalTask(t.PipelineTask.Name) || t.isScheduled():
 		skippingReason = v1beta1.None
 	case facts.IsStopping():
 		skippingReason = v1beta1.StoppingSkip
@@ -349,7 +358,7 @@ func (t *ResolvedPipelineRunTask) IsFinallySkipped(facts *PipelineRunFacts) Task
 	var skippingReason v1beta1.SkippingReason
 
 	switch {
-	case t.isStarted():
+	case t.isScheduled():
 		skippingReason = v1beta1.None
 	case facts.checkDAGTasksDone() && facts.isFinalTask(t.PipelineTask.Name):
 		switch {
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
@@ -402,7 +402,7 @@ func (facts *PipelineRunFacts) DAGExecutionQueue() (PipelineRunState, error) {
 	return tasks, nil
 }
 
-// GetFinalTasks returns a list of final tasks without any taskRun associated with it
+// GetFinalTasks returns a list of final tasks which needs to be executed next
 // GetFinalTasks returns final tasks only when all DAG tasks have finished executing or have been skipped
 func (facts *PipelineRunFacts) GetFinalTasks() PipelineRunState {
 	tasks := PipelineRunState{}
@@ -412,7 +412,7 @@ func (facts *PipelineRunFacts) GetFinalTasks() PipelineRunState {
 	if facts.checkDAGTasksDone() {
 		// return list of tasks with all final tasks
 		for _, t := range facts.State {
-			if facts.isFinalTask(t.PipelineTask.Name) && !t.isScheduled() {
+			if facts.isFinalTask(t.PipelineTask.Name) {
 				finalCandidates.Insert(t.PipelineTask.Name)
 			}
 		}
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
@@ -412,7 +412,7 @@ func (facts *PipelineRunFacts) GetFinalTasks() PipelineRunState {
 	if facts.checkDAGTasksDone() {
 		// return list of tasks with all final tasks
 		for _, t := range facts.State {
-			if facts.isFinalTask(t.PipelineTask.Name) && !t.isSuccessful() {
+			if facts.isFinalTask(t.PipelineTask.Name) && !t.isScheduled() {
 				finalCandidates.Insert(t.PipelineTask.Name)
 			}
 		}
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -665,7 +665,7 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
 	// when expressions, so reset the skipped cache
 	pipelineRunFacts.ResetSkippedCache()
 
-	// GetFinalTasks only returns tasks when a DAG is complete
+	// GetFinalTasks only returns final tasks when a DAG is complete
 	fnextRprts := pipelineRunFacts.GetFinalTasks()
 	if len(fnextRprts) != 0 {
 		// apply the runtime context just before creating taskRuns for final tasks in queue
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
@@ -156,6 +156,12 @@ var trs = []v1beta1.TaskRun{{
 		Name:      "pipelinerun-mytask2",
 	},
 	Spec: v1beta1.TaskRunSpec{},
+}, {
+	ObjectMeta: metav1.ObjectMeta{
+		Namespace: "namespace",
+		Name:      "pipelinerun-mytask4",
+	},
+	Spec: v1beta1.TaskRunSpec{},
 }}
 
 var runs = []v1alpha1.Run{{
@@ -392,6 +398,22 @@ var finalScheduledState = PipelineRunState{{
 	PipelineTask: &pts[1],
 	TaskRunName:  "pipelinerun-mytask2",
 	TaskRun:      makeScheduled(trs[1]),
+	ResolvedTaskResources: &resources.ResolvedTaskResources{
+		TaskSpec: &task.Spec,
+	},
+}}
+
+var retryableFinalState = PipelineRunState{{
+	PipelineTask: &pts[0],
+	TaskRunName:  "pipelinerun-mytask1",
+	TaskRun:      makeSucceeded(trs[0]),
+	ResolvedTaskResources: &resources.ResolvedTaskResources{
+		TaskSpec: &task.Spec,
+	},
+}, {
+	PipelineTask: &pts[3],
+	TaskRunName:  "pipelinerun-mytask4",
+	TaskRun:      makeFailed(trs[2]),
 	ResolvedTaskResources: &resources.ResolvedTaskResources{
 		TaskSpec: &task.Spec,
 	},
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
@@ -190,6 +190,12 @@ var matrixedPipelineTask = &v1beta1.PipelineTask{
 	}},
 }
 
+func makeScheduled(tr v1beta1.TaskRun) *v1beta1.TaskRun {
+	newTr := newTaskRun(tr)
+	newTr.Status = v1beta1.TaskRunStatus{ /* explicitly empty */ }
+	return newTr
+}
+
 func makeStarted(tr v1beta1.TaskRun) *v1beta1.TaskRun {
 	newTr := newTaskRun(tr)
 	newTr.Status.Conditions[0].Status = corev1.ConditionUnknown
@@ -375,6 +381,22 @@ var oneFailedState = PipelineRunState{{
 	},
 }}
 
+var finalScheduledState = PipelineRunState{{
+	PipelineTask: &pts[0],
+	TaskRunName:  "pipelinerun-mytask1",
+	TaskRun:      makeSucceeded(trs[0]),
+	ResolvedTaskResources: &resources.ResolvedTaskResources{
+		TaskSpec: &task.Spec,
+	},
+}, {
+	PipelineTask: &pts[1],
+	TaskRunName:  "pipelinerun-mytask2",
+	TaskRun:      makeScheduled(trs[1]),
+	ResolvedTaskResources: &resources.ResolvedTaskResources{
+		TaskSpec: &task.Spec,
+	},
+}}
+
 var allFinishedState = PipelineRunState{{
 	PipelineTask: &pts[0],
 	TaskRunName:  "pipelinerun-mytask1",
@@ -545,6 +567,13 @@ func TestIsSkipped(t *testing.T) {
 		state: taskCancelled,
 		expected: map[string]bool{
 			"mytask5": false,
+		},
+	}, {
+		name:  "tasks-scheduled",
+		state: finalScheduledState,
+		expected: map[string]bool{
+			"mytask1": false,
+			"mytask2": false,
 		},
 	}, {
 		name: "tasks-parent-failed",
@@ -2653,6 +2682,132 @@ func TestResolvedPipelineRunTask_IsFinallySkipped(t *testing.T) {
 				t.Fatalf("Didn't get expected isFinallySkipped from finally task %s: %s", finallyTaskName, diff.PrintWantGot(d))
 			}
 		}
+	}
+}
+
+func TestResolvedPipelineRunTask_IsFinallySkippedByCondition(t *testing.T) {
+	task := &ResolvedPipelineRunTask{
+		TaskRunName: "dag-task",
+		TaskRun: &v1beta1.TaskRun{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "dag-task",
+			},
+			Status: v1beta1.TaskRunStatus{
+				Status: duckv1beta1.Status{
+					Conditions: []apis.Condition{{
+						Type:   apis.ConditionSucceeded,
+						Status: corev1.ConditionTrue,
+					}},
+				},
+			},
+		},
+		PipelineTask: &v1beta1.PipelineTask{
+			Name:    "dag-task",
+			TaskRef: &v1beta1.TaskRef{Name: "task"},
+		},
+	}
+	for _, tc := range []struct {
+		name  string
+		state PipelineRunState
+		want  TaskSkipStatus
+	}{{
+		name: "task started",
+		state: PipelineRunState{
+			task,
+			{
+				TaskRun: &v1beta1.TaskRun{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "final-task",
+					},
+					Status: v1beta1.TaskRunStatus{
+						Status: duckv1beta1.Status{
+							Conditions: []apis.Condition{{
+								Type:   apis.ConditionSucceeded,
+								Status: corev1.ConditionUnknown,
+							}},
+						},
+					},
+				},
+				PipelineTask: &v1beta1.PipelineTask{
+					Name:    "final-task",
+					TaskRef: &v1beta1.TaskRef{Name: "task"},
+					WhenExpressions: []v1beta1.WhenExpression{
+						{
+							Input:    "$(tasks.dag-task.status)",
+							Operator: selection.In,
+							Values:   []string{"Succeeded"},
+						},
+					},
+				},
+			},
+		},
+		want: TaskSkipStatus{
+			IsSkipped:      false,
+			SkippingReason: v1beta1.None,
+		},
+	}, {
+		name: "task scheduled",
+		state: PipelineRunState{
+			task,
+			{
+				TaskRunName: "final-task",
+				TaskRun: &v1beta1.TaskRun{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "final-task",
+					},
+					Status: v1beta1.TaskRunStatus{
+						Status: duckv1beta1.Status{
+							Conditions: []apis.Condition{ /* explicitly empty */ },
+						},
+					},
+				},
+				PipelineTask: &v1beta1.PipelineTask{
+					Name:    "final-task",
+					TaskRef: &v1beta1.TaskRef{Name: "task"},
+					WhenExpressions: []v1beta1.WhenExpression{
+						{
+							Input:    "$(tasks.dag-task.status)",
+							Operator: selection.In,
+							Values:   []string{"Succeeded"},
+						},
+					},
+				},
+			}},
+		want: TaskSkipStatus{
+			IsSkipped:      false,
+			SkippingReason: v1beta1.None,
+		},
+	}} {
+		t.Run(tc.name, func(t *testing.T) {
+			tasks := v1beta1.PipelineTaskList([]v1beta1.PipelineTask{*tc.state[0].PipelineTask})
+			d, err := dag.Build(tasks, tasks.Deps())
+			if err != nil {
+				t.Fatalf("Could not get a dag from the dag tasks %#v: %v", tc.state[0], err)
+			}
+
+			// build graph with finally tasks
+			var pts v1beta1.PipelineTaskList
+			for _, state := range tc.state[1:] {
+				pts = append(pts, *state.PipelineTask)
+			}
+			dfinally, err := dag.Build(pts, map[string][]string{})
+			if err != nil {
+				t.Fatalf("Could not get a dag from the finally tasks %#v: %v", pts, err)
+			}
+
+			facts := &PipelineRunFacts{
+				State:           tc.state,
+				TasksGraph:      d,
+				FinalTasksGraph: dfinally,
+			}
+
+			for _, state := range tc.state[1:] {
+				got := state.IsFinallySkipped(facts)
+				if d := cmp.Diff(tc.want, got); d != "" {
+					t.Errorf("IsFinallySkipped: %s", diff.PrintWantGot(d))
+				}
+			}
+		})
 	}
 }
 
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate_test.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate_test.go
@@ -1194,6 +1194,15 @@ func TestPipelineRunState_GetFinalTasks(t *testing.T) {
 		DAGTasks:           []v1beta1.PipelineTask{pts[0]},
 		finalTasks:         []v1beta1.PipelineTask{pts[1]},
 		expectedFinalTasks: PipelineRunState{},
+	}, {
+		// tasks: [ mytask1]
+		// finally: [mytask4]
+		name:               "07 - DAG tasks succeeded, return retryable final tasks",
+		desc:               "DAG task (mytask1) finished successfully - retry failed final tasks (mytask4)",
+		state:              retryableFinalState,
+		DAGTasks:           []v1beta1.PipelineTask{pts[0]},
+		finalTasks:         []v1beta1.PipelineTask{pts[3]},
+		expectedFinalTasks: PipelineRunState{retryableFinalState[1]},
 	}}
 	for _, tc := range tcs {
 		dagGraph, err := dag.Build(v1beta1.PipelineTaskList(tc.DAGTasks), v1beta1.PipelineTaskList(tc.DAGTasks).Deps())
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate_test.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate_test.go
@@ -267,6 +267,11 @@ func TestIsBeforeFirstTaskRun_WithStartedRun(t *testing.T) {
 		t.Fatalf("Expected state to be after first taskrun (Run test)")
 	}
 }
+func TestIsBeforeFirstTaskRun_WithSucceededTask(t *testing.T) {
+	if finalScheduledState.IsBeforeFirstTaskRun() {
+		t.Fatalf("Expected state to be after first taskrun")
+	}
+}
 
 func TestGetNextTasks(t *testing.T) {
 	tcs := []struct {
@@ -354,6 +359,26 @@ func TestGetNextTasks(t *testing.T) {
 		state:        oneFailedState,
 		candidates:   sets.NewString("mytask1", "mytask2"),
 		expectedNext: []*ResolvedPipelineRunTask{oneFailedState[1]},
+	}, {
+		name:         "final-task-scheduled-no-candidates",
+		state:        finalScheduledState,
+		candidates:   sets.NewString(),
+		expectedNext: []*ResolvedPipelineRunTask{},
+	}, {
+		name:         "final-task-finished-one-candidate",
+		state:        finalScheduledState,
+		candidates:   sets.NewString("mytask1"),
+		expectedNext: []*ResolvedPipelineRunTask{},
+	}, {
+		name:         "final-task-finished-other-candidate",
+		state:        finalScheduledState,
+		candidates:   sets.NewString("mytask2"),
+		expectedNext: []*ResolvedPipelineRunTask{},
+	}, {
+		name:         "final-task-finished-both-candidate",
+		state:        finalScheduledState,
+		candidates:   sets.NewString("mytask1", "mytask2"),
+		expectedNext: []*ResolvedPipelineRunTask{},
 	}, {
 		name:         "all-finished-no-candidates",
 		state:        allFinishedState,
@@ -1160,6 +1185,15 @@ func TestPipelineRunState_GetFinalTasks(t *testing.T) {
 		DAGTasks:           []v1beta1.PipelineTask{pts[0]},
 		finalTasks:         []v1beta1.PipelineTask{pts[1]},
 		expectedFinalTasks: PipelineRunState{oneFinishedState[1]},
+	}, {
+		// tasks: [ mytask1]
+		// finally: [mytask2]
+		name:               "06 - DAG tasks succeeded, final tasks scheduled - no final tasks",
+		desc:               "DAG task (mytask1) finished successfully - final task (mytask2) scheduled - no final tasks",
+		state:              finalScheduledState,
+		DAGTasks:           []v1beta1.PipelineTask{pts[0]},
+		finalTasks:         []v1beta1.PipelineTask{pts[1]},
+		expectedFinalTasks: PipelineRunState{},
 	}}
 	for _, tc := range tcs {
 		dagGraph, err := dag.Build(v1beta1.PipelineTaskList(tc.DAGTasks), v1beta1.PipelineTaskList(tc.DAGTasks).Deps())


afrittoli:
Thank you @devholic !

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -164,6 +164,15 @@ func (t ResolvedPipelineRunTask) IsCancelled() bool {
 	return c != nil && c.IsFalse() && c.Reason == v1beta1.TaskRunReasonCancelled.String()
 }
 
+// IsScheduled returns true when the PipelineRunTask itself has a TaskRun
+// or Run associated.
+func (t ResolvedPipelineRunTask) IsScheduled() bool {
+	if t.IsCustomTask() {
+		return t.Run != nil
+	}
+	return t.TaskRun != nil
```
pritidesai, line 173: @afrittoli @jerop @vdemeester my understanding was this is not possible 🙃  A `taskRun` or `run` is associated must have a `succeeded` condition with `unknown` i.e. `running` status.

We rely on the same check for the non-final tasks:

https://github.com/tektoncd/pipeline/blob/4ad8f33f8f30ccd4ffc22d57d16dd87420721094/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L207

devholic: I'm still not sure about the root cause (my current assumption is a timing issue), but I think it makes sense to check if `Run` exists only in the `IsStarted()` method like below.

```go
func (t ResolvedPipelineRunTask) IsStarted() bool {
	if t.IsCustomTask() {
		return t.Run != nil

	}
	return t.TaskRun != nil
}
```

Whether it has a `Succeeded-type condition` or not, I think we can assume the task is started since the reconciler found that it has associated `Run`.

What do you think?

devholic:
Could someone give comments on this PR or issue(#4582)? 🙏 
@pritidesai @afrittoli @jerop @vdemeester

jerop (commented):
thank you for the contribution @devholic!

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
jerop: @devholic makes sense to go with that option 👍🏾 

cc @tektoncd/core-maintainers

pritidesai:
There are two separate entities involved, `taskRun` golang object v/s actual `taskRun` CRD.

PipelineRun controller creates a new `taskRun` object if it does not exist followed by creating an actual `taskRun` CRD.

The taskRun controller relies on the fact that the `startTime` of a `taskRun` is either nil or set to zero time to determine whether the `taskRun` has started or not. If `taskRun` has not started, the controller initialized the status condition (setting it to running) and sets the start time.

If you are running into this kind of issue:

```
Currently, the status of PipelineRun with Finally is flakey. For example, PipelineRun's completionTime is earlier than the last
TaskRun's completionTime, and also Running Finally tasks are marked as Skipped.
```

Checking if the `taskRun` golang object exists might not fix this issue with multi-threaded controller. Most likely what's happening here is while one instance of pipelineRun controller is creating the taskRun object followed by `CRD` and the taskRun controller is setting the condition, the another instance of pipelineRun controller is checking for the condition.

We have a test to confirm this does not happen in our CI/CD but this could be cluster/configuration specific and might not be feasible to replicate and/or reproduce it. I would highly recommend adding such multi-threaded e2e test in our CI/CD if possible.

I would recommend to add an additional check if the startTime is set, very similar to the check `(tr *TaskRun) HasStarted()` which taskRun controller relies on. Also, any changes you introduce, please verify them in your cluster to make sure they are solving this issue.

pritidesai:
We highly appreciate any additional best practices or troubleshooting documentation from your experience [here](https://github.com/tektoncd/pipeline/blob/main/docs/tekton-controller-performance-configuration.md).

devholic:
@pritidesai Thanks for the review and comment 🙇‍♂️

> PipelineRun controller creates a new `taskRun` object if it does not exist followed by creating an actual `taskRun` CRD.

I'm not sure if I understand this correctly, but I think TaskRun (Object / CRD) is created in [`createTaskRun`](https://github.com/tektoncd/pipeline/blob/6738db7a179dd93adef40ac29a375cdfdf8823ee/pkg/reconciler/pipelinerun/pipelinerun.go#L722-L777) only - which seems create an actual `TaskRun` only 🤔. Could you give some more details?

> Checking if the `taskRun` golang object exists might not fix this issue with multi-threaded controller. Most likely what's happening here is while one instance of pipelineRun controller is creating the taskRun object followed by `CRD` and the taskRun controller is setting the condition, the another instance of pipelineRun controller is checking for the condition.

As far as I know, race condition will not happen even in the multi-threaded controller. [`workqueue`](https://github.com/tektoncd/pipeline/blob/6738db7a179dd93adef40ac29a375cdfdf8823ee/vendor/k8s.io/client-go/util/workqueue/queue.go), which is used internally, [will not enqueue the request when the same request is processing](https://github.com/tektoncd/pipeline/blob/6738db7a179dd93adef40ac29a375cdfdf8823ee/vendor/k8s.io/client-go/util/workqueue/queue.go#L113-L133) and [`Get()`](https://github.com/tektoncd/pipeline/blob/6738db7a179dd93adef40ac29a375cdfdf8823ee/vendor/k8s.io/client-go/util/workqueue/queue.go#L144-L166) (dequeue) is also a thread-safe call.

Please let me know if my understanding is wrong.

> We have a test to confirm this does not happen in our CI/CD but this could be cluster/configuration specific and might not be feasible to replicate and/or reproduce it. I would highly recommend adding such multi-threaded e2e test in our CI/CD if possible.

I'll try to figure out the method to reproduce this behavior as much as I can 🙏 

> Also, any changes you introduce, please verify them in your cluster to make sure they are solving this issue.

I'll add test results to this thread 🙂

devholic:
@pritidesai I wrote a test script to validate, and I can verify this patch works.
Please let me know if you need more information about this test 🙇‍♂️ 

<details>
<summary>Sample Task</summary>

```yaml
# task.yaml
kind: Task
metadata:
  name: dummy-task
spec:
  params:
    - name: sleep-second
      type: string
      default: "3"
    - name: message
      type: string
      default: "hello, world"
  steps:
    - name: echo
      image: alpine:3.14.3
      script: |
        #!/usr/bin/env sh
        sleep "${sleep-second}"
        echo "${message}"
```
</details>

<details>
<summary>Sample PipelineRun</summary>

```yaml
# pr.yaml 
kind: PipelineRun
metadata:
  generateName: test-
spec:
  pipelineSpec:
    params:
      - name: base-image
        type: string
        default: alpine:3.15.1
      - name: env
        type: string
        default: test
      - name: task
        type: string
        default: deploy
    tasks:
      - name: clone
        when:
          - input: $(params.env)
            operator: in
            values:
              - stage
              - production
        params:
          - name: message
            value: "clone"
        taskRef:
          kind: Task
          name: dummy-task
      - name: deploy
        runAfter:
          - clone
        when:
          - input: $(params.task)
            operator: in
            values:
              - deploy
        params:
          - name: message
            value: "deploy"
        taskRef:
          kind: Task
          name: dummy-task
      - name: deploy-changed-envs
        runAfter:
          - clone
        when:
          - input: $(params.task)
            operator: in
            values:
              - deploy:envs
        params:
          - name: message
            value: "deploy-changed-envs"
        taskRef:
          kind: Task
          name: dummy-task
    finally:
      - name: set-success-deployment-status
        when:
          - input: Succeeded
            operator: in
            values:
              - $(tasks.deploy.status)
              - $(tasks.deploy-changed-envs.status)
        params:
          - name: message
            value: "set-success-deployment-status"
        taskRef:
          kind: Task
          name: dummy-task
      - name: set-failure-deployment-status
        when:
          - input: Failed
            operator: in
            values:
              - $(tasks.deploy.status)
              - $(tasks.deploy-changed-envs.status)
        params:
          - name: message
            value: "set-failure-deployment-status"
        taskRef:
          kind: Task
          name: dummy-task
      - name: set-error-deployment-status
        when:
          - input: $(tasks.clone.status)
            operator: in
            values:
              - None
        params:
          - name: message
            value: "set-error-deployment-status"
        taskRef:
          kind: Task
          name: dummy-task
```
</details>

<details>
<summary>Test script</summary>

```sh
#!/usr/bin/env bash

readonly EXPECTED_SKIPPED_TASK_COUNT="5"
readonly TEST_PIPELINE_COUNT=100

NAMESPACE="${NAMESPACE:-tkn-multi-thread-test}"

script=$(realpath "${0}")
base_dir=$(dirname "${script}")

# Print current controller configurations
echo "[!] Current tekton controller configurations:"
ctrl_deployment=$(kubectl get deploy -n tekton-pipelines tekton-pipelines-controller -o=yaml)

ctrl_image=$(yq '.spec.template.spec.containers[0].image' <<<"${ctrl_deployment}")
echo "Image: ${ctrl_image}"

while read -r flag; do
  echo "Custom flag: ${flag}"
done < <(yq '.spec.template.spec.containers[0].args[] | select(. == "*=*")' <<<"${ctrl_deployment}")

# Ensure namespace exists
echo "[!] Ensuring namespace ${NAMESPACE}..."
kubectl create namespace "${NAMESPACE}" --dry-run=client -o=yaml | kubectl apply -f -

# Ensure task exists
echo "[!] Ensuring task..."
kubectl apply -n="${NAMESPACE}" -f="${base_dir}/task.yaml"

# Create PipelineRuns
echo "[!] Creating PipelineRuns..."
pipeline_runs=()

for ((i = 1; i <= TEST_PIPELINE_COUNT; i++)); do
  created=$(kubectl create -n="${NAMESPACE}" -f="${base_dir}/pr.yaml" -o=name)
  echo "${created} created"
  pipeline_runs+=("${created}")
done

# Validate PipelineRuns
echo "[!] Validating created PipelineRuns..."
flaky_pipeline_runs=()

for pr in "${pipeline_runs[@]}"; do
  printf 'Validating %s... ' "${pr}"

  if ! kubectl wait -n="${NAMESPACE}" --for=condition=Succeeded --timeout=60s "${pr}" >/dev/null 2>&1; then
    printf 'Failed to wait. Skipping...\n'
    continue
  fi

  result="Valid"
  skipped_task_cnt=$(kubectl get -n="${NAMESPACE}" "${pr}" -o yaml | yq '.status.skippedTasks | length')
  if [ "${skipped_task_cnt}" != "${EXPECTED_SKIPPED_TASK_COUNT}" ]; then
    result="Invalid (Actual: ${skipped_task_cnt}, Expected ${EXPECTED_SKIPPED_TASK_COUNT})"
    flaky_pipeline_runs+=("${pr}")
  fi
  printf '%s\n' "${result}"
done

echo "Flaky PipelineRuns: ${#flaky_pipeline_runs[@]}/${TEST_PIPELINE_COUNT}"
```
</details>

| Environment | Flaky PipelineRuns | Log                                                                                         |
| ----------- | ------------------ | ------------------------------------------------------------------------------------------- |
| v0.33.1     | 65/100             | [v0.33.1-result.txt](https://github.com/tektoncd/pipeline/files/8282004/v0.33.1-result.txt) |
| v0.33.3     | 73/100             | [v0.33.3-result.txt](https://github.com/tektoncd/pipeline/files/8282033/v0.33.3-result.txt) |
| Patch       | 0/100              | [patch-result.txt](https://github.com/tektoncd/pipeline/files/8281998/patch-result.txt)     |

---

For e2e-test, I need some time for research and reproducing problematic environment. Can we handle it on https://github.com/tektoncd/plumbing/issues/1017 ?

pritidesai:
thank you @devholic for all the efforts you have put in, very much appreciated 👍 

I am trying to run the tests you have added here against 0.33.1 (with default threads-per-controller).  And here is the result:

```
Events:
  Type    Reason     Age                From         Message
  ----    ------     ----               ----         -------
  Normal  Started    26s (x2 over 26s)  PipelineRun  
  Normal  Running    26s (x2 over 26s)  PipelineRun  Tasks Completed: 0 (Failed: 0, Cancelled 0), Incomplete: 4, Skipped: 2
  Normal  Running    13s                PipelineRun  Tasks Completed: 1 (Failed: 0, Cancelled 0), Incomplete: 2, Skipped: 3
  Normal  Succeeded  13s                PipelineRun  Tasks Completed: 1 (Failed: 0, Cancelled 0), Skipped: 5
  Normal  Running    6s                 PipelineRun  Tasks Completed: 2 (Failed: 0, Cancelled 0), Incomplete: 1, Skipped: 3
  Normal  Succeeded  6s (x2 over 6s)    PipelineRun  Tasks Completed: 3 (Failed: 0, Cancelled 0), Skipped: 3
```

The taskruns:

```
🗂  Taskruns

 NAME                                         TASK NAME                       STARTED         DURATION     STATUS
 ∙ test-dt7bt-set-error-deployment-status     set-error-deployment-status     2 minutes ago   7 seconds    Succeeded
 ∙ test-dt7bt-set-success-deployment-status   set-success-deployment-status   2 minutes ago   6 seconds    Succeeded
 ∙ test-dt7bt-deploy                          deploy                          2 minutes ago   13 seconds   Succeeded

```

The task `clone` is skipped because the `when` expression does not succeed - `test` in `[stage, production]`.  `deploy` runs successfully but `deploy-changed-envs` skipped because of the `when` expression.

The finally task `set-failure-deployment-status` also skipped.

```
 kubectl get pr test-dt7bt -o json | jq .status.skippedTasks 
[
  {
    "name": "clone",
    "whenExpressions": [
      {
        "input": "test",
        "operator": "in",
        "values": [
          "stage",
          "production"
        ]
      }
    ]
  },
  {
    "name": "deploy-changed-envs",
    "whenExpressions": [
      {
        "input": "deploy",
        "operator": "in",
        "values": [
          "deploy:envs"
        ]
      }
    ]
  },
  {
    "name": "set-failure-deployment-status",
    "whenExpressions": [
      {
        "input": "Failed",
        "operator": "in",
        "values": [
          "Succeeded",
          "None"
        ]
      }
    ]
  }
]
```

Are there any parameter setting off here in the above examples compared to your tests?  Also, wondering about the expectation in the results:

```
Validating pipelinerun.tekton.dev/test-9vhtp... Invalid (Actual: 6, Expected 5)
```

What is this evaluating, number of taskRuns?

devholic:
Also, I can confirm that this is happening when the `scopeWhenExpressionsToTask` feature-flag is set to `true`.

<details>
<summary>test-l6jkv.yaml</summary>
```yaml
kind: PipelineRun
metadata:
  creationTimestamp: "2022-03-17T17:53:40Z"
  generateName: test-
  generation: 1
  labels:
    tekton.dev/pipeline: test-l6jkv
  name: test-l6jkv
  namespace: tkn-multi-thread-test
  resourceVersion: "663185561"
  uid: ce0f896a-01fb-44ca-b1a4-d3f903c9362b
spec:
  pipelineSpec:
    finally:
    - name: set-success-deployment-status
      params:
      - name: message
        value: set-success-deployment-status
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: Succeeded
        operator: in
        values:
        - $(tasks.deploy.status)
        - $(tasks.deploy-changed-envs.status)
    - name: set-failure-deployment-status
      params:
      - name: message
        value: set-failure-deployment-status
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: Failed
        operator: in
        values:
        - $(tasks.deploy.status)
        - $(tasks.deploy-changed-envs.status)
    - name: set-error-deployment-status
      params:
      - name: message
        value: set-error-deployment-status
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: $(tasks.clone.status)
        operator: in
        values:
        - None
    params:
    - default: alpine:3.15.1
      name: base-image
      type: string
    - default: test
      name: env
      type: string
    - default: deploy
      name: task
      type: string
    tasks:
    - name: clone
      params:
      - name: message
        value: clone
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: $(params.env)
        operator: in
        values:
        - stage
        - production
    - name: deploy
      params:
      - name: message
        value: deploy
      runAfter:
      - clone
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: $(params.task)
        operator: in
        values:
        - deploy
    - name: deploy-changed-envs
      params:
      - name: message
        value: deploy-changed-envs
      runAfter:
      - clone
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: $(params.task)
        operator: in
        values:
        - deploy:envs
  podTemplate:
    affinity:
      nodeAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
          nodeSelectorTerms:
          - matchExpressions:
            - key: node-role.kubernetes.io/build
              operator: Exists
    tolerations:
    - effect: NoSchedule
      operator: Exists
    - effect: NoExecute
      operator: Exists
  serviceAccountName: default
  timeout: 1h0m0s
status:
  completionTime: "2022-03-17T17:53:57Z"
  conditions:
  - lastTransitionTime: "2022-03-17T17:53:57Z"
    message: 'Tasks Completed: 1 (Failed: 0, Cancelled 0), Skipped: 5'
    reason: Completed
    status: "True"
    type: Succeeded
  pipelineSpec:
    finally:
    - name: set-success-deployment-status
      params:
      - name: message
        value: set-success-deployment-status
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: Succeeded
        operator: in
        values:
        - $(tasks.deploy.status)
        - $(tasks.deploy-changed-envs.status)
    - name: set-failure-deployment-status
      params:
      - name: message
        value: set-failure-deployment-status
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: Failed
        operator: in
        values:
        - $(tasks.deploy.status)
        - $(tasks.deploy-changed-envs.status)
    - name: set-error-deployment-status
      params:
      - name: message
        value: set-error-deployment-status
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: $(tasks.clone.status)
        operator: in
        values:
        - None
    params:
    - default: alpine:3.15.1
      name: base-image
      type: string
    - default: test
      name: env
      type: string
    - default: deploy
      name: task
      type: string
    tasks:
    - name: clone
      params:
      - name: message
        value: clone
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: $(params.env)
        operator: in
        values:
        - stage
        - production
    - name: deploy
      params:
      - name: message
        value: deploy
      runAfter:
      - clone
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: $(params.task)
        operator: in
        values:
        - deploy
    - name: deploy-changed-envs
      params:
      - name: message
        value: deploy-changed-envs
      runAfter:
      - clone
      taskRef:
        kind: Task
        name: dummy-task
      when:
      - input: $(params.task)
        operator: in
        values:
        - deploy:envs
  skippedTasks:
  - name: clone
    whenExpressions:
    - input: test
      operator: in
      values:
      - stage
      - production
  - name: deploy-changed-envs
    whenExpressions:
    - input: deploy
      operator: in
      values:
      - deploy:envs
  - name: set-success-deployment-status
    whenExpressions:
    - input: Succeeded
      operator: in
      values:
      - $(tasks.deploy.status)
      - $(tasks.deploy-changed-envs.status)
  - name: set-failure-deployment-status
    whenExpressions:
    - input: Failed
      operator: in
      values:
      - Succeeded
      - None
  - name: set-error-deployment-status
    whenExpressions:
    - input: $(tasks.clone.status)
      operator: in
      values:
      - None
  startTime: "2022-03-17T17:53:40Z"
  taskRuns:
    test-l6jkv-deploy:
      pipelineTaskName: deploy
      status:
        completionTime: "2022-03-17T17:53:54Z"
        conditions:
        - lastTransitionTime: "2022-03-17T17:53:54Z"
          message: All Steps have completed executing
          reason: Succeeded
          status: "True"
          type: Succeeded
        podName: test-l6jkv-deploy-pod
        startTime: "2022-03-17T17:53:41Z"
        steps:
        - container: step-echo
          imageID: docker.io/library/alpine@sha256:635f0aa53d99017b38d1a0aa5b2082f7812b03e3cdb299103fe77b5c8a07f1d2
          name: echo
          terminated:
            containerID: containerd://4c2c5f425eda7e80ee13e545acc3f8723b68260eba0f91a44d040f8c64634257
            exitCode: 0
            finishedAt: "2022-03-17T17:53:53Z"
            reason: Completed
            startedAt: "2022-03-17T17:53:53Z"
        taskSpec:
          params:
          - default: "3"
            name: sleep-second
            type: string
          - default: hello, world
            name: message
            type: string
          steps:
          - image: alpine:3.14.3
            name: echo
            resources: {}
            script: |
              #!/usr/bin/env sh
              sleep "${sleep-second}"
              echo "${message}"
      whenExpressions:
      - input: deploy
        operator: in
        values:
        - deploy
    test-l6jkv-set-error-deployment-status:
      pipelineTaskName: set-error-deployment-status
      status:
        completionTime: "2022-03-17T17:54:10Z"
        conditions:
        - lastTransitionTime: "2022-03-17T17:54:10Z"
          message: All Steps have completed executing
          reason: Succeeded
          status: "True"
          type: Succeeded
        podName: test-l6jkv-set-error-deployment-status-pod
        startTime: "2022-03-17T17:53:58Z"
        steps:
        - container: step-echo
          imageID: docker.io/library/alpine@sha256:635f0aa53d99017b38d1a0aa5b2082f7812b03e3cdb299103fe77b5c8a07f1d2
          name: echo
          terminated:
            containerID: containerd://b8356179f4d3e4c959e8b1e0cbcf2e6757e9814e3206b43a31f1a0b9428307d8
            exitCode: 0
            finishedAt: "2022-03-17T17:54:09Z"
            reason: Completed
            startedAt: "2022-03-17T17:54:09Z"
        taskSpec:
          params:
          - default: "3"
            name: sleep-second
            type: string
          - default: hello, world
            name: message
            type: string
          steps:
          - image: alpine:3.14.3
            name: echo
            resources: {}
            script: |
              #!/usr/bin/env sh
              sleep "${sleep-second}"
              echo "${message}"
      whenExpressions:
      - input: None
        operator: in
        values:
        - None
    test-l6jkv-set-success-deployment-status:
      pipelineTaskName: set-success-deployment-status
      status:
        completionTime: "2022-03-17T17:54:08Z"
        conditions:
        - lastTransitionTime: "2022-03-17T17:54:08Z"
          message: All Steps have completed executing
          reason: Succeeded
          status: "True"
          type: Succeeded
        podName: test-l6jkv-set-success-deployment-status-pod
        startTime: "2022-03-17T17:53:56Z"
        steps:
        - container: step-echo
          imageID: docker.io/library/alpine@sha256:635f0aa53d99017b38d1a0aa5b2082f7812b03e3cdb299103fe77b5c8a07f1d2
          name: echo
          terminated:
            containerID: containerd://8e8bb171e6e3022a9447d7c3f297c2d4618fdb16bb168b0558ba5dd0d0449ae7
            exitCode: 0
            finishedAt: "2022-03-17T17:54:07Z"
            reason: Completed
            startedAt: "2022-03-17T17:54:07Z"
        taskSpec:
          params:
          - default: "3"
            name: sleep-second
            type: string
          - default: hello, world
            name: message
            type: string
          steps:
          - image: alpine:3.14.3
            name: echo
            resources: {}
            script: |
              #!/usr/bin/env sh
              sleep "${sleep-second}"
              echo "${message}"
      whenExpressions:
      - input: Succeeded
        operator: in
        values:
        - Succeeded
        - None
```
</details>

devholic:
Here is the current feature flag configuration:

```yaml
data:
  disable-affinity-assistant: "true"
  disable-creds-init: "false"
  enable-api-fields: stable
  enable-custom-tasks: "false"
  enable-tekton-oci-bundles: "false"
  require-git-ssh-secret-known-hosts: "false"
  running-in-environment-with-injected-sidecars: "true"
  scope-when-expressions-to-task: "true"
```

pritidesai:
This is very interesting, thank you once again @devholic for all this data! 🤗 

> If scopeWhenExpressionsToTask flag is set to false, only set-failure-deployment-status task is expected to be run. So the skipped task count should be 5, not 6 (all skipped).

So based on the pipeline definition, `set-failure-deployment-status` will not be executed since `FAILED` is not in `[NONE, NONE]`. I think the expectation here is to have `set-error-deployment-status` executed, please confirm?

I am noticing multiple issues here:

You mentioned the context variable is not getting substituted with an actual value which is a clear bug. This should work. @jerop we will have to isolate the issue here, whether the context variable is not being replaced at all or is this issue in the when expression reporting unresolved context variable? 🤔   

```
  # set-error-deployment-status is in the both .status.skippedTasks and .status.tasksRuns
  - name: set-error-deployment-status
    whenExpressions:
    - input: $(tasks.clone.status) # Note that this input is not templated - which I mentioned in the orignal issue
```

I think the `taskRun` created for the `set-error-deployment-status` proves that the `skippedTasks` reporting has a bug. Is this fix solving the `skippedTasks` in the status and the pipelineRun competition time?

devholic:
@pritidesai 

> > If scopeWhenExpressionsToTask flag is set to false, only set-failure-deployment-status task is expected to be run. So the skipped task count should be 5, not 6 (all skipped).
> 
> So based on the pipeline definition, `set-failure-deployment-status` will not be executed since `FAILED` is not in `[NONE, NONE]`. I think the expectation here is to have `set-error-deployment-status` executed, please confirm?

Yes, only `set-error-deployment-status` is expected to be executed 🙏 

> 
> I am noticing multiple issues here:
> 
> You mentioned the context variable is not getting substituted with an actual value which is a clear bug. This should work. @jerop we will have to isolate the issue here, whether the context variable is not being replaced at all or is this issue in the when expression reporting unresolved context variable? 🤔
> 
> ```
>   # set-error-deployment-status is in the both .status.skippedTasks and .status.tasksRuns
>   - name: set-error-deployment-status
>     whenExpressions:
>     - input: $(tasks.clone.status) # Note that this input is not templated - which I mentioned in the orignal issue
> ```

Let me summarize:

### Problem

Flaky PipelineRuns are observed when the high value of `threads-per-controller` is set.

### Observed behaviors

- Even the `Final` task has an associated TaskRun, the controller marks it as skipped.
  - So both `.status.skippedTasks` and `.status.tasksRuns` has the same `Final` task (duplicated).
  - Duplicated `Final` task's variables in `.status.skippedTasks` are not templated.

### Causes

#### Even the `Final` task is running, the controller marks it as skipped.

- Reconciler checks the `Succeeded` condition when calculating the Task is skipped or schedulable (should run, but has no associated `TaskRun` yet)
  - When the `Final` task is `scheduled` has an associated `TaskRun` without the `Succeeded` condition, it can not be assumed as `Started`. ([ref](https://github.com/tektoncd/pipeline/blob/e07ae1b260c71497c789115bd9eb3914c105b5cb/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L206-L214))
  - Suggestion: Use `IsScheduled` instead of `IsStarted` when calculating the Task is skipped or schedulable.

#### Duplicated `Final` task's variables in `.status.skippedTasks` are not templated.

- Task state context is applied to the schedulable `Final` tasks only. [ref](https://github.com/tektoncd/pipeline/blob/62c9b0d861bd2ae01263dfa67ba41fb0fa2bc9a9/pkg/reconciler/pipelinerun/pipelinerun.go#L635-L652)
  - This is a side-effect of the above cause because the returned tasks from the `GetFinalTasks` call affects this.

---

I think templating failure is a side-effect of the TaskRun without the `Succeeded` condition, isolating this might be ambiguous. WDYT?

> I think the `taskRun` created for the `set-error-deployment-status` proves that the `skippedTasks` reporting has a bug. Is this fix solving the `skippedTasks` in the status and the pipelineRun competition time?

Yes, this fixes both issues since they are caused by the TaskRuns without the `Succeeded` condition.

pritidesai:
Thank you @devholic I tried running the same example with the configuration you have mentioned `threads-per-controller` to `16` and `scope-when-expressions-to-task` set to `false` with the changes in this PR, the example PR ran as expected 🎉 

But 😞 the results from running the same example with these changes and setting `threads-per-controller` to `16` and `scope-when-expressions-to-task` set to `true` are still odd:

Skipped Tasks: 

```
 kubectl get pr test-x6n8c -o json | jq .status.skippedTasks
[
  {
    "name": "clone",
    "whenExpressions": [
      {
        "input": "test",
        "operator": "in",
        "values": [
          "stage",
          "production"
        ]
      }
    ]
  },
  {
    "name": "deploy-changed-envs",
    "whenExpressions": [
      {
        "input": "deploy",
        "operator": "in",
        "values": [
          "deploy:envs"
        ]
      }
    ]
  },
  {
    "name": "set-success-deployment-status",
    "whenExpressions": [
      {
        "input": "Succeeded",
        "operator": "in",
        "values": [
          "$(tasks.deploy.status)",
          "$(tasks.deploy-changed-envs.status)"
        ]
      }
    ]
  },
  {
    "name": "set-failure-deployment-status",
    "whenExpressions": [
      {
        "input": "Failed",
        "operator": "in",
        "values": [
          "Succeeded",
          "None"
        ]
      }
    ]
  },
  {
    "name": "set-error-deployment-status",
    "whenExpressions": [
      {
        "input": "$(tasks.clone.status)",
        "operator": "in",
        "values": [
          "None"
        ]
      }
    ]
  }
]
```

The number of `skippedTasks` is equal to 5 out of 6 total tasks with 3 `taskRuns`:

```
 tkn pr describe                                            

🗂  Taskruns

 NAME                                         TASK NAME                       STARTED         DURATION    STATUS
 ∙ test-x6n8c-set-success-deployment-status   set-success-deployment-status   3 minutes ago   1 minute    Succeeded
 ∙ test-x6n8c-set-error-deployment-status     set-error-deployment-status     3 minutes ago   1 minute    Succeeded
 ∙ test-x6n8c-deploy                          deploy                          3 minutes ago   6 seconds   Succeeded
 ```

**Comment by devholic** [comment]
> But 😞 the results from running the same example with these changes and setting `threads-per-controller` to `16` and `scope-when-expressions-to-task` set to `true` are still odd:

Thats seems weird 🤔 
Could you share the whole PipelineRun(`test-x6n8c`)'s yaml or json?

**Comment by devholic** [comment]
I re-tested to figure out this happens with `scope-when-expressions-to-task: "true"` as you commented, but I cannot reproduce the failure as far as I tried 🤔

<details>
<summary>test.sh (validating both TaskRun / Skipped Task count</summary>

```sh
#!/usr/bin/env bash

readonly EXPECTED_SKIPPED_TASK_COUNT="3"
readonly EXPECTED_TASK_RUN_COUNT="3"
readonly TEST_PIPELINE_COUNT=250

NAMESPACE="${NAMESPACE:-tkn-multi-thread-test}"

script=$(realpath "${0}")
base_dir=$(dirname "${script}")

# Print current controller configurations
echo "[!] Current tekton controller configurations:"
ctrl_deployment=$(kubectl get deploy -n tekton-pipelines tekton-pipelines-controller -o=yaml)

ctrl_image=$(yq '.spec.template.spec.containers[0].image' <<<"${ctrl_deployment}")
echo "Image: ${ctrl_image}"

while read -r flag; do
  echo "Custom flag: ${flag}"
done < <(yq '.spec.template.spec.containers[0].args[] | select(. == "*=*")' <<<"${ctrl_deployment}")

# Ensure namespace exists
echo "[!] Ensuring namespace ${NAMESPACE}..."
kubectl create namespace "${NAMESPACE}" --dry-run=client -o=yaml | kubectl apply -f -

# Ensure task exists
echo "[!] Ensuring task..."
kubectl apply -n="${NAMESPACE}" -f="${base_dir}/task.yaml"

# Create PipelineRuns
echo "[!] Creating PipelineRuns..."
pipeline_runs=()

for ((i = 1; i <= TEST_PIPELINE_COUNT; i++)); do
  created=$(kubectl create -n="${NAMESPACE}" -f="${base_dir}/pr.yaml" -o=name)
  echo "${created} created"
  pipeline_runs+=("${created}")
done

# Validate PipelineRuns
echo "[!] Validating created PipelineRuns..."
flaky_pipeline_runs=()

for pr in "${pipeline_runs[@]}"; do
  printf 'Validating %s... ' "${pr}"

  if ! kubectl wait -n="${NAMESPACE}" --for=condition=Succeeded --timeout=600s "${pr}" >/dev/null 2>&1; then
    printf 'Failed to wait. Skipping...\n'
    continue
  fi

  result="Valid"
  pipeline_run=$(kubectl get -n="${NAMESPACE}" "${pr}" -o yaml)
  skipped_task_cnt=$(yq '.status.skippedTasks | length' <<<"${pipeline_run}")
  task_run_cnt=$(yq '.status.taskRuns | length' <<<"${pipeline_run}")

  if [ "${skipped_task_cnt}" != "${EXPECTED_SKIPPED_TASK_COUNT}" ] ||
    [ "${task_run_cnt}" != "${EXPECTED_TASK_RUN_COUNT}" ]; then
    result="Invalid (Skipped: ${skipped_task_cnt}/${EXPECTED_SKIPPED_TASK_COUNT}, Run: ${task_run_cnt}/${EXPECTED_TASK_RUN_COUNT})"
    flaky_pipeline_runs+=("${pr}")
  fi

  printf '%s\n' "${result}"
done

echo "Flaky PipelineRuns: ${#flaky_pipeline_runs[@]}/${TEST_PIPELINE_COUNT}"
```
</details>

[retest-result.txt](https://github.com/tektoncd/pipeline/files/8304201/retest-result.txt)

pritidesai:
> > But 😞 the results from running the same example with these changes and setting `threads-per-controller` to `16` and `scope-when-expressions-to-task` set to `true` are still odd:
> 
> Thats seems weird 🤔  Could you share the whole PipelineRun(`test-x6n8c`)'s yaml or json?

Sure, here it is:

[test-x6n8c-status.txt](https://github.com/tektoncd/pipeline/files/8306299/test-x6n8c-status.txt)

Interesting you are not experiencing the same 🤔 

You are providing very valuable troubleshooting, please feel free to contribute them in another PR, very appreciated 🙏 

I will look into this again next week. Meanwhile if you are sure about this fix, please add more unit tests if possible. thank you once again! 🤗

devholic:
Hello @pritidesai 🙂 
Sorry to bother you, but are there any updates on this issue? 🙏

pritidesai:
> Hello @pritidesai 🙂 Sorry to bother you, but are there any updates on this issue? 🙏

sorry @devholic I haven't been able to look into this yet. I am adding this into the current milestone so that other folks can get a chance to review this.

pritidesai:
meanwhile, please rebase the changes 🙏

abayer:
For what it's worth, I was able to reproduce this, with `-threads-per-controller "32"` and using the `PipelineRun` in https://github.com/tektoncd/pipeline/pull/4583#issuecomment-1070382580, the `test.sh` in https://github.com/tektoncd/pipeline/pull/4583#issuecomment-1072336432, and a slightly modified `Task`:

```
# task.yaml
kind: Task
metadata:
  name: dummy-task
spec:
  params:
    - name: sleep-second
      type: string
      default: "3"
    - name: message
      type: string
      default: "hello, world"
  steps:
    - name: echo
      image: alpine:3.14.3
      script: |
        #!/usr/bin/env sh
        sleep "$(params.sleep-second)"
        echo "$(params.message)"
```

The original `Task` had `sleep "${sleep-second}"`, but that never resolved to anything, since that's shell script substitution syntax, not Pipeline param substitution syntax. Hence the change I made.

I did tweak the test script slightly to just run 100 `PipelineRun`s, since I was doing this on my laptop with `kind` and going the full 250 just took too long for me to be sure any results were legitimate. 

With Pipeline 0.36.0, I generally saw the first 80 or so `PipelineRun`s having the right `skippedTasks`, but then the last 20 or so all were incorrect. When I used a build of this PR, I always saw all 100 being correct.

devholic:
I rebased changes :)

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
afrittoli: Yeah, the `TaskRun` condition is set by the `TaskRun` reconciler on the first reconciliation loop. 
In the `TaskRun` controller, at least within the `ReconcileKind` function, the condition will always be set as initialising the condition is the first thing we do in `ReconcileKind`.
In the `PipelineRun` controller instead there is no guarantee. It may not be very likely to catch a `TaskRun`  with no condition set, but it is certainly possible and I guess it may happen more under high load.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -412,7 +412,7 @@ func (facts *PipelineRunFacts) GetFinalTasks() PipelineRunState {
 	if facts.checkDAGTasksDone() {
 		// return list of tasks with all final tasks
 		for _, t := range facts.State {
-			if facts.isFinalTask(t.PipelineTask.Name) && !t.isSuccessful() {
+			if facts.isFinalTask(t.PipelineTask.Name) && !t.isScheduled() {
```
afrittoli, line 415: I think this would break retries in finally tasks - the current code purposely includes failed tasks because `getNextTasks` will do the filtering again, and it needs failed tasks to be in so that `getRetryableTasks` can get tasks which failed but may be retried. The comment on L405 should be updated to reflect that.

I find the code a bit confusing here because it seems that the filter logic is shared between the `GetFinalTasks` and `getNextTasks`. I think that L415 could be:

```go
   if facts.isFinalTask(t.PipelineTask.Name) {
```

So that we get *all* final tasks and the filtering is done only by `getNextTasks`

devholic:
Thanks for the review @afrittoli 🙂 

As you pointed out, delegating task filtering logic to `getNextTasks` seems better in the point of clarity, and it will work with retrying failed tasks.

I updated code with a test case. Could you take a look?

afrittoli (commented):
Thank you for discovering and fixing this issue!
The new test coverage looks good to me, it catches the issues I highlight in the previous revision.

/approve

afrittoli:
@pritidesai @abayer if you manage to review this today it can still make the release :)

PR #4580, opened 2022-02-15
[TEP-0059] Scope `when` expressions to `Task` only

jerop:
# Changes

In [TEP-0007: Conditions Beta][tep-0007], we introduced `when` expressions to guard execution of `Tasks` in `Pipelines`. To align with `Conditions`, we set scope of `when` expressions to the guarded `Task` and its dependent `Tasks`.

In [TEP-0059: Skipping Strategies][tep-0059], we proposed changing the scope of `when` expressions to the guarded `Task` only. This was implemented in https://github.com/tektoncd/pipeline/pull/4085. We provided a feature flag, `scope-when-expressions-to-task`, to support migration. It defaulted to `false` for 9 months per our [Beta API compatibility policy][policy], meaning that we continued to guard the `Task` and its dependent `Tasks`. In this change, we flip the flag to `true` to guard the `Task` only by default.

[tep-0007]: https://github.com/tektoncd/community/blob/main/teps/0007-conditions-beta.md
[tep-0059]: https://github.com/tektoncd/community/blob/main/teps/0059-skipping-strategies.md
[policy]: https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md

/kind feature

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

```release-note
The `when` expressions in a given `Task` are scoped to guard the `Task` only by default. Set `scope-when-expressions-to-task` to continue guarding the `Task` and its dependent `Tasks`. The `scope-when-expressions-to-task` flag will be removed soon so we recommend migrating `Pipelines` to using `when` expressions scoped to the guarded `Task` only.
```

PR #4584, opened 2022-02-16
Update the `scope-when-expressions-to-task` feature flag docs

jerop:
# Changes

In https://github.com/tektoncd/pipeline/pull/4580, we changed the
flag default from "false" to "true". However, the documentation
above the flag was still describing what setting it to "true" would
do. In this change, we update the documentation to focus on the
non-default option that users can choose to set - "false". We also
add a reference to TEP-0059 and relevant docs for more details.

/kind documentation

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


afrittoli (commented):
Thanks @jerop !
Let's get this PR into the release :)
/approve
/lgtm

Issue #4567, opened 2022-02-11
TestHermeticTaskRun is flakey


jerop:
# Expected Behavior

`TestHermeticTaskRun` should only fail due to actual bugs

# Actual Behavior

`TestHermeticTaskRun` flaked in:
- https://github.com/tektoncd/pipeline/pull/4541#issuecomment-1027998805 and passed on rerun
- https://github.com/tektoncd/pipeline/pull/4564 - https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4564/pull-tekton-pipeline-alpha-integration-tests/1492134427130597378/

```
Error waiting for TaskRun not-hermetic-run-as-root to finish: "not-hermetic-run-as-root" failed
Error executing command: fork/exec /tekton/scripts/script-0-wrvhk: permission denied
```

bobcatfish:
Some more context, It looks like for https://github.com/tektoncd/pipeline/pull/4541 it failed 3 times in a row:

https://tekton-releases.appspot.com/builds/tekton-prow/pr-logs/pull/tektoncd_pipeline/4541/pull-tekton-pipeline-alpha-integration-tests/

![image](https://user-images.githubusercontent.com/432502/154344556-3ca0f4dc-c1ac-4e1d-b2d3-82a03887de89.png)

* https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4541/pull-tekton-pipeline-alpha-integration-tests/1488883003433160704/
* https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4541/pull-tekton-pipeline-alpha-integration-tests/1488840473651974144/
* https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4541/pull-tekton-pipeline-alpha-integration-tests/1488819711146725378/

I think `Error executing command: fork/exec /tekton/scripts/script-0-wrvhk: permission denied` might be a red herring - I think that might actually be from a previous run with hermetic mode on (makes me wonder if the expected failure in hermetic mode is happening for the right reason - maybe script mode doesnt work with hermetic mode? but that's another story!)

It seems like the failing taskrun is timing out:

```
          podName: not-hermetic-run-as-root-pod
          startTime: "2022-02-02T10:36:41Z"
          steps:
          - container: step-access-network
[...]

PR #4590, opened 2022-02-16
Remove lengthly operations from hermetic tests 🧪

bobcatfish:
# Changes

Just WIP for now b/c I want to try out a couple of things first :D

Also use Errorf instead of Fatalf between the two tests (the hermetic
test and the non-hermetic tests) so that if one fails the other will
still run.

In #4567 we see that the hermetic end to end test sometimes fails,
specifically it seems to be the `not-hermetic-run-as-root` version of
the test, and it seems like the failure is hitting the 1 minute timeout.

Looking at the test, it seems to be doing an `apt-get update` which
seems like an operation that would be in grave danger of sometimes
taking a while (especially depending on what version of the latest
ubuntu image is running) so although I'm not sure that's what is causing
the problem, I want to try doing something that is less likely to take
so long but still would require network access, as well as something
that would require priviledged access - which I assume is why the update
was included, to capture the combo of network access and doing something
priviledged. I'm still a bit confused about why both of those elements
are present - I assume both are not allowed in hermetic mode but it
would probably make more sense to test them separately to be sure they
each fail, otherwise only one is covered (i.e. either the network call
is going to fail and halt things, or the priviledged operation)

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


bobcatfish:
My plan is to try a couple of things to make sure the test is still actually testing what it should and then I'm thinking we could try this out and see if the flake stops happening.

bobcatfish:
Didn't get a chance to do as much investigation as I'd hoped, but I suggest we still give this a try - we can always reopen the flakey test issue and continue investigating if it persists.

Fixes #4567

dibyom (approved):

PR #4608, opened 2022-02-22
set the param type based on the default value

pritidesai:
# Changes

`(pp *ParamSpec) SetDefaults` sets the param type based on the default type. But, it is possible that the `default` value is specified without specifying the type. Also, the `else` block sets the `type` to `string` when `default` is `nil` not when the type can not be inferred from the default value (like its explained in the comment): 

https://github.com/tektoncd/pipeline/blob/99b8b196ea753af36befda8c0e0e1eaa9490ae68/pkg/apis/pipeline/v1beta1/param_types.go#L58-L64

The unit tests does not catch this since the test uses `v1beta1.NewArrayOrString` which sets the type explicitly:

https://github.com/tektoncd/pipeline/blob/99b8b196ea753af36befda8c0e0e1eaa9490ae68/pkg/apis/pipeline/v1beta1/param_types.go#L139-L150

This commit checks if the `type` is not specified, infer it from the default value and sets the param type based on the default value.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

For pull requests with a release note:

```release-note
Set the type of the param based on the default value when the type is not specified or cannot be inferred from the default value.
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/param_types.go
+++ b/pkg/apis/pipeline/v1beta1/param_types.go
@@ -57,7 +57,17 @@ func (pp *ParamSpec) SetDefaults(ctx context.Context) {
 	if pp != nil && pp.Type == "" {
 		if pp.Default != nil {
 			// propagate the parsed ArrayOrString's type to the parent ParamSpec's type
-			pp.Type = pp.Default.Type
+			if pp.Default.Type != "" {
+				// propagate the default type if specified
+				pp.Type = pp.Default.Type
+			} else {
+				// determine the type based on the array or string values when default value is specified but not the type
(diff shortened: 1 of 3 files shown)


jerop (approved):
thanks @pritidesai 😁 

/lgtm

jerop:
[workspace-in-sidecar](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4608/pull-tekton-pipeline-alpha-integration-tests/1503799795897077761/#test-testexamplesv1beta1taskrunsalphaworkspace-in-sidecar) is a known flake: https://github.com/tektoncd/pipeline/issues/4169

PR #4617, opened 2022-02-24
Add functions to merge step/sidecar overrides

lbernick:
# Changes
This commit adds methods to merge a slice of Step overrides with a slice of Steps.
This functionality is necessary for the full implementation of Step and Sidecar overrides.
It re-uses the merge process that is used to combine Steps and Step templates;
however, the Step is treated as the "template" so that the Step override takes precedence.
We do not attempt to validate the resource requirements of the resulting pod;
it's possible for the resulting pod to have resource requests > limits, which k8s will reject.
This is the same way we currently handle Step template.
Part 2 of #4326.
This is part of #4598, split out into its own PR.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes
```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/merge.go
+++ b/pkg/apis/pipeline/v1beta1/merge.go
@@ -23,6 +23,15 @@ import (
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 )
 
+// mergeData is used to store the intermediate data needed to merge an object
+// with a template. It's provided to avoid repeatedly re-serializing the template.
+// +k8s:openapi-gen=false
+type mergeData struct {
+	emptyJSON    []byte
+	templateJSON []byte
+	patchSchema  strategicpatch.PatchMetaFromStruct
+}
+
 // MergeStepsWithStepTemplate takes a possibly nil container template and a
 // list of steps, merging each of the steps with the container template, if
 // it's not nil, and returning the resulting list.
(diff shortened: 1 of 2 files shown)


pkg/apis/pipeline/v1beta1/merge.go
```diff
@@ -31,61 +31,111 @@ func MergeStepsWithStepTemplate(template *v1.Container, steps []Step) ([]Step, e
 		return steps, nil
 	}
 
-	// We need JSON bytes to generate a patch to merge the step containers
-	// onto the template container, so marshal the template.
-	templateAsJSON, err := json.Marshal(template)
```
vdemeester, line 36: Probably a bit of an optimization but can we pass the marshall object in `mergeObject` instead of the template string ? Before this merge we would marshall the template once and the use it in the loop. With this change, we will marshall the template for each steps, even though it never changes.

lbernick: Done! Note that the template doesn't change for StepTemplates, but it does change for StepOverrides, so I've provided two separate functions

vdemeester (approved):

pkg/apis/pipeline/v1beta1/merge_test.go
```diff
@@ -117,3 +117,379 @@ func TestMergeStepsWithStepTemplate(t *testing.T) {
...
 	}
 }
+
+func TestMergeStepOverrides(t *testing.T) {
+	tcs := []struct {
+		name          string
+		steps         []Step
+		stepOverrides []TaskRunStepOverride
+		want          []Step
+		wantErr       bool
```
dibyom, line 127: nit: the `wantErr` field does not seem to be used, so remove it. (For error cases, a lot of times its a bit more readable to have a separate Error tests function e.g. TestMergeStepOverridesError)

lbernick: nice catch thanks! done

pkg/apis/pipeline/v1beta1/swagger.json
```diff
@@ -2744,6 +2744,29 @@
           "default": ""
         }
       }
+    },
+    "v1beta1.mergeData": {
```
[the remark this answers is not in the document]
lbernick: you're right thanks!

sbwsg:
Awesome, cheers @lbernick ! 

/lgtm

Issue #3238, opened 2020-09-15
Cancel TaskRuns using entrypoint binary


imjasonh:
## Today

When a TaskRun is cancelled, the TaskRun controller deletes the TaskRun's underlying Pod. This stops execution ~immediately, but also leads Kubernetes to reap the Pod's logs.

## Feature Request

In #2559, we're discussing enforcing TaskRun-level timeouts in the entrypoint binary, so that timed-out TaskRun Pods don't get deleted and any logs lost. Instead of deleting the Pod, the entrypoint binary that runs each step will just stop executing and fail any running step, and not run any subsequent steps.

If we end up doing that, we could also enforce _cancellation_ in the entrypoint binary, which would let us keep Pods and logs around for cancelled TaskRuns too.

To accomplish this, the entrypoint binary could take a new flag `-cancel_file`, which is a Downard API volume populated from a Pod annotation -- this is similar to how we signal the first step to start only after all sidecars are ready. In this model, when a TaskRun is cancelled, the TaskRun controller would annotate the Pod with, for example "cancelled=true", which would update the contents of the projected file, which the entrypoint binary would see, then it can stop executing the currently running step.

This behavior change should be guarded by a feature flag (opt-in at first) since some users might depend on the current behavior. This also gives us an opportunity to compare behavior and timing of cancellation between the two implementations.

/kind feature

bobcatfish:
Makes sense to me to revisit this!

@ImJasonH are there any other options that are worth evaluating for this? For example, we could send a signal that the entrypoint binary could catch - iirc the only reason we haven't relied more on signals is b/c we haven't been able to rely on how some arbitrary process will handle it; we can control the entrypoint binary tho 🤔 

I think @sbwsg has looked into this as well in the context of the initial cancellation feature

ImJasonH:
Can we send signals to containers in the pod from the controller? I didn't think that was an option, so we've relied on file-existence checks backed by Downard volumes instead.

We've also discussed having a sidecar in the Pod that can accept RPCs from the controller, but that's a much larger change, and ultimately sort of orthogonal to how entrypoint stops the user's entrypoint execution.

vdemeester:
/remove-lifecycle stale
/lifecycle frozen

Putting this in the "frozen" box as this is something that is worth exploring 🙃

chengjoey:
Is this still in progress, I am willing to implement this

chengjoey:
Is this still a work in progress, I would love to implement it

[...]

Issue #4035, opened 2021-06-11
PipelineRun timeouts delete TaskRun related pods


rporres:
The current implementation of the PipelineRun timeout is done through deleting the TaskRun associated pods. While this is an understandable implementation as there is no easy way to stop a pod, it makes the debugging of a PipelineRun that has failed completely impossible. It also took me a while to understand why the related pods were being deleted.

In the current shape, the PipelineRun timeout may not be very helpful for people that need to access the logs of the failed TaskRuns. It may be helpful to add a note in the docs about how timeouts are being implemented so that people choose wisely if they want to use the feature.

Since pipeline timeouts are a very desirable feature, they should be implemented in a way that keeps the underlying pods available for debugging purposes.

vdemeester:
/kind feature

One possible exploration to do here would be to make the entrypoint handle the timeout somehow. That way, we could stop the execution of the pod/containers without deleteting it.

afrittoli:
Related to the behaviour of "cancel" - see https://github.com/tektoncd/pipeline/issues/3238

vdemeester:
/lifecycle frozen
Should be more or less taken care of with https://github.com/tektoncd/pipeline/pull/4618, as the timeout is cancelling the Task. Today cancel means delete, but with that proposed feature, we would not delete the pod anymore.

PR #4618, opened 2022-02-24
Cancel taskrun using entrypoint binary

adshmh:
# Changes

The cancellation of taskruns is now done through the entrypoint binary
through a new flag called 'cancel_file'. This removes the need for
deleting the pods to cancel a taskrun, allowing examination of the logs
on the pods from cancelled taskruns. Part of work on issue #3238

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Pods corresponding to a cancelled taskrun are no longer deleted: they are stopped instead.
```


pritidesai:
thanks @adshmh for this! Welcome to the community! 🎉 
/ok-to-test

vdemeester:
/assign
/hold

@adshmh thanks for this PR.  From reading it at a high level, it doesn't change the way a user ask for cancel (through update `spec.status`, but only how the taskrun reconciler handles the cancel.

I am all for that idea, but it might affect current user somehow, so I wonder if we should have this behind a feature flag that we switch on by default after a few release, wdyt ?

adshmh:
> /assign /hold
> 
> @adshmh thanks for this PR. From reading it at a high level, it doesn't change the way a user ask for cancel (through update `spec.status`, but only how the taskrun reconciler handles the cancel.
> 
> I am all for that idea, but it might affect current user somehow, so I wonder if we should have this behind a feature flag that we switch on by default after a few release, wdyt ?

Thank you for pointing this out. I will add a feature flag for the new behavior of cancel.

cmd/entrypoint/waiter.go
```diff
@@ -31,11 +32,14 @@ func (rw *realWaiter) setWaitPollingInterval(pollingInterval time.Duration) *rea
...
 // If a file of the same name with a ".err" extension exists then this Wait
 // will end with a skipError.
-func (rw *realWaiter) Wait(file string, expectContent bool, breakpointOnFailure bool) error {
+func (rw *realWaiter) Wait(ctx context.Context, file string, expectContent bool, breakpointOnFailure bool) error {
 	if file == "" {
 		return nil
 	}
 	for ; ; time.Sleep(rw.waitPollingInterval) {
+		if ctx.Err() != nil {
+			return nil
```
imjasonh, line 41:
```suggestion
			return err
```

Or probably ignore it `if !errors.Is(err, context.Canceled)`

cmd/entrypoint/waiter_test.go
```diff
@@ -38,7 +39,7 @@ func TestRealWaiterWaitMissingFile(t *testing.T) {
 	rw := realWaiter{}
 	doneCh := make(chan struct{})
 	go func() {
-		err := rw.setWaitPollingInterval(testWaitPollingInterval).Wait(tmp.Name(), false, false)
+		err := rw.setWaitPollingInterval(testWaitPollingInterval).Wait(context.Background(), tmp.Name(), false, false)
```
imjasonh, line 42: For readability, maybe `ctx := context.Background()` up at the top of this test, and reuse it throughout.

pkg/entrypoint/entrypointer.go
```diff
@@ -114,7 +117,7 @@ func (e Entrypointer) Go() error {
 	}()
 
 	for _, f := range e.WaitFiles {
-		if err := e.Waiter.Wait(f, e.WaitFileContent, e.BreakpointOnFailure); err != nil {
+		if err := e.Waiter.Wait(context.Background(), f, e.WaitFileContent, e.BreakpointOnFailure); err != nil {
```
imjasonh, line 120: Could we have `Go` take a `context.Context` and use that here? If it's still a `context.Background()` inside `cmd/entrypoint/main.go` that's fine, it's just a bit cleaner and makes it easier to add other stuff like cancelling when we get SIGINT, which we might want to do later.

pkg/entrypoint/entrypointer_test.go
```diff
@@ -160,6 +165,18 @@ func TestEntrypointer(t *testing.T) {
 	}, {
 		desc:                "breakpointOnFailure to wait or not to wait ",
 		breakpointOnFailure: true,
+	}, {
+		desc:       "Runner completes if not cancelled",
+		cancelFile: ".",
```
imjasonh, line 170: Can we make this `cancelfile` or something? `"."` is an odd name for this file.

imjasonh (commented):
Wow, what a great change! 🎉 

Thanks for your contribution, +1 to Vincent's point about a feature flag, but otherwise, this looks great.

adshmh:
> /assign /hold
> 
> @adshmh thanks for this PR. From reading it at a high level, it doesn't change the way a user ask for cancel (through update `spec.status`), but only how the taskrun reconciler handles the cancel.
> 
> I am all for that idea, but it might affect current user somehow, so I wonder if we should have this behind a feature flag that we switch on by default after a few release, wdyt ?
> 
> /cc @tektoncd/core-maintainers

Thank you for the review. The new feature of cancelling using the entrypoint binary is now behind a feature flag, with default set to false.

cmd/entrypoint/waiter.go
adshmh: Thank you for the review. Fixed.

cmd/entrypoint/waiter_test.go
adshmh: Thank you for the review. Fixed.

pkg/entrypoint/entrypointer.go
adshmh: Thank you for the review. Fixed.

pkg/entrypoint/entrypointer_test.go
adshmh: Thank you for the review. Fixed.

cmd/entrypoint/waiter.go
```diff
@@ -31,11 +32,14 @@ func (rw *realWaiter) setWaitPollingInterval(pollingInterval time.Duration) *rea
 //
 // If a file of the same name with a ".err" extension exists then this Wait
 // will end with a skipError.
-func (rw *realWaiter) Wait(file string, expectContent bool, breakpointOnFailure bool) error {
+func (rw *realWaiter) Wait(ctx context.Context, file string, expectContent bool, breakpointOnFailure bool) error {
 	if file == "" {
 		return nil
 	}
 	for ; ; time.Sleep(rw.waitPollingInterval) {
+		if err := ctx.Err(); err != nil {
```
vdemeester, line 40: Do we also want to handle `ctx.Done()` ? 🤔 Not sure if that case would happen but…

pkg/entrypoint/entrypointer.go
```diff
@@ -142,19 +145,46 @@ func (e Entrypointer) Go() error {
...
-			defer cancel()
+		} else {
+			ctx, cancel = context.WithCancel(ctx)
+		}
+		defer cancel()
+
+		errChan := make(chan error, 1)
+		go func() {
+			errChan <- e.Runner.Run(ctx, e.Command...)
+			cancel()
```
vdemeester, line 159: Does it mean we'll run `cancel` twice ? (here and the `defer`)
I don't remember if it's a no-op or if it panics.. If it's a no-op we are fine though.

vdemeester (commented):
2 small questions / comments, otherwise, looks good 👍🏼

cmd/entrypoint/waiter.go
adshmh: Thank you for the review. Fixed.
Looking at the code again, I think using your suggestion, i.e. `ctx.Done()` is more readable. Updated.

On a side note, I increased the polling interval on the unit tests from 10 to 25 milliseconds, as it seems to reduce flakes significantly (from about 5% to around 0,5% in my tests). If this needs to be reverted, please let me know.

`const testWaitPollingInterval = 25 * time.Millisecond`

pkg/entrypoint/entrypointer.go
adshmh: Thank you for the review. Yes, we will end up calling `cancel` twice on the context, but that is no-op ([I tested to be sure](https://go.dev/play/p/PKt2M6X7inR)).
I added the `defer cancel()` for both readability and ensuring clean-up. If removing it is preferred, please let me know.

osherdp:
@adshmh @vdemeester any idea what's left for this change to get in?
If it will have positive effects on #4035 then it will solve lots of problems for us :smiley:

lbernick (commented):
Thanks for this @adshmh! When a pipelineRun is cancelled, it will also cancel its child TaskRuns. I'm not sure if any of the PipelineRun cancellation logic needs to be changed, but could you add some tests to the PipelineRun reconciler for this new cancellation strategy?

In addition, could you please add some docs on the new feature flag, and update the release note to specify that this behavior is controlled by the feature flag?

vdemeester:
> Thanks for this @adshmh! When a pipelineRun is cancelled, it will also cancel its child TaskRuns. I'm not sure if any of the PipelineRun cancellation logic needs to be changed, but could you add some tests to the PipelineRun reconciler for this new cancellation strategy?

From the `PipelineRun` perspective, nothing changes 👼🏼

pkg/entrypoint/entrypointer.go
```diff
@@ -142,19 +145,46 @@ func (e Entrypointer) Go() error {
...
-		err = e.Runner.Run(ctx, e.Command...)
+		err = <-errChan
+
 		if err == context.DeadlineExceeded {
 			output = append(output, v1beta1.PipelineResourceResult{
 				Key:        "Reason",
 				Value:      "TimeoutExceeded",
 				ResultType: v1beta1.InternalTektonResultType,
 			})
+		} else if cancelled {
```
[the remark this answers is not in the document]
vdemeester: @lbernick I think the name is a bit infortunate as it is this struct that we use for returning `Result` I think 🙃 
We could rename it at some point.

afrittoli:
This is unfortunately not ready to merge, so I'll move it to the next milestone

adshmh:
> /assign I'll try to carry this patch later this week 👼🏼

Sorry for the delay on this. I will follow up by addressing the remaining concerns in a few days.

pkg/entrypoint/entrypointer.go
```diff
@@ -142,19 +145,46 @@ func (e Entrypointer) Go() error {
...
+
+		errChan := make(chan error, 1)
+		go func() {
+			errChan <- e.Runner.Run(ctx, e.Command...)
+			cancel()
+		}()
+
+		var cancelled bool
+		if e.CancelFile != "" {
+			if err := e.Waiter.Wait(ctx, e.CancelFile, true, e.BreakpointOnFailure); err != nil {
```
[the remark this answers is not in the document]
adshmh: Thank you for the review. Indeed the logic here could benefit from some simplification. I will do another review to see if it can be simplified without making too many modifications elsewhere.

vdemeester:
Carrying on https://github.com/tektoncd/pipeline/pull/5401

jerop:
doing a clean up of stale pull requests - feel free to reopen if you pick up this work again

newer pull request in https://github.com/tektoncd/pipeline/pull/5401

/close

PR #4327, opened 2021-10-21
TEP-69, Implementation. Support retry for custom tasks.

ScrapCodes:
# Changes
1. Add field Retries to RunSpec.
2. Add field RetriesStatus to RunStatusFields
3. Now pipeline task support retries for custom task by copying it to `Retries` of `RunsSpec`.
4. Removed validation that forbade user from configuring retries for a custom task inside a `pipelineSpec`.

5. Tests included.
6. Documentation,  added a developer guide for custom task developer.

/kind tep

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

```release-note
A custom task specification can be created with `Retries` as follows:

            
             apiVersion: tekton.dev/v1alpha1
             kind: Run
             metadata:
               generateName: simpleexample
             spec:
               retries: 3 # set retries
               params:
                 - name: searching
                   value: the purpose of my existence
               ref:
                 apiVersion: custom.tekton.dev/v1alpha1
                 kind: Example
                 name: exampleName


```

PR #4625, opened 2022-02-25
Update IsFailure method to reflect Run retries

lbernick:
# Changes
Custom Tasks support retries (as of #4327); however, there are some places in the code base that consider
retries only for TaskRuns. This commit updates ResolvedPipelineRunTask.IsFailure to recognize
Run retries and adds tests for this method. The function PipelineRunState.getNextTasks will
be updated in a future commit. No functional changes for TaskRuns.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
[Bug Fix]: Runs that have remaining retries are not failed
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -136,16 +136,46 @@ func (t ResolvedPipelineRunTask) IsSuccessful() bool {
 
 // IsFailure returns true only if the run has failed and will not be retried.
 func (t ResolvedPipelineRunTask) IsFailure() bool {
-	if t.IsCustomTask() {
-		return t.Run != nil && t.Run.IsDone() && !t.Run.IsSuccessful()
+	if t.IsCancelled() {
+		return true
 	}
-	if t.TaskRun == nil {
+	if t.IsSuccessful() {
 		return false
 	}
-	c := t.TaskRun.Status.GetCondition(apis.ConditionSucceeded)
-	retriesDone := len(t.TaskRun.Status.RetriesStatus)
-	retries := t.PipelineTask.Retries
-	return c.IsFalse() && (retriesDone >= retries || c.Reason == v1beta1.TaskRunReasonCancelled.String())
+	var c *apis.Condition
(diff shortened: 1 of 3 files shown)


pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -136,16 +136,46 @@ func (t ResolvedPipelineRunTask) IsSuccessful() bool {
 
 // IsFailure returns true only if the run has failed and will not be retried.
 func (t ResolvedPipelineRunTask) IsFailure() bool {
-	if t.IsCustomTask() {
-		return t.Run != nil && t.Run.IsDone() && !t.Run.IsSuccessful()
+	if t.IsCancelled() {
```
lbernick, line 139: one weird bit of maybe undefined behavior is that for TaskRuns, if it is marked as cancelled but has status "unknown" (rather than "false"), this is considered not failed and not done. I decided to use the same behavior for Runs as for TaskRuns. If we want to change this behavior I think it belongs in a separate commit.

vdemeester (approved):

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
pritidesai: A taskRun is considered as Cancelled if `status` is `false` and `reason` is set to `TaskRunCancelled`. Are you saying the `status` is set to `unknown` when a taskRun is cancelled on demand? I didnt quite follow how thats happening.

pritidesai:
thank you @lbernick 👍 
/lgtm

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
lbernick: What I mean is, if a TaskRun has status "unknown" with reason "cancelled", it will be marked as not failed. It's unclear to me whether such a condition can occur and what the reconciler's behavior should be in this scenario. Our tests implicitly depended on this behavior prior to this commit, and now make this behavior explicit.

PR #4630, opened 2022-02-28
Change Default Metrics Level for Taskrun and Pipelinerun

khrm:
Change default metrics level for Taskrun to task and
Pipelinerun to pipeline. Part of TEP0073

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Metrics Default Changes:
The Level for taskrun and pipelinerun metrics will change in this release.
By default, taskrun and pipelinerun metrics have these values:

``` yaml
    metrics.taskrun.level: "task"
    metrics.taskrun.duration-type: "histogram"
    metrics.pipelinerun.level: "pipeline"
    metrics.pipelinerun.duration-type: "histogram"
```

### DIFF
--- a/pkg/apis/config/metrics.go
+++ b/pkg/apis/config/metrics.go
@@ -38,9 +38,7 @@ const (
 
 	// DefaultTaskrunLevel determines to what level to aggregate metrics
 	// when it isn't specified in configmap
-	// TBD: Change to task in next
-	// relase and taskrun level will be deprecated
-	DefaultTaskrunLevel = "taskrun"
+	DefaultTaskrunLevel = TaskrunLevelAtTask
 	// TaskrunLevelAtTaskrun specify that aggregation will be done at
 	// taskrun level
 	TaskrunLevelAtTaskrun = "taskrun"
@@ -50,9 +48,7 @@ const (
 	TaskrunLevelAtNS = "namespace"
 	// DefaultPipelinerunLevel determines to what level to aggregate metrics
 	// when it isn't specified in configmap
-	// TBD: Change to pipeline in next
-	// relase and pipelinerun level will be deprecated
-	DefaultPipelinerunLevel = "pipelinerun"
+	DefaultPipelinerunLevel = PipelinerunLevelAtPipeline
 	// PipelinerunLevelAtPipelinerun specify that aggregation will be done at
 	// pipelienrun level
 	PipelinerunLevelAtPipelinerun = "pipelinerun"
--- a/config/config-observability.yaml
+++ b/config/config-observability.yaml
@@ -54,7 +54,7 @@ data:
     # charge.  If metrics.backend-destination is not Stackdriver, this is
     # ignored.
     metrics.allow-stackdriver-custom-metrics: "false"
-    metrics.taskrun.level: "taskrun"
+    metrics.taskrun.level: "task"
     metrics.taskrun.duration-type: "histogram"
-    metrics.pipelinerun.level: "pipelinerun"
+    metrics.pipelinerun.level: "pipeline"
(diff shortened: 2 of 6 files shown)


vdemeester (approved):

pkg/apis/config/metrics.go
```diff
@@ -40,7 +40,7 @@ const (
 	// when it isn't specified in configmap
 	// TBD: Change to task in next
 	// relase and taskrun level will be deprecated
```
jerop, line 42: this comment needs an update

pkg/apis/config/metrics.go
```diff
@@ -52,7 +52,7 @@ const (
 	// when it isn't specified in configmap
 	// TBD: Change to pipeline in next
 	// relase and pipelinerun level will be deprecated
```
jerop, line 54: this comment also needs an update

jerop (commented):
metrics is not part of https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md today - i think we should add it in line with k8s conventions: https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-metric

we've briefly discussed adding metrics to the API spec in V1 plans: https://github.com/tektoncd/community/blob/main/teps/0096-pipelines-v1-api.md

pkg/apis/config/metrics.go
```diff
@@ -40,7 +40,7 @@ const (
 	// when it isn't specified in configmap
 	// TBD: Change to task in next
 	// relase and taskrun level will be deprecated
-	DefaultTaskrunLevel = "taskrun"
+	DefaultTaskrunLevel = "task"
```
pritidesai, line 43: NIT: please drop the TBD from here - lines 41 and 42

pkg/apis/config/metrics.go
```diff
@@ -52,7 +52,7 @@ const (
 	// when it isn't specified in configmap
 	// TBD: Change to pipeline in next
 	// relase and pipelinerun level will be deprecated
-	DefaultPipelinerunLevel = "pipelinerun"
+	DefaultPipelinerunLevel = "pipeline"
```
pritidesai, line 55: NIT: please drop the TBD from here - lines 53 and 54

pkg/apis/config/metrics.go
khrm: Done. Thanks

pkg/apis/config/metrics.go
khrm: Resolved. :)

pritidesai:
> metrics is not part of https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md today - i think we should add it in line with k8s conventions: https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-metric
> 
> we've briefly discussed adding metrics to the API spec in V1 plans: https://github.com/tektoncd/community/blob/main/teps/0096-pipelines-v1-api.md
> 
> cc @tektoncd/core-maintainers

@jerop yup we should definitely include it as part of the compatibility policy. Do you think we can add it after we changed defaults right now?

jerop:
> @jerop yup we should definitely include it as part of the compatibility policy. Do you think we can add it after we changed defaults right now?

yes, we can change the defaults now and may also want to make other fixes to metrics before adding it to the policy

also as discussed in API WG, we should highlight this change in the upgrade notice in the next release

khrm:
@jerop I already added the notice in the release note. Is that sufficient for this pr?

jerop:
> @jerop I already added the notice in the release note. Is that sufficient for this pr?

yes, that's sufficient - thanks @khrm!

/lgtm

Issue #3631, opened 2020-12-11
Make it easier to setup SSH authentication for Git


jlpettersson:
### Feature request

The [authentication page in documentation](https://github.com/tektoncd/pipeline/blob/master/docs/auth.md) looks a bit overwhelming at first visit. Unfortunately, setting up authentication is one of the first things a new user needs to do to use Tekton for Continuous Integration. It would be good if the steps (and instructions) on how to setup SSH authentication for Git would be easier to understand and easier to do. If this part is too overwhelming, it might serve as a barrier to start using Tekton.

#### Suggestions

**Prepare a known_hosts file**

Example using github.com

1. Create file with `known_hosts` (you may also want to verify this further)

   ```
   ssh-keyscan github.com > ssh_known_hosts
   ```

2. Create secret from file

    ```
    kubectl create configmap github-known-hosts --from-file=ssh_known_hosts
    ```

**Generate and distribute SSH key pair**

Generate a separate SSH key pair for Tekton. _(The current auth doc describes a process for uploading the users private key from its home dir which might not be a good idea from a security standpoint)_

1. Generate keypair to local file

    ```
    ssh-keygen -t rsa -b 4096 -f id_rsa -q -N ""
    ```

2. Create a secret from the private key

   ```
   kubectl create secret generic github-private-key --from-file=id_rsa
   ```

3. Upload the public key `id_rsa.pub` to GitHub

   Start with copying the content of the public key with
   (this example is for macos)

   ```
   pbcopy < id_rsa.pub
   ```

   And follow [Adding a new SSH key to your GitHub account](https://help.github.com/en/github/authenticating-to-github/adding-a-new-ssh-key-to-your-github-account) alternatively [Managing Deploy Keys](https://docs.github.com/en/free-pro-team@latest/developers/overview/managing-deploy-keys#deploy-keys) for organization

**Usage**

The `ConfigMap` and `Secret` created above, can be exposed to a Task with a Projected Volume:

```
      volumes:
      - name: ssh-auth                  # name of volume - matching name in Task
        projected:
          defaultMode: 0400
          sources:
          - configMap:
              name: github-known-hosts  # name of ConfigMap from Auth setup
          - secret:
              name: github-private-key  # name of Secret from Auth setup
```

#### Summary

The above commands for secret, configmap and ssh key generation can be used by copy-pasting, which makes the process a bit easier do to, and to avoid mistakes.

### Use case

[...]

PR #4631, opened 2022-02-28
Changes to setting up SSH authentication for Git

OctoElla:
# Changes
Modified the "Configuring `ssh-auth` authentication for Git" section of the "Authentication at Run Time" doc. 
According to the issue: Make it easier to set up SSH authentication for Git #3631

/kind documentation

 # Submitter Checklist
 
 * [x]  [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
 * [ ]  [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
 * [x]  Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
 * [x]  Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
   functionality, content, code)
 * [x]  Release notes block below has been filled in or deleted (only if no user facing changes)
# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -235,88 +235,41 @@ Note: Github deprecated basic authentication with username and password. You can
 
 ### Configuring `ssh-auth` authentication for Git
 
-This section describes how to configure an `ssh-auth` type `Secret` for use with Git. In the example below,
-before executing any `Steps` in the `Run`, Tekton creates a `~/.ssh/config` file containing the SSH key
-specified in the `Secret`. When the `Steps` execute, Tekton uses this key to retrieve `PipelineResources`
-specified in the `Run`.
+This section describes how to set up ConfigMap and Secret for use with Git to be able to configuring a git-clone Task to use Tekton Pipelines along with a Tekton trigger connected to a Git repository.
 
-> :warning: **`PipelineResources` are [deprecated](deprecations.md#deprecation-table).**
->
-> Consider using replacement features instead. Read more in [documentation](migrating-v1alpha1-to-v1beta1.md#replacing-pipelineresources-with-tasks)
-> and [TEP-0074](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md).
 
-1. In `secret.yaml`, define a `Secret` that specifies your SSH private key:
+1. Create _known_hosts_ file called _ssh_known_hosts_ with public host keys (example using github.com):
+	```
+	ssh-keyscan github.com > ssh_known_hosts
+	```
+2. Create a configmap using _ssh_known_hosts_ file:
+	```
+	kubectl create configmap github-known-hosts --from-file=ssh_known_hosts
+	```
+3. Generate a public/private rsa key pair to local files:
+	```
+	ssh-keygen -t rsa -b 4096 -f id_rsa -q -N ""
+	```
+4. Create a secret from the file with the private key:
+	```
+	kubectl create secret generic github-private-key --from-file=id_rsa
+	```
+5. Upload the public key id_rsa.pub to GitHub:
 
-   ```yaml
-   apiVersion: v1
-   kind: Secret
-   metadata:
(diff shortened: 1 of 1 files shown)


pritidesai:
thank you @OctoElla for this! Welcome to the community 🎉 
/ok-to-test

docs/auth.md
```diff
@@ -235,88 +235,41 @@ Note: Github deprecated basic authentication with username and password. You can
...
-   ```yaml
-   apiVersion: v1
-   kind: ServiceAccount
-   metadata:
-     name: build-bot
-   secrets:
-     - name: ssh-key
-   ```
-
-1. In `run.yaml`, associate the `ServiceAccount` with your `Run` by doing one of the following:
```
pritidesai, line 286: thank you @OctoElla for updating the instructions!

Can we please keep this part of the doc with the new instructions?

OctoElla:
@sbwsg 
I’ve tested new steps from the instruction.
this step is not working:
```
4. Create a secret from the file with the private key:
        kubectl create secret generic github-private-key --from-file=id_rsa
```
If I use this command I get a secret with the type:
```
type: Opaque
```
But drawing a conclusion from my experiment, we need the secret with the type:
```
type: kubernetes.io/ssh-auth
```
and an annotation:
```
  annotations:
    tekton.dev/git-0: github.com 
```
Only in this case, Tekton creates a ~/.ssh/config file containing the SSH key specified in the Secret. And also I have to specify Service Account for TaskRun.

sbwsg:
Thanks a lot for taking the time to update the doc @OctoElla !

From my pov I don't think we can delete the existing auth content until the ServiceAccount-based approach is removed from Tekton, which is not something we have planned for yet I don't think. I will come back to this in a sec but first I also wanted to address the use of projected volumes :-

One issue with using `volumes` in Tasks and TaskRuns generally is that it requires the user to hard-code some volume configuration in their Task. A `volumeMount` must be added to a Step so that it receives a `volume`. For example, a user would need to take [the `git-clone` Task](https://github.com/tektoncd/catalog/blob/main/task/git-clone/0.5/git-clone.yaml) from the catalog and edit the `clone` Step to include a `volumeMount` with the name of the projected volume they define in their TaskRuns.

A different approach is for a Task to expose a Workspace for credentials. We see this with the `git-clone` Task, which defines an `ssh-directory` Workspace. The user is able to provide a Secret to this workspace in their TaskRun:

```yaml
kind: TaskRun
# ... skip some fields
spec:
  taskRef:
    name: git-clone
  workspaces:
  - name: ssh-directory
    secret:
      secretName: my-ssh-secret
```

This secret is structured just like an ssh dir, so can include a `id_rsa` entry and a `known_hosts` entry, but it doesn't need any of the Tekton-specific annotations or the specific type information.  It can just be a plain Opaque secret, the Workspace doesn't care.

The advantage is that the Task does not need to hard-code any `volumeMount` configuration. It just asks for a general workspace. Then the TaskRun fills in the details of exactly what kind of volume should be used for that workspace.

One down side is that Workspaces do not support projected volumes - this is definitely something we could add but it's not supported right now.

Coming back to the existing `auth.md` doc, I'd like to suggest a slightly different approach to take. Right now this doc only describes the ServiceAccount-based approach (also called `creds-init` in our codebase). What I'd like to suggest is that we add a new docs directory, `docs/auth/`, and then move the existing `auth.md` to `docs/auth/creds-init.md`. Then we could also add a `docs/auth/workspace-credentials.md` to describe the workspace-based approach and a `docs/auth/README.md` which can give a short explanation and links to the other docs. We can also add a `docs/auth/projected-volumes.md` describing that approach too (and maybe describing the pitfall that it requires editing the Task Steps?).

afrittoli:
@OctoElla thank you for your contribution? Are you still pursuing this?

PR #4641, opened 2022-03-03
Build tekton pipelines with golang 1.17

afrittoli:
# Changes

Update the tekton pipelines docker files and build pipeline to build
using golang 1.17.7:

- The windows base images are now built using golang:1.17.7 official
  docker image
- The non-windows bases images are based on distroless and need to
  update. We now reference the docker-buildx base image by sha instead
  of tag which is better for attestation
- Controller images, entrypoint and pipeline resource images are built
  using ko. Updating the ko image in the publish task to one that
  includes golang 1.17.7 (golang:1.17.7-alpine3.15). Reference the ko
  image by sha instead of tag. This will require an extra step to
  update the image in future (plumbing first, pipeline then) but it's
  best for attestations and also it allows updating the image on
  plumbing side and in the registry without direct side effect on
  pipeline.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Tekton Pipelines is built using go 1.17.7
```

### DIFF
--- a/images/windows/entrypoint/Dockerfile
+++ b/images/windows/entrypoint/Dockerfile
@@ -1,6 +1,6 @@
 ARG BASE_IMAGE=mcr.microsoft.com/windows/nanoserver:1809
 
-FROM golang:1.16.13 AS builder
+FROM golang:1.17.7 AS builder
 
 COPY . c:/gopath/src/github.com/tektoncd/pipeline
 
--- a/images/windows/nop/Dockerfile
+++ b/images/windows/nop/Dockerfile
@@ -1,6 +1,6 @@
 ARG BASE_IMAGE=mcr.microsoft.com/windows/nanoserver:1809
 
-FROM golang:1.16.13 AS builder
+FROM golang:1.17.7 AS builder
 
 COPY . c:/gopath/src/github.com/tektoncd/pipeline
 
--- a/tekton/build-push-ma-base-image.yaml
+++ b/tekton/build-push-ma-base-image.yaml
@@ -23,7 +23,7 @@ spec:
       description: The secret that contains a service account authorized to push to the imageRegistry and to the output bucket
   steps:
 
-  - image: gcr.io/tekton-releases/dogfooding/buildx-gcloud:latest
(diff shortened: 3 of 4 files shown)


images/windows/entrypoint/Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG BASE_IMAGE=mcr.microsoft.com/windows/nanoserver:1809
```
imjasonh, line 1: Do we still need these? We should be building entrypoint and nop using ko now, right?

afrittoli: Those are the base images that are fed into the `ko` config.
I believe they are used today in the build pipeline - see https://github.com/tektoncd/pipeline/blob/main/tekton/build-push-ma-base-image.yaml

afrittoli: Uhm, it looks like it's not used here though https://github.com/tektoncd/pipeline/blob/4f8c227794abd6b2157856892415d5c207352295/tekton/publish.yaml#L96-L100

afrittoli: But we still run it as part of the release pipeline: https://github.com/tektoncd/pipeline/blob/4f8c227794abd6b2157856892415d5c207352295/tekton/release-pipeline.yaml#L107-L127

afrittoli: Ok, so summarise, this is the `ko` config for `nop` and `entrypoint`:

https://github.com/tektoncd/pipeline/blob/4f8c227794abd6b2157856892415d5c207352295/tekton/publish.yaml#L107-L108

Where the `$COMBINED_BASE_IMAGE` is the result of:

https://github.com/tektoncd/pipeline/blob/4f8c227794abd6b2157856892415d5c207352295/tekton/publish.yaml#L97-L100

so it doesn't depend on anything else indeed.
The base images for `git-init` and `pullrequest-init`:

https://github.com/tektoncd/pipeline/blob/4f8c227794abd6b2157856892415d5c207352295/tekton/publish.yaml#L111-L113

Are based on the respective base images that are created out of the dockerfiles in `images/git-init` and `images/pullrequest-init`.

So, yeah, we don't need those dockerfiles anymore, and if we can use the latest release pipeline to build earlier releases, we should be able to avoid backport 🎉

PR #4651, opened 2022-03-05
Allow tasks to retry when PipelineRun stops

### LINKED ISSUE #4409: Finally task not executed when tasks define "retries"
# Expected Behavior
Final tasks are always executed

# Actual Behavior
In a pipeline with two parallel tasks where both fail and one of them has "retries=2", final tasks are never executed. Pipeline fails later because of timeout.

1. Create Task: 
```

kind: Task
metadata:
  name: noop
spec:
  params:
  - name: RETURN_CODE
    default: "0"
  steps:
    - name: sleep
      image: registry.access.redhat.com/ubi8/ubi-minimal
      script: |
        #!/usr/bin/env bash
        set -e

        echo "Will sleep for 2 seconds"
        sleep 2

        if [[ $(params.RETURN_CODE) != "0" ]]; then
          echo "Failure"
        else
          echo "Success"
        fi

        exit $(params.RETURN_CODE)
```

2. Create Pipeline

```
kind: Pipeline
metadata:
  name: test-finally
spec:
  tasks:
  - name: test1
    taskRef:
      name: noop
    params:
      - name: RETURN_CODE
        value: "1"
    retries: 2
  - name: test2
    taskRef:
      name: noop
    params:
      - name: RETURN_CODE
        value: "1"
    # retries: 2
  finally:
    - name: cleanup
      taskRef:
        name: noop
```

3. Create pipelinerun

```
kind: PipelineRun
metadata:
  generateName: test-finally-
spec:
  pipelineRef:
    name: test-finally
  timeout: 2m

```

# Additional Info
This pipeline works as expected when any of following changes

1. Task test2 succeeds
2. Both task1 and task2 have retries=2 (or any number)
3. Task test1 has retries=1, retries=0 or retries is omitted

lbernick:
# Changes
Prior to this commit, when a PipelineRun was in a "stopping" state (i.e. a TaskRun failed),
other TaskRuns would not complete their retries. However, TaskRuns with retries would also
not be considered "finished", so no finally Tasks would be scheduled, and the PipelineRun would hang.

This commit updates the PipelineRun reconciler to allow TaskRuns to exhaust their retries even if other
TaskRuns have failed, and documents this behavior.

Closes #4409 (tested locally).

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
[Bug fix]: Allow TaskRuns/Runs to complete retries when PipelineRun is stopped
```


pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -232,31 +232,43 @@ func (state PipelineRunState) getNextTasks(candidateTasks sets.String) []*Resolv
...
 
-				} else {
-					status = t.Run.Status.GetCondition(apis.ConditionSucceeded)
-					isCancelled = t.Run.IsCancelled()
-					if status != nil {
-						isCancelled = isCancelled || status.Reason == v1alpha1.RunReasonCancelled
-					}
+// getRetryableTasks returns a list of tasks which should be executed next when the pipelinerun is stopping, i.e.
+// a list of cancelled/failed tasks from candidateTasks which haven't exhausted their retries
+func (state PipelineRunState) getRetryableTasks(candidateTasks sets.String) []*ResolvedPipelineRunTask {
```
jerop, line 244: nice! thanks for splitting up the function :)

docs/pipelines.md
```diff
@@ -489,7 +489,11 @@ they are referenced in the `Pipeline` definition.
 For each `Task` in the `Pipeline`, you can specify the number of times Tekton
 should retry its execution when it fails. When a `Task` fails, the corresponding
 `TaskRun` sets its `Succeeded` `Condition` to `False`. The `retries` field
-instructs Tekton to retry executing the `Task` when this happens.
+instructs Tekton to retry executing the `Task` when this happens. `retries` are executed
+even when other `Task`s in the `Pipeline` have failed, unless the `PipelineRun` has
+been [cancelled](./pipelineruns.md#cancelling-a-pipelinerun),
+[gracefully cancelled](./pipelineruns.md#gracefully-cancelling-a-pipelinerun), or
+[gracefully stopped](./pipelineruns.md#gracefully-stopping-a-pipelinerun).
```
jerop, line 496: didn't realize we're explicit about not exhausting retries in cancelation and graceful termination cases, but I see we've documented it in the [docs](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#cancelling-a-pipelinerun), now this makes me want for us to be consistent by not exhausting retries across the board instead of having exceptions 🤔 

(apologies for revisiting the discussion about what to do, no one felt strongly in the API WG meeting so hopefully could be an easy switch 🤞🏾)

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -302,29 +314,35 @@ func (facts *PipelineRunFacts) IsGracefullyStopped() bool {
...
-		candidateTasks, err := dag.GetSchedulable(facts.TasksGraph, facts.successfulOrSkippedDAGTasks()...)
-		if err != nil {
-			return tasks, err
-		}
+	var tasks PipelineRunState
+	// when pipelinerun is gracefully cancelled or gracefully stopped, do not schedule any new tasks,
+	// and only wait for all running tasks to complete (without exhausting retries).
+	if facts.IsGracefullyCancelled() || facts.IsGracefullyStopped() {
+		return tasks, nil
+	}
```
jerop, line 322: we've documented that retries are not executed in (non-graceful) cancellation as well, why is that not included in the conditions here? 🤔

jerop (commented):
thanks @lbernick! 

given that we are explicit about not exhausting retries in (non-graceful) cancellation, graceful cancellation and graceful stopping ([docs](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#cancelling-a-pipelinerun)), I wonder if we should be consistent and not exhaust retries in all termination cases including (non-graceful) stopping?

since there were no strong opinions on the options, hopefully it'll be an easy switch to not exhausting retries in all termination cases? (cc @tektoncd/core-maintainers)

pritidesai:
@lbernick @jerop 

Each state has its own set of rules and does not have to align:

Difference between stop and terminate:

**When used as verbs, stop means to cease moving, whereas terminate means to end, especially in an incomplete state.** - quote from https://diffsense.com/diff/stop/terminate 😄

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
lbernick: I think it was never included because it's handled elsewhere-- however it makes sense for this to not return any tasks if the pipeline is cancelled so I've added that in (plus a test case)

lbernick:
/hold

@jerop I am experimenting with not scheduling retries when the pipelinerun is stopping. I think it will be easier if cleanup related changes that exist in both branches go in separately, so I've opened #4667. Once that's merged I'll rebase both branches onto that so we can compare implementations.

jerop:
> @lbernick @jerop
> 
> Each state has its own set of rules and does not have to align:
> 
> Difference between stop and terminate:
> 
> **When used as verbs, stop means to cease moving, whereas terminate means to end, especially in an incomplete state.** - quote from https://diffsense.com/diff/stop/terminate 😄

Thanks for that clarification @pritidesai! in tekton pipelines, we only have "cancel" and "stop" - graceful termination was the title of the TEP but we never used "terminate" as a config option. We have "cancel" meaning "interrupt any executing tasks" and "stop" meaning "complete executing tasks".  We also have "stopping" which is an internal state when a `task` fails, can we rename this internal state "terminating" to distinguish it, and also to indicate "ending in incomplete state"?  what do you think @pritidesai @lbernick?

> @jerop am experimenting with not scheduling retries when the pipelinerun is stopping. I think it will be easier if cleanup related changes that exist in both branches go in separately, so I've opened https://github.com/tektoncd/pipeline/pull/4667. Once that's merged I'll rebase both branches onto that so we can compare implementations.

Thank you for creating a separate PR for the refactor! If it's easier, we can decide on the way forward so you only have to implement one option 😁

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
jerop: sounds good, thanks :)

lbernick:
> Thanks for that clarification @pritidesai! in tekton pipelines, we only have "cancel" and "stop" - graceful termination was the title of the TEP but we never used "terminate" as a config option. We have "cancel" meaning "interrupt any executing tasks" and "stop" meaning "complete executing tasks".

(FWIW I really don't have a strong opinion on this) I wouldn't describe "stopping" as "complete executing tasks" because we wouldn't consider a task having completed execution without exhausting retries if all other tasks succeeded. However since we don't execute retries when the pipelinerun is "gracefully cancelled" or "gracefully stopped", I would take this to mean that both "cancelling" and "stopping" don't require exhausting retries.

> We also have "stopping" which is an internal state when a `task` fails, can we rename this internal state "terminating" to distinguish it, and also to indicate "ending in incomplete state"?
I'm fine with renaming the internal state "stopping" to "terminating" for these reasons but I'd prefer if it didn't block the bug fix.

pritidesai:
Thank you @lbernick and @jerop, I was looking at the doc and thought it will be helpful to put it in the table form: 

| Command         | Workflow     | Retries |
|--------------|-----------|------------|
| `PipelineRunCancelled` |  Running `taskRuns` are cancelled, `pods` are deleted, `retries` not executed.  |  No retires exhausted.  |
| `CancelledRunFinally` |  Running `taskRuns` are cancelled, `pods` are deleted, `retries` not executed. Finally tasks are scheduled.  |    No retries exhausted.   |
| `StoppedRunFinally` |  Running `taskRuns` are completed normally, but no new non-finally task is scheduled. Finally tasks are scheduled. | **Do we know if retries are exhausted in this case?** |

In case of `StoppedRunFinally`, it says running `taskRuns` are completed normally, does it mean that the retries are executed if a task fails but has not exhausted retries?

There has been a lot of [discussion](https://github.com/tektoncd/community/pull/397#issuecomment-829400746) done before deciding the names for these statuses 😭 

The internal `stopping` mode was [introduced](https://github.com/tektoncd/pipeline/pull/2774) such that:

```
When the pipeline run is in stopping mode, tasks that have not been started yet are also counted as skipped in the status message reported to the user.
```

It will help to determine whether `StoppedRunFinally` is allowed to run retries to come to the conclusion on this.

lbernick:
thanks Priti! My guess is that because we don't explicitly say that retries will not be exhausted, we intend for retries to be exhausted during graceful stopping. However, I think in practice, they are not, due to this same bug. (This makes me wonder if we actually have full test coverage for this feature.)

jerop:
> It will help to determine whether `StoppedRunFinally` is allowed to run retries to come to the conclusion on this.

@pritidesai tested it out using this modified version of [an example demonstrating retries](https://github.com/tektoncd/pipeline/blob/f2da9a877f295a6cb865d8dee7b52872089c061d/examples/v1beta1/pipelineruns/using-retries-and-retry-count-variables.yaml):

```yaml
kind: PipelineRun
metadata:
  name: retry-example
spec:
  serviceAccountName: 'default'
  # comment out the line below when the retry-me TaskRun is running and update the PipelineRun
  # status: "StoppedRunFinally"
  pipelineSpec:
    tasks:
    - name: retry-me
      retries: 5
      params:
      - name: pipelineTask-retries
        value: "$(context.pipelineTask.retries)"
      - name: pipelineTask-retry-count
        value: "$(context.task.retry-count)"
      taskSpec:
        params:
        - name: pipelineTask-retries
        - name: pipelineTask-retry-count
        steps:
        - image: alpine:3.12.0
          script: |
            #!/usr/bin/env sh
            if [ "$(params.pipelineTask-retry-count)" == "$(params.pipelineTask-retries)" ]; then
              echo "This is the last retry."
              exit 0;
            fi
            echo "The PipelineTask has retried $(context.task.retry-count) times."
            exit 1
    - name: hello-world
      runAfter:
      - retry-me
      taskSpec:
        steps:
        - image: busybox
          script: |
            echo "hello world"
    finally:
    - name: bye-world
      taskSpec:
        steps:
          - image: busybox
            script: |
              echo "bye world"
```
 > ...we intend for retries to be exhausted during graceful stopping. However, I think in practice, they are not, due to this same bug.

@lbernick is right that it remained hanging due to this bug 🙃 

```shell
$ tkn pr list
NAME            STARTED          DURATION   STATUS
retry-example   10 minutes ago   ---        Running(StoppedRunningFinally)
```

pritidesai:
> @lbernick @pritidesai - given that retries during graceful stopping is also underspecified and it looks like the retries will be exhausted when we fix this bug, it makes sense to allow exhausting retries in the internal stopping mode too (then we can specify what happens to retries during graceful stopping)

Sounds good 🎉  thank you both!

lbernick:
Awesome, thank you both for taking the time to test this out! once #4667 is merged I'll update this to address retries for both normal stopping and graceful stopping- it sounds like we want to exhaust retries for both.

lbernick:
> @lbernick given the hold, may I move this to the next milestone? hoping to release 0.34 today

yeah that makes sense, was hoping to have this in for this release but alas

lbernick:
@jerop @pritidesai I've rebased and updated to address graceful stopping as well -- should be ready for another look!

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -232,31 +232,43 @@ func (state PipelineRunState) getNextTasks(candidateTasks sets.String) []*Resolv
...
+				if status != nil {
+					isCancelled = isCancelled || status.Reason == v1alpha1.RunReasonCancelled
 				}
-				if status != nil && status.IsFalse() {
-					if !(isCancelled || status.Reason == ReasonConditionCheckFailed) {
-						if t.HasRemainingRetries() {
-							tasks = append(tasks, t)
-						}
+			}
+			if status != nil && status.IsFalse() {
```
sbwsg, line 264: `IsFalse` performs a `nil` check [it looks like](https://github.com/tektoncd/pipeline/blob/23ffdba46dca71e30438f8ca5ab4f16d4b3bb8d7/vendor/knative.dev/pkg/apis/condition_types.go#L97-L99), so we don't need the `status != nil` I don't think?

sbwsg:
One small nit but not a blocker.

/lgtm

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
lbernick: nice, thanks! done

dibyom (approved):

rolfedh:
Proposed release note: 
```
Before this update, when a pipeline run was in a "stopping" state, that is, if a task run failed, other task runs might not complete their retries. As a result, no `finally` tasks were scheduled, which caused the pipeline to hang. This update resolves the issue. `TaskRuns` and `Run` objects can retry when a pipeline run has stopped, even by  graceful stopping, so that pipeline runs can complete.
```

Issue #3862, opened 2021-03-31
Cloud Events for Runs


afrittoli:
### Feature request

Cloud events should be sent Runs.

It needs to be identified whether we want this to be a responsibility of the custom task controllers, the pipeline controller or a dedicated one. 

### Use case

We support cloud events for taskruns and pipelineruns, we should support them for runs too.

ImJasonH:
I like the idea of basic cloudevents for Run start/success/failure that everybody gets out of the box, and documentation/samples for Custom Task authors to add their own if they want.

The pipelines controller doesn't currently do anything with Runs, I'm not even sure it has a no-op reconciler for them, so if we decide we want to support this we'd add one and do it there.

jerop:
moved it to milestone for v0.31

afrittoli:
The `PipelineRun` controller creates the `Run` so it could send that event. 
The `PipelineRun` controller also watches for changes to `Runs` so it would need to check if there was a change between the local copy of the status and the one reported by the informer. This could be done in `updateRunsStatusDirectly` - however since we plan on removing the local copy of the status, I think it would not be wise to implement a new feature that relies on that.

My proposal here would be to implement this in the external controller instead. This would then be implemented once the external controller replaced the inbuilt one for sending events.

vdemeester:
I tend to agree on the external controller 👼🏼

pritidesai:
thanks @afrittoli!

Yup implementing as part of an external controller sounds reasonable. This can give freedom to the external controller in designing the events based on their requirements.

afrittoli:
Implemented in three PRs:

- [x] #4658
- [ ] #4659
- [ ] #4663

PR #491, opened 2019-02-12
Make sure resource names do not exceed limit of 63 characters 

pivotal-nader-ziada:
Fixes: #481 

### Proposed changes

 - Generated names for `taskruns` and `build steps` get a random suffix
of five alphanumerics. The string is guaranteed to not exceed the
length of a standard Kubernetes name (63 characters)
 - Update tests using fixed seed for the name generation
 - Remove `rand string` func from test folder and use `GenerateName()`
 - CRD types already have validation for names in `validateObjectMetadata()` 

This PR was co-authored by @bobcatfish, but will not put her name in the commit to avoid CLA hassle, (sorry Christie 🤷‍♂️)

Issue #481, opened 2019-02-07
Resource name too long, cannot create taskrun


afrittoli:
## Expected Behavior
The pipelinerun automatically creates taskruns for every task. It sets up the taskrun name using <pipelinerun-name>-<task-name>. The pod name will be even longer with <pipelinerun-name>-<task-name>-pod<uuid>. When then name is longer then 63 characters it fails to setup the taskrun.

I think we should have some logic to automatically truncated the names, or at least we should create shorter names, e.g. instead of including the whole pipeline name include the uuid. 

This applies to multiple "names" in our system, including:
* CRD names (i.e. TaskRun, Task, Pipeline, PipelineRun, PipelineResource)
   * Generated CRD names (e.g. TaskRuns created by PipelineRuns)
* Names that get translated into resource names, e.g. Build steps become pods

## Actual Behavior
Pipeline fails to reconcile.

```
Events:
  Type     Reason                 Age                    From                 Message
  ----     ------                 ----                   ----                 -------
  Normal   PipelineRunSucceeded   7m20s (x20 over 14m)   pipeline-controller  PipelineRun reconciled successfully
  Warning  TaskRunCreationFailed  6m50s (x3 over 6m50s)  pipeline-controller  Failed to create TaskRun "health-helm-cd-pipeline-run-28jrg-source-to-image-health-frontend": Internal error occurred: admission webhook "webhook.pipeline.knative.dev" denied the request: mutation failed: Invalid resource name: length must be no more than 63 characters: metadata.name
  Warning  PipelineRunFailed      6m50s (x3 over 6m50s)  pipeline-controller  PipelineRun failed to reconcile
```
## Steps to Reproduce the Problem

1. Define a pipelinerun so that pipelinerun name + task name are longer than 63chars

## Additional Info
I initially didn't hit this as I was using static pipelinerun names. When I switched to generateName, it added enough characters to go over 63.

bobcatfish:
`health-helm-cd-pipeline-run-28jrg-source-to-image-health-frontend`

interesting :thinking:

specifically talking about the automatically created taskruns (generated by a pipelinerun) we could probably give them way less meaningful names and get away with it. but we also need a solution names ppl explicitly set :thinking: :thinking::thinking:, tho maybe it's better to tell the user their name is too long and let them pick something else if they are creating a taskrun or pipelinerun manually?

dwnusbaum:
[...]

PR #4659, opened 2022-03-10
Run events

afrittoli:
# Changes

Add CloudEvents for Run definitions

Add the definitions for new CloudEvents for Runs.
The events are not sent yet. The event types are documented and
the cloudevent package supports creating events for Runs.

Partially addresses #3862

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
CloudEvents for Run have been defined, but they are not sent yet.
```


vdemeester (approved):

docs/events.md
```diff
@@ -6,11 +6,12 @@ weight: 700
 -->
 # Events in Tekton
 
-Tekton's task controller emits [Kubernetes events](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#event-v1-core)
```
lbernick, line 9: just some grammar nits: `controllers emits` -> `controllers emit` and `the underlying TaskRun do` -> `the underlying TaskRuns do`

docs/events.md
```diff
@@ -68,6 +69,12 @@ Resource      |Event    |Event Type
 `PipelineRun` | `Condition Change while Running` | `dev.tekton.event.pipelinerun.unknown.v1`
 `PipelineRun` | `Succeed` | `dev.tekton.event.pipelinerun.successful.v1`
 `PipelineRun` | `Failed`  | `dev.tekton.event.pipelinerun.failed.v1`
+`Run`         | `Started` | `dev.tekton.event.run.started.v1`
```
lbernick, line 72: what's the reason not to send a condition change while running event? is this because we don't know what the run's previous status was, and can't make assumptions based on what it currently is? Is this a feature users might want? I'm a bit confused because the PR description describes an event cache but I'm not really seeing where that happens.

pkg/reconciler/events/cloudevent/cloudevent_test.go
```diff
@@ -83,6 +86,34 @@ func getPipelineRunByCondition(status corev1.ConditionStatus, reason string) *v1
...
 
+func getRunByCondition(status corev1.ConditionStatus, reason string) *v1alpha1.Run {
+	myrun := &v1alpha1.Run{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Run",
+			APIVersion: "v1alpha1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      runName,
+			Namespace: "marshmallow",
```
lbernick, line 97: :)

pkg/reconciler/events/cloudevent/cloudevent_test.go
```diff
@@ -83,6 +86,34 @@ func getPipelineRunByCondition(status corev1.ConditionStatus, reason string) *v1
 	}
 }
 
+func getRunByCondition(status corev1.ConditionStatus, reason string) *v1alpha1.Run {
```
lbernick, line 89: nit: a clearer name might be `createRunWithCondition`

pkg/reconciler/events/cloudevent/cloudevent_test.go
```diff
@@ -209,3 +240,62 @@ func TestEventForPipelineRun(t *testing.T) {
...
+	runTests := []struct {
+		desc          string
+		run           *v1alpha1.Run
+		wantEventType TektonEventType
+	}{{
+		desc:          "send a cloud event with unset condition, just started",
+		run:           getRunByCondition("", ""),
+		wantEventType: RunStartedEventV1,
+	}, {
+		desc:          "send a cloud event with unknown status pipelinerun, empty reason",
```
lbernick, line 254: should these test names read `with unknown status run`?

pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -142,12 +156,31 @@ func eventForPipelineRun(pipelineRun *v1beta1.PipelineRun) (*cloudevents.Event,
 	return eventForObjectWithCondition(pipelineRun)
 }
 
+// eventForPipelineRun will create a new event based on a PipelineRun,
+// or return an error if not possible.
+func eventForRun(run *v1alpha1.Run) (*cloudevents.Event, error) {
```
lbernick, line 161: method comment should read `new event based on a Run` and `check if the Run is defined`

pkg/reconciler/events/cloudevent/cloudevent_test.go
```diff
@@ -209,3 +240,62 @@ func TestEventForPipelineRun(t *testing.T) {
...
+		wantEventType: RunSuccessfulEventV1,
+	}, {
+		desc:          "send a cloud event with unknown status pipelinerun",
+		run:           getRunByCondition(corev1.ConditionFalse, "meh"),
+		wantEventType: RunFailedEventV1,
+	}}
+
+	for _, c := range runTests {
+		t.Run(c.desc, func(t *testing.T) {
+			names.TestingSeed()
```
lbernick, line 273: I see `names.TestingSeed` a lot in our tests-- what's the reason it's needed here, just out of curiosity?

docs/events.md
afrittoli: Nice catch, thank you!

docs/events.md
afrittoli: We cannot make assumptions about the `Run` and it will use its `Conditions` beyond what we specify in the custom task interface. So we can see that the resource has been created (start), we can see that an initial condition was set (running) and we can see when it's finished (successfully or not).

pkg/reconciler/events/cloudevent/cloudevent.go
afrittoli: +1

pkg/reconciler/events/cloudevent/cloudevent_test.go
afrittoli: 👍

pkg/reconciler/events/cloudevent/cloudevent_test.go
afrittoli: 👍

pkg/reconciler/events/cloudevent/cloudevent_test.go
afrittoli: Copy paste? I wondered the same but I lazy :P I will try to find out

pritidesai: its there for a reason, I have been wondering about it as well since I started writing unit tests.

It was introduced to use `fixed` seed for the generated names as part of PR https://github.com/tektoncd/pipeline/pull/491. This PR fixed the issue where it was reported the resource names were too long https://github.com/tektoncd/pipeline/issues/481.

@afrittoli some of these might no longer be needed since now we have fixed names for some of the resources as your implemented in https://github.com/tektoncd/pipeline/pull/4361.

I generally comment out one line at a time to see if its breaking my test if I am not sure what that code is doing 😄 Or add the same call twice 🙃 to understand the impact.

afrittoli: Thanks @pritidesai - good idea :)

`names.TestingSeed` is used to initialise the random generator used in names, so that you can get know what name to expect to validate results in tests. As @pritidesai mentioned, the need for this is probably reduced now that children resources have a more predictable name. 

CloudEvent tests only match event types, so the name seed is not strictly required. It does cost anything or do any harm either, but I'll remove it as it's not needed.

PR #4667, opened 2022-03-11
PipelineRunState cleanup: tests + comments

lbernick:
# Changes

This commit cleans up PipelineRunState logic in preparation for a subsequent commit affecting retries.
It adds tests for DAGExecutionQueue and ensures that that function does not return any results if the PipelineRun is cancelled,
since we do not want to continue to schedule tasks if the PipelineRun is cancelled. (This is already documented and implemented elsewhere.)
No functional changes.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
@@ -290,22 +290,28 @@ func (facts *PipelineRunFacts) IsRunning() bool {
 	return false
 }
 
-// IsGracefullyCancelled returns true if the PipelineRun won't be scheduling any new Task because it was gracefully cancelled
+// IsCancelled returns true if the PipelineRun was cancelled
+func (facts *PipelineRunFacts) IsCancelled() bool {
+	return facts.SpecStatus == v1beta1.PipelineRunSpecStatusCancelledDeprecated ||
+		facts.SpecStatus == v1beta1.PipelineRunSpecStatusCancelled
+}
+
+// IsGracefullyCancelled returns true if the PipelineRun was gracefully cancelled
 func (facts *PipelineRunFacts) IsGracefullyCancelled() bool {
 	return facts.SpecStatus == v1beta1.PipelineRunSpecStatusCancelledRunFinally
 }
 
-// IsGracefullyStopped returns true if the PipelineRun won't be scheduling any new Task because it was gracefully stopped
+// IsGracefullyStopped returns true if the PipelineRun was gracefully stopped
 func (facts *PipelineRunFacts) IsGracefullyStopped() bool {
 	return facts.SpecStatus == v1beta1.PipelineRunSpecStatusStoppedRunFinally
 }
 
 // DAGExecutionQueue returns a list of DAG tasks which needs to be scheduled next
 func (facts *PipelineRunFacts) DAGExecutionQueue() (PipelineRunState, error) {
-	tasks := PipelineRunState{}
-	// when pipeline run is stopping, gracefully cancelled or stopped, do not schedule any new task and only
+	var tasks PipelineRunState
+	// when pipeline run is stopping, cancelled, gracefully cancelled or stopped, do not schedule any new task and only
 	// wait for all running tasks to complete and report their status
-	if !facts.IsStopping() && !facts.IsGracefullyCancelled() && !facts.IsGracefullyStopped() {
+	if !facts.IsStopping() && !facts.IsCancelled() && !facts.IsGracefullyCancelled() && !facts.IsGracefullyStopped() {
 		// candidateTasks is initialized to DAG root nodes to start pipeline execution
 		// candidateTasks is derived based on successfully finished tasks and/or skipped tasks
 		candidateTasks, err := dag.GetSchedulable(facts.TasksGraph, facts.successfulOrSkippedDAGTasks()...)
@@ -318,13 +324,12 @@ func (facts *PipelineRunFacts) DAGExecutionQueue() (PipelineRunState, error) {
 }
 
 // GetFinalTasks returns a list of final tasks without any taskRun associated with it
-// GetFinalTasks returns final tasks only when all DAG tasks have finished executing successfully or skipped or
-// any one DAG task resulted in failure
+// GetFinalTasks returns final tasks only when all DAG tasks have finished executing successfully or have been skipped
 func (facts *PipelineRunFacts) GetFinalTasks() PipelineRunState {
 	tasks := PipelineRunState{}
 	finalCandidates := sets.NewString()
-	// check either pipeline has finished executing all DAG pipelineTasks
-	// or any one of the DAG pipelineTask has failed
+	// check either pipeline has finished executing all DAG pipelineTasks,
+	// where "finished executing" means succeeded, failed, or skipped.
 	if facts.checkDAGTasksDone() {
 		// return list of tasks with all final tasks
 		for _, t := range facts.State {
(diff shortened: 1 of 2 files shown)


pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -290,6 +290,12 @@ func (facts *PipelineRunFacts) IsRunning() bool {
 	return false
 }
 
+// IsCancelled returns true if the PipelineRun won't be scheduling any new Task because it was cancelled
+func (facts *PipelineRunFacts) IsCancelled() bool {
```
pritidesai, line 294: Sorry for being very picky here, please update the comment to not include the scheduling aspect but focusing on what this function does.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -302,10 +308,10 @@ func (facts *PipelineRunFacts) IsGracefullyStopped() bool {
 
 // DAGExecutionQueue returns a list of DAG tasks which needs to be scheduled next
 func (facts *PipelineRunFacts) DAGExecutionQueue() (PipelineRunState, error) {
-	tasks := PipelineRunState{}
-	// when pipeline run is stopping, gracefully cancelled or stopped, do not schedule any new task and only
+	var tasks PipelineRunState
```
pritidesai, line 311: why changing to initialized variable to non-initialized?

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
pritidesai: I notice the other functions have the similar style of comments. Since you are cleaning up, do you mind updating those functions (`IsGracefullyStopped` and `IsGracefullyCancelled`) as well. Its good to have additional context around scheduling but this function is just comparing the status.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -318,13 +324,12 @@ func (facts *PipelineRunFacts) DAGExecutionQueue() (PipelineRunState, error) {
 }
 
 // GetFinalTasks returns a list of final tasks without any taskRun associated with it
-// GetFinalTasks returns final tasks only when all DAG tasks have finished executing successfully or skipped or
-// any one DAG task resulted in failure
```
pritidesai, line 322: doesn't this still apply? `isDone` implies skipped, succeeded or failed

pkg/reconciler/pipelinerun/resources/pipelinerunstate_test.go
```diff
@@ -600,6 +600,177 @@ func TestGetNextTaskWithRetries(t *testing.T) {
 	}
 }
 
+// TestDAGExecutionQueue tests the DAGExecutionQueue function for PipelineTasks
+// in different states (without dependencies on each other) and the PipelineRun in different states.
+func TestDAGExecutionQueue(t *testing.T) {
```
pritidesai, line 605: thanks a bunch for adding this unit test 🙏

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -302,10 +308,10 @@ func (facts *PipelineRunFacts) IsGracefullyStopped() bool {
 
 // DAGExecutionQueue returns a list of DAG tasks which needs to be scheduled next
 func (facts *PipelineRunFacts) DAGExecutionQueue() (PipelineRunState, error) {
-	tasks := PipelineRunState{}
-	// when pipeline run is stopping, gracefully cancelled or stopped, do not schedule any new task and only
+	var tasks PipelineRunState
+	// when pipeline run is stopping, cancelled, gracefully cancelled or stopped, do not schedule any new task and only
 	// wait for all running tasks to complete and report their status
-	if !facts.IsStopping() && !facts.IsGracefullyCancelled() && !facts.IsGracefullyStopped() {
+	if !facts.IsStopping() && !facts.IsCancelled() && !facts.IsGracefullyCancelled() && !facts.IsGracefullyStopped() {
```
pritidesai, line 314: I am surprised there is no functional change even after adding this check 🤔 

With this check, if the pipeline run is `cancelled`, it will not attempt to schedule any new task in addition to the check done in the beginning of reconciler:

https://github.com/tektoncd/pipeline/blob/9a2deb3e34ae252d32c017d0155b963aaca8218f/pkg/reconciler/pipelinerun/pipelinerun.go#L211-L215

pritidesai: `stopping` is an internal state and not allowed to set by updating the pipelineRun status (at least not officially). I am wondering why are we checking these other states (cancelled, gracefullycancelled, and gracefullystopped) here? Should we move these checks along with the `cancelled` at the beginning of the reconciler?

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
lbernick: this is mostly personal preference. This change allows me to not have to include `want: PipelineRunState{}` in each test case in TestDAGExecutionQueue.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
lbernick: this is not picky at all, I would make the same comment! Done.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
lbernick: This comment is being updated to reflect the current behavior of GetFinalTasks, since the desired behavior is still under discussion in #4651. GetFinalTasks will not return any tasks if one DAG task resulted in failure but other tasks are still running. Instead, it waits for all tasks to finish before returning.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
pritidesai: yup it is possible to have such gap in the existing test coverage and appreciate your effort of adding more tests 😍 

My understanding is we can definitely move the graceful cancellation out of this check but not `stopping`. The `stopping` state is something which can happen during the same reconcile cycle while `cancellation` cannot be determined until the next reconcile cycle from this point of execution, does it make sense?

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
pritidesai: Nice 👍

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
lbernick: thanks! yeah I understand now that cancellation takes an extra reconcile cycle. I think pulling graceful cancellation out of this block and moving it to where cancellation is might take some extra refactoring (since we still need to attempt retries) so I would prefer it didn't block this PR, but I can create an issue and link it as a TODO if that would work? I tested this PR out using the retries example from [this comment](https://github.com/tektoncd/pipeline/pull/4651#issuecomment-1068147952) with non-graceful cancellation instead of graceful stopping and it appears to still work as intended with this PR.

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
jerop: given that we agreed on the behavior #4651, can we update this comment to reflect that? would rather not have the comment reflecting the buggy behavior

jerop (approved):
thanks @lbernick!

lbernick:
> given that we agreed on the behavior https://github.com/tektoncd/pipeline/pull/4651, can we update this comment to reflect that? would rather not have the comment reflecting the buggy behavior

#4651 actually leaves the behavior of getfinaltasks as is, so no changes are needed to this specific comment. I think it's better to have the comment reflect the actual behavior of the function even if the usage of the function is not correct.

pritidesai:
thanks @lbernick 
/lgtm

PR #4671, opened 2022-03-14
Refactor When Expressions Validation

jerop:
# Changes

In this change, we refactor the validation for `when` expressions
to remove duplications, including:
- moving the validation that "only one of `conditions` or `when`
expressions is used" to a member function of `pipelinetask`
- simplifying the overall `when` expressions validation function to
handle one list of `pipelinetasks`, then reusing it for both `tasks`
and `finally` fields

cc @pritidesai 

/kind cleanup

# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
@@ -255,6 +255,34 @@ func TestPipelineSpec_Validate_Failure(t *testing.T) {
 			Paths:   []string{"tasks[0].conditions", "tasks[0].when"},
 		},
 	}, {
+		name: "invalid pipeline with one pipeline task having both conditions and when expressions in finally",
+		ps: &PipelineSpec{
+			Description: "this is an invalid pipeline with invalid pipeline task",
+			Tasks: []PipelineTask{{
+				Name:    "invalid-pipeline-task",
+				TaskRef: &TaskRef{Name: "foo-task"},
+			}},
+			Finally: []PipelineTask{{
+				Name:    "invalid-pipeline-final-task",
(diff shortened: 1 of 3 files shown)


jerop:
[error](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4671/pull-tekton-pipeline-alpha-integration-tests/1503396511931699200/) creating e2e tests cluster 😕 

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -254,6 +254,34 @@ func TestPipelineSpec_Validate_Failure(t *testing.T) {
...
+					Input:    "foo",
+					Operator: selection.In,
+					Values:   []string{"bar"},
+				}},
+				Conditions: []PipelineTaskCondition{{
+					ConditionRef: "some-condition",
+				}},
+			}},
+		},
+		expectedError: *apis.ErrGeneric("").Also(&apis.FieldError{
```
lbernick, line 278: If conditions are not allowed in finally tasks at all, I think it's almost more confusing to have the error message state that you can't have both conditions and when expressions.

jerop: I see what you're saying, but a user would get both error messages, that `conditions` is not allowed in `finally` and that both `conditions` and `when` expressions cannot be together

jerop: it's all good, wouldn't want to make it more confusing to users when trying to clean up the code to remove duplications

PR #4600, opened 2022-02-18
TEP-0090: Matrix - API Change and Feature Flag

jerop:
# Changes

[TEP-0090][tep-0090] proposed executing a `PipelineTask` in parallel `TaskRuns` and `Runs` with substitutions from combinations of `Parameters` in a `Matrix`.

In this change, we add the `matrix` field to the `PipelineTask` specification, which takes a list of `Parameters`. We gate this field behind the `alpha` feature gate. We add validation and testing that the field is available in `alpha` only - we will add further validation in upcoming changes. We also add documentation the API change - we will incrementally add further details as we iterate on the implementation of [TEP-0090: Matrix][tep-0090].

[tep-0090]: https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md

/kind feature

Issue: https://github.com/tektoncd/pipeline/issues/2050

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including functionality, content, code)
- [x] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

```release-note
Added syntax support for `Matrix` as an alpha feature. It is still in a very early stage of development and is not yet fully functional.
```

PR #4694, opened 2022-03-21
WIP: [TEP-0100] Embedded TaskRun and Run statuses in PipelineRuns

abayer:
# Changes

See https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md

This adds new choices for how `TaskRun` and `Run` statuses are stored/referenced in `PipelineRun`
statuses, depending on how the new feature flag, `embedded-status` is set:
* `full` - the current default, with the `TaskRun` and `Run` statuses embedded in full.
* `minimal` - instead of storing the full embedded statuses, information is stored about the task's
  version, kind, `PipelineTask` name, and the underlying name of the `TaskRun` or `Run`. This can be
  used to look up the `TaskRun` or `Run` itself and get the full status from there. Information
  which is specific to the `PipelineRun`, namely condition checks status and when expressions which
  may have blocked creation of an actual `TaskRun` or `Run` are stored as well.
* `both` - both the full embedded `TaskRun` and `Run` statuses and the minimal references are stored.

The condition check information will be removed from the minimal references once the now-deprecated
condition check functionality is removed from Tekton Pipeline.

As described in https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md#beta-api,
9 months after this change has been released in `v1beta1`, the default value for the
`embedded-status` feature flag will be changed from `full` to `minimal`, and a few months after
that, the `embedded-status` feature flag will be removed entirely and only the `minimal` behavior
will remain.

/kind tep
/kind api-change

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
TODO
```


afrittoli (commented):
🎉  Thanks for implementing this!! 
One question on the "both" option.

pkg/apis/config/feature_flags.go
```diff
@@ -30,6 +30,15 @@ const (
 	StableAPIFields = "stable"
 	// AlphaAPIFields is the value used for "enable-api-fields" when alpha APIs should be usable as well.
 	AlphaAPIFields = "alpha"
+	// FullEmbeddedStatus is the value used for "embedded-status" when the full statuses of TaskRuns and Runs should be
+	// embedded in PipelineRunStatusFields, but ChildReferences should not be used.
+	FullEmbeddedStatus = "full"
+	// BothEmbeddedStatus is the value used for "embedded-status" when full embedded statuses of TaskRuns and Runs as
+	// well as ChildReferences should be used in PipelineRunStatusFields.
+	BothEmbeddedStatus = "both"
```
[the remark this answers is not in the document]
abayer: cc @jerop and @lbernick for that. =)

abayer:
So I _think_ I was comprehensive enough in unit tests - both for the new code, and verifying that existing code doesn't barf when the feature flag is either `both` or ` minimal` rather than the initial default/current behavior of `full`. But I think integration testing of the various feature flags would be good to add, and I will do so.

And then there's a bunch of doc changes to make, of course.

pkg/apis/config/feature_flags.go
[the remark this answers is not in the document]
lbernick: I went back to look at the API WG notes where this was discussed, and it just says "To prevent being broken between versions" :'( The TEP states: "Users can opt in to use `both` at any time, but it is never the default value. It provides a seamless transition for API clients for a short period needed to upgrade to minimal." 
The use case in mind here is if someone wants to update their installation of pipelines and also their use of client libraries: first they would swap to "both", then swap to using ChildRefs, then swap to "minimal".

pkg/apis/config/testdata/feature-flags-invalid-enable-api-fields.yaml
```diff
@@ -18,4 +18,4 @@ metadata:
   name: feature-flags
   namespace: tekton-pipelines
 data:
-  enable-api-fields: "im-not-a-valid-feature-gate"
+  embedded-status: "im-not-a-valid-feature-gate"
```
lbernick, line 21: I think this change was made to the incorrect file.

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -396,6 +398,22 @@ func (pr *PipelineRunStatus) MarkRunning(reason, messageFormat string, messageA
...
+type ChildStatusReference struct {
+	runtime.TypeMeta `json:",inline"`
+	// Name is the name of the TaskRun or Run this is referencing.
+	Name string `json:"name,omitempty"`
+	// PipelineTaskName is the name of the PipelineTask this is referencing.
+	PipelineTaskName string `json:"pipelineTaskName,omitempty"`
+
+	// ConditionChecks is the the list of condition checks, including their names and statuses, for the PipelineTask
+	// +optional
+	ConditionChecks []*PipelineRunChildConditionCheckStatus `json:"conditionChecks,omitempty"`
```
lbernick, line 411: could you add a note that condition checks are deprecated and this will be removed?

lbernick (commented):
thanks @abayer! Would you mind splitting into one PR with the new flag + API fields + docs, and one PR with implementation?

abayer:
@lbernick - I'm not sure I get what the value of splitting into two PRs would be - obviously, my gut instinct is gonna be to avoid the work of teasing apart the current PR. =) But if there's a reason, I am open to persuasion!

pkg/apis/config/testdata/feature-flags-invalid-enable-api-fields.yaml
abayer: Yeah, that's a boo-boo. Fixing!

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
abayer: Done!

lbernick:
> @lbernick - I'm not sure I get what the value of splitting into two PRs would be - obviously, my gut instinct is gonna be to avoid the work of teasing apart the current PR. =) But if there's a reason, I am open to persuasion!

Not specific to this PR, but in general, because smaller PRs are easier to review, merge, and roll back (you will get lots of results for searching for things like "why small pull requests"). The split I've suggested is just one possible way to split this into self contained smaller PRs.

abayer:
@lbernick - Ok, I get the "smaller PRs" angle, but I'm a bit unsure about adding the flag/fields without any implementation. This really is all one atomic unit of changes, and it feels a bit off to split it into multiple PRs if the reason to do so is primarily just to have smaller PRs.

lbernick:
> @lbernick - Ok, I get the "smaller PRs" angle, but I'm a bit unsure about adding the flag/fields without any implementation. This really is all one atomic unit of changes, and it feels a bit off to split it into multiple PRs if the reason to do so is primarily just to have smaller PRs.

FWIW this is the approach taken by #4600, #4502 and #4575 so there's some precedent

abayer:
"precedent" is always a magic word with me, so I'll tease things apart tomorrow. =)

abayer:
Closing this in favor of #4705 and #4706.

PR #4704, opened 2022-03-23
TEP-0090: `Matrix` - `Parameters`

jerop:
# Changes

[TEP-0090][tep-0090] proposed executing a `PipelineTask` in parallel `TaskRuns` and `Runs` with substitutions from combinations of `Parameters` in a `Matrix`.

In this change, we add validation for `Parameters` including:
- `Parameter` must be of type `Array`
- `Parameter` must be in one of `matrix` or `params`, not both
- `Parameter` in `Matrix` must exist in `Parameters` declarations in `Pipeline`

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
`Parameters` in `Matrix`:
- must be of type `Array`
- must not be in `Params` field as well
- must be declared in the `Parameters` in `Pipeline` specification

Note that `Matrix` is not yet fully functional.
```

### DIFF
--- a/docs/matrix.md
+++ b/docs/matrix.md
@@ -9,6 +9,7 @@ weight: 11
 
 - [Overview](#overview)
 - [Configuring a Matrix](#configuring-a-matrix)
+  - [Parameters](#parameters)
   - [Context Variables](#context-variables)
 
 ## Overview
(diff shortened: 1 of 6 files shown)


docs/matrix.md
```diff
@@ -25,8 +26,60 @@ Tekton.
 ## Configuring a Matrix
 
 A `Matrix` supports the following features:
+* [Parameters](#parameters)
 * [Context Variables](#context-variables)
 
+### Parameters
+
+The `Matrix` will take `Parameters` of type `Array` only, which will be supplied to the
```
sbwsg, line 34: Total nit: users would only ever type `"array"` and `"string"` in their yaml I think.  Suggest using these in place of `Array` and `String`.

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -178,6 +178,47 @@ func validatePipelineParametersVariablesInTaskParameters(params []Param, prefix
...
+func (p *Param) validateArrayVariablesInMatrixParameter(prefix string, paramNames sets.String, arrayParamNames sets.String) (errs *apis.FieldError) {
+	for idx, arrayElement := range p.Value.ArrayVal {
+		errs = errs.Also(validateArrayVariable(arrayElement, prefix, paramNames, arrayParamNames).ViaFieldIndex("value", idx).ViaFieldKey("matrix", p.Name))
+	}
+	return errs
+}
+
+func validateParametersInTaskMatrix(matrix []Param) (errs *apis.FieldError) {
+	for _, param := range matrix {
+		errs = errs.Also(param.validateMatrixParameterIsArrayTypeOnly())
```
sbwsg, line 197: Small readability suggestion here:  given `param.validateMatrixParameterIsArrayTypeOnly` and `param.validateArrayVariablesInMatrixParameter` are each only used once I suggest in-lining them directly rather than having separate methods for them.

e.g.

```go
func validateParametersInTaskMatrix(matrix []Param) (errs *apis.FieldError) {
	for _, param := range matrix {
		if param.Value.Type != ParamTypeArray {
			errs = errs.Also(apis.ErrInvalidValue("parameters of type array only are allowed in matrix", "").ViaFieldKey("matrix", p.Name))
		}
	}
	return errs
}
```

The reason for this is that (at least for me, not sure what others think) method names like `validateArrayVariablesInMatrixParameter` are quite difficult to interpret while also keeping the control flow clear in my head.

Similarly for `validatePipelineParametersVariablesInMatrixParameters`, I think I'm able to more quickly understand what the func does when the `for` loops are nested:

```go
func validatePipelineParametersVariablesInMatrixParameters(matrix []Param, prefix string, paramNames sets.String, arrayParamNames sets.String) (errs *apis.FieldError) {
	for _, param := range matrix {
		for idx, arrayElement := range param.Value.ArrayVal {
			errs = errs.Also(validateArrayVariable(arrayElement, prefix, paramNames, arrayParamNames).ViaFieldIndex("value", idx).ViaFieldKey("matrix", p.Name))
		}
	}
	return errs
}
```

Not a big issue from my pov if you decide it's preferable as-is, but thought I'd mention it.

sbwsg (commented):
I left a couple of small style nits but otherwise this looks good to me!

docs/matrix.md
jerop: makes sense, updated them to `"array"` and `"string"`, thanks!

pkg/apis/pipeline/v1beta1/param_types.go
jerop: moved the functions inline, and I agree that it's clearer that way! thank you @sbwsg :)

sbwsg:
/lgtm

Cheers @jerop!

dibyom (approved):
LGTM!

Issue #4657, opened 2022-03-08
Update PipelineRun API documentation


lbernick:
Similar to #4601. We should update the [PipelineRun API docs](https://github.com/tektoncd/pipeline/blob/main/docs/pipelineruns.md#configuring-a-pipelinerun) to reflect all existing fields.

/kind documentation
/help-wanted
/good-first-issue

ssunitaa:
Hi there,

Thank you so much for your reply and for giving me an opportunity to
contribute and learn here. Please give me some time to go through the
requirements and understand the process. I shall soon be coming up with my
list of questions. 🙂
However, to start with -
1. How can I communicate with the team - through this email or GitHub? I am
a novice in GitHub, so quite a lot to learn.
2. Is this a REST API that we have to document?

I'm ready and excited!

Thanks and Regards,
Sunita Arora

PR #4705, opened 2022-03-23
[TEP-0100] Fields/flags/docs for embedded TaskRun and Run statuses in PipelineRuns

abayer:
# Changes

Replaces #4694 (in part - there'll be a followup for the actual implementation)

See https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md

This adds new choices for how `TaskRun` and `Run` statuses are stored/referenced in `PipelineRun`
statuses, depending on how the new feature flag, `embedded-status` is set:
* `full` - the current default, with the `TaskRun` and `Run` statuses embedded in full.
* `minimal` - instead of storing the full embedded statuses, information is stored about the task's
   version, kind, `PipelineTask` name, and the underlying name of the `TaskRun` or `Run`. This can be
   used to look up the `TaskRun` or `Run` itself and get the full status from there. Information
   which is specific to the `PipelineRun`, namely condition checks status and when expressions which
   may have blocked creation of an actual `TaskRun` or `Run` are stored as well.
* `both` - both the full embedded `TaskRun` and `Run` statuses and the minimal references are stored.

The condition check information will be removed from the minimal references once the now-deprecated
condition check functionality is removed from Tekton Pipeline.

As described in https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md#beta-api,
9 months after this change has been released in `v1beta1`, the default value for the
`embedded-status` feature flag will be changed from `full` to `minimal`, and a few months after
that, the `embedded-status` feature flag will be removed entirely and only the `minimal` behavior
will remain.

Note that this also may address #4657, since I had to add documentation for `PipelineRunStatus`,
and while I was there, I added some missing top-level fields to the `PipelineRun` definition.

The actual implementation will follow this PR.

/kind tep
/kind api-change

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Added syntactic support and feature flags for minimal `TaskRun` and `Run` statuses within `PipelineRun` statuses.
```


abayer:
Fun, assign isn't working, I guess.

docs/install.md
```diff
@@ -390,6 +390,11 @@ features](#alpha-features) to be used.
   to "false" to guard a `Task` and its dependent `Tasks`. It defaults to "true". For more information, see [guarding
   `Task` execution using `when` expressions](pipelines.md#guard-task-execution-using-whenexpressions).
 
+- `embedded-status`: set this flag to "full" to enable full embedding of `TaskRun` and `Run` statuses in the 
+ `PipelineRun` status. Set it to "minimal" to populate the `ChildReferences` field in the `PipelineRun` status with
+  name, kind, and API version information for each `TaskRun` and `Run` in the `PipelineRun` instead. Set it to "both" to 
+  do both. For mroe information, see [Configuring usage of `TaskRun` and `Run` embedded statuses](pipelineruns.md#configuring-usage-of-taskrun-and-run-embedded-statuses)
```
lbernick, line 396: typo: mroe -> more

Also I would just add a quick note that this isn't implemented yet, which you can remove in the follow-up PR.

pkg/apis/config/testdata/feature-flags-invalid-embedded-status.yaml
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: feature-flags
+  namespace: tekton-pipelines
+data:
+  enable-api-fields: "im-not-a-valid-feature-gate"
```
lbernick, line 21: I think this should be `embedded-status`

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -396,6 +398,23 @@ func (pr *PipelineRunStatus) MarkRunning(reason, messageFormat string, messageA
...
+// ChildStatusReference is used to point to the statuses of individual TaskRuns and Runs within this PipelineRun.
+type ChildStatusReference struct {
+	runtime.TypeMeta `json:",inline"`
+	// Name is the name of the TaskRun or Run this is referencing.
+	Name string `json:"name,omitempty"`
+	// PipelineTaskName is the name of the PipelineTask this is referencing.
+	PipelineTaskName string `json:"pipelineTaskName,omitempty"`
+
+	// ConditionChecks is the the list of condition checks, including their names and statuses, for the PipelineTask.
+	// Deprecated: This field will be removed when conditions are removed for v1.
```
lbernick, line 410: nit: conditions may be removed in v1beta1 as well according to our compatibility policy so I'd just remove the "v1" bit

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -408,10 +427,12 @@ type PipelineRunStatusFields struct {
 	// +optional
 	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
 
+	// Deprecated - use ChildReferences instead.
```
lbernick, line 430: optional comment: I wonder if it's possible to add an annotation that will allow these fields to display as deprecated in code editors? I did a quick search and see `// +k8s:prerelease-lifecycle-gen:deprecated=1.20` for some stuff in our k8s dependencies so it's probably possible

test/ignore_test.go.ignore
```diff
@@ -0,0 +1,122 @@
...
+		Status: v1alpha1.PipelineRunStatus{Status: duckv1beta1.Status{
+			Conditions: []apis.Condition{{Type: apis.ConditionSucceeded, Status: corev1.ConditionTrue}},
+		}},
+	}
+	ty, err := yaml.Marshal(helloworldResource)
+	require.NoError(t, err)
+	fmt.Printf("yaml:\n%s\n", string(ty))
+}
+
+func TestCompBetaAlpha(t *testing.T) {
```
lbernick, line 45: I'm a bit confused about what this test is for; this almost looks like a test for pipelineresources? why is it called ignore_test? I'm not sure the changes you're introducing need a new e2e test.

lbernick (commented):
Thanks Andrew!

pkg/apis/config/testdata/feature-flags-invalid-embedded-status.yaml
abayer: I'm ashamed of how many YAML files I messed up like this. =)

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
abayer: 👍

test/ignore_test.go.ignore
abayer: Hahaha, this is detritus from who knows when that was still sitting in my local clone and somehow inadvertently got added. Nuking it now!

vdemeester (approved):

jerop (approved):
thank you @abayer!

Issue #4543, opened 2022-02-02
Pipelines don't respect mixed Task API versions


wlynch:
So a weird quirk of the Pipelines API is that it's valid to create a request like:

```yaml
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  generateName: pr-
spec:
  pipelineSpec:
    tasks:
      - name: a
        taskSpec:
          apiVersion: tekton.dev/v1alpha1
          steps:
            - image: alpine
              script: |
                echo "hi!"
      - name: b
        taskSpec:
          steps:
            - image: alpine
              script: |
                echo "hello!"
```

# Expected Behavior

I'm not sure what the behavior should be. I see 2 options:

1. Respect the TypeMeta field and create resources corresponding to the type (e.g. allow a v1beta1 Pipeline to create a v1alpha1 Task) - this might be useful to allow for users to use experimental Tasks within a Pipeline.
2. Enforce that v1beta1 Pipelines must use v1beta1 Tasks. 

# Actual Behavior

The request is accepted, the TypeMeta field is ignored, v1beta1 Tasks are always used. (didn't check what the behavior is for TaskRefs).

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
    Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:18:45Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
  Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:01:33Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

Client version: 0.15.0
Pipeline version: v0.32.0

vdemeester:
[...]

PR #4706, opened 2022-03-23
[TEP-0100] Implementation for embedded TaskRun and Run statuses in PipelineRuns

abayer:
# Changes

See:
* https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
* https://github.com/tektoncd/pipeline/pull/4705
* https://github.com/tektoncd/pipeline/issues/3140

This implements TEP-0100, building on top of the flags/fields/docs changes in #4705.

NOTE: This is on hold until #4728 is merged, so that this PR with the actual implementation is of a more manageable size.

/kind tep
/kind api-change

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Added implementation for minimal `TaskRun` and `Run` statuses within `PipelineRun` statuses.
```

### DIFF
--- a/pkg/reconciler/pipelinerun/cancel_test.go
+++ b/pkg/reconciler/pipelinerun/cancel_test.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"testing"
 
+	"k8s.io/apimachinery/pkg/runtime"
+
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	_ "github.com/tektoncd/pipeline/pkg/pipelinerunmetrics/fake" // Make sure the pipelinerunmetrics are setup
@@ -36,6 +38,7 @@ func TestCancelPipelineRun(t *testing.T) {
 		pipelineRun *v1beta1.PipelineRun
 		taskRuns    []*v1beta1.TaskRun
 		runs        []*v1alpha1.Run
+		wantErr     bool
 	}{{
 		name: "no-resolved-taskrun",
 		pipelineRun: &v1beta1.PipelineRun{
@@ -104,10 +107,67 @@ func TestCancelPipelineRun(t *testing.T) {
 				Status: v1beta1.PipelineRunSpecStatusCancelledDeprecated,
 			},
 		},
+	}, {
+		name: "child-references",
+		pipelineRun: &v1beta1.PipelineRun{
+			ObjectMeta: metav1.ObjectMeta{Name: "test-pipeline-run-cancelled"},
+			Spec: v1beta1.PipelineRunSpec{
+				Status: v1beta1.PipelineRunSpecStatusCancelled,
+			},
+			Status: v1beta1.PipelineRunStatus{PipelineRunStatusFields: v1beta1.PipelineRunStatusFields{
+				ChildReferences: []v1beta1.ChildStatusReference{
+					{
+						TypeMeta:         runtime.TypeMeta{Kind: "TaskRun"},
+						Name:             "t1",
+						PipelineTaskName: "task-1",
+					},
+					{
+						TypeMeta:         runtime.TypeMeta{Kind: "TaskRun"},
+						Name:             "t2",
+						PipelineTaskName: "task-2",
+					},
+					{
+						TypeMeta:         runtime.TypeMeta{Kind: "Run"},
+						Name:             "r1",
+						PipelineTaskName: "run-1",
+					},
+					{
+						TypeMeta:         runtime.TypeMeta{Kind: "Run"},
+						Name:             "r2",
+						PipelineTaskName: "run-2",
+					},
+				},
+			}},
+		},
+		taskRuns: []*v1beta1.TaskRun{
+			{ObjectMeta: metav1.ObjectMeta{Name: "t1"}},
+			{ObjectMeta: metav1.ObjectMeta{Name: "t2"}},
+		},
+		runs: []*v1alpha1.Run{
+			{ObjectMeta: metav1.ObjectMeta{Name: "r1"}},
+			{ObjectMeta: metav1.ObjectMeta{Name: "r2"}},
+		},
+	}, {
+		name: "unknown-kind-on-child-references",
+		pipelineRun: &v1beta1.PipelineRun{
+			ObjectMeta: metav1.ObjectMeta{Name: "test-pipeline-run-cancelled"},
+			Spec: v1beta1.PipelineRunSpec{
+				Status: v1beta1.PipelineRunSpecStatusCancelled,
+			},
+			Status: v1beta1.PipelineRunStatus{PipelineRunStatusFields: v1beta1.PipelineRunStatusFields{
+				ChildReferences: []v1beta1.ChildStatusReference{{
+					TypeMeta:         runtime.TypeMeta{Kind: "InvalidKind"},
+					Name:             "t1",
+					PipelineTaskName: "task-1",
+				}},
+			}},
+		},
+		wantErr: true,
 	}}
 	for _, tc := range testCases {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
+
 			d := test.Data{
 				PipelineRuns: []*v1beta1.PipelineRun{tc.pipelineRun},
 				TaskRuns:     tc.taskRuns,
@@ -117,33 +177,41 @@ func TestCancelPipelineRun(t *testing.T) {
 			ctx, cancel := context.WithCancel(ctx)
 			defer cancel()
 			c, _ := test.SeedTestData(t, ctx, d)
-			if err := cancelPipelineRun(ctx, logtesting.TestLogger(t), tc.pipelineRun, c.Pipeline); err != nil {
-				t.Fatal(err)
-			}
-			// This PipelineRun should still be complete and false, and the status should reflect that
-			cond := tc.pipelineRun.Status.GetCondition(apis.ConditionSucceeded)
-			if cond.IsTrue() {
-				t.Errorf("Expected PipelineRun status to be complete and false, but was %v", cond)
-			}
-			if tc.taskRuns != nil {
-				l, err := c.Pipeline.TektonV1beta1().TaskRuns("").List(ctx, metav1.ListOptions{})
+
+			err := cancelPipelineRun(ctx, logtesting.TestLogger(t), tc.pipelineRun, c.Pipeline)
+			if tc.wantErr {
+				if err == nil {
+					t.Error("expected an error, but did not get one")
+				}
+			} else {
 				if err != nil {
 					t.Fatal(err)
 				}
-				for _, tr := range l.Items {
-					if tr.Spec.Status != v1beta1.TaskRunSpecStatusCancelled {
-						t.Errorf("expected task %q to be marked as cancelled, was %q", tr.Name, tr.Spec.Status)
+				// This PipelineRun should still be complete and false, and the status should reflect that
+				cond := tc.pipelineRun.Status.GetCondition(apis.ConditionSucceeded)
+				if cond.IsTrue() {
+					t.Errorf("Expected PipelineRun status to be complete and false, but was %v", cond)
+				}
+				if tc.taskRuns != nil {
+					for _, expectedTR := range tc.taskRuns {
+						tr, err := c.Pipeline.TektonV1beta1().TaskRuns("").Get(ctx, expectedTR.Name, metav1.GetOptions{})
+						if err != nil {
+							t.Fatalf("couldn't get expected TaskRun %s, got error %s", expectedTR.Name, err)
+						}
+						if tr.Spec.Status != v1beta1.TaskRunSpecStatusCancelled {
+							t.Errorf("expected task %q to be marked as cancelled, was %q", tr.Name, tr.Spec.Status)
+						}
 					}
 				}
-			}
-			if tc.runs != nil {
-				l, err := c.Pipeline.TektonV1alpha1().Runs("").List(ctx, metav1.ListOptions{})
-				if err != nil {
-					t.Fatal(err)
-				}
-				for _, r := range l.Items {
-					if r.Spec.Status != v1alpha1.RunSpecStatusCancelled {
-						t.Errorf("expected Run %q to be marked as cancelled, was %q", r.Name, r.Spec.Status)
+				if tc.runs != nil {
+					for _, expectedRun := range tc.runs {
+						r, err := c.Pipeline.TektonV1alpha1().Runs("").Get(ctx, expectedRun.Name, metav1.GetOptions{})
+						if err != nil {
+							t.Fatalf("couldn't get expected Run %s, got error %s", expectedRun.Name, err)
+						}
+						if r.Spec.Status != v1alpha1.RunSpecStatusCancelled {
+							t.Errorf("expected task %q to be marked as cancelled, was %q", r.Name, r.Spec.Status)
+						}
 					}
 				}
 			}
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -1198,112 +1198,97 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
 		logger.Errorf("could not list TaskRuns %#v", err)
 		return err
 	}
-	updatePipelineRunStatusFromTaskRuns(ctx, logger, pr, taskRuns)
-
 	runs, err := c.runLister.Runs(pr.Namespace).List(k8slabels.SelectorFromSet(pipelineRunLabels))
 	if err != nil {
 		logger.Errorf("could not list Runs %#v", err)
 		return err
 	}
-	updatePipelineRunStatusFromRuns(ctx, logger, pr, runs)
+
+	c.updatePipelineRunStatusFromSlices(ctx, logger, pr, taskRuns, runs)
 
 	return nil
 }
 
-// updatePipelineRunStatusFromTaskRuns takes a PipelineRun and list of TaskRuns within that PipelineRun, and updates the
-// PipelineRun's .Status.TaskRuns and/or .Status.ChildReferences, depending on the value of the "embedded-status" feature
-// flag, with information about each of the given TaskRuns, including ensuring that any (deprecated) condition checks are
-// represented.
-func updatePipelineRunStatusFromTaskRuns(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, trs []*v1beta1.TaskRun) {
-	// If no TaskRun was found, nothing to be done. We never remove taskruns from the status
-	if len(trs) == 0 {
-		return
-	}
+func (c *Reconciler) updatePipelineRunStatusFromSlices(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, taskRuns []*v1beta1.TaskRun, runs []*v1alpha1.Run) {
+	taskNameByPipelineTask := make(map[string]string)
 
 	cfg := config.FromContextOrDefaults(ctx)
 	fullEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.FullEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
 	minimalEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.MinimalEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
 
+	if minimalEmbedded {
+		taskNameByPipelineTask = updatePipelineRunStatusFromChildRefs(logger, pr, taskRuns, runs)
+	}
+	if fullEmbedded {
+		updatePipelineRunStatusFromTaskRuns(logger, pr, taskRuns, taskNameByPipelineTask)
+		updatePipelineRunStatusFromRuns(logger, pr, runs)
+	}
+}
+
+// updatePipelineRunStatusFromTaskRuns takes a PipelineRun, a list of TaskRuns within that PipelineRun, and a map of
+// existing PipelineTask names to TaskRun (or Run) names, and updates the PipelineRun's .Status.TaskRuns, including
+// ensuring that any (deprecated) condition checks are represented.
+func updatePipelineRunStatusFromTaskRuns(logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, trs []*v1beta1.TaskRun, taskRunNameByPipelineTask map[string]string) {
+	// If no TaskRun was found, nothing to be done. We never remove taskruns from the status
+	if len(trs) == 0 {
+		return
+	}
+
+	if taskRunNameByPipelineTask == nil {
+		taskRunNameByPipelineTask = make(map[string]string)
+	}
+
 	// Store a list of Condition TaskRuns for each PipelineTask (by name)
 	conditionTaskRuns := make(map[string][]*v1beta1.TaskRun)
-	// Map PipelineTask names to TaskRun names that were already in the status
-	taskRunByPipelineTask := make(map[string]string)
-	// Map PipelineTask names to TaskRun child references that were already in the status
-	childRefByPipelineTask := make(map[string]*v1beta1.ChildStatusReference)
 
-	if fullEmbedded {
-		if pr.Status.TaskRuns != nil {
-			for taskRunName, pipelineRunTaskRunStatus := range pr.Status.TaskRuns {
-				taskRunByPipelineTask[pipelineRunTaskRunStatus.PipelineTaskName] = taskRunName
-			}
-		} else {
-			pr.Status.TaskRuns = make(map[string]*v1beta1.PipelineRunTaskRunStatus)
+	if pr.Status.TaskRuns != nil {
+		for taskRunName, pipelineRunTaskRunStatus := range pr.Status.TaskRuns {
+			taskRunNameByPipelineTask[pipelineRunTaskRunStatus.PipelineTaskName] = taskRunName
 		}
-	}
-	if minimalEmbedded {
-		for i := range pr.Status.ChildReferences {
-			childRefByPipelineTask[pr.Status.ChildReferences[i].PipelineTaskName] = &pr.Status.ChildReferences[i]
-		}
+	} else {
+		pr.Status.TaskRuns = make(map[string]*v1beta1.PipelineRunTaskRunStatus)
 	}
 
 	// Loop over all the TaskRuns associated to Tasks
-	for _, taskrun := range trs {
+	for _, tr := range trs {
 		// Only process TaskRuns that are owned by this PipelineRun.
 		// This skips TaskRuns that are indirectly created by the PipelineRun (e.g. by custom tasks).
-		if len(taskrun.OwnerReferences) < 1 || taskrun.OwnerReferences[0].UID != pr.ObjectMeta.UID {
-			logger.Debugf("Found a TaskRun %s that is not owned by this PipelineRun", taskrun.Name)
+		if len(tr.OwnerReferences) < 1 || tr.OwnerReferences[0].UID != pr.ObjectMeta.UID {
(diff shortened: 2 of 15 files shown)


pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -104,46 +110,189 @@ func TestCancelPipelineRun(t *testing.T) {
...
+			},
+			Status: v1beta1.PipelineRunStatus{PipelineRunStatusFields: v1beta1.PipelineRunStatusFields{
+				ChildReferences: []v1beta1.ChildStatusReference{{
+					TypeMeta:         runtime.TypeMeta{Kind: "InvalidKind"},
+					Name:             "t1",
+					PipelineTaskName: "task-1",
+				}},
+			}},
+		},
+		expectedErr: errors.New("error(s) from cancelling TaskRun(s) from PipelineRun test-pipeline-run-cancelled: unknown or unsupported kind `InvalidKind` for cancellation"),
```
lbernick, line 185: any way to avoid asserting the exact error message? either a regex or just a boolean wantErr

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -104,46 +110,189 @@ func TestCancelPipelineRun(t *testing.T) {
...
-			if cond.IsTrue() {
-				t.Errorf("Expected PipelineRun status to be complete and false, but was %v", cond)
-			}
-			if tc.taskRuns != nil {
-				l, err := c.Pipeline.TektonV1beta1().TaskRuns("").List(ctx, metav1.ListOptions{})
+
+			err := cancelPipelineRun(ctx, logtesting.TestLogger(t), tc.pipelineRun, c.Pipeline)
+			if tc.expectedErr != nil {
+				if err == nil {
+					t.Fatalf("expected error '%s', but did not get an error", tc.expectedErr)
```
lbernick, line 232: I'd use t.Errorf here and below

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -104,46 +110,189 @@ func TestCancelPipelineRun(t *testing.T) {
...
-				for _, tr := range l.Items {
-					if tr.Spec.Status != v1beta1.TaskRunSpecStatusCancelled {
-						t.Errorf("expected task %q to be marked as cancelled, was %q", tr.Name, tr.Spec.Status)
+				// This PipelineRun should still be complete and false, and the status should reflect that
+				cond := tc.pipelineRun.Status.GetCondition(apis.ConditionSucceeded)
+				if cond.IsTrue() {
+					t.Errorf("Expected PipelineRun status to be complete and false, but was %v", cond)
+				}
+				if tc.taskRuns != nil {
+					l, err := c.Pipeline.TektonV1beta1().TaskRuns("").List(ctx, metav1.ListOptions{
```
lbernick, line 247: listing by label selector is a bit roundabout, could you just query by the names of the ignored/not ignored runs?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -562,11 +562,22 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
...
+	taskRunStatuses := pipelineRunFacts.State.GetTaskRunsStatus(pr)
+	runStatuses := pipelineRunFacts.State.GetRunsStatus(pr)
+
+	if cfg.FeatureFlags.EmbeddedStatus == config.FullEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus {
+		pr.Status.TaskRuns = taskRunStatuses
+		pr.Status.Runs = runStatuses
+	}
+
+	if cfg.FeatureFlags.EmbeddedStatus == config.MinimalEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus {
+		pr.Status.ChildReferences = pipelineRunFacts.State.GetChildReferences(v1beta1.SchemeGroupVersion.String(),
```
lbernick, line 574: no AI for this PR specifically but #4543 is likely relevant here-- we need to clarify what versions of TaskRuns/Runs are run by a PipelineRun

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1230,25 +1262,70 @@ func updatePipelineRunStatusFromTaskRuns(logger *zap.SugaredLogger, pr *v1beta1.
...
-			taskRunName = resources.GetTaskRunName(pr.Status.TaskRuns, pipelineTaskName, pr.Name)
+			taskRunName = resources.GetTaskRunName(pr.Status.TaskRuns, pr.Status.ChildReferences, pipelineTaskName, pr.Name)
 			pr.Status.TaskRuns[taskRunName] = &v1beta1.PipelineRunTaskRunStatus{
 				PipelineTaskName: pipelineTaskName,
 				Status:           nil,
 				ConditionChecks:  nil,
 			}
 		}
+
+		if !ok && minimalEmbedded {
```
lbernick, line 1299: this can be `if minimalEmbedded` since `ok` here will always be false.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1230,25 +1262,70 @@ func updatePipelineRunStatusFromTaskRuns(logger *zap.SugaredLogger, pr *v1beta1.
...
+					Name:             taskRunName,
+					PipelineTaskName: pipelineTaskName,
+				}
+			}
+		}
+
 		// Build the map of condition checks for the taskrun
 		// If there were no other condition, initialise the map
-		conditionChecks := pr.Status.TaskRuns[taskRunName].ConditionChecks
+		conditionChecks := make(map[string]*v1beta1.PipelineRunConditionCheckStatus)
```
lbernick, line 1319: if someone has the feature flag set to "both", I think it would be better to treat the childReferences as a source of truth instead of getting condition checks from both statuses. Also, I think it would be worth pulling the condition check logic into a helper function.

test/featureflags.go
```diff
@@ -32,3 +32,20 @@ func requireAnyGate(gates map[string]string) func(context.Context, *testing.T, *
 		t.Skipf("No feature flag matching %s", strings.Join(pairs, " or "))
 	}
 }
+
+// getEmbeddedStatusValue gets the current value for the "embedded-status" feature flag.
+// If the flag is not set, it returns the default value. This is used in tests which look
```
lbernick, line 37: nit: better not to include in a function's docstring details of how the function is used.

test/v1alpha1/clients.go
```diff
@@ -97,3 +102,20 @@ func newClients(t *testing.T, configPath, clusterName, namespace string) *client
 	c.ConditionClient = cs.TektonV1alpha1().Conditions(namespace)
 	return c
 }
+
+// getEmbeddedStatusValue gets the current value for the "embedded-status" feature flag.
+// If the flag is not set, it returns the default value. This is used in tests which look
+// for something in the "TaskRuns", "Runs", or "ChildReferences" fields of a PipelineRun's
+// status to ensure that those tests check the right things depending on the value of
+// "embedded-status".
+func getEmbeddedStatusValue(ctx context.Context, t *testing.T, c *clients) string {
```
lbernick, line 111: this function is duplicated, can it be reused instead?

pkg/reconciler/pipelinerun/cancel_test.go
abayer: Probably - lemme create a followup to do this systematically, 'cos I've seen this sort of thing in a bunch of places.

pkg/reconciler/pipelinerun/cancel_test.go
abayer: Yup, those are reasonable changes, doing so.

abayer: Actually, check that, this one does make sense as `Fatalf`, since the next check includes `err.Error()`, which would nil pointer on us if `err` is nil. I'll change the second one, but I'm leaving this one.

pkg/reconciler/pipelinerun/cancel_test.go
abayer: I can't remember my reasoning for this approach off the top of my head, but I do remember that this felt like the clearest way to get what we needed.

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -104,46 +110,189 @@ func TestCancelPipelineRun(t *testing.T) {
...
+						Name:             "r1",
+						PipelineTaskName: "run-1",
+					},
+					{
+						TypeMeta:         runtime.TypeMeta{Kind: "Run"},
+						Name:             "r2",
+						PipelineTaskName: "run-2",
+					},
+				},
+				// If ChildReferences is not empty, TaskRuns and Runs should be ignored.
```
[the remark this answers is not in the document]
abayer: This test may not strictly be necessary - it's a scenario that I don't _think_ we can ever actually encounter, since we have solid testing around how `pr.Status.TaskRuns`/`pr.Status.Runs` and `pr.Status.ChildReferences` get populated (depending on the `embedded-status` value) elsewhere. But I decided that I wanted to have testing of the fact that the cancel behavior ignores `pr.Status.TaskRuns` and `pr.Status.Runs` if `pr.Status.ChildReferences` isn't empty, just to be safe. =)

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: Agreed - I banged my head against a wall for a bit on that before deciding it wasn't worth dealing with here.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1171,34 +1186,50 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
+	// Map PipelineTask names to TaskRun child references that were already in the status
+	childRefByPipelineTask := make(map[string]*v1beta1.ChildStatusReference)
+
+	if fullEmbedded {
+		if pr.Status.TaskRuns != nil {
+			for taskRunName, pipelineRunTaskRunStatus := range pr.Status.TaskRuns {
+				taskRunByPipelineTask[pipelineRunTaskRunStatus.PipelineTaskName] = taskRunName
+			}
+		} else {
+			pr.Status.TaskRuns = make(map[string]*v1beta1.PipelineRunTaskRunStatus)
```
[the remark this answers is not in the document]
abayer: I think it is - if I remember correctly (the logic here is a bit of a mess and I can't remember all the details off the top of my head!), `pr.Status.TaskRuns` does start out nil, but everything below here expects it to be a created map, so in the cases where `embedded-status` is either `full` or `both`, we need to initialize `pr.Status.TaskRuns` so it can be properly interacted with later.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1171,34 +1186,50 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
+		}
+	}
+	if minimalEmbedded {
+		for i := range pr.Status.ChildReferences {
+			childRefByPipelineTask[pr.Status.ChildReferences[i].PipelineTaskName] = &pr.Status.ChildReferences[i]
```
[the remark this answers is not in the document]
abayer: Yup, you're right - good catch! Thanks! =)

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: I'm not sure - I know I got bit by not checking `ok` as well somewhere, but looking at the code, I think this is probably not the case. I'll change it and see if anything blows up. =)

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: Good point - I think I may not be consistent on which of `ChildReferences` or `TaskRuns`/`Runs` I use as source of truth if both are populated. I definitely use `ChildReferences` in `cancel.go`, and using the forward-looking one as the source of truth does seem better. I'll rework that, and I'll see if I can tease out the condition check stuff.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1171,34 +1186,50 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
-	updatePipelineRunStatusFromRuns(logger, pr, runs)
+	updatePipelineRunStatusFromRuns(ctx, logger, pr, runs)
 
 	return nil
 }
 
-func updatePipelineRunStatusFromTaskRuns(logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, trs []*v1beta1.TaskRun) {
+func updatePipelineRunStatusFromTaskRuns(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, trs []*v1beta1.TaskRun) {
 	// If no TaskRun was found, nothing to be done. We never remove taskruns from the status
 	if trs == nil || len(trs) == 0 {
```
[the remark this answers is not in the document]
abayer: d'oh! I cargo-culted poorly!

test/v1alpha1/clients.go
abayer: Annoyingly? No. Because `test.v1alpha1.clients` is a different struct than `test.clients`. Though I could change that to pass `kubeClient kubernetes.Interface` instead of `c *clients` - I'll do that.

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
lbernick: I don't think we should be treating nil maps differently from empty maps-- if that's the case that logic should be changed instead of initializing the variable to an empty map.

pkg/reconciler/pipelinerun/pipelinerun.go
lbernick: ah wait my bad I was thinking in the case where you keep the `if ok := ...` then the scope of `ok` would be the if block, but that's not the case here. But I still think this logic could use some changes-- if you have flag = minimal, there's no reason to check the status.TaskRuns, and if you have flag = both, childRefs should be used as the source of truth.

pkg/reconciler/pipelinerun/cancel_test.go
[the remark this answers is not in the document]
lbernick: I also don't think we can ever encounter it, but if it does occur it would be better to alert us via an error rather than silently succeeding. I agree this case should be tested.

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
lbernick: nope, this was already there :)

test/v1alpha1/clients.go
lbernick: that sounds much better!

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: The logic is right, I think, with the `!ok` removed - we still need to be adding to both `TaskRuns` and `ChildReferences` for `both`. But we should just be generating the `taskRunName` once.

abayer: I'm simplifying it, though. =)

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
abayer: ha!

test/featureflags.go
abayer: Okiedokie, cleaning it up!

abayer:
@lbernick - updated to address your comments - thanks!

pkg/reconciler/pipelinerun/cancel_test.go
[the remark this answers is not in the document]
abayer: Ok, I literally can't find a way we ever could encounter this situation, and if we did, it would be cleaned up on reconcile anyway. I feel like my thinking was "what happens with a running PR that somehow gets borked when the feature flag value changes from `full` to `minimal`?" But the situation I put in this test wouldn't result from that case, so the "ignore `.status.taskRuns` and `.status.runs` if `.status.childReferences` is populated, then make sure that we don't do anything to whatever's in `.status.taskRuns` or `.status.runs`" scenario tested here is silly. I'll clean up the test.

pkg/reconciler/pipelinerun/cancel_test.go
lbernick: I'd swap to a bool wantErr (see comment above), which allows you to simplify:
```
if tc.wantErr != (err != nil) {
  t.Errorf("expected err: %s, but got err %s")
}
```

pkg/reconciler/pipelinerun/cancel_test.go
lbernick: Is this approach used elsewhere in the repo? Do you think querying by name could work?

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
lbernick: In that case, maybe it is better to create a new function `updatePipelineRunStatusFromChildRefs` and call that function when the feature flag is set to "minimal" or "both".

test/v1alpha1/retry_test.go
```diff
@@ -43,6 +47,12 @@ func TestTaskRunRetry(t *testing.T) {
 	knativetest.CleanupOnInterrupt(func() { tearDown(ctx, t, c, namespace) }, t.Logf)
 	defer tearDown(ctx, t, c, namespace)
 
+	// Skip if the "embedded-status" feature flag has any value but "full".
```
lbernick, line 50: what's the reason for this?

test/retry_test.go
```diff
@@ -74,12 +78,31 @@ spec:
...
+		// PipelineRunStatus should have 1 TaskRun status, and it should be failed.
+		if len(pr.Status.TaskRuns) != 1 {
+			t.Errorf("Got %d TaskRun statuses, wanted %d", len(pr.Status.TaskRuns), numRetries)
+		}
+		for taskRunName, trs := range pr.Status.TaskRuns {
+			if !isFailed(t, taskRunName, trs.Status.Conditions) {
+				t.Errorf("TaskRun status %q is not failed", taskRunName)
+			}
+		}
+	} else {
```
lbernick, line 91: I think this should be `if embeddedStatusValue != config.FullEmbeddedStatus` so that you're testing that the correct value is present for `both`. also "embeddedstatusvalue" is a bit redundant-- could just be "embeddedstatus".

pkg/reconciler/pipelinerun/cancel_test.go
abayer: I did not, I will do that after a call, and will update here as well.

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
abayer: I'm pushing an update momentarily, but plan to come back to this later this afternoon and see if I can come up with a good way to rewrite the whole `updatePipelineRunStatusFromInformer` to make more sense with `ChildReferences` involved now.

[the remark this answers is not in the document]
abayer: So...I'm not sure if there's a better way to rework this (beyond removing a return I had in at one point but stopped using, which I'm pushing momentarily) - it's going to be messy no matter what, because we're doing very similar things for both the `full` and `minimal` cases, and for `both`, we obviously have to do both. I think I prefer splitting it up into "do the stuff relating to `TaskRun`s" and "do the stuff relating to `Run`s", rather than "do the stuff for `full`/`both`" and "do the stuff for `minimal`/`both`". But I could be wrong, so I'm going to dwell on this a bit longer.

pkg/reconciler/pipelinerun/cancel_test.go
abayer: I forgot again to open the issue! Doing so right now.

abayer:
ping @lbernick and @jerop - just want to remind you that this is (hopefully!) ready for final review. =)

test/v1alpha1/retry_test.go
lbernick: sounds good! did you have a chance to update this test?

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -117,33 +177,41 @@ func TestCancelPipelineRun(t *testing.T) {
...
-				l, err := c.Pipeline.TektonV1beta1().TaskRuns("").List(ctx, metav1.ListOptions{})
+
+			err := cancelPipelineRun(ctx, logtesting.TestLogger(t), tc.pipelineRun, c.Pipeline)
+			if tc.wantErr {
+				if err == nil {
+					t.Error("expected an error, but did not get one")
+				}
+			} else {
 				if err != nil {
 					t.Fatal(err)
```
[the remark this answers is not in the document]
abayer: I think `Fatal` is right here - if we don't expect an error but do hit an error, we already know something's broken and will probably get confusing results from anything below here.

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -117,33 +177,41 @@ func TestCancelPipelineRun(t *testing.T) {
...
+				cond := tc.pipelineRun.Status.GetCondition(apis.ConditionSucceeded)
+				if cond.IsTrue() {
+					t.Errorf("Expected PipelineRun status to be complete and false, but was %v", cond)
 				}
-			}
-			if tc.runs != nil {
-				l, err := c.Pipeline.TektonV1alpha1().Runs("").List(ctx, metav1.ListOptions{})
-				if err != nil {
-					t.Fatal(err)
+				if tc.taskRuns != nil {
```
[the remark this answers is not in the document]
abayer: No, the indentation's right - we're only doing that testing when `wantErr` is false.

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
abayer: I'll need to dig to make sure changing that makes sense.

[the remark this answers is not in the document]
abayer: Ok, yeah, I think initializing the map makes sense in this situation. If we've gotten this far (i.e., `len(trs) > 0` or we would have returned out right away at the top of the function, and `fullEmbedded` is true, meaning `embedded-status` is either `full` or `minimal`, and nothing has initialized `pr.Status.TaskRuns` yet, because it's `nil`), we know that there are `TaskRun`s to process, we know that we will be populating them into `pr.Status.TaskRuns`. So the map will have to be initialized at some point for sure, and it's cleaner to do it here than in the loop, I feel.

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
abayer: I made a run at it, but it just ended up messier than what's already there. No matter what we do, this logic is going to be messy.

test/v1alpha1/retry_test.go
abayer: Yup, it's updated and now matches `test/retry_test.go`.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -581,11 +581,26 @@ func getConditionCheckName(taskRunStatus map[string]*v1beta1.PipelineRunTaskRunS
 			}
 		}
 	}
+	for _, cr := range childRefs {
```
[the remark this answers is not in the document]
abayer: Good catch, this needs to be flipped. Doing so now.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1228,57 +1260,145 @@ func updatePipelineRunStatusFromTaskRuns(logger *zap.SugaredLogger, pr *v1beta1.
...
-					}
-				} else {
-					// The condition name label is missing, so we cannot recover this
-					logger.Warnf("found an orphaned condition taskrun %#v with missing %s label",
-						foundTaskRun, pipeline.ConditionNameKey)
+		pr.Status.ChildReferences = newChildRefs
+	}
+}
+
+func updateConditionChecksForTaskRun(logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, childRefByPipelineTask map[string]*v1beta1.ChildStatusReference,
```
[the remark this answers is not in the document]
abayer: I'll add a comment, but again, when I tried rewriting this to tease apart `TaskRuns` and `ChildReferences` usages completely, it was faaaaar messier than what's there now.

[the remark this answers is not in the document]
abayer: ...which is not to say it's impossible to make it less messy, of course, just that I can't seem to pull that off. I can try again later, and almost certainly will, but I think that if the messiness of these functions ends up as the last thing holding this PR up, this may be a case where we need to plug our nose and accept the smell.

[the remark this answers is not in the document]
abayer: and fwiw, the messiness feels to me to largely be down to the fact that `pr.Status.TaskRuns` and `pr.Status.Runs` are maps, while `pr.Status.ChildReferences` is an array. So while updating `pr.Status.TaskRuns` and `pr.Status.Runs` is just a matter of "oh, this key isn't there yet? Let's add it! Oh, it is? Then update its value!" without having to worry about anything else already in the map, updating `pr.Status.ChildReferences` means we need to make sure that everything already in the array is still in the array when we're done, and we have to make changes to the elements in the array at different points in the process. That's why I added the `childRefByPipelineTask` temporary map - so we can keep all the existing references, add new ones, and make changes to references without needing to do "ok, iterate over `pr.Status.ChildReferences` until we find one with the right `PipelineTaskName`, now make changes to it, etc".

That last bit would admittedly be simpler if `pr.Status.ChildReferences` was a `[]*ChildStatusReference` rather than a `[]ChildStatusReference` as it is now, but it'd still be a mess.

It's definitely better from a Kubernetes API perspective to have an array instead of a map, but the downside is that updating individual elements in an array is always going to be uglier than updating individual elements in a map.

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
abayer: Ok, I got hooked on trying to figure out a less messy way and am maybe making some progress, but I forgot another thing that makes it more complicated (besides what I mentioned in https://github.com/tektoncd/pipeline/pull/4706#discussion_r839942364) - we have to generate the `TaskRun`/`Run` name for new ones, and if we're handling updating `TaskRuns`/`Runs` separately from updating `ChildReferences`, then when we've got `embedded-status='both'`, we need to make sure that we only generate a new name once. So, with `ChildReferences` as our first choice for source of truth, when we go to update `TaskRuns` after updating `ChildReferences`, we have to use the new `TaskRun` name we generated when updating `ChildReferences` again in `TaskRuns`.

I've made real progress in significantly simplifying the condition check handling part, so if I can nail the right way to do the `TaskRun` name generation once-and-only-once, I may be able to refactor this. It'll still be fairly ugly and have a good amount of duplicated code, but I think the current mess is the only way to avoid the duplicated code.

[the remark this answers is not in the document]
abayer: @lbernick Ok! I think I may have a better approach - see https://github.com/tektoncd/pipeline/pull/4706/commits/18521d4dd6e5802d56e6687bcc8ad4c2b6071ac0. It's not perfect yet - needs some cleanup/comments/test changes (specifically tests for the new `updatePipelineRunStatusFromSlices`, `getNewConditionChecksForTaskRun`,  and `updatePipelineRunStatusFromChildRefs`, a test for `updatePipelineRunStatusFromRuns`, and changes to the test for `updatePipelineRunStatusFromTaskRuns`), but it does decouple and simplify things. I'd love your thoughts!

Issue #4711, opened 2022-03-28
`git-init` should specify to run as root when it needs it.


mattmoor:
# Expected Behavior

A default nonroot version of `git-init` can pass e2e testing without issues.

# Actual Behavior

```
Error running git [init /workspace/sourcerepo]: exit status 1
/workspace/sourcerepo/.git: Permission denied
```

# Steps to Reproduce the Problem

I've been experimenting with an `apko` based replacement for https://github.com/tektoncd/pipeline/blob/main/images/git-init/Dockerfile

You can see it here: https://github.com/mattmoor/git-init/blob/main/.apko.yaml

Basically I replaced this base image with `ghcr.io/mattmoor/git-init`: https://github.com/tektoncd/pipeline/blob/30f4051a70e67373d042471c0cd5f546bc5a93b3/.ko.yaml#L4

At which point I see the output above when I `kubectl create` the following task:

```yaml
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  generateName: build-push-run-output-image-
spec:
  serviceAccountName: ""
  taskSpec:
    resources:
      inputs:
      - name: sourcerepo
        type: git
      outputs:
      - name: builtImage
        type: image
        targetPath: /workspace/sourcerepo
    steps:
    - name: build-and-push
      image: busybox
      script: |
        set -e
        cat <<EOF > $(inputs.resources.sourcerepo.path)/index.json
        {
        "schemaVersion": 2,
        "manifests": [
            {
            "mediaType": "application/vnd.oci.image.index.v1+json",
            "size": 314,
            "digest": "sha256:05f95b26ed10668b7183c1e2da98610e91372fa9f510046d4ce5812addad86b5"
            }
        ]
        }
    - name: echo
      image: busybox
      script: cat $(inputs.resources.sourcerepo.path)/index.json
  resources:
    inputs:
    - name: sourcerepo
      resourceSpec:
        type: git
        params:
        - name: revision
          value: v0.32.0
        - name: url
          value: https://github.com/GoogleContainerTools/skaffold
    outputs:
    - name: builtImage
      resourceSpec:
        type: image
        params:
        - name: url
          value: gcr.io/foo/bar
```

sbwsg:
The patch at https://github.com/tektoncd/pipeline/pull/4717 seems to work with this custom `git-init` image.  It just adds an explicit `runAsUser` to the git pipelineresource container.

mattmoor:
[...]

PR #4717, opened 2022-03-29
Git PipelineResource requires root

sbwsg:
# Changes

This PR makes explicit an assumption that the Git PipelineResource makes about how it's run. When a user replaces the git-init image with their own it's possible that the default UID of the container is not 0. Git PipelineResources expect to run as root and when they don't a confusing error is emitted (Permission denied while running git).

This commit updates the git pipelineresource to explicitly set its container `runAsUser` to `0`.

Fixes https://github.com/tektoncd/pipeline/issues/4711

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
The git pipelineresource is updated to explicitly set its `runAsUser` to 0 since it requires root.
```

### DIFF
--- a/pkg/apis/resource/v1alpha1/git/git_resource.go
+++ b/pkg/apis/resource/v1alpha1/git/git_resource.go
@@ -26,6 +26,7 @@ import (
 	resource "github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/names"
 	corev1 "k8s.io/api/core/v1"
+	"knative.dev/pkg/ptr"
 )
 
 var (
@@ -195,6 +196,10 @@ func (s *Resource) GetInputTaskModifier(_ *v1beta1.TaskSpec, path string) (v1bet
 			WorkingDir: pipeline.WorkspaceDir,
 			// This is used to populate the ResourceResult status.
 			Env: env,
+			SecurityContext: &corev1.SecurityContext{
+				// The git pipeline resource only works when running as root.
+				RunAsUser: ptr.Int64(0),
+			},
 		},
 	}
 
--- a/pkg/apis/resource/v1alpha1/git/git_resource_test.go
+++ b/pkg/apis/resource/v1alpha1/git/git_resource_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/ptr"
 
 	"github.com/google/go-cmp/cmp"
 	pipeline "github.com/tektoncd/pipeline/pkg/apis/pipeline"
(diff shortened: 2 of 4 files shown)


sbwsg:
Open to different approaches here.  I figured since PipelineResources have been declared as deprecated this is probably the best route just to ensure the fact they require root is recorded in code.

afrittoli:
> # Changes
> This PR makes explicit an assumption that the Git PipelineResource makes about how it's run. When a user replaces the git-init image with their own it's possible that the default UID of the container is not 0. Git PipelineResources expect to run as root and when they don't a confusing error is emitted (Permission denied while running git).
> 
> This commit updates the git pipelineresource to explicitly set its container `runAsUser` to `0`.
> 
> Fixes #4711
> 
> # Submitter Checklist
> 
> * [x]  Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
> * [x]  Release notes block below has been filled in or deleted (only if no user facing changes)
> 
> # Release Notes
> ```
> The git pipelineresource is updated to explicitly set its `runAsUser` to 0 since it requires root.
> ```

sounds good to me!

pkg/apis/resource/v1alpha1/git/git_resource.go
```diff
@@ -195,6 +197,9 @@ func (s *Resource) GetInputTaskModifier(_ *v1beta1.TaskSpec, path string) (v1bet
 			WorkingDir: pipeline.WorkspaceDir,
 			// This is used to populate the ResourceResult status.
 			Env: env,
+			SecurityContext: &corev1.SecurityContext{
+				RunAsUser: &runAsUser,
```
mattmoor, line 201: Mostly FYI, but we created this package because we were sick of declaring variables for these: https://github.com/knative/pkg/blob/0a1ec2e0d46c7e4ca0d0888510b1663d22a1bed7/ptr/ptr.go#L27-L31

mattmoor (commented):
Thanks for making this requirement explicit! 🎉 

/lgtm

mattmoor:
> Can/should we do a similar thing for the pullrequest-init image, and base its image on :nonroot instead

I think so, but it'd be great to not conflate it here.

sbwsg:
I updated the pull request pipeline resource to run with the explicit uid as well but will actually put that in a separate PR.

As for changing the default user in the pullrequest image I'm generally in favor. I'm not totally sure what the ultimate benefit would be given the feature's deprecation. Worst case perhaps a user who's depending on that image shows up in our issues to complain and we can gently migrate them away from depending on our internal images?

The pull-request catalog task pins to a specific version of that image so hopefully shouldn't be affected.

imjasonh:
> > Can/should we do a similar thing for the pullrequest-init image, and base its image on :nonroot instead
> 
> I think so, but it'd be great to not conflate it here.

Absolutely. I'll send another PR after this one.

dibyom (approved):

PR #4718, opened 2022-03-29
PullRequest PipelineResource expects root

sbwsg:
# Changes

When a user replaces the pullrequest-init image with their own it's possible that the default UID of the container is not 0.

More generally, we've never tested PipelineResources as anything but the root user. Rather than attempt to improve a feature that's now deprecated by adding testing and support for non-root use-cases it makes more sense to simply encode this expectation in to their containers until they're removed.

This commit updates the pullrequest pipelineresource to explicitly set its container `runAsUser` to `0`. It also updates the pullrequest-init image to use our default base of distroless' static:nonroot.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
The pullrequest PipelineResource is updated to explicitly set its runAsUser to 0. PipelineResources aren't tested as anything other than the root user and this change makes that explicit.

The pullrequest-init base image also no longer uses the root user by default. It now defaults to using UID 65532.
```

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -3,6 +3,3 @@ baseImageOverrides:
   # git-init uses a base image that includes Git, and supports running either
   # as root or as user nonroot with UID 65532.
   github.com/tektoncd/pipeline/cmd/git-init: gcr.io/tekton-nightly/github.com/tektoncd/pipeline/git-init-build-base:latest
-
-  # pullrequest-init supports running either as root or as user with UID 65532.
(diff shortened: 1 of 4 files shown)


sbwsg:
fyi @imjasonh @mattmoor this pr removes the rootful distroless base from pullrequest-init.

imjasonh (commented):
tekton/publish.yaml has another reference to remove.

imjasonh (commented):
/lgtm

vdemeester (approved):

PR #4728, opened 2022-03-31
[TEP-0100] Prepare for testing of minimal status implementation

abayer:
# Changes

This is for https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md

This change is to minimize the size of the actual implementation. We need to change a number of
tests in `pkg/reconciler/pipelinerun/pipelinerun_test.go` to be table-based, so that we can test
behavior under each possible value for the new `embedded-status` feature flag. Here, we just
modify the relevant tests to be table-based, with a test case for each value, but we don't actually
change any behavior. Right now, that means that these tests are purely duplicative - they're going
to run the same and check the same things for any value of `embedded-status`, but the implementation
PR will be much, much smaller, due to how many lines end up indented here.

/kind misc
/kind tep

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -100,10 +100,18 @@ var (
 
 	now       = time.Date(2022, time.January, 1, 0, 0, 0, 0, time.UTC)
 	testClock = clock.NewFakePassiveClock(now)
+
+	valuesForEmbeddedStatus = []string{
+		config.DefaultEmbeddedStatus,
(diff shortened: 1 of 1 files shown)


lbernick:
Discussed offline with Andrew-- he is going to try reworking this to parameterize tests/use helpers for checking status

/hold

@abayer just remove the hold when ready for re-review!

abayer:
Closing in favor of an alternative approach to modifying these tests - PR incoming.

PR #4734, opened 2022-04-04
[TEP-0100] Prepare for testing of minimal status implementation

abayer:
# Changes

This change is to minimize the size of the actual implementation. We need to change a number of
tests in `pkg/reconciler/pipelinerun/pipelinerun_test.go` to be table-based, so that we can test
behavior under each possible value for the new `embedded-status` feature flag. Here, we just
modify the relevant tests to be table-based and use a parameterized helper function for the actual
execution, with a test case for each value.

It also adds helper functions for checking the relevant fields in the `PipelineRun` status, which
are hard-coded to always handle the current, "full" embedded status approach.

This also splits out `TestUpdatePipelineRunStatusFromTaskRuns` and `TestUpdatePipelineRunStatusFromRuns`
into a separate file, `pipelinerun_updatestatus_test.go`. When the TEP-0100 implementation lands,
this will also contain additional tests for updating via child references. Splitting like this helps
keep `pipelinerun_test.go` from getting even more bloated than it is currently.

Until the implementation, these table-based tests are purely duplicative - they're going to run the
same and check the same things for any value of `embedded-status`, but the implementation PR will be
cleaner, only adding the implementation and changing the helper functions to take all of the
possible "embedded-status" values into account. The changes which need to be made in the
implementation PR are all marked with `// TODO(abayer): ...` in `pipelinerun_test.go` and
`pipelinerun_updatestatus_test.go`.

/kind tep

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"net/http/httptest"
 	"net/url"
-	"regexp"
 	"testing"
 	"time"
 
@@ -104,9 +103,10 @@ var (
 )
 
 const (
-	apiFieldsFeatureFlag   = "enable-api-fields"
-	customTasksFeatureFlag = "enable-custom-tasks"
-	ociBundlesFeatureFlag  = "enable-tekton-oci-bundles"
+	apiFieldsFeatureFlag      = "enable-api-fields"
+	customTasksFeatureFlag    = "enable-custom-tasks"
+	ociBundlesFeatureFlag     = "enable-tekton-oci-bundles"
+	embeddedStatusFeatureFlag = "embedded-status"
 )
 
 type PipelineRunTest struct {
@@ -240,6 +240,36 @@ func getPipelineRunUpdates(t *testing.T, actions []ktesting.Action) []*v1beta1.P
 }
 
 func TestReconcile(t *testing.T) {
+	testCases := []struct {
+		name              string
+		embeddedStatusVal string
+	}{
+		{
+			name:              "default embedded status",
+			embeddedStatusVal: config.DefaultEmbeddedStatus,
+		},
+		{
+			name:              "full embedded status",
+			embeddedStatusVal: config.FullEmbeddedStatus,
+		},
+		{
+			name:              "both embedded status",
+			embeddedStatusVal: config.BothEmbeddedStatus,
+		},
+		{
+			name:              "minimal embedded status",
+			embeddedStatusVal: config.MinimalEmbeddedStatus,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			runTestReconcileWithEmbeddedStatus(t, tc.embeddedStatusVal)
+		})
+	}
+}
+
+func runTestReconcileWithEmbeddedStatus(t *testing.T, embeddedStatus string) {
 	// TestReconcile runs "Reconcile" on a PipelineRun with one Task that has not been started yet.
 	// It verifies that the TaskRun is created, it checks the resulting API actions, status and events.
 	names.TestingSeed()
@@ -574,6 +604,7 @@ func TestReconcile(t *testing.T) {
 		Tasks:             ts,
 		ClusterTasks:      clusterTasks,
 		PipelineResources: rs,
+		ConfigMaps:        []*corev1.ConfigMap{withEmbeddedStatus(newFeatureFlagsConfigMap(), embeddedStatus)},
 	}
 	prt := newPipelineRunTest(d, t)
 	defer prt.Cancel()
@@ -680,18 +711,15 @@ func TestReconcile(t *testing.T) {
 		t.Errorf("Expected reason %q but was %s", v1beta1.PipelineRunReasonRunning.String(), condition.Reason)
 	}
 
-	if len(reconciledRun.Status.TaskRuns) != 2 {
-		t.Errorf("Expected PipelineRun status to include both TaskRun status items that can run immediately: %v", reconciledRun.Status.TaskRuns)
-	}
-	if _, exists := reconciledRun.Status.TaskRuns["test-pipeline-run-success-unit-test-1"]; !exists {
-		t.Errorf("Expected PipelineRun status to include TaskRun status but was %v", reconciledRun.Status.TaskRuns)
-	}
-	if _, exists := reconciledRun.Status.TaskRuns["test-pipeline-run-success-unit-test-cluster-task"]; !exists {
-		t.Errorf("Expected PipelineRun status to include TaskRun status but was %v", reconciledRun.Status.TaskRuns)
-	}
+	tr1Name := "test-pipeline-run-success-unit-test-1"
+	tr2Name := "test-pipeline-run-success-unit-test-cluster-task"
+
+	verifyTaskRunStatusesCount(t, embeddedStatus, reconciledRun.Status, 2)
+	verifyTaskRunStatusesNames(t, embeddedStatus, reconciledRun.Status, tr1Name, tr2Name)
 
 	// A PVC should have been created to deal with output -> input linking
 	ensurePVCCreated(prt.TestAssets.Ctx, t, clients, expectedTaskRun.GetPipelineRunPVCName(), "foo")
+
 }
 
 // TestReconcile_CustomTask runs "Reconcile" on a PipelineRun with one Custom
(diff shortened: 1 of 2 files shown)


pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -238,6 +238,36 @@ func getPipelineRunUpdates(t *testing.T, actions []ktesting.Action) []*v1beta1.P
 }
 
 func TestReconcile(t *testing.T) {
+	testCases := []struct {
```
lbernick, line 241: I see a lot of tests with these same test cases. They could probably be combined into one, e.g. 
```
testSuites := []func(*testing.T){testReconcileWithEmbeddedStatus, testReconcileWithCustomTask...}
for _, testCase := range testCases {
   for _, testClass := range testSuites {
      t.Run(tc.Name, testClass(t, tc.embeddedStatus))
   }
}

File: pkg/reconciler/pipelinerun/pipelinerun_test.go [hunk 2]
```diff
@@ -946,19 +974,182 @@ func TestReconcile_CustomTask(t *testing.T) {
...
+			verifyRunStatusesNames(t, config.DefaultEmbeddedStatus, reconciledRun.Status, tc.wantRun.Name)
+		})
+	}
+}
+
+// TestReconcile_CustomTaskWithEmbeddedStatus runs "Reconcile" on a PipelineRun
+// with one Custom Task reference that has not been run yet, and each possible
+// value for the "embedded-status" feature flag. It verifies that the Run is
+// created, it checks the resulting API actions, status and events.
+func TestReconcile_CustomTaskWithEmbeddedStatus(t *testing.T) {
```
lbernick, line 987: can this follow the same pattern as Test_Reconcile above?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -8787,3 +8882,107 @@ spec:
...
+}
+
+func shouldHaveMinimalEmbeddedStatus(embeddedVal string) bool {
+	return false
+	// TODO(abayer): Uncomment when implementation of TEP-0100 is done.
+	// return embeddedVal == config.MinimalEmbeddedStatus || embeddedVal == config.BothEmbeddedStatus
+}
+
+func verifyTaskRunStatusesCount(t *testing.T, embeddedStatus string, prStatus v1beta1.PipelineRunStatus, taskCount int) {
+	if shouldHaveFullEmbeddedStatus(embeddedStatus) && len(prStatus.TaskRuns) != taskCount {
```
lbernick, line 8899: rather than creating these functions "shouldhavefullembeddedstatus" etc I think it might be cleaner to not have these branches at all, and just add them when the embedded status functionality is enabled.

lbernick (commented):
Thanks Andrew! I'm still wondering if a pattern like the following could work?
```
func verifyPipelineRunStatus(t *testing.T, embeddedStatus string, prs *PipelineRunStatus, expectedChildStatuses map[string]ChildStatusReference) {
 //pseudocode
 taskRunStatuses := filter(expectedChildStatuses, status.value.kind == "TaskRun")
 runStatuses := filter(expectedChildStatuses, status.value.kind == "Run")
 if embeddedStatus == "full" {
   // verify all taskRuns in pipelineRunStatus have status, conditions, and when expressions matching those in taskRunStatuses, and likewise for runs
   // verify that childRefs has no entries
 }
 // do something similar for both/minimal
}
```

pkg/reconciler/pipelinerun/pipelinerun_test.go
abayer: I’m not sure what you mean? The point of this PR is to cut down on the size of the implementation PR, so if we don’t add the new test behavior for the embedded status functionality here, we have to do it in the implementation PR instead.

pkg/reconciler/pipelinerun/pipelinerun_test.go
abayer: Good question. I’d lean against this for clarity and ease of running individual test cases in an IDE.

abayer:
@lbernick Re that pattern - I feel like that would end up bloating the tests more than it would shrink them, since we’d have to define the full expected statuses for each case, while the existing tests don’t actually check the full statuses currently, just certain aspects. But I’ll give it another shot and see if a smooth enough approach reveals itself.

pkg/reconciler/pipelinerun/pipelinerun_test.go
abayer: Do you mean the separate function thingie? I didn’t bother splitting this one because it’s entirely new, so the diff would have been just as large either way. I guess there is value in consistency, though!

lbernick: yeah I think absent a strong reason to prefer either approach, let's keep things consistent

pkg/reconciler/pipelinerun/pipelinerun_test.go
lbernick: at the very least maybe the test cases struct can be defined only once?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -8787,3 +8882,107 @@ spec:
...
+	// return embeddedVal == config.FullEmbeddedStatus || embeddedVal == config.BothEmbeddedStatus
+}
+
+func shouldHaveMinimalEmbeddedStatus(embeddedVal string) bool {
+	return false
+	// TODO(abayer): Uncomment when implementation of TEP-0100 is done.
+	// return embeddedVal == config.MinimalEmbeddedStatus || embeddedVal == config.BothEmbeddedStatus
+}
+
+func verifyTaskRunStatusesCount(t *testing.T, embeddedStatus string, prStatus v1beta1.PipelineRunStatus, taskCount int) {
```
lbernick, line 8898: just one more comment-- t.Helper() would be useful here

vdemeester (commented):
We should make sure we aim for readability in our tests, no matter the size of the PR. I'd rather have a very readable set of tests in 2000 lines than a not-that-readable set of tests in 1000 lines.

Few things:
- Can we use the `test/pars` package and inline yaml definitions, it's usually easier to read ? Closer to what we "see" in real life.
- If we don't have those for some types (status, …), we could "enhance" that library
- "duplication is far cheaper than the wrong abstraction", especially on tests that we have to maintain over time.

pkg/reconciler/pipelinerun/pipelinerun_test.go
abayer: I considered that, but at least in Goland/IntelliJ, if you don't have the test case structs defined in the test function, you can't launch them individually.

pkg/reconciler/pipelinerun/pipelinerun_test.go
abayer: 👍

pkg/reconciler/pipelinerun/pipelinerun_test.go
abayer: Good catch!

pkg/reconciler/pipelinerun/pipelinerun_updatestatus_test.go
```diff
@@ -0,0 +1,670 @@
...
+
+type updateStatusTaskRunsData struct {
+	withConditions map[string]*v1beta1.PipelineRunTaskRunStatus
+	missingTaskRun map[string]*v1beta1.PipelineRunTaskRunStatus
+	foundTaskRun   map[string]*v1beta1.PipelineRunTaskRunStatus
+	recovered      map[string]*v1beta1.PipelineRunTaskRunStatus
+	simple         map[string]*v1beta1.PipelineRunTaskRunStatus
+}
+
+func getUpdateStatusTaskRunsData() updateStatusTaskRunsData {
```
abayer, line 44: I should note that this as a separate function/struct makes more sense if you see the [tests added in my WIP branch for the implementation](https://github.com/abayer/tektoncd-pipeline/blob/embedded-statuses-impl-mk2/pkg/reconciler/pipelinerun/pipelinerun_updatestatus_test.go#L1159) - there, we want to be able to test the function that calls either `updatePipelineRunStatusFromTaskRuns` and `updatePipelineRunStatusFromRuns` _or_ the added-in-the-impl `updatePipelineRunStatusFromChildRefs` (or all of 'em, for `embedded-status=both`). It didn't make sense to me to copy-paste the structs I end up using in both the existing test and the new one when I could just generate them in a function both the tests can call. I went with a struct as the return from that function rather than a map just because I thought it'd be more clear, but I'm not married to that choice. =)

abayer:
@vdemeester I'm probably going to come back later and see what makes sense to switch to parsed YAML, but I don't want to make this PR any more sprawling/confusing than it already is. =)

pkg/reconciler/pipelinerun/pipelinerun_test.go
lbernick: ok-- whatever you think is best here, although my intention isn't specifically to reduce number of lines of code but number of "things" that a given PR is doing

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -941,19 +969,187 @@ func TestReconcile_CustomTask(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			runTestReconcileCustomTaskWithEmbeddedStatus(t, tc.embeddedStatusVal)
 		})
 	}
 }
 
+func runTestReconcileCustomTaskWithEmbeddedStatus(t *testing.T, embeddedStatus string) {
```
lbernick, line 1012: it looks like this function may accidentally have been duplicated with the one above?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -1674,15 +1866,56 @@ func withOCIBundles(cm *corev1.ConfigMap) *corev1.ConfigMap {
...
 }
 
-func TestReconcileOnCancelledPipelineRun(t *testing.T) {
+func withEmbeddedStatus(cm *corev1.ConfigMap, flagVal string) *corev1.ConfigMap {
+	newCM := cm.DeepCopy()
+	newCM.Data[embeddedStatusFeatureFlag] = flagVal
+	return newCM
+}
+
+func TestReconcileOnCancelledPipelineRunFullEmbeddedStatus(t *testing.T) {
```
lbernick, line 1875: This test case could likely also use the pattern you're establishing (parameterizing the embedded status), because it doesn't really check any parts of the child statuses

pkg/reconciler/pipelinerun/pipelinerun_test.go
abayer: Ah, yeah, I think I added this one while initially prototyping, before the table-based changes. I'm gonna remove this.

pkg/reconciler/pipelinerun/pipelinerun_test.go
abayer: Yeah, good catch.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -702,67 +730,68 @@ func TestReconcile_CustomTask(t *testing.T) {
 	const pipelineRunName = "test-pipelinerun"
 	const pipelineTaskName = "custom-task"
 	const namespace = "namespace"
+
+	simpleCustomTaskPRYAML := `metadata:
```
abayer, line 734: This and `simpleCustomTaskWantRunYAML` are here so that we can use the same input `PipelineRun` and expected `Run` for test cases with different `embedded-status` values without duplicating everything.

pkg/reconciler/pipelinerun/pipelinerun_test.go
lbernick: I can't tell where that links to, but I think the pattern you have going of `func TestReconcile(t *testing.T)` +  `func runTestReconcileWithEmbeddedStatus(t *testing.T, embeddedStatus string)` works well, and it would also be nice to use it here for consistency. can you explain a bit more about the concern with doing it that way?

abayer: Ah, yeah, you have to expand the diff to see it.

So `TestReconcile_CustomTask` is already a table-based test - I didn't want to do the matrix-y thing of running each of its testcases once for each `embedded-status` value, so instead I added two new test cases to `TestReconcile_CustomTask`, using the same input `PipelineRun`/expected `Run` as the existing "custom task with taskref" test case, but with `embedded-status=full` and `embedded-status=minimal`. So it doesn't fit exactly the same pattern as the `runTest...` ones, but I think it's a better fit for this particular test.

abayer: Honestly, in practice, we could probably skip having test cases in `TestReconcile_CustomTask` for different `embedded-status` values, since we test reconciliation of a `Run` with the different `embedded-status` values in `runTestUpdatePipelineRunStatusFromInformer` now anyway.

pkg/reconciler/pipelinerun/pipelinerun_test.go
abayer: Fair enough. =) I guess I just feel like stubbing `shouldHaveFullEmbeddedStatus` and `shouldHaveMinimalEmbeddedStatus` here with the `TODO`s, and full implementations of `verify...` is clearer on the intent, particularly since there (hopefully!) will not be a long gap between this PR landing and the implementation PR landing.

lbernick:
/lgtm

don't forget to squash the commits

abayer:
Thanks, @lbernick! Squashed! =)

PR #4739, opened 2022-04-06
[TEP-0100] Implementation for embedded TaskRun and Run statuses in PipelineRuns

abayer:
# Changes

See:
* https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
* https://github.com/tektoncd/pipeline/pull/4705
* https://github.com/tektoncd/pipeline/pull/4734
* https://github.com/tektoncd/pipeline/pull/4753
* https://github.com/tektoncd/pipeline/pull/4757
* https://github.com/tektoncd/pipeline/pull/4760
* https://github.com/tektoncd/pipeline/issues/3140

This implements TEP-0100, allowing for choosing between the original full embedded `TaskRun` and
`Run` statuses in `PipelineRun` statuses, minimal child references to the underlying `TaskRun` and
`Run`s, or both, building on top of all the other PRs referenced above.

/kind tep
/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Added implementation for minimal `TaskRun` and `Run` statuses within `PipelineRun` statuses. 
```

### DIFF
--- a/pkg/reconciler/pipelinerun/cancel.go
+++ b/pkg/reconciler/pipelinerun/cancel.go
@@ -24,6 +24,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	clientset "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
@@ -99,9 +101,12 @@ func cancelPipelineRun(ctx context.Context, logger *zap.SugaredLogger, pr *v1bet
 func cancelPipelineTaskRuns(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, clientSet clientset.Interface) []string {
 	errs := []string{}
 
-	// Loop over the TaskRuns in the PipelineRun status.
-	// If a TaskRun is not in the status yet we should not cancel it anyways.
-	for taskRunName := range pr.Status.TaskRuns {
+	trNames, runNames, err := getChildObjectsFromPRStatus(ctx, pr.Status)
+	if err != nil {
+		errs = append(errs, err.Error())
+	}
+
+	for _, taskRunName := range trNames {
 		logger.Infof("cancelling TaskRun %s", taskRunName)
 
 		if _, err := clientSet.TektonV1beta1().TaskRuns(pr.Namespace).Patch(ctx, taskRunName, types.JSONPatchType, cancelTaskRunPatchBytes, metav1.PatchOptions{}, ""); err != nil {
@@ -109,8 +114,8 @@ func cancelPipelineTaskRuns(ctx context.Context, logger *zap.SugaredLogger, pr *
 			continue
 		}
 	}
-	// Loop over the Runs in the PipelineRun status.
-	for runName := range pr.Status.Runs {
+
+	for _, runName := range runNames {
 		logger.Infof("cancelling Run %s", runName)
 
 		if err := cancelRun(ctx, runName, pr.Namespace, clientSet); err != nil {
@@ -120,6 +125,43 @@ func cancelPipelineTaskRuns(ctx context.Context, logger *zap.SugaredLogger, pr *
 	}
 
 	return errs
+}
+
+// getChildObjectsFromPRStatus returns taskruns and runs in the PipelineRunStatus's ChildReferences or TaskRuns/Runs,
+// based on the value of the embedded status flag.
+func getChildObjectsFromPRStatus(ctx context.Context, prs v1beta1.PipelineRunStatus) ([]string, []string, error) {
+	cfg := config.FromContextOrDefaults(ctx)
+
+	var trNames []string
+	var runNames []string
+	unknownChildKinds := make(map[string]string)
+
+	if cfg.FeatureFlags.EmbeddedStatus != config.FullEmbeddedStatus {
+		for _, cr := range prs.ChildReferences {
+			switch cr.Kind {
+			case "TaskRun":
+				trNames = append(trNames, cr.Name)
+			case "Run":
+				runNames = append(runNames, cr.Name)
+			default:
+				unknownChildKinds[cr.Name] = cr.Kind
+			}
+		}
+	} else {
+		for trName := range prs.TaskRuns {
+			trNames = append(trNames, trName)
+		}
+		for runName := range prs.Runs {
+			runNames = append(runNames, runName)
+		}
+	}
+
+	var err error
+	if len(unknownChildKinds) > 0 {
+		err = fmt.Errorf("found child objects of unknown kinds: %v", unknownChildKinds)
+	}
+
+	return trNames, runNames, err
 }
 
 // gracefullyCancelPipelineRun marks any non-final resolved TaskRun(s) as cancelled and runs finally.
--- a/pkg/reconciler/pipelinerun/cancel_test.go
+++ b/pkg/reconciler/pipelinerun/cancel_test.go
@@ -20,6 +20,12 @@ import (
 	"context"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+	"github.com/tektoncd/pipeline/test/diff"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	_ "github.com/tektoncd/pipeline/pkg/pipelinerunmetrics/fake" // Make sure the pipelinerunmetrics are setup
@@ -32,12 +38,15 @@ import (
 
 func TestCancelPipelineRun(t *testing.T) {
 	testCases := []struct {
-		name        string
-		pipelineRun *v1beta1.PipelineRun
-		taskRuns    []*v1beta1.TaskRun
-		runs        []*v1alpha1.Run
+		name           string
+		embeddedStatus string
+		pipelineRun    *v1beta1.PipelineRun
+		taskRuns       []*v1beta1.TaskRun
+		runs           []*v1alpha1.Run
+		wantErr        bool
 	}{{
-		name: "no-resolved-taskrun",
+		name:           "no-resolved-taskrun",
+		embeddedStatus: config.DefaultEmbeddedStatus,
 		pipelineRun: &v1beta1.PipelineRun{
 			ObjectMeta: metav1.ObjectMeta{Name: "test-pipeline-run-cancelled"},
 			Spec: v1beta1.PipelineRunSpec{
@@ -45,7 +54,8 @@ func TestCancelPipelineRun(t *testing.T) {
 			},
 		},
 	}, {
-		name: "one-taskrun",
+		name:           "one-taskrun",
+		embeddedStatus: config.DefaultEmbeddedStatus,
 		pipelineRun: &v1beta1.PipelineRun{
 			ObjectMeta: metav1.ObjectMeta{Name: "test-pipeline-run-cancelled"},
 			Spec: v1beta1.PipelineRunSpec{
@@ -61,7 +71,8 @@ func TestCancelPipelineRun(t *testing.T) {
 			{ObjectMeta: metav1.ObjectMeta{Name: "t1"}},
 		},
 	}, {
-		name: "multiple-taskruns",
+		name:           "multiple-taskruns",
+		embeddedStatus: config.DefaultEmbeddedStatus,
 		pipelineRun: &v1beta1.PipelineRun{
 			ObjectMeta: metav1.ObjectMeta{Name: "test-pipeline-run-cancelled"},
 			Spec: v1beta1.PipelineRunSpec{
@@ -79,7 +90,8 @@ func TestCancelPipelineRun(t *testing.T) {
 			{ObjectMeta: metav1.ObjectMeta{Name: "t2"}},
 		},
 	}, {
-		name: "multiple-runs",
+		name:           "multiple-runs",
+		embeddedStatus: config.DefaultEmbeddedStatus,
 		pipelineRun: &v1beta1.PipelineRun{
 			ObjectMeta: metav1.ObjectMeta{Name: "test-pipeline-run-cancelled"},
 			Spec: v1beta1.PipelineRunSpec{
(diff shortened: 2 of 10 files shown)


pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -99,23 +99,48 @@ func cancelPipelineRun(ctx context.Context, logger *zap.SugaredLogger, pr *v1bet
...
-	for taskRunName := range pr.Status.TaskRuns {
-		logger.Infof("cancelling TaskRun %s", taskRunName)
-
-		if _, err := clientSet.TektonV1beta1().TaskRuns(pr.Namespace).Patch(ctx, taskRunName, types.JSONPatchType, cancelTaskRunPatchBytes, metav1.PatchOptions{}, ""); err != nil {
-			errs = append(errs, fmt.Errorf("Failed to patch TaskRun `%s` with cancellation: %s", taskRunName, err).Error())
-			continue
+	// If pr.Status.ChildReferences is populated, use that as source of truth for TaskRun and Run names.
+	if len(pr.Status.ChildReferences) > 0 {
+		// Loop over the ChildReferences in the PipelineRun status.
+		for _, cr := range pr.Status.ChildReferences {
```
lbernick, line 105: I might prefer a helper function which returns the runs and taskruns to cancel based on the embedded status value.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1183,116 +1198,167 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
 	}
-	updatePipelineRunStatusFromRuns(logger, pr, runs)
+
+	updatePipelineRunStatusFromSlices(ctx, logger, pr, taskRuns, runs)
 
 	return nil
 }
 
-func updatePipelineRunStatusFromTaskRuns(logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, trs []*v1beta1.TaskRun) {
+func updatePipelineRunStatusFromSlices(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, taskRuns []*v1beta1.TaskRun, runs []*v1alpha1.Run) {
```
lbernick, line 1212: nit: could this function be `updatePipelineRunStatusFromChildObjects`? IMO including things like "slices" in variable names can be redundant

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1183,116 +1198,167 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
+	fullEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.FullEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
+	minimalEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.MinimalEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
+
+	if minimalEmbedded {
+		taskNameByPipelineTask = updatePipelineRunStatusFromChildRefs(logger, pr, taskRuns, runs)
+	}
+	if fullEmbedded {
+		updatePipelineRunStatusFromTaskRuns(logger, pr, taskRuns, taskNameByPipelineTask)
+		updatePipelineRunStatusFromRuns(logger, pr, runs)
+	} else {
```
lbernick, line 1225: I don't think this else block should be necessary. If something depends on these values being non-nil let's change that code instead of initializing these.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1183,116 +1198,167 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
+		pr.Status.TaskRuns = make(map[string]*v1beta1.PipelineRunTaskRunStatus)
+		pr.Status.Runs = make(map[string]*v1beta1.PipelineRunRunStatus)
+	}
+}
+
+// updatePipelineRunStatusFromTaskRuns takes a PipelineRun, a list of TaskRuns within that PipelineRun, and a map of
+// existing PipelineTask names to TaskRun (or Run) names, and updates the PipelineRun's .Status.TaskRuns, including
+// ensuring that any (deprecated) condition checks are represented.
+func updatePipelineRunStatusFromTaskRuns(logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, trs []*v1beta1.TaskRun, taskRunNameByPipelineTask map[string]string) {
```
lbernick, line 1234: this function should be able to stay the same, no? (except for the small change removing the nil check for trs which I asked for :) )

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1183,116 +1198,167 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
-	updatePipelineRunStatusFromRuns(logger, pr, runs)
+
+	updatePipelineRunStatusFromSlices(ctx, logger, pr, taskRuns, runs)
 
 	return nil
 }
 
-func updatePipelineRunStatusFromTaskRuns(logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, trs []*v1beta1.TaskRun) {
+func updatePipelineRunStatusFromSlices(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, taskRuns []*v1beta1.TaskRun, runs []*v1alpha1.Run) {
+	taskNameByPipelineTask := make(map[string]string)
```
lbernick, line 1213: I don't think this variable (`taskNameByPipelineTask`) is necessary

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1312,3 +1378,136 @@ func updatePipelineRunStatusFromRuns(logger *zap.SugaredLogger, pr *v1beta1.Pipe
 		}
 	}
 }
+
+func updatePipelineRunStatusFromChildRefs(logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, trs []*v1beta1.TaskRun, runs []*v1alpha1.Run) map[string]string {
```
lbernick, line 1382: I don't think this function needs to return anything

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1312,3 +1378,136 @@ func updatePipelineRunStatusFromRuns(logger *zap.SugaredLogger, pr *v1beta1.Pipe
...
+	for i := range pr.Status.ChildReferences {
+		childRefByPipelineTask[pr.Status.ChildReferences[i].PipelineTaskName] = &pr.Status.ChildReferences[i]
+		taskRunNameByPipelineTask[pr.Status.ChildReferences[i].PipelineTaskName] = pr.Status.ChildReferences[i].Name
+	}
+
+	// Loop over all the TaskRuns associated to Tasks
+	for _, tr := range trs {
+		// Only process TaskRuns that are owned by this PipelineRun.
+		// This skips TaskRuns that are indirectly created by the PipelineRun (e.g. by custom tasks).
+		if len(tr.OwnerReferences) < 1 || tr.OwnerReferences[0].UID != pr.ObjectMeta.UID {
```
lbernick, line 1405: this block is repeated in a few places, can it be reused?

pkg/reconciler/pipelinerun/pipelinerun_updatestatus_test.go
```diff
@@ -668,3 +681,789 @@ func TestUpdatePipelineRunStatusFromRuns(t *testing.T) {
...
+					var fixedChildRefs []v1beta1.ChildStatusReference
+					re := regexp.MustCompile(`^[a-z\-]*?-(task|run)-[0-9]`)
+					for _, cr := range actualChildRefs {
+						cr.Name = fmt.Sprintf("%s-xxyyy", re.FindString(cr.Name))
+						fixedChildRefs = append(fixedChildRefs, cr)
+					}
+					actualPrStatus.ChildReferences = fixedChildRefs
+				}
+
+				// Sort the ChildReferences to deal with annoying ordering issues.
```
lbernick, line 1369: rather than sorting the output, I think you can provide extra inputs/flags to cmp.Diff to get it to ignore ordering of some fields.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -585,7 +594,13 @@ func getConditionCheckName(taskRunStatus map[string]*v1beta1.PipelineRunTaskRunS
 }
 
 // GetTaskRunName should return a unique name for a `TaskRun` if one has not already been defined, and the existing one otherwise.
-func GetTaskRunName(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus, ptName, prName string) string {
+func GetTaskRunName(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus, childRefs []v1beta1.ChildStatusReference, ptName, prName string) string {
```
lbernick, line 597: super nit: I think you can import and use constants for the `kind` of Run and TaskRun (here and elsewhere in the PR)

test/custom_task_test.go
```diff
@@ -122,11 +126,25 @@ spec:
 	}
 
 	// Get the Run name.
-	if len(pr.Status.Runs) != 2 {
-		t.Fatalf("PipelineRun had unexpected .status.runs; got %d, want 2", len(pr.Status.Runs))
+	var runNames []string
+	if embeddedStatusValue != config.MinimalEmbeddedStatus {
```
lbernick, line 130: I think the branches are off here, you probably want
```
if embeddedStatusValue != config.MinimalEmbeddedStatus {}
if embeddedStatusValue != config.FullEmbeddedStatus{}
```
both here and below (ensuring config.BothEmbeddedStatus is tested correctly)

test/featureflags.go
```diff
@@ -32,3 +34,17 @@ func requireAnyGate(gates map[string]string) func(context.Context, *testing.T, *
 		t.Skipf("No feature flag matching %s", strings.Join(pairs, " or "))
 	}
 }
+
+// GetEmbeddedStatusValue gets the current value for the "embedded-status" feature flag.
+// If the flag is not set, it returns the default value.
+func GetEmbeddedStatusValue(ctx context.Context, t *testing.T, kubeClient kubernetes.Interface) string {
```
lbernick, line 40: super nit: this function can just be called "getEmbeddedStatus"

test/v1alpha1/retry_test.go
```diff
@@ -44,6 +48,9 @@ func TestTaskRunRetry(t *testing.T) {
 	knativetest.CleanupOnInterrupt(func() { tearDown(ctx, t, c, namespace) }, t.Logf)
 	defer tearDown(ctx, t, c, namespace)
 
+	// Skip if the "embedded-status" feature flag has any value but "full".
```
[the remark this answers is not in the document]
abayer: Nope, that's still relevant in my mind - `v1alpha1` should, in theory, always be `full`. Though in practice because of how we have the type aliases set up, I guess you could mess around there. Eh, I'm fine either way.

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: Good idea! I did not like the name but couldn't think of a better one.

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: Lemme see how much breaks if we don't initialize 'em.

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: The changes here are either minor cleanup/renames (I really didn't like the `taskrun` variable name - I twitch when I see a variable name overriding an imported package name!), necessary signature changes, or the condition check changes with `getNewConditionChecksForTaskRun`. That said, I'll unrename things and clean up one or two inadvertent changes I made.

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: It is - we need to ensure we don't give a condition `TaskRun` different names from each of `...FromChildRefs` and `...FromTaskRuns` in the `both` scenario. It's annoying, but it is what it is.

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: See above at https://github.com/tektoncd/pipeline/pull/4739#discussion_r844421328 =)

pkg/reconciler/pipelinerun/pipelinerun_updatestatus_test.go
abayer: Oh yeaaaah - I just cargo-culted that over from the original `...FromTaskRuns` test blindly.

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: Huh, nothing but tests - nice!

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: It shows up twice for `TaskRun` and twice for `Run` - not sure if it's worth doing that as two functions, but I'm open to it.

pkg/reconciler/pipelinerun/pipelinerun_updatestatus_test.go
abayer: And actually, I'm not sure if `cmpopts.SortSlices` would make much of a difference here - it'd just be moving the same code around to be a parameter rather than standing alone...

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
abayer: I swear that I couldn't figure out where to find that to import. Maybe I missed something?

test/custom_task_test.go
abayer: Good catch!

test/featureflags.go
abayer: 👍

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1183,34 +1198,107 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
+
+	// Verify that the TaskRun and Run statuses match the ChildReferences for "both"
+	if cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus {
+		if len(prs.ChildReferences) != len(prs.TaskRuns)+len(prs.Runs) {
+			return fmt.Errorf("expected the same number of ChildReferences as the number of combined TaskRun and Run statuses (%d), but found %d",
+				len(prs.TaskRuns)+len(prs.Runs),
+				len(prs.ChildReferences))
+		}
+
+		for _, cr := range prs.ChildReferences {
```
lbernick, line 1253: thanks for adding this function! just a few quick comments:
- I would collect the names of child refs not in taskruns/runs and return an error containing all of them; this is more informative
- make sure you're handling the case where childRef isn't a taskrun or a run
- please add some tests for this function

pkg/reconciler/pipelinerun/pipelinerun.go
lbernick: got it-- any way to factor out the condition checks and pass them into the `updatePipelineRunStatus` functions? it seems like the `updatePipelineRunStatus` functions could be made more modular: they should just take some info and use it to update the PR status, without having to return data that will be used in other updates

pkg/reconciler/pipelinerun/pipelinerun.go
lbernick: these two functions in general share a lot of code, could you experiment to see how it can be consolidated?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
lbernick: they're [here](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/controller.go) but actually on second thought importing and using the controller name is a bit weird :/

test/custom_task_test.go
lbernick: same thing on line 218 and in custom_task_test

test/v1alpha1/retry_test.go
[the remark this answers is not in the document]
lbernick: either way makes sense, but the comment here doesn't reflect the code.

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: At this point there's just the one bit of cleanup left - switching `if trs == nil || len(trs) == 0 {` to `if len(trs) == 0 {`.

abayer: (oops accidentally hit resolve!)

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: Maaaaybe? I think it's just a necessary evil at this point - yet another annoying side effect of `conditions`.

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: Lemme see.

abayer: So fwiw on this particular block, I just tried going with a helper function or two (i.e., either one function each for `TaskRun` and `Run`, or one generic one with a type `switch`) and it didn't really make things any clearer, since we still have to have a block here for the `continue`. I'm still seeing if there are any other shared bits I can consolidate, but this particular one doesn't feel worth doing.

abayer: Actually, turns out that once I decided to try doing the filtering of `condition` `TaskRun`s before entering the main loop, I saw the obvious and realized that we could just filter out the "not owned by this `PipelineRun`" `TaskRun`s and `Run`s before entering the main loop as well. 🤦 =) How does https://github.com/tektoncd/pipeline/pull/4739/commits/6c8d61508a0aac2786cec9ae2e53e1114a91ade0 look?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
abayer: Oh right, I _did_ look at that and decided it felt too squirelly - I don't quite trust that the controller names are always going to be guaranteed to map one to one with the `kind`s. Feels like it might be worthwhile to add constants specifically for the `kind` name somewhere...

test/custom_task_test.go
abayer: Well I feel a silly man. Fixed.

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: 👍 on all of that!

pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -122,6 +127,45 @@ func cancelPipelineTaskRuns(ctx context.Context, logger *zap.SugaredLogger, pr *
 	return errs
 }
 
+// getChildObjectsToCancel looks in the PipelineRunStatus for TaskRun and Run names to cancel. It looks in
```
lbernick, line 130: nit: this function comment describes the implementation, and the name describes how the function will be used.
I'd prefer something like:
// getChildObjects returns taskruns and runs owned by the pipelineRun, based on the value of the embedded status flag.

pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -99,18 +101,21 @@ func cancelPipelineRun(ctx context.Context, logger *zap.SugaredLogger, pr *v1bet
 func cancelPipelineTaskRuns(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, clientSet clientset.Interface) []string {
 	errs := []string{}
 
-	// Loop over the TaskRuns in the PipelineRun status.
-	// If a TaskRun is not in the status yet we should not cancel it anyways.
-	for taskRunName := range pr.Status.TaskRuns {
+	trNames, runNames, err := getChildObjectsToCancel(ctx, pr.Status)
```
lbernick, line 104: this is looking much better! :)

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -97,56 +109,275 @@ func TestCancelPipelineRun(t *testing.T) {
...
+		embeddedStatus: config.DefaultEmbeddedStatus,
 		pipelineRun: &v1beta1.PipelineRun{
 			ObjectMeta: metav1.ObjectMeta{Name: "test-pipeline-run-cancelled"},
 			Spec: v1beta1.PipelineRunSpec{
 				Status: v1beta1.PipelineRunSpecStatusCancelledDeprecated,
 			},
 		},
+	}, {
+		name:           "child-references",
+		embeddedStatus: config.BothEmbeddedStatus,
```
lbernick, line 122: could you please add a test case for minimal embedded status?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1183,53 +1198,177 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
 	if err != nil {
 		logger.Errorf("could not list Runs %#v", err)
 		return err
 	}
-	updatePipelineRunStatusFromRuns(logger, pr, runs)
 
-	return nil
+	updatePipelineRunStatusFromChildObjects(ctx, logger, pr, taskRuns, runs)
+
+	return validateChildObjectsInPipelineRunStatus(ctx, pr.Status)
```
lbernick, line 1209: I think this function call should be moved to `updatePipelineRunStatusFromChildObjects`

pkg/reconciler/pipelinerun/pipelinerun.go
lbernick: maybe a better pattern here could be one function for minimal embedded status, one function for full embedded status, and one function for both embedded status which calls both of the first two. I think this way none of them should need to return anything. thoughts?

pkg/reconciler/pipelinerun/cancel.go
abayer: 👍 I'm renaming it to `filterChildObjects`.

pkg/reconciler/pipelinerun/cancel_test.go
abayer: 👍

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: 👍

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: Hmm...lemme give it a shot and see how it feels.

abayer: Actually, we'd still have the same problem in that approach - both the minimal and full functions would need to be able to generate names for `condition` `TaskRun`s, so the "both" function would still need to make sure that the name generated in the "minimal" function is still used in the "full" status. 

There's probably some nifty little hack of an approach to deal with this that I'm just not seeing yet. =)

abayer: The best thing I can come up with so far would be something like when we hit the "oh, we don't already have a `TaskRun` name in `pr.Status.TaskRuns` for this `condition` as part of the `PipelineTask` with this name" in the `...FromTaskRuns` function, we should first iterate over `pr.Status.ChildReferences` to see if there's a child ref there with that `PipelineTaskName`. Would that be preferable to the current approach?

abayer: hahaha, I just realized that `resources.GetTaskRunName` already does exactly that! Cleaning this up now.

abayer: And realizing that actually let me clean the logic up even more - see https://github.com/tektoncd/pipeline/pull/4739/commits/c934e216d1e4f151c4e73dde81de9ce458242598

abayer: (and, for that matter, `kmeta.ChildName` will actually return the same thing every time anyway, so my whole worry about the generated names matching didn't matter in the first place!)

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -97,56 +109,316 @@ func TestCancelPipelineRun(t *testing.T) {
...
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
+
 			d := test.Data{
 				PipelineRuns: []*v1beta1.PipelineRun{tc.pipelineRun},
 				TaskRuns:     tc.taskRuns,
 				Runs:         tc.runs,
 			}
 			ctx, _ := ttesting.SetupFakeContext(t)
+			cfg := config.NewStore(logtesting.TestLogger(t))
```
lbernick, line 230: what's this for?

test/custom_task_test.go
```diff
@@ -295,36 +330,47 @@ spec:
...
-		r, err := c.RunClient.Get(ctx, runName, metav1.GetOptions{})
-		if err != nil {
-			t.Fatalf("Failed to get Run %q: %v", runName, err)
+	if embeddedStatusValue != config.MinimalEmbeddedStatus {
+		if len(pr.Status.Runs) != 1 {
+			t.Fatalf("PipelineRun had unexpected .status.runs; got %d, want 1", len(pr.Status.Runs))
 		}
-		if r.IsDone() {
-			t.Fatalf("Run unexpectedly done: %v", r.Status.GetCondition(apis.ConditionSucceeded))
+		for rn := range pr.Status.Runs {
```
lbernick, line 339: I'm a bit confused, what's going on here?

abayer: We need the run name later in the test.

pkg/reconciler/pipelinerun/cancel_test.go
abayer: Ah, there's a line further down that needs to be removed. We need the context to be populated with the feature flags, hence this bit.

abayer: Oh, no, the later line is fine. Yeah, so what I said above. =)

test/custom_task_test.go
lbernick: oh I see, the slice has length 1-- in that case pr.Status.Runs[0] is more clear

pkg/reconciler/pipelinerun/cancel_test.go
lbernick: I think you can just do `cfg := withCustomTasks(withEmbeddedStatus(newFeatureFlagsConfigMap(), tc.embeddedStatus))`

lbernick (commented):
/lgtm

pkg/reconciler/pipelinerun/cancel_test.go
abayer: That creates a `ConfigMap`, but not a `*config.Store`, which is what we need to call `cfg.ToContext(ctx)`. I tried hand-crafting a `context.Context` so I could skip the `*config.Store` in between, but couldn't make it work right for `config.FromContextOrDefaults(ctx)` to pick it up properly.

test/custom_task_test.go
abayer: Ah, right, I tried to do that, but `pr.Status.Runs` is a map, so `pr.Status.Runs[0]` isn't an option.

abayer:
/hold cancel

Ok, squashed and ready to go!

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -97,56 +109,316 @@ func TestCancelPipelineRun(t *testing.T) {
...
+							t.Errorf("expected task %q to be marked as cancelled, was %q", r.Name, r.Spec.Status)
+						}
 					}
 				}
 			}
 		})
 	}
 }
+
+func TestGetChildObjectsToCancel(t *testing.T) {
```
lbernick, line 278: nit: function needs renaming

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -573,11 +573,22 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
...
+
+	if cfg.FeatureFlags.EmbeddedStatus == config.MinimalEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus {
+		pr.Status.ChildReferences = pipelineRunFacts.State.GetChildReferences(v1beta1.SchemeGroupVersion.String(),
+			v1alpha1.SchemeGroupVersion.String())
+	}
+
 	pr.Status.SkippedTasks = pipelineRunFacts.GetSkippedTasks()
 	if after.Status == corev1.ConditionTrue {
-		pr.Status.PipelineResults = resources.ApplyTaskResultsToPipelineResults(pipelineSpec.Results, pr.Status.TaskRuns, pr.Status.Runs)
+		pr.Status.PipelineResults = resources.ApplyTaskResultsToPipelineResults(pipelineSpec.Results, taskRunStatuses, runStatuses)
```
lbernick, line 591: this looks like a bug: if embedded status is minimal, taskRunStatuses and runStatuses here will be empty slices

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1183,53 +1198,166 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
+		}
+	}
+
+	return err
+}
+
+// filterTaskRunsForPipelineRun goes through the list of TaskRuns, skipping any which aren't owned by the given PipelineRun,
+// and then splitting the remainder into a slice of TaskRuns directly specified in the PipelineRun and a map of condition
+// TaskRuns keyed by the PipelineTask name they relate to.
+func filterTaskRunsForPipelineRun(logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, trs []*v1beta1.TaskRun) ([]*v1beta1.TaskRun, map[string][]*v1beta1.TaskRun) {
```
lbernick, line 1296: could you please reword this comment to describe the inputs and outputs of the function rather than its implementation?

pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -122,6 +127,42 @@ func cancelPipelineTaskRuns(ctx context.Context, logger *zap.SugaredLogger, pr *
 	return errs
 }
 
+// filterChildObjectsFromPRStatus returns taskruns and runs owned by the pipelineRun, based on the value of the embedded status flag.
+func filterChildObjectsFromPRStatus(ctx context.Context, prs v1beta1.PipelineRunStatus) ([]string, []string, error) {
```
lbernick, line 131: I think there is some confusion between this function and `filterTaskRunsForPipelineRun` (and likewise for runs). this function doesn't appear to check owner references at all.

pkg/reconciler/pipelinerun/cancel_test.go
abayer: Fixing!

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: 👍

pkg/reconciler/pipelinerun/cancel.go
abayer: Yeah, lemme clarify.

pkg/reconciler/pipelinerun/pipelinerun.go
lbernick: this doesn't feel right to me. if embedded status is minimal, we're using one function to get the TaskRuns to put into the PipelineRun's status, and a different function to get the TaskRuns to use for TaskRun results?

abayer: Ok, lemme know if https://github.com/tektoncd/pipeline/pull/4739/commits/e886963dfa53d19c9cd61c86b8810f6e93050aa7 feels good?

abayer:
/hold

This is now on top of #4757, so I'll un-hold it once that merges.

abayer:
/hold cancel

#4757 has merged. =)

abayer:
/hold

I spoke too soon - just split this up and created #4760.

abayer:
/hold cancel

Ok, hopefully this is the final form. =)

lbernick (commented):
/lgtm

lbernick:
If build tests are failing this seems unlikely to be a flake?

abayer:
the `pull-tekton-pipeline-build-tests` failure was pre-rebase, and it was a job that appears to not have properly started in the first place, getting marked as a failure for timing out. And literally while I was writing this, the current run passed. =)

abayer:
cc @tektoncd/core-maintainers - I'd be extremely grateful for a final review/approve (well, assuming it's approval-worthy!), so I can _finally_ get TEP-0100 put to bed for once and for all. =)

vdemeester (approved):

JaneSoo:
Hello, thank you all so much for this feature. How do I enable this feature? Should I add the flag in my pipelinerun? Do you have an example?

JaneSoo:
@lbernick I don't think I am subscribed to an email. Could you tell me when is the release date?

lbernick:
Please make sure you are subscribed to https://groups.google.com/g/tekton-users; we just sent out an email update (thanks for the reminder!) This release is scheduled for the 25th.

PR #4740, opened 2022-04-07
Use informer for pod get/list instead of talking to API server.

vaikas:
Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

# Changes
Instead of talking to API server for List/Get related to pods, use informer.

# Submitter Checklist

  functionality, content, code)

# Release Notes

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -51,6 +51,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/clock"
 	"k8s.io/client-go/kubernetes"
 	corev1Listers "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/reconciler/taskrun/taskrun.go
```diff
@@ -402,18 +405,23 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1beta1.TaskRun, rtr *re
...
-		pos, err := c.KubeClientSet.CoreV1().Pods(tr.Namespace).List(ctx, metav1.ListOptions{
-			LabelSelector: labelSelector,
-		})
+		labelSelector := labels.Everything()
+		myTaskLabelSelector, err := labels.NewRequirement(pipeline.TaskRunLabelKey, selection.Equals, []string{tr.Name})
+		if err != nil {
+			logger.Errorf("Failed to construct a label selector for %s = %s", pipeline.TaskRunLabelKey, tr.Name)
+			return err
+		}
+		labelSelector = labelSelector.Add(*myTaskLabelSelector)
```
mattmoor, line 414: I think you can do something tighter here with `labels.SelectorFromSet` where `labels.Set` is just a `map[string]string` where you can say `labels.Set{pipeline.TaskRunLabelKey: tr.Name}` and avoid the error path above entirely.

imjasonh:
@vdemeester @afrittoli Was it a conscious decision to avoid the informer cache here? I seem to recall there being races and flakes, but I think they may have been solved by the change that renamed pods so they were more uniformly mapped to taskrun names. Can we avoid direct API calls here now? 🙏

vdemeester:
> @vdemeester @afrittoli Was it a conscious decision to avoid the informer cache here? I seem to recall there being races and flakes, but I think they may have been solved by the change that renamed pods so they were more uniformly mapped to taskrun names. Can we avoid direct API calls here now? pray

Yeah I don't fully remember the reason, but I have a similar vague idea that it was that, and I also think it might have been fixed by the pod name being uniformly set now.

imjasonh (commented):
/lgtm

Thanks Ville! 👍

PR #4754, opened 2022-04-13
migrate yaml package to sigs.k8s.io/yaml

timonwong:
# Changes

Update `github.com/ghodss/yaml` package to its permanent fork `sigs.k8s.io/yaml`, since the former releasing any new version for a long time, on the other hand, the latter is still under active development (fix CVEs, etc).

Besides, `default-pod-template` and `default-affinity-assistant-pod-template` in `config-defaults` are now deserialized using `UnmarshalStrict` first, print a warning message, and try again with `Unmarshal`, in order to keep the compatibility.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


timonwong:
Since `UnmarshalStrict()` may introduce incompatibility, should we keep using `Unmarshal` instead?

vdemeester:
> Since `UnmarshalStrict()` may introduce incompatibility, should we keep using `Unmarshal` instead?

That's a good point and I think we should indeed, but maybe with a plan (and communication) to move to `UnmarshalStrict()` (with warnings of what types of incompatibility it would bring, and when *according to plan*).

timonwong:
> That's a good point and I think we should indeed, but maybe with a plan (and communication) to move to `UnmarshalStrict()` (with warnings of what types of incompatibility it would bring, and when _according to plan_).

Now the old behavior is preserved. (with a warning message)

vdemeester (approved):

vdemeester:
I think the error comes from `pkg/reconciler/pipelinerun/pipelinerun_updatestatus_test.go` that still calls `github.com/ghodss/yaml` 😅 
@timonwong if you don't have time to fix it, I can fix it in the PR 😉

timonwong:
> I think the error comes from `pkg/reconciler/pipelinerun/pipelinerun_updatestatus_test.go` that still calls `github.com/ghodss/yaml` 😅
> @timonwong if you don't have time to fix it, I can fix it in the PR 😉

@vdemeester thanks, I've submitted a new commit which address that :)

vdemeester:
/hold
@timonwong can you squash your commits ? 
(Also, the UI is telling "cannot be rebased due to conflicts)

abayer:
/retest

One of the sidecar tests flaking, yet again,.

vdemeester:
Thanks @timonwong 🎉

PR #4769, opened 2022-04-18
Omit init containers from limitrange default request calculations

lbernick:
# Changes

This commit modifies how resource requirements are adjusted based on LimitRanges present
in a namespace. Prior to this commit, the container requests for a resource
were determined by dividing the default requests specified in a LimitRange by the number of
app containers plus 1 (to represent the init container), taking the maximum of this value
and the minimum request specified in the LimitRange, and applying this value to all containers.

However, Kubernetes does not sum init container resource requests to app container resource requests
when determining the pod's effective resource requests, as described [here](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#resources). Instead, it takes the larger of
the maximum init container resource requests and the sum of app container resource requests.
As a result, the appropriate behavior is to split the default requests only among app containers, and
to apply the default requests directly to init containers. This commit implements that behavior.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Omit init containers from limitrange default request calculations
```

### DIFF
--- a/pkg/internal/limitrange/transformer.go
+++ b/pkg/internal/limitrange/transformer.go
@@ -47,20 +47,17 @@ func NewTransformer(ctx context.Context, namespace string, lister corev1listers.
 		// computed if there is multiple LimitRange to satisfy the most (if we can).
 		// Count the number of containers (that we know) in the Pod.
 		// This should help us find the smallest request to apply to containers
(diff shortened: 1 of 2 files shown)


vdemeester (approved):

vdemeester:
cc @abayer

PR #4774, opened 2022-04-19
Instrument the kaniko test for logstream.

mattmoor:
We see this flaking a fair amount downstream, and this will make it easier to cross correlate our logs to see what's happening in the reconciler as things go awry.

This also fixes a place where we were improperly using a constant in `pipelinerun_test.go`.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/kaniko_task_test.go
+++ b/test/kaniko_task_test.go
@@ -33,13 +33,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	knativetest "knative.dev/pkg/test"
+	"knative.dev/pkg/test/helpers"
 )
 
 const (
-	kanikoTaskName          = "kanikotask"
-	kanikoTaskRunName       = "kanikotask-run"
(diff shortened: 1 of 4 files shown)


mattmoor:
One subtlety worth calling out: We *always* run `logstream`, but we need to name test resources using the `helpers.ObjectNameForTest` so that it can infer from the resource names, which tagged log lines from the controller to inline into each test's log context.  That's why none of the changed lines mentioned `logstream`, but that's the net effect of the change 🤣

mattmoor:
cc @afrittoli @vdemeester

abayer:
/lgtm

Is there a good reason not to just update all the e2e tests to use `helpers.ObjectNameForTest`? Assuming there isn’t, I’ll get a PR up to do that across the board.

mattmoor:
@abayer That's what we do in Knative, I've just been chipping away as we hit flakes where I wish I had it.  Doing it everywhere would be awesome 🤩

vdemeester (approved):

Issue #4737, opened 2022-04-05
Owning our *Container* spec


vdemeester:
### Feature request

As of today, the `TaskSpec` *Step* is embedding the k8s "upstream" `Container` spec. The "possible" idea here would be to migrate to owning / defining our own `Container` spec instead. This would allow to control some of the "leaks* that comes from k8s in our types. For example, in the [`buildkit-tekton`](https://github.com/vdemeester/buildkit-tekton) experimentation, there is a few fields of the `Container` struct we silently ignore because they don't make sense in a "CI" workflow type.

```
		// Silently ignore Ports
		// Silently ignore LivenessProbe
		// Silently ignore ReadinessProbe
		// Silently ignore StartupProbe
		// Silently ignore Lifecycle
		// Silently ignore TerminationMessagePath
		// Silently ignore TerminationMessagePolicy
		// Silently ignore Resources (for now)
		// Silently ignore ImagePullPolicy
		// Silently ignore Stdin
		// Silently ignore StdinOnce
		// Silently ignore TTY
```

In addition it would make it "cleaner" on openapi schema generation (because we don't get all the fields).
One possible downside might be "duck typing" as, now, the `Task` spec might not be as "PodSpecable" that it could.

If we want to *go* that way, we should do it before `v1`, otherwise, we can ignore that issue 😛. 

/cc @tektoncd/core-maintainers @mattmoor 

### Use case

N/A, it's more a "technical" feature request / open question/discussion

pritidesai:
_I have a limited understanding of the downside here (because of a limited knowledge of `duck typing` 😆 )._ 

Anyways, I like the idea of having our own `Container` spec as it provides us flexibility in introducing pipeline/task specific fields which might not be relevant/available in upstream `Container` spec. It provides us extensibility and decouples from the k8s resources.

lbernick:
I like the idea of owning our own container spec since it could help with issues like #4080. I think we could make this change without breaking existing working yamls but it would break the go client libraries. Priti pointed out that there is precedence for breaking go libraries w/out breaking yamls in #2826.

I don't think we should remove the container fields that can't be implemented in buildkit since users may be depending on them, and our [API spec doc](https://github.com/tektoncd/pipeline/blob/main/docs/api-spec.md#step) states that those fields aren't required for tekton conformance.

I know what duck typing is but I don't quite understand what you're saying about it here, could you elaborate?

vdemeester:
[...]

Issue #4080, opened 2021-07-08
Allow to use variable replacement when defining resource limits and requests


goldmann:
### Feature request

Consider this example Task:

```yaml
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: resource-example
spec:
  description: Sample Task to show issues with passing parameters to resources.
  params:
    - name: RESOURCE_MEMORY_REQUEST
      type: string
      default: 1Gi
    - name: RESOURCE_MEMORY_LIMIT
      type: string
      default: 2Gi
  steps:
    - name: noop
      image: registry.access.redhat.com/ubi8/ubi-minimal:8.2
      resources:
        requests:
          memory: $(params.RESOURCE_MEMORY_REQUEST)
        limits:
          memory: $(params.RESOURCE_MEMORY_LIMIT)
```

When creating such a resource it fails with:

```
Error from server (BadRequest): error when creating "minimal-resource.yaml": admission webhook "webhook.pipeline.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
```

This suggests that webhook is not able to decipher what is the value. And since the value is correct, it should be allowed to do use it this way.

### Use case

Main use case is to dynamically change requests and resources based on the runtime requirements. I consider this as a workaround for general lack of a nice way to do it otherwise. But even the workaround doesn't work currently :)

jcatana:
This is very useful for buildah or other image building tasks. Clusters with namespace resource quotas MUST have resource limits and requests defined.
Having to define resources statically on the task level can be annoying, I currently need to modify the task templates, and create buildah-small-1GB-ram, buildah-large-30GB-ram type tasks.
Some image builds can be very trivial and small, while others may include large scale code compilation and need much more resources. 
It would be ideal to have this functionality.

Duologic:
keywords for search: variable substitution

michaelsauter:
I came across this as I naively hoped to set the resource constraints via parameters. I think this feature is crucial as the place to decide the resource requirements is more likely the pipeline than the task.

I guess the regex validator would need to allow a parameter reference then?

rporres:
Looks like a duplicate of https://github.com/tektoncd/pipeline/issues/1530, which got closed without resolution

rporres:
[...]

PR #4778, opened 2022-04-19
Add Tekton-owned Step, StepTemplate, and Sidecar

lbernick:
# Changes

This commit copies the Container struct from Kubernetes into Tekton into the
existing structs Task.Step.Container, Task.Sidecar.Container, and Task.StepTemplate.
This change breaks the Tekton Go libraries, but does not require any changes to the yaml
that users provide. This should not change anything for users who do not use the Go libraries.

The reason for this change is to have more control over our Task API. For example,
we may want to modify the Container fields we support (e.g. #4737), or change fields to string types
to allow them to be parameterized (e.g. #4080). This commit is the first step towards being able to make changes
to parts of the Task API embedded in Container.

Part 1 of #4737 

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
[Change to Go libraries]: Task.Step, Task.StepTemplate, and Task.Sidecar use Container fields directly instead of embedding the Container struct
```


vdemeester (commented):
Really quick review but few questions:
- Would it make sense to have two different struct for sidecars and steps ? (sidecars container would be step container + some things like readinessprobe, *probe …)
- I see `v1alpha1.Container`, any reason for changing code in `v1alpha1` ? (I guess it might be because we "share" the `Step` struct)

lbernick:
> * Would it make sense to have two different struct for sidecars and steps ? (sidecars container would be step container + some things like readinessprobe, *probe …)

Yes that's definitely where I see this going but I didn't want to remove any fields of the Step container yet (might need a TEP). But I can create two identical structs for sidecars and steps that we can later change, what do you think?

> * I see `v1alpha1.Container`, any reason for changing code in `v1alpha1` ? (I guess it might be because we "share" the `Step` struct)

Yeah pretty much because we share the Step struct. I could update the v1alpha1 Step to just copy over all the fields of the v1beta1 step but then we have to keep them in sync. Open to doing that though.

abayer:
fwiw, I'd lean towards copying over to v1alpha1 normally, but I'm not sure if it's worth it given that [TEP-0105](https://github.com/tektoncd/community/blob/main/teps/0105-remove-pipeline-v1alpha1-api.md) is starting to come together for actually removing v1alpha1. Given that, the additional work involved in copying over and keeping in sync for the time being may not be worth it.

imjasonh:
> I didn't want to remove any fields of the Step container yet (might need a TEP)

Happy to discuss if folks disagree, but in general I don't think deprecating a "feature" like, e.g., `readinessProbe`s on a step container, should require a TEP. We should signal deprecations of these fields, wait 1+ release, and remove them. Especially so for things like `terminationMessage` which just silently don't do anything -- there's no way to use that correctly today.

If we think eventually we'd like to get to a spot where the `Step` and `Sidecar` types just have all the fields they reasonably could need and don't embed a `Container` type (even if it's owned by us), then I think we should just go straight to that and not incur the churn on ourselves and Go-API-consumers in having to account for two breaking type changes.

lbernick:
> fwiw, I'd lean towards copying over to v1alpha1 normally, but I'm not sure if it's worth it given that [TEP-0105](https://github.com/tektoncd/community/blob/main/teps/0105-remove-pipeline-v1alpha1-api.md) is starting to come together for actually removing v1alpha1. Given that, the additional work involved in copying over and keeping in sync for the time being may not be worth it.

That's fair-- I wasn't sure if it would be more work to continue to alias v1alpha1 types to v1beta1 types, or to copy the old v1beta1 struct over to v1alpha1 so I wouldn't have to modify the v1alpha1 tests. I already did most of the grunt work though for pipelines -- when you're talking about the work to support this are you anticipating that this will be more difficult for the CLI team?

> > I didn't want to remove any fields of the Step container yet (might need a TEP)
> 
> Happy to discuss if folks disagree, but in general I don't think deprecating a "feature" like, e.g., `readinessProbe`s on a step container, should require a TEP. We should signal deprecations of these fields, wait 1+ release, and remove them. Especially so for things like `terminationMessage` which just silently don't do anything -- there's no way to use that correctly today.
> 
> If we think eventually we'd like to get to a spot where the `Step` and `Sidecar` types just have all the fields they reasonably could need and don't embed a `Container` type (even if it's owned by us), then I think we should just go straight to that and not incur the churn on ourselves and Go-API-consumers in having to account for two breaking type changes.

Point taken! I'd like to keep the scope of this PR to just swapping over the type definitions, and not deprecating anything. However, I created separate structs for Step and Sidecar containers to hopefully minimize churn. There's still a lot of copypasta so LMK if you'd prefer to see it done differently.

lbernick:
tests pass! 💪

/assign @vdemeester 
/assign @imjasonh

vdemeester:
> That's fair-- I wasn't sure if it would be more work to continue to alias v1alpha1 types to v1beta1 types, or to copy the old v1beta1 struct over to v1alpha1 so I wouldn't have to modify the v1alpha1 tests. I already did most of the grunt work though for pipelines -- when you're talking about the work to support this are you anticipating that this will be more difficult for the CLI team?
> 
> > >

We will remove support for `v1alpha1` in the cli very soon, if this push us to do it a bit early, I am more than fine with it 👼🏼.

> Especially so for things like `terminationMessage` which just silently don't do anything -- there's no way to use that correctly today.

Right, it's even more trickier than this. We use `terminationMessage` for Results, so as of today, this is very confusing to have / see in the spec.

> If we think eventually we'd like to get to a spot where the `Step` and `Sidecar` types just have all the fields they reasonably could need and don't embed a `Container` type (even if it's owned by us), then I think we should just go straight to that and not incur the churn on ourselves and Go-API-consumers in having to account for two breaking type changes.

That's a good point. I think we use the `Container` type today on both because one is not clearly wrapping the other one. With owning the struct, I think we may see that `Sidecar` has all the `Step` type and more. If that's the case, and we feel that's always gonna be the case, we could just merge `Step` and `Container` in the same struct as you suggest.

imjasonh:
> Not sure which is worse, lots of copypasta or the potential for future additional breaking changes to the go libs (probably the latter?)

I'd personally go with copypasting in this case. I don't think there's much conceptual benefit in signalling "sidecars provide a superset of functionality of steps" and it sounds like we might not even believe that will continue to hold. It also means declaring the Go structs to be a little more awkward (`Steps: []Step{{Container: Container{Image: "..."}, Script: "..."}}` vs `Steps: []Step{{Image: "...", Script: "..."}}`).

I won't die on this hill though, if folks disagree.

Thanks for working on this, and for considering all these options as you do so 👍

lbernick:
sounds good-- I think this PR should be good to go then!

lbernick:
> > sounds good-- I think this PR should be good to go then!
> 
> I think with this latest change we should just drop the `StepContainer` and `SidecarContainer` structs, and move the fields directly into `Step` and `Sidecar`. WDYT?

Oh my bad, I understand now what you were suggesting. Will do, I'll comment here when I've made that change.

lbernick:
I'm thinking of keeping StepContainer so it can be reused in StepTemplate, but copying the container fields into sidecar as well, thoughts?

pkg/apis/pipeline/v1beta1/task_types.go
```diff
@@ -172,7 +172,7 @@ type Step struct {
 
 // Sidecar has nearly the same data structure as Step, consisting of a Container and an optional Script, but does not have the ability to timeout.
 type Sidecar struct {
-	corev1.Container `json:",inline"`
+	SidecarContainer `json:",inline"`
```
vdemeester, line 175: The idea was to move all that is to not have this inlined/embedded struct but have the field directly in. So no need for `SidecarContainer`.

lbernick:
> > I'm thinking of keeping StepContainer so it can be reused in StepTemplate, but copying the container fields into sidecar as well, thoughts?
> 
> Good question. Are there fields that make sense in a `Step` but not a `StepTemplate`, or vice versa? Can the type of `StepTemplate` just be `Step`?

I think "Name" wouldn't make sense in StepTemplate since [steps can't have the same name](https://github.com/tektoncd/pipeline/blob/2b8538214169ec1c3af14221f44ea59e6b1b81f7/pkg/apis/pipeline/v1beta1/task_validation.go#L176). In addition, "timeout" is not currently present in StepTemplate, but it could make sense. so maybe these fields should be directly copied into stepTemplate as well, thoughts?

imjasonh:
> I think "Name" wouldn't make sense in StepTemplate since [steps can't have the same name](https://github.com/tektoncd/pipeline/blob/2b8538214169ec1c3af14221f44ea59e6b1b81f7/pkg/apis/pipeline/v1beta1/task_validation.go#L176). In addition, "timeout" is not currently present in StepTemplate, but it could make sense. so maybe these fields should be directly copied into stepTemplate as well, thoughts?

Thanks for looking into it. It sounds like we should just copy these between the types that need them. I'm not sure we should add `timeout` to StepTemplate in this PR, but it's something worth keeping in mind for a future change.

lbernick:
> > I think "Name" wouldn't make sense in StepTemplate since [steps can't have the same name](https://github.com/tektoncd/pipeline/blob/2b8538214169ec1c3af14221f44ea59e6b1b81f7/pkg/apis/pipeline/v1beta1/task_validation.go#L176). In addition, "timeout" is not currently present in StepTemplate, but it could make sense. so maybe these fields should be directly copied into stepTemplate as well, thoughts?
> 
> Thanks for looking into it. It sounds like we should just copy these between the types that need them. I'm not sure we should add `timeout` to StepTemplate in this PR, but it's something worth keeping in mind for a future change.

Agreed! I don't want to change the user facing API at all with this PR-- just mentioning timeout as something that could be added in the future.

lbernick:
@vdemeester @imjasonh I've addressed your feedback

imjasonh (commented):
/lgtm

Thanks for this!

lbernick:
> Thanks for this!

Happy to help!

vdemeester (approved):
/meow

PR #4786, opened 2022-04-22
[TEP-0075] Support Dictionary in Params

chuangw6:
# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes


pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -160,6 +174,14 @@ func NewArrayOrString(value string, values ...string) *ArrayOrString {
 	}
 }
 
+// NewObject creates an ArrayOrString of type ParamTypeObject using the provided key-value pairs
```
ywluogg, line 177: Nice job putting into thoughts about how to do them uniformly. With the considerations that ArrayOrString will be renamed to something else, I'm proposing to call this new struct as StructuredObject, to distinguish this struct, which could be containing strings, arrays or maps, from the "object" (dictionary) concept in json schema.

With this proposal, and because Go doesn't support function overloading, I think it makes sense to actually have three functions, NewStructuredObjectInString, NewStructuredObjectInObject, and NewStructuredObjectInObject. Pinging @vdemeester @pritidesai for opinions.

chuangw6: Great point! Yeah we definitely need to rename the struct `ArrayOrString` to something more general. Thanks for calling this out. The word `object` in `StructuredObject ` might be confusing with the actual `object` type that we will support. How about `StructuredParamValue`?

In terms of splitting `NewArrayOrString` function into three, I don't see any issue with that because this function is mainly used in test and just one place in the implementation code - `resolveResultRef` [function](https://github.com/tektoncd/pipeline/blob/2b8538214169ec1c3af14221f44ea59e6b1b81f7/pkg/reconciler/pipelinerun/resources/resultrefresolution.go#L158). We just need to be very careful with refactor because this function is extensively used in unit tests.

ywluogg: > 

Ah this will be shared with results as well later

chuangw6: > > 
> 
> Ah this will be shared with results as well later

yes, result will use this struct `ArrayOrString ` as well. cc: @Yongxuanzhang

Yongxuanzhang: > > > 
> > 
> > 
> > Ah this will be shared with results as well later
> 
> yes, result will use this struct `ArrayOrString ` as well. cc: @Yongxuanzhang

yes it is already in my next pr, thanks for cc me!

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -122,6 +138,25 @@ func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params
...
+		miss := list.DiffLeft(neededObjectParamsKeys[param.Name], providedObjectParamsKeys[param.Name])
+		if len(miss) != 0 {
+			missingObjectKeysParamNames = append(missingObjectKeysParamNames, param.Name)
+		}
+	}
+
+	if len(missingObjectKeysParamNames) != 0 {
+		return fmt.Errorf("taskrun params miss keys for params %v defined in properties", missingObjectKeysParamNames)
+	}
+
```
ywluogg, line 159: Can you make this and the above code as a separate function like: ValidateParamAgainstSchema(ps ParamSpec, p Param)? This way we can use it elsewhere.
Also wrap ValidateObjectAgainstSchema() against another sharable function such as ValidateObjectAgainstSchema(p PropertySpec, object ArraryOrString) This way, it can be used in Chains and Results in Pipelines

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -107,17 +132,23 @@ type ArrayOrString struct {
...
+	switch value[0] {
+	case '"':
 		arrayOrString.Type = ParamTypeString
 		return json.Unmarshal(value, &arrayOrString.StringVal)
+	case '[':
+		arrayOrString.Type = ParamTypeArray
+		return json.Unmarshal(value, &arrayOrString.ArrayVal)
+	default:
+		arrayOrString.Type = ParamTypeObject
+		return json.Unmarshal(value, &arrayOrString.ObjectVal)
```
ywluogg, line 150: Don't make this as default. Make a case for object and error at default case

pkg/apis/pipeline/v1beta1/param_context.go
```diff
@@ -130,16 +134,26 @@ func getContextParams(ctx context.Context, overlays ...Param) []Param {
...
 				Type:      ParamTypeString,
 				StringVal: fmt.Sprintf("$(params.%s)", ps.Name),
 			}
-		} else {
+		case ParamTypeArray:
 			p.Value = ArrayOrString{
 				Type:     ParamTypeArray,
 				ArrayVal: []string{fmt.Sprintf("$(params.%s[*])", ps.Name)},
 			}
+		default:
```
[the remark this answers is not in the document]
chuangw6: > I think when the ParamSpec reaches here, the type should never be empty as webhook should have set the defaults.

Correct, mutating webhook will be able to infer the default type by its specification and set the type to one of the three types accordingly. And at this point here, the type has already been set to one of the three. Therefore, the default case will be object if the type is not `ParamTypeString` or `ParamTypeArray` and we don't need to throw error here. (In addition, as this function is used in `pkg/apis/pipeline/v1beta1/pipeline_defaults.go`, it seems mutating webhook does not throw errors generally.)

In the original implementation, the if-else statement also works like the default is array type instead of throwing error.

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -38,40 +38,64 @@ type ParamSpec struct {
...
+			pp.Type = pp.Default.Type
+		} else {
+			// determine the type based on the array or string values when default value is specified but not the type
+			if pp.Default.ArrayVal != nil {
+				pp.Type = ParamTypeArray
 			} else {
-				// determine the type based on the array or string values when default value is specified but not the type
-				if pp.Default.ArrayVal != nil {
-					pp.Type = ParamTypeArray
+				// not using else-if to avoid compaint from gocritic
```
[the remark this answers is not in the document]
chuangw6: It will suggest to use `switch` when `if-else` chains is detected, which will fail the build tests. And apparently this code chuck cannot be done with switch because we are checking different variables in the struct.

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -320,6 +336,48 @@ func ValidateResourcesVariables(steps []Step, resources *TaskResources) *apis.Fi
 	return validateVariables(steps, "resources.(?:inputs|outputs)", resourceNames)
 }
 
+func validateObjectUsage(steps []Step, params []ParamSpec) (errs *apis.FieldError) {
```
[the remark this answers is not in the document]
chuangw6: Good point. I'll add a todo for it as it will also validate that entire object is not referenced as part of a string values.

pkg/apis/pipeline/v1beta1/param_types.go
chuangw6: sgtm!

pkg/reconciler/taskrun/validate_resources.go
chuangw6: > Can you make this and the above code as a separate function like: ValidateParamAgainstSchema(ps ParamSpec, p Param)? This way we can use it elsewhere.

sgtm! I'd use `properties` instead of `Schema` just to make sure we use terms consistently. 

> Also wrap ValidateObjectAgainstSchema() against another sharable function such as ValidateObjectAgainstSchema(p PropertySpec, object ArraryOrString)

can you explain this a bit? what kind of another helper function are you looking for?

pkg/apis/pipeline/v1beta1/param_types.go
Yongxuanzhang: I think we can treat all unsupported value to string. This is because the type and func is reused by results. For emitting results users may write integer to results path, in current code this will be read as string.

So to use ArrayOrString's UnmarshalJSON for emitting results and keep current behaviour not changed. I suggest we convert value to string if it doesn't fall to any switch condition or encountered err

I describe the case here:
https://github.com/tektoncd/pipeline/issues/4820

pkg/reconciler/taskrun/validate_resources.go
ywluogg: I can see that `ValidateParamAgainstSchema ` can be helpful when you try to validate params, but this is also a functionality that TaskRun results want to use for objects, so having a separate ValidateObjectAgainstSchema(p PropertySpec, object ArraryOrString) can give us another level of sharable code. Let me know if this clarify what I'm trying to say

ywluogg:
Can you also guard your new changes such as validations with alpha feature flags?

pkg/reconciler/taskrun/validate_resources.go
chuangw6: Can you clarify the difference between the two functions' signatures? Just trying to understand the needs and what level of isolation I should refactor.

For `ValidateParamAgainstProperties `, is it `ValidateParamAgainstProperties(paramSpecs []v1beta1.ParamSpec, params []v1beta1.Param)`? Just thinking whether to make this function parallel to the existing `validateParams` OR to utilizing `ValidateParamAgainstProperties` inside `validateParams` by first filtering out all object params and passing to `ValidateParamAgainstProperties`. 
And what is the second argument `object` of `ValidateObjectAgainstSchema`?

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -38,40 +38,64 @@ type ParamSpec struct {
...
+	// are currently "string", "array" and "object", and "string" is the default.
 	// +optional
 	Type ParamType `json:"type,omitempty"`
 	// Description is a user-facing description of the parameter that may be
 	// used to populate a UI.
 	// +optional
 	Description string `json:"description,omitempty"`
+	// Properties is the JSON Schema properties to support key-value pairs parameter.
+	// +optional
+	Properties map[string]PropertySpec `json:"properties,omitempty"`
```
wlynch, line 50: OpenAPI defines properties as a recursive schema object - https://swagger.io/specification/#schema-object

Should we take a similar approach here with ParamSpec?

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -73,13 +73,29 @@ func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params
 	var neededParams []string
 	paramTypes := make(map[string]v1beta1.ParamType)
 	neededParams = make([]string, 0, len(paramSpecs))
+	neededObjectParamsKeys := make(map[string][]string)
 	for _, inputResourceParam := range paramSpecs {
 		neededParams = append(neededParams, inputResourceParam.Name)
 		paramTypes[inputResourceParam.Name] = inputResourceParam.Type
+
+		// collect needed keys for each parameter of type Object
+		if inputResourceParam.Type == v1beta1.ParamTypeObject {
```
wlynch, line 82: I'm not the biggest fan that we're interleaving general param validation and object-specific validation here - it makes it harder to follow the code because you need to context switch back between what is / isn't object specific. Could we hoist this the object-specific code into a single func that gets called after we've done the param validation that applies to all params?

wlynch (changes requested):

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -552,6 +564,75 @@ func TestTaskSpecValidateError(t *testing.T) {
...
+					"key1": "var1",
+					"key3": "var1",
+				}),
+			}},
+			Steps: validSteps,
+		},
+		expectedError: apis.FieldError{
+			Message: fmt.Sprintf("Required keys %s for the parameter \"%s\" are not provided in default.", []string{"key2"}, "myobjectParam"),
+			Paths:   []string{""},
+		},
```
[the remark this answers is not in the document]
ywluogg: +1 I think this PR is mixed with multiple stuff, for example, default values, but miss a big chunk of other validations. Maybe scope out this PR to a specific module / feature and that would make the PR easier to follow

pkg/reconciler/taskrun/validate_resources.go
ywluogg: +1 to that

pkg/reconciler/taskrun/validate_resources.go
ywluogg: ValidateParamAgainstProperties understanding is correct.
ValidateObjectAgainstSchema is like:
```
ValidateObjectAgainstSchema(p PropertySpec, object ArraryOrString)
```

ywluogg: >  Just thinking whether to make this function parallel to the existing `validateParams` OR to utilizing `ValidateParamAgainstProperties` inside `validateParams` by first filtering out all object params and passing to `ValidateParamAgainstProperties`. And what is the second argument `object` of `ValidateObjectAgainstSchema`?

`ValidateParamAgainstProperties` is for later uses in Tekton Chains, and `ValidateObjectAgainstSchema` is for later usage in Results

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -107,17 +132,23 @@ type ArrayOrString struct {
 	Type      ParamType `json:"type"` // Represents the stored type of ArrayOrString.
 	StringVal string    `json:"stringVal"`
 	// +listType=atomic
-	ArrayVal []string `json:"arrayVal"`
+	ArrayVal  []string          `json:"arrayVal"`
+	ObjectVal map[string]string `json:"objectVal"`
```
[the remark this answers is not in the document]
ywluogg: Originally nested objects wouldn't be supported. Changing it to map[string]ArrayOrString can be a bit tricky for testing and capturing all the corner cases. This would be good stretches to be done after the feature is launched with alpha guard flags

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -38,40 +38,64 @@ type ParamSpec struct {
...
 	// Default is the value a parameter takes if no input value is supplied. If
 	// default is set, a Task may be executed without a supplied value for the
 	// parameter.
 	// +optional
 	Default *ArrayOrString `json:"default,omitempty"`
 }
 
+// PropertySpec defines the struct for object keys
+type PropertySpec struct {
+	Type ParamType `json:"type,omitempty"`
```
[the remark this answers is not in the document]
chuangw6: Good catch! This input indeed needs to be validated also.

The valid value for this type is `string` for now mentioned in this [part](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#defaulting-to-string-types-for-values) of the TEP75.

pkg/reconciler/taskrun/validate_resources.go
chuangw6: yeah, this will be refactored out to a separate function also discussed [here](https://github.com/tektoncd/pipeline/pull/4786#discussion_r863230699)

pkg/apis/pipeline/v1beta1/task_validation_test.go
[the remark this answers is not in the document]
chuangw6: > I think we need more tests where we don't rely on `Defaults` to set the values.

This [validation](https://github.com/tektoncd/pipeline/pull/4786/files/0ca9635e51663e4916a9732edf2b7affc1c3855b#diff-988b00fa4f188a96ff5901c0bbc9f65d4a8355417185d4fe0ae97d8c14e04417R363) is to only check if keys are missed iff the `default` section is provided ([code](https://github.com/tektoncd/pipeline/pull/4786/files/0ca9635e51663e4916a9732edf2b7affc1c3855b#diff-988b00fa4f188a96ff5901c0bbc9f65d4a8355417185d4fe0ae97d8c14e04417R351-R353)). 

With that said, it's not relying on `default` to set values for the object param. `pkg/reconciler/taskrun/validate_resources.go` will be responsible for check if neither taskrun or default provides a value for a param.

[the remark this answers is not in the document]
chuangw6: > +1 I think this PR is mixed with multiple stuff, for example, default values, but miss a big chunk of other validations. Maybe scope out this PR to a specific module / feature and that would make the PR easier to follow

For validation, this pr only includes task&taskrun level validation. Pipeline & pipeline level validation will be in future PRs.

For task&taskrun level validation, this PR includes most of validations except 
1. validate against entire object usage
> When providing values for strings, Task and Pipeline authors can access individual attributes of an object param; they cannot access the object as whole (we could add support for this later). (mentioned [here]((https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#using-objects-in-variable-replacement)) in the tep)

2. restrict the value of type in `PropertySpec` down to `string` only. ([mentioned](https://github.com/tektoncd/pipeline/pull/4786#discussion_r863921383) by billy)

Considering the amount of changes in this pr and amount of work for 1&2, I am thinking to add 1&2 in two different PRs later.

pkg/apis/pipeline/v1beta1/param_types.go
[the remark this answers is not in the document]
chuangw6: But I'll add this validation in a separate PR to avoid making this PR too big. How does that sound?

pkg/apis/pipeline/v1beta1/task_validation_test.go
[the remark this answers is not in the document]
chuangw6: will try to make future prs more modularized :)

pkg/apis/pipeline/v1beta1/param_types.go
chuangw6: IIUC, the tep proposes defaulting to string types for values. [section link](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#defaulting-to-string-types-for-values)

> a. Only string types would initially be supported b. Even when other types are supported, the most common usage of objects will likely use string values

So we might not want to pursue the recursive schema object for now.

pkg/apis/pipeline/v1beta1/param_types.go
chuangw6: In today's Pipeline WG meeting, @pritidesai proposed to create a new struct which embeds `ArrayOrString` and then includes `object` type to avoid breaking changes IICU.

My thoughts:
- If we create the new struct and keep `ArrayOrString` there, we still cannot avoid breaking changes unfortunately. For instance, we have to change the fields which were of `ArrayOrString` type i.e. `ParamSpec`'s [default](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/param_types.go#L52) and `Param`'s [value](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/param_types.go#L85).
- Renaming `ArrayOrString` might make more sense to me too. 

Can you please share more thoughts regarding this here? @pritidesai

pkg/apis/pipeline/v1beta1/param_types.go
Yongxuanzhang, line 136: Now we have 3 options for this struct:
1. Like this current code, we don't rename it
2. Rename it to something else
3. Create a new struct and wrap the ArrayOrString

I think we prefer for 2 right?

chuangw6: correct. see the discussion [here](https://github.com/tektoncd/pipeline/pull/4786#discussion_r862023838)

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -107,17 +132,23 @@ type ArrayOrString struct {
...
 }
 
 // UnmarshalJSON implements the json.Unmarshaller interface.
 func (arrayOrString *ArrayOrString) UnmarshalJSON(value []byte) error {
-	if value[0] == '"' {
+	switch value[0] {
+	case '"':
 		arrayOrString.Type = ParamTypeString
 		return json.Unmarshal(value, &arrayOrString.StringVal)
+	case '[':
```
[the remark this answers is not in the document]
Yongxuanzhang: Is it json array? It looks like yaml syntax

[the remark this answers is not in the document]
chuangw6: > Will this break multi-line arrays? e.g.
> 
> ```
> foo:
>   - a
>   - b
>   - c
> ```

no it will not break array type params.

pkg/apis/pipeline/v1beta1/param_types.go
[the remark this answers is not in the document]
chuangw6: @wlynch that works! Thanks!

pkg/reconciler/taskrun/validate_resources.go
chuangw6: `PropertySpec` is the spec for a ***specific*** key/property of the object param and this spec just contains the type of the key, whereas `ArraryOrString` contains `objectVal` which is a map containing many key-value pairs. They are not comparable not can be validated against each other.

I extracted object-specific validation into a new function called `ValidateObjectParameters` (@wlynch and you suggested).
Happy to split out this further once the result side gets clear needs for another new helper function on this part. Let me know what you think. @Yongxuanzhang @ywluogg

pkg/apis/pipeline/v1beta1/param_types.go
ywluogg: Yeah I think this is also similar to the comment about whether we want to support nested objects for the initial stage.

pkg/apis/pipeline/v1beta1/param_types.go
ywluogg: Ah that makes sense. I think this thread can be closed.

pkg/apis/pipeline/v1beta1/param_context.go
```diff
@@ -130,16 +134,26 @@ func getContextParams(ctx context.Context, overlays ...Param) []Param {
...
+			objVal := make(map[string]string)
+			for k := range ps.Properties {
+				objVal[fmt.Sprintf("$(params.%s.%s)", p.Name, k)] = ""
```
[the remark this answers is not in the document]
chuangw6: > Is it possible to pass through the whole object instead of each field individually? (thinking ahead to if/when objects are a mix of strings/arrays/objects)

sgtm!

As `ObjectVal` is a map, how about this - having only one entry in the `ObjectVal` and passing through the whole object by using the param name as the entry's key and using the whole object as the entry's value. Does it make sense?

pkg/apis/pipeline/v1beta1/task_validation_test.go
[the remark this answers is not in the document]
wlynch: > This validation is to check if keys are missed in default iff the default section is provided (code).

💯  Let's add a test to confirm the behavior and make sure that assumption doesn't break in the future without us knowing.

Perhaps also add tests for TaskRun validation with object params as well? Otherwise totally agree with breaking things up into multiple pull requests!

wlynch (commented):
/lgtm

Pending additional tests + possible rename of ArrayOrString?

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -107,17 +132,23 @@ type ArrayOrString struct {
...
 	// +listType=atomic
-	ArrayVal []string `json:"arrayVal"`
+	ArrayVal  []string          `json:"arrayVal"`
+	ObjectVal map[string]string `json:"objectVal"`
 }
 
 // UnmarshalJSON implements the json.Unmarshaller interface.
 func (arrayOrString *ArrayOrString) UnmarshalJSON(value []byte) error {
-	if value[0] == '"' {
+	switch value[0] {
```
Yongxuanzhang, line 141: I realized we probably need to check the value is not empty? Otherwise get index out of bound error

chuangw6: Seems like this [original code](https://github.com/tektoncd/pipeline/blob/a3caabf08751b3a8515bf2e8528ee728c49280ae/pkg/apis/pipeline/v1beta1/param_types.go#L113-L121) didn't check that. And I guess the reason is that no matter what users provided in yaml, k8s will convert it to json so `value` will not be empty even though users provided nothing for a field i.e. `default: `

Otherwise, can you show me an example yaml that might trigger index out of bound error?

Yongxuanzhang: I will add this in results's PR

pkg/reconciler/taskrun/validate_resources.go
Yongxuanzhang: I will keep this in mind

pkg/apis/pipeline/v1beta1/param_types.go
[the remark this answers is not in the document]
chuangw6: sounds great! go for it @wlynch! Thanks!

pkg/apis/pipeline/v1beta1/param_types.go
chuangw6: sg! if this is specific to Unmarshal issue on the result side, it seems to make more sense to add that check behaviour in relevant pr so that people can track why.

pkg/apis/pipeline/v1beta1/task_validation_test.go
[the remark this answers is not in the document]
chuangw6: > 💯 Let's add a test to confirm the behavior and make sure that assumption doesn't break in the future without us knowing.

Good test case! Just added `myobjWithoutDefault` in `TestTaskValidate` function, which is object type and default is not provided. This valid param test makes sure it doesn't rely on `default` to set values for the object param. 

> Perhaps also add tests for TaskRun validation with object params as well? Otherwise totally agree with breaking things up into multiple pull requests!

Good catch! After adding the alpha flag check, test coverage for `taskrun_validation.go` went down by 1.4% 🤣 . I just added new test to test duplicate params which are object type to bring it back 100%. I think this is the only test needed to add because `taskrun_validation.go` doesn't implement any validation against type alignment etc. Let me know otherwise.

pkg/apis/pipeline/v1beta1/param_types.go
chuangw6: Just agreed on that we'll solve this in another pr. @ywluogg @Yongxuanzhang

pkg/apis/pipeline/v1beta1/param_context.go
```diff
@@ -127,19 +131,26 @@ func getContextParams(ctx context.Context, overlays ...Param) []Param {
...
+		case ParamTypeArray:
 			p.Value = ArrayOrString{
 				Type:     ParamTypeArray,
 				ArrayVal: []string{fmt.Sprintf("$(params.%s[*])", ps.Name)},
 			}
+		default:
+			p.Value = ArrayOrString{
+				Type:      ParamTypeObject,
+				ObjectVal: map[string]string{ps.Name: fmt.Sprintf("$(params.%s[*])", ps.Name)},
+			}
```
ywluogg, line 153: I'm not sure why the old comment gets resolved, but we probably don't want to make the default to be setting the type to be ParamTypeObject. Please error this out. This is different from the other default case in another comment thread

pkg/apis/pipeline/v1beta1/param_types.go
chuangw6: https://github.com/tektoncd/pipeline/pull/4839 might be related!

pkg/apis/pipeline/v1beta1/param_context.go
chuangw6: please follow up on the original thread.

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -122,6 +122,53 @@ func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params
...
+	for p, keys := range providedObjectParamKeys {
+		if _, ok := neededObjectParamKeys[p]; !ok {
+			// Ignore any missing params - this happens when extra params were
+			// passed to the task that aren't being used.
+			continue
+		}
+		missedKeys := list.DiffLeft(neededObjectParamKeys[p], keys)
+		if len(missedKeys) != 0 {
+			missings = append(missings, p)
+		}
```
ywluogg, line 165: I realized the old comment gets outdated, but it shouldn't, but I think refactoring this part should be happening in this PR, rather than in the PR that results need it, as it is pretty straightforward that there is module that's needed in results side that a structure needs to be validated against ArraryOrString. ValidateObjectAgainstSchema(map[string]PropertySpec, object ArraryOrString) should be the one you should be looking for.

This should be a fairly quick restructure of the code that doesn't have to wait until results need to use this part. Let me know if you need help with that.

pkg/apis/pipeline/v1beta1/param_context.go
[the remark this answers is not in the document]
chuangw6: > > In the original implementation, the if-else statement also works like the default is array type instead of throwing error. So I guess it makes more sense to follow the the original implementation.
> 
> I'm fine with either of the implementation. Both LGTM!

Thought we'd agreed on following the original implementation in which no error is thrown in default - `else` because I don't see mutating webhook (setdefaults) which throws errors.

Can you elaborate with an example about how to throw errors?

pkg/reconciler/taskrun/validate_resources.go
chuangw6: Thought we agreed to further extract this helper function in next pr once result side gets clear requirement on this. (see the original thread)
I am cool with having this extracted in this pr. But there is some issues with the function signature

chuangw6: > ValidateObjectAgainstSchema(map[string]PropertySpec, object ArraryOrString) should be the one you should be looking for.

Thought we agreed to extract this helper function further in next pr once result side gets clear requirement on this (see the original thread). I am cool with having this extracted in this pr and just tried with the function signature you just suggested. 
`map[string]PropertySpec, object ArraryOrString` implies we will only find missing keys for ***one pair*** of needed vs provided. That means we have to do something like the following to validate for all params/results of object type in both current module and where results use this, which seems not that elegant.
```
for name, key set in required:
  find the name's corresponding key set in provided:
      call this helper to check missing keys
```

So I just make this function to accept two maps `needed` and `provided`. The map key will be the object name, and corresponding value will be a list of keys for that object.

Let me know if this what you actually need on the result side. @Yongxuanzhang @ywluogg Happy to change otherwise.

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -122,6 +122,59 @@ func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params
...
+// been checked against missing/extra object which been done in `validateParams` in this module.
+func ValidateObjectKeysNeededAgainstProvided(neededObjectKeys, providedObjectKeys map[string][]string) error {
+	missings := []string{}
+	for p, keys := range providedObjectKeys {
+		if _, ok := neededObjectKeys[p]; !ok {
+			// Ignore any missing objects - this happens when extra objects were
+			// passed that aren't being used.
+			continue
+		}
+		missedKeys := list.DiffLeft(neededObjectKeys[p], keys)
```
Yongxuanzhang, line 168: This seems to be a O(n^2) function and not so efficient .

pkg/reconciler/taskrun/validate_resources.go
Yongxuanzhang: I think this function should work, thanks @chuangw6! Left few questions regarding some details.

Yongxuanzhang: Actually I would prefer another implementation:
func ValidateObjectParameters(paramSpecs []v1beta1.ParamSpec, params []v1beta1.Param)
1. Store all 
2. 

```
func ValidateObjectParameters(paramSpecs []v1beta1.ParamSpec, params []v1beta1.Param) error {
	neededKeys := make(map[string][]string)
	// collect needed keys for object parameters
	for _, spec := range paramSpecs {
		if spec.Type == v1beta1.ParamTypeObject {
			for key := range spec.Properties {
				neededKeys[spec.Name] = append(neededKeys[spec.Name], key)
			}
		}
	}

	// collect provided keys for object parameters
	for _, p := range params {
		if p.Value.Type == v1beta1.ParamTypeObject {
			for key := range p.Value.ObjectVal {
				if err:=ValidateObjectAgainstSchema(neededKeys[p.Name], p.Value.ObjectVal);err!=nil{
                                       return err
                                  }
			}
		}
	}

func ValidateObjectAgainstSchema(needKeys[]string, object map[string]string){
     // one loop to store object keys in sets/map
    // one loop on needKeys and if not in sets/map return error

}
```

pkg/reconciler/taskrun/validate_resources.go
chuangw6: This `list.diffleft` is existing implemented function that has been used in [multiple places](https://github.com/tektoncd/pipeline/search?q=list.diffleft) to compare needed vs provided.

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -122,6 +122,59 @@ func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params
...
+		if len(missedKeys) != 0 {
+			missings = append(missings, p)
+		}
+	}
+
+	if len(missings) != 0 {
+		return fmt.Errorf("object provider misses providing the keys required for %v", missings)
+	}
```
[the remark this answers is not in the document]
chuangw6: Because we want to give users more informative error message. For example, you have object param `a`, `b` .. `d`and all of them have keys missing in provider (from taskrun or result). If we did what you just said and directly return error, what will happen is that once key missing is detected for `a`, error will return and tell users "hey, parameter a has missing keys, please fix it", and users realize it and change it and apply their yaml. In the second round, they received error message again "hey param b has missing keys, please fix again.", so on and so forth. You can see, if many object params have missing keys, users need to apply the yaml many many times in order to get what params have missing keys.

I don't think this is user-friendly error handling. So in this implementation, we collect all params once which have missing keys, and tell users  "hey, these object param [a, b, c, d] have missing keys, please fix them".

pkg/reconciler/taskrun/validate_resources.go
Yongxuanzhang: I see why we use it, it is a good way to find out all missing keys, also make sense

pkg/reconciler/taskrun/validate_resources.go
chuangw6: I think this is related to this [comment](https://github.com/tektoncd/pipeline/pull/4786#discussion_r866935643). see response there. let me know if it doesn't make sense.

pkg/apis/pipeline/v1beta1/param_context.go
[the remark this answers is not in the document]
ywluogg: Yes let's let other places to catch this error - at least don't set the value to be an object at default. Or am I looking at an old snapshot? You can just set the type to be ps.Type at default, and don't add anything for the value. This way some other module should catch this issue

pkg/apis/pipeline/v1beta1/param_types.go
ywluogg: Thanks for updating this. Can you add a TODO for it? And you can mark this as resolved

pkg/reconciler/taskrun/validate_resources.go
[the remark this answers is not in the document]
Yongxuanzhang: Yes this makes sense! I think it is a trade-off, like whether we want to save more user's time like this. And throw err in advance will not have significant improvement in our case

pkg/reconciler/taskrun/validate_resources.go
Yongxuanzhang: But this func is not implemented time efficient, maybe we want to save more memory...

pkg/reconciler/taskrun/validate_resources.go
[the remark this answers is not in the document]
chuangw6: > And throw err in advance will not have significant improvement in our case
exactly

pkg/reconciler/taskrun/validate_resources.go
chuangw6: > But this func is not implemented time efficient, maybe we want to save more memory...

are you referring to `list.diffleft` or something else?

Yongxuanzhang: yes

pkg/apis/pipeline/v1beta1/param_context.go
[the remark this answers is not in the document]
chuangw6: > At least don't set the value to be an object at default. Or am I looking at an old snapshot?

Only three possible types: string, array, object. And string & array have been caught in the first two cases and only object type is left. Did I miss some case?

I guess you're referring to unmarshal snippet in this [comment](https://github.com/tektoncd/pipeline/pull/4786#discussion_r863232810) where yongxuan requested default case is string for the purpose of result side. And that has been changed to default to string.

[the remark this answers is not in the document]
chuangw6: > At least don't set the value to be an object at default. Or am I looking at an old snapshot?

Only three possible types: string, array, object. And string & array have been caught in the first two cases and only object type is left. Did I miss some case?

I guess you're referring to unmarshal snippet in this [comment](https://github.com/tektoncd/pipeline/pull/4786#discussion_r863232810) where yongxuan requested default case is string for the purpose of result side. And that has been changed to default to string.

pkg/reconciler/taskrun/validate_resources.go
ywluogg: I think Yongxuan's design for ValidateObjectAgainstSchema() would make more sense, as it gives flexibility that we don't have to bundle all params together. But what I was thinking is like this:
```
func ValidateObjectParameters(paramSpecs []v1beta1.ParamSpec, params []v1beta1.Param) error {
       // this is a map of (pointers to a map)
	neededKeys := make(map[string]map[]*PropertySpec)
	// collect needed keys for object parameters
	for _, spec := range paramSpecs {
		if spec.Type == v1beta1.ParamTypeObject {
                         neededProperties[spec.Name] = &spec.Properties
		}
	}

	// validate param's objectval
	for _, p := range params {
		if p.Value.Type == v1beta1.ParamTypeObject {
				if err:=ValidateObjectAgainstSchema(neededKeys[p.Name], p.Value.ObjectVal);err!=nil{
                                       return err
			}
		}
	}

func ValidateObjectAgainstSchema(p map[]*PropertySpec, a *ArrayOrString){
     // if p is nil, means a is defined, but p doesn have it
    // if p is not nil, blah...

}
```

chuangw6: what do you mean by bundling all params together? See this [comment](https://github.com/tektoncd/pipeline/pull/4786#discussion_r866935643) which is related to the design for this function.

Yongxuanzhang: @ywluogg I think @chuangw6's current implementation can help to capture all missing keys and also aligned with code base, so it should be ok! We should move forward see if we can merge this pr asap

pkg/reconciler/taskrun/validate_resources.go
chuangw6: i see. that's a pre-existing utility function in the code base. Let's look into it later.

pkg/apis/pipeline/v1beta1/param_context.go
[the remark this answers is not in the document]
ywluogg: This should be:
```
case ParamTypeString:
...
case ParamTypeArray:
...
case ParamTypeObject:
...
default:
// set type to string
```
Ideally the default case should be throwing errors to catch this `ideally not should be happening` issue, but I agree that I don't see a good way to throw this error, but the cases should be separate.

pkg/reconciler/taskrun/validate_resources.go
ywluogg: I don't see the reason for merging this ASAP as a good reason to stay on a complicated design of piece of code.

To achieve capturing all the keys, the design can be the following:
```
func ValidateObjectParameters(paramSpecs []v1beta1.ParamSpec, params []v1beta1.Param) error {
       // this is a map of (pointers to a map)
	neededKeys := make(map[string]*map[string]PropertySpec)
	// collect needed keys for object parameters
	for _, spec := range paramSpecs {
		if spec.Type == v1beta1.ParamTypeObject {
                         neededProperties[spec.Name] = &spec.Properties
		}
	}

      missed :=[]

	// validate param's objectval
	for _, p := range params {
		if p.Value.Type == v1beta1.ParamTypeObject {
				if m := ValidateObjectAgainstSchema(neededKeys[p.Name], p.Value); len(missedkeys) !=0 {
                                       missed = append(missed, m)
			}
		}
	}

      if len(missed) != 0 {
       // throw error together
      }

func ValidateObjectAgainstSchema(p *map[string]PropertySpec, a *ArrayOrString) []string {
     // if p is nil, means a is defined, but p doesnt have it
    // if p is not nil, blah...
  // at the end, return all missed keys for this single object
}
```

pkg/apis/pipeline/v1beta1/param_context.go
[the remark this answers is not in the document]
dibyom: so in general I agree with @ywluogg that we should error out for unexpected scenarios but I think in this case this function is being called as part of the mutating webhook? If this is a general purpose function that is reused elsewhere - if so we could return an error from this function but handle it properly in the webhook (i.e. set the default). 

(I might be missing context here so let me know if I'm wrong!)

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -122,6 +122,59 @@ func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params
...
 	}
 
+	return ValidateObjectParameters(paramSpecs, params)
+}
+
+// ValidateObjectParameters checks if all required keys of object type params are provided in taskrun params.
+// To use this function externally, make sure `paramSpecs` and `params` have
+// been checked against general missing/extra params and param type alignment,
+// which been done in `validateParams` in this module.
+func ValidateObjectParameters(paramSpecs []v1beta1.ParamSpec, params []v1beta1.Param) error {
```
ywluogg, line 132: Same logic applies here what if somewhere else wants to validate against one single param instead of the whole param list? I'm not sure if your module is providing that flexibility. You should put the loops in `validateParams`, and have:
```
ValidateObjectParameters(ps *v1beta1.ParamSpec, p *v1beta1.Param) {

}
...
func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params []v1beta1.Param) error {
... 
missedkeys := []
for _, spec := range paramSpecs {
		if spec.Type == v1beta1.ParamTypeObject {
			m = ValidateObjectParameters()
                         if len(m) != 0 {
                           missedkeys.append(m)
                         }
		}
	}
}
// error for all params together

Comment (chuangw6) [reply pkg/reconciler/taskrun/validate_resources.go:132]: Ok, I see the point you're trying to make.

But wouldn't that be too specific? I don't see any scenario in the current code base which is to validate against one single param. I might miss that if there is one, or if you have some scenarios in mind which might need to validate against a single case in future, can you please explain a bit more?

More importantly, to support what you just mentioned - validate against a single item, do you think the existing functions like [validateParams](https://github.com/tektoncd/pipeline/blob/30bed9f1c4fb834baf1cf12fd531b472821a3ce1/pkg/reconciler/taskrun/validate_resources.go#L72-L126), [validateResources](https://github.com/tektoncd/pipeline/blob/30bed9f1c4fb834baf1cf12fd531b472821a3ce1/pkg/reconciler/taskrun/validate_resources.go#L33-L70) should all be changed as well ?

Comment (chuangw6) [reply pkg/reconciler/taskrun/validate_resources.go:132]: Just remember that @wlynch also [commented](https://github.com/tektoncd/pipeline/pull/4786#discussion_r863916027) before that we want to extract all object-specific validation out to not interleave and avoid context switching for readers. I'll let @wlynch to chime in to get more inputs on this :)

Comment (ywluogg) [reply pkg/reconciler/taskrun/validate_resources.go:132]: > 

One use case is that in Chains, we probably wouldn't match each params or objects in groups. They could be matched against each other. When you have module with single item, it's super easy to extend another module to validate against all the params together, and that way it's super easy to wrap those loops within another layer of module. I don't see the problem there

Comment (ywluogg) [reply pkg/reconciler/taskrun/validate_resources.go:132]: > But wouldn't that be too specific? 

I'm not sure about this is being too specific, as this is just providing a flexible module that can be easily extended when we want another level of aggregation.

File: pkg/apis/pipeline/v1beta1/param_context.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (chuangw6) [reply pkg/apis/pipeline/v1beta1/param_context.go:149]: > so in general I agree with @ywluogg that we should error out for unexpected scenarios but I think in this case this function is being called as part of the mutating Webhook?

Agreed! this is part of the mutating webhook.



> This should be:
> 
> ```
> case ParamTypeString:
> ...
> case ParamTypeArray:
> ...
> case ParamTypeObject:
> ...
> default:
> // set type to string
> ```
> 

So do you mean it'd better if we treat the `ParamTypeString ` case as the default (not throwing error)? If so, that sgtm!

File: pkg/reconciler/taskrun/validate_resources.go (hunk 19 above)
Comment (ywluogg) [reply pkg/reconciler/taskrun/validate_resources.go:132]: @dibyom @vdemeester @pritidesai Hi team we probably need more thoughts about making these as sharable modules.

File: pkg/apis/pipeline/v1beta1/param_context.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (ywluogg) [reply pkg/apis/pipeline/v1beta1/param_context.go:149]: Yeah, because I don't think we want to make default to be objects, and this should be better aligned with https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#defaulting-to-string-types-for-values consistently

File: pkg/reconciler/taskrun/validate_resources.go (hunk 19 above)
Comment (chuangw6) [reply pkg/reconciler/taskrun/validate_resources.go]: update: just update the code to align with the refactor for `validateParams` function that @jerop 's made in #4814. 

Can you please have a look at the [objectParam-specific functions](https://github.com/tektoncd/pipeline/pull/4786/files#diff-f6ff4d26a45baec7685e7ad851e9d7fa5901e719b4635d2c7de0298d08b8a71dR150-R198) which is just added and is also part of `validateParams`? @jerop Let me know what you think. Thanks!

Comment (wlynch) [reply pkg/reconciler/taskrun/validate_resources.go]: My 2c - I'm a little skeptical if we would actually need to validate params again in chains? If we can validate that a message came from the controller via spire, would there be a need to re-validate param fields? lmk off-thread if you have something else in mind - would be happy to chat more about this. 👀 

Out of scope of this PR, but in general I think we may be exporting too much internal validation out through the package? e.g. idk if anyone besides the controller would call something like [`ValidateParameterTypes`](https://pkg.go.dev/github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1#ValidateParameterTypes) - I'd expect most users that care about validation to just call through the Validatable interfaces on the CRD types.

For this PR - let's try and unexport as much as possible. This gives us flexibility to make changes with how validation is structured without needing to worry about breaking downstream clients (i.e. we can rewrite this later if we want).

Comment (chuangw6) [reply pkg/reconciler/taskrun/validate_resources.go]: > For this PR - let's try and unexport as much as possible. This gives us flexibility to make changes with how validation is structured without needing to worry about breaking downstream clients (i.e. we can rewrite this later if we want).

Done. Just unexported them.

Agree. Let's leave the function signature as is to align with [other param-related validation functions](https://github.com/tektoncd/pipeline/blob/e586418e798ae9fcc5193d1ed09b810d912bca1b/pkg/reconciler/taskrun/validate_resources.go#L107-L145). We can rewrite it later if needed.

Comment (ywluogg) [reply pkg/reconciler/taskrun/validate_resources.go]: > would there be a need to re-validate param fields? lmk off-thread if you have something else in mind - would be happy to chat more about this. 👀
> 
> Out of scope of this PR, but in general I think we may be exporting too much internal validation out through the package? e.g. idk if anyone besides the controller would call somethin

Can you elaborate a bit more about validating via spire? I think in Tekton Chains, when we have a defined schema, we still need to validate the objects against the defined schema.

I think it makes sense to export the function here, as at least in results there is a clear need to for such sharable module. I do agree that we can rewrite this piece of code later, but when I see a clear need of a piece of code, I think it worths thinking and discussing  about it now rather than later.

File: pkg/apis/pipeline/v1beta1/task_validation.go [hunk 20]
```diff
@@ -268,23 +274,58 @@ func (p ParamSpec) ValidateType() *apis.FieldError {
...
+
+// ValidateObjectType checks that object type parameter does not miss the
+// definition of `properties` section and the type of a PropertySpec is allowed.
+// (Currently, only string is allowed)
+func (p ParamSpec) ValidateObjectType() *apis.FieldError {
+	if p.Type == ParamTypeObject && p.Properties == nil {
+		return &apis.FieldError{
+			Message: "The parameter of object type misses the definition of `properties`",
+			Paths:   []string{fmt.Sprintf("%s.properties", p.Name)},
+		}
```
wlynch, line 290:
```suggestion
		return &apis.ErrMissingField(fmt.Sprintf("%s.properties", p.Name))
```

wlynch (commented):
/lgtm

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -38,39 +38,63 @@ type ParamSpec struct {
...
+	switch {
+	case pp.Type != "":
+		// If param type is provided by the author, do nothing but just set default type for PropertySpec in case `properties` section is provided.
+		pp.setDefaultsForProperties(ctx)
+	case pp.Properties != nil:
+		pp.Type = ParamTypeObject
+		// Also set default type for PropertySpec
+		pp.setDefaultsForProperties(ctx)
+	case pp.Default == nil:
+		pp.Type = ParamTypeString
```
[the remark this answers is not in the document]
chuangw6: sgtm! Thanks for adding the comment. @wlynch

pkg/apis/pipeline/v1beta1/task_validation.go
chuangw6: Done!! Thanks @wlynch.

ywluogg (approved):

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -154,6 +157,51 @@ func wrongTypeParamsNames(params []v1beta1.Param, matrix []v1beta1.Param, needed
 	return wrongTypeParamNames
 }
 
+// missingKeysObjectParamNames checks if all required keys of object type params are provided in taskrun params.
+// TODO: This might be refactored out to support single pair validation.
```
dibyom, line 161: Same as above

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -93,31 +118,40 @@ type ParamType string
...
 // AllParamTypes can be used for ParamType validation.
-var AllParamTypes = []ParamType{ParamTypeString, ParamTypeArray}
+var AllParamTypes = []ParamType{ParamTypeString, ParamTypeArray, ParamTypeObject}
 
 // ArrayOrString is modeled after IntOrString in kubernetes/apimachinery:
 
 // ArrayOrString is a type that can hold a single string or string array.
 // Used in JSON unmarshalling so that a single JSON field can accept
 // either an individual string or an array of strings.
+// TODO: This struct will be renamed or be embedded in a new struct to take into
```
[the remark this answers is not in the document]
chuangw6: > Looks like we are going to rename the struct based on the next resolved comment - any reason to not do it in this PR itself?

I think @ywluogg has sent a headsup email to the community asking for inputs. So we think it might be better to do this a bit later to make sure this rename (or creating new struct) doesn't break others who are using the old struct name?

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -320,6 +358,52 @@ func ValidateResourcesVariables(steps []Step, resources *TaskResources) *apis.Fi
 	return validateVariables(steps, "resources.(?:inputs|outputs)", resourceNames)
 }
 
+// TODO: Make sure an object param is not used as a whole when providing values for strings.
```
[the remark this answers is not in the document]
chuangw6: done! thanks @dibyom

pkg/apis/pipeline/v1beta1/param_types.go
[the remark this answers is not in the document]
dibyom: Ok that's fine

dibyom:
/lgtm

All comments have been addressed

pkg/apis/pipeline/v1beta1/param_types.go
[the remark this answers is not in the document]
ywluogg: > I think @ywluogg has sent a headsup email to the community asking for inputs. So we think it might be better to do this a bit later to make sure this rename (or creating new struct) doesn't break others who are using the old struct name.

I created this [issue](https://github.com/tektoncd/pipeline/issues/4858) for following up the issue

pritidesai:
I am sorry to bring this up @wlynch @dibyom 🙏 

This PR is introducing a new feature and was merged without release notes. Please make sure we check the PR description before its merged!

vdemeester:
I've updated the description to add a release-note block.

PR #4788, opened 2022-04-25
pkg/apis: export more `Task` validation functions 🤹

vdemeester:
# Changes

The reason to export those is mainly to be able to validate
Tasks *ahead* of time in clients (such as `tkn`) and/or to re-use
those function in Custom Tasks if possible.

The current use case I have for those is `TaskGroup`, that *wraps* a
Task, and thus tries to perform the same set of validation.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/task_validation.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation.go
@@ -49,8 +49,8 @@ func (ts *TaskSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 		errs = errs.Also(apis.ErrMissingField("steps"))
 	}
 	errs = errs.Also(ValidateVolumes(ts.Volumes).ViaField("volumes"))
-	errs = errs.Also(validateDeclaredWorkspaces(ts.Workspaces, ts.Steps, ts.StepTemplate).ViaField("workspaces"))
-	errs = errs.Also(validateWorkspaceUsages(ctx, ts))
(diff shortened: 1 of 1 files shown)


afrittoli (commented):
A comment in the code would help understand why these are exposed, since they are not consumed anywhere within `pipeline`.  Else you risk the next maintainer changing this back at some point.

imjasonh:
Re: ahead-of-time validation, what should clients like `tkn` do when their vendored/linked-in validation code doesn't match with the installed Tekton version's? It's potentially a problem if either `tkn` is stricter (valid Tasks get rejected) or if `tkn` is looser (invalid Tasks get sent, disjoint error experience).

Not saying we shouldn't do it, just wondering if there are any big issues with validation version skew we anticipate.

vdemeester:
> Re: ahead-of-time validation, what should clients like `tkn` do when their vendored/linked-in validation code doesn't match with the installed Tekton version's? It's potentially a problem if either `tkn` is stricter (valid Tasks get rejected) or if `tkn` is looser (invalid Tasks get sent, disjoint error experience).
> 
> Not saying we shouldn't do it, just wondering if there are any big issues with validation version skew we anticipate.

I agree but I think it's not necessarily only for clients but also extensions and/or alternative implementation. For example, with `buildkit-tekton` I probably want to add validation very similar to the version I support — today, without this, I need to duplicate code a bit (and that code might be wrong later on). It could also be true for remote resolver or custom task that might "wrap" existing Tekton objects. An example is the [TaskGroup](https://github.com/openshift-pipelines/tekton-task-group) custom task.
Maybe giving the example of `tkn` here was the wrong one 😛

PR #4799, opened 2022-04-26
Validate parameter name format

### LINKED ISSUE #4792: Missing validation against parameter name format
# Expected Behavior
[This doc](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters) mentioned that the parameter names should follow the following rules:
- Must only contain alphanumeric characters, hyphens (-), underscores (_), and dots (.).
- Must begin with a letter or an underscore (_).
> For example, foo.Is-Bar_ is a valid parameter name, but barIsBa$ or 0banana are not.

Therefore, there should be existing validation against parameter names that users specified. For example, the following parameter names are invalid, which should be caught and reported by the admission webhook when the taskrun yaml is applied.
```yaml
kind: TaskRun
metadata:
  generateName: validation-
spec:
  taskSpec:
    params:
      - name: 0foo
        default: "bar1"
      - name: a b
        default: "bar2"
      - name: $x
        default: "bar3"
    results:
      - name: echo-output
      - name: bad-echo
        description: "bad echo behaviour"
    steps:
      - name: echo-params
        image: bash
        script: |
          set -e
          echo $(params.0foo) $(params.a b) $(params.$x) | tee $(results.echo-output.path)
          echo $(params.0fo) | tee $(results.bad-echo.path)
```
# Actual Behavior
`taskResults` output
```
  taskResults:
  - name: bad-echo
    value: |2+

  - name: echo-output
    value: |
      bar1 bar2 bar3
```

1. When you apply the above taskrun yaml, it can be successfully applied, which should've been reported because of invalid param name format instead.
2.  From taskrun result, `echo-output` is populated with values `bar1 bar2 bar3`, which should not happen because the parameter names are all invalid i.e. `0foo` starts with 0, `a b` has a space in the middle, `$x` starts with $
3. In addition, the taskrun result `bad-echo` is shown in the `taskResults` list under status, but has some confusing value `+2`.  (See the reason why this happens [here](https://github.com/tektoncd/pipeline/issues/4792#issuecomment-1109397484))

1. Save above yaml code into a file, and apply it
2. see the taskrun
`kubectl get tr <TASKRUN_NAME> -o yaml`
3. find status -> taskResults 

# Additional Info

## Kubernetes version:

  ```
  Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.1", GitCommit:"86ec240af8cbd1b60bcc4c03c20da9b98005b92e", GitTreeState:"clean", BuildDate:"2021-12-16T11:41:01Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"linux/amd64"}
  Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.6-gke.1503", GitCommit:"2c7bbda09a9b7ca78db230e099cf90fe901d3df8", GitTreeState:"clean", BuildDate:"2022-02-18T03:17:45Z", GoVersion:"go1.16.9b7", Compiler:"gc", Platform:"linux/amd64"}
  WARNING: version difference between client (1.23) and server (1.21) exceeds the supported minor version skew of +/-1
  ```

## Tekton Pipeline version:

  ```
  Client version: 0.21.0
  Pipeline version: devel
  ```

## Also found that a param without a name also works, which is weird 
```yaml
kind: TaskRun
metadata:
  generateName: validation-
spec:
  taskSpec:
    params:
      - name:
        default: "bar1"
    results:
      - name: echo-output
    steps:
      - name: echo-params
        image: bash
        script: |
          set -e
          echo $(params.) | tee $(results.echo-output.path)
```

chuangw6:
# Changes
Fixes #4792

[This doc](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters) mentioned that the parameter names should follow the following rules:
- Must only contain alphanumeric characters, hyphens (-), underscores (_), and dots (.).
- Must begin with a letter or an underscore (_).
> For example, foo.Is-Bar_ is a valid parameter name, but barIsBa$ or 0banana are not.

But the current code has not implemented the validation against the name format, which can cause some problems mentioned in #4792 .

In this PR, we enforce the rules. If invalid name format is found, validation webhook will complain.

# Submitter Checklist

  functionality, content, code)

# Release Notes

### DIFF
--- a/pkg/apis/pipeline/v1beta1/task_validation.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
@@ -33,6 +34,8 @@ import (
 )
 
 var _ apis.Validatable = (*Task)(nil)
+
+const variableNameFormat = "^[_a-zA-Z][_a-zA-Z0-9.-]*$"
 
 // Validate implements apis.Validatable
 func (t *Task) Validate(ctx context.Context) *apis.FieldError {
@@ -351,6 +354,28 @@ func validateStepArrayUsage(step Step, prefix string, vars sets.String) *apis.Fi
 }
 
 func validateVariables(steps []Step, prefix string, vars sets.String) (errs *apis.FieldError) {
+	// validate that the variable name format follows the rules
+	// - Must only contain alphanumeric characters, hyphens (-), underscores (_), and dots (.)
+	// - Must begin with a letter or an underscore (_)
+	re := regexp.MustCompile(variableNameFormat)
+	invalidNames := []string{}
+	// Converting to sorted list here rather than just looping map keys
+	// because we want the order of items in vars to be deterministic for purpose of unit testing
+	for _, name := range vars.List() {
+		if !re.MatchString(name) {
+			invalidNames = append(invalidNames, name)
+		}
+	}
+
+	if len(invalidNames) != 0 {
+		return &apis.FieldError{
+			Message: fmt.Sprintf("The format of following variable names is invalid. %s", invalidNames),
+			Paths:   []string{"params"},
+			Details: "Names: \nMust only contain alphanumeric characters, hyphens (-), underscores (_), and dots (.)\nMust begin with a letter or an underscore (_)",
+		}
+	}
+
+	// We've checked param name format. Now, we want to check if param names are referenced correctly in each step
 	for idx, step := range steps {
 		errs = errs.Also(validateStepVariables(step, prefix, vars).ViaFieldIndex("steps", idx))
 	}
--- a/pkg/apis/pipeline/v1beta1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation_test.go
@@ -481,6 +481,35 @@ func TestTaskSpecValidateError(t *testing.T) {
 			Paths:   []string{"resources.outputs.name"},
 		},
 	}, {
+		name: "invalid param name format",
+		fields: fields{
+			Params: []v1beta1.ParamSpec{{
+				Name:        "_validparam1",
+				Description: "valid param name format",
+			}, {
+				Name:        "valid_param2",
+				Description: "valid param name format",
+			}, {
+				Name:        "",
+				Description: "invalid param name format",
+			}, {
+				Name:        "a^b",
+				Description: "invalid param name format",
+			}, {
+				Name:        "0ab",
+				Description: "invalid param name format",
+			}, {
+				Name:        "f oo",
+				Description: "invalid param name format",
+			}},
+			Steps: validSteps,
+		},
+		expectedError: apis.FieldError{
+			Message: fmt.Sprintf("The format of following variable names is invalid. %s", []string{"", "0ab", "a^b", "f oo"}),
+			Paths:   []string{"params"},
+			Details: "Names: \nMust only contain alphanumeric characters, hyphens (-), underscores (_), and dots (.)\nMust begin with a letter or an underscore (_)",
+		},
+	}, {
 		name: "invalid param type",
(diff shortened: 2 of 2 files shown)


pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -359,6 +362,29 @@ func validateStepArrayUsage(step Step, prefix string, vars sets.String) *apis.Fi
...
+	if len(invalidNames) != 0 {
+		return &apis.FieldError{
+			Message: fmt.Sprintf("The format of following variable names is invalid. %s", invalidNames),
+			// Empty path is required to make the `ViaField`, … work
+			Paths:   []string{""},
+			Details: "Names: \nMust only contain alphanumeric characters, hyphens (-), underscores (_), and dots (.)\nMust begin with a letter or an underscore (_)",
+		}
+	}
+
+	// We've checked param name format. Now, we want to check if param names are referenced correctly in each step
```
chuangw6, line 387: Adding validation here just checks the name format of parameters declared under `taskSpec`. 
```yaml
kind: TaskRun
metadata:
  generateName: name-format-
spec:
  taskSpec:
    params:
      - name: foo
        default: "lower"
    results:
      - name: echo-output
    steps:
      - name: echo-params
        image: bash
        script: |
          set -e
          echo $(params.0foo) | tee $(results.echo-output.path)
```

- When you echo with a non-existing param name but the name follows the [rules](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters) , the validation can detect it's referencing a non-existing param and returns an error when the yaml file is applied.
- But when you echo with a non-existing param name but the name does not follow the [rules](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-parameters), the validation cannot detect it's referencing a non-existing param, and weird behaviour will still happen mentioned in the issue #4792. see the above example in yaml


So there are two tasks to solve #4792:
1. the code that validates against format, which has been implemented by the current code chunk 
2. the current code chunk can just live here or be moved to the beginning of [ValidateVariableP](https://github.com/tektoncd/pipeline/blob/f0f9e32da6b4145ede27a371900e853a3530c1ee/pkg/substitution/substitution.go#L48-L63).
3. now all parameter names format is valid. (if invalid names are found, errors will be returned)
4. change [extractVariablesFromString](https://github.com/tektoncd/pipeline/blob/f0f9e32da6b4145ede27a371900e853a3530c1ee/pkg/substitution/substitution.go#L149-L165) to make sure it extract ***ALL*** chars in `<NAME>` part of`$(params.<NAME>)` no matter if it follows the rules.  (Currently, it only extracts names that follow the rules because of this [regex](https://github.com/tektoncd/pipeline/blob/68f2a66c9b2ffcc23ef262bac95d1eb3856021cf/pkg/substitution/substitution.go#L28))
5. And then just check if the extracted string exists in the set [ValidateVariableP](https://github.com/tektoncd/pipeline/blob/f0f9e32da6b4145ede27a371900e853a3530c1ee/pkg/substitution/substitution.go#L48-L63).

chuangw6 (commented):
additional work to be done with [extractVariablesFromString](https://github.com/tektoncd/pipeline/blob/f0f9e32da6b4145ede27a371900e853a3530c1ee/pkg/substitution/substitution.go#L149-L165) function.

pkg/apis/pipeline/v1beta1/task_validation.go
chuangw6: step 2 is implemented in #4833 !

vdemeester (approved):

pkg/apis/pipeline/v1beta1/task_validation.go
chuangw6: step 2 is solved in this pr https://github.com/tektoncd/pipeline/pull/4857

dibyom (commented):
/lgtm

PR #4803, opened 2022-04-27
Fix variable interpolation on `StepTemplate`

### LINKED ISSUE #4801: Variable interpolation doesn't work in `stepTemplate`
# Expected Behavior

Given the following `TaskRun` (using embedded `Task`, but it's the same with `Task` references), it fails with `InvalidImageName`, as the image name passed to kubernetes is `$(params.image)`(aka not interpolated).

```
kind: TaskRun
metadata:
  generateName: steptemplate-variables-
spec:
  params:
  - name: image
    value: ubuntu
  taskSpec:
    params:
    - name: image
    stepTemplate:
      image: $(params.image)
    steps:
    - name: foo
      script: |
        echo hello from $(params.image)
```

# Actual Behavior

It resolves the param value and "just" work.

1. Create a `TaskRun` similar to above
2. Apply it
3. See it fail

# Additional Info

- Kubernetes version:

```
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:01:33Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

```
Pipeline version: v0.35.0
```

vdemeester:
# Changes

This changes the way we interpolate variables by using `StepTemplate`
directly instead of modifying a de-referenced version of it.

Dereferencing means that changes happening to it (in
`ApplyStepReplacements`) are not reflected to the original object.
Before this change, using variable interpolation on `StepTemplate`
would only work on part of the fields they were supposed to be.
For example, it works on `env` because, it is an array and the
de-referenced version of `StepTemplate` would share the same pointer
to it, making any changes on the array reflect to the initial object.

This is now fixed.

/kind bug
/cc @imjasonh @abayer @jerop 

Fixes #4801 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Variable are now correctly interpolated on `stepTemplate` field for `Task`
```

### DIFF
--- a/examples/v1beta1/taskruns/steptemplate-variable-interop.yaml
+++ b/examples/v1beta1/taskruns/steptemplate-variable-interop.yaml
@@ -1,0 +1,17 @@
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  generateName: steptemplate-variables-
+spec:
+  params:
+  - name: image
+    value: ubuntu
+  taskSpec:
+    params:
+    - name: image
+    stepTemplate:
+      image: $(params.image)
+    steps:
+    - name: foo
+      script: |
+        echo hello from $(params.image)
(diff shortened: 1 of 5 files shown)


afrittoli (commented):
Thanks for this, nice catch!
/approve

PR #4807, opened 2022-04-27
Set git-clone tasks in examples to run as root

abayer:
# Changes

Similar to https://github.com/tektoncd/pipeline/commit/ba2e7f3655d57d3ba3921462f21ece7f86b31488 - with https://github.com/tektoncd/pipeline/pull/4758, `git-init` now uses `ghcr.io/distroless/git` as its base image, and that image needs to run as root. `git-init` by default does _not_ run as root. So the two examples using copy-pasted old versions of the `git-clone` catalog task need to be changed to run as root, or the example tests will keep failing forever.

An equivalent change will be needed in the `git-clone` catalog task once it's bumped to using `git-init:v0.35.0` or later - currently, it's using `v0.29.0`.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
+++ b/examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml
@@ -58,6 +58,8 @@ spec:
   steps:
     - name: clone
       image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:latest
+      securityContext:
+        runAsUser: 0 # This needs root, and git-init is nonroot by default
       script: |
         CHECKOUT_DIR="$(workspaces.output.path)/$(params.subdirectory)"
 
--- a/examples/v1beta1/pipelineruns/pipelinerun.yaml
+++ b/examples/v1beta1/pipelineruns/pipelinerun.yaml
@@ -81,6 +81,8 @@ spec:
   steps:
   - name: clone
     image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:latest
+    securityContext:
+      runAsUser: 0 # This needs root, and git-init is nonroot by default
     script: |
       CHECKOUT_DIR="$(workspaces.output.path)/$(params.subdirectory)"
       cleandir() {


abayer:
@mattmoor - I think this is what you fixed in `test/yamls/v1beta1/pipelineruns/*.yaml` in https://github.com/tektoncd/pipeline/commit/ba2e7f3655d57d3ba3921462f21ece7f86b31488, which we wouldn't see in the examples tests until after a release, since they're using `image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:latest`, while the yamls are using `image: ko://github.com/tektoncd/pipeline/cmd/git-init` (which is why you hit this problem when you changed the base image for `git-init` in `TestYamls` but not in `TestExamples`). 

I assume this means we'll need to update the `git-clone` task in the catalog as well, once it's updated to use `gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.35.0` (or later) as the default (it's currently using `gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/git-init:v0.29.0`).

imjasonh (commented):
/lgtm

lbernick:
Thanks for fixing this!
I understand that this is to unblock current PRs from failing integration tests, but would it make sense to update our [git init dockerfile](https://github.com/tektoncd/pipeline/blob/main/images/git-init/Dockerfile) to run as root? This would avoid breaking users of this image, which we currently list as [part of the API](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#what-is-the-api) (and we should probably backport it to the release we just did). We could have the examples run an earlier version of the image until the image is fixed.

abayer:
@lbernick I'm going to defer to @mattmoor on that, since he opened/fixed #4711. =)

mattmoor (approved):
/lgtm

Sorry!  Do these not run in tests?

abayer:
@mattmoor They run, but they used the latest release image of `git-init`, so won't surface issues there until post release.

abayer:
/hold cancel

I was just trying to see if there was a non-root workaround, but there isn't one that's simple enough to merit going with it over `runAsUser: 0` in the examples to unblock everything else. I've reopened #4711, because I think we need to somehow get rid of the run-as-root requirement, or at least the need to explicitly specify the user when running `git-init` in a step.

dibyom (approved):

seprokof:
It will be awesome to fix task from Tekton Hub (https://hub.tekton.dev/tekton/task/git-clone) as well, as it based on pretty old image and if users just replace image to newer execution will fail.

Issue #4787, opened 2022-04-22
Missing validation when referencing unknown params with bracket notation


chuangw6:
# Expected Behavior
There are [three ways](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#substituting-parameters-and-resources) to reference a parameter in task steps. 
```yaml
# dot notation
$(params.<name>)
# or bracket notation (wrapping <name> with either single or double quotes):
$(params['<name>'])
$(params["<name>"])
```
If task steps use an unknown `<name>` in any 3 reference forms (means the parameter name is not declared in `params`), the webhook **should validate against it and report an error** like the following when applying a taskrun.

```
Error from server (BadRequest): error when creating "validation-test.yaml": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: non-existent variable in "set -e\necho $(params.fooo) | tee $(results.echo-output.path)\n": spec.taskSpec.steps[0].script
```

# Actual Behavior
- Only validate against `$(params.<name>)` dot notation reference
- Not validate against `$(params['<name>'])` `$(params["<name>"])` bracket notation reference. If I use wrong parameter name, the taskrun still succeeds, but the unknown param reference has some bad behaviour i.e. when echo to result. 

# Steps to Reproduce the Problem
1. Apply the following taskrun in which the step `echo-params` references an unknown parameter name `fooo` using the bracket notation.
```yaml
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  generateName: validation-
spec:
  taskSpec:
    params:
      - name: foo
        type: string
        default: "bar"
    results:
      - name: echo-output
        description: "see echo output"
    steps:
      - name: echo-params
        image: bash
        script: |
          set -e
          echo $(params["fooo"]) | tee $(results.echo-output.path)
```

2. Then, you can see nothing is complaint and taskrun is applied successfully. But the expectation is that the taskrun should not be run and an error should be reported.
3. Get the taskrun yaml
`kubectl get tr <TASKRUN_NAME> -o yaml`
4. You can see the `taskResults` `echo-output` has unexpected content because the parameter `fooo` doesn't exits. And this should not happen.
```
...
  taskResults:
  - name: echo-output
    value: |2+

...
```

# Additional Info
## Normal behaviour when using unknown name in dot notation `$(params.<name>)`.
1. Change the `echo` command in step1's yaml to `echo $(params.fooo) | tee $(results.echo-output.path)`
[...]

PR #4811, opened 2022-04-29
Updated regex pattern to allow all three valid bracket notations.

chitrangpatel:
# Changes
Updated regex pattern to allow all three valid bracket notations.

Prior to this, when supplied parameter name was not the same as that
called in the step, webhook raised an error when using the dot notation
as expected. However, it did not when using the bracket notation.

Updated the unit tests to also check for valid bracket notations.
This PR addresses issue https://github.com/tektoncd/pipeline/issues/4787 and https://github.com/tektoncd/pipeline/issues/4792

# Submitter Checklist

  functionality, content, code)

# Release Notes
NONE


pkg/substitution/substitution.go
```diff
@@ -58,6 +67,15 @@ func ValidateVariableP(value, prefix string, vars sets.String) *apis.FieldError
...
 		}
+	} else {
+		// if expected variables but cannot be extracted from the string.
+		if len(vars) > 0 {
+			return &apis.FieldError{
+				Message: fmt.Sprintf("Expected to find variables but none found in %q", value),
+				// Empty path is required to make the `ViaField`, … work
+				Paths: []string{""},
+			}
+		}
```
chuangw6, line 78: Adding this `else` will not work because different values can be passed to `ValidateVariableP` function other than just param names i.e. image name, step name. 

```yaml
kind: TaskRun
metadata:
  generateName: validation-
spec:
  taskSpec:
    params:
      - name: foo
        default: "bar1"
    results:
      - name: echo-output
    steps:
      - name: echo-params
        image: bash
        script: |
          set -e
          echo $(params.foo) | tee $(results.bad-echo.path)
```

Try this example, when you apply, you'll see 
```
Error from server (BadRequest): error when creating "validation.yaml": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: Expected to find variables but none found in "": spec.taskSpec.steps[0].workingDir
Expected to find variables but none found in "bash": spec.taskSpec.steps[0].image
Expected to find variables but none found in "echo-params": spec.taskSpec.steps[0].name
Expected to find variables but none found in "set -e\necho $(params.foo) | tee $(results.bad-echo.path)\n": spec.taskSpec.steps[0].script
```

PR #4815, opened 2022-05-01
Fix validateNoDuplicateNames

### LINKED ISSUE #4814: Case-insensitive duplication check in TaskRunSpec's params
# Background
There is a function named [validateNoDuplicateNames](https://github.com/tektoncd/pipeline/blob/6f39e948ccd2b2b8091fb3b26d84d83412352087/pkg/apis/pipeline/v1beta1/taskrun_validation.go#L149-L165) in place to check **case-insensitive** duplication against ***taskRun spec's params***, WorkspaceBindings, StepOverrides and SidecarOverrides.

For params duplication check, it's only used in ***taskRun spec's params***, but not ***task spec params***. To make sure param names are consistent, #4806 will also call this [validateNoDuplicateNames](https://github.com/tektoncd/pipeline/blob/6f39e948ccd2b2b8091fb3b26d84d83412352087/pkg/apis/pipeline/v1beta1/taskrun_validation.go#L149-L165) to validate names in ***task spec params***. 

# Expected Behavior
We should expect this function to work correctly regardless of the order of case-insensitive duplications.

# Actual Behavior
- If name `foo` is provided in ***taskRun spec's*** params and then `Foo`, it can check duplication.
- If a name `Foo` is provided in ***taskRun spec's*** params and then `foo`, it fails to check duplication.

# Proposed solution
- caused by the implementation of [validateNoDuplicateNames](https://github.com/tektoncd/pipeline/blob/6f39e948ccd2b2b8091fb3b26d84d83412352087/pkg/apis/pipeline/v1beta1/taskrun_validation.go#L149-L165).

# Question
Why do we check ***case-insensitive*** duplication not ***case-sensitive***? In other words, is there any problem with treating name `foo` and `Foo` as two different parameters?

```yaml
kind: TaskRun
metadata:
  generateName: case-
spec:
  params:
    - name: Foo
      value: "x"
    - name: foo
      value: "y"
  taskSpec:
    params:
      - name: foo
      - name: Foo
    results:
      - name: echo-output
    steps:
      - name: echo-params
        image: bash
        script: |
          set -e
          echo $(params.foo) | tee $(results.echo-output.path)
```
1. run the yaml above you can see, it cannot check duplicated `Foo` and `foo` in taskrun spec.
2. change the order of `Foo` and `foo` in taskrun spec (provide value for `foo` first and then `Foo`) and apply again, you can see it validates the duplication
```
Error from server (BadRequest): error when creating "case-sensitivity.yaml": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: expected exactly one, got both: spec.params[Foo].name
```

chuangw6:
# Changes

Fixes #4814

Prior to this fix, `validateNoDuplicateNames` function was intended to check case-insensitive duplications against names, but it saved original names into the set to compare, which can cause that duplicate names cannot be checked correctly - mentioned in #4814.

In this fix:
- `validateNoDuplicateNames` will save names in lower case into the check set and then compare.
- Unit tests are added for this fix.
- Document that parameter names are case insensitive in the `docs/tasks.md` file and add links to this file's `Specifying Parameters` section in other relevant files.

# Submitter Checklist

  functionality, content, code)

# Release Notes

### DIFF
--- a/docs/pipelineruns.md
+++ b/docs/pipelineruns.md
@@ -260,6 +260,8 @@ to all `persistentVolumeClaims` generated internally.
 
 ### Specifying `Parameters`
 
+(See also [Specifying Parameters in Tasks](tasks.md#specifying-parameters))
+
 You can specify `Parameters` that you want to pass to the `Pipeline` during execution,
 including different values of the same parameter for different `Tasks` in the `Pipeline`.
 
--- a/docs/taskruns.md
+++ b/docs/taskruns.md
@@ -189,7 +189,7 @@ spec:
 
 ### Specifying `Parameters`
 
-If a `Task` has [`parameters`](tasks.md#parameters), you can use the `params` field to specify their values:
+If a `Task` has [`parameters`](tasks.md#specifying-parameters), you can use the `params` field to specify their values:
 
 ```yaml
 spec:
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -447,6 +447,8 @@ Parameter names:
 
 For example, `foo.Is-Bar_` is a valid parameter name, but `barIsBa$` or `0banana` are not.
 
+> NOTE: Parameter names are **case insensitive**. For example, `APPLE` and `apple` will be treated as duplicate parameters. If they appear in the same taskspec's params, validation webhook will complain about that.
+
 Each declared parameter has a `type` field, which can be set to either `array` or `string`. `array` is useful in cases where the number
 of compilation flags being supplied to a task varies throughout the `Task's` execution. If not specified, the `type` field defaults to
 `string`. When the actual parameter value is supplied, its parsed type is validated against the `type` field.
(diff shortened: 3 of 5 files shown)


jerop (commented):
thanks for this fix @chuangw6!

please add more context to the commit message and PR description per the [commit message standards](https://github.com/tektoncd/community/blob/main/standards.md#commits), which can include the detail that the function was intended to be case insensitive as demonstrated by lines 150 and 155 below: 

https://github.com/tektoncd/pipeline/blob/6f39e948ccd2b2b8091fb3b26d84d83412352087/pkg/apis/pipeline/v1beta1/taskrun_validation.go#L150-L155

also, please add documentation that the params are case-insensitive (skimmed through the docs and they don't seem to mention this)

dibyom:
LGTM! 
We should definitely update the docs to mention params are case insensitive

chuangw6:
> thanks for this fix @chuangw6!
> 
> please add more context to the commit message and PR description per the [commit message standards](https://github.com/tektoncd/community/blob/main/standards.md#commits), which can include the detail that the function was intended to be case insensitive as demonstrated by lines 150 and 155 below:
> 
> https://github.com/tektoncd/pipeline/blob/6f39e948ccd2b2b8091fb3b26d84d83412352087/pkg/apis/pipeline/v1beta1/taskrun_validation.go#L150-L155
> 
> also, please add documentation that the params are case-insensitive (skimmed through the docs and they don't seem to mention this)

Thanks! Done!

dibyom (approved):

vdemeester (approved):

PR #4818, opened 2022-05-02
[TEP-0076] Add array support for emitting results

Yongxuanzhang:
# Changes
This is part of work in [TEP-0076](https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md)
This commit provides support for emitting array results. Previous to
this commit we have changed the TaskResults Type to support array. But we can only emit string type result.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Support array type for emitting results from a task as an alpha feature.

The type of the result is changed from string to ArrayOrString.

  results:
    - name: array-results
      type: array
      description: The array results

echo -n "[\"hello\",\"world\"]" | tee $(results.array-results.path)

 This feature is part of the TEP-0076 and its in progress to index into the array result while consuming that result.
```


pkg/apis/pipeline/v1beta1/resource_types.go
```diff
@@ -129,7 +129,7 @@ type PipelineResourceBinding struct {
 // PipelineResourceResult used to export the image name and digest as json
 type PipelineResourceResult struct {
 	Key          string `json:"key"`
-	Value        string `json:"value"`
+	Value        ArrayOrString `json:"value"`
 	ResourceName string `json:"resourceName,omitempty"`
```
Yongxuanzhang, line 133: As suggested by @ywluogg ,  we can also to this to separate the usage of result value and other value, since in other cases there is no need to support array. 
```suggestion
	Value        String `json:"value"`
	ResultValue *ArrayOrString `json:"resultValue,omitempty"`
	ResourceName string `json:"resourceName,omitempty"`
```

I'd like to provide this solution first and see how other ppl think of this

Yongxuanzhang: Option 1:  Change Value type from String to ArrayOrString
Pro: backwards compatible without reducing termination msg limit
Con: In some other places we only need String and doesn’t necessarily need to use ArrayOrString

Option 2: Add a new field `ResultValue *ArrayOrString`
Pro: Separate the usage of emitting results (where we need array) and other data(where we only need string) to termination path
Con: reduce the available termination msg size

pkg/entrypoint/entrypointer.go
```diff
@@ -184,30 +184,35 @@ func (e Entrypointer) Go() error {
...
+		if err := e.readResultsFromDisk(pipeline.DefaultResultPath); err != nil {
 			logger.Fatalf("Error while handling results: %s", err)
 		}
 	}
 
 	return err
 }
 
-func (e Entrypointer) readResultsFromDisk() error {
+func (e Entrypointer) readResultsFromDisk(resultDir string) error {
```
Yongxuanzhang, line 195: This is for better unit testing. Without this I think we cannot create "tekton/results" dir to emit results

pkg/apis/pipeline/v1beta1/result_types.go
```diff
@@ -39,7 +39,7 @@ type TaskRunResult struct {
 	Type ResultsType `json:"type,omitempty"`
 
 	// Value the given value of the result
-	Value string `json:"value"`
+	Value ArrayOrString `json:"value"`
```
ywluogg, line 42: This looks like a API change to me. It seems like we need to follow https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#alpha-beta-and-ga for adding new feature and deprecation rules. Pinging @[dibyom](https://github.com/dibyom) for confirmation. This is a field that users can specify. I think we should add a separate field for the new ArrayOrString and implement on the new field

Yongxuanzhang: It is a change, but it won't have effect on users and introduce breaking changes. And our TEP[0076](https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md) requires us to change this `Value` I believe?

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -112,12 +112,22 @@ type ArrayOrString struct {
...
+		if err := json.Unmarshal(value, &arrayOrString.ArrayVal); err == nil {
+			return nil
+		}
+	case '"':
 		arrayOrString.Type = ParamTypeString
-		return json.Unmarshal(value, &arrayOrString.StringVal)
+		if err := json.Unmarshal(value, &arrayOrString.StringVal); err == nil {
+			return nil
+		}
 	}
```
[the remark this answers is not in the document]
Yongxuanzhang: For error and other cases, I use this code as default handling, use a default case will be duplicated 

```
	arrayOrString.Type = ParamTypeString
	arrayOrString.StringVal = string(value)
	arrayOrString.ArrayVal = nil
```

pkg/apis/pipeline/v1beta1/result_types.go
ywluogg: We came to a conclusion that this won't be breaking the API change policy. However, we definitely are blocked by the design about whether we want to have a brand new struct on this or just rename StringOrArray.

Yongxuanzhang: Ok but it doesn't block this PR right, I'm still using StringOrArray. We need to figure out that in Chuang's

pkg/apis/pipeline/v1beta1/param_types_test.go
```diff
@@ -171,6 +171,9 @@ func TestArrayOrString_UnmarshalJSON(t *testing.T) {
 		input  string
 		result v1beta1.ArrayOrString
 	}{
+		{"{\"val\": 123}", *v1beta1.NewArrayOrString("123")},
+		{"{\"val\": [1,2,3]}", *v1beta1.NewArrayOrString("[1,2,3]")},
+		{"{\"val\": [1,\"2\",3]}", *v1beta1.NewArrayOrString("[1,\"2\",3]")},
 		{"{\"val\": \"123\"}", *v1beta1.NewArrayOrString("123")},
 		{"{\"val\": \"\"}", *v1beta1.NewArrayOrString("")},
```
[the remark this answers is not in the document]
ywluogg: > "{\"val\": }",

Yeah. Is it testable?

> 123 is a case starting with '1' not " or [

Ah I missed this test case. Nice!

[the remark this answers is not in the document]
Yongxuanzhang: It will through error and it is expected. Since json won't unmarshal this right? I can add a test to assert error if that is necessary

pkg/entrypoint/entrypointer.go
```diff
@@ -184,31 +184,36 @@ func (e Entrypointer) Go() error {
...
 		if os.IsNotExist(err) {
 			continue
 		} else if err != nil {
 			return err
 		}
+		// unmarshal the filesContents to string or array
+		aos := v1beta1.ArrayOrString{}
+		if err := aos.UnmarshalJSON(fileContents); err != nil {
+			return err
+		}
```
ywluogg, line 211: You can make the new code into a module like `ReadArrayFromFile()`, and just write unit tests on that function

pkg/apis/pipeline/v1beta1/result_types.go
ywluogg: If we end up introducing a new struct, the golang library will be broken twice. Consider not making this particular line of change and be able to test everything else.

pkg/entrypoint/entrypointer.go
Yongxuanzhang: Is it used to avoid changing the signature that I added for `readResultsFromDisk`? I think one advantage of that is we can improve the test coverage and I have already added unit tests for that func and cover this code

pkg/apis/pipeline/v1beta1/result_types.go
Yongxuanzhang: Ok

pkg/apis/pipeline/v1beta1/param_types_test.go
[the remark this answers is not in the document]
ywluogg: > It will through error and it is expected. Since json won't unmarshal this right? I can add a test to assert error if that is necessary

Yeah a test for an assert error would be good.

pkg/entrypoint/entrypointer.go
ywluogg: I think here the module is input: file path; output: ArrayOrString, error. This is different from UnmarshalJSON(). I'm also fine with not having making this module and leave these changes to a later PR. This can help you unblock from being influenced from other PRs

pkg/apis/pipeline/v1beta1/result_types.go
ywluogg: This thread can be closed, as we confirmed that this won't be blocked by introducing a new struct for ArrayOrString, and this is not a API change.

Yongxuanzhang: sure

pkg/apis/pipeline/v1beta1/param_types_test.go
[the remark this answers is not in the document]
Yongxuanzhang: I'm adding a test to assert error

pkg/entrypoint/entrypointer.go
```diff
@@ -184,31 +184,36 @@ func (e Entrypointer) Go() error {
...
 		// if the file doesn't exist, ignore it
 		output = append(output, v1beta1.PipelineResourceResult{
-			Key:        resultFile,
-			Value:      string(fileContents),
-			ResultType: v1beta1.TaskRunResultType,
+			Key:         resultFile,
+			ResultValue: &aos,
+			ResultType:  v1beta1.TaskRunResultType,
```
ywluogg, line 216: Seems like to me you can do:
```
tmp := &v1beta1.PipelineResourceResult{
			Key:         resultFile,
			ResultValue: v1beta1.ArrayOrString{},
			ResultType:  v1beta1.TaskRunResultType,
}
...
if err := *tmp.ResultValue.UnmarshalJSON(fileContents); err != nil {
			return err
		}
...
output = append(output, *tmp)
```
This makes it possible to set ResultValue to be ArrayOrString instead of *ArrayOrString

pkg/apis/pipeline/v1beta1/param_types.go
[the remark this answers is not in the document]
Yongxuanzhang: I think so, for the previous code we will do `string(fileContents)` for result data, assume user write an integer into the file, it will be read as string in the end.

If I want to use ArrayorString, since we don't support int, it will through error. So I suggest for any other type we don't support we just convert them into string (it will be the same as before).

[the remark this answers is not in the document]
ywluogg: I see you explained in #4820. It makes sense to me. You can mark this as resolved.

pkg/apis/pipeline/v1beta1/resource_types_test.go
```diff
@@ -153,17 +153,21 @@ func TestPipelineResourceResult_UnmarshalJSON(t *testing.T) {
...
+		}, {
+			name: "type defined as array - InternalTektonResult",
+			data: "{\"key\":\"resultName\",\"resultValue\":[\"cat\",\"dog\",\"squirrel\"], \"type\": \"InternalTektonResult\"}",
+			pr:   v1beta1.PipelineResourceResult{Key: "resultName", ResultValue: v1beta1.NewArrayOrString("cat", "dog", "squirrel"), ResultType: v1beta1.InternalTektonResultType},
 		}, {
 			name: "type defined as int",
-			data: "{\"key\":\"resultName\",\"value\":\"\", \"type\": 1}",
-			pr:   v1beta1.PipelineResourceResult{Key: "resultName", Value: "", ResultType: v1beta1.TaskRunResultType},
+			data: "{\"key\":\"resultName\",\"resultValue\":\"\", \"type\": 1}",
+			pr:   v1beta1.PipelineResourceResult{Key: "resultName", ResultValue: v1beta1.NewArrayOrString(""), ResultType: v1beta1.TaskRunResultType},
```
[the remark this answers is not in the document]
Yongxuanzhang: oh yes you're right! My bad

pkg/apis/pipeline/v1beta1/swagger.json
```diff
@@ -870,6 +870,9 @@
           "description": "The field ResourceRef should be deprecated and removed in the next API version. See https://github.com/tektoncd/pipeline/issues/2694 for more information.",
           "$ref": "#/definitions/v1beta1.PipelineResourceRef"
         },
+        "resultValue": {
+          "$ref": "#/definitions/v1beta1.ArrayOrString"
+        },
```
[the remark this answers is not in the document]
Yongxuanzhang: Thanks for reminding!

pkg/entrypoint/entrypointer_test.go
```diff
@@ -249,6 +252,85 @@ func TestEntrypointer(t *testing.T) {
...
+				ResultType: 1}},
+	}, {
+		desc:          "read array reault file",
+		results:       []string{"results"},
+		resultContent: []v1beta1.ArrayOrString{*v1beta1.NewArrayOrString("hello", "world")},
+		want: []v1beta1.PipelineResourceResult{
+			{ResultValue: v1beta1.NewArrayOrString("hello", "world"),
+				ResultType: 1}},
+	}, {
+		desc:          "read string and array reault files",
```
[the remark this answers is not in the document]
Yongxuanzhang: oh thx! 😄

pkg/pod/status_test.go
```diff
@@ -44,23 +44,23 @@ func TestSetTaskRunStatusBasedOnStepStatus(t *testing.T) {
 			Name: "step-bar-0",
 			State: corev1.ContainerState{
 				Terminated: &corev1.ContainerStateTerminated{
-					Message: `[{"key":"resultName","value":"resultValue", "type":1}, {"key":"digest","value":"sha256:1234","resourceRef":{"name":"source-image"}}]`,
+					Message: `[{"key":"resultName","resultValue":"resultValue", "type":1}, {"key":"digest","value":"sha256:1234","resourceRef":{"name":"source-image"}}]`,
```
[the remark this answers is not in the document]
Yongxuanzhang: This is a bit different, for these test cases they are mixed of results message and other message. But you're right I should add separate test cases

pkg/entrypoint/entrypointer.go
Yongxuanzhang: if that we, we need to have. There's no way to avoid  *ArrayOrString. Another reason we need to use pointer for ResultValue is that we need to use omitempty for json marshalling, and only pointer can work
```
tmp := &v1beta1.PipelineResourceResult{
			Key:         resultFile,
			ResultValue: &v1beta1.ArrayOrString{},
			ResultType:  v1beta1.TaskRunResultType,
}
```

pkg/apis/pipeline/v1beta1/resource_types_test.go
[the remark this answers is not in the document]
Yongxuanzhang: So I take a look at these test cases again. It's testing InternalTektonResult and TaskRunResult, so for InternalTektonResult we will still use the original Value (type is string), for TaskRunResult we will use ArrayorString. 

I reformat the test cases and also add more cases as well.

Yongxuanzhang:
I just realized that changes of this pr should be gated by alpha flag right? @ywluogg

ywluogg:
> I just realized that changes of this pr should be gated by alpha flag right? @ywluogg

Yes thats right!

pkg/pod/status_test.go
```diff
@@ -1014,9 +1016,110 @@ func TestMakeTaskRunStatus(t *testing.T) {
...
+				if x == nil {
+					return y == nil
+				}
+				return y != nil
+			})
+			if d := cmp.Diff(c.want, got, ignoreVolatileTime, ensureTimeNotNil); d != "" {
+				t.Errorf("Diff %s", diff.PrintWantGot(d))
+			}
+			if tr.Status.StartTime.Time != c.want.StartTime.Time {
+				t.Errorf("Expected TaskRun startTime to be unchanged but was %s", tr.Status.StartTime)
```
Yongxuanzhang, line 1040: This is a mismatching bug from github. I only added one line above

pkg/pod/status_test.go
```diff
@@ -629,7 +631,7 @@ func TestMakeTaskRunStatus(t *testing.T) {
 				}},
 				TaskRunResults: []v1beta1.TaskRunResult{{
 					Name:  "resultName",
-					Value: "resultValue",
+					Value: *v1beta1.NewArrayOrString("resultValue"),
```
ywluogg, line 634: This should be a value not a pointer. Same beflow

ywluogg (commented):
A lot of your unit tests are failing because you added * in front of a ArrayOrString that should be a value

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -112,12 +112,28 @@ type ArrayOrString struct {
 
 // UnmarshalJSON implements the json.Unmarshaller interface.
 func (arrayOrString *ArrayOrString) UnmarshalJSON(value []byte) error {
-	if value[0] == '"' {
+	if len(value) == 0{
```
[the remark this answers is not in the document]
Yongxuanzhang: Thanks! I forgot to format the file

pkg/pod/status_test.go
Yongxuanzhang: unit tests are passing, check this for last commit or all recent runs https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4818/pull-tekton-pipeline-unit-tests/1522392776887504899/

I'm failing build tests because of linting. 😄 

This one is tricky because I'm using a func to create arrayorstring, it returns a pointer, so actually here is a value not pointer

pkg/pod/status.go
```diff
@@ -217,16 +219,28 @@ func createMessageFromResults(results []v1beta1.PipelineResourceResult) (string,
...
 	for _, r := range results {
 		switch r.ResultType {
 		case v1beta1.TaskRunResultType:
+			cfg := config.FromContextOrDefaults(ctx)
+			aos := v1beta1.ArrayOrString{}
+			if cfg.FeatureFlags.EnableAPIFields == config.AlphaAPIFields {
+				err := aos.UnmarshalJSON([]byte(r.Value))
+				if err != nil {
+					continue
+				}
```
ywluogg, line 235: this means there is an error, we continue? Just to confirm here: you want to do if you can't unmarshall as array, you want to keep it as string? If so, should it be another else for setting it to be string?

Yongxuanzhang: Thanks for bring this up!
My original idea is that if we cannot unmarshal the r.Value into ArrayOrString (it can be string, array or object if Chuang's pr got merged), we just continue the for loop and don't throw error. We will add an empty ArrayOrString to the taskRunResult. 

Is it ok if I write it this way? 
```
aos.UnmarshalJSON([]byte(r.Value))
```

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -112,12 +112,27 @@ type ArrayOrString struct {
...
+		arrayOrString.StringVal = string(value)
+		return nil
 	}
-	arrayOrString.Type = ParamTypeArray
-	return json.Unmarshal(value, &arrayOrString.ArrayVal)
+	switch value[0] {
+	case '[':
+		arrayOrString.Type = ParamTypeArray
+		if err := json.Unmarshal(value, &arrayOrString.ArrayVal); err == nil {
+			return nil
```
ywluogg, line 124: I realized the errors didn't get thrown here. This should be `err != nil` then `return nil`

Yongxuanzhang: The reason I did this is because for case like nested lists or lists using not string element, the UnmarshalJson won't work.
Like `[1,2,3]` with int elements, but I think we should Unmarshal it into a whole string. So if we encountered err when trying to unmarshal to array, we will continue and unmarshal it to a string. 

I agree this part of code is not writing very well.

ywluogg: Ah I see. Then this should be fine. We will capture the type mismatch when you validate the results against PropertiesSpec. So it definitely worths adding more comments here and in `filterResultsAndResources()` to explain the details

pkg/pod/status.go
ywluogg: Please see the other comment. You should definitely explain in details about handling cases about unmarshalling here

pkg/apis/pipeline/v1beta1/param_types.go
Yongxuanzhang: Sure! Thanks for the suggestions!

pkg/pod/status_test.go
ywluogg: Ah I see. This thread can be closed

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -120,6 +120,9 @@ func (arrayOrString *ArrayOrString) UnmarshalJSON(value []byte) error {
 	switch value[0] {
 	case '[':
 		arrayOrString.Type = ParamTypeArray
+		// We're tring to Unmarshal to []string, but for cases like []int or other types
+		// of nested array which we don't support yet, we should continue and Unmarshal
+		// it to String.
```
ywluogg, line 125: Can you also mention something like this "if the Type being set doesn't match what it actually should be, it will be captured somewhere else"

pkg/pod/status.go
```diff
@@ -229,6 +229,7 @@ func filterResultsAndResources(ctx context.Context, results []v1beta1.PipelineRe
 			cfg := config.FromContextOrDefaults(ctx)
 			aos := v1beta1.ArrayOrString{}
 			if cfg.FeatureFlags.EnableAPIFields == config.AlphaAPIFields {
+				// Note that for unsupported types such as []int will be Unmarshalled to String
```
ywluogg, line 232: Can you also mention something like this "if the Type being set doesn't match what it actually should be, it will be captured somewhere else"

ywluogg (approved):

pkg/pod/status.go
Yongxuanzhang: Sure! You mean the validation of type (configured vs unmarshalled)? I guess that part also includes schema validation. Correct me if I'm wrong

ywluogg: > Sure! You mean the validation of type (configured vs unmarshalled)? I guess that part also includes schema validation. Correct me if I'm wrong

Yes that's correct, or like schema vs object. This way we would know if the unmarshalled type matches what we expect

Yongxuanzhang: Got it!

Yongxuanzhang:
> hey @Yongxuanzhang this looks good to me. Can users use array results now? In that case, we should also update the docs/ as part of this PR

Oh yes, thanks for reminding me of that! I will add doc and also an alpha example as well!

dibyom (approved):

examples/v1beta1/taskruns/alpha/emit-array-results.yaml
```diff
@@ -0,0 +1,30 @@
+kind: Task
```
pritidesai, line 1: In addition to the `Task`, please add an example `Pipeline`. The results are more of a communication means between two tasks (or pipelineTasks). The example are great way to share the usage of the new features.

Yongxuanzhang: Hi @pritidesai, thanks for your comment! 
In this PR array results passed between tasks are not implemented. This commit focuses on enabling emitting array results from entrypoint->controller. 
I'm planning to add it (also array indexing) as well as examples in the followup PR.  
Or should I add it in this PR?  What do you think?

Yongxuanzhang: Or maybe I should remove this example and add both examples for next pr?

pritidesai: Thank you @Yongxuanzhang for the explanation. 

> In this PR array results passed between tasks are not implemented. 

You mean just the indexing is not supported in this PR or in general, the consumer will get an array result as a string value? I am trying to understand what happens when a consumer refers to such array result in a traditional way i.e. `$tasks.taskA.results.arrayResult`.

Yongxuanzhang: So with this PR, if referred `$tasks.taskA.results.arrayResult` in params and it is an array, the users will get a string of `$tasks.taskA.results.arrayResult` not the array or the string format of the array. 

It is also something related to this [issue](https://github.com/tektoncd/pipeline/issues/4879). 
This is a bit tricky part and it would be better to be addressed in a separate PR. Since `$tasks.taskA.results.arrayResult` or `$tasks.taskA.results.arrayResult[*]` will be treated as `string` type in `ArrayOrString`.

Yongxuanzhang: Oh I realized that the doc I added may be confusing as well, support emitting array results may let users think that they can pass array results between tasks within a pipeline, but it is not implemented in this PR.

pkg/pod/status.go
pritidesai: I am sorry about this I am still catching up ...

Trying to understand why is the error not reported here? 

How are we preventing the users to specify array task results without `alpha` feature flag? Or is this something being addressed somewhere else?

Yongxuanzhang: Sure np! 
This is an add-on feature of this function. It will include all the cases we support before and also support more cases like unmarshal integer/array of integer into string, and enable us to use `ArrayOrString` for results value

So array results is a new case, it is something we previously don't support.
If user specify array task results without alpha feature flag, this function will be unmarshalled to `array` but it will be converted to string here. So the behaviour won't change.
https://github.com/tektoncd/pipeline/blob/0c008a0c0a6d979f65d281f69ea0c60df3d80169/pkg/entrypoint/entrypointer.go#L208-L211

examples/v1beta1/taskruns/alpha/emit-array-results.yaml
pritidesai: > So this PR only enables the Task level results array, not sure if this is a real case.

It might be or might not be, we do not advice such usage but its not restricted either.

Yongxuanzhang: Should I fix the readme to clarify this? @pritidesai

examples/v1beta1/taskruns/alpha/emit-array-results.yaml
```diff
@@ -0,0 +1,30 @@
...
+      description: The array results
+  steps:
+    - name: write-array
+      image: bash:latest
+      script: |
+        #!/usr/bin/env bash
+        echo -n "[\"hello\",\"world\"]" | tee $(results.array-results.path)
+    - name: print-array
+      image: bash:latest
+      script: cat $(results.array-results.path)
```
pritidesai, line 21: this could be a little more robust example by introducing validation that `$(results.array-results.path)` matches the expected value ...

Yongxuanzhang: Sure!

docs/tasks.md
```diff
@@ -684,6 +684,33 @@ or [at the `Pipeline` level](./pipelines.md#configuring-execution-results-at-the
...
+      description: The array results
+  steps:
+    - name: write-array
+      image: bash:latest
+      script: |
+        #!/usr/bin/env bash
+        echo -n "[\"hello\",\"world\"]" | tee $(results.array-results.path)
+    - name: print-array
+      image: bash:latest
+      script: cat $(results.array-results.path)
```
pritidesai, line 711: thank you @Yongxuanzhang for following up, we can drop this step from the doc. The purpose of this section is to share usage on how to emit/write a result as a list.

examples/v1beta1/taskruns/alpha/emit-array-results.yaml
pritidesai: thank you @Yongxuanzhang I added comment ⬆️

docs/tasks.md
Yongxuanzhang: Sure!

examples/v1beta1/taskruns/alpha/emit-array-results.yaml
Yongxuanzhang: I added some check, is it ok?

examples/v1beta1/taskruns/alpha/emit-array-results.yaml
Yongxuanzhang: I have added more clarification in the doc

Yongxuanzhang:
Hi @pritidesai, may you take a look again? Thanks! ❤️

pkg/apis/pipeline/v1beta1/result_validation.go
```diff
@@ -17,22 +17,21 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"knative.dev/pkg/apis"
 )
 
 // Validate implements apis.Validatable
-func (tr TaskResult) Validate(_ context.Context) *apis.FieldError {
+func (tr TaskResult) Validate(ctx context.Context) (errs *apis.FieldError) {
```
Yongxuanzhang, line 25: @ywluogg I made some new changes to this func, so we can fail fast if the alpha gate is not enabled

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -141,17 +141,40 @@ type ArrayOrString struct {
 
 // UnmarshalJSON implements the json.Unmarshaller interface.
 func (arrayOrString *ArrayOrString) UnmarshalJSON(value []byte) error {
+	// ArrayOrString is used for Results Value as well, the results can be any kind of
+	// data so we need to check if it is empty.
+	if len(value) == 0 {
+		arrayOrString.Type = ParamTypeString
+		return nil
+	}
```
pritidesai, line 149: Would this impact parameters with no values? is it covered by alternate validation to make sure a required parameter is specified - please confirm that the parameter validation has a test to check for no empty value for required params.

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -141,17 +141,40 @@ type ArrayOrString struct {
...
 	switch value[0] {
 	case '[':
+		// We're tring to Unmarshal to []string, but for cases like []int or other types
+		// of nested array which we don't support yet, we should continue and Unmarshal
+		// it to String. If the Type being set doesn't match what it actually should be,
+		// it will be captured somewhere else.
```
pritidesai, line 155: thank you for adding comments 🙏 

typo - s/tring/trying

Yongxuanzhang: oh thank you!

pritidesai: NIT: instead of saying it is captured somewhere else, please be explicit about the validation routine which will capture this

pkg/apis/pipeline/v1beta1/param_types.go
Yongxuanzhang: Thanks! Could you point me to a case where parameters has no values?
Correct if I'm wrong, if the parameters has no values then it will be captured by validation check 
For example, this won't work.  And won't reach this function. This case is special for results since user can write custom data to result path and it could be empty

```
kind: TaskRun
metadata:
  name: validation-1
spec:
  taskSpec:
    params:
      - name: foo
        type: string
        default:
    steps:
      - name: echo-params
        image: bash
        args:
          [
            "echo",
           "$(params.foo)"]
```

pkg/apis/pipeline/v1beta1/param_types.go
Yongxuanzhang: sure!

pkg/apis/pipeline/v1beta1/param_types.go
Yongxuanzhang: So validation will check if the param is empty:
https://github.com/tektoncd/pipeline/blob/7aef957effb55a519bba97c9d2c36a28482b35d4/pkg/reconciler/taskrun/validate_resources.go#L110

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -141,17 +141,40 @@ type ArrayOrString struct {
...
+			return nil
+		}
 	}
+	// if failed to unmarshal to above types we will convert the value to string and
+	// marshal it to string
+	arrayOrString.Type = ParamTypeString
+	arrayOrString.StringVal = string(value)
+	arrayOrString.ArrayVal = nil
+	arrayOrString.ObjectVal = nil
+	return nil
```
pritidesai, line 177: This function does look weird. Can we avoid using `switch`:

```
	if value[0] == '[' {
		// We're trying to Unmarshal to []string, but for cases like []int or other types
		// of nested array which we don't support yet, we should continue and Unmarshal
		// it to String. If the Type being set doesn't match what it actually should be,
		// it will be captured somewhere else.
		// if failed to unmarshal to array, we will convert the value to string and marshal it to string
		if err := json.Unmarshal(value, &arrayOrString.ArrayVal); err == nil {
			arrayOrString.Type = ParamTypeArray
			return nil
		}
	}
	if value[0] == '{' {
		// if failed to unmarshal to map, we will convert the value to string and marshal it to string
		if err := json.Unmarshal(value, &arrayOrString.ObjectVal); err == nil {
			arrayOrString.Type = ParamTypeObject
			return nil
		}
	}
	arrayOrString.ArrayVal = nil
	arrayOrString.ObjectVal = nil
	arrayOrString.Type = ParamTypeString
	arrayOrString.StringVal = string(value)
```

Also, the unmarshalling does set the `arrayOrString.ArrayVal` to an array of equal length of the value in case of failure. And that is the reason we need to set it to nil explicitly. Will be great if we find a way unmarshal does not change the `ArrayVal` but not a blocker for this PR.

pkg/apis/pipeline/v1beta1/param_types_test.go
```diff
@@ -282,6 +286,48 @@ func TestArrayOrString_UnmarshalJSON(t *testing.T) {
...
+		input    string
+		expected v1beta1.ArrayOrString
+	}{
+		{desc: "empty value", input: ``, expected: *v1beta1.NewArrayOrString("")},
+		{desc: "int value", input: `1`, expected: *v1beta1.NewArrayOrString("1")},
+		{desc: "int array", input: `[1,2,3]`, expected: *v1beta1.NewArrayOrString("[1,2,3]")},
+		{desc: "nested array", input: `[1,\"2\",3]`, expected: *v1beta1.NewArrayOrString(`[1,\"2\",3]`)},
+		{desc: "string value", input: `hello`, expected: *v1beta1.NewArrayOrString("hello")},
+		{desc: "array value", input: `["hello","world"]`, expected: *v1beta1.NewArrayOrString("hello", "world")},
+	}
```
pritidesai, line 301: thank you @Yongxuanzhang for adding a list of unit tests 🙏 

NIT: please add tests for unmarshalling map as well, not a blocker though.

Yongxuanzhang: Sure! Np!

pkg/apis/pipeline/v1beta1/param_types.go
Yongxuanzhang: I changed it to if, and the json.Unmarshal(value, &arrayOrString.StringVal) is needed, we cannot directly call arrayOrString.StringVal = string(value).

Yes, it is a bit weird that we need to set it to nil since we tried to unmarshal to array and it will create an empty array. It will be worth figuring out a better idea of this as you suggested

pkg/apis/pipeline/v1beta1/result_validation.go
```diff
@@ -17,22 +17,21 @@ import (
...
-		if tr.Type == allowedType {
-			validType = true
-		}
+	// Array and Object is alpha feature
+	if tr.Type == ResultsTypeArray || tr.Type == ResultsTypeObject {
+		return errs.Also(ValidateEnabledAPIFields(ctx, "results type", config.AlphaAPIFields))
 	}
-	if !validType {
+
+	if tr.Type != ResultsTypeString {
```
pritidesai, line 34: wouldn't this condition cause failure for an array and a map results? 

The previous if condition is validating that the type cannot be array/map without `alpha` feature flag. 

So, lets say, the alpha feature flag is set and result is array, this if condition will fail and cause validation error, right? 🤔 

Also, not a blocker for this PR but please add a unit test `result_validation_test.go`

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -317,6 +317,32 @@ func TestTaskSpecValidate(t *testing.T) {
...
+	}, {
+		name: "valid result type array",
+		fields: fields{
+			Steps: []v1beta1.Step{{
+				Image: "my-image",
+				Args:  []string{"arg"},
+			}},
+			Results: []v1beta1.TaskResult{{
+				Name:        "MY-RESULT",
+				Type:        "array",
```
pritidesai, line 329: s/array/v1beta1.ResultsTypeArray/

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -317,6 +317,32 @@ func TestTaskSpecValidate(t *testing.T) {
...
+	}, {
+		name: "valid result type object",
+		fields: fields{
+			Steps: []v1beta1.Step{{
+				Image: "my-image",
+				Args:  []string{"arg"},
+			}},
+			Results: []v1beta1.TaskResult{{
+				Name:        "MY-RESULT",
+				Type:        "object",
```
pritidesai, line 342: s/object/v1beta1.ResultsTypeObject/

pkg/apis/pipeline/v1beta1/result_validation.go
Yongxuanzhang: Thanks for the comment! 
If the type is `ResultsTypeArray` or `ResultsTypeObject`, we will directly return this check and won't reach the following code. So they won't fail validation if  alpha feature flag is set. 
```
return errs.Also(ValidateEnabledAPIFields(ctx, "results type", config.AlphaAPIFields))
```

Yongxuanzhang: This function is tested in `task_validation_test.go`,  but I can add separate test for this func for sure! Maybe save it for another PR?

pkg/apis/pipeline/v1beta1/task_validation_test.go
Yongxuanzhang: Sure! Thank you!

pkg/pod/status.go
```diff
@@ -217,16 +219,30 @@ func createMessageFromResults(results []v1beta1.PipelineResourceResult) (string,
...
+func filterResultsAndResources(ctx context.Context, results []v1beta1.PipelineResourceResult) ([]v1beta1.TaskRunResult, []v1beta1.PipelineResourceResult, []v1beta1.PipelineResourceResult) {
 	var taskResults []v1beta1.TaskRunResult
 	var pipelineResourceResults []v1beta1.PipelineResourceResult
 	var filteredResults []v1beta1.PipelineResourceResult
 	for _, r := range results {
 		switch r.ResultType {
 		case v1beta1.TaskRunResultType:
+			cfg := config.FromContextOrDefaults(ctx)
+			aos := v1beta1.ArrayOrString{}
+			if cfg.FeatureFlags.EnableAPIFields == config.AlphaAPIFields {
```
pritidesai, line 231: Is this check necessary here? We are already validating specifications ⬆️  https://github.com/tektoncd/pipeline/pull/4818/files#diff-2a1291b0e1080edbd9fdcd0c6e4c6bd4bfc2967e2c376815219bb571aeeae70fR30

Yongxuanzhang: Yes we have validated in admission webhook, but I feel like it is no wrong to add more checks in reconciler code if possible.

pkg/pod/status.go
```diff
@@ -217,16 +219,30 @@ func createMessageFromResults(results []v1beta1.PipelineResourceResult) (string,
...
+				// Note that for unsupported types such as []int will be Unmarshalled to String
+				err := aos.UnmarshalJSON([]byte(r.Value))
+				if err != nil {
```
pritidesai, line 234: this if statement will not be reached at all, right? since `UnmarshalJSON` does not return any error.

Yongxuanzhang: Thanks for pointing that out! 
Yes right now we won't reach this, but considering that `UnmarshalJSON` may be changed (maybe one day we need to return err for some cases and don't have to convert them to string), maybe it is ok to keep it?

pkg/apis/pipeline/v1beta1/result_validation.go
pritidesai: thank you @Yongxuanzhang I commented a little early 🙃  thought there was more to this but apparently its only checking the types.

pkg/pod/status.go
pritidesai: The side effect is we are bringing `context` down for no reason. This validation does not add any value and in long run, might require more cycles to understand why its implemented here.

Yongxuanzhang: Yes! Another reason we validate here is that it is possible that users don't specify the result type, it is an optional field and will be set to `string` if user not specified. If the type is required mandatory it will introduce breaking changes. 
So the validation in webhook is not 100% reliable in the case that user creates a task without setting the result type.
And we cannot infer the result type in validation webhook since the results are emitted during runtime.
That's why it may make sense to validate when we get the emitted results at reconciler.

pkg/apis/pipeline/v1beta1/result_types.go
pritidesai: I ran `tkn` which failed with:

```
tkn taskrun list
Error: failed to list TaskRuns from namespace default: unrecognized type: string
```

I haven't ran any troubleshooting on the CLI to confirm if its caused by this change 🙃 .

```
tkn version                                                                                       
Client version: 0.23.1
Pipeline version: devel
```

It might be safe to merge this PR but the CLI might will have to updated before adopting the latest pipelines release (with these changes). @vdemeester @vinamra28

Yongxuanzhang: Thanks for checking! And we also need to change that type for Pipelinerun result in the following PR.

Yongxuanzhang: Yeah sure!

Yongxuanzhang: cc:@ywluogg so we should send emails for this?

Yongxuanzhang: It looks like api mismatch between the pipeline and cli, it should be fixed when we release this change and then bump the pipeline version for cli? Am I correct?

pritidesai: /cc @tektoncd/cli-maintainers

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
```diff
@@ -138,11 +138,13 @@ func resolveResultRef(pipelineState PipelineRunState, resultRef *v1beta1.ResultR
...
 
-	var runName, taskRunName, resultValue string
+	var runName, runValue, taskRunName string
+	var resultValue v1beta1.ArrayOrString
 	var err error
 	if referencedPipelineTask.IsCustomTask() {
 		runName = referencedPipelineTask.Run.Name
-		resultValue, err = findRunResultForParam(referencedPipelineTask.Run, resultRef)
+		runValue, err = findRunResultForParam(referencedPipelineTask.Run, resultRef)
+		resultValue = *v1beta1.NewArrayOrString(runValue)
```
pritidesai, line 147: should this be calling unmarshalling here? Calling NewArrayOrString with `runValue` here returns a `string`. 

Also, why limiting this to just the custom task?

pritidesai: > Also, why limiting this to just the custom task?

So custom task results are still `string`:

```
type RunResult struct {
	// Name the given name
	Name string `json:"name"`
	// Value the given value of the result
	Value string `json:"value"`
}
```

Are we planning on changing the custom task result to string/array? Something coming up ...

Yongxuanzhang: Yes, this is a great question! We don't support structured results for custom task in this PR. So this is intentionally to return string here.

Yongxuanzhang: We plan to add support for task and pipeline first, custom task is not discussed and also not explicitly mentioned in the tep (in tep we only mention task and pipeline) but we will discuss later!

https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md#summary

pritidesai: The first use case explicitly talks about custom task - https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md#use-cases Custom task `params` has use case to consume an array which could be a result of other task and produce a result with an array (a list of hashes/a list of built images).

Thank you @Yongxuanzhang for following up on this 🙏

pkg/apis/pipeline/v1beta1/result_types.go
pritidesai: > It looks like api mismatch between the pipeline and cli, it should be fixed when we release this change and then bump the pipeline version for cli? Am I correct? @pritidesai

I think It will be fixed when CLI bumps the pipelines version (which will have these changes) and update their usage/implementation to change the type of the results from `string` to `StringOrArray`. The error reported here is coming from:

```
./pkg/cmd/taskrun/list.go:                              return fmt.Errorf("failed to list TaskRuns from namespace %s: %v", p.Namespace(), err)
```

pritidesai:
@Yongxuanzhang thank you for this 🙏  I have updated the release notes.

Yongxuanzhang:
> @Yongxuanzhang thank you for this 🙏 I have updated the release notes.

Thank you!! ❤️

pkg/reconciler/pipelinerun/resources/resultrefresolution.go
Yongxuanzhang: Yes! I have added in our tracking issue

pritidesai:
We have opened an issue to track a `type` change from `string` to `StringOrArray`. This might impact CLI and anyone consuming the result type directly.

This change will be part of the GitHub release notes after a release is created.

/lgtm

PR #4833, opened 2022-05-04
Added validation for parameters with bracket notation

chitrangpatel:
Added validation for parameters with bracket notation.

Prior to this, when supplied parameter name was not the same as that
called in the step, webhook raised an error when using the dot notation
as expected. However, it did not when using the bracket notation.

# Changes

Updated the unit tests to also check for valid bracket notations.
This PR addresses issue https://github.com/tektoncd/pipeline/issues/4787

# Submitter Checklist

  functionality, content, code)

# Release Notes
```
Added validation for parameters with bracket notation
```

### DIFF
--- a/pkg/substitution/substitution.go
+++ b/pkg/substitution/substitution.go
@@ -26,8 +26,7 @@ import (
 )
 
 const parameterSubstitution = `[_a-zA-Z][_a-zA-Z0-9.-]*(\[\*\])?`
-
-const braceMatchingRegex = "(\\$(\\(%s\\.(?P<var>%s)\\)))"
+const braceMatchingRegex = "(\\$(\\(%s(\\.(?P<var1>%s)|\\[\"(?P<var2>%s)\"\\]|\\['(?P<var3>%s)'\\])\\)))"
 
 // ValidateVariable makes sure all variables in the provided string are known
 func ValidateVariable(name, value, prefix, locationName, path string, vars sets.String) *apis.FieldError {
@@ -137,7 +136,7 @@ func ValidateVariableIsolatedP(value, prefix string, vars sets.String) *apis.Fie
 // Extract a the first full string expressions found (e.g "$(input.params.foo)"). Return
 // "" and false if nothing is found.
 func extractExpressionFromString(s, prefix string) (string, bool) {
-	pattern := fmt.Sprintf(braceMatchingRegex, prefix, parameterSubstitution)
+	pattern := fmt.Sprintf(braceMatchingRegex, prefix, parameterSubstitution, parameterSubstitution, parameterSubstitution)
 	re := regexp.MustCompile(pattern)
 	match := re.FindStringSubmatch(s)
 	if match == nil {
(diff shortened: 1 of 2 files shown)


pkg/substitution/substitution.go
```diff
@@ -47,6 +46,33 @@ func ValidateVariable(name, value, prefix, locationName, path string, vars sets.
...
+			matches = re.FindAllStringSubmatch(v, -1)
+			if len(matches) > 0 {
+				return &apis.FieldError{
+					Message: fmt.Sprintf("Invalid variable name format in value: %q. Must not start with any character other than a-z, A-Z, 0-9, or _", value),
+					// Empty path is required to make the `ViaField`, … work
+					Paths: []string{""},
+				}
+			}
+		}
+	}
```
chuangw6, line 75: Based on the three steps we discussed (pasted below), in step2, I think we don't need to do any validation here inside `ValidateVariableP`. Because validation against the param names specified in `ParamSpec` should happen inside `task_validation.go`'s `validateVariables` function, which is step 1.

> Step1: validate the format of param names at the beginning of validateVariables in `pkg/apis/pipeline/v1beta1/task_validation.go` (Implemented here https://github.com/tektoncd/pipeline/pull/4799)

> Step2: change regex to extract `NAME` for `$(params.<NAME>)` and `NAME` for `$(params["<NAME>"])` without validating anything
> - Also need to consider bracket notation in regex
> - Make sure the regex change doesn’t break others

> Step3:  Check if the extracted string is one of the param names declared in param spec. (already implemented)

afrittoli:
> # Release Notes
> ```
> NONE
> ```

Perhaps its worth adding a note in the release notes?

pkg/substitution/substitution.go
chitrangpatel: Should be ok now @chuangw6 . Let me know otherwise.

chitrangpatel:
> > # Release Notes
> > ```
> > NONE
> > ```
> 
> Perhaps its worth adding a note in the release notes?

@afrittoli I'm happy to write it. 

> > # Release Notes
> > ```
> > NONE
> > ```
> 
> Perhaps its worth adding a note in the release notes?

@afrittoli Since there are no user facing changes, I skipped it as suggested by the `Submitter Checklist` section. I wrote one anyway.

vdemeester (approved):

lbernick:
@chitrangpatel would you mind changing the PR title to something more descriptive?
In addition, a more informative release note could be something like "Added validation for parameters with bracket notation"

chitrangpatel:
> @chitrangpatel would you mind changing the PR title to something more descriptive?
> In addition, a more informative release note could be something like "Added validation for parameters with bracket notation"

Done!

lbernick:
@chitrangpatel would you mind also changing the commit message title and PR title to be "Add validation for parameters with bracket notation"? other than that looks good :)

chitrangpatel:
> @chitrangpatel would you mind also changing the commit message title and PR title to be "Add validation for parameters with bracket notation"? other than that looks good :)

Done!

lbernick:
> > @chitrangpatel would you mind also changing the commit message title and PR title to be "Add validation for parameters with bracket notation"? other than that looks good :)
> 
> Done!

don't forget the commit message title!

chitrangpatel:
> > > @chitrangpatel would you mind also changing the commit message title and PR title to be "Add validation for parameters with bracket notation"? other than that looks good :)
> > 
> > 
> > Done!
> 
> don't forget the commit message title!

Done now. (I accidentally changed the PR's first line thinking it was the title of the commit message).

Issue #3373, opened 2020-10-12
Explore entrypoint binary redacting secrets from logs


imjasonh:
Stemmed from https://github.com/tektoncd/pipeline/issues/3365 though there have been other cases it's come up.

Today, if the user specifies a `script:` without shebang, we assume `#!/bin/bash` and `set -euxo pipefail` -- we consider this a reasonable default script preamble, but `set -x` especially can surprise some users that might be using secrets in their script, and don't expect those to be emitted to output logs. One solution (as described in #3365) is to set your own preamble  or to otherwise disable `-x` so commands aren't logged.

But we might be able to do better. The entrypoint binary we inject to run the user's command (including `script`s), today [passes the sub-command's stdout directly to its own stdout](https://github.com/tektoncd/pipeline/blob/27c76d24d13e290ec9084746e2644dcf2c7631a4/cmd/entrypoint/runner.go#L38) (same with stderr). It could consider doing something smarter in this case, and take a set of byte sequences to redact from stdout/stderr, which could be passed as secret references into the Pod.

The entrypoint binary could then place this filter around stdout and stderr to replace detected secrets with some descriptive  placeholder like `[REDACTED]`, instead of passing through potentially sensitive valeus to logs.

Considerations:

* Performance overhead passing all user logs through our filter
* Limitations we should impose on filtering -- the target is presumably short-ish byte sequences (passwords and keys), not gigabytes of sensitive text
* Use cases where a user might  _want_ some secret value to be emitted through stdout/stderr, or would otherwise be surprised to find it being redacted
* User configuration to disable filtering, including rollout strategy to make it opt-in (and eventually opt-out?) -- should it be a per-run configuration, or per-installation feature flag?
* e2e tests we could write to check that filtering works

/kind feature

vdemeester:
/remove-lifecycle stale

This still needs investigation

renannprado:
@ImJasonH I'm not sure where this discussion would belong, but would be nice if Tekton also considers "secret parameters".

e.g. so that if I have a task whose output is a token (secret) I can pass to the next step seamlessly without having to create a k8s secret in order to have it `[REDACTED]` later.

ImJasonH:
[...]

PR #4837, opened 2022-05-05
Add credentials filter to entrypoint logger

Useurmind:
# Changes

This PR is related to https://github.com/tektoncd/pipeline/issues/3373. The issue was closed by your bot but the topic is still relevant to us.

It solves the problem that secret values are printed to the output log. All secrets contained in the namespace of the pod will be redacted. The pod needs a service account token with permissions to read out all secret in the namespace from the api server.

I added a credential filter to the runner app of tekton steps. It reads out all secrets from the namespace of the pod and redacts them from the output log.

This is currently a draft, I would be happy to polish it if you have interest in adopting this approach.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes
```release-note
add credential filter that can be activated via a new feature flag `enable-logging-credentials-filter`
```


Useurmind:
Any interest in this feature?

afrittoli:
Thanks @Useurmind for your PR!
I think there is interest in the feature, it may be something that we let user control via a feature flag?
I'm sorry about the slow response - between KubeCon and cdCon it's a busy conference time :D 

/cc @imjasonh @vdemeester

Useurmind:
Then I will start implementing a feature toggle for it and will see that I add some tests.

Useurmind:
So I started looking into this feature and implementing a feature flag for it. Currently I am not sure how a feature flag may be best implemented for this. 

I have found the following points:

1. feature flag config map is only read and used in controller
2. but credential filtering is done in the entrypoint binary during task runtime 
3. credentials filter needs a service account token that can read secrets and that is something which needs to be ensured when a pipeline is run

Point 1 and 2 could be solved handing down an argument or setting an environment variable to the entrypoint from the controller.

Point 3 probably must be solved by documenting the requirements for the credentials filter, so that pipeline creators can pay attention to it.

Does that approach make sense or is there an easier way?

imjasonh:
> The pod needs a service account token with permissions to read out all secret in the namespace from the api server.

This part worries me a bit. Does this feature require that all TaskRun pods now have _more_ permissions than before, to be able to tell what to redact? That seems valuable, but it also may mean that TaskRuns that, e.g., execute `kubectl get secrets` will now be able to get all those secrets, unless we take extra steps to stop them.

Could this somehow query "all accessible secrets" instead, and redact whatever it can find, instead of requiring new access to all secrets in its namespace?

I wonder if this could also lead to more load on the API server, since every TaskRun would start by fetching all secrets in its namespace, which for clusters running lots of TaskRuns could cause problems.

Not saying we shouldn't do this -- redacting secrets would be great -- just curious about some of the implications around limiting access and being mindful to reduce API traffic.

Useurmind:
Some more things to consider:

I would advice to extend this in the future so multiple secret detection algorithms can be used. Secrets have so many sources meanwhile that it becomes hard to configure one that works for everyone. This is a rather simple approach to read all secrets from the namespace. Others could be:

- Read all secrets referenced in environment variables of your pod
- Read all secrets referenced in volume mounts via csi drivers (this could depend on the csi secret provisioner)
- Read secrets from secret parameter values of the task itself
- ... more?

Should we consider implementing one of those? Then the pod would only need access to its own resource and could detect the secrets from files and environment variables. But I am not sure it is as complete as the approach above. It is also more complicated to implement.

Useurmind:
I thought about the implenentation again over the weekend. If there is currently no contact between tekton pipeline run pods and the API server it would be a grave error to require this in the future. It makes the whole design of the pipelines more complicated and the behaviour less stable and harder to predict.

Therefore, I would suggest the following approach, similar to my previous post but not the same:

- The controller knows the created pods ressources exactly, he does not even need to contact the API server to read it
- The controller can extract information about which secrets are linked into the pod, examples (not the actual secrets, just the locations):
  - environment variables: based on the secret ref value in the pod
  - files: based on CSI providers and secret refs volumes
- This information is handed to the entrypoint which can extract the secrets from the locations and use them to redact them from the output

Much more complicated than the current implementation, but more in line with the current design and easier to reason about regarding the permissions because the pipeline pod will not need any new permissions.

imjasonh:
> Btw. as far as I understand the permission concept in k8s a user can either see all secrets in a namespace or none. If you want to hide resources from a user only partially, you need to put them in different namespaces. Seeing can be limited to listing or reading them. But in the end its all or none.

You can grant access to only certain Secrets in a namespace, e.g.:

```
kind: Role
metadata:
  namespace: my-ns
  name: only-this-secret-binding
rules:
- apiGroups: [""]
  resources: ["secrets"]
  resourceNames: ["only-this-secret"]
  verbs: ["get"]
```

In this case, I believe listing secrets for the namespace will only return `only-this-secret`, as you say, and like we want, so no additional permissions are required. It does mean that if the value of `some-other-secret` snuck into logs, it wouldn't be redacted, but I think that's fine and pretty well expected.

> [...] This is a rather simple approach to read all secrets from the namespace. Others could be:
> 
> Should we consider implementing one of those? Then the pod would only need access to its own resource and could detect the secrets from files and environment variables. But I am not sure it is as complete as the approach above. It is also more complicated to implement.

I think what we have here is better than any more complicated solution that does more work to discover associated secrets. More complexity means more opportunity for bugs. 🐛 

> If there is currently no contact between tekton pipeline run pods and the API server it would be a grave error to require this in the future. It makes the whole design of the pipelines more complicated and the behaviour less stable and harder to predict. [...]

👍 Thanks for considering this! I think we should try to avoid contacting the API server if we can avoid it, but having the controller (that already knows secrets) pass them down to the entrypoint also has issues, namely, that the secret values must be passed via Pod spec, and now anybody who has pods.get permissions in that namespace can also read secrets 😢.

There are probably steps we could take to encrypt them and decrypt them using an entrypoint-only secret, but this sure is getting complex quickly 😅.

I'd be interested to hear what other folks think about this approach. @vdemeester @afrittoli @lbernick any concerns with just listing all accessible namespaced secrets, or do you think we should do more to discover them on the controller-side and securely pass them down?

Useurmind:
Today I implemented the code to tell the pod where to find secrets, it has access to. 

The controller looks for secret refs in volumes and env and also for CSI volumes referencing secrets. It saves the information in form of the names of environment variables and paths to files with secrets. The pod can then read those values and redact them.

I see opportunity for bugs but I also think it is simpler to configure and run.

This design does not transmit actual secrets. Only the places where the pod can find secrets it has attached to it. And those are the secrets that need to be redacted. Because only those can be used in the corresponding pipeline step.

What do you think of that?

imjasonh:
> The controller looks for secret refs in volumes and env and also for CSI volumes referencing secrets. It saves the information in form of the names of environment variables and paths to files with secrets. The pod can then read those values and redact them.
> [...]
> This design does not transmit actual secrets. Only the places where the pod can find secrets it has attached to it. And those are the secrets that need to be redacted. Because only those can be used in the corresponding pipeline step.
> 
> What do you think of that?

Oh gotcha, that's great, thanks for doing that! 

So in this model, the entrypoint doesn't have to list secrets at startup, and only gets information from the controller about where to find secret values in env vars, directory paths, etc. That seems like the best possible way to do this. 👍

Useurmind:
Yes the only thing remaining is the documentation and some testing, manually to convince myself and fixing one of the pipelines that does not work. Btw I am clueless what the problem of that pipeline is.

imjasonh:
> Btw I am clueless what the problem of that pipeline is.

That error is just (badly) telling you you need to `gofmt` that file. You can do that with `gofmt -w pkg/credentials/filter/secret_extraction.go`.

Useurmind:
Ok in principal I am done with the code and have validated that it works on a test instance of tekton.

But I am fighting with the build pipeline here. Seems the integration tests and coverage jobs fail (it feels random).

What formal requirements should I fullfill for this feature? I have added some docs.

Useurmind:
> Ok in principal I am done with the code and have validated that it works on a test instance of tekton.
> 
> But I am fighting with the build pipeline here. Seems the integration tests and coverage jobs fail (it feels random).
> 
> What formal requirements should I fullfill for this feature? I have added some docs.

Ok I fixed the bug in the integration tests. Now I need feedback.

Useurmind:
Ping

JeromeJu:
Thanks for the updates and sorry for getting late on this @Useurmind 

- Shall we clean up the commit messages by squashing and coming up with a more descriptive one as the current one seems outdated and not compatible with what comes out from the discussion with Jason?
- Can we have the commits rebased after the fixes?

Also some questions:
- Do we need a TEP file to consolidate the discussions for the previous open questions?

docs/logs.md
```diff
@@ -34,3 +34,10 @@ You can get execution logs using one of the following methods:
 - Get the logs using [Tekton Dashboard](https://github.com/tektoncd/dashboard).
 
 - Configure an external service to consume and display the logs. For example, [ElasticSearch, Beats, and Kibana](https://github.com/mgreau/tekton-pipelines-elastic-tutorials).
+
+
+# Logs Credential Filter
+
+Often it is the case that secret values will sneak into the ouput logs of your pipelines. You can either handle this manually by avoiding to print them in the output or use the feature for filtering secret values from the output log. To enable this feature set the `enable-logging-credentials-filter` value in the `feature-flags` config map to `"true"` (see [Customizing the Pipelines Controller behavior](install.md#customizing-the-pipelines-controller-behavior)).
+
+When the credentials filter is active the tekton entrypoint will redact the secrets found in environment variables and mounted files from secrets. It will replace them with `[REDACTED:<secret-name>]` where secret name is either the environment variable or the file name that was redacted.
```
JeromeJu, line 43: Does it make sense to add the following [design after discussions](https://github.com/tektoncd/pipeline/pull/4837#issuecomment-1142377168) to the documentations?
```
This design does not transmit actual secrets. Only the places where the pod can find secrets it has attached to it. And those are the secrets that need to be redacted. Because only those can be used in the corresponding pipeline step.
```

Useurmind: Makes sense, I will add that.

Useurmind:
> Thanks for the updates and sorry for getting late on this, please let us know if you're still willing to work on the current issue @Useurmind
> 
> * Shall we clean up the commit messages by squashing and coming up with a more descriptive one as the current one seems outdated and not compatible with what comes out from the discussion with Jason?
> * Can we have the commits rebased after the fixes?
> 
> Also some questions:
> 
> * Do we need a TEP file to consolidate the discussions for the previous open questions?

I can do the grunt work. 
Regarding the TEP: I dont know what requires a feature to provide a TEP. This feature does not require any changes to the CRDs so far, if that is a reason to create one.

pkg/credentials/filter/secret_detection.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// DetectedSecret represents a secret that was detected in the entrypoint.
+type DetectedSecret struct {
+	Name  string
+	Value []byte
```
lbernick, line 13: The docs say that the values of the secrets are not transmitted, but it looks like an example DetectedSecret would be Name: "password", Value: "my-super-secret-password"-- am I understanding this correctly? Is the idea here that the entrypoint is reading the secret files to get the contents of the secrets and prevent them from being printed?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4549,6 +4554,49 @@ func Test_validateTaskSpecRequestResources_InvalidResources(t *testing.T) {
 	}
 }
 
+func TestCreatePod_FeatureFlagVarsAreSet(t *testing.T) {
```
lbernick, line 4557: It would be helpful to see a test with a TaskRun that has a mounted secret, and to check that the envvar set on the step containers has the correct value.

pkg/apis/config/feature_flags.go
```diff
@@ -75,7 +77,13 @@ const (
 	enableAPIFields                     = "enable-api-fields"
 	sendCloudEventsForRuns              = "send-cloudevents-for-runs"
 	embeddedStatus                      = "embedded-status"
+	enableLoggingCredentialsFilter      = "enable-logging-credentials-filter"
 	enableSpire                         = "enable-spire"
+
+	// EnvEnableLoggingCredentialsFilter is the name of the environment variable that is set on the step containers
+	// when feature enable-logging-credentials-filter is enabled.
+	// #nosec G101 -- Its just the name of an environment variable that will carry credentials
+	EnvEnableLoggingCredentialsFilter = "TEKTON_FEATURE_ENABLE_LOGGING_CREDENTIALS_FILTER"
```
lbernick, line 86: It seems like this is just a boolean variable and possibly redundant with FeatureFlags.EnableLoggingCredentialsFilter-- can it be removed? Can we pass the boolean as a new optional arg to the entrypoint?

lbernick:
> I can do the grunt work. Regarding the TEP: I dont know what requires a feature to provide a TEP. This feature does not require any changes to the CRDs so far, if that is a reason to create one.

Thank you for continuing to work on this, and I'm sorry this didn't get a response for a while! Please also update the PR title and description to reflect the commit message.

In terms of whether to write a TEP: There's some detail [here](https://github.com/tektoncd/community/blob/main/teps/0001-tekton-enhancement-proposal-process.md#what-type-of-work-should-be-tracked-by-a-tep) on when we find it useful to have a design proposal. It's a bit subjective, but it seems like some design-related questions were tackled during code review which is a sign it might have been helpful to have a TEP. I'm a bit confused about the responsibilities of each of the different components and it would be helpful to see some examples, plus this change has some discussion going on re: secrets access so we want to be a bit careful. I think it would be OK not to have a TEP, but having one would give a bit more clarity on how secrets are detected, how the PR controller and entrypoint are communicating, and make sure we've thought through things like whether we only want to redact secrets coming from Workspace bindings or whether we want to handle any kinds of volumes. (TBH, I'm a bit confused about how we would detect secrets coming from e.g. a CSI volume as is stated in the docs for this change?) It's a bit more work ahead of time, but makes it a lot easier as a reviewer to say confidently that this PR should be merged.

config/config-feature-flags.yaml
```diff
@@ -81,3 +81,7 @@ data:
   # Setting this flag to "true" enables CloudEvents for Runs, as long as a
   # CloudEvents sink is configured in the config-defaults config map
   send-cloudevents-for-runs: "false"
+  # Setting this flag to "true" will enable filtering credentials from the log output
+  # of pipeline runs. This will redact all secret data values found in the pipelines namespace
+  # from the log output.
+  enable-logging-credentials-filter: "false"
```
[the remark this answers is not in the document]
Useurmind: done

docs/install.md
```diff
@@ -442,6 +441,8 @@ features](#alpha-features) to be used.
   name, kind, and API version information for each `TaskRun` and `Run` in the `PipelineRun` instead. Set it to "both" to
   do both. For more information, see [Configuring usage of `TaskRun` and `Run` embedded statuses](pipelineruns.md#configuring-usage-of-taskrun-and-run-embedded-statuses).
 
+- `enable-logging-credentials-filter`: set this flag to "true" to enable filtering credentials from the log output of pipeline runs. This will redact all secret data values found in the pipelines namespace from the log output.
```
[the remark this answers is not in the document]
Useurmind: Well this comment was from the first design. I refactored it and referenced to the detailed documentation.

pkg/credentials/filter/ring_buffer.go
```diff
@@ -0,0 +1,85 @@
...
+// RingBuffer implements a simple ring buffer with a static maximal length.
+type RingBuffer struct {
+	start         int
+	end           int
+	currentLength int
+	buffer        []byte
+}
+
+// NewRingBuffer creates a RingBuffer with the given length.
+func NewRingBuffer(length int) *RingBuffer {
```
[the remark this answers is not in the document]
Useurmind: You need to keep the last N bytes somewhere to check if they contain a secret value and only print out stuff that is clearly no secret. In our case that means everything which was not detected as secret in the ring buffer and drops out when something is added to the ring buffer. As all secret values are known, the maximum ring buffer size will be the maximum secret length.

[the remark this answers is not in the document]
Useurmind: I dont know if there is a good implementation out there that fits the bill here. But there seems to be nothing in the standard lib.

pkg/credentials/filter/credentials_filter_test.go
```diff
@@ -0,0 +1,77 @@
+package filter
```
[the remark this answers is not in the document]
Useurmind: Never heard that you change the package name for unit tests. Usually its enough to just put _test at the end of the files, so that the build ignores those files.
Also its not possible to open up a second package in a folder right?
If that is not what you meant could you explain this in more detail please.

pkg/credentials/filter/secret_detection.go
Useurmind: There are currently to processes in place which could make this a little bit confusing at first. But its quite simple:

- Secret extraction runs on controller side and extracts the environment variables names and file paths where secrets are store in a pod
- Secret detection runs on the entrypoint side and uses the extracted secret locations to actually detect the secret values from the environment variables and files.

This also means that a "detected" secret on entrypoint side contains the full secret that needs to be redacted by the entypoint. But this secret value is not transmitted to it, just the location. The secret value itself comes either from an environment variable or file content.

Useurmind:
> In terms of whether to write a TEP: There's some detail [here](https://github.com/tektoncd/community/blob/main/teps/0001-tekton-enhancement-proposal-process.md#what-type-of-work-should-be-tracked-by-a-tep) on when we find it useful to have a design proposal. It's a bit subjective, but it seems like some design-related questions were tackled ...

So I had a look at this and when looking at your questions and how to answer them a TEP is probably the best solution to do so. Therefore I started writing one. Will get back to you once I have something ready.

Useurmind:
I created a TEP for this in: https://github.com/tektoncd/community/pull/866

pkg/reconciler/taskrun/taskrun_test.go
Useurmind: I created a test for this: TestCreatePod_ExtractedSecretsAppliedToPodEnv

pkg/credentials/filter/secret_detection_test.go
```diff
@@ -0,0 +1,106 @@
+package filter
+
+import (
+	"os"
+	"testing"
+)
+
+func TestDectectionFromLocationsWithSymlinks(t *testing.T) {
```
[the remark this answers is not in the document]
Useurmind: Do you think so? Its quite simple. This is actually the easy part. Just load the secrets from files and environment variables.
The only tricky thing is that in kubernetes the secrets in the file system are made up of symlinks and that requires some little bit of extra logic to handle it. Thats the reason for this test.

pkg/credentials/filter/credentials_filter_test.go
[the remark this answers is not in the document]
lbernick: It's possible! Here's [an example](https://github.com/tektoncd/pipeline/blob/db4c05c0294ad33a2f6e7b17cacf53526139616d/pkg/workspace/apply_test.go#L1)

cmd/entrypoint/runner.go
```diff
@@ -82,30 +84,70 @@ func (rr *realRunner) Run(ctx context.Context, args ...string) error {
 	signal.Notify(rr.signals)
 	defer signal.Reset()
 
+	var err error
```
lbernick, line 87: Could you please add a test for the logic added in this file?

docs/logs.md
```diff
@@ -34,3 +34,69 @@ You can get execution logs using one of the following methods:
...
+
+The tekton controller looks for secret refs in volumes and env and also for CSI volumes referencing secrets in the definition of the pipeline pod. It saves the information in form of the names of environment variables and paths to files with secrets. The pod can then read those values and redact them.
+
+This design does not transmit actual secrets and the pod does not need access to the API sever to retrieve them. It tells the pipeline pod only the places where it can find secrets attached to it. And those are the secrets that will be redacted. Because only those can be used in the corresponding pipeline step.
+
+The credential filter will redact values contained in those secrets from the output log stream and replace them with `[REDACTED:<secret-name>]`.
+
+## Secret Detection Logic
```
lbernick, line 49: Could you also provide an example TaskRun with a secret mounted as a workspace that has the secret scrubbed?

docs/logs.md
```diff
@@ -34,3 +34,69 @@ You can get execution logs using one of the following methods:
...
+
+
+# Logs Credential Filter
+
+Often it is the case that secret values will sneak into the ouput logs of your pipelines. You can either handle this manually by avoiding to print them in the output or use the feature for filtering secret values from the output log. To enable this feature set the `enable-logging-credentials-filter` value in the `feature-flags` config map to `"true"` (see [Customizing the Pipelines Controller behavior](install.md#customizing-the-pipelines-controller-behavior)).
+
+The tekton controller looks for secret refs in volumes and env and also for CSI volumes referencing secrets in the definition of the pipeline pod. It saves the information in form of the names of environment variables and paths to files with secrets. The pod can then read those values and redact them.
+
+This design does not transmit actual secrets and the pod does not need access to the API sever to retrieve them. It tells the pipeline pod only the places where it can find secrets attached to it. And those are the secrets that will be redacted. Because only those can be used in the corresponding pipeline step.
```
lbernick, line 45: These docs include some implementation detail that I don't think is necessary for users to know about. Maybe these details would be better suited to our [developer docs](https://github.com/tektoncd/pipeline/tree/main/docs/developers)? The main detail that users would care about is that enabling this feature doesn't give the TaskRun pod access to all secrets, only the ones mounted to it.

docs/logs.md
```diff
@@ -34,3 +34,69 @@ You can get execution logs using one of the following methods:
...
+The secret locations detected by the controller are transmitted to the entrypoint as environment variable in json format with the following format:
+
+```json
+{
+  "environmentVariables": ["ENV1", "ENV2"],
+  "files": ["/path/to/secre1", "/path/to/secret2"]
+}
+```
+
+The entrypoint can then read the environment variables and file contents and redact them from the output log stream.
```
lbernick, line 64: this also seems better suited to developer docs than user docs.

pkg/apis/config/feature_flags.go
lbernick: Could the entrypoint just check to see if the downward API file is present, and if so, assume that credentials filtering is enabled?

pkg/credentials/filter/ring_buffer.go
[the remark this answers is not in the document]
lbernick: To check my understanding, the entrypoint is replacing the default stdout/stderr buffers with this new filtered credentials writer? would it be possible to just have the whole string in memory and check if it contains secret values before writing to stdout/stderr? obviously this holds a lot more data in memory.. I'm just trying to brainstorm a bit because I think adding a new buffer implementation adds quite a bit of complexity and we should do it only if it's really necessary.

pkg/credentials/filter/secret_detection_test.go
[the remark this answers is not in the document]
lbernick: Would it be possible to extract the test setup logic out of the test itself?

lbernick (commented):
This is a large PR introducing a lot of changes IMO: ring buffer, entrypoint changes, handling symlinks, detecting secret locations from mounted volumes and env vars. It's a lot of code to review in one go, and it's still hard to tell what the responsibility is of the entrypoint vs the controller. Could this be split into multiple PRs? It would also help get these changes merged more quickly.

Useurmind:
I am done changing the code to adapt it to the concept in the TEP.
Any more comments on this or can we merge it?

pkg/credentials/filter/ring_buffer.go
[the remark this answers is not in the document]
Useurmind: Having the whole string in memory would mean that your output is delayed until the run is finished, which is bad for long running jobs where you want the logs timely to get a status of the running job.
This implementation on the other hand will print out everything except the newest N bytes directly.

pkg/credentials/filter/credentials_filter_test.go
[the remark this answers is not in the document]
Useurmind: I still dont get why you do not want to test non public functions. Unit tests can also be written on non public functions right?

pkg/credentials/filter/secret_detection_test.go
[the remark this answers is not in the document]
Useurmind: I extracted the setup symlink logic into a separate function

pkg/apis/config/feature_flags.go
Useurmind: Actually that would mean you have a dynamic feature flag which could influence predictability and stability. And the feature flag is meant to ensure stability by allowing to simply turn something off. I like my feature flags simple and predictable.

cmd/entrypoint/runner.go
Useurmind: done

docs/logs.md
Useurmind: I removed some details but left the information for usage in place. I put the whole article in the developer docs

docs/logs.md
Useurmind: What exactly do you mean with "secret mounted as a workspace"?

pkg/credentials/filter/ring_buffer.go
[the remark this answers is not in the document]
lbernick: Great point. Would the Reader and Writer from bufio work? there's a Peek function (https://pkg.go.dev/bufio#Reader.Peek) that could be used to check if the next n bytes contain a secret

docs/logs.md
Useurmind: Well, I did not know that feature. 
I will have a look into it and extend the implementation.

pkg/credentials/filter/credentials_filter_test.go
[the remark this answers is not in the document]
Useurmind: Change this and used a filter_test package in all test files of this package

docs/logs.md
Useurmind: I extended the code to also test the integration with workspaces. In the end its the same working code as before I only added some more test cases.

I also discovered that I did not yet cover projected volumes with secrets and added that to the logic too.

pkg/credentials/filter/ring_buffer.go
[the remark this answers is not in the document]
Useurmind: The bufio package just implements buffered readers and writers on top of other readers and writers. That itself is not a ringbuffer implementation. A ringbuffer ist just an array that you treat as a circular buffer by jumping from end to start regularly. Its super efficient because you do not need to allocate new arrays when your buffer runs full. I doubt that a general buffered reader implementation does that. 

As I said I do not know any implementation in the standard lib. There are some OS packages like the following:
https://github.com/sahmad98/go-ringbuffer
https://github.com/smallnest/ringbuffer

But the first is too simplistic and the second even uses mutexes which we do not need here I think. Also you would make the whole implementation more "dependent" on a very small package which is perhaps not even cared for anymore which is kind of bad for these few lines of code.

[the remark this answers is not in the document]
lbernick: I agree that we don't want to add a new dependency like the one you listed, but this still adds complexity to an already large PR. Would it be possible to either use a plain slice of bytes as a buffer, or move the ring buffer implementation into its own PR?

cmd/entrypoint/runner.go
```diff
@@ -168,12 +210,19 @@ func (rr *realRunner) Run(ctx context.Context, args ...string) error {
 // be used for multiple streams if desired.
 // The behavior of the Reader is the same as io.TeeReader - reads from the pipe
 // will be written to the file.
-func newTeeReader(pipe func() (io.ReadCloser, error), path string) (*namedReader, error) {
+func newTeeReader(pipe func() (io.ReadCloser, error), path string, writer io.WriteCloser) (*namedReader, error) {
```
lbernick, line 213: This function docstring needs some updates; some of the commentary now belongs on `getFileWriter`.

pkg/credentials/filter/ring_buffer.go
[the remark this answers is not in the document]
Useurmind: I have splitted out the ring buffer into a separate pull request with more tests to cover it.

https://github.com/tektoncd/pipeline/pull/5990

pkg/credentials/filter/credentials_filter.go
```diff
@@ -0,0 +1,86 @@
...
+	_, err = w.out.Write(bytesToOut)
+	if err != nil {
+		return 0, err
+	}
+
+	return len(p), nil
+}
+
+// Close will write remaining bytes from the internal buffer to the output stream.
+// The remaining bytes will not be filtered.
```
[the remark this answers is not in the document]
Useurmind: I had the misconception that there could be no secrets anymore in the remaining bytes. But that is wrong. I implemented a test that proved exactly that and added some filtering to the close method.

cmd/entrypoint/runner.go
Useurmind: updated the doc strings

PR #4838, opened 2022-05-05
[WIP] POC for handling large results.

ScrapCodes:
1. If the result for a step is larger than size X(configurable TODO), then copy results to a remote storage(COS/S3 as configured).

2. The step sees a link embedded as value of results and maybe a flag turned on. Starts downloading result from the link. This is not automated by tekton controller, rationale is, it will consume resources on tekton controller.


wlynch:
Thanks @dibyom for pointing me at this today! cc @chitrangpatel @jerop in case you haven't seen this yet - I think this approach addresses some of the permissions concerns with the log based solution we were talking about in https://github.com/tektoncd/community/pull/745. This is using OIDC from the entrypoint flow (similar to what [I mentioned](https://github.com/tektoncd/community/pull/745#discussion_r908808192)), though it's currently interfacing with an external IdP rather than the in-cluster IdP.

@ScrapCodes great start! Main piece of feedback I have is having the user pod entrypoint interface directly with external OIDC storage buckets is going to make it hard to ensure that every service account has the access and permissions it needs, and doesn't have broad permissions to alter other Task results it shouldn't have access to.

If you haven't seen it already, you might be interested in some of the ideas detailed in [TEP-0086](https://github.com/tektoncd/community/blob/fc1dd1065a7325cee6b3df1b36f6bedb316d7d8d/teps/0086-changing-the-way-result-parameters-are-stored.md#result-sidecar---upload-results-from-sidecar) around the dedicated HTTP API - we could auth from the entrypoint with pod OIDC creds via [service account token projection](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection) like you have here to a Tekton-managed API. That API can verify the JWT matches the correct Task pod, then handle the storage and any additional auth.

Creds can be centralized in the Tekton control plane (so no need to worry about granting user pods access to external resources), and the API could be swapped out for different impl for external storage like this is doing, or updating the TaskRun object directly similar to what @chitrangpatel is doing with the [log based proof of concept](https://github.com/chitrangpatel/pipeline/tree/POC-larger-results-sidecar-logs).

ScrapCodes:
Thank you @wlynch, I agree with auth part, it is not perfect.

In order to confirm, that I have understood correctly.

* The storage should support the same credentials as tekton.
* May be using POD's credentials authenticating against a HTTP auth server running inside the controller might make it perfect. 


Will we be limited to supporting only http transport based stores(e.g. COS) as storages. Do you think that is acceptable?

On the brighter side, It does not need a sidecar !

wlynch:
> In order to confirm, that I have understood correctly.
> 
> * The storage should support the same credentials as tekton.
> * May be using POD's credentials authenticating against a HTTP auth server running inside the controller might make it perfect.
> 
> Will we be limited to supporting only http transport based stores(e.g. COS) as storages. Do you think that is acceptable?

The only bit that would need to be HTTP (though it could also be another transport mechanism like protobuf, etc. - it doesn't really matter as long as the API is agreed upon) would be from the user pod to the tekton controlled storage API. The tekton controlled storage API server can then interface with whatever backing storage it wants once it's authorized the user pod - COS, GCS, it's own local volume, etc.

If an external service supported the same result storage API and federated OIDC creds from the cluster, then you could skip the internal API and go straight to the external service.

PR #4839, opened 2022-05-05
Params: sanity check unmarshalling works with multiline json.

wlynch:
# Changes

This doesn't add any new behavior, but adds additional testing to make
sure that we can safely unmarshal data regardless of whether it is
formatted with indents / multilines.

This sprouted from [a question I had](https://github.com/tektoncd/pipeline/pull/4786#discussion_r863903111) about what would happen if we fed multiline values into this function since we're directly inspecting character values in the unmarshal function - 

https://github.com/tektoncd/pipeline/blob/30bed9f1c4fb834baf1cf12fd531b472821a3ce1/pkg/apis/pipeline/v1beta1/param_types.go#L115-L118

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/param_types_test.go
+++ b/pkg/apis/pipeline/v1beta1/param_types_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1beta1_test
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"reflect"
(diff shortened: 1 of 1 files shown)


chuangw6:
Thanks for improving this test!

pkg/apis/pipeline/v1beta1/param_types_test.go
```diff
@@ -168,23 +169,56 @@ type ArrayOrStringHolder struct {
...
+			input:  map[string]interface{}{"val": "123"},
+			result: *v1beta1.NewArrayOrString("123"),
+		},
+		{
+			input:  map[string]interface{}{"val": ""},
+			result: *v1beta1.NewArrayOrString(""),
+		},
+		{
+			input:  map[string]interface{}{"val": nil},
+			result: v1beta1.ArrayOrString{Type: v1beta1.ParamTypeArray, ArrayVal: nil},
```
Yongxuanzhang, line 185: I know in current code we will unmarshal this to array, but I think this should be a string makes more sense?

vdemeester (approved):

pkg/apis/pipeline/v1beta1/param_types_test.go
wlynch: 🤷 

That feels like a discussion to have in another PR - a null is ambiguous since it's neither a string or an array, but I can see why string might make more sense if we look at this as 1 value (string) vs many values (array).

The other option is we add enforcement to not allow null values, but all of this is outside the scope of what I wanted to do here 🙃

Yongxuanzhang: yes, I added this change in my pr, just want to bring this up here for more thoughts! 😄

PR #4841, opened 2022-05-06
TEP-0090: Refactor Validation of `Parameters` in `TaskRun` Reconciler

jerop:
# Changes

In this change, we refactor the function that validates `Parameters`
in the `TaskRun` reconciler in preparation for changes related to
`Matrix`. The refactor mainly involves splitting up the lengthy
function to make it easier to understand, maintain and reuse parts
of the logic.

There are no functional changes.
# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been filled in
(if there are no user facing changes, use release note "NONE")

# Release Notes

```release-note
NONE
```

PR #4844, opened 2022-05-06
TEP-0090: Include `Matrix` in Validation of `Parameters` in `TaskRun` Reconciler

jerop:
# Changes

The `TaskRun` reconciler validates that the needed `Parameters` are provided,
no extra `Parameters` are provide and the types of `Parameters` are matching.

Prior to this change, the reconciler did the above validation considering the
`params` field in the `PipelineTask` only. In this change, we include `matrix`
field into the validation routine described above. This includes validating
that `Parameters` in the `Matrix` are substituting `Parameters` of type `String`
in the underlying `Task`.

Related:
- https://github.com/tektoncd/pipeline/pull/4704
- https://github.com/tektoncd/pipeline/pull/4841

# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -482,7 +482,7 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
 
 	for _, rprt := range pipelineRunFacts.State {
 		if !rprt.IsCustomTask() {
-			err := taskrun.ValidateResolvedTaskResources(ctx, rprt.PipelineTask.Params, rprt.ResolvedTaskResources)
+			err := taskrun.ValidateResolvedTaskResources(ctx, rprt.PipelineTask.Params, rprt.PipelineTask.Matrix, rprt.ResolvedTaskResources)
 			if err != nil {
 				logger.Errorf("Failed to validate pipelinerun %q with error %v", pr.Name, err)
 				pr.Status.MarkFailed(ReasonFailedValidation, err.Error())
--- a/pkg/reconciler/taskrun/validate_resources.go
+++ b/pkg/reconciler/taskrun/validate_resources.go
@@ -69,16 +69,16 @@ func validateResources(requiredResources []v1beta1.TaskResource, providedResourc
 	return nil
 }
 
-func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params []v1beta1.Param) error {
+func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params []v1beta1.Param, matrix []v1beta1.Param) error {
 	neededParamsNames, neededParamsTypes := neededParamsNamesAndTypes(paramSpecs)
-	providedParamsNames := providedParamsNames(params)
+	providedParamsNames := providedParamsNames(append(params, matrix...))
 	if missingParamsNames := missingParamsNames(neededParamsNames, providedParamsNames, paramSpecs); len(missingParamsNames) != 0 {
 		return fmt.Errorf("missing values for these params which have no default values: %s", missingParamsNames)
 	}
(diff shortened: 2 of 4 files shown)


pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -141,12 +141,26 @@ func wrongTypeParamsNames(params []v1beta1.Param, neededParamsTypes map[string]v
 			wrongTypeParamNames = append(wrongTypeParamNames, param.Name)
 		}
 	}
+	for _, param := range matrix {
```
lbernick, line 144: It feels a bit strange to have code in `reconciler/taskrun` validating that matrix params aren't array params-- ideally the taskrun reconciler shouldn't have knowledge of matrix at all. could this validation be done at the pipelinerun level, and treat the matrix params as normal params by the time they get to the taskrun level? Or is it necessary to ensure that params passed via the matrix aren't declared as array params in the task?

My mental model of how this would work (based on tep 90) is that the params are declared as arrays in the pipeline, passed as matrix params in the pipeline task, and then passed from the pipeline task to the taskrun as a list of string params. If the params were declared as array params in the task, the taskrun reconciler would recognize that the string param passed didn't match the array param declared, and fail.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -482,7 +482,7 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
 
 	for _, rprt := range pipelineRunFacts.State {
 		if !rprt.IsCustomTask() {
```
lbernick, line 484: I don't see much discussion of this in TEP-0090 but I imagine we want to support passing matrixed params into custom tasks. Will they still be passed in, and this is just skipped because the run controller is expected to handle validation?

lbernick: that makes sense to me, thanks Jerop!

pkg/reconciler/taskrun/validate_resources.go
lbernick: Rather than pass in matrix params separately, any reason to not just include them in the existing slice of params to validate?

dibyom: Hmmm... a bit confused since the above comment mentions that the check for param types was removed but I still see it below. Do we still need to resolve that?

dibyom (approved):

dibyom:
I unresolved a thread because I wasn't sure it was fully resolved but otherwise this looks good to me!

pkg/reconciler/taskrun/validate_resources.go
jerop: > Rather than pass in matrix params separately, any reason to not just include them in the existing slice of params to validate?

@lbernick - The existing slice of params is checking if the type of params provided in the pipeline matches the type of params in taskspec (string to string, and array to array). However, we want the type of params provided in the matrix to be array and the type of params they are substituting in the taskspec to be string. Because the types do not match for matrix params, we can't pass it to the existing validation and that's why we have its own validation. Does that clarify?

jerop: > Hmmm... a bit confused since the above comment mentions that the check for param types was removed but I still see it below. Do we still need to resolve that?

@dibyom - Removed the validation that checks that the param in a matrix is of type array - this is already handled in the pipelinerun reconciler which was implemented in https://github.com/tektoncd/pipeline/pull/4704. We still need to validate that the param that's being substituted by the param in a matrix is of type string in the taskrun reconciler - this is the type checking that we still have here.

jerop:
> I unresolved a thread because I wasn't sure it was fully resolved but otherwise this looks good to me!

@dibyom - I'd not seen the follow up questions, addressed them, thank you!

Issue #4802, opened 2022-04-27
Some *useful* Pod status do not bubble up in TaskRun status


vdemeester:
Sligtly related to #4801, when a `Pod` doesn't fail but is in a "long-term" un-reconciliable state, such as `InvalidImageName`, the `TaskRun` controller ignore those and wait forever (aka timeout). 

This happens because it a fixable error from the `Pod` standpoint (you can mutate the `image` of a container in a `Pod`), and thus doesn't seem to be considered as a "terminal" error.

The `TaskRun` controller should pickup these kind of error (at least `InvalidImageName`, but most likely there might be a few other status that make sense) and fail with a clear message/reason based on it.

yuzp1996:
Hi, I am interested in this issue. May I take this one?

If I need any help I will mention it here. Thank you!

yuzp1996:
<img width="618" alt="image" src="https://user-images.githubusercontent.com/23007448/166193850-acac405c-a7d6-42a1-aac1-ba6b54f90066.png">

<img width="648" alt="image" src="https://user-images.githubusercontent.com/23007448/166193874-370605d3-6e1e-461d-8346-72cf069dd55e.png">

I think I have reproduced the error here. The pod is in the status of ImagePullBackOff while the related Taskrun is pending.

I think the expected result is that the taskrun should be failed with the message or reason of ImagePullBackOff.

If  I misunderstand something please tell me. Thank you!

yuzp1996:
Hi, I check the code and find maybe we can do some improvement here. 
<img width="1462" alt="image" src="https://user-images.githubusercontent.com/23007448/166230714-839de56f-3d1f-48d4-85a9-da71d8e46968.png">

I think this is not a bug fix but a new feature. Maybe I need to provide a TEP to describe the goal that we want to achieve and the code change that I need to make.

I will be very happy if you can give me some advice. @vdemeester  Thanks!

yuzp1996:
OK, thanks!  I will do it this week.

yuzp1996:
This is all the [image Error](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/images/types.go#L26) I found in the k8s kubelet code. 

<img width="814" alt="image" src="https://user-images.githubusercontent.com/23007448/167320277-9fe3c4d7-d549-422c-9931-23fe4ab790c0.png">

lbernick:
/close

fixed in #4846, thanks @yuzp1996!

PR #4846, opened 2022-05-08
Bubble up the image related error reason to taskrun status

yuzp1996:
# Changes

Bubble up the image related error reason to taskrun status
    
When pod is pending because of failure of pulling image the users
can not get the clear reason or message from the taskrun status.
    
Now I will check the pod status and judge if the error is caused
by image if so the error reason and message will bubble up to
taskrun status. Then user can know what happened though the taskrun
status.
    
Related issue: https://github.com/tektoncd/pipeline/issues/4802

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/pod/status.go
+++ b/pkg/pod/status.go
@@ -53,6 +53,9 @@ const (
 	// to resource constraints on the node
 	ReasonExceededNodeResources = "ExceededNodeResources"
 
+	// ReasonPullImageFailed indicates that the TaskRun's pod failed to pull image
+	ReasonPullImageFailed = "PullImageFailed"
+
 	// ReasonCreateContainerConfigError indicates that the TaskRun failed to create a pod due to
 	// config error of container
 	ReasonCreateContainerConfigError = "CreateContainerConfigError"
@@ -314,6 +317,8 @@ func updateIncompleteTaskRunStatus(trs *v1beta1.TaskRunStatus, pod *corev1.Pod) 
 			markStatusRunning(trs, ReasonExceededNodeResources, "TaskRun Pod exceeded available resources")
 		case isPodHitConfigError(pod):
 			markStatusFailure(trs, ReasonCreateContainerConfigError, "Failed to create pod due to config error")
+		case isPullImageError(pod):
+			markStatusRunning(trs, ReasonPullImageFailed, getWaitingMessage(pod))
 		default:
 			markStatusRunning(trs, ReasonPending, getWaitingMessage(pod))
 		}
@@ -402,6 +407,34 @@ func IsPodExceedingNodeResources(pod *corev1.Pod) bool {
 func isPodHitConfigError(pod *corev1.Pod) bool {
 	for _, containerStatus := range pod.Status.ContainerStatuses {
 		if containerStatus.State.Waiting != nil && containerStatus.State.Waiting.Reason == ReasonCreateContainerConfigError {
+			return true
+		}
+	}
+	return false
+}
+
+// isPullImageError returns true if the Pod's status indicates there are any error when pulling image
+func isPullImageError(pod *corev1.Pod) bool {
+	for _, containerStatus := range pod.Status.ContainerStatuses {
+		if containerStatus.State.Waiting != nil && isImageErrorReason(containerStatus.State.Waiting.Reason) {
+			return true
+		}
+	}
+	return false
+}
+
+func isImageErrorReason(reason string) bool {
+	// Reference from https://github.com/kubernetes/kubernetes/blob/a1c8e9386af844757333733714fa1757489735b3/pkg/kubelet/images/types.go#L26
+	imageErrorReasons := []string{
+		"ImagePullBackOff",
+		"ImageInspectError",
+		"ErrImagePull",
+		"ErrImageNeverPull",
+		"RegistryUnavailable",
+		"InvalidImageName",
+	}
+	for _, imageReason := range imageErrorReasons {
+		if imageReason == reason {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


vdemeester (approved):

afrittoli:
@yuzp1996 Thanks for your PR! 
I updated the PR description to meet the template, please use the template for next PR, it helps a lot when PRs and producing new releases.

yuzp1996:
> @yuzp1996 Thanks for your PR! I updated the PR description to meet the template, please use the template for next PR, it helps a lot when PRs and producing new releases.

Thanks for your reminder and correction. I will use the template for my next PR.

yuzp1996:
Hi, @vdemeester  @afrittoli @imjasonh I am sorry to bother you guys. Is there anything I need to do to merge this PR? 

I have read the [contributor document](https://github.com/tektoncd/community/blob/main/process.md#contributor) and maybe there needs a label lgtm? 

Is anything I need to do please tell me. Thanks!

pkg/pod/status.go
```diff
@@ -408,6 +413,34 @@ func isPodHitConfigError(pod *corev1.Pod) bool {
...
+	for _, containerStatus := range pod.Status.ContainerStatuses {
+		if containerStatus.State.Waiting != nil && isImageErrorReason(containerStatus.State.Waiting.Reason) {
+			return true
+		}
+	}
+	return false
+}
+
+func isImageErrorReason(reason string) bool {
+	// Reference from https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/images/types.go#L26
```
lbernick, line 427: would you mind using a permalink here?

pkg/pod/status.go
```diff
@@ -314,6 +317,8 @@ func updateIncompleteTaskRunStatus(trs *v1beta1.TaskRunStatus, pod *corev1.Pod)
 			markStatusRunning(trs, ReasonExceededNodeResources, "TaskRun Pod exceeded available resources")
 		case isPodHitConfigError(pod):
 			markStatusFailure(trs, ReasonCreateContainerConfigError, "Failed to create pod due to config error")
+		case isPullImageError(pod):
```
lbernick, line 320: Wondering if it would make sense to mark the taskrun status failed for some of these errors? e.g. "invalidImageName" is unlikely to succeed or be retryable

pkg/pod/status.go
yuzp1996: Sorry, what do you mean about permalink? I guess you mean it is not a  good choice put the link here?

Maybe putting the link in the commit message is a better way?  Putting the message in the comment or the commit message is the only two ways  I can think of. 

If you have a better choice, I‘d love to use it. 😁

pkg/pod/status.go
yuzp1996: Yes, I had thought about this, I tried generating an invalidImageName error for the pod and trying to check the pod's phase. I found that the phase is Pending and not Failed.

image


I think this is because of the [following reason](https://github.com/tektoncd/pipeline/issues/4802#issue-1217467406).  
"This happens because it a fixable error from the Pod standpoint (you can mutate the image of a container in a Pod), and thus doesn't seem to be considered as a "terminal" error."

So I thought maybe we could borrow the status of the pod and not mark the taskrun failed?

lbernick: that could definitely work, and I'm happy to approve the PR as is since that's a reasonable route, however the user isn't really meant to interact with the taskrun pod, i.e. the taskrun controller would be responsible for updating the pod and it wouldn't know how to do so here, so I think it's most reasonable to mark as failed.
Curious what others think about this

lbernick: I thought about this a bit more:

- If we mark the TaskRun as pending still, it's not a great user experience because the user isn't intended to interact with the pod and modify the image name, and it'll eventually fail due to timeout (as I mentioned previously)
- However, if we mark the TaskRun as failed, the pod will still be running, and if in theory someone edited the image, the pod could run to completion. If that happened, we wouldn't want the TaskRun to display as failed.

I think the best approach for right now is to mark the TaskRun as pending, preserving existing behavior. Then, we should add functionality that cancels the TaskRun when the pod encounters an error such as InvalidImageName (not as part of this PR-- can just create an issue to track). I think someone is implementing this functionality to cancel pods when the TaskRuns time out (#4618) which could probably be reused here.

What do you think? If you agree, happy to LGTM.

yuzp1996: Yes! I agree with you.  👍

chmouel: I have started to see this issue a lot on my clusters, where InvalidImageName end up having the PR running forever (or unti default  timeout of 60mn). I think it would be a better user experience if the taskruns would get canceled when encountering those.

lbernick: @chmouel would you mind creating an issue to track this?

chmouel: yep here: https://github.com/tektoncd/pipeline/issues/6105

lbernick: awesome thank you!

afrittoli: Uhm, I don't think I would like to support in Tekton a case where the users edit the underlying `Pod` to fix an `ErrInvalidImageName` error.  I would consider that as a permanent error from Tekton perspective. 
However, I do agree that we should not let a `Pod` running when the `TaskRun` is marked as failed. 
If k8s behaviour is to wait even in case of invalid image name, we would then need to stop the pod somehow because we could fail the `TaskRun`.

PR #4851, opened 2022-05-09
Deprecate unusable Step/StepTemplate fields

### LINKED ISSUE #4737: Owning our *Container* spec

lbernick:
# Changes
Some fields of Step and StepTemplate only exist because they were inherited from
k8s Container, but they don't make sense in a batch workflow context and/or aren't
actually supported in Tekton. No changes to Sidecar.

This commit deprecates the following fields:
- Lifecycle, Ports, ReadinessProbe, LivenessProbe, StartupProbe, StdIn, StdInOnce, TTY:
these don't really have a use in the context of Step
- TerminationMessagePath and TerminationMessagePolicy: these are extra confusing since Results are
currently written as TerminationMessages
- StepTemplate.Name: Steps must have different names.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Deprecate unusable/unsupported fields of Step and StepTemplate
```

Closes #4737

### DIFF
--- a/pkg/apis/pipeline/v1beta1/container_types.go
+++ b/pkg/apis/pipeline/v1beta1/container_types.go
@@ -46,6 +46,7 @@ type Step struct {
 	// Cannot be updated.
 	// +optional
 	WorkingDir string `json:"workingDir,omitempty" protobuf:"bytes,5,opt,name=workingDir"`
+	// Deprecated. This field will be removed in a future release.
 	// List of ports to expose from the container. Exposing a port here gives
 	// the system additional information about the network connections a
 	// container uses, but is primarily informational. Not specifying a port here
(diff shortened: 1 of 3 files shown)


vdemeester:
I am late to the party but 2 things:
- Can we rename the `go` fields prefixing them with `Deprecated` ? This won't have any impact on the API but could be used with something like https://github.com/knative/pkg/pull/2498/files#diff-c24a96cf293a6a5554a2148a1f8eec1e3c1e5dc27119181a565a5239698f6843L39
- Do we want to add an entry to the deprecation file to give a timeframe on when they will be removed ?

lbernick:
> * Can we rename the `go` fields prefixing them with `Deprecated` ? This won't have any impact on the API but could be used with something like https://github.com/knative/pkg/pull/2498/files#diff-c24a96cf293a6a5554a2148a1f8eec1e3c1e5dc27119181a565a5239698f6843L39

Not sure what you mean by this; are you suggesting renaming the struct types? They are owned by k8s. Or are you suggesting renaming the field names but keeping the json tags the same so that serialization/deserialization does not change?

> * Do we want to add an entry to the deprecation file to give a timeframe on when they will be removed ?

I'm actually not sure what an appropriate timeline is here. Our compatibility policy says 9 months but it seems silly to wait 9 months to remove fields that aren't even currently usable.

vdemeester:
> Or are you suggesting renaming the field names but keeping the json tags the same so that serialization/deserialization does not change?

Yes, this is what I am suggesting 👼🏼 

> I'm actually not sure what an appropriate timeline is here. Our compatibility policy says 9 months but it seems silly to wait 9 months to remove fields that aren't even currently usable.

I tend to agree with that, but I don't really think we have a alternative.

lbernick:
> > Or are you suggesting renaming the field names but keeping the json tags the same so that serialization/deserialization does not change?
> 
> Yes, this is what I am suggesting 👼🏼
> 
> > I'm actually not sure what an appropriate timeline is here. Our compatibility policy says 9 months but it seems silly to wait 9 months to remove fields that aren't even currently usable.
> 
> I tend to agree with that, but I don't really think we have a alternative. At least, we need to communicate what is the time schedule for those to be removed.

done in https://github.com/tektoncd/pipeline/pull/4866

PR #4861, opened 2022-05-11
[TEP-0075] Validate against using the whole object in task steps

chuangw6:
# Changes
According to [TEP-0075](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#variable-replacement-with-object-params):
When providing values for strings, Task and Pipeline authors can
access individual attributes of an object param; but they cannot access
the object as whole.

This PR makes sure that the object param cannot be used as whole when providing values for strings in task **steps**.

Also modified `extractVariablesFromString` function a bit to recognize
`params.foo.bar` reference that was treated as invalid previously.

# Submitter Checklist

  functionality, content, code)

# Release Notes
```
Added validation against using the whole object in fields of a task step.
For example, the following reference to the whole object param is not allowed.
steps:
  - name: my-step
     image: $(params.myObjectParam[*])


This is part of TEP-0075. In order to have the changes in this PR to be applied, 
users have to enable the `alpha` feature flag.

```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation_test.go
@@ -194,6 +194,24 @@ func TestTaskSpecValidate(t *testing.T) {
 			}},
 		},
 	}, {
+		name: "valid object template variable",
+		fields: fields{
+			Params: []v1beta1.ParamSpec{{
+				Name: "gitrepo",
+				Type: v1beta1.ParamTypeObject,
+				Properties: map[string]v1beta1.PropertySpec{
+					"url":    {},
+					"commit": {},
+				},
+			}},
+			Steps: []v1beta1.Step{{
+				Name:       "do-the-clone",
+				Image:      "some-git-image",
+				Args:       []string{"-url=$(params.gitrepo.url)", "-commit=$(params.gitrepo.commit)"},
+				WorkingDir: "/foo/bar/src/",
+			}},
+		},
+	}, {
 		name: "valid star array template variable",
 		fields: fields{
 			Params: []v1beta1.ParamSpec{{
@@ -899,6 +917,94 @@ func TestTaskSpecValidateError(t *testing.T) {
 		},
 		expectedError: apis.FieldError{
 			Message: `variable is not properly isolated in "not isolated: $(params.baz[*])"`,
+			Paths:   []string{"steps[0].args[0]"},
+		},
+	}, {
+		name: "object used in a string field",
+		fields: fields{
+			Params: []v1beta1.ParamSpec{{
+				Name: "gitrepo",
+				Type: v1beta1.ParamTypeObject,
+				Properties: map[string]v1beta1.PropertySpec{
+					"url":    {},
+					"commit": {},
+				},
+			}},
+			Steps: []v1beta1.Step{{
+				Name:       "do-the-clone",
+				Image:      "$(params.gitrepo)",
+				Args:       []string{"echo"},
+				WorkingDir: "/foo/bar/src/",
+			}},
+		},
+		expectedError: apis.FieldError{
+			Message: `variable type invalid in "$(params.gitrepo)"`,
+			Paths:   []string{"steps[0].image"},
+		},
+	}, {
+		name: "object star used in a string field",
+		fields: fields{
+			Params: []v1beta1.ParamSpec{{
+				Name: "gitrepo",
+				Type: v1beta1.ParamTypeObject,
+				Properties: map[string]v1beta1.PropertySpec{
+					"url":    {},
+					"commit": {},
+				},
+			}},
+			Steps: []v1beta1.Step{{
+				Name:       "do-the-clone",
+				Image:      "$(params.gitrepo[*])",
+				Args:       []string{"echo"},
+				WorkingDir: "/foo/bar/src/",
+			}},
+		},
+		expectedError: apis.FieldError{
+			Message: `variable type invalid in "$(params.gitrepo[*])"`,
+			Paths:   []string{"steps[0].image"},
+		},
+	}, {
+		name: "object used in a field that can accept array type",
+		fields: fields{
+			Params: []v1beta1.ParamSpec{{
+				Name: "gitrepo",
+				Type: v1beta1.ParamTypeObject,
+				Properties: map[string]v1beta1.PropertySpec{
+					"url":    {},
+					"commit": {},
+				},
+			}},
+			Steps: []v1beta1.Step{{
+				Name:       "do-the-clone",
+				Image:      "myimage",
+				Args:       []string{"$(params.gitrepo)"},
+				WorkingDir: "/foo/bar/src/",
+			}},
+		},
+		expectedError: apis.FieldError{
+			Message: `variable type invalid in "$(params.gitrepo)"`,
+			Paths:   []string{"steps[0].args[0]"},
+		},
+	}, {
+		name: "object star used in a field that can accept array type",
+		fields: fields{
+			Params: []v1beta1.ParamSpec{{
+				Name: "gitrepo",
+				Type: v1beta1.ParamTypeObject,
+				Properties: map[string]v1beta1.PropertySpec{
+					"url":    {},
+					"commit": {},
+				},
+			}},
+			Steps: []v1beta1.Step{{
+				Name:       "do-the-clone",
+				Image:      "some-git-image",
+				Args:       []string{"$(params.gitrepo[*])"},
+				WorkingDir: "/foo/bar/src/",
+			}},
+		},
+		expectedError: apis.FieldError{
+			Message: `variable type invalid in "$(params.gitrepo[*])"`,
 			Paths:   []string{"steps[0].args[0]"},
 		},
 	}, {
--- a/pkg/substitution/substitution.go
+++ b/pkg/substitution/substitution.go
@@ -109,6 +109,31 @@ func ValidateVariableProhibitedP(value, prefix string, vars sets.String) *apis.F
 	return nil
 }
 
+// ValidateEntireVariableProhibitedP verifies that values of object type are not used as whole.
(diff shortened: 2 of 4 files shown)


pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -855,6 +873,50 @@ func TestTaskSpecValidateError(t *testing.T) {
 			Message: `variable is not properly isolated in "not isolated: $(params.baz[*])"`,
 			Paths:   []string{"steps[0].args[0]"},
 		},
```
ywluogg, line 875: Maybe add another test like this for object?

pkg/substitution/substitution_test.go
```diff
@@ -174,6 +174,22 @@ func TestValidateVariablePs(t *testing.T) {
 			vars:   sets.NewString("baz", "foo"),
 		},
 		expectedError: nil,
+	}, {
+		name: "valid usage of an individual attribute of an object param",
+		args: args{
+			input:  "--flag=$(params.objectParam.key1)",
```
ywluogg, line 180: Maybe worths adding more valid tests like `params.['objectParam'].key1` and `params.["objectParam"].key1`

chuangw6: > Maybe worths adding more valid tests like `params.['objectParam'].key1` and `params.["objectParam"].key1`

I think both `params.['objectParam'].key1` and `params.["objectParam"].key1` are invalid reference for object keys according to the [tep](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#variable-replacement-with-object-params).

pkg/apis/pipeline/v1beta1/task_validation_test.go
chuangw6: Just added a test named `object star used in unaccepted field (not isolated)`. 

But I think this might be a bit duplicated with the [existing two tests](https://github.com/tektoncd/pipeline/pull/4861/files#diff-624ee2137fe5ef32422ce97821c958d84a30127121ae8a9183257ed85f417d03R877-R919) that I added before. Unlike array case, using the whole object param will be invalid no matter if it's isolated according to this [tep](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#variable-replacement-with-object-params). 
> When populating a string field, It would be invalid (at least initially) to attempt to do variable replacement on the entire gitrepo object ($(params.gitrepo)). If we choose to support this later we could potentially replace the value with the json representation of the object directly.


Let me know what you think. Happy to remove or keep.

pkg/substitution/substitution_test.go
ywluogg: > I think both params.['objectParam'].key1 and params.["objectParam"].key1 are invalid reference for object keys according to the [tep](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#variable-replacement-with-object-params).

Ah why is that? I the [tep](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#variable-replacement-with-object-params) only covers several examples for referring to keys in params. Ideally we should extend the key references based on what we have for [params](https://github.com/tektoncd/pipeline/blob/main/docs/variables.md#variables-available-in-a-pipeline)

pkg/apis/pipeline/v1beta1/task_validation_test.go
ywluogg: Ah I meant to say putting it in a valid field (like a object type of params) but not isolated

ywluogg:
Can you also add a bit more scope that this PR will be focused on checks of valid/invalid substitutions for keys, and checks of  invalid substitutions for entire object? I think you plan to add checks of valid substitutions for entire objects later, right?

pkg/substitution/substitution_test.go
chuangw6: For existing bracket notation, the [doc](https://github.com/tektoncd/pipeline/blob/main/docs/variables.md#variables-available-in-a-pipeline) seems only mentioning something like `params["<NAME>"]` instead of `params.["<NAME>"]`. [TEP-0080](https://github.com/tektoncd/community/blob/main/teps/0080-support-domainscoped-parameterresult-names.md#proposal) added support for bracket notation, which is meant to support names that contain dots (.).


If we use bracket notation for object, it seems very confusing to users i.e. 
- `params['objectParam'][*]` (array type seems not supporting this?)
- `params['objectParam'].key1` (also do we want to consider `params['objectParam']['key1']`).  

So I'd lean more towards what the [tep-75](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#variable-replacement-with-object-params) suggested (at least initially).

chuangw6:
> Just wanted to confirm that this PR will still be focused on substituted field is a string, and the substituting param is an object, is it?

This PR makes sure object param cannot be used as whole in [any part of a task step](https://github.com/tektoncd/pipeline/pull/4861/files#diff-988b00fa4f188a96ff5901c0bbc9f65d4a8355417185d4fe0ae97d8c14e04417R416-R434). 
As [tep75](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#variable-replacement-with-object-params), the whole object can only be used when **providing value for other objects**. That validation will be part of pipeline validation (not in this pr).

pkg/substitution/substitution_test.go
ywluogg: > For existing bracket notation, the [doc](https://github.com/tektoncd/pipeline/blob/main/docs/variables.md#variables-available-in-a-pipeline) seems only mentioning something like params["<NAME>"] instead of params.["<NAME>"]. [TEP-0080](https://github.com/tektoncd/community/blob/main/teps/0080-support-domainscoped-parameterresult-names.md#proposal) added support for bracket notation, which is meant to support names that contain dots (.).

Ah my bad that there is a typo in the comment. I meant to say `params['objectParam'].key1` and `params["objectParam"].key1`

ywluogg: > > For existing bracket notation, the [doc](https://github.com/tektoncd/pipeline/blob/main/docs/variables.md#variables-available-in-a-pipeline) seems only mentioning something like params[""] instead of params.[""]. [TEP-0080](https://github.com/tektoncd/community/blob/main/teps/0080-support-domainscoped-parameterresult-names.md#proposal) added support for bracket notation, which is meant to support names that contain dots (.).

Ah my bad that there is a typo in the comment. I meant to say `params['objectParam'].key1` and `params["objectParam"].key1`

> params['objectParam']['key1']

Yes I think this should be added

ywluogg:
> This PR makes sure object param cannot be used as whole in [any part of a task step](https://github.com/tektoncd/pipeline/pull/4861/files#diff-988b00fa4f188a96ff5901c0bbc9f65d4a8355417185d4fe0ae97d8c14e04417R416-R434).
As [tep75](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#variable-replacement-with-object-params) proposed, the whole object can only be used when providing value for other objects. That validation will be part of pipeline validation (not in this pr).

Oh I think I meant to ask - if this PR would be focused on validations of whole object substitutions, when the substituted field is a string? Because I don't see the validations of substituted param is an array and substituting param is an object. In `validateStepObjectUsageAsWhole`, all the checks are for those string fields

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -855,6 +873,72 @@ func TestTaskSpecValidateError(t *testing.T) {
...
+				Args:       []string{"$(params.gitrepo)"},
+				WorkingDir: "/foo/bar/src/",
+			}},
+		},
+		expectedError: apis.FieldError{
+			Message: `variable type invalid in "$(params.gitrepo)"`,
+			Paths:   []string{"steps[0].args[0]"},
+		},
+	}, {
+		name: "object star used in unaccepted field",
```
ywluogg, line 899: This is a case for substituted field being an array, and substituting field is an object. Can you also add one for substituted field being a string? "object star used in unaccepted field" also worths changing a name. I think it would be helpful if you can organize the test in such way:
1. object used in array field
2. object used in string field
3. object star used in array field
4. object star used in string field
...
This could help me identifying missing cases easier.

pkg/substitution/substitution_test.go
chuangw6: This is out of the scope of this pr. Let's discuss offline later about whether we want to support chained bracket notation for object name (and keys).

pkg/apis/pipeline/v1beta1/task_validation_test.go
chuangw6: sg. done!

pkg/apis/pipeline/v1beta1/task_validation_test.go
chuangw6: > Ah I meant to say putting it in a valid field (like a object type of params) but not isolated

I'm still confused. Do you mean using object param in a valid field which can accept object as whole?

chuangw6: > Ah I meant to say putting it in a valid field (like a object type of params) but not isolated

I'm still confused. Do you mean using object param in a valid field which can accept object as whole?

ywluogg (approved):

pkg/substitution/substitution.go
```diff
@@ -180,38 +197,53 @@ func extractVariablesFromString(s, prefix string) ([]string, bool, string) {
...
 			}
-
 		}
 	}
 	return vars, true, errString
 }
 
+func extractEntireVariablesFromString(s, prefix string) ([]string, bool) {
+	pattern := fmt.Sprintf(braceMatchingRegex, prefix, parameterSubstitution, parameterSubstitution, parameterSubstitution)
+	re := regexp.MustCompile(pattern)
```
lbernick, line 225: this looks like something that could panic; is there a version that returns an error instead of panicking?

abayer:
This needs a release note.

Issue #4863, opened 2022-05-12
Raise the kaniko tests timeout from `2m`


mattmoor:
# Expected Behavior

`kaniko` tests reliably pass.

# Actual Behavior

We frequently see the `kaniko` tests flake downstream, and most of the ones I've looked at are timeouts.

Quoting @imjasonh "Nobody should expect kaniko to do anything reliably in 2m" 🤣 

cc @abayer since this touches on flakes (not sure if you've seen this in your adventures)

PR #4869, opened 2022-05-13
Increase timeouts for some workspace YAML tests

imjasonh:
I've seen these fail due to timeouts lately ([example](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4865/pull-tekton-pipeline-alpha-integration-tests/1525124401899507713/)), and 60s seems like a very short timeout for a test that might be running in a cluster alongside lots of other tests.

This bumps the timeout from 60s -> 5m.

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/examples/v1beta1/pipelineruns/alpha/isolated-workspaces.yaml
+++ b/examples/v1beta1/pipelineruns/alpha/isolated-workspaces.yaml
@@ -17,7 +17,7 @@ apiVersion: tekton.dev/v1beta1
 metadata:
   generateName: isolated-workspaces-
 spec:
-  timeout: 60s
+  timeout: 5m
   workspaces:
   - name: ssh-credentials
     secret:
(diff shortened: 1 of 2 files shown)


abayer:
I'm actually wondering if we should limit the parallelism of the YAML tests - I'm gonna try that in an experiment PR.

abayer:

I’m running the integration tests a few times to see how the relevant tests perform over multiple runs, similar to what I’m doing on #4841. I’m thinking the likely end result is going to be increased timeouts (also to some non-YAML tests, namely the sidecar ones and kaniko ones, see #4863 for the latter) and an explicit parallelism limit combined. That won’t solve all of our flakes - there are always going to be cases of proxy.golang.org or GitHub API burps, the GKE project hitting quotas, and whatever the root cause is for the sporadic runs taking 1.5+ hours and just barfing out, but the timeout and parallelism changes will, I believe, drastically reduce the recurrence of particular tests flaking out.

abayer:
I’m also gonna run `pull-pipeline-kind-k8s-v1-21-e2e` a bunch on this PR, since that does give us a better sense of how much of the instability is GKE related…annoyingly I can’t run it effectively in my explicit parallelism PR because it doesn’t match the changed files filter!

abayer:
Bleh, the kind run times out waiting for installation of Pipeline to finish anyway.

PR #4867, opened 2022-05-12
[TEP-0075] Extract out the validation of object keys

chuangw6:
# Changes

Prior to this commit, the validation against the required and
provided keys for object type was part of the `validateObjectUsage`
function. That makes the purpose of `validateObjectUsage` confusing.

In this commit, a new helper function is created to validate that
separately.

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ ] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [ ] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [ ] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [ ] Release notes block below has been filled in
(if there are no user facing changes, use release note "NONE")

# Release Notes

PR #4872, opened 2022-05-13
Remove duplicate code for validating params types and defaults

chuangw6:
# Changes
Prior to this commit, `PipelineSpec`'s `ParamSpec` type validation
was re-implemented in `pipeline_validation.go`, which actually has
been implemented in `task_validation.go` for the `ParamSpec`'s type
validation.

So in this commit, we replace the duplicate code with existing function.

# Submitter Checklist

  functionality, content, code)

# Release Notes

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation.go
@@ -59,8 +59,8 @@ func (ps *PipelineSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	errs = errs.Also(validateGraph(ps.Tasks))
 	errs = errs.Also(validateParamResults(ps.Tasks))
 	// The parameter variables should be valid
-	errs = errs.Also(validatePipelineParameterVariables(ps.Tasks, ps.Params).ViaField("tasks"))
-	errs = errs.Also(validatePipelineParameterVariables(ps.Finally, ps.Params).ViaField("finally"))
+	errs = errs.Also(validatePipelineParameterVariables(ctx, ps.Tasks, ps.Params).ViaField("tasks"))
+	errs = errs.Also(validatePipelineParameterVariables(ctx, ps.Finally, ps.Params).ViaField("finally"))
 	errs = errs.Also(validatePipelineContextVariables(ps.Tasks).ViaField("tasks"))
 	errs = errs.Also(validatePipelineContextVariables(ps.Finally).ViaField("finally"))
 	errs = errs.Also(validateExecutionStatusVariables(ps.Tasks, ps.Finally))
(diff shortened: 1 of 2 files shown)


pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -125,28 +125,15 @@ func validatePipelineWorkspacesUsage(wss []PipelineWorkspaceDeclaration, pts []P
...
-		// If a default value is provided, ensure its type matches param's declared type.
-		if (p.Default != nil) && (p.Default.Type != p.Type) {
-			errs = errs.Also(apis.ErrGeneric(fmt.Sprintf("\"%v\" type does not match default value's type: \"%v\"", p.Type, p.Default.Type),
-				"type", "default.type").ViaFieldKey("params", p.Name))
-		}
+	// validates all the types within a slice of ParamSpecs
+	errs = errs.Also(ValidateParameterTypes(ctx, params).ViaField("params"))
 
+	for _, p := range params {
+		// this if chunk will be also removed after https://github.com/tektoncd/pipeline/pull/4806 gets merged.
```
lbernick, line 136: I'd remove this comment and just rebase if that PR is merged first.

lbernick (commented):
Thanks Chuang! I think it could make sense to move parameter validation out of task_validation.go and pipeline_validation.go into a new param_validation.go file since a lot of the param code seems to be duplicated between pipeline and task validation.

chuangw6:
> Thanks Chuang! I think it could make sense to move parameter validation out of task_validation.go and pipeline_validation.go into a new param_validation.go file since a lot of the param code seems to be duplicated between pipeline and task validation.

Hi @lbernick , 
Great point!! I totally agree. I find that duplicate code regarding param validation exists not only here in validation webhook but also in the reconciler level i.e. those [two](https://github.com/tektoncd/pipeline/blob/2e09b385af75d62412f2a9bc13d36ec58790aafc/pkg/reconciler/pipelinerun/resources/validate_params.go#L26-L75) are doing same thing as the code [here](https://github.com/tektoncd/pipeline/blob/2e09b385af75d62412f2a9bc13d36ec58790aafc/pkg/reconciler/taskrun/validate_resources.go#L91-L121). 

Would you recommend we also refactor reconciler level to extract a dedicated file? If so, I am happy to do that after finishing the refactor of this validation webhook level.

chuangw6:
> Thanks Chuang! I think it could make sense to move parameter validation out of task_validation.go and pipeline_validation.go into a new param_validation.go file since a lot of the param code seems to be duplicated between pipeline and task validation.

Just thinking that moving param validation related functions to a new file will be a big change for other existing PRs like #4867 #4861 since they are modifying the param validation functions in the original file. 
How about we move them to the new file after they get merged? @lbernick

lbernick:
> Great point!! I totally agree. I find that duplicate code regarding param validation exists not only here in validation webhook but also in the reconciler level i.e. those [two](https://github.com/tektoncd/pipeline/blob/2e09b385af75d62412f2a9bc13d36ec58790aafc/pkg/reconciler/pipelinerun/resources/validate_params.go#L26-L75) are doing same thing as the code [here](https://github.com/tektoncd/pipeline/blob/2e09b385af75d62412f2a9bc13d36ec58790aafc/pkg/reconciler/taskrun/validate_resources.go#L91-L121). But it's a bit tricky for the reconciler level since they are living in different packages i.e. taskrun, pipelinerun.

> Would you recommend we also refactor reconciler level to extract a dedicated file? If so, I am happy to do that after finishing the refactor of this validation webhook level.

That could definitely make sense! We have something similar for workspaces I think. Agree that it should be a separate PR.


> Just thinking that moving param validation related functions to a new file will be a big change for other existing PRs like #4867 #4861 since they are modifying the param validation functions in the original file. How about we move them to the new file after they get merged? 

Up to you! I think if you are going to the trouble of making a cleanup PR it's worth putting these functions in the place that makes the most sense (right now this PR is having the pipeline validation use a function defined in task validation which feels a bit strange to me). Rebasing and resolving merge conflicts is sort of a fact of life :) but if you'd prefer to wait that's OK too.

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -1500,7 +1501,7 @@ func TestValidatePipelineParameterVariables_Failure(t *testing.T) {
 		}},
 		expectedError: apis.FieldError{
 			Message: `"array" type does not match default value's type: "string"`,
-			Paths:   []string{"params[foo].default.type", "params[foo].type"},
+			Paths:   []string{"params.foo.default.type", "params.foo.type"},
```
ywluogg, line 1504: Maybe we don't want to change the existing tests - we want to add an extra sets of `params.foo.default.type` type of tests? Are we not supporting reference such as `params[foo].type` anymore?

chuangw6: It's not reference. Just hint for path in the error message.

ywluogg: AH I see. You can close this thread

ywluogg:
> Up to you! I think if you are going to the trouble of making a cleanup PR it's worth putting these functions in the place that makes the most sense (right now this PR is having the pipeline validation use a function defined in task validation which feels a bit strange to me). Rebasing and resolving merge conflicts is sort of a fact of life :) but if you'd prefer to wait that's OK too.

@chuangw6 maybe change the PR naming to be specifically to "remove duplicate code for validating params types and defaults" to narrow down the cleanup scope. You can create a separate PR for workspaces

chuangw6:
> @chuangw6 maybe change the PR naming to be specifically to "remove duplicate code for validating params types and defaults" to narrow down the cleanup scope. You can create a separate PR for workspaces

great point! Done!

lbernick (approved):

Issue #4470, opened 2022-01-12
FR: Task-level (and maybe Pipeline-level) resource requests and limits


lbernick:
### Feature request
Currently, users can set resource requests and limits at the Step level, which are summed to determine the resource requests of the pod that runs a TaskRun. However, some users want to be able to directly specify the resource requests for a Task, rather than per Step.

### Related issues

- #2986: a FR for Task-level resource requests (however, the Step resource request behavior described is no longer true)
- #4271: FR for Pipeline-level resource requests
- #4347: Confusion that step-level resource requests are summed

lbernick:
a bit of a wrinkle here. Let's say we want to allow people to specify the total resource requests and limits that should be used by all steps in their task:

- k8s states that containers without resource limits are considered to have higher limits than those with limits configured (https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#resources)
- If the task resource limit is applied to only one container, the pod will therefore not have an effective limit.
- If the task limit is applied to each container, the pod has a much higher limit than desired. This is especially problematic if requests are not set, because the request will then automatically be set to the same value as the limit, and the pod may have difficulty being scheduled.
- If the task limit is spread out among containers, a task where one step is more resource intensive than all the others could get oomkilled or throttled.
- resource requirements [can't be updated](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#resources), so we can't dynamically adjust them as steps run.

One way around this is to support only Task-level resource requests, not limits.

Another option is to run Steps as init containers rather than containers (instead of supporting this feature). As a result, k8s will correctly determine the effective pod resource requirements. We would have to rework our entrypoint but I imagine it would be doable. However, we would have no way to support Sidecars.

dibyom:
> Another option is to run Steps as init containers rather than containers (instead of supporting this feature). As a result, k8s will correctly determine the effective pod resource requirements. We would have to rework our entrypoint but I imagine it would be doable. However, we would have no way to support Sidecars.

We used to run Tekton in init containers before, init containers have some nice characteristics but also come with a bunch of drawbacks:  see https://github.com/tektoncd/pipeline/issues/224

joshuasimon-taulia:
imo, https://github.com/tektoncd/pipeline/pull/4176 made defining cpu/mem requests much less intuitive. assuming steps in a task will never run in parallel, task-level resource definition would definitely alleviate some of that pain.

lbernick:
[...]

PR #4877, opened 2022-05-16
[TEP-0104] Support Task-level Resource Requirements for TaskRun: Part #1 Fields Addition & Validation w/ Docs Updates 

austinzhao-go:
# Changes
/kind feature

The related TEP: 
[TEP-0104: Support Task-level resource limits](https://github.com/tektoncd/community/pull/703/files?short_path=bd66ac0#diff-bd66ac0b7cdb4930cee261f3a042ee9b7dae6c68d2514de56890037176388dab)
[TEP-0104: Scope Updates to Runtime](https://github.com/tektoncd/community/pull/722)
[TEP-0104: Rename "resources" field as "computeResources"](https://github.com/tektoncd/community/pull/726)

The related FR thread: 
[FR: Task-level (and maybe Pipeline-level) resource requests and limits](https://github.com/tektoncd/pipeline/issues/4470)

This PR will include changes as:
- Add the required "computeResources" field for TaskRun
- Add the related validations
- Update the user guide for this new feature 

A follow-up PR will include changes to update compute resources configuration by the specified task-level resource requirements:
- [ref-to-be-added]()

# TODO
- [x] Update this PR only with field addition and webhook validation
- [x] Add alpha FF to guard changes over webhook
- [x] Update related docs
- [x] Confirm approval and merge on [TEP-0104](https://github.com/tektoncd/community/pull/703/files?short_path=bd66ac0#diff-bd66ac0b7cdb4930cee261f3a042ee9b7dae6c68d2514de56890037176388dab), then start merging implementations 

After the scope update & field name update:
- [x] Update field validation  
- [x] Update the related user guidance 

# Submitter Checklist

  functionality, content, code)

# Release Notes
_As changes will be delivered in 2/3 PRs, the release notes will be added in [ref-to-be-added]()._

### DIFF
--- a/docs/compute-resources.md
+++ b/docs/compute-resources.md
@@ -42,7 +42,10 @@ request will be 6.
 Since the sidecar container has no CPU limit, this is treated as the highest CPU limit.
 Therefore, the pod will have no effective CPU limit.
 
-## Task Resource Requirements
+## Task-level Compute Resources Configuration
+
+**([alpha only](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features))**
+(This feature is under development and not functional yet. Stay tuned!)
 
 Tekton allows users to specify resource requirements of [`Steps`](./tasks.md#defining-steps),
 which run sequentially. However, the pod's effective resource requirements are still the
@@ -51,6 +54,86 @@ requirements for `Step` containers, they must be treated as if they are running 
 
 Tekton adjusts `Step` resource requirements to comply with [LimitRanges](#limitrange-support).
 [ResourceQuotas](#resourcequota-support) are not currently supported.
+
+Instead of specifying resource requirements on each `Step`, users can choose to specify resource requirements at the Task-level. If users specify a Task-level resource request, it will ensure that the kubelet reserves only that amount of resources to execute the `Task`'s `Steps`.
+If users specify a Task-level resource limit, no `Step` may use more than that amount of resources.
+
+Each of these details is explained in more depth below.
+
+Some points to note:
+
+- Task-level resource requests and limits do not apply to sidecars which can be configured separately.
+- Users may not configure the Task-level and Step-level resource requirements (requests/limits)  simultaneously.
+
+### Configure Task-level Compute Resources
+
+Task-level resource requirements can be configured in `TaskRun.ComputeResources`, or `PipelineRun.TaskRunSpecs.ComputeResources`.
+
+e.g.
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: foo 
+spec:
+  computeResources:
+    requests:
+      cpu: 1 
+    limits:
+      cpu: 2
+```
+
+The following TaskRun will be rejected, because it configures both step-level and task-level compute resource requirements:
+
+```yaml
+kind: TaskRun
+spec:
+  stepOverrides:
+    - name: foo
+      resources:
+        requests:
+          cpu: 1
+  computeResources:
+    requests:
+      cpu: 2 
+```
+
+```yaml
+kind: PipelineRun
+spec:
+  taskRunSpecs:
+    - pipelineTaskName: foo
+      stepOverrides:
+        - name: foo 
+          resources:
+            requests:
+              cpu: 1 
+      computeResources:
+        requests:
+          cpu: 2
+```
+
+### Configure Resource Requirements with Sidecar
+
+Users can specify compute resources separately for a sidecar while configuring task-level resource requirements on TaskRun.
+
+e.g.
+
+```yaml
+kind: TaskRun
+spec:
+  sidecarOverrides:
+    - name: sidecar
+      resources:
+        requests:
+          cpu: 750m
+        limits:
+          cpu: 1
+  computeResources:
+    requests:
+      cpu: 2
+```
 
 ## LimitRange Support
 
--- a/docs/taskruns.md
+++ b/docs/taskruns.md
@@ -18,6 +18,7 @@ weight: 300
     - [Extra Parameters](#extra-parameters)
   - [Specifying `Resources`](#specifying-resources)
   - [Specifying `Resource` limits](#specifying-resource-limits)
+  - [Specifying Task-level `ComputeResources`](#specifying-task-level-computeresources)
   - [Specifying a `Pod` template](#specifying-a-pod-template)
   - [Specifying `Workspaces`](#specifying-workspaces)
   - [Specifying `Sidecars`](#specifying-sidecars)
@@ -323,6 +324,40 @@ spec:
 Each Step in a Task can specify its resource requirements. See
 [Defining `Steps`](tasks.md#defining-steps). Resource requirements defined in Steps and Sidecars
 may be overridden by a TaskRun's StepOverrides and SidecarOverrides.
+
+### Specifying Task-level `ComputeResources`
+
+**([alpha only](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features))**
+(This feature is under development and not functional yet. Stay tuned!)
+
+Task-level compute resources can be configured in `TaskRun.ComputeResources`, or `PipelineRun.TaskRunSpecs.ComputeResources`.
+
+e.g.
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: task
+spec:
+  steps:
+    - name: foo
+---
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: taskrun 
+spec:
+  taskRef:
+    name: task
+  computeResources:
+    requests:
+      cpu: 1 
+    limits:
+      cpu: 2
+```
+
+Further details and examples could be found in [Compute Resources in Tekton](https://github.com/tektoncd/pipeline/blob/main/docs/compute-resources.md).
 
 ### Specifying a `Pod` template
 
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/tektoncd/pipeline/pkg/apis/config"
@@ -540,6 +541,9 @@ type PipelineTaskRunSpec struct {
 
 	// +optional
 	Metadata *PipelineTaskMetadata `json:"metadata,omitempty"`
+
+	// Compute resources to use for this TaskRun
+	ComputeResources *corev1.ResourceRequirements `json:"computeResources,omitempty"`
 }
 
 // GetTaskRunSpec returns the task specific spec for a given
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
@@ -195,6 +195,9 @@ func validateTaskRunSpec(ctx context.Context, trs PipelineTaskRunSpec) (errs *ap
 		if trs.SidecarOverrides != nil {
 			errs = errs.Also(validateSidecarOverrides(trs.SidecarOverrides).ViaField("sidecarOverrides"))
 		}
+		if trs.ComputeResources != nil {
+			errs = errs.Also(validateTaskRunComputeResources(trs.ComputeResources, trs.StepOverrides))
+		}
 	} else {
 		if trs.StepOverrides != nil {
 			errs = errs.Also(apis.ErrDisallowedFields("stepOverrides"))
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
@@ -610,7 +610,32 @@ func TestPipelineRunSpec_Invalidate(t *testing.T) {
 		},
 		wantErr:     apis.ErrMissingField("taskRunSpecs[0].sidecarOverrides[0].name"),
 		withContext: enableAlphaAPIFields,
+	}, {
+		name: "invalid both step-level (stepOverrides.resources) and task-level (taskRunSpecs.resources) resource requirements configured",
+		spec: v1beta1.PipelineRunSpec{
+			PipelineRef: &v1beta1.PipelineRef{Name: "pipeline"},
+			TaskRunSpecs: []v1beta1.PipelineTaskRunSpec{
+				{
+					PipelineTaskName: "pipelineTask",
+					StepOverrides: []v1beta1.TaskRunStepOverride{{
+						Name: "stepOverride",
+						Resources: corev1.ResourceRequirements{
+							Requests: corev1.ResourceList{corev1.ResourceMemory: corev1resources.MustParse("1Gi")},
+						}},
+					},
+					ComputeResources: &corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{corev1.ResourceMemory: corev1resources.MustParse("2Gi")},
+					},
+				},
+			},
+		},
+		wantErr: apis.ErrMultipleOneOf(
+			"taskRunSpecs[0].stepOverrides.resources",
+			"taskRunSpecs[0].computeResources",
+		),
+		withContext: enableAlphaAPIFields,
 	}}
+
 	for _, ps := range tests {
 		t.Run(ps.name, func(t *testing.T) {
 			ctx := context.Background()
@@ -627,8 +652,9 @@ func TestPipelineRunSpec_Invalidate(t *testing.T) {
 
 func TestPipelineRunSpec_Validate(t *testing.T) {
 	tests := []struct {
-		name string
-		spec v1beta1.PipelineRunSpec
+		name        string
+		spec        v1beta1.PipelineRunSpec
+		withContext func(context.Context) context.Context
 	}{{
 		name: "PipelineRun without pipelineRef",
 		spec: v1beta1.PipelineRunSpec{
@@ -641,10 +667,53 @@ func TestPipelineRunSpec_Validate(t *testing.T) {
 				}},
 			},
 		},
+	}, {
+		name: "valid task-level (taskRunSpecs.resources) resource requirements configured",
+		spec: v1beta1.PipelineRunSpec{
+			PipelineRef: &v1beta1.PipelineRef{Name: "pipeline"},
+			TaskRunSpecs: []v1beta1.PipelineTaskRunSpec{{
+				PipelineTaskName: "pipelineTask",
+				StepOverrides: []v1beta1.TaskRunStepOverride{{
+					Name: "stepOverride",
+				}},
+				ComputeResources: &corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{corev1.ResourceMemory: corev1resources.MustParse("2Gi")},
+				},
+			}},
+		},
+		withContext: enableAlphaAPIFields,
+	}, {
+		name: "valid sidecar and task-level (taskRunSpecs.resources) resource requirements configured",
+		spec: v1beta1.PipelineRunSpec{
+			PipelineRef: &v1beta1.PipelineRef{Name: "pipeline"},
+			TaskRunSpecs: []v1beta1.PipelineTaskRunSpec{{
+				PipelineTaskName: "pipelineTask",
+				StepOverrides: []v1beta1.TaskRunStepOverride{{
+					Name: "stepOverride",
+				}},
+				ComputeResources: &corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{corev1.ResourceMemory: corev1resources.MustParse("2Gi")},
+				},
+				SidecarOverrides: []v1beta1.TaskRunSidecarOverride{{
+					Name: "sidecar",
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							corev1.ResourceMemory: corev1resources.MustParse("4Gi"),
+						},
+					},
+				}},
+			}},
+		},
+		withContext: enableAlphaAPIFields,
 	}}
+
 	for _, ps := range tests {
 		t.Run(ps.name, func(t *testing.T) {
-			if err := ps.spec.Validate(context.Background()); err != nil {
+			ctx := context.Background()
+			if ps.withContext != nil {
+				ctx = ps.withContext(ctx)
+			}
+			if err := ps.spec.Validate(ctx); err != nil {
 				t.Error(err)
 			}
 		})
--- a/pkg/apis/pipeline/v1beta1/taskrun_types.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_types.go
@@ -77,6 +77,8 @@ type TaskRunSpec struct {
 	// +optional
 	// +listType=atomic
 	SidecarOverrides []TaskRunSidecarOverride `json:"sidecarOverrides,omitempty"`
+	// Compute resources to use for this TaskRun
+	ComputeResources *corev1.ResourceRequirements `json:"computeResources,omitempty"`
 }
 
 // TaskRunSpecStatus defines the taskrun spec status the user can provide
--- a/pkg/apis/pipeline/v1beta1/taskrun_validation.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_validation.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/validate"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/apis"
 )
@@ -66,6 +67,7 @@ func (ts *TaskRunSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	if ts.StepOverrides != nil {
 		errs = errs.Also(ValidateEnabledAPIFields(ctx, "stepOverrides", config.AlphaAPIFields).ViaField("stepOverrides"))
 		errs = errs.Also(validateStepOverrides(ts.StepOverrides).ViaField("stepOverrides"))
+		errs = errs.Also(validateTaskRunComputeResources(ts.ComputeResources, ts.StepOverrides))
 	}
 	if ts.SidecarOverrides != nil {
 		errs = errs.Also(ValidateEnabledAPIFields(ctx, "sidecarOverrides", config.AlphaAPIFields).ViaField("sidecarOverrides"))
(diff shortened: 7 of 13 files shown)


pkg/apis/pipeline/v1beta1/resource_types.go
```diff
@@ -70,6 +70,16 @@ type TaskResources struct {
 	// DeclaredPipelineResources to the input PipelineResources required by the Task.
 	// +listType=atomic
 	Outputs []TaskResource `json:"outputs,omitempty"`
+	// Limits describes the maximum amount of compute resources allowed.
```
lbernick, line 73: We'll want to document this feature really well, because it's not intuitive how these resource requirements will be applied. I'd suggest adding a bit of detail here, and a lot of detail in [docs/compute-resources.md](https://github.com/tektoncd/pipeline/blob/main/docs/compute-resources.md).

pkg/apis/pipeline/v1beta1/resource_types.go
```diff
@@ -70,6 +70,16 @@ type TaskResources struct {
...
+	// More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+	// +optional
+	Limits v1.ResourceList `json:"limits,omitempty"`
```
lbernick, line 76: Also, as detailed in the TEP, these fields should be added to TaskRun and PipelineRun.TaskRunSpecs too.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -333,10 +333,27 @@ func (c *Reconciler) prepare(ctx context.Context, tr *v1beta1.TaskRun) (*v1beta1
 		return nil, nil, controller.NewPermanentError(err)
 	}
 
-	if err := validateTaskSpecRequestResources(ctx, taskSpec); err != nil {
-		logger.Errorf("TaskRun %s taskSpec request resources are invalid: %v", tr.Name, err)
+	if isResourcesRequirementsInStepLevel(taskSpec) && isResourcesRequirementsInTaskLevel(taskSpec) {
+		err := fmt.Errorf("TaskRun %s can't be configured with both step-level and task-level resources requirements", tr.Name)
 		tr.Status.MarkResourceFailed(podconvert.ReasonFailedValidation, err)
 		return nil, nil, controller.NewPermanentError(err)
+	} else if isResourcesRequirementsInStepLevel(taskSpec) {
```
lbernick, line 340: We should apply validation in the webhook as well as the reconciler (the reason is to return errors as early as possible -- for example an invalid task should be rejected on creation, not when it's run in a taskrun). I'd suggest:

- in [task_validation.go](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/task_validation.go), ensure that the task does not have both step-level and task-level resource requirements
- in [taskrun_validation.go](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/taskrun_validation.go), make the same check as for tasks
- Here in the reconciler code (probably as a helper function in [validate_resources.go](https://github.com/tektoncd/pipeline/blob/c938ee5667fb64ddcf6e6c26e7732b68dde82104/pkg/reconciler/taskrun/validate_resources.go)), validate that the taskrun resource requirements are compatible with the task resource requirements.

lbernick (commented):
Thanks Austin! I would suggest just adding the fields, docs, and webhook validation in the initial PR, and updating the reconciler in a subsequent PR.

/hold

until https://github.com/tektoncd/community/pull/703 is accepted

austinzhao-go:
> 

agree. thinking of a bit the complexity, was planning to complete the logic `Task-level` first, then for "runtime overwriting" from `TaskRunSpec`. but let me do this perhaps (for an easier review):
- field addition and validation on current PR
- `Task-level` update on 2nd PR 
- "runtime overwriting" on 3rd PR

pkg/apis/pipeline/v1beta1/resource_types.go
austinzhao-go: yap. planned to deliver as a follow-up PR (focusing on "runtime overwriting" part)

pkg/reconciler/taskrun/taskrun.go
austinzhao-go: thanks Lee. let me only keep "validation" logic in the reconciler (removing the resources update part) and add validation in webhook.

lbernick:
ok sounds good! Just one more note, new fields should be guarded behind the alpha feature flag

lbernick:
> got this. So the feature flag checks will be mainly/only needed in reconciler code right? (not changes wekbook code before, let me have another check. perhaps also needed there)

The feature flag checks should be included in the webhook validation, and optionally the reconciler validation, for the same reason as before (fail as fast as possible).

pkg/reconciler/taskrun/taskrun.go
austinzhao-go: closed by moving changes on reconciler into the following PR

docs/compute-resources.md
```diff
@@ -52,6 +52,251 @@ requirements for `Step`containers, they must be treated as if they are running i
...
+    - name: foo1
+    - name: foo2
+  resources:
+    requests:
+      cpu: 1 
+    limits:
+      cpu: 2
+```
+
+### Applying Resources Requirements
```
lbernick, line 77: Some of this commentary could be reworded to focus less on the implementation and more on the user experience.
For example, where you talk about limits being enforced by the container runtime, or validation being performed by the webhook.

docs/compute-resources.md
```diff
@@ -52,6 +52,251 @@ requirements for `Step`containers, they must be treated as if they are running i
...
+        - name: foo 
+          resources:
+            requests:
+              cpu: 1 
+      resources:
+        requests:
+          cpu: 2 
+```
+
+### Interaction with Runtime Overwriting
```
lbernick, line 153: Suggested wording:

Section Name: Specifying compute resources in TaskRuns

Description: You can specify compute resources at the Step level or at the Task level in TaskRuns, but not both.
To specify compute resources at the Step level, use TaskRuns.StepOverrides. <insert example>
To specify compute resources at the Task level, use TaskRuns.Resources. <insert example>

docs/compute-resources.md
```diff
@@ -52,6 +52,251 @@ requirements for `Step`containers, they must be treated as if they are running i
...
+    - name: foo
+  stepTemplate:
+    requests:
+      cpu: 1
+  resources:
+    requests:
+      cpu: 2
+```
+
+Case #3: `TaskRun.spec.stepOverrides.resources` and `TaskRun.spec.taskSpec.resources`
```
lbernick, line 120: This case and the one below should probably be part of the section on TaskRuns.

docs/compute-resources.md
```diff
@@ -52,6 +52,251 @@ requirements for `Step`containers, they must be treated as if they are running i
...
+    resources:
+      requests:
+        cpu: 2
+```
+
+The final task cpu resources request will be 2 by the overwriting.
+
+### Interaction with LimitRanges
+
+Min or max resources requirements can be configured by `LimitRange`. For a min request on certain containers, the task-level request can subtract this amount and assign the rest to one `Step`(container) as stated before. Furthermore, if a container has a value more than the max limit after being set with task-level requests, the request amount will be spread out to each container (the number will be rounded to ignore decimals).
```
lbernick, line 182: Some of this also gets into implementation details. Since there are multiple possible ways to configure container resource requirements to meet task level resource requirements, I wouldn't describe which way we're choosing in the docs.

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -600,6 +601,11 @@ type PipelineTaskRunSpec struct {
 	StepOverrides []TaskRunStepOverride `json:"stepOverrides,omitempty"`
 	// +listType=atomic
 	SidecarOverrides []TaskRunSidecarOverride `json:"sidecarOverrides,omitempty"`
+	// Compute Resources required by the Task/TaskRun (Pod).
```
lbernick, line 604: I'd remove the comment about pods here

pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
```diff
@@ -205,3 +208,15 @@ func validateTaskRunSpec(ctx context.Context, trs PipelineTaskRunSpec) (errs *ap
 	}
 	return errs
 }
+
+// validateResourcesRequirements() validates if only step-level or task-level resources requirements are configured
+func validateResourcesRequirements(trs PipelineTaskRunSpec) (errs *apis.FieldError) {
+	for _, taskRunStepOverride := range trs.StepOverrides {
```
lbernick, line 214: I think the TEP states that we'll consider different resource types separately; so for example you could specify step level for CPU and task-level for memory. (Open to revisiting this decision.)

Also, why use `.Size > 0` instead of a nil check?

pkg/apis/pipeline/v1beta1/resource_types.go
```diff
@@ -61,6 +61,7 @@ var AllResourceTypes = resource.AllResourceTypes
 
 // TaskResources allows a Pipeline to declare how its DeclaredPipelineResources
 // should be provided to a Task as its inputs and outputs.
+// Also task-level(Pod) container computing resources can be configured by limits and requests.
```
lbernick, line 64: Suggested wording:

`TaskResources is overloaded to represent both PipelineResources and compute resources. PipelineResources are deprecated and will be removed, so this field will only represent compute resources in the future.`

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -138,6 +139,19 @@ func validateStepOverrides(overrides []TaskRunStepOverride) (errs *apis.FieldErr
 	return errs
 }
 
+// validateStepOverridesResourcesRequirements() validates if both step-level and task-level resources requirements are configured
+func validateStepOverridesResourcesRequirements(taskSpec *TaskSpec, overrides []TaskRunStepOverride) (errs *apis.FieldError) {
```
lbernick, line 143: I think you also need to add TaskRun.resources to the api

docs/compute-resources.md
```diff
@@ -52,6 +52,251 @@ requirements for `Step`containers, they must be treated as if they are running i
 Tekton adjusts `Step` resource requirements to comply with [LimitRanges](#limitrange-support).
 [ResourceQuotas](#resourcequota-support) are not currently supported.
 
+Kubernetes interprets `Pod` resources requirements by taking the value summed up by each container's resources requirements. It leads Tekton to hoard resources computed by all `Steps`(containers), which can only be used on each sequentially executed `Step`. This observation forms the interest to support a `Task`/`Pod` level resources requirements, which naturally matches Kubernetes computing logic and can help users save significantly on requested resources.
```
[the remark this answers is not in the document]
austinzhao-go: ya. this works with me.

docs/compute-resources.md
austinzhao-go: agree this idea.

so even one of the mentioned fields is on `PipelineRun`, still be good for us to mention "in `TaskRun`" right? as the field is under `PipelineRun.TaskRunSpecs`

docs/compute-resources.md
austinzhao-go: (as a previous comment) let me refine around more under a user doc view. 

thanks this point!

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
austinzhao-go: copy here. will keep as 
`// Compute Resources required by the Task/TaskRun`

pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
austinzhao-go: for ` .Size > 0`, an empty/stub resources field was populated out to fail this check when I did the local test on cluster. so using this.. a bit dummy way.  😵

pkg/apis/pipeline/v1beta1/taskrun_validation.go
austinzhao-go: you mean `TaskRun.spec.resources` right? 
https://github.com/austinzhao-go/pipeline/blob/tep-0104-per-task-resources-requirements/pkg/apis/pipeline/v1beta1/taskrun_types.go#L44

rechecked our "API changes" section, think I understood wrongly as the referred `TaskSpec` field.

lbernick: yes exactly

pkg/apis/pipeline/v1beta1/resource_types.go
austinzhao-go: closed - by updating comments + docs

austinzhao-go:
thanks @lbernick. pushed changes for this round of reviews. 
will follow 
- TEP PR got merged asap (reached out community slack channel, and will check in the following API WG session)
- further refining feedback on the user guide docs

docs/compute-resources.md
austinzhao-go: closed - updated as `Specifying compute resources in TaskRun/PipelineRun`

pkg/apis/pipeline/v1beta1/taskrun_validation.go
austinzhao-go: closed - added the field and related validations

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
austinzhao-go: closed - removed words on `Pod`

docs/compute-resources.md
austinzhao-go: closed - cleaned

docs/compute-resources.md
```diff
@@ -52,6 +52,263 @@ requirements for `Step`containers, they must be treated as if they are running i
 Tekton adjusts `Step` resource requirements to comply with [LimitRanges](#limitrange-support).
 [ResourceQuotas](#resourcequota-support) are not currently supported.
 
+Instead of specifying resource requirements on each `Step`, users can choose to specify resource requirements at the Task-level. If users specify a Task-level resource request, it will ensure that the kubelet reserves only that amount of resources to execute the `Task`'s `Steps`.
```
lbernick, line 55: could you please note this feature is in alpha (you need to enable alpha api fields to get it), and that it is not yet functional? Lastly, docs should be added to taskruns.md as well.

docs/compute-resources.md
```diff
@@ -52,6 +52,263 @@ requirements for `Step`containers, they must be treated as if they are running i
...
+If users specify a Task-level resource limit, no `Step` may use more than that amount of resources.
+
+Each of these details is explained in more depth below.
+
+Some points to note:
+
+- Task-level resource requests and limits do not apply to sidecars which can be configured separately.
+- Users may not configure the Task-level and Step-level resource requirements (requests/limits)  simultaneously.
+
+### Configuring Task-level Resource Requirements
```
lbernick, line 65: Since this PR just introduces the new field to "TaskRun", let's stick with only documenting that for now. Also, the examples should use "computeResources" rather than "resources".

docs/compute-resources.md
```diff
@@ -52,6 +52,263 @@ requirements for `Step`containers, they must be treated as if they are running i
...
+            requests:
+              cpu: 1 
+      resources:
+        requests:
+          cpu: 2 
+```
+
+### Specifying Resource Requirements with LimitRange
+
+User can use `LimitRange` to configure min or max compute resources for a `Task` or `Step`.
```
lbernick, line 190: Could you move this example to the section on LimitRanges? If we're going to include an example like the one you have here, it would be good to show both how it works for step-level and how it works for task-level.

pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -77,6 +77,8 @@ type TaskRunSpec struct {
 	// +optional
 	// +listType=atomic
 	SidecarOverrides []TaskRunSidecarOverride `json:"sidecarOverrides,omitempty"`
+	// Compute resource configurations applied to the task-level
```
lbernick, line 80: suggested docstring: Compute resources to use for this TaskRun

Also, I think this needs to be a pointer to ResourceRequirements, for the same reason as #4913.

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -138,6 +140,24 @@ func validateStepOverrides(overrides []TaskRunStepOverride) (errs *apis.FieldErr
 	return errs
 }
 
+// validateStepOverridesResourceRequirements validates if both step-level and task-level resource requirements are configured
+func validateStepOverridesResourceRequirements(computeResources corev1.ResourceRequirements, overrides []TaskRunStepOverride) (errs *apis.FieldError) {
```
lbernick, line 144: apis.ErrMultipleOneOf is probably more appropriate here

Also, I'd suggest the following function name + docstring:

// validateTaskRunComputeResources ensures that compute resources are not configured at both the step level and the task level

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -138,6 +140,24 @@ func validateStepOverrides(overrides []TaskRunStepOverride) (errs *apis.FieldErr
...
+				Message: "TaskRun can't be configured with both step-level (stepOverrides.resources) and task-level (spec.computeResources) resource requirements",
+				Paths:   []string{"resources"},
+			}
+		}
+	}
+	return nil
+}
+
+// isResourceRequirementsInTaskRun checks if resource requirements are configured in TaskRun under task-level
+func isResourceRequirementsInTaskRun(computeResources corev1.ResourceRequirements) bool {
```
lbernick, line 157: this function is not needed; just check if taskrun.computeResources is a nil pointer

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
```diff
@@ -515,7 +515,31 @@ func TestTaskRunSpec_Invalidate(t *testing.T) {
 		},
 		wantErr: apis.ErrMissingField("sidecarOverrides[0].name"),
 		wc:      enableAlphaAPIFields,
+	}, {
+		name: "invalid both step-level (stepOverrides.resources) and task-level (spec.computeResources) resource requirements",
```
lbernick, line 519: could you please add a test case that ensures that you can configure both sidecarOverrides and computeResources?

lbernick:
/hold cancel

since https://github.com/tektoncd/community/pull/703 is now implementable

docs/compute-resources.md
austinzhao-go: got this. let me do an update.

docs/compute-resources.md
austinzhao-go: yap. (didn't update the doc yet. let me add the field back for PipelineRun and update the doc with our current TEP requirements)

docs/compute-resources.md
austinzhao-go: good catch. thanks this!

pkg/apis/pipeline/v1beta1/taskrun_types.go
austinzhao-go: looked into the details and changes. thanks this ref!

pkg/apis/pipeline/v1beta1/taskrun_validation.go
austinzhao-go: right. this will apply nicely after updating as a pointer.

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
austinzhao-go: sure! thanks

docs/taskruns.md
```diff
@@ -324,6 +324,39 @@ Each Step in a Task can specify its resource requirements. See
 [Defining `Steps`](tasks.md#defining-steps). Resource requirements defined in Steps and Sidecars
 may be overridden by a TaskRun's StepOverrides and SidecarOverrides.
 
+### Specifying Task-level `ComputeResources`
+
+**([alpha only](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features))**
+
+Task-level compute resources can be configured in `TaskRun.ComputeResources`, or `PipelineRun.TaskRunSpecs.ComputeResources`.
```
lbernick, line 331: I forgot to mention, please add this info to pipelineruns.md as well.

docs/compute-resources.md
lbernick: Thanks for marking as alpha! Please also mark that this isn't functional yet.

docs/compute-resources.md
```diff
@@ -52,6 +54,107 @@ requirements for `Step`containers, they must be treated as if they are running i
...
+
+```yaml
+kind: TaskRun
+spec:
+  computeResources:
+    requests:
+      cpu: 2
+```
+
+Rejected when configuring `TaskRun.StepOverrides.Resources` and `TaskRun.ComputeResources`:
```
lbernick, line 105: For here and below, I would just say "the following TaskRun will be rejected, because it configures both step-level and task-level compute resource requirements".

docs/compute-resources.md
```diff
@@ -52,6 +54,107 @@ requirements for `Step`containers, they must be treated as if they are running i
...
+
+Using `TaskRun.Resources`:
+
+```yaml
+kind: TaskRun
+spec:
+  computeResources:
+    requests:
+      cpu: 2
+```
```
lbernick, line 103: this section is a bit redundant with the info above it.

docs/compute-resources.md
```diff
@@ -217,6 +320,113 @@ spec:
...
+    requests:
+      cpu: 1
+```
+
+| Step name | CPU request |
+| --------- | ----------- |
+| step1     | 800m        |
+| step2     | 200m        |
+| step3     | N/A         |
+
```
lbernick, line 363: I'd prefer not to include examples that state what the resource requirements on containers will end up being, since this is more of an implementation detail.

docs/compute-resources.md
```diff
@@ -217,6 +320,113 @@ spec:
 Here, the minimum of the "max" values is the output "max" value, and likewise for "default" and "defaultRequest".
 The maximum of the "min" values is the output "min" value.
 
+### Configure Task-level Compute Resources on TaskRun with LimitRange
```
lbernick, line 323: after thinking about this a bit more, I think it would be better to just leave out this section for now. It works basically the same way as what's already described in the docs.

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
```diff
@@ -641,10 +667,30 @@ func TestPipelineRunSpec_Validate(t *testing.T) {
 				}},
 			},
 		},
+	}, {
+		name: "valid task-level (taskRunSpecs.resources) resource requirements configured",
+		spec: v1beta1.PipelineRunSpec{
+			PipelineRef: &v1beta1.PipelineRef{Name: "pipeline"},
+			TaskRunSpecs: []v1beta1.PipelineTaskRunSpec{{
+				PipelineTaskName: "pipelineTask",
+				StepOverrides: []v1beta1.TaskRunStepOverride{{
```
lbernick, line 676: could you please add a test case with sidecar as well?

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -138,6 +140,19 @@ func validateStepOverrides(overrides []TaskRunStepOverride) (errs *apis.FieldErr
 	return errs
 }
 
+// validateTaskRunComputeResources ensures that compute resources are not configured at both the step level and the task level
+func validateTaskRunComputeResources(computeResources *corev1.ResourceRequirements, overrides []TaskRunStepOverride) (errs *apis.FieldError) {
+	for _, override := range overrides {
+		if override.Resources.Size() > 0 && computeResources != nil {
```
lbernick, line 146: can you use `if len(overrides.Resources) !=0 ...`?

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -138,6 +140,19 @@ func validateStepOverrides(overrides []TaskRunStepOverride) (errs *apis.FieldErr
...
+			return apis.ErrMultipleOneOf(
+				"stepOverrides.resources",
+				"taskRunSpecs.computeResources",
```
lbernick, line 149: the field name here should just be `computeResources`

pkg/apis/pipeline/v1beta1/taskrun_validation.go
lbernick, line 146: can you use `if len(overrides.Resources) !=0 ...`?

docs/compute-resources.md
austinzhao-go: works!

docs/compute-resources.md
austinzhao-go: was thinking if this could help users understand how their configure will take an effect, so better know how to configure their requirements. how do you think about it?

docs/compute-resources.md
austinzhao-go: yap. np for this.

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
austinzhao-go: right. missed it, let me add one case for it.

pkg/apis/pipeline/v1beta1/taskrun_validation.go
austinzhao-go: hmm seems a bit better. != is a bit more specific condition for >

pkg/apis/pipeline/v1beta1/taskrun_validation.go
austinzhao-go: so here mean we leave out the "parent" field, `taskRunSpecs`, for it? 
thinking it perhaps could help users locate the "error" field

pkg/apis/pipeline/v1beta1/taskrun_validation.go
austinzhao-go: ya. +1 as the last comment

austinzhao-go:
cc @ywluogg

docs/compute-resources.md
lbernick: hm, on second thought it might be worth including an example-- my main concern is that there are multiple valid ways to implement task-level resource requirements, so I don't necessarily want to tie ourselves to one way in the documentation. but I'll leave this up to you.

lbernick (approved):

pkg/apis/pipeline/v1beta1/taskrun_validation.go
austinzhao-go: right! let me get rid of it. thanks Lee.

docs/compute-resources.md
austinzhao-go: this is a good point as the updating logic is not completed now and could be adjusted when going into the details. 

I backed these changes. let me remove them now, and see if I can offer one meaningful example after impled the resource config/update logic.

Issue #4723, opened 2022-03-30
[TEP 0075, 0076]: Structured object / arrays in Results and Params


ywluogg:
### Feature request

The issue will track the implementation for [TEP-0075 Propose object (dicitonary) param and result types](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md) and [TEP-0076 Propose array results and indexing](https://github.com/tektoncd/community/commit/e0e7d73e08113ac82e5fe8bc6467d31db51fbec7)

The two docs are in the process of review to get to *Implementable* states (https://github.com/tektoncd/community/pull/661)

### Use case

Will have a more detailed use case summary, but please refer to the use cases mentioned in [TEP-0075 Propose object (dicitonary) param and result types](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md) and [TEP-0076 Propose array results and indexing](https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md)

To summarize, we want to change the params to be supporting the following:

Params will affect the following:
- Pipeline.params
- Pipeline.TaskRef.params
- PipelineRun.params
- PipelineRun.PipelineSpec.params
- PipelineRun.PipelineSpec.tasks.params
- PipelineRun.PipelineSpec.taskSpec.params
- Task.params
- TaskRun.spec.params
- TaskRun.TaskSpec,params

We also want to change the results to be supporting objects and arrays in Pipeline Results and Task Results.
Results will affect the following:
- PipelineRun.Spec.PipelineSpec.Results
- PipelineRun.Spec.PipelineSpec.PipelineSpec.TaskSpec.Results
- Pipeline.Spec.Results
- TaskRun.Spec.TaskSpec.Results
- Task.Spec.Results

Below an example to specify a object param in Tasks and using its key1 and image1 value, and also specify a object in Task results and use script to write the results:
``` yaml
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: task-with-parameters
spec:
  params:
  - name: objectVal1
     description: "description1"
     type: object
     properties:
        key1:
          type: string
        image1:
          type: string
  - name: objectVal2
     description: "description2"
     type: object
     properties:
        key1:
          type: string
        image1:
          type: string
  steps:
  - name: step1
     image: $(objectVal1.image1)
     script: |
     echo $(objectVal1.key1)
  - name: step2
     image: $(objectVal2.image1)
     script: |
[...]

PR #4268, opened 2021-09-30
Support for single-quote bracket notation for params

skaegi:
This change adds single-quote bracket notation to the work done in #4215. This is consistent with how referencing is done elsewhere in K8s in the downwards api.

The original patch used the name subscript notation to describe this however the standard name for this approach is bracket notation and updates the doc accordingly.

This is based on TEP: tektoncd/community#503 -- which also needs to be aligned with the two bracket notation approaches in the implementation. See https://github.com/tektoncd/community/pull/527

# Changes

This change adds single-quote bracket param matching to the params and param results analagous with the double-quote bracket param matching. It also adds similar test cases and updates the doc with the single-quote syntax. 

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

```release-note
Params and result can now be referenced with bracket notation using both single and double quotes in addition to dot notation. For example, the following are equivalent:  $(param.myparam), $(param['myparam']), and $(param["myparam"]). Bracket notation has the additional benefit of allowing users to work with parameter names containing conflicting characters like "." (e.g. $(param['my.param']) or $(param["my.param"]).
```

PR #4878, opened 2022-05-16
[TEP-0075] Validate task result variable of object type

chuangw6:
# Changes
Prior to this commit, when providing param value with task result
variable, it only allowed using the task result as whole in the format
of `tasks.<taskName>.results.<resultName>` since result can
be only of type string previously.

As we are adding support for object type result, we need to support
using the result variable of object type in the format of
`tasks.<taskName>.results.<objectResultName>.<individualAttribute>`.

In this commit, we consider the object case in the validation webhook.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```
add validation and parsing for object result references
For example, $(tasks.mytask.results.myObjectResult[*]) and `$(tasks.mytask.results.myObjectResult.myKey)` 
are valid references to whole object result and a specific key of an object result.
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
@@ -1102,6 +1102,10 @@ func TestValidatePipelineResults_Success(t *testing.T) {
 		Name:        "my-pipeline-result",
 		Description: "this is my pipeline result",
 		Value:       "$(tasks.a-task.results.output)",
+	}, {
+		Name:        "my-pipeline-object-result",
+		Description: "this is my pipeline result",
+		Value:       "$(tasks.a-task.results.gitrepo.commit)",
 	}}
 	if err := validatePipelineResults(results); err != nil {
 		t.Errorf("Pipeline.validatePipelineResults() returned error for valid pipeline: %s: %v", desc, err)
@@ -1118,10 +1122,10 @@ func TestValidatePipelineResults_Failure(t *testing.T) {
 		results: []PipelineResult{{
 			Name:        "my-pipeline-result",
 			Description: "this is my pipeline result",
-			Value:       "$(tasks.a-task.results.output.output)",
+			Value:       "$(tasks.a-task.results.output.key1.extra)",
 		}},
 		expectedError: apis.FieldError{
-			Message: `invalid value: expected all of the expressions [tasks.a-task.results.output.output] to be result expressions but only [] were`,
+			Message: `invalid value: expected all of the expressions [tasks.a-task.results.output.key1.extra] to be result expressions but only [] were`,
 			Paths:   []string{"results[0].value"},
 		},
 	}, {
--- a/pkg/apis/pipeline/v1beta1/resultref.go
+++ b/pkg/apis/pipeline/v1beta1/resultref.go
@@ -28,10 +28,15 @@ type ResultRef struct {
 	PipelineTask string `json:"pipelineTask"`
 	Result       string `json:"result"`
 	ResultsIndex int    `json:"resultsIndex"`
+	Property     string `json:"property"`
 }
 
 const (
 	resultExpressionFormat = "tasks.<taskName>.results.<resultName>"
+	// Result expressions of the form <resultName>.<attribute> will be treated as object results.
+	// If a string result name contains a dot, brackets should be used to differentiate it from an object result.
+	// https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#collisions-with-builtin-variable-replacement
+	objectResultExpressionFormat = "tasks.<taskName>.results.<objectResultName>.<individualAttribute>"
 	// ResultTaskPart Constant used to define the "tasks" part of a pipeline result reference
 	ResultTaskPart = "tasks"
 	// ResultResultPart Constant used to define the "results" part of a pipeline result reference
@@ -55,7 +60,7 @@ var arrayIndexingRegex = regexp.MustCompile(arrayIndexing)
 func NewResultRefs(expressions []string) []*ResultRef {
 	var resultRefs []*ResultRef
 	for _, expression := range expressions {
-		pipelineTask, result, index, err := parseExpression(expression)
+		pipelineTask, result, index, property, err := parseExpression(expression)
 		// If the expression isn't a result but is some other expression,
 		// parseExpression will return an error, in which case we just skip that expression,
 		// since although it's not a result ref, it might be some other kind of reference
@@ -64,6 +69,7 @@ func NewResultRefs(expressions []string) []*ResultRef {
 				PipelineTask: pipelineTask,
 				Result:       result,
 				ResultsIndex: index,
+				Property:     property,
 			})
 		}
 	}
@@ -101,6 +107,11 @@ func GetVarSubstitutionExpressionsForParam(param Param) ([]string, bool) {
 	case ParamTypeString:
 		// string type
 		allExpressions = append(allExpressions, validateString(param.Value.StringVal)...)
+	case ParamTypeObject:
+		// object type
+		for _, value := range param.Value.ObjectVal {
+			allExpressions = append(allExpressions, validateString(value)...)
+		}
 	default:
 		return nil, false
 	}
@@ -129,19 +140,41 @@ func stripVarSubExpression(expression string) string {
 	return strings.TrimSuffix(strings.TrimPrefix(expression, "$("), ")")
 }
 
-func parseExpression(substitutionExpression string) (string, string, int, error) {
+// parseExpression parses "task name", "result name", "array index" (iff it's an array result) and "object key name" (iff it's an object result)
+// Valid Example 1:
+// - Input: tasks.myTask.results.aStringResult
+// - Output: "myTask", "aStringResult", -1, "", nil
+// Valid Example 2:
+// - Input: tasks.myTask.results.anObjectResult.key1
+// - Output: "myTask", "anObjectResult", 0, "key1", nil
+// Valid Example 3:
+// - Input: tasks.myTask.results.anArrayResult[1]
+// - Output: "myTask", "anArrayResult", 1, "", nil
+// Invalid Example 1:
+// - Input: tasks.myTask.results.resultName.foo.bar
+// - Output: "", "", 0, "", error
+// TODO: may use regex for each type to handle possible reference formats
+func parseExpression(substitutionExpression string) (string, string, int, string, error) {
 	subExpressions := strings.Split(substitutionExpression, ".")
-	if len(subExpressions) != 4 || subExpressions[0] != ResultTaskPart || subExpressions[2] != ResultResultPart {
-		return "", "", 0, fmt.Errorf("Must be of the form %q", resultExpressionFormat)
+
+	// For string result: tasks.<taskName>.results.<stringResultName>
+	// For array result: tasks.<taskName>.results.<arrayResultName>[index]
+	if len(subExpressions) == 4 && subExpressions[0] == ResultTaskPart && subExpressions[2] == ResultResultPart {
+		stringIdx := strings.TrimSuffix(strings.TrimPrefix(arrayIndexingRegex.FindString(subExpressions[3]), "["), "]")
+		subExpressions[3] = arrayIndexingRegex.ReplaceAllString(subExpressions[3], "")
+		if stringIdx != "" {
+			intIdx, _ := strconv.Atoi(stringIdx)
+			return subExpressions[1], subExpressions[3], intIdx, "", nil
+		}
+		return subExpressions[1], subExpressions[3], 0, "", nil
 	}
 
-	stringIdx := strings.TrimSuffix(strings.TrimPrefix(arrayIndexingRegex.FindString(subExpressions[3]), "["), "]")
-	subExpressions[3] = arrayIndexingRegex.ReplaceAllString(subExpressions[3], "")
-	if stringIdx != "" {
-		intIdx, _ := strconv.Atoi(stringIdx)
-		return subExpressions[1], subExpressions[3], intIdx, nil
+	// For object type result: tasks.<taskName>.results.<objectResultName>.<individualAttribute>
+	if len(subExpressions) == 5 && subExpressions[0] == ResultTaskPart && subExpressions[2] == ResultResultPart {
+		return subExpressions[1], subExpressions[3], 0, subExpressions[4], nil
 	}
-	return subExpressions[1], subExpressions[3], 0, nil
+
+	return "", "", 0, "", fmt.Errorf("Must be one of the form 1). %q; 2). %q", resultExpressionFormat, objectResultExpressionFormat)
 }
 
 // PipelineTaskResultRefs walks all the places a result reference can be used
--- a/pkg/apis/pipeline/v1beta1/resultref_test.go
+++ b/pkg/apis/pipeline/v1beta1/resultref_test.go
@@ -52,6 +52,17 @@ func TestNewResultReference(t *testing.T) {
 			Result:       "sumResult",
 		}},
 	}, {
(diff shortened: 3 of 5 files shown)


pkg/apis/pipeline/v1beta1/resultref.go
```diff
@@ -122,12 +129,16 @@ func stripVarSubExpression(expression string) string {
 	return strings.TrimSuffix(strings.TrimPrefix(expression, "$("), ")")
 }
 
-func parseExpression(substitutionExpression string) (string, string, error) {
+func parseExpression(substitutionExpression string) (string, string, string, error) {
 	subExpressions := strings.Split(substitutionExpression, ".")
-	if len(subExpressions) != 4 || subExpressions[0] != ResultTaskPart || subExpressions[2] != ResultResultPart {
-		return "", "", fmt.Errorf("Must be of the form %q", resultExpressionFormat)
+	if len(subExpressions) < 4 || len(subExpressions) > 5 || subExpressions[0] != ResultTaskPart || subExpressions[2] != ResultResultPart {
```
lbernick, line 134: this expression is difficult to read-- any chance you can add comments/helpful variable names/split into multiple lines or any of the above?

pkg/apis/pipeline/v1beta1/resultref_test.go
```diff
@@ -188,6 +207,25 @@ func TestHasResultReference(t *testing.T) {
 			PipelineTask: "sumTask2",
 			Result:       "sumResult2",
 		}},
+	}, {
+		name: "Test valid expression in object",
```
lbernick, line 211: just want to make sure I understand what's going on here-- the parameter named "param" is an object param with keys that are each individual results from pipeline tasks?

pkg/apis/pipeline/v1beta1/resultref_test.go
```diff
@@ -41,6 +41,17 @@ func TestNewResultReference(t *testing.T) {
 			PipelineTask: "sumTask",
 			Result:       "sumResult",
```
lbernick, line 42: could you add a test case for an object result with multiple properties? e.g. `(tasks.sumTask3.results.imageresult.digest)` and `(tasks.sumTask3.results.imageresult.tag)`

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -1208,10 +1212,10 @@ func TestValidatePipelineResults_Failure(t *testing.T) {
 	results := []PipelineResult{{
```
lbernick, line 1212: Is the reason this is invalid because the result is not an object type result?

pkg/apis/pipeline/v1beta1/resultref.go
```diff
@@ -26,10 +26,12 @@ import (
 type ResultRef struct {
 	PipelineTask string `json:"pipelineTask"`
 	Result       string `json:"result"`
+	Property     string `json:"property"`
 }
 
 const (
-	resultExpressionFormat = "tasks.<taskName>.results.<resultName>"
+	resultExpressionFormat       = "tasks.<taskName>.results.<resultName>"
+	objectResultExpressionFormat = "tasks.<taskName>.results.<objectResultName>.<individualAttribute>"
```
lbernick, line 34: It may be worth adding a comment here about some of the discussion in this section of the TEP https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#collisions-with-builtin-variable-replacement

pkg/apis/pipeline/v1beta1/resultref_test.go
chuangw6: > just want to make sure I understand what's going on here-- the parameter named "param" is an object param with keys that are each individual results from pipeline tasks?

Yes, correct!

pkg/apis/pipeline/v1beta1/resultref.go
chuangw6: sgtm! Thanks @lbernick

pkg/apis/pipeline/v1beta1/resultref_test.go
chuangw6: > could you add a test case for an object result with multiple properties? e.g. `(tasks.sumTask3.results.imageresult.digest)` and `(tasks.sumTask3.results.imageresult.tag)`

Added. Thanks @lbernick !

pkg/apis/pipeline/v1beta1/resultref.go
chuangw6: > this expression is difficult to read-- any chance you can add comments/helpful variable names/split into multiple lines or any of the above?

Great point. Just separated them and added comments!

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
lbernick: Just to check my understanding, `tasks.a-task.results.output.output` is no longer invalid, as it will be treated as the "output" key on the "output" result, and if such a result doesn't exist, there will be an error? Are the cases where the result or property doesn't exist tested anywhere?

pkg/apis/pipeline/v1beta1/resultref.go
lbernick: I'm still a bit confused here about what this comment is trying to say; is this accurate?

```
Result expressions of the form <resultName>.<attribute> will be treated as object results. If a string result name contains a dot, brackets should be used to differentiate it from an object result.
```

Also, could you link to the specific section of the TEP? (Much easier for someone reading than to have to look through two full TEPs.)

Lastly, has this support already been added? If not, that should probably be done first, so users have an option to express results with dots in the names before the way we interpret that syntax changes.

pkg/apis/pipeline/v1beta1/resultref.go
```diff
@@ -122,12 +133,21 @@ func stripVarSubExpression(expression string) string {
 	return strings.TrimSuffix(strings.TrimPrefix(expression, "$("), ")")
 }
 
-func parseExpression(substitutionExpression string) (string, string, error) {
+func parseExpression(substitutionExpression string) (string, string, string, error) {
 	subExpressions := strings.Split(substitutionExpression, ".")
-	if len(subExpressions) != 4 || subExpressions[0] != ResultTaskPart || subExpressions[2] != ResultResultPart {
-		return "", "", fmt.Errorf("Must be of the form %q", resultExpressionFormat)
+
+	// TODO (@yongxuanzhang): consider array type result here tasks.<taskName>.results.<resultName>[*] and indexing tasks.<taskName>.results.<resultName>[0]
```
lbernick, line 139: could you please link to the issue rather than a contributor? This provides a lot more context for anyone reading this comment.

pkg/apis/pipeline/v1beta1/resultref.go
```diff
@@ -122,12 +133,21 @@ func stripVarSubExpression(expression string) string {
...
+	// For string type result: tasks.<taskName>.results.<resultName>
+	if len(subExpressions) == 4 && subExpressions[0] == ResultTaskPart && subExpressions[2] == ResultResultPart {
+		return subExpressions[1], subExpressions[3], "", nil
 	}
-	return subExpressions[1], subExpressions[3], nil
+
+	// For object type result: tasks.<taskName>.results.<objectResultName>.<individualAttribute>
```
lbernick, line 145: Thanks for breaking this up-- much more readable now

pkg/apis/pipeline/v1beta1/resultref.go
Yongxuanzhang: You can link to https://github.com/tektoncd/pipeline/issues/4723 if you want

pkg/apis/pipeline/v1beta1/resultref.go
```diff
@@ -122,12 +133,21 @@ func stripVarSubExpression(expression string) string {
 	return strings.TrimSuffix(strings.TrimPrefix(expression, "$("), ")")
 }
 
-func parseExpression(substitutionExpression string) (string, string, error) {
+func parseExpression(substitutionExpression string) (string, string, string, error) {
```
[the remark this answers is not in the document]
chuangw6: great point. I'll add a comment for it.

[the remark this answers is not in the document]
Yongxuanzhang: Thanks! Great work on working on this part btw

pkg/apis/pipeline/v1beta1/resultref.go
chuangw6: Let me just remove the todo line since @Yongxuanzhang has implemented this (I guess that will be merged soon) .

Originally I added this just to avoid people's confusion (i.e. why only string result and object result are considered here not array type? 😆

Yongxuanzhang: Oh it is not opened yet, but this is a good reminder for me when I open that PR. We can keep it and link to our issue.

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
chuangw6: > Just to check my understanding, `tasks.a-task.results.output.output` is no longer invalid, as it will be treated as the "output" key on the "output" result, and if such a result doesn't exist, there will be an error? Are the cases where the result or property doesn't exist tested anywhere?

Great questions!

> if such a result doesn't exist, there will be an error?
Yes, this result ref validation is in place on the [reconciler level](https://github.com/tektoncd/pipeline/blob/33b3eac346a37ae7943f210fa13d9c4ea7d571fa/pkg/reconciler/pipelinerun/resources/validate_dependencies.go#L60-L87) (not here in the admission webhook level). Specifically, the error for the case when a result name doesn't exist will be returned from line 84.

> Are the cases where the result or property doesn't exist tested anywhere?
Yes, there will definitely be a validation for that. I already added the `property` field in the [`ResultRef` struct](https://github.com/tektoncd/pipeline/pull/4878/files#diff-49d33732c3f59a974dc47badc9ea139fe28e0e9b059c1f0aa478da3f78deaa09R26-R30). So IIUC the validation for object result properties will be done on the reconciler level too (will be implemented in future pr).

pkg/apis/pipeline/v1beta1/resultref.go
chuangw6: > I'm still a bit confused here about what this comment is trying to say; is this accurate?
> 
> ```
> Result expressions of the form <resultName>.<attribute> will be treated as object results. If a string result name contains a dot, brackets should be used to differentiate it from an object result.
> ```
> 
> Also, could you link to the specific section of the TEP? (Much easier for someone reading than to have to look through two full TEPs.)

Sounds great to me!! Thanks for the suggestion.

chuangw6: > Lastly, has this support already been added? If not, that should probably be done first, so users have an option to express results with dots in the names before the way we interpret that syntax changes.

Do you mean if `If a string result name contains a dot, brackets should be used to differentiate it from an object result.` is already supported?

chuangw6: Another question related to this. Do we really want to support the result name that contains dots(.)? The [result regex ](https://github.com/tektoncd/pipeline/blob/33b3eac346a37ae7943f210fa13d9c4ea7d571fa/pkg/apis/pipeline/v1beta1/resultref.go#L40) seems supporting that, but [parseExpression](https://github.com/tektoncd/pipeline/blob/33b3eac346a37ae7943f210fa13d9c4ea7d571fa/pkg/apis/pipeline/v1beta1/resultref.go#L125-L131) in the validation webhook seems not supporting this, which is inconsistent.

Also https://github.com/tektoncd/pipeline/issues/3590 seems only to proposed supporting param names that contain dots.

If we don't want dots in result names, we can change the [result regex ](https://github.com/tektoncd/pipeline/blob/33b3eac346a37ae7943f210fa13d9c4ea7d571fa/pkg/apis/pipeline/v1beta1/resultref.go#L40) to make sure result name cannot contain dots. This seems the quick solution for the conflicts with `objectResultName.key` name.

Yongxuanzhang: having dot in result name is problematic for parsing.

Yongxuanzhang: I will handle this case later

chuangw6: @Yongxuanzhang thanks!

@lbernick do you think it's okay to merge this pr first? Thanks!

chuangw6: update: talked to @lbernick @Yongxuanzhang @jerop about this offline. [TEP-0080](https://github.com/tektoncd/community/blob/main/teps/0080-support-domainscoped-parameterresult-names.md) proposed to support using dots in both param and result names. Param part was supported but not result. 

#4910 and #4909 are the tickets tracking result names containing dots.

pritidesai: I am a little surprised since this PR https://github.com/tektoncd/pipeline/pull/4268 fixed the results. 🤔

chuangw6: Thanks @pritidesai for the pointer.
Seems like #4268 has implemented the variable replacement part on the reconciler side for result names that contain dots and are referenced via bracket notation. But only the parsing in the validation webhook doesnt recognize `$(tasks.myTask.results["foo.bar"]` (https://github.com/tektoncd/pipeline/issues/4909#issuecomment-1138080594)

lbernick: After chatting with @chuangw6 I realized this is actually not a breaking change-- it allows more types of params to be validated than before, and doesn't change existing validation

lbernick: After chatting with @chuangw6 I realized this is actually not a breaking change-- it allows more types of params to be validated than before, and doesn't change existing validation

lbernick (approved):

ywluogg (commented):
Mostly LGTM except for another test case to be added

pkg/apis/pipeline/v1beta1/resultref.go
```diff
@@ -122,12 +133,32 @@ func stripVarSubExpression(expression string) string {
...
+	// For string type result: tasks.<taskName>.results.<resultName>
+	if len(subExpressions) == 4 && subExpressions[0] == ResultTaskPart && subExpressions[2] == ResultResultPart {
+		return subExpressions[1], subExpressions[3], "", nil
 	}
-	return subExpressions[1], subExpressions[3], nil
+
+	// For object type result: tasks.<taskName>.results.<objectResultName>.<individualAttribute>
+	if len(subExpressions) == 5 && subExpressions[0] == ResultTaskPart && subExpressions[2] == ResultResultPart {
+		return subExpressions[1], subExpressions[3], subExpressions[4], nil
+	}
```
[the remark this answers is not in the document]
chuangw6: sure! added!

pkg/apis/pipeline/v1beta1/resultref.go
```diff
@@ -122,12 +133,32 @@ func stripVarSubExpression(expression string) string {
...
+
+	return "", "", "", fmt.Errorf("Must be one of the form 1). %q; 2). %q", resultExpressionFormat, objectResultExpressionFormat)
```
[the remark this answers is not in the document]
chuangw6: Just added a test case for too many dot-separated components.
The test cases with other invalid format already exist in the function named `TestNewResultReference` which tests its caller, but its caller is expected to return empty resultref rather than an error. So that's why this message will not be printed out.

ywluogg (approved):

ywluogg:
Link this to #4723

abayer:
This needs a release note.

PR #4880, opened 2022-05-17
Extract full parameter name when using dots in bracket notation.

chitrangpatel:
Prior to this, when using `$params["<NAME>"]` or `$params['<NAME>']`,
if `NAME` contained `dots(.)` (eg. `org.foo.blah`), only the first part (ie. `org`) would be
extracted, instead of the entire name`org.foo.blah`. This PR now properly fetches the
entire name and throws an error if using `dots (.)` without `single`
or `double` quotes.

The following references show whats valid and whats not:

`$params.org.foo.blah` : Not Valid: the webhook will throw a validation error.
`$params["org.foo.blah"]`: Valid: Properly extracts the entire `NAME` (`org.foo.blah`).
`$params['org.foo.blah']`: Valid: Properly extracts the entire `NAME` (`org.foo.blah`).

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes
```
Extract full parameter name when using dots inside single/double quotes.
```

### DIFF
--- a/docs/tasks.md
+++ b/docs/tasks.md
@@ -833,8 +833,10 @@ variable values as follows:
 - To reference a parameter in a `Task`, use the following syntax, where `<name>` is the name of the parameter:
   ```shell
   # dot notation
-  $(params.<name>)
+  # Here, the name cannot contain dots (eg. foo.bar is not allowed). If the name contains `dots`, it can only be accessed via the bracket notation.
+  $(params.<name> )
   # or bracket notation (wrapping <name> with either single or double quotes):
+  # Here, the name can contain dots (eg. foo.bar is allowed).
   $(params['<name>'])
   $(params["<name>"])
   ```
--- a/pkg/substitution/substitution.go
+++ b/pkg/substitution/substitution.go
@@ -26,11 +26,12 @@ import (
 )
 
 const parameterSubstitution = `[_a-zA-Z][_a-zA-Z0-9.-]*(\[\*\])?`
+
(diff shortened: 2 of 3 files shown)


vdemeester (approved):

chitrangpatel:
> There's now some overly aggressive validation of previously valid params, that rejects the otherwise valid `resources.input.foo.path`:
> 
> ```
>     cluster_resource_test.go:72: Failed to create Task `cluster-resource-kvkmoqgu`: admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: Invalid referencing of parameters in $(resources.inputs.target-cluster.path)/kubeconfig !!! You can only use the dots inside single or double quotes. eg. $(params["org.foo.blah"]) or $(params['org.foo.blah']) are valid references but NOT $params.org.foo.blah.: spec.steps[0].args[0]
> ```

Hmm. Is the rule `$params.org.foo.blah` invalid only when accessing parameter names (ie. when the prefix is `params`)? Does it not apply to any other case (like `resources.input.` or `resources.output` or anything else that we might use)? If so, I can throw an error only in case of `params...` and not any other type.

imjasonh:
I'd probably say only `resources.*` is allowed to have 4 components, `params.*` must have two components or use bracket notation. Any other prefix might be invalid, unless I'm missing some cases (entirely possible).

chitrangpatel:
@imjasonh All the integration tests now pass. I think it should be ok now 🤞.

imjasonh:
/lgtm

This seems to work in my downstream tests. Thanks for fixing this! 👍

PR #4883, opened 2022-05-18
[TEP-0075] Validate object keys, PipelineRunSpec -> PipelineSpec

chuangw6:
# Changes
Add validation that checks if object param value from `PipelineRunSpec`
misses some keys required for that object param declared in `PipelineSpec`.

# Submitter Checklist

  functionality, content, code)

# Release Notes


ywluogg:
> /assign @ywluogg

I feel like this is a bug since I can get notification from Yongxuan's assignment but not from yours

ywluogg:
Link this to #4723

ywluogg:
Link this to #4723

ywluogg (approved):
Nice!

dibyom (approved):
Thanks @chuangw6 Could you add a release note?

chuangw6:
> Thanks @chuangw6 Could you add a release note?

Added! Thanks

Also added more tests to get coverage rate back 😃

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -1117,6 +1145,26 @@ spec:
...
+  params:
+    - name: some-param
+      value:
+        key1: "a"
+`),
+		reason:         ReasonObjectParameterMissKeys,
+		permanentError: true,
+		wantEvents: []string{
+			"Normal Started",
+			"Warning Failed PipelineRun foo/pipeline-missing-object-param-keys parameters is missing some object keys",
```
lbernick, line 1166: This error message would be more helpful if it returned which keys were missing and for which parameters

chuangw6: sg! Changed the verifier helper function to return a map for param<->missing keys so when the users see the err message from pipelinerun status, they will see which keys were missing and for which parameters.

But for the wantEvent, if I also put the returned err message here, it gives weird error and fails test though the expected and got event reported from the following message are exactly same. I guess this is why other wantEvent cuts off the second half message too.

```
logger.go:130: 2022-06-15T11:43:08.897-0700     ERROR   TestReconcile_InvalidPipelineRuns/invalid-pipeline-missing-object-keys  pipelinerun/reconciler.go:294Returned an error       {"targetMethod": "ReconcileKind", "error": "1 error occurred:\n\t* PipelineRun missing object keys for parameters: map[some-param:[key2]]\n\n"}
        pipelinerun_test.go:1316: error in test pipeline-missing-object-param-keys: expected event "Warning Failed PipelineRun foo/pipeline-missing-object-param-keys parameters is missing object keys required by Pipeline foo/a-pipeline-with-object-params's parameters: PipelineRun missing object keys for parameters: map[some-param:[key2]]" but got "Warning Failed PipelineRun foo/pipeline-missing-object-param-keys parameters is missing object keys required by Pipeline foo/a-pipeline-with-object-params's parameters: PipelineRun missing object keys for parameters: map[some-param:[key2]]" instead
```

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -157,9 +157,9 @@ func wrongTypeParamsNames(params []v1beta1.Param, matrix []v1beta1.Param, needed
 	return wrongTypeParamNames
 }
 
-// missingKeysObjectParamNames checks if all required keys of object type params are provided in taskrun params.
+// MissingKeysObjectParamNames checks if all required keys of object type params are provided in taskrun params.
 // TODO (@chuangw6): This might be refactored out to support single pair validation.
```
lbernick, line 161: could you please link an issue here instead of your GH handle?

pkg/reconciler/pipelinerun/pipelinerun_test.go
lbernick: That's really weird-- can you try it out on the cluster and see what error message is returned to the user?

pkg/reconciler/taskrun/validate_resources.go
chuangw6: Just deleted this todo since that functionality has been supported in the validation webhook part.

pkg/reconciler/pipelinerun/pipelinerun_test.go
chuangw6: sure. Just added.

chuangw6: This is the error I refer to when we also add the rest of the error message to the expected `wantEvent`.
Screen Shot 2022-06-17 at 17 51 37

So I guess we need to remove the rest part to pass the unit test @lbernick

chuangw6: But in the error returned when users miss some keys, we can see the full message without a problem with the information about which keys were missing and for which parameters.

PR #4901, opened 2022-05-23
Fix test cases for validatePipelineParameterVariables function

chuangw6:
# Changes
Prior to this commit, many success tests were meaningless because they
just use the param names in the format of `$(baz)`, `$(foo-is-baz)`.
In validation, without using the format of `$(params.<Name>)`, the
two examples above will be treated as just string literal instead of param
reference, and thus  in fact validation agains using correct params
are not executed. That explains why there is no problem even though
declared param is actually called `baz`, but using `$(bazzzz)`.

In this commit, the usage of param names is corrected to `$(params.<Name>)`
and type for ArrayOrString is added so that the param reference is really
validated.

# Submitter Checklist

  functionality, content, code)

# Release Notes

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
@@ -1239,7 +1239,7 @@ func TestValidatePipelineParameterVariables_Success(t *testing.T) {
 			Name:    "bar",
 			TaskRef: &TaskRef{Name: "bar-task"},
 			Params: []Param{{
-				Name: "a-param", Value: ArrayOrString{StringVal: "$(baz) and $(foo-is-baz)"},
+				Name: "a-param", Value: ArrayOrString{Type: ParamTypeString, StringVal: "$(params.baz) and $(params.foo-is-baz)"},
 			}},
 		}},
 	}, {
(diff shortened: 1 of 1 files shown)


lbernick (approved):
Nice catch! I'm curious if the `type: ParamTypeString` is needed, since users don't have to declare the param type if it is a string

chuangw6:
> Nice catch! I'm curious if the `type: ParamTypeString` is needed, since users don't have to declare the param type if it is a string

Yes, that's needed here in the test because it's used in [`validatePipelineParametersVariablesInTaskParameters `](https://github.com/tektoncd/pipeline/blob/e24df8379e7f85e9305f2710e77ba2190c722532/pkg/apis/pipeline/v1beta1/param_types.go#L239).

In yaml file, users don't have to specify the param type because the [unmarshal](https://github.com/tektoncd/pipeline/blob/e24df8379e7f85e9305f2710e77ba2190c722532/pkg/apis/pipeline/v1beta1/param_types.go#L143-L181) operation will infer its type and set it i.e.

the following declaration will have `string` as the inferred type.
```
params:
- name: myString
   value: "test value"
```
the following declaration will have `array` as the inferred type.
```
params:
- name: myArray
   value: 
     - a
     - b
```
the following declaration will have `object` as the inferred type.
```
params:
- name: myObject
   value: 
     key1: "val1"
     key2: "val2"
```

ywluogg:
Link this to #4723

ywluogg (approved):

PR #4902, opened 2022-05-24
TEP-0075: Validate Pipeline object variables in value, matrix and when

chuangw6:
# Changes
Besides task steps, parameter variables can be used as the value of
another param, matrix or when expression.

- Allowed referencing the whole object value but added validation to make sure it can only be used when providing values
for other object param. example:
```
 	params:
 	- name: arg
           value: $(params.myObject[*])
```
- In all other cases (string/array val, matrix and when expression), only individual object attributes can be referenced i.e. `$(params.myobject.key1)`.

# Submitter Checklist

  functionality, content, code)

# Release Notes

### DIFF
--- a/pkg/apis/pipeline/v1beta1/param_types.go
+++ b/pkg/apis/pipeline/v1beta1/param_types.go
@@ -284,23 +284,32 @@ func ArrayReference(a string) string {
 	return strings.TrimSuffix(strings.TrimPrefix(a, "$("+ParamsPrefix+"."), "[*])")
 }
 
-func validatePipelineParametersVariablesInTaskParameters(params []Param, prefix string, paramNames sets.String, arrayParamNames sets.String) (errs *apis.FieldError) {
+// validatePipelineParametersVariablesInTaskParameters validates param value that
+// may contain the reference(s) to other params to make sure those references are used appropriately.
+func validatePipelineParametersVariablesInTaskParameters(params []Param, prefix string, paramNames sets.String, arrayParamNames sets.String, objectParamNameKeys map[string][]string) (errs *apis.FieldError) {
 	for _, param := range params {
-		if param.Value.Type == ParamTypeString {
-			errs = errs.Also(validateStringVariable(param.Value.StringVal, prefix, paramNames, arrayParamNames).ViaFieldKey("params", param.Name))
-		} else {
+		switch param.Value.Type {
+		case ParamTypeArray:
 			for idx, arrayElement := range param.Value.ArrayVal {
-				errs = errs.Also(validateArrayVariable(arrayElement, prefix, paramNames, arrayParamNames).ViaFieldIndex("value", idx).ViaFieldKey("params", param.Name))
+				errs = errs.Also(validateArrayVariable(arrayElement, prefix, paramNames, arrayParamNames, objectParamNameKeys).ViaFieldIndex("value", idx).ViaFieldKey("params", param.Name))
 			}
+		case ParamTypeObject:
+			for key, val := range param.Value.ObjectVal {
+				errs = errs.Also(validateStringVariable(val, prefix, paramNames, arrayParamNames, objectParamNameKeys).ViaFieldKey("properties", key).ViaFieldKey("params", param.Name))
+			}
+		default:
+			errs = errs.Also(validateParamStringValue(param, prefix, paramNames, arrayParamNames, objectParamNameKeys))
 		}
 	}
 	return errs
 }
 
-func validatePipelineParametersVariablesInMatrixParameters(matrix []Param, prefix string, paramNames sets.String, arrayParamNames sets.String) (errs *apis.FieldError) {
+// validatePipelineParametersVariablesInMatrixParameters validates matrix param value
+// that may contain the reference(s) to other params to make sure those references are used appropriately.
+func validatePipelineParametersVariablesInMatrixParameters(matrix []Param, prefix string, paramNames sets.String, arrayParamNames sets.String, objectParamNameKeys map[string][]string) (errs *apis.FieldError) {
 	for _, param := range matrix {
 		for idx, arrayElement := range param.Value.ArrayVal {
-			errs = errs.Also(validateArrayVariable(arrayElement, prefix, paramNames, arrayParamNames).ViaFieldIndex("value", idx).ViaFieldKey("matrix", param.Name))
+			errs = errs.Also(validateArrayVariable(arrayElement, prefix, paramNames, arrayParamNames, objectParamNameKeys).ViaFieldIndex("value", idx).ViaFieldKey("matrix", param.Name))
 		}
 	}
 	return errs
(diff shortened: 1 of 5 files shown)


ywluogg:
#4723

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -1285,10 +1285,152 @@ func TestValidatePipelineParameterVariables_Success(t *testing.T) {
...
+			TaskRef: &TaskRef{Name: "bar-task"},
+			Params: []Param{{
+				Name: "an-object-param", Value: ArrayOrString{Type: ParamTypeObject, ObjectVal: map[string]string{
+					"url":    "$(params.myObject.key1)",
+					"commit": "$(params.myString)",
+				}},
+			}},
+		}},
+	}, {
+		name: "object param - using individual variables in matrix",
```
ywluogg, line 1376: Do you plan to catch references to objects and params as a whole and invalidate them?

ywluogg:
Can you mention in the description this is for webhook validation in PR description?

ywluogg (commented):
Mostly LGTM, can you explain a bit for matrix that webhook validation works for referencing object / params as a whole, but they are actually not supported

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -236,23 +237,28 @@ func ArrayReference(a string) string {
 	return strings.TrimSuffix(strings.TrimPrefix(a, "$("+ParamsPrefix+"."), "[*])")
 }
 
-func validatePipelineParametersVariablesInTaskParameters(params []Param, prefix string, paramNames sets.String, arrayParamNames sets.String) (errs *apis.FieldError) {
+func validatePipelineParametersVariablesInTaskParameters(params []Param, prefix string, paramNames sets.String, arrayParamNames sets.String, objectParamNameKeys map[string][]string) (errs *apis.FieldError) {
```
lbernick, line 240: Would you mind adding a docstring for this function and the one below it?

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -284,12 +290,48 @@ func validateParameterInOneOfMatrixOrParams(matrix []Param, params []Param) (err
 	return errs
 }
 
-func validateStringVariable(value, prefix string, stringVars sets.String, arrayVars sets.String) *apis.FieldError {
+// validateStringVariableP validates param variable usage in string value
+// This will allow referencing the whole object/array value. example:
+// params:
+// 	- name: arg
+// 		value: $(params.myObject[*])
+// see https://github.com/tektoncd/pipeline/issues/4879#issuecomment-1133120522 for background
```
lbernick, line 298: I think there's not so much context in this issue comment that it makes sense to link to it; it can just be explained here. I would also focus more on what this function does based on its inputs, and if you're going to include example syntax make sure it's formatted correctly.

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -284,12 +290,48 @@ func validateParameterInOneOfMatrixOrParams(matrix []Param, params []Param) (err
...
+func validateStringVariableP(param Param, prefix string, paramNames sets.String, arrayVars sets.String, objectParamNameKeys map[string][]string) (errs *apis.FieldError) {
+	stringValue := param.Value.StringVal
+	singleVariableRegex := regexp.MustCompile(fmt.Sprintf("^\\$\\(%s\\.[_a-zA-Z][_a-zA-Z0-9.-]*\\[\\*\\]\\)$", prefix))
```
lbernick, line 301: similar to previously-- let's avoid `MustCompile` (except in globals) since it can panic

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -284,12 +290,48 @@ func validateParameterInOneOfMatrixOrParams(matrix []Param, params []Param) (err
...
+	// 	params:
+	// 	- name: arg
+	// 		value: $(params.myObject[*])
+	if singleVariableRegex.MatchString(stringValue) {
+		return errs.Also(substitution.ValidateVariableP(stringValue, prefix, paramNames).ViaFieldKey("params", param.Name))
+	}
+
+	// if the provided param value is string literal and/or contains multiple variables
+	// valid example: "$(params.myString) and another $(params.myObject.key1)"
+	// invalid example: "$(params.myString) and another $(params.myObject[*])"
```
lbernick, line 313: I'm a bit confused what's going on here. How is `validateVariableP` different from `validateStringVariable` and `validateStringVariableP`? Can any of these functions be renamed?

also, is the problem with the second example that it is a string which contains both a string param and and object param (without referencing the key)?

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -3054,3 +3334,14 @@ func enableFeatures(t *testing.T, features []string) func(context.Context) conte
 		return s.ToContext(ctx)
 	}
 }
+
+func getContextBasedOnFeatureFlag(featureFlag string) context.Context {
```
lbernick, line 3338: you can use this helper https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go#L844-L855

lbernick (commented):
Could you please update your commit message to make it more clear what is changing as a result of your PR? e.g. "Add validation for usage of parameter variables in Pipeline parameters, matrixes, and when expressions. Valid usage includes.... Invalid usage includes...."

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
chuangw6: It seems like we cannot import `v1beta1_test` package from `pipeline_validation_test.go` that is under `v1beta1` package, can we?

lbernick: oh that's frustrating-- ideally pipeline_validation_test.go should be in the `v1beta1_test` package instead of `v1beta1`, because we want to only test exported functions. So you can either move pipeline_validation_test.go to `v1beta1_test` (may or may not work depending on if we are already testing unexported functions) or you could move the helper to the `v1beta1` package instead which should definitely work.

pkg/apis/pipeline/v1beta1/param_types.go
dibyom, line 301: Question: Don't we have a regex or helper methods we could use for this instead of another regex here?

dibyom (approved):

pkg/apis/pipeline/v1beta1/param_types.go
chuangw6: Done!

pkg/apis/pipeline/v1beta1/param_types.go
chuangw6: Done!

pkg/apis/pipeline/v1beta1/param_types.go
chuangw6: > I'm a bit confused what's going on here. How is `validateVariableP` different from `validateStringVariable` and `validateStringVariableP`? Can any of these functions be renamed?

Thanks. The names were indeed a bit confusing. I changed the name and add comments. Please take a look. Basically, `validateStringVariableP` (now called `validateParamStringValue`) allows using the references to the whole array or object i.e. `$(params.myarray[*])`, or `$(params['myarray'][*])` whereas the existing `validateStringVariable` doesn't allow to have a reference that contains an array param name whatsoever.

> also, is the problem with the second example that it is a string which contains both a string param and and object param (without referencing the key)?

Exactly!

pkg/apis/pipeline/v1beta1/param_types.go
```diff
@@ -284,12 +290,48 @@ func validateParameterInOneOfMatrixOrParams(matrix []Param, params []Param) (err
 	return errs
 }
 
-func validateStringVariable(value, prefix string, stringVars sets.String, arrayVars sets.String) *apis.FieldError {
+// validateStringVariableP validates param variable usage in string value
+// This will allow referencing the whole object/array value. example:
+// params:
+// 	- name: arg
+// 		value: $(params.myObject[*])
```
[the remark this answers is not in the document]
chuangw6: This part is changed. Please have a look.

pkg/apis/pipeline/v1beta1/param_types.go
lbernick: thanks! the docstring should ideally focus on what the function does, not which other functions call it, so I'd remove the reference to PipelineTask (both here and in other docstrings)

pkg/substitution/substitution.go
```diff
@@ -25,8 +25,12 @@ import (
 	"knative.dev/pkg/apis"
 )
 
-const parameterSubstitution = `.*?(\[\*\])?`
-const braceMatchingRegex = "(\\$(\\(%s(\\.(?P<var1>%s)|\\[\"(?P<var2>%s)\"\\]|\\['(?P<var3>%s)'\\])\\)))"
+const (
+	parameterSubstitution = `.*?(\[\*\])?`
+
+	// BraceMatchingRegex is a regex for parameter references including dot notation, bracket notation with single and double quotes.
+	BraceMatchingRegex = "(\\$(\\(%s(\\.(?P<var1>%s)|\\[\"(?P<var2>%s)\"\\]|\\['(?P<var3>%s)'\\])\\)))"
```
lbernick, line 32: could you avoid exporting this, and instead create an exported helper function?

chuangw6: Done!

pkg/apis/pipeline/v1beta1/param_types.go
chuangw6: Changed! Thanks for the suggestion!

pkg/substitution/substitution.go
lbernick: I think the purpose of this package is to validate and parse parameters (it's hard to tell haha-- substitution is a bit messy), so the reason I suggest writing a helper function is to think about what validation and parsing can be abstracted away. The functions exported by a package form a sort of API that other packages can use. Imagine if you encountered an API with this getter function-- I think it would be confusing. This is where a package just for params would be really helpful! Since we don't have that, could you try to write a helper function that's consistent with what the `substitution` package is designed for? I am hoping to avoid the use of regexes in many places in the codebase because they are very hard to understand.

chuangw6: Thank you @lbernick for the explanation. That makes sense to me! Done.

PR #4908, opened 2022-05-25
[TEP-0076]Support Array Results substitution

Yongxuanzhang:
# Changes
This is part of work in [TEP-0076.](https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md)
This commit provides the support to apply array results replacements.
Previous this commit we support emitting array results so users can
write array results to task level, but we cannot pass array results from
tasks within one pipeline. This commit adds the support for this.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes
```release-note
Support array results substitution as an alpha feature.

  results:
    - name: array-results
       type: array
       description: The array results

echo -n "[\"hello\",\"world\"]" | tee $(results.array-results.path)

and we can refer to the array results in param like:
  params:
    - name: foo
      value: "$(tasks.task1.results.array-results[*])"

This feature is part of the TEP-0076. 
```

### DIFF
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -908,7 +908,10 @@ Tasks can emit [`Results`](tasks.md#emitting-results) when they execute. A Pipel
 Sharing `Results` between `Tasks` in a `Pipeline` happens via
 [variable substitution](variables.md#variables-available-in-a-pipeline) - one `Task` emits
 a `Result` and another receives it as a `Parameter` with a variable such as
-`$(tasks.<task-name>.results.<result-name>)`.
+`$(tasks.<task-name>.results.<result-name>)`. Array `Results` is supported as alpha feature and
+can be referer as `$(tasks.<task-name>.results.<result-name>[*])`.
+
+**Note:** Array `Result` cannot be used in `script`.
 
 When one `Task` receives the `Results` of another, there is a dependency created between those
 two `Tasks`. In order for the receiving `Task` to get data from another `Task's` `Result`,
@@ -923,6 +926,8 @@ before this one.
 params:
   - name: foo
     value: "$(tasks.checkout-source.results.commit)"
+  - name: array-params
+    value: "$(tasks.checkout-source.results.array-results[*])"
 ```
 
 **Note:** If `checkout-source` exits successfully without initializing `commit` `Result`,
--- a/docs/variables.md
+++ b/docs/variables.md
@@ -25,6 +25,9 @@ For instructions on using variable substitutions see the relevant section of [th
 | `tasks.<taskName>.results.<resultName>[i]` | The ith value of the `Task's` array result. Can alter `Task` execution order within a `Pipeline`.) |
 | `tasks.<taskName>.results['<resultName>'][i]` | (see above)) |
 | `tasks.<taskName>.results["<resultName>"][i]` | (see above)) |
+| `tasks.<taskName>.results.<resultName>[*]` | The array value of the `Task's` result. Can alter `Task` execution order within a `Pipeline`. Cannot be used in `script`.) |
+| `tasks.<taskName>.results['<resultName>'][*]` | (see above)) |
+| `tasks.<taskName>.results["<resultName>"][*]` | (see above)) |
 | `workspaces.<workspaceName>.bound` | Whether a `Workspace` has been bound or not. "false" if the `Workspace` declaration has `optional: true` and the Workspace binding was omitted by the PipelineRun. |
 | `context.pipelineRun.name` | The name of the `PipelineRun` that this `Pipeline` is running in. |
 | `context.pipelineRun.namespace` | The namespace of the `PipelineRun` that this `Pipeline` is running in. |
--- a/pkg/apis/pipeline/v1beta1/result_types.go
+++ b/pkg/apis/pipeline/v1beta1/result_types.go
@@ -13,6 +13,8 @@ limitations under the License.
 
 package v1beta1
 
+import "strings"
+
 // TaskResult used to describe the results of a task
 type TaskResult struct {
 	// Name the given name
@@ -64,3 +66,8 @@ const (
 
 // AllResultsTypes can be used for ResultsTypes validation.
 var AllResultsTypes = []ResultsType{ResultsTypeString, ResultsTypeArray, ResultsTypeObject}
+
+// ResultsArrayReference returns the reference of the result. e.g. results.resultname from $(results.resultname[*])
+func ResultsArrayReference(a string) string {
+	return strings.TrimSuffix(strings.TrimSuffix(strings.TrimPrefix(a, "$("), ")"), "[*]")
+}
--- a/pkg/apis/pipeline/v1beta1/when_types.go
+++ b/pkg/apis/pipeline/v1beta1/when_types.go
@@ -62,8 +62,11 @@ func (we *WhenExpression) applyReplacements(replacements map[string]string, arra
 	for _, val := range we.Values {
 		// arrayReplacements holds a list of array parameters with a pattern - params.arrayParam1
 		// array params are referenced using $(params.arrayParam1[*])
+		// array results are referenced using $(results.resultname[*])
 		// check if the param exist in the arrayReplacements to replace it with a list of values
 		if _, ok := arrayReplacements[fmt.Sprintf("%s.%s", ParamsPrefix, ArrayReference(val))]; ok {
+			replacedValues = append(replacedValues, substitution.ApplyArrayReplacements(val, replacements, arrayReplacements)...)
+		} else if _, ok := arrayReplacements[ResultsArrayReference(val)]; ok {
 			replacedValues = append(replacedValues, substitution.ApplyArrayReplacements(val, replacements, arrayReplacements)...)
 		} else {
 			replacedValues = append(replacedValues, substitution.ApplyReplacements(val, replacements))
--- a/pkg/apis/pipeline/v1beta1/when_types_test.go
+++ b/pkg/apis/pipeline/v1beta1/when_types_test.go
@@ -249,6 +249,43 @@ func TestApplyReplacements(t *testing.T) {
 			Values:   []string{"barfoo"},
 		},
 	}, {
+		name: "replace array results variables",
+		original: &WhenExpression{
+			Input:    "$(tasks.foo.results.bar)",
+			Operator: selection.In,
+			Values:   []string{"$(tasks.aTask.results.aResult[*])"},
+		},
+		replacements: map[string]string{
+			"tasks.foo.results.bar": "foobar",
+		},
+		arrayReplacements: map[string][]string{
+			"tasks.aTask.results.aResult": {"dev", "stage"},
+		},
+		expected: &WhenExpression{
+			Input:    "foobar",
+			Operator: selection.In,
+			Values:   []string{"dev", "stage"},
+		},
+	}, {
+		name: "invaliad array results replacements",
+		original: &WhenExpression{
+			Input:    "$(tasks.foo.results.bar)",
+			Operator: selection.In,
+			Values:   []string{"$(tasks.aTask.results.aResult[invalid])"},
+		},
+		replacements: map[string]string{
+			"tasks.foo.results.bar":          "foobar",
+			"tasks.aTask.results.aResult[*]": "barfoo",
+		},
+		arrayReplacements: map[string][]string{
+			"tasks.aTask.results.aResult[*]": {"dev", "stage"},
+		},
+		expected: &WhenExpression{
+			Input:    "foobar",
+			Operator: selection.In,
+			Values:   []string{"$(tasks.aTask.results.aResult[invalid])"},
+		},
+	}, {
 		name: "replace array params",
 		original: &WhenExpression{
 			Input:    "$(params.path)",
--- a/pkg/reconciler/pipelinerun/resources/apply_test.go
+++ b/pkg/reconciler/pipelinerun/resources/apply_test.go
@@ -1052,6 +1052,38 @@ func TestApplyTaskResults_MinimalExpression(t *testing.T) {
 			},
 		}},
 	}, {
+		name: "Test array result substitution on minimal variable substitution expression - params",
+		resolvedResultRefs: ResolvedResultRefs{{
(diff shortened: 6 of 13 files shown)


pkg/apis/pipeline/v1beta1/result_types.go
```diff
@@ -60,3 +62,9 @@ const (
 
 // AllResultsTypes can be used for ResultsTypes validation.
 var AllResultsTypes = []ResultsType{ResultsTypeString, ResultsTypeArray, ResultsTypeObject}
+
+// ResultsArrayReference returns the reference of the result from array parameter reference
```
ywluogg, line 66: Comment format

ywluogg (commented):
Mostly LGTM except for the test case

pkg/apis/pipeline/v1beta1/when_types_test.go
```diff
@@ -248,6 +248,25 @@ func TestApplyReplacements(t *testing.T) {
 			Operator: selection.In,
 			Values:   []string{"barfoo"},
 		},
+	}, {
+		name: "replace array results variables",
+		original: &WhenExpression{
+			Input:    "$(tasks.foo.results.bar)",
+			Operator: selection.In,
+			Values:   []string{"$(tasks.aTask.results.aResult[*])"},
```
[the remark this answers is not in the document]
Yongxuanzhang: yes!

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -78,6 +78,11 @@ func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params
 	if extraParamsNames := extraParamsNames(ctx, neededParamsNames, providedParamsNames); len(extraParamsNames) != 0 {
 		return fmt.Errorf("didn't need these params but they were provided anyway: %s", extraParamsNames)
 	}
+	// This is needed to support array replacements in params.Users want to use $(results.resultname[*])
```
ywluogg, line 81: extra space

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -88,6 +93,16 @@ func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params
 	return nil
 }
 
+func correctAoStype(params []v1beta1.Param, neededParamsTypes map[string]v1beta1.ParamType) {
```
ywluogg, line 96: Maybe correctValueType? or correctAoSType

ywluogg, line 96: Can you add a comment to explain what this function does?

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -78,6 +78,11 @@ func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params
 	if extraParamsNames := extraParamsNames(ctx, neededParamsNames, providedParamsNames); len(extraParamsNames) != 0 {
 		return fmt.Errorf("didn't need these params but they were provided anyway: %s", extraParamsNames)
 	}
+	// This is needed to support array replacements in params.Users want to use $(results.resultname[*])
+	// to pass array result to array param, yet in yaml format this will be
+	// unmarshalled to string for ArrayOrString. So we need to check and correct the param type.
+	// Please check issue #4879 for more details and examples.
+	correctAoStype(params, neededParamsTypes)
```
ywluogg, line 85: This actually worths adding a separate test in validate_resource_test for checking the changed params

pkg/reconciler/taskrun/validate_resources.go
Yongxuanzhang: Sure!

pkg/reconciler/taskrun/validate_resources.go
Yongxuanzhang: sure! That's a good point!

pkg/reconciler/taskrun/validate_resources.go
Yongxuanzhang, line 96: My question of this is that is it ok if we put this function here? Since we are doing validation in these functions it may seem weird to add one function to do something different. If we want to move it to higher level then it may be put inside `reconcile`

ywluogg (approved):

examples/v1beta1/pipelineruns/alpha/pipelinerun-array-results.yaml
```diff
@@ -0,0 +1,56 @@
...
+            - name: bar
+              type: string
+              default: "defaultparam1"
+          steps:
+            - name: print-foo
+              image: bash:latest
+              args: [
+                "echo",
+                "$(params.foo[*])"
+              ]
```
Yongxuanzhang, line 42: we can only echo the array this way since it cannot be used in script right now

pkg/apis/pipeline/v1beta1/when_types.go
```diff
@@ -62,9 +62,12 @@ func (we *WhenExpression) applyReplacements(replacements map[string]string, arra
 	for _, val := range we.Values {
 		// arrayReplacements holds a list of array parameters with a pattern - params.arrayParam1
 		// array params are referenced using $(params.arrayParam1[*])
+		// array results are referenced using $(results.resultname[*])
 		// check if the param exist in the arrayReplacements to replace it with a list of values
 		if _, ok := arrayReplacements[fmt.Sprintf("%s.%s", ParamsPrefix, ArrayReference(val))]; ok {
 			replacedValues = append(replacedValues, substitution.ApplyArrayReplacements(val, replacements, arrayReplacements)...)
+		} else if _, ok := arrayReplacements[fmt.Sprintf("%s", ResultsArrayReference(val))]; ok {
```
chuangw6, line 69: Seems like the `if` and `else-if` bodies are doing same thing. Can you merge those two blocks together?

pkg/apis/pipeline/v1beta1/when_types_test.go
```diff
@@ -248,6 +248,44 @@ func TestApplyReplacements(t *testing.T) {
...
+		},
+		replacements: map[string]string{
+			"tasks.foo.results.bar":          "foobar",
+			"tasks.aTask.results.aResult[*]": "barfoo",
+		},
+		arrayReplacements: map[string][]string{
+			"tasks.aTask.results.aResult[*]": {"dev", "stage"},
```
chuangw6, line 263: Why can the whole array reference exist in both string replacement and array replacement?

pkg/reconciler/pipelinerun/resources/apply_test.go
```diff
@@ -662,6 +662,71 @@ func TestApplyTaskResults_MinimalExpression(t *testing.T) {
...
+			FromTaskRun: "aTaskRun",
+		}},
+		targets: PipelineRunState{{
+			PipelineTask: &v1beta1.PipelineTask{
+				Name:    "bTask",
+				TaskRef: &v1beta1.TaskRef{Name: "bTask"},
+				Params: []v1beta1.Param{{
+					Name: "bParam",
+					Value: v1beta1.ArrayOrString{Type: v1beta1.ParamTypeArray,
+						ArrayVal: []string{`$(tasks.aTask.results["a.Result"][*])`},
```
chuangw6, line 682: just want to confirm. The validation webhook doesn't recognize this `$(tasks.aTask.results["a.Result"][*]` reference and will report error, is that right?

pkg/reconciler/taskrun/validate_resources_test.go
```diff
@@ -550,3 +556,65 @@ func TestValidateOverrides(t *testing.T) {
...
+		},
+		expectedTypes: map[string]v1beta1.ParamType{
+			"foo": v1beta1.ParamTypeString,
+		},
+	}, {
+		name: "given string needed array and param value is reference",
+		params: []v1beta1.Param{
+			{
+				Name:  "foo",
+				Value: *v1beta1.NewArrayOrString("$(results.resultname[*])"),
```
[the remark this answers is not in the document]
chuangw6: Also, if the string is `"$(params.stringA) followed by $(params.stringB)"`, `VariableSubstitutionRegex.MatchString` also returns true and the test passes. Is it expected to also **correct** the type in such case?

docs/variables.md
```diff
@@ -25,6 +25,9 @@ For instructions on using variable substitutions see the relevant section of [th
 | `tasks.<taskName>.results.<resultName>[i]` | The ith value of the `Task's` array result. Can alter `Task` execution order within a `Pipeline`.) |
 | `tasks.<taskName>.results['<resultName>'][i]` | (see above)) |
 | `tasks.<taskName>.results["<resultName>"][i]` | (see above)) |
```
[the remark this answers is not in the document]
Yongxuanzhang: > Is there another place we mention that [*] only applies to results that have type array as well as can only be used in params of type array
I can add it to the piplines.md's corresponding section.

> Also the fact that it cannot be used in scripts at the moment
It may be missing in current doc, I will check and add it!

pkg/reconciler/taskrun/validate_resources_test.go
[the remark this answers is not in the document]
Yongxuanzhang: > @Yongxuanzhang if I change `"$(results.resultname[*])"` to `"$(results.resultname)"`, the test still passes. Is that expected? Is there another place that we validate that without a `[*]`, the param should be invalid?

It is supported by the following function so it can pass, for actual use since I didn't add this type, this is not working actually. I'm not sure if we should support `"$(results.resultname)"` for array and object substitution. 
https://github.com/tektoncd/pipeline/blob/3147117bda072f801f03268eb2a6d1379a692962/pkg/substitution/substitution.go#L282-L290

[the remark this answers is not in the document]
Yongxuanzhang: tep 0076 mentions that we should use `*` so we shouldn't support this. It looks like a bit controversial here. 
https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md#to-provide-values-for-array-params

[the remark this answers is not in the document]
Yongxuanzhang: > Also, if the string is `"$(params.stringA) followed by $(params.stringB)"`, `VariableSubstitutionRegex.MatchString` also returns true and the test passes. Is it expected to also **correct** the type in such case?

discussed offline about this, will leave to @chuangw6's PR to handle type stuff. We will skip some validation cases here

pkg/apis/pipeline/v1beta1/when_types.go
Yongxuanzhang: That's true, for better readability it looks clear to split them, or could you suggest changes of merging them?

pkg/apis/pipeline/v1beta1/when_types_test.go
Yongxuanzhang: it should be a error, thanks!

pkg/reconciler/taskrun/validate_resources_test.go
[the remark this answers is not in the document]
Yongxuanzhang: Maybe leave this case as an issue and fix this in another pr.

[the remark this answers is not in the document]
Yongxuanzhang: yes, it will be our cleanup work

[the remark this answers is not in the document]
ywluogg: Can you confirm that this is being handled, as you already rebased the PR?

ywluogg:
#4723

pkg/apis/pipeline/v1beta1/result_types.go
```diff
@@ -60,3 +62,8 @@ const (
 
 // AllResultsTypes can be used for ResultsTypes validation.
 var AllResultsTypes = []ResultsType{ResultsTypeString, ResultsTypeArray, ResultsTypeObject}
+
+// ResultsArrayReference returns the reference of the result. e.g. results.resultname[*] from $(results.resultname[*])
```
dibyom, line 66:
```suggestion
// ResultsArrayReference returns the reference of the result. e.g. results.resultname from $(results.resultname[*])
```

docs/pipelines.md
```diff
@@ -95,7 +95,7 @@ A `Pipeline` definition supports the following fields:
         a `Task` requires.
         - [`from`](#using-the-from-field) - Indicates the data for a [`PipelineResource`](resources.md)
           originates from the output of a previous `Task`.
-      - [`runAfter`](#using-the-runafter-field) - Indicates that a `Task` should execute after one or more other
```
dibyom, line 98: nit: the changes in this file all seem to be about removing spaces. Can we revert them?

dibyom (commented):
Couple of small things, otherwise LGTM!

pkg/reconciler/taskrun/validate_resources_test.go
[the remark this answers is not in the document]
dibyom: ok the test here still passes if I replace $(results.foo[*]) with $(results.foo). Could we add some validation for this or at least a comment that this is upcoming work.

docs/pipelines.md
Yongxuanzhang: sure, let me figure it out. It must be autoformat of my IDE.

pkg/apis/pipeline/v1beta1/result_types.go
Yongxuanzhang: Thank you!

ywluogg:
Except for the existing open comments, other LGTM

pkg/reconciler/taskrun/validate_resources_test.go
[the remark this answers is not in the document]
Yongxuanzhang: After #5007 was merged, the current code supports $(results.foo) to refer to an array. I left one todo in this PR to be validate this use case and it should be invalid.

docs/pipelines.md
Yongxuanzhang: the main branch has been formatted, 😄

dibyom:
/approve

Thanks @Yongxuanzhang :D

PR #4834, opened 2022-05-04
[WIP] Add Metadata Support

austinzhao-go:
# Changes

Implementation for [TEP-0106](https://github.com/tektoncd/community/pull/695)

# TODO
- [ ] Confirm annotations propagating details
- [ ] Update validation, unit tests, and docs
- [ ] Apply similar logics on `labels`

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ ] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [ ] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [ ] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [ ] Release notes block below has been filled in
(if there are no user facing changes, use release note "NONE")

# Release Notes
```release-note
Support Specifying Metadata per Task in Runtime (PipelineRun)
```

PR #4914, opened 2022-05-27
Allow PipelineTaskRunSpec.Metadata to be optional.

### LINKED ISSUE #4913: Regression: PipelineTaskRunSpec.Metadata incompatible with older webhooks
# Expected Behavior

PipelineTaskRunSpec should be safely passed to older webhooks.

# Actual Behavior

```
    serviceaccount_test.go:163: Failed to create PipelineRun `pipeline-run-with-service-accounts-dsofxybx`: admission webhook "defaulting.webhook.pipeline.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "metadata"
    panic.go:661: ############################################
    panic.go:661: ### Dumping objects from arendelle-7ntt4 ###
    panic.go:661: ############################################
```

1. Install Tekton < https://github.com/tektoncd/pipeline/pull/4834
2. Create PipelineRun using client >= https://github.com/tektoncd/pipeline/pull/4834

# Additional Info

We think we've found the root cause (shout out to @n3wscott for finding it):

https://github.com/tektoncd/pipeline/blob/b19a9abdb81ac0d7608a0457348ccb24afa65316/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L605

should be:

```go
Metadata *PipelineTaskMetadata `json:"metadata,omitempty"` 
```

Will send a PR!

cc @austinzhao-go in case you have any thoughts

/kind bug

wlynch:
# Changes

Previously, this was using a non-pointer value, which means that the
field was not actually optional and was being included in client
requests. If this request was sent to an older server, this would break
during unmarshalling because the field doesn't exist yet for older
servers.

This changes the type to use a pointer so it will actually be omitted in
newer clients if it is not set.

Co-authored-by: Scott Nichols <n3wscott@chainguard.dev>

/kind bug
Fixes #4913

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/openapi_generated.go
+++ b/pkg/apis/pipeline/v1beta1/openapi_generated.go
@@ -2883,8 +2883,7 @@ func schema_pkg_apis_pipeline_v1beta1_PipelineTaskRunSpec(ref common.ReferenceCa
 					},
 					"metadata": {
 						SchemaProps: spec.SchemaProps{
-							Default: map[string]interface{}{},
-							Ref:     ref("github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1.PipelineTaskMetadata"),
+							Ref: ref("github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1.PipelineTaskMetadata"),
 						},
 					},
 				},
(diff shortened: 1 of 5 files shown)


jerop (approved):
thanks for the quick fix @wlynch!

cc other reviewer of https://github.com/tektoncd/pipeline/pull/4834 @lbernick

austinzhao-go:
Hi @wlynch, just realized this fix! (I totally missed the cc email, sry...)

thank you for helping correct this and I got the reason here. 
(also thanks @lbernick for referring this changes, as also required for TEP-0104 impl work
https://github.com/tektoncd/pipeline/pull/4877)

PR #4915, opened 2022-05-27
Add dependabot config.

wlynch:
# Changes

Enables dependabot for pipelines.

See
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
for more details.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    vendor: true


wlynch:
cc @imjasonh @afrittoli

vdemeester:
@wlynch this is linked to https://github.com/tektoncd/plumbing/issues/1028 right ?
One "tricky" thing about pipeline is that, we need to do "extra stuff" when we upgrade dependencies : (re)-generate code, …, maybe applying a temporary patch, … We might have to run our own "dependabot-like" way of upgrading dependencies instead of using dependabot direclty.

vdemeester:
> Yeah, though this probably won't make that any harder.
> 
> I think worst case we'll get a few PRs that won't pass unit tests (probably for anything k8s.io related), which will be a signal for additional intervention. Even if we can only use this with 80% of deps, it seems worthwhile to enable it. If this ends up being too noisy we can roll it back.

Fair enough 🙃

vdemeester:
@wlynch I am trying this with `tektoncd/cli`, as you convinced me this could work 🙃. The only *trick* to it is the `release-note` block that will be missing, but, might be ok-ish

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
vdemeester, line 6: Should we put this to `weekly` ? I moved to weekly for `hub` just because with npm it was generating a lot of PR often.. but maybe we are fine here.

vdemeester (approved):
/meow
Let there be dragons 😓 🙃

PR #4920, opened 2022-05-31
[TEP-0076]Validate Results type and object properties 

Yongxuanzhang:
# Changes
This is part of work in TEP-0076.
This commit provides the validation of results in reconciler after we
get the emitted results. It covers the validation of mismatched types
between the results emitted and specified, the results object
properties.
Previous this commit these validations are missing and errors are
ignored.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes
```release-note
Add Properties for TaskResult so user can define what keys are needed for object type TaskResult and leverage this Properties for validation.
```


ywluogg:
Can you add release notes for results?

ywluogg:
#4723

Yongxuanzhang:
> Can you add release notes for results?

Yes!

pkg/pod/status.go
```diff
@@ -140,9 +141,95 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev
...
+		if missingKeysObjectNames := missingKeyObjectResults(tr, trs, resolvedTaskSpec); len(missingKeysObjectNames) != 0 {
+			err := fmt.Errorf("missing keys for these results which are required in TaskResult's properties %v", missingKeysObjectNames)
+			merr = multierror.Append(merr, err)
+		}
+	}
+
 	return *trs, merr.ErrorOrNil()
 }
 
+func missmatchedTypesResults(tr v1beta1.TaskRun, trs *v1beta1.TaskRunStatus, resolvedTaskSpec *v1beta1.TaskSpec) []string {
```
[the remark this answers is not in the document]
Yongxuanzhang: thanks!

pkg/pod/status.go
```diff
@@ -140,9 +141,98 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev
 
 	trs.TaskRunResults = removeDuplicateResults(trs.TaskRunResults)
 
+	if resolvedTaskSpec != nil {
```
lbernick, line 144: the "pod" package doesn't seem like the best place for this functionality; it more likely belongs in the reconciler code.

pkg/pod/status.go
```diff
@@ -140,9 +141,98 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev
...
+
 	return *trs, merr.ErrorOrNil()
 }
 
+// mismatchedTypesResults checks and returns all the mismatched types of emitted results against specified results.
+func mismatchedTypesResults(tr v1beta1.TaskRun, trs *v1beta1.TaskRunStatus, resolvedTaskSpec *v1beta1.TaskSpec) []string {
+	neededTypes := make(map[string][]string)
+	providedTypes := make(map[string][]string)
+	// collect needed keys for object results
+	if tr.Spec.TaskSpec != nil {
```
lbernick, line 166: I think it would make more sense for this function to have only one task spec as the source of truth (and same for the function below it)

pkg/pod/status_test.go
```diff
@@ -1248,6 +1250,51 @@ func TestMakeTaskRunStatusAlpha(t *testing.T) {
...
 		},
+	}, {
+		desc: "test result types mismatch",
+		podStatus: corev1.PodStatus{
+			Phase: corev1.PodSucceeded,
+			ContainerStatuses: []corev1.ContainerStatus{{
+				Name: "step-bar",
+				State: corev1.ContainerState{
+					Terminated: &corev1.ContainerStateTerminated{
+						Message: `[{"key":"resultName","value":"{\"hello\":\"world\"}", "type":1}, {"key":"digest","value":"sha256:1234","resourceRef":{"name":"source-image"}}]`,
```
lbernick, line 1261: what is type: 1 here?

pkg/pod/status_test.go
```diff
@@ -1248,6 +1250,51 @@ func TestMakeTaskRunStatusAlpha(t *testing.T) {
 				CompletionTime: &metav1.Time{Time: time.Now()},
 			},
 		},
+	}, {
+		desc: "test result types mismatch",
```
lbernick, line 1254: are there any test cases for successful matches?

pkg/pod/status.go
Yongxuanzhang: oh okk, I did put them in the reconcile before but the tests seem very hard to add. 😢 . Will try again!

lbernick: Curious what is making adding tests challenging? I think there may be some opportunity here to create a package specifically for results or substitution or something like that

Yongxuanzhang: Did you mean 1) put this code inside the `reconcile` function? Or 2) wrap them as a function and call this function inside the `reconcile`? For 1) the reconcile test code is pretty large, for 2) I think we can test this function separately without testing it from `reconcile`?

lbernick: let's do 2, but the pod package isn't the right place for this IMO

Yongxuanzhang: oh ok, got it!

pkg/pod/status_test.go
Yongxuanzhang: thanks! Just added

pkg/pod/status_test.go
Yongxuanzhang: code is refactored so this is removed

pkg/pod/status.go
Yongxuanzhang: moved them to validate_resources.go

pkg/pod/status.go
Yongxuanzhang: you mean merge them before calling these functions?

Yongxuanzhang: If so, I merged them into one results list

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -283,3 +283,66 @@ func validateSidecarOverrides(ctx context.Context, ts *v1beta1.TaskSpec, trs *v1
 	}
 	return err
 }
+
+// validateResults checks the emitted results type and object properties against the ones defined in spec.
```
ywluogg, line 287: From the discussion, did you mean you need to bypass the type validations here before applying the results? Can you add a TODO if so?

ywluogg: I see. I think the one validation you need to bypass in the validateTaskRunResults() in reconciler taskrun.go. You can close this thread

pkg/pod/status.go
lbernick: no -- I think "resolvedTaskSpec" should be used as the source of truth for what the task spec is

lbernick (commented):
Thanks for moving out of pkg/pod, I think this looks a lot better! I would actually create a new package (maybe pkg/results?) and export a function that validates a task spec against a list of task results. Does that make sense/seem reasonable?

lbernick (approved):

PR #4927, opened 2022-06-01
Refactor the usage of MustCompile and simplify function signatures

chuangw6:
# Changes
Prior to this commit:
1. MustCompile sometimes is used inside a function though it is meant for
initializing global variables holding compiled regular expressions
according to this ([docs](https://pkg.go.dev/regexp#MustCompile)). Using
it in a function could possibly panic.
2. function `extractExpressionFromString` and `extractVariablesFromString`
return too many values i.e. extracted value (slice/string), present (bool)
and errorMessage (string)

In this fix:
1. regex compiled from from MustCompile is changed to be global, or
MustCompile is replaced with Compile so that we can return error rather
than causing potential panic.
2. the two functions signatures are simplified to just return extracted
value and error.

# Submitter Checklist

  functionality, content, code)

# Release Notes

### DIFF
--- a/pkg/apis/pipeline/v1beta1/task_validation.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation.go
@@ -34,9 +34,14 @@ import (
 	"knative.dev/pkg/apis"
 )
 
-var _ apis.Validatable = (*Task)(nil)
-
+// This regex pattern makes sure the variable name format follows the rules
+// - Must only contain alphanumeric characters, hyphens (-), underscores (_), and dots (.)
+// - Must begin with a letter or an underscore (_)
 const variableNameFormat = "^[_a-zA-Z][_a-zA-Z0-9.-]*$"
+
+var paramNameFormatRegex = regexp.MustCompile(variableNameFormat)
+
+var _ apis.Validatable = (*Task)(nil)
 
 // Validate implements apis.Validatable
 func (t *Task) Validate(ctx context.Context) *apis.FieldError {
@@ -451,15 +456,11 @@ func validateStepArrayUsage(step Step, prefix string, vars sets.String) *apis.Fi
 }
 
 func validateVariables(ctx context.Context, steps []Step, prefix string, vars sets.String) (errs *apis.FieldError) {
-	// validate that the variable name format follows the rules
-	// - Must only contain alphanumeric characters, hyphens (-), underscores (_), and dots (.)
-	// - Must begin with a letter or an underscore (_)
-	re := regexp.MustCompile(variableNameFormat)
 	invalidNames := []string{}
 	// Converting to sorted list here rather than just looping map keys
 	// because we want the order of items in vars to be deterministic for purpose of unit testing
 	for _, name := range vars.List() {
-		if !re.MatchString(name) {
+		if !paramNameFormatRegex.MatchString(name) {
 			invalidNames = append(invalidNames, name)
 		}
 	}
--- a/pkg/substitution/substitution.go
+++ b/pkg/substitution/substitution.go
@@ -31,14 +31,20 @@ const braceMatchingRegex = "(\\$(\\(%s(\\.(?P<var1>%s)|\\[\"(?P<var2>%s)\"\\]|\\
 
 // ValidateVariable makes sure all variables in the provided string are known
 func ValidateVariable(name, value, prefix, locationName, path string, vars sets.String) *apis.FieldError {
-	if vs, present, _ := extractVariablesFromString(value, prefix); present {
-		for _, v := range vs {
-			v = strings.TrimSuffix(v, "[*]")
-			if !vars.Has(v) {
-				return &apis.FieldError{
-					Message: fmt.Sprintf("non-existent variable in %q for %s %s", value, locationName, name),
-					Paths:   []string{path + "." + name},
-				}
+	vs, err := extractVariablesFromString(value, prefix)
+	if err != nil {
+		return &apis.FieldError{
+			Message: fmt.Sprintf("fail to extract variables from string: %v", err),
+			Paths:   []string{""},
+		}
+	}
+
+	for _, v := range vs {
+		v = strings.TrimSuffix(v, "[*]")
+		if !vars.Has(v) {
+			return &apis.FieldError{
+				Message: fmt.Sprintf("non-existent variable in %q for %s %s", value, locationName, name),
+				Paths:   []string{path + "." + name},
 			}
 		}
 	}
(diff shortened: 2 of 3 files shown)


pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -34,10 +34,11 @@ import (
 	"knative.dev/pkg/apis"
 )
 
-var _ apis.Validatable = (*Task)(nil)
```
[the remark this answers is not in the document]
chuangw6: This line is not removed. I just move it to line 40 to put two vars declaration together.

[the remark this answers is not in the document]
vdemeester: oh waw, I need glasses 👓 😝 
Thanks @chuangw6 🤗

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -451,15 +452,11 @@ func validateStepArrayUsage(step Step, prefix string, vars sets.String) *apis.Fi
 }
 
 func validateVariables(ctx context.Context, steps []Step, prefix string, vars sets.String) (errs *apis.FieldError) {
-	// validate that the variable name format follows the rules
```
lbernick, line 454: could you keep this comment around, and just move it to where the regex is defined?

pkg/substitution/substitution_test.go
```diff
@@ -174,7 +174,7 @@ func TestValidateVariablePs(t *testing.T) {
 			vars:   sets.NewString("foo.bar.baz"),
 		},
 		expectedError: &apis.FieldError{
-			Message: `Invalid referencing of parameters in --flag=$(params.foo.bar.baz) !!! You can only use the dots inside single or double quotes. eg. $(params["org.foo.blah"]) or $(params['org.foo.blah']) are valid references but NOT $params.org.foo.blah.`,
+			Message: `extractVariablesFromString failed: invalid referencing of parameters in --flag=$(params.foo.bar.baz) !!! You can only use the dots inside single or double quotes. eg. $(params["org.foo.blah"]) or $(params['org.foo.blah']) are valid references but NOT $(params.org.foo.blah)`,
```
lbernick, line 177: I don't think we want the message returned to users to contain the name of the function that failed; it should just explain what is invalid and why.

pkg/substitution/substitution.go
```diff
@@ -87,93 +98,107 @@ func ValidateVariableProhibited(name, value, prefix, locationName, path string,
...
 	}
+
 	return nil
 }
 
 // Extract a the first full string expressions found (e.g "$(input.params.foo)"). Return
-// "" and false if nothing is found.
-func extractExpressionFromString(s, prefix string) (string, bool) {
+// "" and nil if nothing is found.
+func extractExpressionFromString(s, prefix string) (string, error) {
```
lbernick, line 178: I think it would make sense to leave this function signature as is, and move the regexes into global variables instead of using Compile

pkg/apis/pipeline/v1beta1/task_validation.go
chuangw6: sure! sgtm.

pkg/substitution/substitution.go
chuangw6: Since the regex string/pattern needs the input `prefix`, I guess it might not be possible to make it global variables?

pkg/substitution/substitution_test.go
chuangw6: changed. Thanks!

pkg/substitution/substitution.go
lbernick: ah ok makes sense

pritidesai:
> The following is the coverage report on the affected files. Say `/test pull-tekton-pipeline-go-coverage` to re-run this coverage report
> 
> File	Old Coverage	New Coverage	Delta
> [pkg/apis/pipeline/v1beta1/task_validation.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/4927/pull-tekton-pipeline-go-coverage/1532418063528366081/artifacts/line-cov.html#file0)	97.7%	97.7%	-0.0
> [pkg/substitution/substitution.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/4927/pull-tekton-pipeline-go-coverage/1532418063528366081/artifacts/line-cov.html#file1)	57.9%	52.9%	-5.0

@chuangw6 thank you for this 🙏 

I recommend adding more unit test. Impacting 5% coverage during the refactor is not ideal specially with such low coverage to begin with 🔅 .

Please try to make sure we at least maintain the code coverage as is in a refactoring PRs. `substitution`  has been one of the packages with very low coverage 😞.

@vdemeester @lbernick thoughts?

lbernick:
> I recommend adding more unit test. Impacting 5% coverage during the refactor is not ideal specially with such low coverage to begin with 🔅 .
> 
> Please try to make sure we at least maintain the code coverage as is in a refactoring PRs. `substitution` has been one of the packages with very low coverage 😞.

Definitely agree more testing would be nice here for these helpers! Not sure how much to read into the coverage tool output or how much testing of this package is appropriate to tackle for a refactor-- will leave that up to Chuang

pritidesai:
> > I recommend adding more unit test. Impacting 5% coverage during the refactor is not ideal specially with such low coverage to begin with 🔅 .
> > Please try to make sure we at least maintain the code coverage as is in a refactoring PRs. `substitution` has been one of the packages with very low coverage 😞.
> 
> Definitely agree more testing would be nice here for these helpers! Not sure how much to read into the coverage tool output or how much testing of this package is appropriate to tackle for a refactor-- will leave that up to Chuang

Thanks @lbernick!

Please look for the `no coverage` represented by the code in red 🟥 
Screen Shot 2022-06-08 at 10 33 58 AM

Thanks @chuangw6 for your understanding 👍 Appreciate all the simplifications here, the `substitution` package is much more clean now 👏 Please add more coverage for the changes you are introducing.

An alternative could be to introduce the unit tests first in a separate PR to bump the existing coverage.

lbernick:
thanks @pritidesai, sgtm!

lbernick:
@pritidesai in light of our discussion yesterday at the pipelines WG around not testing non-exported functions, what do you think is the best way forward here?

vdemeester:
> @pritidesai in light of our discussion yesterday at the pipelines WG around not testing non-exported functions, what do you think is the best way forward here?

Ideally even if we are not testing non exported functions, we should cover / aim to cover them (through exported ones). If we can’t, this usually mean the unexported function covers more cases that what is exported (and thus is more complex that it needs to be 🤭).

lbernick:
@chuangw6 are you still working on this? would you mind rebasing or closing?

chuangw6:
> @chuangw6 are you still working on this? would you mind rebasing or closing?

Sorry I saw this as non-urgent and switched to other stuff before. Happy to get this cleaned up and move it forward.

PR #4934, opened 2022-06-03
TEP-0090: Add Matrix Package

jerop:
# Changes

[TEP-0090][tep-0090] proposed executing a `PipelineTask` in parallel `TaskRuns` and `Runs` with substitutions from combinations of `Parameters` in a `Matrix`.

In this change, we add the matrix package which generates the combinations of the `Parameters` in a `Matrix`.

The types added are internal, and there are no user-facing changes in this PR.

/kind feature

# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/matrix/matrix.go
+++ b/pkg/matrix/matrix.go
@@ -1,0 +1,27 @@
+/*
+Copyright 2022 The Tekton Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package matrix
+
+import (
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+)
+
+// FanOut produces combinations of Parameters of type String from a slice of Parameters of type Array.
+func FanOut(params []v1beta1.Param) Combinations {
+	var combinations Combinations
+	for _, parameter := range params {
+		combinations = combinations.fanOut(parameter)
+	}
+	return combinations
+}
--- a/pkg/matrix/matrix_types.go
+++ b/pkg/matrix/matrix_types.go
@@ -1,0 +1,82 @@
+/*
+   Copyright 2022 The Tekton Authors
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+       http://www.apache.org/licenses/LICENSE-2.0
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package matrix
+
+import (
+	"strconv"
+
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+)
+
+// Combinations is a slice of combinations of Parameters from a Matrix.
+type Combinations []*Combination
+
+// Combination is a specific combination of Parameters from a Matrix.
+type Combination struct {
+	// MatrixID is an identification of a combination from Parameters in a Matrix.
+	MatrixID string
+
+	// Params is a specific combination of Parameters in a Matrix.
+	Params []v1beta1.Param
+}
+
+func (combinations Combinations) fanOut(param v1beta1.Param) Combinations {
+	if len(combinations) == 0 {
+		return initializeCombinations(param)
+	}
+	return combinations.distribute(param)
+}
+
+func (combinations Combinations) distribute(param v1beta1.Param) Combinations {
+	// when there are existing combinations, this is a non-first parameter in the matrix, and we need to distribute
+	// it among the existing combinations
+	var expandedCombinations Combinations
+	var count int
+	for _, value := range param.Value.ArrayVal {
+		for _, combination := range combinations {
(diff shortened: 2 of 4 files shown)


jerop:
cc @pritidesai

pkg/matrix/matrix_types.go
```diff
@@ -0,0 +1,45 @@
...
+
+package matrix
+
+import "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+
+type Combinations []*Combination
+
+type Combination struct {
+	// Identification of a combination from Parameters in a Matrix.
+	MatrixId string `json:"matrixId"`
```
lbernick, line 22: no need for JSON tags I think

lbernick (approved):
could you make it more clear in the commit message/PR description that these types are internal and won't be added to the API? i.e. no user facing changes

pkg/matrix/matrix.go
```diff
@@ -0,0 +1,56 @@
...
+	var combinations []*Combination
+	for _, parameter := range params {
+		combinations = distributeParameter(parameter, combinations)
+	}
+	return combinations
+}
+
+func distributeParameter(param v1beta1.Param, existingCombinations []*Combination) []*Combination {
+	var expandedCombinations []*Combination
+	if len(existingCombinations) == 0 {
```
abayer, line 32: It might be handy to have a comment explaining why the behavior is different depending on whether there already are existing combinations.

jerop: @abayer added comments explaining the logic

abayer: Thanks!

pkg/matrix/matrix.go
```diff
@@ -0,0 +1,62 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package matrix
+
+import (
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"strconv"
```
pritidesai, line 18: linting might fail here 🙃

jerop: locally, I get a linting failure when I flip the order of imports - let's see what happens here when the tests run 🤔

jerop:
`2022/06/03 16:10:27 main.go:319: Something went wrong: failed to prepare test environment: --provider=gke boskos failed to acquire project: resources not found` 😕 

pkg/matrix/matrix.go
jerop: it [failed](https://tekton-releases.appspot.com/build/tekton-prow/pr-logs/pull/tektoncd_pipeline/4934/pull-tekton-pipeline-build-tests/1532764258557759488/) linting 🙃 

the imports settings in my ide was using gofmt instead of goimports, fixed and updated the imports, thanks @pritidesai!

pkg/matrix/matrix.go
```diff
@@ -0,0 +1,63 @@
...
+package matrix
+
+import (
+	"strconv"
+
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+)
+
+// FanOut produces combinations of Parameters of type String from a slice of Parameters of type Array.
+func FanOut(params []v1beta1.Param) Combinations {
```
pritidesai, line 23: just a styling comment here, it will be great to see this as a member function of Combination, just like you have other member function at https://github.com/tektoncd/pipeline/pull/4934/files#diff-8ca8f243041fb0ce01d5916da5dcde57488b8a94c7aa294eee17a5b3b9cc5eeaR30

In this case, we are initializing combinations compared to reading from it.

pkg/matrix/matrix_types.go
```diff
@@ -0,0 +1,82 @@
...
+	// MatrixID is an identification of a combination from Parameters in a Matrix.
+	MatrixID string
+
+	// Params is a specific combination of Parameters in a Matrix.
+	Params []v1beta1.Param
+}
+
+func (combinations Combinations) fanOut(param v1beta1.Param) Combinations {
+	if len(combinations) == 0 {
+		return initializeCombinations(param)
```
pritidesai, line 36: much cleaner now 🎉 thank you 🤗

pkg/matrix/matrix.go
jerop: good idea, moved them to member functions 👍🏾

pkg/matrix/matrix_types.go
jerop: agreed, thank YOU!

pkg/matrix/matrix_types.go
```diff
@@ -0,0 +1,82 @@
...
+		MatrixID: strconv.Itoa(i),
+		Params: append(parameters, v1beta1.Param{
+			Name:  name,
+			Value: v1beta1.ArrayOrString{Type: v1beta1.ParamTypeString, StringVal: value},
+		}),
+	}
+}
+
+// ToMap converts a list of Combinations to a map where the key is the matrixId and the values are Parameters.
+func (combinations Combinations) ToMap() map[string][]v1beta1.Param {
```
pritidesai, line 76: Can we define a `Combinations` as `map[string][]v1beta1.Param` instead of later converting it to map? I am trying to understand this conversion and how it will be used. 

_Please note that not asking to change in this PR - just wondering if we can simplify_

jerop: @pritidesai - We had discussions about what information to surface in the pipelinerun about the combination used for a given taskrun created from a matrix. We [landed](https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#combinations-of-parameters-in-the-matrix) on providing the matrix only initially in line with providing the "minimal  information needed" as proposed in TEP-0100. Then, if we find that users need the mapping of matrixid to parameters, we can provide the complete information. The Combinations type with MatrixID and Parameters sets us up for this possible extension to align with the Kubernetes API convention of [subobjects over maps](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps). If we find that users don't need the complete information, we can switch to maps only given that it's only an implementation detail now.

PR #4942, opened 2022-06-06
Remove deprecated Conditions CRD/functionality

abayer:
# Changes

Closes #3377

This was deprecated in v0.16.0, and is scheduled to be fully removed in v0.37.0, releasing late in June.

/kind cleanup

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been filled in
(if there are no user facing changes, use release note "NONE")

# Release Notes

```release-note
action required: Deprecated `conditions` in pipelines removed. Existing pipelines using `conditions` will need to be updated.
```

PR #4943, opened 2022-06-07
TEP-0090: Split up ResolvePipelineRunTask 

jerop:
# Changes

Prior to this commit, the logic for resolving `TaskRun` and `Resources` needed for a given `PipelineTask` was included within the `ResolvePipelineRunTask` function.

In this change, we split the logic into their own functions for reuse and readability. We also fix the names of `PipelineTask`
variables to distinguish them from `Tasks`.

No functional changes made in this commit.

/kind feature

# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -438,7 +438,7 @@ func isCustomTask(ctx context.Context, rprt ResolvedPipelineRunTask) bool {
 
 // ResolvePipelineRunTask retrieves a single Task's instance using the getTask to fetch
 // the spec. If it is unable to retrieve an instance of a referenced Task, it  will return
-// an error, otherwise it returns a list of all of the Tasks retrieved.  It will retrieve
+// an error, otherwise it returns a list of all the Tasks retrieved.  It will retrieve
 // the Resources needed for the TaskRun using the mapping of providedResources.
 func ResolvePipelineRunTask(
 	ctx context.Context,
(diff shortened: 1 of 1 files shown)


pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -472,91 +472,124 @@ func ResolvePipelineRunTask(
 	getTaskRun resources.GetTaskRun,
 	getRun GetRun,
 	getCondition GetCondition,
-	task v1beta1.PipelineTask,
+	pipelineTask v1beta1.PipelineTask,
```
pritidesai, line 475: so much cleaner now 🎉 thank you @jerop for renaming!

lbernick (approved):

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -472,91 +472,124 @@ func ResolvePipelineRunTask(
...
 			default:
 				spec = t.TaskSpec()
 				taskName = t.TaskMetadata().Name
 			}
 		}
-		kind = task.TaskRef.Kind
-	} else {
-		spec = task.TaskSpec.TaskSpec
+		kind = pipelineTask.TaskRef.Kind
+	} else if pipelineTask.TaskSpec != nil {
```
pritidesai, line 591: This extra check will be much clear with a unit test. From what I recall, we have `webhook` validation to at least have either `taskRef` or `taskSpec` initialized. So, if `taskRef` is not initialized, its safe to rely on `taskSpec` without checking it being not-nil. At this point, we can avoid introducing such condition check specially with lack of a unit test.

jerop: good point, removed that extra check 👍🏾

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -472,91 +472,124 @@ func ResolvePipelineRunTask(
...
-		taskRun, err := getTaskRun(rprt.TaskRunName)
-		if err != nil {
-			if !kerrors.IsNotFound(err) {
-				return nil, fmt.Errorf("error retrieving TaskRun %s: %w", rprt.TaskRunName, err)
-			}
-		}
-		if taskRun != nil {
-			rprt.TaskRun = taskRun
+		rprt.TaskRunName = GetTaskRunName(pipelineRun.Status.TaskRuns, pipelineRun.Status.ChildReferences, pipelineTask.Name, pipelineRun.Name)
+		if err := resolvePipelineRunTaskWithTaskRun(ctx, &rprt, rprt.TaskRunName, pipelineRun, getTask, getTaskRun, getCondition, pipelineTask, providedResources); err != nil {
```
pritidesai, line 491: is it possible to retrieve `taskRunName` from `rprt`? If so, we can avoid including `taskRunName` from the list of parameters?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -472,91 +472,124 @@ func ResolvePipelineRunTask(
...
+	pipelineTask v1beta1.PipelineTask,
+	providedResources map[string]*resourcev1alpha1.PipelineResource,
+) error {
+	taskRun, err := getTaskRun(taskRunName)
+	if err != nil {
+		if !kerrors.IsNotFound(err) {
+			return fmt.Errorf("error retrieving TaskRun %s: %w", taskRunName, err)
 		}
+	}
+	rprt.TaskRun = taskRun
```
pritidesai, line 515: if condition is missing here:

```
if taskRun != nil {
```

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
jerop: discussed on slack, but adding comment here for future reference:

yes, but we need `taskRunName` to be its own field separate from the rest of the `rprt` is that it sets us up to support matrixed `rprt` which will have list of `taskRunNames` to iterate over

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
jerop: I believe that the check was unnecessary because `rprt.TaskRun` is initialized as nil so if `taskRun` is nil, there's no need to check for that condition as it makes no difference - however, will add back the check just in case

pritidesai:
/lgtm

thank you @jerop 🎉 

/test pull-tekton-pipeline-go-coverage

I am sorry to bring this up again but please add unit tests while introducing any new function 🙏

jerop:
> I am sorry to bring this up again but please add unit tests while introducing any new function 🙏

@pritidesai we have guidelines - https://github.com/tektoncd/community/blob/main/standards.md#tests - to write tests for exported functions only...there was no exported function added, all new functions are unexported functions that are used in the tested exported function (plus there is no functionality change)...or do we need to update that guideline?

pritidesai:
> https://github.com/tektoncd/community/blob/main/standards.md#tests

I think we should, what do you think?

We (not intended to point on anyone) are adding a bunch of unexported functions which is great but its becoming very hard to read and understand the need of so many unexported functions. How are we sure that any changes we introduce are not breaking any functionality. This can only be made sure if we have 100% code coverage which is highly unlikely to maintain in any project.

Also, our standards today says the code coverage must remain same (at least). But thats not something enforced by our CI. One such example - https://github.com/tektoncd/pipeline/pull/4927#issuecomment-1148190842 - Impacting code coverage by 5% while refactoring will burn us down in the future.

jerop:
> > https://github.com/tektoncd/community/blob/main/standards.md#tests
> 
> I think we should, what do you think?
> 
> We (not intended to point on anyone) are adding a bunch of unexported functions which is great but its becoming very hard to read and understand the need of so many unexported functions. How are we sure that any changes we introduce are not breaking any functionality. This can only be made sure if we have 100% code coverage which is highly unlikely to maintain in any project.

that's one option, we could also try out the suggestion in the guidelines to move those functions into packages: "If you find yourself wanting to test an unexported function, consider whether it would make sense to move the test into another package and export it" 

added this topic to the agenda for the next Pipelines working group meeting on 06/14 ([notes](https://docs.google.com/document/d/1AfJfdyd1JN2P4haBdYOxEn6SMENMgvQF9ps7iF2QSG0/edit?resourcekey=0-lm66bE68HA_WVLA6AaCh6w#bookmark=id.3tewyl26ymsi)) so we can discuss with other maintainers

> 
> Also, our standards today says the code coverage must remain same (at least). But thats not something enforced by our CI. One such example - [#4927 (comment)](https://github.com/tektoncd/pipeline/pull/4927#issuecomment-1148190842) - Impacting code coverage by 5% while refactoring will burn us down in the future.

it'd be great if the CI enforced the code coverage standard, maybe for now we can add it to the submitter checklist to make it more explicit? we can bring it up in the working group meeting

jerop:
@abayer @pritidesai - rebased the PR to fix merge conflicts after https://github.com/tektoncd/pipeline/pull/4942 got merged - please take a look

pritidesai:
> that's one option, we could also try out the suggestion in the guidelines to move those functions into packages: "If you find yourself wanting to test an unexported function, consider whether it would make sense to move the test into another package and export it"

This suggestion can not be generalized, specially for the many complex packages we have in Pipelines. Unfortunately, it just resulting in a smelly code.

`runNextSchedulableTask`, `processRunTimeouts`, `updateTaskRunsStatusDirectly`, etc were all  introduced on a similar basis and have grown complex over time without any unit testing. But they do follow the best practices - no unit test for unexported functions. 

https://github.com/tektoncd/pipeline/blob/566e2e3e20c0c6348e86c248310fcc9dbfe85166/pkg/reconciler/pipelinerun/pipelinerun.go#L644

https://github.com/tektoncd/pipeline/blob/566e2e3e20c0c6348e86c248310fcc9dbfe85166/pkg/reconciler/pipelinerun/pipelinerun.go#L616

https://github.com/tektoncd/pipeline/blob/566e2e3e20c0c6348e86c248310fcc9dbfe85166/pkg/reconciler/pipelinerun/pipelinerun.go#L718

I am not pointing at you @jerop, thanks for your understanding ❤️ Also, thank you for adding it to the WG for further discussion.

Didn't expect this to be merged as is, we have many functions with similar (complex) signatures now. At the same time, excited to see this result in much more meaningful and reusable code with matrix 🎉 

https://github.com/tektoncd/pipeline/blob/566e2e3e20c0c6348e86c248310fcc9dbfe85166/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L448-L564

PR #4945, opened 2022-06-07
TEP-0090: Indicate Resolved PipelineRunTask is Matrixed

jerop:
# Changes

In this change, we add support for resolving a `PipelineTask` to indicate that it is matrixed. This is useful for processing the ResolvedPipelineRunTask.

# Submitter Checklist

- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md)(including functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -93,6 +93,11 @@ func (t ResolvedPipelineRunTask) IsCustomTask() bool {
 	return t.CustomTask
 }
 
+// IsMatrixed return true if the PipelineTask has a Matrix.
+func (t ResolvedPipelineRunTask) IsMatrixed() bool {
+	return len(t.PipelineTask.Matrix) > 0
+}
+
 // IsSuccessful returns true only if the run has completed successfully
 func (t ResolvedPipelineRunTask) IsSuccessful() bool {
 	if t.IsCustomTask() {
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
@@ -2705,3 +2705,82 @@ func TestGetRunName(t *testing.T) {
 		})
 	}
 }
+
+func TestIsMatrixed(t *testing.T) {
+	pr := v1beta1.PipelineRun{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "pipelinerun",
+		},
+	}
+	getTask := func(ctx context.Context, name string) (v1beta1.TaskObject, error) { return task, nil }
+	getTaskRun := func(name string) (*v1beta1.TaskRun, error) { return &trs[0], nil }
+	getRun := func(name string) (*v1alpha1.Run, error) { return &runs[0], nil }
+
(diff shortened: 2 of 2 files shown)


pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -461,6 +468,10 @@ func isCustomTask(ctx context.Context, rprt ResolvedPipelineRunTask) bool {
 	return cfg.FeatureFlags.EnableCustomTasks && !invalidSpec && (isTaskRefCustomTask || isTaskSpecCustomTask)
 }
 
+func isMatrixed(rprt ResolvedPipelineRunTask) bool {
```
imjasonh, line 471: I might suggest not making this a separate method, since it can be confusing to see an `isMatrixed(rprt)` method and a `rprt.IsMatrixed()` method, and have them not be the same thing.

It took me a minute to realize `isMatrixed` here is only used to _populate_ `rprt.Matrixed` below, which is read out by `rprt.IsMatrixed()`.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -3554,3 +3554,83 @@ func TestGetRunName(t *testing.T) {
...
 }
+
+func TestIsMatrixed(t *testing.T) {
+	pr := v1beta1.PipelineRun{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "pipelinerun",
+		},
+	}
+	getTask := func(ctx context.Context, name string) (v1beta1.TaskObject, error) { return task, nil }
+	getTaskRun := func(name string) (*v1beta1.TaskRun, error) { return nil, nil }
```
imjasonh, line 3565: Are these methods executed during the test? If they're not, maybe you could pass `nil` so it panics if they're unexpectedly called, or `return nil, errors.New("unimplemented")` so they fail a bit more gracefully.

Having `ResolvePipelineRunTask` take a bunch of methods has always felt like a pretty bad smell, and part of that is making these tests include all this copied boilerplate.

imjasonh (commented):
Just some readability comments, otherwise LGTM

jerop:
@abayer @imjasonh thank you for the reviews, the PR is ready for another look

lbernick (approved):

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -79,6 +79,8 @@ type ResolvedPipelineRunTask struct {
 	ResolvedTaskResources *resources.ResolvedTaskResources
 	// ConditionChecks ~~TaskRuns but for evaling conditions
 	ResolvedConditionChecks TaskConditionCheckState // Could also be a TaskRun or maybe just a Pod?
+
+	Matrixed bool
```
pritidesai, line 83: Do we need to introduce this new field? Can it be implied from the length of `taskRunName` that its a matrix?

jerop: removed the field, will revisit if needed when we add the list of taskRunNames

pritidesai (approved):

PR #4947, opened 2022-06-08
TEP-0090: Matrix - Concurrency Control

jerop:
# Changes

In this change, we implement concurrency controls for the `Matrix` as discussed in [TEP-0090: Matrix - Concurrency Control][cc]. The default value is 256, and users can configure the value for their own installations.

[cc]: https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#concurrency-control

/kind feature

# Submitter Checklist

# Release Notes

```release-note
The default maximum count of `TaskRuns` or `Runs` from a given `Matrix` is 256. Users can configure this value for their installations.
```

### DIFF
--- a/config/config-defaults.yaml
+++ b/config/config-defaults.yaml
@@ -74,3 +74,7 @@ data:
     # but that a TaskRun does not explicitly provide.
     # default-task-run-workspace-binding: |
     #   emptyDir: {}
+
+    # default-max-matrix-combinations-count contains the default maximum number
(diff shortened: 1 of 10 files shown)


lbernick (commented):
just a naming nitpick-- I wonder if it would make sense to call this "maximum-matrix-combinations-count" (without the default) since there's no way to change the max matrix combinations other than this flag as far as I can tell? LGTM other than that

PR #4951, opened 2022-06-09
TEP-0090: Add `TaskRuns` to `ResolvedPipelineRunTask` and implement `isFailure`

jerop:
# Changes

In this change, we add `TaskRuns` field to `ResolvedPipelineRunTask` alongside the `TaskRun` field.
- When a `PipelineTask` does not have a `Matrix`, the `TaskRun` field is populated (as is already being done, no change here).
- When a `PipelineTask` has a `Matrix`, `TaskRuns` will be populated to track all `TaskRuns` from a given matrixed `ResolvedPipelineRunTask`.

When we promote `Matrix` to Beta, we could remove `TaskRun` field and  migrate to using `TaskRuns` field only where it would have one `TaskRun` only when the `PipelineTask` does not have a `Matrix`. However, the current separation is helpful to isolate the Matrix feature in  alpha.

In this change, we also implement the `isFailure` and other member functions of `ResolvedPipelineRunTask` that `isFailure` uses: `isCancelled` and `hasRemainingRetries`. The failure strategy in `Matrix` is to fail fast. We can explore other failure strategies, if needed, before promoting to Beta.

/kind feature

# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -60,6 +60,7 @@ func (e *TaskNotFoundError) Error() string {
 type ResolvedPipelineRunTask struct {
 	TaskRunName string
 	TaskRun     *v1beta1.TaskRun
+	TaskRuns    []*v1beta1.TaskRun
 	// If the PipelineTask is a Custom Task, RunName and Run will be set.
 	CustomTask            bool
 	RunName               string
@@ -115,13 +116,28 @@ func (t ResolvedPipelineRunTask) isFailure() bool {
 	}
 	var c *apis.Condition
 	var isDone bool
-	if t.IsCustomTask() {
+
+	switch {
+	case t.IsCustomTask():
 		if t.Run == nil {
 			return false
 		}
 		c = t.Run.Status.GetCondition(apis.ConditionSucceeded)
 		isDone = t.Run.IsDone()
-	} else {
+	case t.IsMatrixed():
+		if len(t.TaskRuns) == 0 {
+			return false
+		}
+		// is failed on the first failure with no remaining retries to fail fast
+		for _, taskRun := range t.TaskRuns {
+			c = taskRun.Status.GetCondition(apis.ConditionSucceeded)
+			isDone = taskRun.IsDone()
+			if isDone && c.IsFalse() && !t.hasRemainingRetries() {
+				return true
+			}
+		}
+		return false
+	default:
 		if t.TaskRun == nil {
 			return false
 		}
@@ -135,12 +151,25 @@ func (t ResolvedPipelineRunTask) isFailure() bool {
 // is less than the number of retries allowed.
 func (t ResolvedPipelineRunTask) hasRemainingRetries() bool {
 	var retriesDone int
-	if t.IsCustomTask() {
+	switch {
+	case t.IsCustomTask():
 		if t.Run == nil {
 			return true
 		}
 		retriesDone = len(t.Run.Status.RetriesStatus)
-	} else {
+	case t.IsMatrixed():
+		if len(t.TaskRuns) == 0 {
+			return true
+		}
+		// has remaining retries when any TaskRun has a remaining retry
+		for _, taskRun := range t.TaskRuns {
+			retriesDone = len(taskRun.Status.RetriesStatus)
+			if retriesDone < t.PipelineTask.Retries {
+				return true
+			}
+		}
+		return false
+	default:
 		if t.TaskRun == nil {
 			return true
 		}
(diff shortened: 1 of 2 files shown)


pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -115,13 +116,28 @@ func (t ResolvedPipelineRunTask) isFailure() bool {
...
+
+	switch {
+	case t.IsCustomTask():
 		if t.Run == nil {
 			return false
 		}
 		c = t.Run.Status.GetCondition(apis.ConditionSucceeded)
 		isDone = t.Run.IsDone()
-	} else {
+	case t.IsMatrixed():
```
lbernick, line 127: how does this work if a pipeline task is both a custom task and matrixed? Is that out of scope for this PR?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -115,13 +116,28 @@ func (t ResolvedPipelineRunTask) isFailure() bool {
...
+		if len(t.TaskRuns) == 0 {
+			return false
+		}
+		// is failed on the first failure with no remaining retries to fail fast
```
lbernick, line 131: could you please update the function docstring to state that a matrixed pipelinetask is considered failed if any of the pipelineTasks have failed? The comment here is also a bit confusing, could you reword?

Lastly, one of the non goals for this TEP is "Terminating early when one of the TaskRuns or Runs created in parallel fails". It seems like if one of the matrixed Tasks fails, the whole PipelineTask will be considered failed in this implementation (fail fast as you said). I'm concerned that other parts of the codebase might be using this function to decide (for example) when a finally task can be scheduled; is it a concern if the matrixed TaskRuns that haven't failed continue to run? Or will there be some sort of cancellation?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
jerop: yes, that is out of scope here in this PR because it'd be huge otherwise - will add that case when I implement support for matrixed Runs

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
jerop: > could you please update the function docstring to state that a matrixed pipelinetask is considered failed if any of the pipelineTasks have failed? The comment here is also a bit confusing, could you reword?

good catch, will update the docstring and comments 

> Lastly, one of the non goals for this TEP is "Terminating early when one of the TaskRuns or Runs created in parallel fails". It seems like if one of the matrixed Tasks fails, the whole PipelineTask will be considered failed in this implementation (fail fast as you said). I'm concerned that other parts of the codebase might be using this function to decide (for example) when a finally task can be scheduled; is it a concern if the matrixed TaskRuns that haven't failed continue to run? Or will there be some sort of cancellation?

that's right, was planning to add cancel the other taskruns from the matrix upon the first failure - it seems that we have to address the failure strategy in TEP-0090 so will update the TEP shortly with the "fail fast" detail

[matrix in github actions](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures) supports failing fast as the default and there are other strategies - will be proposing that we fail fast then explore other failure strategies later if needed

jerop:
/hold 

need to update the [TEP-0090: Matrix][tep-0090] with fail-fast failure stategy

abayer:
/lgtm

I don’t _love_ this - it’s fine for just supporting matrixed `TaskRun`s, but it feels like it’ll get messy when we add support for matrixed custom tasks (and eventually, matrixed `PipelineRun`s). But given that matrixed custom tasks aren’t in scope for this PR, and matrixed `PipelineRun`s aren’t coming in the near future, this will do.

jerop:
opened https://github.com/tektoncd/community/pull/724 to discuss the failure strategies, will unhold this PR so that we can make progress - matrix is not yet fully functional so we can update this if we land on a different approach other than fail-fast

/hold cancel

pritidesai:
> /lgtm
> 
> I don’t _love_ this - it’s fine for just supporting matrixed `TaskRun`s, but it feels like it’ll get messy when we add support for matrixed custom tasks (and eventually, matrixed `PipelineRun`s). But given that matrixed custom tasks aren’t in scope for this PR, and matrixed `PipelineRun`s aren’t coming in the near future, this will do.

thank you @abayer 👍 

Runs are coming in the next milestone - https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#milestone-2-execute-runs

I haven't gotten chance to review this PR but please feel free to suggest better approach, best to follow @vdemeester's reference quote - _No is temporary, yes is permanent_.

jerop:
> It doesn't look like my comments about the function docstrings were resolved?

@lbernick my bad, forgot to push that update to docstrings, here you go: https://github.com/tektoncd/pipeline/pull/4970

PR #4958, opened 2022-06-09
TEP-0090: Get Names of `TaskRuns`

jerop:
# Changes

In this change, we add a function to get `TaskRuns` names from `TaskRuns` Status or Child References. If none are available, it generates new `TaskRun` names based on the `PipelineRun` and `PipelineTask` names.

/kind feature

# Submitter Checklist

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 
 	"github.com/tektoncd/pipeline/pkg/apis/config"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	resourcev1alpha1 "github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1"
@@ -578,6 +579,47 @@ func GetTaskRunName(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus,
 	}
 
 	return kmeta.ChildName(prName, fmt.Sprintf("-%s", ptName))
+}
+
+// GetNamesofTaskRuns should return unique names for `TaskRuns` if one has not already been defined, and the existing one otherwise.
+func GetNamesofTaskRuns(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus, childRefs []v1beta1.ChildStatusReference, ptName, prName string, combinationCount int) []string {
+	var taskRunNames []string
+	if taskRunNames = getTaskRunNamesFromChildRefs(childRefs, ptName); taskRunNames != nil {
+		return taskRunNames
+	}
+	if taskRunNames = getTaskRunNamesFromTaskRunsStatus(taskRunsStatus, ptName); taskRunNames != nil {
+		return taskRunNames
+	}
+	return getNewTaskRunNames(ptName, prName, combinationCount)
+}
+
+func getTaskRunNamesFromChildRefs(childRefs []v1beta1.ChildStatusReference, ptName string) []string {
+	var taskRunNames []string
+	for _, cr := range childRefs {
+		if cr.Kind == pipeline.TaskRunControllerName && cr.PipelineTaskName == ptName {
+			taskRunNames = append(taskRunNames, cr.Name)
+		}
+	}
+	return taskRunNames
+}
+
+func getTaskRunNamesFromTaskRunsStatus(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus, ptName string) []string {
+	var taskRunNames []string
+	for k, v := range taskRunsStatus {
+		if v.PipelineTaskName == ptName {
+			taskRunNames = append(taskRunNames, k)
+		}
+	}
+	return taskRunNames
+}
+
+func getNewTaskRunNames(ptName, prName string, combinationCount int) []string {
+	var taskRunNames []string
+	for i := 0; i < combinationCount; i++ {
+		taskRunName := kmeta.ChildName(prName, fmt.Sprintf("-%s-%d", ptName, i))
+		taskRunNames = append(taskRunNames, taskRunName)
+	}
+	return taskRunNames
 }
 
 // getRunName should return a unique name for a `Run` if one has not already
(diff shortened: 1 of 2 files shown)


pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -580,6 +580,52 @@ func GetTaskRunName(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus,
 	return kmeta.ChildName(prName, fmt.Sprintf("-%s", ptName))
 }
 
+// GetTaskRunsNames should return unique names for `TaskRuns` if one has not already been defined, and the existing one otherwise.
+func GetTaskRunsNames(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus, childRefs []v1beta1.ChildStatusReference, ptName, prName string, combinationCount int) []string {
```
pritidesai, line 584: NIT: the function name sounds a little weird - `GetTaskRunsNames` v/s `GetTaskRunNames` - get the list of the names of `taskruns`. How about `GetNamesofTaskRuns`?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -580,6 +580,52 @@ func GetTaskRunName(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus,
...
+		return taskRunNames
+	}
+
+	return getNewTaskRunNames(ptName, prName, combinationCount)
+}
+
+func getTaskRunNamesFromChildRefs(childRefs []v1beta1.ChildStatusReference, ptName string) []string {
+	var taskRunNames []string
+	for _, cr := range childRefs {
+		if cr.Kind == "TaskRun" && cr.PipelineTaskName == ptName {
```
pritidesai, line 603: NIT: replace "TaskRun" with const

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -580,6 +581,52 @@ func GetTaskRunName(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus,
 	return kmeta.ChildName(prName, fmt.Sprintf("-%s", ptName))
 }
```
pritidesai, line 583: thanks @jerop for this, I am trying to understand how this existing function will be changed (if at all) after we introduce these new functions. 🙏

jerop: @pritidesai the existing function will not be changed

see related PR: https://github.com/tektoncd/pipeline/pull/4951

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -580,6 +581,52 @@ func GetTaskRunName(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus,
...
+		}
+	}
+	return taskRunNames
+}
+
+func getNewTaskRunNames(ptName, prName string, combinationCount int) []string {
+	taskRunName := kmeta.ChildName(prName, fmt.Sprintf("-%s", ptName))
+	var taskRunNames []string
+	for i := 0; i < combinationCount; i++ {
+		taskRunNames = append(taskRunNames, fmt.Sprintf("%s-%d", taskRunName, i))
```
pritidesai, line 625: This increases the length of the name by at least two more characters (`-` and matrix id). The matrix id could be in 1000...s (hopefully not). But at the least (for 10 count), this results in the name with 65 characters instead of 63 (the current limit).

Can we include the combination count while calling `ChildName`:

```
kmeta.ChildName(prName, fmt.Sprintf("-%s-%d", ptName,i))
```

This will result in a separate suffix all the time. 

The other alternative is to re-write `ChildName` for our purpose which is fine as well.

jerop: @pritidesai - good catch, this is fixed now - and it's okay to have a different suffix, because we have the matrixId

https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#combinations-of-parameters-in-the-matrix

pritidesai: that's a good point @jerop Is this `taskRun` part of minimal pipelineRun status? I am guessing its not which is fine since `taskRun` has `pipelineTask` name. So, the users can classify `taskRun`s belonging to one single `pipelineTask` based on the owner reference.

I am not sure how much matrixId is valuable here in the `taskRun` status since it's a number determined by the tekton controller which cannot be directly mapped to matrix params or a combination of matrix params.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -580,6 +581,52 @@ func GetTaskRunName(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus,
 	return kmeta.ChildName(prName, fmt.Sprintf("-%s", ptName))
 }
 
+// GetNamesofTaskRuns should return unique names for `TaskRuns` if one has not already been defined, and the existing one otherwise.
+func GetNamesofTaskRuns(taskRunsStatus map[string]*v1beta1.PipelineRunTaskRunStatus, childRefs []v1beta1.ChildStatusReference, ptName, prName string, combinationCount int) []string {
+	var taskRunNames []string
+
+	taskRunNames = getTaskRunNamesFromChildRefs(childRefs, ptName)
+	if taskRunNames != nil {
```
pritidesai, line 589: NIT: 

```
if taskRunNames = getTaskRunNamesFromChildRefs(childRefs, ptName);  taskRunNames != nil {
```

pritidesai (approved):
thank you @jerop 🎉

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
jerop: @pierretasci supporting results from fanned out tasks is out of scope for TEP-0090 - https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#results-from-fanned-out-pipelinetasks

results are referenced using the `PipelineTask` names - but in matrixed cases, there are multiple `TaskRuns` - so the matrixId could be useful in gathering the results from matrixed pipelinetasks, such as if we use an object result 

open to updating the TEP to remove that part so that the complete problem will be addressed in the TEP focusing on supporting results from matrixed pipelinetasks

jerop:
@abayer @dibyom @lbernick - please take a look

PR #4971, opened 2022-06-13
Cut PipelineResources from some examples

lbernick:
This commit replaces PipelineResources with workspaces in documentation examples
not related to those features. PipelineResources are deprecated, so our documentation shouldn't be
using them as examples unless it is necessary.

It also cuts the example related to reusing a Task with different
PipelineResources (as the recommended way to do this would be with params).

Lastly, it encourages users to reference the community catalog, rather
than examples tests, for examples of Tasks, because many examples tests
either make use of PipelineResources or aren't realistic.

/kind documentation

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -83,7 +83,7 @@ A `Pipeline` definition supports the following fields:
       - [`tasks`](#adding-tasks-to-the-pipeline) - Specifies the `Tasks` that comprise the `Pipeline`
         and the details of their execution.
 - Optional:
-  - [`resources`](#specifying-resources) - **alpha only** Specifies
+  - [`resources`](#specifying-resources) - **deprecated** Specifies
     [`PipelineResources`](resources.md) needed or created by the `Tasks` comprising the `Pipeline`.
   - [`params`](#specifying-parameters) - Specifies the `Parameters` that the `Pipeline` requires.
   - [`workspaces`](#specifying-workspaces) - Specifies a set of Workspaces that the `Pipeline` requires.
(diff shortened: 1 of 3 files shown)


docs/pipelines.md
```diff
@@ -83,7 +83,7 @@ A `Pipeline` definition supports the following fields:
       - [`tasks`](#adding-tasks-to-the-pipeline) - Specifies the `Tasks` that comprise the `Pipeline`
         and the details of their execution.
 - Optional:
-  - [`resources`](#specifying-resources) - **alpha only** Specifies
+  - [`resources`](#specifying-resources) - **deprecated** Specifies
```
afrittoli, line 86: 🎉

docs/tasks.md
```diff
@@ -992,112 +992,45 @@ The following example `Task` builds and pushes a `Dockerfile`-built image.
...
-
-  # As an implementation detail, this Task mounts the host's daemon socket.
-  volumes:
-    - name: docker-socket
-      hostPath:
-        path: /var/run/docker.sock
-        type: Socket
-```
-
-#### Mounting multiple `Volumes`
```
afrittoli, line 1046: As long as the feature is still there perhaps we shouldn't fully remove this docs, but alter them to say that while this is possible, it is not recommended unless in use cases where no alternative exists.

afrittoli:
Thanks for this @lbernick.
Perhaps we could split the `PipelineResources` part and the volume one, so we could merge the first part right away.
If you prefer to wait for the volume part to be ready that is also ok :)

lbernick:
> Perhaps we could split the `PipelineResources` part and the volume one, so we could merge the first part right away. If you prefer to wait for the volume part to be ready that is also ok :)

Thanks @afrittoli and @pritidesai, I think I'll make this PR just PipelineResources for now

lbernick:
Ok, should be ready for re-review

dibyom (commented):
/lgtm

vdemeester (approved):

PR #4989, opened 2022-06-16
LimitRange calculation should only split Requests for Step Containers

### LINKED ISSUE #4978: LimitRange Memory and CPU Request Distribution does not handle sidecars correctly
The current controller logic for managing "request" resources when LimitRanges are present takes the "default" resource value and divides it by the number of containers in a Pod. This is close, but not quite correct as "sidecars" are run in parallel so should not be part of this calculation 

For...
```
apiVersion: "v1"
kind: "LimitRange"
metadata:
  name: "my-resource-limits" 
spec:
  limits:
    - type: "Container"
      defaultRequest:
        memory: "2Gi"
        cpu: "100m"
---
kind: TaskRun
metadata:
  name: memory-boom-taskrun
spec:
  taskSpec:
    steps:
    - name: memory-boom1
      image: polinux/stress
      script: |
        stress --vm 1 --vm-bytes 20M --vm-hang 10
    - name: memory-boom2
      image: polinux/stress
      script: |
        stress --vm 1 --vm-bytes 20M --vm-hang 10
    sidecars:
    - name: memory-boom-sc
      image: polinux/stress
      command: ["stress"]
      args: ["--vm", "1", "--vm-bytes", "10M", "--vm-hang", "10"]
```
# Expected Behavior
```
Containers:
  step-memory-boom1:
    Requests:
      cpu:                50m
      memory:             1Gi
  step-memory-boom2:
    Requests:
      cpu:                50m
      memory:             1Gi
  sidecar-memory-boom-sc:
    Requests:
      cpu:                100m
      memory:             2Gi
```

# Actual Behavior
```
Containers:
  step-memory-boom1:
    Requests:
      cpu:                33m
      memory:             715827882
  step-memory-boom2:
    Requests:
      cpu:                33m
      memory:             715827882
  sidecar-memory-boom-sc:
    Requests:
      cpu:                33m
      memory:             715827882
```

skaegi:
# Changes
This commit changes  how resource requirements are calculated when one or more LimitRanges are present in the namespace where a TaskRun is.

We now will only transform "step" containers and exclude all other app containers (e.g. sidecars and any other containers) when splitting "requests" from LimitRanges. This behavior is "nearly" consistent with what the current docs say as the split request limits are referred to only with respect to Step containers but they still needed some updates (now included).

This commit also removes the setting of hard "limits" as splitting is not desired and the LimitRanger does this correctly already and we should not reimplement this logic ourselves.

All existing tests have been fixed and two additional tests have been added to verify behavior when a sidecar is present. In addition this patch has been manually tested in our development environment with live pipelines.

Fixes #4978

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Only use step containers for limitrange default request calculations
```

### DIFF
--- a/pkg/internal/limitrange/transformer.go
+++ b/pkg/internal/limitrange/transformer.go
@@ -45,45 +45,32 @@ func NewTransformer(ctx context.Context, namespace string, lister corev1listers.
 
 		// The assumption here is that the min, max, default, ratio have already been
 		// computed if there is multiple LimitRange to satisfy the most (if we can).
-		// Count the number of containers (that we know) in the Pod.
+		// Count the number of "step" containers in the Pod.
 		// This should help us find the smallest request to apply to containers
-		nbContainers := len(p.Spec.Containers)
-		// FIXME(#4230) maxLimitRequestRatio to support later
-		defaultLimits := getDefaultLimits(limitRange)
-		defaultInitRequests := getDefaultInitContainerRequest(limitRange)
-		for i := range p.Spec.InitContainers {
-			// We are trying to set the smallest requests possible
-			if p.Spec.InitContainers[i].Resources.Requests == nil {
-				p.Spec.InitContainers[i].Resources.Requests = defaultInitRequests
-			} else {
-				for _, name := range resourceNames {
-					setRequestsOrLimits(name, p.Spec.InitContainers[i].Resources.Requests, defaultInitRequests)
-				}
-			}
-			// We are trying to set the highest limits possible
-			if p.Spec.InitContainers[i].Resources.Limits == nil {
-				p.Spec.InitContainers[i].Resources.Limits = defaultLimits
-			} else {
-				for _, name := range resourceNames {
-					setRequestsOrLimits(name, p.Spec.InitContainers[i].Resources.Limits, defaultLimits)
-				}
+		stepContainers := []*corev1.Container{}
+		for i, c := range p.Spec.Containers {
+			if pod.IsContainerStep(c.Name) {
+				stepContainers = append(stepContainers, &p.Spec.Containers[i])
 			}
 		}
+		// No step containers, bail early we don't have anything to transform.
+		if len(stepContainers) == 0 {
+			return p, nil
+		}
 
-		defaultRequests := getDefaultAppContainerRequest(limitRange, nbContainers)
-		for i := range p.Spec.Containers {
-			if p.Spec.Containers[i].Resources.Requests == nil {
-				p.Spec.Containers[i].Resources.Requests = defaultRequests
+		defaultStepRequests := getDefaultStepContainerRequest(limitRange, len(stepContainers))
+
+		// Empty LimitRange Requests, bail early we don't have anything to transform.
+		if len(defaultStepRequests) == 0 {
+			return p, nil
+		}
+
+		for _, c := range stepContainers {
+			if c.Resources.Requests == nil {
+				c.Resources.Requests = defaultStepRequests
 			} else {
 				for _, name := range resourceNames {
-					setRequestsOrLimits(name, p.Spec.Containers[i].Resources.Requests, defaultRequests)
-				}
-			}
-			if p.Spec.Containers[i].Resources.Limits == nil {
-				p.Spec.Containers[i].Resources.Limits = defaultLimits
-			} else {
-				for _, name := range resourceNames {
-					setRequestsOrLimits(name, p.Spec.Containers[i].Resources.Limits, defaultLimits)
+					setRequests(name, c.Resources.Requests, defaultStepRequests)
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


imjasonh:
Does this also need to update docs here? https://github.com/tektoncd/pipeline/blob/main/docs/compute-resources.md

pkg/internal/limitrange/transformer.go
```diff
@@ -45,98 +46,78 @@ func NewTransformer(ctx context.Context, namespace string, lister corev1listers.
...
-			// We are trying to set the highest limits possible
-			if p.Spec.InitContainers[i].Resources.Limits == nil {
-				p.Spec.InitContainers[i].Resources.Limits = defaultLimits
-			} else {
-				for _, name := range resourceNames {
-					setRequestsOrLimits(name, p.Spec.InitContainers[i].Resources.Limits, defaultLimits)
-				}
+		stepContainers := []*corev1.Container{}
+		for i, c := range p.Spec.Containers {
+			if strings.HasPrefix(c.Name, "step-") {
```
lbernick, line 53: There's already a function for this purpose: https://github.com/tektoncd/pipeline/blob/d48cfdd1faa831a028b5cbffc99b92fd18a6915e/pkg/pod/entrypoint.go#L275-L277

I'd prefer using the same helper in both places

pkg/internal/limitrange/transformer.go
```diff
@@ -45,98 +46,78 @@ func NewTransformer(ctx context.Context, namespace string, lister corev1listers.
...
+				stepContainers = append(stepContainers, &p.Spec.Containers[i])
 			}
 		}
+		// No step containers, bail early we don't have anything to transform.
```
lbernick, line 57: This will change existing behavior for init containers (currently the defaultrequest and default is applied to init containers, and now it will not be)

lbernick (changes requested):
Please update the [docs](https://github.com/tektoncd/pipeline/blob/main/docs/compute-resources.md) as well

pkg/internal/limitrange/transformer.go
skaegi: This does not change behavior as this is what the LimitRanger does already. We are in effect duplicating the work that this does. We only need to do this work for the Step containers as we are doing request splitting. For limits and for managing the LimitRange's for other containers in the Pod we should just let the LimitRanger do its thing.

pkg/internal/limitrange/transformer.go
```diff
@@ -45,98 +46,78 @@ func NewTransformer(ctx context.Context, namespace string, lister corev1listers.
...
+			return p, nil
+		}
+
+		for _, c := range stepContainers {
+			if c.Resources.Requests == nil {
+				c.Resources.Requests = defaultStepRequests
 			} else {
 				for _, name := range resourceNames {
-					setRequestsOrLimits(name, p.Spec.Containers[i].Resources.Limits, defaultLimits)
+					setRequests(name, c.Resources.Requests, defaultStepRequests)
```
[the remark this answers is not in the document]
skaegi: See comment above.

pkg/internal/limitrange/transformer.go
skaegi: Thanks! I was looking for something like this.

pkg/internal/limitrange/transformer.go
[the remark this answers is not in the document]
lbernick: Are you sure? I think this will work when the step container doesn't have any limits, but if the step or sidecar container does have limits, I think the pod we end up creating could be rejected by the limitrange

pkg/internal/limitrange/transformer.go
lbernick: ah right, I think this will work since init containers won't have any resource requirements set.

skaegi: There is one difference that is worth discussing. With the existing Tekton transformer if a container has set a limit that is "higher" than the max limit the transformer will overwrite with the max limit value. For the same situation the LimitRanger will throw an error on creation -- see https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/cpu-constraint-namespace/#attempt-to-create-a-pod-that-exceeds-the-maximum-cpu-constraint

This is a change in functionality however I think it's unlikely anyone was relying on this silent downgrading. If we can avoid it we probably should stick with the existing LimitRange semantics but I wanted to call it out.

lbernick: Yeah, this is what I was attempting to get at in the other comment thread :) We do document this behavior and it's hard to know if anyone is relying on it or not. There are multiple options for what the "correct" behavior should be, so I'd prefer to keep the scope of this PR to just setting the correct defaults for sidecars, and if we want to change the behavior of limits, let's discuss that separately.

skaegi: ok... let me roll back time a bit here.

skaegi:
Closing this -- I'l re-open as two separate PRs -- one for the sidecar fix, one for limitrange semantics

PR #4990, opened 2022-06-16
TEP-0090: Fan Out `TaskRuns`

jerop:
# Changes

This change implements the fan out of `TaskRuns` from a `PipelineTask` with a `Matrix`. The fanned-out `TaskRuns` are executed in parallel.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
A `PipelineTask` with a `Matrix` is fanned out into parallel `TaskRuns` which are executed in parallel.
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -38,6 +39,7 @@ import (
 	listersv1alpha1 "github.com/tektoncd/pipeline/pkg/client/listers/pipeline/v1alpha1"
 	listers "github.com/tektoncd/pipeline/pkg/client/listers/pipeline/v1beta1"
 	resourcelisters "github.com/tektoncd/pipeline/pkg/client/resource/listers/resource/v1alpha1"
+	"github.com/tektoncd/pipeline/pkg/matrix"
 	"github.com/tektoncd/pipeline/pkg/pipelinerunmetrics"
 	tknreconciler "github.com/tektoncd/pipeline/pkg/reconciler"
 	"github.com/tektoncd/pipeline/pkg/reconciler/events"
@@ -688,7 +690,8 @@ func (c *Reconciler) runNextSchedulableTask(ctx context.Context, pr *v1beta1.Pip
 		if rprt == nil || rprt.Skip(pipelineRunFacts).IsSkipped || rprt.IsFinallySkipped(pipelineRunFacts).IsSkipped {
 			continue
 		}
-		if rprt.IsCustomTask() {
+		switch {
+		case rprt.IsCustomTask():
 			if rprt.IsFinalTask(pipelineRunFacts) {
 				rprt.Run, err = c.createRun(ctx, rprt, pr, getFinallyTaskRunTimeout)
 			} else {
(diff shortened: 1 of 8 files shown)


jerop:
cc @abayer @dibyom @lbernick @pritidesai please take a look

jerop:
`TLS handshake timeout` 🤔 

/retest

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -750,10 +764,24 @@ func (c *Reconciler) updateRunsStatusDirectly(pr *v1beta1.PipelineRun) error {
 
 type getTimeoutFunc func(ctx context.Context, pr *v1beta1.PipelineRun, rprt *resources.ResolvedPipelineRunTask, c clock.PassiveClock) *metav1.Duration
 
-func (c *Reconciler) createTaskRun(ctx context.Context, rprt *resources.ResolvedPipelineRunTask, pr *v1beta1.PipelineRun, storageBasePath string, getTimeoutFunc getTimeoutFunc) (*v1beta1.TaskRun, error) {
+func (c *Reconciler) createTaskRuns(ctx context.Context, rprt *resources.ResolvedPipelineRunTask, pr *v1beta1.PipelineRun, storageBasePath string, getTimeoutFunc getTimeoutFunc) ([]*v1beta1.TaskRun, error) {
+	var taskRuns []*v1beta1.TaskRun
+	matrixCombinations := matrix.FanOut(rprt.PipelineTask.Matrix).ToMap()
+	for i, taskRunName := range rprt.TaskRunNames {
+		params := matrixCombinations[strconv.Itoa(i)]
+		taskRun, err := c.createTaskRun(ctx, taskRunName, params, rprt, pr, storageBasePath, getTimeoutFunc)
```
lbernick, line 772: It seems like this creates each taskrun sequentially and synchronously.

1. would it make sense to create taskruns in goroutines?
2. If some taskruns are successfully created and some are not, will the PipelineRun's status correctly reflect that? If this isn't too complicated, could you add a test case for this scenario?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -59,9 +59,10 @@ func (e *TaskNotFoundError) Error() string {
 // ResolvedPipelineRunTask contains a Task and its associated TaskRun, if it
```
lbernick, line 59: could you please update the docstring for this type?

pkg/reconciler/pipelinerun/pipelinerun.go
jerop: we could explore creating parallel taskruns in goroutines, but i'd prefer that we do that everywhere as opposed to matrix only - see we don't use goroutines when creating taskruns from next schedulable pipelinetasks (we use for loops as currently implemented in matrix):

https://github.com/tektoncd/pipeline/blob/9d60d0adf41e74f78029a0eaab73140fa4f7206a/pkg/reconciler/pipelinerun/pipelinerun.go#L687-L712

if we want to change that behavior, I suggest we do that outside of the scope of this work, what do you think?

jerop: re: `PipelineRun` status - we have not handled yet handled that - this change is focused on creating the `TaskRuns` (and it's already XL lol) - adding the `ChildReferences` for those `TaskRuns` will come right after this

lbernick: ah ok, it's helpful to know that we're already creating multiple taskruns in one reconcile loop, sounds like that's a good starting point for matrix then too.

re: status, my main concern was that we'd end up in an inconsistent state in the scenario I just described (some taskrun creations failing and some succeeding), and that the reconciler would attempt to create the taskruns again in the next reconcile loop. I'd be curious to know how we currently handle this when creating TaskRuns for all the Tasks that are ready to be scheduled. However, it sounds like since matrix isn't usable yet, that's ok for now? Maybe we revisit this concern when you add support for status updates?

lbernick (approved):

pkg/reconciler/pipelinerun/pipelinerun.go
jerop: the same thing applies for parallel taskruns in general, outside of the matrix, so looking at the codeblock referenced above I believe there will be a failure when one of parallel taskruns is created (and others are not created) which causes the rest to complete execution before the pipelinerun exits - we can write tests to show this 

so if we want to change that behavior, we can look into it more broadly for all parallel executions of taskruns/runs

lbernick: SGTM!

pritidesai:
thanks a bunch @jerop for this, appreciate all the efforts you have put in 🎉 ❤️  Can we please add an example and an `e2e` test for this feature, if they do not already exist?

jerop:
> thanks a bunch @jerop for this, appreciate all the efforts you have put in 🎉 ❤️ Can we please add an example and an `e2e` test for this feature, if they do not already exist?

@pritidesai fanning out of taskruns is not yet complete - at this point we create the taskruns but they are not yet passed to the pipelinerun as described in https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#status

if i were to add examples now, the pipelinerun would stay running and timeout 

opening the PRs implementing the status, which would include the examples and end-to-end tests

PR #4996, opened 2022-06-16
LimitRange calculation should only split Requests for Step Containers

### LINKED ISSUE #4978: LimitRange Memory and CPU Request Distribution does not handle sidecars correctly

skaegi:
# Changes
This commit changes how resource requirements are calculated when one or more LimitRanges are present in the namespace where a TaskRun is.

We now will only transform "step" containers and exclude all other app containers (e.g. sidecars and any other containers) when splitting "requests" from LimitRanges.

Existing tests have been fixed to use "Step" containers and two additional tests have been added to verify behavior when a sidecar is present.

Docs have been updated to reflect this change

Cleanup:
1. When computing the defaultStepRequests if a value "isZero" we do not add it to the result as this is not needed and adds clutter
2. If the defaultStepRequests object has no entries we do not add it to the `resources` requirements for a container as this is not needed and adds clutter.

Fixes #4978

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Only use step containers for limitrange default request calculations
```

### DIFF
--- a/docs/compute-resources.md
+++ b/docs/compute-resources.md
@@ -47,7 +47,7 @@ Therefore, the pod will have no effective CPU limit.
 Tekton allows users to specify resource requirements of [`Steps`](./tasks.md#defining-steps),
 which run sequentially. However, the pod's effective resource requirements are still the
 sum of its containers' resource requirements. This means that when specifying resource 
-requirements for `Step`containers, they must be treated as if they are running in parallel.
+requirements for `Step` containers, they must be treated as if they are running in parallel.
 
 Tekton adjusts `Step` resource requirements to comply with [LimitRanges](#limitrange-support).
 [ResourceQuotas](#resourcequota-support) are not currently supported.
@@ -68,26 +68,26 @@ in a `Task's` pod, unless there is a LimitRange present in the namespace.
 (Tekton doesn't allow users to configure init containers for a `Task`.)
 Tekton supports LimitRange minimum, maximum, and default resource requirements for containers,
 but does not support LimitRange ratios between requests and limits ([#4230](https://github.com/tektoncd/pipeline/issues/4230)).
-LimitRange types other than "Container" are not supported.
+LimitRange types other than "Container" are not considered for purposes of resource requirements.
 
 ### Requests
 
-If a `Step` does not have requests defined, the resulting container's requests are the larger of:
+If a container does not have requests defined, the resulting container's requests are the larger of:
 - the LimitRange minimum resource requests 
-- the LimitRange default resource requests, divided among the app containers
+- the LimitRange default resource requests (for init and Sidecar containers) or the LimitRange default resource requests divided by the number of Step containers (for Step containers)
 
-If a `Step` has requests defined, the resulting container's requests are the larger of:
-- the `Step's` requests
+If a container has requests defined, the resulting container's requests are the larger of:
+- the container's requests
 - the LimitRange minimum resource requests
 
 ### Limits
 
-If a `Step` does not have limits defined, the resulting container's limits are the smaller of:
+If a container does not have limits defined, the resulting container's limits are the smaller of:
 - the LimitRange maximum resource limits 
 - the LimitRange default resource limits
 
-If a `Step` has limits defined, the resulting container's limits are the smaller of:
-- the `Step's` limits
+If a container has limits defined, the resulting container's limits are the smaller of:
+- the container's limits
 - the LimitRange maximum resource limits
 
 ### Examples
@@ -116,12 +116,25 @@ A `Task` with 2 `Steps` and no resources specified would result in a pod with th
 
 | Container    | CPU request | CPU limit |
 | ------------ | ----------- | --------- |
-| container 1  | 500m        | 2         |
+| container 1  | 500m        | 2         | 
 | container 2  | 500m        | 2         |
 
-Here, the default CPU request was divided among app containers, and this value was used since it was greater
+Here, the default CPU request was divided among the step containers, and this value was used since it was greater
 than the minimum request specified by the LimitRange.
 The CPU limits are 2 for each container, as this is the default limit specifed in the LimitRange.
+
+If we had a `Task` with 2 `Steps` and 1 `Sidecar` with no resources specified would result in a pod with the following containers:  
+
+| Container    | CPU request | CPU limit |
+| ------------ | ----------- | --------- |
+| container 1  | 500m        | 2         |
+| container 2  | 500m        | 2         |
+| container 3  | 1           | 2         |
+
+For the first two containers, the default CPU request was divided among the step containers, and this value was used since it was greater
+than the minimum request specified by the LimitRange. The third container is a sidecar and since it is not a step container gets the full 
+default CPU request of 1. AS before the CPU limits are 2 for each container, as this is the default limit specifed in the LimitRange.
+
 
 Now, consider a `Task` with the following `Step`s:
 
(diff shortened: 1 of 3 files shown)


docs/compute-resources.md
```diff
@@ -68,26 +68,26 @@ in a `Task's` pod, unless there is a LimitRange present in the namespace.
...
-LimitRange types other than "Container" are not supported.
+LimitRange types other than "Container" are not considered for purposes of resource requirements.
 
 ### Requests
 
-If a `Step` does not have requests defined, the resulting container's requests are the larger of:
+If a container does not have requests defined, the resulting container's requests are the larger of:
 - the LimitRange minimum resource requests 
-- the LimitRange default resource requests, divided among the app containers
+- the LimitRange default resource requests or if a `Step` the result of dividing this value among the other `Step` containers
```
lbernick, line 77: this wording is a bit confusing, maybe something like:

`the LimitRange default resource requests (for init and Sidecar containers) or the LimitRange default resource requests divided by the number of Step containers (for Step containers)`

pkg/internal/limitrange/transformer_test.go
```diff
@@ -805,6 +789,188 @@ func TestTransformerMultipleContainer(t *testing.T) {
...
+				Image: "baz",
+			}, {
+				Name:  "step-bar",
+				Image: "baz",
+			}, {
+				Name:  "sidecar-fizz",
+				Image: "baz",
+			}},
+		},
+		want: corev1.PodSpec{
```
lbernick, line 823: it's hard to tell with this test case which containers are receiving the divided requests vs the full requests. we might need to update the test function to compare the container names as well as the resources, so the test can ensure that the correct containers are getting the correct resources.

skaegi: I've updated the sidecar tests and added support in cmpRequestsAndLimits to also check the Container Name matches but only if present in "wants"

docs/compute-resources.md
skaegi: Done.

docs/compute-resources.md
```diff
@@ -68,26 +68,26 @@ in a `Task's` pod, unless there is a LimitRange present in the namespace.
...
-LimitRange types other than "Container" are not supported.
+LimitRange types other than "Container" are not considered for purposes of resource requirements.
 
 ### Requests
 
-If a `Step` does not have requests defined, the resulting container's requests are the larger of:
+If a container does not have requests defined, the resulting container's requests are the larger of:
 - the LimitRange minimum resource requests 
-- the LimitRange default resource requests, divided among the app containers
+- the LimitRange default resource requests (for init and Sidecar containers) or the LimitRange default resource requests divided by the number of Step containers (for Step containers)
```
pritidesai, line 77: thank you for clarifying this 👍

skaegi: Lee's new and improved verbiage not mine ;)

pkg/internal/limitrange/transformer.go
```diff
@@ -45,33 +45,45 @@ func NewTransformer(ctx context.Context, namespace string, lister corev1listers.
 
 		// The assumption here is that the min, max, default, ratio have already been
 		// computed if there is multiple LimitRange to satisfy the most (if we can).
-		// Count the number of containers (that we know) in the Pod.
+		// Count the number of step containers in the Pod.
 		// This should help us find the smallest request to apply to containers
-		nbContainers := len(p.Spec.Containers)
+		nbStepContainers := 0
+		for _, c := range p.Spec.Containers {
+			if pod.IsContainerStep(c.Name) {
```
pritidesai, line 52: do we need to check for the running containers or any other specific state (not-running) of the container here? or the state of container does not matter at this point since they will be spawned with the configuration being determined here?

pritidesai: alright, so this is only used in pod builder while creating a pod so the state is not relevant here ...

pkg/internal/limitrange/transformer.go
```diff
@@ -45,33 +45,45 @@ func NewTransformer(ctx context.Context, namespace string, lister corev1listers.
...
-					setRequestsOrLimits(name, p.Spec.InitContainers[i].Resources.Limits, defaultLimits)
+					setRequestsOrLimits(name, p.Spec.InitContainers[i].Resources.Limits, defaultContainerLimits)
 				}
 			}
 		}
 
-		defaultRequests := getDefaultAppContainerRequest(limitRange, nbContainers)
-		for i := range p.Spec.Containers {
+		for i, c := range p.Spec.Containers {
+			var defaultRequests = defaultContainerRequests
```
pritidesai, line 82: I hardly see this way of initializing and declaring a variable, interesting 🙃

pkg/internal/limitrange/transformer.go
skaegi: The transform happens before the pod spec is applied so the state of the container does not matter at this point. We use the number of step containers to help choose the optimal request size at configuration generation time.

pkg/internal/limitrange/transformer.go
```diff
@@ -115,19 +127,29 @@ func getDefaultAppContainerRequest(limitRange *corev1.LimitRange, nbContainers i
...
+					result = takeTheMax(request, *resource.NewMilliQuantity(defaultRequest.MilliValue()/int64(nbContainers), defaultRequest.Format), min)
+				}
+				// only set non-zero request values
+				if !isZero(result) {
+					r[name] = result
 				}
 			}
 		}
 	}
+	// return nil if the resource list is empty to avoid setting an empty defaultrequest
```
pritidesai, line 144: NIT: should this be done for the other two other `get` functions - `getDefaultContainerRequest` and `getDefaultLimits`? I havent looked at whether its needed or not but just wondering since we added this here.

pritidesai:
thank you @skaegi, I have one `NIT` left which can be addressed in a follow up PR if needed.

/lgtm

Issue #4923, opened 2022-05-31
Wrong variable propagating Finally Task Results to Pipeline Results


vsinghai:
# Expected Behavior

As described in the [doc](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#cannot-configure-pipeline-result-with-finally), `Finally Task`'s can emit `Results` but they cannot be referenced in `Pipeline Result`'s - but we intend to support this using `$(finally.<finally-task-name>.results.<finally-task-result>)`. This work is tracked in this [issue](https://github.com/tektoncd/pipeline/issues/2710).

**Example**

Modified the `Pipeline` in the provided [example](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml#:~:text=workspace%20after%20done-,apiVersion%3A%20tekton.dev/v1beta1,fi,-%2D%2D%2D) as such:

```
# Pipeline to clone repo into shared workspace and cleanup the workspace after done
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: clone-cleanup-workspace
spec:
  results:
      - name: commit
        value: $(tasks.clone-app-repo.results.commit)
      - name: initialized
        value: $(tasks.check-git-commit.results.init)
  workspaces:
    # common workspace where git repo is cloned and needs to be cleanup after done
    - name: git-source
  tasks:
    # Clone app repo to workspace
    - name: clone-app-repo
      taskRef:
        name: git-clone-from-catalog
      params:
        - name: url
          value: https://github.com/tektoncd/community.git
        - name: subdirectory
          value: application
      workspaces:
        - name: output
          workspace: git-source
  finally:
    # Cleanup workspace
    - name: cleanup
      taskRef:
        name: cleanup-workspace
      workspaces:
        - name: source
          workspace: git-source
    - name: check-git-commit
      params:
        - name: commit1
          value: $(tasks.clone-app-repo.results.commit)
      taskSpec:
        params:
          - name: commit1
        results:
          - name: init
        steps:
          - name: check-commit-initialized
            image: alpine
            script: |
              if [[ ! $(params.commit1) ]]; then
                exit 1
              fi
              printf true | tee /tekton/results/init

```

# Actual Behavior

And this was the created `PipelineRun`, with the `Finally Task`'s `Result` passed on to the `Pipeline`:

```
$ tkn pr describe -L
Name:              write-and-cleanup-workspace
[...]

PR #5000, opened 2022-06-17
Validation for Finally Task Results referenced in Pipeline Results

vsinghai:
# Changes

Prior to this commit, there was no validation for `Finally`
`Task` `Result`'s being referenced in `Pipeline` `Result`'s.

This commit creates validation for that case by making
sure all the `Finally` `Task` `Result`'s reference a `task_name`
within the `Pipeline` `Result`'s. This commit also gathers
all the errors in `Pipeline Results` validation instead
of exiting on the first error. The reason is that if
multiple references exist to non-existant tasks, this
function will only return an error message about one of them,
forcing the person to fix the errors one by one.

Fixes bug [#4923](https://github.com/tektoncd/pipeline/issues/4923)

/kind bug
/cc @jerop 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fixed a bug where Finally Task Result's are being referenced in Pipeline Result's.
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation.go
@@ -69,7 +69,7 @@ func (ps *PipelineSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	errs = errs.Also(validatePipelineWorkspacesUsage(ps.Workspaces, ps.Tasks).ViaField("tasks"))
 	errs = errs.Also(validatePipelineWorkspacesUsage(ps.Workspaces, ps.Finally).ViaField("finally"))
 	// Validate the pipeline's results
-	errs = errs.Also(validatePipelineResults(ps.Results))
+	errs = errs.Also(validatePipelineResults(ps.Results, ps.Tasks))
 	errs = errs.Also(validateTasksAndFinallySection(ps))
 	errs = errs.Also(validateFinalTasks(ps.Tasks, ps.Finally))
 	errs = errs.Also(validateWhenExpressions(ps.Tasks, ps.Finally))
@@ -255,16 +255,17 @@ func filter(arr []string, cond func(string) bool) []string {
 }
 
 // validatePipelineResults ensure that pipeline result variables are properly configured
-func validatePipelineResults(results []PipelineResult) (errs *apis.FieldError) {
+func validatePipelineResults(results []PipelineResult, tasks []PipelineTask) (errs *apis.FieldError) {
+	pipelineTaskNames := getPipelineTasksNames(tasks)
 	for idx, result := range results {
 		expressions, ok := GetVarSubstitutionExpressionsForPipelineResult(result)
 		if !ok {
-			return errs.Also(apis.ErrInvalidValue("expected pipeline results to be task result expressions but no expressions were found",
+			errs = errs.Also(apis.ErrInvalidValue("expected pipeline results to be task result expressions but no expressions were found",
 				"value").ViaFieldIndex("results", idx))
 		}
 
 		if !LooksLikeContainsResultRefs(expressions) {
-			return errs.Also(apis.ErrInvalidValue("expected pipeline results to be task result expressions but an invalid expressions was found",
+			errs = errs.Also(apis.ErrInvalidValue("expected pipeline results to be task result expressions but an invalid expressions was found",
 				"value").ViaFieldIndex("results", idx))
 		}
 
@@ -275,9 +276,40 @@ func validatePipelineResults(results []PipelineResult) (errs *apis.FieldError) {
 				"value").ViaFieldIndex("results", idx))
 		}
 
+		if !taskContainsResult(result.Value.StringVal, pipelineTaskNames) {
+			errs = errs.Also(apis.ErrInvalidValue("referencing a nonexistent task",
+				"value").ViaFieldIndex("results", idx))
+		}
 	}
 
 	return errs
+}
+
+// put task names in a set
+func getPipelineTasksNames(pipelineTasks []PipelineTask) sets.String {
+	pipelineTaskNames := make(sets.String)
+	for _, pipelineTask := range pipelineTasks {
+		pipelineTaskNames.Insert(pipelineTask.Name)
+	}
+
+	return pipelineTaskNames
+}
+
+// taskContainsResult ensures the result value is referenced within the
+// task names
+func taskContainsResult(resultExpression string, pipelineTaskNames sets.String) bool {
+	// split incase of multiple resultExpressions in the same result.Value string
+	// i.e "$(task.<task-name).result.<result-name>) - $(task2.<task2-name).result2.<result2-name>)"
+	split := strings.Split(resultExpression, "$")
+	for _, expression := range split {
+		if expression != "" {
+			pipelineTaskName, _, _, _, _ := parseExpression(stripVarSubExpression("$" + expression))
+			if !pipelineTaskNames.Has(pipelineTaskName) {
+				return false
+			}
+		}
+	}
+	return true
 }
 
 func validateTasksAndFinallySection(ps *PipelineSpec) *apis.FieldError {
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
@@ -1149,7 +1149,7 @@ func TestValidatePipelineResults_Success(t *testing.T) {
 		Description: "this is my pipeline result",
 		Value:       *NewArrayOrString("$(tasks.a-task.results.gitrepo.commit)"),
 	}}
-	if err := validatePipelineResults(results); err != nil {
+	if err := validatePipelineResults(results, []PipelineTask{{Name: "a-task"}}); err != nil {
 		t.Errorf("Pipeline.validatePipelineResults() returned error for valid pipeline: %s: %v", desc, err)
 	}
 }
@@ -1166,10 +1166,8 @@ func TestValidatePipelineResults_Failure(t *testing.T) {
 			Description: "this is my pipeline result",
 			Value:       *NewArrayOrString("$(tasks.a-task.results.output.key1.extra)"),
 		}},
-		expectedError: apis.FieldError{
-			Message: `invalid value: expected all of the expressions [tasks.a-task.results.output.key1.extra] to be result expressions but only [] were`,
-			Paths:   []string{"results[0].value"},
-		},
+		expectedError: *apis.ErrInvalidValue(`expected all of the expressions [tasks.a-task.results.output.key1.extra] to be result expressions but only [] were`, "results[0].value").Also(
+			apis.ErrInvalidValue("referencing a nonexistent task", "results[0].value")),
 	}, {
 		desc: "invalid pipeline result value with static string",
 		results: []PipelineResult{{
@@ -1177,10 +1175,9 @@ func TestValidatePipelineResults_Failure(t *testing.T) {
 			Description: "this is my pipeline result",
 			Value:       *NewArrayOrString("foo.bar"),
 		}},
-		expectedError: apis.FieldError{
-			Message: `invalid value: expected pipeline results to be task result expressions but no expressions were found`,
-			Paths:   []string{"results[0].value"},
-		},
+		expectedError: *apis.ErrInvalidValue(`expected pipeline results to be task result expressions but an invalid expressions was found`, "results[0].value").Also(
+			apis.ErrInvalidValue(`expected pipeline results to be task result expressions but no expressions were found`, "results[0].value")).Also(
+			apis.ErrInvalidValue(`referencing a nonexistent task`, "results[0].value")),
 	}, {
 		desc: "invalid pipeline result value with invalid expression",
 		results: []PipelineResult{{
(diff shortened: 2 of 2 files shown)


pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -1162,6 +1162,82 @@ func TestValidatePipelineResults_Failure(t *testing.T) {
...
 }
 
+func TestFinallyTaskResultsToPipelineResults_Failure(t *testing.T) {
+	tests := []struct {
+		desc          string
+		p             *Pipeline
+		expectedError apis.FieldError
+		wc            func(context.Context) context.Context
+	}{{
+		desc: "invalid propagation of finally task results from pipeline results",
```
lbernick, line 1172: could you please simplify this test case? for example I don't think there's any need for workspaces

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -265,15 +266,53 @@ func validatePipelineResults(results []PipelineResult) (errs *apis.FieldError) {
 		expressions = filter(expressions, looksLikeResultRef)
 		resultRefs := NewResultRefs(expressions)
 		if len(expressions) != len(resultRefs) {
-			errs = errs.Also(apis.ErrInvalidValue(fmt.Sprintf("expected all of the expressions %v to be result expressions but only %v were", expressions, resultRefs),
+			return errs.Also(apis.ErrInvalidValue(fmt.Sprintf("expected all of the expressions %v to be result expressions but only %v were", expressions, resultRefs),
 				"value").ViaFieldIndex("results", idx))
 		}
 
+		if !TaskContainsResult(result.Value, taskMap) {
+			return errs.Also(apis.ErrInvalidValue("invalid propagation of finally task results to pipeline results",
```
lbernick, line 274: This error message is a bit confusing because we are planning to allow finally task results to be propagated to pipeline results. The actual problem seems to be that the "tasks.results.name" syntax is being used to refer to a finally task, so the error message should reflect that.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -265,15 +266,53 @@ func validatePipelineResults(results []PipelineResult) (errs *apis.FieldError) {
...
+
+	if len(sections) < 2 {
+		return false
+	}
+	_, ok := taskMap[sections[1]]
+	return ok
+}
+
+// split result.Value into its respective sections
+func splitValue(value string) []string {
```
lbernick, line 308: poking around some of the helper functions called in `validatePipelineResults`, it looks like the file `resultrefs.go` has some helpers for parsing results of the form `tasks.foo.results.bar`. Can you make use of some of those helpers? @chuangw6 @Yongxuanzhang or @ywluogg might know better how this works

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -265,15 +266,53 @@ func validatePipelineResults(results []PipelineResult) (errs *apis.FieldError) {
...
+			return errs.Also(apis.ErrInvalidValue("invalid propagation of finally task results to pipeline results",
+				"value").ViaFieldIndex("results", idx))
+		}
 	}
 
 	return errs
 }
 
+// put task names in a dictionary
+func convertTaskToTaskMap(tasks []PipelineTask) map[string]string {
```
lbernick, line 283: I don't think you need a map here, just a slice of task names.

lbernick (commented):
I find the commit message a bit confusing because it makes it sound like we already support finally task results in pipeline results. maybe you could clarify that the syntax "tasks.foo.results.bar" can't be used to refer to finally tasks, and that's what this commit adds validation for?

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
vsinghai: agreed, outside of workspaces I don't believe there is anything else that can be taken out to make the test case simplified.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -6644,7 +6644,7 @@ spec:
 		t.Errorf("expected invalid task reference condition but saw: %v", cond1)
 	}
 
-	if cond2.Reason != ReasonInvalidTaskResultReference {
+	if cond2.Reason != ReasonFailedValidation {
```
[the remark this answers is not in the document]
vsinghai: The errors are being caught before it hits the reconciler. So in order to account for that, the reasoning for the error needs to be that the validation failed.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
vsinghai: I am putting the slice of task names into a map for faster processing, otherwise I would have to loop through the task name every time I am trying to validate something.

Do you think this is the best approach, or should I go with the latter?

pkg/apis/pipeline/v1beta1/pipeline_validation.go
vsinghai: does `referencing a nonexistent task` work better?

lbernick: yes that works!

pkg/apis/pipeline/v1beta1/pipeline_validation.go
lbernick: how about sets.string? https://github.com/tektoncd/pipeline/blob/7b9eb0c553dedec05218c01ec8aa99378dee9361/vendor/k8s.io/apimachinery/pkg/util/sets/string.go#L30

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
lbernick: you could do a single pipeline result, no pipeline tasks, and one finally task; also the finally task doesn't need params

pkg/reconciler/pipelinerun/pipelinerun_test.go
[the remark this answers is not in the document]
lbernick: sorry, I still don't understand. pipelinerun.go hasn't been changed, so I don't understand why pipelinerun_test.go is changing. these unit tests should be run completely separately.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
chuangw6: yeah, this [parseExpression](https://github.com/tektoncd/pipeline/blob/e7520a0a2dc78f37c3552e7f22b20d40546df6bd/pkg/apis/pipeline/v1beta1/resultref.go#L157) function can help extract taskName and resultName from the expression `tasks.<taskName>.results.<resultName>`.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
vsinghai: done.

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
vsinghai: yes, everything has been update but in order to have a finally field there must be a pipeline task.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
vsinghai: updated.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
vsinghai: updated to use sets.

Yongxuanzhang:
Looks like test TestExamples/v1beta1/pipelineruns/pipelinerun-results is failing, maybe you could run it locally and see if it can work?

vsinghai:
> Looks like test TestExamples/v1beta1/pipelineruns/pipelinerun-results is failing, maybe you could run it locally and see if it can work?

It does run locally

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -265,15 +266,44 @@ func validatePipelineResults(results []PipelineResult) (errs *apis.FieldError) {
...
+func TaskContainsResult(value string, taskSet sets.String) bool {
+	split := strings.Split(value, "$")
+	for _, val := range split {
+		if val != "" {
+			taskName, _, _, _, _ := parseExpression(stripVarSubExpression("$" + val))
+			if !taskSet.Has(taskName) {
+				return false
+			}
+		}
+	}
```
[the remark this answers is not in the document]
jerop: value --> expression
taskSet --> pipelineTaskNames
taskName --> pipelineTaskName

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -1162,6 +1162,70 @@ func TestValidatePipelineResults_Failure(t *testing.T) {
 	}
 }
 
+func TestFinallyTaskResultsToPipelineResults_Failure(t *testing.T) {
```
jerop, line 1165: please add an example where the expected error is nil

jerop (commented):
thanks @vsinghai!

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
vsinghai: just to clarify, does this mean to write a `success` test?

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -265,15 +266,44 @@ func validatePipelineResults(results []PipelineResult) (errs *apis.FieldError) {
...
+			return errs.Also(apis.ErrInvalidValue("referencing a nonexistent task",
+				"value").ViaFieldIndex("results", idx))
+		}
 	}
 
 	return errs
 }
 
+// put task names in a set
+func convertTaskToTaskSet(tasks []PipelineTask) sets.String {
```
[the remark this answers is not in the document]
vsinghai: thank you for the renaming suggestions

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
jerop: yes

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -265,15 +266,44 @@ func validatePipelineResults(results []PipelineResult) (errs *apis.FieldError) {
...
+	for _, pipelineTask := range pipelineTasks {
+		pipelineTaskNames.Insert(pipelineTask.Name)
+	}
+
+	return pipelineTaskNames
+}
+
+// TaskContainsResult ensures the result value is referenced within the
+// task names
+func TaskContainsResult(value string, pipelineTaskNames sets.String) bool {
```
jerop, line 294: this function is used within this package only so it should be unexported

```suggestion
func taskContainsResult(value string, pipelineTaskNames sets.String) bool {
```

vsinghai: thank you for the suggestion, done.

jerop:
`workspace-in-sidecar` is a known flake - https://github.com/tektoncd/pipeline/issues/4169

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -1152,7 +1150,7 @@ func TestValidatePipelineResults_Failure(t *testing.T) {
 		},
 	}}
 	for _, tt := range tests {
-		err := validatePipelineResults(tt.results)
+		err := validatePipelineResults(tt.results, []PipelineTask{})
```
jerop, line 1153: just so you don't have to modify that test case - that it stays focused on the result expression - you can pass in the pipelinetask name a-task:

```suggestion
		err := validatePipelineResults(tt.results, []PipelineTask{{Name: "a-task"})
```

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -1162,6 +1160,114 @@ func TestValidatePipelineResults_Failure(t *testing.T) {
...
+			ctx := context.Background()
+			if tt.wc != nil {
+				ctx = tt.wc(ctx)
+			}
+			err := tt.p.Validate(ctx)
+			if err == nil {
+				t.Errorf("Pipeline.validatePipelineResults() did not return for invalid pipeline: %s", tt.desc)
+			}
+			if d := cmp.Diff(tt.expectedError.Error(), err.Error(), cmpopts.IgnoreUnexported(apis.FieldError{})); d != "" {
+				t.Errorf("Pipeline.validateParamResults() errors diff %s", diff.PrintWantGot(d))
```
jerop, line 1265: copy pasta?
```suggestion
				t.Errorf("Pipeline.validatePipelineResults() errors diff %s", diff.PrintWantGot(d))
```

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
jerop: and add your own test cases to `TestValidatePipelineResults_Failure` given that you've changed the logic for `validatePipelineResults`

jerop (approved):

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -269,11 +270,42 @@ func validatePipelineResults(results []PipelineResult) (errs *apis.FieldError) {
 				"value").ViaFieldIndex("results", idx))
 		}
 
+		if !taskContainsResult(result.Value.StringVal, pipelineTaskNames) {
+			return errs.Also(apis.ErrInvalidValue("referencing a nonexistent task",
```
lbernick, line 274: It might be better to use `errs = errs.Also` instead of a return statement here. The reason is that if multiple references exist to non-existent tasks, this function will only return an error message about one of them, forcing the person to fix the errors one by one.

vsinghai: done, thank you for the feedback.

jerop:
@vsinghai could you please update the PR description as well (about gathering all the errors, instead of exiting)?

vsinghai:
> 

done

vsinghai:
> @vsinghai could you please update the PR description as well (about gathering all the errors, instead of exiting)?

done

PR #5005, opened 2022-06-20
Remove v1alpha1 Pipeline, PipelineRun, Task, TaskRun, and ClusterTask

### LINKED ISSUE #4982: Removal of v1alpha1 Task, TaskRun, Pipeline, and PipelineRun
This issue tracks the removal of v1alpha1 Task, TaskRun, Pipeline, and PipelineRun, which are deprecated in favor of the v1beta1 versions.

@abayer do you want to be assigned to this?

abayer:
# Changes

Closes #4982

This removes the long deprecated pipeline v1alpha1 APIs for everything that's replicated in v1beta1.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
action required: `v1alpha1` `Pipeline`, `PipelineRun`, `Task`, `TaskRun` removed. Please switch to `v1beta1` for those types.
```


abayer:
I...honestly did not expect everything to pass so soon. =)

vdemeester (approved):
/cc @tektoncd/cli-maintainers as this will have an impact on the CLI (a good one, we'll have to clean our v1alpha1 support as well)

abayer:
cc @tektoncd/core-maintainers

abayer:
/hold

I'm holding this until I get feedback from enough people. =)

pkg/apis/pipeline/v1alpha1/run_types.go
```diff
@@ -47,7 +47,7 @@ type EmbeddedRunSpec struct {
 // RunSpec defines the desired state of Run
 type RunSpec struct {
 	// +optional
-	Ref *TaskRef `json:"ref,omitempty"`
+	Ref *v1beta1.TaskRef `json:"ref,omitempty"`
```
afrittoli, line 50: Any change to the `TaskRef` in beta will impact `Runs` in alpha now.
I guess that should be ok `Run` didn't have its own `TaskRef` definition before anyways.

pkg/apis/pipeline/v1alpha1/run_types.go
```diff
@@ -69,7 +69,7 @@ type RunSpec struct {
 
 	// PodTemplate holds pod specific configuration
 	// +optional
-	PodTemplate *PodTemplate `json:"podTemplate,omitempty"`
+	PodTemplate *v1beta1.PodTemplate `json:"podTemplate,omitempty"`
```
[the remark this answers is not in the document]
abayer: This was actually already the case - https://github.com/tektoncd/pipeline/blob/13a3c5d9612af58382b0fe31622ec0468f311d88/pkg/apis/pipeline/v1alpha1/pod.go#L8

pkg/apis/pipeline/v1alpha1/run_types.go
abayer: Again, already the case: https://github.com/tektoncd/pipeline/blob/13a3c5d9612af58382b0fe31622ec0468f311d88/pkg/apis/pipeline/v1alpha1/pipeline_types.go#L226 =)

afrittoli (approved):
Thank you!!! 

It really nice to see all these alpha types finally go :)

/meow

abayer:
We've overloaded the term "alpha" - in some contexts, it means `v1alpha1`, in others it means alpha vs beta vs stable feature flags, and I often get confused. =) I _think_ @wlynch is talking about `v1alpha1` fields/types, since some do still lurk (namely `Run` related stuff). I don't know what the right thing to do is re serving the current messiness...

afrittoli:
> We've overloaded the term "alpha" - in some contexts, it means `v1alpha1`, in others it means alpha vs beta vs stable feature flags, and I often get confused. =) I _think_ @wlynch is talking about `v1alpha1` fields/types, since some do still lurk (namely `Run` related stuff). I don't know what the right thing to do is re serving the current messiness...

Oh I see, thank you for the clarification.

The ability to specify things like a PipelineResource in a beta Task or Pipeline is beta, but is deprecated. We won't be able to remove it from v1beta1 for some time still.

The ability to have a Run embedded in a Pipeline is an alpha feature in the beta api. we would like to make that a stable feature sometimes soon.

For conditions I'm not quite sure.

afrittoli:
> > We've overloaded the term "alpha" - in some contexts, it means `v1alpha1`, in others it means alpha vs beta vs stable feature flags, and I often get confused. =) I _think_ @wlynch is talking about `v1alpha1` fields/types, since some do still lurk (namely `Run` related stuff). I don't know what the right thing to do is re serving the current messiness...
> 
> Yup! Exactly.
> 
> I'm curious because one of the things I was hoping could happen eventually was renaming `Run` -> `CustomRun` while the feature is still alpha, because we refer to "Run reconcilers" in other places like chains and results to refer to reconcilers that can act on {Pipeline, Task, Custom}Runs. But if you look at how CustomRuns are used in v1beta1 today, there's no indication that these are considered alpha for API stability -

I feel like we discussed about this already, but I cannot remember for sure.
I'm not sure I see enough value in changing the name to justify it, but also I'm not opposed to it.
@imjasonh @vdemeester wdyt?

> 
> https://github.com/tektoncd/pipeline/blob/3147117bda072f801f03268eb2a6d1379a692962/pkg/apis/pipeline/v1beta1/pipeline_types.go#L147-L149
> 
> Mixing different stability policies in the same API makes it a bit confusing what is / isn't in scope - e.g. if we make a breaking change to how CustomRuns are referred to in v1beta1 APIs, does that need to follow beta compatibility deprecation?

The ability to refer custom tasks in a Pipeline is an alpha feature, so if we change that it does not follow the beta compatibility. 

>  While I'm all for simplifying the maintenance burden of v1alpha1, part of me wonders if there's value in keeping a thin layer so we can quickly iterate on experimental features without needing to impact higher stability APIs. 🤔

The idea is that each API exposes by defaults only features that are aligned with the API version in terms of stability.
That means that the `v1beta1` API will not allow using any `alpha` feature unless we explicitly set the API stability level to `alpha` in the config.  Once we have `v1`, it will follow the same pattern, `v1` will not show `alpha` or `beta` features unless configured to do so.

We could introduce a `v1alpha2` which would only be a thin shim i.e. identical to `v1beta1` with `alpha` feature enabled. That `v1alpha2` API would allow getting access to alpha features without changing the config. Do you think that would be useful?

vdemeester:
I tend to agree with @afrittoli. Custom Tasks are only available if we set the api feature flag to `alpha`, so we consider those features alpha (with an alpha timeline). This means if next release we want to remove it (or here rename it), we could (it would be better to do it in a "as little breakage as possible" way but still…).

Custom Tasks are even more "weird" in that, from the user perspective (aka the yaml provide) we never "refer" to the `Run` type. It almost make `Run` an implementation detail that is not visible to the end-user. It is however visible for tools integrating with pipeline.

> I feel like we discussed about this already, but I cannot remember for sure.
> I'm not sure I see enough value in changing the name to justify it, but also I'm not opposed to it.
> @imjasonh @vdemeester wdyt?

I don't have a strong opinion on this, but if we want to change the name, it would be better to do it earlier than later. Changing it today, would mean all the custom task in experimental need to be changed, same with external ones (automatiko manual approval, …).

> We could introduce a `v1alpha2` which would only be a thin shim i.e. identical to `v1beta1` with `alpha` feature enabled. That `v1alpha2` API would allow getting access to alpha features without changing the config. Do you think that would be useful?

That could be a possibility. I wonder what are the cons for this approach, what problems it could "pose".

lbernick:
@wlynch 
> I'm curious because one of the things I was hoping could happen eventually was renaming `Run` -> `CustomRun` while the feature is still alpha

Let's keep "Run" naming out of scope for this PR.

> Mixing different stability policies in the same API makes it a bit confusing what is / isn't in scope

I agree this can be confusing although this is covered in [tep 33](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md)

> - e.g. if we make a breaking change to how CustomRuns are referred to in v1beta1 APIs, does that need to follow beta compatibility deprecation?

Yes, but I don't think that's in scope for this PR.

> While I'm all for simplifying the maintenance burden of v1alpha1, part of me wonders if there's value in keeping a thin layer so we can quickly iterate on experimental features without needing to impact higher stability APIs. 🤔

I'm not sure I understand what you're suggesting here. Is the suggestion to retain the concept of a "v1alpha1 API" which has v1alpha1 versions of all the CRDs you would need to run Tekton? v1alpha1 Run will still exist, at alpha stability, so we can iterate on it. Or is the concern that the v1beta1 Pipeline implicitly declares the spec of Run (currently `runtime.RawExtension`), so there's a beta CRD that depends on Run spec? If the latter, I think this is also out of scope.

wlynch:
> I'm not sure I understand what you're suggesting here. Is the suggestion to retain the concept of a "v1alpha1 API" which has v1alpha1 versions of all the CRDs you would need to run Tekton? 

Yup! (though probably more in line with what @afrittoli was saying with v1alpha2).

> Or is the concern that the v1beta1 Pipeline implicitly declares the spec of Run (currently `runtime.RawExtension`), so there's a beta CRD that depends on Run spec? If the latter, I think this is also out of scope.

The situation I'm worried about is if a change to an alpha field causes a breaking change in beta API clients - it's not immediately obvious in client code whether a beta client is running in "alpha mode", since that's a server-side setting. We kinda get around this today by having alpha checks during validation, though I don't think we have any way to verify what fields are alpha beyond doc comments (and we're not always consistent about labelling these) and whether they are actually enforced as alpha. For me having strict separation of alpha fields only be in alpha APIs makes this distinction clearer for both client and server, then promote fields to beta once we're happy. If we do stick with the alpha fields in beta APIs though, maybe we can do something to improve this like add custom struct tags to label and enforce alpha behavior. 🤔 

That said, agreed that this is out of scope for this PR so...

/lgtm

abayer:
@wlynch fwiw, `v1alpha1` didn't have alpha features, it was just the first version of the API. All the alpha features are in `v1beta1`.

lbernick (commented):
/lgtm

PR #5015, opened 2022-06-22
Check for non-nil `TaskRun` in `GetTaskRunsResults`

jerop:
# Changes

In this change, we handle scenarios where a matrixed `PipelineTask` is successful - so `rprt` is successful but its `TaskRun` field is nil because it has `TaskRuns` instead. We need to add a check that `rprt.TaskRun` is non-nil before attempting to access its field: `rprt.TaskRun.Status.TaskRunResults`.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


jerop:
@abayer @dibyom please take a look

PR #5018, opened 2022-06-22
Reject embedded `Task`s with `APIVersion` and/or `Kind` specified

### LINKED ISSUE #4543: Pipelines don't respect mixed Task API versions

abayer:
# Changes

fixes #4543

If custom tasks are enabled and either of `PipelineTask.TaskSpec.APIVersion` or `PipelineTask.TaskSpec.Kind` are specified, the `PipelineTask` will be treated as a custom task. This is still the case if an actual embedded `Task` is specified in `PipelineTask.TaskSpec.TaskSpec`, so we should be failing validation for those cases.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fail PipelineTask validation if a normal, non-custom embedded task is specified along with `apiVersion` and/or `kind`
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_types.go
@@ -321,6 +321,26 @@ func (pt *PipelineTask) validateMatrixCombinationsCount(ctx context.Context) (er
 	return errs
 }
 
+func (pt PipelineTask) validateEmbeddedOrType() (errs *apis.FieldError) {
+	// Reject cases where APIVersion and/or Kind are specified alongside an embedded Task.
+	// We determine if this is an embedded Task by checking of TaskSpec.TaskSpec.Steps has items.
+	if pt.TaskSpec != nil && len(pt.TaskSpec.TaskSpec.Steps) > 0 {
+		if pt.TaskSpec.APIVersion != "" {
+			errs = errs.Also(&apis.FieldError{
+				Message: "taskSpec.apiVersion cannot be specified when using taskSpec.steps",
+				Paths:   []string{"taskSpec.apiVersion"},
+			})
+		}
+		if pt.TaskSpec.Kind != "" {
+			errs = errs.Also(&apis.FieldError{
+				Message: "taskSpec.kind cannot be specified when using taskSpec.steps",
+				Paths:   []string{"taskSpec.kind"},
+			})
+		}
+	}
+	return
+}
+
 // GetMatrixCombinationsCount returns the count of combinations of Parameters generated from the Matrix in PipelineTask.
 func (pt *PipelineTask) GetMatrixCombinationsCount() int {
 	if len(pt.Matrix) == 0 {
@@ -464,6 +484,9 @@ func (pt PipelineTask) ValidateName() *apis.FieldError {
 // calls the validation routine based on the type of the task
 func (pt PipelineTask) Validate(ctx context.Context) (errs *apis.FieldError) {
 	errs = errs.Also(pt.validateRefOrSpec())
+
+	errs = errs.Also(pt.validateEmbeddedOrType())
+
 	cfg := config.FromContextOrDefaults(ctx)
 	// If EnableCustomTasks feature flag is on, validate custom task specifications
 	// pipeline task having taskRef with APIVersion is classified as custom task
--- a/docs/pipelines.md
+++ b/docs/pipelines.md
@@ -291,7 +291,7 @@ spec:
 
  Your `Pipeline` definition must reference at least one [`Task`](tasks.md).
 Each `Task` within a `Pipeline` must have a [valid](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names)
-`name` and a `taskRef`. For example:
+`name` and a `taskRef` or a `taskSpec`. For example:
 
 ```yaml
 tasks:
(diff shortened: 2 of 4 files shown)


pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -461,6 +461,24 @@ func (pt PipelineTask) ValidateName() *apis.FieldError {
 // calls the validation routine based on the type of the task
 func (pt PipelineTask) Validate(ctx context.Context) (errs *apis.FieldError) {
 	errs = errs.Also(pt.validateRefOrSpec())
+
+	// Reject cases where APIVersion and/or Kind are specified alongside an embedded Task.
+	// We determine if this is an embedded Task by checking of TaskSpec.TaskSpec.Steps has items.
+	if pt.TaskSpec != nil && len(pt.TaskSpec.TaskSpec.Steps) > 0 {
```
abayer, line 467: If anyone has a better way to determine whether the `PipelineTask` is, in fact, an embedded `Task`, I'd love to hear it. =)

vdemeester: We could at least extract this in a function on `PipelineTask`, so that if we change the way to determine it, it won't change that code (and would make that code more readable ? 🙃 )

vdemeester (approved):

pkg/apis/pipeline/v1beta1/pipeline_types.go
afrittoli: We require today that `kind` is specified if `apiVersion` is, but we don't check the content of `kind` and `apiVersion`.

Should we determine if it's an embedded `Task` or an embedded custom task based on the content of `kind` and `apiVersion`? We could say that (`tekton.dev/*`, `Task`) means it's an embedded `Task` and should be validated accordingly, else it's an embedded custom task and we cannot validate further.

Once we implement PIP, we can add `Pipeline` to the list types that we validate.

abayer: Also, the current behavior, with custom tasks enabled, with `apiVersion`/`kind` and `TaskSpec.TaskSpec` populated results in a `Run` getting created and never doing anything, so we're not breaking any existing use cases by being strict.

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -321,6 +321,26 @@ func (pt *PipelineTask) validateMatrixCombinationsCount(ctx context.Context) (er
 	return errs
 }
 
+func (pt PipelineTask) validateEmbeddedOrType() (errs *apis.FieldError) {
+	// Reject cases where APIVersion and/or Kind are specified alongside an embedded Task.
+	// We determine if this is an embedded Task by checking of TaskSpec.TaskSpec.Steps has items.
+	if pt.TaskSpec != nil && len(pt.TaskSpec.TaskSpec.Steps) > 0 {
```
lbernick, line 327: why is checking if there are steps necessary? can we just check `if pt.TaskSpec.TaskSpec != nil`?

lbernick (commented):
thanks @abayer! Could you document that the api version of a task embedded within a pipeline should match the api version of the pipeline?

pkg/apis/pipeline/v1beta1/pipeline_types.go
abayer: `pt.TaskSpec.TaskSpec` is just an inline `TaskSpec`, not a pointer, so this was the best I could come up with.

abayer:
@lbernick Not sure about what to document in that regard - what this change is doing is making specifying `apiVersion` and/or `kind` and a populated `pt.TaskSpec.TaskSpec` invalid. So assuming this goes in, you won't be able to specify a `Task`'s `apiVersion` in a `PipelineTask` at all.

abayer:
@lbernick 👍 - in practice, that's obviously just clarifying what's always been the case, but still worth having. =)

PR #5028, opened 2022-06-23
Cleanup unused arguments - mostly removing unused ctx 🧹

jerop:
# Changes

In this change, we cleanup unused arguments in functions, mostly removing unused "ctx".

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/param_types.go
+++ b/pkg/apis/pipeline/v1beta1/param_types.go
@@ -61,7 +61,7 @@ type PropertySpec struct {
 }
 
 // SetDefaults set the default type
-func (pp *ParamSpec) SetDefaults(ctx context.Context) {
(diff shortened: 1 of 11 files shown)


vdemeester (commented):
The main reason it exists on `SetDefaults` for example, is to make is `Defaultable`. This is not necessary useful for non CRD types though, but I tend to like the consistency. But.. I also do not like when there is function param that are not used 😛

jerop:
> The main reason it exists on `SetDefaults` for example, is to make is `Defaultable`. This is not necessary useful for non CRD types though, but I tend to like the consistency. But.. I also do not like when there is function param that are not used 😛

@vdemeester `SetDefaults` still has `context.Context` - it's just that it's not named because it's not used -- did the same thing for `Validate` to make it `Validatable`

Issue #4446, opened 2021-12-31
CSI volumes as workspaces


emenaceb:
### Feature request

Support kubernetes CSI volumes (https://kubernetes.io/docs/concepts/storage/volumes/#csi) as workspaces

### Use case

My main motivation is to be able to use Secrets Store CSI Driver (https://secrets-store-csi-driver.sigs.k8s.io/introduction.html)
to use Hashicorp Vault (or any other secret provider) . 

I want to be able to define a workspace in the following way:

 ```
spec:
  workspaces:
    - name: my-credentials
      csi:
        driver: secrets-store.csi.k8s.io
        readOnly: true
        volumeAttributes:
          secretProviderClass: "vault-database"
```

Of course, there are other ways to mount secrets into the task's pods, but this way we can define tasks in a standard way (using workspaces) and select later the source of those workspaces.

PR #4825, opened 2022-05-03
WIP: Add support for using CSIVolumeSource as workspaces

vdemeester:
# Changes

This adds support for using kubernetes CSI
volumes (https://kubernetes.io/docs/concepts/storage/volumes/#csi) as
workspaces.

This opens a lot of possibility as it would enable users to use
alternative volumes for their workspace. An example is the Secrets
Store CSI
Driver (https://secrets-store-csi-driver.sigs.k8s.io/introduction.html)
to use Hashicorp Vault.

It could also open doors for Tekton to ship its own CSI driver to
share data between tasks without the need of a PVC.

Signed-off-by: Vincent Demeester <vdemeest@redhat.com>

/kind feature
/cc @tektoncd/core-maintainers @imjasonh @abayer

Closes #4446 

This still needs
- [ ] Figure out if I need to write a TEP or not 🙏🏼 
- [ ] Tests
- [ ] Docs

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ ] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been filled in
(if there are no user facing changes, use release note "NONE")

# Release Notes

```release-note
Add support to use any CSI volume driver as a workspace
```

PR #5030, opened 2022-06-23
Add support for the CSI workspace type

JeromeJu:
# Changes

Adds support for [CSI volumes ](https://kubernetes.io/docs/concepts/storage/volumes/#csi) as a workspace type.

This opens a lot of possibility as it would enable users to use alternative volumes for their workspace. An example is the Secrets Store [CSI Driver](https://secrets-store-csi-driver.sigs.k8s.io/introduction.html) to use Hashicorp Vault.

Thanks to #4825 as guidance.
Signed-off-by: JeromeJu jeromeju@google.com

/kind feature
/cc @dibyom @vdemeester

Fixes: #4446 

TODO:
- [ ] Examples
- [ ] Docs

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Add support to use any CSI volume driver as a workspace
```


pkg/apis/pipeline/v1beta1/workspace_validation.go
```diff
@@ -66,6 +66,11 @@ func (b *WorkspaceBinding) Validate(ctx context.Context) *apis.FieldError {
 		return apis.ErrMissingField("secret.secretName")
 	}
 
+	// For a CSI to work, you must provide the driver to use.
+	if b.CSI != nil && b.CSI.Driver == "" {
+		return apis.ErrMissingField("csi.drive")
```
dibyom, line 71:
```suggestion
		return apis.ErrMissingField("csi.driver")
```

dibyom:
@JeromeJu looks like there is still an extra space in those generated files

examples/v1beta1/pipelineruns/no-ci/csi-workspace.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: tekton.dev/v1beta1
```
lbernick, line 1: Any reason why this is in no-ci?

pkg/apis/pipeline/v1beta1/workspace_types.go
```diff
@@ -77,6 +77,9 @@ type WorkspaceBinding struct {
 	// Secret represents a secret that should populate this workspace.
 	// +optional
 	Secret *corev1.SecretVolumeSource `json:"secret,omitempty"`
+	// CSI (Container Storage Interface) represents ephemeral storage that is handled by certain external CSI drivers (Beta feature).
```
lbernick, line 80: I'm assuming "beta feature" is copied from k8s docs? we will likely want to introduce this as an alpha feature

pkg/apis/pipeline/v1beta1/workspace_validation.go
```diff
@@ -66,6 +66,11 @@ func (b *WorkspaceBinding) Validate(ctx context.Context) *apis.FieldError {
 		return apis.ErrMissingField("secret.secretName")
 	}
 
+	// For a CSI to work, you must provide the driver to use.
```
lbernick, line 69: We'll probably want to put this behind an [alpha feature flag](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features) (and make this clear in the docs)

lbernick (commented):
This is in draft state so I'm guessing it's not meant to work e2e-- but if we're going to introduce support via multiple PRs we should make sure each is self contained (i.e. update the docs to indicate that this isn't fully supported, and make this clear in the commit message). If you do want to add support in a single PR, you'll likely need to make some changes [here](https://github.com/tektoncd/pipeline/blob/main/pkg/workspace/apply.go#L49-L75) (and maybe other places as well)

pkg/apis/pipeline/v1beta1/workspace_types.go
lbernick: Thanks! Here's an [example](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/task_validation.go#L143) of guarding something behind an alpha feature flag.

examples/v1beta1/pipelineruns/no-ci/csi-workspace.yaml
JeromeJu: This would require the respective csi driver to be installed.
(Later we will come up with another hashicorp example with a detailed guide ideally)

pkg/apis/pipeline/v1beta1/workspace_validation_test.go
```diff
@@ -128,11 +144,45 @@ func TestWorkspaceBindingValidateInvalid(t *testing.T) {
...
+				ctx = tc.wc(ctx)
+			}
+			if err := tc.binding.Validate(ctx); err == nil {
 				t.Errorf("expected error for invalid binding but didn't get any!")
 			}
 		})
 	}
 }
+
+func enableAlphaAPIFields(ctx context.Context) context.Context {
```
lbernick, line 177: this helper exists elsewhere, could you use it instead of copy and pasting?

docs/workspaces.md
```diff
@@ -512,6 +512,39 @@ workspaces:
       secretName: my-secret
 ```
 
+##### `csi`
```
lbernick, line 515: could you please document here that this is alpha only?

PR #5043, opened 2022-06-27
Relax result type validation to avoid nightly build failure

Yongxuanzhang:
# Changes
This commit fixes the string result type validation, PR #4779 adds
result type and asumes that the default type should be string via
mutating webhook. PR #4818 adds the validation for this. However,
resources that did already exist in etcd didn't get the default. So this
commit relaxes this validation.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes
```release-note
Relax the result type validation of string type to avoid nightly builds failing
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/result_types.go
+++ b/pkg/apis/pipeline/v1beta1/result_types.go
@@ -25,7 +25,7 @@ type TaskResult struct {
 
 	// Description is a human-readable description of the result
 	// +optional
-	Description string `json:"description"`
+	Description string `json:"description,omitempty"`
 }
 
 // TaskRunResult used to describe the results of a task
--- a/pkg/apis/pipeline/v1beta1/result_validation.go
+++ b/pkg/apis/pipeline/v1beta1/result_validation.go
@@ -31,6 +31,13 @@ func (tr TaskResult) Validate(ctx context.Context) (errs *apis.FieldError) {
 		return errs.Also(ValidateEnabledAPIFields(ctx, "results type", config.AlphaAPIFields))
 	}
 
+	// Resources created before the result. Type was introduced may not have Type set
+	// and should be considered valid
+	if tr.Type == "" {
+		return nil
+	}
+
+	// By default the result type is string
 	if tr.Type != ResultsTypeString {
 		return apis.ErrInvalidValue(tr.Type, "type", fmt.Sprintf("type must be string"))
 	}
--- a/pkg/apis/pipeline/v1beta1/swagger.json
+++ b/pkg/apis/pipeline/v1beta1/swagger.json
@@ -2243,8 +2243,7 @@
       "properties": {
         "description": {
           "description": "Description is a human-readable description of the result",
-          "type": "string",
-          "default": ""
+          "type": "string"
         },
         "name": {
           "description": "Name the given name",
--- a/pkg/apis/pipeline/v1beta1/openapi_generated.go
+++ b/pkg/apis/pipeline/v1beta1/openapi_generated.go
@@ -3977,7 +3977,6 @@ func schema_pkg_apis_pipeline_v1beta1_TaskResult(ref common.ReferenceCallback) c
 					"description": {
 						SchemaProps: spec.SchemaProps{
 							Description: "Description is a human-readable description of the result",
-							Default:     "",
 							Type:        []string{"string"},
 							Format:      "",
 						},
--- a/pkg/apis/pipeline/v1beta1/result_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/result_validation_test.go
@@ -1,0 +1,147 @@
+/*
+Copyright 2019 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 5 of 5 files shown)


pkg/apis/pipeline/v1beta1/result_types.go
```diff
@@ -25,7 +25,7 @@ type TaskResult struct {
 
 	// Description is a human-readable description of the result
 	// +optional
-	Description string `json:"description"`
+	Description string `json:"description,omitempty"`
```
Yongxuanzhang, line 28: @lbernick I check the code before https://github.com/tektoncd/pipeline/pull/4779 and it didn't contain `omitempty`, do you think we should add it or not?

Yongxuanzhang:
@afrittoli

pkg/apis/pipeline/v1beta1/result_validation.go
```diff
@@ -31,9 +31,5 @@ func (tr TaskResult) Validate(ctx context.Context) (errs *apis.FieldError) {
 		return errs.Also(ValidateEnabledAPIFields(ctx, "results type", config.AlphaAPIFields))
 	}
 
-	if tr.Type != ResultsTypeString {
-		return apis.ErrInvalidValue(tr.Type, "type", fmt.Sprintf("type must be string"))
-	}
```
afrittoli, line 36: Oh, this is not what I was expecting, now we will allow for "type" to be anything?
In fact you had to delete this unit test, which is not something we would like to allow.

```yaml
		fields: fields{
			Steps: validSteps,
			Results: []v1beta1.TaskResult{{
				Name:        "MY-RESULT",
				Type:        "wrong",
				Description: "my great result",
			}},
		},
```

The only case we need to allow for is type not being set at all, which might be the case for resources created before type existed.

lbernick: Sorry I missed this, Andrea is right. I think we want something like:
```
if tr.Type == "" {
  return nil
}
if tr.Type != ResultsTypeString {
   return error
}
```

Let's keep the existing test case, and add another one where type is "" that is valid.

dibyom: do we call setDefaults on the result type? One option would be to add the default string type before validation

Yongxuanzhang: > do we call setDefaults on the result type? One option would be to add the default string type before validation

Yes we call setDefaults on the result type and adds `string` as default type. But for existing resources as @afrittoli mentioned it doesn't work.

I will fix it soon!

dibyom: hmm, one way to deal with this is ensure that setDefaults is called even for the existing types. Is there a way we can ensure setDefaults is called for the existing types before we call validation?

dibyom: yup - something like https://github.com/tektoncd/pipeline/blob/067cbb7b71b2d77d5ac6166767b31ee6525447b7/pkg/reconciler/taskrun/taskrun.go#L120-L126

or https://github.com/tektoncd/pipeline/blob/8611c55275955ff4f8a88f57dfd2f0ecf8562470/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L640-L644

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -436,6 +436,54 @@ func TestTaskSpecValidate(t *testing.T) {
 	}
 }
 
+func TestTaskSpecResultsValidate(t *testing.T) {
```
afrittoli, line 439: NIT: I guess this really could be a `result_validation_test.go`, there's nothing relevant to the tests in the steps part oof the `Task` tested.

pkg/apis/pipeline/v1beta1/swagger.json
```diff
@@ -2243,8 +2243,7 @@
       "properties": {
         "description": {
           "description": "Description is a human-readable description of the result",
-          "type": "string",
-          "default": ""
+          "type": "string"
```
afrittoli, line 2246: 🎉

pkg/apis/pipeline/v1beta1/task_validation_test.go
Yongxuanzhang: Yeah sure! I was actually hesitant to add a new file(result_validation_test.go) or not. This test is testing from an upper level--testing the taskspec's validation and it will call result validation.

Yongxuanzhang: I can create a new file and put the test there. Maybe in another PR add more tests?

pkg/apis/pipeline/v1beta1/result_validation.go
Yongxuanzhang: So the question is where should we call it to ensure for the existing resources. 🤔 @afrittoli Does this solution sounds good to you? We still can go with current solution, this may be a quick fix.

pkg/apis/pipeline/v1beta1/task_validation_test.go
afrittoli: It would be great to have them in one PR since I need to back-port this on v0.36 an v0.37, but I can also backport the two PRs if you prefer so

Yongxuanzhang: sure, I will add it in this pr!

afrittoli (commented):
Thank you for all your updates!
/lgtm

Yongxuanzhang:
🤔 why the tests get 404 not found

PR #5055, opened 2022-06-29
Add almost-empty v1 package

lbernick:
# Changes
This commit adds a package for v1 and updates the hack directory to generate
code and OpenAPI spec for v1. The v1 package is left as minimal as possible;
however, the code does not build and pass tests without at least one generated client.
Therefore, a Task type is added with only a Description field, with the rest of the
Task spec to be added in a future commit.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


lbernick:
blocked on https://github.com/tektoncd/pipeline/pull/5066

abayer:
Looks good in general, though I'm inherently a bit wary of adding a `v1` api without it being fully populated. But it should be harmless, I guess?

Would like to get thoughts from others from @tektoncd/core-maintainers

vdemeester:
> Looks good in general, though I'm inherently a bit wary of adding a `v1` api without it being fully populated. But it should be harmless, I guess?
> 
> Would like to get thoughts from others from @tektoncd/core-maintainers

@abayer this is just the code and the package, it is not exported or exposed in any way. This is what we did for `v1beta1`.
I think it's fine, except *maybe* the openapi/swagger generation 👼🏼

abayer:
> > Looks good in general, though I'm inherently a bit wary of adding a `v1` api without it being fully populated. But it should be harmless, I guess?
> > Would like to get thoughts from others from @tektoncd/core-maintainers
> 
> @abayer this is just the code and the package, it is not exported or exposed in any way. This is what we did for `v1beta1`. I think it's fine, except _maybe_ the openapi/swagger generation 👼🏼

@lbernick handled that bit nicely - generating `v1beta1` and `v1` OpenAPI and swagger separately from each other. So given that, I think this is good to go. =)

/approve

Issue #3749, opened 2021-02-04
Results for failed task and pipeline runs


afrittoli:
### Feature request

Results should be available for failed task and pipeline runs.

Today, if a `TaskRun` or `PipelineRun` is set to failed, we skip the code that grabs the results and adds them to status.

### Use case

Results can be used by a failing task or pipeline to provide more context about why the failed, or how to react to the failure.

One specific use case that I'm trying to implement, is to enrich the GitHub update pipelines in Tekton based CI with the ability to publish a comment that gives the developer more context about the failure. Today that can achieved by adding that comment into the logs, however a comment on GitHub provides a more immediate feedback to developers. Specifically I'd like to improve the job that checks for the "kind label".

That would work as follow:
- the CI `TaskRun` is executed. 
- if no valid label is found, a message with the list of valid labels is published to a specific result
- the `TaskRun` fails
- a cloud event is sent to the `tekton-events` trigger. In includes the `TaskRun` status in the payload
- the trigger filters the event
- a binding extracts the result from the payload and passes it the trigger template
- the trigger template runs the github update pipeline, which sets the status and adds the comment

/cc @bobcatfish @vdemeester

sbwsg:
I've discussed this before with users in another issue.

Results are required to be populated.  Pipelines fail if they're not.  If you introduce an "errorMessage" result to a Task then the Task has to populate it whether there was a failure or not.  That means:

```bash
# Success!
echo "" > $(results.errorMessage.path)
```

...in any Tasks that want to provide this.

Further, the error message results will be named differently for every single Task. There could be Tasks that have an `errorMessage` result, an `errorContext` result, an `err` result, etc etc.  That means tooling (like the dashboard or CLI) won't have a consistent way to look up these messages / extra context and surface them differently from other results.

I'd like to propose an alternative for your consideration: Provide an explicit `status.errorResult` field (or named anything else) that is explicitly for the purpose of returning context on Task error. Provide a Tekton variable to its path - It can function almost identically to a Result but will be optional for a Task to return.  This leaves the semantics of Results alone but provides a well-defined outlet for this kind of information.

There are other alternatives which could be considered too.  But I want to make sure that the knock-on effects of supporting Results for failure scenarios are well documented before we decide to support this.

sbwsg:
Another alternative could be to explicitly mark certain results as `error-results`?  Or have a separate section like `results` just for them?

afrittoli:
> I've discussed this before with users in another issue.
[...]

PR #5060, opened 2022-06-30
Publish Pipeline Results from successful TaskRuns in Failed PipelineRuns

QuanZhang-William:
# Changes
Publish `Pipeline Results` from successful `TaskRuns` in Failed `PipelineRuns`: https://github.com/tektoncd/pipeline/issues/3749

Prior to this change a failed `PipelineRun` does not emit any `PipelineResult` even when some `PipelineTasks` referenced by the `PipelineResult` succeeded. This commit changes to publish `Pipeline Results` **only** from successful `PipelineTaskRuns` in the failed `PipelineRun`

Such `Results` are valid because they are generated by successful `TaskRuns`. They can provide more context and successful artifacts to users of the partially successful `PipelinRun`.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -4868,6 +4868,164 @@ spec:
 	}
 }
 
+func TestReconcileWithPipelineResults_OnFailedPipelineRun(t *testing.T) {
+	testCases := []struct {
+		name        string
+		embeddedVal string
+	}{
+		{
+			name:        "default embedded status",
+			embeddedVal: config.DefaultEmbeddedStatus,
+		},
+		{
+			name:        "full embedded status",
+			embeddedVal: config.FullEmbeddedStatus,
+		},
+		{
+			name:        "both embedded status",
+			embeddedVal: config.BothEmbeddedStatus,
+		},
+		{
+			name:        "minimal embedded status",
+			embeddedVal: config.MinimalEmbeddedStatus,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			runTestReconcileWithPipelineResultsOnFailedPipelineRun(t, tc.embeddedVal)
+		})
+	}
+}
+
+func runTestReconcileWithPipelineResultsOnFailedPipelineRun(t *testing.T, embeddedStatus string) {
+	names.TestingSeed()
+	ps := []*v1beta1.Pipeline{parse.MustParsePipeline(t, `
+metadata:
+  name: test-pipeline
+  namespace: foo
+spec:
+  results:
+  - description: pipeline result
+    name: result
+    value: $(tasks.a-task.results.aResult)
+  tasks:
+  - name: a-task
+    taskRef:
+      name: a-task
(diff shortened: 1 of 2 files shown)


jerop (commented):
Note that this is solving the first problem only as discussed in the API WG on 06/27:

"Publishing PipelineResults from successful TaskRuns in Failed PipelineRuns"

@QuanZhang-William please update the commit and pull request description to make this clear

jerop (approved):
Thanks @QuanZhang-William!

You've described the "what" of this change in the commit and PR description, could you please add the "why"?https://github.com/tektoncd/community/blob/main/standards.md#commits

from my perspective, the reason we're making this change is that the results are valid because they are from successful taskruns and the pipelinerun results are only propagating taskrun results (pipelineruns don't produce results themselves)

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -4868,6 +4868,161 @@ spec:
...
+spec:
+  resources: {}
+  serviceAccountName: test-sa
+  taskRef:
+     name: hello-world
+  timeout: 1h0m0s
+status:
+  conditions:
+  - status: "False"
+    type: Succeeded
```
lbernick, line 4953: Looks like this is the failed taskrun -- could you add some results to this one and verify that they aren't propagated to the failed pipelinerun?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -4868,6 +4868,161 @@ spec:
...
+		PipelineRuns: prs,
+		Pipelines:    ps,
+		Tasks:        ts,
+		TaskRuns:     trs,
+		ConfigMaps:   []*corev1.ConfigMap{withEmbeddedStatus(newFeatureFlagsConfigMap(), embeddedStatus)},
+	}
+	prt := newPipelineRunTest(d, t)
+	defer prt.Cancel()
+
+	reconciledRun, _ := prt.reconcileRun("foo", "test-pipeline-run-different-service-accs", []string{}, false)
```
lbernick, line 5016: It looks like some of this test data may be copied from other test cases. Could you change it so it better reflects what this test case is trying to test? for example, "test-failed-pr-with-task-results"

QuanZhang-William: Good catch, I have updated the name to "test-failed-pr-with-task-results"

pkg/reconciler/pipelinerun/pipelinerun_test.go
pritidesai: Failed taskruns does not see results initialized at all.

pritidesai:
I am very excited for this feature, thank you @QuanZhang-William for implementing this 🙏 

@jerop, @lbernick please make sure to have the release notes updated for PRs implementing features. This is a direct user facing change, the users will start seeing pipelineResults even when their pipeline fails. @jerop please make sure to capture this in the `0.38` release notes.

pritidesai:
I updated the comment in this PR description!

PR #5061, opened 2022-06-30
TEP-0090: Matrix - implement `getNextTasks`

jerop:
# Changes

This change implements the `getNextTasks` which is a member function of `PipelineRunState`. This function is used to get the `PipelineTasks` that should be executed yet, considering some candidate `PipelineTasks`.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
@@ -121,6 +121,48 @@ var pts = []v1beta1.PipelineTask{{
 	Name:    "mytask15",
 	TaskRef: &v1beta1.TaskRef{Name: "taskWithReferenceToTaskResult"},
 	Params:  []v1beta1.Param{{Name: "param1", Value: *v1beta1.NewArrayOrString("$(tasks.mytask1.results.result1)")}},
+}, {
+	Name: "mytask16",
+	Matrix: []v1beta1.Param{{
+		Name:  "browser",
+		Value: v1beta1.ArrayOrString{ArrayVal: []string{"safari", "chrome"}},
+	}},
+}, {
+	Name: "mytask17",
+	Matrix: []v1beta1.Param{{
+		Name:  "browser",
+		Value: v1beta1.ArrayOrString{ArrayVal: []string{"safari", "chrome"}},
+	}},
+}, {
+	Name:    "mytask18",
+	TaskRef: &v1beta1.TaskRef{Name: "task"},
+	Retries: 1,
+	Matrix: []v1beta1.Param{{
+		Name:  "browser",
+		Value: v1beta1.ArrayOrString{ArrayVal: []string{"safari", "chrome"}},
+	}},
+}, {
+	Name:    "mytask19",
+	TaskRef: &v1beta1.TaskRef{APIVersion: "example.dev/v0", Kind: "Example", Name: "customtask"},
+	Matrix: []v1beta1.Param{{
+		Name:  "browser",
+		Value: v1beta1.ArrayOrString{ArrayVal: []string{"safari", "chrome"}},
+	}},
+}, {
+	Name:    "mytask20",
+	TaskRef: &v1beta1.TaskRef{APIVersion: "example.dev/v0", Kind: "Example", Name: "customtask"},
+	Matrix: []v1beta1.Param{{
+		Name:  "browser",
+		Value: v1beta1.ArrayOrString{ArrayVal: []string{"safari", "chrome"}},
+	}},
+}, {
+	Name:    "mytask21",
+	TaskRef: &v1beta1.TaskRef{Name: "task"},
+	Retries: 2,
+	Matrix: []v1beta1.Param{{
+		Name:  "browser",
+		Value: v1beta1.ArrayOrString{ArrayVal: []string{"safari", "chrome"}},
+	}},
 }}
 
 var p = &v1beta1.Pipeline{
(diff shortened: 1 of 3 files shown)


jerop:
cc @abayer @dibyom @lbernick please take a look

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -315,19 +315,43 @@ func (state PipelineRunState) getRetryableTasks(candidateTasks sets.String) []*R
...
+				isCancelled = false
+				for _, taskRun := range t.TaskRuns {
+					isDone = isDone && taskRun.IsDone()
+					c := taskRun.Status.GetCondition(apis.ConditionSucceeded)
+					if c.IsFalse() {
+						status = c
+						isCancelled = isCancelled || c.Reason == v1beta1.TaskRunReasonCancelled.String()
+					}
+				}
+				isCancelled = isCancelled && isDone
```
lbernick, line 336: I'm a bit confused about what's going on here. Does this mean a matrixed pipeline task is considered cancelled only if any taskrun is cancelled, and all of them are done? It looks like from the test cases we don't want to schedule any new matrixed tasks if any of them are cancelled. Could you update the docstring of this function to reflect how it works with matrix?

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -498,6 +540,182 @@ var runCancelled = PipelineRunState{{
...
+}, {
+	PipelineTask: &pts[16],
+	TaskRunNames: []string{"pipelinerun-mytask2"},
+	TaskRuns:     []*v1beta1.TaskRun{makeScheduled(trs[1])},
+	ResolvedTaskResources: &resources.ResolvedTaskResources{
+		TaskSpec: &task.Spec,
+	},
+}}
+
+var retryableFinalStateMatrix = PipelineRunState{{
```
lbernick, line 623: is this used anywhere? are there test cases covering remaining retries?

jerop: good catch - removed this unused variable 

yes, the test cases for retries are defined with the tests for retries scenarios - see `TestGetNextTasks`

jerop:
something weird is happening with tekton-robot - it approves PRs when commits are pushed then remove them 🤯 

saw this happen here too: https://github.com/tektoncd/pipeline/pull/5062

flagging this in case it has anything to do with the infra migration happening @afrittoli @abayer @dibyom

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
jerop: @lbernick that's correct - if a pipelinetask has a matrix, iscancelled returns true if any run is cancelled and all other runs are done - this is also aligned with the iscancelled member function of pipelinetask:

https://github.com/tektoncd/pipeline/blob/8a3a8e5070fa99e157999e50c8fbd4711c786a5c/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L237-L280

updated the docstring to reflect this

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
```diff
@@ -309,25 +309,50 @@ func (state PipelineRunState) getNextTasks(candidateTasks sets.String) []*Resolv
 
 // getRetryableTasks returns a list of tasks which should be executed next when the pipelinerun is stopping, i.e.
 // a list of cancelled/failed tasks from candidateTasks which haven't exhausted their retries
+// If the PipelineTask has a Matrix, isCancelled returns true if any run is cancelled and all other runs are done.
```
lbernick, line 312: sorry to keep pushing on this-- this reads more like a docstring for `isCancelled` which makes me wonder if there is some overlap between those two functions?

To clarify, when the pipelinerun is in stopping mode, for a non-matrixed taskrun it will only schedule it if it's cancelled/failed and has remaining retries. For a matrixed taskrun, it will schedule a taskrun if it is cancelled or failed with remaining retries **only if all other matrixed taskruns are done** (which includes being cancelled)?

If any matrixed taskrun is cancelled, does that mean all the others will be cancelled? It seems to me like we should be scheduling a matrixed taskrun in stopping mode if it's cancelled/failed with remaining retries, regardless of the state of the other matrixed taskruns.

lbernick (commented):
/hold

just want to get clarity on this before it is merged

pkg/reconciler/pipelinerun/resources/pipelinerunstate.go
jerop: You were right, switched it to using `t.isCancelled()` instead of repeating the same logic.

The docstring wasn't reflecting what was happening, even for non-matrixed cases.  This function returns a list of pipelinetasks which should be executed next when the pipelinerun is stopping i.e. a list of failed pipelinetasks from candidateTasks which haven't exhausted their retries. If a pipelinetask is cancelled, the retries are not exhausted - they are not retryable.

For matrixed cases, isCancelled returns true if any run is cancelled and all other runs are done. Cancelling one taskrun from a matrix doesn't cancel the others, just the same way that one failing doesn't fail the others.

lbernick: nice thanks this looks better!

PR #5081, opened 2022-07-05
TEP-0111 - Propagating workspaces in taskruns

chitrangpatel:
The verbosity of writing specifications in Tekton Pipelines is a common pain point that creates difficulties in getting-started scenarios. In addition, the verbosity leads to long specifications that are error-prone, harder to maintain, and reach the etcd size limits for CRDs. Automatically propagating Workspaces will lead to better user experience. This addresses features proposed in [TEP-0111](https://github.com/tektoncd/community/blob/main/teps/0111-propagating-workspaces.md). This PR addresses propagating workspaces in `TaskRuns`. A follow up [PR](https://github.com/tektoncd/pipeline/pull/5463) is addressing propagating workspaces in `PipelineRuns`. 

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes
```release-note
Workspaces are propagated in embedded specifications without mutations.
```


docs/pipelineruns.md
```diff
@@ -844,6 +847,518 @@ For more information, see the following topics:
 [`Custom tasks`](pipelines.md#using-custom-tasks) may or may not use workspaces.
 Consult the documentation of the custom task that you are using to determine whether it supports workspaces.
 
+#### Propagated Workspaces
+
+**([alpha only](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features))**
```
jerop, line 852: could we also list this feature in https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features?

examples/v1beta1/pipelineruns/alpha/propagating-workspaces.yaml
```diff
@@ -0,0 +1,76 @@
...
+        items:
+        - key: brownies
+          path: recipe.txt
+    - name: password-vault
+      secret:
+        secretName: secret-password
+  pipelineSpec:
+    tasks:
+    - name: fetch-secure-data
+      taskSpec:
```
jerop, line 57: please add another example with embedded specifications with references -- the last example in the docs

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -66,8 +68,10 @@ func (ps *PipelineSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	errs = errs.Also(validateExecutionStatusVariables(ps.Tasks, ps.Finally))
 	// Validate the pipeline's workspaces.
 	errs = errs.Also(validatePipelineWorkspacesDeclarations(ps.Workspaces))
-	errs = errs.Also(validatePipelineWorkspacesUsage(ps.Workspaces, ps.Tasks).ViaField("tasks"))
-	errs = errs.Also(validatePipelineWorkspacesUsage(ps.Workspaces, ps.Finally).ViaField("finally"))
+	if cfg.FeatureFlags.EnableAPIFields != config.AlphaAPIFields {
+		errs = errs.Also(validatePipelineWorkspacesUsage(ps.Workspaces, ps.Tasks).ViaField("tasks"))
+		errs = errs.Also(validatePipelineWorkspacesUsage(ps.Workspaces, ps.Finally).ViaField("finally"))
+	}
```
jerop, line 74: it'd be cleaner if this check is done in `validatePipelineWorkspacesUsage`

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -7464,11 +7465,62 @@ spec:
 			},
 			expectedError: `expected workspace "not-source" to be provided by pipelinerun for pipeline task "resolved-pipelinetask"`,
 		},
+		{
+			name: "failure propagating workspaces using scripts",
```
jerop, line 7469: could you please use yamls for the tests, it'd be easier to read through it 🙏🏾 

this would involve updating the other 2 test cases that already exist in this function, but it's worth switching over imo

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -7520,10 +7573,60 @@ spec:
 				},
 			},
 		},
+		{
+			name: "propagating workspaces using scripts",
```
jerop, line 7577: same as above regarding using yamls

examples/v1beta1/pipelineruns/alpha/propagating-workspaces.yaml
chitrangpatel: sure!

pkg/apis/pipeline/v1beta1/pipeline_validation.go
chitrangpatel: Can do!

pkg/reconciler/pipelinerun/pipelinerun_test.go
chitrangpatel: Yes, I will do that.

docs/pipelineruns.md
chitrangpatel: sure

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -7420,10 +7420,10 @@ func checkPipelineRunConditionStatusAndReason(t *testing.T, reconciledRun *v1bet
 }
 
 func TestGetTaskrunWorkspaces_Failure(t *testing.T) {
+	ctx := getContextBasedOnFeatureFlag("alpha")
```
dibyom, line 7423: Move this line closer to where its being used

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -2462,8 +2463,13 @@ spec:
 		Kind:     "Task",
 		TaskSpec: &v1beta1.TaskSpec{Steps: simpleTask.Spec.Steps, Workspaces: simpleTask.Spec.Workspaces},
 	}
-	taskSpec := updateTaskSpecParamsContextsResults(context.Background(), taskRun, rtr)
-	pod, err := r.createPod(testAssets.Ctx, taskSpec, taskRun, rtr)
+	ctx := getContextBasedOnFeatureFlag("alpha")
+	workspaceVolumes := workspace.CreateVolumes(taskRun.Spec.Workspaces)
```
dibyom, line 2467: Hmm, the fact the we have to have so much logic here in this test isn't great and indicates we might not be testing this at the right level. Wondering if there are any alternatives here e.g. could we call r.reconcile and check that the created pod has the right volumeMounts? Alternatively, we can group this logic into its own function and test that? What do you think?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -2566,8 +2572,13 @@ spec:
 		TaskSpec: &v1beta1.TaskSpec{Steps: simpleTask.Spec.Steps, Workspaces: simpleTask.Spec.Workspaces},
 	}
 
-	taskSpec := updateTaskSpecParamsContextsResults(context.Background(), taskRun, rtr)
-	_, err := r.createPod(testAssets.Ctx, taskSpec, taskRun, rtr)
+	ctx := getContextBasedOnFeatureFlag("alpha")
+	workspaceVolumes := workspace.CreateVolumes(taskRun.Spec.Workspaces)
```
dibyom, line 2576: same comment as above

pkg/substitution/substitution.go
```diff
@@ -234,7 +234,8 @@ func extractExpressionFromString(s, prefix string) (string, bool) {
 	return match[0], true
 }
 
-func extractVariablesFromString(s, prefix string) ([]string, bool, string) {
+// ExtractVariablesFromString extracts variables from an input string via regex matching.
```
dibyom, line 237:
```suggestion
// ExtractVariablesFromString extracts variables from an input string s with the given prefix via regex matching.
```

pkg/workspace/validate.go
```diff
@@ -53,7 +54,8 @@ func ValidateBindings(ctx context.Context, decls []v1beta1.WorkspaceDeclaration,
 		}
 	}
 	for _, bind := range binds {
-		if !declNames.Has(bind.Name) {
+		if !declNames.Has(bind.Name) && config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields != "alpha" {
```
dibyom, line 57:
```suggestion
		if !declNames.Has(bind.Name) && config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields != config.AlphaAPIFields {
```

pkg/workspace/validate.go
```diff
@@ -53,7 +54,8 @@ func ValidateBindings(ctx context.Context, decls []v1beta1.WorkspaceDeclaration,
 		}
 	}
 	for _, bind := range binds {
```
dibyom, line 56: Let's add a comment that mentions why we are checking for whether alpha fields are enabled or not

pkg/workspace/validate_test.go
```diff
@@ -21,11 +21,13 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 )
 
 func TestValidateBindingsValid(t *testing.T) {
+	ctx := getContextBasedOnFeatureFlag("alpha")
```
dibyom, line 30: same as above - move closer to usage

pkg/workspace/validate_test.go
```diff
@@ -236,3 +243,14 @@ func TestValidateOnlyOnePVCIsUsed_Invalid(t *testing.T) {
 		})
 	}
 }
+
+func getContextBasedOnFeatureFlag(featureFlag string) context.Context {
```
dibyom, line 247: Hmm this function keeps being copied around. Might want a function that takes in a context and returns another one with Alpha set e.g. withAlphaEnabled(ctx)

dibyom: (could be in a different PR)

examples/v1beta1/pipelineruns/alpha/propagating_workspaces_with_referenced_resources.yaml
```diff
@@ -0,0 +1,88 @@
+# PipelineRun attempting to propagate Workspaces to referenced Tasks
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: sensitive-recipe-storage
+data:
+  brownies: |
+    1. Heat oven to 325 degrees F
```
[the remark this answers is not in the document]
chitrangpatel: That's a very good point. Thanks! I will make the necessary changes.

chitrangpatel:
@jerop This is ready for review again.
@dibyom please take a look again as there were a few changes (I incorporated your suggestions as best as I could) after bug fixes related to propagating parameters.

chitrangpatel:
I will work on improving the coverage

pkg/substitution/substitution.go
```diff
@@ -234,7 +234,8 @@ func extractExpressionFromString(s, prefix string) (string, bool) {
 	return match[0], true
 }
 
-func extractVariablesFromString(s, prefix string) ([]string, bool, string) {
+// ExtractVariablesFromString extracts variables from an input string s with the given prefix via regex matching.
+func ExtractVariablesFromString(s, prefix string) ([]string, bool, string) {
```
jerop, line 238: this function is now exported so we need to add unit tests for it

chitrangpatel: Good catch. I will do that.

chitrangpatel: Done now!!

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1013,16 +1014,56 @@ func (c *Reconciler) createRun(ctx context.Context, runName string, params []v1b
 	return c.PipelineClientSet.TektonV1alpha1().Runs(pr.Namespace).Create(ctx, r, metav1.CreateOptions{})
 }
 
-func getTaskrunWorkspaces(pr *v1beta1.PipelineRun, rpt *resources.ResolvedPipelineTask) ([]v1beta1.WorkspaceBinding, string, error) {
+func getTaskrunWorkspaces(ctx context.Context, pr *v1beta1.PipelineRun, rprt *resources.ResolvedPipelineTask) ([]v1beta1.WorkspaceBinding, string, error) {
```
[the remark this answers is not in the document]
chitrangpatel: Oh I think this was `rprt` earlier. It was changed to `rpt` during some other PR and during rebasing with `upstream/main`, I accidentaly kept it as `rprt` instead of changing it to `rpt`.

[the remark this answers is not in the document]
chitrangpatel: I will fix it to be consistent.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -926,16 +927,56 @@ func (c *Reconciler) createRun(ctx context.Context, runName string, params []v1b
...
 	var workspaces []v1beta1.WorkspaceBinding
 	var pipelinePVCWorkspaceName string
 	pipelineRunWorkspaces := make(map[string]v1beta1.WorkspaceBinding)
 	for _, binding := range pr.Spec.Workspaces {
 		pipelineRunWorkspaces[binding.Name] = binding
 	}
-	for _, ws := range rpt.PipelineTask.Workspaces {
-		taskWorkspaceName, pipelineTaskSubPath, pipelineWorkspaceName := ws.Name, ws.SubPath, ws.Workspace
 
+	if config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields == "alpha" {
```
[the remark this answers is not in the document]
chitrangpatel: done.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -764,7 +766,30 @@ func updateTaskSpecParamsContextsResults(ctx context.Context, tr *v1beta1.TaskRu
 	// Apply step exitCode path substitution
 	ts = resources.ApplyStepExitCodePath(ts)
 
-	return ts
+	// Apply workspace resource substitution
+	if config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields == "alpha" {
```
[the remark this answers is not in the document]
chitrangpatel: done.

docs/pipelineruns.md
```diff
@@ -844,6 +847,518 @@ For more information, see the following topics:
...
+            script: cat /workspace/shared-data/recipe.txt
+          workspaces:
+          - name: shared-data
+```
+
+##### Workspace Referenced Resources
+
+The propagation of `Workspaces` will not work for `referenced specifications` because the behavior would become opaque when users can't see the relationship between `Workspaces` declared in the referenced resources and the `Workspaces` supplied in runtime resources. Attempting to propagate `Workspaces` to referenced resources will cause failures.
+
+```yaml
```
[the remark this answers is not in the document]
chitrangpatel: done.

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
chitrangpatel: done.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -926,16 +927,56 @@ func (c *Reconciler) createRun(ctx context.Context, runName string, params []v1b
...
+		// Propagate required workspaces from pipelineRun to the pipelineTasks
+		if rprt.PipelineTask.TaskSpec != nil {
```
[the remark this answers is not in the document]
chitrangpatel: Sounds good! I think that's a good idea. Will do.

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
jerop: forgot to push the change? 🤔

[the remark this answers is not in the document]
chitrangpatel: No, I was fixing a few more things along the way. I pushed the change. I still have some slightly major fixes to do from @dibyom 's suggestions. I suggest reviewing after I fix those too.

jerop:
`authenticating-git-commands` 

docs/pipelineruns.md
```diff
@@ -844,6 +847,518 @@ For more information, see the following topics:
...
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  generateName: recipe-time-
+  ...
+spec:
+  params:
+  - name: filename
+    value: recipe.txt
+  pipelineSpec:
```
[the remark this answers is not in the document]
chitrangpatel: done!

pkg/substitution/substitution.go
```diff
@@ -7,7 +7,7 @@ You may obtain a copy of the License at
 
     http://www.apache.org/licenses/LICENSE-2.0
 
-Unless required by applicable law or agreed to in writing, software
+Unless requirned by applicable law or agreed to in writing, software
```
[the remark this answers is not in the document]
chitrangpatel: fixed.

pkg/reconciler/taskrun/taskrun_test.go
chitrangpatel: Sure.

pkg/reconciler/taskrun/taskrun_test.go
chitrangpatel: Sure.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -10691,3 +10787,14 @@ spec:
...
+func getContextBasedOnFeatureFlag(featureFlag string) context.Context {
+	featureFlags, _ := config.NewFeatureFlagsFromMap(map[string]string{
+		"enable-api-fields": featureFlag,
+	})
+	cfg := &config.Config{
+		FeatureFlags: featureFlags,
+	}
+
+	return config.ToContext(context.Background(), cfg)
+}
```
jerop, line 10800: instead of creating this new function, could you reuse `withEnabledAlphaFields`?

https://github.com/tektoncd/pipeline/blob/5188483f1d2d9fb34d0df910c1e84be30676f941/pkg/reconciler/pipelinerun/pipelinerun_test.go#L1493-L1497

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4908,3 +4858,14 @@ status:
...
+func getContextBasedOnFeatureFlag(featureFlag string) context.Context {
+	featureFlags, _ := config.NewFeatureFlagsFromMap(map[string]string{
+		"enable-api-fields": featureFlag,
+	})
+	cfg := &config.Config{
+		FeatureFlags: featureFlags,
+	}
+
+	return config.ToContext(context.Background(), cfg)
+}
```
jerop, line 4871: same as above on reusing https://github.com/tektoncd/pipeline/blob/5188483f1d2d9fb34d0df910c1e84be30676f941/pkg/reconciler/pipelinerun/pipelinerun_test.go#L1493-L1497

chitrangpatel: done

pkg/reconciler/pipelinerun/pipelinerun_test.go
chitrangpatel: done

pkg/apis/pipeline/v1/task_types.go
```diff
@@ -95,6 +96,46 @@ type TaskSpec struct {
...
+		}
+
+		for i := range step.Command {
+			locationsToCheck.Insert(step.Command[i])
+		}
+	}
+	return locationsToCheck
+}
+
+// FindWorkspaceSubstitutionLocationsInTask returns an array of all the locations in the TaskSpec  where we can apply substitutions.
```
jerop, line 131:
```suggestion
// FindWorkspaceSubstitutionLocationsInTask returns a set of all the locations in the TaskSpec where we can apply substitutions for Workspace variables.
```

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1013,16 +1014,61 @@ func (c *Reconciler) createRun(ctx context.Context, runName string, params []v1b
 	return c.PipelineClientSet.TektonV1alpha1().Runs(pr.Namespace).Create(ctx, r, metav1.CreateOptions{})
 }
 
-func getTaskrunWorkspaces(pr *v1beta1.PipelineRun, rpt *resources.ResolvedPipelineTask) ([]v1beta1.WorkspaceBinding, string, error) {
+func propagateWorkspaces(rpt *resources.ResolvedPipelineTask) (*resources.ResolvedPipelineTask, error) {
+	workspacesUsedInSteps := []string{}
```
jerop, line 1018: can this be a set? it'll make it easier to check "contains" in line 1036 below

pkg/apis/pipeline/v1/task_types.go
chitrangpatel: Thanks

pkg/reconciler/pipelinerun/pipelinerun.go
chitrangpatel: Sure.

docs/pipelineruns.md
```diff
@@ -844,6 +847,378 @@ For more information, see the following topics:
...
+          - image: ubuntu
+            name: fetch-and-write
+            resources: {}
+            script: | # See below: Replacements do not happen.
+      	      echo hi >> $(workspaces.shared-data.path)/recipe.txt
+```
+
+##### Embedded Specifications with Referenced Resources
+
+In the case of embedded specifications that have references, the `Workspaces` will be propagated up to the embedded specification level only. Users need to explicitly declare `Workspaces` wherever they reference resources.
```
lbernick, line 1077:
```suggestion
#### Referenced TaskRuns within Embedded PipelineRuns

As mentioned in the [previous section](link), workspaces can only be propagated from PipelineRuns to embedded Pipeline specs, not Pipeline references. Similarly, workspaces can only be propagated from a Pipeline to embedded Task specs, not referenced Tasks. For example:
```

docs/pipelineruns.md
```diff
@@ -844,6 +847,378 @@ For more information, see the following topics:
...
+  - name: recipe-store
+    configMap:
+      name: sensitive-recipe-storage
+      items:
+      - key: brownies
+        path: recipe.txt
+  - name: shared-data
+    persistentVolumeClaim:
+      claimName: shared-task-storage
+  pipelineSpec:
```
lbernick, line 1119: Can this example be simplified at all? Maybe just one task in the Pipeline, which is referenced, and only one workspace.

examples/v1beta1/taskruns/alpha/propagating_workspaces.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: tekton.dev/v1beta1
```
lbernick, line 1: Can you please update tasks.md with some documentation on propagating taskrun workspaces to embedded tasks?

docs/pipelineruns.md
```diff
@@ -844,6 +847,378 @@ For more information, see the following topics:
...
 
+#### Propagated Workspaces
+
+**([alpha only](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features))**
+
+When using an embedded spec, workspaces from the parent `PipelineRun` will be
+propagated to any inlined specs without needing to be explicitly defined. This
+allows authors to simplify specs by automatically propagating top-level
+workspaces down to other inlined resources.
+**Workspace substutions will only be made for `commands`, `args` and `script` fields of the `steps`.**
```
lbernick, line 858:
```suggestion
**Workspace substutions will only be made for `commands`, `args` and `script` fields of `steps`, `stepTemplates`, and `sidecars`.**
```

pkg/apis/pipeline/v1/task_types_test.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+func TestTaskSpec_FindWorkspaceSubstitutionLocationsInTask(t *testing.T) {
+	tests := []struct {
+		name string
+		ts   *v1.TaskSpec
+		want sets.String
+	}{{
+		name: "completespec",
+		ts: &v1.TaskSpec{
```
lbernick, line 35: should this handle step template as well?

lbernick (commented):
Two overall comments:
- This PR is very large; it might make sense to split into one PR propagating taskrun workspaces -> embedded tasks and a follow-up one propagating pipelinerun workspaces -> embedded pipeline specs
- Should workspaces be propagated to custom tasks as well?

docs/pipelineruns.md
```diff
@@ -844,6 +847,378 @@ For more information, see the following topics:
...
+      # workspaces:
+      #   - name: shared-data 
+      taskSpec:
+        # workspaces:
+        #   - name: shared-data 
+        steps:
+        - name: fetch-and-write-secure
+          image: ubuntu
+          script: |
+	    echo hi >> $(workspaces.shared-data.path)/recipe.txt
```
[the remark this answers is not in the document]
chitrangpatel: Thanks! I will fix this.

docs/pipelineruns.md
```diff
@@ -844,6 +847,378 @@ For more information, see the following topics:
...
+```
+
+##### Workspace Referenced Resources
+
+The propagation of `Workspaces` will not work for `referenced specifications` because the behavior would become opaque when users can't see the relationship between `Workspaces` declared in the referenced resources and the `Workspaces` supplied in runtime resources. Attempting to propagate `Workspaces` to referenced resources will cause failures.
+
+```yaml
+# PipelineRun attempting to propagate Workspaces to referenced Tasks
+apiVersion: v1
+kind: PersistentVolumeClaim
```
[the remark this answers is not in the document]
chitrangpatel: will do!

docs/pipelineruns.md
```diff
@@ -844,6 +847,378 @@ For more information, see the following topics:
...
+            name: print-secrets
+            resources: {}
+            script: cat /workspace/shared-data/recipe.txt
+          workspaces:
+          - name: shared-data
+```
+
+##### Workspace Referenced Resources
+
+The propagation of `Workspaces` will not work for `referenced specifications` because the behavior would become opaque when users can't see the relationship between `Workspaces` declared in the referenced resources and the `Workspaces` supplied in runtime resources. Attempting to propagate `Workspaces` to referenced resources will cause failures.
```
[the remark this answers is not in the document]
chitrangpatel: Thanks

docs/pipelineruns.md
```diff
@@ -844,6 +847,378 @@ For more information, see the following topics:
...
+kind: PipelineRun
+metadata:
+  generateName: recipe-time-
+spec:
+  pipelineRef:
+    name: fetch-and-print-recipe
+  workspaces:
+  - name: shared-data
+    persistentVolumeClaim:
+      claimName: shared-task-storage
```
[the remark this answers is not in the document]
chitrangpatel: Agreed. I will fix that.

docs/pipelineruns.md
chitrangpatel: Thanks!

docs/pipelineruns.md
chitrangpatel: Thanks!

docs/pipelineruns.md
chitrangpatel: Yes. I can do that.

pkg/apis/pipeline/v1/task_types_test.go
chitrangpatel: Yes. Will fix. Thanks!

chitrangpatel:
> Two overall comments:
> 
> * This PR is very large; it might make sense to split into one PR propagating taskrun workspaces -> embedded tasks and a follow-up one propagating pipelinerun workspaces -> embedded pipeline specs

Sure, I think that's a great idea. 

> * Should workspaces be propagated to custom tasks as well?

Good question. As of now TEP https://github.com/tektoncd/community/blob/main/teps/0111-propagating-workspaces.md did not cover that. If we want to then we should first update the TEP and then implement it.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -750,7 +766,7 @@ func (c *Reconciler) createPod(ctx context.Context, ts *v1beta1.TaskSpec, tr *v1
 	return pod, err
 }
 
-func updateTaskSpecParamsContextsResults(ctx context.Context, tr *v1beta1.TaskRun, rtr *resources.ResolvedTaskResources) *v1beta1.TaskSpec {
+func updateTaskSpecParamsContextsResults(ctx context.Context, tr *v1beta1.TaskRun, rtr *resources.ResolvedTaskResources, workspaceVolumes map[string]corev1.Volume) (*v1beta1.TaskSpec, error) {
```
lbernick, line 769: could you please add a docstring for this function? It probably also should be renamed

pkg/substitution/substitution.go
```diff
@@ -50,7 +50,7 @@ var intIndexRegex = regexp.MustCompile(intIndex)
 
 // ValidateVariable makes sure all variables in the provided string are known
 func ValidateVariable(name, value, prefix, locationName, path string, vars sets.String) *apis.FieldError {
-	if vs, present, _ := extractVariablesFromString(value, prefix); present {
+	if vs, present, _ := ExtractVariablesFromString(value, prefix); present {
```
lbernick, line 53: why is this being exported? I don't see any additional usages of it

docs/tasks.md
```diff
@@ -726,6 +726,8 @@ spec:
 For more information, see [Using `Workspaces` in `Tasks`](workspaces.md#using-workspaces-in-tasks)
 and the [`Workspaces` in a `TaskRun`](../examples/v1beta1/taskruns/workspace.yaml) example YAML file.
 
+Workspaces can be propagated to embedded task specs, not referenced Tasks. For more information, see [Referenced TaskRuns within Embedded PipelineRuns](pipelineruns.md#referenced-taskruns-within-embedded-pipelineruns).
```
lbernick, line 729: Can you also add docs for propagating workspaces from taskruns -> inline task specs?

lbernick (commented):
Thank you Chitrang for splitting into multiple PRs! it's a big help

docs/tasks.md
chitrangpatel: Oh shoot. I thought those existed. But I only had them for pipelinerun. Thanks! I will update it right away.

chitrangpatel: Done.

pkg/reconciler/taskrun/taskrun.go
chitrangpatel: Sure.

pkg/substitution/substitution.go
chitrangpatel: It might be because I split it up. I will undo it so that I can fix it in the other PR where it is actually used.

chitrangpatel: done.

pkg/reconciler/taskrun/taskrun.go
chitrangpatel: done.

pkg/workspace/apply_test.go
```diff
@@ -1001,6 +1001,55 @@ func TestApply_IsolatedWorkspaces(t *testing.T) {
 				ReadOnly:  true,
 			}},
 		},
+	}, {
+		name: "add workspaces to taskspec",
+		ts: v1beta1.TaskSpec{
+			Sidecars: []v1beta1.Sidecar{{
+				Name: "conflicting volume mount sidecar",
+				VolumeMounts: []corev1.VolumeMount{{
+					Name:      "mount-path-conflicts",
```
lbernick, line 1010: it looks like this test case is for conflicting mount paths, can you rename to reflect what's being tested here?

pkg/apis/pipeline/v1beta1/task_types.go
```diff
@@ -78,6 +79,61 @@ func (*Task) GetGroupVersionKind() schema.GroupVersionKind {
...
+		}
+		for i := range stepTemplate.Command {
+			locationsToCheck.Insert(stepTemplate.Command[i])
+		}
+	}
+	return locationsToCheck
+}
+
+// FindWorkspaceSubstitutionLocationsInTask returns a set of all the locations in the TaskSpec  where we can apply substitutions.
+func (ts *TaskSpec) FindWorkspaceSubstitutionLocationsInTask() sets.String {
```
lbernick, line 129: Where is this function being used? It looks like this is functionality for replacing `workspaces.foo.path` syntax but I don't see it being used anywhere-- is this replacement happening as part of this PR?
If this function does need to be in this PR, I don't think it should be exported.

pkg/workspace/apply_test.go
chitrangpatel: No it wasn't. Its my mistake. Copy and pasting the above test and then modifying it lead to this confusion. I apologize. I will update it to better reflect the test its actually doing.

pkg/apis/pipeline/v1beta1/task_types.go
chitrangpatel: Good point. Even though it is written in task_types, it might be being called from the PR handling propagated workspaces in pipelinerun.

chitrangpatel: This function was being called in the pipelinerun reconciler to parse the `pipelineTask` and identify which workspaces are being called. The reconciler then propagated those workspaces to the pipelineTasks by creating task bindings. 

The taskRun reconciler (`reconciler/taskrun/taskrun.go`) does a similar thing and propagates workspaces to the taskSpec.
The substitutions happen the way they did before. i.e. via https://github.com/chitrangpatel/pipeline/blob/e302587a9dcbc7af8cc1475dfa24c3c35a6715f9/pkg/reconciler/pipelinerun/resources/apply.go#L173-L185.

docs/tasks.md
lbernick: Could you also update the link here to point to the new docs?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -2467,97 +2468,41 @@ status:
 	}
 }
 
-func TestExpandMountPath(t *testing.T) {
```
lbernick, line 2470: was this test case (ExpandMountPath) accidentally deleted? I see it on main but not on this branch

pkg/reconciler/taskrun/taskrun_test.go
lbernick: I have a similar comment to dibyo here- what were the results of your experimentation?

docs/taskruns.md
```diff
@@ -422,6 +422,86 @@ For more information, see the following topics:
...
+      - echo
+      image: ubuntu
+      name: simple-step
+      resources: {}
+    workspaces:
+    - name: tr-workspace
+
+```
+
+Workspaces can be propagated to embedded task specs, not referenced Tasks. For more information, see [Referenced TaskRuns within Embedded PipelineRuns](pipelineruns.md#referenced-taskruns-within-embedded-pipelineruns).
```
lbernick, line 503: I'd avoid linking to the PipelineRun docs from the TaskRun docs, since TaskRun docs should be able to stand alone. Can you instead provide an example here of an invalid propagation to a referenced Task?

docs/taskruns.md
```diff
@@ -422,6 +422,86 @@ For more information, see the following topics:
...
+        args:                  
+          - $(workspaces.tr-workspace.path)
+  workspaces:                  
+  - emptyDir: {}               
+    name: tr-workspace
+```
+
+Upon execution, the workspaces will be interpolated during resolution.
+
+```yaml
```
lbernick, line 456: Just my personal opinion but I think having the status here isn't as useful as having an english explanation of what will happen when this taskrun is run, since there's a lot of extra info in the status that isn't very relevant.

docs/taskruns.md
chitrangpatel: Yes, Can do!

docs/taskruns.md
chitrangpatel: I agree. I will hide all the non-relevant parts and provide a short description of what happens.

pkg/reconciler/taskrun/taskrun_test.go
chitrangpatel: Thanks for the catch. Yes, this was accidentally deleted.

pkg/reconciler/taskrun/taskrun_test.go
chitrangpatel: I ended up implementing a dedicated unit test `TestUpdateTaskSpecParamsContextsResults` to check the logic until pod creation.

lbernick: I think this test case still has a decent amount of logic in it (getting workspaces from the taskrun spec, calling workspace.CreateVolumes, and passing that into the function being tested), which doesn't really tell us whether the actual code is using this function correctly. I'd prefer to see a test that calls `Reconcile` with a TaskRun spec, and verifies that the output TaskRun has the correct volumes.

chitrangpatel: Sure. I can try to implement that.

docs/tasks.md
chitrangpatel: Actually, this should be added in the PR propagating workspaces in pipeline runs since it references documentation that does not exist in this PR. I will fix it.

pkg/reconciler/taskrun/taskrun_test.go
chitrangpatel: done.

docs/taskruns.md
chitrangpatel: done

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -2545,8 +2590,14 @@ spec:
 		Kind:     "Task",
 		TaskSpec: &v1beta1.TaskSpec{Steps: simpleTask.Spec.Steps, Workspaces: simpleTask.Spec.Workspaces},
 	}
-	taskSpec := updateTaskSpecParamsContextsResults(context.Background(), taskRun, rtr)
-	pod, err := r.createPod(testAssets.Ctx, taskSpec, taskRun, rtr)
+	ctx := config.EnableAlphaAPIFields(context.Background())
```
lbernick, line 2593: would you mind creating an issue to refactor these tests to call `Reconcile` rather than all of these helpers?

lbernick (commented):
/lgtm

pkg/reconciler/taskrun/taskrun_test.go
chitrangpatel: Sure. I can do that.

PR #5054, opened 2022-06-29
[TEP-0104] Add Validation for Step-level Resource Requirements

austinzhao-go:
# Changes

/kind feature

The related FR thread:
https://github.com/tektoncd/pipeline/issues/4470
The related TEP:
[TEP-0104: Task-level Resource Requirements](https://github.com/tektoncd/community/blob/main/teps/0104-tasklevel-resource-requirements.md)

The previous PRs:
https://github.com/tektoncd/pipeline/pull/4877 - Fields Addition & Validation w/ Docs Updates
https://github.com/tektoncd/pipeline/pull/5212 - Populate Task-level Resource Requirements from PipelineRun to TaskRun
The follow-up PR:
https://github.com/tektoncd/pipeline/pull/5082 - Update Pod with Task-level Resource Requirements 

This PR will include logic and tests as:
- Add validation for step-level resource requirements

### TODO
- [x] Resolve initial review feedback
- [x] Add validation for step-level resource requirements

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
(Note: updated docs were included in https://github.com/tektoncd/pipeline/pull/4877/files)
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been filled in
(if there are no user facing changes, use release note "NONE")

# Release Notes
``` release-note
NONE
```

PR #5082, opened 2022-07-05
[TEP-0104] Update Pod with Task-level Resource Requirements

austinzhao-go:
# Changes
/kind feature

Closes https://github.com/tektoncd/pipeline/issues/4470

The related TEP:
[TEP-0104: Task-level Resource Requirements](https://github.com/tektoncd/community/blob/main/teps/0104-tasklevel-resource-requirements.md)

The related impl PRs:
https://github.com/tektoncd/pipeline/pull/4877 - Fields Addition & Validation w/ Docs Updates
https://github.com/tektoncd/pipeline/pull/5054 - Add Validation for Step-level Resource Requirements 
https://github.com/tektoncd/pipeline/pull/5212 - Populate Task-level Resource Requirements from PipelineRun to TaskRun

This PR will include logic and tests as:
- Update the containers in the Pod with the specified task-level resource requirements

# Submitter Checklist

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing (included in https://github.com/tektoncd/pipeline/pull/4877)
  functionality, content, code)

# Release Notes
``` release-note
Supported Task-level Resource Requirements 
```

### DIFF
--- a/docs/compute-resources.md
+++ b/docs/compute-resources.md
@@ -45,7 +45,6 @@ Therefore, the pod will have no effective CPU limit.
 ## Task-level Compute Resources Configuration
 
 **([alpha only](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features))**
-(This feature is under development and not functional yet. Stay tuned!)
 
 Tekton allows users to specify resource requirements of [`Steps`](./tasks.md#defining-steps),
 which run sequentially. However, the pod's effective resource requirements are still the
@@ -63,7 +62,9 @@ Each of these details is explained in more depth below.
 Some points to note:
 
 - Task-level resource requests and limits do not apply to sidecars which can be configured separately.
-- Users may not configure the Task-level and Step-level resource requirements (requests/limits)  simultaneously.
+- If only limits are configured in task-level, it will be applied as the task-level requests.
+- Resource requirements configured in `Step` or `StepTemplate` of the referenced `Task` will be overridden by the task-level requirements.
+- `TaskRun` configured with both `StepOverrides` and task-level requirements will be rejected.
 
 ### Configure Task-level Compute Resources
 
@@ -84,7 +85,7 @@ spec:
       cpu: 2
 ```
 
-The following TaskRun will be rejected, because it configures both step-level and task-level compute resource requirements:
+The following TaskRun will be rejected, because it configures both stepOverrides and task-level compute resource requirements:
 
 ```yaml
 kind: TaskRun
--- a/pkg/internal/computeresources/tasklevel/tasklevel.go
+++ b/pkg/internal/computeresources/tasklevel/tasklevel.go
@@ -1,0 +1,57 @@
+/*
+Copyright 2020 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tasklevel
+
+import (
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// ApplyTaskLevelComputeResources applies the task-level compute resource requirements to each Step.
+func ApplyTaskLevelComputeResources(steps []v1beta1.Step, computeResources *corev1.ResourceRequirements) {
+	if computeResources == nil {
+		return
+	}
+	if computeResources.Requests == nil && computeResources.Limits == nil {
+		return
+	}
+	averageRequests := computeAverageRequests(computeResources.Requests, len(steps))
+	averageLimits := computeAverageRequests(computeResources.Limits, len(steps))
+	for i := range steps {
+		// if no requests are specified in step or task level, the limits are used to avoid
+		// unnecessary higher requests by Kubernetes default behavior.
+		if steps[i].Resources.Requests == nil && computeResources.Requests == nil {
+			steps[i].Resources.Requests = averageLimits
+		} else {
+			steps[i].Resources.Requests = averageRequests
+		}
+		steps[i].Resources.Limits = computeResources.Limits
+	}
+}
+
+// computeAverageRequests computes the average of the requests of all the steps.
+func computeAverageRequests(requests corev1.ResourceList, steps int) corev1.ResourceList {
+	if len(requests) == 0 || steps == 0 {
+		return nil
+	}
+	averageRequests := corev1.ResourceList{}
+	for k, v := range requests {
+		averageRequests[k] = *resource.NewMilliQuantity(v.MilliValue()/int64(steps), requests[k].Format)
+	}
+	return averageRequests
+}
--- a/pkg/internal/computeresources/tasklevel/tasklevel_test.go
+++ b/pkg/internal/computeresources/tasklevel/tasklevel_test.go
@@ -1,0 +1,206 @@
+/*
+Copyright 2019 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tasklevel_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"github.com/tektoncd/pipeline/pkg/internal/computeresources/tasklevel"
+	"github.com/tektoncd/pipeline/test/diff"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func TestApplyTaskLevelResourceRequirements(t *testing.T) {
+	testcases := []struct {
+		desc                     string
+		Steps                    []v1beta1.Step
+		ComputeResources         corev1.ResourceRequirements
+		expectedComputeResources []corev1.ResourceRequirements
+	}{{
+		desc: "only with requests",
+		Steps: []v1beta1.Step{{
+			Name:    "1st-step",
+			Image:   "image",
+			Command: []string{"cmd"},
+		}, {
+			Name:    "2nd-step",
+			Image:   "image",
+			Command: []string{"cmd"},
+		}},
+		ComputeResources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		},
+		expectedComputeResources: []corev1.ResourceRequirements{{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+		}, {
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+		}},
+	}, {
+		desc: "only with limits",
+		Steps: []v1beta1.Step{{
+			Name:    "1st-step",
+			Image:   "image",
+			Command: []string{"cmd"},
+		}, {
+			Name:    "2nd-step",
+			Image:   "image",
+			Command: []string{"cmd"},
+		}},
+		ComputeResources: corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+		},
+		expectedComputeResources: []corev1.ResourceRequirements{{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("250m")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+		}, {
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("250m")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+		}},
+	}, {
+		desc: "both with requests and limits",
+		Steps: []v1beta1.Step{{
+			Name:    "1st-step",
+			Image:   "image",
+			Command: []string{"cmd"},
+		}, {
+			Name:    "2nd-step",
+			Image:   "image",
+			Command: []string{"cmd"},
+		}},
+		ComputeResources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		},
+		expectedComputeResources: []corev1.ResourceRequirements{{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		}, {
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		}},
+	}, {
+		desc: "steps with compute resources are overridden by task-level compute resources",
+		Steps: []v1beta1.Step{{
+			Name:    "1st-step",
+			Image:   "image",
+			Command: []string{"cmd"},
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("100m")},
+				Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+			},
+		}, {
+			Name:    "2nd-step",
+			Image:   "image",
+			Command: []string{"cmd"},
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("200m")},
+				Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+			},
+		}},
+		ComputeResources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		},
+		expectedComputeResources: []corev1.ResourceRequirements{{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		}, {
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		}},
+	}, {
+		desc: "steps with partial compute resources are overridden by task-level compute resources",
+		Steps: []v1beta1.Step{{
+			Name:    "1st-step",
+			Image:   "image",
+			Command: []string{"cmd"},
+			Resources: corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("100m")},
+				Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+			},
+		}, {
+			Name:    "2nd-step",
+			Image:   "image",
+			Command: []string{"cmd"},
+		}},
+		ComputeResources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		},
+		expectedComputeResources: []corev1.ResourceRequirements{{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		}, {
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		}},
+	}, {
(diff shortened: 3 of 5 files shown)


austinzhao-go:
/hold

till https://github.com/tektoncd/pipeline/pull/5054 got merged

pkg/pod/pod.go
```diff
@@ -354,6 +359,26 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec
...
+// applyTaskLevelComputeResources applies the task-level compute resource requirements to the Pod.
+func applyTaskLevelComputeResources(steps []v1beta1.Step, computeResources *corev1.ResourceRequirements, taskSpec v1beta1.TaskSpec) error {
+	if taskSpec.StepTemplate != nil {
+		taskSpec.StepTemplate.Resources = corev1.ResourceRequirements{}
+	}
+
+	// TODO: apply LimitRanges
+
+	for i := range steps {
+		if i == 0 {
```
lbernick, line 371: in the TEP i have the resources applied only to the first container, but I think it would actually be better to split them among the containers. The reason is that limitrange default requests are split among the containers. consider this example:

2 steps
task level memory: 1Gi
limitrange default requests: 600Mi

If you split the task level resources between containers you'll have each container with 500Mi (which is what we want). If you just apply them to one container, the limitrange defaults will kick in for the container without resources and you'll have one container with 1Gi and one with 300Mi (not what we want). 

I'll update the TEP to clarify this.

pkg/pod/pod_test.go
```diff
@@ -2014,6 +2014,215 @@ debug-fail-continue-heredoc-randomly-generated-mz4c7
...
+		trs: v1beta1.TaskRunSpec{
+			ComputeResources: &corev1.ResourceRequirements{
+				Limits: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+			},
+		},
+		expectedComputeResources: []corev1.ResourceRequirements{{
+			Limits: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+		}, {
+			Limits: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+		}},
```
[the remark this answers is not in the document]
austinzhao-go: nice hint. think I need to dig a bit and try some cases to ensure it.

pkg/pod/pod.go
```diff
@@ -361,6 +365,30 @@ func (b *Builder) Build(ctx context.Context, taskRun *v1beta1.TaskRun, taskSpec
 	return newPod, nil
 }
 
+// applyTaskLevelComputeResources applies the task-level compute resource requirements to the Pod.
+func applyTaskLevelComputeResources(steps []v1beta1.Step, computeResources *corev1.ResourceRequirements, taskSpec v1beta1.TaskSpec) {
+	if taskSpec.StepTemplate != nil {
+		taskSpec.StepTemplate.Resources = corev1.ResourceRequirements{}
+	}
```
lbernick, line 372: I don't think you need steptemplate here. Steps are merged with step template and overrides [here](https://github.com/tektoncd/pipeline/blob/3f539724f27a6add8b64d0712e056e4479750975/pkg/pod/pod.go#L136-L149)

pkg/pod/pod_test.go
```diff
@@ -2006,6 +2006,219 @@ debug-fail-continue-heredoc-randomly-generated-mz4c7
...
+		},
+		expectedComputeResources: []corev1.ResourceRequirements{{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		}, {
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2")},
+		}},
+	}, {
+		desc: "overwrite StepTemplate resources requirements",
```
lbernick, line 2088: I think this is a mistake in the TEP :( 

Here, this says it's invalid: https://github.com/tektoncd/community/blob/main/teps/0104-tasklevel-resource-requirements.md#interaction-with-step-resource-requirements

However this example uses it: https://github.com/tektoncd/community/blob/main/teps/0104-tasklevel-resource-requirements.md#example-with-steptemplate

Would you be comfortable taking on a PR to the TEP fixing this mistake?

lbernick (commented):
I know this needs to be rebased onto #5054, but could you keep the scope of this PR to just the changes required to apply the computeResources of a taskrun to a pod? i.e. validation and passing spec from pipelinerun -> taskrun should be covered in #5054. 
Also, I think a better place for some of this logic might be pkg/internal/computeresources.

austinzhao-go:
> 

yap. let me keep validation related in https://github.com/tektoncd/pipeline/pull/5054.

pkg/pod/pod.go
austinzhao-go: I see. thanks this context!
let me clean it up.

pkg/pod/pod_test.go
austinzhao-go: got this. think we still will be good for overwriting resource requirements from StepTemplate while validating/avoiding ones from Step? 

yap. I can do this. (will raise today)

lbernick: Awesome thank you! I think if you can't specify compute resources on a step and a taskrun, you also shouldn't be able to do it on a steptemplate and a taskrun.

austinzhao-go: right! I got your whole point now - as stepTemplate is for step.. 
thanks Lee, will mark it invalid - raise TEP update PR - then update code

austinzhao-go: raised update PR https://github.com/tektoncd/community/pull/763

re-arranged validation in https://github.com/tektoncd/pipeline/pull/5054
and removed the StepTemplate case for pod_test.go

pkg/internal/computeresources/transformer.go
```diff
@@ -99,24 +99,38 @@ func transformPodBasedOnLimitRange(p *corev1.Pod, limitRange *corev1.LimitRange)
...
 	}
 	return p
 }
 
-func setRequestsOrLimits(name corev1.ResourceName, dst, src corev1.ResourceList) {
-	if limitrange.IsZero(dst[name]) && !limitrange.IsZero(src[name]) {
-		dst[name] = src[name]
+// setRequests sets the requests to the max of (currentRequests, defaultRequests)
+// and the min of (currentRequests, defaultLimits)
+func setRequests(name corev1.ResourceName, dst corev1.ResourceList, requests corev1.ResourceList, limits corev1.ResourceList) {
```
austinzhao-go, line 118: think some discussion will come to this line:

my current understanding as 
- if not limitranges, just stay as it is

- if currentRequest < min (of limitrange), keep the min
- if over the limit, set as the limit

for limits
- similarly, but keep the min of currentLimit and limit from limitrange

pkg/internal/computeresources/transformer_test.go
```diff
@@ -313,7 +313,7 @@ func TestTransformerOneContainer(t *testing.T) {
 			Containers: []corev1.Container{{
 				Resources: corev1.ResourceRequirements{
 					Limits: corev1.ResourceList{
-						corev1.ResourceMemory:           resource.MustParse("4Gi"),
+						corev1.ResourceMemory:           resource.MustParse("2Gi"),
```
austinzhao-go, line 316: these updates on previous testcases are to align with updated logics on "setting requests/limits with limitranges"

pkg/internal/computeresources/tasklevel/tasklevel_test.go
```diff
@@ -0,0 +1,222 @@
...
+				Command: []string{"cmd"},
+			}},
+			StepTemplate: &v1beta1.StepTemplate{
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.MustParse("500m"),
+						corev1.ResourceMemory: resource.MustParse("500Mi"),
+					},
+				},
+			},
```
lbernick, line 128: StepTemplate isn't used in this test at all, so this test isn't testing what we want it to. At the point where `ApplyTaskLevelComputeResources` is called, the step template resource requests have already been applied to the input containers [here](https://github.com/tektoncd/pipeline/blob/e8ee8798a6758bd4bfdd41bb23ec63eff1fb7aec/pkg/pod/pod.go#L138). There are two options here:
1. add a test case where the input steps already have resources on them, and make sure they don't make their way to the output. I don't love this solution because it relies on the behavior I linked to.
2. Create an overall pod test with steptemplate resources and task level resources, and make sure the pod created has the right compute resources. This would be a more thorough test but harder to work with.

I'll let you use your best judgment here but I'll also try to figure out if there's any refactoring I can try here.

pkg/internal/computeresources/tasklevel/tasklevel_test.go
```diff
@@ -0,0 +1,222 @@
...
+				Name: "sidecar",
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU: resource.MustParse("750m"),
+					},
+					Limits: corev1.ResourceList{
+						corev1.ResourceCPU: resource.MustParse("1.5"),
+					},
+				},
+			}},
```
lbernick, line 167: Same comment as above; sidecars are ignored in `ApplyTaskLevelComputeResources` so they're not actually being tested here.

pkg/internal/computeresources/transformer_test.go
```diff
@@ -388,6 +388,112 @@ func TestTransformerOneContainer(t *testing.T) {
 				},
 			}},
 		},
+	}, {
+		description: "limitRange with min and max, while resources on containers < min",
```
lbernick, line 392: Was this a bug? Could you make this change in a separate PR?

pkg/internal/computeresources/tasklevel/tasklevel_test.go
```diff
@@ -0,0 +1,222 @@
...
+		},
+		trs: v1beta1.TaskRunSpec{
+			ComputeResources: &corev1.ResourceRequirements{
+				Limits: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+			},
+		},
+		expectedComputeResources: []corev1.ResourceRequirements{{
+			Limits: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+		}, {
+			Limits: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
```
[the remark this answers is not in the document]
austinzhao-go: thanks for more thought on this point. 

I tend to take the "request as the existing limit when only limit is configured" way
- think will align with current k8s behaviours - so ease user's understanding and usages
- and "0 / smallest as possible" request could lead to a "resource hungry" case in the runtime, like the CPU requests acting as weights to acquire CPU time as https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run

pkg/internal/computeresources/tasklevel/tasklevel_test.go
austinzhao-go: ya. this testcase actually came from one of our previous threads
https://github.com/tektoncd/pipeline/pull/4877#discussion_r899154954

pkg/internal/computeresources/transformer_test.go
austinzhao-go: sure. and saw your raised refact PR.

think good for me to keep "updating pod" logic in this PR while leaving the transformer and limitrange changes.

pkg/internal/computeresources/tasklevel/tasklevel_test.go
austinzhao-go: thanks this context. think good for me mark as a follow-up action.
and will remove this case at this moment.

pkg/internal/computeresources/transformer_test.go
austinzhao-go: a follow-up PR I will apply on "setting the requests and limits" after this refact PR was merged.

austinzhao-go: raised here - https://github.com/tektoncd/pipeline/pull/5269

pkg/internal/computeresources/tasklevel/tasklevel_test.go
austinzhao-go: closed - moved this testcase under pod_test.go

austinzhao-go:
/unhold

as depended PRs were merged/closed

pkg/pod/pod_test.go
```diff
@@ -2006,6 +2006,147 @@ debug-fail-continue-heredoc-randomly-generated-mz4c7
 	}
 }
 
+func TestPodBuild_TaskLevelResourceRequirements(t *testing.T) {
```
lbernick, line 2009: this looks great! could you also add a test for where the steps have resource requirements?

pkg/pod/pod_test.go
```diff
@@ -2006,6 +2006,147 @@ debug-fail-continue-heredoc-randomly-generated-mz4c7
...
+
+			if err := verifyTaskLevelComputeResources(tc.expectedComputeResources, gotPod.Spec.Containers); err != nil {
+				t.Errorf("verifyTaskLevelComputeResources: %v", err)
+			}
+		})
+	}
+}
+
+// verifyTaskLevelComputeResources verifies that the given TaskRun's containers have the expected compute resources.
+func verifyTaskLevelComputeResources(expectedComputeResources []corev1.ResourceRequirements, containers []corev1.Container) error {
```
lbernick, line 2138: 👍 could you please also check the container name? This will help us test that the sidecar is getting the right resource requirements, as opposed to the step containers

pkg/internal/computeresources/tasklevel/tasklevel_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/tektoncd/pipeline/pkg/internal/computeresources/tasklevel"
+	"github.com/tektoncd/pipeline/test/diff"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func TestPodBuild_TaskLevelResourceRequirements(t *testing.T) {
+	testcases := []struct {
+		desc                     string
+		ts                       v1beta1.TaskSpec
```
lbernick, line 34: Since the function being tested accepts a slice of steps, I think that's what should be in the test case. (I don't think it makes sense to use a TaskSpec that has other fields including sidecars, because this function doesn't accept task specs or sidecars in its input.

pkg/internal/computeresources/tasklevel/tasklevel.go
```diff
@@ -0,0 +1,50 @@
...
+
+package tasklevel
+
+import (
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// ApplyTaskLevelComputeResources applies the task-level compute resource requirements to the Pod.
```
lbernick, line 25:
```suggestion
// ApplyTaskLevelComputeResources applies the task-level compute resource requirements to each Step
```

pkg/internal/computeresources/tasklevel/tasklevel_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"github.com/tektoncd/pipeline/pkg/internal/computeresources/tasklevel"
+	"github.com/tektoncd/pipeline/test/diff"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func TestPodBuild_TaskLevelResourceRequirements(t *testing.T) {
```
lbernick, line 31:
```suggestion
func TestApplyTaskLevelResourceRequirements(t *testing.T) {
```

pkg/internal/computeresources/tasklevel/tasklevel_test.go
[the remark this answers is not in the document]
lbernick: Good point that this could lead to a pod being more likely to be evicted. I'll update the TEP. in the meantime, I think you can go ahead with implementing this strategy.

austinzhao-go:
/hold

till https://github.com/tektoncd/pipeline/pull/5269 got merged

pkg/internal/computeresources/tasklevel/tasklevel_test.go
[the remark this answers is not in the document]
lbernick: https://github.com/tektoncd/community/pull/776

pkg/internal/computeresources/tasklevel/tasklevel_test.go
austinzhao-go: the tested function here we mean `builder.Build()` right? which needs a TaksRun and a TaskSepc

austinzhao-go: the `Build` function is the one we mentioned in last week to test as it covers logic related to steps, stepTemplate, sidecar to form the container configs to create the Pod. 
`verifyTaskLevelComputeResources` is the help func which we need to verify if the result pod matches our expectation.

lbernick: It looks like this function is testing ApplyTaskLevelResourceRequirements not builder.Build

pkg/internal/computeresources/tasklevel/tasklevel_test.go
austinzhao-go: as above thread - think the tested func is `Build`

pkg/internal/computeresources/tasklevel/tasklevel_test.go
lbernick: I think your approach in pod_test.go (where you're using the TaskSpec and calling builder.Build) is correct!

lbernick: Unresolving this conversation, because while the tests in pod_test.go LGTM (the ones testing builder.Build), this test (in tasklevel_test.go) is testing the function `tasklevel.ApplyTaskLevelComputeResources`, which only takes steps and compute resources. Therefore, I think the test cases in this file should use steps and compute resources as their inputs, not task specs with other fields in addition.

pkg/internal/computeresources/tasklevel/tasklevel_test.go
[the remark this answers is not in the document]
lbernick: Unresolving since I don't think this comment was addressed; just to be clear I'm suggesting implementing what's proposed in the TEP PR.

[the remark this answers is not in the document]
austinzhao-go: thought this will be handled in limitrange (wrongly). but a case will be if there is no limitrange.
let me update

pkg/internal/computeresources/tasklevel/tasklevel_test.go
austinzhao-go: missed these 2 files. sry for this. 
thanks the double check - let me update!

pkg/internal/computeresources/tasklevel/tasklevel_test.go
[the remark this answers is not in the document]
austinzhao-go: added the required logic and updated testcases.

pkg/internal/computeresources/tasklevel/tasklevel_test.go
austinzhao-go: cleaned testcases struct.

pkg/internal/computeresources/tasklevel/tasklevel_test.go
```diff
@@ -0,0 +1,206 @@
...
+		}, {
+			Name:    "2nd-step",
+			Image:   "image",
+			Command: []string{"cmd"},
+		}},
+		ComputeResources: corev1.ResourceRequirements{
+			Limits: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
+		},
+		expectedComputeResources: []corev1.ResourceRequirements{{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("500m")},
```
lbernick, line 71: Since the limits are "500m", we also want to set the total requests to "500m", meaning that each container should have requests of "250m".

austinzhao-go: I see. this (limits) will be applied as the total task-level allowed requests.
let me correct.

austinzhao-go: corrected by "spreading" the limits when no requests.

pkg/internal/computeresources/tasklevel/tasklevel.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	if computeResources.Requests == nil && computeResources.Limits == nil {
+		return
+	}
+	averageRequests := computeAverageRequests(computeResources.Requests, len(steps))
+	for i := range steps {
+		// if no requests are specified in step or task level, the limits are used to avoid
+		// unnecessary higher requests by Kubernetes default behavior.
+		if steps[i].Resources.Requests == nil && computeResources.Requests == nil {
+			steps[i].Resources.Requests = computeAverageRequests(computeResources.Limits, len(steps))
```
lbernick, line 38: nit: no need to compute this on every iteration

lbernick (approved):
This looks great!
A few notes:
- once #5269 is merged I think we should add some unit tests to pod_test.go that use the limitrange transformer
- We will need to wait until https://github.com/tektoncd/community/pull/776 is merged as well
- Please update the docs to clarify the case where a user sets limits but no requests

pkg/internal/computeresources/tasklevel/tasklevel.go
austinzhao-go: replaced with a var - thanks this!

docs/compute-resources.md
```diff
@@ -84,7 +85,20 @@ spec:
       cpu: 2
 ```
 
-The following TaskRun will be rejected, because it configures both step-level and task-level compute resource requirements:
+Either requests or limits can be configured individually:
```
lbernick, line 88: What we specifically want to clarify here is that if no requests are set, the *task-level* requests will be set to the *task-level* limits

docs/compute-resources.md
```diff
@@ -63,7 +63,8 @@ Each of these details is explained in more depth below.
 Some points to note:
 
 - Task-level resource requests and limits do not apply to sidecars which can be configured separately.
-- Users may not configure the Task-level and Step-level resource requirements (requests/limits)  simultaneously.
+- Resource requirements configured in `Step` or `StepTemplate` of the referenced `Task` will be overridden by the task-level requirements.
```
lbernick, line 66: Also, please update line 48 to clarify that this feature is implemented!

docs/compute-resources.md
austinzhao-go: right. misunderstood your original point. - quick update

docs/compute-resources.md
austinzhao-go: yap.

docs/compute-resources.md
```diff
@@ -45,7 +45,7 @@ Therefore, the pod will have no effective CPU limit.
 ## Task-level Compute Resources Configuration
 
 **([alpha only](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features))**
-(This feature is under development and not functional yet. Stay tuned!)
+(This alpha feature is implemented)
```
lbernick, line 48: sorry I should have been more clear-- this line can just be removed. (Users can safely assume everything in our docs is implemented unless specified otherwise.

PR #5083, opened 2022-07-05
[TEP-0075] Support Object Results substitution

Yongxuanzhang:
# Changes
This is part of work in [TEP-0075](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md).
This commit provides the support to apply object results replacements.
Previous this commit we support emitting object results so users can
write object results to task level, but we cannot pass object results from
one task to another within one pipeline. This commit adds the support for this.

Note that this commit does not support `object` and `array` results for `Matrix` 

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes
```release-note
Support object results substitution as an alpha feature.

results:
  - name: object-results
     type: object
     description: The object results
     properties:
        foo: {
          type: string
        }
        hello: {
          type: string
        }

And the task script can populate result in an object form with:

echo -n "{\"foo\":\"bar\",\"hello\":\"world\"}" | tee $(results.object-results.path)

and we can refer to the object results in param like:
  params:
    - name: whole-object
      value: "$(tasks.task1.results.object-results[*])"
    - name: object-element
      value: "$(tasks.task1.results.object-results.hello)"

This feature is part of the TEP-0075. 
```

### DIFF
--- a/examples/v1beta1/pipelineruns/alpha/pipeline-object-results.yaml
+++ b/examples/v1beta1/pipelineruns/alpha/pipeline-object-results.yaml
@@ -1,0 +1,67 @@
+apiVersion: tekton.dev/v1beta1
+kind: PipelineRun
+metadata:
+  name: pipelinerun-object-results
+spec:
+  pipelineSpec:
+    tasks:
+      - name: task1
+        taskSpec:
+          results:
+            - name: object-results
+              type: object
+              description: The object results
+              properties:
+                foo:
+                  type: string
+                hello:
+                  type: string
+          steps:
+            - name: write-object
+              image: bash:latest
+              script: |
+                #!/usr/bin/env bash
+                echo -n "{\"foo\":\"bar\",\"hello\":\"world\"}" | tee $(results.object-results.path)
+      - name: task2
+        params:
+          - name: whole-object
+            value: "$(tasks.task1.results.object-results[*])"
+          - name: object-element
+            value: "$(tasks.task1.results.object-results.hello)"
+        taskSpec:
+          params:
+            - name: whole-object
+              type: object
+              properties:
+                foo:
+                  type: string
+                hello:
+                  type: string
+              default: {
+                hello: "",
+                foo: "",
+              }
+            - name: object-element
+              type: string
+              default: "defaultparam1"
+          steps:
+            - name: print-object-foo
+              image: bash:latest
+              args: [
+                "echo",
+                "$(params.whole-object.foo)"
+              ]
+            - name: print-object-hello
+              image: ubuntu
+              script: |
+                #!/bin/bash
+                VALUE=$(params.object-element)
+                EXPECTED="world"
+                diff=$(diff <(printf "%s\n" "${VALUE[@]}") <(printf "%s\n" "${EXPECTED[@]}"))
+                if [[ -z "$diff" ]]; then
+                    echo "Get expected: ${VALUE}"
+                    exit 0
+                else
+                    echo "Want: ${EXPECTED} Got: ${VALUE}"
+                    exit 1
+                fi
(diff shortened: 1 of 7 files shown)


pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -140,6 +140,13 @@ func wrongTypeParamsNames(params []v1beta1.Param, matrix []v1beta1.Param, needed
 			// passed to the task that aren't being used.
 			continue
 		}
+		// This is needed to support array replacements in params. Users want to use $(tasks.taskName.results.resultname[*])
```
Yongxuanzhang, line 143: This is the same in https://github.com/tektoncd/pipeline/pull/4908, once it is merged we can rebase it

ywluogg:
#4723

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -123,11 +123,12 @@ func ApplyPipelineTaskContexts(pt *v1beta1.PipelineTask) *v1beta1.PipelineTask {
...
 	stringReplacements := resolvedResultRefs.getStringReplacements()
 	arrayReplacements := resolvedResultRefs.getArrayReplacements()
+	objectReplacements := resolvedResultRefs.getObjectReplacements()
 	for _, resolvedPipelineRunTask := range targets {
 		if resolvedPipelineRunTask.PipelineTask != nil {
 			pipelineTask := resolvedPipelineRunTask.PipelineTask.DeepCopy()
-			pipelineTask.Params = replaceParamValues(pipelineTask.Params, stringReplacements, arrayReplacements, nil)
-			pipelineTask.Matrix = replaceParamValues(pipelineTask.Matrix, stringReplacements, arrayReplacements, nil)
+			pipelineTask.Params = replaceParamValues(pipelineTask.Params, stringReplacements, arrayReplacements, objectReplacements)
+			pipelineTask.Matrix = replaceParamValues(pipelineTask.Matrix, stringReplacements, arrayReplacements, objectReplacements)
```
ywluogg, line 131: I think we don't want matrix to support nested values right now. This would essentially mean matrix can be [[array, object], [etc.]], is it?

Yongxuanzhang: done!

examples/v1beta1/pipelineruns/alpha/pipeline-object-results.yaml
```diff
@@ -0,0 +1,71 @@
...
+  pipelineSpec:
+    tasks:
+      - name: task1
+        taskSpec:
+          results:
+            - name: object-results
+              type: object
+              description: The object results
+              properties:
+                foo: {
```
dibyom, line 15: are the curly braces necessary since type: string is already indented

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -123,11 +123,12 @@ func ApplyPipelineTaskContexts(pt *v1beta1.PipelineTask) *v1beta1.PipelineTask {
 func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResultRefs) {
 	stringReplacements := resolvedResultRefs.getStringReplacements()
 	arrayReplacements := resolvedResultRefs.getArrayReplacements()
+	objectReplacements := resolvedResultRefs.getObjectReplacements()
 	for _, resolvedPipelineRunTask := range targets {
 		if resolvedPipelineRunTask.PipelineTask != nil {
 			pipelineTask := resolvedPipelineRunTask.PipelineTask.DeepCopy()
-			pipelineTask.Params = replaceParamValues(pipelineTask.Params, stringReplacements, arrayReplacements, nil)
-			pipelineTask.Matrix = replaceParamValues(pipelineTask.Matrix, stringReplacements, arrayReplacements, nil)
```
dibyom, line 130: why are we removing this?

examples/v1beta1/pipelineruns/alpha/pipeline-object-results.yaml
Yongxuanzhang: yes they are not. I will remove these

pkg/reconciler/pipelinerun/resources/apply.go
dibyom: got it, thanks!

PR #5084, opened 2022-07-06
cmd/entrypoint: do not interpret anything after `--`

### LINKED ISSUE #5080: Tasks are failing when "init" is the first argument followed by two or more other arguments.
# Expected Behavior

Given the following task 

```yaml
kind: Task
metadata:
  name: tkn-arg-test
  labels:
    app.kubernetes.io/version: "0.4"
  annotations:
    tekton.dev/pipelines.minVersion: "0.22.0"
    tekton.dev/tags: cli
spec:
  description: >-
    Test consuming args
  params:
  - name: ARGS
    description: The terraform cli commands to tun
    type: array
    default:
      - "--help"
  - name: USER_HOME
    description: Override home directory to /tekton/home
    type: string
    default: "/tekton/home"
  steps:
    - name: echo-cli
      image: registry.access.redhat.com/ubi9/ubi-minimal:9.0.0-1580
      workingDir: /tekton/home
      args:
        - "$(params.ARGS)"
      command: ["echo"]
      resources:
       limits:
         cpu: 250m
         memory: 1Gi
       requests:
         cpu: 250m
         memory: 500Mi
      env:
        - name: "HOME"
          value: $(params.USER_HOME)
```

The following command should work: 

```bash
$ tkn task start tkn-arg-test "-p=ARGS=init,two" --use-param-defaults --showlog
# Works…
[echo-cli] init two 
$ tkn task start tkn-arg-test "-p=ARGS=init,two,three" --use-param-defaults --showlog
# Doesn't work
[echo-cli] 2022/07/05 17:05:42 init error: open two: no such file or directory
```

# Actual Behavior

It fails as written above

- Create the task
- Issue the `tkn` commands

# Additional Info

This affect any Pipeline release that contains the follow change : 
https://github.com/tektoncd/pipeline/pull/4826. This means it affects 0.32 to 0.37 and the fix should be backport in all.

The main reason why it fails is the way the `entrypoint` binary is written and the way `flag` acts. To simplify a tiny bit, the command created on a task looks like `/ko-app/entrypoint --entrypoint echo -- command and args here` (with a bunch of flags before `--` ommited here). With the above example, it becomes `/ko-app/entrypoint --entrypoint echo -- init foo bar`. The `flag` package, **in that particular case** ignores the `--` and tells the rest of the code that the arguments are `init foo bar`, which is similar to `/ko-app/entrypoint --entrypoint echo init foo bar`. And *this*, in terms, goes into the `init` subcommand.

/assign
/priority critical-urgent

vdemeester:
# Changes

The way the `flag` package works, it can eat the flag
"terminator" (aka the double dash `--`). This means that, in some very
specific cases — where the first item after the `--` is also a
subcommand, it would execute the entrypoint subcommand instead of the
actual command.

For example :

```
$ /ko-app/entrypoint -- init a b
$ /ko-app/entrypoint init a b
```

This is fixed by making sure we remove anything after `--` for the
subcommand processing. And then we pass the rest (after `--`) to the
entrypointer to be executed.

Note for the reviewers: this is one approach (mutate `os.Args`) that is relatively contained. The other way to fix that would be to have an `exec` subcommand that would be used to *exec* things, **but** this means changing all containers commands as well which.. might be a bit much for bug fixes at least.
In any case, I think the entrypoint binary needs a bit of love and refactoring.

Fixes #5080 

/kind bug
/priority critial-urgent

This will have to be cherry-pick in **all** release from 0.35 to 0.37 😅 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fix the entrypoint potentially not executing the right command due to flag parsing
```

### DIFF
--- a/cmd/entrypoint/args.go
+++ b/cmd/entrypoint/args.go
@@ -1,0 +1,38 @@
+/*
+Copyright 2022 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+func extractArgs(initialArgs []string) ([]string, []string) {
+	commandArgs := []string{}
+	args := initialArgs
+	if len(initialArgs) == 0 {
+		return args, commandArgs
+	}
+	// Detect if `--` is present, if it is, parse only the one before.
+	terminatorIndex := -1
+	for i, a := range initialArgs {
+		if a == "--" {
+			terminatorIndex = i
+			break
+		}
+	}
+	if terminatorIndex > 0 {
+		commandArgs = initialArgs[terminatorIndex+1:]
+		args = initialArgs[:terminatorIndex]
+	}
+	return args, commandArgs
+}
--- a/cmd/entrypoint/args_test.go
+++ b/cmd/entrypoint/args_test.go
@@ -1,0 +1,67 @@
+/*
+Copyright 2022 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


vdemeester:
This probably need at least one test 🙃 working on it.

vdemeester:
```
cloudevent: TestSendCloudEventWithRetries/test-send-cloud-event-failed expand_less | 0s
{Failed      cloud_event_controller_test.go:619: error in test test-send-cloud-event-failed: received 0 events but 1 expected. Found events: []string{} WARN	cloudevent/cloud_event_controller.go:162	Failed to send cloudevent: Had to fail. Event ID: 1f6a546b-df34-4e1b-8a72-bfb1fa13414b}
```
🤔 those are new "unit test" flakes ? 🤔

abayer:
I believe the unit test flakes are due to my change decreasing requests/limits for that job - https://github.com/tektoncd/plumbing/pull/1124 reverts that change.

abayer (approved):
I'm a bit worried that this could cause something else to fail, due to all the fun complicated behaviors around `--`, but I can't think of anything better. =)

afrittoli (commented):
/lgtm

PR #5085, opened 2022-07-06
Add support for projected volumes as workspace type

### LINKED ISSUE #5075: Projected volumes as workspaces
### Feature request

Support kubernetes projected volumes (https://kubernetes.io/docs/concepts/storage/projected-volumes/) as workspace type.

### Use case

By using projected volumes a user could pass an arbitrary amount of ConfigMaps or Secrets to a PipelineRun/TaskRun without modifying the actual Task/ClusterTask.

In kubevirt-tekton-tasks we have tasks to [customize](https://github.com/kubevirt/kubevirt-tekton-tasks/tree/main/tasks/disk-virt-customize)/[sysprep](https://github.com/kubevirt/kubevirt-tekton-tasks/tree/main/tasks/disk-virt-sysprep) VM images, where this would allow to pass data in a flexible way. The user could pass multiple ConfigMaps or Secrets to the task, which would also allow to split and reuse them in several tasks.

```
spec:
 workspaces:
   - name: my-customization-data
     projected:
       sources:
        - secret:
            name: my-secrets
        - configMap:
            name: my-configs-1
        - configMap:
            name: my-configs-2
```

0xFelix:
# Changes

Add support for projected volumes as workspace type

By adding support for projected volumes [1] as workspace type an
arbitrary amount of ConfigMaps or Secrets can be passed to a
TaskRun/PipelineRun without modifying the actual Task/ClusterTask.

This allows to pass data to Tasks/Pipelines in a flexible way.

Fixes #5075

[1] https://kubernetes.io/docs/concepts/storage/projected-volumes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Add support for projected volumes as workspace type
```

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -434,6 +434,7 @@ Features currently in "alpha" are:
 | [Matrix](./matrix.md)                                                                                 | [TEP-0090](https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md)                                      |                                                                      |                             |
 | [Embedded Statuses](pipelineruns.md#configuring-usage-of-taskrun-and-run-embedded-statuses)           | [TEP-0100](https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md) |                                                                |                             |
 | [Task-level Resource Requirements](compute-resources.md#task-level-compute-resources-configuration)   | [TEP-0104](https://github.com/tektoncd/community/blob/main/teps/0104-tasklevel-resource-requirements.md)             |                                                                      |                             |
+| [Projected Workspace Type](workspaces.md#projected)                                                   |                  |                                                                |                             |
 | [CSI Workspace Type](workspaces.md#csi)                                                               |                  |                                                                |                             |
 
 ## Configuring High Availability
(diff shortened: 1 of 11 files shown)


vdemeester (approved):

piyush-garg (approved):
PR looks good

0xFelix:
I'd argue it does not to be hidden behind a feature flag. It is a combination of existing functionality (Secrets, ConfigMaps) and is provided by K8S itself without any external dependencies. (like the CSI drivers in #5030)

vdemeester:
I would tend to agree with @0xFelix 👼🏼

0xFelix:
@dibyom What do you think?

dibyom:
@0xFelix @vdemeester I agree that while this is a new field it has fairly well scoped behavior. Where I was coming from was that I thought all new api fields in the beta API had to start behind the alpha feature gate or its own feature flag according to https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#new-api-fields.

(I did not see this documented in our api policy - maybe we should do that. Or we could also relax our rules to allow for some fields to make it directly into beta without going through an alpha first - what do you think @vdemeester ?)

vdemeester:
Given that https://github.com/tektoncd/pipeline/pull/5030 has been merged behind the feature-flag, maybe we should do the same. We can refine this a bit later then.

vdemeester (approved):
/meow

Issue #5001, opened 2022-06-17
Refactor results and params into separate sub directories


ywluogg:
### Feature request

Because the results and params related codebase has become complex and are relatively modularized, so it's worths thinking about refactoring them into sub directories:
current proposal:
struct definitions: result_types.go, param_types.go (done)
validations: TODO
reconciler functionalities: organize into `pipeline/pkg/results` and `pipeline/pkg/params`

### Use case

This is also good for sub directory reviewer setups.

PR #5088, opened 2022-07-06
[TEP-0075]Pipeline results support object

Yongxuanzhang:
# Changes

This is part of work in TEP-0075.
Previous to this commit, we have added support for pipeline array
results. This commit supports object results, so pipeline can emit
object results as whole or emit elements of the object from tasks.
Before this commit the pipeline level result only support string and
array.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes
```release-note
Support object results and object element reference for pipeline level as an alpha feature.

A pipeline can specify a type to create object result, such as:

    results:
      - name: object-results
        type: object
        description: whole object
        value: $(tasks.task1.results.object-results[*])
      - name:  object-results-element
        type: string
        description: object element
        value: $(tasks.task2.results.array-results.keyName)

echo -n "{\"foo\":\"bar\",\"hello\":\"world\"}" | tee $(results.object-results.path)

The pipeline results can refer to task results to collect them.
This feature is part of the TEP-0075. 
```


ywluogg:
#4723

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -251,10 +252,16 @@ func ApplyTaskResultsToPipelineResults(
...
-			if len(variableParts) == 4 && variableParts[0] == "tasks" && variableParts[2] == "results" {
+			if variableParts[0] != v1beta1.ResultTaskPart || variableParts[2] != v1beta1.ResultResultPart {
+				validPipelineResult = false
+				continue
+			}
+			switch len(variableParts) {
+			// For string result: tasks.<taskName>.results.<objectResultName>
+			// For array result: tasks.<taskName>.results.<objectResultName>[*], tasks.<taskName>.results.<objectResultName>[i]
+			// For object result: tasks.<taskName>.results.<objectResultName>[*],
+			case 4:
```
ywluogg, line 264: Maybe make these numbers as const?

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -237,10 +237,11 @@ func ApplyTaskResultsToPipelineResults(
 	results []v1beta1.PipelineResult,
 	taskRunResults map[string][]v1beta1.TaskRunResult,
 	customTaskResults map[string][]v1alpha1.RunResult) []v1beta1.PipelineRunResult {
-
+	// TODO(#4723): Validate array and object results.
```
ywluogg, line 240: Is this TODO for validating the substituted result matches the substituting result? I thought that would be in pipelinerun reconciler?

pkg/reconciler/pipelinerun/resources/apply.go
Yongxuanzhang: sure!

ywluogg:
Do you need to add test cases for these in https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_test.go?

examples/v1beta1/pipelineruns/alpha/pipeline-emitting-results.yaml
```diff
@@ -20,21 +20,36 @@ spec:
...
                 #!/usr/bin/env bash
-                echo -n "[\"4\",\"5\",\"6\"]" | tee $(results.array-results.path)
+                echo -n "{\"foo\":\"bar\",\"hello\":\"world\"}" | tee $(results.object-results.path)
     results:
-      - name: echo-indexing-array-results
+      - name: array-results
+        type: array
+        description: whole array
+        value: $(tasks.task1.results.array-results[*])
+      - name: array-indexing-results
```
ywluogg, line 44: Maybe example like:
``` yaml
- name: array-results
        type: array
        description: whole array
        value: ["$(tasks.task2.results.object-results.foo)"]
```
and
``` yaml
- name: object-results
        type: object
        description: whole object
        value:
           key1: $(tasks.task2.results.object-results.foo)
           key2: $(tasks.task2.results.object-results.bar)
```
would also be valuable?

Yongxuanzhang:
> Do you need to add test cases for these in https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_test.go?

you mean test from `reconcile`?

examples/v1beta1/pipelineruns/alpha/pipeline-emitting-results.yaml
Yongxuanzhang: sure

ywluogg:
> 

I think testing from both places seem more robust, since apply.go will be called there

examples/v1beta1/pipelineruns/alpha/pipeline-emitting-results.yaml
Yongxuanzhang: Just realized that these are not supported before, now they are supported! Thanks for catching these cases! @ywluogg

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -30,6 +30,11 @@ import (
 	"github.com/tektoncd/pipeline/pkg/substitution"
 )
 
+const (
+	resultsParseNumber              = 4
```
dibyom, line 34: Could you please add a comment describing what these are

Yongxuanzhang: sure!

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -237,10 +242,12 @@ func ApplyTaskResultsToPipelineResults(
 	results []v1beta1.PipelineResult,
 	taskRunResults map[string][]v1beta1.TaskRunResult,
 	customTaskResults map[string][]v1alpha1.RunResult) []v1beta1.PipelineRunResult {
-
+	// TODO(#4723): Validate array and object results.
```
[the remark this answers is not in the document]
Yongxuanzhang: I will add more details in the comment, we need to validate the pipelineresults reference is 

1. array indexing reference: not out of bound 
2. object element reference: the key is in the dict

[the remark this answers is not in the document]
dibyom: ok, how hard would it be to add the validation in this PR? I wouldn't want this to be released without the validation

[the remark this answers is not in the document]
ywluogg: I think adding a TODO for validation is to to avoid having those XL PRs, but I agree that we shouldn't hurry when the validations are not there. I will move the issue to next release if the validations can't be merged in time

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -268,19 +288,28 @@ func ApplyTaskResultsToPipelineResults(
 						} else {
 							arrayReplacements[v1beta1.StripStarVarSubExpression(variable)] = resultValue.ArrayVal
 						}
+					case v1beta1.ParamTypeObject:
+						objectReplacements[v1beta1.StripStarVarSubExpression(variable)] = resultValue.ObjectVal
```
ywluogg, line 292: I think the validation to check keys against param definition is that compare the resultValue.ObjectVal with PipelineSpec, which is passed in from function

pkg/reconciler/pipelinerun/resources/apply.go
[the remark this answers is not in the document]
ywluogg: I believe most of TEP 0075, 0076 are finished. The only two features with no validations are: pipeline results for object types and array indexing for both pipelines and tasks. I can remove this two features from v0.38 release.

pkg/reconciler/pipelinerun/resources/apply_test.go
```diff
@@ -1612,12 +1613,63 @@ func TestApplyWorkspaces(t *testing.T) {
...
+			"pt1": {
+				{
+					Name:  "foo",
+					Value: *v1beta1.NewArrayOrString("do", "rae", "mi"),
+				},
+			},
+		},
+		expectedResults: nil,
+	}, {
+		description: "array-index-out-of-bound",
```
lbernick, line 1638: it looks like there are some test cases for arrays being added. Could you please update the PR to just focus on object pipeline results, and make any changes to arrays separately?

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -264,31 +286,58 @@ func ApplyTaskResultsToPipelineResults(
...
+								stringReplacements[variable] = resultValue.ArrayVal[intIdx]
+							} else {
+								invalidPipelineResults = append(invalidPipelineResults, pipelineResult.Name)
+								validPipelineResult = false
+							}
 						} else {
 							arrayReplacements[v1beta1.StripStarVarSubExpression(variable)] = resultValue.ArrayVal
 						}
+					case v1beta1.ParamTypeObject:
+						objectReplacements[v1beta1.StripStarVarSubExpression(variable)] = resultValue.ObjectVal
```
lbernick, line 299: as part of #5001 it would be good to move this out of v1beta1. `StripStarVarSubExpression` doesn't seem like a function we'd want to export from that package.

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -264,31 +286,58 @@ func ApplyTaskResultsToPipelineResults(
...
 			})
 		}
 	}
 
-	return runResults
+	if len(invalidPipelineResults) > 0 {
+		return runResults, fmt.Errorf("invalid pipelineresults %v, the referred results don't exist", invalidPipelineResults)
+	}
+
+	return runResults, nil
```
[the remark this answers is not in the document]
lbernick: This seems like something that would be good to split out into a separate PR

lbernick (approved):

pkg/reconciler/pipelinerun/resources/apply.go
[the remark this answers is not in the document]
Yongxuanzhang: This is the part of the validation, I remember we also need validation for this PR? 
Or should I split the validation into another PR?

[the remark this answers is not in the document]
lbernick: oh I see, that's fine-- also I think it's generally OK to add validation in an earlier PR (as opposed to a later one)

pkg/reconciler/pipelinerun/resources/apply.go
Yongxuanzhang: Yeah sure, I will check and fix this

Yongxuanzhang: move this function to substitution

pkg/reconciler/pipelinerun/resources/apply_test.go
Yongxuanzhang: removed the array validation part

PR #5090, opened 2022-07-06
TEP-0075: Validate object name and key name have no dots

chuangw6:
# Changes

Prior to this commit, dots are allowed to be used in param names to
support domain-scoped names. However, ***object*** params have already
supported domain-scoped names since it has a list of keys. In addition,
using dots in object param names and key names have conflicts.
See more details in https://github.com/tektoncd/community/pull/711.

In this change, we validate against those object names and key names
that contain dots.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Dots are not allowed in object param names and key names.
```


imjasonh:
Is `$(params['com.example.foo'])` still considered a valid way of referencing a param name that contains dots? That's been the case that's caused us breakages in the past.

chuangw6:
> Is `$(params['com.example.foo'])` still considered a valid way of referencing a param name that contains dots? That's been the case that's caused us breakages in the past.

Yes! string/array param name can still use dots. And using bracket notation is the way to reference them like the example you commented.

This PR is to make sure only ***object*** param name and its key names do not contain dots.

imjasonh:
> This PR is to make sure only _**object**_ param name and its key names do not contain dots.
> 
> cc @mattmoor

Thanks for clarifying! 👍

dibyom:
@chuangw6 Can we mention this naming limitation for object params in the docs?

chuangw6:
> @chuangw6 Can we mention this naming limitation for object params in the docs?

Added! Also restructure the section `Specifying Parameters` a bit to make it more readable. Please let me know what you think. @dibyom

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -498,31 +501,57 @@ func validateStepArrayUsage(step Step, prefix string, vars sets.String) *apis.Fi
...
-	// - Must begin with a letter or an underscore (_)
-	re := regexp.MustCompile(variableNameFormat)
+	// We've checked param name format. Now, we want to check if param names are referenced correctly in each step
+	for idx, step := range steps {
+		errs = errs.Also(validateStepVariables(ctx, step, prefix, vars).ViaFieldIndex("steps", idx))
+	}
+	return errs
+}
+
+// validateNameFormat validates that the variable name format follows the rules
```
ywluogg, line 511: I feel you can move the comments to the regex itself?

ywluogg:
Re: check-pr-has-kind-label: `2 valid "kind/*" labels found({'kind/cleanup', 'kind/misc'}), expecting exactly one.`

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -498,31 +501,57 @@ func validateStepArrayUsage(step Step, prefix string, vars sets.String) *apis.Fi
...
+// - Must only contain alphanumeric characters, hyphens (-), underscores (_), and dots (.)
+// - Must begin with a letter or an underscore (_)
+// - Exception: Object param name and key names cannot contain dots (.)
+func validateNameFormat(allParamNames sets.String, objectParams []ParamSpec) (errs *apis.FieldError) {
```
ywluogg, line 515: I'm a bir confused why this is not called in validateStepVariables after this validation has been removed from validateVariables?

pkg/apis/pipeline/v1beta1/task_validation.go
chuangw6: done

pkg/apis/pipeline/v1beta1/task_validation.go
chuangw6: name of parameter names. I mean this function.

chuangw6: Talked to @ywluogg offline. We decided to make the function arguments more explicit i.e. the first argument is only for string&array type since the format requirements for those two types are the same; and the second argument is for object type since it requires not containing dots specifically.

PR #5113, opened 2022-07-08
Move two feature flag related functions to config pkg

chuangw6:
# Changes

Those two feature flag related functions are defined in v1beta1_test
packages but used across v1beta1 and v1beta1_test. Since we cannot
import v1beta1_test from v1beta1, moving it to the config package
will be a solution for it to be used across different packages.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/config/feature_flags.go
+++ b/pkg/apis/config/feature_flags.go
@@ -17,6 +17,7 @@ limitations under the License.
 package config
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strconv"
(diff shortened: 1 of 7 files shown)


pkg/apis/config/feature_flags.go
```diff
@@ -187,3 +188,29 @@ func setEmbeddedStatus(cfgMap map[string]string, defaultValue string, feature *s
 func NewFeatureFlagsFromConfigMap(config *corev1.ConfigMap) (*FeatureFlags, error) {
 	return NewFeatureFlagsFromMap(config.Data)
 }
+
+// GetContextBasedOnFeatureFlag returns a context based on the feature flag
+func GetContextBasedOnFeatureFlag(featureFlag string) context.Context {
```
lbernick, line 193: might be worth stating that these functions are for use in testing

lbernick (approved):

pkg/apis/config/feature_flags.go
chuangw6: Added to EnableAlphaAPIFields function.

Feel like `getContextBasedOnFeatureFlag` is overlapping with `EnableAlphaAPIFields` so I just merged and keep `EnableAlphaAPIFields` only. That looks more clean :)

Issue #5141, opened 2022-07-14
TEP0107 - Parameter propagation doesn't work for `taskSpec-steps-args`


chuangw6:
# Expected Behavior
If users want to propagate parameters that are defined in taskrun to variable replacements, all [places](https://github.com/tektoncd/pipeline/blob/d3b8017e652115ac102a5d10fc455240c1331a49/pkg/reconciler/taskrun/resources/apply.go#L253-L323) where param references can be used should be replaced with the actual value of parameters.

# Actual Behavior
Currently, parameter propagation only works for steps-script field, not other fields i.e. steps-args.

# Steps to Reproduce the Problem
1. Try the following yaml file - a modified version of this [example](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/taskruns/alpha/propagating_params_implicit.yaml).

```yaml
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  generateName: hello-
spec:
  params:
    - name: message
      value: "hello world!"
  taskSpec:
    # There are no explicit params defined here. They are derived from the TaskRun.
    steps:
    - name: default
      image: ubuntu
      args: [
        "echo",
        "$(params.message)",
      ]
```

2. You will see the following error

```
validation failed: non-existent variable in "$(params.message)": spec.taskSpec.steps[0].args[1]
```

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.1", GitCommit:"86ec240af8cbd1b60bcc4c03c20da9b98005b92e", GitTreeState:"clean", BuildDate:"2021-12-16T11:41:01Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"linux/amd64"}
  Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.11-gke.1900", GitCommit:"1f775cfbf564daf2c8f0ba548eff09f58045c4a4", GitTreeState:"clean", BuildDate:"2022-04-13T09:36:54Z", GoVersion:"go1.16.15b7", Compiler:"gc", Platform:"linux/amd64"}
  WARNING: version difference between client (1.23) and server (1.21) exceeds the supported minor version skew of +/-1
  ```

- Tekton Pipeline version:

[...]

PR #5143, opened 2022-07-14
Move parameter validation from `taskspec` to `taskrunspec` when propagating parameters

chitrangpatel:
Prior to this, propagating parameters only skipped webhook validation
for params defined in script. As a result, when users tried to propagate
params to other fields like `args` or `command`, etc. an webhook
validation was raised. This PR address issue https://github.com/tektoncd/pipeline/issues/5141. 
This PR addresses validations in `taskspec` and `taskrunspec`. The changes for validations in `pipelinespec` and `pipelinerunspec` will be added in a follow up [PR](https://github.com/tektoncd/pipeline/pull/5291).

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Move parameter validation from `taskspec` to `taskrunspec` when propagating parameters
```

### DIFF
--- a/pkg/apis/pipeline/v1/task_validation.go
+++ b/pkg/apis/pipeline/v1/task_validation.go
@@ -56,6 +56,7 @@ func (t *Task) Validate(ctx context.Context) *apis.FieldError {
 		return nil
 	}
 	errs := validate.ObjectMetadata(t.GetObjectMeta()).ViaField("metadata")
+	ctx = config.SetValidateParameterVariablesAndWorkspaces(ctx, true)
 	return errs.Also(t.Spec.Validate(apis.WithinSpec(ctx)).ViaField("spec"))
 }
 
@@ -361,7 +362,9 @@ func ValidateParameterVariables(ctx context.Context, steps []Step, params []Para
 	}
 
 	errs = errs.Also(validateNameFormat(stringParameterNames.Insert(arrayParameterNames.List()...), objectParamSpecs))
-	errs = errs.Also(validateVariables(ctx, steps, "params", allParameterNames))
+	if config.ValidateParameterVariablesAndWorkspaces(ctx) == true {
+		errs = errs.Also(validateVariables(ctx, steps, "params", allParameterNames))
+	}
 	errs = errs.Also(validateArrayUsage(steps, "params", arrayParameterNames))
 	errs = errs.Also(validateObjectDefault(objectParamSpecs))
 	return errs.Also(validateObjectUsage(ctx, steps, objectParamSpecs))
@@ -563,9 +566,7 @@ func validateStepVariables(ctx context.Context, step Step, prefix string, vars s
 	errs := validateTaskVariable(step.Name, prefix, vars).ViaField("name")
 	errs = errs.Also(validateTaskVariable(step.Image, prefix, vars).ViaField("image"))
 	errs = errs.Also(validateTaskVariable(step.WorkingDir, prefix, vars).ViaField("workingDir"))
-	if !(config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields == "alpha" && prefix == "params") {
-		errs = errs.Also(validateTaskVariable(step.Script, prefix, vars).ViaField("script"))
-	}
+	errs = errs.Also(validateTaskVariable(step.Script, prefix, vars).ViaField("script"))
 	for i, cmd := range step.Command {
 		errs = errs.Also(validateTaskVariable(cmd, prefix, vars).ViaFieldIndex("command", i))
 	}
--- a/pkg/apis/pipeline/v1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1/task_validation_test.go
@@ -53,6 +53,20 @@ func TestTaskValidate(t *testing.T) {
 			ObjectMeta: metav1.ObjectMeta{Name: "task"},
 		},
 		wc: apis.WithinDelete,
+	}, {
+		name: "valid task",
+		t: &v1.Task{
+			ObjectMeta: metav1.ObjectMeta{Name: "task"},
+			Spec: v1.TaskSpec{
+				Steps: []v1.Step{{
+					Name:  "my-step",
+					Image: "my-image",
+					Script: `
+					#!/usr/bin/env  bash
+					echo hello`,
+				}},
+			},
+		},
 	}}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
@@ -63,6 +77,58 @@ func TestTaskValidate(t *testing.T) {
 			err := tt.t.Validate(ctx)
 			if err != nil {
 				t.Errorf("Task.Validate() returned error for valid Task: %v", err)
+			}
+		})
+	}
+}
+
+func TestTaskSpecValidatePropagatedParamsAndWorkspaces(t *testing.T) {
+	type fields struct {
+		Params       []v1.ParamSpec
+		Steps        []v1.Step
+		StepTemplate *v1.StepTemplate
+		Workspaces   []v1.WorkspaceDeclaration
+		Results      []v1.TaskResult
+	}
+	tests := []struct {
+		name   string
+		fields fields
+	}{{
+		name: "propagating params valid step with script",
+		fields: fields{
+			Steps: []v1.Step{{
+				Name:  "propagatingparams",
+				Image: "my-image",
+				Script: `
+				#!/usr/bin/env bash
+				$(params.message)`,
+			}},
+		},
+	}, {
+		name: "propagating params valid step with args",
+		fields: fields{
+			Steps: []v1.Step{{
+				Name:    "propagatingparams",
+				Image:   "my-image",
+				Command: []string{"$(params.command)"},
+				Args:    []string{"$(params.message)"},
+			}},
+		},
+	}}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ts := &v1.TaskSpec{
+				Params:       tt.fields.Params,
+				Steps:        tt.fields.Steps,
+				StepTemplate: tt.fields.StepTemplate,
+				Workspaces:   tt.fields.Workspaces,
+				Results:      tt.fields.Results,
+			}
+			ctx := config.EnableAlphaAPIFields(context.Background())
+			ts.SetDefaults(ctx)
+			ctx = config.SetValidateParameterVariablesAndWorkspaces(ctx, false)
+			if err := ts.Validate(ctx); err != nil {
+				t.Errorf("TaskSpec.Validate() = %v", err)
 			}
 		})
 	}
@@ -1197,6 +1263,7 @@ func TestTaskSpecValidateError(t *testing.T) {
 			}
 			ctx := config.EnableAlphaAPIFields(context.Background())
 			ts.SetDefaults(ctx)
+			ctx = config.SetValidateParameterVariablesAndWorkspaces(ctx, true)
 			err := ts.Validate(ctx)
 			if err == nil {
 				t.Fatalf("Expected an error, got nothing for %v", ts)
@@ -1301,6 +1368,7 @@ func TestStepAndSidecarWorkspacesErrors(t *testing.T) {
 
 			ctx := config.EnableAlphaAPIFields(context.Background())
 			ts.SetDefaults(ctx)
+			ctx = config.SetValidateParameterVariablesAndWorkspaces(ctx, false)
 			err := ts.Validate(ctx)
 			if err == nil {
 				t.Fatalf("Expected an error, got nothing for %v", ts)
@@ -1352,6 +1420,7 @@ func TestStepOnError(t *testing.T) {
 			}
 			ctx := context.Background()
 			ts.SetDefaults(ctx)
+			ctx = config.SetValidateParameterVariablesAndWorkspaces(ctx, false)
 			err := ts.Validate(ctx)
 			if tt.expectedError == nil && err != nil {
 				t.Errorf("TaskSpec.Validate() = %v", err)
@@ -1445,8 +1514,8 @@ func TestIncompatibleAPIVersions(t *testing.T) {
 				if version == "alpha" {
 					ctx = config.EnableAlphaAPIFields(ctx)
 				}
-
 				ts.SetDefaults(ctx)
+				ctx = config.SetValidateParameterVariablesAndWorkspaces(ctx, false)
 				err := ts.Validate(ctx)
 
 				if tt.requiredVersion != version && err == nil {
@@ -1530,6 +1599,7 @@ func TestSubstitutedContext(t *testing.T) {
 			}
 			ctx := context.Background()
 			ts.SetDefaults(ctx)
+			ctx = config.SetValidateParameterVariablesAndWorkspaces(ctx, true)
 			if tt.fields.SubstitutionContext {
 				ctx = config.WithinSubstituted(ctx)
 			}
--- a/pkg/apis/pipeline/v1beta1/task_validation.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation.go
@@ -56,6 +56,7 @@ func (t *Task) Validate(ctx context.Context) *apis.FieldError {
 		return nil
 	}
 	errs := validate.ObjectMetadata(t.GetObjectMeta()).ViaField("metadata")
+	ctx = config.SetValidateParameterVariablesAndWorkspaces(ctx, true)
 	return errs.Also(t.Spec.Validate(apis.WithinSpec(ctx)).ViaField("spec"))
 }
 
@@ -362,9 +363,10 @@ func ValidateParameterVariables(ctx context.Context, steps []Step, params []Para
 			stringParameterNames.Insert(p.Name)
 		}
 	}
-
 	errs = errs.Also(validateNameFormat(stringParameterNames.Insert(arrayParameterNames.List()...), objectParamSpecs))
-	errs = errs.Also(validateVariables(ctx, steps, "params", allParameterNames))
+	if config.ValidateParameterVariablesAndWorkspaces(ctx) == true {
+		errs = errs.Also(validateVariables(ctx, steps, "params", allParameterNames))
+	}
 	errs = errs.Also(validateArrayUsage(steps, "params", arrayParameterNames))
 	errs = errs.Also(validateObjectDefault(objectParamSpecs))
 	return errs.Also(validateObjectUsage(ctx, steps, objectParamSpecs))
@@ -585,9 +587,7 @@ func validateStepVariables(ctx context.Context, step Step, prefix string, vars s
 	errs := validateTaskVariable(step.Name, prefix, vars).ViaField("name")
 	errs = errs.Also(validateTaskVariable(step.Image, prefix, vars).ViaField("image"))
 	errs = errs.Also(validateTaskVariable(step.WorkingDir, prefix, vars).ViaField("workingDir"))
-	if !(config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields == "alpha" && prefix == "params") {
-		errs = errs.Also(validateTaskVariable(step.Script, prefix, vars).ViaField("script"))
-	}
+	errs = errs.Also(validateTaskVariable(step.Script, prefix, vars).ViaField("script"))
 	for i, cmd := range step.Command {
 		errs = errs.Also(validateTaskVariable(cmd, prefix, vars).ViaFieldIndex("command", i))
 	}
--- a/pkg/apis/pipeline/v1beta1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation_test.go
@@ -67,6 +67,20 @@ func TestTaskValidate(t *testing.T) {
 			ObjectMeta: metav1.ObjectMeta{Name: "task"},
 		},
 		wc: apis.WithinDelete,
+	}, {
+		name: "valid task",
+		t: &v1beta1.Task{
+			ObjectMeta: metav1.ObjectMeta{Name: "task"},
+			Spec: v1beta1.TaskSpec{
+				Steps: []v1beta1.Step{{
+					Name:  "my-step",
+					Image: "my-image",
+					Script: `
+					#!/usr/bin/env  bash
+					echo hello`,
+				}},
+			},
+		},
 	}}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
@@ -77,6 +91,60 @@ func TestTaskValidate(t *testing.T) {
 			err := tt.t.Validate(ctx)
 			if err != nil {
 				t.Errorf("Task.Validate() returned error for valid Task: %v", err)
+			}
+		})
+	}
+}
+
+func TestTaskSpecValidatePropagatedParamsAndWorkspaces(t *testing.T) {
+	type fields struct {
+		Params       []v1beta1.ParamSpec
+		Resources    *v1beta1.TaskResources
+		Steps        []v1beta1.Step
+		StepTemplate *v1beta1.StepTemplate
+		Workspaces   []v1beta1.WorkspaceDeclaration
+		Results      []v1beta1.TaskResult
+	}
+	tests := []struct {
+		name   string
+		fields fields
+	}{{
+		name: "propagating params valid step with script",
+		fields: fields{
+			Steps: []v1beta1.Step{{
+				Name:  "propagatingparams",
+				Image: "my-image",
+				Script: `
+				#!/usr/bin/env bash
+				$(params.message)`,
+			}},
+		},
+	}, {
+		name: "propagating params valid step with args",
+		fields: fields{
+			Steps: []v1beta1.Step{{
+				Name:    "propagatingparams",
+				Image:   "my-image",
+				Command: []string{"$(params.command)"},
+				Args:    []string{"$(params.message)"},
+			}},
+		},
+	}}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ts := &v1beta1.TaskSpec{
+				Params:       tt.fields.Params,
+				Resources:    tt.fields.Resources,
+				Steps:        tt.fields.Steps,
+				StepTemplate: tt.fields.StepTemplate,
+				Workspaces:   tt.fields.Workspaces,
+				Results:      tt.fields.Results,
+			}
+			ctx := config.EnableAlphaAPIFields(context.Background())
+			ts.SetDefaults(ctx)
+			ctx = config.SetValidateParameterVariablesAndWorkspaces(ctx, false)
+			if err := ts.Validate(ctx); err != nil {
+				t.Errorf("TaskSpec.Validate() = %v", err)
 			}
 		})
 	}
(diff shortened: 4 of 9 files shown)


chitrangpatel:
cc @chuangw6

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -571,31 +571,32 @@ func validateNameFormat(stringAndArrayParams sets.String, objectParams []ParamSp
...
 	}
 	return errs
 }
 
-func validateTaskVariable(value, prefix string, vars sets.String) *apis.FieldError {
-	return substitution.ValidateVariableP(value, prefix, vars)
+func validateTaskVariable(ctx context.Context, value, prefix string, vars sets.String) *apis.FieldError {
+	if !(config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields == "alpha" && prefix == "params") {
+		return substitution.ValidateVariableP(value, prefix, vars)
+	}
```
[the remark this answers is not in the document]
chitrangpatel: Yes, so I have a way of validating an inline-spec with propagating params. The solution would be to do the validation in `taskrun_validation` on the `taskrun` instead of `task_validation` on the `taskspec`. For inline spec, all the information is available through the `taskrun` so we can do the validation right there. 

However, for referenced resources where we define `task` separately and pipeline/taskrun has a reference to it, there is no way for the validation to know if the original spec was inline or contained references and so it will skip the validation there.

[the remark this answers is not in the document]
chuangw6: > Yes, so I have a way of validating an inline-spec with propagating params. The solution would be to do the validation in `taskrun_validation` on the `taskrun` instead of `task_validation` on the `taskspec`. For inline spec, all the information is available through the `taskrun` so we can do the validation right there.

Sounds reasonable. So with what you proposed, what will happen in `task_validation`? Will they still validate variable references against params declared in spec? If so, since the spec has no declared params (argument `vars` will be empty), the validation will fail right?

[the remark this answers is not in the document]
Yongxuanzhang: yeah the problem seems is that we want to validate params in `task`, with `propagating` feature it can allow us not setting `params` in `task`, they previous validation will fall for this case. And this should not be skipped based on alpha feature flag. So I wonder if having a new flag for this feature makes sense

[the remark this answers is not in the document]
chitrangpatel: > > Yes, so I have a way of validating an inline-spec with propagating params. The solution would be to do the validation in `taskrun_validation` on the `taskrun` instead of `task_validation` on the `taskspec`. For inline spec, all the information is available through the `taskrun` so we can do the validation right there.
> 
> Sounds reasonable. So with what you proposed, what will happen in `task_validation`? Will they still validate variable references against params declared in spec? If so, since the spec has no declared params (argument `vars` will be empty), the validation will fail right?

Yes, exactly. That's why we cannot validate in `task`.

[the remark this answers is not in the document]
chitrangpatel: @chuangw6 @Yongxuanzhang here is an update. 

I figured out a way, we can keep all the validations in place with/without propagating parameters. 

When a `taskrun` is provided, it tries to validate the steps (in case they are implicitly defined) via the `taskrunspec`. In this case, it skips validating a `taskSpec` because its already validated via the `taskRunSpec`.

In case a `task` is provided separately and referenced in the taskrun, it will validate the `taskSpec` and not skip over it.

The details are in the implementation.

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -351,9 +352,10 @@ func ValidateParameterVariables(ctx context.Context, steps []Step, params []Para
 			stringParameterNames.Insert(p.Name)
 		}
 	}
-
 	errs = errs.Also(validateNameFormat(stringParameterNames.Insert(arrayParameterNames.List()...), objectParamSpecs))
-	errs = errs.Also(validateVariables(ctx, steps, "params", allParameterNames))
+	if !skipVariableValidation {
+		errs = errs.Also(validateVariables(ctx, steps, "params", allParameterNames))
+	}
```
Yongxuanzhang, line 358: just curious if we do so, are we going to skip the following  validateArrayUsage, validateObjectDefault as well?

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -329,7 +330,7 @@ func (p ParamSpec) ValidateObjectType() *apis.FieldError {
 }
 
 // ValidateParameterVariables validates all variables within a slice of ParamSpecs against a slice of Steps
-func ValidateParameterVariables(ctx context.Context, steps []Step, params []ParamSpec) *apis.FieldError {
+func ValidateParameterVariables(ctx context.Context, steps []Step, params []ParamSpec, skipVariableValidation bool) *apis.FieldError {
```
chuangw6, line 333: With the flag bool variable, `ctx` is not needed anymore for its downstream callees right?

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -55,10 +55,16 @@ func (ts *TaskRunSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	}
 	// Validate TaskSpec if it's present.
 	if ts.TaskSpec != nil {
-		errs = errs.Also(ts.TaskSpec.Validate(ctx).ViaField("taskSpec"))
+		skipVariableValidation := false
+		// this is for parameter propagation.
+		if config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields == "alpha" {
+			skipVariableValidation = true
+		}
```
Yongxuanzhang, line 62:
```suggestion
skipVariableValidation := (config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields == config.AlphaAPIFields)
```
instead of using "alpha" we can use config.AlphaAPIFields

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -149,7 +149,8 @@ func validatePipelineParameterVariables(ctx context.Context, tasks []PipelineTas
 			}
 		}
 	}
-	return errs.Also(validatePipelineParametersVariables(tasks, "params", parameterNames, arrayParameterNames, objectParameterNameKeys))
+	errs = errs.Also(validatePipelineParametersVariables(tasks, "params", parameterNames, arrayParameterNames, objectParameterNameKeys))
+	return errs
```
chuangw6, line 153: nit: we don't need to separate it into two lines, do we?

pkg/apis/pipeline/v1beta1/task_validation.go
chitrangpatel: we need it downstream for some other calls. eg. validateVariable

pkg/apis/pipeline/v1beta1/task_validation.go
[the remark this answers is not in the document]
chuangw6: I am wondering if it is a valid case that some params (i.e. `a`, `b`, `c`) are propagated from taskrun and others (i.e. `d`, `e`) are defined explicitly in taskspec.

In this case, with your approach, are we going to just skip validations against `a`, `b`, `c`, but the validation against `d`, `e` is still retained?

pkg/apis/pipeline/v1beta1/taskrun_validation.go
chitrangpatel: Good catch! I will fix that. Thanks

pkg/apis/pipeline/v1beta1/task_validation.go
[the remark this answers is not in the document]
chitrangpatel: No, what will happen is that we will do is attempt to validate `a`, `b` and `c` using `validateInlineParameters` defined in `taskrun_validation.go`. It will also validate `d` and `e` using the regular `task_validation` if the task is referenced.

pkg/apis/pipeline/v1beta1/task_validation.go
chitrangpatel: I don't think so. Those checks are done separately I think.

chitrangpatel:
@jerop this is now ready for review. Thanks in advance 😄

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -1524,11 +1579,11 @@ func TestIncompatibleAPIVersions(t *testing.T) {
 			t.Run(testName, func(t *testing.T) {
 				ts := tt.spec
 				ctx := context.Background()
+				ts.SetDefaults(ctx)
 				if version == "alpha" {
-					ctx = config.EnableAlphaAPIFields(ctx)
```
JeromeJu, line 1528: Do we have any specific reason to move `ts.SetDefaults(ctx)` before the alpha condition block?

pkg/apis/pipeline/v1beta1/pipeline_types_test.go
```diff
@@ -279,6 +279,7 @@ func TestPipelineTask_ValidateRegularTask_Success(t *testing.T) {
 				cfg.FeatureFlags.EnableTektonOCIBundles = true
 			}
 			ctx = config.ToContext(ctx, cfg)
+			ctx = context.WithValue(ctx, ValidateParameterVariablesAndWorkspaces, true)
```
lbernick, line 282: I'm curious why this validation is going in a call to `context.withvalue` (as opposed to being part of (PipelineSpec).Validate), and why it is added to tests?

pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
```diff
@@ -107,14 +110,79 @@ func (ps *PipelineRunSpec) Validate(ctx context.Context) (errs *apis.FieldError)
...
 	}
-
 	for idx, trs := range ps.TaskRunSpecs {
 		errs = errs.Also(validateTaskRunSpec(ctx, trs).ViaIndex(idx).ViaField("taskRunSpecs"))
 	}
 
 	return errs
 }
 
+func (ps *PipelineRunSpec) validateInlineParameters(ctx context.Context) (errs *apis.FieldError) {
```
lbernick, line 120: could you please add some comments for this function describing what is being validated?
Also, can this be consolidated with existing logic for param validation?

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -351,9 +358,11 @@ func ValidateParameterVariables(ctx context.Context, steps []Step, params []Para
 			stringParameterNames.Insert(p.Name)
 		}
 	}
-
 	errs = errs.Also(validateNameFormat(stringParameterNames.Insert(arrayParameterNames.List()...), objectParamSpecs))
-	errs = errs.Also(validateVariables(ctx, steps, "params", allParameterNames))
+	log.Println(ctx.Value(ValidateParameterVariablesAndWorkspaces))
```
lbernick, line 362: extra log line here

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -56,6 +56,8 @@ var invalidSteps = []v1beta1.Step{{
 	Image: "myimage",
 }}
 
+// const v1beta1.ValidateParameterVariablesAndWorkspaces = validateEmbeddedVariables("v1beta1.ValidateParameterVariablesAndWorkspaces")
```
lbernick, line 59: extra comment here

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -93,6 +97,38 @@ func (ts *TaskRunSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	return errs
 }
 
+// validateInlineParameters validates parameters if embedded
```
lbernick, line 100: Same comments as for inline parameter validation-- could you please add more detail to this docstring, and see if it can be combined with any existing parameter validation?

lbernick (commented):
thanks Chitrang! Please rebase, and make sure these changes are reflected in v1 as well

pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
```diff
@@ -107,14 +110,79 @@ func (ps *PipelineRunSpec) Validate(ctx context.Context) (errs *apis.FieldError)
...
+	}
+	for _, p := range ps.Params {
+		pSpec := ParamSpec{
+			Name:    p.Name,
+			Default: &p.Value,
+		}
+		paramSpec = append(paramSpec, pSpec)
+	}
+	for _, p := range ps.PipelineSpec.Params {
+		skip := false
```
JeromeJu, line 133: It seems to me like we can extract out some generic functions for the check here with the skip condition.

pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
```diff
@@ -107,14 +110,79 @@ func (ps *PipelineRunSpec) Validate(ctx context.Context) (errs *apis.FieldError)
...
+		return errs
+	}
+	for _, p := range ps.Params {
+		pSpec := ParamSpec{
+			Name:    p.Name,
+			Default: &p.Value,
+		}
+		paramSpec = append(paramSpec, pSpec)
+	}
+	for _, p := range ps.PipelineSpec.Params {
```
JeromeJu, line 132: It seems that we might be able to extract out some generic helper func for the check here with the skip condition?

chitrangpatel:
@lbernick I accidentally re-requested review. Sorry about that. Please hold off until I implement your suggestions.

chitrangpatel:
> thanks Chitrang! Please rebase, and make sure these changes are reflected in v1 as well

Hi @lbernick, should I only update the files in `V1` that are similar to those in`v1beta1` or also others? I ask because I implemented some changes in files like `pipeline_types.go and pipelinerun_validation.go` which don't exist in `v1`. Should they be added there as well?

lbernick:
> should I only update the files in `V1` that are similar to those in`v1beta1` or also others? I ask because I implemented some changes in files like `pipeline_types.go and pipelinerun_validation.go` which don't exist in `v1`. Should they be added there as well?

Until we release v1 (and stop adding new features to v1beta1), we'll have to keep these two directories in sync. Any changes to v1beta1 should also be made to v1. The pipeline validation can be done in a separate PR though.

chitrangpatel:
> > should I only update the files in `V1` that are similar to those in`v1beta1` or also others? I ask because I implemented some changes in files like `pipeline_types.go and pipelinerun_validation.go` which don't exist in `v1`. Should they be added there as well?
> 
> Until we release v1 (and stop adding new features to v1beta1), we'll have to keep these two directories in sync. (This is a good reason to factor any non-api related logic out of pkg/apis/pipeline.) Any changes to v1beta1 should also be made to v1. The pipeline validation can be done in a separate PR though.

Sure, I can make the changes to the files that exist in v1. What about the files that don't exist? Should I make a copy there and change package names etc.? Or should I make that into a separate PR (I think that's what you were indicating but I wanted to make sure) where I don't need to worry about V1?

lbernick:
> Sure, I can make the changes to the files that exist in v1. What about the files that don't exist? Should I make a copy there and change package names etc.? Or should I make that into a separate PR (I think that's what you were indicating but I wanted to make sure) where I don't need to worry about V1?

Oh I see-- sorry I misunderstood; @JeromeJu is working on pipeline_types for v1 in #5219 so just sync up with him to make sure these changes make their way into the v1 package, depending on which of these PRs is merged first.

chitrangpatel:
After discussion with @JeromeJu, I think I will break this PR into handling task validations and pipeline validations separately so that v1 and v1beta1 can be kept in sync. For now, I will only open a PR for task validations. Once https://github.com/tektoncd/pipeline/pull/5219 is merged, I will sync my changes to the pipeline validations and open a PR for it.

jerop (commented):
/retest

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -55,10 +55,12 @@ func (ts *TaskRunSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	}
 	// Validate TaskSpec if it's present.
 	if ts.TaskSpec != nil {
+		// skip validation since we validate them via taskrunspec below.
```
jerop, line 58:
```suggestion
		// skip validation of parameter and workspaces variables since we validate them via taskrunspec below.
```

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -93,6 +95,43 @@ func (ts *TaskRunSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	return errs
 }
 
+// validateInlineParameters validates parameters that are defined inline.
+// This is cruicial for propagated parameters since the parameters could
```
jerop, line 99:
```suggestion
// This is crucial for propagated parameters since the parameters could
```

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -93,6 +95,43 @@ func (ts *TaskRunSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
...
+// be defined under taskRun and then called directly in the task steps.
+// In this case, paratemers cannot be validated by the underlying taskSpec
```
jerop, line 101:
```suggestion
// In this case, parameters cannot be validated by the underlying taskSpec
```

jerop (approved):
thanks @chitrangpatel!

please add details to the commit that this addresses `taskspec` and `taskrunspec`, and that the changes for `pipelinespec` and `pipelinerunspec` will be added in a follow up PR

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -93,6 +95,43 @@ func (ts *TaskRunSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
...
+// since they may not have the parameters declared because of propagation.
+func (ts *TaskRunSpec) validateInlineParameters(ctx context.Context) (errs *apis.FieldError) {
```
lbernick, line 103: Suggested first line of docstring: "validateInlineParameters validates that any parameters called in the Task spec are declared in the TaskRun. this is crucial for propagated parameters because...." (This makes it clearer what is valid vs invalid)

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -93,6 +95,43 @@ func (ts *TaskRunSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
...
+	errs = errs.Also(ValidateParameters(ctx, ts.Params).ViaField("params"))
+	params := ts.Params
+	var paramSpec []ParamSpec
+	if ts.Params == nil || ts.TaskSpec == nil {
```
lbernick, line 107: it seems like it would be a problem if the task spec used some propagated params, but the taskrun didn't declare them. Is this case tested?

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -93,6 +95,43 @@ func (ts *TaskRunSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
...
+	}
+	for _, p := range params {
+		pSpec := ParamSpec{
+			Name:    p.Name,
+			Default: &p.Value,
+		}
+		paramSpec = append(paramSpec, pSpec)
+	}
+	for _, p := range ts.TaskSpec.Params {
+		skip := false
```
lbernick, line 118: would a set be useful here? we use `sets.string` within the codebase. You could also do a map of param name to default.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -700,6 +700,8 @@ func (c *Reconciler) createPod(ctx context.Context, ts *v1beta1.TaskSpec, tr *v1
 	// Apply workspace resource substitution
 	ts = resources.ApplyWorkspaces(ctx, ts, ts.Workspaces, tr.Spec.Workspaces, workspaceVolumes)
 
+	// By this time, params and workspaces should be propagated down so we can continue the check
```
lbernick, line 703:
```suggestion
	// By this time, params and workspaces should be propagated down so we can validate that all parameter variables and workspaces used in the TaskSpec are declared by the Task
```

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -433,6 +468,7 @@ func TestTaskSpecValidate(t *testing.T) {
 			}
 			ctx := config.EnableAlphaAPIFields(context.Background())
 			ts.SetDefaults(ctx)
+			ctx = context.WithValue(ctx, v1beta1.ValidateParameterVariablesAndWorkspaces, false)
```
lbernick, line 471: curious why any of the tests other than the ones with parameter propagation have validation = false?

JeromeJu:
> After discussion with @JeromeJu, I think I will break this PR into handling task validations and pipeline validations separately so that v1 and v1beta1 can be kept in sync. For now, I will only open a PR for task validations. Once #5219 is merged, I will sync my changes to the pipeline validations and open a PR for it.

Thanks Chitrang!

pkg/apis/pipeline/v1/task_validation.go
```diff
@@ -44,6 +46,9 @@ const (
 	// objectVariableNameFormat is the regext used to validate object name and key names format
 	// The difference with the array or string name format is that object variable names shouldn't contain dots.
 	objectVariableNameFormat = "^[_a-zA-Z][_a-zA-Z0-9-]*$"
+
+	// ValidateParameterVariablesAndWorkspaces provides the toggle to skip validation of parameters when embedded vs referenced.
+	ValidateParameterVariablesAndWorkspaces = validateEmbeddedVariables("ValidateParameterVariablesAndWorkspaces")
```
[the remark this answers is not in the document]
lbernick: thanks for moving to contexts.go; I still have the same question about a string alias vs named constant.

pkg/apis/pipeline/v1/task_validation_test.go
```diff
@@ -77,8 +91,9 @@ func TestTaskSpecValidate(t *testing.T) {
 		Results      []v1.TaskResult
 	}
 	tests := []struct {
-		name   string
```
lbernick, line 80: It might be cleaner to have one test class where you do validate params + workspaces, and one where you don't.

pkg/apis/pipeline/v1beta1/taskrun_validation.go
lbernick: I don't think this comment was addressed but please correct me if I'm wrong

pkg/apis/pipeline/v1beta1/taskrun_validation.go
lbernick: I don't think this comment was addressed.

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
```diff
@@ -64,6 +64,110 @@ func TestTaskRun_Validate(t *testing.T) {
...
+			Spec: v1beta1.TaskRunSpec{
+				Params: []v1beta1.Param{{
+					Name: "task-words",
+					Value: v1beta1.ArrayOrString{
+						ArrayVal: []string{"hello", "task run"},
+					},
+				}},
+				TaskSpec: &v1beta1.TaskSpec{
+					Params: []v1beta1.ParamSpec{{
+						Name: "default-task-words",
```
lbernick, line 102: this test case is a bit confusing-- I'm understanding this as the task declares a totally separate param that happens to not be used, but the name of the separate param is "default" + the name of the param that is used. What's this trying to test?

pkg/apis/pipeline/v1/task_validation.go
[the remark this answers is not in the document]
lbernick: Got it! thanks for the explanation!

pkg/apis/pipeline/v1beta1/taskrun_validation.go
lbernick: I think what I am looking for is a test case like:
```
		taskRun: &v1beta1.TaskRun{
			ObjectMeta: metav1.ObjectMeta{Name: "tr"},
			Spec: v1beta1.TaskRunSpec{
				TaskSpec: &v1beta1.TaskSpec{
					Steps: []v1beta1.Step{{
						Name:    "echo",
						Image:   "ubuntu",
						Command: []string{"echo"},
						Args:    []string{"$(params.task-words[*])"},
					}, 
				},
			},
		},
```
(this is invalid.) Does this exist anywhere?

pkg/apis/pipeline/v1beta1/taskrun_validation.go
lbernick: The idea behind this comment is that you are sort of implementing a set intersection on param names. However I think what you have is fine, not sure if there's a cleaner way.

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
```diff
@@ -64,6 +64,120 @@ func TestTaskRun_Validate(t *testing.T) {
...
+						Type: v1beta1.ParamTypeArray,
+					},
+						{
```
lbernick, line 136: could you fix the formatting here please?

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
lbernick: I think what's confusing is specifically referring to params defined inline as "default" params. (as you know) params can have defaults, which is separate from the concept of a param being a "default param". The idea of param defaults only really makes sense in a task ref (rather than an inline task spec) IMO. I think this test case would be less confusing if you just removed "default" from all the variable names, and changed "default-task-words" and "default-echo" to different names. However if you disagree that's fine.

chitrangpatel: Ah ok. Sure I can remove `default` from there. I will post here once I update the code.

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
chitrangpatel: Yes, of course.

pkg/apis/pipeline/v1beta1/taskrun_validation.go
chitrangpatel: Thanks! it makes sense now. It would just return without throwing an error. I now added a test [here](https://github.com/tektoncd/pipeline/pull/5143/files#diff-1055ae489118d479fe9b003ef2753af20cfe8b2589bf866cb25f5228fa242b68) to capture it.

Thanks a lot!!!

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
chitrangpatel: Done.

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
chitrangpatel: Done

jerop:
TestPipelineRunTimeout again 🙃 

/retest

PR #5134, opened 2022-07-13
Switch PipelineRun timeout -> TaskRun logic to instead signal the TaskRuns to stop

abayer:
# Changes

Fixes #5127, #3460

An alternative approach vs #5133, attempting to eliminate the problem entirely rather than just working around it.

As noted in #5127, the logic around calculating a timeout for a `PipelineRun`'s `TaskRun` to make sure that the `TaskRun`'s timeout is always going to end before the `PipelineRun`'s timeout ends is problematic. It can result in race conditions where a `TaskRun` gets timed out, immediately followed by the `PipelineRun` being reconciled while not yet having hit the end of its own timeout. This change gets rid of that behavior, and instead sets the `TaskRun.Spec.Status` to a new value, `TaskRunTimedOut`, with the `TaskRun` reconciler handling that in the same way it does setting `TaskRun.Spec.Status` to `TaskRunCancelled`.

By doing this, we can unify the behavior for both `TaskRun`s and `Run`s upon `PipelineRun`s timing out, given that we already cancel `Run`s upon `PipelineRun` timeout, and we can kill off a bunch of flaky outcomes for `PipelineRun`s.

Also, rather than setting a 1s timeout on tasks which are created after the `.timeouts.pipeline`, `.timeouts.tasks`, or `.timeouts.finally` timeouts have been passed, this change will skip creation of those `TaskRun`s or `Run`s completely. However, it will still report those "tasks" as failed, not skipped, since attempted creation of tasks after the appropriate timeout has been reached is a failure case, and that's the existing expected behavior for `PipelineTask`s which try to start after timeouts have elapsed.

Huge thanks to @jerop for suggesting this approach!

/kind bug

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
[...]

PR #5145, opened 2022-07-16
Prepare tests for impending knative/pkg version bump

imjasonh:
https://github.com/knative/pkg/pull/2548 changes pkg/changeset.Get to
read VCS information from information embedded by the Go compiler into
built binaries, instead of relying on the convention that a kodata/HEAD
symlink exists pointing to the repo's .git/HEAD, and a kodata/refs
points to .git/refs.

This fails in tests though, since tests don't stamp this information
into the binary where it can be read.

This change removes our tests' usage of changeset.Get and instead
replaces those fake commit SHAs with more obviously placeholder values.
It doesn't seem that the tests cared that it was an actual SHA, let
along the actual current SHA, they just needed any value that looked
like a SHA.

/kind cleanup

/assign @vdemeester 
/assign @abayer 
/release-note-none
cc @dprotaso 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

---

PR friction log:
- scolded for not adding a release note label, because the PR template included a space between ``` and `release-note` -- was this intentional?
- integration test flake: https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5145/pull-tekton-pipeline-alpha-integration-tests/1548197852407140352

### DIFF
--- a/pkg/pod/pod_test.go
+++ b/pkg/pod/pod_test.go
@@ -69,7 +69,7 @@ var (
 	volumeMountSort = cmpopts.SortSlices(func(i, j corev1.VolumeMount) bool { return i.Name < j.Name })
 )
 
-const fakeVersion = "0000000000000000000000000000000000000000"
+const fakeVersion = "a728ce3"
 
 func init() {
 	os.Setenv("KO_DATA_PATH", "./testdata/")
--- a/pkg/pod/pod_test.go
+++ b/pkg/pod/pod_test.go
@@ -41,7 +41,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	fakek8s "k8s.io/client-go/kubernetes/fake"
-	"knative.dev/pkg/changeset"
 	logtesting "knative.dev/pkg/logging/testing"
 	"knative.dev/pkg/system"
 
@@ -63,8 +62,6 @@ var (
 
 	defaultActiveDeadlineSeconds = int64(config.DefaultTimeoutMinutes * 60 * deadlineFactor)
 
-	fakeVersion string
-
 	resourceQuantityCmp = cmp.Comparer(func(x, y resource.Quantity) bool {
 		return x.Cmp(y) == 0
 	})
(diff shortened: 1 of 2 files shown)


imjasonh:
Note to self: After knative/pkg#2548 merges and we pick it up, we can also remove 

https://github.com/tektoncd/pipeline/blob/8a7b0cfa755038f4cfdcd88c314a72a90bcab1a2/pkg/reconciler/taskrun/taskrun_test.go#L444

and 

https://github.com/tektoncd/pipeline/blob/8a7b0cfa755038f4cfdcd88c314a72a90bcab1a2/pkg/pod/pod_test.go#L76

and 

https://github.com/tektoncd/pipeline/tree/8a7b0cfa755038f4cfdcd88c314a72a90bcab1a2/pkg/pod/testdata

and

https://github.com/tektoncd/pipeline/tree/8a7b0cfa755038f4cfdcd88c314a72a90bcab1a2/pkg/reconciler/taskrun/testdata

imjasonh:
Flake

abayer:
> Flake
…and I have a fix for that damned flake at #5134, btw. =)

dprotaso:
You'll want to bump go.mod's go version to `go1.18` the vcs sha embedding happens in that version

imjasonh:
> You'll want to bump go.mod's go version to `go1.18` the vcs sha embedding happens in that version

The version in `go.mod` only controls what module semantics are used, not the version of the compiler that's used.

Our release process seems to already be using an image based on Go 1.18, here:

https://github.com/tektoncd/pipeline/blob/8a7b0cfa755038f4cfdcd88c314a72a90bcab1a2/tekton/publish.yaml#L118

The Go inside that image is Go 1.18.1, which should work just fine for this change.

(That's not to say updating the go version in `go.mod` is a bad idea, just that it won't control whether we get the necessary build info embedded in our binaries, we should have those already)

vdemeester (approved):

abayer:
> Our release process seems to already be using an image based on Go 1.18

Will this have any implications for running tests, `ko`, etc, with Go 1.17? If we do need to move to building with Go 1.18 everywhere to pick up this change, I'd definitely prefer if we bumped `go.mod` too - I know the `go 1.17` line is just specifying module semantics, but it's the best shorthand I know of to tell devs "you need to use this version of Go".

Also, we've got Go 1.17.x on a bunch of our images, such as here in `test-runner`: https://github.com/tektoncd/plumbing/blob/main/tekton/images/test-runner/Dockerfile#L151-L156

...which I'd assume we'll need to bump to Go 1.18.x as well?

Issue #5146, opened 2022-07-16
Reconcilers should rely on knative/pkg to Update TaskRuns and PipelineRuns


imjasonh:
# Background

The TaskRun and PipelineRun controllers are invoked by Knative controller code that calls `ReconcileKind(ctx, tr)` with the result of a K8s watch on resources, then determines if those resources need to be updated, and if so, calls `Update` on them.

This is good, because Knative code handles the watching and updating for us, and only has to assume that any changes Tekton wants to make to resources is expressed in terms of updates to the `tr` object passed to `ReconcileKind`.

knative/pkg expects this:

```
knative/pkg calls Watch
  -> ReconcileKind(tr)
    -> if tr has changed, knative/pkg calls Update
```

However, our `ReconcileKind`s for TaskRuns and PipelineRuns (notably, not Runs) also make their own calls `Update`, as part of `updateLabelsAndAnnotations`:

https://github.com/tektoncd/pipeline/blob/8a7b0cfa755038f4cfdcd88c314a72a90bcab1a2/pkg/reconciler/pipelinerun/pipelinerun.go#L1171-L1186

https://github.com/tektoncd/pipeline/blob/8a7b0cfa755038f4cfdcd88c314a72a90bcab1a2/pkg/reconciler/taskrun/taskrun.go#L548-L573

Effectively, what we're doing is:

```
knative/pkg calls Watch
  -> ReconcileKind(pr)
    -> if pr has changed, Tekton calls Update
      -> if pr has changed (unlikely), knative/pkg calls Update
```

This leads to an unnecessary and unexpected change in the control of the update lifecycle for both PipelineRuns and TaskRuns. A method unsuspiciously called `updateLabelsAndAnnotations` is actually responsible for persisting all the changes we make during a reconciliation.

This isn't a _problem_ per se, it's just unexpected, and not how Knative reconciler code expects to be used, which may cause confusion and bugs later. It may also lead to duplicate calls to `Update`, if diffs sneak in after our call to `updateLabelsAndAnnotations`, which can lead to cluster-destabilizing write load on the K8s API server in heavy users of Tekton.

[...]

PR #5151, opened 2022-07-18
Rely on knative/pkg to update TaskRuns

imjasonh:
This changes the TaskRun reconciler to never call Update itself;
instead, Knative controller infrastructure will be responsible for
Updating TaskRuns if it has determined they've changed.
    
This also changes changeset detection to happen once at init-time, since
it doesn't change for the lifetime of the controller.
    
finishReconcileUpdateEmitEvents is now only responsible for emitting
events. This could be simplified even further in a later change.

Proof of concept for https://github.com/tektoncd/pipeline/issues/5146

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

PR friction log:
- the kind label wasn't set when I created the PR with `/kind cleanup`, causing CI ❌ 
- manually adding the label didn't cause the label check to run, so still ❌

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -20,11 +20,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"reflect"
 	"strings"
 	"time"
 
-	"github.com/hashicorp/go-multierror"
 	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/pod"
@@ -48,7 +46,6 @@ import (
 	_ "github.com/tektoncd/pipeline/pkg/taskrunmetrics/fake" // Make sure the taskrunmetrics are setup
 	"github.com/tektoncd/pipeline/pkg/workspace"
 	resolution "github.com/tektoncd/resolution/pkg/resource"
-	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


vdemeester:
/test all
(Just realized it was skipping CI.. 🙃 )

imjasonh:
It seems something somewhere was depending on this to persist the task labels or something; this fails on the only test that checks that the task label is persisted on the taskrun.

```
    taskrun_test.go:1983: Labels were not added to task run
    taskrun_test.go:1986: Did not get expected label (-want, +got):   string(
        -       "test-task",
        +       "",
          )
```

Anyway, I'll keep plugging away later.

afrittoli:
> PR friction log:
> 
> * the kind label wasn't set when I created the PR with `/kind cleanup`, causing CI ❌
> * manually adding the label didn't cause the label check to run, so still ❌

Thank you for tracking these. They should be fixed now.

vdemeester:
@imjasonh what's the status ? 🙃 (it's red right now because of the rebase required I guess ?)

imjasonh:
Aside from needing to rebase, I need to figure out why label propagation is broken by this.

I _think_ we're relying on there being two updates of the object in etcd during the initial reconcile, so that cloudevent receivers can see the TaskRun has started, then also see it become Pending (I think). The model knative/pkg expects is that each time through `ReconcileKind` is zero-or-one etcd update, which conflicts with our usage. I think their model is simpler to understand than ours, and I'd like to get to it, without breaking cloudevents consumers.

...Then I got distracted with other stuff and haven't made it back to work on this. I'll rebase at least so it's not so red, but any help untangling this or better understanding of how this works would be helpful.

afrittoli:
> Aside from needing to rebase, I need to figure out why label propagation is broken by this.
> 
> I _think_ we're relying on there being two updates of the object in etcd during the initial reconcile, so that cloudevent receivers can see the TaskRun has started, then also see it become Pending (I think). The model knative/pkg expects is that each time through `ReconcileKind` is zero-or-one etcd update, which conflicts with our usage. I think their model is simpler to understand than ours, and I'd like to get to it, without breaking cloudevents consumers.

The `started` event and `running` events are sent during the same reconcile cycle, because I didn't want to rely on two reconcile cycles for that. Not sure why labels would not work... I wonder if it's something about how tests are designed?

vdemeester:
Let me try to tackle this for `TaskRun` *and* `PipelineRun` 🤔

PR #5169, opened 2022-07-19
Add validation for results object properties types

Yongxuanzhang:
# Changes
This commit adds the validation for results object properties types.
Right now we only support string types, so if the type is not string we
should validate and return err for this.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1/result_validation.go
+++ b/pkg/apis/pipeline/v1/result_validation.go
@@ -35,7 +35,9 @@ func (tr TaskResult) Validate(ctx context.Context) (errs *apis.FieldError) {
 	}
 	// Array and Object are alpha features
 	if tr.Type == ResultsTypeArray || tr.Type == ResultsTypeObject {
-		return errs.Also(version.ValidateEnabledAPIFields(ctx, "results type", config.AlphaAPIFields))
+		errs := validateObjectResult(tr)
+		errs = errs.Also(version.ValidateEnabledAPIFields(ctx, "results type", config.AlphaAPIFields))
+		return errs
 	}
 
 	// Resources created before the result. Type was introduced may not have Type set
@@ -49,5 +51,28 @@ func (tr TaskResult) Validate(ctx context.Context) (errs *apis.FieldError) {
 		return apis.ErrInvalidValue(tr.Type, "type", fmt.Sprintf("type must be string"))
 	}
 
+	return nil
+}
+
+// validateObjectResult validates the object result and check if the Properties is missing
+// for Properties values it will check if the type is string.
+func validateObjectResult(tr TaskResult) (errs *apis.FieldError) {
+	if ParamType(tr.Type) == ParamTypeObject && tr.Properties == nil {
+		return apis.ErrMissingField(fmt.Sprintf("%s.properties", tr.Name))
+	}
+
+	invalidKeys := []string{}
+	for key, propertySpec := range tr.Properties {
+		if propertySpec.Type != ParamTypeString {
+			invalidKeys = append(invalidKeys, key)
+		}
+	}
+
+	if len(invalidKeys) != 0 {
+		return &apis.FieldError{
+			Message: fmt.Sprintf("The value type specified for these keys %v is invalid, the type must be string", invalidKeys),
+			Paths:   []string{fmt.Sprintf("%s.properties", tr.Name)},
+		}
+	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


pkg/apis/pipeline/v1/result_validation.go
```diff
@@ -35,7 +35,9 @@ func (tr TaskResult) Validate(ctx context.Context) (errs *apis.FieldError) {
 	}
 	// Array and Object are alpha features
 	if tr.Type == ResultsTypeArray || tr.Type == ResultsTypeObject {
-		return errs.Also(version.ValidateEnabledAPIFields(ctx, "results type", config.AlphaAPIFields))
+		errs := errs.Also(validateObjectResult(tr))
```
lbernick, line 38: nit: there's no need for errs.also here

pkg/apis/pipeline/v1/result_validation.go
```diff
@@ -51,3 +53,25 @@ func (tr TaskResult) Validate(ctx context.Context) (errs *apis.FieldError) {
...
+	invalidKeys := []string{}
+	for key, propertySpec := range tr.Properties {
+		if propertySpec.Type != ParamTypeString {
+			invalidKeys = append(invalidKeys, key)
+		}
+	}
+
+	if len(invalidKeys) != 0 {
+		return &apis.FieldError{
+			Message: fmt.Sprintf("The value type specified for these keys %v is invalid", invalidKeys),
```
lbernick, line 72: maybe include in the message that properties must be strings?

pkg/apis/pipeline/v1/result_validation.go
```diff
@@ -51,3 +53,25 @@ func (tr TaskResult) Validate(ctx context.Context) (errs *apis.FieldError) {
 
 	return nil
 }
+
+
+func validateObjectResult(tr TaskResult) (errs *apis.FieldError) {
```
lbernick, line 58: could you please add a docstring for this function?

pkg/apis/pipeline/v1/result_validation.go
Yongxuanzhang: oh yes, thx!

pkg/apis/pipeline/v1/result_validation.go
Yongxuanzhang: sure

pkg/apis/pipeline/v1/result_validation.go
Yongxuanzhang: np!

pkg/apis/pipeline/v1/result_validation.go
```diff
@@ -51,3 +53,26 @@ func (tr TaskResult) Validate(ctx context.Context) (errs *apis.FieldError) {
 
 	return nil
 }
+
+// validateObjectResult validates the object result and check if the Properties is missing
+// for Properties values it will check if the type is string.
```
lbernick, line 58:
```suggestion
// validateObjectResult validates that the object result has Properties,
// and that all Properties are strings.
```

lbernick: Also, is there a reason PropertySpec has a type if the only supported type is string?

lbernick (approved):

pkg/apis/pipeline/v1/result_validation.go
Yongxuanzhang: yeah it is required in [tep75](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#use-cases) and I think the reason is to support future extension

pkg/apis/pipeline/v1/result_validation.go
```diff
@@ -51,3 +53,26 @@ func (tr TaskResult) Validate(ctx context.Context) (errs *apis.FieldError) {
...
+	invalidKeys := []string{}
+	for key, propertySpec := range tr.Properties {
+		if propertySpec.Type != ParamTypeString {
+			invalidKeys = append(invalidKeys, key)
+		}
+	}
+
+	if len(invalidKeys) != 0 {
+		return &apis.FieldError{
+			Message: fmt.Sprintf("The value type specified for these keys %v is invalid, the type must be string", invalidKeys),
```
ywluogg, line 73:
```suggestion
			Message: fmt.Sprintf("The value type specified for these keys %v are invalid, the type must be string", invalidKeys),
```

ywluogg (commented):
/lgtm

PR #5171, opened 2022-07-19
Fix flaky test by sorting slices

dibyom:
# Changes

TestGetVarSubstitutionExpressionsForPipelineResult was flaky since the expected
test output was in a fixed order while the function under test was iterating
through a map where order is not guaranteed. This commit fixes it by sorting
the inputs to cmp.Diff

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/resultref_test.go
+++ b/pkg/apis/pipeline/v1beta1/resultref_test.go
@@ -754,10 +754,13 @@ func TestGetVarSubstitutionExpressionsForPipelineResult(t *testing.T) {
 		want: []string{"tasks.task1.results.result1", "tasks.task2.results.result2", "tasks.task3.results.result3"},
 	},
 	}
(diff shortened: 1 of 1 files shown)


jerop (approved):

lbernick (approved):
you'll also have to change the v1 package unfortunately

FYI @Yongxuanzhang

dibyom:
> you'll also have to change the v1 package unfortunately

I don't see a resultref in the v1 package. Is that intentional?

lbernick:
/lgtm

> I don't see a resultref in the v1 package. Is that intentional?

oops-- should have looked more carefully

> (I also don't see a CRD struct in result ref - maybe we could just move it out of v1beta1?)

good idea!

Yongxuanzhang:
oh thank you!

Issue #5175, opened 2022-07-20
Parameters should default to type string


jerop:
# Expected Behavior

`Pipelines` with `Parameters` that have no `"type"` specified work with the default value of `"string"`.

# Actual Behavior

`Pipelines` with `Parameters` that have no `"type"` fail. 

Example failures in dogfooding are here: https://dashboard.dogfooding.tekton.dev/#/namespaces/default/pipelineruns

# Steps to Reproduce the Problem

1. Create a `Pipeline` with a `Parameter` without a default value specified.
2. Run the `Pipeline`
3. Observe the `PipelineRun` logs.

# Additional Info

Ran into this error when running the release `Pipeline` which uses the default types for `Parameters` - https://github.com/tektoncd/pipeline/blob/94055d92c120a6010f3d61a821e45dea4f893a74/tekton/release-pipeline.yaml

This was also observed in other `Pipelines` in dogfooding - https://github.com/tektoncd/plumbing/pull/1148

A related issue with `types` in `Results` was fixed in https://github.com/tektoncd/pipeline/pull/5043 and patched in release [v0.37.1](https://github.com/tektoncd/pipeline/releases/tag/v0.37.1)

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  Client Version: v1.24.1
  Kustomize Version: v4.5.4
  Server Version: v1.21.12-gke.1500
  ```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

  ```
  Client version: 0.24.0
  Chains version: v0.8.0
  Pipeline version: v0.37.1
  Triggers version: v0.20.1
  Dashboard version: v0.27.0
  ```

cc @Yongxuanzhang @chuangw6 @lbernick @dibyom @afrittoli @abayer

abayer:
For what it's worth, I would suggest we add a commit on the release-v0.37.x branch adding types to `tekton/publish.yaml` and `tekton/release-pipeline.yaml` so that we can actually release the branch with the fix for this cherry picked.

abayer:
[...]

PR #5176, opened 2022-07-20
Add SetDefaults for pipelineSpec in reconciler 

Yongxuanzhang:
# Changes

This commit adds SetDefaults for pipelineSpec in reconciler
after it is fetched from cluster. This is to avoid failing the resources that
are not mutated and add default values.
# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -565,6 +565,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: unit-test-task
+    kind: Task
   timeout: 1h0m0s
 `)
 	// ignore IgnoreUnexported ignore both after and before steps fields
@@ -2785,6 +2786,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: hello-world
+    kind: Task
   timeout: 1h0m0s
 `)
 
@@ -2922,6 +2924,7 @@ spec:
   serviceAccountName: test-sa-0
   taskRef:
     name: hello-world-task
+    kind: Task
   timeout: 1h0m0s
 `),
 		mustParseTaskRunWithObjectMeta(t,
@@ -2932,6 +2935,7 @@ spec:
   serviceAccountName: test-sa-1
   taskRef:
     name: hello-world-task
+    kind: Task
   timeout: 1h0m0s
 `),
 	}
@@ -3530,6 +3534,7 @@ spec:
   - name: foo
   taskRef:
     name: hello-world
+    kind: Task
   timeout: 1h0m0s
 `)
 
@@ -3758,6 +3763,7 @@ spec:
   serviceAccountName: test-sa-0
   taskRef:
     name: b-task
+    kind: Task
   timeout: 1h0m0s
 `)
 	// Check that the expected TaskRun was created
@@ -3949,6 +3955,7 @@ spec:
   serviceAccountName: test-sa-0
   taskRef:
     name: %s
+    kind: Task
   timeout: 1h0m0s
 `, taskName))
 
@@ -4627,6 +4634,7 @@ spec:
   serviceAccountName: test-sa-0
   taskRef:
     name: b-task
+    kind: Task
   timeout: 1h0m0s
 `)
 	// Check that the expected TaskRun was created
@@ -6258,6 +6266,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: finaltask
+    kind: Task
   timeout: 1h0m0s
 `)
 	// Check that the expected TaskRun was created
@@ -6437,6 +6446,7 @@ spec:
   serviceAccountName: test-sa-0
   taskRef:
     name: final-task
+    kind: Task
   timeout: 1h0m0s
 `)
 
@@ -7574,6 +7584,7 @@ spec:
   serviceAccountName: custom-sa
   taskRef:
     name: hello-world
+    kind: Task
   timeout: 1h0m0s
 `)
 
@@ -7691,6 +7702,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: mytask
+    kind: Task
   timeout: 1h0m0s
 `),
 		mustParseTaskRunWithObjectMeta(t,
@@ -7709,6 +7721,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: mytask
+    kind: Task
   timeout: 1h0m0s
 `),
 		mustParseTaskRunWithObjectMeta(t,
@@ -7727,6 +7740,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: mytask
+    kind: Task
   timeout: 1h0m0s
 `),
 		mustParseTaskRunWithObjectMeta(t,
@@ -7745,6 +7759,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: mytask
+    kind: Task
   timeout: 1h0m0s
 `),
 		mustParseTaskRunWithObjectMeta(t,
@@ -7763,6 +7778,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: mytask
+    kind: Task
   timeout: 1h0m0s
 `),
 		mustParseTaskRunWithObjectMeta(t,
@@ -7781,6 +7797,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: mytask
+    kind: Task
   timeout: 1h0m0s
 `),
 		mustParseTaskRunWithObjectMeta(t,
@@ -7799,6 +7816,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: mytask
+    kind: Task
   timeout: 1h0m0s
 `),
 		mustParseTaskRunWithObjectMeta(t,
@@ -7817,6 +7835,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: mytask
+    kind: Task
   timeout: 1h0m0s
 `),
 		mustParseTaskRunWithObjectMeta(t,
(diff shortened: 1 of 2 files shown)


Yongxuanzhang:
https://github.com/tektoncd/pipeline/issues/5175

Yongxuanzhang:
@jerop

pkg/apis/pipeline/v1/task_validation.go
```diff
@@ -274,6 +274,10 @@ func ValidateParameterTypes(ctx context.Context, params []ParamSpec) (errs *apis
 
 // ValidateType checks that the type of a ParamSpec is allowed and its default value matches that type
 func (p ParamSpec) ValidateType() *apis.FieldError {
+	// For resources not call SetDefaults(), empty type should be valid.
```
afrittoli, line 277: NIT - suggested comment

> A resource may have been created before type was available in `paramSpec` and thus have no type set

Yongxuanzhang: sure!

dibyom: couldn't we call SetDefaults() before calling validation instead of just skipping validation?

pkg/apis/pipeline/v1/task_validation.go
```diff
@@ -274,6 +274,10 @@ func ValidateParameterTypes(ctx context.Context, params []ParamSpec) (errs *apis
...
+	if p.Type == "" {
```
ywluogg, line 278: Can you add a unit test in pipeline reconciler that the param type is null, to make sure this wouldn't introduce any new bugs? Thanks

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -436,6 +436,48 @@ func TestTaskSpecValidate(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ts := &v1beta1.TaskSpec{
+				Params: tt.fields.Params,
+				Steps:  tt.fields.Steps,
+			}
+			if err := ts.Validate(context.Background()); err != nil {
+				t.Errorf("TaskSpec.Validate() = %v", err)
+			}
+		})
```
chuangw6, line 477: In the test, can we compare the returned v.s. expected error like this [example](https://github.com/tektoncd/pipeline/blob/a4f9216e8957e52ce9bdf4af82cfd82148659539/pkg/apis/pipeline/v1beta1/task_validation_test.go#L1304-L1311) rather than just check if err is nil?

pkg/apis/pipeline/v1/task_validation.go
Yongxuanzhang: > couldn't we call SetDefaults() before calling validation instead of just skipping validation?

yeah I would prefer this way!

dibyom: (or at least in this function, we should assume no type == defaultType)

abayer:
It's worth mentioning that when I deleted a problem `Pipeline` and recreated it, with any of `kubectl replace`, `kubectl apply`, or `kubectl create`, params without types specified still had no types specified. This is with Pipeline v0.37.2. So if the current behavior, before this PR, is supposed to be defaulting newly created `Pipeline`s and `Task`s to have default types on their params, well, that's not working either.

chuangw6:
> It's worth mentioning that when I deleted a problem `Pipeline` and recreated it, with any of `kubectl replace`, `kubectl apply`, or `kubectl create`, params without types specified still had no types specified. This is with Pipeline v0.37.2. So if the current behavior, before this PR, is supposed to be defaulting newly created `Pipeline`s and `Task`s to have default types on their params, well, that's not working either.

Hi @abayer,
I tried the following yaml (the type of param `arg` and `mystring` is not specified) with v0.37.2 and it works fine. Can you also try that? If this example works, let's find what are the missing/different part in the problem `pipeline`.

```yaml
kind: Task
metadata:
  name: test-task
spec:
  params:
    - name: arg
  steps:
    - name: echo
      image: bash:latest
      args: ["echo", "$(params.arg)"]
---
kind: Pipeline
metadata:
  name: test-pipeline
spec:
  params:
    - name: mystring
  tasks:
    - name: echo-test
      params:
        - name: arg
          value: $(params.mystring)
      taskRef:
        name: test-task
---
kind: PipelineRun
metadata:
  generateName: test-pipeline-run
spec:
  params:
    - name: mystring
      value: abc.com
  pipelineRef:
    name: test-pipeline

```

abayer:
Ok, turns out the problem was incredibly stupid and specific - in the `default` namespace on the dogfooding cluster, we had `webhooks.knative.dev/exclude: "true"` in the labels, so it was skipping defaulting entirely. 🤦

ywluogg:
> Ok, turns out the problem was incredibly stupid and specific - in the `default` namespace on the dogfooding cluster, we had `webhooks.knative.dev/exclude: "true"` in the labels, so it was skipping defaulting entirely. 🤦

Haha lol! But this actually helped us catching some bugs in webhook and strengthen our checks in reconciler! Effort not wasted!

ywluogg:
> Ok, turns out the problem was incredibly stupid and specific - in the `default` namespace on the dogfooding cluster, we had `webhooks.knative.dev/exclude: "true"` in the labels, so it was skipping defaulting entirely. 🤦

🤦🤦🤦

chuangw6:
> Ok, turns out the problem was incredibly stupid and specific - in the `default` namespace on the dogfooding cluster, we had `webhooks.knative.dev/exclude: "true"` in the labels, so it was skipping defaulting entirely. 🤦

Thanks for finding out this. I am curious if this field `webhooks.knative.dev/exclude` was set from false to true just recently OR it has been `true` all the time?

Yongxuanzhang:
> It's worth mentioning that when I deleted a problem `Pipeline` and recreated it, with any of `kubectl replace`, `kubectl apply`, or `kubectl create`, params without types specified still had no types specified. This is with Pipeline v0.37.2. So if the current behavior, before this PR, is supposed to be defaulting newly created `Pipeline`s and `Task`s to have default types on their params, well, that's not working either.



> Ok, turns out the problem was incredibly stupid and specific - in the `default` namespace on the dogfooding cluster, we had `webhooks.knative.dev/exclude: "true"` in the labels, so it was skipping defaulting entirely. 🤦

Is it something very easy to fix? We're not so familiar with the dogfooding cluster configuration. Is it public and can be viewed somewhere?

abayer:
Someone set it by hand to be able to remove resources that couldn't be deleted due to removed fields - not sure exactly when it was done, but more than a week or so ago. It's fixed now, so we're all good on dogfooding regarding defaulting.

chuangw6:
> Someone set it by hand to be able to remove resources that couldn't be deleted due to removed fields - not sure exactly when it was done, but more than a week or so ago. It's fixed now, so we're all good on dogfooding regarding defaulting.

Got it. Good lesson to learn. Didn't know there is a field `webhooks.knative.dev/exclude` that can disable the wekbhook in the cluster. Thanks @abayer

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -348,6 +348,10 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
 	}
 
 	pipelineMeta, pipelineSpec, err := rprp.GetPipelineData(ctx, pr, getPipelineFunc)
+	if pipelineSpec != nil {
```
ywluogg, line 351: There should be a test case testing this and should be corresponds to this change?

Yongxuanzhang: yes I haven't finished all tests yet! Once done I will unhold this pr

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -565,6 +565,7 @@ spec:
   serviceAccountName: test-sa
   taskRef:
     name: unit-test-task
+    kind: Task
```
lbernick, line 568: What's the reason to add `kind: Task` to all of these taskRefs?

lbernick (commented):
I'm a bit confused about what this PR is doing given the original PR message and the conversation that's happened. It looks like this sets defaults in the reconciler to ensure that defaults are applied to resources that were created before these code changes were applied to the server. Is this still an issue with params without types, disregarding the annotations that were mistakenly kept around on our dogfooding cluster?

Yongxuanzhang:
> I'm a bit confused about what this PR is doing given the original PR message and the conversation that's happened. It looks like this sets defaults in the reconciler to ensure that defaults are applied to resources that were created before these code changes were applied to the server. 

Yes,  that's the current goal

> Is this still an issue with params without types, disregarding the annotations that were mistakenly kept around on our dogfooding cluster?

There are some other discussions in this issue https://github.com/tektoncd/pipeline/issues/5175. And for taskrun when we get the taskspec, it will call `taskSpec.SetDefaults(ctx)`, 
https://github.com/tektoncd/pipeline/blob/10a5ae69b35bd7cc4542234f395c30ca214ca5a6/pkg/reconciler/taskrun/resources/taskspec.go#L48-L54
but for `pipelinerun` we don't have it. It may make sense to add for pipelinerun as well
https://github.com/tektoncd/pipeline/blob/10a5ae69b35bd7cc4542234f395c30ca214ca5a6/pkg/reconciler/pipelinerun/pipelinespec/pipelinespec.go#L40-L47

pkg/reconciler/pipelinerun/pipelinerun_test.go
Yongxuanzhang: setdefaults will add kind for task as well, if not adding them the current tests will fail

lbernick (approved):

Issue #5174, opened 2022-07-19
Release Pipeline failing


dibyom:
# Expected Behavior
Release pipeline works

# Actual Behavior

Pipeline fails at the run-ko step: 

```
2022/07/19 18:39:13 Unexpected error running "go build": exit status 2
# github.com/tektoncd/pipeline/cmd/entrypoint
cmd/entrypoint/main.go:142:4: unknown field 'stdoutPath' in struct literal of type realRunner
cmd/entrypoint/main.go:143:4: unknown field 'stderrPath' in struct literal of type realRunner
```
# Steps to Reproduce the Problem

1.
2.
3.

# Additional Info

Full logs: https://gist.github.com/dibyom/81be5cb43e342cbcb77b9bea0926e0a3

Run: https://dashboard.dogfooding.tekton.dev/#/namespaces/default/pipelineruns/pipeline-release-run-z8vht?pipelineTask=publish-images

- Tekton Pipeline version: v0.37.1

Go version: 1.17.8

dibyom:
I tried the failing command `ko resolve --platform=linux/amd64,linux/arm,linux/arm64,linux/s390x,linux/ppc64le,windows/amd64 --preserve-import-paths -t v0.38.0 -R -f config/` locally on go version ` go1.18.3 darwin/amd64` and it worked

dibyom:
cc @jerop

jerop:
another release pipeline failure reported in https://github.com/tektoncd/pipeline/issues/5175

abayer:
Ok, via `docker run -it gcr.io/tekton-releases/dogfooding/ko@sha256:aafcfa5c22281f9e882664c3e5f234a2069a0050674c0c8d9908ccae6f0e8457` and manually replicating the commands leading up to the `ko resolve` and then running `ko resolve ...` with `--push=false`, I've reproduced the error, and narrowed it down to having `windows/amd64` in the platforms. And more specifically, it's only when `.ko.yaml` is updated to use `gcr.io/tekton-releases/github.com/tektoncd/pipeline/combined-base-image@sha256:cabda55b027488777f06576dcb615779640ea9e768784b2720215b91e14cc9e0` as the base image for `cmd/entrypoint`, which is done by https://github.com/tektoncd/pipeline/blob/94055d92c120a6010f3d61a821e45dea4f893a74/tekton/publish.yaml#L97-L113

That said, it doesn't even try to build Windows images without that base image override. 

[...]

PR #5178, opened 2022-07-20
Include Windows image builds in CI build tests

### LINKED ISSUE #5177: CI build tests should use combined build image and build Windows images
As revealed by #5174, we're not actually build Windows images in `pull-tekton-pipeline-build-tests`, which resulted in code that can't compile on Windows slipping through in a PR. We should be setting Windows-supporting combined build images in `ko_resolve` in `test/presubmit-tests.sh` in the same sort of way we are in https://github.com/tektoncd/pipeline/blob/94055d92c120a6010f3d61a821e45dea4f893a74/tekton/publish.yaml#L96-L113

We can probably just use `gcr.io/tekton-releases/github.com/tektoncd/pipeline/combined-base-image:latest` for that combined base image, rather than generating a new one each time.

/kind bug

abayer:
# Changes

closes #5177

Note that this PR will fail `pull-tekton-pipeline-build-tests` (assuming the change does what it's supposed to!) until #5174 is dealt with.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/presubmit-tests.sh
+++ b/test/presubmit-tests.sh
@@ -59,6 +59,19 @@ function check_yaml_lint() {
 
 function ko_resolve() {
   header "Running `ko resolve`"
+
+  cat <<EOF > .ko.yaml
+    defaultBaseImage: ghcr.io/distroless/static
+    baseImageOverrides:
+      # Use the combined base image for images that should include Windows support.
+      # NOTE: Make sure this list of images to use the combined base image is in sync with what's in tekton/publish.yaml's 'create-ko-yaml' Task.
+      github.com/tektoncd/pipeline/cmd/entrypoint: gcr.io/tekton-releases/github.com/tektoncd/pipeline/combined-base-image:latest
+      github.com/tektoncd/pipeline/cmd/nop: gcr.io/tekton-releases/github.com/tektoncd/pipeline/combined-base-image:latest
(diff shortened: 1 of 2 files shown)


test/presubmit-tests.sh
```diff
@@ -59,6 +59,20 @@ function check_yaml_lint() {
 
 function ko_resolve() {
   header "Running `ko resolve`"
+
+  cat <<EOF > .ko.yaml
+    # This matches the value configured in .ko.yaml
```
imjasonh, line 64: Do we have any guardrails in place to help keep these in sync? I can definitely imagine someone (me, probably) updating publish.yaml and not making the same update here.

This comment is a little misleading, since in this context it's actually matching values from publish.yaml, which is trying to match values in .ko.yaml

abayer: Yeah, I just copy-pasted directly. And no, no guardrails - this is literally just the quick-and-dirty "lemme make sure this fails like I expect" approach for the moment.

abayer: Oh, and actually, that comment is accurate - we're only overriding `baseImageOverrides`. But it's a confusing comment, so I'll remove it and the other down a few lines.

abayer: Comment removed, added new comments here and in `tekton/publish.yaml` to remind people to keep them in sync.

vdemeester (approved):

afrittoli (commented):
Cool :cool: thank you!
/lgtm

PR #5198, opened 2022-07-22
Fail taskrun when results validation fails

Yongxuanzhang:
# Changes

This commit adds the ReasonFailedValidation into taskrun status and fail
the taskrun when results validation fails. Before this commit users can
only see error messages in logs but the taskrun are completed like no
errors happen.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -132,6 +132,21 @@ var (
 			Steps: []v1beta1.Step{simpleStep},
 		},
 	}
+	resultsTask = &v1beta1.Task{
+		ObjectMeta: objectMeta("test-results-task", "foo"),
+		Spec: v1beta1.TaskSpec{
+			Steps: []v1beta1.Step{simpleStep},
+			Results: []v1beta1.TaskResult{{
+				Name: "aResult",
+				Type: v1beta1.ResultsTypeArray,
+			}, {
+				Name:       "objectResult",
+				Type:       v1beta1.ResultsTypeObject,
+				Properties: map[string]v1beta1.PropertySpec{"url": {Type: "string"}, "commit": {Type: "string"}},
+			},
+			},
+		},
+	}
 
 	clustertask = &v1beta1.ClusterTask{
 		ObjectMeta: metav1.ObjectMeta{Name: "test-cluster-task"},
@@ -430,6 +445,9 @@ func runVolume(i int) corev1.Volume {
 
 func createServiceAccount(t *testing.T, assets test.Assets, name string, namespace string) {
 	t.Helper()
+	if name == "" {
+		name = "default"
+	}
 	if _, err := assets.Clients.Kube.CoreV1().ServiceAccounts(namespace).Create(assets.Ctx, &corev1.ServiceAccount{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
@@ -1184,9 +1202,6 @@ spec:
 			c := testAssets.Controller
 			clients := testAssets.Clients
 			saName := tc.taskRun.Spec.ServiceAccountName
-			if saName == "" {
-				saName = "default"
-			}
 			createServiceAccount(t, testAssets, saName, tc.taskRun.Namespace)
 
 			if err := c.Reconciler.Reconcile(testAssets.Ctx, getRunName(tc.taskRun)); err == nil {
@@ -1342,9 +1357,6 @@ spec:
 			c := testAssets.Controller
 			clients := testAssets.Clients
 			saName := tc.taskRun.Spec.ServiceAccountName
-			if saName == "" {
-				saName = "default"
-			}
 			createServiceAccount(t, testAssets, saName, tc.taskRun.Namespace)
 
 			if err := c.Reconciler.Reconcile(testAssets.Ctx, getRunName(tc.taskRun)); err == nil {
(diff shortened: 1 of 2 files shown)


Yongxuanzhang:
@ywluogg

ywluogg (commented):
/lgtm

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4657,3 +4672,96 @@ func objectMeta(name, ns string) metav1.ObjectMeta {
...
+			if d := cmp.Diff(err.Error(), tc.expectedError.Error()); d != "" {
+				t.Errorf("Expected: %v, but Got: %v", tc.expectedError, err)
+			}
+			tr, err := clients.Pipeline.TektonV1beta1().TaskRuns(tc.taskRun.Namespace).Get(testAssets.Ctx, tc.taskRun.Name, metav1.GetOptions{})
+			if err != nil {
+				t.Fatalf("getting updated taskrun: %v", err)
+			}
+			condition := tr.Status.GetCondition(apis.ConditionSucceeded)
+			if condition.Type != apis.ConditionSucceeded || condition.Status != corev1.ConditionFalse || condition.Reason != tc.wantFailedReason {
+				t.Errorf("Expected TaskRun to \"%s\" but it did not. Final conditions were:\n%#v", tc.wantFailedReason, tr.Status.Conditions)
```
lbernick, line 4762: suggested error message: "Expected TaskRun to fail with reason %s but it did not...<etc>"

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4657,3 +4672,96 @@ func objectMeta(name, ns string) metav1.ObjectMeta {
...
+		t.Run(tc.name, func(t *testing.T) {
+			testAssets, cancel := getTaskRunController(t, d)
+			defer cancel()
+			c := testAssets.Controller
+			clients := testAssets.Clients
+			saName := tc.taskRun.Spec.ServiceAccountName
+			if saName == "" {
+				saName = "default"
+			}
+			createServiceAccount(t, testAssets, saName, tc.taskRun.Namespace)
```
lbernick, line 4750: can any of this be moved into a testing helper or deleted?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4657,3 +4672,96 @@ func objectMeta(name, ns string) metav1.ObjectMeta {
...
+	for _, tc := range []struct {
+		name             string
+		taskRun          *v1beta1.TaskRun
+		wantFailedReason string
+		expectedError    error
+	}{{
+		name:             "taskrun results type mismatched",
+		taskRun:          taskRunResultsTypeMismatched,
+		wantFailedReason: podconvert.ReasonFailedValidation,
+		expectedError:    fmt.Errorf("1 error occurred:\n\t* missmatched Types for these results, map[aResult:[array]]\n\n"),
```
lbernick, line 4734: This is a confusing error message because it's not clear which result (taskrun result or result declared in task spec) is an array and which is not, and what the type of the result that is not an array is. Can it be improved?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4657,3 +4672,96 @@ func objectMeta(name, ns string) metav1.ObjectMeta {
...
+		TaskRuns: taskruns,
+		Tasks:    []*v1beta1.Task{resultsTask},
+		ConfigMaps: []*corev1.ConfigMap{{
+			ObjectMeta: metav1.ObjectMeta{Namespace: system.Namespace(), Name: config.GetFeatureFlagsConfigName()},
+			Data: map[string]string{
+				"enable-api-fields": config.AlphaAPIFields,
+			},
+		}},
+	}
+	for _, tc := range []struct {
```
lbernick, line 4725: could you add a test case for success if there isn't one already in taskrun_test.go?

pkg/reconciler/taskrun/taskrun_test.go
Yongxuanzhang: Sure! Thanks!

pkg/reconciler/taskrun/taskrun_test.go
Yongxuanzhang: sure! just added one test for success

pkg/reconciler/taskrun/taskrun_test.go
Yongxuanzhang: yes, I move saName default into the createServiceAccount and remove other lines

pkg/reconciler/taskrun/taskrun_test.go
Yongxuanzhang: The current code logic will only collect the wrong results/params names and log them, and same for other parts. You're 100% correct that this can be improved! I will think about it and maybe address it in a new PR? If that is ok I will open an issue to track

https://github.com/tektoncd/pipeline/blob/8a7b0cfa755038f4cfdcd88c314a72a90bcab1a2/pkg/reconciler/pipelinerun/resources/validate_params.go#L82-L90

pkg/reconciler/taskrun/taskrun_test.go
lbernick: nice thanks!

pkg/reconciler/taskrun/taskrun_test.go
lbernick: It looks like you might have accidentally just copied the failure test case?

pkg/reconciler/taskrun/taskrun_test.go
lbernick: Sure, you can definitely change in a separate PR-- it's a small enough change I'll leave it up to you whether to create an issue or just open a follow-up PR

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4657,3 +4669,169 @@ func objectMeta(name, ns string) metav1.ObjectMeta {
...
+		name    string
+		taskRun *v1beta1.TaskRun
+	}{{
+		name:    "taskrun results type mismatched",
+		taskRun: taskRunResultsTypeMismatched,
+	}, {
+		name:    "taskrun results object miss key",
+		taskRun: taskRunResultsObjectMissKey,
```
lbernick, line 4726: could you please rename these test cases and variables so it's clear what's being tested?

Yongxuanzhang: sure!

pkg/reconciler/taskrun/taskrun_test.go
Yongxuanzhang: oh yes, sorry about that. >_< I only updated the yaml name. Should be correct now!

lbernick (approved):

Issue #4710, opened 2022-03-28
Promote Remote Resolution to beta


ghost:
### Feature request

Remote resolution is currently behind the alpha feature gate in Tekton Pipelines. This issue is intended to track the work required to promote resolution into beta, removing the feature gate.

### Use case

Platforms and users would like to access Tekton resources in locations other than OCI registries or the cluster their taskruns and pipelineruns execute in.

### Criteria for Promotion

- [ ] Poll Pipeline maintainers for their preferred direction with resolution: integrate with pipelines + add out-of-box resolvers, keep entirely separate, go a different direction entirely.
- [ ] Once ^ is done, document the rest of the criteria for promotion here.

sbwsg:
This is a Work-in-Progress comment describing some of the changes that would be expected if we move remote resolution to a feature of pipelines rather than an external project.

Currently Tekton Resolution maintains a "core" reconciler whose responsibility is to monitor _all_ `ResolutionRequests` and mark them Succeeded after a resolver successfully patches in the content of a pipeline/task. This core reconciler also performs book-keeping like initializing the ResolutionRequest's status and enforcing global timeouts so that a resolution request doesn't run forever.

Here is approximately the order of operations today for a `ResolutionRequest`:

```mermaid
sequenceDiagram
    Tekton Pipelines->>ResolutionRequest: CREATE resolution request for git resource
    ResolutionRequest-->>Tekton Resolution: WATCH triggers reconcile of Resolution's "core" reconciler
    Tekton Resolution->>ResolutionRequest: UPDATE with initial status
    ResolutionRequest-->>GitResolver: WATCH triggers reconcile of GitResolver
    GitResolver->>Git Repo: Clone & get file contents
    GitResolver->>ResolutionRequest: PATCH with contents of file from git
    ResolutionRequest-->>Tekton Resolution: WATCH triggers reconcile of Resolution's "core" reconciler
    Tekton Resolution->>ResolutionRequest: UPDATE with Succeeded status
    ResolutionRequest-->>Tekton Pipelines: WATCH triggers reconcile of pipeline w/ yaml from file
```

In a world where Tekton Pipelines merges remote resolution directly into its codebase, this flow would be simplified quite a bit:

```mermaid
  sequenceDiagram
    Tekton Pipelines->>ResolutionRequest: CREATE resolution request for git resource with initial status
    ResolutionRequest-->>GitResolver: WATCH triggers reconcile of GitResolver
[...]

PR #5200, opened 2022-07-22
Add types and client for Resolution

abayer:
# Changes

Part of #4710

In order to get remote resolution functionality to beta, we're moving it from a separate repository/release to be part of Pipeline. This is the first PR in a sequence moving the code from the Resolution repository into Pipeline.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
action required: Remote resolver `resource` field has changed to `params`.
```


abayer:
/assign @vdemeester @dibyom 
cc @chitrangpatel

abayer:
/hold

Putting this on hold until we've got v0.38.0 cut, 'cos I'd really like to have all of the relevant commits in one release together.

pkg/apis/resolution/v1alpha1/resolution_request_types.go
```diff
@@ -0,0 +1,86 @@
...
+
+// ResolutionRequestSpec are all the fields in the spec of the
+// ResolutionRequest CRD.
+type ResolutionRequestSpec struct {
+	// Parameters are the runtime attributes passed to
+	// the resolver to help it figure out how to resolve the
+	// resource being requested. For example: repo URL, commit SHA,
+	// path to file, the kind of authentication to leverage, etc.
+	// +optional
+	Parameters map[string]string `json:"params,omitempty"`
```
pritidesai, line 64: I was sharing the remote resource spec with a team and came across two different syntaxes. It seems we are implementing it using `params`:

```
taskRef:
    resolver: git
    resource:
    - name: url
      value: https://github.com/tekton/catalog.git
    - name: path
      value: /task/golang-fuzz/0.1/golang-fuzz.yaml
    - name: branch
      value: main
```

The [proposal](https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md#proposal) in the Remote Resolution TEP has listed it as:

```
taskRef:
  resolver: git
  resource:
    repository_url: https://github.com/tektoncd/catalog.git
    branch: main
    path: /task/golang-fuzz/0.1/golang-fuzz.yaml
```

Do we need to update the proposal to reflect `params`? or is it documented somewhere else?

abayer: Good question! I didn't do the initial implementation, but while I like the `repository_url` etc syntax better, the `- name: url` syntax is simpler to implement, since we don't need to use a `runtime.RawExtension` here for arbitrary YAML. 

So I'd say the proposal should be updated, yeah.

vdemeester: @abayer to play devil's advocate, we should aim for the API to be the easiest to use (for end-user) than the easiest to implement 😝

abayer: Fair point. =) I spent some time this weekend thinking about this - in practice, it probably is easier to go with a simple `map[string]string` than I'd been thinking, given that what we've got now is already just a slice of string name/value pairs anyway. Then, if we hit a situation down the road where we have a need for more complex resolver parameters, we could change that to something like `unstructured.Unstructured` without breaking API compatibility.

abayer: ...though it is worth mentioning that changing the syntax will break compatibility with the existing syntax, but that's alpha syntax and probably not at all widely used right now anyway.

vdemeester: > ...though it is worth mentioning that changing the syntax will break compatibility with the existing syntax, but that's alpha syntax and probably not at all widely used right now anyway.

yeah that's the only reason that was "against" changing, but this is behind the alpha feature-flag for that sole reason, so if we properly communicate, I think that is fine.

abayer: 👍 Lemme do that!

abayer:
@pritidesai @vdemeester I've changed the `resource:` syntax to be 
```
resource:
  foo: bar
```
instead of
```
resource:
  - name: foo
     value: bar
```

vdemeester (approved):

pkg/apis/pipeline/v1beta1/resolver_types.go
```diff
@@ -33,16 +33,5 @@ type ResolverRef struct {
...
-
-// ResolverParam is a single parameter passed to a resolver.
-type ResolverParam struct {
-	// Name is the name of the parameter that will be passed to the
-	// resolver.
-	Name string `json:"name"`
-	// Value is the string value of the parameter that will be
-	// passed to the resolver.
-	Value string `json:"value"`
+	Resource map[string]string `json:"resource,omitempty"`
```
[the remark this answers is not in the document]
abayer: Very good question! I am not a fan of `resource` here - it's not really clear what it means in the first place. I feel like just `params` would work, though since `params` is used in a bunch of places with different syntax than here, I'm not sure if that's ideal...

abayer:
/hold

Putting on hold again until we nail down whether we keep the `resource` field name or not.

pkg/apis/pipeline/v1beta1/resolver_types.go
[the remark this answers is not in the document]
vdemeester: In the TEP, we used `resource` ? I'd rather go with `params` or `values` than `resource` indeed.

abayer:
I've added a commit changing `resource:` to `params:` - if it seems good to everyone, I'll squash it, and open a corresponding PR to update the TEP.

lbernick:
OK I'm sorry I don't really have a better suggestion here, but `params` might be confusing since the syntax was changed to be non-params-like

abayer:
@lbernick Yeah, I don't like `params` either, but can't think of anything better. `values` doesn't feel right...maybe `args`? Though it's not the same as a container/step's `args`...

abayer:
You know what? I'm just going to go with `paramsMap` for now - yeah, it's still got `params` in it, but I feel like being a different string, with `map` explicitly in the string, is good enough.

lbernick (changes requested):
Sorry to push back on this but I think we should discuss naming at API WG if reviewers for this PR are able to attend-- I'll try to brainstorm some ideas before then but I think getting naming correct is important

abayer:
No worries! =) I'm not in love with any option at this point, though I do prefer `paramsMap` to either `params` or `resource`, hence me updating the PR.

abayer:
Another option could be `resolverParams` - which stutters a bit being under the `resolver` field. Again, I'm open to literally any suggestion.

abayer:
So after discussion with @lbernick and her discovery of https://github.com/tektoncd/pipeline/pull/4502#discussion_r803199842, I've come to the conclusion that we should stick with the subobject syntax, with the field renamed to `params`. As discussed in that comment, this is more consistent with Kubernetes in general, and it also will make it much easier to add support for array/object parameters in the future if we so desire. That will entail a change to `ResolutionRequestSpec.Parameters` to not be a `map[string]string` as well, and an update to the TEP.

cc @pritidesai @vdemeester

pkg/apis/resolution/v1alpha1/resolution_request_validation.go
```diff
@@ -0,0 +1,51 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"context"
+
+	"github.com/tektoncd/resolution/pkg/common"
```
dibyom, line 22: not for this PR but at some point we should get rid of the "common" package and just have the helper function in better named packages

dibyom (approved):

pkg/apis/resolution/v1alpha1/resolution_request_validation.go
abayer: Good idea - I'll do that in the PR where I move that package in.

lbernick (approved):
Thanks Andrew! LGTM but I'll let Vincent or Priti give the final lgtm just to make sure they are also on board with this syntax.

Please don't forget the action required release note!

vdemeester:
> So after discussion with @lbernick and her discovery of [#4502 (comment)](https://github.com/tektoncd/pipeline/pull/4502#discussion_r803199842), I've come to the conclusion that we should stick with the subobject syntax, with the field renamed to `params`. As discussed in that comment, this is more consistent with Kubernetes in general, and it also will make it much easier to add support for array/object parameters in the future if we so desire. That will entail a change to `ResolutionRequestSpec.Parameters` to not be a `map[string]string` as well, and an update to the TEP.
> 
> cc @pritidesai @vdemeester

So, I agree that, if we stick with `params` field for this, we should be consistent with the shape of `params` in the other parts of the API. I disagree that we should be necessarily consistent with Kubernetes on this. We should *aim* for the best UX for our users not because Kubernetes does it — and we are trying to not be too much "tied up" with Kubernetes as well. If we aim for end-users to author `Pipeline`, `Task`, … we should aim for making the API tailored for them, and not necessarily follow what *idiomatic* Kubernetes API does/looks like.

The `params` syntax is quite verbose. If we use the `bundle` for comparison :

```yaml
# from
taskRef:
  bundle: quay.io/vdemeest/foo:bar
  name: baz
# to
taskRef:
  resolver: bundle
  params:
  - name: image
    value: quay.io/vdemeest/foo:bar
  - name: name
    value: baz
  - name: type
    value: Task
```

From a user perspective, an "author", I have the feeling that the following feels better.

```yaml
taskRef:
  resolver: bundle
  params:
    image: quay.io/vdemeest/foo:bar
    name: baz
    type: Task
```

That *might* be one of the reason the [TEP](https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md) uses `resource` instead, to not have to be consistent with `params`.

On "additionnal" point here, the resulting `ResolutionRequest` type will have a different `params` syntax as the one specified in `resolver`. Do we want to change this as well ?

```yaml
apiVersion: resolution.tekton.dev/v1alpha1
kind: ResolutionRequest
# […]
spec:
  params:
    serviceAccount: default
    bundle: gcr.io/tekton-releases/catalog/upstream/golang-build:0.1
    name: golang-build
    kind: task
EOF
```

In a gist, I am not a huge fan of using `params` and the `task.params` syntax, and I would rather stick to the proposed TEP syntax 😅 ; mainly because, from a user/author perspective, it's easier to write 😇. *Sorry for the back and forth 😓*

abayer:
> On "additionnal" point here, the resulting `ResolutionRequest` type will have a different `params` syntax as the one specified in `resolver`. Do we want to change this as well ?

Eventually, yes, but I think that will involve moving `ArrayOrString` out of `pkg/apis/pipeline` so we can use it in `pkg/apis/resolution` as well. Since that will be a decent amount of work, and `ResolutionRequest`s aren't user-authored anyway, I thought that work could wait til later.

vdemeester:
> it's "use the same admittedly subpar/overly verbose syntax for `resolver.[params|resource]` that we do elsewhere or use a different syntax unique to this one field to accomplish the same end result".

I agree with that 💯 % 😝. But *if* we didn't name it `params`, but something else, maybe `spec` 😈. Then we would have something completely "new" and "unique", and more closely to the `spec` you tend to see in Kubernetes. For example `volumeClaimTemplate`, etc.. I agree it's a very thin line between parameters and the "spec" of the resolver though 😝.

abayer:
> I agree with that 💯 % 😝. But _if_ we didn't name it `params`, but something else, maybe `spec` 😈. Then we would have something completely "new" and "unique", and more closely to the `spec` you tend to see in Kubernetes. For example `volumeClaimTemplate`, etc.. I agree it's a very thin line between parameters and the "spec" of the resolver though 😝.

Right - so I think that's a legitimate argument for `ResolutionRequest`'s `Params` field, since that's in a different CRD. But within the context of a `pipelineRef` or `taskRef`, I think it would be a mistake to have a different syntax for `.resolver.[params|resource]` than we do elsewhere in the same `Pipeline`/`PipelineRun`/`TaskRun` CRD.

abayer:
In general, I'd argue that we should only have one syntax in Pipeline for arbitrarily-named key/value pairs, and since we already have one (`params` and `results` etc using `- name: ...`), we should keep using that one syntax in all cases.

pkg/apis/resolution/v1alpha1/resolution_request_types.go
jerop: the reason why it was implemented as such was because of K8s convention: "Lists of named subobjects preferred over maps" - as discussed in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps and https://github.com/kubernetes/kubernetes/issues/2004

and we listed in https://github.com/tektoncd/community/blob/main/design-principles.md#api-conventions that Tekton API should comply with the K8s API conventions

jerop: haha just saw the comments below

abayer:
FYI, I'm reworking `[PipelineRef|TaskRef].ResolverRef.Params` and `ResolutionRequestSpec.Params` to be `[]v1beta1.Param` now, since it feels a little silly to not do so, with validation, replacements, etc supported.

abayer:
For what it's worth, in the process of switching over to using `[]v1beta1.Param` everywhere, I realized another reason for using the `Param` syntax (and therefore using `v1beta1.Param` behind the scenes - it makes param/result substitution simpler, and that's something I think users will end up wanting. By using `v1beta1.Param`, we just have to add the appropriate calls to the standard replacement functions for `v1beta1.Param` without having to write new substitution logic specifically for a new type.

abayer:
I've just switched `ResolutionRequest`'s spec to keep the map for the moment, so that we can have the CRD be the same as it is in Resolution currently, and we'll follow that up with a deprecation and switch to `.params` syntax there later. Note that the `pipelineRef` and `taskRef` syntax is still changing from `resource` as a map of strings to `params` as a slice of `v1beta1.Param`.

abayer:
So this still needs a squash, but I've avoided doing so until there's buy-in from @vdemeester. =)

I deliberately decided to _not_ change the `ResolutionRequest` syntax for now, because there's already the v1alpha1 version in Resolution. Once this chain lands, I'll deprecate v1alpha1 and replace it with v1beta1 using `[]Param` rather than `map[string]string`, but I didn't want to break compatibility for the existing CRD. We still need to have the v1alpha1 definition here so that we can add conversion to it, so that's why it's a bit messy. =)

pkg/apis/resolution/v1alpha1/resolution_request_defaults.go
```diff
@@ -0,0 +1,33 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import "context"
+
+// ManagedByLabelKey is the label key used to mark what is managing this resource
+const ManagedByLabelKey = "app.kubernetes.io/managed-by"
```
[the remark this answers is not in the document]
abayer: We have - I tried to avoid having `pkg/apis/resolution` depend on anything from `pkg/apis/pipeline`, which may have been overkill.

pkg/apis/resolution/v1alpha1/resolution_request_types.go
abayer: ...and yeah, I reverted back to the old form for `ResolutionRequest.Spec` so as not to break compatibility. See https://github.com/tektoncd/pipeline/pull/5200#issuecomment-1219463898 =)

pkg/apis/pipeline/v1beta1/pipeline_types_test.go
```diff
@@ -337,9 +337,16 @@ func TestPipelineTask_ValidateRegularTask_Failure(t *testing.T) {
 		},
 		expectedError: *apis.ErrDisallowedFields("taskref.resolver"),
 	}, {
-		name: "pipeline task - use of resource without the feature flag set",
+		name: "pipeline task - use of resolver params without the feature flag set",
 		task: PipelineTask{
-			TaskRef: &TaskRef{Name: "boo", ResolverRef: ResolverRef{Resource: []ResolverParam{{}}}},
+			TaskRef: &TaskRef{Name: "boo", ResolverRef: ResolverRef{Params: []Param{{
+				Name: "bar",
+				Value: ArrayOrString{
```
jerop, line 344: `ArrayOrString` was renamed to `ParamValue` in https://github.com/tektoncd/pipeline/pull/5304 -- we need to rename it in this PR as well

jerop (commented):
`ArrayOrString` was renamed to `ParamValue` in https://github.com/tektoncd/pipeline/pull/5304 -- we need to rename it in this PR as well

abayer:
> `ArrayOrString` was renamed to `ParamValue` in #5304 -- we need to rename it in this PR as well

Done!

Issue #4986, opened 2022-06-15
V1 version of Pipeline


lbernick:
This issue tracks the release of a v1 version of the Pipeline CRD. More details in #4983.

PR #5219, opened 2022-07-26
Add V1 Pipeline Golang structs

JeromeJu:
# Changes

This commit adds structs and validation for v1 Pipeline, including when expressions,
resultRef. This is copied from v1beta1 with the following exceptions:

- omitted PipelineResources
- omitted Bundle and ClusterTask fields
- change PipelineTask `whenExpressions` Go field name from to `When`
due to api naming violation

Part of #4986
/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


pkg/apis/pipeline/v1/resultref.go
```diff
@@ -0,0 +1,213 @@
...
+	// If a string result name contains a dot, brackets should be used to differentiate it from an object result.
+	// https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#collisions-with-builtin-variable-replacement
+	objectResultExpressionFormat = "tasks.<taskName>.results.<objectResultName>.<individualAttribute>"
+	// ResultTaskPart Constant used to define the "tasks" part of a pipeline result reference
+	ResultTaskPart = "tasks"
+	// ResultResultPart Constant used to define the "results" part of a pipeline result reference
+	ResultResultPart = "results"
+	// TODO(#2462) use one regex across all substitutions
+	// variableSubstitutionFormat matches format like $result.resultname, $result.resultname[int] and $result.resultname[*]
+	variableSubstitutionFormat = `\$\([_a-zA-Z0-9.-]+(\.[_a-zA-Z0-9.-]+)*(\[([0-9]+|\*)\])?\)`
```
JeromeJu, line 46: Noticed that exactVariableSubstitutionRegex has been moved to pkg/apis/pipeline/v1/param_types.go.
May I ask if we have any plan to move it back to resultref.go?

lbernick: I think it's ok to keep in param_types.go

lbernick:
We do need to keep our alpha fields around (including matrix); they should just stay behind an alpha feature flag. the v1 TEP is the source of truth for what should be in the v1 api

JeromeJu:
> We do need to keep our alpha fields around (including matrix); they should just stay behind an alpha feature flag. the v1 TEP is the source of truth for what should be in the v1 api

Oh sorry for missing this. I drafted earlier before the v0.38 release. Will bring that up. Thanks @lbernick

JeromeJu:
Shall we exclude the files generated by `./hack/update-codegen.sh` under pkg/client?

pkg/apis/pipeline/v1/pipeline_types_test.go
```diff
@@ -0,0 +1,1048 @@
...
+				Inputs: []PipelineTaskInputResource{{
+					From: []string{"task-1", "task-2"},
+				}},
+			}},
+		},
+		expectedDeps: map[string][]string{
+			"task-3": {"task-1", "task-2"},
+		},
+	}, {
+		name: "valid pipeline with resource deps - Task Results",
```
JeromeJu, line 457: Can I confirm that theses should be deleted as resources are deprecated?
(Keeping the when expression cases.)

lbernick: PipelineResources are deprecated, but results are not, and it looks like this test case is just using results. it should probably be renamed.

pkg/apis/pipeline/v1/pipeline_types.go
```diff
@@ -0,0 +1,654 @@
...
+	// RunAfter is the list of PipelineTask names that should be executed before
+	// this Task executes. (Used to force a specific ordering in graph execution.)
+	// +optional
+	// +listType=atomic
+	RunAfter []string `json:"runAfter,omitempty"`
+
+	// Resources declares the resources given to this task as inputs and
+	// outputs.
+	// +optional
+	Resources *PipelineTaskResources `json:"resources,omitempty"`
```
lbernick, line 194: this is deprecated and should be removed

pkg/apis/pipeline/v1/taskref_types.go
```diff
@@ -0,0 +1,49 @@
...
+	// Name of the referent; More info: http://kubernetes.io/docs/user-guide/identifiers#names
+	Name string `json:"name,omitempty"`
+	// TaskKind indicates the kind of the task, namespaced or cluster scoped.
+	Kind TaskKind `json:"kind,omitempty"`
+	// API version of the referent
+	// +optional
+	APIVersion string `json:"apiVersion,omitempty"`
+	// Bundle url reference to a Tekton Bundle.
+	// +optional
+	Bundle string `json:"bundle,omitempty"`
```
lbernick, line 30: this field won't be present in v1

pkg/apis/pipeline/v1/taskref_types.go
```diff
@@ -0,0 +1,49 @@
...
+// Check that Pipeline may be validated and defaulted.
+
+// TaskKind defines the type of Task used by the pipeline.
+type TaskKind string
+
+const (
+	// NamespacedTaskKind indicates that the task type has a namespaced scope.
+	NamespacedTaskKind TaskKind = "Task"
+	// ClusterTaskKind indicates that task type has a cluster scope.
+	ClusterTaskKind TaskKind = "ClusterTask"
```
lbernick, line 48: cluster tasks are also not coming to v1, I'd delete the whole type

pkg/apis/pipeline/v1/status_validation.go
```diff
@@ -0,0 +1,36 @@
...
+	"context"
+	"fmt"
+
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+	"knative.dev/pkg/apis"
+)
+
+// ValidateEmbeddedStatus checks that the embedded-status feature gate is set to the wantEmbeddedStatus value and,
+// if not, returns an error stating which feature is dependent on the status and what the current status actually is.
+func ValidateEmbeddedStatus(ctx context.Context, featureName, wantEmbeddedStatus string) *apis.FieldError {
```
lbernick, line 29: this file shouldn't be needed for Pipeline (only relevant to PipelineRun)

lbernick:
> Shall we exclude the files generated by `./hack/update-codegen.sh` under pkg/client?

No, these files are code generated by knative, which we need

pkg/apis/pipeline/v1/taskref_types.go
JeromeJu: I'm sorry, this was brought back by restoring some files after one squash and revert. 
Will double-check against the TEP96 later before any push.

pkg/apis/pipeline/v1/status_validation.go
JeromeJu: I think this is used in the call stack for validateMatrix, which uses the following func:
https://github.com/JeromeJu/pipeline/blob/5e7bc86121ed814b7066b57d5686b5338326abcd/pkg/apis/pipeline/v1/pipeline_types.go#L289

pkg/apis/pipeline/v1/pipeline_types_test.go
JeromeJu: Thanks Lee! Renamed at https://github.com/tektoncd/pipeline/pull/5228.

pkg/apis/pipeline/v1/status_validation.go
lbernick: Got it that makes sense! In v1, we will only allow this feature flag to be set to "minimal". I think this is OK to stay, but we should also add a similar validation function for PipelineRun (not in this PR) that prevents a v1 PipelineRun from being created unless the feature flag is set to "minimal".

JeromeJu: Thanks Lee. I've noted this down and will be applying to the PipelineRun change.

pkg/apis/pipeline/v1/pipeline_types.go
```diff
@@ -0,0 +1,566 @@
...
+// PipelineSpec returns the Pipeline's Spec, implementing PipelineObject
+func (p *Pipeline) PipelineSpec() PipelineSpec {
+	return p.Spec
+}
+
+// // Copy returns a deep copy of the Pipeline, implementing PipelineObject
+// func (p *Pipeline) Copy() PipelineObject {
+// 	return p.DeepCopy()
+// }
+
```
lbernick, line 81: should this be uncommented?

pkg/apis/pipeline/v1/pipeline_types.go
```diff
@@ -0,0 +1,566 @@
...
+	// +optional
+	TaskRef *TaskRef `json:"taskRef,omitempty"`
+
+	// TaskSpec is a specification of a task
+	// +optional
+	TaskSpec *EmbeddedTask `json:"taskSpec,omitempty"`
+
+	// WhenExpressions is a list of when expressions that need to be true for the task to run
+	// +optional
+	WhenExpressions WhenExpressions `json:"whenExpressions,omitempty"`
```
lbernick, line 178: Please don't change the json tag here; this changes the API which there's not a good reason to do here. Instead, change the name of the struct field to "when"

pkg/apis/pipeline/v1/taskref_types.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Check that Pipeline may be validated and defaulted.
+
+// TaskKind defines the type of Task used by the pipeline.
+type TaskKind string
+
+const (
+	// NamespacedTaskKind indicates that the task type has a namespaced scope.
+	NamespacedTaskKind TaskKind = "Task"
+)
```
lbernick, line 44: this can probably be removed

pkg/apis/pipeline/v1/taskref_types.go
```diff
@@ -0,0 +1,44 @@
...
+	APIVersion string `json:"apiVersion,omitempty"`
+
+	// ResolverRef allows referencing a Task in a remote location
+	// like a git repo. This field is only supported when the alpha
+	// feature gate is enabled.
+	// +optional
+	ResolverRef `json:",omitempty"`
+}
+
+// Check that Pipeline may be validated and defaulted.
```
lbernick, line 36: not sure what this comment is for, but it can probably be removed.

pkg/apis/pipeline/v1/pipeline_types.go
JeromeJu: I noticed in [`v1beta1/task_types.go`](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/task_types.go/#L71-L74) there is the deepcopy while in v1 task_types there is not. Thought we should do this as we don't have the interface moved?

pkg/apis/pipeline/v1/taskref_types.go
JeromeJu: I think these are referenced at [pipeline_defaults.go ](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/pipeline_defaults.go/#L41)

pkg/apis/pipeline/v1/pipeline_types.go
JeromeJu: Thanks Lee. In order to show it's a list, would it make any sense to rename it as `Whens`?

lbernick: What you have now is correct! It looks like you just need to re-run codegen/openapigen.

pkg/apis/pipeline/v1/pipeline_types.go
lbernick: I'm not sure what this is for actually but if it's not needed let's remove it rather than leaving it commented. we can always add it later.

pkg/apis/pipeline/v1/taskref_types.go
lbernick: Good catch. For v1 we'll want to get rid of `TaskRef.Kind` since all Tasks are Tasks (not ClusterTasks)

pkg/apis/pipeline/v1/pipeline_types.go
JeromeJu: Thanks @lbernick. It always got me that the `update-codegen` won't trigger `openapigen`, but it should.

pkg/apis/pipeline/v1/taskref_types.go
JeromeJu: Resolving this as CustomTask would also require the TaskRef.Kind

JeromeJu:
> Thanks Jerome! could you please:
> 
> * update the PR description to specify that the Go field name for `When` was changed, not the API
> * add a commit message body
> * double check whether there have been any changes to pkg/apis affecting these files since you opened this PR, and make sure they're incorporated here?

Thanks Lee:) Updated.

jerop (commented):
/lgtm

pkg/apis/pipeline/v1/pipeline_types.go
```diff
@@ -0,0 +1,561 @@
...
+		errs = errs.Also(pt.validateTask(ctx))
+	}
+	return
+}
+
+// Deps returns all other PipelineTask dependencies of this PipelineTask, based on ordering
+func (pt PipelineTask) Deps() []string {
+	deps := []string{}
+
+	deps = append(deps, pt.orderingDeps()...)
```
afrittoli, line 475: It seems to me that the removal of resource dependencies was a bit too eager here.
We do not need in `v1` dependencies related to `PipelineResources` but we still do need dependencies associated with `results` references.

PR #5234, opened 2022-07-28
Add V1 version to Task CRD

lbernick:
# Changes

This commit adds a v1 version of the Task CRD, and support to the webhook.
Since this version is not served, it will not be available to users.

/kind misc

# Submitter Checklist

- n/a Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/config/300-task.yaml
+++ b/config/300-task.yaml
@@ -43,6 +43,25 @@ spec:
     # starts to increment
     subresources:
       status: {}
+  - name: v1
+    served: false
+    storage: false
+    schema:
+      openAPIV3Schema:
+        type: object
+        # TODO(#1461): Add OpenAPIV3 schema
+        # OpenAPIV3 schema allows Kubernetes to perform validation on the schema fields
+        # and use the schema in tooling such as `kubectl explain`.
+        # Using "x-kubernetes-preserve-unknown-fields: true"
+        # at the root of the schema (or within it) allows arbitrary fields.
+        # We currently perform our own validation separately.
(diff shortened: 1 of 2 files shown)


lbernick:
Needs https://github.com/tektoncd/pipeline/pull/5202

jerop (commented):
`workspace-in-sidecar` is a known flake https://github.com/tektoncd/pipeline/issues/4169 

/retest

config/300-task.yaml
```diff
@@ -43,6 +43,18 @@ spec:
     # starts to increment
     subresources:
       status: {}
+  - name: v1
+    served: false
+    storage: false
+    schema:
+      openAPIV3Schema:
+        type: object
```
JeromeJu, line 51: Wondering if we might need the comments block from v1beta1 in #L34-L40, just in case when we remove v1beta1, this context is available in the v1 entry?
```
        # One can use x-kubernetes-preserve-unknown-fields: true
        # at the root of the schema (and inside any properties, additionalProperties)
        # to get the traditional CRD behaviour that nothing is pruned, despite
        # setting spec.preserveUnknownProperties: false.
        #
        # See https://kubernetes.io/blog/2019/06/20/crd-structural-schema/
        # See issue: https://github.com/knative/serving/issues/912
```

jerop (approved):

config/300-task.yaml
lbernick: I added a small explanation that is more up to date.

Issue #734, opened 2019-04-05
Schedule Pods in resource-constrained environments


ImJasonH:
# Expected Behavior

In a resource-constrained environment like a namespace with [resource limits](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/) imposed (or just an insufficiently provisioned cluster), creating a TaskRun (Pod) that exceeds those limits should not fail the TaskRun, but should instead continually try to create the Pod until it either succeeds or times out.

# Actual Behavior

Pods fail to start and the TaskRun is failed ~immediately.

# Steps to Reproduce the Problem

1. Define a namespace with resource constraints (e.g., 10 CPU, 10 GB RAM)
2. Create 15 TaskRuns each requesting 1 CPU and 1 GB RAM, running `hello world` or something simple
3. ~10 of those will be scheduled and will succeed, the rest will fail due to insufficient resources.

# Additional Info

This is similar to how [Jobs can handle Pod scheduling failures by retrying until they are successful](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#handling-pod-and-container-failures).

It's unclear whether users would expect TaskRuns waiting for sufficient resources to queue in order of the time they were created, or whether they'd expect the Kubernetes scheduler to do whatever it needs to do to schedule the Pods. As an initial implementation it's probably fine to have Kubernetes schedule Pods, and not have to worry about enforcing FIFO.

sbwsg:
Been working through some of the implementation details in a POC but want to drop current working notes here since I likely won't be able to work on it more until tomorrow.

1. Catching a pod failure is relatively straightforward; checking the error message produced by the `createPod()` func in `pkg/reconciler/v1alpha1/taskrun/taskrun.go` reveals the reason. From here it's quick to parse out the error message and look for e.g. "exceeded quota" in the string. This relies on a somewhat brittle contract though. I'll also need to check for the different messages generated both by LimitRanges as well as ResourceQuotas since it looks like they both enforce resource limits on a pod. I'm currently looking around to see if there's a less brittle approach to this error checking.

[...]

PR #5252, opened 2022-08-02
Fix for ResourceQuotaConflictError

yachna:
# Changes
Fix for the bug ResourceQuotaConflictError https://github.com/kubernetes/kubernetes/issues/67761

From @SaschaSchwarze0 
>We face the problem that when you create a resource (for example a Pod) that is constrained by a resource quota, then the creation can fail because of conflicts while updating the resource quota status (that's the long-standing Kubernetes issue that @yachna mentioned, https://github.com/kubernetes/kubernetes/issues/67761).

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Tekton will retry the creation of the Pod if it fails due to a conflict and results in ResourceQuotaConflictError while updating a ResourceQuota.
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -575,6 +575,11 @@ func (c *Reconciler) updateLabelsAndAnnotations(ctx context.Context, tr *v1beta1
 
 func (c *Reconciler) handlePodCreationError(tr *v1beta1.TaskRun, err error) error {
 	switch {
+	case isResourceQuotaConflictError(err):
+		// Requeue if it runs into ResourceQuotaConflictError Error i.e https://github.com/kubernetes/kubernetes/issues/67761
+		tr.Status.StartTime = nil
+		tr.Status.MarkResourceOngoing(podconvert.ReasonPending, fmt.Sprint("tried to create pod, but it failed with ResourceQuotaConflictError"))
+		return controller.NewRequeueAfter(time.Second)
 	case isExceededResourceQuotaError(err):
 		// If we are struggling to create the pod, then it hasn't started.
 		tr.Status.StartTime = nil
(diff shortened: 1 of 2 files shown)


dibyom:
Shouldn't this already already be handled in https://github.com/tektoncd/pipeline/blob/94055d92c120a6010f3d61a821e45dea4f893a74/pkg/reconciler/taskrun/taskrun.go#L577?

Also see https://github.com/tektoncd/pipeline/issues/734

SaschaSchwarze0:
> Shouldn't this already already be handled in
> 
> https://github.com/tektoncd/pipeline/blob/94055d92c120a6010f3d61a821e45dea4f893a74/pkg/reconciler/taskrun/taskrun.go#L577
> ?
> 
> Also see #734

Hi @dibyom, we are not facing an exceeded resource quota. We face the problem that when you create a resource (for example a Pod) that is constrained by a resource quota, then the creation can fail because of conflicts while updating the resource quota status (that's the long-standing Kubernetes issue that @yachna mentioned, https://github.com/kubernetes/kubernetes/issues/67761).

But maybe that function is where we need to move the retry logic @yachna, with maybe a requeue after one second.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -731,6 +736,10 @@ func (c *Reconciler) createPod(ctx context.Context, ts *v1beta1.TaskSpec, tr *v1
 	// to fetch it.
 	podName := pod.Name
 	pod, err = c.KubeClientSet.CoreV1().Pods(tr.Namespace).Create(ctx, pod, metav1.CreateOptions{})
+
+	// Requeue if it runs into ResourceQuotaConflictError Error i.e https://github.com/kubernetes/kubernetes/issues/67761
+	c.handlePodCreationError(tr, err)
```
SaschaSchwarze0, line 741: @yachna this addition needs to go away

SaschaSchwarze0 (changes requested):

pkg/reconciler/taskrun/taskrun.go
yachna: Removed

SaschaSchwarze0:
@dibyom we are ready here, code has been moved :-)

dibyom:
Thanks @yachna and @SaschaSchwarze0 The code looks fine to me. One request - could we update the commit message with a bit more description around what the issue was that this commit solves (like the description [here](https://github.com/tektoncd/pipeline/pull/5252#issuecomment-1204855613) so that it meets our commit guidelines: https://github.com/tektoncd/community/blob/main/standards.md#commits

vdemeester (approved):

yachna:
> Thanks @yachna and @SaschaSchwarze0 The code looks fine to me. One request - could we update the commit message with a bit more description around what the issue was that this commit solves (like the description [here](https://github.com/tektoncd/pipeline/pull/5252#issuecomment-1204855613) so that it meets our commit guidelines: https://github.com/tektoncd/community/blob/main/standards.md#commits

Hi @dibyom 

Thanks for the suggestion. I have updated the commit message accordingly.

Issue #4546, opened 2022-02-04
Handle deprecated fields during conversion from v1beta1 to v1


lbernick:
This issue is for discussing how to implement conversion between v1beta1 and v1 CRDs, as proposed in [TEP-0096](https://github.com/tektoncd/community/blob/main/teps/0096-pipelines-v1-api.md). This conversation came up in tektoncd/community#587.

One question we will need to answer is how to handle the "resources" field of Task/TaskRun/Pipeline/PipelineRun, since these will be removed in v1.

Copying over a comment from @sbwsg:

We serialize the Finally section of a v1beta1 Pipeline into an annotation if a client requests the v1alpha1 version. When a client submits a v1alpha1 version of a Pipeline with that annotation present the process reverses and the v1beta1 object is recreated with the Finally section in-tact: https://github.com/tektoncd/pipeline/blob/7a0a988cafaa64ff3e08969fa66f8521638d9b52/pkg/apis/pipeline/v1alpha1/pipeline_conversion.go#L142-L177

We don't do this for any other v1beta1-only fields. So, for example, when a v1alpha1 client requests a v1beta1 Pipeline with When Expressions, we simply drop them on the floor. If that client then re-applies the v1alpha1 version the stored v1beta1 version will be updated and the when expressions will be gone.

For a while we were [working on an approach](https://github.com/tektoncd/pipeline/pull/3854) where we take the entire v1beta1 object and serialize it into an annotation when clients request v1alpha1, but it proved to be a complex and imperfect workaround.

abayer:
Should this be folded into #4983?

vdemeester:
probably 🙃

lbernick:
Added it to the linked issues in that epic! I think this still has value remaining open as a subtask, to discuss what we want to do when we update the stored version to v1 and need to implement conversion from v1beta1 to v1.

lbernick:
@afrittoli if you have any context about past pitfalls we ran into when supporting deprecated fields on conversion from v1alpha1 -> v1beta1 this issue is probably the best place for it!

vdemeester:
> We don't do this for any other v1beta1-only fields. So, for example, when a v1alpha1 client requests a v1beta1 Pipeline with When Expressions, we simply drop them on the floor. If that client then re-applies the v1alpha1 version the stored v1beta1 version will be updated and the when expressions will be gone.

Note that this can happen no matter what we do 🙃. If the user / tool cleans the annotation(s), the situation is the same 👼🏼

PR #5253, opened 2022-08-02
V1: Add conversion for Task.Resources

lbernick:
# Changes

This commit adds support for Task.Resources when converting between v1beta1 and v1
versions of Tasks. This will allow us to release v1 Task in a backwards compatible way,
by ensuring that v1beta1 Tasks with Resources will have the Resources serialized into
annotations on the v1 Task on conversion.

Part n of #4546 

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/version/conversion.go
+++ b/pkg/apis/version/conversion.go
@@ -1,0 +1,57 @@
+/*
+Copyright 2022 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package version
+
+import (
+	"encoding/json"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SerializeToMetadata serializes the input field and adds it as an annotation to
+// the metadata under the input key.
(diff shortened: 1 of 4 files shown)


lbernick:
/retest

filed #5254

pkg/apis/version/conversion.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SerializeToMetadata serializes the input field and adds it as an annotation to
+// the metadata under the input key.
+func SerializeToMetadata(meta *metav1.ObjectMeta, field interface{}, key string) error {
```
JeromeJu, line 28: - Shall we have tests for the two exported functions?
- Do we need to test with `v1.Task`?

lbernick: > * Shall we have tests for the two exported functions?

Good point, thanks! Added a test.

> * Do we need test cases with `v1.Task` in task_conversion_tests?

No; here we are ensuring that a field of v1beta1 task that is not present in v1 is still there when converted to v1 and back. There's no equivalent for v1, since it's our highest version.

pkg/apis/version/version_validation_test.go
```diff
@@ -14,19 +14,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package version
+package version_test
```
jerop, line 17: why are the changes in this file included in this commit? seems to me that it's an unrelated cleanup

lbernick: Split into a separate PR: https://github.com/tektoncd/pipeline/pull/5319

jerop: happy to lgtm this when these changes in this file are removed from this commit

lbernick: my bad-- for some reason I forgot it was possible to have some tests in a "version" package and some in a "version_test" package so I just assumed this would wait for 5319 to merge. done.

PR #5222, opened 2022-07-26
[TEP-0075] Add variable usage and links to examples in docs

chuangw6:
# Changes

- Added instructions on how object variables can be used on Pipeline
and Task level.
- Added links to example yaml files in `docs/tasks.md`.

/kind misc

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ ] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [ ] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [ ] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [ ] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [ ] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ ] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
Update docs
```

PR #5255, opened 2022-08-02
Clean up Regexes

chuangw6:
# Changes

Prior to this commit, the regexes in substitution.go file were hard to
understand and hard to reuse because of the complexity of the regexes.
For example, we have to consider
  - all the three notations (dot notation, bracket with single quote,
  and bracket with double quote)
  - the operator suffix including star `[*]` operator and
  int indexing that was not considerd before.
Also the old regex coupled each notation part with the operator
part, which leads to extra step to be done in the extract-related functions
i.e. have to trim the [*] every time.

This PR
  - cleans up the regexes and make it as much reable and reuseble as possible
  - makes the method `extractVariablesFromString` be capable of extracting
  both variable name and the operator (indexing operator or star operator)
  examples:
    - $(params.myString)       - name: "myString", operator: ""
    - $(params['myArray'][2])  - name: "myArray", operator: "1"
    - $(params.myObject[\*])    - name: "myObject", operator: "*"

Enabling `extractVariablesFromString` to extract both name and operator
is particularly useful for array indexing validation. There will be more
PRs to come to clean up considering the complexity of the all regexes.
Also results-related regexes in resultref.go file need cleanup too.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1/param_types.go
+++ b/pkg/apis/pipeline/v1/param_types.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"regexp"
 	"strings"
 
 	resource "github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1"
@@ -28,12 +27,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/apis"
 )
-
-// exactVariableSubstitutionFormat matches strings that only contain a single reference to result or param variables, but nothing else
-// i.e. `$(result.resultname)` is a match, but `foo $(result.resultname)` is not.
-const exactVariableSubstitutionFormat = `^\$\([_a-zA-Z0-9.-]+(\.[_a-zA-Z0-9.-]+)*(\[([0-9]+|\*)\])?\)$`
-
-var exactVariableSubstitutionRegex = regexp.MustCompile(exactVariableSubstitutionFormat)
 
 // ParamsPrefix is the prefix used in $(...) expressions referring to parameters
 const ParamsPrefix = "params"
@@ -228,7 +221,7 @@ func (arrayOrString *ArrayOrString) applyOrCorrect(stringReplacements map[string
 
 	// if the stringVal is a string literal or a string that mixed with var references
 	// just do the normal string replacement
-	if !exactVariableSubstitutionRegex.MatchString(stringVal) {
+	if !isolatedResultVariableSubstitutionRegex.MatchString(stringVal) && !substitution.IsIsolatedArrayOrObjectParamRef(stringVal) {
 		arrayOrString.StringVal = substitution.ApplyReplacements(arrayOrString.StringVal, stringReplacements)
 		return
 	}
@@ -350,7 +343,7 @@ func validateParamStringValue(param Param, prefix string, paramNames sets.String
 	stringValue := param.Value.StringVal
 
 	// if the provided param value is an isolated reference to the whole array/object, we just check if the param name exists.
-	isIsolated, errs := substitution.ValidateWholeArrayOrObjectRefInStringVariable(param.Name, stringValue, prefix, paramNames)
+	isIsolated, errs := substitution.ValidateIsolatedArrayOrObjectRefInStringVariable(param.Name, stringValue, prefix, paramNames)
 	if isIsolated {
 		return errs
 	}
--- a/pkg/apis/pipeline/v1beta1/resultref.go
+++ b/pkg/apis/pipeline/v1beta1/resultref.go
@@ -44,20 +44,20 @@ const (
 	// ResultResultPart Constant used to define the "results" part of a pipeline result reference
 	ResultResultPart = "results"
 	// TODO(#2462) use one regex across all substitutions
-	// variableSubstitutionFormat matches format like $result.resultname, $result.resultname[int] and $result.resultname[*]
-	variableSubstitutionFormat = `\$\([_a-zA-Z0-9.-]+(\.[_a-zA-Z0-9.-]+)*(\[([0-9]+|\*)\])?\)`
-	// exactVariableSubstitutionFormat matches strings that only contain a single reference to result or param variables, but nothing else
+	// resultVariableSubstitutionFormat matches format like $result.resultname, $result.resultname[int] and $result.resultname[*]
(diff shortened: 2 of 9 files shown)


abayer:
/approve

Nice! Getting out of obfuscated regex hell is a good thing, to say the least.

pkg/substitution/substitution.go
```diff
@@ -199,33 +214,30 @@ func ValidateVariableIsolatedP(value, prefix string, vars sets.String) *apis.Fie
 	return nil
 }
 
-// ValidateWholeArrayOrObjectRefInStringVariable validates if a single string field uses references to the whole array/object appropriately
+// ValidateIsolatedArrayOrObjectRefInStringVariable validates if a single string field uses references to the whole array/object appropriately
```
lbernick, line 217: what is an isolated reference?

chuangw6: An isolated reference means a string that contains only a reference to a variable but nothing else i.e. no extra characters, no reference to other variables. Some examples

- Isolated reference: `"$(params.myParam)"`, `"$(params['myArray'][*])"`
- Non-isolated reference: `"this is param $(params.myParam)"`, `"the first param is $(params.first) and the second param is $(params["second"])"`

lbernick: Makes sense! could you add that as a comment in the code?

chuangw6: SG! Added! Thanks Lee.

abayer:
Ok, pull-tekton-pipeline-alpha-integration-tests keeps failing on the same example tests, which makes me think there is a legit issue.

chuangw6:
> Ok, pull-tekton-pipeline-alpha-integration-tests keeps failing on the same example tests, which makes me think there is a legit issue.

Thanks @abayer for helping retest. https://github.com/tektoncd/pipeline/pull/5222 that only updates doc files failed same 3 tests in alpha integration test. Let me retry.

abayer:
@chuangw6 The fact they keep failing in exactly the same way every time really makes me think there's a problem with the regex changes. Might be worth trying to replicate param patterns in those examples as unit tests?

abayer:
fwiw, I've verified locally that the `Pipeline` in examples/v1beta1/alpha/pipelinerun-param-array-indexing.yaml fails validation. So yeah, there's an issue. =)

abayer:
Ok, I took your branch and added this as a test case to `pkg/apis/pipeline/v1beta1/pipeline_validation_test.go`'s `TestValidatePipelineParameterVariables` test:

```
{
	name: "array index param - using the value of an array index as a string param",
	params: []ParamSpec{{
		Name: "myArray",
		Type: ParamTypeArray,
	}},
	tasks: []PipelineTask{{
		Name:    "bar",
		TaskRef: &TaskRef{Name: "bar-task"},
		Params: []Param{{
			Name: "a-param-intended-to-be-from-an-array-index", Value: ArrayOrString{Type: ParamTypeString, StringVal: $(params.myArray[0])"},
		}},
	}},
}
```
...and the test case fails on your branch, while passing on main. So yeah, something is definitely broken there.

chuangw6:
> fwiw, I've verified locally that the `Pipeline` in examples/v1beta1/alpha/pipelinerun-param-array-indexing.yaml fails validation. So yeah, there's an issue. =)

Thanks @abayer for helping find out this. I'll fix it.

chuangw6:
Hi @abayer,
I've solved the issue and also added the test. Please take a look. Thanks.

chuangw6:
@lbernick Since this PR has been changed a bit, please take a look again.
Thanks!

pkg/apis/pipeline/v1/param_types.go
```diff
@@ -228,7 +221,7 @@ func (arrayOrString *ArrayOrString) applyOrCorrect(stringReplacements map[string
 
 	// if the stringVal is a string literal or a string that mixed with var references
 	// just do the normal string replacement
-	if !exactVariableSubstitutionRegex.MatchString(stringVal) {
+	if !isolatedResultVariableSubstitutionRegex.MatchString(stringVal) && !substitution.IsIsolatedArrayOrObjectParamRef(stringVal) {
```
lbernick, line 224: hm it seems like the `substitution` package is responsible for determining whether a reference is an isolated reference or not. Can the regex being used be moved to that package instead (and put into a more useful helper function instead of exporting the regex)? With the existing code, each package is only partially responsible for determining how the param reference should be parsed, and I'd like to see better separation of concerns.

pkg/apis/pipeline/v1beta1/resultref.go
```diff
@@ -44,20 +44,20 @@ const (
...
 	arrayIndexing = `\[([0-9])*\*?\]`
 	// ResultNameFormat Constant used to define the the regex Result.Name should follow
 	ResultNameFormat = `^([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$`
 )
 
-// VariableSubstitutionRegex is a regex to find all result matching substitutions
-var VariableSubstitutionRegex = regexp.MustCompile(variableSubstitutionFormat)
-var exactVariableSubstitutionRegex = regexp.MustCompile(exactVariableSubstitutionFormat)
+// ResultVariableSubstitutionRegex is a regex to find all result matching substitutions
+var ResultVariableSubstitutionRegex = regexp.MustCompile(resultVariableSubstitutionFormat)
```
lbernick, line 59: does this need to be exported? based on its usage in the TaskRun reconciler, it looks like it would be more useful as part of a validation helper function.

pkg/substitution/substitution.go
```diff
@@ -366,3 +416,13 @@ func ExtractIndex(s string) (int, error) {
 func StripStarVarSubExpression(s string) string {
 	return strings.TrimSuffix(strings.TrimSuffix(strings.TrimPrefix(s, "$("), ")"), "[*]")
 }
+
+// parseIndex parses integer from a string.
+// If the string is not numeric, the second returned value will be false.
+func parseIndex(s string) (int, bool) {
```
JeromeJu, line 422: Wondering if we should stick with strconv.Atoi here? This func seems to be triming the error message to boolean.

JeromeJu: And also this func doesn't seem to be used elsewhere at the moment?

chuangw6:
Hey @lbernick @JeromeJu @abayer 
Thank you so much for taking your time to review this PR. I saw this as non-urgent and was distracted by other tasks, so I didn't work on this after then.

Just temporally mark this as closed. Will pick this up later and get it ready for review. Thanks again!

Issue #5152, opened 2022-07-18
flake: TestRealRunnerStdoutAndStderr* unit tests


abayer:
I've seen a couple of these fail a few times recently on CI with this output:

```
=== RUN   TestRealRunnerStdoutAndStderrPaths
hello world
    runner_test.go:51: stderr: got: , wanted: hello world
--- FAIL: TestRealRunnerStdoutAndStderrPaths (0.02s)
```

...and...

```
=== RUN   TestRealRunnerStdoutAndStderrSamePath
    runner_test.go:78: got: 0, wanted: 24
--- FAIL: TestRealRunnerStdoutAndStderrSamePath (0.01s)
```

I have not been able to reproduce locally.

/kind flake

lbernick:
Seeing this as well, I think it was added recently in #4882; @bradbeck would you mind taking a look at this? It might help to use a fake clock ([example in taskrun reconciler tests](https://github.com/tektoncd/pipeline/blob/3f539724f27a6add8b64d0712e056e4479750975/pkg/reconciler/taskrun/taskrun_test.go#L420)) as opposed to a real one

abayer:
From @dibyom on #5250 
> TestRealRunnerStdoutAndStderrPaths is testing out the expected stdout and stderr strings are written out to files. However, the writes happen [asynchronously](https://github.com/tektoncd/pipeline/blob/94055d92c120a6010f3d61a821e45dea4f893a74/cmd/entrypoint/runner.go#L139) which means that sometimes the output that we get is empty leading to the test being flaky.

dibyom:
One fix might be to have the test wait in a loop till the file has content before comparing (with a time limit of a few seconds say)

wlynch:
Took a crack at this in https://github.com/tektoncd/pipeline/pull/5261 since this blocked one of my other PRs 🙃 - would appreciate any feedback, in particular since this is going against what a comment in the original PR said not to do (but I think it's fine anyway?)

PR #5261, opened 2022-08-02
Implement stderr/stdout copying with exec pipes.

wlynch:
# Changes

This implements stdout/stderr with builtin exec command pipes, rather
than our own async writer.

There was a previous comment that mentioned the Std*Pipe funcs may not
work because a potential deadlock from a spawned child process (I think
in particular if they share the same fd? - not 100% sure). I don't think
we should really optimize for this case - if processes want to have
child processes my expectation is that they would handle synchronization
to make sure the logs make it back to the parent stdout/stderr, or
otherwise make sure the logs get exported somewhere else that would be
visible.
Not sure how controversial this is - @bradbeck @chhsia0 would appreciate feedback here 🙏 

Fixes https://github.com/tektoncd/pipeline/issues/5152 (did my best to verify this with `go test -race -count 100 .`

/kind flake

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


abayer:
/approve

Nice!

abayer:
Ok, the integration tests failed in the same on two consecutive runs, so I think those failures are legit. =)

bradbeck:
@wlynch I just resurrected the work that @chhsia0 had done and polished it. If you think there is a better way to handle the redirection I'm completely open to that.

wlynch:
This one's real to - e2e tests expect the alpha gate to be set locally 🤦

PR #5263, opened 2022-08-03
[TEP-0056] Cancellation support for pipeline-in-pipeline

abayer:
# Changes

See https://github.com/tektoncd/community/blob/main/teps/0056-pipelines-in-pipelines.md

This is the first of a series of PRs to prepare for implementing TEP-0056. These initial PRs will be focused on changes around reading `PipelineRun.Status.ChildReferences` to handle entries being `PipelineRun`s, without any changes to any existing behavior or additional fields, etc. This means that these PRs will not impact runtime behavior until the actual implementation of specifying and executing child `PipelineRun`s is done. Also, these changes don't need to be behind a feature flag, since they don't do anything without child `PipelineRun`s being executed, and that functionally will itself be behind the `alpha` feature flag once it's implemented.

Finally, there's no release note for this change since it's not something actually exposed to users at this point.

Also adds a test for `gracefullyCancelPipelineRun`.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/cancel.go
+++ b/pkg/reconciler/pipelinerun/cancel.go
@@ -37,7 +37,7 @@ import (
 	"knative.dev/pkg/apis"
 )
 
-var cancelTaskRunPatchBytes, cancelRunPatchBytes []byte
+var cancelTaskRunPatchBytes, cancelRunPatchBytes, cancelPipelineRunPatchBytes []byte
 
 func init() {
 	var err error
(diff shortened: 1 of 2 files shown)


abayer:
Note that the coverage drop in `cancel.go` is because the functionality handling failures to cancel a child `PipelineRun` isn't tested, which we can't _really_ test until we actually have child `PipelineRun`s. =)

Issue #5248, opened 2022-08-01
TaskRun retries: Improve separation of concerns between PipelineRun and TaskRun reconciler


lbernick:
Today, retries in Pipeline Tasks are implemented via the following process:
- If a TaskRun fails and the Pipeline Task has retries specified, the PipelineRun controller will clear the status of the TaskRun, add its old status to `taskRun.status.retriesStatus`, and mark it as running. (happens [here](https://github.com/tektoncd/pipeline/blob/b8e244a66b40da3d15e025eed92642461845ccee/pkg/reconciler/pipelinerun/pipelinerun.go#L810-L813))
- The TaskRun reconciler will use the length of its RetriesStatus to determine what to name the pod (happens [here](https://github.com/tektoncd/pipeline/blob/860b6285fe2494e9cb520d3b44ac6ada047aa836/pkg/pod/pod.go#L311-L313)

This pattern is a bit awkward because it means both reconcilers are partially responsible for implementing retries, and only the reconciler for a given CRD should be the one updating that CRD's status.

One way around this would be to create a new TaskRun for each retry instead. This was discussed as a potential implementation strategy when retries were created (in issue [221](https://github.com/tektoncd/pipeline/issues/221), initial PR [510](https://github.com/tektoncd/pipeline/pull/510), and final PR [658](https://github.com/tektoncd/pipeline/pull/658)), but it's not clear why the final decision was made.

If we'd like to go this route, here's what I think we should do:
- update the PipelineRun reconciler to create a new TaskRun for each retry, without changing the old TaskRun's status
- The new TaskRun can still have `RetriesStatus` based on the status of the old TaskRun.
- Because the new TaskRun will need a different name, it will create a different pod name anyway, so the logic around naming the pod based on the length of the retries status can be deleted.
- Now, RetriesStatus is solely informational. It serves no purpose to the TaskRun reconciler, so it can be removed from TaskRun.Status.

lbernick:
@XinruZhang I was just talking about this with @jerop: An equally valid way to support TaskRun retries and have separation of concerns is to add `retries` to `taskrun.spec`, and have the TaskRun reconciler implement retries. The advantage of this strategy is that we'd be consistent with how we handle retries for runs. This might need a bit more discussion before you start implementing--sorry!

XinruZhang:
Thanks @lbernick for bringing up this issue! Indeed the behavior here is a little bit weird 😅. I'm more than happy to discuss it more :)

[...]

PR #5269, opened 2022-08-03
Fix Existing Requests and Limits with LimitRange

austinzhao-go:
# Changes

This PR is to update `limitrange` logic to ensure container's requests and limits under the min and max range which is configured by limitrange.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
Fixed Existing Requests and Limits with LimitRange
```

### DIFF
--- a/pkg/internal/computeresources/transformer.go
+++ b/pkg/internal/computeresources/transformer.go
@@ -68,6 +68,9 @@ func transformPodBasedOnLimitRange(p *corev1.Pod, limitRange *corev1.LimitRange)
 
 	// FIXME(#4230) maxLimitRequestRatio to support later
 	defaultContainerLimits := getDefaultLimits(limitRange)
+	maxContainerLimits := getMaxLimits(limitRange)
+	minContainerLimits := getMinLimits(limitRange)
+
 	defaultContainerRequests := getDefaultContainerRequest(limitRange)
 	defaultStepContainerRequests := getDefaultStepContainerRequest(limitRange, nbStepContainers)
 
@@ -77,7 +80,7 @@ func transformPodBasedOnLimitRange(p *corev1.Pod, limitRange *corev1.LimitRange)
 			p.Spec.InitContainers[i].Resources.Requests = defaultContainerRequests
 		} else {
 			for _, name := range resourceNames {
-				setRequestsOrLimits(name, p.Spec.InitContainers[i].Resources.Requests, defaultContainerRequests)
+				setRequests(name, p.Spec.InitContainers[i].Resources.Requests, defaultContainerRequests)
 			}
 		}
 		// We are trying to set the highest limits possible
@@ -85,8 +88,12 @@ func transformPodBasedOnLimitRange(p *corev1.Pod, limitRange *corev1.LimitRange)
 			p.Spec.InitContainers[i].Resources.Limits = defaultContainerLimits
 		} else {
 			for _, name := range resourceNames {
-				setRequestsOrLimits(name, p.Spec.InitContainers[i].Resources.Limits, defaultContainerLimits)
+				setLimits(name, p.Spec.InitContainers[i].Resources.Limits, defaultContainerLimits, minContainerLimits, maxContainerLimits)
 			}
+		}
+		// ensure the requests are not higher than the limits
+		for _, name := range resourceNames {
+			setRequestsToLimits(name, p.Spec.InitContainers[i].Resources.Requests, p.Spec.InitContainers[i].Resources.Limits)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


austinzhao-go:
/hold 

till https://github.com/tektoncd/pipeline/pull/5082 is merged

pkg/internal/computeresources/transformer.go
```diff
@@ -100,24 +100,38 @@ func transformPodBasedOnLimitRange(p *corev1.Pod, limitRange *corev1.LimitRange)
...
+		dst[name] = compare.MinLimit(dst[name], limits[name])
+	}
+}
+
+// setLimits sets the limits to the min of (currentLimits, defaultLimits)
+func setLimits(name corev1.ResourceName, dst corev1.ResourceList, limits corev1.ResourceList) {
+	if compare.IsZero(limits[name]) {
+		return
+	}
+	dst[name] = compare.MinLimit(dst[name], limits[name])
```
austinzhao-go, line 134: main changes related to this PR (just to exclude the depended changes in `tasklevel.go`)

pkg/internal/computeresources/transformer_test.go
```diff
@@ -946,13 +1052,13 @@ func TestTransformerMultipleContainer(t *testing.T) {
 				Name: "sidecar-fizz",
 				Resources: corev1.ResourceRequirements{
 					Limits: corev1.ResourceList{
-						corev1.ResourceCPU:              resource.MustParse("4"),
+						corev1.ResourceCPU:              resource.MustParse("2"),
```
lbernick, line 1055: The LimitRange in this test case has a "default" of 2 but not a "max" of 2, so I think we want to leave the limit at 4.

pkg/internal/computeresources/transformer_test.go
```diff
@@ -310,7 +310,7 @@ func TestTransformerOneContainer(t *testing.T) {
 			Containers: []corev1.Container{{
 				Resources: corev1.ResourceRequirements{
 					Limits: corev1.ResourceList{
-						corev1.ResourceMemory:           resource.MustParse("4Gi"),
+						corev1.ResourceMemory:           resource.MustParse("2Gi"),
```
lbernick, line 313: Thanks Austin! Nice job catching this bug, it looks like the behavior here wasn't consistent with our [documentation](https://github.com/tektoncd/pipeline/blob/main/docs/compute-resources.md#limits). @vdemeester could you double check me here?

pkg/internal/computeresources/transformer_test.go
```diff
@@ -385,6 +385,112 @@ func TestTransformerOneContainer(t *testing.T) {
 				},
 			}},
 		},
+	}, {
+		description: "limitRange with min and max, while resources on containers < min",
```
lbernick, line 389:
```suggestion
		description: "limitRange with min and max, with requests on containers < min",
```

pkg/internal/computeresources/transformer_test.go
```diff
@@ -385,6 +385,112 @@ func TestTransformerOneContainer(t *testing.T) {
...
+						corev1.ResourceCPU: resource.MustParse("1"),
+					},
+					Limits: corev1.ResourceList{
+						corev1.ResourceCPU: resource.MustParse("2"),
+					},
+				},
+			}},
+		},
+	}, {
+		description: "limitRange with min and max, while resources on containers > max",
```
lbernick, line 442:
```suggestion
		description: "limitRange with min and max, with requests on containers > max",
```

pkg/internal/computeresources/transformer.go
```diff
@@ -100,24 +100,38 @@ func transformPodBasedOnLimitRange(p *corev1.Pod, limitRange *corev1.LimitRange)
...
 		}
 	}
 	return p
 }
 
-func setRequestsOrLimits(name corev1.ResourceName, dst, src corev1.ResourceList) {
-	if compare.IsZero(dst[name]) && !compare.IsZero(src[name]) {
-		dst[name] = src[name]
+// setRequests sets the requests to the max of (currentRequests, defaultRequests)
+// and the min of (currentRequests, defaultLimits)
```
lbernick, line 118: I think what actually needs to happen is that the requests are:
- greater than the min request
- smaller than the max limit
- set to the requests if they exist
- smaller than whatever limit we end up with

Using default limits here is not right if we have both requests and limits > default limits-- in that case we'd just want to use the specified requests and limits.

Also, this docstring reflects details of how the function is called (i.e. the "limits" are the default limits of the limitRange); could you reword it to focus on what the function does?

lbernick (commented):
Thanks Austin! Would you mind splitting the limitrange commit into a separate PR?

pkg/internal/computeresources/transformer_test.go
lbernick: I think the init container memory limits need to be changed too

pkg/internal/computeresources/transformer.go
austinzhao-go: I see. thanks for details about the logic. 

so requests will be ensured to have a value larger than Min of LimitRange while leaving limits set by the lower end of limits (of container) and LimitRange.

pkg/internal/computeresources/transformer_test.go
austinzhao-go: got this. so it means we actually want to overwrite the default limits whenever there are limits from containers right? 
I will update the setLimits as the "overwrite with default only if there are no limits on containers"

lbernick: yes, that's right. the [docs](https://github.com/tektoncd/pipeline/blob/main/docs/compute-resources.md#limits) describe our intended behavior

austinzhao-go: checked the code - the "default" in LimitRange was used to compute the `defaultContainerLimits`
https://github.com/tektoncd/pipeline/blob/99b6692a5cca86926650387830d05551236c323d/pkg/internal/computeresources/transformer.go#L193

in this case, by the doc, we still should keep the min of limits right?
so here - the defaultLimits as "4" and sidecar limits as "2", the expected is "2"

austinzhao-go:
/hold

till https://github.com/tektoncd/pipeline/pull/5279 got merged

pkg/internal/computeresources/transformer_test.go
lbernick: I think it's actually the other way around here. The default limits are 2 and the sidecar limits are 4.

lbernick:
Could you please remove the TEP title from the commit message + PR description? since this is an unrelated bug fix.

lbernick:
Fixes #5248 
Please also add a release note!

pkg/internal/computeresources/transformer_test.go
austinzhao-go: your are correct. the "4" is from the sidecar. 
so the result "2" comes from the lower end of both limits.

lbernick: We don't want to allow default limits to override limits set on the container. Only the limitrange max should override limits set on the container.

lbernick (commented):
We also need to address the test cases "limitRange with min and limits on containers < min" and "limitRange with max and requests on containers > max". These test cases are outputting invalid pods (requests > limits).

pkg/internal/computeresources/transformer_test.go
austinzhao-go: right, think your understanding matches the doc we are referring. 
just the code causing me a bit confused: https://github.com/tektoncd/pipeline/blob/4c775883dcb54431574343ac85be7423711f3e72/pkg/internal/computeresources/transformer.go#L197

here the max of LimitRange is used to compute the default limit (if there is no default) - so if we only allow override limits with default, think will need to split into 2 funcs - getDefaultLimit and getMaxLimit from LimitRange - so clear the "default" here we used

lbernick: got it-- let's make sure we have the test cases how we want them, and then we can change the code any way that's needed so we get the right results. If you need to split into 2 functions that's OK; for the review right now I'm just focusing on getting the test cases right.

austinzhao-go: understood your logic. thanks Lee!

austinzhao-go: kept the default as it was (since the max will be considered as the default if only max)
add a maxLimit getter func - so only take the min when max exists

austinzhao-go:
thanks @abayer for the review!

pushed an update to ensure `requests` got "capped" by `limits` finally

pkg/internal/computeresources/transformer_test.go
```diff
@@ -1140,7 +1140,7 @@ func TestTransformerMultipleContainer(t *testing.T) {
 					},
 					Requests: corev1.ResourceList{
 						corev1.ResourceCPU:              resource.MustParse("1"),
-						corev1.ResourceMemory:           resource.MustParse("400Mi"),
+						corev1.ResourceMemory:           resource.MustParse("200Mi"),
```
lbernick, line 1143: I think there was another comment thread about this, but this line should stay "400Mi".

austinzhao-go:
/hold 

till collecting further feedback from community

austinzhao-go:
closed the PR for this moment after a check with @lbernick 

the expected fix will be updated in another PR.

Issue #5278, opened 2022-08-05
LimitRange transformer behavior inconsistent with our documentation


lbernick:
# Behavior described in [docs](https://github.com/tektoncd/pipeline/blob/main/docs/compute-resources.md#limits)

If a taskrun is created with resource requirements inconsistent with a limitrange (e.g. requests < min or limits > max),
the resource requirements will be adjusted to fit within the limitrange constraints.

# Actual Behavior

If a taskrun is created with resource requirements inconsistent with a limitrange, the resulting pod will fail to schedule.
The pod may also have requests > limits and be invalid.

lbernick:
@skaegi curious about your opinion on whether we should change the docs or the behavior

lbernick:
Even if we decide to allow TaskRuns to fail when they don't meet limitrange requirements, we should definitely stop creating invalid pods. (This happens when e.g. a taskrun's requests are larger than the limitrange max or its limits are smaller than the limitrange min.)

skaegi:
My feeling is that other than the resource spreading algorithm for step "requests" we should not interfere. We should not change the "requests" for init containers or side-car containers, and we should not change "limits" in any containers.

We should create a valid schema-wise pod but don't try to second guess the admin's intent for limitranges and what other policy they might have in place. Let the pod or limitrange errors happen as they would for a Job, Deployment, etc.

PR #5279, opened 2022-08-05
Add tests for limitrange transformer

lbernick:
# Changes
This commit adds test cases for pods where container resource requirements
are inconsistent with limitranges already deployed in the namespace (for example,
with requests lower than the limitrange minimums).

This commit doesn't change existing behavior of the limitrange transformer.
The existing behavior is inconsistent with the behavior we have documented,
and may result in in an invalid pod (i.e. requests > limits). (#5278)
Either the code or the documentation will be fixed in a later PR.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/internal/computeresources/transformer_test.go
+++ b/pkg/internal/computeresources/transformer_test.go
@@ -385,6 +385,194 @@ func TestTransformerOneContainer(t *testing.T) {
 				},
 			}},
 		},
+	}, {
+		description: "limitRange with min and requests on containers < min",
+		limitranges: []corev1.LimitRangeItem{{
+			Type: corev1.LimitTypeContainer,
+			Min: corev1.ResourceList{
+				corev1.ResourceCPU: resource.MustParse("1"),
+			},
+		}},
+		podspec: corev1.PodSpec{
+			InitContainers: []corev1.Container{{
+				Name:  "bar",
+				Image: "foo",
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU: resource.MustParse("800m"),
+					},
+				},
+			}},
+			Containers: []corev1.Container{{
+				Name:  "step-foo",
+				Image: "baz",
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU: resource.MustParse("800m"),
+					},
+				},
+			}},
+		},
+		want: corev1.PodSpec{
+			InitContainers: []corev1.Container{{
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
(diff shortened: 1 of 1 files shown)


pkg/internal/computeresources/transformer_test.go
```diff
@@ -385,6 +385,194 @@ func TestTransformerOneContainer(t *testing.T) {
...
+						corev1.ResourceCPU: resource.MustParse("800m"),
+					},
+				},
+			}},
+		},
+		want: corev1.PodSpec{
+			InitContainers: []corev1.Container{{
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU: resource.MustParse("800m"),
```
austinzhao-go, line 420: here mean we don't want to reset requests as min right?

lbernick: I think we probably will want that, but my intention with this PR is not to change the existing behavior, just to have the test cases in place so we can decide what to do about them.

pkg/internal/computeresources/transformer_test.go
```diff
@@ -385,6 +385,194 @@ func TestTransformerOneContainer(t *testing.T) {
...
+						corev1.ResourceCPU: resource.MustParse("3"),
+					},
+				},
+			}},
+		},
+		want: corev1.PodSpec{
+			InitContainers: []corev1.Container{{
+				Resources: corev1.ResourceRequirements{
+					Limits: corev1.ResourceList{
+						corev1.ResourceCPU: resource.MustParse("3"),
```
austinzhao-go, line 514: do we need here to have the lower bound of the limits by doc - so "2"?

pkg/internal/computeresources/transformer_test.go
austinzhao-go: I see. just checked this as this leads to the failed TaskRun.

pkg/internal/computeresources/transformer_test.go
lbernick: Yes, I think we'll also want that in a follow-up.

pkg/internal/computeresources/transformer_test.go
austinzhao-go: closed - will discuss it later with the logic changes

pkg/internal/computeresources/transformer_test.go
austinzhao-go: closed - same as above

thanks Lee:)

abayer:
/approve

Seems solid to me - having tests verifying the existing behavior probably should have been there from the start, so we would have known the existing behavior wasn't what we thought it was! =)

Issue #5283, opened 2022-08-08
`runTestReconcileWithPipelineResults` test `PipelineRun` status set to True and is not allowing for validation


vsinghai:
# Expected Behavior
The test, `runTestReconcileWithPipelineResults`, is meant to test the reconciler with `Pipeline` `Results` passed in from `Tasks`. However, a problem with this test is the `PipelineRuns` `status.conditions.status` is set too `True`. This essentially makes the reconciler to see that this `PipelineRun` has already run successfully and does not need any form of validation, shown [here](https://github.com/tektoncd/pipeline/blob/deb0fec6e861ed9c74c7f43791b6e52823a62e0d/pkg/reconciler/pipelinerun/pipelinerun.go#L185) in the code.

Furthermore, the test does not fully test all the `embedded.status` fields. It does not have `childReference` fields for the `both embedded status` and `minimal embedded status`.

This test must be rewritten in order to properly test the reconciler with `Pipeline` `Results` passed from `Tasks` and to test the `embedded.status`. 

# Actual Behavior

Although the test passes, this is because the `PipelineRun` `status.conditions.status` is set to `True`. Since it is set to true, the reconciler thinks everything is successful and passed (since the status is set to True), and returns back no errors to the test.

# Additional Info

- Kubernetes version:

```
Client Version: v1.24.1
Kustomize Version: v4.5.4
Server Version: v1.22.10-gke.600
```

- Tekton Pipeline version:

```
Client version: 0.23.1
Pipeline version: devel
```

This test can be found in `pipelinerun_test.go`

vsinghai:
issue has been resolved.

PR #5284, opened 2022-08-08
Modified test to allow for validation

vsinghai:
# Changes
Prior to this commit, the previous iteration
of this test set the `PipelineRun` status to `True`,
essentially allowing the `PipelineRun` execute
without any validation. Furthermore, the test did
not test for the various `embedded.status`.

In this commit, we add implementation to test for
`Pipeline` `Results` which allow for validation to
execute by setting the `PipelineRun` status to `Unknown`.
Also, the test accounts for the different `embedded.status`.

Fixes bug [#5283](https://github.com/tektoncd/pipeline/issues/5283)

/kind bug
/cc @jerop

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -5146,78 +5146,74 @@ metadata:
   namespace: foo
 spec:
   results:
-  - description: pipeline result
-    name: result
-    value: $(tasks.a-task.results.aResult)
+    - description: pipeline result
+      name: result
+      value: $(tasks.a-task.results.a-Result)
+    - description: custom task pipeline result
+      name: custom-result
+      value: $(tasks.b-task.results.b-Result)
   tasks:
+  - name: b-task
+    taskRef:
+      apiVersion: example.dev/v0
+      kind: Example
+      name: b-task
   - name: a-task
     taskRef:
       name: a-task
-  - name: b-task
-    params:
-    - name: bParam
-      value: $(tasks.a-task.results.aResult)
-    taskRef:
-      name: b-task
 `)}
 	trs := []*v1beta1.TaskRun{mustParseTaskRunWithObjectMeta(t,
-		taskRunObjectMeta("test-pipeline-run-different-service-accs-a-task", "foo",
-			"test-pipeline-run-different-service-accs", "test-pipeline", "a-task", true),
+		taskRunObjectMeta("test-pipeline-run-results-task-run-a", "foo",
+			"test-pipeline-run-results", "test-pipeline", "a-task", true),
 		`
 spec:
-  resources: {}
-  serviceAccountName: test-sa
   taskRef:
     name: hello-world
-  timeout: 1h0m0s
 status:
   conditions:
   - status: "True"
     type: Succeeded
   taskResults:
-  - name: aResult
+  - name: a-Result
     value: aResultValue
+`)}
+
+	rs := []*v1alpha1.Run{mustParseRunWithObjectMeta(t,
+		taskRunObjectMeta("test-pipeline-run-results-task-run-b", "foo",
+			"test-pipeline-run-results", "test-pipeline", "b-task", true),
+		`
+spec:
+  ref:
+    apiVersion: example.dev/v0
+    kind: Example
+status:
+  conditions:
+  - status: "True"
+    type: Succeeded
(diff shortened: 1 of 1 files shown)


pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -5230,12 +5233,203 @@ spec:
...
+          value: aResultValue
+    test-pipeline-run-results-task-run-b:
+      pipelineTaskName: b-task
+      status:
+        conditions:
+        - status: "True"
+          type: Succeeded
+        taskResults:
+        - name: b-Result
+          value: bResultValue
```
jerop, line 5295: thought `b-task` should be a `Run`

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -5230,12 +5233,203 @@ spec:
...
+  - name: result
+    value: aResultValue
+  - name: custom-result
+    value: bResultValue
+  taskRuns: {}
+  childReferences:
+  - apiVersion: tekton.dev/v1beta1
+    kind: TaskRun
+    name: test-pipeline-run-results-task-run-b
+    pipelineTaskName: b-task
```
jerop, line 5411: thought `b-task` should be a `Run` -- the `kind` should be `Example`

vsinghai: it should be, however in `yaml.go` if you try to create a function called `MustParseExample` and set the kind to `Example`, it complains since `Example` is not a valid type within `v1beta1`

vsinghai:
TestPipelineRunTimeout is a known flake - https://github.com/tektoncd/pipeline/issues/3460

pkg/reconciler/pipelinerun/pipelinerun_test.go
vsinghai: i think regardless of what function you use, the main problems are:
1. `apiVersion: example.dev/v0` is not a valid package when used in `yaml.go`. I have tried it before, forgot what file it routes to, however it did not allow the `example.dev` package to be run.
2. As a result of above, trying to write `kind: Example` does not work either.

abayer: So `example.dev/v0` isn't on the `Run`, it's only on the `taskRef`. The actual `Run` is a standard `tekton.dev/v1alpha1` `Run`. See https://github.com/tektoncd/pipeline/blob/78f706bcdf100bc394de402315e910d1cdbeb91d/pkg/reconciler/pipelinerun/pipelinerun_test.go#L599 (and the `"simple custom task with minimal embedded status"` test case specifically, since you're using minimal status here) for an example of how to use `Run` in this way.

pkg/reconciler/pipelinerun/pipelinerun_test.go
vsinghai: done

pkg/reconciler/pipelinerun/pipelinerun_test.go
vsinghai: thank you for the clarification and feedback @abayer ! I believe my latest commit should resolve this conversation.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -5146,78 +5146,75 @@ metadata:
...
-    test-pipeline-run-different-service-accs-a-task:
-      pipelineTaskName: a-task
-      status:
-        conditions:
-        - status: "True"
-          type: Succeeded
-        taskResults:
-        - name: aResult
-          value: aResultValue
+    reason: Succeeded
```
jerop, line 5208: we should change this `reason` to `"running"` because when the status is `"unknown"` it means that the pipelinerun is running....we could also just remove the `reason` field to keep it simple

jerop (approved):
thank you for fixing this test @vsinghai!

planning to create a separate commit fixing the custom task in pipeline results with finally?

pkg/reconciler/pipelinerun/pipelinerun_test.go
vsinghai: deleted to keep it simple, thank you for the feedback!

PR #5288, opened 2022-08-09
Make pipeline cancel robust to missing resources

afrittoli:
# Changes

When a pipelinerun is cancelled, it may be that some of the resources
managed by the pipelinerun are missing (NotFound). In this case we
should not fail, and let the pipelinerun be cancelled.

Discussion about this was initially triggered by
https://github.com/tektoncd/pipeline/issues/5282, but this feature
makes sense regardless of the bug that exposed its need.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
A PipelineRun can be cancelled even if some of its owned resources have been deleted.
```

/kind feature

### DIFF
--- a/pkg/reconciler/pipelinerun/cancel.go
+++ b/pkg/reconciler/pipelinerun/cancel.go
@@ -32,6 +32,7 @@ import (
 	"go.uber.org/zap"
 	jsonpatch "gomodules.xyz/jsonpatch/v2"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/pkg/apis"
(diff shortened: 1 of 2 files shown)


abayer (approved):

pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -73,6 +74,21 @@ func init() {
 
 func cancelRun(ctx context.Context, runName string, namespace string, clientSet clientset.Interface) error {
 	_, err := clientSet.TektonV1alpha1().Runs(namespace).Patch(ctx, runName, types.JSONPatchType, cancelRunPatchBytes, metav1.PatchOptions{}, "")
+	if errors.IsNotFound(err) {
```
pritidesai, line 77: Should we instead first check if the resource exists before cancelling it?

abayer: I personally prefer just ignoring not found errors - it keeps the code simpler and achieves the same end result.

pritidesai: I agree but that causes unnecessary error showing up in the logs which can be confusing. I would rather avoid such error which will be ignored later by the same function

abayer: I don't think it would show any error - it just returns nil if the `TaskRun` or `Run` doesn't exist.

Tomcli:
thanks @afrittoli will this be part of the Tekton 0.38.3 release as well?

vdemeester (approved):
/lgtm

PR #5291, opened 2022-08-09
Move validation from `pipelinespec` to `pipelinerunspec` when propagating parameters

chitrangpatel:
# Changes

Prior to this, propagating parameters only skipped webhook validation for params defined in script. As a result, when users tried to propagate params to other fields like `args` or `command`, etc. an webhook validation was raised. This PR address issue https://github.com/tektoncd/pipeline/issues/5141. This PR addresses validations in `pipelinespec` and `pipelinerunspec`. The changes for validations in `taskspec` and `taskrunspec` are in a separate [PR](https://github.com/tektoncd/pipeline/pull/5143).

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Move parameter validation from `pipelinespec` to `pipelinerunspec` when propagating parameters
```

### DIFF
--- a/pkg/apis/pipeline/v1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1/pipeline_validation.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/validate"
 	"github.com/tektoncd/pipeline/pkg/reconciler/pipeline/dag"
 	"github.com/tektoncd/pipeline/pkg/substitution"
@@ -38,6 +39,7 @@ func (p *Pipeline) Validate(ctx context.Context) *apis.FieldError {
 	if apis.IsInDelete(ctx) {
 		return nil
 	}
+	ctx = config.SkipValidationDueToPropagatedParametersAndWorkspaces(ctx, false)
 	return errs.Also(p.Spec.Validate(apis.WithinSpec(ctx)).ViaField("spec"))
 }
 
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
@@ -62,8 +62,11 @@ func (ps *PipelineRunSpec) Validate(ctx context.Context) (errs *apis.FieldError)
 
 	// Validate PipelineSpec if it's present
 	if ps.PipelineSpec != nil {
+		ctx = config.SkipValidationDueToPropagatedParametersAndWorkspaces(ctx, true)
 		errs = errs.Also(ps.PipelineSpec.Validate(ctx).ViaField("pipelineSpec"))
 	}
+	// Validate propagated parameters
+	errs = errs.Also(ps.validateInlineParameters(ctx))
 
 	if ps.Timeout != nil {
 		// timeout should be a valid duration of at least 0.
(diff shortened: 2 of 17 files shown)


pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
```diff
@@ -107,14 +110,84 @@ func (ps *PipelineRunSpec) Validate(ctx context.Context) (errs *apis.FieldError)
...
 	}
 
 	return errs
 }
 
+// validateInlineParameters validates parameters that are defined inline.
+// This is cruicial for propagated parameters since the parameters could
+// be defined under pipelineRun and then called directly in the task steps.
+// In this case, paratemers cannot be validated by the underlying pipelineSpec
+// or taskSpec since they may not have the parameters declared because of propagation.
```
jerop, line 124:
```suggestion
// validateInlineParameters validates parameters that are defined inline.
// This is crucial for propagated parameters since the parameters could
// be defined under pipelineRun and then called directly in the task steps.
// In this case, parameters cannot be validated by the underlying pipelineSpec
// or taskSpec since they may not have the parameters declared because of propagation.
```

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
```diff
@@ -167,9 +271,15 @@ func TestPipelineRun_Validate(t *testing.T) {
...
+				Params: []v1beta1.Param{{
+					Name: "pipeline-words",
+					Value: v1beta1.ArrayOrString{
+						ArrayVal: []string{"hello", "pipeline"},
+					},
+				}},
 				PipelineSpec: &v1beta1.PipelineSpec{
 					Params: []v1beta1.ParamSpec{{
-						Name: "pipeline-words",
+						Name: "default-pipeline-words",
```
jerop, line 282: why did you rename this param?

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
```diff
@@ -182,20 +292,50 @@ func TestPipelineRun_Validate(t *testing.T) {
 						}},
 						TaskSpec: &v1beta1.EmbeddedTask{TaskSpec: v1beta1.TaskSpec{
 							Params: []v1beta1.ParamSpec{{
-								Name: "task-words",
+								Name: "default-task-words",
```
jerop, line 295: why did you rename this param?

jerop (approved):
thanks @chitrangpatel!

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
chitrangpatel: Same reason as above. Let me know if you need more clarification.

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
jerop: yes, calling it `default` when there's a default field was what was confusing -- yes please rename

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
jerop: the discussion above covered it, thanks

pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
chitrangpatel: Thanks!

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
chitrangpatel: done

pkg/apis/pipeline/v1/pipeline_validation.go
```diff
@@ -38,6 +39,7 @@ func (p *Pipeline) Validate(ctx context.Context) *apis.FieldError {
 	if apis.IsInDelete(ctx) {
 		return nil
 	}
+	ctx = config.SetValidateParameterVariablesAndWorkspaces(ctx, true)
```
abayer, line 42: I should have mentioned this on the other PR, but this name is confusing to me. Just based on the name alone, it's not really clear what changes when you set this. It feels like it's saying "ok, validate parameters and workspaces" in general, which seems like something you'd never actually want to turn off. Could we get this a more clear name to better signify exactly what it's doing?

chitrangpatel: Yes, that's right. How about I have a flag that is defaulted to `true` and we only call it when we intend to make it `false`? 

e.g.

we could have `config.DoNotValidateParameterVariablesAndWorkspaces(ctx)` which would set it the underlying context field to `false`. 

We also have a function `config.ValidateParameterVariablesAndWorkspaces(ctx)` which returns a `true` or `false` based on the value of the context field. This function is called by `task_validation` to decide whether to validate parameters and workspaces. We can have it return `true` by default.

Does that look like a good solution?

abayer: I'd really like the name to be more clear than `ValidateParameterVariablesAndWorkspaces` - that name is just too vague. What does it relate to? Which parameter variables and workspaces? If we're just using it for parameter propagation cases, it feels like it would make more sense to have the name be something regarding parameter propagation validation, i.e., _why_ we're setting this flag, not what the result of setting this flag is.

chitrangpatel: I see what you mean. Its trying to skip validation when it does not contain the full information. e.g. When propagating parameters, a user could provide the params at the task run or the pipeline run level and directly call them in the steps. However, when we validate the underlying taskSpec or pipelineSpec, we don't have these params declared there and so the web hook would throw an error. This flag allows us to specify where to skip validation of params due to lack of information.

I could call the it `SkipValidationDueToPropagatedParametersAndWorkspaces` (a bit of a mouthful, but ok I guess)?

abayer: That would definitely be better, yeah.

chitrangpatel: I will make the appropriate changes.

chitrangpatel: @abayer done! Thanks for the suggestion.

JeromeJu:
It seems that some `ArrayOrString` params in https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go/#L130 should be updated to `ParamValue`.  

cc @Yongxuanzhang @chitrangpatel Could you please help confirm here?

chitrangpatel:
I agree. I don't recall which PR was merged first. There is backwards compatibility which is why the tests work. However, we should make it ParamValue. I can do that in one shot in a separate PR.

JeromeJu:
Thanks @chitrangpatel

PR #5292, opened 2022-08-09
Prevent infinite recursion in skipBecauseResultReferencesAreMissing

### LINKED ISSUE #5271: restrict PipelineRunSpec.Params with a task result reference
# Expected Behavior
The `alpha` features become available

# Actual Behavior
The tekton-pipelines-controller crashed with this panic message: `fatal error: stack overflow`. 

I created a gist with the full stack trace
https://gist.github.com/RafaeLeal/00c944a670bdba9a178b23da9eeb0e7c

It seems to have entered a 3-function infinite recursion loop:
- On function [skipBecauseResultReferencesAreMissing we call Skip](https://github.com/tektoncd/pipeline/blob/v0.35.1/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L330)
- then on function [Skip we call skip](https://github.com/tektoncd/pipeline/blob/v0.35.1/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L275)
- then on function [skip we call back to skipBecauseResultReferencesAreMissing](https://github.com/tektoncd/pipeline/blob/v0.35.1/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L249)

But I don't know how we got to this loop, or if this logic is indeed meant to be recursive.

I'm not sure exactly how to reproduce this problem.. but I had a cluster with around ~400 pipelineruns running while I updated the config map `feature-flags` to set `enable-api-fields: alpha`. 
Even after rolling back this change (setting `enable-api-fields` back to `stable`), the problem remained. Only after we deleted all pipelineruns we were able to restore the `tekton-pipelines-controller` health.

# Additional Info

- Kubernetes version:
  ```
  Server Version: version.Info{Major:"1", Minor:"22+", GitVersion:"v1.22.9-eks-a64ea69", GitCommit:"540410f9a2e24b7a2a870ebfacb3212744b5f878", GitTreeState:"clean", BuildDate:"2022-05-12T19:15:31Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
  ```

- Tekton Pipeline version: `v0.35.1`

abayer:
# Changes

fixes #5271

While I have not been able to determine why setting `enable-api-fields` to `alpha` while `PipelineRun`s using results are in flight hits this, the immediate symptom in the linked issue is that an infinite recursive loop is hit in `skipBecauseResultReferencesAreMissing`, due to `skipBecauseResultReferencesAreMissing` calling `rpt.Skip(facts)`, which calls `rpt.skip(facts)`, which ends up calling `rpt.skipBecauseResultReferencesAreMissing` again, etc. Changing the call to `rpt.Skip(facts).SkippingReason == v1beta1.WhenExpressionsSkip` to `rpt.skipBecauseWhenExpressionsEvaluatedToFalse(facts)` not only gets rid of the potential for infinite recursion, it also more directly checks the condition we care about there in the first place.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fixes a potential infinite loop when a Pipeline task is missing needed results.
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -412,7 +412,7 @@ func (t *ResolvedPipelineTask) skipBecauseResultReferencesAreMissing(facts *Pipe
 		resolvedResultRefs, pt, err := ResolveResultRefs(facts.State, PipelineRunState{t})
 		rpt := facts.State.ToMap()[pt]
 		if rpt != nil {
-			if err != nil && (t.IsFinalTask(facts) || rpt.Skip(facts).SkippingReason == v1beta1.WhenExpressionsSkip) {
+			if err != nil && (t.IsFinalTask(facts) || rpt.skipBecauseWhenExpressionsEvaluatedToFalse(facts)) {
 				return true
 			}
 		}


pritidesai:
thanks @abayer, will appreciate some unit tests, its easier to review 😆

abayer:
> thanks @abayer, will appreciate some unit tests, its easier to review 😆

I completely failed to reproduce the conditions for the infinite recursion, so there's no new unit test, just the existing ones still passing without any changes, demonstrating that this doesn't result in any behavioral changes.

lbernick (approved):

pritidesai:
Unfortunately, there are no unit tests for `skip`, `skipBecauseWhenExpressionsEvaluatedToFalse`, `skipBecauseResultReferencesAreMissing`, etc.

abayer:
> Unfortunately, there are no unit tests for `skip`, `skipBecauseWhenExpressionsEvaluatedToFalse`, `skipBecauseResultReferencesAreMissing`, etc.

True, but they are actually well-covered by the tests for `Skip` - I checked. =)

pritidesai:
Yup, I will it take back 🙃 I just ran `gocover` locally and they are very well covered:
Screen Shot 2022-08-10 at 12 24 16 PM

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -412,7 +412,7 @@ func (t *ResolvedPipelineTask) skipBecauseResultReferencesAreMissing(facts *Pipe
 		resolvedResultRefs, pt, err := ResolveResultRefs(facts.State, PipelineRunState{t})
 		rpt := facts.State.ToMap()[pt]
 		if rpt != nil {
-			if err != nil && (t.IsFinalTask(facts) || rpt.Skip(facts).SkippingReason == v1beta1.WhenExpressionsSkip) {
+			if err != nil && (t.IsFinalTask(facts) || rpt.skipBecauseWhenExpressionsEvaluatedToFalse(facts)) {
```
pritidesai, line 415: Is `skipBecauseWhenExpressionsEvaluatedToFalse` direct substitution of  `rpt.Skip(facts).SkippingReason == v1beta1.WhenExpressionsSkip`?

In this function `skipBecauseResultReferencesAreMissing`, the task results in `finally` tasks or `whenExpressions` in `dag` tasks, or `whenExpressions` in `finally` tasks  are resolved.

If the result resolution fails and returns an error, then it checks if it's a `finally` task or if this task appears in the list of  `Skipped` tasks with `skipping` reason set to `WhenExpressionsSkip`.

`skipBecauseWhenExpressionsEvaluatedToFalse` just checks if the `WhenExpression` evaluates to `true` or `false` irrespective of if it’s cached in the list of skipped tasks.

Do we need to maintain such `cache` if it's not useful? The `cache` was introduced for huge pipelines with large set of dependencies and it sure improved the task scheduling.

pritidesai:
I don't mind merging this change but I haven't got chance to run any additional tests. 

I don't think this change is fixing https://github.com/tektoncd/pipeline/issues/5271 since the proposed change does not execute any different with or without `alpha` flag. Does it?

It will be great to at least know what is the pipeline configuration while running into https://github.com/tektoncd/pipeline/issues/5271.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
abayer: Also, the cache is still very much useful, it's just that something (and again, I haven't been able to figure out what exactly) is happening when you switch `enable-api-fields` to `alpha` mid-`PipelineRun` that's resulting in infinite recursion. It's definitely an edge case.

Maybe the answer here is actually just to flip the order of the `case`s for `t.skipBecauseResultReferencesAreMissing(facts)` and `t.skipBecauseWhenExpressionsEvaluatedToFalse(facts)` in `skip`? Hmm, probably not, because you need to be sure you've got all results that could be referenced in when expressions before evaluating the when expressions...this one's a bit ugly.

abayer:
/hold

> I don't mind merging this change but I haven't got chance to run any additional tests.

I'd actually like to hold it a bit - I think I oversimplified this a bit in my head.

> I don't think this change is fixing #5271 since the proposed change does not execute any different with or without `alpha` flag. Does it?

It's meant to fix the immediate issue in #5271 of the infinite recursion, but not whatever the root cause was.

> It will be great to at least know what is the pipeline configuration while running into #5271.

I've got a dump of all `PipelineRun`s (with inline pipeline specs) running on the user's setup when they changed to `alpha`, but there are a lot of them, and they're quite big, so trying to narrow down what exactly the trigger is has been...a challenge. I'll come back to that tomorrow and just spend the day trying to reproduce the loop in a canned test case.

pritidesai:
thanks a bunch @abayer for all the efforts on this, appreciate your guidance and troubleshooting this 🙏

abayer:
Closing this because I am 99% sure I figured out the problem in https://github.com/tektoncd/pipeline/issues/5271#issuecomment-1218341861, and this is not the way to prevent the scenario caused by the now-removed implicit parameters behavior from happening again in the future. =)

Issue #5241, opened 2022-07-30
PipelineRun 's PodTemplate support affinity


lookis:
### Feature request

currently, PipelineRun support PodTemplate only three feature: tolerations, nodeSelector, and imagePullSecrets. and I would like to see affinity be supported.

### Use case

In my case, I prefer all the task running on the high performance node which I will label it. but PipelineRun scheduler doesn't support affinity.nodeAffinity. 

and I cannot guarantee the high performance node will avalible all the time, so nodeSelector makes assumption too strong.

lookis:
any update? @yuzp1996

yuzp1996:
> any update? @yuzp1996

Hi! I will start working on this issue today.

yuzp1996:
Hi @lookis,  Podtemplate supports affinity now. This document [podtemplates](https://tekton.dev/docs/pipelines/podtemplates/) could help you.

But if your PipelineRun contains PVC then affinity will be overwritten by affinity-assistant. 

I tried and find two solutions to avoid being overwritten by affinity-assistant.
1.  delete PVC in your pipelinerun
2. disable affinity-assistant by setting feature-flag disable-affinity-assistant to be true 
<img width="1380" alt="image" src="https://user-images.githubusercontent.com/23007448/183874665-cda14793-4225-42cc-9367-e3e7c8e31b95.png">

yuzp1996:
@lookis  For more information about Affinity Assistant, you can refer to this document https://tekton.dev/vault/pipelines-v0.16.3/workspaces/#specifying-workspace-order-in-a-pipeline-and-affinity-assistants.

lookis:
@yuzp1996 I think it would be better to merge two configurations, that's what nodeAffinity.weight for.

It's no reason for pvc case not to prefer a better node, thought it may failed, that's why it's nodeAffinity rather than nodeSelector

yuzp1996:
@lookis  Yes, I agree with you that merging the two configurations is the better approach. I made changes in my dev environment and tested some simple cases and everything seems to be working fine. Maybe tomorrow I will create the PullRequest because I need to add some unit tests and test some more complex cases.

yuzp1996:
PR https://github.com/tektoncd/pipeline/pull/5306 has merged 
/close

lookis:
Thx for the merge.

PR #5306, opened 2022-08-11
Merge affinity from podtempalte and affinity-assistant

yuzp1996:
# Changes

No matter user provide affinity in podtemplate or not the
affinity-assistant will overwrite the affinity in pod.

However if user set affinity in podtemplate merge
podtemplate's affinity with affinity-assistant's affinity
is better way.

So we inject the affinity of the affinity-assitant into the
affinity of the podtempalte as the affinity of the pod to
make sure that all the affinities work properly.

Related Issue: https://github.com/tektoncd/pipeline/issues/5241
# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
If the user provide affinity in podtempalte it will merge with affinity-assistant's affinity

action required: Need to check podtemplate make sure the change will not cause unexpected behaviour
```

### DIFF
--- a/pkg/internal/affinityassistant/transformer.go
+++ b/pkg/internal/affinityassistant/transformer.go
@@ -28,32 +28,39 @@ import (
 // NewTransformer returns a pod.Transformer that will pod affinity if needed
 func NewTransformer(_ context.Context, annotations map[string]string) pod.Transformer {
 	return func(p *corev1.Pod) (*corev1.Pod, error) {
-		// Using node affinity on taskRuns sharing PVC workspace, with an Affinity Assistant
-		// is mutually exclusive with other affinity on taskRun pods. If other
-		// affinity is wanted, that should be added on the Affinity Assistant pod unless
-		// assistant is disabled. When Affinity Assistant is disabled, an affinityAssistantName is not set.
+		// Using node affinity on taskRuns sharing PVC workspace.  When Affinity Assistant
+		// is disabled, an affinityAssistantName is not set.
 		if affinityAssistantName := annotations[workspace.AnnotationAffinityAssistantName]; affinityAssistantName != "" {
-			p.Spec.Affinity = nodeAffinityUsingAffinityAssistant(affinityAssistantName)
+			if p.Spec.Affinity == nil {
+				p.Spec.Affinity = &corev1.Affinity{}
+			}
+			mergeAffinityWithAffinityAssistant(p.Spec.Affinity, affinityAssistantName)
 		}
 		return p, nil
 	}
 }
 
-// nodeAffinityUsingAffinityAssistant achieves Node Affinity for taskRun pods
-// sharing PVC workspace by setting PodAffinity so that taskRuns is
-// scheduled to the Node were the Affinity Assistant pod is scheduled.
-func nodeAffinityUsingAffinityAssistant(affinityAssistantName string) *corev1.Affinity {
-	return &corev1.Affinity{
-		PodAffinity: &corev1.PodAffinity{
-			RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{{
-				LabelSelector: &metav1.LabelSelector{
-					MatchLabels: map[string]string{
-						workspace.LabelInstance:  affinityAssistantName,
-						workspace.LabelComponent: workspace.ComponentNameAffinityAssistant,
-					},
-				},
-				TopologyKey: "kubernetes.io/hostname",
-			}},
+func mergeAffinityWithAffinityAssistant(affinity *corev1.Affinity, affinityAssistantName string) {
+	podAffinityTerm := podAffinityTermUsingAffinityAssistant(affinityAssistantName)
+
+	if affinity.PodAffinity == nil {
+		affinity.PodAffinity = &corev1.PodAffinity{}
+	}
+
+	affinity.PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution =
+		append(affinity.PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution, *podAffinityTerm)
+}
+
+// podAffinityTermUsingAffinityAssistant achieves pod Affinity term for taskRun
+// pods so that the taskRun is scheduled to the Node where the Affinity Assistant pod
+// is scheduled.
+func podAffinityTermUsingAffinityAssistant(affinityAssistantName string) *corev1.PodAffinityTerm {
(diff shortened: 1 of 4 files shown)


yuzp1996:
Need to add unit test.

lbernick:
I'm not sure whether this is a good idea or not (maybe it is! I'm just not sure). It seems like in this PR, if a user has set affinity from the pod template, both that affinity and the affinity assistant affinity will be applied to the resulting pod. I'm worried this might result in some confusing behavior. What if the user applies a pod affinity that isn't really compatible with the affinity assistant?

I'm also wondering if any node affinity specified by the user should be applied to the affinity assistant pod as well since it's scheduled first (again, not sure what the right answer is here).

yuzp1996:
> I'm not sure whether this is a good idea or not (maybe it is! I'm just not sure). It seems like in this PR, if a user has set affinity from the pod template, both that affinity and the affinity assistant affinity will be applied to the resulting pod. I'm worried this might result in some confusing behavior. What if the user applies a pod affinity that isn't really compatible with the affinity assistant?

I think maybe I understand your concern and here is my thoughts.

If I set affinity in pipelinerun's podtemplate I think the affinity priority in podtemplate should be the highest and it should not be overwritten by anything. On the other hand,  I think affinity-assistant should only serve as an aid, helping to find  better node but not change the way to find a better node.

And if the user applies a pod affinity that isn't really compatible with the affinity assistant the result is that pod will not be scheduled and I guess it should be correct. In this case, the user can disable affinity-assistant or change the affinity in podtemplate to ensure a better node can be found.



> I'm also wondering if any node affinity specified by the user should be applied to the affinity assistant pod as well since it's scheduled first (again, not sure what the right answer is here).

Yes, I totally agree with you, we should set the affinity of podtemplate to affinity-assistant pod.



Sorry, even with the help of google translate I don't know if I describe it clearly! If there is anything I have not described clearly, please point it out to me. Thanks!

lbernick:
> If I set affinity in pipelinerun's podtemplate I think the affinity priority in podtemplate should be the highest and it should not be overwritten by anything. On the other hand, I think affinity-assistant should only serve as an aid, helping to find better node but not change the way to find a better node.
> 
> And if the user applies a pod affinity that isn't really compatible with the affinity assistant the result is that pod will not be scheduled and I guess it should be correct. In this case, the user can disable affinity-assistant or change the affinity in podtemplate to ensure a better node can be found.

Ok, this makes sense to me.

> Yes, I totally agree with you, we should set the affinity of podtemplate to affinity-assistant pod.

Ok! The affinity assistant statefulset code lives [here](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/affinity_assistant.go).

yuzp1996:
> > If I set affinity in pipelinerun's podtemplate I think the affinity priority in podtemplate should be the highest and it should not be overwritten by anything. On the other hand, I think affinity-assistant should only serve as an aid, helping to find better node but not change the way to find a better node.
> > And if the user applies a pod affinity that isn't really compatible with the affinity assistant the result is that pod will not be scheduled and I guess it should be correct. In this case, the user can disable affinity-assistant or change the affinity in podtemplate to ensure a better node can be found.
> 
> Ok, this makes sense to me.
> 
> > Yes, I totally agree with you, we should set the affinity of podtemplate to affinity-assistant pod.
> 
> Ok! The affinity assistant statefulset code lives [here](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/affinity_assistant.go).

Hi @lbernick, I have added the logic of changing the affinity of affinity-assistant pod and some unit tests.

pkg/internal/affinityassistant/transformer.go
```diff
@@ -28,32 +28,42 @@ import (
...
-		// is mutually exclusive with other affinity on taskRun pods. If other
-		// affinity is wanted, that should be added on the Affinity Assistant pod unless
-		// assistant is disabled. When Affinity Assistant is disabled, an affinityAssistantName is not set.
+		// Using node affinity on taskRuns sharing PVC workspace.  When Affinity Assistant
+		// is disabled, an affinityAssistantName is not set.
 		if affinityAssistantName := annotations[workspace.AnnotationAffinityAssistantName]; affinityAssistantName != "" {
-			p.Spec.Affinity = nodeAffinityUsingAffinityAssistant(affinityAssistantName)
+			if p.Spec.Affinity == nil {
+				p.Spec.Affinity = &corev1.Affinity{}
+			}
```
lbernick, line 36: I think this block is duplicated with the similar block in `mergeAffinityWithAffinityAssistant`-- we don't need both

pkg/internal/affinityassistant/transformer.go
```diff
@@ -28,32 +28,42 @@ import (
...
+	if affinity.PodAffinity == nil {
+		affinity.PodAffinity = &corev1.PodAffinity{}
+	}
+
+	affinity.PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution =
+		append(affinity.PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution, *podAffinityTerm)
+}
+
+// podAffinityTermUsingAffinityAssistant achieves pod Affinity term for taskRun
+// pods so that taskRuns is scheduled to the Node were the Affinity Assistant pod
```
lbernick, line 58:
```suggestion
// pods so that the taskRun is scheduled to the Node where the Affinity Assistant pod
```

pkg/internal/affinityassistant/transformer_test.go
```diff
@@ -88,3 +88,143 @@ func TestNewTransformer(t *testing.T) {
...
+							workspace.LabelInstance:  "baz",
+							workspace.LabelComponent: workspace.ComponentNameAffinityAssistant,
+						},
+					},
+					TopologyKey: "kubernetes.io/hostname",
+				}},
+			},
+		},
+	}, {
+		description: "affinity annotation with a different name and pod contains podAffinity",
```
lbernick, line 164: it might also be useful to have a test case for pod affinity with preferredduringschedulingignoredduringexecution

lbernick (approved):

vdemeester (approved):

yuzp1996:
@vdemeester  sorry I force push because of rebasing main branch but cause removing lgtm. Could you please take another look again when you are fine? Thanks!

vdemeester:
@yuzp1996 no worries, both would have remove the lgtm 😉 
/lgtm

PR #5307, opened 2022-08-11
support parameterization for `steps[].onError`

pritidesai:
# Changes

In addition to the static values supported (`continue` and `stopAndFail`) for `onError`, allow specifying these valid values as a task parameter for example, `onError: $(params.CONTINUE)`. Also, validate the specified value is supported otherwise return an error for an invalid value.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Support variables in steps[].onError, for example, $(params.CONTINUE)
```

/kind feature

### DIFF
--- a/pkg/apis/pipeline/v1beta1/task_validation.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation.go
@@ -242,8 +242,9 @@ func validateStep(ctx context.Context, s Step, names sets.String) (errs *apis.Fi
 		}
 	}
 
+	// validate static values in onError if specified - onError can only be set to continue or stopAndFail
 	if s.OnError != "" {
-		if s.OnError != "continue" && s.OnError != "stopAndFail" {
+		if !isParamRefs(s.OnError) && s.OnError != "continue" && s.OnError != "stopAndFail" {
 			errs = errs.Also(&apis.FieldError{
 				Message: fmt.Sprintf("invalid value: %v", s.OnError),
 				Paths:   []string{"onError"},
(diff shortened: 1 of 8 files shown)


pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -242,16 +242,6 @@ func validateStep(ctx context.Context, s Step, names sets.String) (errs *apis.Fi
...
-	if s.OnError != "" {
-		if s.OnError != "continue" && s.OnError != "stopAndFail" {
-			errs = errs.Also(&apis.FieldError{
-				Message: fmt.Sprintf("invalid value: %v", s.OnError),
-				Paths:   []string{"onError"},
-				Details: "Task step onError must be either continue or stopAndFail",
-			})
-		}
-	}
-
```
afrittoli, line 254: NIT: I think we could still do an early validation is no variables are used in `onError` - and keep the test below?

afrittoli (commented):
Thank you for this!
It's a nice feature to have, it allows running a task/pipeline in different modes depending on the environment.
The only NIT I have is that it would still be possible to have early validation of the values when no variables are used, but I'll leave it up to you if you want to add it back.

/approve

afrittoli:
The plumbing change that broke the build test has been reverted now, but there is still one issue left:

```
=========================================
==== TESTING IF GOLINT HAS BEEN DONE ====
=========================================
pkg/pod/entrypoint.go:142:19: errorf: should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...) (revive)
						return nil, errors.New(fmt.Sprintf("task step onError must be either %s or %s but it is set to an invalid value %s",
							entrypoint.ContinueOnError, entrypoint.FailOnError, taskSpec.Steps[i].OnError))
```

pkg/apis/pipeline/v1beta1/task_validation.go
pritidesai: sure sounds good, I will add them back

PR #5313, opened 2022-08-11
Check and/or add to sent cloud event cache in one operation

### LINKED ISSUE #5160: flake: Cloud events unit tests
Both `TestReconcile_CloudEvents` in `pkg/reconciler/pipelinerun/pipelinerun_test.go` with:
```
{Failed      pipelinerun_test.go:6018: config maps: [&ConfigMap{ObjectMeta:{config-defaults  knative-testing    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},Data:map[string]string{default-cloud-events-sink: http://synk:8080,},BinaryData:map[string][]byte{},Immutable:nil,}]
          }
          }}
```

...and `TestReconcile_CloudEvents/Run_with_no_condition` in `pkg/reconciler/run/run_test.go` with:
```
{Failed      run_test.go:246: error in test Run with no condition: timed out waiting for 1 more events: []string{"(?s)dev.tekton.event.run.started.v1.*test-run"}
    run_test.go:255: error in test Run with no condition: extra event received: "Context Attributes,\n  specversion: 1.0\n  type: dev.tekton.event.run.started.v1\n  source: /apis///namespaces/foo//test-run\n  subject: test-run\n  id: 3d7cd7c6-777f-466d-9c0e-99bf34c99a3e\n  datacontenttype: application/json\nData,\n  {\n    \"run\": {\n      \"metadata\": {\n        \"name\": \"test-run\",\n        \"namespace\": \"foo\",\n        \"resourceVersion\": \"00001\",\n        \"creationTimestamp\": null\n      },\n      \"spec\": {\n        \"serviceAccountName\": \"\"\n      },\n      \"status\": {\n        \"extraFields\": null\n      }\n    }\n  }\n"
INFO	TestReconcile_CloudEvents/Run_with_no_condition	cloudevent/cloud_event_controller.go:157	cloudevent Context Attributes,
  specversion: 1.0
  type: dev.tekton.event.run.started.v1
  source: /apis///namespaces/foo//test-run
  subject: test-run
  id: 318d26ad-b6a3-4944-9ca4-47918e2b9028
  datacontenttype: application/json
Data,
  {
    "run": {
      "metadata": {
        "name": "test-run",
        "namespace": "foo",
        "resourceVersion": "00001",
        "creationTimestamp": null
      },
      "spec": {
        "serviceAccountName": ""
      },
      "status": {
        "extraFields": null
      }
    }
  }
 already sent}
```

...have been failing sporadically in CI lately.

/kind flake

abayer:
# Changes

fixes (hopefully) #5160

At least some of the `TestReconcile_CloudEvents` failures in `pkg/reconciler/run/run_test.go` we've been seeing are due to a race condition around the event cache used to prevent double-sending `Run` events. This is because we check the cache, then send the event, and finally write to the cache, creating a race where an initial attempt to send an event may not get to writing to the cache until after a second attempt has gotten past checking the cache. This fixes that by writing to the cache immediately after checking the cache.

In my test bed, with this fix (and a 50ms sleep added to `CheckEventsOrdered` and `CheckEventsUnordered` to make sure that we leave some leeway for asynchronous event delivery to complete before checking events), `TestReconcile_CloudEvents` went from failing a relatively consistent one out of every ten or so runs to failing 20 times out of 1,800 runs over the weekend. So there's still some flakiness there, but much, much less than there was previously

/kind flake

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/events/cache/cache.go
+++ b/pkg/reconciler/events/cache/cache.go
@@ -31,21 +31,8 @@ type eventData struct {
 	Run *v1alpha1.Run `json:"run,omitempty"`
 }
 
-// AddEventSentToCache adds the particular object to cache marking it as sent
-func AddEventSentToCache(cacheClient *lru.Cache, event *cloudevents.Event) error {
-	if cacheClient == nil {
-		return errors.New("cache client is nil")
-	}
-	eventKey, err := EventKey(event)
-	if err != nil {
-		return err
-	}
-	cacheClient.Add(eventKey, nil)
-	return nil
-}
-
-// IsCloudEventSent checks if the event exists in the cache
-func IsCloudEventSent(cacheClient *lru.Cache, event *cloudevents.Event) (bool, error) {
+// ContainsOrAddCloudEvent checks if the event exists in the cache
+func ContainsOrAddCloudEvent(cacheClient *lru.Cache, event *cloudevents.Event) (bool, error) {
 	if cacheClient == nil {
 		return false, errors.New("cache client is nil")
 	}
@@ -53,7 +40,8 @@ func IsCloudEventSent(cacheClient *lru.Cache, event *cloudevents.Event) (bool, e
 	if err != nil {
 		return false, err
 	}
-	return cacheClient.Contains(eventKey), nil
+	isPresent, _ := cacheClient.ContainsOrAdd(eventKey, nil)
+	return isPresent, nil
 }
 
 // EventKey defines whether an event is considered different from another
--- a/pkg/reconciler/events/cache/cache_test.go
+++ b/pkg/reconciler/events/cache/cache_test.go
@@ -146,8 +146,8 @@ func TestAddCheckEvent(t *testing.T) {
 	for _, tc := range testcases {
 		t.Run(tc.name, func(t *testing.T) {
 			testCache, _ := lru.New(10)
-			AddEventSentToCache(testCache, tc.firstEvent)
-			found, _ := IsCloudEventSent(testCache, tc.secondEvent)
+			_, _ = ContainsOrAddCloudEvent(testCache, tc.firstEvent)
+			found, _ := ContainsOrAddCloudEvent(testCache, tc.secondEvent)
 			if d := cmp.Diff(tc.wantFound, found); d != "" {
 				t.Errorf("Cache check failure %s", diff.PrintWantGot(d))
 			}
--- a/pkg/reconciler/events/cloudevent/cloud_event_controller.go
+++ b/pkg/reconciler/events/cloudevent/cloud_event_controller.go
@@ -149,7 +149,7 @@ func SendCloudEventWithRetries(ctx context.Context, object runtime.Object) error
 		logger.Debugf("Sending cloudevent of type %q", event.Type())
 		// In case of Run event, check cache if cloudevent is already sent
 		if isRun {
-			cloudEventSent, err := cache.IsCloudEventSent(cacheClient, event)
+			cloudEventSent, err := cache.ContainsOrAddCloudEvent(cacheClient, event)
 			if err != nil {
 				logger.Errorf("error while checking cache: %s", err)
 			}
@@ -163,14 +163,9 @@ func SendCloudEventWithRetries(ctx context.Context, object runtime.Object) error
 			recorder := controller.GetEventRecorder(ctx)
 			if recorder == nil {
 				logger.Warnf("No recorder in context, cannot emit error event")
+				return
 			}
 			recorder.Event(object, corev1.EventTypeWarning, "Cloud Event Failure", result.Error())
-		}
-		// In case of Run event, add to the cache to avoid duplicate events
-		if isRun {
-			if err := cache.AddEventSentToCache(cacheClient, event); err != nil {
-				logger.Errorf("error while adding sent event to cache: %s", err)
-			}
 		}
 	}()
 
--- a/test/events/events.go
+++ b/test/events/events.go
@@ -26,6 +26,8 @@ import (
 // in the same order.
 func CheckEventsOrdered(t *testing.T, eventChan chan string, testName string, wantEvents []string) error {
 	t.Helper()
+	// Sleep 50ms to make sure events have delivered
+	time.Sleep(50 * time.Millisecond)
 	err := eventsFromChannel(eventChan, wantEvents)
 	if err != nil {
 		return fmt.Errorf("error in test %s: %v", testName, err)
@@ -37,6 +39,8 @@ func CheckEventsOrdered(t *testing.T, eventChan chan string, testName string, wa
 // were received via the given chan, in any order.
 func CheckEventsUnordered(t *testing.T, eventChan chan string, testName string, wantEvents []string) error {
 	t.Helper()
+	// Sleep 50ms to make sure events have delivered
+	time.Sleep(50 * time.Millisecond)
 	err := eventsFromChannelUnordered(eventChan, wantEvents)
 	if err != nil {
 		return fmt.Errorf("error in test %s: %v", testName, err)


abayer:
Ok, the flake count went down considerably, but I also saw a case of an event being set twice even though it was theoretically cached. Digging into that now...

abayer:
/retest

There's definitely something else wrong with the event cache - I'm seeing two attempts to send an identical event, which it shouldn't be doing. Digging further!

lbernick:
What's the actual race condition here? This seems like it might be a bandaid over most of the test flakes, but is it possible to fix the underlying problem?

abayer:
> What's the actual race condition here? This seems like it might be a bandaid over most of the test flakes, but is it possible to fix the underlying problem?

That's part of what I'm still trying to figure out. If I am right about the retries (which I'm less sure of now!), then the race would be that sometimes, under load, sending the cloud event fails the first time and gets resent 10ms later, which is fine. That's why we have the retries there - the end result is correct, it's just the test scenario that has problems due to expecting all the events to be delivered by the time the `PipelineRun` reconciliation is over...which may not be the case if it had to retry.

But I have now seen at least one case (with a bunch of extra debug logging tossed in to give me better visibility into what's actually happening) of an event getting sent successfully, then when the second reconcile happens, it shouldn't be sending the event again because it's already in the cloud event controller cache, but it is, in fact, being sent again. So that's where I'm poking around in right now.

abayer:
Also, to make it more fun, in my test setup, `TestReconcile_InvalidPipelineRuns` in `pkg/reconciler/pipelinerun/pipelinerun_test.go` is flaking out fairly often, which I've never seen in CI and am assuming is just a weird artifact of the resource-limited setup I'm running in...which, of course, could throw off all my theories around the cloud events tests. Fun!

abayer:
So I've added https://github.com/tektoncd/pipeline/pull/5313/commits/1cdc4fa81100a8c3525069aa5d0caf9bd27f8a36 to this PR - as I detailed in #5160, I've narrowed down the cause of the `run_test` cloud events failures to multiple events for the same `Run` being sent before the first call finishes, so adding a mutex to cloud event dispatch in the `Run` reconciler should solve that. This particular issue is very specific to `Run`s - we only do caching to prevent duplicate event dispatch for `Run`s. This mutex should also deal with [the race condition](https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5134/pull-tekton-pipeline-unit-tests/1557430461020508160) that shows up sometimes in the various cloud event tests in `pkg/reconciler/run/run_test.go`, I believe.

That said, I think the sleep in the original commit here is still valuable for other event test flakiness, such as [`TestSendKubernetesEvents`](https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5312/pull-tekton-pipeline-unit-tests/1557762412231266304), [`TestSendCloudEventWithRetries/test-send-cloud-event-taskrun`](https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5082/pull-tekton-pipeline-unit-tests/1557723601816260608), and [`TestReconcile_CloudEvents` in `taskrun_test.go`](https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5293/pull-tekton-pipeline-unit-tests/1557421293962268672) - these event-related flakes are all cases where events, either k8s events or cloud events, have not yet been received when we go to check them, rather than the `Run`/cloud events flakes, which are caused by a duplicate event being received. So I do still think that a delay should cut down on the "missing event" flakes.

afrittoli:
Thanks for this. I wonder if we should add the mutex further down in the code structure though, because this will hold the entire reconcile cycle, so it will basically force reconciles to not happen in parallel as long as long as cloudevents for run is enabled. It feels like this would slow down things a lot because the event will be taken from the queue and sit there waiting.

Perhaps it would be enough to add locks to the methods in https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/events/cache/cache.go, similar to what the original cache class does?

abayer:
> Thanks for this. I wonder if we should add the mutex further down in the code structure though, because this will hold the entire reconcile cycle, so it will basically force reconciles to not happen in parallel as long as long as cloudevents for run is enabled. It feels like this would slow down things a lot because the event will be taken from the queue and sit there waiting.

Yeah, I don't love this location, but the thing is that we do actually need to lock the whole of `events.EmitCloudEvents`, or at least all of `cloudevent.SendCloudEventWithRetries`. That said, I'm 100% sure there's a better way to do it than what I've done here, I just haven't landed on it yet.

> Perhaps it would be enough to add locks to the methods in https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/events/cache/cache.go, similar to what the original cache class does?

No, that wouldn't work, because it's not that we're having problems because of the cache calls directly. The problem is that https://github.com/tektoncd/pipeline/blob/45b7c378985f0baaab751980511228fdce1e0801/pkg/reconciler/events/cloudevent/cloud_event_controller.go#L148-L174 isn't locked, so two different calls can be somewhere in that block at the same time. What's happening is that call 1 gets past `cache.IsCloudEventSent(...)` and into `ceClient.Send(...)`, and before call 1 gets past that to get to `cache.AddEventSentToCache(...)`, call 2 gets to `cache.IsCloudEventSent(...)`. So really, we want a lock on that `go func()...`.

afrittoli:
> > Thanks for this. I wonder if we should add the mutex further down in the code structure though, because this will hold the entire reconcile cycle, so it will basically force reconciles to not happen in parallel as long as long as cloudevents for run is enabled. It feels like this would slow down things a lot because the event will be taken from the queue and sit there waiting.
> 
> Yeah, I don't love this location, but the thing is that we do actually need to lock the whole of `events.EmitCloudEvents`, or at least all of `cloudevent.SendCloudEventWithRetries`. That said, I'm 100% sure there's a better way to do it than what I've done here, I just haven't landed on it yet.
> 
> > Perhaps it would be enough to add locks to the methods in https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/events/cache/cache.go, similar to what the original cache class does?
> 
> No, that wouldn't work, because it's not that we're having problems because of the cache calls directly. The problem is that
> 
> https://github.com/tektoncd/pipeline/blob/45b7c378985f0baaab751980511228fdce1e0801/pkg/reconciler/events/cloudevent/cloud_event_controller.go#L148-L174
> 
> isn't locked, so two different calls can be somewhere in that block at the same time. What's happening is that call 1 gets past `cache.IsCloudEventSent(...)` and into `ceClient.Send(...)`, and before call 1 gets past that to get to `cache.AddEventSentToCache(...)`, call 2 gets to `cache.IsCloudEventSent(...)`. So really, we want a lock on that `go func()...`.

So, I'm thinking that we should change this function to add the event to the cache *before* we invoke the actual send function, since that might do retries and take an arbitrary amount of time to complete.

I would propose to:
- Change the `AddEventSentToCache` function to, lock, read, write, unlock, and return whether the event was already there
- Call the `AddEventSentToCache` where the `IsCloudEventSent` is now, and use the result to decide whether to send the event or not

abayer:
Yeah, that all sounds good to me. Lemme do that.

vdemeester (approved):

abayer:
ping @afrittoli @jerop @lbernick =)

pkg/reconciler/events/cache/cache.go
```diff
@@ -31,19 +31,6 @@ type eventData struct {
...
-	eventKey, err := EventKey(event)
-	if err != nil {
-		return err
-	}
-	cacheClient.Add(eventKey, nil)
-	return nil
-}
-
 // IsCloudEventSent checks if the event exists in the cache
 func IsCloudEventSent(cacheClient *lru.Cache, event *cloudevents.Event) (bool, error) {
```
jerop, line 35: looks like we need to update this function name and description 

```suggestion
// ContainsOrAddCloudEvent checks if the event exists in the cache, and if not, adds the event
func ContainsOrAddCloudEvent(cacheClient *lru.Cache, event *cloudevents.Event) (bool, error) {

abayer: Whoops! Doing so now.

jerop (commented):
/lgtm 

thanks @abayer!

PR #5322, opened 2022-08-15
Convert Step OnError from string to "OnErrorType" type

QuanZhang-William:
# Changes
Prior to this commit, the ```step.OnError``` field is defined as string and the 2 supported string values "continue" and "stopAndFail" are directly used across the codebase. This is error-prone and it introduces maintenance difficulty. This commit updates the ```OnError``` field to be typed string ```OnErrorType```, with constants defined for the 2 supported values, and updates all the related references

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Convert step.OnError from string to type: OnErrorType
```

### DIFF
--- a/pkg/apis/pipeline/v1/container_types.go
+++ b/pkg/apis/pipeline/v1/container_types.go
@@ -129,9 +129,7 @@ type Step struct {
 
 	// OnError defines the exiting behavior of a container on error
 	// can be set to [ continue | stopAndFail ]
-	// stopAndFail indicates exit the taskRun if the container exits with non-zero exit code
-	// continue indicates continue executing the rest of the steps irrespective of the container exit code
-	OnError string `json:"onError,omitempty"`
+	OnError OnErrorType `json:"onError,omitempty"`
 	// Stores configuration for the stdout stream of the step.
 	// +optional
 	StdoutConfig *StepOutputConfig `json:"stdoutConfig,omitempty"`
@@ -139,6 +137,16 @@ type Step struct {
 	// +optional
 	StderrConfig *StepOutputConfig `json:"stderrConfig,omitempty"`
 }
+
+// OnErrorType defines a list of supported exiting behavior of a container on error
+type OnErrorType string
+
+const (
+	// StopAndFail indicates exit the taskRun if the container exits with non-zero exit code
+	StopAndFail OnErrorType = "stopAndFail"
+	// Continue indicates continue executing the rest of the steps irrespective of the container exit code
+	Continue OnErrorType = "continue"
+)
 
 // StepOutputConfig stores configuration for a step output stream.
 type StepOutputConfig struct {
--- a/pkg/apis/pipeline/v1beta1/container_types.go
+++ b/pkg/apis/pipeline/v1beta1/container_types.go
@@ -188,9 +188,7 @@ type Step struct {
 
 	// OnError defines the exiting behavior of a container on error
 	// can be set to [ continue | stopAndFail ]
-	// stopAndFail indicates exit the taskRun if the container exits with non-zero exit code
-	// continue indicates continue executing the rest of the steps irrespective of the container exit code
-	OnError string `json:"onError,omitempty"`
+	OnError OnErrorType `json:"onError,omitempty"`
 
 	// Stores configuration for the stdout stream of the step.
 	// +optional
@@ -199,6 +197,16 @@ type Step struct {
 	// +optional
 	StderrConfig *StepOutputConfig `json:"stderrConfig,omitempty"`
 }
+
+// OnErrorType defines a list of supported exiting behavior of a container on error
+type OnErrorType string
+
+const (
+	// StopAndFail indicates exit the taskRun if the container exits with non-zero exit code
+	StopAndFail OnErrorType = "stopAndFail"
+	// Continue indicates continue executing the rest of the steps irrespective of the container exit code
+	Continue OnErrorType = "continue"
+)
 
 // StepOutputConfig stores configuration for a step output stream.
 type StepOutputConfig struct {
--- a/docs/pipeline-api.md
+++ b/docs/pipeline-api.md
@@ -445,6 +445,117 @@ TaskSpec
 </tr>
 </tbody>
 </table>
+<h3 id="tekton.dev/v1.EmbeddedTask">EmbeddedTask
+</h3>
+<p>
+(<em>Appears on:</em><a href="#tekton.dev/v1.PipelineTask">PipelineTask</a>)
+</p>
+<div>
+<p>EmbeddedTask is used to define a Task inline within a Pipeline&rsquo;s PipelineTasks.</p>
+</div>
+<table>
+<thead>
+<tr>
+<th>Field</th>
+<th>Description</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+<td>
+<code>spec</code><br/>
+<em>
+k8s.io/apimachinery/pkg/runtime.RawExtension
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>Spec is a specification of a custom task</p>
+<br/>
+<br/>
+<table>
+<tr>
+<td>
+<code>-</code><br/>
+<em>
+[]byte
+</em>
+</td>
+<td>
+<p>Raw is the underlying serialization of this object.</p>
+<p>TODO: Determine how to detect ContentType and ContentEncoding of &lsquo;Raw&rsquo; data.</p>
+</td>
+</tr>
+<tr>
+<td>
+<code>-</code><br/>
+<em>
+k8s.io/apimachinery/pkg/runtime.Object
+</em>
+</td>
+<td>
+<p>Object can hold a representation of this extension - useful for working with versioned
+structs.</p>
+</td>
+</tr>
+</table>
+</td>
+</tr>
+<tr>
+<td>
+<code>metadata</code><br/>
+<em>
+<a href="#tekton.dev/v1.PipelineTaskMetadata">
+PipelineTaskMetadata
+</a>
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+</td>
+</tr>
+<tr>
+<td>
+<code>TaskSpec</code><br/>
+<em>
+<a href="#tekton.dev/v1.TaskSpec">
+TaskSpec
+</a>
+</em>
+</td>
+<td>
+<p>
+(Members of <code>TaskSpec</code> are embedded into this type.)
+</p>
+<em>(Optional)</em>
+<p>TaskSpec is a specification of a task</p>
+</td>
+</tr>
+</tbody>
+</table>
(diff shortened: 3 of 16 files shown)


pkg/apis/pipeline/v1beta1/container_types.go
```diff
@@ -200,6 +200,14 @@ type Step struct {
 	StderrConfig *StepOutputConfig `json:"stderrConfig,omitempty"`
 }
 
+// OnErrorType defines a list of supported exiting behavior of a container on error
+type OnErrorType string
+
```
JeromeJu, line 205: Wondering if there's specific reason that we need to change from using the reference in the pkg/entrypoint/engrypointer.go:
```
	ContinueOnError = "continue"
	FailOnError     = "stopAndFail"
)
```

QuanZhang-William: In general we prefer a unified way to define the OnErrorType to reduce maintenance cost. And we can benefit from the compile time check by the typed string.

It's also a bit weird to me to add the entrypoint package as an extra dependency to v1 and v1beta1 packages just to refer to a constant.

pkg/entrypoint/entrypointer.go
```diff
@@ -77,7 +75,7 @@ type Entrypointer struct {
 	// OnError defines exiting behavior of the entrypoint
 	// set it to "stopAndFail" to indicate the entrypoint to exit the taskRun if the container exits with non zero exit code
 	// set it to "continue" to indicate the entrypoint to continue executing the rest of the steps irrespective of the container exit code
-	OnError string
+	OnError v1beta1.OnErrorType
```
wlynch, line 78: I'd keep the entrypoint types decoupled from the API types as much as possible - the entrypoint needs to support a superset of states across all versions. e.g. if there's a new OnError state in a v1alpha2 API, the entrypoint will need to add it (and it'll be easier to support if we don't have to add it to stable APIs and deal with potential deprecation policies).

wlynch (changes requested):
Mostly LGTM! Just one change I think we should make.

pkg/entrypoint/entrypointer.go
QuanZhang-William: Yep, this makes sense! I have converted converted the entrypoint types back. Please take a look!

wlynch (approved):
/lgtm

Yongxuanzhang (commented):
Great work! 
Maybe adding a release note?

QuanZhang-William:
> Great work! Maybe adding a release note?

My understanding of a Release Note is to notify users about a behaviour change. There is actually no change from the user's perspective and there is nothing to "notify" (since we already have the OnError input validation before this PR). This is more like a "clean up " work. I don't think a Release Note is necessary in this case?

wlynch:
> My understanding of a Release Note is to notify users about a behaviour change. There is actually no change from the user's perspective and there is nothing to "notify" (since we already have the OnError input validation before this PR). This is more like a "clean up " work. I don't think a Release Note is necessary in this case?

Good catch @Yongxuanzhang! - This should be called out as a breaking change to the API - this will cause build failures for clients that were previously using the string based type, i.e.

```go
s := "asdf"
v1beta1.Step{
	OnError: s,
}
```

pkg/apis/pipeline/v1/container_types.go
```diff
@@ -129,9 +129,7 @@ type Step struct {
 
 	// OnError defines the exiting behavior of a container on error
 	// can be set to [ continue | stopAndFail ]
-	// stopAndFail indicates exit the taskRun if the container exits with non-zero exit code
-	// continue indicates continue executing the rest of the steps irrespective of the container exit code
-	OnError string `json:"onError,omitempty"`
+	OnError OnErrorType `json:"onError,omitempty"`
```
pritidesai, line 132: we added support for variables in addition to the two constants - https://github.com/tektoncd/pipeline/pull/5307

```
steps[].onError: $(params.CONTINUE)
steps[].onError: continue
steps[].onError: stopAndFail
```

QuanZhang-William:
> > My understanding of a Release Note is to notify users about a behaviour change. There is actually no change from the user's perspective and there is nothing to "notify" (since we already have the OnError input validation before this PR). This is more like a "clean up " work. I don't think a Release Note is necessary in this case?
> 
> Good catch @Yongxuanzhang! - This should be called out as a breaking change to the API - this will cause build failures for clients if they were previously using the string based type, i.e.
> 
> ```go
> s := "asdf"
> v1beta1.Step{
> 	OnError: s,
> }
> ```

Ohh, got what you mean. I didn't think about that Tekton can be used in this way programmatically. Will update the Release Note in this case

pkg/apis/pipeline/v1/container_types.go
QuanZhang-William: Hi @pritidesai , we will eventually parse the variables to be actual values, and we have validation for the variable scenario. Is there any concern with it?

QuanZhang-William: Hi @pritidesai, thanks for letting me know. Was wondering if there is any concern when using variables with "OnErrorType". We will eventually parse the variable to a real value here.

pritidesai: How?

I haven't reviewed the code changes in this PR. I am not sure how can it be restricted to the type of `enum` (`continue` and `stopAndFail`) when it can have a parameter of type `string`.

QuanZhang-William: This is also what I found interesting that ```enum``` is not that restricted in Go, even with a Typed string ```OnErrorType``` you can still do something like

```
step.OnError = "this is undefined value in enum OnErrorType"
```

and there is no compilation or runtime error. So there is essentially no difference between a regular string or a typed string ```OnErrorType```when calling [ApplyReplacement function](https://github.com/tektoncd/pipeline/blob/7ba5090c37ddc09be4c7ff62cc49a2428af22dc7/pkg/container/step_replacements.go#L27) other than some type casting.

pritidesai: thank you @QuanZhang-William for all the efforts but declaring `onError` to as `enum` but allowing it to specify variables sounds very confusing. If we are declaring a field as enum, it will be simpler and easier to follow if it's restricted to the specified enum values. 

I suggest keeping the type to `string`, thanks!

vdemeester:
> Ohh, got what you mean. I didn't think about that Tekton can be used in this way programmatically. Will update the Release Note in this case

Mainly the `cli` or go client might be affected 🙃

pkg/apis/pipeline/v1/container_types.go
vdemeester: @pritidesai I am not sure I follow here ? The `type` is still `string`, as `OnErrorType` is still just an alias to a string. There is no concept of `enum` in Go anyway 🙃. Also, this is the exact same thing with *a lot* of other types we are using (in status, …), it is mainly useful for developer and consumer of the go library to use a predefined set of values, it doesn't do much more than this.

vdemeester (approved):

pritidesai:
> @pritidesai I am not sure I follow here ? The type is still string, as OnErrorType is still just an alias to a string. There is no concept of enum in Go anyway 🙃. Also, this is the exact same thing with a lot of other types we are using (in status, …), it is mainly useful for developer and consumer of the go library to use a predefined set of values, it doesn't do much more than this.

I was a little surprised to discover this was merged while going over the release notes for 0.39 🤔 

`step.onError` was recently changed to support `params` in addition to constants. `step.onError` is something user specified through `params` unlike other types in `status` which are mostly set by the controller or specified as static values by the users (unless that's not true) in their `YAML` specifications.

`OnErrorType` represents a false impression that it only accepts static values (`continue` and `stopAndFail`). Why unnecessary creating such type and then converting it back to `string` in many places?

vdemeester:
> `OnErrorType` represents a false impression that it only accepts static values (`continue` and `stopAndFail`). Why unnecessary creating such type and then converting it back to `string` in many places?

From a API consumer point of view, nothing changes, it's still "just a string" nothing else. 99% of users of Tekton won't read code, just docs, and `OnErrorType` doesn't bubble up anywhere in the API or the doc. This is a pure "implementation detail" change that can help developer code overall when working on the controller's code.

pritidesai:
thank you @vdemeester for being patient here, appreciate it 🙏 

I don't know how are these interpreted - https://tekton.dev/docs/pipelines/pipeline-api/#tekton.dev/v1.Step  This API doc has listed `onError` as `OnErrorType`.

The source for this doc is here - https://github.com/tektoncd/pipeline/blob/main/docs/pipeline-api.md#tekton.dev/v1.Step

QuanZhang-William:
> thank you @vdemeester for being patient here, appreciate it 🙏
> 
> I don't know how are these interpreted - https://tekton.dev/docs/pipelines/pipeline-api/#tekton.dev/v1.Step This API doc has listed `onError` as `OnErrorType`.
> 
> The source for this doc is here - https://github.com/tektoncd/pipeline/blob/main/docs/pipeline-api.md#tekton.dev/v1.Step

Thanks for your feedback @pritidesai! The ```OnErrorType``` is documented as ```(string Alias)``` here: https://tekton.dev/docs/pipelines/pipeline-api/#tekton.dev/v1.OnErrorType. I think the question is essentially how does a user interpret a ```string alias``` in GO. My understanding is that a ```string alias``` is not a ```enum``` and there is no real ```enum``` in GO: https://github.com/golang/go/issues/19814. 

The string value "continue" and "stopAndFail" are being directly used across the codebase and it is in general not a good practice. We can also benefit a better compile time check from such typed string, for example, ```pipelineTask.OnError = step.OnError``` (after [TEP-0050](https://github.com/tektoncd/community/pull/785/files)) cannot happen as they are different things semantically.

I think we can keep the ```OnErrorType``` for a better practice. And I can update the API doc explicitly saying something like:  ```The value can be set dynamically from task parameter for example, onError: $(params. CONTINUE)``` to resolve user's confusion if any. What do you think?

/cc @jerop, @dibyom

dibyom:
> I think we can keep the OnErrorType for a better practice. And I can update the API doc explicitly saying something like: The value can be set dynamically from task parameter for example, onError: $(params. CONTINUE) to resolve user's confusion if any. What do you think?

https://tekton.dev/docs/pipelines/variables/#fields-that-accept-variable-substitutions already lists `onError` - I think that's sufficient? (There are other non "string" string types that also accept variable substitutions e.g. `imagePullPolicy`)

QuanZhang-William:
> > I think we can keep the OnErrorType for a better practice. And I can update the API doc explicitly saying something like: The value can be set dynamically from task parameter for example, onError: $(params. CONTINUE) to resolve user's confusion if any. What do you think?
> 
> https://tekton.dev/docs/pipelines/variables/#fields-that-accept-variable-substitutions already lists `onError` - I think that's sufficient? (There are other non "string" string types that also accept variable substitutions e.g. `imagePullPolicy`)

Yeah, I agree it should be sufficient in this case :)

PR #5325, opened 2022-08-15
Add "beta" value to `enable-api-fields`

lbernick:
# Changes

Add "beta" value to `enable-api-fields`
This commit continues the implementation plan described in
[TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md)
by adding a "beta" option to the "enable-api-fields" feature flag.

Features moving from "alpha" to "beta" stability should be gated behind the "beta" value of
this feature flag in v1 until they are ready to move to a "stable" stability level.
Maintainers can make exceptions to this policy on a case-by-case basis.
While the Tekton community may decide to make changes to the API versioning process,
features should follow this "graduation" process until the new process, if any,
is reflected in a TEP.

No functional changes.
/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -23,6 +23,7 @@ This guide explains how to install Tekton Pipelines. It covers the following top
 - [Customizing basic execution parameters](#customizing-basic-execution-parameters)
     - [Customizing the Pipelines Controller behavior](#customizing-the-pipelines-controller-behavior)
     - [Alpha Features](#alpha-features)
+    - [Beta Features](#beta-features)
 - [Configuring High Availability](#configuring-high-availability)
 - [Configuring tekton pipeline controller performance](#configuring-tekton-pipeline-controller-performance)
 - [Creating a custom release of Tekton Pipelines](#creating-a-custom-release-of-tekton-pipelines)
@@ -450,6 +451,7 @@ Alpha features are still in development and their syntax is subject to change.
 To enable these, set the `enable-api-fields` feature flag to `"alpha"` in
 the `feature-flags` ConfigMap alongside your Tekton Pipelines deployment via
 `kubectl patch cm feature-flags -n tekton-pipelines -p '{"data":{"enable-api-fields":"alpha"}}'`.
+Setting `enable-api-fields` to "alpha" also enables [beta features](#beta-features).
 
 Features currently in "alpha" are:
 
@@ -471,6 +473,15 @@ Features currently in "alpha" are:
 | [CSI Workspace Type](workspaces.md#csi)                                                               | N/A                                                                                                                        | [v0.38.0](https://github.com/tektoncd/pipeline/releases/tag/v0.38.0) |                             |
 | [Object Params and Results](pipelineruns.md#specifying-parameters)                                    | [TEP-0075](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md)                     | [v0.38.0](https://github.com/tektoncd/pipeline/releases/tag/v0.38.0) |                             |
 | [Array Results](pipelineruns.md#specifying-parameters)                                                | [TEP-0076](https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md)                                | [v0.38.0](https://github.com/tektoncd/pipeline/releases/tag/v0.38.0) |                             |
+
+### Beta Features
+
+Beta features are fields of stable CRDs that follow our "beta" [compatibility policy](../api_compatibility_policy.md).
+To enable these features, set the `enable-api-fields` feature flag to `"beta"` in
+the `feature-flags` ConfigMap alongside your Tekton Pipelines deployment via
+`kubectl patch cm feature-flags -n tekton-pipelines -p '{"data":{"enable-api-fields":"beta"}}'`.
+
+For beta versions of Tekton CRDs, setting `enable-api-fields` to "beta" is the same as setting it to "stable".
 
 ## Configuring High Availability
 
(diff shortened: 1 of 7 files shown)


docs/install.md
```diff
@@ -439,6 +446,23 @@ Features currently in "alpha" are:
...
+
+Beta features are fields of stable CRDs that follow our "beta" [compatibility policy](../api_compatibility_policy.md).
+To enable these features, set the `enable-api-fields` feature flag to `"beta"` in
+the `feature-flags` ConfigMap alongside your Tekton Pipelines deployment:
+
+```sh
+kubectl edit configmap feature-flags -n tekton-pipelines
+```
+
+For beta versions of Tekton CRDs, setting `enable-api-fields` to "beta" is the same as setting it to "stable".
```
lbernick, line 459: One thing that might be confusing here is that these docs don't distinguish between fields that are in beta because they're newly introduced, and fields that are in beta because they were in v1beta1 but we don't want to stabilize them. Open to suggestions on how best to document this.

pkg/apis/config/feature_flags.go
```diff
@@ -196,10 +198,19 @@ func NewFeatureFlagsFromConfigMap(config *corev1.ConfigMap) (*FeatureFlags, erro
...
 }
 
-// EnableAlphaAPIFields enables alpha feature in an existing context (for use in testing)
+// EnableAlphaAPIFields enables alpha features in an existing context (for use in testing)
 func EnableAlphaAPIFields(ctx context.Context) context.Context {
+	return setEnableAPIFields(ctx, "alpha")
+}
+
+// EnableBetaAPIFields enables beta features in an existing context (for use in testing)
+func EnableBetaAPIFields(ctx context.Context) context.Context {
```
JeromeJu, line 207: nit: wondering if we need to include this in the test? As noticed the coverage off.

JeromeJu:
LGTM
Thank you Lee.

vdemeester (commented):
I am not really sure I understand that change. I think in [TEP-0119](https://github.com/tektoncd/community/pull/786/files), we are proposing to deprecate/remove this field for v1 right ? If that's the case, then it shouldn't be there in the `v1` types, at all.

The more we use the `enable-api-fields` feature flag, the more confusing it is getting for me. Aside from the all-or-nothing behavior slight problem, my feeling is that we are abusing the feature-flag for a lot of things.
Shouldn't we make `v1` with the field we only want to support instead ? There is going to be some time when we will (**have to**) support `v1` and `v1beta1` and thus users will still be able to use `step.resources` from `v1beta1` object. Having this "it's in the API (so documented in openapi "possibly", …) but only when the `beta` feature-flag is enable is very confusing to me. It's even more confusing to me as we *aim* to get read of it, but as an end-user, I am mostly assuming what would be under the `beta` feature-flag is on the way to be "promoted".

lbernick:
> The more we use the `enable-api-fields` feature flag, the more confusing it is getting for me. Aside from the all-or-nothing behavior slight problem, my feeling is that we are abusing the feature-flag for a lot of things. Shouldn't we make `v1` with the field we only want to support instead ? There is going to be some time when we will (**have to**) support `v1` and `v1beta1` and thus users will still be able to use `step.resources` from `v1beta1` object. Having this "it's in the API (so documented in openapi "possibly", …) but only when the `beta` feature-flag is enable is very confusing to me. It's even more confusing to me as we _aim_ to get read of it, but as an end-user, I am mostly assuming what would be under the `beta` feature-flag is on the way to be "promoted".

TBH I agree, I would also assume that anything in beta is on its way to be promoted rather than deprecated. Maybe we should reopen discussion around maintaining multiple api versions rather than the enable-api-fields feature flag? @dibyom 

For this PR, would it be helpful to have it just add the beta flag and not use it for any fields right now? (especially since compute resources are still under discussion) Or does it not make sense to add dead code?

vdemeester:
> TBH I agree, I would also assume that anything in beta is on its way to be promoted rather than deprecated. Maybe we should reopen discussion around maintaining multiple api versions rather than the enable-api-fields feature flag? @dibyom

I am not sure I understand the "reopen discussion around maintaining multiple api versions". We are/will maintain multiple api versions, this is a *given*. For a given period of time (according to API policies, …), we will support and serve `v1beta1` as well as `v1`. Which one is stored, and how the controller does its magic is an implementation detail, but from a user perspective, we will support both version, we **have to**.

> For this PR, would it be helpful to have it just add the beta flag and not use it for any fields right now? (especially since compute resources are still under discussion) Or does it not make sense to add dead code?

It kinds of makes it dead code, but I think it's fine though, as it provides the "framework" for `beta` fields to be added / promoted later one.

vdemeester:
> Sorry, I should have been more clear-- I meant instead of adding features gated behind an alpha feature flag to a beta CRD, creating an alpha CRD and adding features there, and later on adding them to the beta CRD. This discussion is probably out of scope for this PR though.

Right, now I understand it a bit more 🙃. Indeed this is a bit out of the scope of this PR, and it comes with its "set" of challenges.

pkg/apis/config/feature_flags.go
lbernick: Done.

lbernick:
@dibyom mentioned wanting to discuss potential updates to TEP 33 (whether there is value in a "beta" feature flag at all as opposed to alpha and stable) before merging this PR, so I've added it to the next api wg agenda and I think we can put it on hold for now. FYI @vdemeester I think he might have chatted with you about this as well.

/hold

lbernick:
/hold cancel

discussed at api wg-- don't want to block moving some existing features to beta on the larger discussion about better api versioning, so I'm reopening this pr

jerop (approved):
@lbernick please add more context/details to the commit message, based on the discussions in this PR and the API WG today

lbernick:
Thanks for the feedback @jerop, done!

PR #5331, opened 2022-08-16
V1: Add conversion for Pipeline.Resources

JeromeJu:
This commit adds support for Pipeline.Resources when converting between v1beta1 and v1
versions of Pipelines. This will allow us to release v1 Pipeline in a backwards compatible way,
by ensuring that v1beta1 Pipelines with Resources will have the Resources serialized into
annotations on the v1 Pipeline on conversion.

# Changes
This commit adds support for Pipeline.Resources when converting between v1beta1 and v1
versions of Pipelines. This will allow us to release v1 Pipeline in a backwards compatible way,
by ensuring that v1beta1 Pipelines with Resources will have the Resources serialized into
annotations on the v1 Pipeline on conversion.

Part of #4986 
/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/pkg/apis/version/conversion_test.go
+++ b/pkg/apis/version/conversion_test.go
@@ -52,3 +52,28 @@ func TestSerializationRoundTrip(t *testing.T) {
 		t.Errorf("Unexpected diff after serialization/deserialization round trip: %s", d)
 	}
 }
+
+func TestSliceSerializationRoundTrip(t *testing.T) {
+	meta := metav1.ObjectMeta{}
+	source := []testStruct{{Field: "foo"}, {Field: "bar"}}
+	key := "my-key"
+	err := version.SerializeToMetadata(&meta, source, key)
+	if err != nil {
+		t.Fatalf("Serialization error: %s", err)
+	}
+
+	sink := []testStruct{}
+	err = version.DeserializeFromMetadata(&meta, &sink, key)
+	if err != nil {
+		t.Fatalf("Deserialization error: %s", err)
+	}
+
+	_, ok := meta.Annotations[key]
+	if ok {
+		t.Errorf("Expected key %s not to be present in annotations but it was", key)
+	}
+
+	if d := cmp.Diff(source, sink); d != "" {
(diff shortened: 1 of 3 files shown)


pkg/apis/version/conversion_test.go
```diff
@@ -52,3 +52,28 @@ func TestSerializationRoundTrip(t *testing.T) {
 		t.Errorf("Unexpected diff after serialization/deserialization round trip: %s", d)
 	}
 }
+
+func TestArraySerializationRoundTrip(t *testing.T) {
```
lbernick, line 56: can this be a test case in the existing test class instead?

lbernick (approved):

pkg/apis/version/conversion_test.go
```diff
@@ -29,26 +30,51 @@ type testStruct struct {
...
+		source: []testStruct{{Field: "foo"}, {Field: "bar"}},
+		key:    "my-key",
+	}}
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			err := version.SerializeToMetadata(&tc.meta, tc.source, tc.key)
+			if err != nil {
+				t.Fatalf("Serialization error: %s", err)
+			}
+			var sink interface{}
```
JeromeJu, line 57: Wondering if there's any better way to change the variable type here?

lbernick: ah sorry I didn't realize this was why you created two separate test classes. You could put the sink as part of the test case?

Also I think lines 60 and 74 can be removed.

JeromeJu: It doesn't seem to work fine if I put the test case as 
```
	testcases := []struct {
		name   string
		meta   metav1.ObjectMeta
		sink   interface{}
		source interface{}
		key    string
	}```
The result would be a map instead of the struct.

jerop (commented):
/lgtm

PR #5332, opened 2022-08-16
TEP-0114: Custom Task E2E Test with A Controller Installed

XinruZhang:
This PR introduces a `wait-task` controller into test/ folder to
better support Custom Task e2e testing.

Previously we update the Run object to mimic the behavior of a custom
task controller. This pr introduces a real controller to reconcile
the custom task run.

# Changes

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```


test/wait-task/tekton/README.md
```diff
@@ -0,0 +1,32 @@
+# Tekton Resources
+
+This folder contains tekton resources used build and release the wait custom task.
+The resources included in `kustomization.yaml` are installed nightly on the `dogfooding`
+cluster, and used to run nightly releases.
```
jerop, line 5: as @dibyom mentioned, we should remove this folder that contains resources to make releases - this is needed in the experimental wait task that actually makes releases

test/wait-task/README.md
```diff
@@ -0,0 +1,73 @@
...
+## Install
+
+Install and configure `ko`.
+
+```
+ko apply -f controller.yaml
+```
+
+This will build and install the controller on your cluster, in the namespace
+`wait-task`.
```
jerop, line 34: please update this readme, this custom task is meant for testing only -- not for users to install and use

test/wait-task/tekton/README.md
XinruZhang: Oh thx for the comment! I may need some clarification here, do you mean remove the entire wait-test folder under test/ with only one yaml file being used to install the wait-task?

XinruZhang: NVM, removed the tekton folder :) Thx for the quick review!

test/wait-task/README.md
XinruZhang: yeah that's true, thx for catching this, updated.

test/wait-task/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
```
dibyom, line 2: not sure we need the LICENSE file since the root of the repo already contains one

XinruZhang: Oh yeah, the two licenses are the same, removed.

test/custom_task_test.go
```diff
@@ -401,3 +402,59 @@ spec:
...
+  - name: duration
+    value: 1s
+`, runName, apiVersion, kind)), metav1.CreateOptions{}); err != nil {
+		t.Fatalf("Failed to create TaskRun %q: %v", runName, err)
+	}
+
+	// Wait for the Run to Finish.
+	if err := WaitForRunState(ctx, c, runName, time.Minute, Succeed(runName), "RunFinished"); err != nil {
+		t.Fatalf("Waiting for Run to finish running: %v", err)
+	}
```
jerop, line 459: could you please add a check that the `run` is what we expect it to be? that is, compare the resolved run with the expected run?

bendory:
> This cl introduces a real controller to reconcile the custom task run.

Nit: "This **PR** introduces..." `CL`s (changelists) are a Google thing, `PR`s are a GitHub thing. 😺

test/custom_task_test.go
XinruZhang: Yeah SG! Done.

test/custom_task_test.go
```diff
@@ -401,3 +403,130 @@ spec:
...
+    status: %q
+    type: %s
+  observedGeneration: 1
+  `, runName, namespace, apiVersion, kind, tc.duration, tc.wantConditionReason, tc.wantConditionStatus, tc.wantConditionType)
+			wantRun := parse.MustParseRun(t, runYAML)
+			var (
+				filterTypeMeta   = cmpopts.IgnoreFields(v1alpha1.Run{}, "TypeMeta")
+				filterObjectMeta = cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion", "UID", "CreationTimestamp", "Generation", "ManagedFields")
+				filterCondition  = cmpopts.IgnoreFields(apis.Condition{}, "LastTransitionTime.Inner.Time", "Message")
+				filterRunStatus  = cmpopts.IgnoreFields(v1alpha1.RunStatusFields{}, "StartTime", "CompletionTime")
```
jerop, line 520: could this be moved to the top level so that they can be reused in other end to end tests?

test/custom_task_test.go
```diff
@@ -401,3 +403,130 @@ spec:
...
+
+func cleanUpController(t *testing.T) {
+	t.Log("Tearing down Wait Custom Task Controller...")
+	cmd := exec.Command("ko", "delete", "-f", "./config/controller.yaml")
+	cmd.Dir = "./wait-task"
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		t.Fatalf("Failed to tear down Wait Custom Task Controller: %s, Output: %s", err, out)
+	}
+}
```
jerop, line 425: the other tests in this file are "simulating behavior of runs" can we update them to use the wait custom task? 

this can be done in a follow up PR

test/custom_task_test.go
```diff
@@ -401,3 +403,130 @@ spec:
...
+
+func TestCustomTask_NoCRD(t *testing.T) {
```
jerop, line 427:
```suggestion
func TestWaitCustomTask(t *testing.T) {
```

test/wait-task/pkg/reconciler/reconciler.go
```diff
@@ -0,0 +1,95 @@
...
+	dur, err := time.ParseDuration(expr.Value.StringVal)
+	if err != nil {
+		r.Status.MarkRunFailed("InvalidDuration", "The duration param was invalid: %v", err)
+		return nil
+	}
+
+	if r.Status.StartTime == nil {
+		now := metav1.Now()
+		r.Status.StartTime = &now
+		r.Status.MarkRunRunning("Waiting", "Waiting for duration to elapse")
```
jerop, line 79: can we mark this as "Running" instead of "Waiting"? this way, we won't have to modify "IsRunning" function

```suggestion
		r.Status.MarkRunRunning("Running", "Waiting for duration to elapse")
```

test/wait.go
```diff
@@ -268,7 +268,8 @@ func Running(name string) ConditionAccessorFn {
 		if c != nil {
 			if c.Status == corev1.ConditionTrue || c.Status == corev1.ConditionFalse {
 				return true, fmt.Errorf(`%q already finished`, name)
-			} else if c.Status == corev1.ConditionUnknown && (c.Reason == "Running" || c.Reason == "Pending") {
+			} else if c.Status == corev1.ConditionUnknown &&
+				(c.Reason == "Running" || c.Reason == "Pending" || c.Reason == "Waiting") {
```
jerop, line 272: see above comment about reusing "Running" so that we don't have to change function which is used in other types besides "Run"

jerop (commented):
@XinruZhang thanks for making updates from the initial offline review -- added a few comments, almost there 😁

test/custom_task_test.go
XinruZhang: Yeah SG! What I can think of is there are two major things:
1. Test Custom Task in Pipeline
2. Test Custom Task Run with CRD
I'll open two prs for the two sub tasks above. wdyt?

Thanks for catching it!

test/wait.go
XinruZhang: Makes sense, done!

test/wait-task/pkg/reconciler/reconciler.go
XinruZhang: Done

test/wait-task/pkg/reconciler/reconciler_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/apis"
+	"knative.dev/pkg/controller"
+)
+
+var validRef = &v1alpha1.TaskRef{
+	APIVersion: "example.dev/v0",
```
jerop, line 32: can we use something more descriptive here than `example.dev`? 

maybe `wait.tekton.dev/v1alpha1`?

jerop: this will also be useful if/when we decide to add other custom tasks for testing

XinruZhang: I'm under the impression taht the new controller is meant to be the only one that is used for testing? say if we have a new feature to be tested, simply adds a new fields and related func in the reconciler? I'm okay with having more descriptive name here, probably customtask.tekton.test?

jerop: agreed that it's the only controller and we'd add functionality to it...but that could change in the future, for example if we want to test a custom task that has a CRD (instead of this one that doesn't have a CRD)

even if we remain with this as the only custom task, it's still helpful to make the name more descriptive than "example"

XinruZhang: SG! To reflect what we discussed offline: We propose to recommend people to use `*.testing.tekton.dev/<version>` when defining new resources for testing purpose. For this particular case, we'll use `wait.testing.tekton.dev/v1alpha1`.

XinruZhang: And I'm not sure the right place to document the naming convention, do you think it's a good idea to bring it up in the pipeline working group? or probably doc WG?

test/custom_task_test.go
XinruZhang: Totally agree that this is a great idea after we add more test cases using those vars in other funcs. For now I'm trying to follow https://wiki.c2.com/?DeclareVariablesAtFirstUse to declare it where it is first used. wdyt?

test/wait-task/pkg/reconciler/reconciler_test.go
jerop: you can bring it up in the pipelines wg next week, but it shouldn't be a blocker for this PR because this not user-facing

test/wait-task/pkg/reconciler/reconciler_test.go
```diff
@@ -0,0 +1,166 @@
...
+			Name: "run",
+		},
+		Spec: v1alpha1.RunSpec{
+			Ref: validRef,
+			Params: []v1beta1.Param{{
+				Name:  "duration",
+				Value: *v1beta1.NewArrayOrString("1s"),
+			}},
+		},
+	}
```
jerop, line 50: let's use yaml helpers here and the rest of this file

test/custom_task_test.go
jerop: sounds good 👍

test/wait-task/pkg/reconciler/reconciler_test.go
XinruZhang: Oh need extra confirmation here. The yaml helpers is outside of this wait-task/ module, if we want to use it, need to copy paste related funcs in test/parse/yaml.go to somewhere in this wait-task module, is that what you mean?

jerop: ah yea, you're right - let's not copy paste it

jerop (approved):
Thank you @XinruZhang, and welcome to Tekton! 🎉 

/meow

dibyom (commented):
/lgtm

PR #5350, opened 2022-08-19
Fail validation for PipelineRun params with results in value

### LINKED ISSUE #5271: restrict PipelineRunSpec.Params with a task result reference

abayer:
# Changes

fixes #5271

To prevent confusion, we should reject `PipelineRun` parameter values containing task result references, like `$(tasks.some-task.results.someResult)`, since such values will not have any substitution performed on them. Additionally, we weren't doing the basic parameter validation for `PipelineRun`s that we have been doing for `TaskRun`s, namely checking for duplicate names and use of object params without `alpha` being set, so let's turn that on here as well.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
PipelineRuns will now fail validation for duplicate parameter names or task result references in parameter values
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
@@ -65,6 +65,10 @@ func (ps *PipelineRunSpec) Validate(ctx context.Context) (errs *apis.FieldError)
 		ctx = config.SkipValidationDueToPropagatedParametersAndWorkspaces(ctx, true)
 		errs = errs.Also(ps.PipelineSpec.Validate(ctx).ViaField("pipelineSpec"))
 	}
+
+	// Validate PipelineRun parameters
+	errs = errs.Also(ps.validatePipelineRunParameters(ctx))
+
 	// Validate propagated parameters
 	errs = errs.Also(ps.validateInlineParameters(ctx))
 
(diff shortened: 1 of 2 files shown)


pritidesai (approved):
thank you @abayer for adding restriction such that pipelineRun `params` cannot use a task result reference as a value 👍 

This will not prevent a reference to `pipelineRun` result but I think it should not be a concern at this point.

PR #5351, opened 2022-08-19
Fix docs and remove redundant code for LimitRanges

lbernick:
# Changes
Prior to this commit, LimitRange transformer behavior did not match our documentation.
Our documentation stated that Tekton adjusts compute resources of containers to fit within
LimitRange minimum requests and maximum limits. This is not true. Instead, Tekton divides LimitRange
defaultRequests among Step containers, and considers the minimum only if the defaultRequest divided
by the number of Steps would be less than the minimum.

In addition, the LimitRange transformer does several things that would be done by Kubernetes anyway,
which are removed by this commit.
When a LimitRange is created, Kubernetes automatically sets defaults and defaultRequests to the max
if a user did not specify them. Kubernetes will automatically apply these defaults and defaultRequests
to any pods created. Therefore, there is no need for Tekton to set container limits to the default limits
(or in fact, to consider limits at all). There's also no need for Tekton to consider defaults/defaultRequests
for Sidecars and init containers, as we are happy to accept Kubernetes' default behavior here.
The only resource requirements we need to modify are to split a LimitRange's defaultRequests among Steps that
don't specify requests, so all other LimitRange-based transformations are removed in this commit.

Note that Tekton makes no guarantees that a TaskRun's pod will be valid (i.e. requests <= limits).
If a user configures a TaskRun in a way that is inconsistent with any LimitRanges, we allow Kubernetes
to reject the resulting pod.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
[Bug fix] Clarify limitrange documentation and remove functionality that's provided by k8s anyway
```

### DIFF
--- a/docs/compute-resources.md
+++ b/docs/compute-resources.md
@@ -158,23 +158,18 @@ will be applied to the init containers that Tekton injects into a `TaskRun`'s po
 
 ### Requests
 
-If a container does not have requests defined, the resulting container's requests are the larger of:
-- the LimitRange minimum resource requests 
-- the LimitRange default resource requests (for init and Sidecar containers) or the LimitRange default resource requests divided by the number of Step containers (for Step containers)
+If a Step container does not have requests defined, Tekton will divide a LimitRange's `defaultRequests` by the number of Step containers and apply these requests to the Steps.
+This results in a TaskRun with overall requests equal to LimitRange `defaultRequests`.
+If this value is less than the LimitRange minimum, the LimitRange minimum will be used instead.
+LimitRange `defaultRequests` are applied as-is to init containers or Sidecar containers that don't specify requests.
 
-If a container has requests defined, the resulting container's requests are the larger of:
-- the container's requests
-- the LimitRange minimum resource requests
+Containers that do specify requests will not be modified. If these requests are lower than LimitRange minimums, Kubernetes will reject the resulting TaskRun's pod.
 
 ### Limits
 
-If a container does not have limits defined, the resulting container's limits are the smaller of:
-- the LimitRange maximum resource limits 
-- the LimitRange default resource limits
-
-If a container has limits defined, the resulting container's limits are the smaller of:
-- the container's limits
-- the LimitRange maximum resource limits
+Tekton does not adjust container limits, regardless of whether a container is a Step, Sidecar, or init container.
+If a container does not have limits defined, Kubernetes will apply the LimitRange `default` to the container's limits.
+If a container does define limits, and they are less than the LimitRange `default`, Kubernetes will reject the resulting TaskRun's pod.
 
 ### Examples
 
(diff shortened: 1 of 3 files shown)


lbernick:
Closes #5278

vdemeester (commented):
> Instead, Tekton divides LimitRange defaultRequests among Step containers, and considers the minimum only if the defaultRequest divided by the number of Steps would be less than the minimum.

This one is important one in my view and shouldn't be removed. We do not need to do anything smart, but we did add this "support" to, at least, "try" to consume as little as possible, while still being valid. 

We've been up and down with this support, but I am not sure reverting some of our decision is the best move — mainly, as of today, I "feel" we are doing probably the least amount of work that we could/should do — **but** it could make sense to be even less smart and just *heavily* document how to use `LimitRange` with Tekton.

I think one of the main "trick" here is that end-users might want to use the same `LimitRange` for Tekton workloads and their own workloads, which might not work for the best.

lbernick:
@vdemeester I'm a bit confused-- this PR isn't intended to cause any changes in behavior; I think I might have just misunderstood what the limitrange transformer does when I wrote these docs. I agree we don't want to be repeatedly changing this behavior.

> I think one of the main "trick" here is that end-users might want to use the same LimitRange for Tekton workloads and their own workloads, which might not work for the best.

Yeah that is tricky :/

vdemeester:
> @vdemeester I'm a bit confused-- this PR isn't intended to cause any changes in behavior; I think I might have just misunderstood what the limitrange transformer does when I wrote these docs. I agree we don't want to be repeatedly changing this behavior.

Oh I see. Well I *think* it does, [here](https://github.com/tektoncd/pipeline/pull/5351/files#diff-791144a666d26bdcbdb9f0810388c771054f96e3f4cb59f1570dcf9751816c1cL163). I don't think the LimitRange controller sets requests or limit to the specified min version in case a default is not present.

lbernick:
> > I'm a bit confused-- this PR isn't intended to cause any changes in behavior; I think I might have just misunderstood what the limitrange transformer does when I wrote these docs. I agree we don't want to be repeatedly changing this behavior.
> 
> Oh I see. Well I _think_ it does, [here](https://github.com/tektoncd/pipeline/pull/5351/files#diff-791144a666d26bdcbdb9f0810388c771054f96e3f4cb59f1570dcf9751816c1cL163). I don't think the LimitRange controller sets requests or limit to the specified min version in case a default is not present.

I think a default will always be present-- k8s always creates the default and sets it to the max as in [this example](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/cpu-constraint-namespace/#create-a-limitrange-and-a-pod)

vdemeester:
> I think a default will always be present-- k8s always creates the default and sets it to the max as in [this example](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/cpu-constraint-namespace/#create-a-limitrange-and-a-pod)

Ah, interesting 🙃, for some reason I didn't think that was the case 🤔 (maybe on some arcane cluster that maybe didn't have that behavior…). In that case, we might be fine with it 😅

vdemeester (approved):

jerop (approved):
/lgtm

PR #5356, opened 2022-08-23
TEP-0114: Tests Wait Custom Task without CRD in PipelineRun

XinruZhang:
# Changes

This commit starts to test the behavior of Custom Task Run **without
CRD** in a **PipelineRun** against **a Custom Task Controller**.

Future Work: test Custom Task Run **with CRD** against the Wait Custom
Task Controller.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/test/custom_task_test.go
+++ b/test/custom_task_test.go
@@ -51,10 +51,18 @@ const (
 	kind       = "Wait"
 )
 
-var supportedFeatureGates = map[string]string{
-	"enable-custom-tasks": "true",
-	"enable-api-fields":   "alpha",
-}
+var (
+	supportedFeatureGates = map[string]string{
+		"enable-custom-tasks": "true",
+		"enable-api-fields":   "alpha",
+	}
+
+	filterTypeMeta          = cmpopts.IgnoreFields(metav1.TypeMeta{}, "Kind", "APIVersion")
+	filterObjectMeta        = cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion", "UID", "CreationTimestamp", "Generation", "ManagedFields")
+	filterCondition         = cmpopts.IgnoreFields(apis.Condition{}, "LastTransitionTime.Inner.Time", "Message")
+	filterRunStatus         = cmpopts.IgnoreFields(v1alpha1.RunStatusFields{}, "StartTime", "CompletionTime")
+	filterPipelineRunStatus = cmpopts.IgnoreFields(v1beta1.PipelineRunStatusFields{}, "StartTime", "CompletionTime")
+)
 
 func TestCustomTask(t *testing.T) {
 	ctx := context.Background()
@@ -424,7 +432,7 @@ func cleanUpController(t *testing.T) {
 	}
 }
 
-func TestWaitCustomTask(t *testing.T) {
+func TestWaitCustomTask_Run(t *testing.T) {
 	ctx := context.Background()
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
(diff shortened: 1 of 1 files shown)


lbernick:
I'm a bit confused what the goal of these tests are-- what do you mean by with/without CRD? The name of the test class is "TestWaitCustomTask_PipelineRun" but the the test cases are PipelineRun running/succeeded/timed out and it seems like it's testing interaction between pipelinerun timeout and wait task duration. Could you please just give a few sentences about what the behavior being tested is?

XinruZhang:
> I'm a bit confused what the goal of these tests are-- what do you mean by with/without CRD? The name of the test class is "TestWaitCustomTask_PipelineRun" but the the test cases are PipelineRun running/succeeded/timed out and it seems like it's testing interaction between pipelinerun timeout and wait task duration. Could you please just give a few sentences about what the behavior being tested is?

Thanks for the comment Lee @lbernick, great point. I think the main purpose is to test how the run status is reflected upon different PipelineRun statuses. I'll update the testing name to make it clearer.

About with/without CRD, the difference is whether the Custom Task has it's own spec&status defined. We can take the existing  `wait-task` controller as an example, we only need to specify `apiversion`, `kind` and an input param `duration` to make it work. The `wait.testing.tekton.dev/v1alpha1/Wait` doesn't have its own spec and status fields. Therefore the Run status doesn't change upon Wait.status changes.

Shortly speaking, we are basically mimic the actual usage of Custom Task Run.

wdyt? Let me know there's still confusion :)

jerop:
@lbernick @XinruZhang adding a comment to clarify further.

There are some `Custom Tasks` that don't define CRDs - they are only used inline in `Runs` and `PipelineRuns` - these are referred to as "unnamed" `Custom Tasks` in [TEP-0002](https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md). Examples of these are [`Wait`](https://github.com/tektoncd/experimental/tree/f60e1cd8ce22ed745e335f6f547bb9a44580dc7c/wait-task) and [`CEL`](https://github.com/tektoncd/experimental/tree/f60e1cd8ce22ed745e335f6f547bb9a44580dc7c/cel). 

There are some `Custom Tasks` that define CRDs e.g. [`TaskLoop`](https://github.com/tektoncd/experimental/tree/f60e1cd8ce22ed745e335f6f547bb9a44580dc7c/task-loops) and [`PipelineLoop`](https://github.com/tektoncd/experimental/tree/f60e1cd8ce22ed745e335f6f547bb9a44580dc7c/pipeline-loops).

These end-to-end tests cover the first scenario (without CRDs) and this may be sufficient for now. If needed later, we may consider adding another `Custom Task` for testing that covers the second scenario.

test/custom_task_test.go
```diff
@@ -51,10 +51,18 @@ const (
...
-	"enable-custom-tasks": "true",
-	"enable-api-fields":   "alpha",
-}
+var (
+	supportedFeatureGates = map[string]string{
+		"enable-custom-tasks": "true",
+		"enable-api-fields":   "alpha",
+	}
+
+	filterTypeMeta          = cmpopts.IgnoreFields(metav1.TypeMeta{}, "Kind", "APIVersion")
```
jerop, line 60: Why do we filter `Kind` and `APIVersion`? 

Makes sense to filter the random fields, such as `CreationTimestamp` and `CompletionTime`, but it's not clear why we filter these fields that are known

XinruZhang: The actual resources retrieved from API server don't specify Kind and APIVersion. Do you think It's better to manually set the two fields in wantResource (like wantPipelineRun or wantRun) as empty strings?

jerop: ah makes sense, checked that other tests also do the same, thanks!

jerop (approved):
Thanks @XinruZhang!

lbernick:
ok thanks @XinruZhang  @jerop for the explanation, this makes sense.

/lgtm

Issue #1744, opened 2019-12-12
Feature Request: Using ConfigMap Values For Params


tomfrenken:
# Expected Behavior
When defining parameters for a `Task` in my `Pipeline`, I would like to use values for the parameters which I defined in a `ConfigMap`, instead of hardcoding them in the `Pipeline`.

A cool solution would look like the use of `ConfigMaps` when defining environment variables in a `Pod`, there I am able to do this:

```yaml
        env:
        - name: SOME_ENV_VARIABLE
          valueFrom:
            configMapKeyRef:
              name: my-config
              key: my-key
```

if params could work the same way it would be like this:

```yaml
          params:
          - name: some-parameter
            valueFrom:
              configMapKeyRef:
                name: pipeline-config
                key: some-key
```

# Actual Behavior
If I want to use a `ConfigMap` I can only use it on a `Step` level, and I can't directly access its values, this becomes an issue in the following `Task`:

```yaml
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: pmd-test
spec:
  inputs:
    resources:
      - name: git-source-code
        type: git
    params:
      - name: image
        type: string
        description: The image that will be used in the test
        default: rawdee/pmd
      - name: command
        type: string
        description: The command that will be used in the test
        default: pmd
      - name: args
        type: string
        description: The args that will be used in the test
        default: -language java -dir /workspace/git-source-code -rulesets ./my-rulesets.xml
  steps:
    - name: test-pmd
      image: $(input.params.image)
      command: ["$(input.params.command)"]
      args: ['$(input.params.args)']
```

in this example, I want to have the possibility configure the entire `Step` based on the `ConfigMap`, but currently it would look like this:

```yaml
apiVersion: tekton.dev/v1alpha1
kind: Task
metadata:
  name: pmd-test
spec:
  volumes:
    - name: config-volume
      configMap:
        name: pipeline-config
  inputs:
    resources:
      - name: git-source-code
        type: git
  steps:
    - name: test-pmd
      volumeMounts:
        - name: config-volume
          mountPath: /my-path
      image: $(cat /my-path/my-image)
      command: ["$(cat /my-path/my-command)"]
      args: ['$(cat /my-path/my-args)]
```
Note that this wouldn't work for the `image` since I'm only able to use variable substitution there.

ImJasonH:
[...]

PR #5358, opened 2022-08-24
feat: support task spec param get value from configmap

chengjoey:
# Changes

support task spec param get value from configmap

Part of #1744 

/kind feature

# API changes
adding a field `valueFrom` within the task.params: valueFrom

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
`valueFrom` can use values for the parameters which defined in  `ConfigMap`
```


chengjoey:
#1744

chengjoey:
@vdemeester can you take a look when you have time

vdemeester:
Naive question : do we want a TEP for this ? I would think it's fine not having it but, I'd rather ask the rest of the maintainer @tektoncd/core-maintainers 🙃

lbernick:
> Naive question : do we want a TEP for this ? I would think it's fine not having it but, I'd rather ask the rest of the maintainer 

I think so-- some points that could use more discussion:
- this affects task reusability as it ties the task definition to runtime configuration (the name of the configmap). maybe the configmap should be specified in taskrun instead.
- this affects conformance, as it's only supported for a k8s implementation of tekton
- I'm curious what the use case is here, and why a workspace bound to a configmap doesn't work for it. Can we improve workspaces in some way to meet this use case?
- syntax wise, I would imagine people would want to put all keys into the same configmap so I doubt they want to include the configmap reference for each param definition

Issue #5140, opened 2022-07-14
TEP0107 & TEP0075 - Issues with propagating object parameters


chuangw6:
In TEP0107's implementation #4845 , `Parameters` are propagated directly from taskrun/pipelinerun to the variable replacements in embedded specifications without  having to define param spec in `TaskSpec.Params[*]`.

In TEP0075's implementation, we asks users to provide the object schema (keys) in `TaskSpec.Params[*].Properties` section as per [the TEP0075 proposal](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#adding-additional-json-schema-properties).

In short, object params have to define param specs but in param propagation, param specs are missed. Talked to @chitrangpatel about this offline. We landed on an idea that parameter propagation might not be applicable for object params. 

If anyone has any comments/ideas/concerns, please feel free to add them here.

chuangw6:
cc @jerop @dibyom @lbernick @vdemeester

vdemeester:
> In short, object params have to define param specs but in param propagation, param specs are missed. Talked to @chitrangpatel about this offline. We landed on an idea that parameter propagation might not be applicable for object params.

Having parameter propagation working *depending* on the type of parameters doesn't make too much sense to me. Either propagation works for all or none. We need to find a way for it to work for object params as well 🤞🏼

jerop:
fixed in https://github.com/tektoncd/pipeline/pull/5371

jerop:
Thank you @chitrangpatel!

Issue #5270, opened 2022-08-03
More flexible way to provide value for object keys


chuangw6:
### Feature request

When users provide object param value in `taskSpec.Params.default` (or `taskrunSpec.Params.Value`), it might be beneficial if it's allowed that users can provide values for only a subset of the keys instead of having to provide values for all keys which is now enforced by the validation webhook (or reconciler for taskrun level). 

For example, the following example is valid now, but if one of `url` and `commit` is missed in the `default`, validation webhook will complain about missing keys. But it would be great if it's allowed that only `url` is provided in `default`, and `commit` is provided from taskrun.

```yaml
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: object-task
spec:
  params:
    - name: giturl
      type: object
      properties:
        url:
          type: string
        commit:
          type: string
      default:
        url: "abc.com"
        commit: "sha123"
  steps:
    ...
```

### Use case

Say we want to have 10 taskruns to run the above taskrun, and we want all 10 taskruns use same value for the key `url`, but use different value for the `commit` key.
If it's allowed that only the shared value for `url` is provided in the `default`, and each taskrun only provides the value for the key `commit`, there will be no redundant lines to give value for `commit` from taskrun level.

chuangw6:
If you have any comments/feedback/questions about this, please add them here. Thanks

chuangw6:
cc @wlynch @vdemeester @chitrangpatel @dibyom @jerop @lbernick @ywluogg @Yongxuanzhang

wlynch:
sgtm! I think what you described matches what I would expect the behavior to be.

It might be helpful to add a simple TaskRun spec that doesn't work today, but perhaps should?

chuangw6:
> sgtm! I think what you described matches what I would expect the behavior to be.

Yeah, in the initial implementation, we just think the object param as whole i.e. if default/taskrun wants to provide value for it, they must provide value for all the keys. But just thinking it might be beneficial if they just provide a subset of keys as long as they cover all the keys together.


> It might be helpful to add a simple TaskRun spec that doesn't work today, but perhaps should?

SG! Added!

PR #5371, opened 2022-08-24
Propagating object params

chitrangpatel:
# Changes

Prior to this, object parameters were not being propagated down the spec during resolution.
After following the same logic applied for propagating parameters in PRs https://github.com/tektoncd/pipeline/pull/5291 and https://github.com/tektoncd/pipeline/pull/5143, we can not also propagate object params. This PR address issue https://github.com/tektoncd/pipeline/issues/5140

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Propagating object params
```

### DIFF
--- a/pkg/apis/pipeline/v1/taskrun_validation.go
+++ b/pkg/apis/pipeline/v1/taskrun_validation.go
@@ -113,30 +113,80 @@ func (ts *TaskRunSpec) validateInlineParameters(ctx context.Context) (errs *apis
 	if ts.TaskSpec == nil {
 		return errs
 	}
-	var paramSpec []ParamSpec
+	paramSpecForValidation := make(map[string]ParamSpec)
 	for _, p := range ts.Params {
-		pSpec := ParamSpec{
-			Name:    p.Name,
-			Default: &p.Value,
-		}
-		paramSpec = append(paramSpec, pSpec)
+		paramSpecForValidation = createParamSpecFromParam(p, paramSpecForValidation)
 	}
+
 	for _, p := range ts.TaskSpec.Params {
-		skip := false
-		for _, ps := range paramSpec {
-			if ps.Name == p.Name {
-				skip = true
-				break
-			}
+		var err *apis.FieldError
+		paramSpecForValidation, err = combineParamSpec(p, paramSpecForValidation)
+		if err != nil {
+			errs = errs.Also(err)
 		}
-		if !skip {
-			paramSpec = append(paramSpec, p)
-		}
+	}
+	var paramSpec []ParamSpec
+	for _, v := range paramSpecForValidation {
+		paramSpec = append(paramSpec, v)
 	}
 	if ts.TaskSpec != nil && ts.TaskSpec.Steps != nil {
+		errs = errs.Also(ValidateParameterTypes(ctx, paramSpec))
 		errs = errs.Also(ValidateParameterVariables(config.SkipValidationDueToPropagatedParametersAndWorkspaces(ctx, false), ts.TaskSpec.Steps, paramSpec))
 	}
 	return errs
+}
+
+func createParamSpecFromParam(p Param, paramSpecForValidation map[string]ParamSpec) map[string]ParamSpec {
+	value := p.Value
+	pSpec := ParamSpec{
+		Name:    p.Name,
+		Default: &value,
+		Type:    p.Value.Type,
+	}
+	if p.Value.ObjectVal != nil {
+		pSpec.Properties = make(map[string]PropertySpec)
+		prop := make(map[string]PropertySpec)
+		for k := range p.Value.ObjectVal {
+			prop[k] = PropertySpec{Type: ParamTypeString}
+		}
+		pSpec.Properties = prop
+	}
+	paramSpecForValidation[p.Name] = pSpec
+	return paramSpecForValidation
+}
+
+func combineParamSpec(p ParamSpec, paramSpecForValidation map[string]ParamSpec) (map[string]ParamSpec, *apis.FieldError) {
+	if pSpec, ok := paramSpecForValidation[p.Name]; ok {
+		// Merge defaults with provided values in the taskrun.
+		if p.Default != nil && p.Default.ObjectVal != nil {
+			for k, v := range p.Default.ObjectVal {
+				if pSpec.Default.ObjectVal == nil {
+					pSpec.Default.ObjectVal = map[string]string{k: v}
+				} else {
+					pSpec.Default.ObjectVal[k] = v
+				}
+			}
+			// If Default values of object type are provided then Properties must also be fully declared.
+			if p.Properties == nil {
+				return paramSpecForValidation, apis.ErrMissingField(fmt.Sprintf("%s.properties", p.Name))
+			}
+		}
+
+		// Properties must be defined if paramSpec is of object Type
+		if pSpec.Type == ParamTypeObject {
+			if p.Properties == nil {
+				return paramSpecForValidation, apis.ErrMissingField(fmt.Sprintf("%s.properties", p.Name))
+			}
+			// Expect Properties to be complete
+			pSpec.Properties = p.Properties
+		}
+		paramSpecForValidation[p.Name] = pSpec
+	} else {
+		// No values provided by task run but found a paramSpec declaration.
+		// Expect it to be fully speced out.
+		paramSpecForValidation[p.Name] = p
+	}
+	return paramSpecForValidation, nil
 }
 
 // validateDebug
--- a/pkg/apis/pipeline/v1/taskrun_validation_test.go
+++ b/pkg/apis/pipeline/v1/taskrun_validation_test.go
@@ -63,6 +63,61 @@ func TestTaskRun_Invalidate(t *testing.T) {
 			Paths:   []string{"spec.steps[0].args[0]"},
 		},
 		wc: config.EnableAlphaAPIFields,
+	}, {
+		name: "propagating object params not provided but used by step",
+		taskRun: &v1.TaskRun{
+			ObjectMeta: metav1.ObjectMeta{Name: "tr"},
+			Spec: v1.TaskRunSpec{
+				TaskSpec: &v1.TaskSpec{
+					Steps: []v1.Step{{
+						Name:    "echo",
+						Image:   "ubuntu",
+						Command: []string{"echo"},
+						Args:    []string{"$(params.task-words.hello)"},
+					}},
+				},
+			},
+		},
+		want: &apis.FieldError{
+			Message: `non-existent variable in "$(params.task-words.hello)"`,
+			Paths:   []string{"spec.steps[0].args[0]"},
+		},
+		wc: config.EnableAlphaAPIFields,
+	}, {
+		name: "propagating object properties not provided",
+		taskRun: &v1.TaskRun{
+			ObjectMeta: metav1.ObjectMeta{Name: "tr"},
+			Spec: v1.TaskRunSpec{
+				Params: []v1.Param{{
+					Name: "task-words",
+					Value: v1.ParamValue{
+						Type:      v1.ParamTypeObject,
+						ObjectVal: map[string]string{"hello": "task run"},
+					},
+				}},
+				TaskSpec: &v1.TaskSpec{
+					Params: []v1.ParamSpec{{
+						Name: "task-words",
+						Type: v1.ParamTypeObject,
+						Default: &v1.ParamValue{
+							Type:      v1.ParamTypeObject,
+							ObjectVal: map[string]string{"hello": "task run def"},
+						},
+					}},
+					Steps: []v1.Step{{
+						Name:    "echo",
+						Image:   "ubuntu",
+						Command: []string{"echo"},
+						Args:    []string{"$(params.task-words.hello)"},
+					}},
+				},
+			},
+		},
+		want: &apis.FieldError{
+			Message: `missing field(s)`,
+			Paths:   []string{"spec.task-words.properties"},
+		},
+		wc: config.EnableAlphaAPIFields,
 	}}
 	for _, ts := range tests {
 		t.Run(ts.name, func(t *testing.T) {
@@ -97,6 +152,7 @@ func TestTaskRun_Validate(t *testing.T) {
 				Params: []v1.Param{{
 					Name: "task-words",
 					Value: v1.ParamValue{
+						Type:     v1.ParamTypeArray,
 						ArrayVal: []string{"hello", "task run"},
 					},
 				}},
@@ -112,6 +168,62 @@ func TestTaskRun_Validate(t *testing.T) {
 		},
 		wc: config.EnableAlphaAPIFields,
 	}, {
+		name: "propagating object params from taskrun to steps",
+		taskRun: &v1.TaskRun{
+			ObjectMeta: metav1.ObjectMeta{Name: "tr"},
+			Spec: v1.TaskRunSpec{
+				Params: []v1.Param{{
+					Name: "task-words",
+					Value: v1.ParamValue{
+						Type:      v1.ParamTypeObject,
+						ObjectVal: map[string]string{"hello": "task run"},
+					},
+				}},
+				TaskSpec: &v1.TaskSpec{
+					Steps: []v1.Step{{
+						Name:    "echo",
+						Image:   "ubuntu",
+						Command: []string{"echo"},
+						Args:    []string{"$(params.task-words.hello)"},
+					}},
+				},
+			},
+		},
+		wc: config.EnableAlphaAPIFields,
+	}, {
+		name: "propagating object params with taskrun mixing value types",
+		taskRun: &v1.TaskRun{
+			ObjectMeta: metav1.ObjectMeta{Name: "tr"},
+			Spec: v1.TaskRunSpec{
+				Params: []v1.Param{{
+					Name: "task-words",
+					Value: v1.ParamValue{
+						Type: v1.ParamTypeObject,
+					},
+				}},
+				TaskSpec: &v1.TaskSpec{
+					Params: []v1.ParamSpec{{
+						Name: "task-words",
+						Type: v1.ParamTypeObject,
+						Properties: map[string]v1.PropertySpec{
+							"hello": {Type: v1.ParamTypeString},
+						},
+						Default: &v1.ParamValue{
+							Type:      v1.ParamTypeObject,
+							ObjectVal: map[string]string{"hello": "task run def"},
+						},
+					}},
+					Steps: []v1.Step{{
+						Name:    "echo",
+						Image:   "ubuntu",
+						Command: []string{"echo"},
+						Args:    []string{"$(params.task-words.hello)"},
+					}},
+				},
+			},
+		},
+		wc: config.EnableAlphaAPIFields,
+	}, {
 		name: "propagating partial params with different provided and default names",
 		taskRun: &v1.TaskRun{
 			ObjectMeta: metav1.ObjectMeta{Name: "tr"},
@@ -119,6 +231,7 @@ func TestTaskRun_Validate(t *testing.T) {
 				Params: []v1.Param{{
 					Name: "task-words",
 					Value: v1.ParamValue{
+						Type:     v1.ParamTypeArray,
 						ArrayVal: []string{"hello", "task run"},
 					},
 				}},
@@ -143,6 +256,40 @@ func TestTaskRun_Validate(t *testing.T) {
 		},
 		wc: config.EnableAlphaAPIFields,
 	}, {
+		name: "propagating object params with one declared in taskspec and other provided by taskrun",
+		taskRun: &v1.TaskRun{
+			ObjectMeta: metav1.ObjectMeta{Name: "tr"},
+			Spec: v1.TaskRunSpec{
+				Params: []v1.Param{{
+					Name: "task-words",
+					Value: v1.ParamValue{
+						Type:      v1.ParamTypeObject,
+						ObjectVal: map[string]string{"hello": "task run"},
+					},
+				}},
+				TaskSpec: &v1.TaskSpec{
+					Params: []v1.ParamSpec{{
+						Name: "task-words-2",
+						Type: v1.ParamTypeObject,
+						Properties: map[string]v1.PropertySpec{
+							"hello": {Type: v1.ParamTypeString},
+						},
+						Default: &v1.ParamValue{
+							Type:      v1.ParamTypeObject,
+							ObjectVal: map[string]string{"hello": "task run def"},
+						},
+					}},
+					Steps: []v1.Step{{
+						Name:    "task-echo",
+						Image:   "ubuntu",
+						Command: []string{"echo"},
+						Args:    []string{"$(params.task-words.hello)", "$(params.task-words-2.hello)"},
+					}},
+				},
+			},
+		},
+		wc: config.EnableAlphaAPIFields,
+	}, {
 		name: "propagating partial params in taskrun",
 		taskRun: &v1.TaskRun{
 			ObjectMeta: metav1.ObjectMeta{Name: "tr"},
@@ -150,6 +297,7 @@ func TestTaskRun_Validate(t *testing.T) {
 				Params: []v1.Param{{
 					Name: "task-words",
 					Value: v1.ParamValue{
+						Type:     v1.ParamTypeArray,
 						ArrayVal: []string{"hello", "task run"},
 					},
 				}},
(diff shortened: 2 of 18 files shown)


chitrangpatel:
@chuangw6 Please take a look when you can

chitrangpatel:
I will work on getting the coverage to its original value/better tomorrow

pkg/apis/pipeline/v1/taskrun_validation.go
```diff
@@ -116,15 +116,44 @@ func (ts *TaskRunSpec) validateInlineParameters(ctx context.Context) (errs *apis
...
+		for i, ps := range paramSpec {
 			if ps.Name == p.Name {
+				if p.Default != nil && p.Default.ObjectVal != nil {
+					for k, v := range p.Default.ObjectVal {
+						if paramSpec[i].Default.ObjectVal == nil {
+							paramSpec[i].Default.ObjectVal = map[string]string{k: v}
+						} else {
+							paramSpec[i].Default.ObjectVal[k] = v
+						}
+						if fmt.Sprintf("%T", v) == "string" {
```
lbernick, line 146: Can you explain a bit what's going on here or maybe give a yaml example? this is if the literal value of an object param property is "string"? I don't understand why this is a special case or why we're using `sprintf` in this way

chitrangpatel: Thanks for catching that @lbernick! I put it there originally thinking that the object params also allowed us to provide arrays as values. After going through the TEPs regarding object params yesterday, I figured out that we currently only support string values. I missed updating the conditional later. The tests passed because nothing really triggered a failure. I will fix it ASAP.

chuangw6 (commented):
Thanks @chitrangpatel !

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -319,10 +319,6 @@ func (p ParamSpec) ValidateType() *apis.FieldError {
 // definition of `properties` section and the type of a PropertySpec is allowed.
 // (Currently, only string is allowed)
 func (p ParamSpec) ValidateObjectType() *apis.FieldError {
-	if p.Type == ParamTypeObject && p.Properties == nil {
-		return apis.ErrMissingField(fmt.Sprintf("%s.properties", p.Name))
-	}
-
```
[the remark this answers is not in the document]
chitrangpatel: That's right. We cannot have this check because there might be no `properties` field defined when users are propagating parameters. Good catch though. When we explicitly declare the task without parameter propagation we want this check. I will fix accordingly and ping you later to verify.

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -119,6 +119,17 @@ func TestTaskSpecValidatePropagatedParamsAndWorkspaces(t *testing.T) {
...
+		name: "propagating object params valid step with script",
+		fields: fields{
+			Steps: []v1beta1.Step{{
+				Name:  "propagatingobjectparams",
+				Image: "my-image",
+				Script: `
+				#!/usr/bin/env bash
+				$(params.message.foo)`,
+			}},
+		},
```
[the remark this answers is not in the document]
chitrangpatel: I can do that to avoid confusion. Thanks!

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
```diff
@@ -177,6 +245,36 @@ func TestTaskRun_Validate(t *testing.T) {
...
+					Steps: []v1beta1.Step{{
+						Name:    "echo",
+						Image:   "ubuntu",
+						Command: []string{"echo"},
+						Args:    []string{"$(params.task-words.hello)"},
+					}},
+				},
+			},
+		},
+		wc: config.EnableAlphaAPIFields,
```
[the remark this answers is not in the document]
chitrangpatel: Will do!

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -117,15 +117,44 @@ func (ts *TaskRunSpec) validateInlineParameters(ctx context.Context) (errs *apis
...
+		}
+		if p.Value.ObjectVal != nil {
+			prop := make(map[string]PropertySpec)
+			for k, v := range p.Value.ObjectVal {
+				if fmt.Sprintf("%T", v) == "string" {
+					prop[k] = PropertySpec{Type: ParamTypeString}
+				}
+			}
+			pSpec.Type = ParamTypeObject
+			pSpec.Properties = prop
```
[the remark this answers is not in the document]
chitrangpatel: @chuangw6 did you mean I replace `prop[k] = PropertySpec{Type: ParamTypeString}` with `prop[k] = PropertySpec{Type: p.Value.Type}`?

chitrangpatel:
@lbernick @chuangw6 feel free to review again. I have incorporated your changes.

pkg/apis/pipeline/v1/taskrun_validation_test.go
```diff
@@ -111,6 +131,28 @@ func TestTaskRun_Validate(t *testing.T) {
 			},
 		},
 		wc: config.EnableAlphaAPIFields,
+	}, {
+		name: "propagating partial params with different provided and default names",
```
lbernick, line 135: this test case has the same name as the next one, can you make one or both more descriptive?

pkg/apis/pipeline/v1/taskrun_validation_test.go
```diff
@@ -142,6 +184,32 @@ func TestTaskRun_Validate(t *testing.T) {
 			},
 		},
 		wc: config.EnableAlphaAPIFields,
+	}, {
+		name: "propagating partial object params with different provided and default names",
```
lbernick, line 188: I think I left a similar comment on a previous PR but params don't have default names (just default values), so I think this is a bit confusing-- it looks like here the task spec is declaring another param task-words-2 and not using it (which I'm sort of surprised is valid); the test case name should reflect that. Also what's a "partial" param?

pkg/apis/pipeline/v1/taskrun_validation_test.go
```diff
@@ -176,6 +244,36 @@ func TestTaskRun_Validate(t *testing.T) {
 			},
 		},
 		wc: config.EnableAlphaAPIFields,
+	}, {
+		name: "propagating partial object params in taskrun",
```
lbernick, line 248:
```suggestion
		name: "propagating object params with multiple keys",
```

pkg/apis/pipeline/v1/taskrun_validation_test.go
```diff
@@ -202,6 +300,32 @@ func TestTaskRun_Validate(t *testing.T) {
 			},
 		},
 		wc: config.EnableAlphaAPIFields,
+	}, {
+		name: "propagating object params with taskrun same names",
```
lbernick, line 304: this is technically not propagated params right? Just normal params.

pkg/apis/pipeline/v1/taskrun_validation_test.go
```diff
@@ -176,6 +244,36 @@ func TestTaskRun_Validate(t *testing.T) {
...
+						Default: &v1.ParamValue{
+							Type:      v1.ParamTypeObject,
+							ObjectVal: map[string]string{"hello": "task run def", "hi": "task"},
+						},
+					}},
+					Steps: []v1.Step{{
+						Name:    "echo",
+						Image:   "ubuntu",
+						Command: []string{"echo"},
+						Args:    []string{"$(params.task-words.hello)", "$(params.task-words.hi)"},
```
lbernick, line 271: this is interesting-- it seems like the task spec has defined a default value for the object param which contains the key "hi", and the default value is overridden by the value defined in the taskrun, which doesn't contain the key "hi". Why is it valid to use the key "hi" in the args? When a taskrun overrides the default value for an object param, is that on a per-key basis or for the param as a whole? Is this documented anywhere?

pkg/apis/pipeline/v1/taskrun_validation_test.go
chitrangpatel: Yes, I had that for increasing coverage. I will use it to avoid confusion. Partial params in my mind (which you appropriately worded in the next comment below) is really params where some of the keys are provided by the task run/pipelinerun and others are defaulted in the task/pipeline spec. (i.e. a case where only some of the params are explicitly provided).

pkg/apis/pipeline/v1/taskrun_validation_test.go
chitrangpatel: Sure.

pkg/apis/pipeline/v1/taskrun_validation_test.go
chitrangpatel: That's a good question. Below is an example that should explain a few things. In the original [TEP](https://github.com/tektoncd/community/blob/main/teps/0107-propagating-parameters.md) on propagating params, we captured the usage for string and array types. But the main idea was that we wanted to have a [scoping mechanism](https://github.com/tektoncd/community/blob/main/teps/0107-propagating-parameters.md#scope-and-precedence). In this case, the keys provided at the task run will replace the default keys during substitution.  The scoping part is documented in the context of string and array types. Thanks for bringing this up. It seems like we should also update the documentation to show object replacements. 

```yaml
kind: TaskRun
metadata:
...
spec:
  params:
  - name: gitrepo
    value:
      commit: sha123
      url: xyz.com
  serviceAccountName: default
  taskSpec:
    params:
    - default:
        branch: main
        commit: sha456
        url: abc.com
      name: gitrepo
      type: object
    steps:
    - args:
      - echo
      - --url=$(params.gitrepo.url)
      - --commit=$(params.gitrepo.commit)
      - --branch=$(params.gitrepo.branch)
...
status:
...
  taskSpec:
    params:
    - default:
        branch: main
        commit: sha456
        url: abc.com
      name: gitrepo
      type: object
    steps:
    - args:
      - echo
      - --url=xyz.com
      - --commit=sha123
      - --branch=main
...
```

pkg/apis/pipeline/v1beta1/taskrun_validation.go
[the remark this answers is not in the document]
chuangw6: I meant to say `pSpec.Type = p.value.Type` to make the type setting easier & more accurate for the fake spec, and we should keep `prop[k] = PropertySpec{Type: ParamTypeString}`. 

Chatted with @chitrangpatel offline. we aligned on this.

pkg/apis/pipeline/v1/taskrun_validation_test.go
lbernick: I think this is actually more related to TEP-0075 than TEP-0107. In TEP-0075, under [declaring object results and params](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#declaring-object-results-and-params), it states "there will be no behavior to merge the default with the provided value, if for example the provided value specified some fields but not others". This sounds like if a user wants to replace the default value of an object param, any keys they do not specify will not have values, unlike the status.taskSpec.steps in your example.

@chuangw6 do you know if this behavior is validated anywhere? Would one of you be able to add documentation for it? I think it would also be reasonable to support the behavior shown in this example, but we would need to update TEP-0075 first.

chuangw6: Yes, this is validated in [webhook](https://github.com/tektoncd/pipeline/blob/af4d0e6eb018a34c950180939a7f6a2f24941da6/pkg/apis/pipeline/v1beta1/task_validation.go#L408-L426). Since TEP75 proposes that `properties` section is required for either inline or ref definition of the spec, the validation makes sure the key used in steps is indeed declared in `properties`. The way I interpreted this line *"there will be no behavior to merge the default with the provided value, if for example the provided value specified some fields but not others"* in the TEP75 was that if an object `default` is provided, the value for all keys must be provided. Same for the value provided in taskrun. 

When it comes to propagating object params, I agree that it makes sense and provides more flexibility to propagate on   a per-key basis like the example @chitrangpatel shows above. @chitrangpatel also proposed that `properties` section is not required when it comes to propagating object param (like the example above), which I think makes sense too. This is because param propagating only happens for inline spec. As long keys being referenced exist in either `default` or `value`, it would be valid usage. (@chitrangpatel Please correct me if I'm wrong.)

So now it's about updating the documentation like TEP75 and [tasks.md](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#parameter-type):
- the merge behaviour between `default` and `value` is allowed. And `value` from taskruns takes precedence. (FYI: #5270 might be related too.)
- `properties` section is not required when the spec is inline (embedded).

Did I miss anything here?

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -120,17 +120,49 @@ func (ts *TaskRunSpec) validateInlineParameters(ctx context.Context) (errs *apis
...
+						}
+						if paramSpec[i].Properties == nil {
+							paramSpec[i].Properties = map[string]PropertySpec{
+								k: {Type: ParamTypeString},
+							}
+						} else {
+							paramSpec[i].Properties[k] = PropertySpec{Type: ParamTypeString}
+						}
+					}
+				}
```
chuangw6, line 153: I am wondering if we can change `paramSpec` from `[]ParamSpec` to `map[string]ParamSpec` type (paramname:paramSpec) to avoid the 3-layer for loops. When all ParamSpecs are needed at the end for validation, we can just use a single for loop to get all the values of the map. wdyt?

nit: maybe also rename `paramSpec` to `paramSpecsForValidation` or just add a comment to explain what is this variable for?

pkg/apis/pipeline/v1/taskrun_validation_test.go
lbernick: Could you point me to any test cases that cover this?

I agree that having defaults overridden on a per-key basis is much more useful behavior, but that's not how I interpreted TEP-0075. It does look like #5270 is the feature we're describing. Can we update TEP-0075 as you've suggested?

No strong opinions about whether `properties` should be required in param propagation, but that should likely also be covered in TEP-0107.

chuangw6: > Could you point me to any test cases that cover this?

Sure [here](https://github.com/tektoncd/pipeline/blob/af4d0e6eb018a34c950180939a7f6a2f24941da6/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go#L1989-L2132) is the tests that cover this. I realized that the test for covering task steps was missed. I just added that https://github.com/tektoncd/pipeline/pull/5411. PTAL. Thanks @lbernick 


> I agree that having defaults overridden on a per-key basis is much more useful behavior, but that's not how I interpreted TEP-0075. It does look like https://github.com/tektoncd/pipeline/issues/5270 is the feature we're describing. Can we update TEP-0075 as you've suggested?

Sure. I'll update the TEP.

pkg/apis/pipeline/v1beta1/taskrun_validation.go
chitrangpatel: I can change it to a map, however, it will require changing many other calls too since `[]ParamSpec` is required by `ValidatePipelineParams` and many other functions. They will all need to be changed. Might be outside the scope of just this PR.

chitrangpatel: I can certainly change `paramSpec` to `paramSpecForValidation`.

chuangw6: In this function, we just called `ValidateParameterVariables` at the end to delegate the validation job. Since it expects a slice of paramspec for the third argument, we can just collect the map's values using a single loop right before calling this func. This may help us avoid layered loops?

chitrangpatel: I see what you mean. So basically just convert the map to a slice of param spec at the end before passing it for validation.

chuangw6: Yes!

chitrangpatel: Done!

pkg/apis/pipeline/v1/taskrun_validation_test.go
chitrangpatel: I updated the names. Should be ok now.

pkg/apis/pipeline/v1beta1/taskrun_validation.go
```diff
@@ -114,27 +114,64 @@ func (ts *TaskRunSpec) validateInlineParameters(ctx context.Context) (errs *apis
...
 			}
+		} else {
+			if p.Default != nil && p.Default.ObjectVal != nil {
+				prop := make(map[string]PropertySpec)
+				for k := range p.Default.ObjectVal {
+					prop[k] = PropertySpec{Type: ParamTypeString}
+				}
+				p.Properties = prop
+			}
+			paramSpecForValidation[p.Name] = p
```
chuangw6, line 166: Feel like this part is a bit hard to read. How about decomposing them into helper function that does the merge job? 

Two benefits I can think of: 1. easy to understand, 2. it can be reused in `pipelinerun_validation.go` as well.
```
=========
                var the_map map[string]string
		if v, ok := paramSpecForValidation[p.Name]; ok {
			the_map = mergeObjectPartOfParamValue(v.Default, p.Default)
		} else {
			the_map = mergeParamValueObjectPart(nil, p.Default)
		}
		paramSpecForValidation[p.Name].Default = the_map
		paramSpecForValidation[p.Name].Properties = composePropertySectionFromAMap(the_map)
		
==========	
helpers
==========
		
func mergeObjectPartOfParamValue(input1 ParamValue, input2 ParamValue) map[string]string {
             // check nil of input1 and input2
             // if same key exist in both, value in input1 will take precedence 
}
		
func composePropertySectionFromAMap(input map[string]string) map[string]PropertySpec {
}
```

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
```diff
@@ -177,6 +284,37 @@ func TestTaskRun_Validate(t *testing.T) {
...
+					Steps: []v1beta1.Step{{
+						Name:    "echo",
+						Image:   "ubuntu",
+						Command: []string{"echo"},
+						Args:    []string{"$(params.task-words.hello)", "$(params.task-words.hi)"},
+					}},
+				},
+			},
+		},
+		wc: config.EnableAlphaAPIFields,
```
chuangw6, line 317: Without `properties`, this tricks the [check](https://github.com/tektoncd/pipeline/blob/0fe293118eccdbf4dd478f00ef8d41e0bff06af5/pkg/reconciler/taskrun/validate_resources.go#L86-L88) on the reconciler side to pass because the check compares what are provided on the taskrun level against what are needed in the spec's `properties` section. And the needed is empty so the check is passed.

Although taskrun will not need to provide values for all keys required in `properties` because of https://github.com/tektoncd/pipeline/issues/5270 , the reconciler side will still need to compare things (combined key set provided in taskrun and spec's default) against `properties` section anyway.

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
```diff
@@ -112,6 +133,59 @@ func TestTaskRun_Validate(t *testing.T) {
...
+					Steps: []v1beta1.Step{{
+						Name:    "echo",
+						Image:   "ubuntu",
+						Command: []string{"echo"},
+						Args:    []string{"$(params.task-words.hello)"},
+					}},
+				},
+			},
+		},
+		wc: config.EnableAlphaAPIFields,
```
chuangw6, line 188: I am a bit doubt if this a real case for propagating param. If no value is provided in taskrun, why does user write something about that param in taskrun?

pkg/apis/pipeline/v1beta1/taskrun_validation.go
chitrangpatel: Agreed. I was about to do that after I got all the tests passing.

pkg/apis/pipeline/v1beta1/task_validation.go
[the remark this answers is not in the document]
chitrangpatel: @chuangw6, based on our discussion,  I updated this line to:
```
if p.Type == ParamTypeObject && p.Default != nil && p.Properties == nil {
```
This will enforce that if the defaults are specified then properties must also be provided.

pkg/apis/pipeline/v1beta1/task_validation.go
```diff
@@ -366,10 +362,10 @@ func ValidateParameterVariables(ctx context.Context, steps []Step, params []Para
 	errs = errs.Also(validateNameFormat(stringParameterNames.Insert(arrayParameterNames.List()...), objectParamSpecs))
 	if config.ValidateParameterVariablesAndWorkspaces(ctx) == true {
 		errs = errs.Also(validateVariables(ctx, steps, "params", allParameterNames))
+		errs = errs.Also(validateObjectDefault(objectParamSpecs))
+		errs = errs.Also(validateObjectUsage(ctx, steps, objectParamSpecs))
 	}
-	errs = errs.Also(validateArrayUsage(steps, "params", arrayParameterNames))
-	errs = errs.Also(validateObjectDefault(objectParamSpecs))
-	return errs.Also(validateObjectUsage(ctx, steps, objectParamSpecs))
+	return errs.Also(validateArrayUsage(steps, "params", arrayParameterNames))
```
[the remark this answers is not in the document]
chitrangpatel: I'm not sure. The tests were passing without including it so I didn't want to change the default behaviour.

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
chitrangpatel: I now provide properties after our discussion today.

chitrangpatel: Resolve this if you are now satisfied.

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
chitrangpatel: I had this in place to increase coverage. Because of the behaviour of propagated params, when we combine defaults with provided values, I need to cover different cases.

pkg/apis/pipeline/v1beta1/taskrun_validation.go
chitrangpatel: Done now.

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
```diff
@@ -360,6 +544,7 @@ func TestPipelineRun_Validate(t *testing.T) {
 						Params: []v1beta1.Param{{
 							Name: "task-words",
 							Value: v1beta1.ParamValue{
+								Type:     v1beta1.ParamTypeArray,
```
jerop, line 547: why is type array added in some of these test cases? it doesn't seem related to the changes in this commit -- would rather not add them in this commit if not necessary

jerop (commented):
could you please examples with propagated object params to the docs? existing examples only show propagated string params and it'd be helpful to make it clear that propagated object params are supported too

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
chitrangpatel: That's because Chuang pointed out that we should be ParamType checking as well when we are propagating parameters. I had overlooked it and so my original test cases didn't have that. When I added the Type checking in line 189 of `pipelinerun_validation.go` to do it properly for object params, the arrays tests failed as well. Hence this change.

pkg/apis/pipeline/v1beta1/pipelinerun_validation.go
```diff
@@ -156,66 +156,62 @@ func (ps *PipelineRunSpec) validateInlineParameters(ctx context.Context) (errs *
...
 		}
 	}
+	var paramSpec []ParamSpec
+	for _, v := range paramSpecForValidation {
+		paramSpec = append(paramSpec, v)
+	}
 	if ps.PipelineSpec != nil && ps.PipelineSpec.Tasks != nil {
 		for _, pt := range ps.PipelineSpec.Tasks {
 			if pt.TaskSpec != nil && pt.TaskSpec.Steps != nil {
+				errs = errs.Also(ValidateParameterTypes(ctx, paramSpec))
```
chitrangpatel, line 189: @jerop this is the line that caused the array test to fail and so I needed to add the Type field.

pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go
chitrangpatel: Let me know.

pkg/apis/pipeline/v1/taskrun_validation_test.go
lbernick: @chitrangpatel I think this test case should result in an error until we resolve https://github.com/tektoncd/pipeline/pull/5427

chitrangpatel: In that case, can we put this PR on hold until https://github.com/tektoncd/pipeline/pull/5427 is resolved? A lot of parameter propagation validation is around combining provided values with the defaults and comparing against the Properties field.

lbernick: sure, fine with me

vdemeester (approved):

chitrangpatel:
The following is the coverage report on the affected files.
Say `/test chitrang-test-1` to re-run this coverage report

File | Old Coverage | New Coverage | Delta
---- |:------------:|:------------:|:-----:
[pkg/apis/pipeline/v1/pipeline_validation.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/5371/chitrang-test-1/51421370456/artifacts/line-cov.html#file0) | 99.4% | 98.4% | -1.1
[pkg/apis/pipeline/v1/task_validation.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/5371/chitrang-test-1/51421370456/artifacts/line-cov.html#file1) | 97.0% | 97.4% | 0.4
[pkg/apis/pipeline/v1/taskrun_validation.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/5371/chitrang-test-1/51421370456/artifacts/line-cov.html#file2) | 98.4% | 99.2% | 0.8
[pkg/apis/pipeline/v1beta1/pipeline_validation.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/5371/chitrang-test-1/51421370456/artifacts/line-cov.html#file3) | 99.6% | 100.0% | 0.4
[pkg/apis/pipeline/v1beta1/pipelinerun_validation.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/5371/chitrang-test-1/51421370456/artifacts/line-cov.html#file4) | 97.4% | 98.5% | 1.1
[pkg/apis/pipeline/v1beta1/task_validation.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/5371/chitrang-test-1/51421370456/artifacts/line-cov.html#file5) | 97.1% | 97.6% | 0.4
[pkg/apis/pipeline/v1beta1/taskrun_validation.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/5371/chitrang-test-1/51421370456/artifacts/line-cov.html#file6) | 98.4% | 99.2% | 0.8

PR #5374, opened 2022-08-25
Replace time.After with NewTimers

### LINKED ISSUE #5362: Avoid using time.After() in select statements
# Expected Behavior

` time.After()` is not used in select statements

# Actual Behavior

For instance: https://github.com/tektoncd/pipeline/blob/99b8b196ea753af36befda8c0e0e1eaa9490ae68/pkg/pipelinerunmetrics/metrics.go#L290-L300

# Additional Info

Tekton Pipeline version: v0.33 -> v0.38

From the security audit:

> Calls to time.After in for/select statements can lead to memory leaks because the garbage collector does not clean up the underlying Timer object until the timer fires. A new timer, which requires resources, is initialized at each iteration of the for loop (and, hence, the select statement). As a result, many routines originating from the time.After call could lead to overconsumption of the memory.

> Short term, consider refactoring the code that uses the time.After function in for/select loops using tickers. This will prevent memory leaks and crashes caused by memory exhaustion.
> Long term, ensure that the time.After method is not used in for/select routines. Periodically use the Semgrep query to check for and detect similar patterns.

abayer:
# Changes

fixes #5362

See https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/ for more details.

/kind security

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Replace usages of time.After
```

### DIFF
--- a/cmd/entrypoint/waiter_test.go
+++ b/cmd/entrypoint/waiter_test.go
@@ -44,11 +44,16 @@ func TestRealWaiterWaitMissingFile(t *testing.T) {
 		}
 		close(doneCh)
 	}()
+
+	delay := time.NewTimer(2 * testWaitPollingInterval)
 	select {
+	case <-delay.C:
+		// Success
 	case <-doneCh:
 		t.Errorf("did not expect Wait() to have detected a file at path %q", tmp.Name())
-	case <-time.After(2 * testWaitPollingInterval):
-		// Success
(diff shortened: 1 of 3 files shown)


imjasonh (commented):
/lgtm

vdemeester (approved):

Issue #5320, opened 2022-08-12
Refactor pipeline related function in taskrun_types.go


JeromeJu:
This issue address that the [`IsPartOfPipeline` func](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/taskrun_types.go/#L468-L480) in `taskrun_types` does not seem to belong in the file.  This should be refactored and essentially moved to pipeline related files.

Thanks @lbernick 
_Originally posted by @lbernick in https://github.com/tektoncd/pipeline/pull/5264#discussion_r943580056_

PR #5377, opened 2022-08-25
Move bespoke functionality out of TaskRun API

xchapter7x:
The TaskRun API was maintaining a `isPartOfPipeline`
function, which was only consumed by a single package (taskrunmetrics).

As the only consumer, taskrunmetrics should own this interface, and
pull the taskrun meta data it needs on its own,
rather than depending on the package to maintain a compatible API
for its bespoke use case.

closes https://github.com/tektoncd/pipeline/issues/5320

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/taskrun_types_test.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_types_test.go
@@ -22,7 +22,6 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/pipeline/test/diff"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


pkg/apis/pipeline/v1beta1/taskrun_types_test.go
```diff
@@ -217,50 +216,6 @@ func TestTaskRunHasStarted(t *testing.T) {
 	}
 }
 
-func TestTaskRunIsOfPipelinerun(t *testing.T) {
```
lbernick, line 220: why not move these into metrics_test.go?

pkg/taskrunmetrics/metrics.go
```diff
@@ -473,6 +474,17 @@ func (r *Recorder) CloudEvents(ctx context.Context, tr *v1beta1.TaskRun) error {
 	return nil
 }
 
+func isPartOfPipeline(tr *v1beta1.TaskRun) (bool, string, string) {
```
lbernick, line 477: could you add a docstring stating that this returns the values of the pipeline label and pipelinerun label if it is part of a pipeline?

lbernick (commented):
thanks John! looks good, happy to approve once tests are moved over

pkg/taskrunmetrics/metrics.go
```diff
@@ -473,6 +474,17 @@ func (r *Recorder) CloudEvents(ctx context.Context, tr *v1beta1.TaskRun) error {
 	return nil
 }
```
JeromeJu, line 476:
```suggestion
// IsPartOfPipeline return true if TaskRun is a part of a Pipeline.
// It also return the name of Pipeline and PipelineRun
```

JeromeJu:
Thanks @xchapter7x for fixing this.

pkg/apis/pipeline/v1beta1/taskrun_types_test.go
xchapter7x: currently, the same scenarios are tested through the public API of the `taskrunmetrics` package.

I also felt that adding a test for a un-exported function would make the tests a little brittle, as I feel the implementation should be able to change below the public API without modifying the tests.

Im happy to add specific tests for this function, but would prefer to also then make it exported if we go that path.

lbernick: ok that makes sense-- it does seem like there's test coverage [here](https://github.com/tektoncd/pipeline/blob/c245f3ecf9eafae7ccef1afdefbe234093baf438/pkg/taskrunmetrics/metrics_test.go#L275)

lbernick (approved):

jerop (commented):
/lgtm

thanks @xchapter7x!

bendory:
LGTM, and `+1` for technical hygiene!

Issue #4985, opened 2022-06-15
V1 version of TaskRun


lbernick:
This issue tracks the release of a v1 version of the TaskRun CRD. More details in #4983.

PR #5381, opened 2022-08-25
V1: Add conversion for TaskRunSpec.TaskRef.Bundle

JeromeJu:
This commit adds support for TaskRef.Bundle when converting between
v1beta1 and v1 versions of TaskRef. This allows us to release v1 TaskRun
 in a backwards compatible way by ensuring that v1beta1 TaskRuns and
PipelineRuns with Bundle converted to v1 TaskRuns conversion.

part of #4985 
/kind misc

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/taskref_conversion.go
+++ b/pkg/apis/pipeline/v1beta1/taskref_conversion.go
@@ -10,18 +10,38 @@ func (tr TaskRef) convertTo(ctx context.Context, sink *v1.TaskRef) {
 	sink.Name = tr.Name
 	sink.Kind = v1.TaskKind(tr.Kind)
 	sink.APIVersion = tr.APIVersion
-	// TODO: handle bundle in #4546
 	new := v1.ResolverRef{}
 	tr.ResolverRef.convertTo(ctx, &new)
 	sink.ResolverRef = new
+	tr.convertBundleToResolver(sink)
 }
 
 func (tr *TaskRef) convertFrom(ctx context.Context, source v1.TaskRef) {
 	tr.Name = source.Name
 	tr.Kind = TaskKind(source.Kind)
 	tr.APIVersion = source.APIVersion
-	// TODO: handle bundle in #4546
 	new := ResolverRef{}
 	new.convertFrom(ctx, source.ResolverRef)
 	tr.ResolverRef = new
+}
+
+// convertBundleToResolver converts v1beta1 bundle string to a remote reference with the bundle resolver in v1.
+// The conversion from Resolver to Bundle is not being supported since remote resolution would be turned on by
+// default and it will be in beta before the stored version of CRD getting swapped to v1.
+func (tr TaskRef) convertBundleToResolver(sink *v1.TaskRef) {
+	if tr.Bundle != "" {
+		sink.ResolverRef = v1.ResolverRef{
+			Resolver: "bundles",
+			Params: []v1.Param{{
+				Name:  "bundle",
+				Value: v1.ParamValue{StringVal: tr.Bundle},
+			}, {
+				Name:  "name",
+				Value: v1.ParamValue{StringVal: tr.Name},
+			}, {
+				Name:  "kind",
+				Value: v1.ParamValue{StringVal: "Task"},
+			}},
+		}
+	}
 }
--- a/pkg/apis/pipeline/v1beta1/taskrun_conversion_test.go
+++ b/pkg/apis/pipeline/v1beta1/taskrun_conversion_test.go
@@ -280,6 +280,39 @@ func TestTaskRunConversionFromDeprecated(t *testing.T) {
 				},
 			},
 		},
+	}, {
+		name: "bundle",
+		in: &v1beta1.TaskRun{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "foo",
+				Namespace: "bar",
+			},
+			Spec: v1beta1.TaskRunSpec{
+				TaskRef: &v1beta1.TaskRef{
+					Name:   "test-bundle-name",
+					Bundle: "test-bundle",
+				},
+			},
+		},
+		want: &v1beta1.TaskRun{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "foo",
+				Namespace: "bar",
+			},
+			Spec: v1beta1.TaskRunSpec{
+				TaskRef: &v1beta1.TaskRef{
+					Name: "test-bundle-name",
+					ResolverRef: v1beta1.ResolverRef{
+						Resolver: "bundles",
+						Params: []v1beta1.Param{
+							{Name: "bundle", Value: v1beta1.ParamValue{StringVal: "test-bundle"}},
+							{Name: "name", Value: v1beta1.ParamValue{StringVal: "test-bundle-name"}},
+							{Name: "kind", Value: v1beta1.ParamValue{StringVal: "Task"}},
+						},
+					},
+				},
+			},
+		},
 	}}
 	for _, test := range tests {
 		for _, version := range versions {


JeromeJu:
kind /misc

JeromeJu:
kind /misc

JeromeJu:
kind /misc

JeromeJu:
kind /misc

JeromeJu:
kind /misc

JeromeJu:
kind /misc

Yongxuanzhang:
kind /misc

lbernick (commented):
Instead of serializing/deserializing to metadata, we should instead convert this to a remote reference w/ the bundle resolver, since the bundle resolver will be packaged with pipelines. The reason is that we still need to support v1beta1 for 9 months, but we'd like to swap our stored version to v1 as soon as we can. When we swap the stored version to v1 and use it in the reconciler, it's way easier to use the existing functionality for remote resolution than it is to deserialize the bundle reference from annotations.

lbernick:
I'm not sure though if there's a good way to parse the `name` to pass to the resolver from the bundle string name (i.e. if we can assume the image name works here). maybe @vdemeester can comment on whether that would work?

vdemeester:
> Instead of serializing/deserializing to metadata, we should instead convert this to a remote reference w/ the bundle resolver, since the bundle resolver will be packaged with pipelines. The reason is that we still need to support v1beta1 for 9 months, but we'd like to swap our stored version to v1 as soon as we can. When we swap the stored version to v1 and use it in the reconciler, it's way easier to use the existing functionality for remote resolution than it is to deserialize the bundle reference from annotations.

I am not sure that I agree with this too much or not 😅, mainly because we don't have any precedent yet on this (we never really converted to new features before). But it's probably fine.

> I'm not sure though if there's a good way to parse the `name` to pass to the resolver from the bundle string name (i.e. if we can assume the image name works here). maybe @vdemeester can comment on whether that would work?

I am not sure I understand this question, it should be straightforward. Today we have the following…

```
taskRef:
  bundle: foo.bar/baz # image reference
  name: taskName
```

… that translates to …

```
taskRef:
  resolver: bundles
  params:
  - name: bundle
    value: foo.bar/baz
  - name: name
    value: taskName
  - name: kind
    value: Task
```

The `kind` param should be straightforward to get from the code, the rest is the same.

pkg/apis/pipeline/v1beta1/taskrun_conversion_test.go
```diff
@@ -280,6 +280,40 @@ func TestTaskRunConversionFromDeprecated(t *testing.T) {
...
+				Namespace: "bar",
+			},
+			Spec: v1beta1.TaskRunSpec{
+				TaskRef: &v1beta1.TaskRef{
+					Name:   "test-bundle-name",
+					Bundle: "test-bundle",
+				},
+			},
+		},
+		want: &v1beta1.TaskRun{
```
JeromeJu, line 297: @abayer Could you please help confirm the expected output here in the test? Wondering if we would want to introduce extra Params when converting back to v1beta1.

pkg/apis/pipeline/v1beta1/taskref_conversion.go
```diff
@@ -10,18 +10,51 @@ func (tr TaskRef) convertTo(ctx context.Context, sink *v1.TaskRef) {
 	sink.Name = tr.Name
 	sink.Kind = v1.TaskKind(tr.Kind)
 	sink.APIVersion = tr.APIVersion
-	// TODO: handle bundle in #4546
 	new := v1.ResolverRef{}
 	tr.ResolverRef.convertTo(ctx, &new)
 	sink.ResolverRef = new
+	tr.convertBundleToResolver(sink)
```
JeromeJu, line 16: May I ask if it's correct to put `convertBundletoResolver` after conversion for `ResolverRef` as there should be only 1 ResolverRef?

lbernick: we should have validation preventing a user from creating a taskref with both bundle and resolver, but if you'd like you can use conditionals here to ensure we don't convert both

pkg/apis/pipeline/v1beta1/taskrun_conversion_test.go
lbernick: This output is an invalid TaskRef-- you can't specify both resolver and bundle. I think we have 2 options here:

1. Convert all v1 bundle resolverrefs back to oci bundle syntax. This means while v1beta1 is the stored version, if the user creates a v1 CRD, the thing fetching bundles will be the oci bundle code even if the user specified a resolver ref. If the user creates a v1beta1 CRD, the resolver ref syntax will stay as is. I don't love this solution because it will result in different behavior depending on whether the user created a v1beta1 or v1 CRD (while v1beta1 is the stored version). AFAIK the bundle resolver doesn't have any additional features compared to the oci bundle syntax, but it might be more performant.

2. Turn on bundle resolution by default (i.e. remove the feature flag `enable-bundles-resolver`) and don't convert v1 resolver refs back to bundles. I prefer this solution but it might be weird to have the bundles resolver turned on by default while others are not.

@abayer I'd also like to hear your thoughts here

pkg/apis/pipeline/v1beta1/taskref_conversion.go
```diff
@@ -10,18 +10,51 @@ func (tr TaskRef) convertTo(ctx context.Context, sink *v1.TaskRef) {
...
+			Resolver: "bundles",
+			Params: []v1.Param{{
+				Name:  "bundle",
+				Value: v1.ParamValue{StringVal: tr.Bundle},
+			}, {
+				Name:  "name",
+				Value: v1.ParamValue{StringVal: tr.Name},
+			}, {
+				Name:  "kind",
+				Value: v1.ParamValue{StringVal: tr.Name},
```
lbernick, line 42: `kind` here should be "Task"-- that's the thing being resolved

lbernick (changes requested):

pkg/apis/pipeline/v1beta1/taskrun_conversion_test.go
abayer: `name` and `bundle` should both not be there in the `TaskRef` if `ResolverRef` is populated, yes.

I'd lean towards option 2 - I kinda want to turn on the bundle, hub, cluster, and (current, i.e., anonymous) git resolvers by default in v0.41, as part of promoting remote resolution to beta. That is, I think that we can still reserve the right to make changes to the remote resolution API, but am pretty sure we're locked in on the `pipelineRef`/`taskRef` syntax, so enabling them by default seems fine to me, with the ability to disable them via the feature flags still there.

lbernick: Sounds good. I think we don't need to wait for remote resolution to be in beta and the flag to be removed before merging this PR, since we will be moving remote resolution to beta before swapping the stored version of our crds to v1. So @JeromeJu you can just remove the conversion from resolver -> bundle. Andrew, lmk if any concerns about this

abayer: Works for me!

JeromeJu: Thanks @lbernick and @abayer.

pkg/apis/pipeline/v1beta1/taskref_conversion.go
```diff
@@ -10,18 +10,36 @@ func (tr TaskRef) convertTo(ctx context.Context, sink *v1.TaskRef) {
...
 	tr.Name = source.Name
 	tr.Kind = TaskKind(source.Kind)
 	tr.APIVersion = source.APIVersion
-	// TODO: handle bundle in #4546
 	new := ResolverRef{}
 	new.convertFrom(ctx, source.ResolverRef)
 	tr.ResolverRef = new
 }
+
+// convertBundleToResolver converts v1beta1 bundle string to a remote reference with the bundle resolver in v1.
```
lbernick, line 28: Can you add a comment about why we are not doing the reverse conversion?

lbernick (approved):
thanks Jerome!

pkg/apis/pipeline/v1beta1/taskref_conversion.go
JeromeJu: Thank you Lee for the comment. Updated.

PR #5387, opened 2022-08-27
Support parameter substitution for resolver params

abayer:
# Changes

Part of #4710

Until we do `v1beta1` of `ResolutionRequest`, it will still have parameters of `map[string]string`, so we're only doing limited substitution.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1/taskref_validation.go
+++ b/pkg/apis/pipeline/v1/taskref_validation.go
@@ -18,6 +18,7 @@ package v1
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/version"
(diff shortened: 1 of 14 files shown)


lbernick:
I'm a bit confused-- we're allowing substitution of array and object params in resolver ref, but when they're passed to the resolution request we can only do string params? it seems like it would make more sense to not allow array and object params in resolver ref if we can't pass them to resolution request.

abayer:
Yeah, I'm planning to add support for array/object params to `ResolutionRequest` in `v1beta1`, so I debated whether to block them for now or just shrug and only pass on strings. I opted for the latter, since it's fairly short term, but am willing to be persuaded. =)

abayer:
> what's the timeline for v1beta1? if v1beta1 resolutionrequest won't be in the same release I'd definitely say reject. If they'll be in the same release it would still be good for the commit to be self consistent, but I wouldn't push back as much

So resolution will be fully merged for v0.40, but I plan to let it wait til v0.42 (in November) for `v1beta1` to give it a month in the wild to make sure no critical issues pop up. I'll rework this to reject array/object params shortly.

abayer:
@lbernick Updated!

lbernick (commented):
/lgtm

PR #5389, opened 2022-08-29
TEP-0114: Timeout the Testing Wait Custom Task

XinruZhang:
Implements Timeout of  the Testing Wait Custom Task

Prior to this PR, the testing custom task `wait-task`, under `test/wait-task`,
can only wait for a duration of time. This PR adds a new functionality
`timeout` to its controller, so that we can test the behavior of Custom
Task Run time out.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```


abayer:
/retest

Docker Hub is having issues.

test/wait-task/pkg/reconciler/reconciler.go
```diff
@@ -67,27 +74,67 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, r *v1alpha1.Run) kreconc
...
+			r.Status.MarkRunFailed("ParentPipelineRunTimedOut", WaitTaskCancelledByPipelineRunTimeoutMsg)
+		}
 		return nil
 	}
 
-	if r.Status.StartTime == nil {
-		now := metav1.Now()
-		r.Status.StartTime = &now
+	if !r.HasStarted() {
+		logger.Info("Run hasn't stated, start it")
```
lbernick, line 90:
```suggestion
		logger.Info("Run hasn't started, start it")
```

test/wait-task/pkg/reconciler/reconciler_test.go
```diff
@@ -18,149 +18,247 @@ package reconciler
...
+				filterCondition,
+				filterRunStatus,
+			); d != "" {
+				t.Errorf("-got +want: %v", d)
 			}
 		})
 	}
 }
+
+func mustParseRun(t *testing.T, yaml string) *v1alpha1.Run {
```
lbernick, line 251: this function exists [here](https://github.com/tektoncd/pipeline/blob/cd8f4f07d033c51d059aee65d20ccb95b709692b/test/parse/yaml.go#L38)

test/wait-task/pkg/reconciler/reconciler.go
```diff
@@ -67,27 +74,67 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, r *v1alpha1.Run) kreconc
...
-		r.Status.MarkRunFailed("InvalidDuration", "The duration param was invalid: %v", err)
+	// Skip if the Run is cancelled.
+	if r.IsCancelled() {
+		logger.Infof("The Custom Task Run %v has been cancelled", r.GetName())
+		// Check if the Run has been cancelled by its Parent PipelineRun
+		if r.Spec.Status == v1alpha1.RunSpecStatusCancelled && r.Spec.StatusMessage == v1alpha1.RunCancelledByPipelineMsg {
+			now := metav1.Now()
+			r.Status.CompletionTime = &now
+			r.Status.MarkRunFailed("ParentPipelineRunTimedOut", WaitTaskCancelledByPipelineRunTimeoutMsg)
+		}
```
lbernick, line 85: I think the status message should just be propagated to the run status-- runs don't need to know about pipelineruns

test/wait-task/pkg/reconciler/reconciler.go
```diff
@@ -67,27 +74,67 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, r *v1alpha1.Run) kreconc
...
+	dur, err := time.ParseDuration(expr.Value.StringVal)
+	if err != nil {
+		r.Status.MarkRunFailed("InvalidDuration", "The duration param was invalid: %v", err)
+		return nil
+	}
 	done := r.Status.StartTime.Time.Add(dur)
 
-	if time.Now().After(done) {
+	to := time.Minute * config.DefaultTimeoutMinutes
+	if r.GetTimeout() != 0 {
```
lbernick, line 103: I think `GetTimeout` already handles defaults

test/wait-task/pkg/reconciler/reconciler.go
```diff
@@ -67,27 +74,67 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, r *v1alpha1.Run) kreconc
...
+		to, err = time.ParseDuration(r.GetTimeout().String())
+		if err != nil {
+			r.Status.MarkRunFailed("InvalidTimeOut", "Spec.Timeout was invalid: %v", err)
```
lbernick, line 106: possibly worth validating that the timeout is not equal to the duration?

test/wait-task/pkg/reconciler/reconciler.go
```diff
@@ -67,27 +74,67 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, r *v1alpha1.Run) kreconc
...
+			return nil
+		}
+	}
+	timeout := r.Status.StartTime.Time.Add(to)
+
+	// Custom Task is running and not timed out
+	if r.Status.StartTime != nil && time.Now().Before(done) && time.Now().Before(timeout) {
```
lbernick, line 113: instead of using time.Now, it might make sense to use Clock like the other reconcilers do-- this makes it easy to provide a fake clock for unit testing.

lbernick (commented):
I think we should refer to this as the "testing" custom task controller or something similar-- this will help distinguish it from the wait task in experimental that users can install independently and make it clear that the sole purpose of this controller is for testing

test/wait-task/pkg/reconciler/reconciler.go
XinruZhang: Oh right, thx!

test/wait-task/pkg/reconciler/reconciler.go
XinruZhang: Great point, thx!

test/wait-task/pkg/reconciler/reconciler.go
XinruZhang: SG, updated

test/wait-task/pkg/reconciler/reconciler_test.go
XinruZhang: yeah I copied it from the file. This reconciler_test.go belongs to wait-task module while the yaml helpers is outside of this wait-task/ and belongs to pipeline. Would it be better to make the two modules decoupled as much as possible?

lbernick: I think in this instance it's better to use the helper than have duplicate code. I don't see a benefit to having the wait task reconciler use separate code from the regular reconciler, as this wait task reconciler is used only in tekton pipelines for integration testing (can't be installed separately).

XinruZhang: makes sense, updated

XinruZhang:
> I think we should refer to this as the "testing" custom task controller or something similar-- this will help distinguish it from the wait task in experimental that users can install independently and make it clear that the sole purpose of this controller is for testing

Updated the PR Title to clarify it.

test/wait-task/pkg/reconciler/reconciler.go
```diff
@@ -67,27 +73,56 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, r *v1alpha1.Run) kreconc
...
 
-	done := r.Status.StartTime.Time.Add(dur)
+	dur, err := time.ParseDuration(expr.Value.StringVal)
+	if err != nil {
+		r.Status.MarkRunFailed("InvalidDuration", "The duration param was invalid: %v", err)
+		return nil
+	}
+	to := r.GetTimeout()
+	if dur == to {
+		r.Status.MarkRunFailed("InvalidTimeOut", "Spec.Timeout shouldn't equal to duration")
```
lbernick, line 98:
```suggestion
		r.Status.MarkRunFailed("InvalidTimeOut", "Spec.Timeout shouldn't equal duration")
```

test/wait-task/pkg/reconciler/reconciler.go
```diff
@@ -67,27 +73,56 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, r *v1alpha1.Run) kreconc
...
 	}
 
-	dur, err := time.ParseDuration(expr.Value.StringVal)
-	if err != nil {
-		r.Status.MarkRunFailed("InvalidDuration", "The duration param was invalid: %v", err)
+	// Skip if the Run is cancelled.
+	if r.IsCancelled() {
+		logger.Infof("The Custom Task Run %v has been cancelled", r.GetName())
+		r.Status.CompletionTime = &metav1.Time{Time: c.Clock.Now()}
+		r.Status.MarkRunFailed("Cancelled", "The Wait Task is cancelled")
```
lbernick, line 80:
```suggestion
                var reason string
                if r.spec.StatusMessage == "" {
                    reason = "Cancelled"
                } else {
                    reason = r.Spec.StatusMessage
                }
		r.Status.MarkRunFailed(reason, "The Wait Task is cancelled")
```
although I'm not sure if `StatusMessage` should be the "reason" or the "message"-- maybe check what the taskrun reconciler does

lbernick (approved):

test/wait-task/pkg/reconciler/reconciler.go
XinruZhang: https://github.com/tektoncd/pipeline/blob/af4d0e6eb018a34c950180939a7f6a2f24941da6/pkg/apis/pipeline/v1beta1/taskrun_types.go#L100-L106

The `StatusMessage` is message, do you think distinguishing the different cancellation is a better idea?

lbernick: Ah OK in that case, I would do `r.Status.MarkRunFailed("Cancelled", r.spec.StatusMessage)` (if status message != "")-- does that seem reasonable?

XinruZhang: SG! Thanks for catching this~ updated!

test/wait-task/pkg/reconciler/reconciler.go
XinruZhang: 😅 thx! done.

abayer:
Could you rename `test/wait-task` to `test/custom-task-for-testing` or `test/testing-wait-task` or whatever as well?

XinruZhang:
> Could you rename `test/wait-task` to `test/custom-task-for-testing` or `test/testing-wait-task` or whatever as well?

@abayer  Thanks for your comment! I feel the naming `testing` under folder `test/` is a little bit redundant, but I guess I get your point here, it can be better if we clarify that this is a custom task controller. What about creating a new folder called `custom-task-controllers`, `custom-task-ctrls` for short, and moving `wait-task` in it?

test/custom-task-ctrls/wait-task/pkg/reconciler/reconciler.go
```diff
@@ -67,27 +74,60 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, r *v1alpha1.Run) kreconc
...
 		r.Status.MarkRunRunning("Running", "Waiting for duration to elapse")
 	}
 
-	done := r.Status.StartTime.Time.Add(dur)
+	dur, err := time.ParseDuration(expr.Value.StringVal)
+	if err != nil {
+		r.Status.MarkRunFailed("InvalidDuration", "The duration param was invalid: %v", err)
+		return nil
+	}
+	to := r.GetTimeout()
```
jerop, line 101: "The basic rule: the further from its declaration that a name is used, the more descriptive the name must be." - https://github.com/golang/go/wiki/CodeReviewComments#variable-names

`to` is used far from its declaration so it's better to call it `timeout`

```suggestion
	timeout := r.GetTimeout()
```

test/custom-task-ctrls/wait-task/pkg/reconciler/reconciler.go
```diff
@@ -67,27 +74,60 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, r *v1alpha1.Run) kreconc
...
-		r.Status.StartTime = &now
+	if !r.HasStarted() {
+		logger.Info("Run hasn't started, start it")
+		r.Status.InitializeConditions()
+		r.Status.StartTime = &metav1.Time{Time: c.Clock.Now()}
 		r.Status.MarkRunRunning("Running", "Waiting for duration to elapse")
 	}
 
-	done := r.Status.StartTime.Time.Add(dur)
+	dur, err := time.ParseDuration(expr.Value.StringVal)
```
jerop, line 96: "The basic rule: the further from its declaration that a name is used, the more descriptive the name must be." - https://github.com/golang/go/wiki/CodeReviewComments#variable-names

`dur` is used far from its declaration so it's better to call it `duration`

```suggestion
	duration, err := time.ParseDuration(expr.Value.StringVal)
```

jerop (approved):

test/custom-task-ctrls/wait-task/pkg/reconciler/reconciler.go
XinruZhang: SG, updated!

test/custom-task-ctrls/wait-task/pkg/reconciler/reconciler.go
XinruZhang: Done

Issue #5158, opened 2022-07-18
TEP-0114: Custom Tasks Beta - Documentation


jerop:
In [TEP-0114: Custom Tasks Beta][tep-0114], we proposed migrating `Custom Tasks` and `Runs` to beta. 

Ahead of the promotion, we agreed to expand the documentation for `Custom Tasks` and `Runs`. This issue tracks the documentation work. 

[tep-0114]: https://github.com/tektoncd/community/blob/main/teps/0114-custom-tasks-beta.md

XinruZhang:
To extend the TODOs listed above, we need to
- [ ] Migration guide for v1alpha1.Run --> v1beta1.CustomRun
  - [ ] Prepare a migration doc, refer to the migration doc in the runs.md and pipelines.md
  - [ ] Remove illustration about how to use alpha run
  - [ ] Add illustration about how to use beta customrun
    - [ ] Update [runs.md](https://github.com/tektoncd/pipeline/blob/main/docs/runs.md) e.g. adding missing fields in "configuration" section
    - [ ] Update [pipelines.md#using-custom-tasks](https://github.com/tektoncd/pipeline/blob/main/docs/pipelines.md#using-custom-tasks) e.g. providing practical examples with references to end-to-end examples that use wait custom task
- [ ] UX improvement
  - [ ] Add a table of known Custom Tasks to [runs.md](https://github.com/tektoncd/pipeline/blob/main/docs/runs.md) for visibility - can get started with the list of eight in [TEP-0114](https://github.com/tektoncd/community/blob/main/teps/0114-custom-tasks-beta.md#motivation)

PR #5390, opened 2022-08-29
TEP-0114: Updating runs.md Documentation

vsinghai:
# Changes
Prior to this commit, the previous iteration
of the runs.md doc had some fields missing from
the API and the level of detail and clarification.

In this commit, we aim to clarify the API and update
it to make sure it is clear, upto date, and has useful
examples.

Fixes issue [#5158](https://github.com/tektoncd/pipeline/issues/5158)

/cc @jerop
/kind documentation

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/runs.md
+++ b/docs/runs.md
@@ -54,12 +54,16 @@ A `Run` definition supports the following fields:
 - Optional:
   - [`params`](#specifying-parameters) - Specifies the desired execution
     parameters for the custom task.
+  - [`status`](#) - Used for cancelling a `Run`.
+  - [`statusMessage`](#) - Status message for cancellation.
+  - [`retries`](#) - Specifies propagating retries count to custom tasks.
   - [`serviceAccountName`](#specifying-a-serviceaccount) - Specifies a `ServiceAccount`
     object that provides custom credentials for executing the `Run`.
-  - [`workspaces`](#specifying-workspaces) - Specifies the physical volumes to use for the
-    [`Workspaces`](workspaces.md) required by a custom task.
   - [`podTemplate`](#specifying-a-pod-template) - Specifies a [`Pod` template](podtemplates.md) to use
     to configure pods created by the custom task.
+  - [`timeout`](#) - Time after which the custom-task times out.
+  - [`workspaces`](#specifying-workspaces) - Specifies the physical volumes to use for the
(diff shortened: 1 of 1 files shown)


docs/runs.md
```diff
@@ -54,12 +54,16 @@ A `Run` definition supports the following fields:
 - Optional:
   - [`params`](#specifying-parameters) - Specifies the desired execution
     parameters for the custom task.
+  - [`status`](#) - Used for cancelling a `Run`.
```
lbernick, line 57: I think a few of these links are broken.

docs/runs.md
```diff
@@ -54,12 +54,16 @@ A `Run` definition supports the following fields:
...
+  - [`statusMessage`](#) - Status message for cancellation.
+  - [`retries`](#) - Specifies propagating retries count to custom tasks.
```
lbernick, line 59:
```suggestion
  - [`retries`](#) - Specifies the number of times the `Run` should retry, if the custom task controller supports retries.
```

lbernick (commented):
Thanks Varun! This is great to have, but I don't think we should close #5158 when this is merged (I think "fixes #" in your PR description will [result in that happening](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)).

I think there's a lot more documentation for runs that would be nice to have, such as realistic examples of ci/cd pipelines containing runs and links to useful existing custom tasks. @jerop maybe worth updating the issue with some ideas, and possibly @geriom you might have some thoughts here as well? Of course, out of scope for this PR.

vsinghai:
Thanks for the feedback @lbernick! There is still currently a lot that I have left to do for this PR, including (but not limited to) having examples for the custom fields and providing more clarity to the API. The reason I created a PR was so that @jerop can just look over the incremental progress that I am making to make sure we are on the same path regarding how to fix #5158 . I apologize for the confusion.

lbernick:
no problem, thanks for the explanation! I think creating a draft PR can be useful in situations like these- it lets reviewers know that it's still a WIP (hold does this to some extent but I wasn't sure why the PR was on hold)

PR #5391, opened 2022-08-29
Add combined remote resolvers binary

abayer:
# Changes

Part of #4710

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


vdemeester (commented):
Few comments
- Any reason for not making the `bundle` resolver disabled (even though it would be enabled by default) ? I think one of the main reason for resolvers is to be able to selectively decide which resolvers are enable and which are not (https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md)
- Naive question : any reason to go for several feature-flags instead of one called `enable-resolvers: []` that would take a list of resolvers ? It might be harder to document and validate maybe, but it could be nice as well.

vdemeester:
> > * Any reason for not making the `bundle` resolver disabled (even though it would be enabled by default) ? I think one of the main reason for resolvers is to be able to selectively decide which resolvers are enable and which are not (https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md)
> 
> It's got a feature flag - it's just reusing the existing `enable-tekton-oci-bundles` one rather than getting a new one.
> 
> > * Naive question : any reason to go for several feature-flags instead of one called `enable-resolvers: []` that would take a list of resolvers ? It might be harder to document and validate maybe, but it could be nice as well.
> 
> Good question. I'd lean towards `enable-foo-resolver` flags rather than a list - not as ergonomic, maybe, but much simpler for users since they don't need to know the exact name of the resolver, they just need to change a flag from `false` to `true`. But I'm persuadable. =)

Fair enough 🙃 then I would lean to add `enable-bundle-resolver` as `enable-tekton-oci-bundles` will be deprecated 👼🏼 Just to be more consistent.

Code LGTM 👼🏼 
/approve

lbernick:
thanks Andrew! would you mind updating the commit message to say that the hub resolver, git resolver, and bundle resolver are being added, and some detail on the feature flags being used? happy to LGTM once vincent's comment is resolved

abayer:
@lbernick Done!

lbernick:
hmm @abayer I thought you fixed the cloud events flake? it's back 😢

abayer:
/retest

@lbernick It's actually a different flake in the same test. =) The previous ones were duplicate events, this is a more general events-related flake timing out waiting for all the events to be delivered. It popped up sometimes before the duplicate event fix, and I've seen it show up on `TestSendKubernetesEvents` too. It's less frequent than the duplicate events flakes we used to see on cloud events, at least...

lbernick:
ah ok sg-- also it looks like your commit message refers to `enable-tekton-oci-bundles` instead of `enable-bundle-resolver`

abayer:
> ah ok sg-- also it looks like your commit message refers to `enable-tekton-oci-bundles` instead of `enable-bundle-resolver`

Whoops, fixing - `enable-tekton-oci-bundles` will still work, but yeah, the docs etc should reference `enable-bundle-resolver`.

abayer:
@lbernick Fixed, and switched to `enable-bundles-resolver` since `bundles` is the `resolver` value for the resolver, so might as well be consistent.

PR #5393, opened 2022-08-30
TEP-0114: Clarify the Behavior of Custom Task Retries

XinruZhang:
# Changes

Prior to this PR, it is not clear enough how retries works when
the custom task run is timeout. This PR refines the developer
guide for custom controllers supporting `retries` to clarify it.

/kind doc

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/docs/runs.md
+++ b/docs/runs.md
@@ -224,11 +224,8 @@ Supporting retries is optional but recommended.
 
 #### Developer guide for custom controllers supporting `retries`
 
-1. Tekton controllers will use the entire `timeout` duration for exhausting
-    all the retries the custom task is configured to run with. In other words,
-    tekton does not discriminate, if the failure was due to a stuck up process or
-    exhausting all the retries. Custom task developers are recommended to use the
-    same strategy for implementing their timeout for the custom task.
+1. Timeout is for each Custom Task Run retries, meaning if retry balance is
+    not 0 and the run fails because of timeout, it should be retried.
 2. Those custom task who do not wish to support retry, can simply ignore it.
 3. It is recommended, that custom task should update the field `RetriesStatus`
    of a `Run` on each retry performed by the custom task.


docs/runs.md
```diff
@@ -224,11 +224,9 @@ Supporting retries is optional but recommended.
...
 #### Developer guide for custom controllers supporting `retries`
 
-1. Tekton controllers will use the entire `timeout` duration for exhausting
-    all the retries the custom task is configured to run with. In other words,
-    tekton does not discriminate, if the failure was due to a stuck up process or
-    exhausting all the retries. Custom task developers are recommended to use the
-    same strategy for implementing their timeout for the custom task.
+1. Timeout is configured as the maximum running time of a Custom Task Run,
+    meaning if a run is timeout while retires havn't been exhaused, developers
+    should fail the custom task.
```
Yongxuanzhang, line 229:
```suggestion
1. Timeout is configured as the maximum running time of a Custom Task Run,
    meaning if a run is timeout while retries haven't been exhausted, developers
    should fail the custom task.
```

XinruZhang: Oppps, thx! updated

lbernick:
I don't think it's true that `timeout` should be considered for the sum of retries. The timeout value passed to runs comes from the Pipeline Task timeout. For TaskRuns, we've decided that the pipeline task timeout should apply to each retry independently, and the `tasks` and `pipeline` timeouts on the PipelineRun take into account time spent during retries. We should be consistent between TaskRuns and Runs here.

lbernick:
Although I guess if we've been documenting that it applies to all retries people have already been implementing it in this way? FYI @jerop there's some inconsistency here between taskruns and runs

lbernick:
Thanks @XinruZhang for the updates! my only concern is that even though this is a documentation change, it's sort of backwards incompatible because it reflects the contract we expect run controllers to implement (defined in [TEP-0069](https://github.com/tektoncd/community/blob/main/teps/0069-support-retries-for-custom-task-in-a-pipeline.md#proposal)). I think it's worth getting some more opinions on whether we'd prefer to keep the run reconciler contract as is, or whether we'd prefer consistency between TaskRuns and Runs. In the meantime, it's worth documenting that the Pipeline Task timeout is applied independently to each retry of a TaskRun.

dibyom:
>  I think it's worth getting some more opinions on whether we'd prefer to keep the run reconciler contract as is, or whether we'd prefer consistency between TaskRuns and Runs. In the meantime, it's worth documenting that the Pipeline Task timeout is applied independently to each retry of a TaskRun.

I think we should at least mark this as a breaking change with an action-required - probably worth sending to the mailing lists as well if we decide to go ahead

lbernick:
we could probably just discuss here-- @tektoncd/core-maintainers

XinruZhang:
Though I do think it would be better if we keep the behaviors of Runs & TaskRuns aligned, then the action item would be what's the process of making breaking changes to Custom Task Run.

XinruZhang:
Opened a TEP (TEP-0121 https://github.com/tektoncd/community/pull/816) to fully discuss the behavior of retries, we can continue our discussions there.

XinruZhang:
As discussed in the [Tekton API Working Group](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#bookmark=id.jmjb8ox1pxln) today, we'd like to modify the development guidance of Custom Task, suggesting users to set timeout for each retry attempt.

XinruZhang:
I'm closing this PR because we've made changes to `v1beta1.CustomRun`. We can leave it as what it is for `v1alpha1.Run`. Please feel free to comment if there's any other concerns.

PR #5396, opened 2022-08-30
Fix TestYamls for change in `ko create`

abayer:
# Changes

With https://github.com/google/ko/pull/750, which we picked up when we bumped `ko` to `v0.12.0` in the test-runner image, `-n (namespace)` is no longer an argument for `ko create`. So we need to ensure that the namespace for the `TestYamls` CRDs is actually set properly, and get rid of the `-n (namespace)` from our invocation of `ko create`.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/examples_test.go
+++ b/test/examples_test.go
@@ -77,8 +77,7 @@ func substituteEnv(input []byte, namespace string) ([]byte, error) {
 	}
 	output := defaultKoDockerRepoRE.ReplaceAll(input, []byte(val))
 
-	// Strip any "namespace: default"s, all resources will be created in
-	// the test namespace using `ko create -n`
+	// Replace any "namespace: default"s with the test namespace.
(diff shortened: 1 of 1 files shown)


vdemeester (commented):
Why not using « ko create ... -- --namespace=foo » instead ? I think it is the supported way isn’t it ?

abayer:
> Why not using « ko create ... -- --namespace=foo » instead ? I think it is the supported way isn’t it ?

Huh, didn't think of that. Well, let's see if this works anyway? =)

abayer:
Well, that worked, but lemme try the `-- -namespace (ns)` approach...

vdemeester (approved):

afrittoli (commented):
Thanks!
/approve

PR #5397, opened 2022-08-30
Populate the `*Run.Status.Provenance.ConfigSource` field

chuangw6:
# Changes

***Before:*** 
Prior, remote `ResolutionRequest` CRD supports **recording** the source information
in its Status that identifies where the remote resource came from.
Similarly, `TaskRun/PipelineRun` CRD supports **receiving** the source information
via a field in its status from remote ResolutionRequest. Specifically,
this field named `ConfigSource` is a subfield of the `Provenance` field in status.

***Now:*** 
In this PR, we are trying to pass the data from `ResolutionRequest` to pipeline
reconciler so that the data can be captured in TaskRun/PipelineRun's `Status.Provenance.ConfigSource`.
The implication of this change is that many functions' interface has been
changed because we are passing this extra source data alongside the remote resoure.

***Note:***
- The `provenance` field in Run.status is behind a feature flag named `enable-provenance-in-status`
, which was introduced in https://github.com/tektoncd/pipeline/pull/5670. The field will be populated iff
the flag is set to `true`.
- If a pipeline yaml is from remote place A, and the individual tasks
are from other remote places, pipelinerun status will only record the source
for the pipeline, and the individual taskrun status will record the
source for the corresponding task.

***Related PRs/Issues:***
- https://github.com/tektoncd/pipeline/pull/5580
- https://github.com/tektoncd/pipeline/pull/5551
- https://github.com/tektoncd/pipeline/pull/5670
- https://github.com/tektoncd/pipeline/issues/5522

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Populate the TaskRun/PipelineRun's Status.Provenance.ConfigSource field with the value from the remote ResolutionRequest Status. 

Note: the feature flag `enable-provenance-in-status` needs to be set to "true" to enable this provenance field to be populated & available in *Run.Status.
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -40,6 +40,7 @@ import (
 	listersv1alpha1 "github.com/tektoncd/pipeline/pkg/client/listers/pipeline/v1alpha1"
 	listers "github.com/tektoncd/pipeline/pkg/client/listers/pipeline/v1beta1"
 	resourcelisters "github.com/tektoncd/pipeline/pkg/client/resource/listers/resource/v1alpha1"
+	resolutionutil "github.com/tektoncd/pipeline/pkg/internal/resolution"
 	"github.com/tektoncd/pipeline/pkg/matrix"
 	"github.com/tektoncd/pipeline/pkg/pipelinerunmetrics"
 	tknreconciler "github.com/tektoncd/pipeline/pkg/reconciler"
@@ -392,7 +393,7 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
 		return controller.NewPermanentError(err)
 	default:
 		// Store the fetched PipelineSpec on the PipelineRun for auditing
-		if err := storePipelineSpecAndMergeMeta(pr, pipelineSpec, pipelineMeta); err != nil {
+		if err := storePipelineSpecAndMergeMeta(ctx, pr, pipelineSpec, pipelineMeta); err != nil {
 			logger.Errorf("Failed to store PipelineSpec on PipelineRun.Status for pipelinerun %s: %v", pr.Name, err)
 		}
 	}
@@ -522,7 +523,7 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
 	if len(pipelineSpec.Finally) > 0 {
 		tasks = append(tasks, pipelineSpec.Finally...)
 	}
-	pipelineRunState, err := c.resolvePipelineState(ctx, tasks, pipelineMeta, pr, providedResources)
+	pipelineRunState, err := c.resolvePipelineState(ctx, tasks, pipelineMeta.ObjectMeta, pr, providedResources)
 	switch {
 	case errors.Is(err, remote.ErrorRequestInProgress):
 		message := fmt.Sprintf("PipelineRun %s/%s awaiting remote resource", pr.Namespace, pr.Name)
@@ -1234,10 +1235,13 @@ func (c *Reconciler) updateLabelsAndAnnotations(ctx context.Context, pr *v1beta1
 	return newPr, nil
 }
 
-func storePipelineSpecAndMergeMeta(pr *v1beta1.PipelineRun, ps *v1beta1.PipelineSpec, meta *metav1.ObjectMeta) error {
+func storePipelineSpecAndMergeMeta(ctx context.Context, pr *v1beta1.PipelineRun, ps *v1beta1.PipelineSpec, meta *resolutionutil.ResolvedObjectMeta) error {
 	// Only store the PipelineSpec once, if it has never been set before.
 	if pr.Status.PipelineSpec == nil {
 		pr.Status.PipelineSpec = ps
+		if meta == nil {
+			return nil
+		}
 
 		// Propagate labels from Pipeline to PipelineRun. PipelineRun labels take precedences over Pipeline.
 		pr.ObjectMeta.Labels = kmap.Union(meta.Labels, pr.ObjectMeta.Labels)
@@ -1247,6 +1251,19 @@ func storePipelineSpecAndMergeMeta(pr *v1beta1.PipelineRun, ps *v1beta1.Pipeline
 		pr.ObjectMeta.Annotations = kmap.Union(meta.Annotations, pr.ObjectMeta.Annotations)
 
 	}
+
+	// Propagate ConfigSource from remote resolution to PipelineRun Status
+	// This lives outside of the status.spec check to avoid the case where only the spec is available in the first reconcile and source comes in next reconcile.
+	cfg := config.FromContextOrDefaults(ctx)
+	if cfg.FeatureFlags.EnableProvenanceInStatus && meta != nil && meta.ConfigSource != nil {
+		if pr.Status.Provenance == nil {
+			pr.Status.Provenance = &v1beta1.Provenance{}
+		}
+		if pr.Status.Provenance.ConfigSource == nil {
+			pr.Status.Provenance.ConfigSource = meta.ConfigSource
+		}
+	}
+
 	return nil
 }
 
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -37,6 +37,7 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	resourcev1alpha1 "github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1"
+	resolutionutil "github.com/tektoncd/pipeline/pkg/internal/resolution"
 	"github.com/tektoncd/pipeline/pkg/reconciler/events/cloudevent"
 	"github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun/resources"
 	ttesting "github.com/tektoncd/pipeline/pkg/reconciler/testing"
@@ -5523,7 +5524,7 @@ status:
 	}
 }
 
-func Test_storePipelineSpec(t *testing.T) {
+func Test_storePipelineSpecAndConfigSource(t *testing.T) {
 	pr := parse.MustParseV1beta1PipelineRun(t, `
 metadata:
   name: test-pipeline-run-success
@@ -5532,6 +5533,13 @@ metadata:
   annotations:
     io.annotation: value
 `)
+	configSource := &v1beta1.ConfigSource{
+		URI: "abc.com",
+		Digest: map[string]string{
+			"sha1": "a123",
+		},
+		EntryPoint: "foo/bar",
+	}
 
 	ps := v1beta1.PipelineSpec{Description: "foo-pipeline"}
 	ps1 := v1beta1.PipelineSpec{Description: "bar-pipeline"}
@@ -5540,24 +5548,75 @@ metadata:
 	want.Status = v1beta1.PipelineRunStatus{
 		PipelineRunStatusFields: v1beta1.PipelineRunStatusFields{
 			PipelineSpec: ps.DeepCopy(),
+			Provenance: &v1beta1.Provenance{
+				ConfigSource: configSource.DeepCopy(),
+			},
 		},
 	}
 	want.ObjectMeta.Labels["tekton.dev/pipeline"] = pr.ObjectMeta.Name
 
-	// The first time we set it, it should get copied.
-	if err := storePipelineSpecAndMergeMeta(pr, &ps, &pr.ObjectMeta); err != nil {
-		t.Errorf("storePipelineSpec() error = %v", err)
-	}
-	if d := cmp.Diff(pr, want); d != "" {
-		t.Fatalf(diff.PrintWantGot(d))
+	type args struct {
+		pipelineSpec       *v1beta1.PipelineSpec
+		resolvedObjectMeta *resolutionutil.ResolvedObjectMeta
 	}
 
-	// The next time, it should not get overwritten
-	if err := storePipelineSpecAndMergeMeta(pr, &ps1, &metav1.ObjectMeta{}); err != nil {
-		t.Errorf("storePipelineSpec() error = %v", err)
+	var tests = []struct {
+		name            string
+		reconcile1Args  *args
+		reconcile2Args  *args
+		wantPipelineRun *v1beta1.PipelineRun
+	}{
+		{
+			name: "spec and source are available in the same reconcile",
+			reconcile1Args: &args{
+				pipelineSpec: &ps,
+				resolvedObjectMeta: &resolutionutil.ResolvedObjectMeta{
+					ObjectMeta:   &pr.ObjectMeta,
+					ConfigSource: configSource.DeepCopy(),
+				},
+			},
+			reconcile2Args: &args{
+				pipelineSpec:       &ps1,
+				resolvedObjectMeta: &resolutionutil.ResolvedObjectMeta{},
+			},
+			wantPipelineRun: want,
+		},
+		{
+			name: "spec comes in the first reconcile and source comes in next reconcile",
+			reconcile1Args: &args{
+				pipelineSpec: &ps,
+				resolvedObjectMeta: &resolutionutil.ResolvedObjectMeta{
+					ObjectMeta: &pr.ObjectMeta,
+				},
+			},
+			reconcile2Args: &args{
+				pipelineSpec: &ps,
+				resolvedObjectMeta: &resolutionutil.ResolvedObjectMeta{
+					ConfigSource: configSource.DeepCopy(),
+				},
+			},
+			wantPipelineRun: want,
+		},
 	}
-	if d := cmp.Diff(pr, want); d != "" {
-		t.Fatalf(diff.PrintWantGot(d))
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := ttesting.EnableFeatureFlagField(context.Background(), t, "enable-provenance-in-status")
+			// mock first reconcile
+			if err := storePipelineSpecAndMergeMeta(ctx, pr, tc.reconcile1Args.pipelineSpec, tc.reconcile1Args.resolvedObjectMeta); err != nil {
+				t.Errorf("storePipelineSpec() error = %v", err)
+			}
+			if d := cmp.Diff(pr, tc.wantPipelineRun); d != "" {
+				t.Fatalf(diff.PrintWantGot(d))
+			}
+
+			// mock second reconcile
+			if err := storePipelineSpecAndMergeMeta(ctx, pr, tc.reconcile2Args.pipelineSpec, tc.reconcile2Args.resolvedObjectMeta); err != nil {
+				t.Errorf("storePipelineSpec() error = %v", err)
+			}
+			if d := cmp.Diff(pr, tc.wantPipelineRun); d != "" {
+				t.Fatalf(diff.PrintWantGot(d))
+			}
+		})
 	}
 }
 
@@ -5573,7 +5632,9 @@ func Test_storePipelineSpec_metadata(t *testing.T) {
 		ObjectMeta: metav1.ObjectMeta{Name: "foo", Labels: pipelinerunlabels, Annotations: pipelinerunannotations},
 	}
 	meta := metav1.ObjectMeta{Name: "bar", Labels: pipelinelabels, Annotations: pipelineannotations}
-	if err := storePipelineSpecAndMergeMeta(pr, &v1beta1.PipelineSpec{}, &meta); err != nil {
+	if err := storePipelineSpecAndMergeMeta(context.Background(), pr, &v1beta1.PipelineSpec{}, &resolutionutil.ResolvedObjectMeta{
+		ObjectMeta: &meta,
+	}); err != nil {
 		t.Errorf("storePipelineSpecAndMergeMeta error = %v", err)
 	}
 	if d := cmp.Diff(pr.ObjectMeta.Labels, wantedlabels); d != "" {
--- a/pkg/reconciler/pipelinerun/pipelinespec/pipelinespec.go
+++ b/pkg/reconciler/pipelinerun/pipelinespec/pipelinespec.go
@@ -22,32 +22,37 @@ import (
 	"fmt"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	resolutionutil "github.com/tektoncd/pipeline/pkg/internal/resolution"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // GetPipeline is a function used to retrieve Pipelines.
-type GetPipeline func(context.Context, string) (v1beta1.PipelineObject, error)
+type GetPipeline func(context.Context, string) (v1beta1.PipelineObject, *v1beta1.ConfigSource, error)
 
 // GetPipelineData will retrieve the Pipeline metadata and Spec associated with the
 // provided PipelineRun. This can come from a reference Pipeline or from the PipelineRun's
 // metadata and embedded PipelineSpec.
-func GetPipelineData(ctx context.Context, pipelineRun *v1beta1.PipelineRun, getPipeline GetPipeline) (*metav1.ObjectMeta, *v1beta1.PipelineSpec, error) {
+func GetPipelineData(ctx context.Context, pipelineRun *v1beta1.PipelineRun, getPipeline GetPipeline) (*resolutionutil.ResolvedObjectMeta, *v1beta1.PipelineSpec, error) {
 	pipelineMeta := metav1.ObjectMeta{}
+	var configSource *v1beta1.ConfigSource
 	pipelineSpec := v1beta1.PipelineSpec{}
 	switch {
 	case pipelineRun.Spec.PipelineRef != nil && pipelineRun.Spec.PipelineRef.Name != "":
 		// Get related pipeline for pipelinerun
-		t, err := getPipeline(ctx, pipelineRun.Spec.PipelineRef.Name)
+		p, source, err := getPipeline(ctx, pipelineRun.Spec.PipelineRef.Name)
 		if err != nil {
 			return nil, nil, fmt.Errorf("error when listing pipelines for pipelineRun %s: %w", pipelineRun.Name, err)
 		}
-		pipelineMeta = t.PipelineMetadata()
-		pipelineSpec = t.PipelineSpec()
+		pipelineMeta = p.PipelineMetadata()
+		pipelineSpec = p.PipelineSpec()
+		configSource = source
 	case pipelineRun.Spec.PipelineSpec != nil:
 		pipelineMeta = pipelineRun.ObjectMeta
 		pipelineSpec = *pipelineRun.Spec.PipelineSpec
+		// TODO: if we want to set source for embedded pipeline, set it here.
+		// https://github.com/tektoncd/pipeline/issues/5522
 	case pipelineRun.Spec.PipelineRef != nil && pipelineRun.Spec.PipelineRef.Resolver != "":
-		pipeline, err := getPipeline(ctx, "")
+		pipeline, source, err := getPipeline(ctx, "")
 		switch {
 		case err != nil:
 			return nil, nil, err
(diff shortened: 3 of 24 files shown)


pkg/reconciler/taskrun/resources/taskref_test.go
```diff
@@ -532,11 +534,15 @@ func TestGetTaskFunc_RemoteResolution(t *testing.T) {
 		t.Fatalf("failed to get task fn: %s", err.Error())
 	}
 
-	resolvedTask, err := fn(ctx, taskRef.Name)
+	resolvedTask, annotation, err := fn(ctx, taskRef.Name)
 	if err != nil {
 		t.Fatalf("failed to call pipelinefn: %s", err.Error())
 	}
 
+	if !reflect.DeepEqual(expectedAnnotation, annotation) {
```
abayer, line 542: We tend to prefer `if d := cmp.Diff(a, b); d != "" { t.Errorf("annotations did not match: %s", diff.PrintWantGot(d)) }`

pkg/resolution/resolver/git/annotations.go
```diff
@@ -17,7 +17,10 @@ limitations under the License.
 package git
 
 const (
-	// AnnotationKeyRevision is the commit hash that was fetched
-	// from git
+	// AnnotationKeyURL is the git repo url that was fetched
+	AnnotationKeyURL = "url"
```
abayer, line 21: If we're passing resolver annotations on to the `TaskRun`/`PipelineRun`, we should probably have namespaced annotations, like, say, `resolution.tekton.dev/url`.

chuangw6: Thanks for reviewing the PR @abayer !!

This is a very good idea! If we use namespaced annotation key like `resolution.tekton.dev/xxxx`, are we assured that the annotation will not overlap with or be overrode by remote task/pipeline's annotations?

I am asking because I had some concern about the case where the remote task/pipeline has the annotation field same as the annotations we are adding here. If the namespaced annotation can mitigate the issue. That's great.

Some background: when [merging](https://github.com/tektoncd/pipeline/blob/05f28f2cbc22bd4eafadc34a263637cb16ef00f9/pkg/reconciler/taskrun/taskrun.go#L868-L874) taskspec and metadata to taskrun, taskrun annotations precedences over task in the existing code. Since we are trying to pass through resolver annotations, resolver annotations and remote task annotations are merged first, and then the combined annotations are merged to taskrun.

abayer: The namespacing as `resolution.tekton.dev/xxx` seems like the best way to assure it doesn't get overridden, yeah.

abayer:
Oh, and this should be done for `Pipeline`s, not just `Task`s. I'd do all of that in one PR.

chuangw6:
> Oh, and this should be done for `Pipeline`s, not just `Task`s. I'd do all of that in one PR.

Yeah, that's what I planned to do in this PR. I just did for `Task`s for now to gather some thoughts first.

pkg/resolution/resolver/git/annotations.go
abayer: FYI, I'm going to open a PR (after #5405 merges) to switch all the existing resolver annotations to be namespaced, so you can wait for that if you want. =)

dibyom: The annotation works for a PoC but for the provenance use case,  I think we should put it in the status instead!

chuangw6: SG! I've updated this to make sure the information from gitresolver is stored under `status.annotations` instead of `metadata.annotations` for taskrun. Please take a look.

abayer:
FYI, the git resolver undergoes some significant changes in #5450.

chuangw6:
> FYI, the git resolver undergoes some significant changes in #5450.

Thanks for the heads up @abayer

alhuizenga:
Also, really nicely written PR @chuangw6, even I could understand it!

chuangw6:
> Also, really nicely written PR @chuangw6, even I could understand it!

Thanks @alhuizenga !! Appreciate it!

chuangw6:
@abayer Please take a look since this is closely related to https://github.com/tektoncd/pipeline/pull/5450. Thanks.

Not sure `change 1` is the correct way to get the real url if users only provide a value for `repo` and `org`. In github example, it would be `https://api.github.com/<org>/<repo>.git` Thoughts?

pkg/resolution/resolver/git/annotations.go
chuangw6: PipelineRun level work is done, and all annotations are passed to `status.annotations`. Please take a look @dibyom.

FYI, I think we will also need to make sure the `ResolutionRequest` status is signed with SPIRE integration https://github.com/tektoncd/pipeline/issues/5474.

abayer:
/approve

I'd change the title to reflect that this isn't git resolver specific, since it should be doing this for all resolvers now, right?

chuangw6:
> /approve
> 
> I'd change the title to reflect that this isn't git resolver specific, since it should be doing this for all resolvers now, right?

Thanks for reviewing this PR @abayer !
Correct, this change will be applied to other resolvers. This PR for now only implements the gitresolvers's part. I'll open other PRs to implement that for oci and catalog resolvers.

dibyom:
Also while we are at it, can we update the gitresolver docs to mention the additional status.annotations that it will always provide?

chuangw6:
> @chuangw6 would you mind opening an issue to track this?

Sure! Created here https://github.com/tektoncd/pipeline/issues/5522

chuangw6:
> Also while we are at it, can we update the gitresolver docs to mention the additional status.annotations that it will always provide?

SG! Added to `What's Supported?` section in the git-resolver.md file. Please let me know if there is better place for documenting this. Thanks

wlynch (changes requested):
Swooping in from https://github.com/tektoncd/chains/pull/554, so consider these comments non-blocking if you want to go ahead and submit.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -4907,12 +4907,12 @@ metadata:
   namespace: foo
   labels:
     tekton.dev/pipeline: test-pipeline
-  annotations: {}
+  annotations: null
```
dibyom, line 4910: why is this change necessary?

docs/git-resolver.md
```diff
@@ -145,6 +145,10 @@ spec:
   * Gitea
   * BitBucket Server
   * BitBucket Cloud
+- As users can specify branch name or tag name for the param `revision `, the actual `commit sha` being used at the moment of resolving the resource is recorded in the `status.annotations` of both `ResolutionRequest` and `PipelineRun`/`TaskRun` object along with the `url` and `path`. The annotations keys are
```
dibyom, line 148: Can we put this under its own heading with an example?

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -961,3 +945,20 @@ func isResourceQuotaConflictError(err error) bool {
 	}
 	return k8ErrStatus.Details != nil && k8ErrStatus.Details.Kind == "resourcequotas"
 }
+
+// AppendMap appends the second map to the first map.
+// If the first map is nil, return whatever the second map is.
+// If both maps have a common key, the value in the first map takes precedence.
+func AppendMap(original, added map[string]string) map[string]string {
```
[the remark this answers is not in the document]
dibyom: Can we use kmap.Union: https://pkg.go.dev/knative.dev/pkg@v0.0.0-20220914154704-5f66ecf267fe/kmap#Union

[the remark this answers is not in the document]
chuangw6: @dibyom great point! Thanks for the suggestion!

docs/git-resolver.md
chuangw6: Sure! Added PTAL1

pkg/resolution/resolver/git/resolver.go
```diff
@@ -154,12 +155,14 @@ func (r *Resolver) resolveAPIGit(ctx context.Context, params map[string]string)
...
 		return nil, fmt.Errorf("no content for resource in %s/%s %s", params[orgParam], params[repoParam], params[pathParam])
 	}
+
 	return &resolvedGitResource{
 		Content:  content.Data,
 		Revision: content.Sha,
 		Org:      params[orgParam],
 		Repo:     params[repoParam],
 		Path:     content.Path,
+		URL:      fmt.Sprintf("%s%s/%s.git", scmClient.BaseURL.String(), params[orgParam], params[repoParam]),
```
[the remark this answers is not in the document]
chuangw6: > Be very careful with this, since I'm not sure if this will work with every URL type - e.g. will this match the URL if I give it ssh:// or git://? You may want to just pass through the user provided value.

In https://github.com/tektoncd/pipeline/pull/5450, provider api was introduced to allow users to only provide `repo` and `org` instead of a complete `url`. So what I am trying to do here is to construct the real url based on user input `repo` and `org` because the url is needed in the provenance configSource.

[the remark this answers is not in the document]
chuangw6: I do have some doubts here about what is the right way to get the real url using the provider api. In current implementation, the url for github example is `https://api.github.com/<org>/<repo>.git` if users use Authenticated API
approach by providing `org` and `repo`.

@abayer do you have any idea or better suggestion here? Thanks

pkg/resolution/common/annotations.go
```diff
@@ -22,4 +22,18 @@ const (
...
+	// AnnotationKeyPath is the path used
+	AnnotationKeyPath = resolution.GroupName + "/path"
+
+	// AnnotationKeyRevision is the commit hash that was fetched
+	// from remote location
+	AnnotationKeyRevision = resolution.GroupName + "/revision"
+
+	// AnnotationKeyHashAlgo is the cryptographic hash algorithm being used
+	// to calculate the actual sha
+	AnnotationKeyHashAlgo = resolution.GroupName + "/hashAlgo"
```
chuangw6, line 38: @wlynch RE: https://github.com/tektoncd/chains/pull/554#discussion_r974576447

I am trying to add new annotation here to pass the hash algo from resolver side. And for git source, I just created a constant with value `sha1` to set this annotation. Do you have any other better suggestion please? Thanks

pkg/resolution/resolver/git/resolver.go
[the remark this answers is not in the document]
abayer: What do we intend to use the value of the URL annotation for? Is it supposed to be the clone URL, or a web link? Or just a consistent identifier of the source of the resolved resource? If it's meant to be the clone URL or a web link, then we'd probably want to add 
```
repo, _, err := scmClient.Repositories.Get(ctx, fmt.Sprintf("%s/%s", params[orgParam], params[repoParam]"))
if err != nil { ... }

...

return &resolvedGitResource{
...
  URL: repo.Clone, // or repo.Link if we want the web link
}
```

If it's just meant to be a consistent identifier, then adding a new annotation like `resolution.tekton.dev/server` and using `serverURL` for that with nothing set for `resolution.tekton.dev/url` when using the API-based resolution feels appropriate.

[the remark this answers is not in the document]
chuangw6: Thanks @abayer !! very helpful!
The intention is for SLSA provenance to record the [identifier of the config source](https://slsa.dev/provenance/v0.2#invocation.configSource.uri). I think `repo.Link` should be what is desired here.

pkg/remote/resolver.go
```diff
@@ -31,5 +31,5 @@ type ResolvedObject struct {
 //   - Get:      retrieves a specific object with the given Kind and name.
 type Resolver interface {
 	List(ctx context.Context) ([]ResolvedObject, error)
-	Get(ctx context.Context, kind, name string) (runtime.Object, error)
+	Get(ctx context.Context, kind, name string) (runtime.Object, map[string]string, error)
```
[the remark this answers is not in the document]
chuangw6: Added document for get method! And just created https://github.com/tektoncd/pipeline/issues/5529 to discuss adding structured sourceref to remote `ResolutionRequest`. 

Happy to change that later in another PR if we get agreement on this.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1216,35 +1216,18 @@ func (c *Reconciler) updateLabelsAndAnnotations(ctx context.Context, pr *v1beta1
...
-		for key, value := range meta.Annotations {
-			// Do not override duplicates between PipelineRun and Pipeline
-			// PipelineRun labels take precedences over Pipeline
-			if _, ok := pr.ObjectMeta.Annotations[key]; !ok {
-				pr.ObjectMeta.Annotations[key] = value
-			}
-		}
+		pr.ObjectMeta.Annotations = taskrun.AppendMap(pr.ObjectMeta.Annotations, meta.Annotations)
+		// Propagate annotations from resolutionrequest to PipelineRun's status.
+		pr.Status.Annotations = taskrun.AppendMap(pr.Status.Annotations, annotationsInStatus)
```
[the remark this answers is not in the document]
chuangw6: I am thinking it might be better to do it after the ResolutionRequest supports the ConfigSource https://github.com/tektoncd/pipeline/issues/5529. 

Otherwise, we have to do `annotations from resolution` (map) -> `resolvedObjectMeta.ConfigSource` (struct) -> `annotations in run status` (map).

With https://github.com/tektoncd/pipeline/issues/5529 to be implemented, it makes it easier/cleaner. wdyt?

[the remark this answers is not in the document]
dibyom: +1 to returning a configSource vs a generic map

docs/git-resolver.md
chuangw6: Hey @dibyom ,

I breakdown this PR for easier review, and the change for the git resolver is now in https://github.com/tektoncd/pipeline/pull/5664. I've implemented this feedback in that PR. PTAL. Thank you!!

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
chuangw6: Thank you for the feedback and idea @wlynch @dibyom !

With the structured fields supported in ResolutionRequest.Status and Run.Status, I've implemented the feedback so that the data is passed through those structured fields instead of annotation maps. The suggested `resolvedObjectMeta` is also implemented to pass the metadata together.

PTAL! Thank you very much!

pkg/remote/resolver.go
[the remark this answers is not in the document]
chuangw6: Thank you @wlynch! This has been implemented. PTAL

chuangw6:
Hey @wlynch @dibyom, 

With the structured fields for source information supported in `ResolutionRequest.Status` and `Run.Status`, I've changed this PR to pass the data through the structured fields instead of annotation maps. I also breakdown the PR to make sure this PR only focuses on passing the data and to extract out other changes to other PRs. 

Please take a look. Thank you very much!

chuangw6:
cc @Yongxuanzhang PTAL b/c you mentioned that url information is also desired for trusted resource work. Thank you!

cc @jagathprakash PTAL and let me know if you have any comments/feedback! Thanks!!

chuangw6:
/hold

waiting for the featured flag from https://github.com/tektoncd/pipeline/pull/5670

pkg/reconciler/pipelinerun/resources/pipelineref.go
```diff
@@ -98,28 +104,32 @@ type LocalPipelineRefResolver struct {
 
 // GetPipeline will resolve a Pipeline from the local cluster using a versioned Tekton client. It will
 // return an error if it can't find an appropriate Pipeline for any reason.
-func (l *LocalPipelineRefResolver) GetPipeline(ctx context.Context, name string) (v1beta1.PipelineObject, error) {
+// TODO (@chuangw6): if we want to set source for in-cluster pipeline, set it here.
```
Yongxuanzhang, line 107: It may be better to link TODO to an issue instead of a person

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -777,7 +777,9 @@ func resolveTask(
 			spec = *taskRun.Status.TaskSpec
 			taskName = pipelineTask.TaskRef.Name
 		} else {
-			t, err = getTask(ctx, pipelineTask.TaskRef.Name)
+			// Following minimum status principle, no need to propagate the source about PipelineTask up to PipelineRun status.
```
Yongxuanzhang, line 780: 👍

pkg/reconciler/pipelinerun/resources/pipelineref.go
```diff
@@ -42,36 +42,42 @@ func GetPipelineFunc(ctx context.Context, k8s kubernetes.Interface, tekton clien
...
 	namespace := pipelineRun.Namespace
-	// if the spec is already in the status, do not try to fetch it again, just use it as source of truth
+	// if the spec is already in the status, do not try to fetch it again, just use it as source of truth.
+	// Same for the Source field in the Status.Provenance.
 	if pipelineRun.Status.PipelineSpec != nil {
-		return func(_ context.Context, name string) (v1beta1.PipelineObject, error) {
+		return func(_ context.Context, name string) (v1beta1.PipelineObject, *v1beta1.ConfigSource, error) {
+			var configSource *v1beta1.ConfigSource
+			if pipelineRun.Status.Provenance != nil {
+				configSource = pipelineRun.Status.Provenance.ConfigSource
```
Yongxuanzhang, line 51: This is not tested. Can we test it?

Yongxuanzhang (commented):
Thank you for the work! It looks good to me!

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1253,10 +1253,21 @@ func (c *Reconciler) updateLabelsAndAnnotations(ctx context.Context, pr *v1beta1
...
+func storePipelineSpecAndMergeMeta(pr *v1beta1.PipelineRun, ps *v1beta1.PipelineSpec, meta *tresources.ResolvedObjectMeta) error {
 	// Only store the PipelineSpec once, if it has never been set before.
 	if pr.Status.PipelineSpec == nil {
 		pr.Status.PipelineSpec = ps
+		if meta == nil {
+			return nil
+		}
+
+		// Propagate ConfigSource from remote resolution to PipelineRun Status
+		if meta.ConfigSource != nil {
```
jagathprakash, line 1265: Is there a chance that, this might get populated in a reconcile after the first reconcile, in which case this code will not be reached because pr.Status.PipelineSpec will not be nil.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -5552,20 +5553,38 @@ metadata:
...
+					EntryPoint: "foo/bar",
+				},
+			},
 		},
 	}
 	want.ObjectMeta.Labels["tekton.dev/pipeline"] = pr.ObjectMeta.Name
 
 	// The first time we set it, it should get copied.
-	if err := storePipelineSpecAndMergeMeta(pr, &ps, &pr.ObjectMeta); err != nil {
+	if err := storePipelineSpecAndMergeMeta(pr, &ps, &tresources.ResolvedObjectMeta{
```
jagathprakash, line 5570: So according to my comment in pipelinerun_test.go, if storePipelineSpecAndMergeMeta is called twice once without the ResolvedObject and 2nd time with the resolvedObjectMeta then the ConfigSource should not get stored.
Either this is expected or not, please add a test in either case as a documentation of the behavior.

pkg/reconciler/pipelinerun/resources/pipelineref.go
chuangw6: I think it's covered in [TestGetPipelineFuncSpecAlreadyFetched](https://github.com/tektoncd/pipeline/pull/5397/files/074210dd2eee8df89d5a15d9ceb42e3a89fec178#diff-3864741a4c79caf7dcb0d45d46fd2780275657d519554f0fd14dd56adc9f1fdaL209-R282) same for taskrun level in [TestGetTaskFuncFromTaskRunSpecAlreadyFetched](https://github.com/tektoncd/pipeline/pull/5397/files/074210dd2eee8df89d5a15d9ceb42e3a89fec178#diff-bd1e986450e0c75da7f92a1a33b2bff6a133d63c3d584dedb90d1e8abce509d5L451-R533). 

PTAL and let me know if this is not what you meant. Thank you!!

pkg/reconciler/pipelinerun/resources/pipelineref.go
chuangw6: SG! The issue link was already on the next line. I'll remove my name and just leave the link there instead :D. Thanks!

pkg/reconciler/pipelinerun/resources/pipelineref.go
Yongxuanzhang: The coverage report says it is not tested. I didn't look into the details maybe you're right

pkg/reconciler/pipelinerun/pipelinerun.go
chuangw6: Good point @jagathprakash! I guess this is something not impossible to happen but not sure how likely it will be. In normal cases, the source and the spec should be available from remote resolution at the same time. Agreed that we should avoid that if this can be the case.

What I did in the recent update is move out the configsource chunk out of the status.spec != nil check and also added tests for it. Please take a look and let me know what you think. Thank you!!

pkg/reconciler/pipelinerun/pipelinerun_test.go
chuangw6: Thanks @jagathprakash. SG. Added the tests. PTAL!

pkg/reconciler/pipelinerun/pipelinespec/pipelinespec.go
```diff
@@ -22,32 +22,37 @@ import (
 	"fmt"
 
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	tresources "github.com/tektoncd/pipeline/pkg/reconciler/taskrun/resources"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // GetPipeline is a function used to retrieve Pipelines.
-type GetPipeline func(context.Context, string) (v1beta1.PipelineObject, error)
+type GetPipeline func(context.Context, string) (v1beta1.PipelineObject, *v1beta1.ConfigSource, error)
```
dibyom, line 30: maybe a terrible idea but PipelineObject is an interface right? Can we add a function to that GetConfigSource() that returns a ConfigSource object instead of changing the function signature?

dibyom (commented):
I think we need to check the feature flag before writing the provenance; Also, I think given that this involves both remote resolution + the pipeline reconciler, an e2e test that checks if the provenance was written would be good

pkg/reconciler/pipelinerun/pipelinespec/pipelinespec.go
wlynch: We probably don't want to embed location data in to pipelines themselves, and I'd lean towards not adding extra interface methods that the base pipeline type would need to `return nil` for.

You'd probably want another interface type that embeds this one that would wrap the Pipeline with the provenance data, but that would also modify the func signature so I don't feel too strongly vs leaving it as-is. 🤷

wlynch (approved):
/lgtm

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1266,6 +1269,18 @@ func storePipelineSpecAndMergeMeta(pr *v1beta1.PipelineRun, ps *v1beta1.Pipeline
 		pr.ObjectMeta.Annotations = kmap.Union(meta.Annotations, pr.ObjectMeta.Annotations)
 
 	}
+
+	// Propagate ConfigSource from remote resolution to PipelineRun Status
```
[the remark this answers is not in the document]
chuangw6: Yes! That's right.

pkg/reconciler/pipelinerun/pipelinespec/pipelinespec.go
```diff
@@ -57,10 +62,14 @@ func GetPipelineData(ctx context.Context, pipelineRun *v1beta1.PipelineRun, getP
...
 			pipelineSpec = pipeline.PipelineSpec()
 		}
+		configSource = source
 	default:
 		return nil, nil, fmt.Errorf("pipelineRun %s not providing PipelineRef or PipelineSpec", pipelineRun.Name)
 	}
 
 	pipelineSpec.SetDefaults(ctx)
-	return &pipelineMeta, &pipelineSpec, nil
+	return &tresources.ResolvedObjectMeta{
```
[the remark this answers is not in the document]
chuangw6: https://github.com/tektoncd/pipeline/pull/5397#discussion_r974607056 might be related. I think the main reason is that this helps us pass in all metadata about the resolved object in 1 struct rather than splitting into separate fields.

> Also, the taskRun reconciler package seems like an odd place for this struct to be defined in

I concur with that given the fact that it is used in both taskrun and pipelinerun reconciler, and this struct not only has *metav1.ObjectMeta but also has the source metadata. Do you have better suggestion on where we should define this struct if we go down that road? I am open to discussion here.

cc @wlynch

pkg/reconciler/pipelinerun/pipelinerun.go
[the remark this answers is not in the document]
chuangw6: the check was added. PTAL! Thank you!

chuangw6:
> I think we need to check the feature flag before writing the provenance; 

Yes indeed. Since https://github.com/tektoncd/pipeline/pull/5670 was merged recently, this pr was rebased and the feature flag check was added. PTAL.

> Also, I think given that this involves both remote resolution + the pipeline reconciler, an e2e test that checks if the provenance was written would be good

Good point! Agree. I've added the e2e test in my local but not pushed because the source value in all resolvers is still nil, which will lead us to compare if nil == nil. Happy to add this in another PR (or hold this pr) once the prs of setting value in resolver are merged. wdyt?

dibyom:
> Good point! Agree. I've added the e2e test in my local but not pushed because the source value in all resolvers is still nil, which will lead us to compare if nil == nil. Happy to add this in another PR (or hold this pr) once the prs of setting value in resolver are merged. wdyt? @dibyom

Slight preference for merging the PRs setting the source value in resolver first and then merging this with the e2e test (since this pr enables the feature).

pkg/reconciler/pipelinerun/pipelinespec/pipelinespec.go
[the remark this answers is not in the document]
dibyom: Hmm, looks like there are many instances of the pipelinerun reconciler depending on taskrun reconciler's resources package :/

Maybe pkg/internal/resolved_meta.go?

pkg/reconciler/pipelinerun/pipelinespec/pipelinespec.go
dibyom: Hmm ok...I did say it might be terrible idea 😛 

I think what we have is fine for now, though it'd be nice if we didn't have to pass around the ConfigSource as much as we do now

pkg/reconciler/pipelinerun/pipelinespec/pipelinespec.go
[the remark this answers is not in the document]
chuangw6: Thanks for the idea. Putting into internal SGTM. I put it under `pkg/internal/resolution/resolved_meta.go` file since the internal folder has subdirectories. PTAL. Let me know if this doesn't make sense or if you have other feedback. Many Thanks!

chuangw6:
> Slight preference for merging the PRs setting the source value in resolver first and then merging this with the e2e test (since this pr enables the feature).

That makes perfect sense to me! PR for most resolvers were merged and e2e test is added. Please take a look! Thank you.

dibyom:
> Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing

This seems to be already documented

dibyom:
Thank you for all the work through the many iterations on this @chuangw6 🎉 

/lgtm

dibyom:
Coverage drop is oci/resolver is due to one line being split into two (thanks @chuangw6 ). So again:

/lgtm

chuangw6:
Thank you so much everyone for the thorough review on this PR. @abayer @jagathprakash @Yongxuanzhang @alhuizenga @dibyom @wlynch!!

Started with the simple approach which passed the data through unstructured annotations to the current approach which passes through the structured fields in status, this PR enables Chains to capture the important information in the provenance in a way that is *extensible and easy to maintain*. Special thanks to @wlynch @dibyom for the suggestion of introducing the structured fields into `ResolutionRequest.Status` and `Run.Status`.  👍

Issue #5420, opened 2022-09-02
Validation of pipelines with hundreds of tasks and links fails on default webhook timeout (10s)


rafalbigaj:
# Expected Behavior

`Pipeline` DAG validation should work efficiently for hundreds of tasks and links between them. Tekton Pipelines webhook performance should neither be impacted by DAG validation nor fail with default timeout (10s).

# Actual Behavior

`PipelineRun` with hundreds of tasks and links creation fails with:

```
failed calling webhook "validation.webhook.pipeline.tekton.dev":
failed to call webhook: 
  Post "https://tekton-pipelines-webhook.tekton-pipelines.svc:443/resource-validation?timeout=10s": 
  context deadline exceeded
```

# Steps to Reproduce the Problem

1. Create a YAML file with `PipelineRun` containing hundreds of tasks and links like this one: [pipeline.hundreds-of-tasks.yaml](https://ibm.box.com/shared/static/ciifdw2kcxf23615jj13quk0yahsdapf.yaml)
2. Apply the YAML file on Tekton cluster
3. tekton-pipelines-webhook fails with timeout

# Additional Info

- Kubernetes version:

```
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.0", GitCommit:"4ce5a8954017644c5420bae81d72b09b735c21f0", GitTreeState:"clean", BuildDate:"2022-05-03T13:36:49Z", GoVersion:"go1.18.1", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.9+IKS", GitCommit:"1dc6043da4c077ebf5fa7a44d98695719495ec79", GitTreeState:"clean", BuildDate:"2022-07-14T04:23:53Z", GoVersion:"go1.17.11", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:

```
 v0.38.3
```

vdemeester:
@rafalbigaj I think this is fixed by #5421 isn't it ?

rafalbigaj:
@vdemeester indeed. Thanks

PR #5421, opened 2022-09-02
Improve DAG validation for pipelines with hundreds of tasks

rafalbigaj:
DAG validation rewritten using Kahn's algorithm to find cycles in task dependencies.

Original implementation, as pointed at https://github.com/tektoncd/pipeline/issues/5420 is a root cause of poor validation webhook performance, which fails on default timeout (10s).

# Changes

# Submitter Checklist

  functionality, content, code)
- [X] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)

/kind bug

# Release Notes

```release-note
bug fixes:
- https://github.com/tektoncd/pipeline/issues/5420 - Improve DAG validation for pipelines with hundreds of tasks (validation wehbook performance)
```

### DIFF
--- a/pkg/reconciler/pipeline/dag/dag.go
+++ b/pkg/reconciler/pipeline/dag/dag.go
@@ -19,6 +19,7 @@ package dag
 import (
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/tektoncd/pipeline/pkg/list"
@@ -79,6 +80,11 @@ func Build(tasks Tasks, deps map[string][]string) (*Graph, error) {
 		}
 	}
 
+	// Ensure no cycles in the graph
+	if err := findCyclesInDependencies(deps); err != nil {
+		return nil, fmt.Errorf("cycle detected; %w", err)
+	}
+
 	// Process all from and runAfter constraints to add task dependency
 	for pt, taskDeps := range deps {
 		for _, previousTask := range taskDeps {
@@ -120,41 +126,72 @@ func GetCandidateTasks(g *Graph, doneTasks ...string) (sets.String, error) {
 	return d, nil
 }
 
-func linkPipelineTasks(prev *Node, next *Node) error {
-	// Check for self cycle
-	if prev.Task.HashKey() == next.Task.HashKey() {
-		return fmt.Errorf("cycle detected; task %q depends on itself", next.Task.HashKey())
-	}
-	// Check if we are adding cycles.
-	path := []string{next.Task.HashKey(), prev.Task.HashKey()}
-	if err := lookForNode(prev.Prev, path, next.Task.HashKey()); err != nil {
-		return fmt.Errorf("cycle detected: %w", err)
-	}
+func linkPipelineTasks(prev *Node, next *Node) {
 	next.Prev = append(next.Prev, prev)
 	prev.Next = append(prev.Next, next)
-	return nil
 }
 
-func lookForNode(nodes []*Node, path []string, next string) error {
-	for _, n := range nodes {
-		path = append(path, n.Task.HashKey())
-		if n.Task.HashKey() == next {
-			return errors.New(getVisitedPath(path))
+// use Kahn's algorithm to find cycles in dependencies
+func findCyclesInDependencies(deps map[string][]string) error {
+	independentTasks := sets.NewString()
+	dag := make(map[string]sets.String, len(deps))
+	childMap := make(map[string]sets.String, len(deps))
+	for task, taskDeps := range deps {
+		if len(taskDeps) == 0 {
+			continue
 		}
-		if err := lookForNode(n.Prev, path, next); err != nil {
-			return err
+		dag[task] = sets.NewString(taskDeps...)
+		for _, dep := range taskDeps {
+			if len(deps[dep]) == 0 {
+				independentTasks.Insert(dep)
+			}
+			if children, ok := childMap[dep]; ok {
+				children.Insert(task)
+			} else {
+				childMap[dep] = sets.NewString(task)
+			}
 		}
 	}
-	return nil
+
+	for {
+		parent, ok := independentTasks.PopAny()
+		if !ok {
+			break
+		}
+		children := childMap[parent]
+		for {
+			child, ok := children.PopAny()
+			if !ok {
+				break
+			}
+			dag[child].Delete(parent)
+			if dag[child].Len() == 0 {
+				independentTasks.Insert(child)
+				delete(dag, child)
+			}
+		}
+	}
+
+	return getInterdependencyError(dag)
 }
 
-func getVisitedPath(path []string) string {
-	// Reverse the path since we traversed the Graph using prev pointers.
-	for i := len(path)/2 - 1; i >= 0; i-- {
-		opp := len(path) - 1 - i
-		path[i], path[opp] = path[opp], path[i]
+func getInterdependencyError(dag map[string]sets.String) error {
+	if len(dag) == 0 {
+		return nil
 	}
-	return strings.Join(path, " -> ")
+	firstChild := ""
+	for task := range dag {
+		if firstChild == "" || firstChild > task {
+			firstChild = task
+		}
+	}
+	deps := dag[firstChild].List()
+	depNames := make([]string, 0, len(deps))
+	sort.Strings(deps)
+	for _, dep := range deps {
+		depNames = append(depNames, fmt.Sprintf("%q", dep))
+	}
+	return fmt.Errorf("task %q depends on %s", firstChild, strings.Join(depNames, ", "))
 }
 
 func addLink(pt string, previousTask string, nodes map[string]*Node) error {
@@ -163,9 +200,7 @@ func addLink(pt string, previousTask string, nodes map[string]*Node) error {
 		return fmt.Errorf("task %s depends on %s but %s wasn't present in Pipeline", pt, previousTask, previousTask)
 	}
 	next := nodes[pt]
-	if err := linkPipelineTasks(prev, next); err != nil {
-		return fmt.Errorf("couldn't create link from %s to %s: %w", prev.Task.HashKey(), next.Task.HashKey(), err)
-	}
+	linkPipelineTasks(prev, next)
 	return nil
 }
 
--- a/pkg/reconciler/pipeline/dag/dag_test.go
+++ b/pkg/reconciler/pipeline/dag/dag_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package dag_test
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


pkg/reconciler/pipeline/dag/dag.go
```diff
@@ -121,40 +126,62 @@ func GetCandidateTasks(g *Graph, doneTasks ...string) (sets.String, error) {
...
 	return nil
 }
 
-func lookForNode(nodes []*Node, path []string, next string) error {
-	for _, n := range nodes {
-		path = append(path, n.Task.HashKey())
-		if n.Task.HashKey() == next {
-			return errors.New(getVisitedPath(path))
+// use Kahn's algorithm to find cycles in dependencies
+func findCyclesInDependencies(deps map[string][]string) error {
```
Udiknedormin, line 135: This function wrongly errors on:
```go
deps := map[string][]string{
  "a": []string{},
  "b": []string{"c", "d"},
  "c": []string{},
  "d": []string{},
}
// valid graph, topo-order: a, c, d, b
// but the call erros: `task "c" depends on `
```

pkg/reconciler/pipeline/dag/dag.go
```diff
@@ -121,40 +126,62 @@ func GetCandidateTasks(g *Graph, doneTasks ...string) (sets.String, error) {
...
+	independentTasks := sets.NewString()
+	dag := make(map[string]sets.String, len(deps))
+	childMap := make(map[string]sets.String, len(deps))
+	for task, taskDeps := range deps {
+		dag[task] = sets.NewString(taskDeps...)
+		for _, dep := range taskDeps {
+			if len(deps[dep]) == 0 {
+				independentTasks.Insert(dep)
+			}
```
Udiknedormin, line 144: I can see it's an optimization originating from the fact that no topo-sort order is actually built or printed (as "independent" tasks that are also "final" ones --- are not even put into the `independentTasks` set).

pkg/reconciler/pipeline/dag/dag.go
```diff
@@ -121,40 +126,62 @@ func GetCandidateTasks(g *Graph, doneTasks ...string) (sets.String, error) {
...
+		for {
+			child, ok := children.PopAny()
+			if !ok {
+				break
+			}
+			dag[child].Delete(parent)
+			if dag[child].Len() == 0 {
+				independentTasks.Insert(child)
+				delete(dag, child)
+			}
```
Udiknedormin, line 168: Note how it's only the children that are removed from the stack. The original `independentTasks` are not. So at the end, `dag` contains at least all the original `independentTasks`.

pkg/reconciler/pipeline/dag/dag.go
```diff
@@ -121,40 +126,62 @@ func GetCandidateTasks(g *Graph, doneTasks ...string) (sets.String, error) {
...
-func getVisitedPath(path []string) string {
-	// Reverse the path since we traversed the Graph using prev pointers.
-	for i := len(path)/2 - 1; i >= 0; i-- {
-		opp := len(path) - 1 - i
-		path[i], path[opp] = path[opp], path[i]
+func getInterdependencyError(dag map[string]sets.String) error {
+	for parent, children := range dag {
+		// report first interdependency
+		var childNames []string
+		for child := range children {
```
Udiknedormin, line 179: ...which is why `children` here can be empty and therefore the first independent task will be reported in the error.

It should be enough to either remove the independent tasks from the dag (just like children are removed after no entering edges are left), or adding a simple `if` here:
```go
if len(children) == 0 {
  continue
}
```

The Kahn's algorithm is defined for edges, so it doesn't really matter which method will be used as long as it's actually only triggered for existing edges. However, it's easier to define the invariant of `dag` if it only contains tasks that have some edges entering into them. That would require adding:
```go
delete(dag, parent)
```
...right after the `if !ok`. `delete(dag, child)` may be removed then, but it should be safe to leave it be too, as Go supports delete-non-existing for maps.

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -549,6 +550,78 @@ func TestBuild_InvalidDAG(t *testing.T) {
 	}
 }
 
+func TestBuildGraphWithHundredsOfTasks_Success(t *testing.T) {
```
Udiknedormin, line 553: Maybe some performance check too?

Udiknedormin (changes requested):

pkg/reconciler/pipeline/dag/dag.go
rafalbigaj: @Udiknedormin thanks for pointing. I have guarded the code to be safe in such case, even though graph builder does not include tasks with empty dependencies.

Udiknedormin (approved):
Looks ok to me now.

pkg/reconciler/pipeline/dag/export_test.go
```diff
@@ -0,0 +1,21 @@
+/*
+Copyright 2019 The Tekton Authors
```
afrittoli, line 2: 2022?

afrittoli:
Thanks @rafalbigaj for this! Do you have any number about the improvement in performance with the new algorithm?
This change adds complexity to the code, which is fine as long as it is justified by the performance improvement.

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -630,3 +703,69 @@ func assertSameDAG(t *testing.T, l, r *dag.Graph) {
 		}
 	}
 }
+
+func TestFindCyclesInDependencies(t *testing.T) {
```
afrittoli, line 707: ❤️

rafalbigaj:
@afrittoli let me provide as an example results from `TestBuildGraphWithHundredsOfTasks_Success` test in both implementations:
- **0.04s** - new implementation based on cycles detection using Kahn's algorithm (`findCyclesInDependencies`)
- **did not finish in 1h!** - old implementation based on cycle detection on every link addition (`lookForNode` in `linkPipelineTasks`)

Benchmarked on: 2,3 GHz 8-Core Intel Core i9; 16 GB 2400 MHz DDR4

afrittoli (commented):
Thanks @rafalbigaj for this. The code and test coverage looks good to me.
Just one possible NIT, but nothing blocking for this PR.
/approve

pkg/reconciler/pipeline/dag/dag_test.go
pritidesai: yes, please so that we do not run into it in future, thanks!

pkg/reconciler/pipeline/dag/dag.go
```diff
@@ -163,9 +200,7 @@ func addLink(pt string, previousTask string, nodes map[string]*Node) error {
 		return fmt.Errorf("task %s depends on %s but %s wasn't present in Pipeline", pt, previousTask, previousTask)
 	}
 	next := nodes[pt]
-	if err := linkPipelineTasks(prev, next); err != nil {
```
pritidesai, line 166: I had troubleshooted till this point, thanks for taking it further, appreciate your efforts 🙏

pkg/reconciler/pipeline/dag/dag_test.go
afrittoli: The problem with performance checks is that they could be inherently flaky because of the variable performance of the test nodes. If we do add performance checks, I would suggest for now to only log the execution time. 
We can collect such timings over time for some time and then decide on an acceptable bar for the execution time.

pkg/reconciler/pipeline/dag/dag.go
```diff
@@ -163,9 +200,7 @@ func addLink(pt string, previousTask string, nodes map[string]*Node) error {
 		return fmt.Errorf("task %s depends on %s but %s wasn't present in Pipeline", pt, previousTask, previousTask)
 	}
 	next := nodes[pt]
```
pritidesai, line 202: This is holding a huge struct (object) in real world applications. It is set to the entire `pipelineTask` specification along with a list of `params`, `when` expressions, the entire specification when `taskSpec` is specified, etc. The `pipelineTask` specification is not required at this point since all the dependencies (`runAfter`, task results in params and when expressions, and from) are calculated before calling `dag.Build`. All we need is just the `pipelineTask.Name` i.e. `HashKey()`.  We have a room of improvement in future to avoid passing the entire blob around.

pkg/reconciler/pipeline/dag/dag_test.go
pritidesai: That's a great idea @afrittoli, where do we store those timings? 

I was running the test you had introduced https://github.com/tektoncd/pipeline/pull/3524 (a huge thanks to you for introducing such test):

https://github.com/tektoncd/pipeline/blob/99404d532ae9e9149cd941d11b754035cfb646c9/pkg/reconciler/pipelinerun/resources/pipelinerunstate_test.go#L1404

I ran it locally yesterday multiple times which took 60 seconds 😲 (without this PR). 

At the time introducing this test, it took less than the default timeout 30 seconds based on the PR description (if I am reading it right):

```This changes adds a unit test that reproduces the issue in https://github.com/tektoncd/pipeline/issues/3521, which
used to fail (with timeout 30s) and now succeedes for pipelines roughly
up to 120 tasks / 120 links. On my laptop, going beyond 120 tasks/links
takes longer than 30s, so I left the unit test at 80 to avoid
introducing a flaky test in CI. There is still work to do to improve
this further, some profiling / tracing work might help.```

There has been many changes introduced since then, we really need a way to flag us (nightly performance tests) when we introduce any delay.

pritidesai: Nightly performance test or like you are suggesting to collect timings over time. I am fine with either option.

afrittoli: Yes. The `TestBuildGraphWithHundredsOfTasks_Success` in a way is a kind of performance test (like the one I added at the time), because if things slow down significantly, the tests will eventually timeout.
If we collected test execution times and graphed them over time - or if we had some dedicated nightly performance test, we would be able to see a change in execution time sooner than just waiting for the tests to timeout.

That is something that we would need to setup as part of the CI infrastructure. Would you like to create an issue about that?

afrittoli:
@tektoncd/core-maintainers I added the "needs-cherry-pick" label for this item. While there was no explicit functional regression, there has been a performance degradation over time already since v0.36.x, which is significantly eased (solved) by this PR. I would propose doing a new series of minor releases to include this.

pkg/reconciler/pipeline/dag/dag_test.go
pritidesai: yup definitely, we had a PR from @guillaumerose https://github.com/tektoncd/pipeline/pull/4378 which didn't materialize but at least most of us including @vdemeester and @imjasonh were on board with the idea of running nightly performance tests.

PR #5427, opened 2022-09-02
TEP-0075: More flexible ways to provide values for object param keys

chuangw6:
# Changes

Related to https://github.com/tektoncd/pipeline/issues/5270.

Prior to this change, if users want to provide default value for an
object param, they have to provide values for ALL keys, but can't provide
values for a subset of required keys that are declared in `properties`
section. Same restriction applies to the `value` provided from run level.

In this PR, we relax the validation for object param keys to allow users
to provide values for an object param in a more flexible way i.e. a subset
of keys are provided in the spec's default, and the rest of the keys are
provided in run level's value.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
More flexible ways to provide values for object param keys: a subset of keys can be provided from default, and the rest is provided at runtime.
```

### DIFF
--- a/pkg/apis/pipeline/v1/task_validation.go
+++ b/pkg/apis/pipeline/v1/task_validation.go
@@ -27,7 +27,6 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/validate"
 	"github.com/tektoncd/pipeline/pkg/apis/version"
-	"github.com/tektoncd/pipeline/pkg/list"
 	"github.com/tektoncd/pipeline/pkg/substitution"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -366,7 +365,6 @@ func ValidateParameterVariables(ctx context.Context, steps []Step, params []Para
 		errs = errs.Also(validateVariables(ctx, steps, "params", allParameterNames))
 	}
 	errs = errs.Also(validateArrayUsage(steps, "params", arrayParameterNames))
-	errs = errs.Also(validateObjectDefault(objectParamSpecs))
 	return errs.Also(validateObjectUsage(ctx, steps, objectParamSpecs))
 }
 
@@ -402,45 +400,6 @@ func validateObjectUsage(ctx context.Context, steps []Step, params []ParamSpec) 
 	}
 
 	return errs.Also(validateObjectUsageAsWhole(steps, "params", objectParameterNames))
-}
-
-// validateObjectDefault validates the keys of all the object params within a
-// slice of ParamSpecs are provided in default iff the default section is provided.
-func validateObjectDefault(objectParams []ParamSpec) (errs *apis.FieldError) {
-	for _, p := range objectParams {
-		errs = errs.Also(ValidateObjectKeys(p.Properties, p.Default).ViaField(p.Name))
-	}
-	return errs
-}
-
-// ValidateObjectKeys validates if object keys defined in properties are all provided in its value provider iff the provider is not nil.
-func ValidateObjectKeys(properties map[string]PropertySpec, propertiesProvider *ParamValue) (errs *apis.FieldError) {
-	if propertiesProvider == nil || propertiesProvider.ObjectVal == nil {
-		return nil
-	}
-
-	neededKeys := []string{}
-	providedKeys := []string{}
-
-	// collect all needed keys
-	for key := range properties {
-		neededKeys = append(neededKeys, key)
-	}
-
-	// collect all provided keys
-	for key := range propertiesProvider.ObjectVal {
-		providedKeys = append(providedKeys, key)
-	}
-
-	missings := list.DiffLeft(neededKeys, providedKeys)
-	if len(missings) != 0 {
-		return &apis.FieldError{
-			Message: fmt.Sprintf("Required key(s) %s are missing in the value provider.", missings),
-			Paths:   []string{"properties", "default"},
-		}
-	}
-
-	return nil
 }
 
 // validateObjectUsageAsWhole makes sure the object params are not used as whole when providing values for strings
--- a/pkg/apis/pipeline/v1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1/task_validation_test.go
@@ -189,6 +189,17 @@ func TestTaskSpecValidate(t *testing.T) {
 					"key1": {},
 					"key2": {},
 				},
+			}, {
+				Name:        "myobjWithDefaultPartialKeys",
+				Type:        v1.ParamTypeObject,
+				Description: "param",
+				Properties: map[string]v1.PropertySpec{
+					"key1": {},
+					"key2": {},
+				},
+				Default: v1.NewObject(map[string]string{
+					"key1": "default",
+				}),
 			}},
 			Steps: validSteps,
 		},
@@ -685,27 +696,6 @@ func TestTaskSpecValidateError(t *testing.T) {
 		expectedError: apis.FieldError{
 			Message: fmt.Sprintf("The value type specified for these keys %v is invalid", []string{"key1"}),
 			Paths:   []string{"params.task.properties"},
-		},
-	}, {
-		name: "keys defined in properties are missed in default",
-		fields: fields{
-			Params: []v1.ParamSpec{{
-				Name:        "myobjectParam",
-				Description: "param",
-				Properties: map[string]v1.PropertySpec{
-					"key1": {},
-					"key2": {},
-				},
-				Default: v1.NewObject(map[string]string{
-					"key1": "var1",
-					"key3": "var1",
-				}),
-			}},
-			Steps: validSteps,
-		},
-		expectedError: apis.FieldError{
-			Message: fmt.Sprintf("Required key(s) %s are missing in the value provider.", []string{"key2"}),
-			Paths:   []string{"myobjectParam.properties", "myobjectParam.default"},
 		},
 	}, {
 		name: "invalid step",
--- a/pkg/apis/pipeline/v1beta1/task_validation_test.go
+++ b/pkg/apis/pipeline/v1beta1/task_validation_test.go
@@ -225,6 +225,17 @@ func TestTaskSpecValidate(t *testing.T) {
 					"key1": {},
 					"key2": {},
 				},
+			}, {
+				Name:        "myobjWithDefaultPartialKeys",
+				Type:        v1beta1.ParamTypeObject,
+				Description: "param",
+				Properties: map[string]v1beta1.PropertySpec{
+					"key1": {},
+					"key2": {},
+				},
+				Default: v1beta1.NewObject(map[string]string{
+					"key1": "default",
+				}),
 			}},
 			Steps: validSteps,
 		},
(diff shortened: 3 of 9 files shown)


chuangw6:
cc @lbernick @wlynch @chitrangpatel @ywluogg @Yongxuanzhang 
please take a look thanks!

afrittoli:
Thank you @chuangw6 - could you please provide release notes for this feature, and check the submitter checklist?

chuangw6:
> Thank you @chuangw6 - could you please provide release notes for this feature, and check the submitter checklist?

Added! Thanks @afrittoli !

pkg/apis/pipeline/v1/task_validation_test.go
```diff
@@ -686,27 +686,6 @@ func TestTaskSpecValidateError(t *testing.T) {
 			Message: fmt.Sprintf("The value type specified for these keys %v is invalid", []string{"key1"}),
 			Paths:   []string{"params.task.properties"},
 		},
-	}, {
-		name: "keys defined in properties are missed in default",
```
lbernick, line 690: let's keep this test case around but just have it result in a successful validation.

lbernick (commented):
I think this is actually two separate changes:
1. if default is provided for all keys, the values from the run will be merged with the defaults
2. the default does not have to be provided for all keys

Is it clear that we want both of these features?
We might also want to update tep 75 before merging this

chuangw6:
> * if default is provided for all keys, the values from the run will be merged with the defaults

Can you elaborate a bit on what merge means in case default is provided for all keys? @lbernick

pkg/apis/pipeline/v1/task_validation_test.go
chuangw6: Sounds great! Done. Thanks @lbernick

chuangw6:
> 2. the default does not have to be provided for all keys

Not only for default but also the value on the runtime. Both of them do not have to be provided for all keys. (Prior, the validation enforces that they have to provide values for all keys if they provide value for the object param.)

> 1. if default is provided for all keys, the values from the run will be merged with the defaults

This is true. But I'd think this as the implication of that both default and runtime do not need to provide value for all keys of an object param. 

Currently [tep75](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md#:~:text=If%20a%20value%20is%20provided%20for%20the%20param%2C%20the%20default%20will%20not%20be%20used%20(i.e.%20there%20will%20be%20no%20behavior%20to%20merge%20the%20default%20with%20the%20provided%20value%2C%20if%20for%20example%20the%20provided%20value%20specified%20some%20fields%20but%20not%20others).) says "If a value is provided for the param, the default will not be used (i.e. there will be no behavior to merge the default with the provided value, if for example the provided value specified some fields but not others).". If we do not have the merge behaviour, what will happen is that the keys missed at runtime value will have empty value even though the `default` provides the value for these keys, which I think probably is a bit unintuitive. 

For example, and object param named `foo` has two keys `key1` and `key2`. The `default` provides value for both keys (i.e. "default1" for `key1`, and "default2" for `key2`), but at runtime, only the value of `key1` is provided (i.e. "value-runtime"). 
- Without merging behaviour, the resolved object will have "value-runtime" as the value of `key1`, but empty value for `key2`.
- With merging behaviour, the resolved object will have "value-runtime" as the value of `key1`, and "default2" as the value of `key2`.

> Is it clear that we want both of these features? We might also want to update tep 75 before merging this

IMHO, the 2nd case might make more sense, but I am open to the discussion here. And I agree that we need to update the tep before merging this.

lbernick:
Thanks for the response @chuangw6!

Let's say we start with not requiring the default to specify all keys. In order for this to be useful, we will either need the merge behavior (so that unspecified defaults can be provided at runtime), or we will need to allow object keys to be optional. Probably both!

We could start with just the merge behavior, but still require the default to specify all keys. There's definitely some examples where this could be useful, e.g. a git object param with a default url like the example you have in this PR. 

In the example you provided, without merging behavior, one of the keys would not be set, and the taskrun would fail because we don't support optional keys, correct? Merging behavior would be backwards incompatible in the sense that this would no longer fail, but because we don't support optional keys we don't need to provide a mechanism to "unset" a key at runtime.

I think we might want to start with just the merge behavior to avoid getting into optional vs required keys, WDYT?

chuangw6:
> I think we might want to start with just the merge behavior to avoid getting into optional vs required keys, WDYT?

Starting with merge behaviour to avoid optional v.s. required keys SGTM! 
But it seems to me that default does not have to specify all keys, which is not related to optional/required keys. Currently, the tep proposed that all keys specified in properties are required. Therefore, in the following example, an error will be thrown because we find out that neither runtime or default provides a value for the required key `commit`.

```yaml
kind: TaskRun
metadata:
  generateName: object-param-test-
spec:
  params:
    - name: gitrepo
      value:
        url: "xyz.com"
  taskSpec:
    params:
      - name: gitrepo
        properties:
          url: {}
          commit: {}
        default:
          url: "default.com"
```

If the above example provides the value (i.e. "sha123") for key `commit` at runtime, this would be valid and resolved object param `gitrepo` will be like `{url:"xyz.com", commit:"sha123"}`.

lbernick:
ok -- I'm convinced that this doesn't require optional keys and isn't backwards incompatible (because we were previously requiring the taskrun to provide a value for all keys if it overrode the default)

pkg/apis/pipeline/v1/task_validation_test.go
lbernick: I think you might not have pushed these changes?

pkg/reconciler/pipelinerun/resources/validate_params_test.go
```diff
@@ -262,13 +268,18 @@ func TestValidateObjectParamRequiredKeys_Valid(t *testing.T) {
 					"key1": {Type: "string"},
 					"key2": {Type: "string"},
 				},
+				Default: &v1beta1.ParamValue{
```
lbernick, line 271: I think we'd want to keep the existing test case and add a new one where one of the keys is provided by the default

pkg/reconciler/taskrun/validate_resources_test.go
```diff
@@ -155,6 +155,13 @@ func TestValidateResolvedTaskResources_ValidParams(t *testing.T) {
 						"key1": {},
 						"key2": {},
 					},
+					Default: &v1beta1.ParamValue{
```
lbernick, line 158: again could you keep the existing test case and add a new one?

pkg/reconciler/taskrun/validate_resources_test.go
```diff
@@ -174,9 +181,7 @@ func TestValidateResolvedTaskResources_ValidParams(t *testing.T) {
 	}, {
 		Name: "myobj",
 		Value: *v1beta1.NewObject(map[string]string{
-			"key1":      "val1",
-			"key2":      "val2",
-			"extra_key": "val3",
+			"key2": "my value",
```
lbernick, line 184: could you add a test case where the default defines a value for one of the keys and that value is overridden by the taskRun?

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -798,27 +809,6 @@ func TestTaskSpecValidateError(t *testing.T) {
 			Message: fmt.Sprintf("The value type specified for these keys %v is invalid", []string{"key1"}),
 			Paths:   []string{"params.task.properties"},
 		},
-	}, {
-		name: "keys defined in properties are missed in default",
```
lbernick, line 802: we might still want to validate for taskruns with inline task specs that all the keys have values defined somewhere.

chuangw6:
> ok -- I'm convinced that this doesn't require optional keys and isn't backwards incompatible (because we were previously requiring the taskrun to provide a value for all keys if it overrode the default)

Correct. Now taskrun can still provide a value for all keys, but they don't have to as long as default provides a value the keys that are missed at tasrkun.

pkg/apis/pipeline/v1beta1/task_validation_test.go
chuangw6: It's under the valid test cases called valid params type explicit and the param name is myobjWithDefaultPartialKeys.

pkg/apis/pipeline/v1/task_validation_test.go
chuangw6: It's under the valid test cases called `valid params type explicit` and the param name is `myobjWithDefaultPartialKeys`.

pkg/reconciler/pipelinerun/resources/validate_params_test.go
chuangw6: SG! Reverted the old test back and add this new one.

pkg/reconciler/taskrun/validate_resources_test.go
chuangw6: This newly added test `myObjWithDefault` might be also the one that you mentioned "we might still want to validate for taskruns with inline task specs that all the keys have values defined somewhere." It is simulating the taskspec has 3 keys and default provides a value for `key1` and `key3`, taskrun provides a value for `key2` and `key3`.

pkg/reconciler/taskrun/validate_resources_test.go
chuangw6: [this](https://github.com/tektoncd/pipeline/pull/5427/files#diff-e5039733a366256d36ca2680d4f7ea1fab71fd2fc78d0526ba9a13ba511af4a2R158-R202) might cover this case. Please let me know otherwise.

pkg/apis/pipeline/v1beta1/task_validation_test.go
chuangw6: see this [comment](https://github.com/tektoncd/pipeline/pull/5427#discussion_r967250399)

chuangw6:
I also updated the tep https://github.com/tektoncd/community/pull/817. Please take a look @lbernick. Thanks!

pkg/reconciler/taskrun/validate_resources_test.go
lbernick: I was actually thinking we could validate in the admission webhook in addition to the validation that exists in the reconciler for taskruns with inline task specs. This would mean the admission webhook can reject the taskrun instead of having the reconciler fail the taskrun. but what you have here is OK

lbernick (approved):

pkg/reconciler/taskrun/validate_resources_test.go
chuangw6: Aha I see what you mean. Thanks for clarification.
Yeah, the validation against params across task and taskrun is done on the reconciler side specifically `pkg/reconciler/taskrun/validate_resources.go` file i.e. checking if a param `foo` of string type declared in task spec is provided with a value of an array type, checking missing params and missing keys for object param across default and taskrun value etc.

vdemeester (approved):

pkg/apis/pipeline/v1/task_validation.go
```diff
@@ -404,45 +402,6 @@ func validateObjectUsage(ctx context.Context, steps []Step, params []ParamSpec)
...
-// slice of ParamSpecs are provided in default iff the default section is provided.
-func validateObjectDefault(objectParams []ParamSpec) (errs *apis.FieldError) {
-	for _, p := range objectParams {
-		errs = errs.Also(ValidateObjectKeys(p.Properties, p.Default).ViaField(p.Name))
-	}
-	return errs
-}
-
-// ValidateObjectKeys validates if object keys defined in properties are all provided in its value provider iff the provider is not nil.
-func ValidateObjectKeys(properties map[string]PropertySpec, propertiesProvider *ParamValue) (errs *apis.FieldError) {
```
ywluogg, line 417: Ah can you leave this function as is? I'm planning to use it in Chains. I think initially you are making this function public because I need to do keys comparisons in Chains, is it?

ywluogg: Oh NVM because you are not using this function anymore in Chains, I will create this one in Chains then. Please ignore the previous commnent

chuangw6:
Tep 75 was updated to reflect the change as well. https://github.com/tektoncd/community/pull/817

PR #5446, opened 2022-09-07
de-dupe order and resource dependencies

pritidesai:
# Changes

In case of a pipeline with a huge list of redundant dependencies, the list of dependencies is growing exponentially. This way of calculating the list of dependencies is causing extra delay in the validation cycle. This delay sometimes hit the webhook timeout during validation. This is one of the changes being proposed to make the validation cycle efficient and avoid unnecessary delay.

Introducing a test case which was running until a timeout was hit (without @rafalbigaj's changes in https://github.com/tektoncd/pipeline/pull/5421). Thanks @rafalbigaj for the efficient implementation of building a graph. This test now runs to competition, on average taking 3 minutes. With the changes in this PR, this test runs under two seconds.  

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
De-dupe task dependencies - order and resource dependencies all together. It's very common to have a task with multiple when expressions referring to the same task but different results. Maintain a set of dependencies and add only a new parent.
```


rafalbigaj (approved):
I like this simplification very much! Thanks for catching the next bottleneck 🙏

afrittoli:
Thanks @pritidesai - could you apply the change to the `v1` API as well?

pritidesai:
thank you @afrittoli for pointing that out 👍 I have applied these changes to `v1`. Thanks!

afrittoli:
Do we need to cherry-pick this one too? @pritidesai

Issue #5373, opened 2022-08-25
Emit warnings when users create CRDs with deprecated fields


lbernick:
### Feature request

It would be great to have our admission webhook emit a warning when a user creates a CRD with a deprecated field but not reject it entirely. Knative allows us to set an error "level" (via [func(*FieldError).At](https://pkg.go.dev/knative.dev/pkg@v0.0.0-20220824011017-a6afcab5bd78/apis#FieldError.At)) which is meant for this use case but we likely want https://github.com/knative/pkg/issues/2514 to be resolved before doing this.

Issue #5111, opened 2022-07-08
Put all `pkg/apis/pipeline/v1beta1/*_test.go` files to `v1beta1_test` package


chuangw6:
### Feature request

Under `pkg/apis/pipeline/v1beta1` folder, there are a number of `*_test.go` files under `v1beta1` package, whereas some are under `v1beta1_test` package. Problem with this is that in some test files under `v1beta1` package, if we want to use some test helper functions that are defined in `v1beta1_test` package, we cannot import them i.e. [getContextBasedOnFeatureFlag](https://github.com/tektoncd/pipeline/blob/dc9ea6cfb4aab26909ac365fbaa248df2191b534/pkg/apis/pipeline/v1beta1/task_validation_test.go#L1523), [enableAlphaAPIFields](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/pipelinerun_validation_test.go#L844-L855).

So it might be great to move those `*_test.go` files under `v1beta1_test`  package. However, a couple of pain points 
- this will change thousand of lines bc we need to import `v1beta1` and prefix all functions/structs defined under it. 
- there are LOTS of unexported functions defined in `v1beta1` but used in these `*_test.go` files i.e. example [`validateParamResults`](https://github.com/tektoncd/pipeline/blob/dc9ea6cfb4aab26909ac365fbaa248df2191b534/pkg/apis/pipeline/v1beta1/pipeline_validation_test.go#L1114). If we change `*_test.go` from package `v1beta1` to `v1beta1_test`, we are not sure if it's appropriate to export those functions OR just remove testings for those unexported functions. 

Talked to @lbernick about this today. Loop in @JeromeJu for visibility. 

So this issue is created to track the thoughts/comments/suggestions about this.

lbernick:
The motivation here is based on our [code standards for tests](https://github.com/tektoncd/community/blob/main/standards.md#tests), which state that we should aim to only test exported functions. FYI @pritidesai

vdemeester:
Note that the main point on the "code standards" is that exported function are tested, the *only*, for me, is to take with a grain of salt. In an ideal world we would only test exported method ; which tends to make the code base be composed of smaller packages. But we are not in an ideal world 😅.

Also, if there is, sometimes, cases where we would like to exported a function or a set of function but we are not really prepared to support go package users to use those, there is always the possibility to use the special `internal` package 👼🏼

JeromeJu:
[...]

PR #5457, opened 2022-09-08
Emit warning for deprecated fields in v1beta1

JeromeJu:
# Changes

[WIP]This commit adds warnings when a user creates a CRD with a deprecated field.

Needs: https://github.com/knative/pkg/issues/2514
Fixes: #5373 
/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1beta1/pipeline_validation.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/validate"
+	"github.com/tektoncd/pipeline/pkg/apis/version"
 	"github.com/tektoncd/pipeline/pkg/list"
 	"github.com/tektoncd/pipeline/pkg/reconciler/pipeline/dag"
 	"github.com/tektoncd/pipeline/pkg/substitution"
@@ -79,7 +80,8 @@ func (ps *PipelineSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	errs = errs.Also(validateMatrix(ctx, ps.Finally).ViaField("finally"))
 	errs = errs.Also(validateResultsFromMatrixedPipelineTasksNotConsumed(ps.Tasks, ps.Finally))
 	// Emit warnings for deprecated fields
-	errs = errs.Also(warnPipelineFieldsDeprecation(ctx, ps))
+	errs = errs.Also(validatePipelineResourcesDeprecation(ctx, ps))
+	errs = errs.Also(validatePipelineBundleDeprecation(ctx, ps))
 	return errs
 }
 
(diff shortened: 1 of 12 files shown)


pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -78,6 +78,8 @@ func (ps *PipelineSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	errs = errs.Also(validateMatrix(ctx, ps.Tasks).ViaField("tasks"))
 	errs = errs.Also(validateMatrix(ctx, ps.Finally).ViaField("finally"))
 	errs = errs.Also(validateResultsFromMatrixedPipelineTasksNotConsumed(ps.Tasks, ps.Finally))
+	// Emit warnings for deprecated fields
+	errs = errs.Also(warnPipelineFieldsDeprecation(ctx, ps))
```
lbernick, line 82: I think individual validation functions should be responsible for emitting warnings, rather than one overall function responsible for emitting all deprecation warnings.

pkg/apis/pipeline/v1beta1/pipeline_validation.go
```diff
@@ -529,3 +531,27 @@ func validateResultsFromMatrixedPipelineTasksNotConsumed(tasks []PipelineTask, f
 	}
 	return errs
 }
+
+// Instead of rejecting the deprecated field entirely, warnPipelineFieldsDeprecation
+// emits a WarningLevel FieldError to notify users that the CRD created contains one
+// or more fields that are deprecated.
+func warnPipelineFieldsDeprecation(ctx context.Context, ps *PipelineSpec) (errs *apis.FieldError) {
+	if ps.Resources != nil {
+		errs = errs.Also(&apis.FieldError{
```
lbernick, line 540: We'll probably want to define a helper function (e.g. `DeprecatedError`) that returns a message along the lines of "X field is deprecated" at a warning level. let's remove the reference to the api version, it's not necessary.

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -659,8 +659,9 @@ func TestPipelineSpec_Validate_Failure(t *testing.T) {
 			}},
 		},
 		expectedError: apis.FieldError{
-			Message: `invalid value: pipeline declared resources didn't match usage in Tasks: Didn't provide required values: [missing-great-resource missing-wonderful-resource]`,
-			Paths:   []string{"resources"},
+			Message: `Resources field is deprecated in v1 Pipeline: Resources
+invalid value: pipeline declared resources didn't match usage in Tasks: Didn't provide required values: [missing-great-resource missing-wonderful-resource]`,
+			Paths: []string{"resources"},
```
lbernick, line 664: This test will need to ensure the error returned is at a warning level, otherwise the Pipeline will actually be rejected. This will probably require addressing https://github.com/knative/pkg/issues/2514

pkg/apis/pipeline/v1beta1/pipeline_validation_test.go
```diff
@@ -3574,3 +3575,73 @@ func enableFeatures(t *testing.T, features []string) func(context.Context) conte
...
+			}
+			if d := cmp.Diff(tt.expectedError.Error(), err.Error(), cmpopts.IgnoreUnexported(apis.FieldError{})); d != "" {
+				t.Errorf("TaskSpec.Validate() errors diff %s", diff.PrintWantGot(d))
+			}
+		})
+	}
+}
+
+// TODO refactor in #5111
+func enableTektonOCIBundles(t *testing.T) func(context.Context) context.Context {
```
lbernick, line 3636: might be easier to start with a single deprecated field such as timeout. Also, this function is defined elsewhere and should be imported instead of copy/pasted.

JeromeJu: This is because in pipeline_validation_test.go the package is v1beta1, while elsewhere in pipelineref_validation_test.go it is in v1beta1_test. I put the TODO comment as to resolve this in https://github.com/tektoncd/pipeline/issues/5111 if that makes sense to refactor in a separate PR.

lbernick: you can move it to https://github.com/tektoncd/pipeline/blob/main/pkg/apis/config/feature_flags.go-- that's where we put `EnableAlphaAPIFields` after a similar discussion

PR #5435, opened 2022-09-06
TEP-0114: Implements Retries in The Testing Wait Custom Task

XinruZhang:
# Changes

This PR implements `retries` in the testing Wait custom task,
enabling e2e test to cover test cases of Custom Task `retries`
in the future.

/kind feature

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

``` release-note
NONE
```

PR #5462, opened 2022-09-09
Avoid Duplicate Reconcile Logic in Testing Wait Custom Run Controller

XinruZhang:
# Changes

The existing test cases in Wait Custom Run reconciler_test.go are reusing some reconciliation logic in reconciler.go, this may lead some functionality deviation in the future.

This commit rewrites some test cases in reconciler_test.go to reuse ReconcileKind in reconciler.go

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/test/custom-task-ctrls/wait-task/pkg/reconciler/reconciler_test.go
+++ b/test/custom-task-ctrls/wait-task/pkg/reconciler/reconciler_test.go
@@ -19,7 +19,6 @@ package reconciler
 import (
 	"context"
 	"fmt"
-	"math"
 	"strings"
 	"testing"
 	"time"
@@ -59,6 +58,7 @@ func TestReconcile(t *testing.T) {
 		refName                string
 		timeout                string
 		params                 string
+		startTime              *metav1.Time
 		wantRunConditionType   apis.ConditionType
 		wantRunConditionStatus corev1.ConditionStatus
 		wantRunConditionReason string
@@ -70,6 +70,7 @@ func TestReconcile(t *testing.T) {
   - name: duration
     value: 1s
 `,
+		startTime:              &metav1.Time{Time: time.Unix(testClock.Now().Unix()-2, 0)},
 		wantRunConditionType:   apis.ConditionSucceeded,
 		wantRunConditionStatus: corev1.ConditionTrue,
 		wantRunConditionReason: "DurationElapsed",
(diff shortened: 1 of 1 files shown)


test/custom-task-ctrls/wait-task/pkg/reconciler/reconciler_test.go
```diff
@@ -305,12 +333,23 @@ spec:
...
 				r.Spec.Status = v1alpha1.RunSpecStatusCancelled
 			}
+			if tc.startTime != nil {
+				r.Status.StartTime = tc.startTime
+			}
 
-			fakeReconcile(t, ctx, rec, r)
+			err := rec.ReconcileKind(ctx, r)
+			if err != nil {
+				if !strings.Contains(err.Error(), "requeue") {
```
lbernick, line 349: what's the reason that requeue errors are ignored? could you add a comment?

lbernick: This seems a bit weird to me-- returning a requeue error is not the same as returning successfully; why do the tests treat them the same?

XinruZhang: Ignoring the requeue error because we are testing a single round of reconciliation given Run Spec and Status. 

If we want to deal with the requeue event, then we need to duplicate some logics of deciding when to requeue, as discussed here: https://github.com/tektoncd/pipeline/pull/5435#discussion_r964113450.

Or do you mean we should also compare the error with expected error?

XinruZhang: Correct me if I'm wrong... as far as I understand, requeuing means we need to reconcile it again at some time, meaning there's actually no other runtime error.

abayer: fwiw, in, say, `pkg/reconciler/pipelinerun/pipelinerun_test.go`, we do this too, albeit in a slightly different way:
https://github.com/tektoncd/pipeline/blob/0484584c518de3497dac5fa680888fdbc07683e5/pkg/reconciler/pipelinerun/pipelinerun_test.go#L2311

lbernick: ah ok-- good to know

PR #5463, opened 2022-09-09
TEP-0111 - Propagating workspaces in pipelinerun

chitrangpatel:
The verbosity of writing specifications in Tekton Pipelines is a common pain point that creates difficulties in getting-started scenarios. In addition, the verbosity leads to long specifications that are error-prone, harder to maintain, and reach the etcd size limits for CRDs. Automatically propagating Workspaces will lead to better user experience. This addresses features proposed in [TEP-0111](https://github.com/tektoncd/community/blob/main/teps/0111-propagating-workspaces.md). This is a follow up PR that addresses propagating workspaces in `PipelineRuns`. Propagating workspaces in Taskruns is being addressed by https://github.com/tektoncd/pipeline/pull/5081 and should be merged first.

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Workspaces are propagated in embedded specifications of pipelinerun without mutations.
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -928,7 +928,7 @@ func (c *Reconciler) createTaskRun(ctx context.Context, taskRunName string, para
 
 	var pipelinePVCWorkspaceName string
 	var err error
-	tr.Spec.Workspaces, pipelinePVCWorkspaceName, err = getTaskrunWorkspaces(pr, rpt)
+	tr.Spec.Workspaces, pipelinePVCWorkspaceName, err = getTaskrunWorkspaces(ctx, pr, rpt)
 	if err != nil {
 		return nil, err
 	}
@@ -999,7 +999,7 @@ func (c *Reconciler) createRun(ctx context.Context, runName string, params []v1b
 	}
 	var pipelinePVCWorkspaceName string
 	var err error
-	r.Spec.Workspaces, pipelinePVCWorkspaceName, err = getTaskrunWorkspaces(pr, rpt)
+	r.Spec.Workspaces, pipelinePVCWorkspaceName, err = getTaskrunWorkspaces(ctx, pr, rpt)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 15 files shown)


pkg/apis/pipeline/v1/task_types.go
```diff
@@ -52,6 +53,61 @@ func (*Task) GetGroupVersionKind() schema.GroupVersionKind {
...
+		}
+		for i := range stepTemplate.Command {
+			locationsToCheck.Insert(stepTemplate.Command[i])
+		}
+	}
+	return locationsToCheck
+}
+
+// FindWorkspaceSubstitutionLocationsInTask returns a set of all the locations in the TaskSpec  where we can apply substitutions.
+func (ts *TaskSpec) FindWorkspaceSubstitutionLocationsInTask() sets.String {
```
lbernick, line 103: Can this function be moved out of the apis package?

pkg/substitution/substitution.go
```diff
@@ -234,7 +234,8 @@ func extractExpressionFromString(s, prefix string) (string, bool) {
 	return match[0], true
 }
 
-func extractVariablesFromString(s, prefix string) ([]string, bool, string) {
+// ExtractVariablesFromString extracts variables from an input string s with the given prefix via regex matching.
+func ExtractVariablesFromString(s, prefix string) ([]string, bool, string) {
```
lbernick, line 238: can we avoid exporting this function and instead make use of one of the already exported functions in this package? If we have to export it, can you please document the return values, and what happens if the string does not have the given prefix?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1014,16 +1015,54 @@ func (c *Reconciler) createRun(ctx context.Context, runName string, params []v1b
 	return c.PipelineClientSet.TektonV1alpha1().Runs(pr.Namespace).Create(ctx, r, metav1.CreateOptions{})
 }
 
-func getTaskrunWorkspaces(pr *v1beta1.PipelineRun, rpt *resources.ResolvedPipelineTask) ([]v1beta1.WorkspaceBinding, string, error) {
+func propagateWorkspaces(rpt *resources.ResolvedPipelineTask) (*resources.ResolvedPipelineTask, error) {
```
lbernick, line 1018: can you please add a docstring for this function?

pkg/substitution/substitution.go
lbernick: Maybe an alternative option would be to write a function that gets all the names of workspaces used in a task spec?

pkg/apis/pipeline/v1/task_types.go
chitrangpatel: I think it can. I can move it closer to where it is called.

chitrangpatel: Would it make more sense to move it to `pkg/workspaces`?

pkg/reconciler/pipelinerun/pipelinerun.go
chitrangpatel: will do!

pkg/apis/pipeline/v1/task_types.go
lbernick: that could work! and instead of "workspace substitution locations", maybe it could return a list of workspace names used in the task spec

chitrangpatel: That works too. But, to get the workspace names that are used by the task spec, I will have to parse the strings (args, command etc.) that could contain workspace substitutions so I think I will still have to extract variables names using the function defined in substitution.

The reason for doing this is that because of propagated workspaces, the workspaces might not be declared in the pipeline tasks and instead, directly be used in the steps of those tasks. We only want to propagate workspaces to this task that it needs.

pkg/substitution/substitution.go
chitrangpatel: I might have to write a dedicated function e.g.`ExtractWorkspaceFromString` that calls this function internally. 

The only exported functions in this package are `Validate...` and `Apply...` which I cannot use here. I could move `FindWorkspaceSubstitutionLocationsInTask` in here instead of in `pkg/workspaces` and combine the usage to a single function e.g. `FindWorkspacesRequiredbyTask`.

chitrangpatel: However, `substitution` seems like a more general purpose package than `workspace` so moving workspace specific changes to substitution does not make sense.

chitrangpatel: For now, I have left `ExtractVariablesFromString ` as is since it needs to be called by `pkg/workspaces` which now contains `FindWorkspacesUsedByTask`. I have documented the return values and what happens if the string does not have the given prefix.

pkg/apis/pipeline/v1/task_types.go
chitrangpatel: I implemented your suggestion.

lbernick: Thanks! I understand why you're exporting the function ExtractVariablesFromString in the substitution package, but I still don't understand why it's necessary to have the function `FindWorkspaceSubstitutionLocationsInTask` in the workspace pkg instead of a function like `FindWorkspaceNamesUsedInTaskSpec`. The actual string substitutions can happen in the TaskRun reconciler instead

chitrangpatel: I think I implemented what you are suggesting. I wrote an exported function called `FindWorkspacesUsedByTask` in `pkg/workspaces` which actually returns the names of the workspaces that are used in the Task Spec. It internally uses `ExtractVariablesFromString` to parse the strings that may contain things like (`$(workspaces.foo.path)`) and extract the workspace names. Note, `ExtractVariablesFromString` is only parsing the strings, not doing the actual substitutions. The actual substitution is handled by the TaskRun reconciler.

pkg/substitution/substitution.go
lbernick: Thanks! One last comment; I think the boolean variable is redundant with the string slice that's returned (i.e. the bool is just len(stringslice) > 0 )

lbernick (approved):

pkg/substitution/substitution.go
chitrangpatel: I agree. I can create an issue on it and clean it up in a small PR.

jerop (approved):
thanks @chitrangpatel!

/lgtm

Issue #4643, opened 2022-03-03
Test Flake: Unit test cmd/entrypoint TestRealRunnerTimeout


lbernick:
Sadly not much info present in log file:
```
=== RUN   TestRealRunnerTimeout
    runner_test.go:37: step didn't timeout
--- FAIL: TestRealRunnerTimeout (0.03s)
```

/kind flake
/priority important-soon

abayer:
/lifecycle frozen

I hit this one while trying to reproduce the various events-related flakes, so it's still a thing. I thought that maybe increasing the timeout in the test would help, but I seem to still be hitting it...

abayer:
Yeah, it failing or not depended on what I had for `-count` and the overall load on my system, not on anything flaky. Sigh. Well, it's still a flake, just one I can't come up with a fix for...

Yongxuanzhang:
I'm looking into this issue and I'm not sure what the testcase wants to test. From the test description it wants to test that the runner will be killed after a millisecond (though it wants to sleep 10 ms?),
https://github.com/tektoncd/pipeline/blob/0b8349b770a76877051c9c790c94bf9ed897c75c/cmd/entrypoint/runner_test.go#L123
and the DeadlineExceeded error can be retuned in 2 places:
https://github.com/tektoncd/pipeline/blob/f83cd1f63bb216c76a3af46d58614d517113d1c5/cmd/entrypoint/runner.go#L120-L125
https://github.com/tektoncd/pipeline/blob/f83cd1f63bb216c76a3af46d58614d517113d1c5/cmd/entrypoint/runner.go#L154-L158
It seems that the DeadlineExceeded can still be returned if the runner really sleep 10 ms?

lbernick:
> I'm looking into this issue and I'm not sure what the testcase wants to test. From the test description it wants to test that the runner will be killed after a millisecond (though it wants to sleep 10 ms?),

The command here that the entrypoint is supposed to be running is "sleep 10ms". This test ensures that the runner can be timed out in the appropriate time. i.e. it wants to ensure that the runner times out after 1ms even though the command is not done.

> It seems that the DeadlineExceeded can still be returned if the runner really sleep 10 ms?

[...]

PR #5465, opened 2022-09-09
hardening `looksLikeResultRef` - params and when expressions

pritidesai:
# Changes

`looksLikeResultRef` checks for a prefix `task` and if a string contains `.result`. Any expression satisfying these two conditions are considered as a result reference, for example, all of the following expressions are considered as a result reference:

- `tasks.a-task.results.bResult` (valid reference)
- `tasks.a-task.resultTypo.bResult` (invalid reference)
- `task.result` (invalid reference)
- `tasks.a-task.b-task.c-task.results.result.result` (invalid reference)
-  etc. 

But, this way of checking for a result reference is weak. `looksLikeResultRef` returns true for some invalid expressions but not others, for example:

- `tasks.a-task.typoresults.bResult` (invalid reference)
- `task.typoresult` (invalid reference)
- etc

A pipeline with such expressions `tasks.a-task.resultsTypo.bResult` results in validation failure with a vague error `"invalid value: expected all of the expressions [tasks.a-task.resultTypo.bResult] to be result expressions but only [] were: tasks[0].params[a-param].value"` v/s a pipeline with an another invalid expression `tasks.a-task.typoresults.bResult` does not recognize this expression as an invalid result reference but considers it as a string constant. This PR drops the vague validation error for such expressions. 

Also, the validation function `validateParamResults` is also relying on the same weak `looksLikeResultRef` which is redundant with `validateGraph`. 

The validation function `validateGraph` which is called prior to `validateParamResults` is using `NewResultRef` to discover the dependencies and make sure a valid `DAG` can be constructed. The way we have implemented reporting `validation` errors with `errs.Also(`, the pipeline validation calls  the same function `NewResultRef` with the same set of expressions. This redundancy impacts the amount of time spent in validation. This PR deletes these redundant checks.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Hardening looksLikeResultRef to only report a valid result references, "tasks.<taskName>.results.<resultName>" and "tasks.<taskName>.results.<objectResultName>.<individualAttribute>". Any other invalid look-a-like-result-references are treated as constants. 
```

### DIFF
--- a/pkg/apis/pipeline/v1/pipeline_validation.go
+++ b/pkg/apis/pipeline/v1/pipeline_validation.go
@@ -53,7 +53,6 @@ func (ps *PipelineSpec) Validate(ctx context.Context) (errs *apis.FieldError) {
 	errs = errs.Also(ValidatePipelineTasks(ctx, ps.Tasks, ps.Finally))
 	// Validate the pipeline task graph
 	errs = errs.Also(validateGraph(ps.Tasks))
-	errs = errs.Also(validateParamResults(ps.Tasks))
 	// The parameter variables should be valid
 	errs = errs.Also(validatePipelineParameterVariables(ctx, ps.Tasks, ps.Params).ViaField("tasks"))
 	errs = errs.Also(validatePipelineParameterVariables(ctx, ps.Finally, ps.Params).ViaField("finally"))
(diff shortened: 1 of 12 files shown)


pritidesai:
thank you @abayer and @vdemeester for the reviews 👍 

Unit test failing with a known flake - https://github.com/tektoncd/pipeline/issues/4643

Issue #5154, opened 2022-07-18
TEP-0114: Custom Tasks Beta - References and Specifications


jerop:
In [TEP-0114: Custom Tasks Beta][tep-0114], we proposed migrating `Custom Tasks` and `Runs` to beta. This issue tracks this API changes to references and specifications in `Custom Tasks` and `Runs`.

Today, [embedded specs][runs-spec-docs] are in `Run.Spec.Spec`. In [TEP-0114: Custom Tasks Beta][tep-0114], we agreed to align the embedded specs with the rest of the API such that it's `CustomRun.Spec.CustomSpec`.

  ```yaml
  # before
  apiVersion: tekton.dev/v1alpha1
  kind: Run
  metadata:
    name: run-with-spec
  spec:
    spec:
      apiVersion: example.dev/v1alpha1
      kind: Example
      spec:
        field1: value1
        field2: value2
  # after
  apiVersion: tekton.dev/v1beta1
  kind: CustomRun
  metadata:
    name: customrun-with-spec
  spec:
    customSpec:
      apiVersion: example.dev/v1beta1
      kind: Example
      spec:
        field1: value1
        field2: value2
  ```

Today, [references][runs-spec-docs] are in `Run.Spec.Ref`. In [TEP-0114: Custom Tasks Beta][tep-0114], we agreed to align the references with the rest of the API such that it's `CustomRun.Spec.CustomRef`.

  ```yaml
  # before
  apiVersion: tekton.dev/v1alpha1
  kind: Run
  metadata:
    name: run-with-reference
  spec:
    ref:
      apiVersion: example.dev/v1alpha1
      kind: Example
      name: my-example
  ---
  # after
  apiVersion: tekton.dev/v1beta1
  kind: CustomRun
  metadata:
    name: customrun-with-reference
  spec:
    customRef:
      apiVersion: example.dev/v1beta1
      kind: Example
      name: my-customtask
  ```

Note that the API for `TaskRuns` and `PipelineRuns` is:

```yaml
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  name: pipelinerun-with-ref
spec:
  pipelineRef:
    name: my-pipeline
---
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  name: taskrun-with-reference
spec:
  taskRef:
    name: my-task
---
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  name: pipelinerun-with-spec
[...]

Issue #5153, opened 2022-07-18
TEP-0114: Custom Tasks Beta - Support `v1beta1.CustomRun`


jerop:
In [TEP-0114: Custom Tasks Beta][tep-0114], we proposed migrating `Custom Tasks` and `Runs` to beta. 

Moving `v1alpha1.Run` to `v1beta1.CustomRun` is a required API change in [TEP-0114: Custom Tasks Beta][tep-0114]. This issue tracks this API change.

[tep-0114]: https://github.com/tektoncd/community/blob/main/teps/0114-custom-tasks-beta.md

XinruZhang:
Putting a note here as a reminder: all the references about `v1alpha1.Run` should be migrated to `v1beta1.CustomRun`, for example: https://github.com/tektoncd/pipeline/blob/f1b68123b2de6c5fe05b0db3a98ca77b56e8a91f/pkg/apis/pipeline/v1alpha1/run_types.go#L101-L107

Thanks for @lbernick 's input!

pritidesai:
Pipeline WG - Blockers:

- What should go in RunSpec?
- How to support v1alpha1 Run along with v1beta1 CustomRun?

/cc @ScrapCodes

vdemeester:
On the second question, I would ask a question to the question : do we even want to support both ? `v1alpha1.Run` is alpha, we can pretty much do whatever. As already mentionned during a WG, if we give a timeline when `tektoncd/pipeline` will start creating `v1beta1.CustomRun` instead of `v1alpha1.Run`, announce it, it should give enough time for current `CustomTask` project to update to handle `v1beta1.CustomRun` (in addition to `v1alpha1.Run` or not — but that's on those project, not `tektoncd/pipeline` to support both APIs, or not).

PR #5470, opened 2022-09-12
Conversion code from v1alpha1.Run to v1beta1.CustomRun

vsinghai:
# Changes

Prior to this commit, there was no conversion for
v1alpha1.Run to v1beta1.CustomRun. Since the integration of CustomRun, it is pivotal to convert future uses of Run to a CustomRun.

In this commit, we create the conversion code to allow for the conversion of v1alpha1.Run to v1beta1.CustomRun.

Fixes issue [#5443](https://github.com/tektoncd/pipeline/issues/5443)

/kind feature
/cc @jerop 

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- n/a Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ ] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

PR #5471, opened 2022-09-12
Add v1beta1 version of CustomRun CRD

vsinghai:
# Changes

This commit adds a v1beta1 version of the CustomRun CRD, and support to the webhook.

Fixes issue [#5154](https://github.com/tektoncd/pipeline/issues/5154)
Fixes issue [#5153](https://github.com/tektoncd/pipeline/issues/5153)
Fixes issue [#5532](https://github.com/tektoncd/pipeline/issues/5532)

/cc @jerop
/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


vsinghai:
do I need `conversions`, like stated [here](https://github.com/tektoncd/pipeline/blob/df7abd26b9800c767fb49897aaaf945dd76ced3c/config/300-pipeline.yaml#L70-L77), for v1alpha1.Runs to v1beta1.CustomRuns?

XinruZhang:
> do I need `conversions`, like stated [here](https://github.com/tektoncd/pipeline/blob/df7abd26b9800c767fb49897aaaf945dd76ced3c/config/300-pipeline.yaml#L70-L77), for v1alpha1.Runs to v1beta1.CustomRuns?

From the API compatibility policy perspective, it's okay that we don't convert it (per `Rule #2` in [K8s API Deprecation Policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/)) (meaning we deprecate the entire alpha track of the Custom Task Run).

The reason that PipelineRun needs that conversion is we want to support both v1beta1 and v1 version at the same time.

lbernick:
> > do I need `conversions`, like stated [here](https://github.com/tektoncd/pipeline/blob/df7abd26b9800c767fb49897aaaf945dd76ced3c/config/300-pipeline.yaml#L70-L77), for v1alpha1.Runs to v1beta1.CustomRuns?
> 
> From the API compatibility policy perspective, it's okay that we don't convert it (per `Rule #2` in [K8s API Deprecation Policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/)) (meaning we deprecate the entire alpha track of the Custom Task Run).
> 
> The reason that PipelineRun needs that conversion is we want to support both v1beta1 and v1 version at the same time.

I think we would ideally like to have conversion here because it will allow us to support v1alpha1 Runs and v1beta1 CustomRuns at the same time. Technically it's not necessary since v1alpha1 Run is in alpha but it feels a bit user hostile to drop support for Runs at the same time we're adding support for CustomRuns. It will force people to copy all their Run definitions somewhere, delete them, upgrade tekton pipelines, and then recreate their old run definitions as CustomRuns. (This would also be the case if we rename the "Kind" of the existing CRD to "CustomRun".)

Unfortunately I'm not sure if there's an easy way to rename a CRD/implement conversion between two different "Kinds". One thing we could try is to add a v1beta1 version of Run that deserializes to the CustomRun go struct, in addition to adding a CustomRun CRD with a v1beta1 version, and probably deprecate the Run CRD after giving people a release or two to convert to the new CRD. (I'm not totally sure if this would work)

Another idea would be to have the new CRD and update the pipelinerun reconciler code to allow reconciling both v1alpha1 Runs and v1beta1 custom runs, and not implement conversion. We would probably want to remove support for v1alpha1 runs after a release or two. I'm fairly sure this would work.

happy to help here but want to hear others' thoughts first -- @tektoncd/core-maintainers is this making sense/do you know how to rename a CRD?

abayer:
I've never actually seen anyone rename a CRD - I'm pretty sure you can't convert between two different `Kind`s, yeah. Some research is definitely needed here.

XinruZhang:
I would say the main focus is how do we want our users to conceive the "alpha" feature.

IMHO alpha version track is designed to serve as experimental features, meaning we really don't want our users to rely on alpha features in their prod env. They are only suppose to use it for experiments and they should expect the alpha features can be different, backward incompatible across different releases.

However, IIUC, this "alpha" feature has existed for a long time and some users have already heavily relied on it (have they?), we can't ignore this real life situation, in this case, I strongly recommend we need to support both for a while, giving them enough buffer time to migrate.

But in order to form a formal standard around **stability** of each version track, I would suggest it might be better if we explicitly tell users that **supporting both alpha and beta versions this time is an exception**, this decision is not based on the guidance we'd like to follow, but based on the reality that we haven't actively promoting alpha features to beta in the past. While we are planning to do it in the future.

vsinghai:
Thanks for the great conversation @lbernick @XinruZhang @abayer . I would like to ask a couple of questions regarding this situation. To my understanding, `v1alpha1.Runs` and `v1beta1.CustomRuns`, in theory, should be mutually exclusive. Therefore, I am not clear as to why we should add the conversion, when we should rather push the user to start using `CustomRuns` while still providing support for `Runs` (for a limited amount of time). In the end, the whole point of the K8 version control is to deprecate the less stable versions, i.e. `v1alpha1` (where `Runs` is stored), and keep upgrading to stable versions. If we are trying to upgrade `Runs` to `CustomRuns` from `v1alpha` to `v1beta1`, but still allow the users to be able to use the `v1alpha1` feature, doesn't that negate the whole point of constantly trying to upgrade our K8 versioning? I understand @lbernick point that it would be harsh to the user to suddenly drop `Runs` out of the blue and force everyone to update their code, but would a slow push of `CustomRuns` paired with telling users that `Runs` will no longer be supported help solve that issue?

Just providing some food for thought here, let me know what you think!

lbernick:
> Thanks for the great conversation @lbernick @XinruZhang @abayer . I would like to ask a couple of questions regarding this situation. To my understanding, `v1alpha1.Runs` and `v1beta1.CustomRuns`, in theory, should be mutually exclusive. Therefore, I am not clear as to why we should add the conversion, when we should rather push the user to start using `CustomRuns` while still providing support for `Runs` (for a limited amount of time).

The main purpose of implementing conversion is so that we only have to support one golang version in our reconciler code, because each served version (version that users can use) gets converted into the same stored version (version stored in etcd and used by our reconciler code). If we aren't able to implement conversion, our reconciler code will have to support both v1alpha1 Run and v1beta1 CustomRun. Contrast this to PipelineRun: our stored version is v1beta1, but once we start serving v1, it will be converted into a v1beta1 PipelineRun for use in our reconciler, so the reconciler code only needs to handle one version at a time.

I agree we'd like to ask users to swap to customruns while still supporting runs for a short window.

> In the end, the whole point of the K8 version control is to deprecate the less stable versions, i.e. `v1alpha1` (where `Runs` is stored), and keep upgrading to stable versions. If we are trying to upgrade `Runs` to `CustomRuns` from `v1alpha` to `v1beta1`, but still allow the users to be able to use the `v1alpha1` feature, doesn't that negate the whole point of constantly trying to upgrade our K8 versioning? I understand @lbernick point that it would be harsh to the user to suddenly drop `Runs` out of the blue and force everyone to update their code, but would a slow push of `CustomRuns` paired with telling users that `Runs` will no longer be supported help solve that issue?

I agree we'd like to remove support for Runs (I'm not sure exactly what you mean by a "slow push" here). The question is just how best to create a release where we support both Runs and CustomRuns, before creating a release where we remove support for Runs.

vsinghai:
/cc @dibyom 

please review

jerop (approved):
There's no conversion in this PR, it's just adding a new CRD instead of a new version, so I think the conversion conversation should not be a blocker here - we can move it to https://github.com/tektoncd/pipeline/pull/5470

(I believe we don't need conversion because this is a new CRD instead of a new version)

lbernick:
I'm not sure this PR should go in yet-- we're exposing a CRD that doesn't actually do anything when you create it

config/300-customrun.yaml
```diff
@@ -0,0 +1,66 @@
...
+    app.kubernetes.io/part-of: tekton-pipelines
+    pipeline.tekton.dev/release: "devel"
+    version: "devel"
+spec:
+  group: tekton.dev
+  preserveUnknownFields: false
+  versions:
+  - name: v1beta1
+    served: true
+    storage: true
```
jerop, line 30: @vsinghai can we set both of these fields to false?

vsinghai: done

vsinghai:
Here is a [doc](https://docs.google.com/document/d/1sJGHtrPq8R2FuIpbUTjWUxQHSssVArapJuFwRC1FD04/edit#) with the roadmap for this PR. The doc summarizes the lengthy conversation we have had in this PR and what the team has agreed upon for next steps for the PR. 

Once again, thank you @lbernick @XinruZhang and @abayer for the great discussion!

PR #5476, opened 2022-09-12
Replace `$(context...)` values in resolver parameters

### LINKED ISSUE #5475: PipelineRef resolver part doesn't support `context` variable replacement
# Expected Behavior

Using the following `PipelineRun` should create the following `ResolutionRequest`.

```
kind: PipelineRun
metadata:
  name: simple-pipelinerun
spec:
  pipelineRef:
    resolver: foo
    params:
    - name: bar
      value: $(context.pipelinerun.name)
---
apiVersion: resolution.tekton.dev/v1alpha1
kind: ResolutionRequest
metadata:
  labels:
    resolution.tekton.dev/type: foo
  name: foo-bf6f3d56876eb1d6948a50c52c0d1d87
spec:
  params:
    bar: simple-pipelinerun
```

# Actual Behavior

It doesn't *resolve* the `context.*` variable, resulting in the following `ResolutionRequest`.

```
apiVersion: resolution.tekton.dev/v1alpha1
kind: ResolutionRequest
metadata:
  labels:
    resolution.tekton.dev/type: foo
  name: foo-bf6f3d56876eb1d6948a50c52c0d1d87
spec:
  params:
    bar: $(context.pipelinerun.name)
```

1. Create the above `PipelineRun`
2. See what `ResolutionRequest` it creates

# Additional Info

- Tekton Pipeline version: `main`

/assign @abayer

abayer:
# Changes

fixes #5475

I did not think of this previously, but now it works.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/resources/pipelineref.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelineref.go
@@ -74,6 +74,9 @@ func GetPipelineFunc(ctx context.Context, k8s kubernetes.Interface, tekton clien
 		return func(ctx context.Context, name string) (v1beta1.PipelineObject, error) {
 			params := map[string]string{}
 			stringReplacements, arrayReplacements, objectReplacements := paramsFromPipelineRun(ctx, pipelineRun)
+			for k, v := range getContextReplacements("", pipelineRun) {
+				stringReplacements[k] = v
+			}
 			replacedParams := replaceParamValues(pr.Params, stringReplacements, arrayReplacements, objectReplacements)
 			for _, p := range replacedParams {
 				params[p.Name] = p.Value.StringVal
--- a/pkg/reconciler/pipelinerun/resources/apply.go
+++ b/pkg/reconciler/pipelinerun/resources/apply.go
@@ -140,16 +140,19 @@ func paramsFromPipelineRun(ctx context.Context, pr *v1beta1.PipelineRun) (map[st
 	return stringReplacements, arrayReplacements, objectReplacements
 }
 
-// ApplyContexts applies the substitution from $(context.(pipelineRun|pipeline).*) with the specified values.
-// Currently supports only name substitution. Uses "" as a default if name is not specified.
-func ApplyContexts(ctx context.Context, spec *v1beta1.PipelineSpec, pipelineName string, pr *v1beta1.PipelineRun) *v1beta1.PipelineSpec {
-	replacements := map[string]string{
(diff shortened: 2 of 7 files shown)


vdemeester (approved):

pkg/reconciler/pipelinerun/resources/pipelineref.go
```diff
@@ -74,6 +74,9 @@ func GetPipelineFunc(ctx context.Context, k8s kubernetes.Interface, tekton clien
 		return func(ctx context.Context, name string) (v1beta1.PipelineObject, error) {
 			params := map[string]string{}
 			stringReplacements, arrayReplacements, objectReplacements := paramsFromPipelineRun(ctx, pipelineRun)
+			for k, v := range getContextReplacements("", pipelineRun) {
```
lbernick, line 77: Is it possible to get the Pipeline name here? Do all resolvers need a param that is the pipeline name, or can we not make that guarantee?

abayer: We can't make any guarantee, no, in almost all cases. In practice, the hub or bundles resolver params include what we expect will be the `Pipeline` name, but even then, we can't _guarantee_ it - the actual `Pipeline` could have a different name than its key, e.g. The cluster resolver does know the name, but the git resolver has no idea.

lbernick: Could we add a new field to resolver for Pipeline name instead of passing it in as a param? I would imagine almost all resolvers would need it

abayer: I don't think we should - we never create an actual `Pipeline`, we just load the spec in to `PipelineRun.Spec.PipelineSpec`...

vdemeester: I don't think we should either.

abayer: That and while the name happens to be a "key" for three of the current resolvers, there's no reason to assume that'll be the case for all or even necessarily a majority of future resolvers. Plus adding another `name`-related field to `pipelineRef`/`taskRef` would be pretty dang confusing.

lbernick: ok, sounds good

Issue #5302, opened 2022-08-10
FR: pipelineRun.spec.taskRunTemplate


lbernick:
### Feature request

We already have [pipelineRun.spec.taskRunSpecs](https://github.com/tektoncd/pipeline/blob/bf8be59f43dd81ef2f26102ae5b3c91e22e30bdb/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L229), but it might be nice to be able to specify a TaskRun configuration that applies to every TaskRun in the PipelineRun  by default. An analogous feature is [task.spec.stepTemplate](https://github.com/tektoncd/pipeline/blob/bf8be59f43dd81ef2f26102ae5b3c91e22e30bdb/pkg/apis/pipeline/v1beta1/task_types.go#L113).

### Example use cases

- I would like to use the same service account for all of my pipeline tasks except one.
- I want to specify compute resources that each TaskRun in my PipelineRun should run with, and don't want to have to specify them individually for each TaskRun.

yuzp1996:
taskRunTemplate should be a field of pipelineRun.spec or a field of pipeline.spec ?

lbernick:
I think we'd want this to be on PipelineRun, since the use cases I mentioned (service account and compute resources) are probably things you'd want to specify for a pipelinerun rather than a Pipeline. However (I should have mentioned this in the initial issue) I think we should have a small TEP (design proposal) for this before implementing it just to make sure there are no concerns from maintainers. Is that something you'd be willing to do, with help from me?

yuzp1996:
> I think we'd want this to be on PipelineRun, since the use cases I mentioned (service account and compute resources) are probably things you'd want to specify for a pipelinerun rather than a Pipeline. However (I should have mentioned this in the initial issue) I think we should have a small TEP (design proposal) for this before implementing it just to make sure there are no concerns from maintainers. Is that something you'd be willing to do, with help from me?

Of course! I would like to do it!  Thanks for your help！@lbernick     

Maybe I should create a TEP PR in [Tekton/community](https://github.com/tektoncd/community/tree/main/teps) first?

lbernick:
Thank you, that would be amazing! The instructions for creating a new TEP are [here](https://github.com/tektoncd/community/tree/main/teps#creating-and-merging-teps). Please let me know if you need any help with this.

yuzp1996:
[...]

PR #5466, opened 2022-09-09
Move and update v1beta1 to v1 migration doc from Indexed

JeromeJu:
# Changes

This commit follows up on #5415 and resolves the respective examples and docs on the migration.

Fixes: #5330 
/kind documentation

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [n/a] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
The document has been added that describes the changes needed when migrating from v1beta1 to v1.
```

PR #5491, opened 2022-09-13
TEP-119 - Add TaskRun Template in v1 PipelineRun

JeromeJu:
# Changes

This commit adds the TaskRunTemplate in v1 PipelineRun Spec. It moves
the serviceAccountName and podTemplate from PipelineRun Spec. The rest
of the TaskRunTemplate fields are to be implemented after releasing v1.

part of #5302 
Implementing [TEP-119](https://github.com/tektoncd/community/blob/main/teps/0119-add-taskrun-template-in-pipelinerun.md)
/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Add `pipelineRun.spec.pipelineTaskRunTemplate` with `serviceAccountName` and `podTemplate`.
```


JeromeJu:
Would like to confirm that since for this step we are just moving two fields into a new struct, keeping the existing test cases shall be sufficient with no new unit tests introduced?

JeromeJu:
/assign @lbernick 
cc @yuzp1996

pkg/apis/pipeline/v1/pipelinerun_types.go
```diff
@@ -563,3 +563,13 @@ func (pr *PipelineRun) GetTaskRunSpec(pipelineTaskName string) PipelineTaskRunSp
 	}
 	return s
 }
+
+// PipelineTaskRunTemplate is used to specify run specifications for all Task in pipelinerun.
+type PipelineTaskRunTemplate struct {
+	// +optional
+	Metadata *PipelineTaskMetadata `json:"metadata,omitempty"`
```
lbernick, line 570: let's not add metadata yet, since we don't have a way of implementing it until we swap the stored version over to v1. does that make sense?

JeromeJu:
> can you please also update the v1beta1 -> v1 migration doc?

Thanks @lbernick. Was right about to ask if I can put those changes in this #5466 PR?

pkg/apis/pipeline/v1/pipelinerun_types.go
JeromeJu: Yes I see. Thanks!

lbernick:
> > can you please also update the v1beta1 -> v1 migration doc?
> 
> Thanks @lbernick. Was right about to ask if I can put those changes in this #5466 PR?

I'd rather not-- PRs should be self contained and have all the code + docs related to a single change. If you add those docs to your other PR but keep the code changes in this one, when that PR is merged the docs will not match the code.

docs/migrating-v1beta1-to-v1.md
```diff
@@ -89,3 +92,38 @@ spec:
...
+  pipelineRef:
+    name: clone-test-build
+  taskRunSpecTemplate:
+    serviceAccountName: build
+    podTemplate:
+      securityContext:
+        fsGroup: 65532
+```
+
+*Currently, only `TaskRunTemplate.ServiceAccountName` and `TaskRunTemplate.PodTemplate` are implemented in v1 and the rest of the fields would be added after v1 release.
```
lbernick, line 128: I think this line can just be removed

lbernick (approved):

vdemeester:
> The rest of the TaskRunTemplate fields are to be implemented after releasing v1.

I am not sure I understand why "after" ?

JeromeJu:
Thanks @vdemeester for the comment. My take is that `PodTemplate` and `ServiceAccountName` are moved before the release to avoid back compatible issues as we would not want to support them in v1 pipelinerun.spec. Could you help confirm my understanding here @lbernick ?

lbernick:
> Thanks @vdemeester for the comment. My take is that `PodTemplate` and `ServiceAccountName` are moved before the release to avoid back compatible issues as we would not want to support them in v1 pipelinerun.spec. Could you help confirm my understanding here @lbernick ?

Yes. The distinction here is that we don't have a great way to support the planned new fields `taskRunTemplate.metadata`, `taskRunTemplate.computeResources`, etc, until we swap the stored version of our CRDs to v1. However, this isn't something the user needs to know about, so I suggest removing the line from the docs about when those features will be added.

PR #5501, opened 2022-09-15
POC: Canceling concurrent PipelineRuns

lbernick:
# Changes

Alternative to https://github.com/tektoncd/triggers/pull/1446.

This commit adds a proof of concept for "concurrency primitives" in Pipelines that will allow higher-level controllers to implement concurrency controls. It adds a new CRD, "ConcurrencyControl". PipelineRuns that should have their concurrency controlled should be created with a label referencing the ConcurrencyControl. The PipelineRun controller will create another label on the PipelineRun specifying its "concurrency key". Any PipelineRuns with the same concurrency key belong to the same concurrency group and use the same concurrency strategy.

This solution doesn't require PipelineRuns to be created as Pending. The reason for this is that knative does not allow a reconciler to edit the spec of an object being reconciled. In other words, the PipelineRun controller cannot "start" Pending PipelineRuns after canceling other PipelineRuns in the same concurrency group. Instead, concurrency controls are just handled before execution starts.

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```


dibyom:
> This solution doesn't require PipelineRuns to be created as Pending. The reason for this is that knative does not allow a reconciler to edit the spec of an object being reconciled. In other words, the PipelineRun controller cannot "start" Pending PipelineRuns after canceling other PipelineRuns in the same concurrency group. Instead, concurrency controls are just handled before execution starts.

The ReconcileKind interface does not update changes to the spec, but one can always make a call to update the status from within the ReconcileKind function.

lbernick:
> > This solution doesn't require PipelineRuns to be created as Pending. The reason for this is that knative does not allow a reconciler to edit the spec of an object being reconciled. In other words, the PipelineRun controller cannot "start" Pending PipelineRuns after canceling other PipelineRuns in the same concurrency group. Instead, concurrency controls are just handled before execution starts.
> 
> The ReconcileKind interface does not update changes to the spec, but one can always make a call to update the status from within the ReconcileKind function.

for sure! I mention this because I didn't want to "run" a pending pipelinerun by editing its status, because I thought it would be confusing/introduce bugs to have spec status = Pending but then have the PR actually executing. so I made this not dependent on creating pipelineruns as pending.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -704,6 +707,47 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
...
+		// Assume concurrency key was added in a previous reconcile loop, and other pipelineRuns have already been canceled.
+		return nil
+	}
+	cc, err := r.concurrencyControlLister.ConcurrencyControls(pr.Namespace).Get(concurrencyRef)
+	if err != nil {
+		return fmt.Errorf("error getting concurrency control for PR %s: %s", pr.Name, err)
+	}
+	key := resources.ApplyParametersToConcurrency(ctx, cc, pr)
+	pr.Labels["tekton.dev/concurrency-key"] = key
+	if cc.Spec.Strategy != "Cancel" {
```
abayer, line 726: I'm assuming we'd create constants for supported strategies?

lbernick: definitely!

abayer: Just making sure. =)

abayer:
This feels like it's on the right track!

docs/concurrency.md
```diff
@@ -0,0 +1,47 @@
...
+
+```yaml
+apiVersion: tekton.dev/v1alpha1
+kind: ConcurrencyControl
+metadata:
+  name: my-concurrency-control
+spec:
+  params:
+  - name: param-1
+  - name: param-2
```
vdemeester, line 14: Those would be "populated" by the `PipelineRun` params ?
What happens if there is a mismatch between `PipelineRun`'s params and that `ConcurrencyControl` params?

vdemeester (commented):
I do like that the `ConcurrencyControl` CRD holds the strategy and "how" the key is computed.
What is "slightly" weird though, is the *implicitness* of *how* it is computed. With default value, we can probably get away with it, but it might lead to weird behavior from time to time, if, for example, a parameter is renamed and thus, doesn't map to the `ConcurrencyControl` anymore.

lbernick:
> I do like that the `ConcurrencyControl` CRD holds the strategy and "how" the key is computed. What is "slightly" weird though, is the _implicitness_ of _how_ it is computed. With default value, we can probably get away with it, but it might lead to weird behavior from time to time, if, for example, a parameter is renamed and thus, doesn't map to the `ConcurrencyControl` anymore.

I agree! Do you have any suggestions for what might work better? We could remove "key" and "params" entirely, but I think this will be challenging if we want to sort things into concurrency groups based on parameters from a trigger or workflow (which I think we'll want to do).

lbernick:
Closing in favor of https://github.com/tektoncd/experimental/pull/895

Issue #5500, opened 2022-09-15
Conversion error for parameters when pipelineref.bundle is used


lbernick:
Webhook error messages show:

```
{"severity":"error","timestamp":"2022-09-15T14:38:48.420Z","logger":"tekton-pipelines-webhook","caller":"conversion/conversion.go:59","message":"Conversion failed","commit":"698f8a7-dirty","uid":"1c56f711-20c2-454d-90d1-0210280ee8f5","desiredAPIVersion":"tekton.dev/v1","error":"unable to marshal output: json: error calling MarshalJSON for type v1.ParamValue: impossible ParamValues.Type: \"\"","stacktrace":"knative.dev/pkg/webhook/resourcesemantics/conversion.(*reconciler).Convert\n\tknative.dev/pkg@v0.0.0-20220818004048-4a03844c0b15/webhook/resourcesemantics/conversion/conversion.go:59\nknative.dev/pkg/webhook.conversionHandler.func1\n\tknative.dev/pkg@v0.0.0-20220818004048-4a03844c0b15/webhook/conversion.go:64\nnet/http.HandlerFunc.ServeHTTP\n\tnet/http/server.go:2109\nnet/http.(*ServeMux).ServeHTTP\n\tnet/http/server.go:2487\nknative.dev/pkg/webhook.(*Webhook).ServeHTTP\n\tknative.dev/pkg@v0.0.0-20220818004048-4a03844c0b15/webhook/webhook.go:262\nknative.dev/pkg/network/handlers.(*Drainer).ServeHTTP\n\tknative.dev/pkg@v0.0.0-20220818004048-4a03844c0b15/network/handlers/drain.go:113\nnet/http.serverHandler.ServeHTTP\n\tnet/http/server.go:2947\nnet/http.(*conn).serve\n\tnet/http/server.go:1991"}
```

I believe the issue is that [param value conversion](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/param_conversion.go#L63-L75) doesn't handle param type "" (i.e. param type not set-- corresponds to string type)

vdemeester:
@lbernick what version does it affect ?

JeromeJu:
> @lbernick what version does it affect ?

Looks like it's the stored version as of v1beta1 according to https://github.com/tektoncd/pipeline/pull/5506#discussion_r972979172.

lbernick:
[...]

PR #5506, opened 2022-09-15
Fix empty paramValueType conversion

JeromeJu:
# Changes

This commit fixes the paramValueType that is not handled if input is an empty string,
which could lead to issues for 'marshal' in knative.

FIx: #5500 
/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/param_conversion.go
+++ b/pkg/apis/pipeline/v1beta1/param_conversion.go
@@ -8,7 +8,11 @@ import (
 
 func (p ParamSpec) convertTo(ctx context.Context, sink *v1.ParamSpec) {
 	sink.Name = p.Name
-	sink.Type = v1.ParamType(p.Type)
+	if p.Type != "" {
+		sink.Type = v1.ParamType(p.Type)
+	} else {
+		sink.Type = v1.ParamType(ParamTypeString)
+	}
 	sink.Description = p.Description
 	var properties map[string]v1.PropertySpec
 	if p.Properties != nil {
@@ -28,7 +32,11 @@ func (p ParamSpec) convertTo(ctx context.Context, sink *v1.ParamSpec) {
 
 func (p *ParamSpec) convertFrom(ctx context.Context, source v1.ParamSpec) {
 	p.Name = source.Name
-	p.Type = ParamType(source.Type)
+	if source.Type != "" {
+		p.Type = ParamType(source.Type)
+	} else {
+		p.Type = ParamTypeString
+	}
 	p.Description = source.Description
 	var properties map[string]PropertySpec
 	if source.Properties != nil {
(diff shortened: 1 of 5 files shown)


pkg/apis/pipeline/v1beta1/param_conversion.go
```diff
@@ -61,13 +69,22 @@ func (p *Param) convertFrom(ctx context.Context, source v1.Param) {
 }
 
 func (v ParamValue) convertTo(ctx context.Context, sink *v1.ParamValue) {
-	sink.Type = v1.ParamType(v.Type)
+	if v.Type != "" {
```
JeromeJu, line 72: Thanks to the discussion with @Yongxuanzhang I changed both the conversion for `paramValue` and `paramSpec`. But I was wondering in which case would we bump into the conversion of an empty string input? Thought the webhook should already taken care of that in the earlier stage?

lbernick: This can happen with resources already stored in etcd without a type set (that were created before we added defaulting logic for param values)

pkg/apis/pipeline/v1beta1/taskrun_conversion_test.go
```diff
@@ -78,13 +78,15 @@ func TestTaskrunConversion(t *testing.T) {
 					Name: "param-task-1",
 					Value: v1beta1.ParamValue{
 						ArrayVal: []string{"value-task-1"},
+						Type:     "string",
```
lbernick, line 81: Do you know why tests passed before you added this?

JeromeJu: I think previously these remain empty string for `paramSpec` and `paramValue` so the round trip would pass.

JeromeJu: And this also made wonder if we need extra tests with the empty string for paramType?

lbernick: yes, I think we do need a new test case. I thought about it some more and I think your explanation is not quite right-- if the test input is an empty string, and the type cast is what's failing, ConvertTo would have failed. I think my original hypothesis in the bug was wrong; it's not the type cast that failed, it's that the output was type = "" which knative isn't able to marshal (serialize). Anyway your approach here is still correct, please just add the new test case.

vdemeester (approved):

lbernick:
@JeromeJu it looks like you might have accidentally included a go binary-- happy to lgtm once that's removed

JeromeJu:
> @JeromeJu it looks like you might have accidentally included a go binary-- happy to lgtm once that's removed

Thanks @lbernick Sorry that I'm not sure how my vscode generated this.

vdemeester:
Well, I realize I did the release without this fix 😓 
So I am moving it to 0.40.0 and will cherry-pick for a 0.40.1 near the end of the week 🙃

PR #5511, opened 2022-09-16
Update API spec doc to reflect the new fields introduced from TEP75&76

chuangw6:
# Changes

Implementation of TEP75&76 supported object type param, object type result and array type result, which introduced some changes to the API spec i.e. `ParamSpec`, `TaskResult` and `TaskRunResult`.

This PR updates the doc to reflect the changes.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/api-spec.md
+++ b/docs/api-spec.md
@@ -244,7 +244,8 @@ List responses have the following fields (based on [`meta.v1/ListMeta`](https://
 |---------------|------------|-------------|
 | `name`        | string     | REQUIRED    |
 | `description` | string     | REQUIRED    |
-| `type`        | Enum: <br>- `"string"` (default) <br>- `"array"` | REQUIRED |
+| `type`      | Enum:<br>- `"string"` (default)<br>- `"array"` <br>- `"object"` | REQUIRED (The values `string` and `array` for this field are REQUIRED, and the value `object` is RECOMMENDED.) |
+| `properties`  | map<string,PropertySpec> |RECOMMENDED <br><br>note: `PropertySpec` is a type that defines the spec of an individual key. See how to define the `properties` section in the [example](../examples/v1beta1/taskruns/alpha/object-param-result.yaml).|
 | `default`     | `ParamValue` | REQUIRED |
 
 ### `Step`
@@ -279,13 +280,16 @@ List responses have the following fields (based on [`meta.v1/ListMeta`](https://
 |--------------|------------|-------------|
 | `name`        | string    | REQUIRED    |
 | `description` | string    | REQUIRED    |
+| `type`      | Enum:<br>- `"string"` (default)<br>- `"array"` <br>- `"object"` | RECOMMENDED (Each of the values is RECOMMENDED.)|
+| `properties`  | map<string,PropertySpec> | RECOMMENDED <br><br>note: `PropertySpec` is a type that defines the spec of an individual key. See how to define the `properties` section in the [example](../examples/v1beta1/taskruns/alpha/object-param-result.yaml).|
 
 ### `TaskRunResult`
 
 | Field Name | Field Type | Requirement |
 |------------|------------|-------------|
 | `name`     | string     | REQUIRED    |
-| `value`    | string     | REQUIRED    |
+| `type`      | Enum:<br>- `"string"` (default)<br>- `"array"` <br>- `"object"` | RECOMMENDED (Each of the values is RECOMMENDED.) |
+| `value`    | `ParamValue`     | REQUIRED    |
 
 ### `TaskSpec`
 


chuangw6:
@Yongxuanzhang Please take a look. Let me know if I miss anything. Thanks

lbernick:
I don't think we want to mark array results and object params + results as required for conformance especially since these fields are still in alpha. [TEP-0012](https://github.com/tektoncd/community/blob/main/teps/0012-api-spec.md) says that we may add "an optional process to very mature supported features, to potentially mark them as required by the spec", and [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#design-evaluation) states that "Fields should not be added to the Tekton conformance surface until they are promoted to the same stability level as the CRD impacted". Maybe "recommended" would be better.

chuangw6:
> I don't think we want to mark array results and object params + results as required for conformance especially since these fields are still in alpha. [TEP-0012](https://github.com/tektoncd/community/blob/main/teps/0012-api-spec.md) says that we may add "an optional process to very mature supported features, to potentially mark them as required by the spec", and [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#design-evaluation) states that "Fields should not be added to the Tekton conformance surface until they are promoted to the same stability level as the CRD impacted". Maybe "recommended" would be better.

Thanks for the pointers @lbernick ! We are not marking specifically array result, object param + result as required. It's the `type` field. And I agree that this is indeed not a required field since the mutating webhook will set it if not provided. In addition, I feel like some other fields marked as `required` should also be changed to `recommended` or `optional` i.e. `description`, paramSpec's `default`. wdyt?

chuangw6:
@lbernick I see. I've updated to mark the type field in `TaskResult` and `TaskRunResult` as `recommended` for the object and array.
PTAL! Thanks

docs/api-spec.md
```diff
@@ -244,7 +244,8 @@ List responses have the following fields (based on [`meta.v1/ListMeta`](https://
 |---------------|------------|-------------|
 | `name`        | string     | REQUIRED    |
 | `description` | string     | REQUIRED    |
-| `type`        | Enum: <br>- `"string"` (default) <br>- `"array"` | REQUIRED |
+| `type`      | Enum:<br>- `"string"` (default)<br>- `"array"` <br>- `"object"` | REQUIRED for `string` and `array`type <br> RECOMMENDED for `object` type |
```
lbernick, line 247: I think it would make sense to phrase this as the "type" field is required, and the values "string" and "array" for this field are required, and the value "object" is recommended.

docs/api-spec.md
```diff
@@ -279,13 +280,16 @@ List responses have the following fields (based on [`meta.v1/ListMeta`](https://
 |--------------|------------|-------------|
 | `name`        | string    | REQUIRED    |
 | `description` | string    | REQUIRED    |
+| `type`      | Enum:<br>- `"string"` (default)<br>- `"array"` <br>- `"object"` | REQUIRED for `string` type <br> RECOMMENDED for `array` and `object` |
```
lbernick, line 283: "type" is new, so I think type has to be "recommended" for results. each of the supported values should be "recommended" as well.

chuangw6: SG! Updated!

docs/api-spec.md
```diff
@@ -279,13 +280,16 @@ List responses have the following fields (based on [`meta.v1/ListMeta`](https://
...
+| `properties`  | map<string,PropertySpec> | RECOMMENDED <br><br>note: `PropertySpec` is a type that defines the spec of an individual key. See how to define the `properties` section in the [example](../examples/v1beta1/taskruns/alpha/object-param-result.yaml).|
 
 ### `TaskRunResult`
 
 | Field Name | Field Type | Requirement |
 |------------|------------|-------------|
 | `name`     | string     | REQUIRED    |
-| `value`    | string     | REQUIRED    |
+| `type`      | Enum:<br>- `"string"` (default)<br>- `"array"` <br>- `"object"` | RECOMMENDED (Each of the values is RECOMMENDED.) |
+| `value`    | `ParamValue`     | REQUIRED    |
```
lbernick, line 292: it's a bit unfortunate that we're changing the type of a required field but since we only have one primary implementation of tekton currently it's probably OK?

lbernick (approved):

docs/api-spec.md
chuangw6: I guess so because it's a backward compatible change. String type result is still supported as the way it was supported before.

lbernick: Just to clarify, if there were other implementations of the Tekton API, this would be a problem. The reason is that they might support the field "value" as a string, but with this change they are "required" to support value as a "ParamValue". This would mean an implementation that was previously conformant would not be conformant after this change. But since we have one main implementation, it doesn't matter too much.

chuangw6: Thanks for the clarification!

Issue #5514, opened 2022-09-19
Deprecate `TaskRef.Bundle` and `PipelineRef.Bundle` in favor of the bundles resolver


abayer:
Relates to #4710 

With remote resolution coming out of `alpha` in v0.41.0, the legacy `taskRef.bundle` and `pipelineRef.bundle` syntax should be deprecated in favor of using the `bundles` resolver.

vdemeester:
@abayer is it https://github.com/tektoncd/pipeline/pull/5131 ?

abayer:
Basically, but I think it's worth doing it as a new PR, since I'm also going to change the legacy bundle syntax to be handled by the bundles resolver.

PR #5436, opened 2022-09-06
Move the remote resolvers deployment to a separate namespace

abayer:
# Changes

In https://github.com/tektoncd/community/blob/main/teps/0060-remote-resource-resolution.md#3-create-a-new-tekton-resolution-project, we say:
```
Webhook and controller deployments for ResolutionRequest objects will run in the tekton-remote-resolution namespace by default. The namespace is intentionally separate from tekton-pipelines to allow RBAC that isolates the remote resolution machinery.
```

While the actual `ResolutionRequest` reconciler/webhooks/etc are now part of the core Pipeline controller/webhook deployments, the actual resolvers (i.e., `git`, `hub`, `bundles`) should probably still be moved to a separate namespace from `tekton-pipelines` to properly isolate the actual resolver executions. So let's do that.

Note that no release note is needed, because this should be merged in before the first Pipeline release with remote resolution incorporated into it.

/kind feature

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ ] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

PR #5515, opened 2022-09-19
Move remote resolution out of alpha

### LINKED ISSUE #4710: Promote Remote Resolution to beta

abayer:
# Changes

Closes #4710

There are two major aspects to this change:

* Creating a `v1beta1` for `ResolutionRequest`, changing from `parameters` as a `map[string]string` to `params` (field name rename for `openapi-gen` API rule validation reasons) as a `[]pipelinev1beta1.Param`. This is being done to match `ResolverRef.Params`, which is a `[]pipelinev1beta1.Param`. We originally left `ResolutionRequest.Parameters` as `map[string]string` since it already was that in `v1alpha1` and changing it would require `ResolutionRequest` `v1beta1`, hence doing this now. Changing to `[]pipelinev1beta1.Param` is needed to support array and object params for resolvers, and we're going with a slice rather than a map because that's how you're supposed to do this in Kubernetes. =)
* Promoting the remote resolution functionality from `alpha` to `beta`, meaning on by default for `v1beta1` `Pipeline`s etc, and requiring `enable-api-fields: beta` for `v1` `Pipeline`s etc. Additionally, we're undoing the questionable decision to split the resolvers out into `resolvers.yaml`, separate from `release.yaml`, and enabling the built-in resolvers.

Deprecation of the legacy `taskRef.bundle` and `pipelineRef.bundle` syntax will be handled in a separate PR (#5514).

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
action required: Any resolvers being used other than built-in resolvers will need to be updated to use ResolutionRequest v1beta1. Remote resolution of Pipelines and Tasks promoted to beta, and ResolutionRequest v1beta1 introduced to support array and object parameters for resolvers.
```


abayer:
/hold

After discussion with @vdemeester about whether we should stop having the separate `resolvers.yaml` and instead include `config/resolvers` in `release.yaml` (which will let us assume that the resolvers are installed and enabled by default, which is important for things like the deprecation/eventual removal of `taskRef.bundle` and `ClusterTask` in favor of their resolver equivalents), I've made some changes (currently in a separate commit):
* `release.yaml` does now include `config/resolvers`.
* A new `minimal-release.yaml` will be created, with `config/resolvers` excluded from it, matching the contents of `release.yaml` as of v0.40.x.
* `resolvers.yaml` will still be created, but is only needed/used if you want to add the built-in resolvers after installing Pipelines with `minimal-release.yaml`.
* Docs are updated accordingly.

abayer:
/hold cancel

Ok, squashed the `release.yaml` changes in, removing the hold.

config/300-resolutionrequest.yaml
```diff
@@ -33,8 +33,8 @@ spec:
       - resolutionrequests
   versions:
     - name: v1alpha1
-      served: true
-      storage: true
+      served: false
```
lbernick, line 36: would it make sense to keep served: true and add deprecated: true? Is it possible that people have resolutionrequests saved in their cluster and this could break them?

docs/resolution-getting-started.md
```diff
@@ -24,16 +24,11 @@ with a Pipeline stored in a git repo.
...
-Tekton Pipelines currently has its integration with remote resolution behind
-the alpha feature gate, and enabling specific resolvers is controlled by feature 
-flags as well:
+Remote resolvers for Tekton Pipelines are enabled by feature flags:
 
 ```sh
-# update the feature-flags configmap in the tekton-pipelines namespace
-kubectl patch -n tekton-pipelines configmap feature-flags -p '{"data":{"enable-api-fields":"alpha","enable-git-resolver":"true"}}'
-
 # update the resolvers-feature-flags configmap in the tekton-pipelines-resolvers namespace
```
lbernick, line 32: wait, do we have two configmaps where users need to enable resolvers? Or was this a docs mistake?

test/tektonbundles_test.go
```diff
@@ -54,6 +54,10 @@ var requireFeatureFlags = requireAnyGate(map[string]string{
 	"enable-api-fields":         "alpha",
 })
```
lbernick, line 55: can this variable (requireFeatureFlags) be removed now?

docs/resolution-getting-started.md
abayer: The `enable-git-resolver` bit there should have been removed in #5436.

test/tektonbundles_test.go
abayer: That'll be part of #5514

config/300-resolutionrequest.yaml
lbernick: When you say they are short-lived, do you mean they're not referred to again after resolution, or they are deleted after resolution? If the latter, I think we still need to serve v1alpha1 because there will still be resolution requests present on some clusters. But ofc that doesn't matter post change-- thanks!

abayer: Not referred to again - we probably should be cleaning them up automatically at some point, though. But yeah, thanks for bringing this up!

vdemeester: @abayer aren't they "owned" (through ownerRef) by the `PipelineRun` that created it ? If yes, then yes they should be deleted when the `PipelineRun` is deleted (same applies for `TaskRun`).

vdemeester:
> Why should we skip beta and go to stable right away?

What do you mean by `beta` ?

vdemeester:
@afrittoli as of today "promoted to stable" means to `v1beta1`, which is *beta* anyway. I don't think we want to have 3 levels of `enable-api-fields` as it's gonna become increasingly weird and confusing to manage. We probably need to refine the TEP that introduced that and maybe use `unstable` and `stable` term to not create any confusion between API version and this feature-flag.

Do we want to keep it in beta a little more before doing a `v1` is a good question, and I would vote for as, for me, remote resolution *should be* a `v1` requirement.

afrittoli:
> Based on discussions around promoting features last week, I'm opting to skip the need for a `beta` flag and take remote resolution straight to `stable`.

From the PR description:

> Based on discussions around promoting features last week, I'm opting to skip the need for a beta flag and take remote resolution straight to stable.

I understand this would mean removing this from the feature flag so that it would be available right away in `v1beta1` and in `v1`.

vdemeester:
@afrittoli yes indeed, but mainly, because I don't think we want to have a `beta` value for that feature-flag (as written in my previous comment 🙃 )

afrittoli:
> @afrittoli as of today "promoted to stable" means to `v1beta1`, which is _beta_ anyway. I don't think we want to have 3 levels of `enable-api-fields` as it's gonna become increasingly weird and confusing to manage. 

Having three levels of `enable-api-fields` was the plan when we wrote the TEP, and I think it's still a valid approach. 
Using `enable-api-fields=alpha` means that alpha features are available in the `v1beta1` and in the `v1` API. 
Using `enable-api-fields=beta` means that beta features are available in the `v1` API (they are available in the `v1beta` API by default).

As things are defined today, once we have `v1`, the road of a new API feature would be:
- introduced as `alpha` - only available when `enable-api-fields=alpha` is set, since we do not serve any `alpha` API. When the feature flag is set, the feature becomes available for `v1beta1` and `v1` resources alike
- promoted to `beta` - it becomes available for `v1beta` resources by default. Only available for `v1` resources when `enable-api-fields=beta`
- promoted to `stable` - generally available in all APIs, no flag required anymore 

The only alternative for such a promotion path that I can see, would be to implement new API features in proper API versions, which would mean that users need to use a different resource version to access unstable API features and that we need to maintain more API versions, with all the conversion etc.

/cc @wlynch

vdemeester:
> > @afrittoli as of today "promoted to stable" means to `v1beta1`, which is _beta_ anyway. I don't think we want to have 3 levels of `enable-api-fields` as it's gonna become increasingly weird and confusing to manage.
> 
> Having three levels of `enable-api-fields` was the plan when we wrote the TEP, and I think it's still a valid approach. Using `enable-api-fields=alpha` means that alpha features are available in the `v1beta1` and in the `v1` API. Using `enable-api-fields=beta` means that beta features are available in the `v1` API (they are available in the `v1beta` API by default).
> 
> As things are defined today, once we have `v1`, the road of a new API feature would be:
> 
> * introduced as `alpha` - only available when `enable-api-fields=alpha` is set, since we do not serve any `alpha` API. When the feature flag is set, the feature becomes available for `v1beta1` and `v1` resources alike
> * promoted to `beta` - it becomes available for `v1beta` resources by default. Only available for `v1` resources when `enable-api-fields=beta`
> * promoted to `stable` - generally available in all APIs, no flag required anymore

Yes, and looking back, I think/feel we made a mistake there. We are effectively bypassing API version rules with this feature-flag, so that we can end-up keep enriching one API version instead of writing more.

> The only alternative for such a promotion path that I can see, would be to implement new API features in proper API versions, which would mean that users need to use a different resource version to access unstable API features and that we need to maintain more API versions, with all the conversion etc.

Yes, and I think / feel this is what we should aim for. We should probably look at other Kubernetes APIs and API providers, but I feel it would be "simpler" from the user standpoint. As an end-user, you have a stable API you consume and you can choose to consume a less stable one. Once we create `v1`, we iterate on a new API version, maintaining `v1` and growing usage and feature of a new version. It *does* mean more code, generation, conversion work, but if that's the cost to provide a non-confusing set of APIs (and feature-flags, …) to the user, I would gladly pay it.

If we see Kubernetes API (and CRDs) as a "domain specific database", CRDs are table/schemas. What we are currently doing with the `enable-api-fields` feature-flag, is to dynamically change the schema "at runtime"/"during execution". It is a bit confusing and it is a bit error prone (what if you revert the flag, …).

I prefer to pay the price in code duplication, generation, sharing, … than making the user pay it. If we need to have another reconciler specifically for a new version (instead of conversion webhook), so be it. If we need to use an internal version that can be convert back-and-forth with all published version, so be it as well.

afrittoli:
> I prefer to pay the price in code duplication, generation, sharing, … than making the user pay it. If we need to have another reconciler specifically for a new version (instead of conversion webhook), so be it. If we need to use an internal version that can be convert back-and-forth with all published version, so be it as well.

I believe we tried that before and did not really find a good solution for these conversions. 
If we can find a good way of implementing this, it sounds like a good way forward, but we should clearly weight out benefits and implementation costs before doing it.

One advantage of the current approach is that it allows platform maintainers to restrict experimental features. If we decided to rely on API versions, the same could probably achieved by blocking specific API versions.

In any case, even if we decide that the original approach was not ideal, I believe we should stick with it until we have an alternate solution we agree on and implemented.

abayer:
@afrittoli I've pushed a new commit that adds the `beta` option and makes `v1` fail validation for remote resolution if `enable-api-fields` isn't either `beta` or `alpha`. Is that what you're looking for?

abayer:
Note that I didn't add anything in docs about resolution, v1, and beta, because we don't actually talk about v1 there yet, so adding a beta features section seems a bit pointless.

lbernick:
should we resurrect https://github.com/tektoncd/pipeline/pull/5325?

abayer:
> should we resurrect #5325?

Sounds reasonable - I forgot that existed and ended up partially recreating it here. =)

pkg/apis/version/version_validation.go
```diff
@@ -30,20 +30,20 @@ import (
...
-	message := `%s requires "enable-api-fields" feature gate to be %q but it is %q`
+	message := `%s requires "enable-api-fields" feature gate to be %s but it is %q`
 	switch wantVersion {
 	case config.StableAPIFields:
 		// If the feature is stable, it doesn't matter what the current version is
 	case config.BetaAPIFields:
 		// If the feature requires "beta" fields to be enabled, the current version may be "beta" or "alpha"
 		if currentVersion != config.BetaAPIFields && currentVersion != config.AlphaAPIFields {
-			message = fmt.Sprintf(message, featureName, fmt.Sprintf("%s or %s", config.AlphaAPIFields, config.BetaAPIFields), currentVersion)
+			message = fmt.Sprintf(message, featureName, fmt.Sprintf("%q or %q", config.AlphaAPIFields, config.BetaAPIFields), currentVersion)
```
abayer, line 40: @lbernick - I tweaked the error message here because it was coming out as `to be "alpha or beta" but it is...` and that obviously wasn't what we wanted. =) I can split this into a separate PR if so desired.

abayer:
I've rebased after #5325 merged and moved out of WIP since this seems to be the consensus approach for beta handling for now.

abayer:
@afrittoli Mind taking a look? =)

afrittoli:
> # Changes
> Closes #4710
> 
> Based on discussions around promoting features last week, I'm opting to skip the need for a `beta` flag and take remote resolution straight to `stable`. I'm confident in the API stability, so I don't think there's a reason to wait since v1 hasn't happened yet. And for that matter, v1 is expecting remote resolution to be on by default anyway.

It might make sense to update the PR description

config/resolvers/config-feature-flags.yaml
```diff
@@ -23,18 +23,10 @@ metadata:
     app.kubernetes.io/part-of: tekton-pipelines
 data:
   # Setting this flag to "true" enables remote resolution of Tekton OCI bundles.
```
afrittoli, line 25: I guess the comment sounds a bit strange now that the default is "true".

config/resolvers/config-feature-flags.yaml
```diff
@@ -23,18 +23,10 @@ metadata:
...
-  # This is an experimental feature and thus should still be considered
-  # an alpha feature.
-  enable-bundles-resolver: "false"
+  enable-bundles-resolver: "true"
```
afrittoli, line 26: Why are we changing this as part of making this feature `v1beta1`? 
I'm not entirely sure why they were disabled by default in alpha, but if we want to change that it feels like it should go in a different PR

config/resolvers/config-feature-flags.yaml
abayer: Good catch!

config/resolvers/config-feature-flags.yaml
abayer: Ok, I can do that in a separate PR, if so desired. I wasn't a big fan of having them off by default in the first place, but the more I thought about it, the more I decided that it was a mistake to have them off. I just thought the `alpha`->`beta` transition was a good time to do that, particularly since we're also including the resolvers deployment in `release.yaml` in this PR. I'd need to make some changes to e2e test setup if I yank this out of the PR - let me know if you think I should.

config/resolvers/config-feature-flags.yaml
abayer: Waiting on acting on this until it's confirmed that the switch to the built-in resolvers being on by default should be moved out of this PR.

chuangw6:
> cc @chuangw6 - I've updated this to incorporate the `ConfigSource` changes from #5551.

@abayer Thank you for letting me know this. Sounds good to me!

config/resolvers/config-feature-flags.yaml
lbernick: I think it's OK to make this swap in the same PR, but I'll defer to @afrittoli here. Just FYI that for v1 we're hoping to have resolution in beta and these resolvers (at least the cluster and bundle ones) enabled by default, as replacements for the OCI bundles syntax and ClusterTask. (ClusterTask might be less necessary since it's a separate CRD, but either way, I think we're waiting on https://github.com/tektoncd/pipeline/pull/5579 until this is merged.)

afrittoli: I understand the reasoning for switching those flags by default, and since this PR is already baked (and rebased enough times) I think it's fine to keep them together - but generally I would prefer if we avoided grouping changes in one PR. I think a more natural flow would have been switching the flag default first and moving to beta then.

config/resolvers/config-feature-flags.yaml
afrittoli: let's keep the one PR

config/resolvers/config-feature-flags.yaml
abayer: 👍 Will do in the future.

afrittoli (commented):
I'm going through the review - so far so good, and sorry about the late review 😓 

I'm slightly concerned about moving to beta a feature that we have not tried in dogfooding.
The standard resolution mechanism is still there, so this is probably ok. 
Before we use this to fully replace  existing features like `ClusterTasks` we'll need to get a bit more experience with the feature though.

Also I think we need slightly better YAML test coverage.

We have no single yaml test in CI which with a `PipelineTask` that uses remote resolution.
We do have tests of pipelines that use remote resolution but they are excluded form CI. At least for the git resolver it should be easy to host a test pipeline in the pipeline repo?

abayer:
> We have no single yaml test in CI which with a `PipelineTask` that uses remote resolution.
> We do have tests of pipelines that use remote resolution but they are excluded form CI. At least for the git resolver it should be easy to host a test pipeline in the pipeline repo?

Whoops! I meant to add a test using `test/git-resolvers/remote-task.yaml` - I'll do that in a separate PR I'll open shortly.

go.sum
```diff
@@ -828,8 +828,6 @@ github.com/imdario/mergo v0.3.10/go.mod h1:jmQim1M+e3UYxmgPu/WyfjB3N3VflVyUjjjwH
 github.com/imdario/mergo v0.3.11/go.mod h1:jmQim1M+e3UYxmgPu/WyfjB3N3VflVyUjjjwH0dnCYA=
 github.com/imdario/mergo v0.3.12 h1:b6R2BslTbIEToALKP7LxUvijTsNI9TAe80pLWN2g/HU=
 github.com/imdario/mergo v0.3.12/go.mod h1:jmQim1M+e3UYxmgPu/WyfjB3N3VflVyUjjjwH0dnCYA=
-github.com/in-toto/in-toto-golang v0.3.4-0.20211211042327-af1f9fb822bf h1:FU8tuL4IWx/Hq55AO4+13AZn3Kd6uk3Z44OCIZ9coTw=
-github.com/in-toto/in-toto-golang v0.3.4-0.20211211042327-af1f9fb822bf/go.mod h1:twl9XmClqj6/h/HANQQYaJZVKPPW/Mz53bd2t6UXGQA=
```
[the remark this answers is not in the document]
abayer: I think that may be rebase cruft, will do a `go mod tidy`.

docs/resolution-getting-started.md
```diff
@@ -24,16 +24,11 @@ with a Pipeline stored in a git repo.
 See [the installation instructions for Tekton Pipeline](./install.md#installing-tekton-pipelines-on-kubernetes), and
 [the installation instructions for the built-in resolvers](./install.md#installing-and-configuring-remote-task-and-pipeline-resolution).
 
-## Step 2: Configure Pipelines to enable alpha features and resolvers
+## Step 2: Configure Pipelines to enable resolvers
 
-Tekton Pipelines currently has its integration with remote resolution behind
-the alpha feature gate, and enabling specific resolvers is controlled by feature 
-flags as well:
+Remote resolvers for Tekton Pipelines are enabled by feature flags:
```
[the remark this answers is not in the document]
abayer: Yeah, I'll reword it to "can be disabled by..."

tekton/publish.yaml
```diff
@@ -189,9 +191,9 @@ spec:
         IMAGES="${IMAGES} ${IMAGES_PATH}/cmd/${cmd}:$(params.versionTag)"
       done
 
-      # Parse the built images from the release.yaml generated by ko
+      # Parse the built images from the minimal-release.yaml generated by ko
       koparse \
-        --path $OUTPUT_RELEASE_DIR/release.yaml \
+        --path $OUTPUT_RELEASE_DIR/minimal-release.yaml \
```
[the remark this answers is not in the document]
abayer: @vdemeester is the one to ask. =)

go.sum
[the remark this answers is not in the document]
abayer: Ah, yeah, `go mod tidy` on `main` removes those - the actual dependency was removed in #5551

tekton/publish.yaml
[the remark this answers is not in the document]
vdemeester: So, today, we are shipping `release.yaml` and `resolvers.yaml`. This would make it so that `release.yaml` has the resolvers by default, but I would like to keep a release.yaml that doesn't have those as well (and thus also keep resolvers.yaml).

Maybe instead of using `ko resolve` multiple time, we could use `yq` to generate new files with the thing we want to filter out. (And this would also be true for the `-t` as we could remove the tag part from it using `yq`).

pkg/remote/resolution/resolver.go
```diff
@@ -34,7 +36,7 @@ type Resolver struct {
 	requester       remoteresource.Requester
 	owner           kmeta.OwnerRefable
 	resolverName    string
-	params          map[string]string
```
afrittoli, line 37: Could you add context in the commit/PR message about this change?
It doesn't seem to be strictly required for the `v1beta1` version of remote resolution.
It would have been nice and easier to review to have that as a separate change before the `v1beta1` 🙏

afrittoli (commented):
Still going through the PR, still looking good, I left some comments though.
I'm a bit confused about the alpha/beta flags in the reconciler and reconciler testing.
Also I would like to see in the PR/commit message/release notes all relevant changes introduced by the PR please 🙏

pkg/apis/resolution/v1beta1/resolution_request_types.go
```diff
@@ -62,6 +62,7 @@ type ResolutionRequestSpec struct {
 	// resource being requested. For example: repo URL, commit SHA,
 	// path to file, the kind of authentication to leverage, etc.
 	// +optional
+	// +listType=atomic
```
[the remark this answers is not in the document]
abayer: Hey, this was me too! See #4402 - dealing with API rule violations from `openapi-gen`. I went with `atomic` because that seemed like the sanest default choice.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -7399,10 +7399,7 @@ spec:
   serviceAccountName: default
 `)
 
-	cms := []*corev1.ConfigMap{withEnabledAlphaAPIFields(newFeatureFlagsConfigMap())}
-
 	d := test.Data{
-		ConfigMaps:   cms,
```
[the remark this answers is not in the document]
abayer: 👍

pkg/reconciler/pipelinerun/pipelinespec/pipelinespec.go
```diff
@@ -49,7 +47,7 @@ func GetPipelineData(ctx context.Context, pipelineRun *v1beta1.PipelineRun, getP
 	case pipelineRun.Spec.PipelineSpec != nil:
 		pipelineMeta = pipelineRun.ObjectMeta
 		pipelineSpec = *pipelineRun.Spec.PipelineSpec
-	case cfg.FeatureFlags.EnableAPIFields == config.AlphaAPIFields && pipelineRun.Spec.PipelineRef != nil && pipelineRun.Spec.PipelineRef.Resolver != "":
+	case pipelineRun.Spec.PipelineRef != nil && pipelineRun.Spec.PipelineRef.Resolver != "":
```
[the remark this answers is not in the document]
abayer: Good call, will do.

pkg/reconciler/pipelinerun/resources/pipelineref.go
```diff
@@ -70,18 +70,14 @@ func GetPipelineFunc(ctx context.Context, k8s kubernetes.Interface, tekton clien
 			resolver := oci.NewResolver(pr.Bundle, kc)
 			return resolvePipeline(ctx, resolver, name)
 		}, nil
-	case cfg.FeatureFlags.EnableAPIFields == config.AlphaAPIFields && pr != nil && pr.Resolver != "" && requester != nil:
```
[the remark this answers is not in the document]
abayer: I'll put in a comment...or maybe I can check the version of the `PipelineRun`?

pkg/resolution/resource/name.go
```diff
@@ -33,22 +35,33 @@ func nameHasher() hash.Hash {
...
 		}
-		if _, err := hasher.Write([]byte(params[key])); err != nil {
-			return "", err
+		switch p.Value.Type {
+		case v1beta1.ParamTypeString:
+			if _, err := hasher.Write([]byte(p.Value.StringVal)); err != nil {
+				return "", err
+			}
+		case v1beta1.ParamTypeArray, v1beta1.ParamTypeObject:
+			asJSON, err := p.Value.MarshalJSON()
```
afrittoli, line 58: Nice :)

test/e2e-common.sh
```diff
@@ -21,18 +21,14 @@ source $(git rev-parse --show-toplevel)/vendor/github.com/tektoncd/plumbing/scri
...
       | sed -e 's%loglevel.controller: "info"%loglevel.controller: "debug"%' \
       | sed -e 's%loglevel.webhook: "info"%loglevel.webhook: "debug"%' \
       | kubectl apply -R -f - || fail_test "Build pipeline installation failed"
 
   verify_pipeline_installation
-
-  if [ "${PIPELINE_FEATURE_GATE}" == "alpha" ]; then
-    ko apply -f config/resolvers || fail_test "Resolvers installation failed"
-    verify_resolvers_installation
-  fi
```
afrittoli, line 35: Happy to see this one go

test/e2e-common.sh
```diff
@@ -42,7 +38,8 @@ function install_pipeline_crd_version() {
   echo ">> Deploying Tekton Pipelines of Version $1"
   kubectl apply -f "https://github.com/tektoncd/pipeline/releases/download/$1/release.yaml" || fail_test "Build pipeline installation failed of Version $1"
 
-  if [ "${PIPELINE_FEATURE_GATE}" == "alpha" ]; then
+  # If the version to be installed is v0.40.x, we need to install resolvers.yaml separately.
+  if [[ "${1}" == v0.40.* ]]; then
```
afrittoli, line 42: I had forgotten about this function... I guess it's used for update testing, which is not really working but good to keep it up to date as much as possible

pkg/reconciler/pipelinerun/resources/pipelineref.go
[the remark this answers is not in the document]
afrittoli: Good point, we usually don't use flag checks in the reconciler for API-driven features, only for back-end ones.

pkg/reconciler/pipelinerun/pipelinerun_test.go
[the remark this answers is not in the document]
abayer: Actually, let's open a separate issue for this, since we're not testing `v1` in the reconciler tests anywhere at the moment - that seems like a separate requirement for `v1` rather than specific to remote resolution beta, and will involve a lot of tests.

pkg/reconciler/pipelinerun/pipelinespec/pipelinespec.go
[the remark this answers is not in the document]
abayer: For the record, we're not gonna add a comment - see https://github.com/tektoncd/pipeline/pull/5515#discussion_r987871976

pkg/reconciler/taskrun/resources/taskref.go
```diff
@@ -100,11 +100,10 @@ func GetTaskFunc(ctx context.Context, k8s kubernetes.Interface, tekton clientset
 
 			return resolveTask(ctx, resolver, name, kind)
 		}, nil
-	case cfg.FeatureFlags.EnableAPIFields == config.AlphaAPIFields && tr != nil && tr.Resolver != "" && requester != nil:
```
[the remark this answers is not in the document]
abayer: (for the record - see discussion at https://github.com/tektoncd/pipeline/pull/5515#discussion_r987871976)

pkg/reconciler/pipelinerun/pipelinerun_test.go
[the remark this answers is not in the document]
afrittoli: ok, sounds good

tekton/publish.yaml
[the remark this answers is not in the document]
afrittoli: My question was not about `koparse`, not `ko resolve` actually.
The `ko resolve` calls should be fine, since the built images will be in the cache so it won't actually build the same image more than once. 

The `koparse` is not a big deal, but I would prefer a simpler script / logic where possible.
The split was done because of v0.40, now we could revert that and go back to one `koparse` invocation and fewer params in the pipeline/task.

tekton/publish.yaml
```diff
@@ -189,9 +191,9 @@ spec:
...
         --base ${IMAGES_PATH} --images ${IMAGES} > /workspace/built_images
```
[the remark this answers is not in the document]
abayer: We are - see below this.

config/300-resolutionrequest.yaml
afrittoli: They are, I tested it earlier, and they're deleted together with the `TaskRun` / `PipelineRun`.

afrittoli: @abayer Since this is likely to be in v0.41 which is still 10+ days away, I think it's still worth sending a notice on the mailing list, and we should have this mentioned in the release notes as well, so that if anyone implemented they're own resolver they'll know they must updated it before switching to v0.41.

abayer: 👍 I'll send that as soon this merges.

lbernick:
> > We have no single yaml test in CI which with a `PipelineTask` that uses remote resolution.
> > We do have tests of pipelines that use remote resolution but they are excluded form CI. At least for the git resolver it should be easy to host a test pipeline in the pipeline repo?
> 
> Whoops! I meant to add a test using `test/git-resolvers/remote-task.yaml` - I'll do that in a separate PR I'll open shortly.

There are a few examples tests that use a git clone copied directly from the catalog, e.g. https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/pipelinerun-with-final-tasks.yaml. (It has the comment "This can be deleted after we add support to refer to the remote Task in a registry (Issue #1839) or add support for referencing task in git directly (issue #2298)". Maybe those can be replaced with git resolution?

abayer:
@lbernick Yup, that's a followup. Right now I just have to figure out why #5604 is failing in CI but working for me locally. =)

abayer:
@afrittoli Ok, so we _do_ test a remote-resolved `PipelineTask` in https://github.com/tektoncd/pipeline/blob/main/test/resolvers_test.go#L202-L260 - I'm not sure why #5604 (adding an example with `PipelineTask` remote resolution) is failing, since it's passing for me locally. But I'll keep digging into it. =)

tekton/publish.yaml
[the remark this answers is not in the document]
vdemeester: yes :)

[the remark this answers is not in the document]
abayer: I've opened #5607 for that discussion.

afrittoli:
Thanks @abayer for working on this and for the updates.
Should I be worried about https://github.com/tektoncd/pipeline/pull/5604 failing CI?

/hold cancel

abayer:
@afrittoli Nah, pretty sure that's just due to GitHub webhooks having problems.

/retest

abayer:
Rather, GH webhook issues are why the retest isn't firing - the unit test failure is just one of the family of event-related unit test flakes.

PR #5523, opened 2022-09-19
TEP-0114: Tests Wait Custom Task Retries and TimeOut

XinruZhang:
# Changes
Prior to this pr, we are not able to test cases where Custom Task TimesOut or retries is specified because we don't have a controller ready. Now that we have Wait Task Custom Controller, we are able to write e2e tests to ensure the behavior of timeout and retries.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```

### DIFF
--- a/test/custom_task_test.go
+++ b/test/custom_task_test.go
@@ -40,8 +40,11 @@ import (
 	"go.opencensus.io/trace"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
+	v1 "knative.dev/pkg/apis/duck/v1"
+	duckv1beta1 "knative.dev/pkg/apis/duck/v1beta1"
 	knativetest "knative.dev/pkg/test"
 	"knative.dev/pkg/test/helpers"
 )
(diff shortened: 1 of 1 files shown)


test/custom_task_test.go
```diff
@@ -552,6 +592,8 @@ func TestWaitCustomTask_PipelineRun(t *testing.T) {
 	for _, tc := range []struct {
 		name                      string
 		runDuration               string
+		runTimeout                string
+		runRetries                int
 		prTimeout                 string
 		prConditionAccessorFn     func(string) ConditionAccessorFn
 		wantPrConditionType       apis.ConditionType
```
dibyom, line 599: Instead of a bunch of different expected outputs, would it be possible to define a wantStatus that contains the expected custom run status?

test/custom_task_test.go
```diff
@@ -661,6 +741,9 @@ spec:
           type: %s
         observedGeneration: %v
 `, runName, tc.wantRunConditionReason, tc.wantRunConditionStatus, tc.wantRunConditionType, tc.wantRunObservedGeneration)
+					if tc.wantRetriesStatus != "" {
```
dibyom, line 744: This is more of a comment on the previous lines - but getting the status and then appending to a string seems unmaintainable (if we change the status, someone would have to know to update these strings) - is there a reason we cannot just use the go structs here?

test/custom_task_test.go
XinruZhang: The `runStatus` differs upon different feature flags (FullEmbeddedStatus, BothEmbeddedStatus, MinimalEmbeddedStatus), therefore we need to compose it in runtime accordinlgy. wdyt?

test/custom_task_test.go
XinruZhang: Oh I'm actually not getting the status but composing the expected status, added some comments, hope this can clear some of the confusion here. 

> if we change the status, someone would have to know to update these strings

I may need more clarification here. Do you mean we change the structure of the status or the value of the fields in status? 

About why do we prefer using YAML, as far as I remember, using YAMLs can give our users examples for how to use Custom Runs. @jerop may have other answers that I don't remember clearly.

test/custom_task_test.go
dibyom: ok, in that case - shouldn't the feature flag value be one of the test inputs?

test/custom_task_test.go
dibyom: > I may need more clarification here. Do you mean we change the structure of the status or the value of the fields in status?

Structure of the status field

> About why do we prefer using YAML, as far as I remember, using YAMLs can better instruct our users on how to use Custom Runs. @jerop may have other answers that I don't remember clearly 😅.

Sure - I get that the inputs being yaml make it easier to read the pipeline/run under test , but composing large strings  like the pipelineRun below with 14 args that all have to be specified in exactly the right order makes the test harder to read and maintain

XinruZhang: Updated as we discussed offline. The basic principle is to make the code more readable, so that it would be easier to debug future test failures.

test/custom_task_test.go
XinruZhang: Updated as suggested, use `wantRetriesStatus` to group related fields together.

dibyom (approved):

jerop (approved):
/lgtm

Happy to see these e2e tests for custom tasks, thank you @XinruZhang!

Issue #4112, opened 2021-07-22
PodSecurityPolicy deprecated in Kubernetes 1.21


pietervincken:
# Expected Behavior
The released deployment manifest doesn't have deprecated resources listed in it.

# Actual Behavior
PodSecurityPolicy tekton-pipelines is listed in the deployment manifest which is a deprecated resource as of 1.21.

# Steps to Reproduce the Problem

1. Download the latest release yaml (tested with 0.26.0)

# Additional Info

- Kubernetes version: 1.21

  **Output of `kubectl version`:**
N/A

- Tekton Pipeline version:

0.26.0

vdemeester:
@pietervincken thanks for the issue. Indeed, I think we should *just* remove this from the `config` folder (and thus from the released yamls). 

/cc @tektoncd/core-maintainers

dibyom:
Some background reading: https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/

afrittoli:
@vdemeester I will move this to the next milestone. Let me know if there is any concern.

sbwsg:
Looks like the replacement feature in Kubernetes will be called ["PodSecurity Admission"](https://github.com/kubernetes/enhancements/issues/2579). Released as alpha in 1.22

afrittoli:
Discussed during the Pipeline WG:
- we shall had support for the new mechanism but keep PSP in parallel
- both methods can stay together until Pod Security Admission is stable

vdemeester:
If `PodSecurityPolicy` are removed in 1.25, then we need to either remove it for 0.36 or 0.37 (any release that would be in the 1.25 timeframe more or less) or be able to generate release with and without it.
I can re-open https://github.com/tektoncd/pipeline/pull/4122 if need be.

dibyom:
I think we need to do this soon - we could point users to https://appvia.github.io/psp-migration/ to migrate to an alternate policy engine. We could also see if we can enable https://kubernetes.io/docs/concepts/security/pod-security-admission/

abayer:
Resurrecting this again - looks like `PodSecurityPolicy` has been removed in k8s 1.25, so we need to push this. I'm putting this in the v0.39 milestone, but am going to assume it'll end up slipping to v0.40 just due to time.

abayer:
/priority critical-urgent

Actually, I just put it in v0.40, but am also bumping the priority.

pritidesai:
[...]

PR #5528, opened 2022-09-20
[Deprecated] PodSecurityPolicy

JeromeJu:
# Changes

This commit removes the deprecated PodSeucrityPolicy and provide a guide moving forward for users.

Fixes: https://github.com/tektoncd/pipeline/issues/4112
/kind bug

TODO:
Apply PSA & OPA that could cover what previously PSP has achieved.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
PodSecurityPolicy is deprecated.
```


dibyom:
> How are we going to test the PSA implementation against PSP?

I'm not sure if I fully understand the question but from reading your doc, it looks like we'd have to use PSA level "privileged" . You also provide the [3 policies](https://github.com/tektoncd/pipeline/pull/5528/files#diff-b0def43104df1bd58421e6db6aeec42dae7eb1d24a1e089d1604c18e2dafb266R57-R65) that one would have to implement using an external policy engine like OPA. I think that's sufficient

JeromeJu:
> > How are we going to test the PSA implementation against PSP?
> 
> I'm not sure if I fully understand the question but from reading your doc, it looks like we'd have to use PSA level "privileged" . You also provide the [3 policies](https://github.com/tektoncd/pipeline/pull/5528/files#diff-b0def43104df1bd58421e6db6aeec42dae7eb1d24a1e089d1604c18e2dafb266R57-R65) that one would have to implement using an external policy engine like OPA. I think that's sufficient

Yes I would like to ask as pointed in the WG, do I need to provide the implementation of a PSA controller as an example, which should be an equivalence of 101-podsecuritypolicy.yaml as we are currently removing from pipeline in this PR?
If that is the case, do we need to have test cases that compares each specification between those two?

JeromeJu:
Goal: When we remove psp, the security specs should be the same for webhooks and controllers service account.

Issue #5524, opened 2022-09-20
Projected volumes to beta


vdemeester:
### Feature request
For the latest release, Projected Volumes is in Alpha in v0.38. It shall be promoted to beta.

PR #5530, opened 2022-09-21
Remove alpha feature gate from projected workspaces

0xFelix:
# Changes

This removes the alpha feature gate that is currently guarding projected workspaces so they become a beta/stable feature.

Closes https://github.com/tektoncd/pipeline/issues/5524

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Projected workspaces are promoted to beta/stable API
```


afrittoli:
> # Changes
> This removes the alpha feature gate that is currently guarding projected workspaces so they become a beta/stable feature.

What does beta/stable means?  `beta` is *not* `stable`, we don't have a `stable` release yet.

This change means that the feature will implicitly be marked as stable once `v1` is released.
Our current feature gate design requires features to go through alpha, then beta and then stable (once `v1` is released).

@tektoncd/core-maintainers

0xFelix:
It means the feature is now useable in `beta` and `stable` without a feature gate.

I did not add a `beta` feature gate because the `stable` / `v1` API is not release yet and the [docs](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#new-api-fields) say a `beta` feature gate is not needed until then.

No concerns were raised before merging it, so I thought this is following our processes.

vdemeester:
@afrittoli if `v1` wasn't created yet, we would have no `beta` value for the `enable-api-fields` for example. So we would straight up promoting to *beta*. One or two release appear and we *create* / *promote* / *publish* `v1`, in that case, did we "promote" to both `beta` and `stable` at the same time ? When we would create that `v1` (stable), would we decide to not include that feature but it was recently promoted to `beta` which was the latest thing available at that time ?

*Stable* doesn't exist, it is in the making (as we are starting to put code for `v1`, …) but it doesn't exist. This is also why, as commented on https://github.com/tektoncd/pipeline/pull/5515, I feel the flag (and its value) is, looking back, confusing to say the least.

0xFelix:
If the goal is to stop development on the `v1beta` version of the API once `v1` is out, why not get rid of the `alpha` / `beta` terminology and use something like feature gates for specific features instead? I think this is way more intuitive, as for example with the way it currently is implemented I can only enable none or all alpha/beta features, even if I only want to use one specific feature that is guarded.

We do that in kubevirt for example:

https://github.com/kubevirt/kubevirt/blob/main/docs/deprecation.md#feature-gates
https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/feature-gates.go

afrittoli:
The reason we opted for a global flag, as opposed to feature-specific ones, is that we wanted to avoid exploding the test matrix. We do have feature flags, we but only use them for features that are not API driven, since those may have an impact across the API surface.

I don't think there is any concern with regards to the feature flags, the options for API fields are:
- keep using the existing flag
- tie new features to API versions 

If we decide to go for the latter, we will need to find a way to manage the conversions.

Issue #5487, opened 2022-09-13
Add support for using multiple remote resolution Git providers in the same cluster


abayer:
With #5450 about to merge, we're adding support for remote resolution of `Task`s and `Pipeline`s using Git providers' APIs. Which is great! But it does come with a sizable limitation - you can only have one provider type/server/token per cluster. While this is fine to start with, we really should add the ability for admins to configure multiple supported providers, so that you can use both an internal GitHub Enterprise and github.com, or any other combination of the various providers supported through go-scm.

abayer:
I'm kicking around a couple ideas at the moment. Both revolve around having the provider configured in the `git-resolver-config` `ConfigMap` as the "default", with an optional parameter for the `git` resolver available to specify the "name" of a different provider configuration. The question, in my mind, is how to store those additional provider configurations. My current ideas:

* `ConfigMap`s in the `tekton-pipelines-resolvers` namespace, containing the relevant fields from `git-resolver-config` and the unique "name" as an identifier for the provider config. For discovery of these `ConfigMap`s, I'm thinking requiring a label, like `resolution.tekton.dev/git-provider-config: "true"`. Then the resolver will use watches to load/update those configurations into a map keyed on the "name".
* A new CRD, something like `ResolutionGitProvider`, with instances of the CRD also living in the `tekton-pipelines-resolvers` namespace. This would also have the config fields from `git-resolver-config`, albeit as actual fields rather than map keys, with the `name` of the instance being used as the identifier. Again, they'd be loaded up via watches.

The `ConfigMap` approach is probably the right one, but I just really like the CRD approach - less things to document (key names, the required label, etc), and it just feels nice? I'm giving myself a week or two to come back to this and see how I feel about the options then.

chuangw6:
> ConfigMaps in the tekton-pipelines-resolvers namespace, containing the relevant fields from git-resolver-config and the unique "name" as an identifier for the provider config.

Agreed that this is probably the good path to explore. Depending on what is best for the users, we might want to consider how to reuse `api-token-secret-name` that could store the tokens for different providers in a single secret under different `api-token-secret-key`. wdyt?

abayer:
Ok, I ran into some walls with the config map approach - namely, how to know when a new provider's config map has been added so we can add it to the config map watch. I've prototyped a new approach:

[...]

PR #5531, opened 2022-09-21
Add the ability to mock more SCM data for git resolver tests

abayer:
# Changes

Specifically, it adds the ability to cheat a "fake" list of repositories into the SCM client for use in adding the clone URL to the resolved resource annotations, as needed in #5397.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/resolution/resolver/git/resolver.go
+++ b/pkg/resolution/resolver/git/resolver.go
@@ -31,6 +31,7 @@ import (
 	gitcfg "github.com/go-git/go-git/v5/config"
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/go-git/go-git/v5/storage/memory"
+	"github.com/jenkins-x/go-scm/scm"
 	"github.com/jenkins-x/go-scm/scm/factory"
 	resolverconfig "github.com/tektoncd/pipeline/pkg/apis/config/resolver"
 	resolutioncommon "github.com/tektoncd/pipeline/pkg/resolution/common"
@@ -75,6 +76,9 @@ type Resolver struct {
 	logger     *zap.SugaredLogger
 	cache      *cache.LRUExpireCache
 	ttl        time.Duration
+
+	// Used in testing
+	clientFunc func(string, string, string, ...factory.ClientOptionFunc) (*scm.Client, error)
 }
 
 // Initialize performs any setup required by the gitresolver.
@@ -83,6 +87,9 @@ func (r *Resolver) Initialize(ctx context.Context) error {
 	r.logger = logging.FromContext(ctx)
 	r.cache = cache.NewLRUExpireCache(cacheSize)
 	r.ttl = ttl
+	if r.clientFunc == nil {
+		r.clientFunc = factory.NewClient
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


vdemeester (approved):

pkg/resolution/resolver/git/resolver.go
```diff
@@ -147,19 +152,32 @@ func (r *Resolver) resolveAPIGit(ctx context.Context, params map[string]string)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create SCM client: %w", err)
 	}
+
+	// If we have fake data for the scm client, replace the scm client with a fake client with the fake data
+	if r.fakeData != nil && scmClient.Driver == scm.DriverFake {
```
lbernick, line 157: would it make more sense to pass a client in as an argument to a helper function, or store the client on the resolver or something like that? (i.e. more like the dependency injection we use elsewhere in the codebase). It feels strange to mix test functionality and real functionality in the resolver code.

abayer: I made the deliberate choice to create the client at resolution time every time. That way we don't have to specifically watch for changes to the token secret and then create a new client, and the work I'm doing for #5487 will mean we support multiple providers, which means different clients will be needed at runtime for each client, etc. 

I guess we could have a field on the resolver that's a function for creating the client, with the default implementation being what we do now, and tests can override that?

lbernick: maybe the resolution function can accept a client creation function as an argument? kind of like [getpipelinefunc](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L355)

abayer: It would need to be on the resolver, not the function, since the actual path to calling `Resolve(...)` in the tests is off in `frtesting.RunResolverReconcileTest`. Lemme try something.

abayer:
/hold

@lbernick How's https://github.com/tektoncd/pipeline/pull/5531/commits/69171b7362b96d25786172fe01b5c0b4981fe3d0?

lbernick:
I like that approach!

abayer:
woo! Squashing it down now.

pkg/resolution/resolver/git/resolver.go
```diff
@@ -83,6 +87,9 @@ func (r *Resolver) Initialize(ctx context.Context) error {
 	r.logger = logging.FromContext(ctx)
 	r.cache = cache.NewLRUExpireCache(cacheSize)
 	r.ttl = ttl
+	if r.clientFunc == nil {
+		r.clientFunc = factory.NewClient
+	}
```
lbernick, line 92: nit: is this conditional needed, or is this function called only from real code?

abayer: Correct, that's to make sure we actually do the normal client creation outside of tests.

lbernick: sorry I mean can you just do `r.clientFunc = factory.NewClient` (without the if)

abayer: No - I tried that - I forgot that `Initialize` is called before each `Resolve` call, so it gets overwritten etc.

PR #5536, opened 2022-09-21
[Deprecated] PodSecurityPolicy

JeromeJu:
This commit removes the deprecated PodSeucrityPolicy and provide a guide moving forward for users.

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

``` release-note
NONE
```


JeromeJu:
```seLinuxOptions:
            role: "RunAsAny"
```
works for the seLinux configurations in controller.yaml

config/controller.yaml
```diff
@@ -126,6 +126,9 @@ spec:
           # User 65532 is the distroless nonroot user ID
           runAsUser: 65532
           runAsGroup: 65532
+          fsGroup: 65535
+          seLinuxOptions:
+            role: "RunAsAny"
```
lbernick, line 131: Does the controller truly need `RunAsAny`? If not, we would be able to use a restricted PodSecurityAdmission which would be preferable

config/101-podsecuritypolicy.yaml
```diff
@@ -1,57 +0,0 @@
...
-  supplementalGroups:
-    rule: 'MustRunAs'
-    ranges:
-    - min: 1
-      max: 65535
-  fsGroup:
-    rule: 'MustRunAs'
-    ranges:
-    - min: 1
-      max: 65535
```
lbernick, line 57: these rules will need to be applied to the webhook deployment too.

config/controller.yaml
JeromeJu: I tested with RunAsUser and it shall work.

lbernick: Great! In that case, can you add a "restricted" pod security admission label to the tekton-pipelines namespace? We also can remove the docs, since there won't be any additional action needed for cluster operators.

config/101-podsecuritypolicy.yaml
JeromeJu: The[1, 65535] range for the user ID here restricts the use of root user, which is the default value of 0 if unspecified. Used `runAsNonRoot` instead.

config/101-podsecuritypolicy.yaml
```diff
@@ -1,57 +0,0 @@
...
-metadata:
-  name: tekton-pipelines
-  labels:
-    app.kubernetes.io/instance: default
-    app.kubernetes.io/part-of: tekton-pipelines
-  annotations:
-    seccomp.security.alpha.kubernetes.io/allowedProfileNames: 'docker/default,runtime/default'
-    seccomp.security.alpha.kubernetes.io/defaultProfileName: 'runtime/default'
-    apparmor.security.beta.kubernetes.io/allowedProfileNames: 'runtime/default'
-    apparmor.security.beta.kubernetes.io/defaultProfileName: 'runtime/default'
```
lbernick, line 26: Do these annotations need to be replaced with any changes to the deployment spec?

config/webhook.yaml
```diff
@@ -140,6 +140,10 @@ spec:
           initialDelaySeconds: 5
           periodSeconds: 10
           timeoutSeconds: 5
+      securityContext:
+        seLinuxOptions:
```
lbernick, line 144: I'm actually wondering if we need seLinuxOptions here since it's basically redundant with the container security context?

lbernick (commented):
This looks basically right to me but I think @wlynch or @skaegi might be better reviewers-- could either of you take a look?

@JeromeJu it might be helpful to make a list (either in your doc or the commit message) of the pod security policy configuration options we're using and what replaces them. (For most, it'll probably be replaced by the pod security admission label you've added.)

Lastly, the release note should indicate changes relevant to users, and I don't think replacing our podsecuritypolicy should be relevant to them since you added the pod security admission. However, I think users might need to set a feature gate to enable pod security admission for k8s 1.22 to 1.24 (https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-graduated-or-deprecated-features), so please include "action required" in the release note with instructions on how to do that.

I'd also change the commit title to "Replace deprecated PodSecurityPolicy with Pod Security Admission enforcement".

config/controller.yaml
JeromeJu: I was wondering if we should remove the docs. Thinking about the use where there are some customized enforcement by PSP but not covered by our tekton/pipeline version of 101-podsecuritypolicy.yaml. Shall we provide any instructions?

lbernick: Are there any aspects of our old PSP that are not covered by the configuration you've added? I don't think there are, but having a table as I suggested would be really helpful to verify that that's the case.

config/webhook.yaml
JeromeJu: Looking at the `seLinux` setup from https://kubernetes.io/docs/reference/access-authn-authz/psp-to-pod-security-standards/ it seems that with the restricted label, it specified the podTemplate to be `seLinux: mustRunAs` with `user` as undefined. I was wondering if this essentially mean it could be 0 as root?

config/101-podsecuritypolicy.yaml
wlynch: I think I'm okay with this - I don't think we were ever requiring a strict set of apparmor / seccomp policies - we just wanted to make sure that there was some policy set. The restricted policy still enforces this, though loosens it a bit in that it's easier to load in other policies present on the node.

We may want to double check that this is working on a few common k8s setups (since if a policy isn't defined this will fail), but otherwise this seems okay.

config/101-podsecuritypolicy.yaml
```diff
@@ -1,57 +0,0 @@
...
-spec:
-  privileged: false
-  allowPrivilegeEscalation: false
-  requiredDropCapabilities:
-    - ALL
-  volumes:
-  - 'emptyDir'
-  - 'configMap'
-  - 'secret'
```
wlynch, line 35: The values covered in PSS is technically larger than we we previously set, but I think I'm okay with the increase in scope. This is the list PSS looks for:

- configMap
- csi
- downwardAPI
- emptyDir
- ephemeral
- persistentVolumeClaim
- projected
- secret

config/101-podsecuritypolicy.yaml
```diff
@@ -1,57 +0,0 @@
...
-  runAsUser:
-    rule: 'MustRunAsNonRoot'
-  runAsGroup:
-    rule: 'MustRunAs'
-    ranges:
-    - min: 1
-      max: 65535
-  seLinux:
-    rule: 'RunAsAny'
-  supplementalGroups:
```
wlynch, line 48: FWIW, it doesn't look like this is included in PSS so it's something we'll need to be mindful about (I'm not sure if you can add root as a supplemental group, though I hope not).

As is I think this is okay, since we're not setting any supplemental groups and we're dropping capabilities like SETUID/SETGID that could be possible escalation points.

wlynch (approved):
/lgtm

Some minor comments, but nothing blocking.

config/webhook.yaml
wlynch: Yeah typically root uid/gid is 0.

If you want a simple test to add to the e2e test try to add a simple a pod that runs as root to ensure that PSS is correctly blocking the pod.

config/101-podsecuritypolicy.yaml
JeromeJu: I thought the point of `supplementalGroups` in PSP is to not allow root access. So will `runAsNonRoot` to be set as true here cover this?

wlynch: Reading https://pkg.go.dev/k8s.io/api/core/v1#SecurityContext + the PSS docs it looks like runAsNonRoot doesn't restrict GID, just UID. That said, we're not granting root GID in any of our yamls, so it's probably fine (just something to keep in mind since PSS won't stop us from setting this).

Issue #4955, opened 2022-06-09
TEP-0107: Propagated `Parameters` Migration


jerop:
[TEP-0107: Propagating Parameters][tep-0107] proposed propagating `Parameters` in embedded specifications without mutations. This feature was released in [v0.36.0][36] gated behind the `alpha` feature flag. 

This issue tracks the removal of that feature flag and promotion to Beta - targeting release [v0.40.0][40].

- [x] https://github.com/tektoncd/pipeline/issues/5141
- [x] https://github.com/tektoncd/pipeline/issues/5140
- [x] https://github.com/tektoncd/pipeline/issues/5588
- [x] https://github.com/tektoncd/pipeline/pull/5540

[tep-0107]: https://github.com/tektoncd/community/blob/main/teps/0107-propagating-parameters.md
[36]: https://github.com/tektoncd/pipeline/releases/tag/v0.36.0
[40]: https://github.com/tektoncd/pipeline/releases/tag/v0.40.0

jerop:
Completed in https://github.com/tektoncd/pipeline/pull/5540 - Thank you @chitrangpatel!

PR #5540, opened 2022-09-22
Propagated parameters to beta

chitrangpatel:
# Changes
[Propagated Parameters](#4955) was deployed as an alpha feature in release [v0.36.0](https://github.com/tektoncd/pipeline/releases/tag/v0.36.0). This PR promotes it to beta by removing the constraints applied under the `alpha` feature. This PR addresses issue https://github.com/tektoncd/pipeline/issues/4955.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Promote propagated parameters to beta.
```


abayer:
/hold

Holding this pending beta/stable/etc discussions in other places, such as #5530 and #5515. Probably will get covered at the API WG on Monday.

pkg/apis/pipeline/v1/task_validation_test.go
```diff
@@ -142,7 +142,7 @@ func TestTaskSpecValidatePropagatedParamsAndWorkspaces(t *testing.T) {
 				Workspaces:   tt.fields.Workspaces,
 				Results:      tt.fields.Results,
 			}
-			ctx := config.EnableAlphaAPIFields(context.Background())
+			ctx := context.Background()
```
jerop, line 145: we need to check if beta api fields is enabled in v1

jerop (commented):
thank you @chitrangpatel!

we need to guard the propagation behind beta feature flag in v1 - https://github.com/tektoncd/pipeline/pull/5325

jerop:
/hold cancel

we agreed to support alpha/beta/stable to unblock progress - https://github.com/tektoncd/pipeline/pull/5325

pkg/apis/pipeline/v1/task_validation_test.go
chitrangpatel: sounds good!

pkg/apis/pipeline/v1beta1/taskrun_validation_test.go
```diff
@@ -63,7 +63,7 @@ func TestTaskRun_Invalidate(t *testing.T) {
 			Message: `non-existent variable in "$(params.task-words[*])"`,
 			Paths:   []string{"spec.steps[0].args[0]"},
 		},
-		wc: config.EnableAlphaAPIFields,
+		wc: config.EnableBetaAPIFields,
```
jerop, line 66: we do not need this check in v1beta1 - see https://github.com/tektoncd/pipeline/pull/5628#discussion_r993724518

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -126,18 +123,6 @@ func missingParamsNames(neededParams []string, providedParams []string, paramSpe
...
-	// Extra params are needed to support
-	// https://github.com/tektoncd/community/blob/main/teps/0023-implicit-mapping.md
-	// So that parent params can be passed down to subresources (even if they
-	// are not explicitly used).
-	if config.FromContextOrDefaults(ctx).FeatureFlags.EnableAPIFields != "alpha" {
-		return list.DiffLeft(providedParams, neededParams)
-	}
-	return nil
-}
-
```
jerop, line 140: this is not relevant to the promotion, right? i recommend that this goes in a separate commit & PR

jerop (changes requested):
believe the "extra params" change should be in a separate commit & PR

pkg/reconciler/taskrun/validate_resources.go
chitrangpatel: Leaving this in makes the tests fail. It seems like the default action was to check if there are extra parameters and throw an error. When the alpha feature flag was enabled, this would not be the case. Now that propagated params is being promoted to beta, we cannot leave this filter in just for "alpha", we need to do the same for "beta" as well. That's why I thought it belonged in the same PR.

chitrangpatel: I'm happy to open a separate PR just for this and fix it there. However, it will have to be merged in for propagated params to beta to work.

docs/pipelineruns.md
```diff
@@ -326,8 +326,6 @@ go through the complexity of checking each `Pipeline` and providing only the req
 
 #### Propagated Parameters
 
-**([alpha only](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features))**
```
JeromeJu, line 329: Sorry for running late on this review.
Wondering for this do we also need to add it to the [list of beta features](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#beta-features)?

jerop: didn't see a list there and thought we'd create the list when we make the v1 release

PR #5552, opened 2022-09-23
[TEP-0091] Trusted resources alpha add sigstore packages

Yongxuanzhang:
# Changes
This commit is part of the work of trusted resources, it introduces the
dependency of [sigstore/sigstore](https://github.com/sigstore/sigstore) by adding signing and verifying
functions to the code base. Those functions are not invoked in this
commit but will be used in the following work.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


Yongxuanzhang:
What I'm worried about is the huge added lines of code into vendor. 😟 
Any suggestions or help would be appreciated! 🙏

pkg/client/resource/clientset/versioned/fake/register.go
```diff
@@ -37,14 +37,14 @@ var localSchemeBuilder = runtime.SchemeBuilder{
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
 // of clientsets, like in:
 //
-//   import (
-//     "k8s.io/client-go/kubernetes"
-//     clientsetscheme "k8s.io/client-go/kubernetes/scheme"
-//     aggregatorclientsetscheme "k8s.io/kube-aggregator/pkg/client/clientset_generated/clientset/scheme"
-//   )
+//	import (
```
JeromeJu, line 40: Looks like these are unnecessary tabs introduced.

Yongxuanzhang: Thanks!!

pkg/reconciler/trustedresources/sign.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/sigstore/cosign/pkg/cosign"
+	"github.com/sigstore/sigstore/pkg/signature"
+)
+
+// signInterface returns the encoded signature for the given object.
+func signInterface(signer signature.Signer, i interface{}) (string, error) {
```
Yongxuanzhang, line 33: This function is not used in other packages at this commit, I guess we should make them unexported?

pkg/reconciler/trustedresources/sign.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/sigstore/cosign/pkg/cosign"
```
wlynch, line 28: I'd drop the signing from this PR, since the pipeline controller will primarily be interested in verification only. (signing will likely be more relevant for CLI / CLI plugin).

In generaly, I'd avoid taking on the cosign dependency unless it's absolutely necessary - as you can see from the vendor directory, there's a ton of dependencies.

wlynch (changes requested):

pkg/reconciler/trustedresources/sign.go
wlynch: I'm remembering now - you're using this for testing.

It looks like you're using cosign mostly for the key generation - you can drop this in favor of generating + loading the key directly via one of the load functions in https://pkg.go.dev/github.com/sigstore/sigstore/pkg/signature

e.g. you could replace most of this with something like:

```go
key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
if err ...
sv, err := signature.LoadECDSASignerVerifier(key, crypto.SHA256)
if err ...
return sv
```

pkg/reconciler/trustedresources/verify_test.go
```diff
@@ -0,0 +1,143 @@
...
+	taskSpecTest = &v1beta1.TaskSpec{
+		Steps: []v1beta1.Step{{
+			Image: "ubuntu",
+			Name:  "echo",
+		}},
+	}
+)
+
+func init() {
+	os.Setenv("SYSTEM_NAMESPACE", namespace)
```
wlynch, line 49: I think the preferred way of doing this is is to import `knative.dev/pkg/testing`.

I'm also curious why you need this if we're not verifying any controller functionality yet 🤔

pkg/reconciler/trustedresources/sign.go
wlynch: if/when you need the multi-algo sv creator, use https://pkg.go.dev/github.com/sigstore/sigstore/pkg/signature#LoadSignerVerifier

Yongxuanzhang: Sure! Thanks! I will try it out!

pkg/reconciler/trustedresources/verify_test.go
Yongxuanzhang: This should be the piece of code I forgot to remove 🥲

wlynch (approved):
/lgtm

PR #5565, opened 2022-09-27
Properly cleanup injected sidecars

### LINKED ISSUE #4731: istio-proxy container not stopped after running as taskrun
Right now I have tekton-pipelines installed through the cncf tekton-pipelines chart here https://github.com/cdfoundation/tekton-helm-chart using v0.23.0 and triggers v0.13.0. Currently when my task runs spin up pods the istio-proxy image is replaced by the nop image and everything exits successfully. I am currently trying to change my pipeline version to v0.34.0 using the install process listed in this repo and triggers to v0.19.1. Now when I run a taskrun the istio-proxy container is no longer being replaced by the nop image and the pod hangs around for 45 mins or so before finally exiting. Can I configure something to get the previous behavior or is there some way to exit the istio-proxy container after my task run code completes? I'd prefer NOT to send the /quitquitquit message to istio-proxy or specify the istio-proxy as a sidecar for every task if I don't need to.

Comparing versions of taskrun.go from v0.23.0 to current v0.34.0 there seems to be a decent amount of logic in the stopSidecars method. In the old version any non-step containers were considered as sidecars but in the newer version it seems the sidecars must be listed on the taskrun to be considered. Specifically this change added in v0.30.0 with the following commit
https://github.com/tektoncd/pipeline/commit/2a70aedf09b0279a7e77a27e317ccb725ce94a9a
```
// do not continue if the TaskSpec had no sidecars
	if tr.Status.TaskSpec != nil && len(tr.Status.TaskSpec.Sidecars) == 0 {
		return nil
	}
```
Is this change really worth breaking anyone depending on automatically injected sidecars? Is there a more convenient way to keep the same behavior as the older version?

tomkennedy513:
# Changes

Injected sidecars are only cleaned up if sidecars are defined in the task spec because of a check that exits early if the TaskRun does not include any sidecars. This breaks users that are running in environments with injected sidecars, but are not defining sidecars in their TaskRuns.

This change will remove the early exit condition when there are no sidecars defined on the TaskRun Spec. This will cause the podconvert.StopSidecars method to run every time, properly cleaning up sidecars the TaskRun doesn't know about.

/kind bug

Resolves #4731

# Submitter Checklist

- [X] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
  functionality, content, code)
- [X] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [X] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
Fixed istio-proxy container not stopped after running as taskrun #4731
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -253,11 +253,6 @@ func (c *Reconciler) stopSidecars(ctx context.Context, tr *v1beta1.TaskRun) erro
 		return nil
 	}
 
-	// do not continue if the TaskSpec had no sidecars
-	if tr.Status.TaskSpec != nil && len(tr.Status.TaskSpec.Sidecars) == 0 && len(tr.Status.Sidecars) == 0 {
-		return nil
-	}
-
 	// do not continue if the TaskRun was canceled or timed out as this caused the pod to be deleted in failTaskRun
 	condition := tr.Status.GetCondition(apis.ConditionSucceeded)
 	if condition != nil {
@@ -265,11 +260,6 @@ func (c *Reconciler) stopSidecars(ctx context.Context, tr *v1beta1.TaskRun) erro
 		if reason == v1beta1.TaskRunReasonCancelled || reason == v1beta1.TaskRunReasonTimedOut {
 			return nil
 		}
-	}
-
-	// do not continue if there are no Running sidecars.
-	if !podconvert.IsSidecarStatusRunning(tr) {
-		return nil
 	}
 
 	pod, err := podconvert.StopSidecars(ctx, c.Images.NopImage, c.KubeClientSet, tr.Namespace, tr.Status.PodName)
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -4078,65 +4078,85 @@ status:
 	}
 }
 
-func TestStopSidecars_NoClientGetPodForTaskSpecWithoutRunningSidecars(t *testing.T) {
-	for _, tc := range []struct {
-		desc string
-		tr   *v1beta1.TaskRun
-	}{{
-		desc: "no sidecars",
-		tr: parse.MustParseV1beta1TaskRun(t, `
+func TestStopSidecars_WithInjectedSidecarsNoTaskSpecSidecars(t *testing.T) {
+	taskRun := parse.MustParseV1beta1TaskRun(t, `
 metadata:
-  name: test-taskrun
+  name: test-taskrun-injected-sidecars
   namespace: foo
+spec:
+  taskRef:
+    name: test-task
 status:
+  podName: test-taskrun-injected-sidecars-pod
   conditions:
-  - status: "True"
+  - message: Build succeeded
+    reason: Build succeeded
+    status: "True"
     type: Succeeded
-  podName: test-taskrun-pod
-  startTime: "2000-01-01T01:01:01Z"
-`),
-	}, {
-		desc: "sidecars are terminated",
-		tr: parse.MustParseV1beta1TaskRun(t, `
-metadata:
-  name: test-taskrun
-  namespace: foo
-status:
-  conditions:
-  - status: "True"
-    type: Succeeded
-  podName: test-taskrun-pod
-  sidecars:
-  - terminated:
-      exitCode: 0
-      finishedAt: "2000-01-01T01:01:01Z"
-      startedAt: "2000-01-01T01:01:01Z"
-  startTime: "2000-01-01T01:01:01Z"
-`),
-	}} {
-		t.Run(tc.desc, func(t *testing.T) {
-			d := test.Data{
-				TaskRuns: []*v1beta1.TaskRun{tc.tr},
-			}
+`)
 
-			testAssets, cancel := getTaskRunController(t, d)
-			defer cancel()
-			c := testAssets.Controller
-			clients := testAssets.Clients
-			reconcileErr := c.Reconciler.Reconcile(testAssets.Ctx, getRunName(tc.tr))
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-taskrun-injected-sidecars-pod",
+			Namespace: "foo",
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name:  "step-do-something",
+					Image: "my-step-image",
+				},
+				{
+					Name:  "injected-sidecar",
+					Image: "some-image",
+				},
+			},
+		},
+		Status: corev1.PodStatus{
+			Phase: corev1.PodRunning,
+			ContainerStatuses: []corev1.ContainerStatus{
+				{
+					Name:  "step-do-something",
+					State: v1.ContainerState{Terminated: &corev1.ContainerStateTerminated{}},
+				},
+				{
+					Name:  "injected-sidecar",
+					State: v1.ContainerState{Running: &corev1.ContainerStateRunning{}},
+				},
+			},
+		},
+	}
 
-			// We do not expect an error
-			if reconcileErr != nil {
-				t.Errorf("Expected no error to be returned by reconciler: %v", reconcileErr)
-			}
+	d := test.Data{
+		Pods:     []*corev1.Pod{pod},
+		TaskRuns: []*v1beta1.TaskRun{taskRun},
+		Tasks:    []*v1beta1.Task{simpleTask},
+	}
 
-			// Verify that the pod was not retrieved
-			for _, action := range clients.Kube.Actions() {
-				if action.Matches("get", "pods") {
-					t.Errorf("expected the pod not to be retrieved because the TaskRun has no sidecars")
-				}
-			}
-		})
+	testAssets, cancel := getTaskRunController(t, d)
+	defer cancel()
+	c := testAssets.Controller
+	clients := testAssets.Clients
+
+	reconcileErr := c.Reconciler.Reconcile(testAssets.Ctx, getRunName(taskRun))
+
+	// We do not expect an error
+	if reconcileErr != nil {
+		t.Errorf("Expected no error to be returned by reconciler: %v", reconcileErr)
+	}
+
+	retrievedPod, err := clients.Kube.CoreV1().Pods(pod.Namespace).Get(testAssets.Ctx, pod.Name, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("error retrieving pod: %s", err)
+	}
+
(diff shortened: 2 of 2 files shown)


dibyom:
> Side effects of this change are that non-user-defined sidecars will show in the TaskRun status.

@chuangw6 @wlynch does this affect chains at all?

tomkennedy513:
> > Side effects of this change are that non-user-defined sidecars will show in the TaskRun status.
> 
> @chuangw6 @wlynch does this affect chains at all?

If this does end up having a negative effect, an alternate approach could be relaxing the early exit optimizations (I believe these were added to minimize api calls to retrieve pods) for the stopSidecar method in the case where the `running-in-environment-with-injected-sidecars` feature flag is true so that we don't check to the task status to find running containers and call the api to get the pod. In the event that `running-in-environment-with-injected-sidecars` is false (this would have to be manually changed since it defaults to true), then we would proceed as normal

wlynch:
> @chuangw6 @wlynch does this affect chains at all?

Chains shouldn't need additional changes. 🎉 

For Tekton based provenance we just spit out the status: https://github.com/tektoncd/chains/blob/bc296d1ee1491df7204211cf6d5bd8df9a89c157/pkg/chains/formats/tekton/tekton.go#L34-L37

In-toto is WIP - @chitrangpatel pointed out we aren't capturing sidecars and other fields in the predicate at the moment, but we'll look to populate from the Run status once we support it.

dibyom:
Sorry for the delay on this @tomkennedy513 - This looks good to me but wanted to see if any of the other @tektoncd/core-maintainers had any opinions on this

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -248,11 +248,6 @@ func (c *Reconciler) stopSidecars(ctx context.Context, tr *v1beta1.TaskRun) erro
 		return nil
 	}
 
-	// do not continue if the TaskSpec had no sidecars
```
XinruZhang, line 251: Do we need to delete this considering the change in pkg/pod/status.go?

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -248,11 +248,6 @@ func (c *Reconciler) stopSidecars(ctx context.Context, tr *v1beta1.TaskRun) erro
...
-	if tr.Status.TaskSpec != nil && len(tr.Status.TaskSpec.Sidecars) == 0 {
-		return nil
-	}
-
```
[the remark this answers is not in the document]
tomkennedy513: I don;t think we need to delete it, but do to the changes in status.go, [this check](https://github.com/tektoncd/pipeline/blob/8f686f868f6935d426e8e5eb5d47ad98f1ba9dac/pkg/reconciler/taskrun/taskrun.go#L261) should be sufficient to exit early imo.

mmlac:
Is this only blocked because `parse.MustParseTaskRun` needs to be changed to `parse.MistParseV1beta1TaskRun`?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4788,6 +4788,66 @@ status:
 	}
 }
 
+func TestStopSidecars_WithInjectedSidecarsNoTaskSpecSidecars(t *testing.T) {
+	taskRun := parse.MustParseTaskRun(t, `
```
pritidesai, line 4792: please replace this with `MustParseV1beta1TaskRun`

tomkennedy513: done!

tomkennedy513:
@abayer Is there anything else needed to get this approved?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4438,6 +4438,66 @@ status:
...
+		t.Errorf("Expected no error to be returned by reconciler: %v", reconcileErr)
+	}
+
+	// Verify that the pod was retrieved.
+	getPodFound := false
+	for _, action := range clients.Kube.Actions() {
+		if action.Matches("get", "pods") {
+			getPodFound = true
+			break
+		}
```
lbernick, line 4494: it might be simpler to use `clients.Kube.CoreV1().Pods("foo").Get("test-taskrun-injected-sidecars-pod")` (not sure of exact syntax) since this relies less on implementation details and is just a bit more readable

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4438,6 +4438,66 @@ status:
...
+	}
+	if !getPodFound {
```
lbernick, line 4496: It looks like this doesn't actually test that the pod stopped successfully; it just tests that the pod exists. Am I missing something?

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -258,11 +258,6 @@ func (c *Reconciler) stopSidecars(ctx context.Context, tr *v1beta1.TaskRun) erro
 		return nil
 	}
 
-	// do not continue if the TaskSpec had no sidecars
-	if tr.Status.TaskSpec != nil && len(tr.Status.TaskSpec.Sidecars) == 0 && len(tr.Status.Sidecars) == 0 {
```
lbernick, line 262: since the injected sidecars are now populated in the taskrun status, could this be
```
if len(tr.Status.Sidecars) == 0 {
```
?

pkg/pod/status.go
```diff
@@ -135,7 +135,7 @@ func MakeTaskRunStatus(ctx context.Context, logger *zap.SugaredLogger, tr v1beta
 	for _, s := range pod.Status.ContainerStatuses {
 		if IsContainerStep(s.Name) {
 			stepStatuses = append(stepStatuses, s)
-		} else if isContainerSidecar(s.Name) {
+		} else {
```
lbernick, line 138: suggest adding a comment about why we add all other containers to sidecar statuses

afrittoli (commented):
Thanks for this change.
The list of containers running in the `Pod` should be available when we stop the sidecars, so I wonder if we really need to add injected sidecars in the `TaskRun` status to be able to stop them. Couldn't we instead stop both Tekton and non-Tekton sidecars and avoid increasing the size of the `TaskRun` status with data already available in the `Pod`?

tomkennedy513:
> Thanks for this change. The list of containers running in the `Pod` should be available when we stop the sidecars, so I wonder if we really need to add injected sidecars in the `TaskRun` status to be able to stop them. Couldn't we instead stop both Tekton and non-Tekton sidecars and avoid increasing the size of the `TaskRun` status with data already available in the `Pod`?

I think that it is totally feasible. The reason why I added it to the TaskRun Status was because it felt more in line with the SidecarsReady Method which checks that all non-step containers are ready, implying that every non-step container is a sidecar and we want to wait for all of them, so it felt odd to me to leave these other sidecars out of the status even though they are treated the same as defined sidecars from the spec.

pkg/reconciler/taskrun/taskrun.go
tomkennedy513: Yes, I can add that back if needed, but I'm sure its providing much value. We check to see if the sidecars are running a few lines down and that would return immediately if there were no sidecars to iterate over so it felt like we were adding unnecessary complexity for negligible performance gains by not checking the TaskRun condition.

tomkennedy513: Alternatively, I think the best path forward might be pushing the check for running sidecars above the TaskRun condition check due to the fact that the state of the TaskRun doesn't matter if there are no sidecars anyways

pkg/reconciler/taskrun/taskrun_test.go
tomkennedy513: I took that implementation from the test above for consistency, but it is checking that the reconciler actually reached out to get the actual pod which means that it made it to the point where it had sidecars to stop

pkg/reconciler/taskrun/taskrun_test.go
tomkennedy513: same as https://github.com/tektoncd/pipeline/pull/5565/files/9d97e72d72f727d684ab01dc552b170f44f2babc#r1136216449

lbernick:
> > Thanks for this change. The list of containers running in the `Pod` should be available when we stop the sidecars, so I wonder if we really need to add injected sidecars in the `TaskRun` status to be able to stop them. Couldn't we instead stop both Tekton and non-Tekton sidecars and avoid increasing the size of the `TaskRun` status with data already available in the `Pod`?
> 
> I think that it is totally feasible. The reason why I added it to the TaskRun Status was because it felt more in line with the SidecarsReady Method which checks that all non-step containers are ready, implying that every non-step container is a sidecar and we want to wait for all of them, so it felt odd to me to leave these other sidecars out of the status even though they are treated the same as defined sidecars from the spec.

I agree with @afrittoli's comment; we could always add injected sidecar containers into the taskrun status if needed later on, but once there, it's harder to remove them.

pkg/reconciler/taskrun/taskrun_test.go
lbernick: I think our existing tests could probably use some improvement here; they seem to test implementation details rather than functionality visible to users. I would create a test that has a pod with completed step containers and running injected sidecar containers, and tests that the injected containers are replaced with the nop image.

pritidesai:
This PR did not make it 0.46, moving it to 0.47.

@tomkennedy513 @afrittoli @lbernick are we aligned on the suggested changes?

tomkennedy513:
> This PR did not make it 0.46, moving it to 0.47.
> 
> 
> 
> @tomkennedy513 @afrittoli @lbernick are we aligned on the suggested changes?

I can make the requested changes this coming week

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4078,65 +4078,85 @@ status:
...
+	reconcileErr := c.Reconciler.Reconcile(testAssets.Ctx, getRunName(taskRun))
+
+	// We do not expect an error
+	if reconcileErr != nil {
+		t.Errorf("Expected no error to be returned by reconciler: %v", reconcileErr)
+	}
+
+	retrievedPod, err := clients.Kube.CoreV1().Pods(pod.Namespace).Get(testAssets.Ctx, pod.Name, metav1.GetOptions{})
+	if err != nil {
+		t.Errorf("error retrieving pod: %s", err)
```
lbernick, line 4150:
```suggestion
		t.Fatalf("error retrieving pod: %s", err)
```

lbernick (approved):

tomkennedy513:
@lbernick I made those test changes. Thank you for the suggestions!

afrittoli (commented):
Thanks for this fix. See my comment about the release notes.
/lgtm

PR #5573, opened 2022-09-28
tekton: make sure the git workingdir is not dirty…

vdemeester:
# Changes

… when building the released binaries/images.

As of today, we are modifying the source and building. This means `knative.dev/pkg/changeset` will require the commit hash suffixed by `-dirty`. This can be a bit confusing the users as, well, this means the source is not really the source of truth of the binary :)

This changes that:
- it generates `.ko.yaml` in `/workspace` so that it's not in the git repository. It uses `KO_CONFIG_PATH` environment variable to force `ko` to use it.
- it does the devel->version rewrite on generated yamls, so after the build is done.

/kind cleanup
/cc @imjasonh @mattmoor @afrittoli @abayer 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fix the `-dirty` suffix in `pipeline.tekton.dev/release` annotation
```

### DIFF
--- a/tekton/publish.yaml
+++ b/tekton/publish.yaml
@@ -104,7 +104,7 @@ spec:
         ${CONTAINER_REGISTRY}/$(params.package)/combined-base-image:latest)
 
       # NOTE: Make sure this list of images to use the combined base image is in sync with what's in test/presubmit-tests.sh's 'ko_resolve' function.
-      cat <<EOF > ${PROJECT_ROOT}/.ko.yaml
+      cat <<EOF > /workspace/.ko.yaml
       # This matches the value configured in .ko.yaml
       defaultBaseImage: cgr.dev/chainguard/static
       baseImageOverrides:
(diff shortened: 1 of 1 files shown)


tekton/publish.yaml
```diff
@@ -130,6 +130,9 @@ spec:
       #!/usr/bin/env sh
       set -ex
 
+      # Use the generated `.ko.yaml`
+      KO_CONFIG_PATH=/workspace/.ko.yaml
```
imjasonh, line 134: I was surprised to find this accepted a filepath and not the directory.

I was _even more surprised_ to find I reviewed and approved the PR that made it accept being a filepath: https://github.com/ko-build/ko/pull/731 🙃

abayer (approved):

imjasonh:
Would it make sense to have the release pipeline fail if its working dir is dirty, instead of silently pushing images that have `-dirty` versions?

vdemeester:
> Would it make sense to have the release pipeline fail if its working dir is dirty, instead of silently pushing images that have `-dirty` versions?

good point, and I would say yes. let me update the PR to add this as well.

imjasonh (approved):

Issue #5537, opened 2022-09-22
Taskrun not working with volumeClaimTemplate


piyush-garg:
# Expected Behavior

Taskrun Should Work

# Actual Behavior

Taskrun is throwing an error

```
task task-with-workspace has failed: failed to create task run pod "task-with-workspace-run-j622l": Pod "task-with-workspace-run-j622l-pod" is invalid: [spec.volumes[10].name: Required value, spec.containers[0].volumeMounts[0].name: Required value, spec.containers[0].volumeMounts[0].name: Not found: ""]. Maybe missing or invalid Task tekton-pipelines/task-with-workspace
Error: pod for taskrun task-with-workspace-run-j622l not available yet
```

# Steps to Reproduce the Problem

1. Create a task

```
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: task-with-workspace
spec:
  steps:
  - name: list-files
    image: ubuntu
    script: ls $(workspaces.read-allowed.path)
  workspaces:
  - name: read-allowed
```

2. Create taskrun

```
apiVersion: tekton.dev/v1beta1
kind: TaskRun
metadata:
  generateName: task-with-workspace-run-
spec:
  taskRef:
    kind: Task
    name: task-with-workspace
  workspaces:
    - name: read-allowed
      volumeClaimTemplate:
        metadata:
          name: pvc
        spec:
          accessModes:
            - ReadWriteOnce
          resources:
            requests:
              storage: 1Gi
          volumeMode: Filesystem
```

3. See the taskrun getting failed.

# Additional Info

- Kubernetes version:

```
  kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.1", GitCommit:"e4d4e1ab7cf1bf15273ef97303551b279f0920a9", GitTreeState:"clean", BuildDate:"2022-09-14T19:49:27Z", GoVersion:"go1.19.1", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
[...]

PR #5576, opened 2022-09-28
Write TaskRun.Status.TaskSpec with replaced spec on every reconcile run

### LINKED ISSUE #5574: TaskRun.Status.TaskSpec is sometimes overwritten without replacements applied
Discovered this due to `TestPipelineRunStatusSpec/pipeline_status_spec_updated` e2e test being flaky:

```
init_test.go:138: Create namespace arendelle-t9lb4 to deploy to
    init_test.go:166: Verify SA "default" is created in namespace "arendelle-t9lb4"
    pipelinerun_test.go:110: Setting up test resources for "pipeline status spec updated" test in namespace arendelle-t9lb4
    pipelinerun_test.go:120: Waiting for PipelineRun pipeline-task-update in namespace arendelle-t9lb4 to complete
    pipelinerun_test.go:124: Making sure the expected TaskRuns [task1] were created
    pipelinerun_test.go:164: Checking if parameter replacements have been updated in the spec.
    pipelinerun_test.go:171: Expected replaced parameter value : Hello World! in Script: #!/usr/bin/env bash
        echo "$(params.HELLO)"
         But not found
```

Seen on https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5573/pull-tekton-pipeline-alpha-integration-tests/1575135529287028736 among others.

/kind bug

abayer:
# Changes

fixes #5574

With #5537, the call to `applyParamsContextsResultsAndWorkspaces` and setting of `tr.Status.TaskSpec` to the output of that call was moved from happening with every call to `reconcile` to only happening if there wasn't already a pod created. This didn't cause any problems with execution, but it sometimes resulted in `tr.Status.TaskSpec` not getting set properly or getting reset to `nil` on a subsequent `reconcile` run, with the end result that, sometimes, `tr.Status.TaskSpec` would contain the original `TaskSpec` without parameter, result, context, and workspace references being replaced with the corresponding value. This caused flaky failures in the `TestPipelineRunStatusSpec/pipeline_status_spec_updated` e2e test, and integration test failures in Chains (https://github.com/tektoncd/chains/issues/577).

This change moves the PVC creation out of the `if pod == nil {` block that creates the pod if needed, while still checking if `pod` is `nil` before creating the PVC, and brings the `applyParamsContextsResultsAndWorkspaces` call and setting of `tr.Status.TaskSpec` back out of the pod creation block, but after the possible PVC creation.

Note that I never managed to figure out what conditions resulted in the flaky behavior - I was deep in that rabbit hole when @wlynch flagged the Chains issue and I realized what the root cause was. Since I never nailed down the flake trigger, I don't have a new e2e test to add here - but then again, since I started looking at this in the first place due to `TestPipelineRunStatusSpec/pipeline_status_spec_updated` flaking, we do know that it's covered to some extent. =)

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fix TaskRun parameter etc replacement logic to persist in the TaskRun's Status properly
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -481,31 +481,37 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1beta1.TaskRun, rtr *re
 		}
 	}
 
+	// Please note that this block is required to run before `applyParamsContextsResultsAndWorkspaces` is called the first time,
+	// and that `applyParamsContextsResultsAndWorkspaces` _must_ be called on every reconcile.
+	if pod == nil && tr.HasVolumeClaimTemplate() {
+		if err := c.pvcHandler.CreatePersistentVolumeClaimsForWorkspaces(ctx, tr.Spec.Workspaces, *kmeta.NewControllerRef(tr), tr.Namespace); err != nil {
+			logger.Errorf("Failed to create PVC for TaskRun %s: %v", tr.Name, err)
+			tr.Status.MarkResourceFailed(volumeclaim.ReasonCouldntCreateWorkspacePVC,
+				fmt.Errorf("Failed to create PVC for TaskRun %s workspaces correctly: %s",
+					fmt.Sprintf("%s/%s", tr.Namespace, tr.Name), err))
+			return controller.NewPermanentError(err)
+		}
+
+		taskRunWorkspaces := applyVolumeClaimTemplates(tr.Spec.Workspaces, *kmeta.NewControllerRef(tr))
+		// This is used by createPod below. Changes to the Spec are not updated.
+		tr.Spec.Workspaces = taskRunWorkspaces
+	}
+
+	// Get the randomized volume names assigned to workspace bindings
+	workspaceVolumes := workspace.CreateVolumes(tr.Spec.Workspaces)
+
+	ts, err := applyParamsContextsResultsAndWorkspaces(ctx, tr, rtr, workspaceVolumes)
+	if err != nil {
+		logger.Errorf("Error updating task spec parameters, contexts, results and workspaces: %s", err)
+		return err
+	}
+	tr.Status.TaskSpec = ts
+
+	if len(tr.Status.TaskSpec.Steps) > 0 {
+		logger.Infof("set taskspec for %s/%s - script: %s", tr.Namespace, tr.Name, tr.Status.TaskSpec.Steps[0].Script)
+	}
+
 	if pod == nil {
-		if tr.HasVolumeClaimTemplate() {
-			if err := c.pvcHandler.CreatePersistentVolumeClaimsForWorkspaces(ctx, tr.Spec.Workspaces, *kmeta.NewControllerRef(tr), tr.Namespace); err != nil {
-				logger.Errorf("Failed to create PVC for TaskRun %s: %v", tr.Name, err)
-				tr.Status.MarkResourceFailed(volumeclaim.ReasonCouldntCreateWorkspacePVC,
-					fmt.Errorf("Failed to create PVC for TaskRun %s workspaces correctly: %s",
-						fmt.Sprintf("%s/%s", tr.Namespace, tr.Name), err))
-				return controller.NewPermanentError(err)
-			}
-
-			taskRunWorkspaces := applyVolumeClaimTemplates(tr.Spec.Workspaces, *kmeta.NewControllerRef(tr))
-			// This is used by createPod below. Changes to the Spec are not updated.
-			tr.Spec.Workspaces = taskRunWorkspaces
-		}
-
-		// Get the randomized volume names assigned to workspace bindings
-		workspaceVolumes := workspace.CreateVolumes(tr.Spec.Workspaces)
-
-		ts, err := applyParamsContextsResultsAndWorkspaces(ctx, tr, rtr, workspaceVolumes)
-		if err != nil {
-			logger.Errorf("Error updating task spec parameters, contexts, results and workspaces: %s", err)
-			return err
-		}
-		tr.Status.TaskSpec = ts
-
 		pod, err = c.createPod(ctx, ts, tr, rtr, workspaceVolumes)
 		if err != nil {
 			newErr := c.handlePodCreationError(tr, err)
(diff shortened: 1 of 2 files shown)


vdemeester (approved):
/cherry-pick release-v0.40.x

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -481,31 +481,35 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1beta1.TaskRun, rtr *re
...
-		}
-		tr.Status.TaskSpec = ts
+	// Get the randomized volume names assigned to workspace bindings
+	workspaceVolumes := workspace.CreateVolumes(tr.Spec.Workspaces)
 
+	ts, err := applyParamsContextsResultsAndWorkspaces(ctx, tr, rtr, workspaceVolumes)
+	if err != nil {
+		logger.Errorf("Error updating task spec parameters, contexts, results and workspaces: %s", err)
+		return err
+	}
```
wlynch, line 505: Do these need to be computed each time, or does this only need to happen if pod == nil?

This might help clean things up a little bit if we can preserve the old structure so we can keep all the `if pod == nil` logic together - i.e.

```go
if pod == nil {
  if tr.HasVolumeClaimTemplate() {
    ...
  }
 
  workspaceVolumes := workspace.CreateVolumes(tr.Spec.Workspaces)
  ...

  pod, err = c.createPod(ctx, ts, tr, rtr, workspaceVolumes)
  ...
}

tr.Status.TaskSpec = ts
```

wlynch (commented):
Would it be worth doing a deeper comparison here instead of just checking that the reconcile succeeded? 🤔 

https://github.com/tektoncd/pipeline/blob/3e6e4dd516bb43d796334be4634647cfc91c4c97/pkg/reconciler/taskrun/taskrun_test.go#L1316-L1326

pkg/reconciler/taskrun/taskrun.go
abayer: The problem is that we need `workspaceVolumes` as input to `applyParamsContextsResultsAndWorkspaces`, and the whole issue here stemmed from that not getting called and `tr.Status.TaskSpec` not getting updated, so I think we do need to the `createVolumes` call every time. That's what was being done before https://github.com/tektoncd/pipeline/commit/9cf7590959acd9ad9518e62d56f5941885614446

wlynch:
/lgtm
/hold

Holding in case you wanted to make any changes to taskrun_test.go before submission, else feel free to remove and submit.

abayer:
> Holding in case you wanted to make any changes to taskrun_test.go before submission, else feel free to remove and submit.

I think I'm going to write some new tests to exercise all of this properly, but that'll probably take me a day or two, so we can merge this before then if it's urgent enough. But let's see what I can get done in the next hour or two...

abayer:
(the reason for new tests is that that particular test, for example, would always pass even without this fix, even if it was actually doing param/etc replacement, because it's only getting reconciled once and is creating a pod - the problem was reconciling when there _isn't_ a pod)

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4925,3 +4925,74 @@ status:
...
+metadata:
+  name: test-task-with-replacements
+  namespace: foo
+spec:
+  params:
+  - default: mydefault
+    name: myarg
+    type: string
+  steps:
+  - script: echo $(inputs.params.myarg)
```
abayer, line 4940: I'm only testing for replacement of a param here because that's good enough to prove the point, and is simplest to do. Note that this issue does _not_ show up with an inline `tr.Spec.TaskSpec`, interestingly.

wlynch (approved):
/lgtm

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -4925,3 +4925,74 @@ status:
...
+metadata:
+  name: test-taskrun-with-replacements
+  namespace: foo
+spec:
+  params:
+  - name: myarg
+    value: foo
+  taskRef:
+    name: test-task-with-replacements
+status:
```
abayer, line 4954: I initially tried this with a pre-set `tr.Status.TaskSpec`, but that kept passing when it shouldn't have. I'm guessing it might have eventually flaked into failure, but it's very possible that the specific conditions that triggered the flaky behavior would never actually show up in a unit test, so I just went with testing to make sure that a `nil` `tr.Status.TaskSpec` gets written properly.

abayer:
@wlynch Ok, I think I'm good with this test. It's not exercising whatever the specific trigger is that sets off the flaky behavior where `tr.Status.TaskSpec` ends up nil on reconcile, but it is making sure that the replacements are being applied to `tr.Status.TaskSpec` when there already is a pod.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -481,31 +481,35 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1beta1.TaskRun, rtr *re
...
-				tr.Status.MarkResourceFailed(volumeclaim.ReasonCouldntCreateWorkspacePVC,
-					fmt.Errorf("Failed to create PVC for TaskRun %s workspaces correctly: %s",
-						fmt.Sprintf("%s/%s", tr.Namespace, tr.Name), err))
-				return controller.NewPermanentError(err)
-			}
-
-			taskRunWorkspaces := applyVolumeClaimTemplates(tr.Spec.Workspaces, *kmeta.NewControllerRef(tr))
-			// This is used by createPod below. Changes to the Spec are not updated.
-			tr.Spec.Workspaces = taskRunWorkspaces
+	if pod == nil && tr.HasVolumeClaimTemplate() {
```
afrittoli, line 484: I think it might be worth adding a comment here on the conditions to avoid the conditions moving around again when the next change happens in the code.

abayer: 👍

abayer:
@wlynch ping? =)

wlynch (approved):
/lgtm

abayer:
@wlynch Sorry to bug you again, but I didn't want to remove the hold without your approval.

wlynch:
/hold cancel 

Hold was more for you than for me. :)

/gogogo

Issue #5541, opened 2022-09-22
Release v1 versions of Task, TaskRun, Pipeline, and PipelineRun CRDs


lbernick:
This issue tracks the release of v1 api versions for Task, TaskRun, Pipeline, and PipelineRun.

Before doing this, we would like to wait for remote resolution beta (https://github.com/tektoncd/pipeline/issues/4710), since we want to provide an alternative to taskref.bundle and pipelineref.bundle.

It would be nice to wait for custom tasks beta (https://github.com/tektoncd/pipeline/issues/4313) so we can release v1 versions of these CRDs along with custom task beta, but there's still more work to do there and I'm not sure this should be a blocker.

Steps:
1. Create examples tests for v1 versions of our CRDs (probably just copy pasting existing examples tests, swapping api versions, and removing any that aren't relevant for v1 such as those that test pipelineresources). This will probably require step 2 but it would be nice if there's a way to do it first. It will probably be easiest to do all the CRDs at once, but it's not necessary to do so.
2. Update CRDs to serve v1 versions. At this point, we're "committing" to the structure of the v1 api-- no more changes.
3. Update CRDs to use v1 as the stored version, and in the same PR, update the reconcilers to reconcile the v1 versions of the CRDs. Again, we don't have to do all the CRDs at once here, but we might find it easier to do so. One catch is that we must deserialize PipelineResources from annotations in v1 CRDs if they are present. (They may have been created by conversion from v1beta1 CRDs referencing PipelineResources.) Failing to support PipelineResources in this way would be violating our compatibility policy by not supporting a v1beta1 feature for 9 months following the v1 release. After completing step 3, we'll be able to add new features in v1 and should stop making any changes to v1beta1.

lbernick:
Discussed in today's api wg: folks seem to be OK with going forward with this plan without blocking on custom tasks beta but would like to block on remote resolution beta. We should review our plan for supporting deprecated fields (#4546) before swapping our stored versions over to v1.

JeromeJu:
Would like to confirm the understanding on the part iii.

[...]

Issue #5504, opened 2022-09-15
Improve API versioning


lbernick:
Currently, we use the "enable-api-fields" feature flag to allow users to opt into features at different stability levels, as described in [tep-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md).

We have discussed moving towards maintaining multiple versions of CRDs in the future (for example, a v1 version with stable features and a v2alpha1 version with stable and alpha features), but the main reason we don't do this is [because it creates a large maintenance burden for contributors](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#maintain-multiple-verisons-of-impacted-crds).

This issue can be used to brainstorm ways to make it easier to maintain multiple API versions of our CRDs.
Some ideas that have already come up:

- move as much code as possible out of the apis package, such as validation functions, so that it doesn't need to be maintained in multiple places
- create a CRD version that's stored but not served, and all other CRD versions are converted to this one. this would be the version used in the reconciler.
- encouraging users to use the [dynamic client](https://pkg.go.dev/k8s.io/client-go/dynamic) instead of the tekton versioned client. however, we'd still need to support all versions of the CRD, so I don't think this is really easier from a contributor effort perspective.
- some kind of tooling to be able to easily view the diff between pipeline version packages?

lbernick:
@dibyom has written up a proposal for a new API versioning strategy: https://docs.google.com/document/d/15vD5vbK5UKCDFMoOqOwEc_NYACthtCZJlTJiiEEoljY

lbernick:
Looking at the [scheduler-plugins repo](https://github.com/kubernetes-sigs/scheduler-plugins), they've found a way to register "internal versions" of the types their controller operates on (example: https://github.com/kubernetes-sigs/scheduler-plugins/blob/7da686c01cf3bc5d8fe852334698468e046bb000/apis/config/register.go#L26), and then all the versioned types are automatically converted into the unversioned internal type. 

@dibyom I know you originally suggested something like this and it seems like it's actually possible which is very cool! We'd have to change our storage version over to this "unversioned" type, but once we do this, I would guess we wouldn't have to do it again (e.g. if we want to create a v2 api). @tektoncd/core-maintainers thoughts on this?

FYI @JeromeJu

dibyom:
[...]

PR #5579, opened 2022-09-28
V1 CRD release

JeromeJu:
# Changes
This commit releases the v1 pipeline CRDs including task, taskrun, pipeline and
piplinerun by the following:
- Adds the example tests for the release of v1 pipeline CRD.
It copies the example tests from v1beta1 with exceptions for those deprecated
fields eg. `resources`. Reference can be found at https://github.com/tektoncd/pipeline/blob/14eba83bf31277dc3f8c91fa798b38e30d0a8e6a/dev/docs/migrating-v1beta1-to-v1.md.
- Sets `served` in v1 CRDs to true.

The second commit adds the integration test for v1 CRD:
- This covers all the deprecated fields from v1beta1.
- It first creates v1beta1 CRD with v1beta1Client and requests it by v1client,
and then creates v1 CRD with v1Client and requests it by v1beta1client to 
compare if the conversion webhook.

Part i & ii of #5541 
Fixes: #5627 

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Releases v1 CRD for task, taskrun, pipeline and pipelinerun. Served v1 CRDs are included in this release for users as a preview, and also for users to start preparing for the migration. v1 is served but CLI, Dashboard, Chains etc support won't be available until the following release. v1 will be the official version starting with the next release when it has been released as the storage version.
```


examples/v1/pipelineruns/pipelinerun-with-final-results.yaml
```diff
@@ -0,0 +1,77 @@
+apiVersion: tekton.dev/v1beta1
```
lbernick, line 1: the CRDs in this file should be v1

examples/v1/pipelineruns/pipelinerun-with-final-tasks.yaml
```diff
@@ -0,0 +1,204 @@
+# Copied from https://github.com/tektoncd/catalog/blob/v1beta1/git/git-clone.yaml :(
+# This can be deleted after we add support to refer to the remote Task in a registry (Issue #1839) or
+# add support for referencing task in git directly (issue #2298)
```
lbernick, line 3: can this be changed? these features have been added

examples/v1/pipelineruns/pipelinerun.yaml
```diff
@@ -0,0 +1,334 @@
...
+    script: |
+      # The intention behind this example Task is to run unit test, however we
+      # currently do nothing to ensure that a unit test issue doesn't cause this example
+      # to fail unnecessarily. In the future we could re-introduce the unit tests (since
+      # we are now pinning the version of Skaffold we pull) or use Tekton Pipelines unit tests.
+      echo "pass"
+---
+# Copied from https://github.com/tektoncd/catalog/blob/v1beta1/git/git-clone.yaml
+# With a few fixes being ported over in https://github.com/tektoncd/catalog/pull/290
+# Post #1839 we can refer to the remote Task in a registry or post #2298 in git directly
```
lbernick, line 41: same here

examples/v1/pipelineruns/pipelinerun.yaml
```diff
@@ -0,0 +1,334 @@
...
+      then
+        exit $EXIT_CODE
+      fi
+      # Make sure we don't add a trailing newline to the result!
+      echo -n "$RESULT_SHA" > $(results.commit.path)
+---
+# Copied from https://github.com/tektoncd/catalog/blob/v1beta1/kaniko/kaniko.yaml
+# with a few fixes that will be port over in https://github.com/tektoncd/catalog/pull/291
+# Post #1839 we can refer to the remote Task in a registry or post #2298 in git directly
+apiVersion: tekton.dev/v1
```
lbernick, line 125: same here

lbernick (commented):
could you please move the v1 migration docs into the docs folder?

1
```diff
@@ -0,0 +1,30 @@
...
+# m, merge [-C <commit> | -c <commit>] <label> [# <oneline>]
+# .       create a merge commit using the original merge commit's
+# .       message (or the oneline, if no original merge commit was
+# .       specified); use -c <commit> to reword the commit message
+#
+# These lines can be re-ordered; they are executed from top to bottom.
+#
+# If you remove a line here THAT COMMIT WILL BE LOST.
+#
+# However, if you remove everything, the rebase will be aborted.
```
chuangw6, line 29: nit: we might want to delete this? 👀

examples/v1/pipelineruns/pipelinerun-with-final-results.yaml
JeromeJu: Sorry was still debugging on this as if I changed to v1, will have the error
```Error from server (BadRequest): error when creating "examples/v1/pipelineruns/pipelinerun-with-final-results.yaml": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: invalid value: referencing a nonexistent task: spec.results[1].value
```

1
JeromeJu: Thanks Chuang. Sorry my local vscode config introduces this.

examples/v1/pipelineruns/pipelinerun-with-final-tasks.yaml
lbernick: I saw that you removed the code comment here, but what I'm suggesting is actually swapping to using the catalog task (pin the version) so this can serve as an example. Might need to wait on resolvers being enabled by default (part of RR beta PR)

lbernick (approved):

lbernick:
@abayer FYI since the only thing blocking this PR is https://github.com/tektoncd/pipeline/pull/5515-- you might need to coordinate with Jerome here

vdemeester:
Are we sure 100% we want `v1` API in 0.41 ? Because as soon as we publish them, we will not be able to change it that much. And we should probably agree on things like [Tekton Pipeline V1 policy](https://docs.google.com/document/d/15vD5vbK5UKCDFMoOqOwEc_NYACthtCZJlTJiiEEoljY/edit#heading=h.lfq10wr4d4eq) prior to getting this in.

lbernick:
From discussion at yesterday's api wg, we want to tackle two tasks before merging this:
- https://github.com/tektoncd/pipeline/issues/5504 (dibyo is working on this)
- https://github.com/tektoncd/pipeline/issues/5627: Might have to be part of this PR, since we can't have integration tests for new api versions until they are served, but we don't want to serve them until they are tested

test/clients.go
```diff
@@ -57,14 +58,18 @@ import (
...
-	PipelineResourceClient  resourcev1alpha1.PipelineResourceInterface
-	RunClient               v1alpha1.RunInterface
-	ResolutionRequestclient resolutionv1alpha1.ResolutionRequestInterface
+	V1beta1PipelineClient    v1beta1.PipelineInterface
+	V1beta1ClusterTaskClient v1beta1.ClusterTaskInterface
+	V1beta1TaskClient        v1beta1.TaskInterface
+	V1beta1TaskRunClient     v1beta1.TaskRunInterface
+	V1beta1PipelineRunClient v1beta1.PipelineRunInterface
+	PipelineResourceClient   resourcev1alpha1.PipelineResourceInterface
+	RunClient                v1alpha1.RunInterface
```
XinruZhang, line 67: Let's keep the naming consistent, what about `V1alpha1RunClient` and `V1alpha1ResolutionRequestClient`

XinruZhang (commented):
Should we delete `examples/v1/pipelineruns/no-ci/.keep`?

test/clients.go
XinruZhang: An example: 

https://github.com/tektoncd/pipeline/blob/1b941d31d7a47c8710cc19bc3b161ce265aa96dc/test/retry_test.go#L50

https://github.com/tektoncd/pipeline/blob/1b941d31d7a47c8710cc19bc3b161ce265aa96dc/test/featureflags.go#L93-L103

test/conversion_test.go
```diff
@@ -0,0 +1,879 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	knativetest "knative.dev/pkg/test"
+	"knative.dev/pkg/test/helpers"
+)
+
+var (
+	// supportedConversionFeatureGates = map[string]string{
+	// 	"enable-tekton-oci-bundles": "true",
+	// }
```
XinruZhang, line 37: I'm thinking if we want to keep this conversion test in both `pull-tekton-pipeline-build-tests` and `pull-tekton-pipeline-alpha-integration-tests`, which means we may need to write different test cases for different feature-gate values.

XinruZhang: An example: 

https://github.com/tektoncd/pipeline/blob/1b941d31d7a47c8710cc19bc3b161ce265aa96dc/test/retry_test.go#L50

https://github.com/tektoncd/pipeline/blob/1b941d31d7a47c8710cc19bc3b161ce265aa96dc/test/featureflags.go#L93-L103

test/conversion_test.go
```diff
@@ -0,0 +1,879 @@
...
+	if err != nil {
+		t.Fatalf("Couldn't get expected Task for %s: %s", taskName, err)
+	}
+
+	if d := cmp.Diff(v1TaskGot, v1Task, filterTypeMeta, filterObjectMeta); d != "" {
+		t.Fatalf("-got, +want: %v", d)
+	}
+}
+
+func TestTaskRunCRDConversion(t *testing.T) {
```
XinruZhang, line 398: It's better to add comments here (as well as other two tests) to clarify that this test is testing the conversion from v1beta1 to v1, similar to what you did for `TestTaskCRDConversion`.

JeromeJu:
Currently separating the second commit to another PR for v1-client additon.

test/clients.go
JeromeJu: Thanks Xinru, changed in https://github.com/tektoncd/pipeline/pull/5701

pritidesai:
Pipeline WG - no blockers

JeromeJu:
Thanks for the catch @XinruZhang , I was wondering what's point of keeping it in v1beta1 also?
> Should we delete examples/v1/pipelineruns/no-ci/.keep?

XinruZhang:
> Thanks for the catch @XinruZhang , I was wondering what's point of keeping it in v1beta1 also?
> 
> > Should we delete examples/v1/pipelineruns/no-ci/.keep?

learning: https://www.sololearn.com/Discuss/1336098/what-are-keep-and-gitkeep-used-for-in-git

Since the folder is not empty, we don't need that file anymore.

examples/v1/pipelineruns/alpha/pipelinerun-with-final-tasks.yaml
```diff
@@ -0,0 +1,103 @@
...
+  tasks:
+    # Clone app repo to workspace
+    - name: clone-app-repo
+      taskRef:
+        resolver: git
+        params:
+        - name: url
+          value: https://github.com/tektoncd/catalog.git
+        - name: revision
+          value: v1beta1
```
abayer, line 49: This should be `main`

examples/v1/pipelineruns/alpha/pipelinerun-with-final-tasks.yaml
```diff
@@ -0,0 +1,103 @@
...
+        - name: pathInRepo
+          value: git/git-clone.yaml
```
abayer, line 51: This should be `task/git-clone/0.9/git-clone.yaml`

JeromeJu:
/retest 
CI not the latest

JeromeJu:
/retest
codebase not updated in CI

test/conversion_test.go
```diff
@@ -0,0 +1,794 @@
...
+func TestTaskCRDConversion(t *testing.T) {
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	t.Parallel()
+
+	c, namespace := setup(ctx, t)
+	knativetest.CleanupOnInterrupt(func() { tearDown(ctx, t, c, namespace) }, t.Logf)
+	defer tearDown(ctx, t, c, namespace)
```
lbernick, line 609: Curious why both `cleanuponinterrupt` and `defer tearDown` are needed?

lbernick (commented):
thanks @JeromeJu for adding the conversion tests, they look great! @afrittoli could you take a look and see if they are what you had in mind? I think they were able to catch some issues with remote resolution conversion which is awesome.

Jerome, can you please revert the changes to the examples tests in your second commit? I think it's ok to have 2 commits in this PR, since they both need to be shipped together but have a logical split, but if possible the first one should just be the examples tests and the second one should just be the conversion tests, rather than mixing them.

@tektoncd/core-maintainers PTAL

test/conversion_test.go
```diff
@@ -0,0 +1,794 @@
...
+)
+
+var (
+	supportedV1ConversionFeatureGates = map[string]string{
+		"enable-api-fields":         "alpha",
+		"enable-tekton-oci-bundles": "true",
+	}
+
+	filterLabels        = cmpopts.IgnoreFields(metav1.ObjectMeta{}, "Labels")
+	filterAnnotations   = cmpopts.IgnoreFields(metav1.ObjectMeta{}, "Annotations")
```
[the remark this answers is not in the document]
JeromeJu: I am willing to discard that if there's any way I could circumvent the following release being generated:

```
± |v1-release → origin {5} U:1 ✗| → go test -v -tags=e2e -count=1 ./test -run ^TestTaskRunCRDConversion
=== RUN   TestTaskRunCRDConversion
=== PAUSE TestTaskRunCRDConversion
=== CONT  TestTaskRunCRDConversion
    init_test.go:138: Create namespace arendelle-ttvl6 to deploy to
    init_test.go:166: Verify SA "default" is created in namespace "arendelle-ttvl6"
    stream.go:254: E 15:55:34.467 tekton-pipelines-controller-8588544f45-jxvhz [github.com.tektoncd.pipeline.pkg.reconciler.taskrun.Reconciler] [arendelle-ttvl6/task-run-c-r-d-conversion-auxzmupl] TaskRun prepare error: resource request in progress
    conversion_test.go:692: -want, +got:   &v1beta1.TaskRun{
                ... // 1 ignored field
                ObjectMeta: v1.ObjectMeta{
                        ... // 5 ignored and 6 identical fields
        -               Annotations:     nil,
        +               Annotations:     map[string]string{"pipeline.tekton.dev/release": "c6dc3cd-dirty"},
                        OwnerReferences: nil,
                        Finalizers:      nil,
                        ... // 1 ignored field
                },
```

But if there is not I assume this should be fine as it is only ignored in the v1beta1 yaml expected, which is converted from the annotations of v1TaskRun yaml (in line #L346-348). So I think the annotations are not required in this specific test case of v1beta1 expected. And the expected v1 annotation case is not ignored in #L674.

test/conversion_test.go
JeromeJu: My take is `cleanuponinterrupt` only cleans up the namespace when the test fails https://github.com/tektoncd/pipeline/blob/main/vendor/knative.dev/pkg/test/cleanup.go#L57-L65 while defer tearDown works at the success of the test case? 

Also what's used in the test folders right now as in https://github.com/tektoncd/pipeline/blob/main/test/artifact_bucket_test.go#L61-L62 are deprecated. Created [issue](https://github.com/tektoncd/pipeline/issues/5728) to track down

test/conversion_test.go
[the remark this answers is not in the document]
lbernick: It sounds like you're saying we're testing conversion from annotations in v1 to resources in v1beta1, which is true, but we need to make sure conversion from v1beta1 resources -> v1 annotations is also covered. Maybe there is something in cmpopts that will allow you to implement a function that compares only the keys of interest, or an example of doing so in the codebase? Or maybe you can continue ignoring the annotations in the initial check, but add another check that does compare the annotations of interest?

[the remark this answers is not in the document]
lbernick: oh I see-- I think this is fine

XinruZhang (approved):

afrittoli:
Thanks, @JeromeJu!

If I understand correctly, the process that we are following is:
- (this PR/release) start serving the `v1` CRD
- (next PR/release) setting `v1` as the stored version, deprecate `v1beta`, define EOL for `v1beta`

Is this correct? If so we need to clarify what this means for users and include it in the release notes.

We could say something like `v1` is included in this release for users as a preview, for users to start preparing for the migration, and that `v1` will be the official version starting with the next release.

The gap release would give the rest of the Tekton projects a released `v1` client for integration (at least CLI, dashboard, triggers, chains and operator should be ready to work with `v1` before we do the next release) - so we need to make sure we sync with the release plan of other Tekton projects.

FYI @tektoncd/core-maintainers 

I'm restoring the hold to make sure we have enough eyes on this before it merges and also that we agree on the process.

/hold

afrittoli:
@JeromeJu with `v1` introduced, the feature flag gets a new valid value of `beta`, so we'll need a new CI job `pull-tekton-pipeline-beta-integration-tests` to test features that are marked as `beta` in the `v1` API. 
Since we still serve and test `v1beta`, the beta features will be on by default there and thus tested there, but we'll need to make sure we test them within the `v1` API too. I think it's probably ok to do that as a follow-up to this PR, but we'll need that CI job before we switch `v1` to being stored.

JeromeJu:
> 

Thanks Andrea and yes the steps you described are exactly those.

JeromeJu:
Much appreciate it with the clarifications. @dibyom and I also had this conversation earlier and agreed this should be in a separate PR as it would be easier to review and maintain and also because of that I have not added any beta folder in the examples. Created #5734 as a follow-up issue.

> @JeromeJu with `v1` introduced, the feature flag gets a new valid value of `beta`, so we'll need a new CI job `pull-tekton-pipeline-beta-integration-tests` to test features that are marked as `beta` in the `v1` API. Since we still serve and test `v1beta`, the beta features will be on by default there and thus tested there, but we'll need to make sure we test them within the `v1` API too. I think it's probably ok to do that as a follow-up to this PR, but we'll need that CI job before we switch `v1` to being stored.

dibyom:
@afrittoli Thanks for pointing out the release notes  - I think we should go beyond just release notes and send a note to tekton-dev and tekton-users. I think it is also important to mention that v1beta1 is not going away yet. 

We haven't decided on the EOL for v1beta1 I think but I think your plan makes sense - we should probably put that in an issue or a TEP for better visibility and to discuss it more broadly. I do think that we can wait on finalizing when the EOL for v1beta1 will be until we move the stored version to v1.

examples/v1/pipelineruns/pipelinerun-with-final-tasks.yaml
```diff
@@ -0,0 +1,203 @@
+# Copied from https://github.com/tektoncd/catalog/blob/v1/git/git-clone.yaml :(
```
afrittoli, line 1: Link in the comment should be https://github.com/tektoncd/catalog/blob/v1beta1/git/git-clone.yaml - at least until we have a `v1` tag on the catalog

examples/v1/pipelineruns/pipelinerun.yaml
```diff
@@ -0,0 +1,334 @@
...
+      value: /workspace/go
+    workingDir: $(workspaces.source.path)
+    script: |
+      # The intention behind this example Task is to run unit test, however we
+      # currently do nothing to ensure that a unit test issue doesn't cause this example
+      # to fail unnecessarily. In the future we could re-introduce the unit tests (since
+      # we are now pinning the version of Skaffold we pull) or use Tekton Pipelines unit tests.
+      echo "pass"
+---
+# Copied from https://github.com/tektoncd/catalog/blob/v1/git/git-clone.yaml
```
afrittoli, line 39: v1 -> v1beta1

examples/v1/pipelineruns/pipelinerun.yaml
```diff
@@ -0,0 +1,334 @@
...
+      RESULT_SHA="$(git rev-parse HEAD | tr -d '\n')"
+      EXIT_CODE="$?"
+      if [ "$EXIT_CODE" != 0 ]
+      then
+        exit $EXIT_CODE
+      fi
+      # Make sure we don't add a trailing newline to the result!
+      echo -n "$RESULT_SHA" > $(results.commit.path)
+---
+# Copied from https://github.com/tektoncd/catalog/blob/v1/kaniko/kaniko.yaml
```
afrittoli, line 122: v1 -> v1beta1

examples/v1/taskruns/no-ci/pull-private-image.yaml
```diff
@@ -0,0 +1,50 @@
+# Copyright 2019 The Tekton Authors
```
afrittoli, line 1: NIT: might want to use 2022

afrittoli:
I'm comparing the examples, and I would like to confirm the differences I see between beta1 and v1.
Most of the examples removed are specific to `PipelineResources` so that's fine.
A few I did not understand though:

Files moved in `v1`:
- pipeline run `git-resolver.yaml` moved under `no-ci` in `v1` 

Files removed in `v1` / pipelinerun:
- the entire `alpha` folder 
- `no-ci/pipeline-timeout.yaml` - I imagine this is related to the changes in timeout, but I would have expected it to move to the new syntax
- `clustertask-pipelinerun.yaml` -> deprecated. Perhaps it could be translated to the cluster resolver, but maybe we already have coverage for it
- `propagating_params_*`  

Files removed in `v1` / taskrun:
- the entire `alpha` folder 
- `propagating_params_implicit.yaml` 
-

JeromeJu:
Thank you @afrittoli . This is indeed helpful! This has also been put in discussion and I have opened https://github.com/tektoncd/pipeline/issues/5740 earlier to track down. And since that this PR has been opened for a long time, it is hard to keep synced with the latest v1beta1 folder. Also weighing that it could be even harder for reviewing, does it make sense here to have those examples synced in a follow up PR?

> I'm comparing the examples, and I would like to confirm the differences I see between beta1 and v1. Most of the examples removed are specific to `PipelineResources` so that's fine. A few I did not understand though:
> 
> Files moved in `v1`:
> 
> * pipeline run `git-resolver.yaml` moved under `no-ci` in `v1`
> 
> Files removed in `v1` / pipelinerun:
> 
> * the entire `alpha` folder
> * `no-ci/pipeline-timeout.yaml` - I imagine this is related to the changes in timeout, but I would have expected it to move to the new syntax
> * `clustertask-pipelinerun.yaml` -> deprecated. Perhaps it could be translated to the cluster resolver, but maybe we already have coverage for it
> * `propagating_params_*`
> 
> Files removed in `v1` / taskrun:
> 
> * the entire `alpha` folder
> * `propagating_params_implicit.yaml`

JeromeJu:
Added the migration doc https://github.com/tektoncd/community/pull/876 into the TEP according to the comment.
> Thanks, @JeromeJu!
> 
> If I understand correctly, the process that we are following is:
> 
> * (this PR/release) start serving the `v1` CRD
> * (next PR/release) setting `v1` as the stored version, deprecate `v1beta`, define EOL for `v1beta`
> 
> Is this correct? If so we need to clarify what this means for users and include it in the release notes.
> 
> We could say something like `v1` is included in this release for users as a preview, for users to start preparing for the migration, and that `v1` will be the official version starting with the next release.
> 
> The gap release would give the rest of the Tekton projects a released `v1` client for integration (at least CLI, dashboard, triggers, chains and operator should be ready to work with `v1` before we do the next release) - so we need to make sure we sync with the release plan of other Tekton projects.
> 
> FYI @tektoncd/core-maintainers
> 
> I'm restoring the hold to make sure we have enough eyes on this before it merges and also that we agree on the process.
> 
> /hold

examples/v1/taskruns/no-ci/tekton-bundles.yaml
```diff
@@ -0,0 +1,9 @@
+# TODO: Move the example image to a tekton owned repo.
+apiVersion: tekton.dev/v1
```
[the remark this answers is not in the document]
JeromeJu: Removed. Thanks @afrittoli

examples/v1/pipelineruns/no-ci/cluster-resolver.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: tekton.dev/v1
```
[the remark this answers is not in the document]
JeromeJu: This is copied from `v1beta1` examples, and I assume it is because that we have to create some tasks in the cluster ourselves to use this as for the `some-pipeline` in the `pipelineRef`.

JeromeJu:
As discussed in the pipeline WG today, this PR shall get merged in the current delayed v0.42 release once the [GA API policy and compatibility](https://github.com/tektoncd/pipeline/pull/5634) gets in. Therefore please take a look and chime in if there are any other comments.

FYI https://github.com/orgs/tektoncd/teams/core-maintainers

afrittoli:
Sorry to keep asking questions 🙏 
I don't remember if we finally agreed on a way forward for `PipelineRun` creating `Runs` vs `CustomRuns`.

XinruZhang:
> Sorry to keep asking questions 🙏 I don't remember if we finally agreed on a way forward for `PipelineRun` creating `Runs` vs `CustomRuns`.

Hi Andrea, thanks for asking! Sorry I didn't notice if you were in the [Pipeline WG](https://docs.google.com/document/d/1AfJfdyd1JN2P4haBdYOxEn6SMENMgvQF9ps7iF2QSG0/edit?resourcekey=0-lm66bE68HA_WVLA6AaCh6w#bookmark=id.ih19lyrsv965) this Tuesday where we thought people agreed with moving forward with supporting both versions that gated by a new feature flag `custom-task-version`.

lbernick:
We're not serving CustomRuns as part of this PR, so I think we can do it as a separate discussion (happening in https://github.com/tektoncd/community/pull/867). I don't think any of the proposed solutions we've discussed involve backwards incompatible changes to the 4 crds moved to v1 in this PR.

pritidesai:
@afrittoli @vdemeester @dibyom please take another look

pritidesai:
I haven't looked at the comments so far, apologies if this is already answered somewhere 🙏 

> Sets served in v1 CRDs to true.

Are we serving both `v1beta1` and `v1`?

afrittoli:
> I haven't looked at the comments so far, apologies if this is already answered somewhere 🙏
> 
> > Sets served in v1 CRDs to true.
> 
> Are we serving both `v1beta1` and `v1`?

Yes, that's the plan for when this merges. 
See https://github.com/tektoncd/pipeline/pull/5579#issuecomment-1309214799

afrittoli:
> > Sorry to keep asking questions 🙏 I don't remember if we finally agreed on a way forward for `PipelineRun` creating `Runs` vs `CustomRuns`.
> 
> Hi Andrea, thanks for asking! Sorry I didn't notice if you were in the [Pipeline WG](https://docs.google.com/document/d/1AfJfdyd1JN2P4haBdYOxEn6SMENMgvQF9ps7iF2QSG0/edit?resourcekey=0-lm66bE68HA_WVLA6AaCh6w#bookmark=id.ih19lyrsv965) this Tuesday where we thought people agreed with moving forward with supporting both versions that gated by a new feature flag `custom-task-version` based on discussions in [tektoncd/community#867](https://github.com/tektoncd/community/pull/867)

Yes, indeed, thank you

afrittoli:
> We're not serving CustomRuns as part of this PR, so I think we can do it as a separate discussion (happening in [tektoncd/community#867](https://github.com/tektoncd/community/pull/867)). I don't think any of the proposed solutions we've discussed involve backwards incompatible changes to the 4 crds moved to v1 in this PR.

ok, thank you.

pritidesai:
I do not see a separate check for `v1` or `v1beta1` integration tests. Does `pull-tekton-pipeline-integration-tests` cover both CRDs?

There are some conversion tests skipped, any reason for them being skipped?

https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5579/pull-tekton-pipeline-integration-tests/1591088955242057728

JeromeJu:
Thanks for the comment @pritidesai . Yes, currently `pull-tekton-pipeline-integration-tests` and `pull-tekton-pipeline-alpha-integration-tests` are covering the CRDs. And the reason why the tests for the two CRDs (specifically taskRun and pipelineRun) in `pull-tekton-pipeline-integration-tests` is that the deprecated fields need alpha-feature flag to be enabled and this is made on purpose to cover those fields to make sure compatibility of the migration.


> https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5579/pull-tekton-pipeline-integration-tests/1591088955242057728

test/init_test.go
```diff
@@ -101,11 +101,20 @@ func tearDown(ctx context.Context, t *testing.T, cs *clients, namespace string)
...
-			t.Errorf("Error getting TaskRun list %s", err)
+			t.Errorf("Error getting v1beta1TaskRun list %s", err)
 		}
-		for _, tr := range taskruns.Items {
+		for _, tr := range v1beta1Taskruns.Items {
+			if tr.Status.PodName != "" {
+				CollectPodLogs(ctx, cs, tr.Status.PodName, namespace, t.Logf)
+			}
+		}
+		v1Taskruns, err := cs.V1TaskRunClient.List(ctx, metav1.ListOptions{})
```
pritidesai, line 113: I would assume that this test file covers V1 Tasks, V1 Pipeline, V1 PipelineRun along with v1beta1 TaskRun 🤔 

https://github.com/tektoncd/pipeline/blob/main/test/init_test.go#L217

test/conversion_test.go
```diff
@@ -0,0 +1,794 @@
...
+    description: param des
+    type: string
+    default: "1"
+  workspaces:
+  - name: workspace
+    description: description
+    mountPath: /foo
+    readOnly: true
+    optional: true
+  resources:
```
pritidesai, line 140: I am confused seeing `resources` here 🙃

test/conversion_test.go
```diff
@@ -0,0 +1,794 @@
...
+        description: "The output is a Docker image"
+        optional: true
+`
+
+	v1TaskYaml = `
+metadata:
+  name: %s
+  namespace: %s
+  annotations: {
+    tekton.dev/v1beta1Resources: '{"inputs":[{"name":"git-repo","type":"git","description":"The input is code from a git repository","optional":true}],"outputs":[{"name":"optionalimage","type":"image","description":"The output is a Docker image","optional":true}]}'
```
pritidesai, line 158: beta resources?

test/conversion_test.go
JeromeJu: `resources` is deprecated in v1beta1, so this needs to be covered by the conversion integration test to make sure it does not get dropped on the floor. 

For this specific case, on line #L612 it is being created by v1beta1TaskClient and converted to v1(which serializing the resources into annotations, which would correspond to the following comment at line #L158).

Please let me know if this makes any sense:)

test/init_test.go
JeromeJu: Updated. Thanks Priti for the catch!

test/conversion_test.go
JeromeJu: as in https://github.com/tektoncd/pipeline/pull/5579#discussion_r1028441687

test/conversion_test.go
pritidesai: Thank you @JeromeJu 

Yup, I understand `resources` are deprecated and also, `resources` were not promoted to `beta`.

We also want to make sure `resources` are surfaced when we create V1 as part of the proposed migration [strategy](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md#upgrade--migration-strategy).

I understand they are being stored in `annotations` but aren't `annotations` part of the same API policies.

The only concern I have here is, when the users will notice pipeline resources in annotations as part of V1 CRDs, it might create a false expectation that the resources will be part of the V1 and not getting dropped anytime soon. I have been a huge fan of `resources` and would love that 🤣 

The pipeline resources were announced as deprecated in  `v0.30.0` and safe to remove since 12/20//2021.

https://github.com/tektoncd/pipeline/blob/main/docs/deprecations.md

From the policy perspective, I think, it's safe to delete the resources from the code base before we release V1 CRDs. Or at least drop them on the floor while converting to V1 but like you pointed out, it could lead to an incomplete/invalid CRD.

JeromeJu: I do understand the concerns here. So that to suggest that we need to make sure this is not misleading in the context of conversion_test.go. And I do agree that we would still need them in the examples here to be aligned with
https://github.com/tektoncd/pipeline/issues/4546. Therefore, would that makes sense to add the comments here in/ above the yamls where `resources` are referenced?

JeromeJu: Also wondering if this is out of scope for the current v1-CRD-release PR. The way dealing with the deprecated fields are discussed and addressed at https://github.com/tektoncd/pipeline/issues/4546. While I think in this PR what's added is the coverage for integration test of conversion, which shall not be the false expectation as this serves as a part of the [integration test ](https://github.com/tektoncd/pipeline/issues/5627).

pritidesai: @jerop trying to understand, why not dropping `resources` on the floor while converting `beta` pipelines to `v1` instead adding them into `annotations`?

https://github.com/tektoncd/pipeline/pull/5331

FYI, no intention to block this as long as we comply with the policies.

My understanding is, the `annotations` approach is more applicable for the fields which are present in the higher versions e.g. `beta` but not in the lower versions e.g. `alpha`. Like `finally` and `when` expressions are introduced in `beta` CRDs but did not exist in `alpha` CRDs. These fields were serialized in `alpha` annotations to get the valid objects while converting up.

vdemeester: @pritidesai I think one of the main intention (long-term) is to not lost *any* input from the user. 

In addition, we will support `v1beta1` and `v1` for a little while. As we do, we *will* support resource on `v1beta1` types, even when we switch the storage version to `v1`. If a user creates a `v1beta1` resource with a `PipelineResource`, it should work, as long as we support and serve `v1beta1` (or at least as long as the deprecation notice / 9 months thingy is). This means we don't really have any choice of dropping it on the floor if we want to keep our API policy promise.

JeromeJu: Thanks Vincent for chiming in and thanks Priti for the clarification question.

I think for the question it might be divided to 2 parts: 
- "whether we can drop `resources` on the floor":
   since `resources` is still in v1beta1 fields (though it has passed the time as you pointed in the deprecation.md to be removed), we might still have to support it.
- why going the `annotation` way
   I think the purpose for serializing into the `annotation` as in v1 there is no `resources` field and we do not want to apply it so that it can be converted back into v1beta1 before v1 gets to the stored version.

vdemeester: @pritidesai we could do that but what about the next field that we deprecate ? The problem will be similar. 
Serializing into an annotation doesn't necessarily we support it "always". But it means we "keep" what the user sent us. If tomorrow we remove the `resources` field, then we would stop serializing, and for things where it was serialized it wouldn't be taken into account. As of today, someone who provide a `resources` in a `v1beta1` object has something working, if we release `v1` tomorrow and still serve `v1beta1`, it should still work for that user using `v1beta1` — there is nothing really more to it. The day we remove it to then we stop supporting it from `v1beta1`. In any case, today, a user using a `v1` wouldn't be able to user `resources` so it's relatively clear we are not supporting it.

lbernick: @JeromeJu and I have been looking through previous discussions, and although we discussed this quite a bit, it's actually not clear that we ever came to a decision. These alternatives [were proposed in API WG](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit?disco=AAAAkU3A2Kg); [TEP-0105 to remove v1alpha1](https://github.com/tektoncd/community/blob/main/teps/0105-remove-pipeline-v1alpha1-api.md#proposal) doesn't set a proposal for removing PipelineResources, and [TEP-0074](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md#proposal) states that we can remove pipelineresources after 9 months, but not necessarily that we will. 

I believe we chose this approach after contributors expressed concerns about making sure that a v1 release doesn't include backwards incompatible changes to v1beta1, and I opened https://github.com/tektoncd/pipeline/issues/4546 for discussion. I didn't realize that this issue had not actually been decided; I'm sorry about that.

I agree with Vincent that the fact that you can't create a v1 CRD with resources makes it clear that we won't continue to support them, and prefer to keep the current approach. Would it be helpful for me to create an issue for removing PipelineResources completely, and we can target that for the next release?

JeromeJu: Thanks Priti, Vincent and Lee, added https://github.com/tektoncd/pipeline/issues/5785#issue-1460683293 for further discussion. Also added to the WG for next week.

test/conversion_test.go
```diff
@@ -0,0 +1,787 @@
...
+// executed by the webhook. And then it creates the v1 TaskRun CRD using v1Clients
+// and requests it by v1beta1Clients to compare with v1beta1.
+func TestTaskRunCRDConversion(t *testing.T) {
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	t.Parallel()
+
+	c, namespace := setup(ctx, t, requireAnyGate(supportedV1ConversionFeatureGates))
```
pritidesai, line 650: Any reason for including `alpha` feature gate for taskRun and pipelineRun conversion here? Do we have any alpha API in the test case? I suggest keeping alpha APIs as a separate test case if that's not the case.

JeromeJu: Agreed with the separation for better delegations.
This is another deprecated field `bundle` that needs git resolver bundle here. The `bundle` would requires the `alpha` feature flag here.

pritidesai:
Is this expected or may be just my local instance misbehaving? 🤔 Tekton controller logs has this run as completed with the duration.  Also, noticing the same with v1beta1 resources 😞 

```
$ kubectl create -f examples/v1/pipelineruns/ignore-step-error.yaml

$ kubectl get all                                                  
NAME                                                        READY   STATUS      RESTARTS   AGE
pod/pipelinerun-with-failing-step-and-ws-bvqp6-reader-pod   0/1     Completed   0          94s
pod/pipelinerun-with-failing-step-and-ws-bvqp6-writer-pod   0/1     Completed   0          2m35s
pod/pipelinerun-with-failing-step-l76mw-task1-pod           0/2     Completed   0          2m35s
pod/pipelinerun-with-failing-step-l76mw-task2-pod           0/3     Completed   0          91s

$ kubectl get pr 
NAME                                         SUCCEEDED   REASON   STARTTIME   COMPLETIONTIME
pipelinerun-with-failing-step-and-ws-bvqp6                                    
pipelinerun-with-failing-step-l76mw

$ tkn pr describe
tkn pr describe pipelinerun-with-failing-step-l76mw
Name:        pipelinerun-with-failing-step-l76mw
Namespace:   default
Labels:
 tekton.dev/pipeline=pipelinerun-with-failing-step-l76mw

🌡️  Status

STARTED   DURATION   STATUS
---       ---        ---

⚓ Params

 NAME         VALUE
 ∙ CONTINUE   continue
```

vdemeester:
@pritidesai maybe your tkn version is a bit old ? Also I am not entirely sure how tkn act with v1 types today (it probably just queries v1beta1)

JeromeJu:
/retest
codebase not updated

JeromeJu:
Thanks for the answer Vincent, can I confirm what are the exact reproducible steps by `act with v1 types` I should try with?

> @pritidesai maybe your tkn version is a bit old ? Also I am not entirely sure how tkn act with v1 types today (it probably just queries v1beta1)

afrittoli:
I think this looks good from my perspective - @vdemeester @pritidesai is there any concern left? 
We might want to include in the release notes that `v1` is served but CLI, Dashboard, Chains etc support won't be available until the following release.

@JeromeJu you mentioned you have an issue tracking the work needed after this is merged and before we can switch the stored version to v1. We should include in there one issue per Tekton project to introduce v1 support so we can easily track that work.

JeromeJu:
Thanks @afrittoli , currently drafting the issues.

> I think this looks good from my perspective - @vdemeester @pritidesai is there any concern left? We might want to include in the release notes that `v1` is served but CLI, Dashboard, Chains etc support won't be available until the following release.
> 
> @JeromeJu you mentioned you have an issue tracking the work needed after this is merged and before we can switch the stored version to v1. We should include in there one issue per Tekton project to introduce v1 support so we can easily track that work.

pritidesai:
thank you @JeromeJu @afrittoli @vdemeester 

I am not much worried about `cli` with `v1` CRDs at this point but two concerns I have:

- CLI should still work for `v1beta1` CRDs
- `kubectl` describe (Kuberentes APIs ) must show the status including reason, start time, completion time, etc for both CRDs.

It could be my local cluster misbehaving. I have not got chance to look into Tekton CI tests. Yes CLI must support both versions eventually but not a blocker at this point.

We have pipelines `0.42` created yesterday. Please feel free to merge this given that we have enough time until the next release to fix anything if at all broken.

JeromeJu:
After testing against tkn with `served` switching back from true for v1 (where tkn does not work) to false, the STATUS etc was back to normal again. So agreed with @afrittoli this should probably fixed after v1 is a supported version after https://github.com/tektoncd/cli/issues/1804.

> thank you @JeromeJu @afrittoli @vdemeester
> 
> I am not much worried about `cli` with `v1` CRDs at this point but two concerns I have:
> 
> * CLI should still work for `v1beta1` CRDs
> * `kubectl` describe (Kuberentes APIs ) must show the status including reason, start time, completion time, etc for both CRDs.
> 
> It could be my local cluster misbehaving. I have not got chance to look into Tekton CI tests. Yes CLI must support both versions eventually but not a blocker at this point.
> 
> We have pipelines `0.42` created yesterday. Please feel free to merge this given that we have enough time until the next release to fix anything if at all broken.

afrittoli (commented):
Thank you @JeromeJu for your work on this and all the updates over time!!

When you have a link to the issue that includes all the required follow-ups, could you share it on this PR?

/approve

@vdemeester ok to merge?

JeromeJu:
Yes, thanks @afrittoli. 

- https://github.com/tektoncd/pipeline/issues/5740

> Thank you @JeromeJu for your work on this and all the updates over time!!
> 
> When you have a link to the issue that includes all the required follow-ups, could you share it on this PR?
> 
> /approve
> 
> @vdemeester ok to merge?

JeromeJu:
I think what @pritidesai brought up is also needs some more eyes. Yesterday I also saw this at my local so maybe still need to fix that?
/hold

vdemeester:
Hum, indeed there is something wrong with the CRDs as `kubectl get …` doesn't show the status

vdemeester:
Yes, the conversion doesn't work correctly.

```
~/s/t/cli main λ k get pipelineruns.v1.tekton.dev/write-and-cleanup-workspace
NAME                          SUCCEEDED   REASON   STARTTIME   COMPLETIONTIME
write-and-cleanup-workspace                                    

~/s/t/cli main λ k get pipelineruns.v1beta1.tekton.dev/write-and-cleanup-workspace
NAME                          SUCCEEDED   REASON      STARTTIME   COMPLETIONTIME
write-and-cleanup-workspace   True        Succeeded   9m20s       6m59s
```

This also might have an impact on `tkn`.

vdemeester:
Yeah, we are not handling `status` at all in the conversion.
See https://github.com/tektoncd/pipeline/pull/5005/files#diff-9085e8f15471a04a7656a958b04fa6568f97ff8b4e3596cf849e3d0ab293eca8L37 on `v1alpha1, we need to copy/handle the status, otherwise, we are loosing some informations 👼🏼

test/conversion_test.go
```diff
@@ -0,0 +1,978 @@
...
+	if d := cmp.Diff(v1beta1TaskExpected, v1beta1TaskGot, filterTypeMeta, filterObjectMeta); d != "" {
+		t.Fatalf("-want, +got: %v", d)
+	}
+}
+
+// TestTaskRunCRDConversion first creates a v1beta1 TaskRun CRD using v1beta1Clients
+// and requests it by v1Clients to compare with v1 if the conversion has been correctly
+// executed by the webhook. And then it creates the v1 TaskRun CRD using v1Clients
+// and requests it by v1beta1Clients to compare with v1beta1.
+func TestTaskRunCRDConversion(t *testing.T) {
```
XinruZhang, line 738: Sorry for the very late check, leaving this comment to track: we need to update `TestTaskRunCRDConversion`, `TestTaskRunCRDConversionAlpha`, `TestPipelineRunCRDConversion` and `TestPipelineRunCRDConversionAlpha`:
1. Add status to all **expected** YAMLs of runtime objects -- PipelineRun and TaskRun.
2. Create `v1beta1` Object, wait until it finishes (can use [updated](https://github.com/tektoncd/pipeline/issues/5793) version of [WaitForPipelineRunState](https://github.com/tektoncd/pipeline/blob/b3e4f35712fca0235e3ef80cc3f03b2600f40cb1/test/wait.go#L159) and [WaitForTaskRunState](https://github.com/tektoncd/pipeline/blob/b3e4f35712fca0235e3ef80cc3f03b2600f40cb1/test/wait.go#L85)), compare **finished** TaskRuns and PipelineRuns (of both `v1` and `v1beta1` versions) with **expected** objects.
3. [optional but highly recommended] create `v1` Objects, wait until it finishes and compare.

JeromeJu: Thanks @XinruZhang , this is very helpful.
Added to https://github.com/tektoncd/pipeline/pull/5579#issuecomment-1326617948.

JeromeJu:
Currently ready for review 🙏PTAL🙏:
- https://github.com/tektoncd/pipeline/pull/5799
- https://github.com/tektoncd/pipeline/pull/5797

bendory:
This PR is good evidence that we need a better way to decouple versioning across repositories and components -- for example, ensuring cross-compatibility across CLI and API such that the CLI does not need to be updated concurrently with the API version.
🤔🤔🤔

JeromeJu:
/retest
Codebase not updated in Prow for the fix.

test/conversion_test.go
```diff
@@ -0,0 +1,1031 @@
...
+		t.Fatalf("-want, +got: %v", d)
+	}
+}
+
+// TestPipelineRunCRDConversion first creates a v1beta1 PipelineRun CRD using v1beta1Clients and
+// requests it by v1Clients to compare with v1 if the conversion has been
+// correctly executed by the webhook. And then it creates the v1 PipelineRun CRD using v1Clients
+// and requests it by v1beta1Clients to compare with v1beta1.
+func TestPipelineRunCRDConversion(t *testing.T) {
+	ctx := withFullEmbeddedStatus(context.Background())
```
JeromeJu, line 946: alpha-test failing due to `embedded-status` set as `minimal`
Currently investigating why the withFullEmbeddedStatus here does not take effect.
https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5579/pull-tekton-pipeline-alpha-integration-tests/1600939404950507520

test/parse/yaml.go
```diff
@@ -34,6 +35,16 @@ kind: TaskRun
 	return &tr
 }
 
+// MustParseV1TaskRun takes YAML and parses it into a *v1.TaskRun
```
dibyom, line 38: Once we deprecate v1beta1 I think we should just modify MustParseTaskRun to use v1 by default. (Again, not necessary for this PR, but we should track this)

examples/v1/taskruns/no-ci/pull-private-image.yaml
```diff
@@ -0,0 +1,50 @@
...
...UVUwNVpWME5hV0hoRWNGTklTVEU0WkZ4dWFrdHJiMEZpZHpKTmIzbDNVVWxzYTJWMVFXNHhaRmhHWVdReGVuTllVVWRrVkhKdFdIbDJOMDVRVUNzNFIxaENhMjVDVEdrelkzWjRWR2xzU2tsVGVWeHVkV05PY2tOTmFYRk9RVk51TDJSeE4yTlhSRVpWUVVKbmFsZ3hOa3BJTWtST1Jsb3ZiQzlWVmtZelRrUkJTbXBZUTNNeFdEZDVTVXA1V0VJMmIzZ3ZlbHh1WVZOc2NXeEpUVll6TlVSQ1JEZHhlRko1ZFV0eloyczlYRzR0TFMwdExVVk9SQ0JRVWtsV1FWUkZJRXRGV1MwdExTMHRYRzRpTEFvZ0lDSmpiR2xsYm5SZlpXMWhhV3dpT2lBaWNIVnNiQzF6WldOeVpYUXRkR1Z6ZEdsdVowQmlkV2xzWkMxamNtUXRkR1Z6ZEdsdVp5NXBZVzB1WjNObGNuWnBZMlZoWTJOdmRXNTBMbU52YlNJc0NpQWdJbU5zYVdWdWRGOXBaQ0k2SUNJeE1EYzVNelU0TmpJd016TXdNalV5TlRFek5USWlMQW9nSUNKaGRYUm9YM1Z5YVNJNklDSm9kSFJ3Y3pvdkwyRmpZMjkxYm5SekxtZHZiMmRzWlM1amIyMHZieTl2WVhWMGFESXZZWFYwYUNJc0NpQWdJblJ2YTJWdVgzVnlhU0k2SUNKb2RIUndjem92TDJGalkyOTFiblJ6TG1kdmIyZHNaUzVqYjIwdmJ5OXZZWFYwYURJdmRHOXJaVzRpTEFvZ0lDSmhkWFJvWDNCeWIzWnBaR1Z5WDNnMU1EbGZZMlZ5ZEY5MWNtd2lPaUFpYUhSMGNITTZMeTkzZDNjdVoyOXZaMnhsWVhCcGN5NWpiMjB2YjJGMWRHZ3lMM1l4TDJObGNuUnpJaXdLSUNBaVkyeHBaVzUwWDNnMU1EbGZZMlZ5ZEY5MWNtd2lPaUFpYUhSMGNITTZMeTkzZDNjdVoyOXZaMnhsWVhCcGN5NWpiMjB2Y205aWIzUXZkakV2YldWMFlXUmhkR0V2ZURVd09TOXdkV3hzTFhObFkzSmxkQzEwWlhOMGFXNW5KVFF3WW5WcGJHUXRZM0prTFhSbGMzUnBibWN1YVdGdExtZHpaWEoyYVdObFlXTmpiM1Z1ZEM1amIyMGlDbjA9In19
```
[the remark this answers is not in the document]
JeromeJu: Thanks, added https://github.com/tektoncd/pipeline/issues/5867 and linked it in the comment.

dibyom:
/lgtm

I think the last outstanding issue was around errors while handling status conversions - I verified that the both v1beta1 and v1 example pipelineruns run successfully and the status is displayed when using kubectl

PR #5597, opened 2022-10-04
Merge Labels and Annotations instead of override

vdemeester:
# Changes

When the `PipelineRun` controller updates labels, it does override whatever it fetched using the client. This means, if the labels where updated during the reconciliation, the controller will override them (possibly loosing some information). This can happen on "not so quick" reconciliation loop, for example, when fetching bundles.

This is also true for `TaskRun`, and this fixes it for both.

/kind bug

This affects all version of Tekton since 2020 🙃. Because this is a race, this is *very* hard to actually test 🙃.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
The `PipelineRun` and `TaskRun` controller will not override label set by other tools during the reconciler loop, and will merge them instead
```


imjasonh (commented):
/lgtm

abayer (approved):

PR #5599, opened 2022-10-04
Propagated Parameters e2e tests

chitrangpatel:
# Changes

Prior to this, we only had examples for propagated parameters. This PR adds e2e tests for testing this feature instead of simply relying on the examples.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
e2e tests added for propagated parameters.
```


test/propagated_params_test.go
```diff
@@ -0,0 +1,161 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2019 The Tekton Authors
```
JeromeJu, line 5: nit: year to be updated

test/propagated_params_test.go
```diff
@@ -0,0 +1,161 @@
...
+		pipelineRunFunc: getPropagatedParamWithScopePipelineRun,
+	}, {
+		name:            "propagated parameters with default inner scope",
+		pipelineName:    "propagated-parameters-default-inner-scope",
+		taskName:        "echo-hello",
+		pipelineRunFunc: getPropagatedParamWithDefaultScopePipelineRun,
+	}}
+
+	for _, td := range tds {
+		td := td // capture range variable
```
JeromeJu, line 60: sorry can you elaborate what this is for?

test/propagated_params_test.go
chitrangpatel: 😁 Thats what happens when you copy a file and start replacing its contents. Will fix

chitrangpatel: done

test/propagated_params_test.go
chitrangpatel: Honestly, it isnt required. I was probably required in the test that I copied and edited. I cleaned it up to avoid confusion.

dibyom: I think we do need that actually - see https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables and https://internals.tidb.io/t/topic/391

chitrangpatel: Oh I see. And because we're running the tests in parallel, we need it.

test/propagated_params_test.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+
+			t.Logf("Waiting for PipelineRun %s in namespace %s to complete", prName, namespace)
+			if err := WaitForPipelineRunState(ctx, c, prName, timeout, PipelineRunSucceed(prName), "PipelineRunSuccess"); err != nil {
+				t.Fatalf("Error waiting for PipelineRun %s to finish: %s", prName, err)
+			}
+			t.Log("Checking if parameter replacements have been updated in the spec.")
+			cl, _ := c.PipelineRunClient.Get(ctx, prName, metav1.GetOptions{})
+			if cl.Status.PipelineSpec.Tasks[0].TaskSpec.Steps[0].Script != "echo Hello World!" {
+				t.Fatalf(`Expected replaced parameter value %s but found %s`, "echo Hello World!", cl.Status.PipelineSpec.Tasks[0].TaskSpec.Steps[0].Script)
```
jerop, line 90: could we check the resolved pipelinerun vs expected pipelinerun instead of the script only? this will help us ensure that the spec was not modified (while ignoring the fields that can't be pre-determined like start time)

test/propagated_params_test.go
```diff
@@ -0,0 +1,334 @@
...
+  - name: HELLO
+    value: "Hello World!"
+  pipelineSpec:
+    tasks:
+      - name: %s
+        taskSpec:
+          steps:
+            - name: echo
+              image: ubuntu
+              script: echo $(params.HELLO)
```
jerop, line 140: please add a finally task with propagation

chitrangpatel: done

jerop (approved):
thank you @chitrangpatel!

test/propagated_params_test.go
```diff
@@ -0,0 +1,382 @@
...
+        taskSpec:
+          steps:
+            - name: echo
+              image: ubuntu
+              script: echo Hello World!
+`, namespace))
+	return pipelineRun, expectedPipelineRun
+}
+
+func getPropagatedParamWithScopePipelineRun(t *testing.T, namespace string) (*v1beta1.PipelineRun, *v1beta1.PipelineRun) {
```
dibyom, line 234: What does Scope mean here?

test/propagated_params_test.go
```diff
@@ -0,0 +1,382 @@
...
+			embeddedStatus := GetEmbeddedStatus(ctx, t, c.KubeClient)
+			knativetest.CleanupOnInterrupt(func() { tearDown(ctx, t, c, namespace) }, t.Logf)
+			defer tearDown(ctx, t, c, namespace)
+
+			t.Logf("Setting up test resources for %q test in namespace %s", td.name, namespace)
+			pipelineRun, expectedResolvedPipelineRun := td.pipelineRunFunc(t, namespace)
+			if embeddedStatus == config.MinimalEmbeddedStatus {
+				expectedResolvedPipelineRun.Status.ChildReferences = []v1beta1.ChildStatusReference{}
+				for k, v := range expectedResolvedPipelineRun.Status.TaskRuns {
+					k := k
```
dibyom, line 99: I don't think this is necessary since this is running in the same goroutine unlike the above for loop where each t.Run is running in parallel

test/propagated_params_test.go
chitrangpatel: The scope here means that if we have a parameter provided in PipelineRun and pipelineTask, use the value from pipelineTask.

test/propagated_params_test.go
chitrangpatel: Hmm. I did try it without that and ran into errors occasionally when I ran it multiple times. I thought this might have had something to do with it. I can retry a few more times to validate that.

chitrangpatel: You are right. I get this error even after setting `k := k, ...` because the order of the taskruns (the the case where I have a regular and a finally task changes) and so when I compare, there is a difference. I will fix it. Thanks!

chitrangpatel: fixed this.

test/propagated_params_test.go
```diff
@@ -0,0 +1,462 @@
...
+   steps:
+     - name: echo
+       container: step-echo
+   taskSpec:
+     steps:
+       - name: echo
+         image: ubuntu
+         script: echo Hello World!
+`, namespace))
+	finallyRun := parse.MustParseTaskRun(t, fmt.Sprintf(`
```
jerop, line 231:
```suggestion
	finallyTaskRun := parse.MustParseTaskRun(t, fmt.Sprintf(`
```

test/propagated_params_test.go
```diff
@@ -0,0 +1,462 @@
...
+   taskSpec:
+     steps:
+       - name: echo
+         image: ubuntu
+         script: echo Hello World!
+`, namespace))
+	return pipelineRun, expectedPipelineRun, []*v1beta1.TaskRun{taskRun, finallyRun}
+}
+
+func getPropagatedParamWithScopePipelineRun(t *testing.T, namespace string) (*v1beta1.PipelineRun, *v1beta1.PipelineRun, []*v1beta1.TaskRun) {
```
jerop, line 257:
```suggestion
func getPropagatedParamTaskLevelPipelineRun(t *testing.T, namespace string) (*v1beta1.PipelineRun, *v1beta1.PipelineRun, []*v1beta1.TaskRun) {
```

test/propagated_params_test.go
```diff
@@ -0,0 +1,462 @@
...
+      status:
+        podName: propagated-parameters-inner-scope-echo-hello-pod
+        steps:
+          - name: echo
+            container: step-echo
+        taskSpec:
+          steps:
+            - name: echo
+              image: ubuntu
+              script: echo Hello World!
```
jerop, line 312: this should be removed 🤔

test/propagated_params_test.go
```diff
@@ -0,0 +1,462 @@
...
+            container: step-echo
+        taskSpec:
+          params:
+            - name: HELLO
+              type: string
+              default: "Default Hello World"
+          steps:
+            - name: echo
+              image: ubuntu
+              script: echo Hello World!
```
jerop, line 415: this should be removed 🤔

jerop (approved):

test/propagated_params_test.go
chitrangpatel: Yes.

test/propagated_params_test.go
chitrangpatel: Yes.

test/propagated_params_test.go
chitrangpatel: sounds good!

Issue #5262, opened 2022-08-03
CSI Volumes to beta


JeromeJu:
### Feature request

For the latest release, CSI Volumes is in Alpha in v0.38. It shall be promoted to beta.

PR #5628, opened 2022-10-11
CSI workspace to Beta

JeromeJu:
# Changes
This commit removes the alpha feature gate for the csi workspace so that it becomes a beta feature.

Fixes: #5262 
/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
CSI workspaces are promoted to beta/stable API
```


pkg/apis/pipeline/v1/workspace_validation.go
```diff
@@ -73,16 +71,9 @@ func (b *WorkspaceBinding) Validate(ctx context.Context) (errs *apis.FieldError)
...
 	// For a CSI to work, you must provide and have installed the driver to use.
-	if b.CSI != nil {
-		errs := version.ValidateEnabledAPIFields(ctx, "csi workspace type", config.AlphaAPIFields).ViaField("workspaces")
-		if errs != nil {
-			return errs
-		}
-		if b.CSI.Driver == "" {
-			return apis.ErrMissingField("csi.driver")
-		}
+	if b.CSI != nil && b.CSI.Driver == "" {
```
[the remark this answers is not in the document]
dibyom: And once we do that, we should also add it to the list of beta features:https://github.com/tektoncd/pipeline/blob/main/docs/install.md#beta-features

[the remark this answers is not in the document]
JeromeJu: Thanks Dibyo!

docs/install.md
```diff
@@ -490,6 +489,12 @@ the `feature-flags` ConfigMap alongside your Tekton Pipelines deployment via
 
 For beta versions of Tekton CRDs, setting `enable-api-fields` to "beta" is the same as setting it to "stable".
 
+Features currently in "beta" are:
+
+| Feature                                                                                               | TEP                                                                                                                        | Release                                                              | Individual Flag             |
+|:------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------|:----------------------------|
+| [CSI Workspace Type](workspaces.md#csi)                                                               | N/A                                                                                                                        | [v0.38.0](https://github.com/tektoncd/pipeline/releases/tag/v0.38.0) |                             |
```
dibyom, line 496: This should be updated to v0.41 right?
Assuming this makes it to 0.41?

docs/workspaces.md
```diff
@@ -534,7 +534,7 @@ workspaces:
 
 ##### `csi`
 
-This is an alpha feature. The `enable-api-fields` feature flag [must be set to `"alpha"`](./install.md)
+This is a beta feature. The `enable-api-fields` feature flag [must be set to `"beta"`](./install.md)
```
dibyom, line 537: So, I think we only need to set it to "beta" for the v1 API version, not the v1beta1 api version

pkg/apis/pipeline/v1beta1/workspace_validation.go
```diff
@@ -73,10 +73,10 @@ func (b *WorkspaceBinding) Validate(ctx context.Context) (errs *apis.FieldError)
 		return apis.ErrMissingField("projected.sources")
 	}
 
-	// The csi workspace is only supported when the alpha feature gate is enabled.
+	// The csi workspace is only supported when the beta feature gate is enabled.
 	// For a CSI to work, you must provide and have installed the driver to use.
 	if b.CSI != nil {
-		errs := version.ValidateEnabledAPIFields(ctx, "csi workspace type", config.AlphaAPIFields).ViaField("workspaces")
+		errs := version.ValidateEnabledAPIFields(ctx, "csi workspace type", config.BetaAPIFields).ViaField("workspaces")
```
dibyom, line 79: for v1beta1 apiVersion, there is no beta i.e. beta fields are "stable" per TEP-033: 
https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md

Does this mean we need to set this to stable instead of v1beta1? Basically how can we ensure this field is enabled by default in v1beta1 but not in v1? (@pritidesai @vdemeester @afrittoli @jerop @lbernick)

lbernick: we can just remove this validation check in v1beta1, no? alternatively we can modify [ValidateEnableAPIFields](https://github.com/tektoncd/pipeline/blob/933e4f667c19eaf0a18a19557f434dbabe20d063/pkg/apis/version/version_validation.go#L30) to take a crd version as well but that's a bit awkward

JeromeJu: Would like to confirm, by 'remove this validation check in v1beta1' means we only keep validate 'alpha' features gates?

And for the alternative solution, would it require a conversion to be implemented for feature flags of specific features?

lbernick: I just meant this line of code can be removed in v1beta1. The alternative solution shouldn't involve feature flags related to specific features.

JeromeJu: > For example, with `enable-api-field: stable`, one could use resolvers using the `v1beta1` API but **not** the `v1` API ? This might be slightly (or very) confusing for our users.

Thanks Vincent, can we interpret this as  we need to assure that users do understand "beta is the equivalent of stable so that "beta" features are enabled by default in v1"? If so I think this is covered in #5634 [#5634](https://github.com/tektoncd/pipeline/pull/5634/files) as:

```
The `apiVersion` field in a Tekton CRD determines whether the overall API (and its default behaviors) is considered to be in `alpha`, `beta`, or `GA`. Within a stable CRD, certain opt-in features or API fields gated may be considered `alpha` or `beta`. Similarly, within a beta CRD, certain opt-in features may be considered `alpha`. See the section on Feature Gates for details.
```

vdemeester: Ok so even in v1beta1 we would need `enable-api-fields: beta` to use CSI workspace and project volume, am I right ?
@afrittoli @JeromeJu @lbernick

JeromeJu: @vdemeester sorry for the confusion. I think the line in this comment shall be already removed. So no `enable-api-fields: beta` needed as `beta` are the defaulted 'stable' in v1beta1.

vdemeester: > @vdemeester sorry for the confusion. I think the line in this comment shall be already removed. So no `enable-api-fields: beta` needed as `beta` are the defaulted 'stable' in v1beta1.

It confuses me even more 😅. 

What is proposed here is (with the example of CSI workspace):
- To use CSI workspace:
  - in `v1` the user need to set `enable-api-fields: beta`
  - in `v1beta1` the user doesn't need to set `enable-api-fields: beta` (aka it would also work with `enable-api-fields: stable`)

am I right ?

JeromeJu: Yes totally!

vdemeester: Just to make sure, pinging @tektoncd/core-maintainers, does that behavior make sense to every one ?
Is it inlined with https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#promotion-to-beta-and-beyond ?

dibyom: I do find it a bit confusing but I think that is in line with the feature gates TEP

lbernick: agreed with dibyo-- it's confusing and I'd be open to changing the TEP but I think this is in line with what the TEP states.

docs/workspaces.md
```diff
@@ -534,8 +534,8 @@ workspaces:
 
 ##### `csi`
 
-This is an alpha feature. The `enable-api-fields` feature flag [must be set to `"alpha"`](./install.md)
-for csi volume source to function.
+This is a beta feature in v1. The `enable-api-fields` feature flag [must be set to `"beta"`](./install.md)
+for csi volume source to function in v1.
```
jerop, line 538: could it be confusing to have this in docs when we haven't released v1 yet?

jerop (approved):

docs/workspaces.md
JeromeJu: Makes sense, thanks @jerop 
I will delete this for now. 
Also noted this in https://github.com/tektoncd/pipeline/issues/5663

vdemeester (approved):

jerop (commented):
/lgtm

Issue #5633, opened 2022-10-11
Update API compatibility policy for V1/GA


dibyom:
### Feature request

Our current policy describes alpha and beta. We need to update it for GA.

### Use case

1. As a Tekton user, I'd like to know what API guarantees Tekton V1 provides
2. As a Tekton developer, I'd like to know how to experiment with new features post the V1 release.

We had some discussion on this in the [API WG](https://docs.google.com/document/d/17PodAxG8hV351fBhSu7Y_OIPhGTVgj6OJ2lPphYYRpU/edit#heading=h.d3j2qywgio47) and I also wrote up a [doc based on my reading of the Kubernetes V1 policies](https://docs.google.com/document/d/15vD5vbK5UKCDFMoOqOwEc_NYACthtCZJlTJiiEEoljY/edit#)

(join https://groups.google.com/forum/#!forum/tekton-dev for access)

This issue is only for tracking changes/updates to our policy. Development guidelines for post v1 dev work is tracked in https://github.com/tektoncd/pipeline/issues/5504 and any potential changes to our feature gates implementation is tracked in https://github.com/tektoncd/pipeline/issues/5632

PR #5634, opened 2022-10-11
Add GA API policy and describe feature gates

dibyom:
# Changes

[Add GA API policy and describe feature gates](https://github.com/tektoncd/pipeline/pull/5634/commits/3a91ae83b56c0470ff501fcb7dbc65493d7457b7)

This commit describes our API policiy for our GA APIs (e.g. v1). It also
clarifies how TEP-33 Feature Gates interacts with CRD API versions.

Fixes https://github.com/tektoncd/pipeline/issues/5633
# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Updates API compatibility policy for the V1 api version
```

### DIFF
--- a/api_compatibility_policy.md
+++ b/api_compatibility_policy.md
@@ -1,13 +1,9 @@
 # API compatibility policy
 
-This document proposes a policy regarding making API updates to the CRDs in this
+This document proposes a policy regarding making updates to the Tekton API surface in this
 repo. Users should be able to build on the APIs in these projects with a clear
 idea of what they can rely on and what should be considered in progress and
 therefore likely to change.
-
-The use of `alpha`, `beta` and `GA` in this document is meant to correspond
-roughly to
-[the kubernetes API deprecation policies](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-a-flag-or-cli).
 
 ## What is the API?
 
@@ -29,35 +25,95 @@ A backwards incompatible change would be a change that requires a user to update
 existing instances of these CRDs in clusters where they are deployed and/or cause them
 to change their CRD definitions to account for changes in the above.
 
-## Alpha, Beta and GA
+## CRD API Versions
 
-Some of our CRDs and features are considered alpha, some are beta, and we are working
-toward GA.
+The `apiVersion` field in a Tekton CRD determines whether the overall API (and its default behaviors) is considered to be in `alpha`, `beta`, or `GA`. The use of `alpha`, `beta` and `GA` (aka `stable`) follows the corresponding Kubernetes [API stages definition](https://kubernetes.io/docs/reference/using-api/#api-versioning).
+
+Within a stable CRD, certain opt-in features or API fields gated may be considered `alpha` or `beta`. Similarly, within a beta CRD, certain opt-in features may be considered `alpha`. See the section on Feature Gates for details.
 
 The following CRDs are considered beta, though features may be introduced that are
 alpha:
 
-* `Task`
-* `TaskRun`
-* `ClusterTask`
-* `Pipeline`
-* `PipelineRun`
+- `Task`
+- `TaskRun`
+- `ClusterTask`
+- `Pipeline`
+- `PipelineRun`
 
-We are following [the Kubernetes definitions of these stages](https://kubernetes.io/docs/reference/using-api/api-overview/#api-versioning)
-and we are following [the Kubernetes deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/).
+### Alpha CRDs
 
-* Alpha:
-  * These features may be dropped at any time, though you will be given at least
-    one release worth of warning.
-* Beta:
-  * These features will not be dropped, though the details may change.
-  * Any [backwards incompatible changes](#backwards-incompatible-changes) must be
-    introduced in a backwards compatible manner first, with a deprecation warning
-    in the release notes and migration instructions.
-  * You will be given at least 9 months to migrate before a backward incompatible
-    change is made. This means an older beta API version will continue to be
-    supported in new releases for a period of at least 9 months from the time a
-    newer version is made available.
+- For Alpha CRDs, the `apiVersion` contains the `alpha` (e.g. `v1alpha1`)
+  
+- Features may be dropped at any time, though you will be given at least one release worth of warning.
+
+### Beta CRDs
+
+- The `apiVersion` field of the CRD contains contain `beta` (for example, `v1beta1`).
+
+- These features will not be dropped, though the details may change.
+
+- Any [backwards incompatible changes](#backwards-incompatible-changes) must be introduced in a backwards compatible manner first, with a   deprecation warning in the release notes and migration instructions.
+
+- Users will be given at least 9 months to migrate before a backward incompatible change is made. This means an older beta API version will continue to be supported in new releases for a period of at least 9 months from the time a newer version is made available.
+
+
+### GA CRDs
+
+- The `apiVersion` field of the CRD is `vX` where `X` is an integer.
+
+- Stable API versions remain available for all future releases within a [semver major version](https://semver.org/#summary).
+
+- Stable features may be marked as deprecated but may only be removed by incrementing the api version (i.e v1 to v2).
+
+- We will not make any backwards incompatible changes to fields that are stable without incrementing the api version.
+
+- Alpha and Beta features may be present within a stable API version. However, they will not be enabled by default and must be enabled by setting `enable-api-fields` to `alpha` or `beta`.
+
+## Feature Gates
+
+CRD API versions gate the overall stability of the CRD and its default behaviors. Within a particular CRD version, certain opt-in features may be at a lower stability level as described in [TEP-33](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md). These fields may be disabled by default and can be enabled by setting the right `enable-api-fields` feature-flag as described in TEP-33:
+
+* `stable` (default) - This value indicates that only fields of the highest stability level are enabled; For `beta` CRDs, this means only     beta stability fields are enabled, i.e. `alpha` fields are not enabled. For  `GA` CRDs, this means only `GA` fields are enabled by defaultd, i.e. `beta` and `alpha` fields would not be enabled.
+* `beta` - This value indicates that only fields which are of `beta` (or greater) stability are enabled, i.e. `alpha` fields are not enabled. 
+* `alpha` - This value indicates that fields of all stability levels are enabled, specifically `alpha`, `beta` and `GA`.
+
+
+| Feature Versions -> | v1 | beta | alpha |
+|---------------------|----|------|-------|
+| stable              | x  |      |       |
+| beta                | x  | x    |       |
+| alpha               | x  | x    | x     |
+
+
+See the current list of [alpha features](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features) and [beta features](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#beta-features).
+
+
+### Alpha features
+
+- Alpha feature in beta or GA CRDs are disabled by default and must be enabled by [setting `enable-api-fields` to `alpha`](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features)
+
+- These features may be dropped or backwards incompatible changes made at any time though will be given at least one release worth of warning
+
+- Alpha features are reviewed for promotion to beta at a regular basis. However, there is no guarantee that they will be promoted to beta.
+
+### Beta features
+
+- Beta features in GA CRDs are disabled by default and must be enabled by [setting `enable-api-fields` to `beta`](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#beta-features). In beta API versions, beta features are enabled by default.
+
+- Beta features may be deprecated or changed in a backwards incompatible way by following the same process as [Beta CRDs](#beta-crds) 
+  i.e. by providing a 9 month support period.
+
+- Beta features are reviewed for promotion to GA/Stable on a regular basis. However, there is no guarantee that they will be promoted to GA/stable.
+
+- For beta API versions, beta is the highest level of stability possible for any feature.
+  
+### GA/Stable features
+
+- GA/Stable features are present in a [GA CRD](#ga-crds) only.
+
+- GA/Stable features are enabled by default
+
+- GA/Stable features will not be removed or changed in a backwards incompatible manner without incrementing the API Version.
 
 ## Approving API changes
 


api_compatibility_policy.md
```diff
@@ -43,21 +40,41 @@ alpha:
...
+
+- For Alpha CRDs, the `apiVersion` contains the `alpha` (e.g. `v1alpha1`)
+  
+- Alpha feature in beta or stable CRDs are disabled by default and must be enabled by [setting `enable-api-fields` to `alpha`](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features)
+
+- These features may be dropped at any time, though you will be given at least one release worth of warning.
+
+### Beta
+
+- The `apiVersion` field of the CRD contains contain `beta` (for example, `v1beta1`).
```
JeromeJu, line 53: nit: we might want one less contain here

api_compatibility_policy.md
```diff
@@ -43,21 +40,41 @@ alpha:
...
-    in the release notes and migration instructions.
-  * You will be given at least 9 months to migrate before a backward incompatible
-    change is made. This means an older beta API version will continue to be
-    supported in new releases for a period of at least 9 months from the time a
-    newer version is made available.
+### Alpha
+
+- For Alpha CRDs, the `apiVersion` contains the `alpha` (e.g. `v1alpha1`)
+  
+- Alpha feature in beta or stable CRDs are disabled by default and must be enabled by [setting `enable-api-fields` to `alpha`](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features)
```
JeromeJu, line 47: Wondering if it make sense and need to mention "alpha" feature is a superset of "beta"?

api_compatibility_policy.md
dibyom: indeed

api_compatibility_policy.md
dibyom: do you mean if you enable alpha, beta features are also enabled?

lbernick:
@dibyom is this meant to clarify the existing api policy, or to propose the api policy discussed in this week's WG and your [API versioning doc](https://docs.google.com/document/d/15vD5vbK5UKCDFMoOqOwEc_NYACthtCZJlTJiiEEoljY)? (or is this just not ready for review yet? :) )

dibyom:
@tektoncd/core-maintainers PTAL!

api_compatibility_policy.md
```diff
@@ -43,21 +40,83 @@ alpha:
...
+
+
+See the current list of [alpha features](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features) and [beta features](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#beta-features).
+
+
+### Alpha features
+
+- Alpha feature in beta or GA CRDs are disabled by default and must be enabled by [setting `enable-api-fields` to `alpha`](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features)
+
+- These features may be dropped or backawards incompatible changes made at any time though will be given at least one release worth of warning
```
lbernick, line 98:
```suggestion
- These features may be dropped or backwards incompatible changes made at any time though will be given at least one release worth of warning
```

api_compatibility_policy.md
```diff
@@ -43,21 +40,83 @@ alpha:
...
+
+- The `apiVersion` field of the CRD is `vX` where `X` is an integer.
+
+- Stable API versions remain available for all future releases within a Tekton major version.
+
+- Stable features may be marked as deprecated but may only be removed by incrementing the api version (i.e v1 to v2).
+
+- We will not make any backwards incompatible changes to fields that are stable without incrementing the api version.
+
+- Stable APIs do not follow the [Go libraries compatibility policy](#go-libraries-compatibility-policy) i.e. backwards incompatible changes to the Go struct is not allowed.
```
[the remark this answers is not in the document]
dibyom: Hmm, I was thinking of only a subset of the go packages - esp. the ones in pkg/api/v1 and pkg/client. I'm not sure we need to have all exported functions under this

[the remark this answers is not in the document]
lbernick: Honestly maybe everything that's not in apis should be in an internal package?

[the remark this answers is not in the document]
vdemeester: > Honestly maybe everything that's not in apis should be in an internal package?

In theory yes 😬. Like in theory most of what is in `pkg` today should be in `internal/pkg` or something unless required otherwise (like `pkg/apis`, the generated clients, …). In practice using `internal` is not always super known by go dev.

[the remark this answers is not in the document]
wlynch: @dibyom I'd say anything visible in https://pkg.go.dev/github.com/tektoncd/pipeline is fair game.

+1 for moving things to internal.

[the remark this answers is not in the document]
lbernick: Any reason not to move things to internal? Vincent it sounds like you're saying we haven't done this just because most people aren't aware of the option? If that's the case I'm happy to make this change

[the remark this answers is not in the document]
lbernick: tracked in https://github.com/tektoncd/pipeline/issues/5679.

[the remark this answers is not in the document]
vdemeester: There is more to it than just "move" though. We did not use it too much initially, because it's not that common to use in the OSS golang community. But now, the project is there for a couple of years already, there is probably quite some project depending on some parts of our go packages, even only in the `tektoncd` org (triggers, …). We need to make sure we are not doing a "switch off" on those and, for each package, look at what is exported, why, is it useful, and is it used elsewhere ?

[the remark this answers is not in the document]
dibyom: At least for triggers, we only use pkg/client and pkg/apis (we have one test relying on pkg/names but that can probably be fixed easily). I do agree with your statement @vdemeester. My initial suggestion was just keeping only pkg/apis and pkg/client under the API policy - I don't think its feasible for us to promise no breaking changes to all of pkg/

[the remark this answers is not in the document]
lbernick: I agree with Billy that things indexed on pkg.go.dev make sense to cover in our compatibility policy. I would find it confusing to have a compatibility policy that covers some of our public go packages but not others. I agree we don't want to promise no changes to pkg/, but we could say we won't remove/rename a deprecated function/const for x time?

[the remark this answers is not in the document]
dibyom: > we could say we won't remove/rename a deprecated function/const for x time?

We could perhaps provide that but backwards compatibility technically also includes things like adding new arguments or return values to a function: https://go.dev/blog/module-compatibility 

@wlynch I'm not sure how we can both guarantee go module compatibility while also making rapid changes to alpha APIs if all of pkg/ is under scope.

I think clearly stating out which packages are in scope and which are not is a decent middle ground. Over time, we could move the the out of scope packages into internal. Another option would be to clearly write out what subset of compatibility we can provide e.g. we can support not deprecating/removing functions but we'd probably want the ability modify function signatures in most of pkg/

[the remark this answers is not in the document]
afrittoli: I'm a bit confused about how the policy can be different for stable APIs only - as go libraries are not necessarily bound to a specific API version. 

Does this summary make sense:
- go APIs specific to stable Tekton APIs are subject to the new policy, i.e. we cannot do backwards incompatible changes. Also, a future `v2` public API must be separated in code from the `v1` public API
- go APIs specific to alpha/beta Tekton APIs can be subject to backwards-incompatible changes
- go APIs which are shared between alpha/beta and stable that needs to stay public would have to be separated,  one copy under the stable tree, one copy under the alpha/beta one
- go APIs that we don't need to be public would have to become private, and we could (1) move them to private (2) create a public shim (3) deprecate the shim right away. This would have to happen before v1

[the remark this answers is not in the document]
dibyom: The first two points make sense to me. 

- For public go apis shared between alpha and beta - why do we need to copy unless we are making a breaking change? There is quite a bit of complexity involved in this so I'm sure we should do this unless we really need to

- I support moving them to private incrementally but I don't think we need to block the v1 API release on this. I'm unsure why we'd need to deprecate the shim.

We had some discussion on this in the last API WG - the conclusion was that I'd update the policy to mention add  guidelines for making changes to the go libraries. Specifically this would be to avoid making backwards incompatible changes to the go APIs pertaining to the V1 API version.

dibyom:
I updated and split up the PR to separate out the bit around go libraries compatibility since that seems to be more controversial. Happy to move forward without the changes to Go libraries if that is easier! @vdemeester @lbernick Let me know if the update captures the discussion and what you had in mind

Also, @afrittoli and any other @tektoncd/core-maintainers this should be ready for another review!

api_compatibility_policy.md
```diff
@@ -29,35 +25,96 @@ A backwards incompatible change would be a change that requires a user to update
...
+
+- Users will be given at least 9 months to migrate before a backward incompatible change is made. This means an older beta API version will continue to be supported in new releases for a period of at least 9 months from the time a newer version is made available.
+
+- Beta API versions follow the [Go libraries compatibility policy](#go-libraries-compatibility-policy).
+
+### GA CRDs
+
+- The `apiVersion` field of the CRD is `vX` where `X` is an integer.
+
+- Stable API versions remain available for all future releases within a Tekton major version.
```
lbernick, line 65: does "tekton major version" here refer to semver?

api_compatibility_policy.md
```diff
@@ -29,35 +25,96 @@ A backwards incompatible change would be a change that requires a user to update
 existing instances of these CRDs in clusters where they are deployed and/or cause them
 to change their CRD definitions to account for changes in the above.
 
-## Alpha, Beta and GA
+## CRD API Versions
+
+The `apiVersion` field in a Tekton CRD determines whether the overall API (and its default behaviors) is considered to be in `alpha`, `beta`, or `GA`. The use of `alpha`, `beta` and `GA` (aka `stable`) follows the corresponding Kubernetes [API stages definition](https://kubernetes.io/docs/reference/using-api/#api-versioning) and its [deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/).
```
lbernick, line 30: I think saying we follow k8s deprecation policy is a bit confusing actually (the main differences I see are the ability to mark a beta api as deprecated immediately, rather than waiting 9 months in addition to the 9 months required to remove it) and our feature flags policy.

api_compatibility_policy.md
```diff
@@ -29,35 +25,96 @@ A backwards incompatible change would be a change that requires a user to update
...
+
+- Stable API versions remain available for all future releases within a Tekton major version.
+
+- Stable features may be marked as deprecated but may only be removed by incrementing the api version (i.e v1 to v2).
+
+- We will not make any backwards incompatible changes to fields that are stable without incrementing the api version.
+
+- Alpha and Beta features may be present within a stable API version. However, they will not be enabled by default and must be enabled by setting `enable-api-fields` to `alpha` or `beta`.
+
+## Features Gates
```
lbernick, line 73:
```suggestion
## Feature Gates
```

api_compatibility_policy.md
```diff
@@ -29,35 +25,96 @@ A backwards incompatible change would be a change that requires a user to update
...
+- Alpha features are reviewed for promotion to beta at a regular basis. However, there is no guarantee that they will be promoted to beta.
+
+### Beta features
+
+- Beta features in GA CRDs are disabled by default and must be enabled by [setting `enable-api-fields` to `beta`](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#beta-features). In beta API versions, beta features are enabled by default.
+
+- Beta features may be deprecated or changed in a backwards incompatible way by following the same process as [Beta CRDs](#beta-crds) 
+  i.e. by providing a 9 month support period.
+
+- Beta features are reviewed for promotion to beta at a GA/Stable basis. However, there is no guarantee that they will be promoted to GA/stable.
```
lbernick, line 107:
```suggestion
- Beta features are reviewed for promotion to GA/Stable at a regular basis. However, there is no guarantee that they will be promoted to GA/stable.
```

api_compatibility_policy.md
dibyom: yes - will update the wording to include semver major version

api_compatibility_policy.md
dibyom: ok, yeah the k8s deprecation policy also includes things like cli flags that are not relevant to us at the moment

api_compatibility_policy.md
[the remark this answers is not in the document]
afrittoli: > The first two points make sense to me.
> 
> * For public go apis shared between alpha and beta - why do we need to copy unless we are making a breaking change? There is quite a bit of complexity involved in this so I'm sure we should do this unless we really need to

I meant shared between alpha/beta and stable. If we can make sure (through testing) that we would catch those cases, then it would be fine to not copy the code until a breaking change is needed on alpha or beta side.

> * I support moving them to private incrementally but I don't think we need to block the v1 API release on this. I'm unsure why we'd need to deprecate the shim.

We don't need to block v1 only on this as long as our API stability policy excludes them upfront. 
If we create a shim, it would only be to give existing users more time to move away from those APIs that we want to become private. But we would have to deprecate the shim to signal our users that that is only a temporary solution.
Alternatively we could deprecate the API before making it private, and when it moves to private provide no shim at all.

> 
> We had some discussion on this in the last API WG - the conclusion was that I'd update the policy to mention add guidelines for making changes to the go libraries. Specifically this would be to avoid making backwards incompatible changes to the go APIs pertaining to the V1 API version.

api_compatibility_policy.md
```diff
@@ -89,9 +145,16 @@ These changes must be approved by [more than half of the project OWNERS](OWNERS)
...
-## Go libraries compatibility policy
+## Go Libraries
+
+The following guidelines describe backwards incompatible changes and how they relate to release versioning for the public Go packages under `github.com/tektoncd/pipeline/pkg`.
+
+Packages that are part of the Go client library for a particular CRD version should adhere to the stability guarantees for that version. This includes the Go structs for the API types in `pkg/apis/pipeline/$apiVersion` and the client libraries generated from those in `pkg/client`. This means:
+
+- Changing or removing a field in a go struct in `pkg/apis/pipeline/v1alpha1` (or another alpha API) is not considered a backwards incompatible change.
+
+- Renaming a field in `pkg/apis/pipeline/v1beta1` (or another beta API) is not considered a backwards incompatible change. However, removing a field is considered a backwards incompatible change and requires a major version bump.
```
afrittoli, line 156: Similar comment as for alpha. We cannot really define what is and is not backwards compatible.
The [reference](https://go.dev/blog/module-compatibility) you linked below looks like a good one to explain what is and what is not backwards compatible. 

I believe what should change between alpha and beta is the amount of notice we give users for backward incompatible changes. Additionally, we can define changes that require a new major beta version no matter the amount of notice - for instance, we can say that removing a field requires a new major version.

It seems to me that separating renaming from removing is artificial. Renaming is at all effects removal of the old name + the addition of the new name. While the same content is stored, old clients won't find the new field without changes.

I think the way to do renaming is to provide the new name and keep the old one but deprecate it. 
After the agreed deprecation period is met, we remove the old field.

In stable APIs instead, we can never remove a field. 
We could add the new name, which is backwards compatible, but only a new stable version could remove the field

afrittoli (commented):
Thank you for this PR - see some comments on the go library side.

api_compatibility_policy.md
lbernick: I agree that separating removing from renaming is a bit artificial (even though it was partially my idea 😅 ) but I'm not sure how realistic it is to announce to users in advance a change we want to make to the go libs. (maybe it is?)
Do you think it would be OK to just allow backwards incompatible changes in v1beta1 go libs? since the definition of beta is that features won't be removed, but the details may change-- I think if we're not changing the yaml/json definitions it's a reasonable policy.

afrittoli: I think it's ok to allow backwards incompatible changes in the beta API, but I think we should provide notice.
Similar to the API policy:
- alpha we can do backwards incompatible without notice 
- beta we can do backwards incompatible changes with notice (i.e. going through deprecation)
- stable we cannot do backwards incompatible changes, they have to go in a new version

dibyom: I tend to agree...what I'm a bit worried about is how well we can actually implement providing notice for the beta  changes - things like adding an extra arg or a return value to a function are not backwards incompatible. Do you see us providing notice for changes like those across all exported functions in pkg/? If so, how long would the notice period be? (Personally, I'd be ok with a `action-required` release note)

afrittoli: In terms of how to deprecate, I think action required release notes are good, and comment in the code, so that's shown by IDEs 

In terms of duration, what about three releases? I think our current API policy of 9 months is too much, especially it will be once we have a v1 out, so definitely I would go shorter than that.

dibyom: updated to 3 releases

api_compatibility_policy.md
```diff
@@ -29,35 +25,95 @@ A backwards incompatible change would be a change that requires a user to update
...
+
+- For beta API versions, beta is the highest level of stability possible for any feature.
+  
+### GA/Stable features
+
+- GA/Stable features are present in a [GA CRD](#ga-crds) only.
+
+- GA/Stable features are enabled by default
+
+- GA/Stable features will not be removed or changed in a backwards incompatible manner.
```
lbernick, line 116:
```suggestion
- GA/Stable features will not be removed or changed in a backwards incompatible manner without incrementing the API version
```

api_compatibility_policy.md
```diff
@@ -89,9 +145,16 @@ These changes must be approved by [more than half of the project OWNERS](OWNERS)
...
+
+The following guidelines describe backwards incompatible changes and how they relate to release versioning for the public Go packages under `github.com/tektoncd/pipeline/pkg`.
+
+Packages that are part of the Go client library for a particular CRD version should adhere to the stability guarantees for that version. This includes the Go structs for the API types in `pkg/apis/pipeline/$apiVersion` and the client libraries generated from those in `pkg/client`. This means:
+
+- [Backwards incompatible changes](https://go.dev/blog/module-compatibility) such as changing or removing a field in a go struct in `pkg/apis/pipeline/v1alpha1` (or another alpha API) does not require a major version bump.
+
+- Backwards incompatible changes in `pkg/apis/pipeline/v1beta1` (or another beta API) require providing a notice of the upcoming change for at least 3 releases before making the change. However, they do not require a major version bump.
+
+- For `pkg/apis/pipeline/v1` (or any other stable APIs), any [backwards incompatible change](https://go.dev/blog/module-compatibility) will require a major version bump.
```
vdemeester, line 158: By *major* we mean `{major}.{minor}.{bugfix}` ?

lbernick: I think this was intended to refer to semver since it's talking about the go libraries-- @dibyom can you update to clarify?

dibyom: I was referring to semver. I do think we need to bump the API version if we make a breaking change to the API itself, but if we are just renaming a function or something, I think we can just update the semver version.

afrittoli: I'm confused. We have the API version, which is not semver, and the version of the go module, for which we only have used minor and patch, major is still 0 (e.g. 0.41.1).  So we can do a breaking change in the code in `pkg/apis/pipeline/v1`, as long as we release a v1.0.0, v2.0.0 etc?

About renaming, for an API in a package associated to a stable version, I would prefer we never did it in one go, regardless of the version. We should create a new field / function with the name we want and deprecate the old one.

dibyom: > So we can do a breaking change in the code in pkg/apis/pipeline/v1, as long as we release a v1.0.0, v2.0.0 etc?

Technically yes, the intent is to strongly discourage these kinds of changes since major version bumps should be rare. (If it changes the user facing API and not just a function, we'd need to change the API version as well)

> About renaming, for an API in a package associated to a stable version, I would prefer we never did it in one go, regardless of the version. We should create a new field / function with the name we want and deprecate the old one.

I agree, I don't think anything here is saying we should not do that. (I can include that explicitly if that helps)

dibyom:
[Splitting off the discussion around Go libraries and releases to a separate PR]

afrittoli (commented):
Thank you!
/approve

PR #5664, opened 2022-10-19
Set ConfigSource in gitresolver

chuangw6:
# Changes

/kind feature

Prior, a field named Source was introduced to ResolutionRequest status to record the source where the remote resource came from. And the individual resolvers need to implement the Source function to set the correct source value. But gitresolver just returns nil.

Now, we return correct source value with the 3 subfields: url, digest and entrypoint
- If users choose to use anonymous cloning, the url is just user-provided value for the `url` param. If scm api is used, it would be the clone URL of the repo available from scm repo service.
- Digest algorithm name is fixed value "sha1" and the hex-encoded value is the actual commit sha at the moment of resolving even if users provide a tag/branch name for the `revision` param.
- Entrypoint is the user-provided value for the `path` param.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Set ConfigSource value for gitresolver to record the source where the remote resource came from. 
```


chuangw6:
/assign @wlynch 

Please take a look at if the value set for configsource in gitresolver meets our needs on Chains side. Thank you!!

docs/git-resolver.md
```diff
@@ -146,6 +146,55 @@ spec:
   * BitBucket Server
   * BitBucket Cloud
 
+## `ResolutionRequest` Status
+`ResolutionRequest.Status.Source` field captures the source where the remote resource came from. It includes the 3 subfields: `url`, `digest` and `entrypoint`.
+- `url`
```
jagathprakash, line 151: Is this data enough to uniquely identify the file? Does the branch have to be encoded somewhere?

docs/git-resolver.md
```diff
@@ -146,6 +146,55 @@ spec:
...
+  - If users choose to use anonymous cloning, the url is just user-provided value for the `url` param. 
+  - If scm api is used, it would be the clone URL of the repo fetched from scm repository service.
```
jagathprakash, line 153: Could you add an example.

pkg/resolution/resolver/git/resolver_test.go
```diff
@@ -765,3 +780,27 @@ func toParams(m map[string]string) []pipelinev1beta1.Param {
...
 }
+
+func revisionToCommitSHA(reqURL string, useNthCommit int, revision, defaultRevision, specificCommit string, commitsByBranch map[string][]string, commitByTag map[string]string) string {
+	switch {
+	case useNthCommit > 0:
+		return commitsByBranch[plumbing.Master.Short()][useNthCommit]
+	case specificCommit != "":
+		return specificCommit
+	case revision == "" && reqURL != "":
+		return revisionByBranchOrTag(commitsByBranch, commitByTag, plumbing.Master.Short())
```
wlynch, line 791: This has me a little worried, since this could be defaulting to branch names. Instead of trying to plumb all of the repo data through commitByBranch / commitByTag, could we have createTestRepo return the https://pkg.go.dev/github.com/go-git/go-git/v5#Repository and we can use https://pkg.go.dev/github.com/go-git/go-git/v5#Repository.ResolveRevision instead?

wlynch (changes requested):

docs/git-resolver.md
```diff
@@ -146,6 +146,55 @@ spec:
...
+- `digest`
+  - The algorithm name is fixed "sha1".
```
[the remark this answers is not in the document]
chuangw6: Thanks for calling out this! I added a note here and I will keep in mind this.

pkg/resolution/resolver/git/resolver_test.go
chuangw6: Good point. I've changed createTestRepo as suggested. PTAL. Thank you!!

pkg/resolution/resolver/git/resolver.go
```diff
@@ -156,29 +156,67 @@ func (r *Resolver) resolveAPIGit(ctx context.Context, params map[string]string)
...
+
+	// find the commit sha if the provided revision is a branch name
+	ref, _, err := scmClient.Git.FindBranch(ctx, orgRepo, revision)
+	if err == nil && ref != nil {
+		return ref.Sha, nil
+	}
+
+	return "", fmt.Errorf("the provided revision %q is not a real commit sha or a name of a branch or a tag.", revision)
+}
+
```
chuangw6, line 219: Hi @abayer ,
This is for scm cloning approach only. What I am trying to do here is retrieve the actual commit sha rather than content sha regardless of the user-provided revision param which can be commit sha / branch name / tag name.

The approach in `fetchCommitSHAForAPIGit` seems the best solution for now but the fake test utilities in the scm package do no implement GitService's [FindBranch](https://github.com/jenkins-x/go-scm/blob/648bf6dd5dd722041d215869a547045f20984d29/scm/driver/fake/git.go#L37-L39), which blocks unit test in our case. 

The alternative I tried is to use GitService's [FindRef](https://github.com/jenkins-x/go-scm/blob/338e0fcd608a59f137b4aa9f97d4761082cf8729/scm/git.go#L93-L94) to find commit sha, but this is not optional because the 3rd argument expected differs from provider to provider. For example, [github](https://docs.github.com/en/rest/git/refs#get-a-reference)
- we need to prepend "heads/" to the revision if it's branch name
- we need to prepend "tags/"  to the revision if it's tag name 


wdyt of the approach in `fetchCommitSHAForAPIGit`? If it sounds good, I am happy to submit a PR to implement the missing fake [FindBranch](https://github.com/jenkins-x/go-scm/blob/648bf6dd5dd722041d215869a547045f20984d29/scm/driver/fake/git.go#L37-L39) in scm package to help us do unit test.

pkg/resolution/resolver/git/resolver_test.go
chuangw6: Discussed with @wlynch offline. By this approach, we seem to compare value returned by [ResolveRevision](https://pkg.go.dev/github.com/go-git/go-git/v5#Repository.ResolveRevision) in test with the value returned by the same function in implementation code which uses same function for anonymous cloning.

Here is the latest update:
- for anonymous cloning: 
I changed createTestRepo to return a slice of commit SHAs. In the test table, I added a new field `expectedNthCommit` so that we can compare the commit SHA.
- for scm api cloning:
The implementation code sets the the digest to be the content sha atm, but we need to change to actual commit sha. See the discussion [here](https://github.com/tektoncd/pipeline/pull/5664#discussion_r1011858087)

wdyt?

pkg/resolution/resolver/git/resolver.go
```diff
@@ -323,17 +362,20 @@ func (r *resolvedGitResource) Annotations() map[string]string {
...
 // file came from including the url, digest and the entrypoint.
 func (r *resolvedGitResource) Source() *pipelinev1beta1.ConfigSource {
-	return nil
+	return &pipelinev1beta1.ConfigSource{
+		URI: spdxGit(r.URL),
+		Digest: map[string]string{
+			"sha1": r.Revision,
+		},
+		EntryPoint: r.Path,
+	}
```
chuangw6, line 378: Hi @wlynch ,
For the url field, I am trying to format the url to the spdx download format as recommended by intoto [here](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#:~:text=ResourceURI%20(string)). Do you think we should do it as is in pipeline or we should format it in Chains?

Please let me know if you have any concerns or feedback. Thanks you!!!

pkg/resolution/resolver/git/resolver_test.go
```diff
@@ -489,16 +509,19 @@ func TestResolve(t *testing.T) {
...
-					default:
-						expectedStatus.Annotations[AnnotationKeyRevision] = tc.revision
+
+					var expectedCommitSHAInStatus string
+					// Anonymous cloning
+					if len(tc.commits) > 0 {
+						expectedCommitSHAInStatus = commitSHAs[tc.expectedNthCommit]
+					} else if tc.revision != "" {
+						// scm cloning
+						expectedCommitSHAInStatus = tc.revision
```
wlynch, line 519: Not blocking if you want to push forward with this as-is, but wdyt about simplifying this further? I think something like this would make it easier to see what's going on:

```go
repo := createTestRepo(...)

// map revision -> resolved SHA
// SHAs should be deterministic if timestamps are static.
tc := map[string]string{
  "": "abcd",
  "abcd": "abc",
  "main": "abc",
  "other-branch": "def",
  "main^": "def",
  "refs/tags/v1": "def",
  "refs/other/asdf": "xyz",
}

for k, v := range want {
  req := createRequest(repo, k)
  
 resp := new(v1beta1.ResolutionRequestStatus)
 RunResolverReconcileTest(..., got, ...)  // or call resolver directly

 want  := &v1beta1.ResolutionRequestStatus{
    ...
    Annotations: map[string]string{
      AnnotationKeyRevision: v,
      ...
    }
  }

  if cmp.Diff(want, got) {
     ...
  }
}
```

pkg/resolution/resolver/git/resolver.go
wlynch: Fine with me!

pkg/resolution/resolver/git/resolver.go
abayer: I'd say go with adding the fake `FindBranch` - cc me on the PR and I'll review it and ping the maintainer to make sure it gets merged. =)

pkg/resolution/resolver/git/resolver.go
```diff
@@ -156,29 +156,67 @@ func (r *Resolver) resolveAPIGit(ctx context.Context, params map[string]string)
...
 		Org:      params[orgParam],
 		Repo:     params[repoParam],
 		Path:     content.Path,
 		URL:      repo.Clone,
 	}, nil
 }
 
+// fetchCommitSHAForAPIGit fetches actual commit SHA using scm api based on
+// the user provided revision which can be the name of commitSHA/branch/tag.
+func fetchCommitSHAForAPIGit(ctx context.Context, scmClient *scm.Client, orgRepo, revision string) (string, error) {
```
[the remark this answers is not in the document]
chuangw6: Thanks @wlynch !! Yay git service's [FindCommit](https://pkg.go.dev/github.com/jenkins-x/go-scm/scm#GitService) in scm api does the job.

pkg/resolution/resolver/git/resolver_test.go
chuangw6: Thanks a lot for the suggestion here @wlynch !!! Really appreciate it!

> repo := createTestRepo(...)

Creating a local test repo in advance with some pre-defined commits rather than per-test commits is a good idea. Done!  (Note this setting up local test repo is for anonymous clone approach only. SCM local repo set up is a bit different.)

> // map revision -> resolved SHA. SHAs should be deterministic if timestamps are static.

SG! Done!

> RunResolverReconcileTest(..., got, ...)  // or call resolver directly

Perhaps we should keep using `RunResolverReconcileTest` because I found in a previous commit in which a unit test [calls `resolve` directly](https://github.com/tektoncd/pipeline/blob/6074fa43e2381201ba94576ad1cfe815312ae228/pkg/resolution/resolver/git/resolver_test.go#L268), but they all consolidated to `RunResolverReconcileTest` at some point. So it might be a good idea to keep using RunResolverReconcileTest, which is consistent with how other resolvers are tested.


With regards to the test table, I think we should have keep existing fields other than having the revision itself because we do need to test some failure cases i.e path not found, scm token etc. 

That said, I've made some changes to the test table to make it look a bit cleaner. Please take a look and let me know wdyt. ty!

pkg/resolution/resolver/git/resolver.go
chuangw6: Thanks for the thoughts @abayer.

@wlynch just helped me realize the git service's [FindCommit](https://pkg.go.dev/github.com/jenkins-x/go-scm/scm#GitService) in scm api can also fetch a commit by commit sha/branch/tag name, which I've used in `fetchCommitSHAForAPIGit` but I thought it only recognizes an commit sha but not tag/branch name :D

pkg/resolution/resolver/git/resolver_test.go
```diff
@@ -191,7 +190,43 @@ func TestResolveNotEnabled(t *testing.T) {
...
+		Content:  "released content in main branch and in tag v1",
+		Tag:      "v1",
+	}}
+
+	repoPath := createTestRepo(t, commits)
+
+	// NOTE: these 3 commit shas are static because of the static content, commit msg, author info and time.
+	firstCommitSHA := "24f5ed4a93a11148ef3ebd2bda567413585b5c55"
+	secondCommitSHA := "117c5a12426a0ec9780953244e89c1e219d2e9e9"
+	thirdCommitSHA := "c618acf88328cdd2e6f5282a96529065d59ab288"
```
wlynch, line 226: nit: `commits := []string{`?

pkg/resolution/resolver/git/resolver_test.go
```diff
@@ -481,33 +517,30 @@ func TestResolve(t *testing.T) {
...
-						expectedStatus.Annotations[AnnotationKeyURL] = reqParams[urlParam]
+					if tc.args.url != "" {
+						expectedStatus.Annotations[AnnotationKeyURL] = tc.args.url
 					} else {
-						expectedStatus.Annotations[AnnotationKeyOrg] = reqParams[orgParam]
-						expectedStatus.Annotations[AnnotationKeyRepo] = reqParams[repoParam]
+						expectedStatus.Annotations[AnnotationKeyOrg] = tc.args.org
+						expectedStatus.Annotations[AnnotationKeyRepo] = tc.args.repo
 						expectedStatus.Annotations[AnnotationKeyURL] = fakeClone
 					}
```
wlynch, line 534: Shouldn't this data always be the same? 🤔

pkg/resolution/resolver/git/resolver_test.go
```diff
@@ -212,240 +247,241 @@ func TestResolve(t *testing.T) {
...
 					}}
 
+					// git service
+					scmData.Commits = map[string]*scm.Commit{
+						"main":  {Sha: revisionToSHA["main"]},
+						"other": {Sha: revisionToSHA["other"]},
+					}
 					return scmClient, nil
 				},
 			}
```
wlynch, line 483: nit: This data all looks like it doesn't change. If you don't need a unique resolver for each test case, consider hoisting this out of the test loop.

wlynch (changes requested):
Some minor changes - but otherwise LGTM!

pkg/resolution/resolver/git/resolver_test.go
chuangw6: 👍  SGTM!!!

pkg/resolution/resolver/git/resolver_test.go
chuangw6: theoretically yes, args in the test cases are using those constants, and I used the value from test case to be consistent.
I just changed to the constant. But we cannot omit the if-else because annotation keys are different between scm cloning and anonymous cloning. Please let me know if you wanted to suggest something else :D

pkg/resolution/resolver/git/resolver_test.go
chuangw6: SGTM. Also changed to dynamic commits that were returned from createRestRepo as we discussed offline. PTAL. tyty!

pkg/resolution/resolver/git/resolver_test.go
```diff
@@ -212,240 +247,241 @@ func TestResolve(t *testing.T) {
...
 
+	revisionToSHA := map[string]string{
+		"":             thirdCommitSHA,
+		"v1":           thirdCommitSHA,
+		"refs/tags/v1": thirdCommitSHA,
+		"test-branch":  secondCommitSHA,
+		firstCommitSHA: firstCommitSHA,
+		"main":         firstCommitSHA,
+		"other":        thirdCommitSHA,
+	}
```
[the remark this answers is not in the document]
chuangw6: > keeping the input/output pairings closer together is probably more important

Done! Concur with that.

wlynch (approved):
/lgtm

PR #5666, opened 2022-10-19
[TEP-0115] Support Artifact Hub in Hub Resolver

QuanZhang-William:
# Changes

Part of [issues/667].
This commit adds support to resolve catalog from the [Artifact Hub] and keep current functionality of fetching catalog from the Tekton Hub.

- Change 1:

The commit adds a new field `type` to the hub resolver indicating the type of the Hub to pull the resource from. The value can be set to `tekton` or `artifact`. By default, the resolver fetches resources from `https://artifacthub.io/` when setting `type` to  `artifact`. 

- Change 2:

Prior to this change, the hub resolver only supports pulling resources from the Tekton Hub. This commit updates the default hub type to `artifact` since the [Artifact Hub][Artifact Hub] will be the main entrypoint for Tekton Catalogs in the future.

- Change 3:

Prior to this change, the default Tekton Hub URL is: `https://api.hub.tekton.dev`. This commit removes the default value of the Tekton Hub URL and enforces users to configure their own instance of Tekton Hub since the public instance `https://api.hub.tekton.dev` will be deprecated after the migration to Artifact Hub is done.

/kind feature

[Artifact Hub]: https://artifacthub.io/
[issues/667]: https://github.com/tektoncd/hub/issues/667

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
The Hub Resolver will have a new `type` field to indicate the type of Hub from where to pull the resource. The default hub type is updated from the Tekton Hub to the Artifact Hub. Please see more details in TEP-0115
```

### DIFF
--- a/cmd/resolvers/main.go
+++ b/cmd/resolvers/main.go
@@ -33,21 +33,26 @@ import (
 
 func main() {
 	ctx := filteredinformerfactory.WithSelectors(signals.NewContext(), v1alpha1.ManagedByLabelKey)
-
-	apiURL := os.Getenv("HUB_API")
-	hubURL := hub.DefaultHubURL
-	if apiURL == "" {
-		hubURL = hub.DefaultHubURL
-	} else {
-		if !strings.HasSuffix(apiURL, "/") {
-			apiURL += "/"
-		}
-		hubURL = apiURL + hub.YamlEndpoint
-	}
+	tektonHubURL := buildHubURL(os.Getenv("TEKTON_HUB_API"), "", hub.TektonHubYamlEndpoint)
+	artifactHubURL := buildHubURL(os.Getenv("ARTIFACT_HUB_API"), hub.DefaultArtifactHubURL, hub.ArtifactHubYamlEndpoint)
 
 	sharedmain.MainWithContext(ctx, "controller",
 		framework.NewController(ctx, &git.Resolver{}),
-		framework.NewController(ctx, &hub.Resolver{HubURL: hubURL}),
+		framework.NewController(ctx, &hub.Resolver{TektonHubURL: tektonHubURL, ArtifactHubURL: artifactHubURL}),
 		framework.NewController(ctx, &bundle.Resolver{}),
 		framework.NewController(ctx, &cluster.Resolver{}))
+}
+
+func buildHubURL(configAPI, defaultURL, yamlEndpoint string) string {
+	var hubURL string
+	if configAPI == "" {
+		hubURL = defaultURL
+	} else {
+		if !strings.HasSuffix(configAPI, "/") {
+			configAPI += "/"
+		}
+		hubURL = configAPI + yamlEndpoint
+	}
+
+	return hubURL
 }
--- a/docs/hub-resolver.md
+++ b/docs/hub-resolver.md
@@ -6,10 +6,13 @@ Use resolver type `hub`.
 
 | Param Name       | Description                                                                   | Example Value                                              |
 |------------------|-------------------------------------------------------------------------------|------------------------------------------------------------|
-| `catalog`        | The catalog from where to pull the resource (Optional)                        | Default:  `Tekton`                                         |
-| `kind`           | Either `task` or `pipeline`                                                   | `task`                                                     |
+| `catalog`        | The catalog from where to pull the resource (Optional)                        | Default:  `tekton-catalog-tasks` (for `task` kind);  `tekton-catalog-pipelines` (for `pipeline` kind)                                        |
+| `type`           | The type of Hub from where to pull the resource (Optional). Either `artifact` or `tekton` | Default:  `artifact`                                         |
+| `kind`           | Either `task` or `pipeline` (Optional)                                        | Default: `task`                                                     |
 | `name`           | The name of the task or pipeline to fetch from the hub                        | `golang-build`                                             |
-| `version`        | Version of task or pipeline to pull in from hub. Wrap the number in quotes!   | `"0.5"`                                                    |
+| `version`        | Version of task or pipeline to pull in from hub. Wrap the number in quotes!   | `"0.5.0"`                                                    |
+
+The Catalogs in the Artifact Hub follows the semVer (i.e.` <major-version>.<minor-version>.0`) and the Catalogs in the Tekton Hub follows the simplified semVer (i.e. `<major-version>.<minor-version>`). Both full and simplified semantic versioning will be accepted by the `version` parameter. The Hub Resolver will map the version to the format expected by the target Hub `type`.
 
 ## Requirements
 
@@ -26,24 +29,43 @@ for the name, namespace and defaults that the resolver ships with.
 
 ### Options
 
-| Option Name       | Description                                          | Example Values     |
-|-------------------|------------------------------------------------------|--------------------|
-| `default-catalog` | The default catalog from where to pull the resource. | `tekton`           |
-| `default-kind`    | The default object kind for references.              | `task`, `pipeline` |
+| Option Name                 | Description                                          | Example Values         |
+|-----------------------------|------------------------------------------------------|------------------------|
+| `default-tekton-hub-catalog`| The default tekton hub catalog from where to pull the resource.| `Tekton`               |
+| `default-artifact-hub-task-catalog`| The default artifact hub catalog from where to pull the resource for task kind.| `tekton-catalog-tasks`               |
+| `default-artifact-hub-pipeline-catalog`| The default artifact hub catalog from where to pull the resource for pipeline kind.  | `tekton-catalog-pipelines`               |
+| `default-kind`              | The default object kind for references.              | `task`, `pipeline`     |
+| `default-type`              | The default hub from where to pull the resource.     | `artifact`, `tekton`   |
 
 
 ### Configuring the Hub API endpoint
 
-By default this resolver will hit the public hub api at https://hub.tekton.dev/
+The Hub Resolver supports to resolve resources from the [Artifact Hub](https://artifacthub.io/) and the [Tekton Hub](https://hub.tekton.dev/),
+which can be configured by setting the `type` field of the resolver. 
+
+*(Please note that the [Tekton Hub](https://hub.tekton.dev/) will be deprecated after [migration to the Artifact Hub](https://github.com/tektoncd/hub/issues/667) is done.)*
+
+When setting the `type` field to `artifact`, the resolver will hit the public hub api at https://artifacthub.io/ by default
 but you can configure your own (for example to use a private hub
-instance) by setting the `HUB_API` environment variable in
+instance) by setting the `ARTIFACT_HUB_API` environment variable in
 [`../config/resolvers/resolvers-deployment.yaml`](../config/resolvers/resolvers-deployment.yaml). Example:
 
 ```yaml
 env
-- name: HUB_API
-  value: "https://api.hub.tekton.dev/"
+- name: ARTIFACT_HUB_API
+  value: "https://artifacthub.io/"
 ```
+
+When setting the `type` field to `tekton`, you **must** configure your own instance of the Tekton Hub by setting the `TEKTON_HUB_API` environment variable in
+[`../config/resolvers/resolvers-deployment.yaml`](../config/resolvers/resolvers-deployment.yaml). Example:
+
+```yaml
+env
+- name: TEKTON_HUB_API
+  value: "https://api.private.hub.instance.dev"
+```
+
+The Tekton Hub deployment guide can be found [here](https://github.com/tektoncd/hub/blob/main/docs/DEPLOYMENT.md).
 
 ## Usage
 
@@ -59,7 +81,9 @@ spec:
     resolver: hub
     params:
     - name: catalog # optional
-      value: Tekton
+      value: tekton-catalog-tasks
+    - name: type # optional
+      value: artifact 
     - name: kind
       value: task
     - name: name
@@ -80,7 +104,9 @@ spec:
     resolver: hub
     params:
     - name: catalog # optional
-      value: Tekton 
+      value: tekton-catalog-pipelines 
+    - name: type # optional
+      value: artifact
     - name: kind
       value: pipeline
     - name: name
--- a/examples/v1beta1/taskruns/hub-resolver.yaml
+++ b/examples/v1beta1/taskruns/hub-resolver.yaml
@@ -1,7 +1,7 @@
 apiVersion: tekton.dev/v1beta1
 kind: TaskRun
 metadata:
-  generateName: hub-resolver
+  generateName: hub-resolver-simple-semver-
 spec:
   workspaces:
     - name: output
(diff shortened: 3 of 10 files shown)


abayer:
Nice!

jerop (commented):
please add examples e.g. https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/taskruns/hub-resolver.yaml
(ensure that we cover the minimal and full versions of resources)

@tektoncd/core-maintainers added this to the v0.41 milestone so that the first release of hub resolver will support artifact hub (and don't have to deal with deprecation/migration of default behavior)

QuanZhang-William:
> please add examples e.g. https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/taskruns/hub-resolver.yaml (ensure that we cover the minimal and full versions of resources)
> 
> @tektoncd/core-maintainers added this to the v0.41 milestone so that the first release of hub resolver will support artifact hub (and don't have to deal with deprecation/migration of default behavior)

Thanks @jerop ! Will add the example.

To confirm my understanding, I have [this code](https://github.com/tektoncd/pipeline/blob/ee293745b6c0f59c4ac2bc6c39ac59b43af69334/pkg/resolution/resolver/hub/resolver.go#L260) to deal with backward compatibility for users who specifically set Catalog to `"Tekton"` for [example](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/taskruns/hub-resolver.yaml#L20) (without such code, the resolver will try to find the Catalog named `"Tekton"` in the Artifact Hub which does not exist). 

If we don't need to deal the migration/deprecation of the default behavior, I can drop the code?

examples/v1beta1/taskruns/hub-resolver.yaml
```diff
@@ -17,11 +17,39 @@ spec:
...
+    - name: url
+      value: https://github.com/tektoncd/pipeline.git
+    - name: revision
+      value: main
+  taskRef:
+    resolver: hub
+    params:
+      - name: type  #optional
+        value: artifact    
+      - name: kind
```
jerop, line 50: `kind` is optional too, right?

can we have an example with the required fields only?

jerop:
known flake: https://github.com/tektoncd/pipeline/issues/4643

/retest

pkg/resolution/resolver/hub/resolver.go
```diff
@@ -107,14 +136,19 @@ func (r *Resolver) Resolve(ctx context.Context, params []pipelinev1beta1.Param)
...
+		if typeString, ok := conf[ConfigType]; ok {
+			paramsMap[ParamType] = typeString
 		} else {
-			return nil, fmt.Errorf("default catalog was not set during installation of the hub resolver")
+			return nil, fmt.Errorf("default type was not set during installation of the hub resolver")
 		}
 	}
+	if paramsMap[ParamType] != ArtifactHubType && paramsMap[ParamType] != TektonHubType {
+		return nil, fmt.Errorf("type param must be artifact or tekton")
+	}
```
chuangw6, line 149: I feel like we might be able to extract a helper function from `ValidateParams` and `Resolve` regarding params i.e. [one good example from gitresolver](https://github.com/tektoncd/pipeline/blob/7189162fd3ff5d9aa11b7090134aab5566f1cfbc/pkg/resolution/resolver/git/resolver.go#L406) and this helper function does 3 things: 1/. populate the params with default from configmap and from user-provided spec, 2/. validate param values, 3/. find missing params. 

This can bring 2 benefits to us 1/. make sure we do the 3 things mentioned above consistently throughout both `ValidateParams` and `Resolve` functions. 2/. collect all the missing params together and tell users only once the names of all missing params instead of finding one missed -> return error.

chuangw6 (commented):
Thank you @QuanZhang-William !! 👍

pkg/resolution/resolver/hub/resolver.go
```diff
@@ -174,7 +298,16 @@ func (*ResolvedHubResource) Annotations() map[string]string {
...
+	h := sha256.New()
+	h.Write(rr.Content)
+	sha256CheckSum := hex.EncodeToString(h.Sum(nil))
+
+	return &pipelinev1beta1.ConfigSource{
+		URI: rr.URL,
+		Digest: map[string]string{
+			"sha256": sha256CheckSum,
+		},
+	}
```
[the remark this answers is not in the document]
QuanZhang-William: > Just curious why we don't use https://pkg.go.dev/crypto/sha256#Sum256 here? Any difference between Sum256 and the current approach we used?

Thanks for the comments, @chuangw6! I think the difference is that it used hex encoding of the check sum. It could be easier to interpret if we want to display the value  based on the post here: https://stackoverflow.com/a/10701951.

Happy to remove the encoding part if not necessary to you  :)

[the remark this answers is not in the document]
wlynch: @chuangw6 I wouldn't worry about it too much - these are functional equivalent.

pkg/resolution/resolver/hub/resolver.go
QuanZhang-William: Good point. I agree the point that we can returning all missing values in one error message, that definitely helps users. I get the point you want to do the mutation first before validation, but I'm not sure if doing the three things in 1 helper function really helps (as I personally don't expect value population functionality in a `validate()` helper method).

I can refactor this part to 2 helper functions: `validate()` and `populateDefaultParams()`? And we can also apply the same idea in git resolver later separately.

pkg/resolution/resolver/hub/resolver.go
```diff
@@ -126,36 +160,126 @@ func (r *Resolver) Resolve(ctx context.Context, params []pipelinev1beta1.Param)
...
+// the Artifact Hub follows the semVer (i.e. <major-version>.<minor-version>.0)
+// the Tekton Hub follows the simplified semVer (i.e. <major-version>.<minor-version>)
+// for resolution request with "artifact" type, we append ".0" suffix if the input version is simplified semVer
+// for resolution request with "tekton" type, we only use <major-version>.<minor-version> part of the input if it is semVer
+func resolveVersion(version, hubType string) (string, error) {
+	semVer := strings.Split(version, ".")
+	resVer := version
+
+	if hubType == ArtifactHubType && len(semVer) == 2 {
+		resVer = version + ".0"
```
wlynch, line 272: I'm a bit concerned about doing this - i.e. if someone takes this version and tries to fetch the config, will this guarantee to always work? i.e. if artifacthub publishes `0.1`, will I *always* be able to fetch `0.1.0` and will this *always* map to `0.1`? Or could a different `0.1.0` be added after?

There's an argument that maybe the resolver shouldn't try to interpret semver and just always handle versions as-is.

pkg/resolution/resolver/hub/resolver.go
```diff
@@ -174,7 +311,16 @@ func (*ResolvedHubResource) Annotations() map[string]string {
 // Source is the source reference of the remote data that records where the remote
 // file came from including the url, digest and the entrypoint.
 func (rr *ResolvedHubResource) Source() *pipelinev1beta1.ConfigSource {
```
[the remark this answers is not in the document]
wlynch: This seems fine! 🎉

wlynch (changes requested):
M

pkg/resolution/resolver/hub/resolver_test.go
```diff
@@ -110,16 +141,167 @@ func TestValidateParamsMissing(t *testing.T) {
...
+				}
+			} else {
+				if err != nil {
+					t.Fatalf("unexpected error resolving, %v", err)
+				} else {
+					if d := cmp.Diff(tc.expectedVer, resVer); d != "" {
+						t.Fatalf("expected version '%v' but got '%v'", tc.expectedVer, resVer)
+					}
+				}
+			}
```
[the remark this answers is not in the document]
QuanZhang-William: Sure, I can use a helper method for the `tc.expectedErr != nil` branch to save a bit duplicate code 😄 . But I don't see much value to refactor the  `tc.expectedErr == nil` branch 🤔

pkg/resolution/resolver/hub/resolver.go
QuanZhang-William: Hi @wlynch, thanks for your input and this is a good question 😀 . The short answer is **yes**. If artifacthub publishes 0.1, and I will  always be able to fetch 0.1.0

The artifact hub will always convert the [resource yaml version](https://github.com/tektoncd/catalog/blob/962f9005350b1acb359558b6c5aafc420e0beb46/task/git-clone/0.8/git-clone.yaml#L6) to semVer (i.e. `0.6` -> `0.6.0`) and the Hub uses the semVer to index the corresponding resource.
 
So technically you can publish a resource with yaml version `0.6` first (which is converted to `0.6.0` in the Hub), and then publish another yaml version of `0.6.0`, it just overrides the previous `0.6.0` version in the Hub. But the API will always return the latest `0.6.0`

The main reason I put such conversion is for backward compatibility, so that current users can keep version 0.6 in the resolver, and the type is default to `artifact`

QuanZhang-William:
> 

🤔

pkg/resolution/resolver/hub/resolver_test.go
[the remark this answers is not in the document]
chuangw6: SGTM!

pkg/resolution/resolver/hub/resolver.go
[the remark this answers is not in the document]
chuangw6: SG! Keeping it as is makes sense to me! Thanks @QuanZhang-William @wlynch !

pkg/resolution/resolver/hub/resolver.go
chuangw6: SGTM! 2 helpers will make things even more clear. 👍 

echo what you said - I couldn't tell that the populateDefaultParams in gitresolver does two jobs until I saw the implementation returns validation error.

abayer: > echo what you said - I couldn't tell that the populateDefaultParams in gitresolver does two jobs until I saw the implementation also returns validation error.

Yeah, that's my bad - I'm prone to seeing code that does much of the work for two different things and just combining the code a bit more often than I should be. =)

chuangw6: @abayer NOT A PROBLEM AT ALL! It's still a very good example for how to reuse code for other resolvers 👍 !!! 😄

pkg/resolution/resolver/hub/resolver.go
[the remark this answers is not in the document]
QuanZhang-William: Cool! @chuangw6 , added the rr.URL value, PTAL.

docs/hub-resolver.md
```diff
@@ -26,25 +29,42 @@ for the name, namespace and defaults that the resolver ships with.
...
 
-| Option Name       | Description                                          | Example Values     |
-|-------------------|------------------------------------------------------|--------------------|
-| `default-catalog` | The default catalog from where to pull the resource. | `tekton`           |
-| `default-kind`    | The default object kind for references.              | `task`, `pipeline` |
+| Option Name       | Description                                          | Example Values         |
+|-------------------|------------------------------------------------------|------------------------|
+| `default-catalog` | The default catalog from where to pull the resource. | `tekton-catalog-tasks` |
+| `default-kind`    | The default object kind for references.              | `task`, `pipeline`     |
+| `default-type`    | The default hub from where to pull the resource.     | `artifact`, `tekton`   |
```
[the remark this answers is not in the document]
QuanZhang-William: Good catch, fixed!

pkg/resolution/resolver/hub/resolver.go
QuanZhang-William: @abayer, @chuangw6. I refactored the code based on discussion here, PTAL 😁

QuanZhang-William:
@jerop @abayer @wlynch @chuangw6 . Could you please review the code again. Thanks 😁

test/resolvers_test.go
```diff
@@ -190,7 +190,7 @@ spec:
 	if err := WaitForPipelineRunState(ctx, c, prName, timeout,
 		Chain(
 			FailedWithReason(pod.ReasonCouldntGetTask, prName),
-			FailedWithMessage("requested resource 'https://api.hub.tekton.dev/v1/resource/Tekton/task/git-clone-this-does-not-exist/0.7/yaml' not found on hub", prName),
+			FailedWithMessage("requested resource 'https://artifacthub.io/api/v1/packages/tekton-task/tekton-catalog-tasks/git-clone-this-does-not-exist/0.7.0' not found on hub", prName),
```
jerop, line 193: instead of changing this test, can we leave it as is to use the tekton hub, and add `type: tekton` to the test case?

then we can add another test case that defaults to artifact hub? 

want to make sure that both scenarios are covered in the end-to-end tests

jerop: ah I see, how soon are we planning to remove it? so users can use hub resolver with their own tekton hub instances?

jerop:
/approve

cc @abayer

test/resolvers_test.go
QuanZhang-William: Yeah, exactly! Based on the current progress (https://github.com/tektoncd/hub/issues/667), I think we can deprecate `https://api.hub.tekton.dev` in late Dec 2022 or early 2023. And yes, user must configure their own instance of Tekton Hub to use `tekton type` 😄

pkg/resolution/resolver/hub/resolver.go
wlynch: I'd lean towards dropping this for the new artifacthub impl:

1. semver with a missing minor/patch version generally uses the latest version matching the prefix (i.e. the golang 1.19 docker image refers to the latest patch version 1.19.2) - I'm not sure we want to have the resolver implement this kind of lookup behavior.
2. I'm worried about version skew for ArtifactHub URLs - e.g. https://artifacthub.io/packages/tekton-task/tekton-catalog-tasks/git-clone/0.8.0 (even though this points to the 0.8 git tag), but https://artifacthub.io/packages/tekton-task/tekton-catalog-tasks/git-clone/0.8 404s.

It'll probably be simpler long term for the resolver to interpret the user provided version tag as-is, so it always matches the upstream state (up to you if you want to preserve the old behavior for compatibility).

wlynch:
Would it be worth adding docs around configuring the different hub impls and the differences between them?

pkg/resolution/resolver/hub/resolver.go
QuanZhang-William: For 2, I don't think there is version skew for the Artifact Hub URL since all the valid URLs will use semVer no matter how do you tag the repo, and URLs like https://artifacthub.io/packages/tekton-task/tekton-catalog-tasks/git-clone/0.8 will always return 404s.

QuanZhang-William: Thoughts? @jerop @abayer @vdemeester?

cmd/resolvers/main.go
```diff
@@ -33,21 +33,26 @@ import (
...
+
+func buildHubURL(configAPI, defaultURL, yamlEndpoint string) string {
+	var hubURL string
+	if configAPI == "" {
+		hubURL = defaultURL
+	} else {
+		if !strings.HasSuffix(configAPI, "/") {
+			configAPI += "/"
+		}
+		hubURL = configAPI + yamlEndpoint
```
wlynch, line 54: non-blocking: When Tekton bumps to Go 1.19, you can replace this with https://pkg.go.dev/net/url#JoinPath 😎

wlynch (approved):
/lgtm

PR #5580, opened 2022-09-28
Add Provenance field in TaskRun&PipelineRun status

chuangw6:
# Changes

/kind feature

resolve https://github.com/tektoncd/pipeline/issues/5550

**Change 1:** Add a `Provenance` field in TaskRun&PipelineRun status. This field
currently only contains a subfield named `ConfigSource`, but can be extended later to
have more provenance-related fields.

**Change 2:** Prior, https://github.com/tektoncd/pipeline/pull/5551 introduced the `ConfigSource` to api/resolution alpha & beta 
package. In this PR, we moved the `ConfigSource` to api/pipeline alpha & beta package so that 
the `Provenance` struct can reuse the `ConfigSource` type (Otherwise, we cannot import it from the api/resolution alpha because of the import cycle).

**Why:** See the motivation and discussions in https://github.com/tektoncd/pipeline/issues/5550.
The tl;dr is that it helps pass provenance-related data in a more structured way
ConfigSource is one example.

Signed-off-by: Chuang Wang <chuangw@google.com>

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [ ] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ ] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
Add Provenance field in TaskRun&PipelineRun status that wraps all the information we might need from pipeline side.

It only contains ConfigSource at the moment, but it can be extended to have more subfields in future.
```

PR #5670, opened 2022-10-20
Feature flag for provenance field in status

chuangw6:
# Changes

/kind feature 

(*This is [an additive change](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#additive-changes)*)

Introduced a dedicated feature flag of boolean type named "enable-provenance-in-status"
in feature-flags configmap to enable the provenance field in status. The `provenance` field 
was introducted to *run status in https://github.com/tektoncd/pipeline/pull/5580 to record authenticated metadata about how a 
software artifact was built i.e. the source where remote resource came from.

By default, this feature flag is false.

***Related PRs***:
- #5580 
- #5397 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Added a new boolean feature flag named "enable-provenance-in-status" in feature-flags configmap to enable the provenance field in status to be populated. This field in status aims to record authenticated metadata about how a software artifact was built i.e. the source where remote resource came from.
```

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -422,6 +422,11 @@ features](#alpha-features) to be used.
 
 - `resource-verification-mode`: Setting this flag to "enforce" will enforce verification of tasks/pipeline. Failing to verify will fail the taskrun/pipelinerun. "warn" will only log the err message and "skip" will skip the whole verification.
 
+- `enable-provenance-in-status`: set this flag to "true" to enable recording 
+  the `provenance` field in `TaskRun` and `PipelineRun` status. The `provenance` 
+  field contains metadata about resources used in the TaskRun/PipelineRun such as the 
+  source from where a remote Task/Pipeline definition was fetched.
+
 For example:
 
 ```yaml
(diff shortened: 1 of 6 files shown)


JeromeJu:
Wondering if this would be better to be part of #5580?

chuangw6:
> Thanks Chuang! Wondering if this would be better to be part of #5580?

Good point. I didn't include in that PR for 2 reasons: 1/ tried to make that PR as smaller as possible 2/ wanted to add the feature flag when the code is actually trying to populate the field with data. But it turns out the PR doing that becomes large as well, so I just created this PR to add feature flag only to make review a bit easier for reviewers :D

JeromeJu:
Sounds good. Also, LGTM.

> > Thanks Chuang! Wondering if this would be better to be part of #5580?
> 
> Good point. I didn't include in that PR for 2 reasons: 1/ tried to make that PR as smaller as possible 2/ wanted to add the feature flag when the code is actually trying to populate the field with data. But it turns out the PR doing that is quite large already as well, so I just created this PR to add feature flag only to make review a bit easier for reviewers :D

chuangw6:
Hey @jerop @afrittoli @dibyom, 

I've added this dedicated feature flag for the provenance field in status that was recently introduced in #5580. PTAL. Thank you!!!

docs/install.md
```diff
@@ -442,6 +442,11 @@ features](#alpha-features) to be used.
   name, kind, and API version information for each `TaskRun` and `Run` in the `PipelineRun` instead. Set it to "both" to
   do both. For more information, see [Configuring usage of `TaskRun` and `Run` embedded statuses](pipelineruns.md#configuring-usage-of-taskrun-and-run-embedded-statuses).
 
+- `enable-provenance-in-status`: set this flag to "true" to enable recording 
+  the `provenance` field in `TaskRun` and `PipelineRun` status. The `provenance` field contains 
+  the key authenticated metadata about how a software artifact was built i.e. the source
+  where remote resource came from.
+
```
jerop, line 449: we should also list this feature among the alpha features with its individual feature flag - https://github.com/tektoncd/pipeline/blob/main/docs/install.md#alpha-features

jerop:
we shouldn't release v0.41 with https://github.com/tektoncd/pipeline/pull/5580 without the feature flag which is added in this PR, so adding this PR to the mileston -- cc @tektoncd/core-maintainers

dibyom:
#5580 just adds the field - I don't think its actually being used yet

jerop:
> #5580 just adds the field - I don't think its actually being used yet

ah that's good to know, the conversation above btwn @chuangw6 and @JeromeJu made me think it was, in that case we don't need to have it in the milestone

docs/install.md
chuangw6: Thanks @jerop ! Good idea. I've added this to the table. PTAL!

chuangw6: I've also updated the paragraph above the alpha features table. PTAL. Thank you very much!

cc @dibyom

chuangw6:
/hold

make sure this PR and #5397 are in the same release.

chuangw6:
/unhold 

b/c https://github.com/tektoncd/pipeline/releases/tag/v0.41.0 was cut.

docs/pipelineruns.md
```diff
@@ -1386,6 +1386,7 @@ Your `PipelineRun`'s `status` field can contain the following fields:
     - [`kind`][kubernetes-overview] - Generally either `TaskRun` or `Run`.
     - [`apiVersion`][kubernetes-overview] - The API version for the underlying `TaskRun` or `Run`.
     - [`whenExpressions`](pipelines.md#guard-task-execution-using-when-expressions) - The list of when expressions guarding the execution of this task.
+  - `provenance` - A list of fields recording the key authenticated metadata about how a software artifact was built i.e. the source where remote resource came from.
```
dibyom, line 1389: Nit: provenance is not a list

How about something like  -  Metadata about resources used in the PipelineRun such as the source from where a remote pipeline definition was fetched

dibyom (approved):
one minor thing about the docs, otherwise LGTM

docs/pipelineruns.md
chuangw6: SGTM. Updated. Thanks for the suggestion here. @dibyom

vdemeester (approved):

docs/pipelineruns.md
dibyom: Can we also update the descriptions for provenance in install.md and the feature flag comment itself to match?

chuangw6: Sounds good. Updated! Thanks @dibyom !

Issue #5138, opened 2022-07-14
API: References and Specifications


jerop:
Tekton Pipelines has the following fields:
- `TaskRef` - references to `Tasks` and `Custom Tasks`
- `TaskSpec` - embedded specifications of `Tasks` and `CustomTasks`
- `PipelineRef` - references to `Pipelines`
- `PipelineSpec` - embedded specifications of `Pipelines`

We have discussed replacing them with `ref` and `spec`, especially when we discussed supporting `Pipelines` in `Pipelines`. We descoped this work from the TEPs where it came up, so opening this issue to track this discussion and gather more feedback.

Below is the most promising API for using `ref` and `spec` - thanks to @afrittoli!

This example below demonstrates the API change to `Pipelines`, but the same can be applied to other types.

```yaml
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: clone-scan-notify
spec:
  tasks:

    # task using reference
    - name: pt-0
      ref:
         apiVersion: tekton.dev/v1beta1
         kind: Task
         name: git-clone

    # task using specification
    - name: pt-1
      spec:
         apiVersion: tekton.dev/v1beta1
         kind: Task
         spec:
           params: (...)
           steps:
             - (...)

    # pipeline using reference
    - name: pt-2
      ref:
         apiVersion: tekton.dev/v1beta1
         kind: Pipeline
         name: golang-pipeline

    # pipeline using specification
    - name: pt-3
      spec:
         apiVersion: tekton.dev/v1beta1
         kind: Pipeline
         spec:
           params: (...)
           steps:
             - (...)

    # custom task using reference
    - name: pt-4
      ref:
         apiVersion: example.dev/v1
         kind: Example
         name: example-pipeline

    # custom task using specification
    - name: pt-5
      spec:
         apiVersion: example.dev/v1
         kind: Example
         spec:
           params: (...)
           steps:
             - (...)
             
    # anything using remote resolution (kind is specified in resource field)
    - name: pt-6
      ref:
        resolver: git
        resource: (...)
```

Notes
- We could default `apiVersion` such that it matches the top level value (such that users have to specify it for Custom Tasks only)
- `ChildReferences` already takes this shape - where it lists the references with `apiVersion` and `kind` - so this change will align the specifications with the status
- We can support both the existing API and this API while we migrate to this new version, but we should go into V1 with this API if we decide this is the direction we want to go

/kind feature
/area api

pritidesai:
[...]

PR #5677, opened 2022-10-24
TEP-0114: Custom Task Beta - User Guide

XinruZhang:
# Changes

This PR provides a documentation for how to define and implement `v1beta1.CustomRun`.

Note that the migration doc, and using `v1beta1.CustomRun` within a `PipelineRun` is not included. Those docs will be introduced after the `PipelineRun controller` have a better support for migration.

/kind documentation

# Submitter Checklist

- [x] ~Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed~
  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/customruns.md
+++ b/docs/customruns.md
@@ -1,0 +1,469 @@
+<!--
+---
+linkTitle: "CustomRuns"
+weight: 800
+---
+-->
+
+# CustomRuns
+
+- [Overview](#overview)
+- [Configuring a `CustomRun`](#configuring-a-customrun)
+  - [Specifying the target Custom Task](#specifying-the-target-custom-task)
+  - [Cancellation](#cancellation)
+  - [Specifying `Timeout`](#specifying-timeout)
+  - [Specifying `Retries`](#specifying-retries)
+  - [Specifying Parameters](#specifying-parameters)
+  - [Specifying Workspaces](#specifying-workspaces)
+  - [Specifying Service Account](#specifying-a-serviceaccount)
+- [Monitoring execution status](#monitoring-execution-status)
+  - [Status Reporting](#status-reporting)
+  - [Monitoring `Results`](#monitoring-results)
+- [Code examples](#code-examples)
+  - [Example `CustomRun` with a referenced custom task](#example-customrun-with-a-referenced-custom-task)
+  - [Example `CustomRun` with an unnamed custom task](#example-customrun-with-an-unnamed-custom-task)
+  - [Example of specifying parameters](#example-of-specifying-parameters)
+
+# Overview
+
+A `CustomRun` allows you to instantiate and execute a [Custom
+Task](https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md),
+which can be implemented by a custom task controller running on-cluster. Custom
+Tasks can implement behavior that's independent of how Tekton TaskRun is implemented.
+
+In order for a `CustomRun` to actually execute, there must be a custom task
+controller running on the cluster that is responsible for watching and updating
+`CustomRun`s which reference their type. If no such controller is running, `CustomRun`s
+will have no `.status` value and no further action will be taken.
+
+## Configuring a `CustomRun`
+
+A `CustomRun` definition supports the following fields:
+
+- Required:
+  - [`apiVersion`][kubernetes-overview] - Specifies the API version. Currently,only
+    `tekton.dev/v1beta1` is supported.
+  - [`kind`][kubernetes-overview] - Identifies this resource object as a `CustomRun` object.
+  - [`metadata`][kubernetes-overview] - Specifies the metadata that uniquely identifies the
+    `CustomRun`, such as a `name`.
+  - [`spec`][kubernetes-overview] - Specifies the configuration for the `CustomRun`.
+    - [`customRef`](#1-specifying-the-target-custom-task-with-customref) - Specifies the type and
+      (optionally) name of the custom task type to execute.
+    - [`customSpec`](#2-specifying-the-target-custom-task-by-embedding-its-spec) - Embed the custom task resource spec
+      directly in a `CustomRun`.
+- Optional:
+  - [`timeout`](#specifying-timeout) - specifies the maximum duration of a single execution
+    of a `CustomRun`.
+  - [`retries`](#specifying-retries) - specifies the number of retries to execute upon `CustomRun`failure.
+  - [`params`](#specifying-parameters) - Specifies the desired execution
+    parameters for the custom task.
+  - [`serviceAccountName`](#specifying-a-serviceaccount) - Specifies a `ServiceAccount`
+    object for executing the `CustomRun`.
+  - [`workspaces`](#specifying-workspaces) - Specifies the physical volumes to use for the
+    [`Workspaces`](workspaces.md) required by a custom task.
+
+[kubernetes-overview]:
+  https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#required-fields
+
+### Specifying the target Custom Task
+
+A custom task resource's `CustomSpec` may be directly embedded in the `CustomRun` or it may
+be referred to by a `CustomRef`. But, not both at the same time.
+
+1. [Specifying the target Custom Task with customRef](#specifying-the-target-custom-task-with-customref)
+   Referring a custom task (i.e. `CustomRef` ) promotes reuse of custom task definitions.
+
+2. [Specifying the target Custom Task by embedding its spec](#specifying-the-target-custom-task-by-embedding-its-customspec)
+   Embedding a custom task (i.e. `CustomSpec` ) helps in avoiding name collisions with other users within the same namespace.
+   Additionally, in a pipeline with multiple embedded custom tasks, the details of entire pipeline can be fetched in a
+   single API request.
+
+#### Specifying the target Custom Task with customRef
+
+To specify the custom task type you want to execute in your `CustomRun`, use the
+`customRef` field as shown below:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: CustomRun
+metadata:
+  name: my-example-run
+spec:
+  customRef:
+    apiVersion: example.dev/v1beta1
+    kind: MyCustomKind
+  params:
+  - name: duration
+    value: 10s
+```
+
+When this `CustomRun` is created, the Custom Task controller responsible for
+reconciling objects of kind "MyCustomKind" in the "example.dev/v1beta1" api group
+will execute it based on the input params.
+
+You can also specify the `name` and optional `namespace` (default is `default`)
+of a custom task resource object previously defined in the cluster.
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: CustomRun
+metadata:
+  name: my-example-run
+spec:
+  customRef:
+    apiVersion: example.dev/v1beta1
+    kind: Example
+    name: an-existing-example-task
+```
+
+If the `customRef` specifies a name, the custom task controller should look up the
+`Example` resource with that name, and use that object to configure the
+execution.
+
+If the `customRef` does not specify a name, the custom task controller might support
+some default behavior for executing unnamed tasks.
+
+In either case, if the named resource cannot be found, or if unnamed tasks are
+not supported, the custom task controller should update the `CustomRun`'s status to
+indicate the error.
+
+####  Specifying the target Custom Task by embedding its customSpec
+
+To specify the custom task spec, it can be embedded directly into a
+`CustomRun`'s spec as shown below:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: CustomRun
+metadata:
+  name: embedded-run
+spec:
+  customSpec:
+    apiVersion: example.dev/v1beta1
+    kind: Example
+    spec:
+      field1: value1
+      field2: value2
+```
+
+When this `CustomRun` is created, the custom task controller responsible for
+reconciling objects of kind `Example` in the `example.dev` api group will
+execute it.
+
+#### Developer guide for custom controllers supporting `customSpec`.
+
+1. A custom controller may or may not support a `Spec`. In cases where it is
+   not supported the custom controller should respond with proper validation
+   error.
+
+2. Validation of the fields of the custom task is delegated to the custom
+   task controller. It is recommended to implement validations as asynchronous
+   (i.e. at reconcile time), rather than part of the webhook. Using a webhook
+   for validation is problematic because, it is not possible to filter custom
+   task resource objects before validation step, as a result each custom task
+   resource has to undergo validation by all the installed custom task
+   controllers.
+
+3. A custom task may have an empty spec, but cannot have an empty
+   `ApiVersion` and `Kind`. Custom task controllers should handle
+   an empty spec, either with a default behaviour, in a case no default
+   behaviour is supported then, appropriate validation error should be
+   updated to the `CustomRun`'s status.
+
+### Cancellation
+
+The custom task is responsible for implementing `cancellation` to support pipelineRun level `timeouts` and `cancellation`. If the Custom Task implementor does not support cancellation via `.spec.status`, `Pipeline` **can not** timeout within the specified interval/duration and **can not** be cancelled as expected upon request.
+
+Pipeline Controller sets the `spec.Status` and `spec.StatusMessage` to signal `CustomRuns` about the `Cancellation`, while `CustomRun` controller updates its `status.conditions` as following once noticed the change on `spec.Status`.
+
+```yaml
+status
+  conditions:
+  - type: Succeeded
+    status: False
+    reason: CustomRunCancelled
+```
+
+### Specifying `Timeout`
+
+A custom task specification can be created with `Timeout` as follows:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: CustomRun
+metadata:
+  generateName: simpleexample
+spec:
+  timeout: 10s # set timeouts here.
+  params:
+    - name: searching
+      value: the purpose of my existence
+  customRef:
+    apiVersion: custom.tekton.dev/v1alpha1
+    kind: Example
+    name: exampleName
+```
+
+Supporting timeouts is optional but recommended.
+
+#### Developer guide for custom controllers supporting `Timeout`
+
+1. Tekton controllers will never directly update the status of the `CustomRun`, 
+   it is the responsibility of the custom task controller to support timeout.
+   If timeout is not supported, it's the responsibility of the custom task
+   controller to reject `CustomRun`s that specify a timeout value.
+2. When `CustomRun.Spec.Status` is updated to `RunCancelled`, the custom task controller
+   MUST cancel the `CustomRun`. Otherwise, pipeline-level `Timeout` and
+   `Cancellation` won't work for the Custom Task.
+3. A Custom Task controller can watch for this status update
+   (i.e. `CustomRun.Spec.Status == RunCancelled`) and or `CustomRun.HasTimedOut()`
+   and take any corresponding actions (i.e. a clean up e.g., cancel a cloud build,
+   stop the waiting timer, tear down the approval listener). 
+4. Once resources or timers are cleaned up, while it is **REQUIRED** to set a
+   `conditions` on the `CustomRun`'s `status` of `Succeeded/False` with an optional
+   `Reason` of `CustomRunTimedOut`.
+5. `Timeout` is specified for each `retry attempt` instead of all `retries`.
+
+### Specifying `Retries`
+
+A custom task specification can be created with `Retries` as follows:
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: CustomRun
+metadata:
+  generateName: simpleexample
+spec:
+  retries: 3 # set retries
+  params:
+    - name: searching
+      value: the purpose of my existence
+  ref:
+    apiVersion: custom.tekton.dev/v1alpha1
+    kind: Example
+    name: exampleName
+```
+
+Supporting retries is optional but recommended.
+
+#### Developer guide for custom controllers supporting `retries`
+
+1. Teton controller only depends on `ConditionSucceeded` to determine the 
+   termination status of a `CustomRun`, therefore Custom task implementors
+   MUST NOT set `ConditionSucceeded` to `False` until all retries are exhausted.
+2. Those custom tasks who do not wish to support retry, can simply ignore it.
+3. It is recommended, that custom task should update the field `RetriesStatus`
+   of a `CustomRun` on each retry performed by the custom task.
+4. Tekton controller does not validate that number of entries in `RetriesStatus`
+   is same as specified value of retries count.
+
+### Specifying `Parameters`
+
+If a custom task supports [`parameters`](tasks.md#parameters), you can use the
+`params` field in the `CustomRun` to specify their values:
+
+```yaml
+spec:
+  params:
+    - name: my-param
+      value: chicken
+```
+
+If the custom task controller knows how to interpret the parameter value, it
+will do so. It might enforce that some parameter values must be specified, or
+reject unknown parameter values.
+
+### Specifying workspaces
+
+If the custom task supports it, you can provide [`Workspaces`](workspaces.md) to share data with the custom task .
+
+```yaml
+spec:
+  workspaces:
+    - name: my-workspace
+      emptyDir: {}
+```
+
+Consult the documentation of the custom task that you are using to determine whether it supports workspaces and how to name them.
+
+### Specifying a ServiceAccount
+
+If the custom task supports it, you can execute the `CustomRun` with a specific set of credentials by
+specifying a `ServiceAccount` object name in the `serviceAccountName` field in your `CustomRun`
+definition. If you do not explicitly specify this, the `CustomRun` executes with the service account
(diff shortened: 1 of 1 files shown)


docs/migrating-v1alpha1.Run-to-v1beta1.CustomRun.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Changes to fields
+
+Comparing `v1alpha1.Run` with `v1beta1.CustomRun`, the following fields have been changed:
+
+| Old field              | New field         |
+| ---------------------- | ------------------|
+| `spec.spec`            | [`spec.customSpec`](#changes-to-the-specification) |
+| `spec.ref`             | [`spec.customRef`](#changes-to-the-reference) |
+| `spec.podTemplate`     | removed |
```
jerop, line 20: please add section showing that users can support `podtemplate` in their own custom task spec, and reference that section in this table

for example:

```yaml
kind: CustomRun
metadata:
  name: customrun-with-podtemplate
spec:
  customSpec:
    apiVersion: example.dev/v1beta1
    kind: Example
    spec:
      podTemplate: 
        securityContext:
        runAsUser: 1001


```

docs/pipelines.md
```diff
@@ -1564,21 +1564,20 @@ no `runAfter` can be specified in `finally` tasks.
...
+
 **Note: This is only allowed if `enable-custom-tasks` is set to
 `"true"` in the `feature-flags` configmap, see [`install.md`](./install.md#customizing-the-pipelines-controller-behavior)**
 
 [Custom Tasks](https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md)
 can implement behavior that doesn't correspond directly to running a workload in a `Pod` on the cluster.
 For example, a custom task might execute some operation outside of the cluster and wait for its execution to complete.
 
-A PipelineRun starts a custom task by creating a [`Run`](https://github.com/tektoncd/pipeline/blob/main/docs/runs.md) instead of a `TaskRun`.
+A PipelineRun starts a custom task by creating a [`CustomRun`](https://github.com/tektoncd/pipeline/blob/main/docs/customruns.md) instead of a `TaskRun`.
```
jerop, line 1576: we need to rename runs.md to customruns.md for this link to work

docs/pipelines.md
```diff
@@ -1721,9 +1720,32 @@ Pipelines do not support the following items with custom tasks:
...
+|:---|:--- |
+| [Task Loops][task-loops]| Runs a `Task` in a loop with varying `Parameter` values. 
+| [Pipeline Loops][pipeline-loops]| Runs a `Pipeline` in a loop with varying `Parameter` values.
+| [Common Expression Language][cel]| Provides Common Expression Language support in Tekton Pipelines.
+| [Wait][wait]| Waits a given amount of time, specified by a `Parameter` named "duration", before succeeding.
+| [Approvals][approvals]| Pauses the execution of `PipelineRuns` and waits for manual approvals.
+| [Pipelines in Pipelines][pipelines-in-pipelines]| Defines and executes a `Pipeline` in a `Pipeline`. 
+| [Task Group][task-group]| Groups `Tasks` together as a `Task`.
+| [Pipeline in a Pod][pipeline-in-pod]| Runs `Pipeline` in a `Pod`.
+
```
jerop, line 1735: Let's add a note at the end encouraging implementors of custom tasks to add their custom tasks to this table

XinruZhang: NP, done!

docs/pipelines.md
XinruZhang: Ops, updated!

docs/migrating-v1alpha1.Run-to-v1beta1.CustomRun.md
XinruZhang: SG, added, thanks!

vdemeester (commented):
I am probably late to the party but why is `ref` rename to `customRef` and `spec` renamed to `customeSpec` ? `Custom` is already in the name of the object (`CustomRun`), why the repetition ? 🤔

XinruZhang:
> I am probably late to the party but why is `ref` rename to `customRef` and `spec` renamed to `customeSpec` ? `Custom` is already in the name of the object (`CustomRun`), why the repetition ? 🤔

Hi Vincent, the main reason is being aligned with TaskRun and PipelineRun, their APIs on the references and specifications are:

```
kind: PipelineRun
metadata:
  name: pipelinerun-with-ref
spec:
  pipelineRef:
    name: my-pipeline
---
kind: TaskRun
metadata:
  name: taskrun-with-reference
spec:
  taskRef:
    name: my-task
---
kind: PipelineRun
metadata:
  name: pipelinerun-with-spec
spec:
  pipelineSpec:
    params:
      - name: param1
    tasks:
      - name: my-task
---
kind: TaskRun
metadata:
  name: taskrun-with-spec
spec:
  taskSpec:
    params:
      - name: param1
    steps:
      - name: my-step
```

Worthing mentioning here: simplifying all the references to use ref and embedded specifications to use spec is tracked in https://github.com/tektoncd/pipeline/issues/5138.

LMK if you have further concern!

docs/pipelines.md
```diff
@@ -1721,9 +1720,34 @@ Pipelines do not support the following items with custom tasks:
 
 For a better understanding of `Pipelines`, study [our code examples](https://github.com/tektoncd/pipeline/tree/main/examples).
 
+## Known Custom Tasks
+
+| Custom Task | Description |
+|:---|:--- |
+| [Task Loops][task-loops]| Runs a `Task` in a loop with varying `Parameter` values.
```
lbernick, line 1727: we have matrix now, so task-loops probably isn't very useful

docs/pipelines.md
```diff
@@ -1721,9 +1720,34 @@ Pipelines do not support the following items with custom tasks:
 
 For a better understanding of `Pipelines`, study [our code examples](https://github.com/tektoncd/pipeline/tree/main/examples).
 
+## Known Custom Tasks
```
lbernick, line 1723: sort of unfortunate that these all implement v1alpha1.Run... maybe we could have one section for "old" custom tasks implementing v1alpha1.Run and one (empty) section for those implementing customrun?

lbernick (commented):
thanks Xinru, this looks great! the "fix ####" in the PR description will close the linked bug when this is merged but I think we probably don't want to close the bug until we have at least an example or two.

XinruZhang (commented):
Thanks for your review @lbernick!

As for the example, do you mean we want to implement a controller for `v1beta1.CustomRun` as an example or we need to create example yamls under the `examples` folder?

lbernick:
> As for the example, do you mean we want to implement a controller for `v1beta1.CustomRun` as an example or we need to create example yamls under the `examples` folder?

I think we'll need to implement a v1beta1 controller for testing, just like the existing wait task one we already have for v1alpha1. we can list that as an example.

XinruZhang:
Created a new doc `customruns.md` to illustrate how to define and implement Custom Task Beta. PTAL. @jerop @abayer

docs/customruns.md
```diff
@@ -0,0 +1,419 @@
...
+If the custom task controller knows how to interpret the parameter value, it
+will do so. It might enforce that some parameter values must be specified, or
+reject unknown parameter values.
+
+### Specifying Workspaces and Service Account
+
+A `CustomRun` object can specify workspaces and a service account name.
+These can be used to share data, or specify the identity to run the `CustomRun`.
+Not all custom tasks will support these values.
+Consult the documentation of the custom task that you are using to determine whether these values apply.
```
jerop, line 273: why are these two grouped together? seems that they should be their own sections 🤔

docs/customruns.md
```diff
@@ -0,0 +1,419 @@
...
+```
+
+The following tables shows how to read the overall status of a `CustomRun`:
+
+`status`|Description
+:-------|-----------:
+<unset>|The custom task controller has not taken any action on the Run.
+Unknown|The custom task controller has started execution and the Run is ongoing.
+True|The CustomRun completed successfully.
+False|The CustomRun completed unsuccessfully.
```
jerop, line 331: clarify here as well that all retries must be completed

docs/customruns.md
XinruZhang: There were another `podTemplate` grouped with these two, I deleted it. You are right, we should split them~ Updated.

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+After the `CustomRun` completes, the custom task controller can report output
+values in the `results` field:
+
+```
+results:
+- name: my-result
+  value: chicken
+```
+
+## Code examples
```
jerop, line 341: @XinruZhang it'd be helpful to implement CustomRun for the Wait Custom Task and reference them in the code examples, but that could be done in another change

jerop (approved):
Thank you @XinruZhang!

Planning to add the migration guide in another PR?

docs/customruns.md
XinruZhang: Agreed, I created an issue to track the example controller implementation: https://github.com/tektoncd/pipeline/issues/5754.

XinruZhang:
@jerop  The migration plan was removed intentionally.

I'm thinking of adding the Migration Plan (including the feature flag story) after Custom Task Beta is supported in Pipeline, because It doesn't make sense to `v1alpha1.Run` custom task implementors to migrate to `v1beta1.CustomRun` while defining inline Custom Task in pipeline is not supported.

Wdyt?

docs/customruns.md
jerop: sounds great!

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+   `Reason` of `CustomRunTimedOut`.
+6. `Timeout` is specified for each `retry attempt` instead of all `retries`.
+
+### Specifying `Retries`
+
+A custom task specification can be created with `Retries` as follows:
+
+```yaml
+apiVersion: tekton.dev/v1alpha1
+kind: Run
```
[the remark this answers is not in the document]
XinruZhang: Good catch! Thanks Yongxuan

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+Note: Custom Task implementors **MUST NOT** update the value of `spec.StatusMessage` and `spec.Status`, because Pipeline Controller depends the two fields.
+
+### Specifying `Timeout`
+
+A custom task specification can be created with `Timeout` as follows:
+
+```yaml
+apiVersion: tekton.dev/v1alpha1
+kind: Run
+metadata:
```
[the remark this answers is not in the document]
XinruZhang: Updated

Yongxuanzhang:
any difference we have between v1beta1 customrun and v1alpha1 run? 
If user wants to update run->customrun, any changes they need to make?

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+  - [Example `CustomRun` with an unnamed custom task](#example-customrun-with-an-unnamed-custom-task)
+  - [Example of specifying parameters](#example-of-specifying-parameters)
+
+# Overview
+
+A `CustomRun` allows you to instantiate and execute a [Custom
+Task](https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md),
+which can be implemented by a custom task controller running on-cluster. Custom
+Tasks can implement behavior that doesn't correspond directly to running a
+workload in a `Pod` on the cluster. For Pod-based on-cluster workloads, you
```
lbernick, line 32: Is there any way we can reword these guidelines around when to use a TaskRun vs CustomRun that don't focus on implementation details? For example, in the buildkit implementation of Tekton, TaskRuns don't run in a pod. (Maybe focusing on containers would be OK though.)

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+controller running on the cluster that is responsible for watching and updating
+`CustomRun`s which reference their type. If no such controller is running, `CustomRun`s
+will have no `.status` value and no further action will be taken.
+
+## Configuring a `CustomRun`
+
+A `CustomRun` definition supports the following fields:
+
+- Required:
+  - [`apiVersion`][kubernetes-overview] - Specifies the API version, for example
```
lbernick, line 45:
```suggestion
  - [`apiVersion`][kubernetes-overview] - Specifies the API version. Currently, only `tekton.dev/v1beta1` is supported.
```

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+    `CustomRun`, such as a `name`.
+  - [`spec`][kubernetes-overview] - Specifies the configuration for the `CustomRun`.
+    - [`customRef`](#1-specifying-the-target-custom-task-with-customref) - Specifies the type and
+      (optionally) name of the custom task type to execute.
+    - [`customSpec`](#2-specifying-the-target-custom-task-by-embedding-its-spec) - Embed the custom task resource spec
+      directly in a `CustomRun`.
+- Optional:
+  - [`timeout`](#specifying-timeout) - specifies the maximum duration of a single execution
+    of a `CustomRun`.
+  - [`retries`](#specifying-retries) - specifies the number of retures upon `CustomRun`failure.
```
lbernick, line 58:
```suggestion
  - [`retries`](#specifying-retries) - specifies the number of retries to execute upon `CustomRun` failure.
```

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+  - [`params`](#specifying-parameters) - Specifies the desired execution
+    parameters for the custom task.
+  - [`serviceAccountName`](#specifying-a-serviceaccount) - Specifies a `ServiceAccount`
+    object that provides custom credentials for executing the `CustomRun`.
```
lbernick, line 62: I'd suggest removing "provides custom credentials" since service accounts get permissions to do things via RoleBindings and ClusterRoleBindings or workload identity.

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+
+To specify the custom task type you want to execute in your `CustomRun`, use the
+`customRef` field as shown below:
+
+```yaml
+spec:
+  customRef:
+    apiVersion: example.dev/v1alpha1
+    kind: Example
+```
```
lbernick, line 92: Would it be possible to use more realistic examples? For some custom tasks, it might make sense to define the custom task once on the cluster and reference it multiple times, and for others, creating a new CustomRun might make more sense. Also, apiVersion should be v1beta1, and I suggest including `kind: CustomRun`.

XinruZhang:
> any difference we have between v1beta1 customrun and v1alpha1 run? If user wants to update run->customrun, any changes they need to make?

Thanks @Yongxuanzhang , there are differences between `v1beta1` customrun (you can find it in TEP-0114), but we do not expect existing `v1alpha1.Run` customers migrate to `v1beta1.CustomRun` until https://github.com/tektoncd/community/pull/867 gets merged :)

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+
+This initiates the execution of a `CustomRun` of a custom task of type `Example`, in
```
lbernick, line 94: Suggest clarifying the wording here, as this sounds a bit like Tekton will initiate the execution of a customrun when it is created, rather than the custom task controller being responsible for executing it.

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+If the `customRef` specifies a name, the custom task controller should look up the
+`Example` resource with that name, and use that object to configure the
+execution.
+
+If the `customRef` does not specify a name, the custom task controller might support
+some default behavior for executing unnamed tasks.
+
+In either case, if the named resource cannot be found, or if unnamed tasks are
+not supported, the custom task controller should update the `CustomRun`'s status to
+indicate the error.
```
lbernick, line 117: Could this be separated into a "Developer Guide for supporting CustomRef section"?

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+
+3. A custom task may have an empty spec, but cannot have an empty
+   `ApiVersion` and `Kind`. Custom task controllers should handle
+   an empty spec, either with a default behaviour, in a case no default
+   behaviour is supported then, appropriate validation error should be
+   updated to the `CustomRun`'s status.
+
+### Cancellation
+
+Implementing cancellation is now `required` instead of `optional`. If the Custom Task implementor does not support cancellation via `.spec.status`, `Pipeline` timeouts and `PipelineRun` cancellation will not work properly.
```
lbernick, line 163:
```suggestion
Implementing cancellation is `required`. If the Custom Task implementor does not support cancellation via `.spec.status`, `Pipeline` timeouts and `PipelineRun` cancellation will not work properly.
```

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+
+Pipeline Controller sets the `spec.Status` and `spec.StatusMessage` to signal `CustomRuns` about the `Cancellation`, while `CustomRun` controller **MUST** update its `status.conditions` as following once noticed the change on `spec.Status`.
```
lbernick, line 165: We could probably use some background on what custom task controllers are expected to do with status in general. (probably copied from https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md#status-reporting)

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+
+| Pipeline Signal| `spec.Status` | `spec.StatusMessage` | `status.Conditions` | 
+|:--- |:--- |:--- |:---|
+|Parent `PipelineRun` is cancelled | `RunCancelled` | `CustomRun cancelled as the PipelineRun it belongs to has been cancelled. `| `Type: Succeeded`<br>`Status:False`<br>`Reason:CustomRunCancelled` |
+|`Pipeline` timeouts | `RunCancelled` | `CustomRun cancelled as the PipelineRun it belongs to has timed out.`| `Type: Succeeded`<br>`Status:False`<br>`Reason:CustomRunCancelled`|
+
+Note: Custom Task implementors **MUST NOT** update the value of `spec.StatusMessage` and `spec.Status`, because Pipeline Controller depends the two fields.
```
lbernick, line 172: this is a bit confusing-- custom task controllers are responsible for changing the status of the customrun, and shouldn't have a reason to change its spec.

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+
+Supporting timeouts is optional but recommended.
+
+#### Developer guide for custom controllers supporting `Timeout`
+
+1. Tekton controllers will never directly update the status of the `CustomRun`, 
+   it is the responsibility of the custom task controller to support timeout.
+   If timeout is not supported, it's the responsibility of the custom task
+   controller to reject `CustomRun`s that specify a timeout value.
+2. On a `pipelineRun` or `pipelineTask` timeout, the status of the
```
lbernick, line 202: I think this is an implementation detail that custom task controllers don't need to know.

docs/customruns.md
```diff
@@ -0,0 +1,413 @@
...
+2. On a `pipelineRun` or `pipelineTask` timeout, the status of the 
+   `CustomRun.Spec.Status` is updated to `RunCancelled`. The custom task controller
+   MUST cancel the `CustomRun`. Otherwise, pipeline-level `Timeout` and
+   `Cancellation` won't work for the Custom Task.
+3. A Custom Task controller can watch for this status update
+   (i.e. `CustomRun.Spec.Status == RunCancelled`) and or `CustomRun.HasTimedOut()`
+   and take any corresponding actions (i.e. a clean up e.g., cancel a cloud build,
+   stop the waiting timer, tear down the approval listener).
+4. A Custom Task author can differentiate between a `CustomRun` cancelled by an user
+   from a Run cancellation initiated by a PipelineRun by looking at the
```
lbernick, line 211: I'm not sure the custom task controller needs to know why the custom task was canceled (beyond info present in the message) or what canceled it. Is this StatusMessage part of the api?

lbernick (commented):
This documentation mixes "configuring custom tasks" and "developing custom task controllers." I think it would be helpful to write one markdown file aimed at each audience, as it may be confusing for someone just trying to use a custom task to learn about how to implement them.

docs/customruns.md
XinruZhang: Thanks Lee! It would be great if you have any example in your mind XD. Aslo @ScrapCodes Would like to hear if you have any existing Custom Task in your mind that we usually use them by referring a predefined Custom Task instead of embedded one? vice versa!

docs/customruns.md
XinruZhang: Agreed, this would be part of the doc separation, will be tracked by https://github.com/tektoncd/pipeline/issues/5775, I'd prefer to merge this PR in to get some momentum XD, but I can also be convinced that we should do it in this PR~

docs/customruns.md
XinruZhang: Updated~ Thanks!

XinruZhang (commented):
Thanks @lbernick for taking time to review! ❤️

docs/customruns.md
XinruZhang: SG, rephrased it~

docs/customruns.md
lbernick: SG!

docs/customruns.md
lbernick: I'd prefer to do it in this PR, because the docs are being introduced in this PR, and because I don't think there's anything directly blocked by these docs right now, but could be convinced otherwise.

docs/customruns.md
lbernick: I'd prefer adding it in this PR-- it's sort of confusing to have instructions about what to do with the status in case of retries or cancelation without understanding the basics of the status in the first place

docs/customruns.md
lbernick: Sorry, I should have been more clear, the field `StatusMessage` is indeed part of the API, but do we make guarantees about what that message will be? I think it's intended to be human readable and there's a reasonable chance we might want to change it. I'd just put here that "StatusMessage displays a human-readable reason for the cancelation."

docs/customruns.md
lbernick: Thinking about this comment a bit more.. Giving good examples is a bit tricky. We don't want to write something that looks like you can actually use it, because we don't have any CustomRun controllers and we don't want users copy pasting the example and getting annoyed when it doesn't work. But at the same time I think the variable naming can "clue" ways we intend users to interact with this feature.

For example, here's a wait Task run (copied from experimental repo):
```
kind: Run
metadata:
  generateName: wait-run-
spec:
  ref:
    apiVersion: example.dev/v0
    kind: Wait
  params:
  - name: duration
    value: 10s
```
Maybe our example here could be:
```
kind: CustomRun
spec:
  ref:
    apiVersion: example.dev/v0
    kind: MyCustomKind
  params:
  - name: duration
    value: 10s
```
Here, I'd imagine the user wouldn't want to specify a "name" in the custom task ref, because it's more useful to define duration on the run. (Duration is more of a runtime feature than an authoring time feature.)

The `pipeline-loop` custom task has some examples where it is useful to specify a name in the custom task ref (e.g. https://github.com/tektoncd/experimental/blob/main/pipeline-loops/examples/simplepipelineloop.yaml) and could be adapted similarly.

WDYT? This might be something we can iterate on in the future.

docs/customruns.md
XinruZhang: If I'm not mistaken, per [the Pipelines WG](https://docs.google.com/document/d/1AfJfdyd1JN2P4haBdYOxEn6SMENMgvQF9ps7iF2QSG0/edit?resourcekey=0-lm66bE68HA_WVLA6AaCh6w#bookmark=id.kzvlbp8pb72) last Tuesday, we need to have documentation ready before releasing `v1beta1.CustomTask` in `v0.42` right?

XinruZhang: I can also try my best to separate the guide before v0.42 :) But I'm not sure if that can be merged in before scheduled release 🤔

docs/customruns.md
XinruZhang: Thanks Lee @lbernick, this is a very great discussion, I'm thinking about what the biggest benefit of defining unamed CustomRun would be:

I would say it's a benefit for Custom Task authors instead of Custom Task end users:

It enables Custom Task authors reconcile on `v1beta1.CustomRun` without defining a new CRD for the custom kind `MyCustomKind`.

Imagine the following case: I'm a Custom Task end user, I want to use `PipelineLoop`. I don't think It's possible for us to use `CustomRef` without specifying a name of an existing `PipelineLoop`.

wdyt?

docs/customruns.md
XinruZhang: Makes sense! 👍 Even though Pipeline Controller relies on the `spec.StatusMessage` to distinguish the different cancellation, it's an unnecessary info for CustomRun controllers, we shouldn't guarantee what we'd put in this field and Custom Run controller shouldn't rely on this field.

docs/customruns.md
XinruZhang: By any means, this can be a document improvement as part of https://github.com/tektoncd/pipeline/issues/5775:
- For end users, clarify when to use `unnamed CustomRun` and `named CustomRun`
- For custom task authors, clarify whether they should support `unnamed CustomRun`

docs/customruns.md
XinruZhang: Thanks Lee~ @lbernick I'd prefer we make incremental updates :P

pritidesai:
Ready for review, please take another look, thanks!

docs/customruns.md
XinruZhang: SG, updated!

Issue #5679, opened 2022-10-24
Move non-API related code into internal


lbernick:
Based on this [discussion thread](https://github.com/tektoncd/pipeline/pull/5634#discussion_r997411776).
Everything in the [apis packages](https://github.com/tektoncd/pipeline/tree/main/pkg/apis) and the [client packages](https://github.com/tektoncd/pipeline/tree/main/pkg/client) should be public, and all other code should be moved into the internal folder where possible to avoid usage outside of the repo and reduce our Go API surface.

vdemeester:
> Everything in the [apis packages](https://github.com/tektoncd/pipeline/tree/main/pkg/apis) and the [client packages](https://github.com/tektoncd/pipeline/tree/main/pkg/client) should be public, and all other code should be moved into the internal folder where possible to avoid usage outside of the repo and reduce our Go API surface.

Note that today, other tektoncd component might depend on those, for example the cli, or the operator, or triggers or even chains. I think for each package we want to move we need to see if it make sense or not to be "public" (and exported) or not. 

See https://github.com/tektoncd/pipeline/pull/5681#pullrequestreview-1154364716

aThorp96:
I think this relates to https://github.com/tektoncd/pipeline/issues/10072 @vdemeester. LLM code "understanding" and vulnerability reports have almost always been, in my experience, lower quality due to internal APIs being public. Even outside of LLMs this will be very helpful, but LLMs I think raise the priority here if we want to curb false-positive vulnerability reports. 

We'll need to be conservative about which APIs we trust to be internal, since other Tekton components or third party extensions (custom resolvers, for example) may depend on some of the APIs as well. Still, chipping away at this will be helpful I think.

vdemeester:
@aThorp96 agreed. I definitely think we need to make sure (and enforce) that by default we put any new package under `internal` if there is no valid reason to make it *public*.. Moving current public ones is a bit harder (as we may broke dependencies) so we need to be careful while doing it. But I 100% agree, especially in light of recent vulnerability reports (Q1/Q2 2026).

PR #5681, opened 2022-10-24
Move several packages into internal

lbernick:
# Changes
This commit moves several packages in pkg/ into pkg/internal. This reduces the Pipelines API surface by ensuring these packages can't be imported and relied on by clients, as we only intend for pkg/apis and pkg/client to be imported outside of Pipelines. Other packages that contain exported functions used outside of pkg/ will be refactored in a later commit.

/kind cleanup

# Submitter Checklist

- n/a Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
  functionality, content, code)

# Release Notes

```release-note
Several packages have been moved into pkg/internal and can no longer be imported.
```


lbernick:
part 1 of https://github.com/tektoncd/pipeline/issues/5679

/assign @dibyom @vdemeester @wlynch

vdemeester (changes requested):
Note that today, other tektoncd component might depend on those, for example the cli, or the operator, or triggers or even chains. I think for each package we want to move we need to see if it make sense or not to be "public" (and exported) or not. 

For example, `github.com/tektoncd/pipeline/pkg/pod` is used by `results` (`tektoncd` org) in `experimental` (notifiers and task-loops) and in `buildkit-tekton` (external, in my own repositories). In `results` the code that uses it looks like : 

```go
        case pod.ReasonCouldntGetTask, pod.ReasonFailedResolution, pod.ReasonFailedValidation, pod.ReasonExceededResourceQuota, pod.ReasonExceededNodeResources, pod.ReasonCreateContainerConfigError, pod.ReasonPodCreationFailed:
                return rpb.RecordSummary_FAILURE
        case pod.ReasonPending:
                return rpb.RecordSummary_UNKNOWN
```

Those "reason" do make sense to be public, maybe in another package or not, but they are definitely useful.
In `buildkit-tekton`, the code is more "involved", for example `pod.IsContainerStep(c.Name)`, but it's still valuable to export this kind of functions. `buildkit-tekton` also uses `github.com/tektoncd/pipeline/pkg/substitution` (at least the exported ones) because it is useful to re-use (to be able to perform substitution à-la-tekton) and that package is also probably used in some custom task or resolvers I wrote. 
`github.com/tektoncd/pipeline/pkg/workspace` is used in `pipeline-in-pod` for example. `github.com/tektoncd/pipeline/pkg/list` is used in triggers, in catlin, in results, in operator, in chains, in buildkit-tekton and in some resolvers I wrote as well.

lbernick:
Thanks for pointing this out! This change was probably a bit too hasty. Is there a good way to find out which functions are in use by other projects? (I know we can look at package importers on pkg.go.dev, but doesn't tell us which functions/consts are in use.) Maybe a good place to start would be moving the reasons out of the pod package and into apis, and make the rest of the pod package internal? Going to close this out for now.

vdemeester:
> Is there a good way to find out which functions are in use by other projects? (I know we can look at package importers on pkg.go.dev, but doesn't tell us which functions/consts are in use.) 

Not that I know of. Looking into package importers on pkg.go.dev helps us a bit at least.

> Maybe a good place to start would be moving the reasons out of the pod package and into apis, and make the rest of the pod package internal? 

I think it's gonna be both. There is probably a lot of things that needs to move in `pkg/apis`, and some in very specific, non internal package. I think we need to investigate and look into this as time goes.

Issue #5522, opened 2022-09-19
Passing ref data from `ResolutionRequest` to taskrun/pipelinerun through `status.Annotations`


chuangw6:
### Feature request

Currently, remote `ResolutionRequest` carries some data in the `status.Annotations`, but this data is not used anywhere / not passed down further. 

We need to edit/add extra data in the `ResolutionRequest`'s `status.Annotations` to make sure it contains the fields (url, commit, path) required by the SLSA provenance, and finally pass this data to taskrun/pipelinerun's `status.Annotations` so that Chains can pick up the data. We need to do this for all resolvers:
- git resolver
- oci bundle resolver
- hub resolver
- cluster resolver (TBD because cluster def is not in source control)

### Use case

SLSA provenance is required to capture the source of the remote config files in `predicate.invocation.configSource` section in which three fields are required: url, revision (actual commit sha) and path in source. Therefore, we need to make these information available from resolvers given that user-provided params in resolvers might not be deterministic.  
In git resolver, for example, users can specify the branch name or tag name for the `revision` param other than commit sha. To capture the actual commit sha at the moment of resolving the resource, we need to pass down the resolved ref information and make it available in run object's status.annotations so that Chains can capture the data in the provenance.

chuangw6:
## Implementation

- Git resolver: https://github.com/tektoncd/pipeline/pull/5397

chuangw6:
/close

done!

PR #5687, opened 2022-10-25
Set ConfigSource in clusterresolver

chuangw6:
# Changes

/kind feature

Related to https://github.com/tektoncd/pipeline/issues/5522

Prior, a field named Source was introduced to ResolutionRequest status to record the source where the remote resource came from. And the individual resolvers need to implement the Source function to set the correct source value. But the method in clusterresolver returns a nil value.

Now, we return correct source value with the 3 subfields: url, digest and entrypoint
- url: [Kubernetes CRD namespace-scoped resource URI](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-uris) appended with UID.
Example: /apis/GROUP/VERSION/namespaces/NAMESPACE/RESOURCETYPE/NAME@UID.
- digest: hex-encoded sha256 checksum of the in-cluster resource
- entrypoint: ***empty*** because the path is already available in url field.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Set source value for cluster resource to link back its origin i.e. url and checksum.
```

### DIFF
--- a/docs/cluster-resolver.md
+++ b/docs/cluster-resolver.md
@@ -74,6 +74,60 @@ spec:
       value: namespace-containing-pipeline
 ```
 
+## `ResolutionRequest` Status
+`ResolutionRequest.Status.Source` field captures the source where the remote resource came from. It includes the 3 subfields: `url`, `digest` and `entrypoint`.
+- `url`: url is the unique full identifier for the resource in the cluster. It is in the format of `<resource uri>@<uid>`. Resource URI part is the namespace-scoped uri i.e. `/apis/GROUP/VERSION/namespaces/NAMESPACE/RESOURCETYPE/NAME`. See [K8s Resource URIs](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-uris) for more details.
+- `digest`: hex-encoded sha256 checksum of the content in the in-cluster resource's spec field. The reason why it's the checksum of the spec content rather than the whole object is because the metadata of in-cluster resources might be modified i.e. annotations. Therefore, the checksum of the spec content should be sufficient for source verifiers to verify if things have been changed maliciously even though the metadata is modified with good intentions.
+- `entrypoint`: ***empty*** because the path information is already available in the url field.
+
+Example:
+- TaskRun Resolution
+
+```yaml
+apiVersion: tekton.dev/v1beta1
+kind: TaskRun
+metadata:
+  name: cluster-demo
+spec:
+  taskRef:
+    resolver: cluster
+    params:
+    - name: kind
+      value: task
+    - name: name
+      value: a-simple-task
+    - name: namespace
+      value: default
+```
+
+
+- `ResolutionRequest`
+```yaml
+apiVersion: resolution.tekton.dev/v1beta1
+kind: ResolutionRequest
+metadata:
+  labels:
+    resolution.tekton.dev/type: cluster
+  name: cluster-7a04be6baa3eeedd232542036b7f3b2d
+  namespace: default
+  ownerReferences: ...
+spec:
+  params:
+  - name: kind
+    value: task
+  - name: name
+    value: a-simple-task
+  - name: namespace
+    value: default
+status:
+  annotations: ...
+  conditions: ...
+  data: xxx
+  source:
+    digest:
+      sha256: 245b1aa918434cc8195b4d4d026f2e43df09199e2ed31d4dfd9c2cbea1c7ce54
+    uri: /apis/tekton.dev/v1beta1/namespaces/default/task/a-simple-task@3b82d8c4-f89e-47ea-a49d-3be0dca4c038
+```
 ---
 
 Except as otherwise noted, the content of this page is licensed under the
--- a/pkg/resolution/resolver/cluster/resolver.go
+++ b/pkg/resolution/resolver/cluster/resolver.go
@@ -18,11 +18,14 @@ package cluster
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"strings"
 
 	resolverconfig "github.com/tektoncd/pipeline/pkg/apis/config/resolver"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	pipelinev1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	clientset "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
 	pipelineclient "github.com/tektoncd/pipeline/pkg/client/injection/client"
@@ -101,6 +104,8 @@ func (r *Resolver) Resolve(ctx context.Context, origParams []pipelinev1beta1.Par
 	}
 
 	var data []byte
+	var spec []byte
+	var uid string
 	groupVersion := pipelinev1beta1.SchemeGroupVersion.String()
 
 	switch params[KindParam] {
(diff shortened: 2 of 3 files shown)


pkg/resolution/resolver/cluster/resolver.go
```diff
@@ -122,8 +128,9 @@ func (r *Resolver) Resolve(ctx context.Context, origParams []pipelinev1beta1.Par
 			logger.Infof("failed to load pipeline %s from namespace %s: %v", params[NameParam], params[NamespaceParam], err)
 			return nil, err
 		}
+		uid = string(pipeline.UID)
 		pipeline.Kind = "Pipeline"
-		pipeline.APIVersion = "tekton.dev/v1beta1"
+		pipeline.APIVersion = groupVersion
```
XinruZhang, line 133: Out of curiosity, are we also going to support `v1` in the remote resolution after V1 release?

abayer: Yeah, we will. Basically, this should always be whatever version is used in the `PipelineRun` reconciler by default, but we should probably add a param for `apiVersion` to allow overriding that...

pkg/resolution/resolver/cluster/resolver.go
```diff
@@ -101,6 +104,8 @@ func (r *Resolver) Resolve(ctx context.Context, origParams []pipelinev1beta1.Par
 	}
 
 	var data []byte
+	var uid string
+	groupVersion := "tekton.dev/v1beta1"
```
abayer, line 108: We can probably just use `v1beta1.SchemeGroupVersion.String()` here.

abayer (changes requested):

pkg/resolution/resolver/cluster/resolver.go
chuangw6: > Out of curiosity, are we also going to support `v1` in the remote resolution after V1 release?

Yeah agreed. We will need to support v1 in remote resolution. Thanks for asking @XinruZhang .

chuangw6:
/hold

Since resources in cluster are not in version control which is desired for slsa provenance, the source value might be meaningless. We need more discussions around this.

pkg/resolution/resolver/cluster/resolver.go
chuangw6: Sounds good. I changed it in this pr https://github.com/tektoncd/pipeline/pull/5705 because I am not sure if we really need to move this PR forward. FYI: The reason why this PR might not be moved forward is that resource in cluster is not in version control which is desired for slsa provenance. This implies that setting source value for in-cluster resources might be meaningless here.

PTAL that pr. Thank you!! cc @XinruZhang Thanks!

pkg/resolution/resolver/cluster/resolver.go
XinruZhang: +1 for a new param for apiVersion, it looks like a future refactor work XD

pkg/resolution/resolver/cluster/resolver.go
```diff
@@ -183,7 +197,16 @@ func (r *ResolvedClusterResource) Annotations() map[string]string {
 // Source is the source reference of the remote data that records where the remote
 // file came from including the url, digest and the entrypoint.
 func (r ResolvedClusterResource) Source() *pipelinev1beta1.ConfigSource {
-	return nil
+	h := sha256.New()
+	h.Write(r.Content)
+	sha256CheckSum := hex.EncodeToString(h.Sum(nil))
```
Yongxuanzhang, line 202: I'm curious why using hex not base64?

chuangw6: Thanks for asking.
Just trying to follow how to calculate for hub resource, and this [discussion](https://github.com/tektoncd/pipeline/pull/5666#discussion_r1002301051) might be related.
I think those are functional equivalent, and hex approach is mainly for displaying purpose as called out [here](https://stackoverflow.com/questions/10701874/generating-the-sha-hash-of-a-string-using-golang/10701951#10701951).

Yongxuanzhang: I mean base64 encoding, hex encoding is actually base16. And it seems base64 is more efficient?
https://stackoverflow.com/questions/3183841/base64-vs-hex-for-sending-binary-content-over-the-internet-in-xml-doc#:~:text=The%20difference%20between%20Base64%20and,it's%20more%20efficient%20than%20hex.

docs/cluster-resolver.md
```diff
@@ -74,6 +74,60 @@ spec:
       value: namespace-containing-pipeline
 ```
 
+## `ResolutionRequest` Status
+`ResolutionRequest.Status.Source` field captures the source where the remote resource came from. It includes the 3 subfields: `url`, `digest` and `entrypoint`.
+- `url`: [Kubernetes CRD namespace-scoped resource URI](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-uris) appended with UID i.e. `/apis/GROUP/VERSION/namespaces/NAMESPACE/RESOURCETYPE/NAME@UID`
+- `digest`: sha256 checksum of the content in the in-cluster resource's spec field. The reason why it's the checksum of the spec content rather than the whole object is because the metadata of in-cluster resources might be modified i.e. annotations. Therefore, the checksum of the spec content should be sufficient for source verifiers to verify if things have been changed maliciously even though the metadata is modified with good intentions.
+- `entrypoint`: ***empty*** because the path information is already available in the url field.
```
chuangw6, line 81: Heyy @wlynch ,

I've changed to store the checksum of the spec of in-cluster task/pipeline as we discussed before, and also documented this here. Please take a look and please let me know if the documentation needs to be rephrased here.
Thanks!!

pkg/resolution/resolver/cluster/resolver.go
Yongxuanzhang: Oh I'm not sure which one we should use here. Or what is the standard way of doing this. Just want to discuss this

chuangw6: I think it just doesn't matter that much as long as we can compare if anything is tempered by checking digest. Given base64 can help save space, I think it's fine to use.

chuangw6: Just realized that intoto recommended the [hex-encoded value](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#DigestSet:~:text=map%20from%20algorithm%20name%20to%20lowercase%20hex%2Dencoded%20value) in the DigestSet. So we might lean on hex for now. Thanks for sharing hex v.s. base64 though. That's good to know!

pkg/resolution/resolver/cluster/resolver.go
```diff
@@ -161,9 +182,17 @@ func (r *Resolver) isDisabled(ctx context.Context) bool {
...
+	// Content is the actual resolved resource data.
+	Content []byte
+	// Spec is the data in the resolved task/pipeline CRD spec.
+	Spec []byte
+	// Name is the resolved resource name in the cluster
+	Name string
+	// Namespace is the namespace in the cluster under which the resolved resource was created.
 	Namespace string
+	// ResourceURI is in the format of namespace-scoped resources API.
+	// https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-uris
```
wlynch, line 194: nit: This isn't quite a resource URI (it's `<resource uri>@<uid>`). We should update the comment and/or name to reflect this.

wlynch (changes requested):
/lgtm

pkg/resolution/resolver/cluster/resolver.go
chuangw6: Good catch. Updated both comment and name, and also updated the clusterresolver doc. Thank you!

abayer (approved):

wlynch (approved):
/lgtm

chuangw6:
Thanks for the lgtm @wlynch.
Sorry. Just resolved conflict with main, which caused lgtm to be removed..

wlynch (commented):
/lgtm

Issue #4404, opened 2021-12-03
Refactor reconciler events package to separate events and cloudevents


lbernick:
The /pkg/reconciler/events package is [responsible](https://github.com/tektoncd/pipeline/blob/07adee706154ce7b79045dd109f1382e997b3163/pkg/reconciler/events/event.go#L45-L49) for functionality related to events and functionality related to cloudevents. However, there is a separate cloudevents sub-package in the events package, which contains test code for both cloudevents and [events](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/events/cloudevent/testing.go#L27-L28). These should either be refactored to separate responsibilities or combined into one package, avoiding circular dependencies in tests.

cc: @sbwsg 
/kind cleanup

vdemeester:
cc @afrittoli 
One of the goal here is to extract the cloudevent functionnality outside of pipeline, see [here](https://github.com/tektoncd/experimental/tree/main/cloudevents)

PR #5690, opened 2022-10-26
fix cloud event flaky unit tests by adding waitgroup to fakeclient

Yongxuanzhang:
# Changes

This commit adds waitgroup to fakeclient to avoid that some goroutines are not done when we want to collect the events.
The tests are flaky because the cloud events are sent with goroutine
but we don't wait until all goroutines done to check the events. So it
is possible that some events are not collected. The waitGroup will count when each goroutine is created and decrease the count when the goroutine is done.
This change has no impact on current code.

/kind flake

Signed-off-by: Yongxuanzhang yongxuanzhang@google.com

related issues:
https://github.com/tektoncd/pipeline/issues/5160
to reproduce this issue, adding time.Sleep(100 * time.Millisecond) to https://github.com/tektoncd/pipeline/blob/b648a5b2eaebf0ece7199dcf2bd89c65137973d5/pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go#L53  
related PR:
https://github.com/tektoncd/pipeline/pull/5313
Recent failing tests:

- https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5666/pull-tekton-pipeline-unit-tests/1582795530906374144
- https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5662/pull-tekton-pipeline-unit-tests/1583441668340715520
- https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5647/pull-tekton-pipeline-unit-tests/1583130831700889600

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fix cloud event flacky unit tests by adding EventSender
```

### DIFF
--- a/pkg/reconciler/customrun/customrun_test.go
+++ b/pkg/reconciler/customrun/customrun_test.go
@@ -28,7 +28,6 @@ import (
 	"github.com/tektoncd/pipeline/pkg/reconciler/events/cloudevent"
 	ttesting "github.com/tektoncd/pipeline/pkg/reconciler/testing"
 	"github.com/tektoncd/pipeline/test"
-	eventstest "github.com/tektoncd/pipeline/test/events"
 	"github.com/tektoncd/pipeline/test/names"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -47,6 +46,7 @@ import (
 
 func initializeCustomRunControllerAssets(t *testing.T, d test.Data) (test.Assets, func()) {
 	ctx, _ := ttesting.SetupFakeContext(t)
+	ctx = ttesting.SetupFakeCloudClientContext(ctx, d.ExpectedCloudEventCount)
 	ctx, cancel := context.WithCancel(ctx)
 	test.EnsureConfigurationConfigMapsExist(&d)
 	c, informers := test.SeedTestData(t, ctx, d)
@@ -158,8 +158,9 @@ func TestReconcile_CloudEvents(t *testing.T) {
 			customRuns := []*v1beta1.CustomRun{&customRun}
 
 			d := test.Data{
-				CustomRuns: customRuns,
-				ConfigMaps: cms,
+				CustomRuns:              customRuns,
+				ConfigMaps:              cms,
+				ExpectedCloudEventCount: len(tc.wantCloudEvents),
 			}
 			testAssets, cancel := getCustomRunController(t, d)
 			defer cancel()
@@ -187,19 +188,13 @@ func TestReconcile_CloudEvents(t *testing.T) {
 			}
 
 			ceClient := clients.CloudEvents.(cloudevent.FakeClient)
-			err = eventstest.CheckEventsUnordered(t, ceClient.Events, tc.name, tc.wantCloudEvents)
-			if err != nil {
-				t.Errorf(err.Error())
-			}
+			ceClient.CheckCloudEventsUnordered(t, tc.name, tc.wantCloudEvents)
 
 			// Try and reconcile again - expect no event
 			if err := c.Reconciler.Reconcile(testAssets.Ctx, getCustomRunName(customRun)); err != nil {
 				t.Fatal("Didn't expect an error, but got one.")
 			}
-			err = eventstest.CheckEventsUnordered(t, ceClient.Events, tc.name, []string{})
-			if err != nil {
-				t.Errorf(err.Error())
-			}
+			ceClient.CheckCloudEventsUnordered(t, tc.name, []string{})
 		})
 	}
 }
@@ -296,10 +291,7 @@ func TestReconcile_CloudEvents_Disabled(t *testing.T) {
 			}
 
 			ceClient := clients.CloudEvents.(cloudevent.FakeClient)
-			err = eventstest.CheckEventsUnordered(t, ceClient.Events, tc.name, []string{})
-			if err != nil {
-				t.Errorf(err.Error())
-			}
+			ceClient.CheckCloudEventsUnordered(t, tc.name, []string{})
 		})
 	}
 }
--- a/pkg/reconciler/events/cloudevent/cloud_event_controller.go
+++ b/pkg/reconciler/events/cloudevent/cloud_event_controller.go
@@ -145,7 +145,10 @@ func SendCloudEventWithRetries(ctx context.Context, object runtime.Object) error
 	_, isCustomRun := object.(*v1beta1.CustomRun)
 
 	wasIn := make(chan error)
+
+	ceClient.addCount()
 	go func() {
+		defer ceClient.decreaseCount()
 		wasIn <- nil
 		logger.Debugf("Sending cloudevent of type %q", event.Type())
 		// In case of Run event, check cache if cloudevent is already sent
--- a/pkg/reconciler/events/cloudevent/cloudevent.go
+++ b/pkg/reconciler/events/cloudevent/cloudevent.go
@@ -88,8 +88,15 @@ func (t TektonEventType) String() string {
 	return string(t)
 }
 
-// CEClient matches the `Client` interface from github.com/cloudevents/sdk-go/v2/cloudevents
-type CEClient cloudevents.Client
+// CEClient wraps the `Client` interface from github.com/cloudevents/sdk-go/v2/cloudevents
+// and has methods to count the cloud events being sent, those methods are for testing purposes.
+type CEClient interface {
+	cloudevents.Client
+	// addCount increments the count of events to be sent
+	addCount()
+	// decreaseCount decrements the count of events to be sent, indicating the event has been sent
+	decreaseCount()
+}
 
 // TektonCloudEventData type is used to marshal and unmarshal the payload of
 // a Tekton cloud event. It can include a TaskRun or a PipelineRun
--- a/pkg/reconciler/events/cloudevent/cloudeventclient.go
+++ b/pkg/reconciler/events/cloudevent/cloudeventclient.go
@@ -21,6 +21,9 @@ import (
 	"net/http"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
+	"github.com/cloudevents/sdk-go/v2/client"
+	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/cloudevents/sdk-go/v2/protocol"
 	"k8s.io/client-go/rest"
 	"knative.dev/pkg/injection"
 	"knative.dev/pkg/logging"
@@ -59,7 +62,38 @@ func withCloudEventClient(ctx context.Context) context.Context {
 		logger.Panicf("Error creating the cloudevents client: %s", err)
 	}
 
-	return context.WithValue(ctx, ceKey{}, cloudEventClient)
+	celient := CloudClient{
+		client: cloudEventClient,
+	}
+	return context.WithValue(ctx, ceKey{}, celient)
+}
+
+// CloudClient is a wrapper of CloudEvents client and implements addCount and decreaseCount
+type CloudClient struct {
+	client client.Client
+}
+
+// AddCount does nothing
+func (c CloudClient) addCount() {
+}
+
+// DecreaseCount does nothing
+func (c CloudClient) decreaseCount() {
+}
+
+// Send invokes call client.Send
+func (c CloudClient) Send(ctx context.Context, event cloudevents.Event) protocol.Result {
+	return c.client.Send(ctx, event)
+}
+
+// Request invokes client.Request
+func (c CloudClient) Request(ctx context.Context, event event.Event) (*cloudevents.Event, protocol.Result) {
+	return c.client.Request(ctx, event)
+}
+
+// StartReceiver invokes client.StartReceiver
+func (c CloudClient) StartReceiver(ctx context.Context, fn interface{}) error {
+	return c.client.StartReceiver(ctx, fn)
 }
 
 // Get extracts the cloudEventClient client from the context.
--- a/pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
+++ b/pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
@@ -19,12 +19,13 @@ package cloudevent
 import (
 	"context"
 	"fmt"
+	"regexp"
+	"sync"
+	"testing"
 
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"github.com/cloudevents/sdk-go/v2/protocol"
 )
-
-const bufferSize = 100
 
 // FakeClientBehaviour defines how the client will behave
 type FakeClientBehaviour struct {
@@ -37,13 +38,17 @@ type FakeClient struct {
 	behaviour *FakeClientBehaviour
 	// Modelled after k8s.io/client-go fake recorder
 	Events chan string
+	// waitGroup is used to block until all events have been sent
+	waitGroup *sync.WaitGroup
 }
 
 // newFakeClient is a FakeClient factory, it returns a client for the target
-func newFakeClient(behaviour *FakeClientBehaviour) cloudevents.Client {
+func newFakeClient(behaviour *FakeClientBehaviour, expectedEventCount int) CEClient {
 	return FakeClient{
 		behaviour: behaviour,
-		Events:    make(chan string, bufferSize),
+		// set buffersize to length of want events to make sure no extra events are sent
+		Events:    make(chan string, expectedEventCount),
+		waitGroup: &sync.WaitGroup{},
 	}
 }
 
@@ -52,8 +57,12 @@ var _ cloudevents.Client = (*FakeClient)(nil)
 // Send fakes the Send method from cloudevents.Client
 func (c FakeClient) Send(ctx context.Context, event cloudevents.Event) protocol.Result {
 	if c.behaviour.SendSuccessfully {
-		c.Events <- fmt.Sprintf("%s", event.String())
-		return nil
+		// This is to prevent extra events are sent. We don't read events from channel before we call CheckCloudEventsUnordered
+		if len(c.Events) < cap(c.Events) {
+			c.Events <- fmt.Sprintf("%s", event.String())
+			return nil
+		}
+		return fmt.Errorf("Channel is full of size:%v, but extra event wants to be sent:%v", cap(c.Events), event)
 	}
 	return fmt.Errorf("Had to fail. Event ID: %s", event.ID())
 }
@@ -61,8 +70,11 @@ func (c FakeClient) Send(ctx context.Context, event cloudevents.Event) protocol.
 // Request fakes the Request method from cloudevents.Client
 func (c FakeClient) Request(ctx context.Context, event cloudevents.Event) (*cloudevents.Event, protocol.Result) {
 	if c.behaviour.SendSuccessfully {
-		c.Events <- fmt.Sprintf("%v", event.String())
-		return &event, nil
+		if len(c.Events) < cap(c.Events) {
+			c.Events <- fmt.Sprintf("%v", event.String())
+			return &event, nil
+		}
+		return nil, fmt.Errorf("Channel is full of size:%v, but extra event wants to be sent:%v", cap(c.Events), event)
 	}
 	return nil, fmt.Errorf("Had to fail. Event ID: %s", event.ID())
 }
@@ -72,7 +84,56 @@ func (c FakeClient) StartReceiver(ctx context.Context, fn interface{}) error {
 	return nil
 }
 
-// WithClient adds to the context a fake client with the desired behaviour
-func WithClient(ctx context.Context, behaviour *FakeClientBehaviour) context.Context {
-	return context.WithValue(ctx, ceKey{}, newFakeClient(behaviour))
+// addCount can be used to add the count when each event is going to be sent
+func (c FakeClient) addCount() {
+	c.waitGroup.Add(1)
+}
+
+// decreaseCount can be used to the decrease the count when each event is sent
+func (c FakeClient) decreaseCount() {
+	c.waitGroup.Done()
+}
+
+// WithClient adds to the context a fake client with the desired behaviour and expectedEventCount
+func WithClient(ctx context.Context, behaviour *FakeClientBehaviour, expectedEventCount int) context.Context {
+	return context.WithValue(ctx, ceKey{}, newFakeClient(behaviour, expectedEventCount))
+}
+
+// CheckCloudEventsUnordered checks that all events in wantEvents, and no others,
+// were received via the given chan, in any order.
+// Block until all events have been sent.
+func (c *FakeClient) CheckCloudEventsUnordered(t *testing.T, testName string, wantEvents []string) {
+	t.Helper()
+	c.waitGroup.Wait()
+	expected := append([]string{}, wantEvents...)
+	channelEvents := len(c.Events)
+
+	// extra events are prevented in FakeClient's Send function.
+	// fewer events are detected because we collect all events from channel and compare with wantEvents
+	for eventCount := 0; eventCount < channelEvents; eventCount++ {
+		event := <-c.Events
+		if len(expected) == 0 {
+			t.Errorf("extra event received: %q", event)
+		}
+		found := false
+		for wantIdx, want := range expected {
+			matching, err := regexp.MatchString(want, event)
+			if err != nil {
+				t.Errorf("something went wrong matching an event: %s", err)
+			}
+			if matching {
+				found = true
+				// Remove event from list of those we expect to receive
+				expected[wantIdx] = expected[len(expected)-1]
+				expected = expected[:len(expected)-1]
+				break
+			}
+		}
+		if !found {
+			t.Errorf("unexpected event received: %q", event)
+		}
+	}
+	if len(expected) != 0 {
+		t.Errorf("%d events %#v are not received", len(expected), expected)
+	}
 }
--- a/pkg/reconciler/events/cloudevent/cloudeventsfakeclient_test.go
+++ b/pkg/reconciler/events/cloudevent/cloudeventsfakeclient_test.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2022 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cloudevent
+
+import (
+	"testing"
+
+	"github.com/cloudevents/sdk-go/v2/event"
+	rtesting "knative.dev/pkg/reconciler/testing"
+)
+
+func TestSend_Success(t *testing.T) {
+	sendEvents := []event.Event{
+		{
+			Context: event.EventContextV1{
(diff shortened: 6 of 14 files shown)


Yongxuanzhang:
cc @lbernick @vdemeester @imjasonh

lbernick:
based on [this comment](https://github.com/tektoncd/pipeline/issues/4404#issuecomment-985656290) from @vdemeester  it seems like we're thinking of moving cloudevents functionality out of pipelines, so it might not make sense to combine these two packages into one, but I'm not sure if this comment is still true. @afrittoli can you comment on this?

pkg/reconciler/events/cloudevent/cloud_event_controller.go
```diff
@@ -144,7 +145,9 @@ func SendCloudEventWithRetries(ctx context.Context, object runtime.Object) error
 	_, isRun := object.(*v1alpha1.Run)
 
 	wasIn := make(chan error)
+	e.WaitGroup.Add(1)
```
lbernick, line 148: One thing that I'm noticing here is that the interface is really just a wrapper around a waitGroup, i.e. it is pretty similar to the original implementation that just used a plain waitGroup. I'm sorry if I led you down a rabbit hole with my suggestions about how this interface should be implemented, but now that I'm seeing it in a PR we might just want a wrapper around the CloudEventClient. Something like

```
type OurOwnCloudEventClient struct {
   client *cloudEvents.Client
   eventCount int
}

func initialize(ce *cloudEvents.Client) *OurOwnCloudEventClient {
   client = ce
   eventCount = 0
}

func (c *OurOwnCloudEventClient) SendCloudEventWithRetries(ctx, object) {
  <do some stuff>
  eventCount += 1
  go func() {
    c.client.Send()
  }
}

func (c *OurOwnCloudEventClient) checkEventsOrdered(t *testing.T, <some other args>) {
    eventsSent := []string{}
    for i := 0; i < c.eventCount, i ++ {
      e <- c.client.Events
      eventsSent = append(eventsSent, e)
    }
   // Compare expected list of events with actual list of events
}
```
In this example, eventCount can also be replaced with a waitgroup. WDYT?

pkg/reconciler/events/eventsender/events_order_check.go
```diff
@@ -24,10 +24,11 @@ import (
 
 // CheckEventsOrdered checks that the events received via the given chan are the same as wantEvents,
 // in the same order.
-func CheckEventsOrdered(t *testing.T, eventChan chan string, testName string, wantEvents []string) error {
+func CheckEventsOrdered(t *testing.T, eventChan chan string, testName string, wantEvents []string, e EventSenderInterface) error {
 	t.Helper()
 	// Sleep 50ms to make sure events have delivered
 	time.Sleep(50 * time.Millisecond)
+	e.Wait()
```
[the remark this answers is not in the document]
Yongxuanzhang: I'm not sure if the fix can 100% replace the sleep here (Though I believe it is). Maybe if we could wait the fix is merged and then remove the sleep in another PR?

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -72,7 +79,70 @@ func (c FakeClient) StartReceiver(ctx context.Context, fn interface{}) error {
...
+}
+
 // WithClient adds to the context a fake client with the desired behaviour
 func WithClient(ctx context.Context, behaviour *FakeClientBehaviour) context.Context {
 	return context.WithValue(ctx, ceKey{}, newFakeClient(behaviour))
 }
+
+// CheckCloudEventsUnordered checks that all events in wantEvents, and no others,
+// were received via the given chan, in any order.
+func (c *FakeClient) CheckCloudEventsUnordered(t *testing.T, eventChan chan string, testName string, wantEvents []string) error {
```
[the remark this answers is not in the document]
Yongxuanzhang: oh yes!!

pkg/reconciler/events/cloudevent/cloud_event_controller.go
```diff
@@ -144,7 +144,11 @@ func SendCloudEventWithRetries(ctx context.Context, object runtime.Object) error
 	_, isRun := object.(*v1alpha1.Run)
 
 	wasIn := make(chan error)
+
+	// fix flacky cloud client related units
```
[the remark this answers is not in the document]
Yongxuanzhang: ok I will remove this

pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -79,7 +79,11 @@ func (t TektonEventType) String() string {
 }
 
 // CEClient matches the `Client` interface from github.com/cloudevents/sdk-go/v2/cloudevents
-type CEClient cloudevents.Client
+type CEClient interface {
+	cloudevents.Client
+	AddCount()
+	DecreaseCount()
```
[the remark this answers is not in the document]
Yongxuanzhang: yes, thanks!

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -72,7 +79,70 @@ func (c FakeClient) StartReceiver(ctx context.Context, fn interface{}) error {
...
+		return fmt.Errorf("error in test %s: %v", testName, err)
+	}
+	return nil
+}
+
+// eventsFromChannelUnordered takes a chan of string and a list of events that a test
+// expects to receive. The events can be received in any order. Any extra or too few
+// events are both considered errors.
+func eventsFromChannelUnordered(c chan string, wantEvents []string) error {
+	timer := time.NewTimer(10 * time.Millisecond)
```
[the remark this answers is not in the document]
Yongxuanzhang: I think we can not remove it, it is not related to the blocking, but a timer used to read from channel

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -72,7 +79,70 @@ func (c FakeClient) StartReceiver(ctx context.Context, fn interface{}) error {
...
+	expected := append([]string{}, wantEvents...)
```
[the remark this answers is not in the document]
Yongxuanzhang: I tried this method and it probably doesn't work, in this function it will do regex matching from expected events to got events, the reason is the events from channel is much more complex so we cannot simply do diff between two list of strings.

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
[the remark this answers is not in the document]
lbernick: Why do the tests hang if the timer is not used? The only reason this should happen is if the expected number of events is less than the actual number of events received.

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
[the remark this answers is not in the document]
lbernick: Can you try defining your own comparison function? Here's an [example](https://github.com/tektoncd/pipeline/blob/main/pkg/internal/computeresources/compare/compare.go#L56-L79)

pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -78,8 +78,15 @@ func (t TektonEventType) String() string {
...
-type CEClient cloudevents.Client
+// CEClient wraps the `Client` interface from github.com/cloudevents/sdk-go/v2/cloudevents
+// and has methods to count goroutines that send cloud events
+type CEClient interface {
+	cloudevents.Client
+	// addCount can be implemented to add count for goroutine
+	addCount()
+	// addCount can be implemented to decrease count for goroutine
+	decreaseCount()
+}
```
lbernick, line 89:
```suggestion
	// addCount increments the count of events to be sent
	addCount()
	// decreaseCount decrements the count of events to be sent, indicating the event has been sent
	decreaseCount()
}
```

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -72,7 +79,69 @@ func (c FakeClient) StartReceiver(ctx context.Context, fn interface{}) error {
...
+// AddCount increases 1 count for waitGroup
+func (c FakeClient) addCount() {
+	c.waitGroup.Add(1)
+}
+
+// DecreaseCount decreases 1 count for waitGroup
+func (c FakeClient) decreaseCount() {
+	c.waitGroup.Done()
+}
+
```
lbernick, line 91: I wonder if `decreaseCount` is even necessary? you could call c.waitGroup.Done() within`Send`

Yongxuanzhang: Probably not? Because the fakeclient `Send` is not only used in `SendCloudEventWithRetries`'s gorountine. There is another function `SendCloudEvents` also used it and it doesn't need the waitgroup to block.

pkg/reconciler/events/cloudevent/cloudevent.go
Yongxuanzhang: Thanks!

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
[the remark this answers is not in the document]
Yongxuanzhang: It basically means we will reuse the current code to compare the two list? If so I don't think it adds any values here. And it may has less readability than the current version.

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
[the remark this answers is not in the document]
Yongxuanzhang: It's because the previous code will loop +1 times, and during the extra loop since we don't have any events from channel, and also we don't have a timer, it will just hand there,
```
loop len(expected) + 1 times to catch extra erroneous events received that the test is not expecting
```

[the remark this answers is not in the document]
Yongxuanzhang: To fix it we can just reduce the loop to len(expected) not len(expected)+1. But it would be a change to this function's behaviour. I don't know why or who implement this logic. I suggest we if we want to do it, we should leave it to another clean up PR to avoid extra changes in this PR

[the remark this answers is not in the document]
lbernick: Oh I see-- the reason is because we don't want to make sure extra events have been sent, so we wait a little extra to see if any more events are received. What we really need is for the fake client to be able to indicate that it's done sending events (e.g. by closing the channel), but the fake client doesn't know when events are done being sent because they can only be sent one at a time so I'm not sure if this is possible. Let me think about if there's a better way to do this. Keeping this logic here means that our tests are still not deterministic.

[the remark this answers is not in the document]
Yongxuanzhang: Sure! Thanks!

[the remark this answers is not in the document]
lbernick: ok! I have an idea! We should update [newFakeClient](https://github.com/tektoncd/pipeline/blob/14eba83bf31277dc3f8c91fa798b38e30d0a8e6a/pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go#L43-L48) to initialize the fake client buffer size with len(wantEvents). 

The reason is that sends block on a buffered channel when it is full. We are currently waiting (via waitGroup.Wait()) until all events are sent before reading any events from the channel. This means if the controller sends more events than expected, these sends will block because the channel is full, and Wait() will wait indefinitely.

With this solution, if the client sends more or fewer events than expected, Wait() will block indefinitely, but if the client sends the correct number of events, Wait() will wait until they are all sent, and no longer! a.k.a. no more flakes!

Does this make sense? WDYT? If you implement this please remove the timer and add some commentary on this strategy.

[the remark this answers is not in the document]
Yongxuanzhang: Maybe? Let me test this solution. Thank you for coming up with this idea!!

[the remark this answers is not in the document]
Yongxuanzhang: I tried to implement this in the latest commit, plz lmk if that is aligned what you're suggesting.

lbernick (approved):
Thanks @Yongxuanzhang for addressing this flake that has been annoying everyone since... before I joined Tekton :) Approving since this fixes the flake, and the codebase is improved by this change, but please make sure to address the readability and style comments before merging.

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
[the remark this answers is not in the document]
Yongxuanzhang: Sure! Let me give it a try. Thanks for your advice!

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
Yongxuanzhang: I'm not sure since that's confusing. 😂 We don't have gorontine in `SendCloudEvents `

pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -88,8 +88,15 @@ func (t TektonEventType) String() string {
 	return string(t)
 }
 
-// CEClient matches the `Client` interface from github.com/cloudevents/sdk-go/v2/cloudevents
-type CEClient cloudevents.Client
+// CEClient wraps the `Client` interface from github.com/cloudevents/sdk-go/v2/cloudevents
+// and has methods to count goroutines that send cloud events
```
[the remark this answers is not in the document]
Yongxuanzhang: Sure!

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -37,13 +40,17 @@ type FakeClient struct {
 	behaviour *FakeClientBehaviour
 	// Modelled after k8s.io/client-go fake recorder
 	Events chan string
+	// waitGroup is used to block until all events have been sent
+	waitGroup *sync.WaitGroup
 }
 
 // newFakeClient is a FakeClient factory, it returns a client for the target
-func newFakeClient(behaviour *FakeClientBehaviour) cloudevents.Client {
+func newFakeClient(behaviour *FakeClientBehaviour, clientChanbuffersize int) CEClient {
```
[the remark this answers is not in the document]
Yongxuanzhang: yesm thanks!

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -72,7 +79,66 @@ func (c FakeClient) StartReceiver(ctx context.Context, fn interface{}) error {
...
+
+// DecreaseCount decreases 1 count for waitGroup
+func (c FakeClient) decreaseCount() {
+	c.waitGroup.Done()
+}
+
 // WithClient adds to the context a fake client with the desired behaviour
-func WithClient(ctx context.Context, behaviour *FakeClientBehaviour) context.Context {
-	return context.WithValue(ctx, ceKey{}, newFakeClient(behaviour))
+func WithClient(ctx context.Context, behaviour *FakeClientBehaviour, clientChanbuffersize int) context.Context {
```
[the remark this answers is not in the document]
Yongxuanzhang: sure!

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -72,7 +79,66 @@ func (c FakeClient) StartReceiver(ctx context.Context, fn interface{}) error {
...
+
+// eventsFromChannelUnordered takes a chan of string and a list of events that a test
+// expects to receive. The events can be received in any order. Any extra or too few
+// events are both considered errors.
+func eventsFromChannelUnordered(c chan string, wantEvents []string) error {
+	expected := append([]string{}, wantEvents...)
+	// fakeclient's channel buffersize equals to the size of wantEvents, so no extra events will be sent
+	maxEvents := len(expected)
+	for eventCount := 0; eventCount < maxEvents; eventCount++ {
+		select {
```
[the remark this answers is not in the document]
Yongxuanzhang: removed! Thanks!

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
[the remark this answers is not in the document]
Yongxuanzhang: I think the possible issue is that, we cannot guarantee that the two sorted lists have the same order. 
e.g.:
[a,b,c,] -> [a,b,c]
[za,by,xc]->[by,xc,za]
If we iterate them and compare on by one then this is not working.

pkg/reconciler/events/cloudevent/cloudeventclient.go
```diff
@@ -59,7 +62,38 @@ func withCloudEventClient(ctx context.Context) context.Context {
...
 	}
 
-	return context.WithValue(ctx, ceKey{}, cloudEventClient)
+	celient := CloudClient{
+		client: cloudEventClient,
+	}
+	return context.WithValue(ctx, ceKey{}, celient)
+}
+
+// CloudClient is wrapper of cloud client and implements AddCount and DecreaseCount
```
afrittoli, line 71: NIT: CloudClient is a wrapper of CloudEvents client and implements addCount and decreaseCount

pkg/reconciler/events/cloudevent/cloudevent.go
```diff
@@ -88,8 +88,15 @@ func (t TektonEventType) String() string {
 	return string(t)
 }
 
-// CEClient matches the `Client` interface from github.com/cloudevents/sdk-go/v2/cloudevents
-type CEClient cloudevents.Client
+// CEClient wraps the `Client` interface from github.com/cloudevents/sdk-go/v2/cloudevents
+// and has methods to count the cloud events being sent
```
afrittoli, line 92: NIT: perhaps it's worth commenting that the extra methods are for testing purposes only

pkg/reconciler/customrun/customrun_test.go
```diff
@@ -187,7 +188,7 @@ func TestReconcile_CloudEvents(t *testing.T) {
 			}
 
 			ceClient := clients.CloudEvents.(cloudevent.FakeClient)
-			err = eventstest.CheckEventsUnordered(t, ceClient.Events, tc.name, tc.wantCloudEvents)
+			ceClient.CheckCloudEventsUnordered(t, tc.name, tc.wantCloudEvents)
 			if err != nil {
 				t.Errorf(err.Error())
 			}
```
afrittoli, line 194: There is no `err` returned anymore to check!

pkg/reconciler/customrun/customrun_test.go
```diff
@@ -196,7 +197,7 @@ func TestReconcile_CloudEvents(t *testing.T) {
 			if err := c.Reconciler.Reconcile(testAssets.Ctx, getCustomRunName(customRun)); err != nil {
 				t.Fatal("Didn't expect an error, but got one.")
 			}
-			err = eventstest.CheckEventsUnordered(t, ceClient.Events, tc.name, []string{})
+			ceClient.CheckCloudEventsUnordered(t, tc.name, []string{})
 			if err != nil {
 				t.Errorf(err.Error())
 			}
```
afrittoli, line 203: There is no `err` returned anymore to check!

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -37,13 +38,17 @@ type FakeClient struct {
...
+	waitGroup *sync.WaitGroup
 }
 
 // newFakeClient is a FakeClient factory, it returns a client for the target
-func newFakeClient(behaviour *FakeClientBehaviour) cloudevents.Client {
+func newFakeClient(behaviour *FakeClientBehaviour, expectedEventCount int) CEClient {
 	return FakeClient{
 		behaviour: behaviour,
-		Events:    make(chan string, bufferSize),
+		// set buffersize to length of want events to make sure no extra events are sent
```
afrittoli, line 49: What happens if the controller tried to send more events than expected? Would the test fail?

pkg/reconciler/events/cloudevent/cloudeventclient.go
Yongxuanzhang: Thank you!

pkg/reconciler/events/cloudevent/cloudevent.go
Yongxuanzhang: Sure!

pkg/reconciler/customrun/customrun_test.go
Yongxuanzhang: thanks! I have fixed all of them

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
Yongxuanzhang: Yes it will fail and this is also the previous behaviour.

Yongxuanzhang: Oh I see the problem, this will lead to deadlock, I added a buffer size check before sending the event, if extra event wants to be sent it will return error. Plz lmk if that works, thanks!

afrittoli: OK - it sounds good - it's important for the test to fail in that case otherwise we would miss error conditions.
Since the test method there is not trivial, I wonder if we should have a test that verifies this case.

Yongxuanzhang: Sure! I will add a test for this!

Yongxuanzhang: Added a test for this, ptal! 🙏

lbernick:
/hold cancel
since this is ready for review

lbernick:
@Yongxuanzhang could you please update your commit message title + body to match the PR description?

Yongxuanzhang:
> @Yongxuanzhang could you please update your commit message title + body to match the PR description?

Oh thanks! I just updated it. Sorry I made some changes after @afrittoli's reviews. you may want to remove your approval and review again? 

1. I added a check in `Send` so if the channel is full we will return error, this should prevent the extra events.
2. I loop the channel size not the len(expectedEvents), this will avoid blocking on reading more events when we send less events. This should work in this PR since we will wait to ensure no events are missed

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -72,7 +83,67 @@ func (c FakeClient) StartReceiver(ctx context.Context, fn interface{}) error {
...
+	err := eventsFromChannelUnordered(c.Events, wantEvents)
+	if err != nil {
+		t.Fatalf("error getting events from fake client: %v", err)
+	}
+}
+
+// eventsFromChannelUnordered takes a chan of string and a list of events that a test
+// expects to receive. The events can be received in any order. Any extra or too few
+// events are both considered errors.
+func eventsFromChannelUnordered(c chan string, wantEvents []string) error {
```
lbernick, line 118: It might make sense to merge this function with `CheckCloudEventsUnordered`, since it's not used anywhere else as far as I can tell, and it needs the call to waitgroup.Wait in order to work correctly.

pkg/reconciler/events/cloudevent/cloudeventsfakeclient_test.go
```diff
@@ -0,0 +1,74 @@
...
+		if err != nil {
+			t.Fatalf("got err %v", err)
+		}
+
+	}
+	fakeClient.CheckCloudEventsUnordered(t, "send cloud events", wantEvents)
+}
+
+func TestSend_Error(t *testing.T) {
+	sendEvents := []event.Event{{
```
lbernick, line 57: nit: this doesn't need to be a slice

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -72,7 +83,67 @@ func (c FakeClient) StartReceiver(ctx context.Context, fn interface{}) error {
...
+	expected := append([]string{}, wantEvents...)
+	channelEvents := len(c)
+	// fakeclient's channel buffersize equals to the size of wantEvents, so no extra events will be sent
```
lbernick, line 121: This comment is slightly misleading. Buffered channels can send once events have been pulled off of them. The reason no extra events will be sent is because the fake client returns an error in this case, but that logic lives far away from here. 

I'd suggest adding commentary to the fake client where you return an error (to explain why this prevents extra events from being sent) and some commentary here about why this block detects the case where too few events are sent.

pkg/reconciler/events/cloudevent/cloudeventsfakeclient_test.go
Yongxuanzhang: 👍

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
Yongxuanzhang: 👍

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
Yongxuanzhang: 👍

pkg/reconciler/events/cloudevent/cloudeventsfakeclient_test.go
```diff
@@ -0,0 +1,74 @@
...
+	// Setup the context and seed test event
+	ctx, _ := rtesting.SetupFakeContext(t)
+	ctx = WithClient(ctx, &FakeClientBehaviour{SendSuccessfully: true}, 0)
+	fakeClient := Get(ctx).(FakeClient)
+
+	// the channel size is 0 so no more events can be sent
+	err := fakeClient.Send(ctx, sendEvent)
+	if err == nil {
+		t.Fatalf("want err but got nil")
+	}
```
afrittoli, line 72: Thank you 🙏

pkg/reconciler/testing/logger.go
```diff
@@ -24,11 +24,15 @@ import (
 // SetupFakeContext sets up the Context and the fake filtered informers for the tests.
 func SetupFakeContext(t *testing.T) (context.Context, []controller.Informer) {
 	ctx, _, informer := setupFakeContextWithLabelKey(t)
+	return WithLogger(ctx, t), informer
+}
+
+// SetupFakeCloudClientContext sets up the fakeclient to context
+func SetupFakeCloudClientContext(ctx context.Context, expectedEventCount int) context.Context {
```
afrittoli, line 31: NIT: Why does this need to be a separate function?

test/events/events.go
```diff
@@ -35,19 +35,6 @@ func CheckEventsOrdered(t *testing.T, eventChan chan string, testName string, wa
 	return nil
```
afrittoli, line 35: NIT: It's a bit inconsistent now that the unordered check has been moved inside to the fake client but the ordered one is in this test module. It might we worth keeping this logic in the same place, for maintenance and to help test developers. BTW, do you think the ordered check suffers from the same flake as the other one?

afrittoli: This is something we could fix as a follow-up.

afrittoli (commented):
Thank you!
/lgtm

test/events/events.go
Yongxuanzhang: Yes I'm looking into the k8s event as well, but I don't think they are the same root cause. 
Some prs:
https://github.com/tektoncd/pipeline/pull/5752
https://github.com/tektoncd/pipeline/pull/5770

PR #5705, opened 2022-11-01
Use SchemeGroupVersion for tekton objects in cluster resolver

chuangw6:
# Changes

/kind cleanup

Prior, the cluster resolver sets the APIVersion of the retrieved tekton object to be a fixed string "tekton.dev/v1beta1".

Now, we changed this to `v1beta1.SchemeGroupVersion.String()` for easier transition to v1 once v1 is released.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Use v1beta1.SchemeGroupVersion.String() for the APIVersion field in the tekton object retrieved by cluster resolver.
```

### DIFF
--- a/pkg/resolution/resolver/cluster/resolver.go
+++ b/pkg/resolution/resolver/cluster/resolver.go
@@ -101,6 +101,7 @@ func (r *Resolver) Resolve(ctx context.Context, origParams []pipelinev1beta1.Par
 	}
 
 	var data []byte
+	groupVersion := pipelinev1beta1.SchemeGroupVersion.String()
 
 	switch params[KindParam] {
 	case "task":
@@ -110,7 +111,7 @@ func (r *Resolver) Resolve(ctx context.Context, origParams []pipelinev1beta1.Par
 			return nil, err
 		}
 		task.Kind = "Task"
-		task.APIVersion = "tekton.dev/v1beta1"
+		task.APIVersion = groupVersion
 		data, err = yaml.Marshal(task)
 		if err != nil {
 			logger.Infof("failed to marshal task %s from namespace %s: %v", params[NameParam], params[NamespaceParam], err)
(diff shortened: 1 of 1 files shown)


pkg/resolution/resolver/cluster/resolver.go
```diff
@@ -101,6 +101,7 @@ func (r *Resolver) Resolve(ctx context.Context, origParams []pipelinev1beta1.Par
 	}
 
 	var data []byte
+	groupVersion := pipelinev1beta1.SchemeGroupVersion.String()
```
JeromeJu, line 104: Thanks Chuang! I was wondering if there is any way we can link this to the storage version?

JeromeJu: What could happen if the version is not correctly specified?

vdemeester (approved):
/lgtm

pkg/resolution/resolver/cluster/resolver.go
chuangw6: Sorry for the delay @JeromeJu .
> I was wondering if there is any way we can link this to the storage version?

Not sure how we can get the storage version. I think `pipelinev1beta1.SchemeGroupVersion.String()` might be sufficient here as andrew suggested [here](https://github.com/tektoncd/pipeline/pull/5687#discussion_r1010491191). Once we switched to v1, we need to change the imported pkg and this version will be changed accordingly I think. 

> What could happen if the version is not correctly specified?

That's a good question. I just played around this. If the typemeta i.e. the version or kind is not set correctly, the resolution will still succeed, but [the code of converting the retrieved content to the runtime object](https://github.com/tektoncd/pipeline/blob/b16a5b5ea65a4dae2401415fb498c340170557c2/pkg/remote/resolution/resolver.go#L80-L83) will fail. And it will complain "no kind "Task" is registered for version "v1" in scheme ...", which makes sense because we don't serve the v1 CRD in the cluster.

JeromeJu: Thanks for the findings Chuang, and yeah I think that is very reasonable.

PR #4759, opened 2022-04-14
[TEP-0089] - Phase 1 Signed TaskRun Results

pxp928:
Signed-off-by: pxp928 <parth.psu@gmail.com>

# Changes

Authors - @pxp928 and @lumjjb 

In association with [TEP-0089: Non-falsifiable provenance support](https://github.com/tektoncd/community/blob/main/teps/0089-nonfalsifiable-provenance-support.md)

This PR is the implementation of Phase 1 of the [TEP-0089: Non-falsifiable provenance support](https://github.com/tektoncd/community/blob/main/teps/0089-nonfalsifiable-provenance-support.md)

Phase 2 of the PR. Thats build on top of this to add Signed TaskRun Status can be found here - #4828

Phase 1

* Add support for Signed Results with SPIRE (this will primarily involve modifications to the entrypointer image)
* Add support for tekton-pipelines-controller verifying Signed Results

Taking the work that @dlorenc started a while back and adding improvements to get the spire integrated with Tekton pipeline.

The integration of spire k8s workload registrar automatically allows for tekton created pods to be registered into the spire-server.

Currently spire-server and spire-agent needs to be running in your cluster in order for tekton pipelines to integrate. Please follow the Spire documentation that is part of this PR to set up for local testing.

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Release notes block below has been filled in or deleted (only if no user facing changes)

# Release Notes

- Added Spiffe-CSI driver to the controller and allow for spire workload API communication over CSI driver (not using hostpath)
- Added in [k8s-workload-registrar](https://github.com/spiffe/spire/tree/main/support/k8s/k8s-workload-registrar) into the tekton pipeline controller. This will allow pods to be auto registered into the spire server based on pod annotations. 
- Added fetch SVID based on the pod running the TaskRun and use the private key to sign payload and attach the corresponding SVID to the termination messages
- Add in a RESULT_MANIFEST to register all outputs of the TaskRun
[...]

PR #5715, opened 2022-11-03
[TEP-0089] Enable SPIRE for signing taskrun results in alpha.

jagathprakash:
[TEP-0089] Enable SPIRE for signing taskrun results in alpha. Breaking down PR #4759 originally proposed by @pxp928 to address TEP-0089 according @lumjjb suggestions.e Plan for breaking down PR is PR 1.1: api PR 1.2: entrypointer (+cmd line + test/entrypointer) Entrypoint takes results and signs the results (termination message). PR 1.3: reconciler + pod + cmd/controller + integration tests Controller will verify the signed result. This commit corresponds to 1.3 above.

# Changes
This PR is one of the 3 which addresses non-falsifiable provenance in Tekton Taskruns.
Specifically it uses SPIRE to attest a TR and create unique certificates for TRs.
TRs use these certificates to sign the results which ensures that the TR results cannot be modified without detection by another workload.
This PR in specific has 
1. Changes to enable Pipelines to talk to Spire server and agent to request a certificate.
2. TR to sign the results that it creates.
3. TR to verify that the results are indeed what it has created.

The next PR would enable Chains to verify that a result created by a Task is indeed created by the task.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note

- Added Spiffe-CSI driver to the controller and allow for spire workload API communication over CSI driver (not using hostpath)
- Added fetch SVID based on the pod running the TaskRun and use the private key to sign payload and attach the corresponding SVID to the termination messages
- Add in a RESULT_MANIFEST to register all outputs of the TaskRun. Tekton Pipelines verifies the Results against the SPIRE SVID and Trust Bundle and sets the SignedResultsVerified condition to True
- Added a configmap config-spire which will have settings to enable Pipelines to talk to SPIRE.

The exact details of how to use SPIRE are in the doc docs/spire.md.
```


jagathprakash:
> @jagathprakash Seems like the tests are breaking as we might need to update them based on changes made to pipelines. Let me know if you need help debugging it.

I did update the tests according to latest pipeline changes but might have missed some issues. I will try to fix these and if I get stuck will ping you for help. Thanks @pxp928!

jagathprakash:
> @jagathprakash Seems like the tests are breaking as we might need to update them based on changes made to pipelines. Let me know if you need help debugging it.

I did update the tests according to latest pipeline changes but might have missed some issues. I will try to fix these and if I get stuck will ping you for help. Thanks @pxp928!

chitrangpatel (approved):
apart from the comment, it looks good!

pxp928:
/approve

Looks good. We made updates to the tests.

cmd/controller/main.go
```diff
@@ -62,12 +62,20 @@ func main() {
 	flag.StringVar(&opts.Images.ImageDigestExporterImage, "imagedigest-exporter-image", "", "The container image containing our image digest exporter binary.")
 	flag.StringVar(&opts.Images.WorkingDirInitImage, "workingdirinit-image", "", "The container image containing our working dir init binary.")
 
+	flag.StringVar(&opts.SpireConfig.TrustDomain, "spire-trust-domain", "example.org", "Experimental: The SPIRE Trust domain to use.")
+	flag.StringVar(&opts.SpireConfig.SocketPath, "spire-socket-path", "unix:///spiffe-workload-api/spire-agent.sock", "Experimental: The SPIRE agent socket for SPIFFE workload API.")
+	flag.StringVar(&opts.SpireConfig.ServerAddr, "spire-server-addr", "spire-server.spire.svc.cluster.local:8081", "Experimental: The SPIRE server address for workload/node registration.")
+	flag.StringVar(&opts.SpireConfig.NodeAliasPrefix, "spire-node-alias-prefix", "/tekton-node/", "Experimental: The SPIRE node alias prefix to use.")
```
afrittoli, line 68: These are not valid default values it seems, and need to be customised to be able to use this feature.
Is there any reason these cannot be loaded from a config map like other configuration parameters? 
Setting them here implies to me that these values should not be changed at runtime. If that is the case, I think it's fine to keep the configs in here. I would ask you then to include them in the controller manifest, perhaps commented out, so that if someone would like to use the feature they can simply uncomment in the manifests and set the correct value. I know it's in the `spire.md` docs but I think having it commented in the controller manifest would help.

jagathprakash: @afrittoli Thanks for the review comment. I think it should be possible to change these flags into a configmap. I am trying that out now. Will update once I am done.

docs/spire.md
```diff
@@ -0,0 +1,285 @@
...
+│   SPIRE Agent    ( Runs as   )                                           │
+│   + CSI Driver   ( Daemonset )                                           │
+│                                                                          │
+└──────────────────────────────────────────────────────────────────────────┘
+```
+
+Initial Setup:
+1. As part of the SPIRE deployment, the SPIRE server attests the agents running on each node in the cluster.
+2. The Tekton Controller is configured to have workload identity entry creation permissions to the SPIRE server.
+3. As part of the Tekton Controller operations, the Tekton Controller will retrieve an identity that it can use to talk to the SPIRE server to register TaskRun workloads.
```
afrittoli, line 95: NIT: using 1. 2. 3. here and then 1. 1. 1. below - perhaps it should be 1. 1. 1. here too

docs/spire.md
```diff
@@ -0,0 +1,285 @@
...
+...
+📝 Results
+ NAME                    VALUE
+ ∙ commit       aa79de59c4bae24e32f15fda467d02ae9cd94b01
+ ∙ url          https://github.com/buildpacks/samples
+```
+
+## Architecture Overview
+
+This feature relies on a SPIRE installation. This is how it integrates into the architecture of Tekton:
```
afrittoli, line 60: Thank you for including the diagram, nice! ❤️

docs/spire.md
```diff
@@ -0,0 +1,285 @@
...
+1. The entrypointer receives an x509 SVID, containing the x509 certificate and associated private key. 
+1. The entrypointer signs the results of the TaskRun and emits the signatures and x509 certificate to the TaskRun results for later verification.
+
+## Enabling TaskRun result attestations
+
+To enable TaskRun attestations:
+1. Make sure `enable-spire` is set to `"true"` in the `feature-flags` configmap, see [`install.md`](./install.md#customizing-the-pipelines-controller-behavior) for details
+1. Create a SPIRE deployment containing a SPIRE server, SPIRE agents and the SPIRE CSI driver, for convenience, [this sample single cluster deployment](https://github.com/spiffe/spiffe-csi/tree/main/example/config) can be used.
+1. Register the SPIRE workload entry for Tekton with the "Admin" flag, which will allow the Tekton controller to communicate with the SPIRE server to manage the TaskRun identities dynamically.
+    ```
```
afrittoli, line 111: NIT - There should be an empty line between the text and a code block (below too)

docs/spire.md
```diff
@@ -0,0 +1,285 @@
...
+1. The SPIRE agent will attest the workload and request an SVID.
+1. The entrypointer receives an x509 SVID, containing the x509 certificate and associated private key. 
+1. The entrypointer signs the results of the TaskRun and emits the signatures and x509 certificate to the TaskRun results for later verification.
+
+## Enabling TaskRun result attestations
+
+To enable TaskRun attestations:
+1. Make sure `enable-spire` is set to `"true"` in the `feature-flags` configmap, see [`install.md`](./install.md#customizing-the-pipelines-controller-behavior) for details
+1. Create a SPIRE deployment containing a SPIRE server, SPIRE agents and the SPIRE CSI driver, for convenience, [this sample single cluster deployment](https://github.com/spiffe/spiffe-csi/tree/main/example/config) can be used.
+1. Register the SPIRE workload entry for Tekton with the "Admin" flag, which will allow the Tekton controller to communicate with the SPIRE server to manage the TaskRun identities dynamically.
```
afrittoli, line 110: I'm not familiar with SPIRE enough to judge here, but does "admin" comply with the principle of least privilege?

docs/spire.md
```diff
@@ -0,0 +1,285 @@
...
+    ```
+    # This example is assuming use of the above SPIRE deployment
+    # Example where trust domain is "example.org" and cluster name is "example-cluster"
+    
+    # Register a node alias for all nodes of which the Tekton Controller may reside
+    kubectl -n spire exec -it \
+        deployment/spire-server -- \
+        /opt/spire/bin/spire-server entry create \
+            -node \
+            -spiffeID spiffe://example.org/allnodes \
```
afrittoli, line 120: Does this mean all nodes in the cluster or is "allnodes" the name of the alias

bobcatfish (commented):
Very excited to see this moving forward! I've so far taken a look primarily at the commits and briefly the tests and docs

* The PR description says no release notes, but it seems like this PR is adding user facing functionality so I think it would make sense to have an accompanying release note? (maybe I'm missing something though!)
* A couple of requests around the 2 commits: one of them seems to be just a merge (https://github.com/tektoncd/pipeline/pull/5715/commits/2c805750660064012d5d82393388eaef18d3176e), can it be squashed together with the other one? The other request, could the commit message be updated with a bit more detail? I'm finding the description a bit hard to follow, it seems like it assumes a lot of context - and I can't quite tell which of the PRs this one is supposed to be. I'd really like to see a specific description of what the changes are in this PR (i.e. what's changing and why) - there are some more docs on how to do this in our commit message standard at https://github.com/tektoncd/community/blob/main/standards.md#commits
* Can you explain a bit around the strategy behind the new end to end tests? It looks like a lot of existing tests have been modified and I'm wondering if it might be easier to reason about and maintain the tests if instead of updating existing tests to have new cases woven through them, we maybe add a targeted test or two specifically for the SPIRE functionality?  The `if spireEnabled` code that seems to have shown up a lot in the tests reminds me a lot of the 'code smell' around boolean flags (e.g. https://ardalis.com/are-boolean-flags-on-methods-a-code-smell/)

cmd/imagedigestexporter/main.go
```diff
@@ -31,6 +34,8 @@ import (
 var (
 	images                 = flag.String("images", "", "List of images resources built by task in json format")
 	terminationMessagePath = flag.String("terminationMessagePath", "/tekton/termination", "Location of file containing termination message")
+	enableSpire            = flag.Bool("enable_spire", false, "If specified by configmap, this enables spire signing and verification")
+	socketPath             = flag.String("spire_socket_path", "unix:///spiffe-workload-api/spire-agent.sock", "Experimental: The SPIRE agent socket for SPIFFE workload API.")
```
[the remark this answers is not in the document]
pxp928: Ah yes. I think this should be able to be removed. Would need to test to make sure no functionality breaks.

docs/spire.md
pxp928: "allnodes" is just the alias specified for the spiffeID.

pxp928:
> Very excited to see this moving forward! I've so far taken a look primarily at the commits and briefly the tests and docs
> 
> * The PR description says no release notes, but it seems like this PR is adding user facing functionality so I think it would make sense to have an accompanying release note? (maybe I'm missing something though!)
> * A couple of requests around the 2 commits: one of them seems to be just a merge ([2c80575](https://github.com/tektoncd/pipeline/commit/2c805750660064012d5d82393388eaef18d3176e)), can it be squashed together with the other one? The other request, could the commit message be updated with a bit more detail? I'm finding the description a bit hard to follow, it seems like it assumes a lot of context - and I can't quite tell which of the PRs this one is supposed to be. I'd really like to see a specific description of what the changes are in this PR (i.e. what's changing and why) - there are some more docs on how to do this in our commit message standard at https://github.com/tektoncd/community/blob/main/standards.md#commits
> * Can you explain a bit around the strategy behind the new end to end tests? It looks like a lot of existing tests have been modified and I'm wondering if it might be easier to reason about and maintain the tests if instead of updating existing tests to have new cases woven through them, we maybe add a targeted test or two specifically for the SPIRE functionality?  The `if spireEnabled` code that seems to have shown up a lot in the tests reminds me a lot of the 'code smell' around boolean flags (e.g. https://ardalis.com/are-boolean-flags-on-methods-a-code-smell/)

Yes, release notes on the original PR https://github.com/tektoncd/pipeline/pull/4759 can be copied over to this PR. 

As for the e2e tests, the thought was that spire should be able to support each use case and not break the original functionality. That is the reason why I updated the tests to add in the spire use case. I do agree with your assessment that the boolean flags might cause tech debt in the future. So happy to change this based on recommendations.

docs/spire.md
pxp928: The admin flag is needed for the Tekton controller to allow for dynamic registration of the taskruns. The spire server itself should be running in another more secure k8s cluster with limited access. Thus, only the registered tekton controller is allowed to dynamically register taskruns. @lumjjb any other thoughts?

docs/spire.md
jagathprakash: Done.

docs/spire.md
jagathprakash: Done.

docs/spire.md
jagathprakash: Thanks to @pxp928 and @lumjjb who added this diagram.

bendory:
Prakash, can you please update the PR description to better adhere to the [standards](https://github.com/tektoncd/community/blob/main/standards.md#commits)?

Specifically -- what changes are actually happening in this PR? Do any of these changes justify release notes?

Thanks!

cmd/imagedigestexporter/main.go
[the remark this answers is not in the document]
jagathprakash: Done. 
I removed the changes to imagedigestexporter.

cmd/controller/main.go
jagathprakash: I moved the spire config to a config map. These changes are in a separate commit so we could probably get them out if not ok.
Pros of configmap
1. The config can be changed after pipelines controller starts. So if a spire server goes down and a new one is instantiated, the config map can be changed and Tekton will pick up the change.

BTW the above is also a con of the configmap as now the SPIRE server can be changed after the pipelines comes up.

@pxp928 @lumjjb @afrittoli PTAL and comment on the config map approach.

jagathprakash:
> Prakash, can you please update the PR description to better adhere to the [standards](https://github.com/tektoncd/community/blob/main/standards.md#commits)?
> 
> Specifically -- what changes are actually happening in this PR? Do any of these changes justify release notes?
> 
> Thanks!

Thanks for the review. 
I have updated the PR description to reflect
1. What problem this PR addresses.
2. Top level description of how it addresses this.

I have also updated the release notes. The release note links to spire.md which is not complete (i.e it does not completely address how this PR can be used) and I am addressing that now.

cmd/imagedigestexporter/main.go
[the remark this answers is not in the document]
pxp928: @jagathprakash Will have to do some testing to make sure this did not break any functionality.

jagathprakash:
Thanks for the review! Please find my comments inlined.
> Very excited to see this moving forward! I've so far taken a look primarily at the commits and briefly the tests and docs
> 


> * The PR description says no release notes, but it seems like this PR is adding user facing functionality so I think it would make sense to have an accompanying release note? (maybe I'm missing something though!)
This does bring new user facing functionality. As such I have updated the release notes. The release notes point to spire.md. The release notes have been copied from PR-4759 as @pxp928 suggested.

> * A couple of requests around the 2 commits: one of them seems to be just a merge ([2c80575](https://github.com/tektoncd/pipeline/commit/2c805750660064012d5d82393388eaef18d3176e)), can it be squashed together with the other one? The other request, could the commit message be updated with a bit more detail? I'm finding the description a bit hard to follow, it seems like it assumes a lot of context - and I can't quite tell which of the PRs this one is supposed to be. I'd really like to see a specific description of what the changes are in this PR (i.e. what's changing and why) - there are some more docs on how to do this in our commit message standard at https://github.com/tektoncd/community/blob/main/standards.md#commits

I have updated the PR description to reflect what is being done in this PR. As for the commits, I will squash them once the configmap method of initializing pipelines to talk to SPIRE is approved.

> * Can you explain a bit around the strategy behind the new end to end tests? It looks like a lot of existing tests have been modified and I'm wondering if it might be easier to reason about and maintain the tests if instead of updating existing tests to have new cases woven through them, we maybe add a targeted test or two specifically for the SPIRE functionality?  The `if spireEnabled` code that seems to have shown up a lot in the tests reminds me a lot of the 'code smell' around boolean flags (e.g. https://ardalis.com/are-boolean-flags-on-methods-a-code-smell/)

From @pxp928 (which seems to have got lost). As for the e2e tests, the thought was that spire should be able to support each use case and not break the original functionality. That is the reason why I updated the tests to add in the spire use case. I do agree with your assessment that the boolean flags might cause tech debt in the future. So happy to change this based on recommendations.

Looking into if we can change this to not use spire-enabled inside the tests while still checking if SPIRE integration works for all tests.

cmd/imagedigestexporter/main.go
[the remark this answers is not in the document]
jagathprakash: Agree.
I ran the unit tests and they are ok.

The E2E tests are stuck with some errors. Trying to find the issue.

bobcatfish:
> As for the e2e tests, the thought was that spire should be able to support each use case and not break the original functionality. That is the reason why I updated the tests to add in the spire use case.
> Looking into if we can change this to not use spire-enabled inside the tests while still checking if SPIRE integration works for all tests.

I wonder if one difficulty is that we've kinda got two different things going on:
1. We want to make sure the SPIRE features are working as expected (adding expected signatures, etc.)
2. We want to make sure that everything else still works when Tekton Pipelines is using SPIRE

(1) seems like something we'd want an explict test for, (2) kinda feels like something like, like it's more about environment setup. I'm wondering if we want to do something like always have SPIRE on for end to end tests (or the opposite) and have a less frequently run test (e.g. nightly) that runs the tests the other way (e.g. run tests on PRs with SPIRE enabled, have a nightly run that runs the end to end tests with SPIRE off to make sure that still works) <-- we're actually going to run into a lot of questions like this as our testing gets more complicated with per feature flags in general + @dibyom 

I'd be happy to just do (1) for now (i.e. have one focused test with SPIRE on that makes sure that the functionality works)

pxp928:
> > As for the e2e tests, the thought was that spire should be able to support each use case and not break the original functionality. That is the reason why I updated the tests to add in the spire use case.
> > Looking into if we can change this to not use spire-enabled inside the tests while still checking if SPIRE integration works for all tests.
> 
> I wonder if one difficulty is that we've kinda got two different things going on:
> 
> 1. We want to make sure the SPIRE features are working as expected (adding expected signatures, etc.)
> 2. We want to make sure that everything else still works when Tekton Pipelines is using SPIRE
> 
> (1) seems like something we'd want an explict test for, (2) kinda feels like something like, like it's more about environment setup. I'm wondering if we want to do something like always have SPIRE on for end to end tests (or the opposite) and have a less frequently run test (e.g. nightly) that runs the tests the other way (e.g. run tests on PRs with SPIRE enabled, have a nightly run that runs the end to end tests with SPIRE off to make sure that still works) <-- we're actually going to run into a lot of questions like this as our testing gets more complicated with per feature flags in general + @dibyom
> 
> I'd be happy to just do (1) for now (i.e. have one focused test with SPIRE on that makes sure that the functionality works)

Sure that sounds like a plan. We can change to have focused tests rather than having them all run spire. As for (2), a less frequent nightly e2e test with spire running is something I can help create as needed.

config/config-feature-flags.yaml
```diff
@@ -90,3 +90,7 @@ data:
   # in the TaskRun/PipelineRun such as the source from where a remote Task/Pipeline
   # definition was fetched.
   enable-provenance-in-status: "false"
+  # Setting this flag to "true" enables spire integration with pipeline.
+  # This is an experimental feature and thus should still be considered
+  # an alpha feature.
+  enable-spire: "false"
```
afrittoli, line 96: NIT: it would be nice to use a name associated with the feature from a Tekton user point of view, e.g. `enable-taskrun-attestation` or so

docs/spire.md
```diff
@@ -0,0 +1,287 @@
...
+    - name: spiffe-workload-api
+      mountPath: /spiffe-workload-api
+      readOnly: true
+    
+    # Add the following to the volumes of the controller pod
+    - name: spiffe-workload-api
+      csi:
+        driver: "csi.spiffe.io"
+    ```
+1. (Optional) Modify the controller (`config/controller.yaml`) to configure non-default SPIRE options by adding arguments to the CLI.
```
afrittoli, line 149: Isn't this done via config map now?

docs/spire.md
```diff
@@ -0,0 +1,287 @@
+<!--
+---
+linkTitle: "TaskRun Result Attestation"
+weight: 1660
```
afrittoli, line 4: We're in the process of restructuring the documentation, so we don't have a good framework for assigning these IDs right now. I imaging 1660 might end up towards the end of the list.

docs/spire.md
jagathprakash: Thats correct. Updated the doc to reflect the change.
Thanks for pointing this out.

cmd/imagedigestexporter/main.go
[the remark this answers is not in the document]
jagathprakash: @afrittoli, @pxp928 this does break functionality. 
TaskRun converts this to a step. As such if we do not add signing in imagedigestextractor, the taskrun fails as the signature does not exist for this step.
One way is to add "if else logic" and not check for the signature for only these steps.
The other way is to just change imagedigestextractor, which seems cleaner. 

So reverted back to adding SPIRE feature to PipelineResources. 
Please comment on whether this is ok or we should change it to not check in the taskrun.

cmd/imagedigestexporter/main.go
```diff
@@ -76,6 +81,21 @@ func main() {
 
 	}
 
+	if enableSpire != nil && *enableSpire && socketPath != nil && *socketPath != "" {
```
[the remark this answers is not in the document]
jagathprakash: Done.

cmd/imagedigestexporter/main.go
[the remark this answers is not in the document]
pxp928: @jagathprakash we dont not want to skip any checks. Can you walk me through the error and we can come up with a path forward?

pkg/spire/controller.go
```diff
@@ -45,6 +47,20 @@ func init() {
...
+// OnStore stores the changed spire config into the SpireClientApi
+func OnStore(ctx context.Context, logger *zap.SugaredLogger) func(name string, value interface{}) {
+	return func(name string, value interface{}) {
+		if name == config.GetSpireConfigName() {
+			cfg, ok := value.(*spireconfig.SpireConfig)
+			if !ok {
+				logger.Error("Failed to do type insertion for extracting metrics config")
+				return
+			}
+			GetControllerAPIClient(ctx).SetConfig(*cfg)
```
pxp928, line 59: Need to close connection - `	GetControllerAPIClient(ctx).Close()` and set them to:

```
sc.entryClient = nil, sc.workloadConn = nil, sc.workloadAPI = nil,  sc.serverConn = nil
```

config/config-feature-flags.yaml
jagathprakash: +1. Agree that this should be named from a Tekton user point of view. 
In the near future we might have SPIRE attesting the results generated by other objects like Pipelines.
enable-taskrun-attestation would be incorrect then.
How about enable-non-falsifiable-provenance ?

docs/spire.md
jagathprakash: This is good point. But I am not sure how to address this?  Should we lower the weight to say somewhere below 1000s?

jagathprakash:
@afrittoli @bobcatfish I have addressed all the comments and made the following high level changes
1. Changed the spire initialization in the controller to a config map.
2. Addressed the concern that we are modifying code for a to be deprecated feature PipelineResources.
3. Modified the tests so that we are not adding a new SPIRE test for every previous feature.
4. Added one e2e test which tests this new feature.
5. Updated the doc to reflect the new config map based changes.

Please do take a look.

test/e2e-tests.sh
```diff
@@ -40,6 +40,21 @@ header "Setting up environment"
...
+  fi
+  if [ "$gate" == "alpha" ] ; then
+    DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+    printf "Setting up environement for alpha features"
+    install_spire
+    patch_pipline_spire
+    kubectl patch configmap config-spire -n tekton-pipelines -f "$DIR"/testdata/spire/config-spire.yaml
+    failed=0
+  fi
+}
```
Yongxuanzhang, line 57: why do we create a new func here, can it be merged into set_feature_gate() ?

examples/v1beta1/pipelineruns/4808-regression.yaml
```diff
@@ -92,4 +92,4 @@ spec:
     name: result-test
   params:
   - name: RESULT_STRING_LENGTH
-    value: "3000"
+    value: "2000"
```
Yongxuanzhang, line 95: why changing this to 2000? will 3000 exceed the limit in test due to more data in termination msg?

chitrangpatel: I agree. No need to change it.

chitrangpatel: termination message allows up to 4096 bytes.

test/e2e-tests.sh
jagathprakash: This is just to keep the size of the set_feature_gate small and also keep this only for SPIRE initialization.
We could merge it  into set_feature_gate.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -527,6 +530,16 @@ func (c *Reconciler) reconcile(ctx context.Context, tr *v1beta1.TaskRun, rtr *re
...
 	if podconvert.SidecarsReady(pod.Status) {
+		if spireEnabled {
+			// TTL for the entry is in seconds
+			ttl := time.Duration(config.FromContextOrDefaults(ctx).Defaults.DefaultTimeoutMinutes) * time.Minute
+			if err = c.SpireClient.CreateEntries(ctx, tr, pod, ttl); err != nil {
+				logger.Errorf("Failed to create workload SPIFFE entry for taskrun %v: %v", tr.Name, err)
+				return err
+			}
+			logger.Infof("Created SPIFFE workload entry for %v/%v", tr.Namespace, tr.Name)
+		}
```
chitrangpatel, line 541: I wonder if this can be tested via unit tests. Coverage is missing for this. Maybe it can only be done via e2e tests?

pkg/spire/controller.go
```diff
@@ -45,6 +47,22 @@ func init() {
...
+				logger.Error("Failed to do type insertion for extracting metrics config")
+				return
+			}
+			controllerAPIClient := GetControllerAPIClient(ctx)
+			controllerAPIClient.Close()
+			controllerAPIClient.SetConfig(*cfg)
+		}
+	}
+}
+
```
chitrangpatel, line 65: Any way to unit test this?

docs/spire.md
```diff
@@ -0,0 +1,273 @@
...
+1. When the TaskRun pod is ready, the Tekton Controller registers an identity with the information of the pod to the SPIRE server. This will tell the SPIRE server the identity of the TaskRun to use as well as how to attest the workload/pod.
+1. After the TaskRun steps complete, as part of the entrypointer code, it requests an SVID from SPIFFE workload API (via the SPIRE agent socket)
+1. The SPIRE agent will attest the workload and request an SVID.
+1. The entrypointer receives an x509 SVID, containing the x509 certificate and associated private key. 
+1. The entrypointer signs the results of the TaskRun and emits the signatures and x509 certificate to the TaskRun results for later verification.
+
+## Enabling TaskRun result attestations
+
+To enable TaskRun attestations:
+1. Make sure `enable-spire` is set to `"true"` in the `feature-flags` configmap, see [`install.md`](./install.md#customizing-the-pipelines-controller-behavior) for details
```
bobcatfish, line 108: Drive-by requests re. this option:
1. The name of this option and the values don't match the name in the TEP (https://github.com/tektoncd/community/blob/main/teps/0089-nonfalsifiable-provenance-support.md#enabling-spire-on-the-entrypointer-image) - i.e. this should be `--enforce-nonfalsifiablity=spire` - and in general please audit this functionality against TEP-0089 which describes the design that was proposed and accepted
2. Please also document new config options in https://github.com/tektoncd/pipeline/blob/main/docs/install.md

pkg/reconciler/taskrun/taskrun.go
pxp928: Creating the entries cannot be checked as this is an interaction with the spire server directly. We do have some mocked spire tests in the spire package but this has to be done via e2e test.

pkg/apis/config/spire_config_test.go
```diff
@@ -0,0 +1,72 @@
...
+	type testCase struct {
+		expectedConfig *sc.SpireConfig
+		fileName       string
+	}
+
+	testCases := []testCase{
+		{
+			expectedConfig: &sc.SpireConfig{
+				TrustDomain:     "test.com",
+				SocketPath:      "unix:///spiffe-workload-api/test-spire-agent.sock",
```
bendory, line 39: Reference the defined `const` values instead of repeating the strings, or use bogus values that are clearly test values like `unix:///somewhere.sock`.

pkg/apis/config/spire_config_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+	test "github.com/tektoncd/pipeline/pkg/reconciler/testing"
+	sc "github.com/tektoncd/pipeline/pkg/spire/config"
+	"github.com/tektoncd/pipeline/test/diff"
+)
+
+func TestNewSpireConfigFromConfigMap(t *testing.T) {
+	type testCase struct {
+		expectedConfig *sc.SpireConfig
```
bendory, line 31: Go style favors brevity: just call this `want`.

pkg/apis/config/spire_config_test.go
```diff
@@ -0,0 +1,72 @@
...
+		SocketPath:      "unix:///spiffe-workload-api/spire-agent.sock",
+		ServerAddr:      "spire-server.spire.svc.cluster.local:8081",
+		NodeAliasPrefix: "/tekton-node/",
+	}
+	verifyConfigFileWithExpectedSpireConfig(t, SpireConfigEmptyName, expectedConfig)
+}
+
+func verifyConfigFileWithExpectedSpireConfig(t *testing.T, fileName string, expectedConfig *sc.SpireConfig) {
+	cm := test.ConfigMapFromTestFile(t, fileName)
+	if ab, err := config.NewSpireConfigFromConfigMap(cm); err == nil {
```
bendory, line 65: Go style is `got, err` (instead of `ab`) -- then later in the test you compare `got` with `want`.

pkg/apis/config/spire_config_test.go
jagathprakash: Done.

pkg/apis/config/spire_config_test.go
jagathprakash: Done.

pkg/apis/config/spire_config_test.go
jagathprakash: Done.

docs/spire.md
jagathprakash: > Drive-by requests re. this option:
> 
> 1. The name of this option and the values don't match the name in the TEP (https://github.com/tektoncd/community/blob/main/teps/0089-nonfalsifiable-provenance-support.md#enabling-spire-on-the-entrypointer-image) - i.e. this should be `--enforce-nonfalsifiablity=spire` - and in general please audit this functionality against TEP-0089 which describes the design that was proposed and accepted
Agree.. Modified the user facing flag enable-spire to enforce-nonfalsifiability.

> 2. Please also document new config options in https://github.com/tektoncd/pipeline/blob/main/docs/install.md

pkg/spire/controller.go
jagathprakash: Added a test for this.

pkg/pod/status.go
```diff
@@ -529,6 +576,36 @@ func markStatusSuccess(trs *v1beta1.TaskRunStatus) {
 	})
 }
 
+// markStatusResultsVerified sets taskrun status to
```
bendory, line 579:
```suggestion
// markStatusResultsVerified sets taskrun status to verified
```

pkg/apis/config/feature_flags.go
```diff
@@ -157,6 +165,25 @@ func NewFeatureFlagsFromMap(cfgMap map[string]string) (*FeatureFlags, error) {
...
 
+	setStringFeature := func(key string, defaultValue string, acceptedValues []string, feature *string) error {
+		value := defaultValue
+		if cfg, ok := cfgMap[key]; ok {
+			value = strings.ToLower(cfg)
+		}
+		accepted := false
+		for _, s := range acceptedValues {
+			if s == value {
+				accepted = true
```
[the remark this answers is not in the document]
jagathprakash: Replace this altogether with a set i.e map[string]struct{}, so that we can get rid of this iteration.

pkg/apis/config/store_test.go
```diff
@@ -74,6 +78,7 @@ func TestStoreLoadWithContext_Empty(t *testing.T) {
 	artifactPVC, _ := config.NewArtifactPVCFromMap(map[string]string{})
 	metrics, _ := config.NewMetricsFromConfigMap(&corev1.ConfigMap{Data: map[string]string{}})
 	trustedresources, _ := config.NewTrustedResourcesConfigFromMap(map[string]string{})
+	spireConfig, _ := config.NewSpireConfigFromMap(map[string]string{})
 
 	expected := &config.Config{
```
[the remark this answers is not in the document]
jagathprakash: Done.

pkg/apis/config/spire_config_test.go
```diff
@@ -0,0 +1,72 @@
...
+				TrustDomain:     "test.com",
+				SocketPath:      "unix:///test-spire-api/test-spire-agent.sock",
+				ServerAddr:      "test-spire-server.spire.svc.cluster.local:8081",
+				NodeAliasPrefix: "/test-tekton-node/",
+			},
+			fileName: config.GetSpireConfigName(),
+		},
+	}
+
+	for _, tc := range testCases {
```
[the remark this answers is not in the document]
jagathprakash: Merged the empty config and the new config tests into a single Test with a new entry in the testCases slice.

pkg/pod/pod_test.go
```diff
@@ -2344,6 +2406,171 @@ func TestPodBuild_TaskLevelResourceRequirements(t *testing.T) {
...
+			}
+
+			got, err := builder.Build(store.ToContext(context.Background()), tr, c.ts)
+			if err != nil {
+				t.Fatalf("builder.Build: %v", err)
+			}
+
+			expectedName := kmeta.ChildName(tr.Name, "-pod")
+			if d := cmp.Diff(expectedName, got.Name); d != "" {
+				t.Errorf("Pod name does not match: %q", d)
```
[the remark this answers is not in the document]
jagathprakash: Done

docs/spire.md
jagathprakash: Regarding the install.md, this feature is not yet done. We have 2 more PRs (one in pipelines and the other in chains) before this can be done.
The install.md is a user facing doc. I think we should do this around in the final PR. WDYT?

pkg/apis/config/feature_flags_test.go
```diff
@@ -141,15 +140,15 @@ func TestNewFeatureFlagsFromConfigMap(t *testing.T) {
...
+				EnforceNonfalsifiability:         "spire",
 				ResourceVerificationMode:         config.DefaultResourceVerificationMode,
 				RunningInEnvWithInjectedSidecars: config.DefaultRunningInEnvWithInjectedSidecars,
 				AwaitSidecarReadiness:            config.DefaultAwaitSidecarReadiness,
 				ResultExtractionMethod:           config.DefaultResultExtractionMethod,
 				MaxResultSize:                    config.DefaultMaxResultSize,
 				CustomTaskVersion:                config.DefaultCustomTaskVersion,
 			},
-			fileName: "feature-flags-enable-spire",
+			fileName: "feature-flags-enforce-nonfalsifiability-spire",
```
chitrangpatel, line 151: I wonder if you can also test with a bad value `ie. enforce-nonfalsifiability: 'not spire or none'`. I did that for larger results if you want to look at an example.

pkg/pod/status.go
```diff
@@ -165,7 +193,7 @@ func setTaskRunStatusBasedOnStepStatus(ctx context.Context, logger *zap.SugaredL
 			merr = multierror.Append(merr, err)
 		}
 		// populate task run CRD with results from sidecar logs
-		taskResults, pipelineResourceResults, _ := filterResultsAndResources(sidecarLogResults)
+		taskResults, pipelineResourceResults, _ := filterResultsAndResources(sidecarLogResults, spireEnabled)
```
chitrangpatel, line 196: Thanks for catching this!!!

pkg/spire/controller.go
```diff
@@ -45,14 +47,30 @@ func init() {
 // controllerKey is a way to associate the ControllerAPIClient from inside the context.Context
 type controllerKey struct{}
 
+// OnStore stores the changed spire config into the SpireClientApi
+func OnStore(ctx context.Context, logger *zap.SugaredLogger) func(name string, value interface{}) {
+	return func(name string, value interface{}) {
+		if name == config.GetSpireConfigName() {
+			cfg, ok := value.(*spireconfig.SpireConfig)
+			if !ok {
+				logger.Error("Failed to do type insertion for extracting SPIRE config")
```
Yongxuanzhang, line 56: is this  `type assertion`?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -1268,12 +1272,30 @@ spec:
...
+				name:       "mycontainer",
+				image:      "myimage",
+				stdoutPath: "stdout.txt",
+				cmd:        "/mycmd",
+			}}, true),
+			[]corev1.VolumeMount{
+				{
+					Name:      spire.WorkloadAPI,
+					MountPath: spire.VolumeMountPath,
+					ReadOnly:  true,
```
Yongxuanzhang, line 1295: why one use readonly(which is set to true), and the other uses `true`?

pkg/apis/config/feature_flags.go
```diff
@@ -207,15 +232,15 @@ func NewFeatureFlagsFromMap(cfgMap map[string]string) (*FeatureFlags, error) {
 	if tc.EnableAPIFields == AlphaAPIFields {
 		tc.EnableTektonOCIBundles = true
 		tc.EnableCustomTasks = true
-		tc.EnableSpire = true
+		tc.EnforceNonfalsifiability = EnforceNonfalsifiabilityWithSpire
```
Yongxuanzhang, line 235: does this mean the spire feature is enabled by default for alpha api flag?

pkg/apis/config/feature_flags.go
```diff
@@ -80,8 +80,12 @@ const (
 	DefaultSendCloudEventsForRuns = false
 	// DefaultEmbeddedStatus is the default value for "embedded-status".
 	DefaultEmbeddedStatus = FullEmbeddedStatus
-	// DefaultEnableSpire is the default value for "enable-spire".
-	DefaultEnableSpire = false
+	// EnforceNonfalsifiabilityWithSpire is the value used for  "enable-nonfalsifiability" when SPIRE is used to enable non-falsifiability.
+	EnforceNonfalsifiabilityWithSpire = "spire"
+	// EnforceNonfalsifiabilityNone is the value used for  "enable-nonfalsifiability" when non-falsifiability is not enabled.
+	EnforceNonfalsifiabilityNone = ""
```
Yongxuanzhang, line 86: why don't we consider Spire feature flag as a bool `EnforceNonfalsifiabilityWithSpire`? Will there be any other choices?

pkg/apis/config/feature_flags_test.go
jagathprakash: Done. 

Added it in TestNewFeatureFlagsConfigMapErrors.

test/init_test.go
```diff
@@ -319,3 +325,19 @@ func getCRDYaml(ctx context.Context, cs *clients, ns string) ([]byte, error) {
...
+	if tr.IsTaskRunResultVerified() {
+		t.Errorf("Taskrun `%s` status condition should not be verified as taskrun failed", tr.Name)
+	}
+	t.Logf("Taskrun `%s` status results condition verified by spire as false, which is valid", tr.Name)
+}
+
+// Verifies if the taskrun results are verified by spire
+func spireShouldPassTaskRunResultsVerify(tr *v1beta1.TaskRun, t *testing.T) {
+	if !tr.IsTaskRunResultVerified() {
+		t.Errorf("Taskrun `%s` status condition not verified. Spire taskrun results verification failure", tr.Name)
```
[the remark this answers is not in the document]
jagathprakash: Good catch. Done.

pkg/spire/controller.go
jagathprakash: Fixed. :D

pkg/apis/config/feature_flags.go
jagathprakash: yes.

pkg/apis/config/feature_flags.go
jagathprakash: Not sure, but that was what was decided in the TEP-0089. 
There are other ways to enforce non-falsifiable provenance. As of now there is only SPIRE based way.

pkg/reconciler/taskrun/taskrun_test.go
jagathprakash: CSIVolumeSource (https://pkg.go.dev/k8s.io/api/core/v1#CSIVolumeSource)'s ReadOnly type is *bool, while VolumeMount it is bool.
Ofcourse, we could use readonly in VolumeMount too.

docs/spire.md
```diff
@@ -0,0 +1,273 @@
...
+1. When the TaskRun pod is ready, the Tekton Controller registers an identity with the information of the pod to the SPIRE server. This will tell the SPIRE server the identity of the TaskRun to use as well as how to attest the workload/pod.
+1. After the TaskRun steps complete, as part of the entrypointer code, it requests an SVID from SPIFFE workload API (via the SPIRE agent socket)
+1. The SPIRE agent will attest the workload and request an SVID.
+1. The entrypointer receives an x509 SVID, containing the x509 certificate and associated private key. 
+1. The entrypointer signs the results of the TaskRun and emits the signatures and x509 certificate to the TaskRun results for later verification.
+
+## Enabling TaskRun result attestations
+
+To enable TaskRun attestations:
+1. Make sure `enforce-nonfalsifiability` is set to `"spire"` in the `feature-flags` configmap, see [`install.md`](./install.md#customizing-the-pipelines-controller-behavior) for details
```
bobcatfish, line 108: i dont see any updates to install.md in this PR, could it be updated to mention the new config option as well?

test/e2e-tests.sh
```diff
@@ -40,6 +40,21 @@ header "Setting up environment"
 install_pipeline_crd
 
 failed=0
+function alpha_gate() {
```
bobcatfish, line 43: it looks like the function `set_feature_gate` has an existing section for ` if [ "$gate" == "alpha" ];` - does it maek sense to do this setup there instead? or make a new function for alpha setup in general and move the logic from `set_feature_gate` into there as well?

test/e2e-common.sh
```diff
@@ -48,6 +48,65 @@ function install_pipeline_crd_version() {
...
+}
+
+function patch_pipline_spire() {
+  kubectl patch \
+      deployment tekton-pipelines-controller \
+      -n tekton-pipelines \
+      --patch-file "$DIR"/testdata/patch/pipeline-controller-spire.json
+      
+  verify_pipeline_installation
+}
```
bobcatfish, line 108: i wonder if it would work to make a new spire.sh for the spire specific logic instead of making e2e-common.sh even longer? ideally i wish this could be in a Tekton Task for Spire setup but there are a lot of things we'd need to update before we could do that 🙃  - maybe isolating the script into its own file would at least be a step in the direction of modularizing this logic a bit?

test/e2e-common.sh
```diff
@@ -48,6 +48,65 @@ function install_pipeline_crd_version() {
...
+    -selector k8s_psat:agent_sa:spire-agent \
+    -node
+  spire_apply \
+    -spiffeID spiffe://example.org/ns/tekton-pipelines/sa/tekton-pipelines-controller \
+    -parentID spiffe://example.org/ns/spire/node/example \
+    -selector k8s:ns:tekton-pipelines \
+    -selector k8s:pod-label:app:tekton-pipelines-controller \
+    -selector k8s:sa:tekton-pipelines-controller \
+    -admin
+}
```
bobcatfish, line 99: i have a lot of trouble telling from any of these scripts, but what happens if any of these commands fail? will it fail the script or will the script keep going? i.e. does there need to be any more error checking?

jagathprakash:
Closing this pull request in lieu of the newer smaller PRS
1. https://github.com/tektoncd/pipeline/pull/5902
2. https://github.com/tektoncd/pipeline/pull/6160

PR #5717, opened 2022-11-03
Remove logging.request-log-template from resolvers config-observability CM

### LINKED ISSUE #5707: 0.41.0 new configmap: config-observability contains helm specific syntax
We are using our own custom helm chart to deploy new versions of tekton and in version 0.41.0 there is a new configmap named "config-observability" and in the example the following syntax is used: 

{{.Request.Method}}

and more of this in the same example.

By doing so helm will try to resolve this eventhough it is just an example. It would be nice if the entire example could be commented out

abayer:
# Changes

Fixes #5707

When I copied `config/resolvers/config-observability.yaml` from resolution.git, I didn't notice that, unlike our existing `config/config-observability.yaml`, it contained a rather elaborate `logging.request-log-template` example. That example ends up causing problems with Helm, due to it containing syntax that Helm tries to resolve, like `{{.Request.Method}}`.

There's no real reason to have that example there, and we don't have it in the `tekton-pipelines` namespace's `config-observability`, so I'm removing it from `config/resolvers/config-observability.yaml` now.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Clean up example configuration in config-observability configmap for tekton-pipelines-resolvers namespace
```

### DIFF
--- a/config/resolvers/config-observability.yaml
+++ b/config/resolvers/config-observability.yaml
@@ -39,35 +39,6 @@ data:
     # this example block and unindented to be in the data block
     # to actually change the configuration.
 
-    # If non-empty, this enables queue proxy writing request logs to stdout.
-    # The value determines the shape of the request logs and it must be a valid go text/template.
-    # It is important to keep this as a single line. Multiple lines are parsed as separate entities
-    # by most collection agents and will split the request logs into multiple records.
-    #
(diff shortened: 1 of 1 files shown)


vdemeester (approved):
We’ll have to backport it in 0.41.1 I think 😇

PR #5746, opened 2022-11-11
[TEP-0124] implement opentelemetry Jaeger tracing

kmjayadeep:
Adds opentelemetry instrumentation code to pipelinerun and taskrun reconcilers. Also made required changes in the main.go to include jaeger as tracing backend

# Changes

# Submitter Checklist

  functionality, content, code)

Note to reviewers: There are pipeline failures (looks like they are not related to the changes). Once the review is complete and ready to approve, I will address the pipeline issues, rebase with main and squash all commits into one.

# Release Notes

```release-note
Added support for tracing using Jaeger and OpenTelemetry. It can be enabled by adding the following environment variables to controller manifest
* `OTEL_EXPORTER_JAEGER_ENDPOINT` is the HTTP endpoint for sending spans directly to a collector.
* `OTEL_EXPORTER_JAEGER_USER` is the username to be sent as authentication to the collector endpoint. (optional)
* `OTEL_EXPORTER_JAEGER_PASSWORD` is the password to be sent as authentication to the collector endpoint. (optional)
```


pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -715,10 +736,42 @@ func (c *Reconciler) reconcile(ctx context.Context, pr *v1beta1.PipelineRun, get
...
+			}
+		}
+	}
+	if len(errs) > 0 {
+		e := strings.Join(errs, "\n")
+		return fmt.Errorf("error(s) from processing cancel request for timed out Run(s) of PipelineRun %s: %s", pr.Name, e)
+	}
+	return nil
+}
+
```
afrittoli, line 768: I don't believe this code belongs to this PR?

kmjayadeep: nice find. Looks like this was removed from main and wasnt removed from this branch while rebasing

afrittoli:
From the error in the build tests it might be that something went wrong in the rebase again, but you could try and run the codegen script to see if it fixes the error

kmjayadeep:
Hi @afrittoli I tried the codegen script already. It has introduced some formatting changes, but didn't fix the build.

afrittoli:
> Hi @afrittoli I tried the codegen script already. It has introduced some formatting changes, but didn't fix the build.

I had a look and your commit actually includes formatting changes to generated files, which causes the build job to fail. 

```
pkg/client/clientset/versioned/fake/register.go                                               |   14 +-
pkg/client/clientset/versioned/scheme/register.go                                             |   14 +-
pkg/client/resolution/clientset/versioned/fake/register.go                                    |   14 +-
pkg/client/resolution/clientset/versioned/scheme/register.go                                  |   14 +-
pkg/client/resource/clientset/versioned/fake/register.go                                      |   14 +-
pkg/client/resource/clientset/versioned/scheme/register.go                                    |   14 +-
```

I tried reverting those and running the codegen again and they appear back, so I suspect this might be related to a change in go version between CI and your/mine local go version.

My go version is:
```
go version go1.19.1 darwin/arm64
```
while CI runs on 1.18.

For this specific CI, I would recommend reverting those changes and adding them to your commit. You can skip the codegen script since there is no change in the PR which requires it:

```
git checkout origin/main -- pkg/client/clientset/versioned/fake/register.go
git checkout origin/main -- pkg/client/clientset/versioned/scheme/register.go
git checkout origin/main -- pkg/client/resolution/clientset/versioned/fake/register.go
git checkout origin/main -- pkg/client/resolution/clientset/versioned/scheme/register.go
git checkout origin/main -- pkg/client/resource/clientset/versioned/fake/register.go
git checkout origin/main -- pkg/client/resource/clientset/versioned/scheme/register.go
```

afrittoli:
This build failure looks more like a failure of the test itself, I wonder if we expect to guarantee the order of modules in modules.txt... @vdemeester wdyt?

```
Diffing /home/prow/go/src/github.com/tektoncd/pipeline against freshly generated codegen
diff -Naupr /home/prow/go/src/github.com/tektoncd/pipeline/vendor/modules.txt /home/prow/go/src/github.com/tektoncd/pipeline/tmpdiffroot.WBnLEp/vendor/modules.txt
--- /home/prow/go/src/github.com/tektoncd/pipeline/vendor/modules.txt	2022-11-15 10:59:53.216770693 +0000
+++ /home/prow/go/src/github.com/tektoncd/pipeline/tmpdiffroot.WBnLEp/vendor/modules.txt	2022-11-15 10:50:19.054106984 +0000
@@ -782,8 +782,8 @@ golang.org/x/sys/internal/unsafeheader
 golang.org/x/sys/plan9
 golang.org/x/sys/unix
 golang.org/x/sys/windows
-golang.org/x/sys/windows/registry
 # golang.org/x/term v0.2.0
+golang.org/x/sys/windows/registry
 ## explicit; go 1.17
 golang.org/x/term
 # golang.org/x/text v0.4.0
/home/prow/go/src/github.com/tektoncd/pipeline is out of date. Please run hack/update-codegen.sh.
```

cmd/controller/main.go
```diff
@@ -102,17 +116,84 @@ func main() {
 		log.Fatal(http.ListenAndServe(":"+port, mux))
 	}()
 
+	// initialize opentelemetry
```
vdemeester, line 119: nit: we could probably extract this in a function to return the object we'll pass to the controller as well as the `defer` function for "cleanly shutdown".

config/controller.yaml
```diff
@@ -124,6 +124,13 @@ spec:
           value: /etc/ssl/certs
         - name: METRICS_DOMAIN
           value: tekton.dev/pipeline
+       # The following variables can be uncommented with correct values to enable Jaeger tracing
+       #- name: OTEL_EXPORTER_JAEGER_ENDPOINT
+       #  value: http://jaeger-collector.jaeger:14268/api/traces
+       #- name: OTEL_EXPORTER_JAEGER_USER
+       #  value: username
+       #- name: OTEL_EXPORTER_JAEGER_PASSWORD
+       #  value: password
```
vdemeester, line 133: If those are not set, is it a "no-op" or do we still go through some opentelemetry code (just without any endpoint) ? Asking this for a "performance" point of view.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -159,6 +163,13 @@ var (
 func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun) pkgreconciler.Event {
 	logger := logging.FromContext(ctx)
 	ctx = cloudevent.ToContext(ctx, c.cloudEventClient)
+	ctx = initTracing(ctx, c.tracerProvider, pr)
+	ctx, span := c.tracerProvider.Tracer(Tracer).Start(ctx, "PipelineRun:ReconcileKind")
+	defer span.End()
+
+	span.SetAttributes(
+		attribute.String("pipeline", pr.Name), attribute.String("namespace", pr.Namespace),
```
vdemeester, line 171: Shouldn't this be `pipelinerun` instead ? `pipeline` might be "confusing".

pkg/reconciler/pipelinerun/tracing.go
```diff
@@ -0,0 +1,106 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/trace"
+	"knative.dev/pkg/logging"
+)
+
+const (
+	// Tracer is the name of the tracer
+	Tracer = "PipelineRunReconciler"
```
vdemeester, line 34: Could we name this `TrackerName` or something ? It did confusing me a bit when reading the code (wasn't sure if it was a constant for the name, a global variable, …)

vdemeester (commented):
This uses annotations to track the trackerid for a given `PipelinerRun` (or `TaskRun`), am I right ? If that's correct, wouldn't it make more sense to use a field in the `status`. Mainly asking that because, usually, a reconciler for a given object only updates the `status`, and not `spec` or `metadata`. This has caused few bugs in `tektoncd/pipeline` already and we (at least I) are looking into removing that part (aka updating the labels and annotations of the currently reconciled `PipelineRun` or `TaskRun`).

Other than this, and a few "nit" comments, this looks good to me 👼🏼

afrittoli:
> This uses annotations to track the trackerid for a given `PipelinerRun` (or `TaskRun`), am I right ? If that's correct, wouldn't it make more sense to use a field in the `status`. Mainly asking that because, usually, a reconciler for a given object only updates the `status`, and not `spec` or `metadata`. This has caused few bugs in `tektoncd/pipeline` already and we (at least I) are looking into removing that part (aka updating the labels and annotations of the currently reconciled `PipelineRun` or `TaskRun`).
> 
> Other than this, and a few "nit" comments, this looks good to me 👼🏼

The challenge with using the status is that when a `PipelineRun` creates a `TaskRun` it must pass down the trace context to the `TaskRun` - the `PipelineRun` controller should not create a `TaskRun` with the status already pre-set, so the best alternative is to set it into an annotation.

One alternative could be for the  `PipelineRun` to store the trace context in its own status and for `TaskRun` to lookup the trace context from the owner's status, and then write it to its own status, but that makes it harder to create a `PipelineRun` or `TaskRun` that belong to some distribute trace that starts before the `PipelineRun` started. Thinking forward I imagine that we could have a `CloudEvent` sent to a Trigger event listener that includes a [trace context](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/cloudevents/). The `PipelineRun` created by that trigger could continue the existing distributed trace.

pkg/reconciler/pipelinerun/tracing.go
afrittoli: `TracerName` rather than `TrackerName`?

vdemeester: @afrittoli yes, typo on my part 🙃

vdemeester:
> The challenge with using the status is that when a `PipelineRun` creates a `TaskRun` it must pass down the trace context to the `TaskRun` - the `PipelineRun` controller should not create a `TaskRun` with the status already pre-set, so the best alternative is to set it into an annotation.

Ah good point. For passing information from the `PipelineRun` to `TaskRun`, annotation is fine, because *I feel*, the `TaskRun` wouldn't update it. But the `PipelineRun` itself shouldn't add annotation on its own object, so it could store the tracerid on its status.

> One alternative could be for the `PipelineRun` to store the trace context in its own status and for `TaskRun` to lookup the trace context from the owner's status, and then write it to its own status, but that makes it harder to create a `PipelineRun` or `TaskRun` that belong to some distribute trace that starts before the `PipelineRun` started. Thinking forward I imagine that we could have a `CloudEvent` sent to a Trigger event listener that includes a [trace context](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/cloudevents/). The `PipelineRun` created by that trigger could continue the existing distributed trace.

Ah good point as well. Then I think objects should write their own tracerid to their own status, but should read the annotation in case it's part of a bigger "trace". This mean to create a `PipelineRun` or a `TaskRun` part of a distrubuted trace, one could just set the annotation and the reconciler would use it instead of generating one, and would write it in their status.

afrittoli:
> > The challenge with using the status is that when a `PipelineRun` creates a `TaskRun` it must pass down the trace context to the `TaskRun` - the `PipelineRun` controller should not create a `TaskRun` with the status already pre-set, so the best alternative is to set it into an annotation.
> 
> Ah good point. For passing information from the `PipelineRun` to `TaskRun`, annotation is fine, because _I feel_, the `TaskRun` wouldn't update it. But the `PipelineRun` itself shouldn't add annotation on its own object, so it could store the tracerid on its status.
> 
> > One alternative could be for the `PipelineRun` to store the trace context in its own status and for `TaskRun` to lookup the trace context from the owner's status, and then write it to its own status, but that makes it harder to create a `PipelineRun` or `TaskRun` that belong to some distribute trace that starts before the `PipelineRun` started. Thinking forward I imagine that we could have a `CloudEvent` sent to a Trigger event listener that includes a [trace context](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/cloudevents/). The `PipelineRun` created by that trigger could continue the existing distributed trace.
> 
> Ah good point as well. Then I think objects should write their own tracerid to their own status, but should read the annotation in case it's part of a bigger "trace". This mean to create a `PipelineRun` or a `TaskRun` part of a distrubuted trace, one could just set the annotation and the reconciler would use it instead of generating one, and would write it in their status.

OK, this makes sense to me. 
The annotation one would be for inherited trace context only, and the status one for be written in any case.
The consequence of this is also that annotations would be for the sole use of the controllers, while the status one is what users would rely on.

@kmjayadeep what do you think?

vdemeester:
> The consequence of this is also that annotations would be for the sole use of the controllers, while the status one is what users would rely on.

Yep 👍🏼

cmd/controller/main.go
vdemeester: I was thinking something like :

```go
trPipelineRun, trTaskRun, deferFn := initializeTracers(…)
defer deferFn()

func initializeTracers(…) (…) {
    # …
    return trPipelineRun, trTaskRun, func() { 
        // the content of the defer below
    }
}
```

And do all "tracer" related code in there. But it's a nit, I'm happy to keep it as is as well 😉

kmjayadeep:
I like the idea of using status for storing tracecontext :+1: . That makes it clear whether the trace was propogated from upstream or the trace starts in the CR.

So the new logic would be like the following
* Check if CR status has the trace context already set, if yes use it for tracing
* Check if trace context annotation exists for the CR, if yes use it for tracing and update it in status subresource
* If not present in annotation or status, create a new tracecontext and update in status

Upstream services can add the trace context in pipelinerun/taskrun annotations (eg: triggers), and downstream services can use the trace context from status field (eg: notifications)

If we all agree on this approach, I will make the changes along with documentation changes.

kmjayadeep:
@vdemeester @afrittoli I have updated the implementation as per the discussion. Looking forward to your feedback!

jerop:
@kmjayadeep - moving this PR to the next mileston

cmd/controller/main.go
afrittoli: It sounds like something we could fix in a follow-up?

docs/pipeline-api.md
```diff
@@ -4569,6 +4569,9 @@ reasons that emerge from underlying resources are not included here</p>
 </tr><tr><td><p>&#34;TaskRunImagePullFailed&#34;</p></td>
 <td><p>TaskRunReasonImagePullFailed is the reason set when the step of a task fails due to image not being pulled</p>
 </td>
+</tr><tr><td><p>&#34;TaskRunResultLargerThanAllowedLimit&#34;</p></td>
+<td><p>TaskRunReasonResultLargerThanAllowedLimit is the reason set when one of the results exceeds its maximum allowed limit of 1 KB</p>
+</td>
```
afrittoli, line 4574: It looks like these were missed in some previous PR, but I wonder by CI did not catch them.
@vdemeester @abayer any idea?

docs/pipeline-api.md
```diff
@@ -6347,15 +6350,17 @@ string
...
 </td>
 <td>
 <em>(Optional)</em>
 <p>KMS contains the KMS url of the public key
-Supported formats differ based on the KMS system used.</p>
+Supported formats differ based on the KMS system used.
+One example of a KMS url could be:
+gcpkms://projects/[PROJECT]/locations/[LOCATION]&gt;/keyRings/[KEYRING]/cryptoKeys/[KEY]/cryptoKeyVersions/[KEY_VERSION]
+For more examples please refer <a href="https://docs.sigstore.dev/cosign/kms_support">https://docs.sigstore.dev/cosign/kms_support</a>.
+Note that the KMS is not supported yet.</p>
```
afrittoli, line 6363: Ditto

pkg/reconciler/pipelinerun/tracing.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// SpanContext was created already
+	if pr.Status.SpanContext != nil && len(pr.Status.SpanContext) > 0 {
+		return pro.Extract(ctx, propagation.MapCarrier(pr.Status.SpanContext))
+	}
+
+	spanContext := make(map[string]string)
+
+	// SpanContext was propogated through annotations
+	if pr.Annotations != nil && pr.Annotations[SpanContextAnnotation] != "" {
```
afrittoli, line 55: Nice! These means that a `PipelineRun` too can receive an input context, which means we can support existing contexts for pipelines, like in the case of pipeline in pipeline, or a pipeline triggered via CloudEvent with a tracing context in it.

afrittoli (commented):
Thank you @kmjayadeep 
/lgtm

Issue #5478, opened 2022-09-12
Flake: PipelineRun unittests "received 2 events but 3 expected"


lbernick:
https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5466/pull-tekton-pipeline-unit-tests/1569343520609669120

TestReconcile_InvalidPipelineRuns/invalid-pipeline-run-resources-not-bound-shd-stop-reconciling:

```
{Failed            }
          }
    pipelinerun_test.go:1320: error in test pipeline-resources-not-bound: received 2 events but 3 expected. Found events: []string{"Normal Started ", "Warning Failed PipelineRun foo/pipeline-resources-not-bound doesn't bind Pipeline foo/a-fine-pipeline's PipelineResources correctly: Pipeline's declared required resources are missing from the PipelineRun: [a-resource]"}}
```

While PipelineResources are deprecated and will be removed eventually, I think this is a more general issue with our PipelineRun unit tests that rely on events. We do have a [FakeRecorded](https://github.com/tektoncd/pipeline/blob/3733e13697674076d28977fc97fd8ff8143dfbe1/vendor/k8s.io/client-go/tools/record/fake.go#L28) used in tests to record events, but we don't have an implementation of sending events for tests. (I think the issue is that the controller retries sending events, and the recorder may record duplicate events, which is what happened here.)

We could rewrite tests to filter out duplicate events, or we could provide a testing implementation of something that sends events and is allowed to block.

abayer:
Yeah, the bulk of our flaky unit tests that I've seen are either in `cmd/entrypoint` or event-based.

lbernick:
closed in #6063

Yongxuanzhang:
Hope https://github.com/tektoncd/pipeline/pull/6063 can be a fix for this issue, we will continue to monitor the ci and see if it works

PR #5752, opened 2022-11-14
fix flaky k8s event tests

Yongxuanzhang:
This commits fixes flaky k8s event test, #5478. The tests are flaky because we have a timer `eventsFromChannel` and it may lead to some events are not collected and cause errors. 

The flaky test can be reproduced if we set the timer to a smaller value:
https://github.com/tektoncd/pipeline/blob/38c739a5d0d9faff8c362462f8b6beee0458d978/test/events/events.go#L60
change it to ` timer := time.NewTimer(1 * time.Microsecond) `

/kind flake

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/events/events.go
+++ b/test/events/events.go
@@ -52,38 +52,26 @@ func CheckEventsUnordered(t *testing.T, eventChan chan string, testName string, 
 // expects to receive. The events must be received in the same order they appear in the
 // wantEvents list. Any extra or too few received events are considered errors.
 func eventsFromChannel(c chan string, wantEvents []string) error {
-	// We get events from a channel, so the timeout is here to avoid waiting
-	// on the channel forever if fewer than expected events are received.
-	// We only hit the timeout in case of failure of the test, so the actual value
-	// of the timeout is not so relevant, it's only used when tests are going to fail.
-	// on the channel forever if fewer than expected events are received
-	timer := time.NewTimer(10 * time.Millisecond)
+	// we loop the channel to collect all events, if the collected events are not
+	// expected events we will return error
+
 	foundEvents := []string{}
-	for ii := 0; ii < len(wantEvents)+1; ii++ {
-		// We loop over all the events that we expect. Once they are all received
-		// we exit the loop. If we never receive enough events, the timeout takes us
-		// out of the loop.
-		select {
-		case event := <-c:
-			foundEvents = append(foundEvents, event)
-			if ii > len(wantEvents)-1 {
-				return fmt.Errorf("received event \"%s\" but not more expected", event)
+	channelEvents := len(c)
+	for ii := 0; ii < channelEvents; ii++ {
+		event := <-c
+		foundEvents = append(foundEvents, event)
+		wantEvent := wantEvents[ii]
+		matching, err := regexp.MatchString(wantEvent, event)
+		if err == nil {
+			if !matching {
+				return fmt.Errorf("expected event \"%s\" but got \"%s\" instead", wantEvent, event)
 			}
-			wantEvent := wantEvents[ii]
-			matching, err := regexp.MatchString(wantEvent, event)
-			if err == nil {
-				if !matching {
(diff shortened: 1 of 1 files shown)


test/events/events.go
```diff
@@ -52,39 +52,27 @@ func CheckEventsUnordered(t *testing.T, eventChan chan string, testName string,
...
-					return fmt.Errorf("expected event \"%s\" but got \"%s\" instead", wantEvent, event)
-				}
-			} else {
-				return fmt.Errorf("something went wrong matching the event: %s", err)
-			}
-		case <-timer.C:
-			if len(foundEvents) != len(wantEvents) {
-				return fmt.Errorf("received %d events but %d expected. Found events: %#v", len(foundEvents), len(wantEvents), foundEvents)
+	channelEvents := len(c)
+	for ii := 0; ii < channelEvents; ii++ {
```
lbernick, line 60: I think this doesn't actually address the flake, since there's still the `time.Sleep` in `CheckEventsUnordered` that is buying time until the events occur. If that `Sleep` is removed, do these tests still pass?

This loops over the number of events already in the channel, rather than looping over the number of events we want to be in the channel (which would block until that number of events is sent). It's still possible to have a flake if the events are not sent in time.

I think it's also possible that extra events could be sent and we might not catch them (i.e. if another event is sent after the call to `len(c)`). We prevented this in your other PR by setting the length of the buffered channel equal to the expected number of events, but here we're [using 1000 as our buffer size](https://github.com/tektoncd/pipeline/blob/f6f9f3eb2109f8221dacf7eaf0d7cd3f919654a1/vendor/knative.dev/pkg/reconciler/testing/context.go#L53). If we don't want to change this upstream we might want a different way of ensuring that extra events aren't sent.

Yongxuanzhang: Another way to mitigate this flaky is to increase the timer value. It would be similar to adding sleep before the check. But it gives more time to collect the events from channel. If that is the direct reason of this flaky

Yongxuanzhang: So my guess is that the flaky is not coming from `we haven't sent the events`(this is why we add time.sleep before), it may be `we have sent the events in the channel, since we have a short time timer in the loop so we are not able to get the events within the time from the channel`

Yongxuanzhang: I guess this is the issue by adding more logging in this PR https://github.com/tektoncd/pipeline/pull/5747.
We do add the events in the channel but we don't get all of them

lbernick: I don't think we'd want to "fix" this flake by just increasing the timer value, since that doesn't address the underlying problem. One approach would be to remove the `time.sleep` and the timer and loop over the expected number of events. This would block forever if fewer events were sent than expected, but remove the flake when the correct number of events were sent. The main problem would be that it would not catch the case where extra events are sent (even if we set the buffer size equal to the expected number of events-- more events could be sent as soon as we start reading from the channel). Instead, would an approach similar to #5690 work here?

Yongxuanzhang: yes that's why I think it is a "mitigation" not a "fix". 
I prefer to remove the sleep and timer since I think timer might be the cause of flaky. 

I think it may work if we loop over the channel. The only concern you raised is that we may have some events not sent before we check them.  Correct me if I'm wrong:

1. We don't have any async process to send the k8s event, there is less possibility that this issue will happen
2. If this issue is true, then current code is still not able to fix that, current code only adds sleep time, they are just mitigation of this.

Yongxuanzhang: Oh sorry I was wrong, we don't use goroutines to send k8s events, but we do have one place to use the recorder to send failure event.https://github.com/tektoncd/pipeline/blob/f6f9f3eb2109f8221dacf7eaf0d7cd3f919654a1/pkg/reconciler/events/cloudevent/cloud_event_controller.go#L169

Yongxuanzhang:
https://github.com/tektoncd/pipeline/pull/5770

PR #5783, opened 2022-11-22
Exclude tags from being reported by knative.dev/test-infra/tools/go-ls-tags tool

cardil:
# Changes

Fixes https://github.com/knative/pkg/issues/2643
Fixes https://github.com/knative/actions/issues/86

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/.gotagsignore
+++ b/.gotagsignore
@@ -1,0 +1,2 @@
+tools
+windows


imjasonh:
This feels like an abstraction violation. Why should this repo need to maintain a file so that tests in some other repo don't fail? People adding new build tag scenarios to Tekton might (even "should") have no idea that some remote test could be affected by it.

Can this be maintained in knative/pkg instead? Or, should we just disable knative's failing downstream test of Tekton?

cardil:
> Can this be maintained in knative/pkg instead? Or, should we just disable knative's failing downstream test of Tekton?

Ofc it is violation. That's the purpose of the downstream test.

The downstream test benefit Tekton because you are using knative/pkg. If someone refactors something from knative/pkg that you are using, it will not pass the downstream test. So, it saves you from upstream breaking changes.

So, I'd say your call...

cardil:
Worth adding that knative/pkg doesn't have tags, so you always rely on latest commit being merged. That's why there is this downstream test in the first place.

vdemeester:
> Worth adding that knative/pkg doesn't have tags, so you always rely on latest commit being merged. That's why there is this downstream test in the first place.

We do not rely on the last commit though, most of the time, we "rely" on the *versionned* branches (release-*).

cardil:
@vdemeester said:

> _We do not rely on the last commit though, most of the time, we "rely" on the versionned branches (release-*)._

If so, maybe we can remove the downstream test for tektoncd/pipeline?

vdemeester:
> @vdemeester said:
> 
> > _We do not rely on the last commit though, most of the time, we "rely" on the versionned branches (release-*)._
> 
> If so, maybe we can remove the downstream test for tektoncd/pipeline?

Probably yes.

dibyom:
I don't think this PR is necessary anymore since the downstream tests got removed. Feel free to reopen if that's not the case

/close

Issue #5782, opened 2022-11-22
Add client/server compatibility tests


JeromeJu:
This issue tracks the improvement of upgrade test for preventing regression as noted in #5193:

```
Recently, we have had a few issues of regressions that were not detected by CI.

For example, https://github.com/tektoncd/pipeline/pull/4834 introduced changes that caused a new client to be incompatible with an old server.

In addition, some breaking changes were introduced where resources already stored on a cluster became invalid when the server version was upgraded. https://github.com/tektoncd/pipeline/pull/4779 added new defaults via the mutating webhook, and then https://github.com/tektoncd/pipeline/pull/4818 validated the existence of these defaults. However, existing resources didn't receive these defaults, so they failed validation.
```

We should add a step where we run integration tests on the older server version (before upgrading the server). This will address the first type of regression detailed here. When we do make a breaking change purposefully (after deprecating a field), I would imagine we'd remove tests for this functionality, so I don't think adding this check would stop us from removing deprecated fields.

- duplicate: [x] #6728 
- [ ] followup: to add LTS and supported releases for the e2e tests

dibyom:
>  the first type of regression detailed here.

@JeromeJu could you describe what the regression we are trying to catch is?

PR #5784, opened 2022-11-22
Add integration upgrade test on older version

JeromeJu:
# Changes
This commit adds the integration test on older version to prevent regression of current changes.

To prove that this PR could catch the regression:
- Reproducing #4913 and catch it with the existing change.
- Remove the pointer expression of [`pipelineRunSpec.taskRunSpecs.metadata`](https://github.com/tektoncd/pipeline/blob/f83cd1f63bb216c76a3af46d58614d517113d1c5/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L568).
- The previous version is set to the release version (v0.35.1) before the PR that introduces the regression.

Test case for regression:
- tested at https://github.com/JeromeJu/pipeline/tree/upgrade-regression-test
- Change ./test/e2e-tests-upgrade to install the previous version of #4913.
- run `./test/e2e-tests-upgrade.sh`, expect 
```
serviceaccount_test.go:163: Failed to create PipelineRun `pipeline-run-with-service-accounts-xtlybuvy`: admission webhook "webhook.pipeline.tekton.dev" denied the request: mutation failed: cannot decode incoming new object: json: unknown field "metadata"
```

/kind misc
fixes: #5782 case i

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e-tests-upgrade.sh
+++ b/test/e2e-tests-upgrade.sh
@@ -46,12 +46,15 @@ set +o pipefail
 header "Install the previous release of Tekton pipeline $PREVIOUS_PIPELINE_VERSION"
 install_pipeline_crd_version $PREVIOUS_PIPELINE_VERSION
 
+failed=0
+# The integration test for the older version to prevent regression on existing changes.
+go_test_e2e -timeout=20m ./test || failed=1
+
 # Upgrade to the current release.
 header "Upgrade to the current release of Tekton pipeline"
 install_pipeline_crd
 
 # Run the integration tests.
-failed=0
 go_test_e2e -timeout=20m ./test || failed=1
 
 # Run the post-integration tests.


JeromeJu:
Current status:
-   Trying to catch the regression of https://github.com/tektoncd/pipeline/issues/4913 as an example.
-   Reverting fixes https://github.com/tektoncd/pipeline/pull/4914/files to test if this shall be catched.

JeromeJu:
Updated the test case to prove this workable, and test case updated at https://github.com/JeromeJu/pipeline/tree/upgrade-regression-test. PTAL @lbernick :pray:

test/e2e-tests-upgrade.sh
```diff
@@ -46,12 +46,15 @@ set +o pipefail
 header "Install the previous release of Tekton pipeline $PREVIOUS_PIPELINE_VERSION"
 install_pipeline_crd_version $PREVIOUS_PIPELINE_VERSION
 
+failed=0
+# The integration test for the older version to prevent regression on existing changes.
+go_test_e2e -timeout=20m ./test || failed=1
```
afrittoli, line 51: It would be nice to update the initial comments (lines 21+) and describe what this script aims to test.

If I read this correctly, what you are doing here is to running the tests from `main` against the latest release of `Tekton`.
Unless tests are smart enough to skip themselves if a feature they rely on is not available, this test has a good chance of failing.

What we might consider doing is to run the tests from the latest release against `main`, which would ensure that no backward incompatible change was done.
Since there are cases where we do backward incompatible changes (in alpha and beta API versions, after the deprecation period), in these few cases we would have to call this out specifically in the tests, for instance we could have a mechanism to tell the tests to skip if the version of the target Tekton installation is greater than a certain value. 

This would still require to go backk and update the tests from the released versions though. An alternative could be to maintain in this script a version dependent exclusion list, to be passed to the `go test` command, with comments explaining the reason for the skip.

vdemeester: Note that we don't necessarily need to run all the test in the old "released" version 👼🏼. We just need to run enough testing to make sure it works (because ideally it runs a released version, that should have had all its own test run and successful when we released), upgrade and run the `main` test on the upgraded instance to make sure it behave as intended.

PR #5794, opened 2022-11-24
Add TaskRunStatus Conversion

JeromeJu:
# Changes
This commmit adds the conversion for taskRunStatus.
The `ResourcesResult` and `CloudEvents` are handled by being serialized to
annotations from v1beta1 to v1 and being deserialized to recpective
fields back into v1beta1 from v1.

/kind misc

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [n/a] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [n/a] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [n/a] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

PR #5797, opened 2022-11-24
Add PipelineRunStatus Conversion

JeromeJu:
# Changes
This commit adds the conversion for PipelineRunStatus. For the deprecated fields `PipelineRunStatus.Status.TaskRuns` and `PipelineRunStatus.Status.Runs`, they are converted into ChildStatusReference when converting from `v1beta1` to `v1`. The `taskRuns` and `runs` are serialized into annotations and would be populated by deserializations from v1 PipelineRunStatus to v1beta1 with `Full` or `Both` embeddedStatus.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
None
```

### DIFF
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
@@ -28,6 +28,15 @@ import (
 
 var _ apis.Convertible = (*PipelineRun)(nil)
 
+const (
+	// taskRunsAnnotationKey is used to restore PipelineRunStatus.Status.TaskRuns when converting
+	// down from v1 to v1beta1 as `status.taskRuns` is deprecated in v1 PipelineRunStatus.
+	taskRunsAnnotationKey = "tekton.dev/v1beta1TaskRuns"
+	// runsAnnotationKey is used to restore PipelineRunStatus.Status.Run when converting
+	// down from v1 to v1beta1 as `status.runs` is deprecated in v1 PipelineRunStatus.
+	runsAnnotationKey = "tekton.dev/v1beta1Runs"
+)
+
 // ConvertTo implements apis.Convertible
 func (pr *PipelineRun) ConvertTo(ctx context.Context, to apis.Convertible) error {
 	if apis.IsInDelete(ctx) {
@@ -37,6 +46,15 @@ func (pr *PipelineRun) ConvertTo(ctx context.Context, to apis.Convertible) error
 	case *v1.PipelineRun:
 		sink.ObjectMeta = pr.ObjectMeta
 		if err := serializePipelineRunResources(&sink.ObjectMeta, &pr.Spec); err != nil {
+			return err
+		}
+		if err := serializePipelineRunStatusTaskRuns(&sink.ObjectMeta, &pr.Status); err != nil {
+			return err
+		}
+		if err := serializePipelineRunStatusRuns(&sink.ObjectMeta, &pr.Status); err != nil {
+			return err
+		}
+		if err := pr.Status.ConvertTo(ctx, &sink.Status); err != nil {
 			return err
 		}
 		return pr.Spec.ConvertTo(ctx, &sink.Spec)
@@ -97,6 +115,15 @@ func (pr *PipelineRun) ConvertFrom(ctx context.Context, from apis.Convertible) e
 	case *v1.PipelineRun:
 		pr.ObjectMeta = source.ObjectMeta
 		if err := deserializePipelineRunResources(&pr.ObjectMeta, &pr.Spec); err != nil {
+			return err
+		}
+		if err := deserializePipelineRunStatusTaskRuns(&pr.ObjectMeta, &pr.Status); err != nil {
+			return err
+		}
+		if err := deserializePipelineRunStatusRuns(&pr.ObjectMeta, &pr.Status); err != nil {
+			return err
+		}
+		if err := pr.Status.ConvertFrom(ctx, &source.Status); err != nil {
 			return err
 		}
 		return pr.Spec.ConvertFrom(ctx, &source.Spec)
@@ -208,6 +235,155 @@ func (ptrs *PipelineTaskRunSpec) convertFrom(ctx context.Context, source v1.Pipe
 	ptrs.ComputeResources = source.ComputeResources
 }
 
+// ConvertTo implements apis.Convertible
+func (prs *PipelineRunStatus) ConvertTo(ctx context.Context, sink *v1.PipelineRunStatus) error {
+	sink.Status = prs.Status
+	sink.StartTime = prs.StartTime
+	sink.CompletionTime = prs.CompletionTime
+	sink.Results = nil
+	for _, pr := range prs.PipelineResults {
+		new := v1.PipelineRunResult{}
+		pr.convertTo(ctx, &new)
+		sink.Results = append(sink.Results, new)
+	}
+	if prs.PipelineSpec != nil {
+		sink.PipelineSpec = &v1.PipelineSpec{}
+		err := prs.PipelineSpec.ConvertTo(ctx, sink.PipelineSpec)
+		if err != nil {
+			return err
+		}
+	}
+	sink.SkippedTasks = nil
+	for _, st := range prs.SkippedTasks {
+		new := v1.SkippedTask{}
+		st.convertTo(ctx, &new)
+		sink.SkippedTasks = append(sink.SkippedTasks, new)
+	}
+	sink.ChildReferences = nil
+	for _, cr := range prs.ChildReferences {
+		new := v1.ChildStatusReference{}
+		cr.convertTo(ctx, &new)
+		sink.ChildReferences = append(sink.ChildReferences, new)
+	}
+	sink.FinallyStartTime = prs.FinallyStartTime
+	if prs.Provenance != nil {
+		new := v1.Provenance{}
+		prs.Provenance.convertTo(ctx, &new)
+		sink.Provenance = &new
+	}
+	// If embedded-status is set to "both", both ChildReferences and TaskRuns/Runs
+	// will be populated. In this case, use the value from ChildReferences.
+	if sink.ChildReferences == nil {
+		if prs.TaskRuns != nil {
+			sink.ChildReferences = append(sink.ChildReferences, convertTaskRunsToChildReference(ctx, prs.TaskRuns)...)
+		}
+		if prs.Runs != nil {
+			sink.ChildReferences = append(sink.ChildReferences, convertRunsToChildReference(ctx, prs.Runs)...)
+		}
+	}
+	return nil
+}
+
+// ConvertFrom implements apis.Convertible
+func (prs *PipelineRunStatus) ConvertFrom(ctx context.Context, source *v1.PipelineRunStatus) error {
+	prs.Status = source.Status
+	prs.StartTime = source.StartTime
+	prs.CompletionTime = source.CompletionTime
+	prs.PipelineResults = nil
+	for _, pr := range source.Results {
+		new := PipelineRunResult{}
+		new.convertFrom(ctx, pr)
+		prs.PipelineResults = append(prs.PipelineResults, new)
+	}
+	if source.PipelineSpec != nil {
+		newPipelineSpec := PipelineSpec{}
+		err := newPipelineSpec.ConvertFrom(ctx, source.PipelineSpec)
+		if err != nil {
+			return err
+		}
+		prs.PipelineSpec = &newPipelineSpec
+	}
+	prs.SkippedTasks = nil
+	for _, st := range source.SkippedTasks {
+		new := SkippedTask{}
+		new.convertFrom(ctx, st)
+		prs.SkippedTasks = append(prs.SkippedTasks, new)
+	}
+	prs.ChildReferences = nil
+	for _, cr := range source.ChildReferences {
+		new := ChildStatusReference{}
+		new.convertFrom(ctx, cr)
+		prs.ChildReferences = append(prs.ChildReferences, new)
+	}
+	prs.FinallyStartTime = source.FinallyStartTime
+	if source.Provenance != nil {
+		new := Provenance{}
+		new.convertFrom(ctx, *source.Provenance)
+		prs.Provenance = &new
+	}
+	return nil
+}
+
+func (prr PipelineRunResult) convertTo(ctx context.Context, sink *v1.PipelineRunResult) {
+	sink.Name = prr.Name
+	newValue := v1.ParamValue{}
+	prr.Value.convertTo(ctx, &newValue)
+	sink.Value = newValue
+}
+
+func (prr *PipelineRunResult) convertFrom(ctx context.Context, source v1.PipelineRunResult) {
+	prr.Name = source.Name
+	newValue := ParamValue{}
+	newValue.convertFrom(ctx, source.Value)
+	prr.Value = newValue
+}
+
+func (st SkippedTask) convertTo(ctx context.Context, sink *v1.SkippedTask) {
+	sink.Name = st.Name
+	sink.Reason = v1.SkippingReason(st.Reason)
+	sink.WhenExpressions = nil
+	for _, we := range st.WhenExpressions {
+		new := v1.WhenExpression{}
+		we.convertTo(ctx, &new)
+		sink.WhenExpressions = append(sink.WhenExpressions, new)
+	}
+}
+
+func (st *SkippedTask) convertFrom(ctx context.Context, source v1.SkippedTask) {
+	st.Name = source.Name
+	st.Reason = SkippingReason(source.Reason)
+	st.WhenExpressions = nil
+	for _, we := range source.WhenExpressions {
+		new := WhenExpression{}
+		new.convertFrom(ctx, we)
+		st.WhenExpressions = append(st.WhenExpressions, new)
+	}
+}
+
+func (csr ChildStatusReference) convertTo(ctx context.Context, sink *v1.ChildStatusReference) {
+	sink.TypeMeta = csr.TypeMeta
+	sink.Name = csr.Name
+	sink.PipelineTaskName = csr.PipelineTaskName
+	sink.WhenExpressions = nil
+	for _, we := range csr.WhenExpressions {
+		new := v1.WhenExpression{}
+		we.convertTo(ctx, &new)
+		sink.WhenExpressions = append(sink.WhenExpressions, new)
+	}
+}
+
+func (csr *ChildStatusReference) convertFrom(ctx context.Context, source v1.ChildStatusReference) {
+	csr.TypeMeta = source.TypeMeta
+	csr.Name = source.Name
+	csr.PipelineTaskName = source.PipelineTaskName
+	csr.WhenExpressions = nil
+	for _, we := range source.WhenExpressions {
+		new := WhenExpression{}
+		new.convertFrom(ctx, we)
+		csr.WhenExpressions = append(csr.WhenExpressions, new)
+	}
+}
+
 func serializePipelineRunResources(meta *metav1.ObjectMeta, spec *PipelineRunSpec) error {
 	if spec.Resources == nil {
 		return nil
@@ -223,6 +399,87 @@ func deserializePipelineRunResources(meta *metav1.ObjectMeta, spec *PipelineRunS
 	}
 	if len(resources) != 0 {
 		spec.Resources = resources
+	}
+	return nil
+}
+
+// convertTaskRunsToChildReference handles the deprecated `status.taskRuns` field PipelineRunTaskRunStatus.
+func convertTaskRunsToChildReference(ctx context.Context, taskRuns map[string]*PipelineRunTaskRunStatus) []v1.ChildStatusReference {
+	csrs := []v1.ChildStatusReference{}
+	for name, status := range taskRuns {
+		csr := v1.ChildStatusReference{}
+		// The apiVersion is defaulted to 'v1beta1' all taskRuns are populated by v1beta1
+		// PipelineRunStatus reconcilers when `embedded-status` is `Full` or `Both`.
+		csr.TypeMeta.APIVersion = "v1beta1"
+		csr.TypeMeta.Kind = "TaskRun"
+		csr.Name = name
+		csr.PipelineTaskName = status.PipelineTaskName
+		csr.WhenExpressions = nil
+		for _, we := range status.WhenExpressions {
+			new := v1.WhenExpression{}
+			we.convertTo(ctx, &new)
+			csr.WhenExpressions = append(csr.WhenExpressions, new)
+		}
+		csrs = append(csrs, csr)
+	}
+	return csrs
+}
+
+// convertRunsToChildReference handles the deprecated `status.runs`` field PipelineRunRunStatus.
+func convertRunsToChildReference(ctx context.Context, runs map[string]*PipelineRunRunStatus) []v1.ChildStatusReference {
+	csrs := []v1.ChildStatusReference{}
+	for name, status := range runs {
+		csr := v1.ChildStatusReference{}
+		// The apiViersion is set to 'v1alpha1' as Run is only in `v1alpha1`.
+		csr.TypeMeta.APIVersion = "v1alpha1"
+		csr.TypeMeta.Kind = "Run"
+		csr.Name = name
+		csr.PipelineTaskName = status.PipelineTaskName
+		csr.WhenExpressions = nil
+		for _, we := range status.WhenExpressions {
+			new := v1.WhenExpression{}
+			we.convertTo(ctx, &new)
+			csr.WhenExpressions = append(csr.WhenExpressions, new)
+		}
+		csrs = append(csrs, csr)
+	}
+	return csrs
+}
+
+func serializePipelineRunStatusTaskRuns(meta *metav1.ObjectMeta, status *PipelineRunStatus) error {
+	if status.TaskRuns == nil {
+		return nil
+	}
+	return version.SerializeToMetadata(meta, status.TaskRuns, taskRunsAnnotationKey)
+}
+
+func deserializePipelineRunStatusTaskRuns(meta *metav1.ObjectMeta, status *PipelineRunStatus) error {
+	taskRuns := make(map[string]*PipelineRunTaskRunStatus)
+	err := version.DeserializeFromMetadata(meta, &taskRuns, taskRunsAnnotationKey)
+	if err != nil {
+		return err
+	}
+	if len(taskRuns) != 0 {
+		status.TaskRuns = taskRuns
+	}
+	return nil
+}
+
+func serializePipelineRunStatusRuns(meta *metav1.ObjectMeta, status *PipelineRunStatus) error {
+	if status.Runs == nil {
+		return nil
+	}
+	return version.SerializeToMetadata(meta, status.Runs, runsAnnotationKey)
+}
+
+func deserializePipelineRunStatusRuns(meta *metav1.ObjectMeta, status *PipelineRunStatus) error {
+	runs := make(map[string]*PipelineRunRunStatus)
+	err := version.DeserializeFromMetadata(meta, &runs, runsAnnotationKey)
+	if err != nil {
+		return err
+	}
+	if len(runs) != 0 {
+		status.Runs = runs
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -208,6 +214,144 @@ func (ptrs *PipelineTaskRunSpec) convertFrom(ctx context.Context, source v1.Pipe
...
+	st.Reason = SkippingReason(source.Reason)
+	st.WhenExpressions = nil
+	for _, we := range source.WhenExpressions {
+		new := WhenExpression{}
+		new.convertFrom(ctx, we)
+		st.WhenExpressions = append(st.WhenExpressions, new)
+	}
+}
+
+func (csr ChildStatusReference) convertTo(ctx context.Context, sink *v1.ChildStatusReference) {
```
JeromeJu, line 333: Do we need to convert the TypeMeta for all the fields or only for ChildStatusReference?

vdemeester: Naively I would have say "for all" 🤔, this might be a "miss" for `task.Typemeta`

pkg/apis/pipeline/v1beta1/pipelinerun_conversion_test.go
```diff
@@ -143,6 +145,76 @@ func TestPipelineRunConversion(t *testing.T) {
...
+								Values:   []string{"foo", "bar"},
+							}},
+						}, {
+							Name:   "skipped-2",
+							Reason: v1beta1.MissingResultsSkip,
+						},
+					},
+					ChildReferences: []v1beta1.ChildStatusReference{
+						{
+							TypeMeta:         runtime.TypeMeta{Kind: "TaskRun"},
```
JeromeJu, line 194: Shall we have this kind here or that's what `typeMeta` should handle inline?

XinruZhang: We need to keep this

XinruZhang: Agreed with what @vdemeester mentioned, we'd need to convert all the `TypeMeta` **except for** those top-level CRDs (`Task`, `TaskRun`, `Pipeline`, `PipelineRun`, `Run`). The idea is those `TypeMeta` (outside of none-top-level CRDs) would be mainly for referring to a resource of a specific type and version (though there might be different case by case).

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -226,3 +377,41 @@ func deserializePipelineRunResources(meta *metav1.ObjectMeta, spec *PipelineRunS
 	}
 	return nil
 }
+
+func convertTaskRunsToChildReference(ctx context.Context, taskRuns map[string]*PipelineRunTaskRunStatus) []v1.ChildStatusReference {
+	csrs := []v1.ChildStatusReference{}
+	for name, status := range taskRuns {
+		csr := v1.ChildStatusReference{}
+		// TOOD: is apiVersion required here? and is it set to be `v1beta1`?
```
JeromeJu, line 385: Does it hold that `v1beta1` are always for TaskRunStatus type as a ChildStatusReference and `v1alpha1` for `RunStatus`?

abayer: Ok, so for `Run`s, yeah, it should always be `v1alpha1`, since there's never going to be a `v1beta1` for `Run` - we won't be converting from `v1alpha1.Run` to `v1beta1.CustomRun`, so the `v1alpha1.Run` object will still exist. For `TaskRun`s, I think the `apiVersion` should maybe be based on what the `PipelineRun` reconciler is creating - if the `v1beta1.TaskRun` has already been created, which is the case for any `TaskRun` referenced in `ChildReferences`, that's how we should refer to it. It may be _stored_ as a `v1.TaskRun` via conversion, but, if I'm remembering correctly, we can still reference it as `v1beta1`?

JeromeJu: Thanks Andrew, that makes perfect sense to me

pkg/apis/pipeline/v1beta1/provenance_conversion.go
```diff
@@ -0,0 +1,44 @@
...
+
+package v1beta1
+
+import (
+	"context"
+
+	v1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+)
+
+func (p Provenance) convertTo(ctx context.Context, sink *v1.Provenance) {
```
JeromeJu, line 22: This is to be kept for the completeness of PR function only.

Will be removed and rebased if https://github.com/tektoncd/pipeline/pull/5794 is merged first.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
abayer: I'd actually say we shouldn't convert the `TypeMeta` of the actual `ChildStatusReference`s, since that's identifying information about the resource being referenced, and that resource has already been created by the time it ends up in `ChildReferences`.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -208,6 +216,159 @@ func (ptrs *PipelineTaskRunSpec) convertFrom(ctx context.Context, source v1.Pipe
 	ptrs.ComputeResources = source.ComputeResources
 }
 
+// ConvertTo implements apis.Convertible
+func (prs *PipelineRunStatus) ConvertTo(ctx context.Context, sink *v1.PipelineRunStatus) error {
+	// TODO: handle ducktype conversion using knative func once https://github.com/knative/pkg/issues/2655 is fixed
```
lbernick, line 221: It looks like we're going to address this in pipelines now, no? can you link to the pipelines issue?

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -208,6 +216,159 @@ func (ptrs *PipelineTaskRunSpec) convertFrom(ctx context.Context, source v1.Pipe
...
+		new := v1.Provenance{}
+		prs.Provenance.convertTo(ctx, &new)
+		sink.Provenance = &new
+	}
+	if prs.TaskRuns != nil {
+		sink.ChildReferences = append(sink.ChildReferences, convertTaskRunsToChildReference(ctx, prs.TaskRuns)...)
+	}
+	if prs.Runs != nil {
+		sink.ChildReferences = append(sink.ChildReferences, convertRunsToChildReference(ctx, prs.Runs)...)
+	}
```
lbernick, line 264: If `embedded-status` is set to "both", both ChildReferences and TaskRuns/Runs will be populated, resulting in duplicate info in ChildReferences, which isn't what we want. Please also add a test for this

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -208,6 +216,159 @@ func (ptrs *PipelineTaskRunSpec) convertFrom(ctx context.Context, source v1.Pipe
...
+		new := SkippedTask{}
+		new.convertFrom(ctx, st)
+		prs.SkippedTasks = append(prs.SkippedTasks, new)
+	}
+	prs.ChildReferences = nil
+	for _, cr := range source.ChildReferences {
+		new := ChildStatusReference{}
+		new.convertFrom(ctx, cr)
+		prs.ChildReferences = append(prs.ChildReferences, new)
+	}
```
lbernick, line 302: I think we'd want to convert to TaskRuns/Runs if the feature flag specifies this, but unfortunately there's not enough information in ChildReferences to populate those fields without making api calls (which we don't want to do in a webhook). I'm concerned the PipelineRun controller might not function correctly if we don't populate the correct section of PR status. We can remove TaskRuns and Runs as early as January (and I think we should) but not sure what to do until then... @abayer wdyt?

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -226,3 +387,41 @@ func deserializePipelineRunResources(meta *metav1.ObjectMeta, spec *PipelineRunS
 	}
 	return nil
 }
+
+func convertTaskRunsToChildReference(ctx context.Context, taskRuns map[string]*PipelineRunTaskRunStatus) []v1.ChildStatusReference {
+	csrs := []v1.ChildStatusReference{}
+	for name, status := range taskRuns {
+		csr := v1.ChildStatusReference{}
+		csr.TypeMeta.APIVersion = "v1beta1"
```
lbernick, line 395: can you add a comment about why v1beta1 is the right version to use here?

pkg/apis/pipeline/v1beta1/pipelinerun_conversion_test.go
```diff
@@ -263,6 +402,105 @@ func TestPipelineRunConversionFromDeprecated(t *testing.T) {
 				},
 			},
 		},
+	}, {
+		name: "taskRuns",
```
lbernick, line 406: I would think the desired output here depends on the value of the embedded-status feature flag

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
abayer: So I think I may be thinking of "convert" differently - what I mean is that we should _preserve_ the `TypeMeta`s in `ChildStatusReference` when we convert the `PipelineRunStatus`, as I mentioned elsewhere on this PR, but somehow I got confused by this thread and repeated myself in a confusing manner. =)

XinruZhang: It seems like the term "convert" is used somehow differently 😅. 

As far as I understand, what vincent means, which is the idea I agreed, is we don't need to touch `Typemeta` of primitive CRDs (task, taskrun, pipeline, pipelinerun), but we need to **manually deal with** (which is called `convert` here, can be done by copying over etc) the `TypeMeta` that appear in other places. We are pretty much aligned if I'm not mistaken.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
JeromeJu: Thanks Lee, added.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -208,6 +214,158 @@ func (ptrs *PipelineTaskRunSpec) convertFrom(ctx context.Context, source v1.Pipe
...
+	sink.FinallyStartTime = prs.FinallyStartTime
+	if prs.Provenance != nil {
+		new := v1.Provenance{}
+		prs.Provenance.convertTo(ctx, &new)
+		sink.Provenance = &new
+	}
+	// TODO: to remove when TaskRuns and Runs are removed from TaskRunStatus.
+	// Currently only populate ChildReferences if `embedded-status` is set to `both`,
+	// where we have both ChildReferences and TaskRuns/Runs.
+	// Assumptions: childReferences contains the same status of the status TaskRuns and
```
JeromeJu, line 256: To confirm here this assumption holds that when `embedded-status` is set to both, the childReferences shall always matches the same status populated by `taskRuns` and `runs`.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -208,6 +214,158 @@ func (ptrs *PipelineTaskRunSpec) convertFrom(ctx context.Context, source v1.Pipe
...
+	// Runs populate.
```
lbernick, line 257:
```suggestion
	// If embedded-status is set to "both", both ChildReferences and TaskRuns/Runs
	// will be populated. In this case, use the value from ChildReferences.
```

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -226,3 +384,45 @@ func deserializePipelineRunResources(meta *metav1.ObjectMeta, spec *PipelineRunS
 	}
 	return nil
 }
+
+// convertTaskRunsToChildReference handles the deprecated TaskRuns field PipelineRunTaskRunStatus.
+// The apiViersion is set to 'v1beta1' as default, it could be changed if v1 is swapped as stored.
```
lbernick, line 389:
```suggestion
// The apiVersion is set to 'v1beta1' because we treat all TaskRuns within a PipelineRun
// as having the same API version as the PipelineRun.
```

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
JeromeJu: Thanks @lbernick .
https://github.com/tektoncd/pipeline/issues/5813 is created to track this down and coming up with a solution with testing and evidence.

JeromeJu:
https://github.com/tektoncd/pipeline/issues/5813 is created to track down the problems caused by the full `embedded-status` issue with `ChildReferences` in v1 being converted to `taskRuns` and `runs` status. This needs to be resolved before swapping the storage version to v1 as in a followup PR. It would be great if this could be confirmed @abayer @vdemeester @lbernick .

Currently no blockers to this PR if the findings and statements in #5813 is valid.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion_test.go
JeromeJu: Put in the related #5813, this shall be resolved with the decision made in #5813 about the handling of the full `embedded-status` in v1.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
JeromeJu: Thanks for the pointers moving forward, put the TODO at #L292

JeromeJu: Currently addressed this comment, but would love to have more inputs here @abayer @vdemeester :pray:

lbernick: My preferred approach is before merging this PR, to fail conversion if the embedded status is set to "full" or "both". the TODO should be for supporting conversion if the embedded status is set to either of those values.

My reasoning is that allowing conversion for the wrong value of embedded status might lead to confusing bugs in the PipelineRun controller, and I'm not sure yet whether this is the case. Failing conversion will at least make it clearer to devs what the problem is.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion_test.go
lbernick: Same as above: I think this does need to be addressed in this PR, but we can just fail the conversion if the feature flag is set to the incorrect value. The TODO is for creating a better solution in those cases.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
abayer: Yeah, failing seems reasonable to me. We're removing full/both embedded status in the near future (earliest possible date for switching the default from `full` to `minimal` is Jan 25, 2023, and then we can remove the flag and support for `full` and `both` completely a release later), so this isn't going to be a long-lived problem.

JeromeJu: Thanks @lbernick and @abayer . Updated.
Sorry for mistaking the failing as a follow-up option in the issue.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion_test.go
JeromeJu: Thanks Lee. That makes sense, also updated the issue.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
JeromeJu: Just to confirm this error handling would not require a release note.

JeromeJu:
/retest
flake: TestExamples/v1beta1/taskruns/dind-sidecar 
[link](https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5797/pull-tekton-pipeline-integration-tests/1598374107882196992)

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -208,6 +215,160 @@ func (ptrs *PipelineTaskRunSpec) convertFrom(ctx context.Context, source v1.Pipe
...
+	}
+	prs.SkippedTasks = nil
+	for _, st := range source.SkippedTasks {
+		new := SkippedTask{}
+		new.convertFrom(ctx, st)
+		prs.SkippedTasks = append(prs.SkippedTasks, new)
+	}
+	cfg := config.FromContextOrDefaults(ctx)
+	if cfg.FeatureFlags.EmbeddedStatus == config.FullEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus {
+		return fmt.Errorf("Invalid EmbeddedStatus value in v1: 'taskRuns' and 'runs' status are deprecated in v1, and cannot be populated by conversion from v1 childReferences")
```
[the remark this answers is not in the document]
abayer: I'd maybe word this as `PipelineRun .status.taskRuns and .status.runs fields do not exist in v1, and cannot be populated by conversion from a v1 PipelineRun to a v1beta1 PipelineRun when the "embedded-status" feature flag is not set to "minimal" or "both"`.

Also might be worth finally remembering to put `embedded-status` in `config/config-feature-flags.yml`, which I for some reason never remembered to do, so that the flag shows up in `kubectl get configmap -n tekton-pipelines feature-flags -o yaml`, but that can be a separate PR.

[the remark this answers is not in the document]
JeromeJu: Thanks, updated the error message. And created https://github.com/tektoncd/pipeline/issues/5824 to track this down.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
afrittoli: As a side note, having a mix of different embedded status in `PipelineRuns` on the sane cluster is possible today.
A user can use full status, and then switch config to minimal, which will lead to both kind of `PipelineRun` to exist in the cluster.
The controller needs to be able to handle those `PipelineRuns` regardless of the current status of the flag.
That said, I cannot think of anything that the `PipelineRun` controller would have to do with the status of completed `PipelineRuns`, so perhaps the only issue could be with `PipelineRuns` that are running when the configuration change happens. We probably don't test that scenario.

When doing conversions we should consider the current setting of the "embedded-status" flag at all, and provide full convertion back and forth between versions.
For the full status specifically, a way to address that could be to use annotations like we do for `PipelineResources` for instance.
Like `PipelineResources`, full status is something that is going away rather soon, but we are stuck with supporting it for a little while longer.

If we decide instead to drop full-status in the v1beta -> v1 convertion (which would be fine by me) we should document that in the release notes and in the v1beta -> v1 migration docs, so that users may know what to expect.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -208,6 +215,160 @@ func (ptrs *PipelineTaskRunSpec) convertFrom(ctx context.Context, source v1.Pipe
...
+	}
+	sink.FinallyStartTime = prs.FinallyStartTime
+	if prs.Provenance != nil {
+		new := v1.Provenance{}
+		prs.Provenance.convertTo(ctx, &new)
+		sink.Provenance = &new
+	}
+	// If embedded-status is set to "both", both ChildReferences and TaskRuns/Runs
+	// will be populated. In this case, use the value from ChildReferences.
+	// TODO: to handle conversion when `embedded-status` feature flag is set to full or both as in #5813
```
lbernick, line 256: I think it might make sense to return an error for this conversion as well, because I'm not sure how the PipelineRun controller would behave in this scenario.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion_test.go
```diff
@@ -285,3 +409,263 @@ func TestPipelineRunConversionFromDeprecated(t *testing.T) {
...
+
+			err := v1beta1PipelineRun.ConvertFrom(ctx, test.in)
+			if err.Error() != test.expectedError.Error() {
+				t.Errorf("expected error %v received %v", test.expectedError, err)
+			}
+		})
+	}
+}
+
+func withMinialEmbeddedStatus(ctx context.Context) context.Context {
```
lbernick, line 665:
```suggestion
func withMinimalEmbeddedStatus(ctx context.Context) context.Context {
```

pkg/apis/pipeline/v1beta1/pipelinerun_conversion_test.go
```diff
@@ -285,3 +409,263 @@ func TestPipelineRunConversionFromDeprecated(t *testing.T) {
...
+						TypeMeta:         runtime.TypeMeta{Kind: "TaskRun", APIVersion: "v1beta1"},
+						Name:             "tr-0",
+						PipelineTaskName: "ptn",
+						WhenExpressions:  []v1.WhenExpression{{Input: "default-value", Operator: "in", Values: []string{"val"}}},
+					}},
+				},
+			},
+		},
+		embeddedStatus: config.BothEmbeddedStatus,
+		expectedError:  fmt.Errorf("PipelineRun .status.taskRuns and .status.runs fields do not exist in v1, and cannot be populated by conversion from a v1 PipelineRun to a v1beta1 PipelineRun when the 'embedded-status' feature flag is not set to 'minimal'"),
```
lbernick, line 645: nit: I prefer asserting only that there is an error, and maybe a regex match on part of the error string, but not the entire error string itself. The reason is that it makes tests brittle: they will fail if we change only the error string and no other behavior, which doesn't provide much useful signal. However this is subjective/up to you

lbernick (approved):

pkg/apis/pipeline/v1beta1/pipelinerun_conversion_test.go
JeromeJu: Thanks Lee for the suggestion. Tried to refactor into this way but couldn't find a single phrase that's best suited for the regex matching piece here. And it seems the reason to have such long error message here is both sentences in the message is required:/ Would love to hear if any piece is preferred to be in the regex or I can keep an eye out for the refactorization on this during the pipelineRunStatus migration.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
JeromeJu: This is indeed a concern once we are changing the reconciler when swapping the storage version. Noted in #5813  For now I suspect if this is necessary as `v1beta1` is still the storage version, so when converting up would be fine in all cases:
- Full: we are converting `taskruns` and `runs` into `childReferences`
- Minimal:  `childReferences` -> `childReferences`
- Both: `childReferences` with `taskruns` and `runs` are converted into `childReferences`, but the underneath logics here are mostly on `childReferences`
Therefore, there seems no information loss here when converting up. But I did add into the migration doc that this would require notifications for users at https://github.com/tektoncd/pipeline/pull/5827.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
JeromeJu: Thanks @afrittoli for the conversation and the clarifications offline as well which are very helpful pinpointing the concerns moving forward. I have noted down the concerns about the knative way and controllers’ behaviours in the issue #5813 and the investigation would be done within the time frame of that , which is when swapping the storage version from v1beta1 to v1.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
JeromeJu: So failing the conversion here might be limiting the usage of v1 pipelineRunStatus here if `status.taskRuns` and `status.runs` are not required for the conversion up here.

pkg/apis/pipeline/v1beta1/pipelinerun_conversion_test.go
lbernick: I think it would be OK to just have `wantErr: bool`

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
lbernick: In that case, you can probably remove the TODO since this logic doesn't need to change?

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
dibyom: > When doing conversions we should not consider the current setting of the "embedded-status" flag at all, and provide full convertion back and forth between versions.

+1 - Conversion should not cause loss of information.

> If we decide instead to drop full-status in the v1beta -> v1 convertion (which would be fine by me) we should document that in the release notes and in the v1beta -> v1 migration docs, so that users may know what to expect.

This sounds good to me!

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
XinruZhang: @lbernick wanna check if my understanding is correct here:

Let say the storage version is `v1beta1`, an end user submits a `v1.PipelineRun`. The process that we handle it would be
1. The conversion webhook controller converts `v1` resource to `v1beta1` and store it in the etcd
2. The pipelinerun reconciler picks up the `v1beta1.Resource` (because it listens to the beta resource changes), and starts to reconcile it, store the updated `v1beta1` resource back to etcd
3. The end user wants to get the `v1.PipelineRun` resource back, the k8s api server receives the request, dispatch it to the conversion webhook, and conversion webhook convert the stored `v1beta1.PipelineRun` to `v1.PipelineRun`.

It this is how the request flow looks like, it seems like we don't have information loss from `v1beta1` to `v1` no matter what the feature flag `embedded-status`. Is that correct?

lbernick: Yeah that sounds right to me!

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
JeromeJu: And meanwhile for this PR, currently while the storage version is still in `v1beta1` and we might want to provide users with a headup when converting down from v1 to v1beta1 with `embedded-status` not set to `minimal`. And I will make sure to have the test cases for changing the `embedded-status` as in a followup PR addressing #5813

pkg/apis/pipeline/v1beta1/pipelinerun_conversion_test.go
JeromeJu: Thanks Lee, updated!

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
JeromeJu: Thanks for chiming in @XinruZhang for the clarifications!
And this TODO is removed @lbernick :)

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
JeromeJu: Updated the release note and the respective migration doc change is initiated at https://github.com/tektoncd/pipeline/pull/5827.

JeromeJu:
It is found in the integration tests that once we fail the conversion if there is `Full` `embedded`, it will error out all the v1 examples test since we are trying to create v1 CRD that needs to be converted to v1beta1 as storage version while the default embedded-status flag is full,

dibyom:
@JeromeJu PR description needs an update since we are converting vs returning an error now?

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
```diff
@@ -28,6 +28,11 @@ import (
 
 var _ apis.Convertible = (*PipelineRun)(nil)
 
+const (
+	taskRunsAnnotationKey = "tekton.dev/v1beta1TaskRuns"
```
dibyom, line 32: can we add a docstring that explains what these annotations do?

pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go
dibyom: @JeromeJu - I think it makes more sense to add the comment explaining why we need v1beta1 about the line that sets the TypeMeta vs adding it to the function signature

dibyom (approved):
Thanks @JeromeJu - Looks good to me now!
@afrittoli @vdemeester @abayer any other comments to resolve or can we merge this now?

abayer:
/lgtm

Let's do it!

Issue #5793, opened 2022-11-24
Improve `./test/examples_test.go` for V1 CRDS


JeromeJu:
The [examples_test](https://github.com/tektoncd/pipeline/blob/main/test/examples_test.go) needs to be updated with the V1 client version to have v1 Status testing:

For these to add status succeeded check:
- https://github.com/tektoncd/pipeline/blob/main/test/wait.go#L91
- https://github.com/tektoncd/pipeline/blob/main/test/wait.go#L167

Thanks @XinruZhang for the catch, confirmation and findings on this.

PR #5799, opened 2022-11-25
Add v1 clients for examples_test

JeromeJu:
# Changes
This PR contains 2 commits which accomplish the following:
- Add regarding v1 clients for the prerequisites of the v1-CRD release testing
- Update example_tests wait status setup for v1-CRD examples tests on `Status`

Fixes: https://github.com/tektoncd/pipeline/issues/5793
/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/test/wait.go
+++ b/test/wait.go
@@ -60,8 +60,9 @@ import (
 )
 
 const (
-	interval = 1 * time.Second
-	timeout  = 10 * time.Minute
+	interval       = 1 * time.Second
+	timeout        = 10 * time.Minute
+	v1beta1Version = "v1beta1"
 )
 
 // ConditionAccessorFn is a condition function used polling functions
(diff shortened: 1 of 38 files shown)


JeromeJu:
codebase not updated
/retest

test/wait.go
```diff
@@ -96,6 +96,25 @@ func WaitForTaskRunState(ctx context.Context, c *clients, name string, inState C
 	})
 }
 
+// WaitForV1TaskRunState polls the status of the TaskRun called name from client every
+// interval until inState returns `true` indicating it is done, returns an
+// error or timeout. desc will be used to name the metric that is emitted to
+// track how long it took for name to get into the state checked by inState.
+// TODO: needs to change all WaitForTaskRunState to v1 when switching the stored version
+func WaitForV1TaskRunState(ctx context.Context, c *clients, name string, inState ConditionAccessorFn, desc string) error {
```
lbernick, line 104: Instead of creating a separate function here, could the version be passed as an argument to `WaitForTaskRunState`? (same for PipelineRun)

JeromeJu: In most tests under ./test, `WaitForTaskRunState` and `WaitForPipelineRunState` are being used. And as put in #L103, they will all be swapped to `WaitForV1TaskRunState` and `WaitForV1PipelineRunState` once we change the stored version to v1. So I think this is just an intermediate state where we keep it and it shall not be necessary to add another param in this func's declaration?

lbernick: The timeline for removing these tests would be when we remove support for v1beta1, not when we swap the stored version to v1. I'd prefer to choose the option that keeps the codebase in a cleaner state the entire time. Removing an argument isn't really harder than removing a function and I don't think it's desirable to optimize for that compared to minimizing duplication.

JeromeJu:
/retest
codebase not updated

lbernick:
@JeromeJu what do you mean by "codebase not updated"? Are you experiencing issues with CI not running the latest changes you've pushed? or is the issue that the previous CI run posts a failed status while the newer CI run is still going? If the latter, I think the newer CI run should post its status when it completes, so you shouldn't need to run "retest". (If you do need to run "retest", this is an issue worth looking into.)

test/wait.go
JeromeJu: Thanks Lee. This makes perfect sense to me. Updated.

test/wait.go
```diff
@@ -82,17 +83,28 @@ func pollImmediateWithContext(ctx context.Context, fn func() (bool, error)) erro
...
-		if err != nil {
-			return true, err
+		switch version {
+		case "v1":
+			r, err := c.V1TaskRunClient.Get(ctx, name, metav1.GetOptions{})
+			if err != nil {
+				return true, err
+			}
+			return inState(&r.Status)
+		// TODO: default version needs to be changed to v1 once swapped to stored version
```
lbernick, line 100: I don't think this TODO is necessary (here or below)-- we'll probably remove v1beta1 support long before adding a v2 or other version

PR #5814, opened 2022-11-30
string results that can be parsed as json should not fail

### LINKED ISSUE #5803: result value type is not same as defined
# Expected Behavior
When defining the result type as string, the result value should be a string no matter what data it contains.
# Actual Behavior
If the string data is JSON then the result type will be recognized as object and cause pipelineRun to fail.

1. Deploy pipeline with the latest commit
2. Apply task pipeline and pipelineRun to cluster
```
kind: Pipeline
metadata:
  name: wrong-type-result-pipeline
spec:
  tasks:
    - name: wrongtype
      taskRef:
        name: wrong-type-task
  results:
    - name: string0
      value: $(tasks.wrongtype.results.string0)
      type: string
    - name: string1
      value: $(tasks.wrongtype.results.string1)
      type: string
---
kind: Task
metadata:
  name: wrong-type-task
spec:
  results:
    - name: string0
      type: string
    - name: string1
      type: string
  steps:
    - name: add
      image: alpine:latest
      script: |
        #!/bin/sh
        echo -n '{"key": "value"}' | tee $(results.string0.path);
        echo -n "key:value" | tee $(results.string1.path)
        # echo -n "key:value" | tee $(results.string0.path);
---
kind: PipelineRun
metadata:
  generateName: wrong-type-result-pipeline-
spec:
  pipelineRef:
    name: wrong-type-result-pipeline
```
3. Check the result
image

# Additional Info

- Kubernetes version:

```
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.2", GitCommit:"5835544ca568b757a8ecae5c153f317e5736700e", GitTreeState:"clean", BuildDate:"2022-09-21T14:33:49Z", GoVersion:"go1.19.1", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.2", GitCommit:"5835544ca568b757a8ecae5c153f317e5736700e", GitTreeState:"clean", BuildDate:"2022-09-21T14:27:13Z", GoVersion:"go1.19.1", Compiler:"gc", Platform:"linux/amd64"}
```

- Tekton Pipeline version:
```
Client version: 0.19.1
Pipeline version: devel
Dashboard version: v0.30.0
```
Deploy with latest commit b3e4f35712f on branch main

Yongxuanzhang:
# Changes

This commit fixes #5803. In tep75&76 we introduced object results. The object results is inferred from the emitted message, and if the message is of json format, then the result type could be object or array. We later validate the type with the type we defined in taskspec. So if the taskpec type is string it will fail validation. Based on the discussions in #5803, if the type in taskpec is string, we should keep the results as is and don't convert it to object or array.

/kind bug

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Task can emit string type results that can be parsed as json.  
```

### DIFF
--- a/pkg/pod/status.go
+++ b/pkg/pod/status.go
@@ -109,7 +109,7 @@ func SidecarsReady(podStatus corev1.PodStatus) bool {
 }
 
 // MakeTaskRunStatus returns a TaskRunStatus based on the Pod's status.
-func MakeTaskRunStatus(ctx context.Context, logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev1.Pod, kubeclient kubernetes.Interface) (v1beta1.TaskRunStatus, error) {
+func MakeTaskRunStatus(ctx context.Context, logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev1.Pod, kubeclient kubernetes.Interface, ts *v1beta1.TaskSpec) (v1beta1.TaskRunStatus, error) {
 	trs := &tr.Status
 	if trs.GetCondition(apis.ConditionSucceeded) == nil || trs.GetCondition(apis.ConditionSucceeded).Status == corev1.ConditionUnknown {
 		// If the taskRunStatus doesn't exist yet, it's because we just started running
@@ -141,7 +141,7 @@ func MakeTaskRunStatus(ctx context.Context, logger *zap.SugaredLogger, tr v1beta
 	}
 
 	var merr *multierror.Error
-	if err := setTaskRunStatusBasedOnStepStatus(ctx, logger, stepStatuses, &tr, pod.Status.Phase, kubeclient); err != nil {
+	if err := setTaskRunStatusBasedOnStepStatus(ctx, logger, stepStatuses, &tr, pod.Status.Phase, kubeclient, ts); err != nil {
 		merr = multierror.Append(merr, err)
 	}
 
@@ -152,9 +152,18 @@ func MakeTaskRunStatus(ctx context.Context, logger *zap.SugaredLogger, tr v1beta
 	return *trs, merr.ErrorOrNil()
 }
 
-func setTaskRunStatusBasedOnStepStatus(ctx context.Context, logger *zap.SugaredLogger, stepStatuses []corev1.ContainerStatus, tr *v1beta1.TaskRun, podPhase corev1.PodPhase, kubeclient kubernetes.Interface) *multierror.Error {
+func setTaskRunStatusBasedOnStepStatus(ctx context.Context, logger *zap.SugaredLogger, stepStatuses []corev1.ContainerStatus, tr *v1beta1.TaskRun, podPhase corev1.PodPhase, kubeclient kubernetes.Interface, ts *v1beta1.TaskSpec) *multierror.Error {
 	trs := &tr.Status
 	var merr *multierror.Error
+
+	// collect results from taskrun spec and taskspec
+	specResults := []v1beta1.TaskResult{}
+	if tr.Spec.TaskSpec != nil {
+		specResults = append(specResults, tr.Spec.TaskSpec.Results...)
+	}
+	if ts != nil {
+		specResults = append(specResults, ts.Results...)
+	}
 
 	// Extract results from sidecar logs
 	sidecarLogsResultsEnabled := config.FromContextOrDefaults(ctx).FeatureFlags.ResultExtractionMethod == config.ResultExtractionMethodSidecarLogs
@@ -164,8 +173,9 @@ func setTaskRunStatusBasedOnStepStatus(ctx context.Context, logger *zap.SugaredL
 		if err != nil {
 			merr = multierror.Append(merr, err)
 		}
+
 		// populate task run CRD with results from sidecar logs
-		taskResults, pipelineResourceResults, _ := filterResultsAndResources(sidecarLogResults)
+		taskResults, pipelineResourceResults, _ := filterResultsAndResources(sidecarLogResults, specResults)
 		if tr.IsSuccessful() {
 			trs.TaskRunResults = append(trs.TaskRunResults, taskResults...)
 			trs.ResourcesResult = append(trs.ResourcesResult, pipelineResourceResults...)
@@ -191,7 +201,8 @@ func setTaskRunStatusBasedOnStepStatus(ctx context.Context, logger *zap.SugaredL
 					logger.Errorf("error extracting the exit code of step %q in taskrun %q: %v", s.Name, tr.Name, err)
 					merr = multierror.Append(merr, err)
 				}
-				taskResults, pipelineResourceResults, filteredResults := filterResultsAndResources(results)
+
+				taskResults, pipelineResourceResults, filteredResults := filterResultsAndResources(results, specResults)
 				if tr.IsSuccessful() {
 					trs.TaskRunResults = append(trs.TaskRunResults, taskResults...)
 					trs.ResourcesResult = append(trs.ResourcesResult, pipelineResourceResults...)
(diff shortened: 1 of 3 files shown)


vdemeester (approved):
/meow

pkg/pod/status.go
```diff
@@ -170,7 +170,15 @@ func setTaskRunStatusBasedOnStepStatus(logger *zap.SugaredLogger, stepStatuses [
 					logger.Errorf("error extracting the exit code of step %q in taskrun %q: %v", s.Name, tr.Name, err)
 					merr = multierror.Append(merr, err)
 				}
-				taskResults, pipelineResourceResults, filteredResults := filterResultsAndResources(results)
+				// collect results from taskspec
+				specResults := []v1beta1.TaskResult{}
+				if tr.Spec.TaskSpec != nil {
+					specResults = append(specResults, tr.Spec.TaskSpec.Results...)
```
chitrangpatel, line 176: Just wondering if this works if `tr.Spec.TaskSpec.Results` is nil?

pkg/pod/status.go
```diff
@@ -170,7 +170,15 @@ func setTaskRunStatusBasedOnStepStatus(logger *zap.SugaredLogger, stepStatuses [
...
+				}
+				if ts != nil {
+					specResults = append(specResults, ts.Results...)
```
chitrangpatel, line 179: same point as above here.

chitrangpatel: just thinking if you also need `&& ts.Results != nil` in the if condition.

pkg/pod/status.go
Yongxuanzhang: yes, tr.Spec.TaskSpec.Results is a list, so if it is an empty list it should be fine.

pkg/pod/status.go
Yongxuanzhang: ts.Results is also a list, so it should be fine if we have an empty list

pkg/pod/status.go
chitrangpatel: SGTM!

pkg/pod/status.go
chitrangpatel: SGTM!

chitrangpatel (approved):

Yongxuanzhang:
@chuangw6 sorry I forgot to cc you. please take a look

pkg/pod/status.go
```diff
@@ -224,22 +232,36 @@ func createMessageFromResults(results []v1beta1.PipelineResourceResult) (string,
...
+				v := v1beta1.ResultValue{}
+				err := v.UnmarshalJSON([]byte(r.Value))
+				if err != nil {
+					continue
+				}
+				taskRunResult = v1beta1.TaskRunResult{
+					Name:  r.Key,
+					Type:  v1beta1.ResultsType(v.Type),
+					Value: v,
+				}
```
chuangw6, line 264: It's interesting to see that UnmarshalJSON never returns an error. For example, if one writes an invalid json string to an object result, UnmarshalJSON will not return error, and instead set the result type to string, and return nil. This will cause the task run to fail because of `missmatched Types for these results` - declared is object but the taskrun  parses as string.

The following taskrun is an example. We do need to return some error and fail the taskrun, but I am curious if the `missmatch` error is desired/appropriate, or we should handle this case differently.
```yaml
kind: TaskRun
metadata:
  name: obj-result
spec:
  taskSpec:
    results:
      - name: echo-output
        type: object
        properties:
          key: {}
    steps:
      - name: write-invalid-object
        image: bash
        script: |
          set -e
          echo -n '{"key": "value"' | tee $(results.echo-output.path)
```

chuangw6 (commented):
Thanks @Yongxuanzhang !!

pkg/pod/status.go
Yongxuanzhang: yes very good point!
I think the previous problem is that we we do UnmarshalJSON we don't know what the type we want to Unmarshal to, so if it is an invalid json, it should work if the expected type is string (no error return).

But I think we may improve this, since in this PR we pass the results type to this function. And we may use the type to guide the UnmarshalJSON

pkg/pod/status.go
```diff
@@ -224,22 +232,36 @@ func createMessageFromResults(results []v1beta1.PipelineResourceResult) (string,
 	return string(bytes), nil
 }
 
-func filterResultsAndResources(results []v1beta1.PipelineResourceResult) ([]v1beta1.TaskRunResult, []v1beta1.PipelineResourceResult, []v1beta1.PipelineResourceResult) {
+func filterResultsAndResources(results []v1beta1.PipelineResourceResult, specResults []v1beta1.TaskResult) ([]v1beta1.TaskRunResult, []v1beta1.PipelineResourceResult, []v1beta1.PipelineResourceResult) {
 	var taskResults []v1beta1.TaskRunResult
 	var pipelineResourceResults []v1beta1.PipelineResourceResult
 	var filteredResults []v1beta1.PipelineResourceResult
+	neededTypes := make(map[string]string)
```
afrittoli, line 239: NIT: I guess this could be `map[string]v1beta1.ResultsTypeString`

afrittoli (commented):
Thank you for this.
A small nit but nothing blocking.
Would you mind updating the release notes to include what changes as opposed to what should be.
/approve

zhouhaibing089:
Thanks for fixing this. This is a blocker for us to upgrade to the latest release.

I wonder whether it makes sense to add notes to the released version (since the validation was added) with additional description like **Known Issues**.

```
Task can't emit string type results that can be parsed as json 
```

Yongxuanzhang:
> Thanks for fixing this. This is a blocker for us to upgrade to the latest release.
> 
> I wonder whether it makes sense to add notes to the released version (since the validation was added) with additional description like **Known Issues**.
> 
> ```
> Task can't emit string type results that can be parsed as json 
> ```

Sure. @zhouhaibing089  What did you mean by adding notes to the released version?

Yongxuanzhang:
@jerop Can we merge this into v0.43 btw? Thanks!

pkg/pod/status.go
```diff
@@ -152,10 +152,19 @@ func MakeTaskRunStatus(ctx context.Context, logger *zap.SugaredLogger, tr v1beta
 	return *trs, merr.ErrorOrNil()
 }
 
-func setTaskRunStatusBasedOnStepStatus(ctx context.Context, logger *zap.SugaredLogger, stepStatuses []corev1.ContainerStatus, tr *v1beta1.TaskRun, podPhase corev1.PodPhase, kubeclient kubernetes.Interface) *multierror.Error {
+func setTaskRunStatusBasedOnStepStatus(ctx context.Context, logger *zap.SugaredLogger, stepStatuses []corev1.ContainerStatus, tr *v1beta1.TaskRun, podPhase corev1.PodPhase, kubeclient kubernetes.Interface, ts *v1beta1.TaskSpec) *multierror.Error {
 	trs := &tr.Status
 	var merr *multierror.Error
 
+	// collect results from taskrun spec and taskspec
+	specResults := []v1beta1.TaskResult{}
```
bendory, line 160: maybe shorted to just `results`?

pkg/pod/status.go
```diff
@@ -245,22 +256,35 @@ func createMessageFromResults(results []v1beta1.PipelineResourceResult) (string,
 	return string(bytes), nil
 }
 
-func filterResultsAndResources(results []v1beta1.PipelineResourceResult) ([]v1beta1.TaskRunResult, []v1beta1.PipelineResourceResult, []v1beta1.PipelineResourceResult) {
+func filterResultsAndResources(results []v1beta1.PipelineResourceResult, specResults []v1beta1.TaskResult) ([]v1beta1.TaskRunResult, []v1beta1.PipelineResourceResult, []v1beta1.PipelineResourceResult) {
```
bendory, line 259: suggest `taskResults` b/c that's what it actually is?

pkg/pod/status.go
```diff
@@ -245,22 +256,35 @@ func createMessageFromResults(results []v1beta1.PipelineResourceResult) (string,
...
+			taskRunResult := v1beta1.TaskRunResult{}
+			if neededTypes[r.Key] == v1beta1.ResultsTypeString {
+				taskRunResult = v1beta1.TaskRunResult{
+					Name:  r.Key,
+					Type:  v1beta1.ResultsTypeString,
+					Value: *v1beta1.NewStructuredValues(r.Value),
+				}
+			} else {
+				v := v1beta1.ResultValue{}
+				err := v.UnmarshalJSON([]byte(r.Value))
```
bendory, line 279:
```suggestion
			if err := v.UnmarshalJSON([]byte(r.Value)); err != nil {
			    continue
			}
```

bendory: See Go style guide: https://go.dev/doc/effective_go#control-structures

pkg/pod/status.go
Yongxuanzhang: realized that taskResults is used for TaskRunResult

PR #5828, opened 2022-12-02
TEP-0127: Larger results using sidecar logs - feature flags

chitrangpatel:
Prior to this, we were extracting results from tasks via the termination messages which had a limit of only 4 KB per pod. If users had multiple results outputted by a task then they would have to share the 4KB limit.

We now run a dedicated sidecar that has access to the results of all the steps. This sidecar prints out the result and its content to stdout. The logs of the sidecar are parsed by the taskrun controller and the results updated instead of termination logs. We enforce an upper limit of 4KB per result (configurable) but there is no overall cumulative limit shared the results (users can have as many 4KB (or as configured) results as they desire). This PR addresses the feature-flags related changes of https://github.com/tektoncd/community/pull/887.
# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


docs/tasks.md
```diff
@@ -847,6 +848,17 @@ available size will less than 4096 bytes.
...
+
+This feature allows users to store up to 4 KB per result by default. Because we are not limited by the size of the termination messages, users can have as many results as they require (or until the CRD reaches its limit). If the size of a result exceeds this limit, then the TaskRun will be placed into a failed state with the following message: `Result exceeded the maximum allowed limit.`
+
+**Note**: If you require even larger results, you can specify a different upper limit per result by setting `max-result-size` feature flag to your desired size in bytes ([see instructions](./install.md#enabling-larger-results-using-sidecar-logs)). **CAUTION**: the larger you make the size, more likely will the CRD reach its max limit enforced by the `etcd` server leading to bad user experience.
+  
+
```
jerop, line 861: since we're only adding the feature flags in this PR, we can remove this documentation and only leave in the install.md docs

chitrangpatel: sure.

docs/install.md
```diff
@@ -477,6 +479,24 @@ the `feature-flags` ConfigMap alongside your Tekton Pipelines deployment via
...
+
+**Note**: to enable this feature, you need to grant `get` access to all `pods/log` to the `Tekton pipeline controller`. This means that the tekton pipeline controller has the ability to access the pod logs.
+
+1. Create a cluster role and rolebinding by applying the following spec to provide log access to `tekton-pipelines-controller`. 
+
+```
+kubectl apply -f config/enable-log-access-to-controller/
+```
+
+2. Set the `results-from` feature flag to use sidecar logs by setting `results-from: sidecar-logs` in the [configMap](#customizing-the-pipelines-controller-behavior).
```
jerop, line 496: please add the way to set it:

```suggestion
2. Set the `results-from` feature flag to use sidecar logs by setting `results-from: sidecar-logs` in the [configMap](#customizing-the-pipelines-controller-behavior).

kubectl patch cm feature-flags -n tekton-pipelines -p '{"data":{"results-from":"sidecar-logs"}}'
```

docs/install.md
```diff
@@ -477,6 +479,24 @@ the `feature-flags` ConfigMap alongside your Tekton Pipelines deployment via
...
+
+3. If you want the size per result to be something other than 4096 bytes, you can set the `max-result-size` feature flag in bytes by setting `max-result-size: 8192(whatever you need here)`. **Note:** The value you can set here cannot exceed the size of the CRD limit of 1.5 MB.
```
jerop, line 502: add the way to set it:

```shell
kubectl patch cm feature-flags -n tekton-pipelines -p '{"data":{"max-result-size":"<VALUE-IN-BYTES>"}}'
```

docs/install.md
jerop, line 496: add the way to set it:

```shell
kubectl patch cm feature-flags -n tekton-pipelines -p '{"data":{"results-from":"sidecar-logs"}}'
```

pkg/apis/config/feature_flags_test.go
```diff
@@ -201,6 +215,12 @@ func TestNewFeatureFlagsConfigMapErrors(t *testing.T) {
 		fileName: "feature-flags-invalid-embedded-status",
 	}, {
 		fileName: "feature-flags-invalid-resource-verification-mode",
+	}, {
+		fileName: "feature-flags-invalid-results-from",
+	}, {
+		fileName: "feature-flags-invalid-max-result-size-too-large",
+	}, {
+		fileName: "feature-flags-invalid-max-result-size-bad-value",
```
jerop, line 223: add a valid config file as well

jerop (commented):
@chitrangpatel thank you for splitting up the PR!

docs/install.md
chitrangpatel: done

docs/install.md
chitrangpatel: done.

pkg/apis/config/feature_flags_test.go
chitrangpatel: sure

jerop (approved):

pkg/apis/config/testdata/feature-flags-invalid-max-result-size-bad-value.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2021 The Tekton Authors
```
afrittoli, line 1: NIT: 2022

pkg/apis/config/testdata/feature-flags-invalid-max-result-size-too-large.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2021 The Tekton Authors
```
afrittoli, line 1: NIT: 2022

pkg/apis/config/testdata/feature-flags-invalid-results-from.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2021 The Tekton Authors
```
afrittoli, line 1: NIT: 2022

pkg/apis/config/testdata/feature-flags-results-via-sidecar-logs.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright 2021 The Tekton Authors
```
afrittoli, line 1: NIT: 2022

docs/install.md
```diff
@@ -477,6 +479,32 @@ the `feature-flags` ConfigMap alongside your Tekton Pipelines deployment via
 
 For beta versions of Tekton CRDs, setting `enable-api-fields` to "beta" is the same as setting it to "stable".
 
+## Enabling larger results using sidecar logs
```
afrittoli, line 482: The documentation here makes it look like the feature is already available - but it's actually not implemented yet.

docs/install.md
```diff
@@ -477,6 +479,32 @@ the `feature-flags` ConfigMap alongside your Tekton Pipelines deployment via
...
+
+To exceed this limit of 4096 bytes, you can enable larger results using sidecar logs. By enabling this feature, you will have a configurable limit (with a default of 4096 bytes) per result with no restriction on the number of results. The results are still stored in the taskrun crd so they should not exceed the 1.5MB CRD size limit.
+
+**Note**: to enable this feature, you need to grant `get` access to all `pods/log` to the `Tekton pipeline controller`. This means that the tekton pipeline controller has the ability to access the pod logs.
+
+1. Create a cluster role and rolebinding by applying the following spec to provide log access to `tekton-pipelines-controller`. 
+
+```
+kubectl apply -f config/enable-log-access-to-controller/
+```
```
afrittoli, line 494: The Tekton release is produced doing:

```shell
ko resolve -R -f config/
```

We process all subfolders of `config` recursively. 
In fact, you can check the [latest nightly](https://storage.googleapis.com/tekton-releases-nightly/pipeline/latest/release.yaml) and you'll notice that the new `Role` and `RoleBinding` are already there and they would be installed by the standard Tekton installation.

This new `Role` to access logs was meant to be an opt-in, but it's installed by default right now.
We'll need to move those files to some other folder and also we will need to CI machinery to install them for testing this feature in CI on the alpha job.

chitrangpatel: Yes, this new role was meant to be opt-in. I'm happy to move it to another directory. Thanks for the catch. I was following how the remote resolvers were set up.

docs/install.md
chitrangpatel: We broke the PR into smaller ones. That must be why it got slipped in. I will remove this change and submit a PR. I will add it back in the very last PR where I'm wiring everything up.

docs/install.md
afrittoli: The remote resolvers are installed by default :)

chitrangpatel: aha! Makes sense then. I will move it to a parallel folder `optional_config`?

docs/install.md
jerop: thought the docs would be okay in the split up PR because the full implementation will be completed before the v0.43 release

PR #5831, opened 2022-12-02
Rename ConfigSource and Source to `RefSource`

### LINKED ISSUE #6350: Rename `ConfigSource`
In v0.42, we introduced a new field [`Provenance`](https://github.com/tektoncd/pipeline/blob/main/docs/pipeline-api.md#provenance-1) into TaskRun / PipelineRun status to pipe through some data needed for the provenance generation. It's an alpha feature gated by the dedicated feature flag `enable-provenance-in-status`.

For example, one subfield of it is currently named `configSource` for recording the source information of the remote task/pipeline definition.  But the issue with this naming is that it came from SLSA 0.2 spec. Since SLSA is evolving, we want to change this name to something not specific to SLSA.

chuangw6:
# Changes

fixes #6350 

/kind misc

Rename ConfigSource and Source to `RefSource`

Recently, we introduced a new field named provenance in TaskRun/PipelineRun CRD
status, and one of its subfields is named `ConfigSource`. Meanwhile, we introduced
`source` field into `ResolutionRequest` CRD status to pipe the value.

In this commit, we renamed both `configSource` and `source` to `RefSource`.

Reasoning: 
  - ConfigSource is the SLSA name and ties to a specific SLSA version.
It also makes this a leaky abstraction, i.e. we are naming fields in our API
after how we want to use them. Additionally, `config` isn't a concept that
exists in Tekton. 
  - The name `source` in ResolutionRequest status is too generic.

Backward compatibility: This PR doesn't remove the old field. Instead, it
keeps the old field while introducing the new name to give a release of warning.
The old name will be removed in future release for a smooth transition.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
The subfield `ConfigSource` of the `provenance` field in the PipelineRun/TaskRun Status and the field `source` in ResolutionRequest status are both renamed to `RefSource`.
Note: The old name was not removed from this release, but it has been marked as deprecated and will be removed in future release.
```


chuangw6:
Please take a look @bobcatfish . Thanks a lot!!

cc @jagathprakash @dibyom @wlynch

pkg/apis/pipeline/v1beta1/provenance.go
```diff
@@ -13,22 +13,20 @@ limitations under the License.
...
 
-// ConfigSource identifies the source where a resource came from.
-// This can include Git repositories, Task Bundles, file checksums, or other information
-// that allows users to identify where the resource came from and what version was used.
-type ConfigSource struct {
-	// URI indicates the identity of the source of the config.
+// Source contains the information that can uniquely identify where a remote
+// built definition came from i.e. Git repositories, Tekton Bundles in OCI registry
+// and hub.
+type Source struct {
```
wlynch, line 28: I'm worried this is going to be confused as an input source type - This probably need to be clearer that it's for output. If we're not okay with the Config naming, perhaps we should renaming this to something like `RefProvenance`, or consider moving the fields in ConfigSource directly into provenance?

What were other fields we were considering adding to provenance?

chuangw6: > I'm worried this is going to be confused as an input source type - This probably need to be clearer that it's for output.

that's a valid point. What about renaming the struct to be `RemoteSource`?  Similarly the field in provenance struct is also called `remoteSource`. 

> What were other fields we were considering adding to provenance?

Based on the discussion in Chains WG - Sep 29, 2022 ([meeting note](https://docs.google.com/document/d/1UVPSCDyNO-TzEFSv8jrqrEOF_FmV8NFuXncFm1gwmeY/edit#bookmark=id.rjj1d2yrtqfn)), we may extend this provenance field by adding fields like `input` and `output`, originally proposed in [TEP-109](https://github.com/tektoncd/community/blob/main/teps/0109-better-structured-provenance-retrieval-in-tekton-chains.md#using-run-status-to-generate-provenance-metadata).

jagathprakash: +1 to not naming it as source.
RemoteSource has the same problem as source, how about RemoteResource?

chuangw6: I feel like RemoteResource sounds more like the resource itself instead of the source information. What about `ResolvedSource`?

jagathprakash: Agree that RemoteResouce does not seem to suggest that it is source information.
ResolvedSource still feels like source code but resolved. 
Maybe ResourceSource conveys the information that this is the source of the resource but misses out on the fact that this source is remote.

afrittoli: ResourceSource conveys the meaning but also a bit of alliteration :D 
In terms of provenance, one could argue that any resources that is not embedded in the `*Run` object is remote, including those stored on the cluster.

dibyom: From this bunch I like ResolvedSource or ResolverSource

Yongxuanzhang, line 28: How about using type alias to keep backward compatible? 
like
https://github.com/tektoncd/pipeline/blob/a127323da31bcb933a04a6a1b5dbb6e0411e3dc1/pkg/apis/pipeline/v1beta1/param_types.go#L141

chuangw6: cc @bobcatfish for thoughts. Thanks!

chuangw6: It's true that it's the same name as SLSA provenance, but it's only a name in SLSA v0.2 specification, and SLSA v1 will be very likely to not have a field named 'configSource'. 

That said, I feel like it's worth making this field more generic instead of attaching it to a specific SLSA version.

Yongxuanzhang: Good to know! Then what is the name in V1?

dibyom: I think it's ok to do since this is a newly released alpha feature. We should definitely call it out in the release notes as a breaking change though

chuangw6: Sorry for the delay @Yongxuanzhang 

> Good to know! Then what is the name in V1?
It's likely to be part of [external parameters](https://github.com/MarkLodato/slsa/blob/03a066000033fad54af5ce5d88f8aa672b130237/docs/provenance/v1.0.proto#L56-L104).

> what we want to do in this PR is to seek a more generic name which works no matter how SLSA changes its' naming?
yes. I think that's the intention here.

bobcatfish: hey all! dropping in some ideas from a convo that @chuangw6 and I had about this today, pros and cons of some options for the name:

- If naming the one field isn't clear, maybe we could make a more complex object to make things clearer? isntead of `status.provenance.configsource` something like:
  - `status.provenance.task/pipeline.source` <-- not great to have a different field name depending on whether the run is a pipelinerun or taskrun 🙃 
  - status.provenance.tekton.source
- Config/ConfigSource
  - Con: not clear what "config" means in this context
  - Con: SLSA provenance format leaking into Tekton Pipelines API (and that format itself is shifting, e.g. the changes @chuangw6 has described around this field) / coupling Tekton Pipelines API to to SLSA provenance format
- Source
  - Con: Potential confusion for source used to build artifact
- RemoteSource & ResolvedSource & ResolverSource
  - Con: similar to Source (could be artifact source)
  - Pro: linked to remote resolution feature
- Build definition / build script
  - Pro: fits the way folks talk about 'builds'
  - Con: Pipeline/Task might not be building at all, could be just tests and other automation (and b/c of this 'build' is a bit of an anachronism)
- SpecSource / ResourceSource / DefinitionSource
- TektonSource

Maybe something like `TektonSource` could be a nice balance of being specific and not being confused with the source for the artifact?

dibyom: Is Chains using this yet? We should also reach out to the Chains WG to make sure they are aware!

chuangw6: `TektonSource`

chuangw6: `Source`

chuangw6: `SpecSource`

chuangw6: `TektonRemoteSource`

(to reflect that's from remote place.)

chuangw6: I've changed the name to `RefSource` based on the discussion in https://github.com/tektoncd/pipeline/issues/6350. PTAL.

chuangw6:
Update: I've changed the name to `RefSource` based on the discussion in https://github.com/tektoncd/pipeline/issues/6350. PTAL.

cc @dibyom , @Yongxuanzhang , @wlynch , @jagathprakash , @bobcatfish 

Thanks!

bobcatfish:
> Backward compatibility: This field was just introduced recently as an alpha
> feature gated by a dedicated feature flag. It should be okay being unstable :D.

Note that [technically we do try to give one release of warning](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#alpha-crds) for backwards incompatible changes 🙏  This does make it much more annoying to implement but can make the transition a bit smoother

pritidesai:
@tektoncd/core-maintainers, @chuangw6 is looking for reviews on this one!

chuangw6:
> > Backward compatibility: This field was just introduced recently as an alpha
> > feature gated by a dedicated feature flag. It should be okay being unstable :D.
> 
> Note that [technically we do try to give one release of warning](https://github.com/tektoncd/pipeline/blob/main/api_compatibility_policy.md#alpha-crds) for backwards incompatible changes 🙏 This does make it much more annoying to implement but can make the transition a bit smoother

Good point. Thanks @bobcatfish for the feedback. I've added the old field back and will remove it in future release then. Please take a look.

pkg/trustedresources/verify.go
```diff
@@ -45,9 +45,9 @@ const (
 // Skip the verification when no policies are found and trusted-resources-verification-no-match-policy is set to ignore or warn
 // Return an error when no policies are found and trusted-resources-verification-no-match-policy is set to fail,
 // or the resource fails to pass matched enforce verification policy
-// source is from ConfigSource.URI, which will be used to match policy patterns. k8s is used to fetch secret from cluster
-func VerifyTask(ctx context.Context, taskObj *v1beta1.Task, k8s kubernetes.Interface, source string, verificationpolicies []*v1alpha1.VerificationPolicy) error {
-	matchedPolicies, err := getMatchedPolicies(taskObj.TaskMetadata().Name, source, verificationpolicies)
+// refSourceURI is from RefSource.URI, which will be used to match policy patterns. k8s is used to fetch secret from cluster
```
Yongxuanzhang, line 48: Thank you for this change!

pkg/reconciler/taskrun/resources/taskref.go
```diff
@@ -217,9 +217,9 @@ type LocalTaskRefResolver struct {
 
 // GetTask will resolve either a Task or ClusterTask from the local cluster using a versioned Tekton client. It will
 // return an error if it can't find an appropriate Task for any reason.
-// TODO: if we want to set source for in-cluster task, set it here.
+// TODO: if we want to set RefSource for in-cluster task, set it here.
```
Yongxuanzhang, line 220: This issue is closed, do we still need it?

pkg/reconciler/taskrun/resources/taskref_test.go
```diff
@@ -476,9 +476,9 @@ func TestGetTaskFunc(t *testing.T) {
 				t.Error(diff)
 			}
 
-			// local cluster task and  bundle task have empty source for now. This may be changed in future.
-			if configSource != nil {
-				t.Errorf("expected configsource is nil, but got %v", configSource)
+			// local cluster task and  bundle task have empty RefSource for now. This may be changed in future.
```
Yongxuanzhang, line 479: Is this mentioned in any issues/design teps?

Yongxuanzhang (commented):
Thanks! I think RefSource is a good naming. 😄

pkg/reconciler/taskrun/resources/taskref.go
chuangw6: Thanks @Yongxuanzhang ! I've removed it.

pkg/reconciler/taskrun/resources/taskref_test.go
chuangw6: I should have made this clear in the comment.
- `bundle` task is meant to say the tasks that are resolved via the `taskref.bundle` approach instead of remote bundle resolver. Since that approach will be [deprecated](https://github.com/tektoncd/pipeline/issues/5514), I think it should be ok to have empty refSource. 
- `local cluster task` refers to the tasks living in the local cluster that are resolved by the localcluster resolver via `taskref.kind + name + apiversion` approach instead of remote cluster resolver.

PR #5834, opened 2022-12-05
TEP-0127: Larger results using sidecar logs - entrypoint, sidecar binary and injecting results sidecar

chitrangpatel:
This PR implements a part of [TEP-0127](https://github.com/tektoncd/community/blob/main/teps/0127-larger-results-via-sidecar-logs.md) - entrypoint, sidecar binary and injection of the results sidecar (see https://github.com/tektoncd/pipeline/pull/5695).

The sidecar binary relies on the Postfiles generated by the steps in `/tekton/run/<step>` to determine when a step has finished running. Once all the steps are done or if one of the step fails, the sidecar binary logs results to stdout.

Before approving this PR, review https://github.com/tektoncd/pipeline/pull/5828 

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Implemented sidecar binary and entrypoint related changes.
```

### DIFF
--- a/cmd/entrypoint/main.go
+++ b/cmd/entrypoint/main.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/containerd/containerd/platforms"
 	"github.com/tektoncd/pipeline/cmd/entrypoint/subcommands"
+	featureFlags "github.com/tektoncd/pipeline/pkg/apis/config"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
 	"github.com/tektoncd/pipeline/pkg/credentials"
 	"github.com/tektoncd/pipeline/pkg/credentials/dockercreds"
@@ -52,9 +53,10 @@ var (
 	breakpointOnFailure = flag.Bool("breakpoint_on_failure", false, "If specified, expect steps to not skip on failure")
 	onError             = flag.String("on_error", "", "Set to \"continue\" to ignore an error and continue when a container terminates with a non-zero exit code."+
 		" Set to \"stopAndFail\" to declare a failure with a step error and stop executing the rest of the steps.")
-	stepMetadataDir = flag.String("step_metadata_dir", "", "If specified, create directory to store the step metadata e.g. /tekton/steps/<step-name>/")
-	enableSpire     = flag.Bool("enable_spire", false, "If specified by configmap, this enables spire signing and verification")
-	socketPath      = flag.String("spire_socket_path", "unix:///spiffe-workload-api/spire-agent.sock", "Experimental: The SPIRE agent socket for SPIFFE workload API.")
+	stepMetadataDir        = flag.String("step_metadata_dir", "", "If specified, create directory to store the step metadata e.g. /tekton/steps/<step-name>/")
+	enableSpire            = flag.Bool("enable_spire", false, "If specified by configmap, this enables spire signing and verification")
+	socketPath             = flag.String("spire_socket_path", "unix:///spiffe-workload-api/spire-agent.sock", "Experimental: The SPIRE agent socket for SPIFFE workload API.")
+	resultExtractionMethod = flag.String("result_from", featureFlags.ResultExtractionMethodTerminationMessage, "The method using which to extract results from tasks. Default is using the termination message.")
 )
 
 const (
@@ -154,13 +156,14 @@ func main() {
 			stdoutPath: *stdoutPath,
 			stderrPath: *stderrPath,
 		},
-		PostWriter:          &realPostWriter{},
-		Results:             strings.Split(*results, ","),
-		Timeout:             timeout,
-		BreakpointOnFailure: *breakpointOnFailure,
-		OnError:             *onError,
-		StepMetadataDir:     *stepMetadataDir,
-		SpireWorkloadAPI:    spireWorkloadAPI,
+		PostWriter:             &realPostWriter{},
+		Results:                strings.Split(*results, ","),
+		Timeout:                timeout,
+		BreakpointOnFailure:    *breakpointOnFailure,
+		OnError:                *onError,
+		StepMetadataDir:        *stepMetadataDir,
+		SpireWorkloadAPI:       spireWorkloadAPI,
+		ResultExtractionMethod: *resultExtractionMethod,
 	}
 
 	// Copy any creds injected by the controller into the $HOME directory of the current
--- a/cmd/sidecarlogresults/main.go
+++ b/cmd/sidecarlogresults/main.go
@@ -1,0 +1,47 @@
+/*
+Copyright 2019 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"flag"
+	"log"
+	"os"
+	"strings"
+
+	"github.com/tektoncd/pipeline/internal/sidecarlogresults"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline"
+	"github.com/tektoncd/pipeline/pkg/pod"
+)
+
+func main() {
+	var resultsDir string
+	var resultNames string
+	flag.StringVar(&resultsDir, "results-dir", pipeline.DefaultResultPath, "Path to the results directory. Default is /tekton/results")
+	flag.StringVar(&resultNames, "result-names", "", "comma separated result names to expect from the steps running in the pod. eg. foo,bar,baz")
+	flag.Parse()
+	if resultNames == "" {
+		log.Fatal("result-names were not provided")
+	}
+	expectedResults := []string{}
+	for _, s := range strings.Split(resultNames, ",") {
+		expectedResults = append(expectedResults, s)
+	}
+	err := sidecarlogresults.LookForResults(os.Stdout, pod.RunDir, resultsDir, expectedResults)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
--- a/internal/sidecarlogresults/sidecarlogresults.go
+++ b/internal/sidecarlogresults/sidecarlogresults.go
@@ -1,0 +1,127 @@
+/*
+Copyright 2019 The Tekton Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sidecarlogresults
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"golang.org/x/sync/errgroup"
+)
+
+// SidecarLogResult holds fields for storing extracted results
+type SidecarLogResult struct {
+	Name  string
+	Value string
+}
+
+func fileExists(filename string) (bool, error) {
+	info, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false, nil
+	} else if err != nil {
+		return false, fmt.Errorf("error checking for file existence %w", err)
+	}
+	return !info.IsDir(), nil
+}
+
+func encode(w io.Writer, v any) error {
+	return json.NewEncoder(w).Encode(v)
+}
+
+func waitForStepsToFinish(runDir string) error {
+	steps := make(map[string]bool)
+	files, err := os.ReadDir(runDir)
+	if err != nil {
+		return fmt.Errorf("error parsing the run dir  %w", err)
+	}
+	for _, file := range files {
+		steps[filepath.Join(runDir, file.Name(), "out")] = true
+	}
+	for len(steps) > 0 {
+		for stepFile := range steps {
+			// check if there is a post file without error
+			exists, err := fileExists(stepFile)
+			if err != nil {
+				return fmt.Errorf("error checking for out file's existence %w", err)
+			}
+			if exists {
+				delete(steps, stepFile)
+				continue
+			}
+			// check if there is a post file with error
+			// if err is nil then either the out.err file does not exist or it does and there was no issue
+			// in either case, existence of out.err marks that the step errored and the following steps will
+			// not run. We want the function to break out with nil error in that case so that
+			// the existing results can be logged.
+			if exists, err = fileExists(fmt.Sprintf("%s.err", stepFile)); exists || err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// LookForResults waits for results to be written out by the steps
+// in their results path and prints them in a structured way to its
+// stdout so that the reconciler can parse those logs.
+func LookForResults(w io.Writer, runDir string, resultsDir string, resultNames []string) error {
+	if err := waitForStepsToFinish(runDir); err != nil {
+		return fmt.Errorf("error while waiting for the steps to finish  %w", err)
+	}
+	results := make(chan SidecarLogResult)
+	g := new(errgroup.Group)
+	for _, resultFile := range resultNames {
+		resultFile := resultFile
+
+		g.Go(func() error {
+			value, err := os.ReadFile(filepath.Join(resultsDir, resultFile))
+			if os.IsNotExist(err) {
+				return nil
+			} else if err != nil {
+				return fmt.Errorf("error reading the results file %w", err)
+			}
+			newResult := SidecarLogResult{
+				Name:  resultFile,
+				Value: string(value),
+			}
+			results <- newResult
+			return nil
+		})
+	}
+	channelGroup := new(errgroup.Group)
+	channelGroup.Go(func() error {
+		if err := g.Wait(); err != nil {
+			return fmt.Errorf("error parsing results: %w", err)
+		}
+		close(results)
+		return nil
+	})
+
+	for result := range results {
+		if err := encode(w, result); err != nil {
+			return fmt.Errorf("error writing results: %w", err)
+		}
+	}
+	if err := channelGroup.Wait(); err != nil {
+		return err
+	}
+	return nil
+}
--- a/internal/sidecarlogresults/sidecarlogresults_test.go
+++ b/internal/sidecarlogresults/sidecarlogresults_test.go
@@ -1,0 +1,145 @@
+package sidecarlogresults
+
+import (
+	"bytes"
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"sort"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/pipeline/test/diff"
+)
+
+func TestLookForResults_FanOutAndWait(t *testing.T) {
+	for _, c := range []struct {
+		desc    string
+		results []SidecarLogResult
+	}{{
+		desc: "multiple results",
+		results: []SidecarLogResult{{
+			Name:  "foo",
+			Value: "bar",
+		}, {
+			Name:  "foo2",
+			Value: "bar2",
+		}},
+	}} {
+		t.Run(c.desc, func(t *testing.T) {
+			dir := t.TempDir()
+			resultNames := []string{}
+			wantResults := []byte{}
+			for _, result := range c.results {
+				createResult(t, dir, result.Name, result.Value)
+				resultNames = append(resultNames, result.Name)
+				encodedResult, err := json.Marshal(result)
+				if err != nil {
+					t.Error(err)
+				}
+				// encode adds a newline character at the end.
+				// We need to do the same before comparing
+				encodedResult = append(encodedResult, '\n')
+				wantResults = append(wantResults, encodedResult...)
+			}
+			dir2 := t.TempDir()
+			go createRun(t, dir2, false)
+			got := new(bytes.Buffer)
+			err := LookForResults(got, dir2, dir, resultNames)
+			if err != nil {
+				t.Fatalf("Did not expect any error but got: %v", err)
+			}
+			// sort because the order of results is not always the same because of go routines.
+			sort.Slice(wantResults, func(i int, j int) bool { return wantResults[i] < wantResults[j] })
+			sort.Slice(got.Bytes(), func(i int, j int) bool { return got.Bytes()[i] < got.Bytes()[j] })
+			if d := cmp.Diff(wantResults, got.Bytes()); d != "" {
+				t.Errorf(diff.PrintWantGot(d))
+			}
+		})
+	}
+}
+
+func TestLookForResults(t *testing.T) {
+	for _, c := range []struct {
+		desc         string
+		resultName   string
+		resultValue  string
+		createResult bool
+		stepError    bool
+	}{{
+		desc:         "good result",
+		resultName:   "foo",
+		resultValue:  "bar",
+		createResult: true,
+		stepError:    false,
+	}, {
+		desc:         "empty result",
+		resultName:   "foo",
+		resultValue:  "",
+		createResult: true,
+		stepError:    true,
+	}, {
+		desc:         "missing result",
+		resultName:   "missing",
+		resultValue:  "",
+		createResult: false,
+		stepError:    false,
+	}} {
+		t.Run(c.desc, func(t *testing.T) {
+			dir := t.TempDir()
+			if c.createResult == true {
+				createResult(t, dir, c.resultName, c.resultValue)
+			}
+			dir2 := t.TempDir()
+			createRun(t, dir2, c.stepError)
+
+			var want []byte
+			if c.createResult == true {
+				// This is the expected result
+				result := SidecarLogResult{
+					Name:  c.resultName,
+					Value: c.resultValue,
+				}
+				encodedResult, err := json.Marshal(result)
+				if err != nil {
+					t.Error(err)
+				}
+				// encode adds a newline character at the end.
+				// We need to do the same before comparing
+				encodedResult = append(encodedResult, '\n')
+				want = encodedResult
+			}
+			got := new(bytes.Buffer)
+			err := LookForResults(got, dir2, dir, []string{c.resultName})
+			if err != nil {
+				t.Fatalf("Did not expect any error but got: %v", err)
+			}
+			if d := cmp.Diff(want, got.Bytes()); d != "" {
+				t.Errorf(diff.PrintWantGot(d))
+			}
+		})
+	}
+}
+
+func createResult(t *testing.T, dir string, resultName string, resultValue string) {
+	t.Helper()
+	resultFile := filepath.Join(dir, resultName)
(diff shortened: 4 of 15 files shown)


cmd/entrypoint/main.go
```diff
@@ -52,9 +53,10 @@ var (
...
-	stepMetadataDir = flag.String("step_metadata_dir", "", "If specified, create directory to store the step metadata e.g. /tekton/steps/<step-name>/")
-	enableSpire     = flag.Bool("enable_spire", false, "If specified by configmap, this enables spire signing and verification")
-	socketPath      = flag.String("spire_socket_path", "unix:///spiffe-workload-api/spire-agent.sock", "Experimental: The SPIRE agent socket for SPIFFE workload API.")
+	stepMetadataDir        = flag.String("step_metadata_dir", "", "If specified, create directory to store the step metadata e.g. /tekton/steps/<step-name>/")
+	enableSpire            = flag.Bool("enable_spire", false, "If specified by configmap, this enables spire signing and verification")
+	socketPath             = flag.String("spire_socket_path", "unix:///spiffe-workload-api/spire-agent.sock", "Experimental: The SPIRE agent socket for SPIFFE workload API.")
+	resultExtractionMethod = flag.String("result_extraction_method", featureFlags.ResultExtractionMethodTerminationMessage, "The method using which to extract results from tasks. Default is using the termination message.")
```
jerop, line 59: we renamed the feature flag:

```suggestion
	resultExtractionMethod = flag.String("result_from", featureFlags.ResultExtractionMethodTerminationMessage, "The method using which to extract results from tasks. Default is using the termination message.")
```

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// LookForResults waits for results to be written out by the steps
+// in their results path and prints them in a structured way to its
+// stdout so that the reconciler can parse those logs.
+func LookForResults(runDir string, resultsDir string, resultNames string) error {
+	err := waitForStepsToFinish(runDir)
+	if err != nil {
+		return err
+	}
```
jerop, line 82:
```suggestion
	if err := waitForStepsToFinish(runDir); err != nil {
		return err
	}
```

jerop (commented):
@chitrangpatel please add more details in the commit message and PR description about the entrypoint and sidecar binary

cmd/entrypoint/main.go
chitrangpatel: thanks!

internal/sidecarlogresults/sidecarlogresults.go
chitrangpatel: done

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -0,0 +1,71 @@
...
+		resultName   string
+		resultValue  string
+		createResult bool
+		runError     bool
+	}{{
+		desc:         "good result",
+		resultName:   "foo",
+		resultValue:  "bar",
+		createResult: true,
+		runError:     false,
```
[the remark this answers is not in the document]
chitrangpatel: Oh I think the word runError is mieading. It's not trying to say that the function call errors out. Sidecar log binary relies on the step status to know when a step is completed. The way it does that is using PostFiles generated by the step upon completion. It can be /tekton/run/step/out or /tekton/run/step/out.err ( in case the step errors out). runError is providing a test case such that the step successfully finishes or errors out. 

The results should be logged regardless.

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,111 @@
...
+	return nil
+}
+
+// LookForResults waits for results to be written out by the steps
+// in their results path and prints them in a structured way to its
+// stdout so that the reconciler can parse those logs.
+func LookForResults(runDir string, resultsDir string, resultNames string) error {
+	if err := waitForStepsToFinish(runDir); err != nil {
+		return err
```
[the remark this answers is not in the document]
chitrangpatel: I tried to but I couldn't produce an error that os.Stat finds. To trigger it, we need the file to be faulty (on the file system) or some other type of error (I'm not sure what).

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,111 @@
...
+	g := new(errgroup.Group)
+	for _, resultFile := range expectedResults {
+		resultFile := resultFile
+
+		g.Go(func() error {
+			value, err := ioutil.ReadFile(filepath.Join(resultsDir, resultFile))
+			if os.IsNotExist(err) {
+				return nil
+			} else if err != nil {
+				return err
```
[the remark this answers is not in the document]
chitrangpatel: same issue as above

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,111 @@
...
+				Value: string(value),
+			}
+			res, _ := json.Marshal(&newResult)
+			// print the result to stdout
+			fmt.Println(string(res))
+			return nil
+		})
+	}
+	if err := g.Wait(); err != nil {
+		return err
```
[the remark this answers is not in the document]
chitrangpatel: you mean add some in-code documentation?

chitrangpatel:
Thanks for your comments @Yongxuanzhang ! Much appreciated.

internal/sidecarlogresults/sidecarlogresults.go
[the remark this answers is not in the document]
Yongxuanzhang: like return fmt.Errof("some context about this err %v", err)

internal/sidecarlogresults/sidecarlogresults_test.go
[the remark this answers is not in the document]
Yongxuanzhang: Oh thanks! Sorry that is my mistake.

Yongxuanzhang:
Is there a tracking issue of TEP-0127? I can see there are several PRs but it's a bit confused to me to understand their relationship. Thanks!

chitrangpatel:
There is a TEP for it. You can see the original PR https://github.com/tektoncd/pipeline/pull/5695 from which we decided to split it. The PR description shows the order of PRs to review

internal/sidecarlogresults/sidecarlogresults.go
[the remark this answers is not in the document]
chitrangpatel: done!

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,111 @@
...
+	Name  string
+	Value string
+}
+
+func fileExists(filename string) (bool, error) {
+	info, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false, nil
+	} else if err != nil {
+		return false, fmt.Errorf("error checking for file existence %v", err)
```
wlynch, line 41:
```suggestion
		return false, fmt.Errorf("error checking for file existence %w", err)
```

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	return !info.IsDir(), nil
+}
+
+func waitForStepsToFinish(runDir string) error {
+	steps := make(map[string]bool)
+	files, err := ioutil.ReadDir(runDir)
```
wlynch, line 48: ioutil is deprecated - prefer using os. ReadDir instead https://pkg.go.dev/io/ioutil#ReadDir

internal/sidecarlogresults/sidecarlogresults.go
wlynch: (and also else where in this PR)

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,111 @@
...
+			if os.IsNotExist(err) {
+				return nil
+			} else if err != nil {
+				return fmt.Errorf("error reading the results file %v", err)
+			}
+			newResult := SidecarLogResult{
+				Name:  resultFile,
+				Value: string(value),
+			}
+			res, _ := json.Marshal(&newResult)
```
wlynch, line 101: check error

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,111 @@
...
+				return fmt.Errorf("error checking for out file's existence %v", err)
+			}
+			if exists {
+				delete(steps, stepFile)
+				continue
+			}
+			// check if there is a post file with error
+			if exists, err = fileExists(fmt.Sprintf("%s.err", stepFile)); exists || err != nil {
+				return err
+			}
```
wlynch, line 69: Double checking if this is intentional - if the error file exists nil is returned here.

At first glance this looks like it wants to return an error, so just want to make sure this is okay. (you may want to add some comments explaining the expected behavior).

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,111 @@
...
+			} else if err != nil {
+				return fmt.Errorf("error reading the results file %v", err)
+			}
+			newResult := SidecarLogResult{
+				Name:  resultFile,
+				Value: string(value),
+			}
+			res, _ := json.Marshal(&newResult)
+			// print the result to stdout
+			fmt.Println(string(res))
```
wlynch, line 103: I'd have the func take in a io.Writer - this will let you test that you're generating the correct output.

wlynch: you could also get fancier and use the writer to encode the output directly - https://go.dev/play/p/NSYd5DrDnd_R

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+}
+
+func createResult(dir string, resultName string, resultValue string, fileMode fs.FileMode) {
+	resultFile := filepath.Join(dir, resultName)
+	err := os.WriteFile(resultFile, []byte(resultValue), fileMode)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
```
wlynch, line 58: Pass in a testing.T + use t.Helper() so that any errors are captured in the right test output:

```suggestion
func createResult(t *testing.T, dir string, resultName string, resultValue string) {
	t.Helper()
	resultFile := filepath.Join(dir, resultName)
	err := os.WriteFile(resultFile, []byte(resultValue), 0644)
	if err != nil {
		t.Fatal(err)
	}
}
```

same with createRun.

pkg/pod/entrypoint.go
```diff
@@ -42,7 +42,10 @@ const (
 	entrypointBinary = binDir + "/entrypoint"
 
 	runVolumeName = "tekton-internal-run"
-	runDir        = "/tekton/run"
+
+	// RunDir is the directory in the containers where the step writes its
+	// post files to let the next step know to continue.
```
wlynch, line 47: nit: (since this directory contains more than just post files)

```suggestion
	// RunDir is the directory that contains runtime variable data for TaskRuns.
	// This includes files for handling container ordering, exit status codes, and more.
	// See [https://github.com/tektoncd/pipeline/blob/main/docs/developers/taskruns.md#tekton]
	// for more details.
```

pkg/apis/pipeline/reservedsidecar.go
```diff
@@ -0,0 +1,18 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package pipeline
+
+// ReservedResultsSidecarName is the name of the results sidecar that outputs the results to stdout
+// when the results-from feature-flag is set to "sidecar-logs".
+const ReservedResultsSidecarName = "tekton-log-results"
```
wlynch, line 18: Perhaps this should live in `pkg/pod` instead of `pkg/pipeline`? Most of the consts in `pkg/pipeline` deal with values that users may need to interact with in Tasks, while this feels like an internal implementation detail of how we construct pods.

wlynch (changes requested):
Thanks for breaking this out!

wlynch:
> There is a TEP for it. You can see the original PR #5695 from which we decided to split it. The PR description shows the order of PRs to review

It might still be useful to create a tracking issue in the pipelines repo for all of these. It can sometimes be tricky to track changes across repos.

jerop:
@chitrangpatel @Yongxuanzhang @wlynch - opened an issue to track the implementation: https://github.com/tektoncd/pipeline/issues/5851

internal/sidecarlogresults/sidecarlogresults.go
chitrangpatel: Thanks!

internal/sidecarlogresults/sidecarlogresults.go
chitrangpatel: Updated. Thanks for that link!

internal/sidecarlogresults/sidecarlogresults.go
chitrangpatel: done!

internal/sidecarlogresults/sidecarlogresults.go
chitrangpatel: Thanks! That indeed allowed me to test the expected output.

internal/sidecarlogresults/sidecarlogresults_test.go
chitrangpatel: done! Thanks!

pkg/apis/pipeline/reservedsidecar.go
chitrangpatel: good call! Done.

pkg/pod/entrypoint.go
chitrangpatel: thanks. I added it.

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,124 @@
...
+				Name:  resultFile,
+				Value: string(value),
+			}
+			fmt.Println(encode(os.Stdout, &newResult))
+			/*res, err := json.Marshal(&newResult)
+			if err != nil {
+				return fmt.Errorf("error marshaling the result as json %w", err)
+			}
+			// print the result to stdout
+			fmt.Println(string(res))*/
```
jerop, line 116: needs to be deleted?

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,118 @@
...
+	if err := waitForStepsToFinish(runDir); err != nil {
+		return fmt.Errorf("error while waiting for the steps to finish  %w", err)
+	}
+	expectedResults := []string{}
+	for _, s := range strings.Split(resultNames, ",") {
+		expectedResults = append(expectedResults, s)
+	}
+	g := new(errgroup.Group)
+	for _, resultFile := range expectedResults {
+		resultFile := resultFile
```
Yongxuanzhang, line 97: curious why do we need this

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,118 @@
...
+func waitForStepsToFinish(runDir string) error {
+	steps := make(map[string]bool)
+	files, err := os.ReadDir(runDir)
+	if err != nil {
+		return fmt.Errorf("error parsing the run dir  %w", err)
+	}
+	for _, file := range files {
+		steps[filepath.Join(runDir, file.Name(), "out")] = true
+	}
+	for len(steps) > 0 {
```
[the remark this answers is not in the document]
chitrangpatel: Yes. Its in an infinite loop until all steps are deleted by the inner loop. We need to constantly keep checking if the steps are done.

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,118 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+// LookForResults waits for results to be written out by the steps
+// in their results path and prints them in a structured way to its
+// stdout so that the reconciler can parse those logs.
+func LookForResults(runDir string, resultsDir string, resultNames string) error {
```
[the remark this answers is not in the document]
chitrangpatel: resultsDir can be specified to be something else https://github.com/tektoncd/pipeline/blob/3cc0be1fcaa00921f617866e59b178eda0c7c391/pkg/entrypoint/entrypointer.go#L87. Which is why I could not assume it to be /tekton/results. Default is `/tekton/results`.

internal/sidecarlogresults/sidecarlogresults.go
chitrangpatel: I think its because of the goRoutine below.

Yongxuanzhang: is that error able to be caught by test?

internal/sidecarlogresults/sidecarlogresults.go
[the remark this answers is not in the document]
chitrangpatel: I kept runDir similar to resultsDir for consistency. It also helps me with testing since I can create temp directories to write results to.

internal/sidecarlogresults/sidecarlogresults.go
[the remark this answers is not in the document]
Yongxuanzhang: is this tested in unit test or e2e test(or is it possible)? I tried to remove this and the unit test can pass

internal/sidecarlogresults/sidecarlogresults.go
chitrangpatel: I think the reason is similar to this: https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721#how-to-solve-this

(when we run table tests in parallel)

internal/sidecarlogresults/sidecarlogresults.go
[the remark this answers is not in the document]
chitrangpatel: The real flow will be tested in an e2e test but that will be in a separate PR. I currently have it as an example test but not an actual e2e test.

The current unit test does something similar (it creates runDir).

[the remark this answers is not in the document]
Yongxuanzhang: 👍

internal/sidecarlogresults/sidecarlogresults.go
Yongxuanzhang: oh yes! Thanks!

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -0,0 +1,105 @@
...
+			got := new(bytes.Buffer)
+			encode(got, &c.result)
+			want, err := json.Marshal(&c.result)
+			if err != nil {
+				t.Error(err)
+			}
+			// encode adds a newline character at the end.
+			// We need to do the same before comparing
+			want = append(want, []byte("\n")...)
+			if d := cmp.Diff(want, []byte(got.String())); d != "" {
```
wlynch, line 35: There's also https://pkg.go.dev/bytes#Buffer.Bytes, but I'm curious if the output would be easier to read as a byte buffer or a string? 🤔

wlynch (changes requested):

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}{{
+		desc: "good result",
+		result: SidecarLogResult{
+			Name:  "foo",
+			Value: "bar",
+		},
+	}} {
+		t.Run(c.desc, func(t *testing.T) {
+			got := new(bytes.Buffer)
+			encode(got, &c.result)
```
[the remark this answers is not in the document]
chitrangpatel: yes!

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,118 @@
...
+	expectedResults := []string{}
+	for _, s := range strings.Split(resultNames, ",") {
+		expectedResults = append(expectedResults, s)
+	}
+	g := new(errgroup.Group)
+	for _, resultFile := range expectedResults {
+		resultFile := resultFile
+
+		g.Go(func() error {
+			value, err := ioutil.ReadFile(filepath.Join(resultsDir, resultFile))
```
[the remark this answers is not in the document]
chitrangpatel: done

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,118 @@
...
+			if os.IsNotExist(err) {
+				return nil
+			} else if err != nil {
+				return fmt.Errorf("error reading the results file %w", err)
+			}
+			newResult := SidecarLogResult{
+				Name:  resultFile,
+				Value: string(value),
+			}
+			fmt.Println(encode(os.Stdout, &newResult))
```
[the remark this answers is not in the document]
chitrangpatel: good point! I will do that.

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -0,0 +1,105 @@
...
+			want, err := json.Marshal(&c.result)
+			if err != nil {
+				t.Error(err)
+			}
```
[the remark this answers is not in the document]
chitrangpatel: good point!

internal/sidecarlogresults/sidecarlogresults.go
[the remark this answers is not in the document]
chitrangpatel: good call!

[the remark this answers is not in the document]
chitrangpatel: I think this allowed me to put in a realistic test case where the binary actually wait for a postfile from the step.

pkg/apis/pipeline/reservedsidecar.go
chitrangpatel: Actually, I take it back. The reason I added it `pkg/pipeline` is because I need for validation for admission web hook where I ensure that the users don't provide the same name as that reserved by this. By keeping it in pkg/pod I was running into circular dependency issues. Because pkg/pod imports pkg/pipeline/v1beta1 and vice versa.

chitrangpatel: The part about admission webhook is in the very last PR: https://github.com/tektoncd/pipeline/pull/5695

chitrangpatel: Let me know if `pkg/pipeline` is better in this case and I will move it back.

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -0,0 +1,148 @@
...
+				// We need to do the same before comparing
+				encodedResult = append(encodedResult, '\n')
+				want = encodedResult
+			}
+			got := new(bytes.Buffer)
+			err := LookForResults(got, dir2, dir, c.resultName)
+			if err != nil {
+				t.Fatalf("Did not expect any error but got: %v", err)
+			}
+			if d := cmp.Diff(want, []byte(got.String())); d != "" {
```
wlynch, line 117:
```suggestion
			if d := cmp.Diff(want, got.Bytes()); d != "" {
```

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -0,0 +1,148 @@
...
+				}
+				// encode adds a newline character at the end.
+				// We need to do the same before comparing
+				encodedResult = append(encodedResult, '\n')
+				wantResults = append(wantResults, string(encodedResult))
+			}
+			dir2 := t.TempDir()
+			go createRun(t, dir2, false, true)
+			got := new(bytes.Buffer)
+			err := LookForResults(got, dir2, dir, strings.Join(resultNames, ","))
```
wlynch, line 49: If we're exporting this and think other clients will use this I'd lean towards taking in a []string rather than making clients need to encode as a csv.

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -0,0 +1,148 @@
...
+			if err != nil {
+				t.Fatalf("Did not expect any error but got: %v", err)
+			}
+			for _, res := range wantResults {
+				if !strings.Contains(got.String(), res) {
+					t.Errorf("Expected result %s to be in %s", res, got.String())
+				}
+			}
```
wlynch, line 57: prefer sorting the results and using cmp.Diff

internal/sidecarlogresults/sidecarlogresults_test.go
wlynch: or if you want to keep the behavior of being able to pass in a single name you could use `...string`

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+func createRun(t *testing.T, dir string, causeErr bool, sleep bool) {
+	t.Helper()
+	stepFile := filepath.Join(dir, "1")
+	_ = os.Mkdir(stepFile, 0755)
+	stepFile = filepath.Join(stepFile, "out")
+	if causeErr {
+		stepFile += ".err"
+	}
+	if sleep {
```
wlynch, line 141: This is always false. Remove?

(since this is setting up the results this won't actually cause variation in the goroutine behavior later)

wlynch (changes requested):
Mostly just final nits.

Otherwise /lgtm

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -0,0 +1,132 @@
...
+				Value: string(value),
+			}
+			results <- newResult
+			return nil
+		})
+	}
+	channelGroup := new(errgroup.Group)
+	channelGroup.Go(func() error {
+		if err := g.Wait(); err != nil {
+			return fmt.Errorf("error logging results to stdout %w", err)
```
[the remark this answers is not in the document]
chitrangpatel: done

internal/sidecarlogresults/sidecarlogresults_test.go
chitrangpatel: I agree, []string is much better. I had it as a csv because that's the input I was getting via cmd line flags. I moved the split logic into cmd/sidecarlogresults/main.go and pass in the []string to LookForResults.

internal/sidecarlogresults/sidecarlogresults_test.go
chitrangpatel: sure. Sounds good.

internal/sidecarlogresults/sidecarlogresults_test.go
chitrangpatel: sure

jerop (commented):
@chitrangpatel please update the commit message

jerop (approved):

cmd/sidecarlogresults/main.go
```diff
@@ -0,0 +1,47 @@
...
+	flag.StringVar(&resultsDir, "results-dir", pipeline.DefaultResultPath, "Path to the results directory. Default is /tekton/results")
+	flag.StringVar(&resultNames, "result-names", "", "comma separated result names to expect from the steps running in the pod. eg. foo,bar,baz")
+	flag.Parse()
+	if resultNames == "" {
+		log.Fatal("result-names were not provided")
+	}
+	expectedResults := []string{}
+	for _, s := range strings.Split(resultNames, ",") {
+		expectedResults = append(expectedResults, s)
+	}
```
wlynch, line 42:
```suggestion
	expectedResults := strings.Split(resultNames, ",")
```

wlynch: (I'm not going to make you wait for another round of review for this - you can follow up with another PR 🙏 )

wlynch (approved):
/lgtm

PR #5838, opened 2022-12-06
make stdout and stderror paths readable

chitrangpatel:
Prior to this, permission of the stdout and stderr file was 0600. Users would set this path to be the same as result path (see [here](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/taskruns/alpha/step-stream-results.yaml) for example). When using `sidecar-logs` to extract results, the sidecar would run into read permission error because of this. This PR loosens the permissions of the files generated this way to `0644` so that they can be read by other containers in the pod.

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Make stdout and stderr files readable by all.
```
/kind feature

### DIFF
--- a/cmd/entrypoint/runner.go
+++ b/cmd/entrypoint/runner.go
@@ -177,7 +177,7 @@ func newTeeReader(pipe func() (io.ReadCloser, error), path string) (*namedReader
 	if err := os.MkdirAll(filepath.Dir(path), os.ModePerm); err != nil {
 		return nil, fmt.Errorf("error creating parent directory: %w", err)
 	}
-	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
+	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644)
 	if err != nil {
 		return nil, fmt.Errorf("error opening %s: %w", path, err)
 	}


jerop (approved):

wlynch:
The main risk here is it allows steps to read the stdout/stderr of other steps, which may include secrets, other PII, etc. However, we weren't really preventing this before - IIUC this would work as is if the container user ids happen to match (which AFAIK we don't enforce any policies on).

Based on https://github.com/tektoncd/community/blob/main/teps/0011-redirecting-step-output-streams.md#motivation it looks like being able to read stdout across steps was intentional, but not sure if this opinion has changed (esp. given that based on this PR it hasn't been fully working for awhile now).

If we're okay with containers having full read access to other steps output, this change is fine. Otherwise we'd need to think about a more complex GID / mounting process and grant access that way.

afrittoli:
@wlynch makes a good point that we don't actively prevent this today. 
However we don't make it easier either, and this change would do so.

What makes me slightly nervous here is that it is possible to inject sidecars on pod creation (i.e. like Istio does) and those sidecars would have access to logs as well. That said, if someone has access to injecting sidecars, they probably have access to pod logs too, so maybe it's not an issue after all.

vdemeester:
> @wlynch makes a good point that we don't actively prevent this today. However we don't make it easier either, and this change would do so.
> 
> What makes me slightly nervous here is that it is possible to inject sidecars on pod creation (i.e. like Istio does) and those sidecars would have access to logs as well. That said, if someone has access to injecting sidecars, they probably have access to pod logs too, so maybe it's not an issue after all.

I tend to agree with @wlynch and @afrittoli, not really nervous about this.
So 👍🏼 from me.

bradbeck (approved):

dibyom (approved):

dibyom:
Merging this since we don't seem to have any concerns at the moment
/lgtm

PR #5840, opened 2022-12-06
TEP-0127: Larger results using sidecar logs - parsing sidecar logs to extract results

chitrangpatel:
# Changes

This PR addresses a part of [TEP-0127](https://github.com/tektoncd/community/blob/main/teps/0127-larger-results-via-sidecar-logs.md) - parsing logs to extract results into the task run CRD.

The taskrun reconciler fetches the logs of the injected sidecar container `sidecar-tekton-log-results`. The logs are already structured and can be un-marshalled via json. The reconciler parses these logs and adds the results to the task run CRD. If the size of the result is greater than the `max-result-size` configured by the user then the TaskRun fails with reason `TaskRunResultLargerThanAllowedLimit`.

Before approving this PR, review https://github.com/tektoncd/pipeline/pull/5834

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Parsing sidecar logs to extract results into the task run CRD.
```
/kind feature

### DIFF
--- a/internal/sidecarlogresults/sidecarlogresults.go
+++ b/internal/sidecarlogresults/sidecarlogresults.go
@@ -17,14 +17,24 @@ limitations under the License.
 package sidecarlogresults
 
 import (
+	"bufio"
+	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"golang.org/x/sync/errgroup"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
 )
+
+// ErrSizeExceeded indicates that the result exceeded its maximum allowed size
+var ErrSizeExceeded = errors.New("results size exceeds configured limit")
 
 // SidecarLogResult holds fields for storing extracted results
 type SidecarLogResult struct {
@@ -124,4 +134,60 @@ func LookForResults(w io.Writer, runDir string, resultsDir string, resultNames [
 		return err
 	}
 	return nil
+}
+
+// GetResultsFromSidecarLogs extracts results from the logs of the results sidecar
+func GetResultsFromSidecarLogs(ctx context.Context, clientset kubernetes.Interface, namespace string, name string, container string, podPhase corev1.PodPhase) ([]v1beta1.PipelineResourceResult, error) {
+	sidecarLogResults := []v1beta1.PipelineResourceResult{}
+	if podPhase == corev1.PodPending {
+		return sidecarLogResults, nil
+	}
+	podLogOpts := corev1.PodLogOptions{Container: container}
+	req := clientset.CoreV1().Pods(namespace).GetLogs(name, &podLogOpts)
+	sidecarLogs, err := req.Stream(ctx)
+	if err != nil {
+		return sidecarLogResults, err
+	}
+	defer sidecarLogs.Close()
+	maxResultLimit := config.FromContextOrDefaults(ctx).FeatureFlags.MaxResultSize
+	return extractResultsFromLogs(sidecarLogs, sidecarLogResults, maxResultLimit)
+}
+
+func extractResultsFromLogs(logs io.Reader, sidecarLogResults []v1beta1.PipelineResourceResult, maxResultLimit int) ([]v1beta1.PipelineResourceResult, error) {
+	scanner := bufio.NewScanner(logs)
+	buf := make([]byte, maxResultLimit)
+	scanner.Buffer(buf, maxResultLimit)
+	for scanner.Scan() {
+		result, err := parseResults(scanner.Bytes(), maxResultLimit)
+		if err != nil {
+			return nil, err
+		}
+		sidecarLogResults = append(sidecarLogResults, result)
+	}
+
+	if scanner.Err() != nil {
+		if errors.Is(scanner.Err(), bufio.ErrTooLong) {
+			return sidecarLogResults, ErrSizeExceeded
+		}
+		return nil, scanner.Err()
+	}
+	return sidecarLogResults, nil
+}
+
+func parseResults(resultBytes []byte, maxResultLimit int) (v1beta1.PipelineResourceResult, error) {
+	result := v1beta1.PipelineResourceResult{}
+	if len(resultBytes) > maxResultLimit {
+		return result, ErrSizeExceeded
+	}
+
+	var res SidecarLogResult
+	if err := json.Unmarshal(resultBytes, &res); err != nil {
+		return result, fmt.Errorf("Invalid result %w", err)
+	}
+	result = v1beta1.PipelineResourceResult{
+		Key:        res.Name,
+		Value:      res.Value,
+		ResultType: v1beta1.TaskRunResultType,
+	}
+	return result, nil
 }
--- a/internal/sidecarlogresults/sidecarlogresults_test.go
+++ b/internal/sidecarlogresults/sidecarlogresults_test.go
@@ -2,14 +2,22 @@ package sidecarlogresults
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/pipeline/test/diff"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	fakekubeclientset "k8s.io/client-go/kubernetes/fake"
 )
 
 func TestLookForResults_FanOutAndWait(t *testing.T) {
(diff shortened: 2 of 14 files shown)


internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -17,15 +17,26 @@ limitations under the License.
...
+	"github.com/tektoncd/pipeline/pkg/apis/config"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"golang.org/x/sync/errgroup"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 )
 
+// ErrorReasonMaxResultSizeExceeded indicates that the result exceeded its maximum allowed size
+var ErrorReasonMaxResultSizeExceeded = fmt.Errorf("%s", "MaxResultSizeExceeded")
```
bendory, line 38: 1) Go style prefers brief names, and convention is to end errors with `Err`: `SizeExceededErr`.
2) Give the user helpful details -- like the max size: `SizeExceededErr  = fmt.Errorf("results size >%d limit", maxResultLimit)`

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -186,6 +187,11 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, tr *v1beta1.TaskRun) pkg
 	// updates regardless of whether the reconciliation errored out.
 	if err = c.reconcile(ctx, tr, rtr); err != nil {
 		logger.Errorf("Reconcile: %v", err.Error())
+		if errors.Is(err, sidecarlogresults.ErrorReasonMaxResultSizeExceeded) {
+			message := fmt.Sprintf("TaskRun %q failed to finish because atleast one of its results exceeded the max allowed limit.", tr.Name)
```
bendory, line 191: type: `at least`

bendory, line 191: Prefer brevity for readability: "TaskRun %q failed: results exceeded size limit %d bytes"

internal/sidecarlogresults/sidecarlogresults.go
chitrangpatel: Thanks! I updated the message to "result size > configured limit". I cannot quantify it because it can change based on a value in the configMap.

pkg/reconciler/taskrun/taskrun.go
chitrangpatel: done

internal/sidecarlogresults/sidecarlogresults.go
chitrangpatel: Hmm.
I got a linter error in the build tests:
```
=========================================
==== TESTING IF GOLINT HAS BEEN DONE ====
=========================================
internal/sidecarlogresults/sidecarlogresults.go:38:5: error-naming: error var ResultSizeExceededErr should have name of the form ErrFoo (revive)
var ResultSizeExceededErr = fmt.Errorf("%s", "results size > configured limit")
    ^
==============================
==== GO LINT TESTS FAILED ====
==============================
```

Seems like I need to rename it to `ErrResultSizeExceeded` to follow `ErrFoo` convention.

bendory: Oops, looks like I got the convention backwards, it's `ErrFoo` (not `FooErr`) -- sorry about that!

But I still recommend brevity: `ErrSizeExceeded`.

pkg/pod/status.go
```diff
@@ -147,15 +152,26 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev
...
+			merr = multierror.Append(merr, err)
+		}
+		sidecarLogResults = slr
+	}
 	for _, s := range stepStatuses {
 		if s.State.Terminated != nil && len(s.State.Terminated.Message) != 0 {
 			msg := s.State.Terminated.Message
 
 			results, err := termination.ParseMessage(logger, msg)
+			results = append(results, sidecarLogResults...)
```
jerop, line 174: I'd have thought that it's either results from termination message or results from sidecar logs, not that we need to combine the results from both extraction methods 🤔

pkg/pod/status.go
```diff
@@ -175,12 +191,14 @@ func setTaskRunStatusBasedOnStepStatus(logger *zap.SugaredLogger, stepStatuses [
...
-				} else {
-					s.State.Terminated.Message = msg
+				if len(sidecarLogResults) == 0 {
+					msg, err = createMessageFromResults(filteredResults)
+					if err != nil {
+						logger.Errorf("%v", err)
+						merr = multierror.Append(merr, err)
+					} else {
+						s.State.Terminated.Message = msg
+					}
```
jerop, line 201: this check based on length of `sidecarLogResults` is confusing after already combining results from both extraction methods

pkg/pod/status.go
chitrangpatel: Its because termination messages can contain other information too. Like Start time etc. We haven't changed that logic here. Its only the results produced by steps that are not in the termination message.

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -121,6 +129,184 @@ func TestLookForResults(t *testing.T) {
...
+		Value:      "foo",
+		ResultType: 1,
+	}, {
+		Key:        "result2",
+		Value:      `{"IMAGE_URL":"ar.com", "IMAGE_DIGEST":"sha234"}`,
+		ResultType: 1,
+	}, {
+		Key:        "result3",
+		Value:      `["hello","world"]`,
+		ResultType: 1,
```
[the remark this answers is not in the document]
chitrangpatel: agreed!

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -125,3 +136,60 @@ func LookForResults(w io.Writer, runDir string, resultsDir string, resultNames [
 	}
 	return nil
 }
+
+// GetResultsFromSidecarLogs extracts results from the logs of the results sidecar
+func GetResultsFromSidecarLogs(ctx context.Context, clientset kubernetes.Interface, namespace string, name string, container string) ([]v1beta1.PipelineResourceResult, error) {
+	sidecarLogResults := []v1beta1.PipelineResourceResult{}
+	p, _ := clientset.CoreV1().Pods(namespace).Get(ctx, name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
chitrangpatel: thanks for point that out.

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -121,6 +129,184 @@ func TestLookForResults(t *testing.T) {
...
+	}
+	podLogs := []string{}
+	for _, r := range results {
+		res, _ := json.Marshal(&r)
+		podLogs = append(podLogs, string(res))
+	}
+	want := []v1beta1.PipelineResourceResult{{
+		Key:        "result1",
+		Value:      "foo",
+		ResultType: 1,
```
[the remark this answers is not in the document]
chitrangpatel: 👍

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -125,3 +136,60 @@ func LookForResults(w io.Writer, runDir string, resultsDir string, resultNames [
...
+	scanner.Buffer(buf, maxResultLimit)
+	for scanner.Scan() {
+		result, err := parseResults(scanner.Bytes(), maxResultLimit)
+		if err != nil {
+			return nil, err
+		}
+		sidecarLogResults = append(sidecarLogResults, result)
+	}
+
+	if err := scanner.Err(); err != nil {
```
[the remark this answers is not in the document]
chitrangpatel: 👍

pkg/pod/status.go
chitrangpatel: For context, this was done because the CI tests run the same test environment for all example tests and e2e tests. As a result, certain alpha features were failing this test because sidecar-logs was activated. The solution for now is to disable `sidecar-logs` for all alpha e2e and example tests. Only trigger it for the specific e2e tests. We will figure out how to handle examples next.

pkg/pod/status.go
```diff
@@ -147,15 +152,26 @@ func MakeTaskRunStatus(logger *zap.SugaredLogger, tr v1beta1.TaskRun, pod *corev
...
 
-func setTaskRunStatusBasedOnStepStatus(logger *zap.SugaredLogger, stepStatuses []corev1.ContainerStatus, tr *v1beta1.TaskRun) *multierror.Error {
+func setTaskRunStatusBasedOnStepStatus(ctx context.Context, logger *zap.SugaredLogger, stepStatuses []corev1.ContainerStatus, tr *v1beta1.TaskRun, kubeclient kubernetes.Interface) *multierror.Error {
 	trs := &tr.Status
 	var merr *multierror.Error
 
+	sidecarLogResults := []v1beta1.PipelineResourceResult{}
+	sidecarLogsResultsEnabled := config.FromContextOrDefaults(ctx).FeatureFlags.ResultExtractionMethod == config.ResultExtractionMethodSidecarLogs
+	if sidecarLogsResultsEnabled && tr.Status.TaskSpec.Results != nil {
+		resultSidecarName := fmt.Sprintf("sidecar-%s", sidecarlogsvalidation.ReservedResultsSidecarName)
```
jerop, line 162: I find it confusing that there's a reserved sidecar name defined elsewhere, but then it's modified (prefix added) here: 

https://github.com/tektoncd/pipeline/blob/f9021b1385c0d5bfa4e61b9473597ca03d173cb6/internal/sidecarlogsvalidation/sidecarlogs.go#L20-L22

test/larger_results_sidecar_logs_test.go
```diff
@@ -0,0 +1,385 @@
...
+	knativetest "knative.dev/pkg/test"
+)
+
+var (
+	ignoreTaskRunStatusFields = cmpopts.IgnoreFields(v1beta1.TaskRunStatusFields{}, "Steps", "TaskRunResults")
+	ignoreSidecarState        = cmpopts.IgnoreFields(v1beta1.SidecarState{}, "ImageID")
+
+	requireSidecarLogResultsGate = map[string]string{
+		"embedded-status": "minimal",
+	}
```
jerop, line 44:
```suggestion
	requireSidecarLogResultsGate = map[string]string{
	        "results-from": "sidecar-logs",
	}
```

chitrangpatel: Thanks!

pkg/pod/status.go
chitrangpatel: I think when a container is created from the sidecar name, `sidecar-` prefix is added by default. So the name of the sidecar is `tekton-log-results` but the name of the container is `sidecar-tekton-log-results`.

chitrangpatel: I could also create another reserved const `ReservedResultsSidecarContainerName`.
How about that?

chitrangpatel: I could also name the variable better to avoid confusion.

jerop: Yes please, that would be clearer!

jerop: or rename the variable `resultSidecarName` to `resultSidecarContainerName`

chitrangpatel: I created a new const `ReservedResultsSidecarContainerName` and I use that to avoid confusion. Thanks for pointing this out!

pkg/pod/status.go
jerop: ah I see, it may make more sense for this line to go within the else block (after the error check), maybe above line 173? 🤔

chitrangpatel: Agreed. Discussed on GVC. Move line 173 to 188.

jerop (approved):

pkg/pod/entrypoint_test.go
```diff
@@ -651,11 +721,23 @@ func TestStopSidecars(t *testing.T) {
 		Name:  injectedSidecar.Name,
 		Image: nopImage,
 	}
+	// This is a container that is added by the controller for accessing sidecar logs.
+	// This should not be stopped as long as results-from is set to sidecar-logs.
```
dibyom, line 725: Can we inject the sidecar only if results-from is set to sidecar-logs?

dibyom (commented):
Mostly looks good to me - just a few small comments!

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -17,15 +17,26 @@ limitations under the License.
 package sidecarlogresults
```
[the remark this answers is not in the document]
chitrangpatel: I did have it in pkg/internal before but then I was unable to import it in cms/sidecarlogresults/main.go which is why I had to move it here. 

the reason they are two different packages is because in my final PR I will be doing webhook validation to ensure that a user does not create a sidecar with the same name as the one that is injected. I run into cyclic imports there where sidecarlogresults would import v1beta1 and vice versa (to get access to the reservedSidecarName.

internal/sidecarlogsvalidation/sidecarlogs.go
```diff
@@ -20,4 +20,8 @@ const (
 	// ReservedResultsSidecarName is the name of the results sidecar that outputs the results to stdout
 	// when the results-from feature-flag is set to "sidecar-logs".
 	ReservedResultsSidecarName = "tekton-log-results"
+
```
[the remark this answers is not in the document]
chitrangpatel: The goal is to also add validation of sidecar name logic here in the final PR. v1beta1 and v1 will call the validation functions in here. 

Like I mentioned above, cyclic imports were the main culprit. I did move it to pkg/api/pipeline before but this seemed unrelated to any API changes so we ended up creating a sidecarlogvalidation package.

pkg/pod/entrypoint.go
```diff
@@ -252,6 +254,12 @@ func StopSidecars(ctx context.Context, nopImage string, kubeclient kubernetes.In
 	updated := false
 	if newPod.Status.Phase == corev1.PodRunning {
 		for _, s := range newPod.Status.ContainerStatuses {
+			// If the results-from is set to sidecar logs,
+			// a sidecar container with name `sidecar-log-results` is injected by the reconiler.
+			// Do not kill this sidecar. Let it exit gracefully.
```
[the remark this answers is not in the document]
chitrangpatel: Yes, that's correct!

pkg/pod/entrypoint_test.go
chitrangpatel: Yes, this is already the case : https://github.com/tektoncd/pipeline/blob/963aa40829045a349c8f78d4fa3dd10ffca46282/pkg/pod/pod.go#L155

It will also ensure that the sidecar is only injected if the task is producing results. It will not blindly inject a sidecar for all taskruns.

internal/sidecarlogresults/sidecarlogresults_test.go
```diff
@@ -121,6 +129,184 @@ func TestLookForResults(t *testing.T) {
...
+	podLogs := ""
+	for _, r := range inputResults {
+		res, _ := json.Marshal(&r)
+		podLogs = fmt.Sprintf("%s%s\n", podLogs, string(res))
+	}
+	logs := strings.NewReader(podLogs)
+
+	_, err := extractResultsFromLogs(logs, []v1beta1.PipelineResourceResult{}, 4096)
+	if err != ErrSizeExceeded {
+		t.Fatal(fmt.Sprintf("Expexted error %v but got %v", ErrSizeExceeded, err))
```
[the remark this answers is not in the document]
chitrangpatel: Thanks!

internal/sidecarlogresults/sidecarlogresults.go
```diff
@@ -125,3 +136,63 @@ func LookForResults(w io.Writer, runDir string, resultsDir string, resultNames [
 	}
 	return nil
 }
+
+// GetResultsFromSidecarLogs extracts results from the logs of the results sidecar
+func GetResultsFromSidecarLogs(ctx context.Context, clientset kubernetes.Interface, namespace string, name string, container string) ([]v1beta1.PipelineResourceResult, error) {
+	sidecarLogResults := []v1beta1.PipelineResourceResult{}
+	p, err := clientset.CoreV1().Pods(namespace).Get(ctx, name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
chitrangpatel: Perfect!

abayer:
/approve

Looks good to me!

internal/sidecarlogsvalidation/sidecarlogs.go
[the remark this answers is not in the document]
dibyom: To me, adding to `pkg/apis/pipeline` makes sense. Similar to https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/paths.go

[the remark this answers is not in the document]
chitrangpatel: Ok. I will move it back unless @jerop has any concerns.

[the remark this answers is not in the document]
jerop: no concerns from me

internal/sidecarlogresults/sidecarlogresults.go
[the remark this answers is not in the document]
chitrangpatel: moving it to pkg/apis/pipeline again :)

internal/sidecarlogsvalidation/sidecarlogs.go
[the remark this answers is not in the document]
chitrangpatel: done.

internal/sidecarlogresults/sidecarlogresults.go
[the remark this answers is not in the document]
dibyom: > the reason they are two different packages is because in my final PR I will be doing webhook validation to ensure that a user does not create a sidecar with the same name as the one that is injected.

I think putting reserved names (such as the sidecar name) in pkg/apis/pipeline makes sense. Does that result in import cycles as well?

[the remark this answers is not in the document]
chitrangpatel: No, Moving it to pkg/apis/pipeline does not result in any import cycle issue. Moved it there so I think we're good here.

dibyom (commented):
/lgtm

pkg/apis/pipeline/v1beta1/taskrun_types.go
```diff
@@ -186,6 +186,8 @@ const (
 	TaskRunReasonsResultsVerificationFailed TaskRunReason = "TaskRunResultsVerificationFailed"
 	// AwaitingTaskRunResults is the reason set when waiting upon `TaskRun` results and signatures to verify
 	AwaitingTaskRunResults TaskRunReason = "AwaitingTaskRunResults"
+	// TaskRunReasonResultLargerThanAllowedLimit is the reason set when one of the results exceeds its maximum allowed limit of 1 KB
+	TaskRunReasonResultLargerThanAllowedLimit TaskRunReason = "TaskRunResultLargerThanAllowedLimit"
```
JeromeJu, line 190: Thanks for the PR @chitrangpatel !

I was wondering if there is any reason this is kept in v1beta1 only? If not I can open up a PR to keep v1 and v1beta1 synced for the reconciler changes.

chitrangpatel: I think thats because the reconciler currently uses v1beta1. This reason is only called there. Its ok to keep them both I sync I suppose. That way, when we update the reconciler to point to v1, it works.

PR #5844, opened 2022-12-07
TEP-0121: Retries - Reconciler Implementation

XinruZhang:
# Changes
This is the second peice of TEP-0121 implementation, targeting at merging it as part in release `v0.43.0`.

Prior to this PR, `retries` logic for TaskRun is handled by Tekton `PipelineRun` reconciler. This PR delegates the retries implementation for `TaskRun` to the `TaskRun` reconciler.

The major change is we stopped relying on `len(retriesStatus)` to decide if a target `TaskRun` failed or not. Instead, we use status `ConditionSuceeded` to gate the completion of a `TaskRun`. Even a `TaskRun` failed on one execution, as long as it has remaining retries, the TaskRun won't be stored in etcd with its status set as `Failed`. Instead, the status will be:

```yaml
Type: Succeeded
Status: Unknown
Reason: ToBeRetried
```

It is worth noticing that, we still need to use `len(retriesStatus)` for Custom Tasks because currently we still support the alpha version, once we fully migrate custom task to the beta, we can safely remove that reliance.

Note that the integration test [retry_test.go](https://github.com/tektoncd/pipeline/blob/main/test/retry_test.go) still works.

This is [a demo video in an API WG meeting](https://drive.google.com/corp/drive/u/0/folders/1HtbupUIIeTOi77Exv-tFUYNrHjUlf19L) [18:05-21:00]

/close #5756 

/kind feature

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

PR #5853, opened 2022-12-08
TEP-0121: Modify metrics recorder for TaskRun Retries

XinruZhang:
# Changes

Openning this PR originally because a `DATA RACE` (Dirty Read) error 
occurs when I was implementing https://github.com/tektoncd/pipeline/pull/5844, see [error msg](
https://prow.tekton.dev/view/gs/tekton-prow/pr-logs/pull/tektoncd_pipeline/5844/pull-tekton-pipeline-unit-tests/1600899786926460928).

Basically, the reason is that [durationAndCountMetrics()](https://github.com/tektoncd/pipeline/blob/3279a4310399655b78df0bfeb805a931b0e84481/pkg/reconciler/taskrun/taskrun.go#L448) is reading 
the `taskRun` object in one goroutine, while retryTaskRun() (which is
called in [finishReconcileUpdateEmitEvents](https://github.com/tektoncd/pipeline/blob/3279a4310399655b78df0bfeb805a931b0e84481/pkg/reconciler/taskrun/taskrun.go#L292)) is writing the `taskRun` object.

After digging into it, there's one more concern which looks like a bug:
[durationAndCountMetrics()](https://github.com/tektoncd/pipeline/blob/3279a4310399655b78df0bfeb805a931b0e84481/pkg/reconciler/taskrun/taskrun.go#L448) in [reconcile()](https://github.com/tektoncd/pipeline/blob/3279a4310399655b78df0bfeb805a931b0e84481/pkg/reconciler/taskrun/taskrun.go#L447) only counts the taskrun numbers
where the taskruns that have no preparation error / not cancelled right after
scheduled / not timed out before calling `reconcile()` (for example,
be pending for a long time.).

To address the problem mentioned above, this PR moves 
`durationAndCountMetrics()` outside of [reconcile()](https://github.com/tektoncd/pipeline/blob/3279a4310399655b78df0bfeb805a931b0e84481/pkg/reconciler/taskrun/taskrun.go#L447) to [ReconcileKind()](https://github.com/tektoncd/pipeline/blob/3279a4310399655b78df0bfeb805a931b0e84481/pkg/reconciler/taskrun/taskrun.go#L98)
to make sure it is reading the taskRun object that has no further update.

---
**Related Background Info**

The recording metrics logic was originally in the ReconcileKind() function,
but the commit https://github.com/tektoncd/pipeline/commit/137521906ce3e1df9440a3248f932ddb5a653e35 addressing the recount issue **moves that logic
to `reconcile()`** [1] (probably for better code style? 🤔). 

This is a blocker for https://github.com/tektoncd/pipeline/pull/5844.

[1] https://github.com/tektoncd/pipeline/commit/137521906ce3e1df9440a3248f932ddb5a653e35#diff-6e67e9c647bbe2a08807ff5ccbdd7dc9036df373e56b9774d3996f92ab7ceabaL138-L145

# Submitter Checklist

- [x] ~Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing~
  functionality, content, code)
- [x] ~Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)~
- [x] ~Release notes contains the string "action required" if the change requires additional action from users switching to the new release~

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/taskrun/taskrun.go
+++ b/pkg/reconciler/taskrun/taskrun.go
@@ -96,6 +96,7 @@ var (
 // converge the two. It then updates the Status block of the Task Run
 // resource with the current status of the resource.
 func (c *Reconciler) ReconcileKind(ctx context.Context, tr *v1beta1.TaskRun) pkgreconciler.Event {
+	defer c.durationAndCountMetrics(ctx, tr)
 	logger := logging.FromContext(ctx)
 	ctx = cloudevent.ToContext(ctx, c.cloudEventClient)
 	// By this time, params and workspaces should not be propagated for embedded tasks so we cannot
@@ -445,7 +446,6 @@ func (c *Reconciler) prepare(ctx context.Context, tr *v1beta1.TaskRun) (*v1beta1
 // error but it does not sync updates back to etcd. It does not emit events.
 // `reconcile` consumes spec and resources returned by `prepare`
 func (c *Reconciler) reconcile(ctx context.Context, tr *v1beta1.TaskRun, rtr *resources.ResolvedTaskResources) error {
-	defer c.durationAndCountMetrics(ctx, tr)
 	logger := logging.FromContext(ctx)
 	recorder := controller.GetEventRecorder(ctx)
 	var err error


XinruZhang:
/assign @jerop 

cc @khrm 
cc @tektoncd/core-collaborators

JeromeJu:
>After digging into it, there's one more concern:
[durationAndCountMetrics()](https://github.com/tektoncd/pipeline/blob/3279a4310399655b78df0bfeb805a931b0e84481/pkg/reconciler/taskrun/taskrun.go#L448) in [reconcile()](https://github.com/tektoncd/pipeline/blob/3279a4310399655b78df0bfeb805a931b0e84481/pkg/reconciler/taskrun/taskrun.go#L447) only counts the taskrun numbers
where the taskrun has no preparation error / not cancelled right after
scheduled / not timed out before calling reconcile() (for example,
be pending for a long time.).

Wondering if adding the `status` check in [DurationAndCount](https://github.com/tektoncd/pipeline/blob/main/pkg/taskrunmetrics/metrics.go#L307) will help only counts the taskRuns that are not failed from https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/taskrun/taskrun.go#L110-L180?

XinruZhang:
Thanks @JeromeJu  for taking time to looking into the issue and commenting on the issue :)

If I'm not mistaken, are you suggesting that we should not count those failed taskruns for metrics? If so, why would we want to do that? 

> If this route is valid, shall we also change the similar logics for reconciler/pipelinerun.go

That's true, I do think we also need to change the logic for reconciler/pipelinerun.go

JeromeJu:
> If I'm not mistaken, are you suggesting that we should not count those failed taskruns for metrics? If so, why would we want to do that?

That was referring to the previous logics of putting [durationAndCountMetrics()](https://github.com/tektoncd/pipeline/blob/3279a4310399655b78df0bfeb805a931b0e84481/pkg/reconciler/taskrun/taskrun.go#L448) in reconcile function. But I totally agree that we should have it at `ReconcileKind` level which we can keep count of failed taskRuns(taking the current route).

khrm:
We need to check whether count in this logical route is correct or not. I will check why this is so.

XinruZhang:
@khrm  I believe this commit https://github.com/tektoncd/pipeline/commit/137521906ce3e1df9440a3248f932ddb5a653e35  is addressing the recount issue by comparing `beforeCondition` and `afterCondition`, only record when the two conditions are different). 

The recording metrics logic was originally in the ReconcileKind() function, but the commit https://github.com/tektoncd/pipeline/commit/137521906ce3e1df9440a3248f932ddb5a653e35 addressing the recount issue **moves that logic to `reconcile()`** [1] (probably for better code style? 🤔). 


cc @afrittoli @jerop @dibyom 

[1] https://github.com/tektoncd/pipeline/commit/137521906ce3e1df9440a3248f932ddb5a653e35#diff-6e67e9c647bbe2a08807ff5ccbdd7dc9036df373e56b9774d3996f92ab7ceabaL138-L145

chitrangpatel (approved):

dibyom:
Makes sense, thanks for digging in @XinruZhang 

/approve


> We need to check whether the count in this logical route is correct or not. I will check why this is so. We want to avoid recount issues.

@khrm - Agreed as well. Will the current set of tests count this?

jerop (approved):

afrittoli:
NIT Submitter checklist is not filled in

afrittoli (commented):
Thank you @XinruZhang for addressing this!

It looks like we don't have any reconcile unit test that checks metrics, which is problematic because we must rely on our understanding of the code alone to know whether the counting of `TaskRuns` is ok or duplicates will exist or cases will be missed.

The current logic for counting metrics relies on the assumption that we shall only count a `TaskRun` when `isDone()` is `true` at the end of the reconciliation cycle and some change happened in the `ConditionSucceeded` condition. These two conditions identify a single reconciliation cycle as long as no change is done to the `ConditionSucceeded` after a `TaskRun` is marked as done.

These two conditions do not change when moving the logic from `reconcile` to `ReconcileKind`, and in fact, the new location of the code captures some cases that were missing before, as @XinruZhang noted already.

There are a couple of improvements that could be made, but we could do them as a follow-up:
- the [code under the `isDone` check](https://github.com/tektoncd/pipeline/blob/e55c0dad5cdd23f5a356d0c8ce39f035618336a9/pkg/reconciler/taskrun/taskrun.go#L128) should never touch the conditions of the`TaskRun`. We invoke `setDefaults` and later `updateLabelsAndAnnotations` neither of which will touch the condition, so it's ok now, but it would be nice to enforce that somehow to avoid inadvertently breaking this logic here in the future. We could start adding a comment to it. Reconcile  metrics tests would ensure that this is not broken in future
- The code in the [`durantionAndCountMetrics`](https://github.com/tektoncd/pipeline/blob/e55c0dad5cdd23f5a356d0c8ce39f035618336a9/pkg/reconciler/taskrun/taskrun.go#L233) uses a lister to get the before condition. We should not do that, listers are expensive for the API server, besides [we already store the `before` condition](https://github.com/tektoncd/pipeline/blob/e55c0dad5cdd23f5a356d0c8ce39f035618336a9/pkg/reconciler/taskrun/taskrun.go#L107), so we should pass that to `durantionAndCountMetrics`, similar to what we do for events.

/approve

afrittoli (commented):
/lgtm

PR #5864, opened 2022-12-12
[cmd/entrypoint] remove ioutil for new go version

my-git9:
The `io/ioutil` has been deprecated from go 1.16 version, we should remove it.
https://go.dev/doc/go1.16#ioutil

ref: https://github.com/tektoncd/pipeline/issues/5874

### DIFF
--- a/cmd/entrypoint/subcommands/decode_script_test.go
+++ b/cmd/entrypoint/subcommands/decode_script_test.go
@@ -19,7 +19,7 @@ package subcommands
 import (
 	"encoding/base64"
 	"errors"
-	"io/ioutil"
+	"io"
 	"os"
 	"path/filepath"
 	"testing"
@@ -40,7 +40,7 @@ echo "Hello World!"
 			t.Errorf("temporary script file %q was not cleaned up: %v", src, err)
 		}
 	}()
-	if err := ioutil.WriteFile(src, []byte(encoded), mode); err != nil {
+	if err := os.WriteFile(src, []byte(encoded), mode); err != nil {
 		t.Fatalf("error writing encoded script: %v", err)
 	}
 
@@ -58,7 +58,7 @@ echo "Hello World!"
 		t.Fatalf("unexpected error statting decoded script: %v", err)
 	}
 	mod := info.Mode()
-	b, err := ioutil.ReadAll(file)
+	b, err := io.ReadAll(file)
 	if err != nil {
 		t.Fatalf("unexpected error reading content of decoded script: %v", err)
 	}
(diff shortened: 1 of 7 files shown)


JeromeJu:
Thanks @my-git9 
/kind cleanup
/ok-to-test

cmd/entrypoint/subcommands/decode_script_test.go
```diff
@@ -84,7 +84,7 @@ func TestDecodeScriptInvalidBase64(t *testing.T) {
 	invalidData := []byte("!")
 	expectedError := base64.CorruptInputError(0)
 
-	src, err := ioutil.TempFile("", "decode-script-test-*")
+	src, err := os.MkdirTemp("", "decode-script-test-*")
```
JeromeJu, line 87: I believe, also according to https://go.dev/doc/go1.16#ioutil:~:text=TempFile%20%3D%3E%20os.CreateTemp, this shall be CreateTemp. That's what's been failing the build test.

JeromeJu:
Also would you mind opening up an issue to track this as I suspect if we have any plan getting to 1.16 (or soon?).

cmd/entrypoint/subcommands/decode_script_test.go
my-git9: It was my oversight, thanks

my-git9:
> Thanks for the PR @my-git9 . Could you please help elaborate the reason why the `ioutil` in some specific tests are changed instead of all in the ./entrypoint folder?
> 
> /kind cleanup /ok-to-test

The ./entrypoint folder has been updated now

my-git9:
> Also wondering it would be helpful to open up an issue to track this as I suspect if we have any plan getting to 1.16 (or maybe in the near future if I didn't have the full picture?).

Sure, i will do it later

bendory (approved):
LGTM, thanks!

chitrangpatel (approved):
Thanks for doing this!

cmd/entrypoint/runner.go
```diff
@@ -8,7 +8,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
JeromeJu, line 11: NIT: I wonder if it would be better to restore this as in this PR this might a bit confusing upfront seeing that we're changing files that does not relate with `ioutil`.

JeromeJu:
Thanks @my-git9 for dividing up the migration in different directories, this is very helpful for keeping the PR small and easy to review 😃 . Accordingly, I think it might be helpful to update the PR title/ comment to indicate that the scope of the cleanup is within the `./cmd/entrypoint` directory?

I also updated in the issue https://github.com/tektoncd/pipeline/issues/5874#issuecomment-1353273568 that helps are favoured in other directories.

cmd/entrypoint/runner.go
my-git9: Updated, thanks

vdemeester (approved):

PR #5876, opened 2022-12-15
Stop Metric Recorder Listing TaskRuns When Records Data

XinruZhang:
# Changes

    Prior to this PR, we retrieve the `TaskRun` back from shared cache in
    `durationAndCountMetrics` in order to get the condition before
    this reconcile cycle, while actually we have already got the
    before condition in the main reconcile loop.
    
    This PR removes the lister logic and reuses the `before` condition
    from the `ReconcileKind()` loop.

/kind cleanup

Thanks @afrittoli  for catching this in https://github.com/tektoncd/pipeline/pull/5853#pullrequestreview-1219108535 ❤️

# Submitter Checklist

- [x] ~Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing~
- [x] ~Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed~
  functionality, content, code)
- [x] ~Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)~
- [x] ~Release notes contains the string "action required" if the change requires additional action from users switching to the new release~

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -230,7 +230,7 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 
 	// Reconcile this copy of the pipelinerun and then write back any status or label
 	// updates regardless of whether the reconciliation errored out.
-	if err = c.reconcile(ctx, pr, getPipelineFunc); err != nil {
+	if err = c.reconcile(ctx, pr, getPipelineFunc, before); err != nil {
 		logger.Errorf("Reconcile error: %v", err.Error())
 	}
 
@@ -256,26 +256,13 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 	return nil
 }
 
-func (c *Reconciler) durationAndCountMetrics(ctx context.Context, pr *v1beta1.PipelineRun) {
+func (c *Reconciler) durationAndCountMetrics(ctx context.Context, pr *v1beta1.PipelineRun, beforeCondition *apis.Condition) {
 	logger := logging.FromContext(ctx)
 	if pr.IsDone() {
-		// We get latest pipelinerun cr already to avoid recount
-		newPr, err := c.pipelineRunLister.PipelineRuns(pr.Namespace).Get(pr.Name)
-		if err != nil && !kerrors.IsNotFound(err) {
-			logger.Errorf("Error getting PipelineRun %s when updating metrics: %w", pr.Name, err)
-			return
-		} else if kerrors.IsNotFound(err) || newPr == nil {
-			logger.Debugf("Pipelinerun %s not found when updating metrics: %w", pr.Name, err)
-			return
+		err := c.metrics.DurationAndCount(pr, beforeCondition)
+		if err != nil {
+			logger.Warnf("Failed to log the metrics : %v", err)
 		}
-
-		before := newPr.Status.GetCondition(apis.ConditionSucceeded)
-		go func(metrics *pipelinerunmetrics.Recorder) {
-			err := metrics.DurationAndCount(pr, before)
-			if err != nil {
-				logger.Warnf("Failed to log the metrics : %v", err)
-			}
-		}(c.metrics)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/reconciler/taskrun/taskrun.go
```diff
@@ -228,27 +230,15 @@ func (c *Reconciler) checkPodFailed(tr *v1beta1.TaskRun) (bool, v1beta1.TaskRunR
...
-			if err := metrics.CloudEvents(ctx, tr); err != nil {
-				logger.Warnf("Failed to log the metrics : %v", err)
-			}
-		}(c.metrics)
+		if err := c.metrics.DurationAndCount(ctx, tr, before); err != nil {
+			logger.Warnf("Failed to log the metrics : %v", err)
+		}
+		if err := c.metrics.CloudEvents(ctx, tr); err != nil {
+			logger.Warnf("Failed to log the metrics : %v", err)
+		}
```
Yongxuanzhang, line 241: By reading this change, since we pass the `tr` to this function, and the `before` is got from the `tr`, why don't we just pass `tr` and inside the function get the `before` there?

Is it because the `before` may change later in `tr`? since `durationAndCountMetrics` is deferred. If so, in previous code we fetch the tr again from the lister, so it should be the latest `tr` not the previous `tr`. This seems to be a behaviour change?

XinruZhang: Thanks for the detailed review Yongxuan! Yes you are right, tr will be modified later on in this reconcile loop. It's worth noticing that though the value of this tr is changed, but it hasn't been saved into the etcd, therefore the condition retrieved from the lister is the same as the "before" condtion we get from tr initially. does this make sense to you?

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -256,26 +256,13 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 	return nil
 }
 
-func (c *Reconciler) durationAndCountMetrics(ctx context.Context, pr *v1beta1.PipelineRun) {
+func (c *Reconciler) durationAndCountMetrics(ctx context.Context, pr *v1beta1.PipelineRun, before *apis.Condition) {
```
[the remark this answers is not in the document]
XinruZhang: Yeah makes sense, updated

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -228,27 +230,15 @@ func (c *Reconciler) checkPodFailed(tr *v1beta1.TaskRun) (bool, v1beta1.TaskRunR
...
-			if err := metrics.CloudEvents(ctx, tr); err != nil {
-				logger.Warnf("Failed to log the metrics : %v", err)
-			}
-		}(c.metrics)
+		if err := c.metrics.DurationAndCount(ctx, tr, beforeCondition); err != nil {
+			logger.Warnf("Failed to log the metrics : %v", err)
+		}
+		if err := c.metrics.CloudEvents(ctx, tr); err != nil {
+			logger.Warnf("Failed to log the metrics : %v", err)
+		}
```
Yongxuanzhang, line 241: these 2 warn logs the same message (though err is different). This may be confusing for debugging. Could we change them into different msgs?

XinruZhang: makes sense, done!

pkg/reconciler/taskrun/taskrun.go
Yongxuanzhang: yes that makes sense! I think the PR's approach is good. 
Just one question, I read Andrea's comment and I don't think lister will talk to api server/etcd, it is a cache and shouldn't fetch data from etcd?

Yongxuanzhang: ref: https://aly.arriqaaq.com/kubernetes-informers/

XinruZhang: Yes yes, you are right, I believe so, it is fetching data from the shared informer instead of from the API server (which makes it less costy). Nevetheless, this is still an unnecessary api call, while every API call that eats network bandwith. Though it looks nother for one API call, but that's a big difference if we schedule taskruns at large scale :)

XinruZhang: Anyway I think we are aligned, thanks for clarifying the technical detail! Appreciate your expertise here!

Yongxuanzhang: It should be a local storage and we just fetch the data from cache, my understanding is that it doesn't send request to api server.  What was the api call before (sorry I may miss that)?  Did you mean 
```
c.taskRunLister.TaskRuns(tr.Namespace).Get(tr.Name)
```

XinruZhang: Oh true, sorry I was being curtly on "network bandwith". The point I wantted to make is this is an unnecessary "call" :P

Yongxuanzhang: sounds good! Shall we change the commit msg and pr description?

Yongxuanzhang (commented):
/lgtm
suggest update the commit message and PR description

pkg/reconciler/taskrun/taskrun.go
XinruZhang: SG! updated.

XinruZhang:
cc @tektoncd/core-collaborators PTAL

Issue #5527, opened 2022-09-20
TEP-0091: Trusted Resources Alpha


Yongxuanzhang:
This issue tracks the progress of Trusted Resources for Alpha.
TEP:  https://github.com/tektoncd/community/blob/main/teps/0091-trusted-resources.md

**Feature Version:**
`enable-api-fields: alpha`

Checklist:
Step 1: https://github.com/tektoncd/pipeline/pull/5552
- [x] Signing functions
- [x] Verification functions

Step 2: https://github.com/tektoncd/pipeline/pull/5581
- [x] Verification of resolved Task/Pipeline at reconciler 
- [x] Public Keys in Configmap
- [x] Examples and Documentation

Tekton/Cli: 
- [x] Sign and Verify at cmd

Step 3: https://github.com/tektoncd/pipeline/pull/5714
- [x] Public Keys in `VerificationPolicy` 

Step 4: KMS keys support
- [x] add KMS libraries: https://github.com/tektoncd/pipeline/pull/5890 
- [x] add KMS field: https://github.com/tektoncd/pipeline/pull/5891
- [x] use KMS (docs and mock tests): https://github.com/tektoncd/pipeline/pull/5965

Future work (will move to a new issue to track):
- [ ] Verification of API Task/Pipeline at webhook
- [ ] Secret lister: https://github.com/tektoncd/pipeline/issues/5884
- [ ] deprecate config map to store keys: https://github.com/tektoncd/pipeline/issues/5852
- [ ] test refactoring: https://github.com/tektoncd/pipeline/issues/5820
- [ ] v1 task and pipeline supports

Yongxuanzhang:
future work is moved to https://github.com/tektoncd/pipeline/issues/5980

PR #5890, opened 2022-12-19
[TEP-0091] add KMS libraries

Yongxuanzhang:
# Changes

This commit adds kms libraries into trusted resources verifier package, so that we could use kms for verification. Before this commit we only supports loading keys from files. 

This is similar to what we have done in [Chains](https://github.com/tektoncd/chains/blob/7897fa4a8be4d54c8aef625416fee4a106f07af2/pkg/chains/signing/kms/kms.go#L25-L28)

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


Yongxuanzhang:
tracking issue: https://github.com/tektoncd/pipeline/issues/5527

wlynch (commented):
/lgtm

chuangw6 (approved):
/lgtm

pkg/trustedresources/verifier/verifier.go
```diff
@@ -26,6 +26,10 @@ import (
 
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 	"github.com/sigstore/sigstore/pkg/signature"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/aws"        // pull in aws kms libraries
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/azure"      // pull in azure kms libraries
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/gcp"        // pull in gcp kms libraries
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/hashivault" // pull in hashivault kms libraries
```
jerop, line 32: are these blank identifiers here for now because the packages are not yet used? if so, why not import the packages in the PR where they are actually used?

vdemeester: @jerop they are imported here for the [`init`](https://github.com/sigstore/sigstore/blob/main/pkg/signature/kms/gcp/client.go#L44) of those packages to be executed, to "register" themselves on some other "package" that is/might be used. 
For example, this is the same technique used by the k8s credential package helper.

vdemeester: (I am usually not a fan of that approach, but oh well, sometimes there is no other way around 🙃 )

Yongxuanzhang: Thank you so much!! @vdemeester. Sorry I was ooo before and not able to answer this comment.

lbernick: The comments here aren't very informative, consider removing them or including a one-line comment that we need to execute these packages' init functions for xyz reasons.

lbernick (commented):
Note to reviewers: This new dependency is covered under the [flexibility](https://github.com/tektoncd/community/blob/main/teps/0091-trusted-resources.md#flexibility) section of TEP-0091.

Yongxuanzhang:
> Chains

Thanks! We need to have comment for each line, so I changed the comments as you suggested

Yongxuanzhang:
> Note to reviewers: This new dependency is covered under the [flexibility](https://github.com/tektoncd/community/blob/main/teps/0091-trusted-resources.md#flexibility) section of TEP-0091.

Thanks! We need to have comment for each line, so I changed the comments as you suggested

PR #5891, opened 2022-12-19
[TEP-0091] add KMS field

Yongxuanzhang:
# Changes

This commit add KMS into v1alpha1.VerificationPolicy and validation. KMS field is used to configure the KMS path and can be used to resolve keys stored in key management system by cloud providers. Before this commit we only support loading keys from file or raw data.

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Add KMS filed into v1alpha1.VerificationPolicy
```


pkg/apis/pipeline/v1alpha1/verificationpolicy_validation_test.go
```diff
@@ -117,7 +117,74 @@ func TestVerificationPolicy_Invalid(t *testing.T) {
...
+								Name:      "name",
+								Namespace: "namespace",
+							},
+							KMS: "kms://key/path",
+						},
+					},
+				},
+			},
+		},
+		want: apis.ErrMultipleOneOf("data", "kms", "secretref").ViaFieldIndex("key", 0),
```
wlynch, line 187: This file may not be the right place, but we should add a test that loads and uses a KMS provider. https://pkg.go.dev/github.com/sigstore/sigstore@v1.5.0/pkg/signature/kms/fake be likely useful here (or feel free to write your own).

wlynch (changes requested):

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation_test.go
Yongxuanzhang: Yes this PR only adds the filed in api and not really use it. 
Did you suggest add tests to mock kms when we use it?

Yongxuanzhang: I plan to divide this feature into 3 small PRs, 1 adds libraries, 2 adds field and 3 uses it .
Should I use the `fakekms` to mock in the 3rd pr?

wlynch:
Is there a tracking issue we can associate this with?

Yongxuanzhang:
> Is there a tracking issue we can associate this with?

Yes! I have this issue to track all the PRs and milestones
https://github.com/tektoncd/pipeline/issues/5527

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation.go
```diff
@@ -61,8 +61,16 @@ func (vs *VerificationPolicySpec) Validate(ctx context.Context) (errs *apis.Fiel
...
+
+	if key.Data != "" {
+		if key.KMS != "" || key.SecretRef != nil {
+			errs = errs.Also(apis.ErrMultipleOneOf("data", "kms", "secretref"))
+		}
+	} else if key.KMS != "" && key.SecretRef != nil {
+		errs = errs.Also(apis.ErrMultipleOneOf("data", "kms", "secretref"))
 	}
 
 	if key.Data != "" && key.SecretRef != nil {
```
jagathprakash, line 76: nit:
This entire piece can be recoded as 
num_key_types := 0
if key.Data != "" {
   num_key_types += 1
}
if key.SecretRef != nil {
   num_key_types +=1 
}
..
if num_key_types == 0 {
   errs = errs.Also(apis.ErrMissingOneOf("data", "secretref"))
}
if num_key_types > 1 {
   errs = errs.Also(apis.ErrMultipleOneOf("data", "kms", "secretref"))
}

Which would be more concise and clear.

Yongxuanzhang: thank! I agree this has more readability, I will change the code!

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation.go
```diff
@@ -61,8 +61,22 @@ func (vs *VerificationPolicySpec) Validate(ctx context.Context) (errs *apis.Fiel
...
+	}
+
+	if keyCount == 0 {
+		errs = errs.Also(apis.ErrMissingOneOf("data", "kms", "secretref"))
+	}
+	if keyCount > 1 {
+		errs = errs.Also(apis.ErrMultipleOneOf("data", "kms", "secretref"))
 	}
 
 	if key.Data != "" && key.SecretRef != nil {
```
chuangw6, line 82: Since multiple keys are checked via keyCount, `if key.Data != "" && key.SecretRef != nil {` seems not needed anymore.

pkg/apis/pipeline/v1alpha1/verificationpolicy_types.go
```diff
@@ -90,6 +90,10 @@ type KeyRef struct {
 	// Data contains the inline public key.
 	// +optional
 	Data string `json:"data,omitempty"`
+	// KMS contains the KMS url of the public key
+	// Supported formats differ based on the KMS system used.
+	// +optional
+	KMS string `json:"kms,omitempty"`
```
chuangw6, line 96: We should briefly comment what the format the kms url should follow, and also document in [trusted-resources.md](https://github.com/tektoncd/pipeline/blob/main/docs/trusted-resources.md). Are we planning to use sigstore package to do the signing? If so, https://docs.sigstore.dev/cosign/kms_support/ might be a good ref to link to.

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation.go
Yongxuanzhang: thanks I missed this one

pkg/apis/pipeline/v1alpha1/verificationpolicy_types.go
bendory: In addition to the format, I suggest providing a concrete example. "A picture is worth 100 words." :-)

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation.go
```diff
@@ -61,13 +61,24 @@ func (vs *VerificationPolicySpec) Validate(ctx context.Context) (errs *apis.Fiel
 // Validate KeyRef will check if one of KeyRef's Data or SecretRef exists, and the
 // Supported HashAlgorithm is in supportedSignatureAlgorithms.
 func (key *KeyRef) Validate(ctx context.Context) (errs *apis.FieldError) {
-	if key.Data == "" && key.SecretRef == nil {
-		errs = errs.Also(apis.ErrMissingOneOf("data", "secretref"))
+	keyCount := 0
```
bendory, line 64: Please add a comment explaining that you are validating that one and only one of Data, SecretRef, KMS is defined. (As the code exists today, I have to read and understand the code to figure that out myself -- please make my life easier!)

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation.go
```diff
@@ -61,13 +61,24 @@ func (vs *VerificationPolicySpec) Validate(ctx context.Context) (errs *apis.Fiel
...
+	if key.SecretRef != nil {
+		keyCount++
+	}
+	if key.KMS != "" {
+		keyCount++
 	}
 
-	if key.Data != "" && key.SecretRef != nil {
-		errs = errs.Also(apis.ErrMultipleOneOf("data", "secretref"))
+	if keyCount == 0 {
```
bendory, line 75:
```suggestion
switch keyCount {
    case 0:
		errs = errs.Also(apis.ErrMissingOneOf("data", "kms", "secretref"))

    case 1:
        // do nothing -- a single key definition is valid
     
     default:
		errs = errs.Also(apis.ErrMultipleOneOf("data", "kms", "secretref"))
} ```

File: pkg/apis/pipeline/v1alpha1/verificationpolicy_validation_test.go [hunk 7]
```diff
@@ -95,7 +95,7 @@ func TestVerificationPolicy_Invalid(t *testing.T) {
 				},
 			},
 		},
-		want: apis.ErrMissingOneOf("key[0].data", "key[0].secretref"),
+		want: apis.ErrMissingOneOf("data", "kms", "secretref").ViaFieldIndex("key", 0),
```
bendory, line 98: Can we make these test cases less fragile? If the given error changes in the implementation, this test will start failing. Maybe there's a way to untie them such that the test case is less fragile?
For example, what if someone has good reason to change the case from `kms` to `KMS` in the implementation -- as written, they now have to change the test code as well -- but this is not a functional change, it's just a change in the error message. Can we test for functional change instead of error message?

bendory: This comment also applies to the ordering of the error message in index `0`. Can we be order-independent such that the test remains valid even if the order of validation changes in the implementation?

Yongxuanzhang: This is how we return err for validation in current code base, take this for example: https://github.com/tektoncd/pipeline/blob/17b705c0efbc62741e41f4c99015db743552fb9b/pkg/apis/pipeline/v1beta1/pipelineref_validation.go#L37-L54
I think we should keep it consistent until we want to change them all

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation.go
Yongxuanzhang: sure!

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation.go
Yongxuanzhang: thanks! Great suggestion!

pkg/apis/pipeline/v1alpha1/verificationpolicy_types.go
Yongxuanzhang: > In addition to the format, I suggest providing a concrete example. "A picture is worth 100 words." :-)

a potential example makes sense! I also added a note as you suggested

pkg/apis/pipeline/v1alpha1/verificationpolicy_types.go
```diff
@@ -90,6 +90,13 @@ type KeyRef struct {
 	// Data contains the inline public key.
 	// +optional
 	Data string `json:"data,omitempty"`
+	// KMS contains the KMS url of the public key
+	// Supported formats differ based on the KMS system used.
+	// One example of a KMS url could be:
+	// gcpkms://projects/[PROJECT]/locations/[LOCATION]>/keyRings/[KEYRING]/cryptoKeys/[KEY]/cryptoKeyVersions/[KEY_VERSION]
+	// Note that the KMS is not supported yet.
+	// +optional
+	KMS string `json:"kms,omitempty"`
```
dibyom, line 99: Consider using a pointer for an optional type : https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

dibyom (approved):

pkg/apis/pipeline/v1alpha1/verificationpolicy_types.go
Yongxuanzhang: I cannot find a pointer string being used in our api, I think we should keep it consistent?

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation_test.go
lbernick: I almost have the opposite take from Bendory-- I often give the feedback he is giving, but here I think it actually makes sense to include the text of the error message rather than `apis.ErrMultipleOneOf`. The reason is that sometimes these helper functions can lead to confusing error message strings, and you don't really realize what output the user is actually seeing until it's reflected in a test. This does have the downside of brittle tests, though.

Re: consistency, I think the section you're referring to is in non-test code, while the comment David is making is a separate point about brittleness of tests.

pkg/apis/pipeline/v1alpha1/verificationpolicy_types.go
lbernick: The API conventions are a bit confusing here, it says "Using a pointer allows distinguishing unset from the zero value for that type" but also "we ask that pointers always be used with optional fields that do not have a built-in nil value" (assuming that "nil value" means "zero value" in this context).

I agree with Yongxuan here, I think having an empty string implies that the URL is unset, not that the URL is "", so there's not a strong need to distinguish.

pkg/apis/pipeline/v1alpha1/verificationpolicy_types.go
```diff
@@ -90,6 +90,13 @@ type KeyRef struct {
 	// Data contains the inline public key.
 	// +optional
 	Data string `json:"data,omitempty"`
+	// KMS contains the KMS url of the public key
+	// Supported formats differ based on the KMS system used.
+	// One example of a KMS url could be:
```
lbernick, line 95: Are there docs you could link to here instead?

Yongxuanzhang: There are docs in Chains, I remember David suggest adding an example here to show how does a sample kms url look like.

lbernick: Is sigstore the authority on what's a valid KMS URL? or do you have to look at gcp, aws, azure, etc? Can you say something like "see sigstore/your cloud provider KMS docs for URL specification" along with the example you have here?

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation_test.go
Yongxuanzhang: > I almost have the opposite take from Bendory-- I often give the feedback he is giving, but here I think it actually makes sense to include the text of the error message rather than `apis.ErrMultipleOneOf`. The reason is that sometimes these helper functions can lead to confusing error message strings, and you don't really realize what output the user is actually seeing until it's reflected in a test. This does have the downside of brittle tests, though.
> 
> Re: consistency, I think the section you're referring to is in non-test code, while the comment David is making is a separate point about brittleness of tests.

Are you suggesting we should avoid using `apis.ErrMissingOneOf`?. It seems convenient and also used a lot in other validation functions. https://github.com/tektoncd/pipeline/search?q=apis.ErrMissingOneOf

pkg/apis/pipeline/v1alpha1/verificationpolicy_types.go
Yongxuanzhang: Oh yes, I found one in sigstore docs https://docs.sigstore.dev/cosign/kms_support. Is this ok?

Yongxuanzhang: I added this link in the comment

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation_test.go
lbernick: Only in tests. For example, in https://github.com/tektoncd/pipeline/issues/5673 I noted a confusing error message "expected exactly one, got both: spec.taskRef.name, spec.taskRef.params, spec.taskRef.resolver", which comes from `ErrMultipleOneOf`. These helpers are super useful in validation code but in tests IMO it's useful to be able to see the string error message the user encounters so you can check that it's not confusing.

The downside as David pointed out is that if you change "kms" in the code to "KMS", your tests will fail, but not in a way that gives you useful signal about the code. (The comment about indexing here I'm not sure I agree with, since the "0" refers to the index of the problematic field in the test data rather than an artifact of validation ordering.)

I may be going more in depth than is necessary; feel free to use your best judgment here although I'd just note our tests aren't actually that consistent ([example](https://github.com/tektoncd/pipeline/blob/b27ec733a7debf40a31f9628bc749603963b8fbb/pkg/apis/pipeline/v1beta1/task_validation_test.go#L789-L792))

pkg/apis/pipeline/v1alpha1/verificationpolicy_types.go
lbernick: Great that's perfect

lbernick:
/lgtm

I wouldn't actually use a release note here because you're introducing a field that isn't usable yet, so I'm not sure there's actionable info for users.

pkg/apis/pipeline/v1alpha1/verificationpolicy_validation_test.go
Yongxuanzhang: Ohh I see your point. Thanks!

PR #5902, opened 2022-12-20
[TEP-0089] Add a config map to support SPIRE initialization.

jagathprakash:
This PR is one among the set of PRs done to implement TEP-0089.
This PR has been derived from the larger PRs  PR#5715 and PR#4759 by
@pxp928 and @lumjjb.

This PR is addressing the problem of non-falisfiable provenance discussed in [TEP-0089](https://github.com/tektoncd/community/blob/main/teps/0089-nonfalsifiable-provenance-support.md)
[SPIRE](https://github.com/spiffe/spire) is a tool which provides cryptographic identities to workloads in a cluster.
These identities are also associated with a key pair, which can be used to
sign TaskRun results to track if they have been modified by other workloads.

Using SPIRE is one of the ways to address non-falsifiable provenance.

This PR has the following changes
1. Modify the feature flag to control non-falsifiability from enable-spire to enforce-nonfalsifiability="spire".
   This is in accordance to the approved TEP.
2. Add a configmap config-spire to initialize SPIRE.

Signed-off-by: jagathprakash <31057312+jagathprakash@users.noreply.github.com>

# Changes

# Submitter Checklist

- [] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
  functionality, content, code)

# Release Notes

-  A new configMap spire-config has been introduced in this PR. This configMap has the following details
    - spire-trust-domain: specifies the SPIRE trust domain to use.
    - spire-socket-path: specifies the SPIRE agent socket for SPIFFE workload API.
    - spire-server-addr: specifies the SPIRE server address for workload/node registration.
    - spire-node-alias-prefix: specifies the SPIRE node alias prefix to use.
These are the details of an existing SPIRE deployment. This implies that SPIRE should be up and running if the SPIRE features has to be used. For now this PR is just adding a configMap and as such does nothing, but later the configMap will be used to initialize the SPIRE client.
- The feature flag gating this feature was previously called "enable-spire" and it was a boolean flag. This has now been changed to "enforce-nonfalsifiablity" which is a string and takes the strings "" (which is none) or "spire". This is in conformance to [TEP-0089](https://github.com/tektoncd/community/blob/main/teps/0089-nonfalsifiable-provenance-support.md).


pkg/apis/config/spire_config_test.go
```diff
@@ -0,0 +1,71 @@
...
+		verifyConfigFileWithExpectedSpireConfig(t, tc.fileName, tc.want)
+	}
+}
+
+func verifyConfigFileWithExpectedSpireConfig(t *testing.T, fileName string, want *sc.SpireConfig) {
+	t.Helper()
+	cm := test.ConfigMapFromTestFile(t, fileName)
+	if got, err := config.NewSpireConfigFromConfigMap(cm); err == nil {
+		if d := cmp.Diff(got, want); d != "" {
+			t.Errorf("Diff:\n%s", diff.PrintWantGot(d))
```
Yongxuanzhang, line 66: this should be `cmp.Diff(want, got)` since we use `PrintWantGot`?

Yongxuanzhang (commented):
just one minor comment

pkg/apis/config/spire_config_test.go
jagathprakash: Agree. Thanks for finding this.

pkg/apis/config/store_test.go
```diff
@@ -74,21 +78,23 @@ func TestStoreLoadWithContext_Empty(t *testing.T) {
...
+		SpireConfig:      spireConfig,
 	}
 
 	store := config.NewStore(logtesting.TestLogger(t))
 
-	cfg := config.FromContext(store.ToContext(context.Background()))
+	got := config.FromContext(store.ToContext(context.Background()))
 
-	if d := cmp.Diff(cfg, expected); d != "" {
+	if d := cmp.Diff(want, got); d != "" {
```
bobcatfish, line 97: nice fix!

pkg/apis/config/store.go
```diff
@@ -56,13 +58,16 @@ func FromContextOrDefaults(ctx context.Context) *Config {
 	artifactPVC, _ := NewArtifactPVCFromMap(map[string]string{})
 	metrics, _ := newMetricsFromMap(map[string]string{})
 	trustedresources, _ := NewTrustedResourcesConfigFromMap(map[string]string{})
+	spireconfig, _ := NewSpireConfigFromMap(map[string]string{})
```
bobcatfish, line 61: do you know why the returned errors are being ignored here from this function call and the ones above? it seems like we should probably handle them?

pkg/apis/config/spire_config_test.go
```diff
@@ -0,0 +1,71 @@
...
+			fileName: "config-spire-empty",
+		},
+	}
+
+	for _, tc := range testCases {
+		verifyConfigFileWithExpectedSpireConfig(t, tc.fileName, tc.want)
+	}
+}
+
+func verifyConfigFileWithExpectedSpireConfig(t *testing.T, fileName string, want *sc.SpireConfig) {
```
bobcatfish, line 61: nit: not sure it's helping much with readability to break this into a separate function - also i think that by calling [t.Helper](https://pkg.go.dev/testing#T.Helper), it makes it so that any errors are going to be reported at line 57 above where verifyConfigFileWithExpectedSpireConfig is called instead of at the specific line where they occur within the verifyConfigFileWithExpectedSpireConfig function, i think that might be making things less clear instead of more clear

pkg/apis/config/spire_config_test.go
```diff
@@ -0,0 +1,71 @@
...
+				TrustDomain:     "example.org",
+				SocketPath:      "unix:///spiffe-workload-api/spire-agent.sock",
+				ServerAddr:      "spire-server.spire.svc.cluster.local:8081",
+				NodeAliasPrefix: "/tekton-node/",
+			},
+			fileName: "config-spire-empty",
+		},
+	}
+
+	for _, tc := range testCases {
```
bobcatfish, line 56: you can combine the lines before this (declaring the testCase structure and the testCases list) and the for loop into one line which is a) a bit less verbose and b) a common approach in a lot of other tests in the repo

for example:

https://github.com/tektoncd/pipeline/blob/892198bb0ff5cad9e3260e417697bae90a3ba2a0/pkg/workspace/apply_test.go#L20-L43

bobcatfish (commented):
Thanks for this smaller PR, this is definitely easier to review!

Initial feedback based on the checklist in the PR description:
* Can there be some initial docs added to start to flesh this out?
* Can there be a release note explaining this as well?

Although the feature is incomplete, adding the config option that supports it is something that is user facing, so in order to make sure we have a complete release at any time we should include some initial docs - just explaining what the config option is for and that it doesn't yet do anything.

It would also be great if the commit message could be updated to be closer to [the commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits), specifically to explain some more of the context e.g. what is SPIRE, why is it being added (it reads a bit like a TODO list at the moment)

It also looks like this PR is changing the name of an existing feature flag if I'm understanding correctly so that is probably worth explaining in the release notes and commit message as well (and technically is a backward incompatible change but since I'm assuming that it didn't actually do anything (is that true?) it should be okay? (Flagging @tektoncd/core-maintainers fyi)

pkg/apis/config/spire_config_test.go
jagathprakash: Done.

pkg/apis/config/spire_config_test.go
jagathprakash: Done.

pkg/apis/config/store_test.go
jagathprakash: Thanks!

pkg/apis/config/store.go
jagathprakash: +1 to handling the errors.
This is more a design question. If there is an error in one of the configs then we could return nil from this function stating that the function failed. This might not work well as and error in config should not be reason for stalling the initialization.
These should not be propagated upwards and handling these could just be logging the error and continuing.

jagathprakash:
> Thanks for this smaller PR, this is definitely easier to review!
> 
> Initial feedback based on the checklist in the PR description:
> 
> * Can there be some initial docs added to start to flesh this out?
> * Can there be a release note explaining this as well?
> 
> Although the feature is incomplete, adding the config option that supports it is something that is user facing, so in order to make sure we have a complete release at any time we should include some initial docs - just explaining what the config option is for and that it doesn't yet do anything.
> 
> It would also be great if the commit message could be updated to be closer to [the commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits), specifically to explain some more of the context e.g. what is SPIRE, why is it being added (it reads a bit like a TODO list at the moment)
> 
> It also looks like this PR is changing the name of an existing feature flag if I'm understanding correctly so that is probably worth explaining in the release notes and commit message as well (and technically is a backward incompatible change but since I'm assuming that it didn't actually do anything (is that true?) it should be okay? (Flagging @tektoncd/core-maintainers fyi)

- Updated the commit message according to the Tekton guidelines.
- Updated the PR description on what is being done here and also added release notes.
- Added spire.md to the docs. Added a caveat at the top mentioning that this is a WIP.

docs/spire.md
```diff
@@ -0,0 +1,273 @@
+<!--
+---
+linkTitle: "TaskRun Result Attestation"
+weight: 1660
+---
+-->
+# TaskRun Result Attestations
```
bobcatfish, line 7: @jagathprakash I think you mentioned adding a caveat to this doc indicating it is a WIP, at first glance here I'm not seeing it - it might be more obvious if you remove more of this content initially? im not sure where to draw the line, i could see including some of the spire setup info and how to configure the config map that this PR is adding but i think including docs on how this would actually work when it doesn't yet work might be misleading (and would make more sense to review in the same PR that adds the functionality)

curious if @afrittoli has any thoughts on this too since we've discussed how to add docs incrementally a bit

pkg/apis/config/feature_flags.go
```diff
@@ -202,12 +227,12 @@ func NewFeatureFlagsFromMap(cfgMap map[string]string) (*FeatureFlags, error) {
 	// defeat the purpose of having a single shared gate for all alpha features.
 	if tc.EnableAPIFields == AlphaAPIFields {
 		tc.EnableTektonOCIBundles = true
-		tc.EnableSpire = true
+		tc.EnforceNonfalsifiability = EnforceNonfalsifiabilityWithSpire
```
bobcatfish, line 230: does this mean that if alpha fields are on, enforcing with spire will be turned on also? i think we'd want to configure spire enforcement on _only_ explicitly - i.e. its only configurable if alpha fields are turned on, but by default it's off

docs/spire.md
```diff
@@ -0,0 +1,273 @@
...
+    apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: config-spire
+      namespace: tekton-pipelines
+    labels:
+      app.kubernetes.io/instance: default
+      app.kubernetes.io/part-of: tekton-pipelines
+    data:
+      # spire-trust-domain specifies the SPIRE trust domain to use.
```
bobcatfish, line 160: is it possible to provide some more details around how to know what values to provide here? e.g. links to spire docs maybe? (i know i wouldn't know what value to put for 'trust domain' if i was trying to set this up 😅 )

pxp928: Yes that is a good suggestion. Here is a link to it: https://spiffe.io/docs/latest/deploying/spire_server/#server-configuration-file

pkg/apis/config/feature_flags.go
pxp928: +1 agree that Spire should be explicitly turned on as it has an external dependency on Spire running in the environment. The reason I did this originally was because of the following comment:
```
	// Note: the user cannot enable "alpha" while disabling bundles or custom tasks - that would
	// defeat the purpose of having a single shared gate for all alpha features.
```
"single shared gate for all alpha features" can this be removed?

jagathprakash: Though the comment "single shared gate for all alpha features" remains, the code is modified so that SPIRE is enabled only when alpha is enabled.
If SPIRE flag is set in the stable/beta modes, SPIRE will still not be enabled.

bobcatfish: To be clear setting the alpha flag should enable the ability to _use_ all alpha features but not necessarily configure them and turn them on. I think there are two things getting mixed together here:
1. Features have stability levels associated with them (alpha/beta/stable) and we are letting operators opt into making these features available via the EnableAPIFields flag
2. Some features are optional (regardless of their stability level), including using SPIRE

SPIRE support will initially be alpha, so the EnableAPIFields should be required to be configured at alpha in order for the SPIRE functionality to be available. However there should still be more configuration required (EnforceNonfalsifiability = SPIRE) in order to turn it on

> Changing the contract that setting alpha flag will enable all alpha features

Hopefully this shouldn't be a change in the contract and should match the description of the alpha feature flag in [TEP-0033](https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md) (or maybe TEP-0033 needs to be updated with more details about a case like this where a feature is alpha but also needs to be opted into? on the other hand it seems like the trend is toward per-feature flags in the future based on some recent discussions)

> Enable SPIRE only if alpha is set and EnforceNonfalsifiability is set to spire.

+1

> Even if the EnforceNonfalsifiability flag is set to spire, if alpha is not enabled then do not enable spire.

Yes this makes sense to me - this could even be treated as an error

docs/spire.md
bobcatfish: > But just adding a doc for a configmap seems hard. More so since it does not give any context on why we are adding the configMap.

While the implementation is ongoing, I think it would be reasonable to:
* Add documentation that links to the TEP in progress (while I don't think the final documentation should rely on the TEP, I think it's reasonable to do before the feature is complete enough to use)
* Make sure the config map itself is well documented

specifically looking at spire.md, I would suggest that this commit includes the "Architecture Overview" section and maybe "Enabling TaskRun result attestations" as well

pkg/apis/config/feature_flags.go
pxp928: Yes that makes sense.

jagathprakash: As mentioned in the previous comment, https://github.com/tektoncd/pipeline/pull/5902#discussion_r1072875138, this has been fixed as suggested, i.e. SPIRE can only be enabled if alpha is turned on and SPIRE will not be enabled even if it is set but alpha is not turned on.

docs/spire.md
jagathprakash: @bobcatfish I added the following to the documentation

-  Link to the TEP. This is added as part of the caveat that this document is a work in progress.
- More documentation on the conffig map. Specifically how the config map comes into play and the explanation of the fields in the config map.

Unchanged sections - "Architecture overview".
Retained sections - all, which includes "Architecture overview" and "Enabling TaskRun result attestations".

bobcatfish: hmm there's still a lot of documentation here that doesn't really have to do with the features being added in this PR - here's a gist with an example of what I had in mind: https://gist.github.com/bobcatfish/69a66c5296e3f012b7d271c0d03c3415

^^ trying to add documentation which:
* Clearly indicates the feature is not implemented
* Clearly points the reader to where the complete feature is described (would be great if that included a phased implementation approach so they could know how this will be incrementally merged but that's another story :D)
* Has documentation for what this PR includes: support for configuring Tekton Pipelines to connect to SPIRE

pkg/apis/config/feature_flags.go
```diff
@@ -138,6 +142,14 @@ func GetFeatureFlagsConfigName() string {
 	return "feature-flags"
 }
 
+func getEnforceNonfalsifiabilityValues() map[string]struct{} {
```
bobcatfish, line 145: this function seems to have added with this very specific signature in order to be able to be passed to setStringFeature which is a function declared inline within NewFeatureFlagsFromMap and only used with getEnforceNonfalsifiabilityValues - would it be clearer to instead combine setStringFeature and this function into one that is specific to the getEnforceNonfalsifiability feature?

it seems like setStringFeature is created with the  idea that it would be reused, however it's declared inline within the NewFeatureFlagsFromMap function so it doesn't have the benefit of being a separate function that could be in its own library with its own tests, and it's not actually used anywhere else, so it seems like it having the indirection of getEnforceNonfalsifiabilityValues + setStringFeature at the moment doesn't give us much gain and makes it a bit harder to read the code

pkg/apis/config/feature_flags.go
```diff
@@ -202,14 +227,16 @@ func NewFeatureFlagsFromMap(cfgMap map[string]string) (*FeatureFlags, error) {
 	// defeat the purpose of having a single shared gate for all alpha features.
 	if tc.EnableAPIFields == AlphaAPIFields {
 		tc.EnableTektonOCIBundles = true
-		tc.EnableSpire = true
-	} else {
-		if err := setFeature(enableTektonOCIBundles, DefaultEnableTektonOciBundles, &tc.EnableTektonOCIBundles); err != nil {
+		// Only consider SPIRE if alpha is on.
```
bobcatfish, line 230: should we consider it an error if someone tries to configure SPIRE when alpha is off?

pkg/apis/config/store.go
bobcatfish: would it make sense to log something in the case of an error at least? (or maybe that is already happening?)

pkg/apis/config/feature_flags.go
jagathprakash: I thought we could have a internal function setStringFeature, similar to setFeature above, but as pointed out this is currently being used only for setting the EnforceNonFalsifiablity flag. 
So as suggested, moved all the functionality into one function getEnforceNonfalsifiabilityFeature.

pkg/apis/config/feature_flags.go
jagathprakash: Maybe, my earlier thought was adding an error will just throw a message for an impossible flow.
But if we do not then user will not know why SPIRE was not enabled in this flow even if the flag was set.
So added an error here.

docs/spire.md
jagathprakash: Thanks for the example doc. I have updated spire.md as suggested.
Also added a list of phases for the implementation.

jagathprakash:
Addressed the comment https://github.com/tektoncd/pipeline/pull/5902#discussion_r1086051535

by logging the errors.

docs/spire.md
bobcatfish: that's great! thanks

bobcatfish: oh actually im surprised to still see so much content in the doc, in https://gist.github.com/bobcatfish/69a66c5296e3f012b7d271c0d03c3415 i was suggesting that you remove a lot of it. it seems like you feel strongly about including that here?

pkg/apis/config/feature_flags.go
```diff
@@ -138,6 +142,24 @@ func GetFeatureFlagsConfigName() string {
...
+	}
+	var value = DefaultEnforceNonfalsifiability
+	if cfg, ok := cfgMap[enforceNonfalsifiability]; ok {
+		value = strings.ToLower(cfg)
+	}
+	if _, ok := acceptedValues[value]; !ok {
+		*feature = DefaultEnforceNonfalsifiability
+		return fmt.Errorf("invalid value for feature flag %q: %q", enforceNonfalsifiability, value)
+	}
+	*feature = value
```
bobcatfish, line 159: the interface here might be a bit simpler if you returned the string value to use vs. changing the value of a pointer (i.e. more of a functional interface vs. a function with side effects)

i think this might also simplify the code below where you check if SPIRE is configured in non-alpha mode; all you need to do is assert that the string returned is 'none'/ no error

pkg/apis/config/store.go
```diff
@@ -50,19 +53,43 @@ func FromContextOrDefaults(ctx context.Context) *Config {
...
+		logging.FromContext(ctx).Info(errMetrics)
+	}
+	trustedresources, errTrustedresources := NewTrustedResourcesConfigFromMap(map[string]string{})
+	if errTrustedresources != nil {
+		logging.FromContext(ctx).Info(errTrustedresources)
+	}
+	spireconfig, errSpireconfig := NewSpireConfigFromMap(map[string]string{})
+	if errSpireconfig != nil {
+		logging.FromContext(ctx).Info(errSpireconfig)
+	}
```
bobcatfish, line 83: thanks for doing this! I apologize but I have a few more nits here:
* instead of renaming the err variable every time its clearer to just use `err` in all cases (usually we don't want to reuse variables and change their meaning so your instinct is right but imo error variables are an exception to this - similar to what you are doing with `ok` in NewSpireConfigFromMap)
* test coverage of this file went way down after this change b/c the error conditions are not being tested
* Can you add more context to the log message, e.g. something like "Spire configuration was invalid: %v"?

Testing log message handling is super ugly and I don't recommend having tests that assert on the log content

One way that you could address most of the above would be to collect a list of errors and log once at the end - we are already using `go-multierror` in other parts of the project, e.g. in https://github.com/tektoncd/pipeline/blob/2ee148d7af881588fc50846344ff5b4d63bbf07c/pkg/pod/status.go#L155 you can see that one error object `merr` is initialized and errors are appended as the function executes. If you collect errors as you go you could log once at the end. then to test, you can have a few cases where some/all/one value is invalid, and maybe don't assert that the log function was called, but assert that the object was constructed correctly in those cases

bobcatfish (commented):
I think we're almost there! Thanks for all the back and forth on this

pkg/apis/config/store.go
pxp928: I think this should be made into an issue and moved to a separate PR as this is out of scope for this PR.

docs/spire.md
pxp928: @jagathprakash you can move some of the documentation out from here and make it part of the other 2 remaining PRs. Let me know if you need help to where it needs to be split up.

pkg/apis/config/feature_flags.go
jagathprakash: +1,  changed to a functional interface.

pkg/apis/config/store.go
jagathprakash: +1 to @pxp928 to make this an issue.
2 reasons

1. This requires more thought. Do we need to handle the errors? The initialization is for default values so we will never get errors.
2. This is a larger fix which is unrelated to this PR.

I can add an issue to track this and other locations where errors are getting dropped.
@bobcatfish WDYT

docs/spire.md
jagathprakash: Thanks @bobcatfish @pxp928 for the suggestions.
I have reduced the documentation as @bobcatfish suggested. 

Please take a look now.

pkg/apis/config/store.go
bobcatfish: Sounds good to me! Looks like you already undid the update to change the error handling; my request would be that we either do nothing (which it looks like is what we're doing, and np there since it's definitely scope creep) or if we do make changes, we do so with test coverage (and addressing variable handling, etc., as discussed above)

np with leaving this out of scope and opening an issue - could be a good 'help wanted' 'good first issue' if all we want done is adding some logging

bobcatfish:
Thanks for all the back and forth!

/approve

bobcatfish:
Looks like there is a rebasing problem stopping this from merging @jagathprakash 

Also looks like there are a bunch of commits in this PR that should probably be squashed together before merging, plus a merge commit that we wouldn't want to include. Putting a hold on this just to make sure that gets resolved before merging, feel free to remove after the commits are fixed:

/hold

PR #5904, opened 2022-12-20
Enabled `gosimple` linter.

bendory:
# Changes

This linter detects use of over-complex Go and suggests idiomatic replacements.

There are no functional changes expected in this PR.

Context: #5899 

/kind cleanup

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [N/A] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [N/A] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [N/A] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [N/A] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

Issue #5899, opened 2022-12-20
Cleanup: uplevel our technical hygiene by enabling additional Golang linters


bendory:
### Feature request

`golangci-lint` which we have configured supports additional helpful linters that we are not using.
This FR suggests that we uplevel our technical hygiene by enabling more linters and cleaning up
the issues they identify.

Documentation: https://golangci-lint.run/usage/linters

bendory:
I would assign this issue to me, but I don't seem to have the authority to do so. 🙃

bendory:
#5871 
#5881 
#5889

vdemeester:
/assign @bendory 
(btw, if you comment `/assign`, it should automatically assign the issue to you 😉 )

vdemeester:
(ah lol, @jerop manually assigned you already 😂 )

bendory:
See also: https://github.com/tektoncd/community/pull/908

PR #5916, opened 2022-12-22
Cleanup unit tests to use idiomatic Go.

bendory:
# Changes

This PR is a follow-on from https://github.com/tektoncd/pipeline/pull/5904#issuecomment-1361398213 to make spire tests more readable, easier to understand, and conformant with [idiomatic Golang style](https://goo.gle/go-style).

There are no expected functional changes in this PR.

Changes:
- `spire_mock_test.go`: Renamed test functions from `TestSpireMock_*` --> `TestMock_*`
- `spire_test.go`:
  - Renamed test functions from `TestSpire_*` --> `Test*`
  - use concise helper function names -([style guide](https://google.github.io/styleguide/go/decisions#getters))
  - in-lined test data in place of `testSingleTaskRun()`, `testSinglePipelineResourceResults()` helpers
  - eliminated unnecessary helper functions (`genPr()` and `success := func()...`)
  - localized setup to just the tests that need it ([style guide](https://google.github.io/styleguide/go/best-practices#keep-setup-code-scoped-to-specific-tests))
  - use naming conventions `got`, `want` in tests

/kind cleanup

Ref: https://github.com/tektoncd/pipeline/issues/5899

# Submitter Checklist

- [N/A] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
  functionality, content, code)
- [N/A] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [N/A] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/spire/spire_test.go
+++ b/pkg/spire/spire_test.go
@@ -18,6 +18,7 @@ package spire
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"testing"
 	"time"
@@ -33,10 +34,10 @@ import (
 	"knative.dev/pkg/logging"
 )
 
+const trustDomain = "example.org"
+
 var (
-	trustDomain  = "example.org"
 	td           = spiffeid.RequireTrustDomainFromString(trustDomain)
-	fooID        = spiffeid.RequireFromPath(td, "/foo")
 	controllerID = spiffeid.RequireFromPath(td, "/controller")
 )
 
@@ -47,7 +48,7 @@ func init() {
 	backoff = 100 * time.Millisecond
 }
 
-func TestSpire_TaskRunSign(t *testing.T) {
+func TestTaskRunSign(t *testing.T) {
 	ctx, _ := ttesting.SetupDefaultContext(t)
 
 	ca := test.NewCA(t, td)
@@ -58,7 +59,7 @@ func TestSpire_TaskRunSign(t *testing.T) {
 
 	resp := &fakeworkloadapi.X509SVIDResponse{
 		Bundle: ca.X509Bundle(),
-		SVIDs:  makeX509SVIDs(ca, controllerID),
+		SVIDs:  x509svids(ca, controllerID),
 	}
 	wl.SetX509SVIDResponse(resp)
 
@@ -87,7 +88,7 @@ func TestSpire_TaskRunSign(t *testing.T) {
 	}
 }
 
-func TestSpire_CheckSpireVerifiedFlag(t *testing.T) {
+func TestCheckSpireVerifiedFlag(t *testing.T) {
 	ctx, _ := ttesting.SetupDefaultContext(t)
 
 	ca := test.NewCA(t, td)
@@ -98,7 +99,7 @@ func TestSpire_CheckSpireVerifiedFlag(t *testing.T) {
 
 	resp := &fakeworkloadapi.X509SVIDResponse{
 		Bundle: ca.X509Bundle(),
-		SVIDs:  makeX509SVIDs(ca, controllerID),
+		SVIDs:  x509svids(ca, controllerID),
 	}
 	wl.SetX509SVIDResponse(resp)
 
@@ -127,7 +128,7 @@ func TestSpire_CheckSpireVerifiedFlag(t *testing.T) {
 	}
 }
 
-func TestSpire_CheckHashSimilarities(t *testing.T) {
+func TestCheckHashSimilarities(t *testing.T) {
 	ctx, _ := ttesting.SetupDefaultContext(t)
 
 	ca := test.NewCA(t, td)
@@ -138,7 +139,7 @@ func TestSpire_CheckHashSimilarities(t *testing.T) {
 
 	resp := &fakeworkloadapi.X509SVIDResponse{
 		Bundle: ca.X509Bundle(),
-		SVIDs:  makeX509SVIDs(ca, controllerID),
+		SVIDs:  x509svids(ca, controllerID),
 	}
 	wl.SetX509SVIDResponse(resp)
 
@@ -181,7 +182,7 @@ func TestSpire_CheckHashSimilarities(t *testing.T) {
 }
 
 // Task run sign, modify signature/hash/svid/content and verify
-func TestSpire_CheckTamper(t *testing.T) {
+func TestCheckTamper(t *testing.T) {
 	ctx, _ := ttesting.SetupDefaultContext(t)
 
 	ca := test.NewCA(t, td)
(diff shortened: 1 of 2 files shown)


bendory:
ping @ijschwabacher

lbernick (approved):

pkg/spire/spire_test.go
```diff
@@ -455,34 +455,27 @@ func TestSpire_TaskRunResultsSignTamper(t *testing.T) {
...
+		Key:          "foo",
+		Value:        "foo-value",
+		ResourceName: "source-image",
+		ResultType:   v1beta1.TaskRunResultType,
+	}, {
+		Key:          "bar",
+		Value:        "bar-value",
+		ResourceName: "source-image2",
+		ResultType:   v1beta1.TaskRunResultType,
+	}}
```
ijschwabacher, line 468: I notice that the preexisting tests consistently generate new copies of test data for each test. Is that out of concern that the test might accidentally modify its inputs? If there's a real concern that the code *under* test might do so, I think it would be better to get a chance of exposing it by making the tests share test data, but even better would be detecting the changes on purpose.

In any case, I think it's important to be consistent here. If you're going to change this, you'll also want to change `testTaskRuns()` and `testPipelineResourceResults()`.

pkg/spire/spire_test.go
```diff
@@ -625,79 +618,47 @@ func TestSpire_TaskRunResultsSignTamper(t *testing.T) {
...
-				sigResults, err := ec.Sign(ctx, results)
-				if err != nil {
-					return false
-				}
+				var err error
+				var sigResults []v1beta1.PipelineResourceResult
+				if sigResults, err = ec.Sign(ctx, results); err == nil {
+					results = append(results, sigResults...)
+					if tt.tamperFn != nil {
+						results = tt.tamperFn(results)
```
ijschwabacher, line 641: ...in fact it looks like this test *does* end up modifying `results` in place, because many of the `tamperFn`s do! This will only translate into a modification of `pr` if `sigResults` is empty, because a slice literal's capacity is guaranteed to equal its length, and this might not be compatible with `err == nil`, but at that point we're relying on the correctness of the code *under* test to show the correctness of the test *itself*, which is a problem. (Alternatively, one could argue that the length of this chain of reasoning is already a problem.)

pkg/spire/spire_test.go
```diff
@@ -625,79 +618,47 @@ func TestSpire_TaskRunResultsSignTamper(t *testing.T) {
...
+					SVIDs:  x509svids(ca, spiffeid.RequireFromPath(td, taskrunPath(tr))),
+				})
 
-				sigResults, err := ec.Sign(ctx, results)
-				if err != nil {
-					return false
-				}
+				var err error
+				var sigResults []v1beta1.PipelineResourceResult
+				if sigResults, err = ec.Sign(ctx, results); err == nil {
```
ijschwabacher, line 638: I don't think needing to split the `wantErr` check is a strong enough reason to disregard https://google.github.io/styleguide/go/decisions#indent-error-flow.

pkg/spire/spire_test.go
```diff
@@ -356,86 +368,74 @@ func TestSpire_TaskRunResultsSign(t *testing.T) {
...
+	cfg := &config.SpireConfig{
+		SocketPath:  wl.Addr(),
+		TrustDomain: trustDomain,
+	}
+
+	ec := NewEntrypointerAPIClient(cfg)
+	cc := GetControllerAPIClient(ctx)
+	cc.SetConfig(*cfg)
+	defer cc.Close()
+	defer ec.Close()
```
ijschwabacher, line 413: *extremely* optional: Almost every test in this file has this same setup. Is it worth factoring this out into a test helper?

pkg/spire/spire_test.go
```diff
@@ -336,7 +337,18 @@ func TestSpire_CheckTamper(t *testing.T) {
...
 }
 
-func TestSpire_TaskRunResultsSign(t *testing.T) {
+func TestNoSVID(t *testing.T) {
+	ctx := context.Background()
+	cfg := &config.SpireConfig{SocketPath: "tcp://127.0.0.1:12345"} // bogus SocketPath
+	ec := NewEntrypointerAPIClient(cfg)
+	defer ec.Close()
+
+	if _, err := ec.Sign(ctx, nil); !errors.Is(err, context.DeadlineExceeded) {
```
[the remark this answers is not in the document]
bendory: The implementation as written has its own embedded timeout, and IDK the implications of changing that implementation to move the timeout up to the caller. This was a pre-existing test that I have rewritten into idiomatic Go using `errors.Is` rather than generically checking that an error was returned (which is what the old test did).

I agree that this implementation could be better, but that's not my current focus. 🙈

pkg/spire/spire_test.go
```diff
@@ -356,86 +368,74 @@ func TestSpire_TaskRunResultsSign(t *testing.T) {
...
+	wl.SetX509SVIDResponse(&fakeworkloadapi.X509SVIDResponse{
+		Bundle: ca.X509Bundle(),
+		SVIDs:  x509svids(ca, spiffeid.RequireFromPath(td, "/bogus")),
+	})
+
+	for _, tr := range trList {
+		for _, results := range rsrcResults {
+			if sigResults, err := ec.Sign(ctx, results); err != nil {
+				t.Errorf("ec.Sign error: %v", err)
+			} else {
```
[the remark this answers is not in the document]
bendory: Unwinding this tangle into idiomatic Go is definitely an iterative process. Thanks for this one.

pkg/spire/spire_test.go
bendory: Well doh! This is obvious now; fixed.

pkg/spire/spire_test.go
```diff
@@ -356,86 +368,74 @@ func TestSpire_TaskRunResultsSign(t *testing.T) {
...
+	for _, tr := range trList {
+		for _, results := range rsrcResults {
+			wl.SetX509SVIDResponse(&fakeworkloadapi.X509SVIDResponse{
+				Bundle: ca.X509Bundle(),
+				SVIDs:  x509svids(ca, spiffeid.RequireFromPath(td, taskrunPath(tr))),
+			})
+
+			if sigResults, err := ec.Sign(ctx, results); err != nil {
+				t.Errorf("sign error: %v", err)
+			} else {
```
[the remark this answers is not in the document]
bendory: ditto `fixed`

pkg/spire/spire_test.go
bendory: this is resolved by properly in-lining the definition of the `results` local var

pkg/spire/spire_test.go
bendory: Much cleaner this way!

pkg/spire/spire_test.go
bendory: not today 😸
/meow

PR #5943, opened 2023-01-04
[WIP] Swap TaskRun to storage version

JeromeJu:
# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


pkg/taskrunmetrics/metrics.go
```diff
@@ -489,9 +491,14 @@ func IsPartOfPipeline(tr *v1beta1.TaskRun) (bool, string, string) {
 	return false, "", ""
 }
 
-func sentCloudEvents(tr *v1beta1.TaskRun) int64 {
+func sentCloudEvents(tr *v1.TaskRun) int64 {
 	var sent int64
-	for _, event := range tr.Status.CloudEvents {
+	var v1beta1TaskRun v1beta1.TaskRun
```
JeromeJu, line 496: I've tried the following two ways: (taking the taskrun.status.cloudEvents as an example)
- make DeserializeCloudEvents public
- convert the v1.TaskRun object to v1beta1 at https://github.com/tektoncd/pipeline/blob/main/pkg/taskrunmetrics/metrics.go#L492

Personally I felt that converting v1 objects to v1beta1 and reuse the logics at L492 makes more sense, can I ask if that looks like the right way that I picked?

JeromeJu: cc @lbernick can I ask what do you think about this?

pkg/reconciler/taskrun/resources/apply.go
```diff
@@ -130,36 +130,36 @@ func paramsFromTaskRun(ctx context.Context, tr *v1beta1.TaskRun) (map[string]str
...
 	}
 
 	// We always add replacements for 'path'
 	if spec.Resources != nil && spec.Resources.Inputs != nil {
 		for _, r := range spec.Resources.Inputs {
-			replacements[fmt.Sprintf("resources.inputs.%s.path", r.Name)] = v1beta1.InputResourcePath(r.ResourceDeclaration)
-			// FIXME(vdemeester) Remove that with deprecating v1beta1
-			replacements[fmt.Sprintf("inputs.resources.%s.path", r.Name)] = v1beta1.InputResourcePath(r.ResourceDeclaration)
+			replacements[fmt.Sprintf("resources.inputs.%s.path", r.Name)] = v1.InputResourcePath(r.ResourceDeclaration)
+			// FIXME(vdemeester) Remove that with deprecating v1
```
lbernick, line 147: this comment should be left intact. We shouldn't have `InputResourcePath` in v1. Instead, pipelineresources here would need to be deserialized from v1 annotations.

pkg/taskrunmetrics/metrics.go
lbernick: I think for handling PipelineResources in general you'll need a function to determine if a v1 CRD has PipelineResources serialized during conversion from v1beta1. However, I'd prefer to avoid exporting these as public functions in the apis/pipeline/v1 package since we're going to remove them in 9 months, and it would be better to avoid the backwards incompatible change to the go libraries if we can. I think it would be fine to export PipelineResource related stuff in the v1beta1 package though.

I wouldn't export DeserializeCloudEvents; instead, I'd write a public helper function in v1beta1 that takes a v1 CRD and returns the PipelineResources (or cloudevents, etc), if any. Between that and doing conversion everywhere you need to, I don't have a strong preference; just pick whichever way you feel is more readable.

lbernick (commented):
Any reason to convert both Task and TaskRun in the same PR? Just curious

pkg/reconciler/taskrun/resources/apply.go
JeromeJu: Thanks for the comments Lee. 

> Any reason to convert both Task and TaskRun in the same PR? Just curious
This PR is made only to raise the questions on the conversion question. Will be putting up a PR on v1-release branch later. Also I would avoid opening branches redundantly in the future. Thanks for the pointers!

pkg/taskrunmetrics/metrics.go
JeromeJu: Much appreciated for the pointers! And to confirm and summarize, I would:

- [ ] Not export any conversion/versioning related functions related with deprecated fields BECAUSE they would be deprecated in 9 months
- [ ] Exporting helpers from v1beta1 and converting everywhere are both options.

lbernick: Basically yes. We shouldn't have anything in pkg/apis/version with reference to deprecated fields, since that package is more generically for converting between versions. All I'm saying is not to add anything to the v1 package that references fields from v1beta1. After 9 months, it's OK to delete anything in the v1beta1 package, but we can't do the same for the v1 package.

PR #5950, opened 2023-01-04
refactor trusted resources testing functions

### LINKED ISSUE #5820: Refactor trusted resources test code
Test helper functions in https://github.com/tektoncd/pipeline/blob/main/test/trustedresources.go should be refactored into files where they are called, this will improved the readability 

1. try to remove the https://github.com/tektoncd/pipeline/blob/main/test/trustedresources.go by adding helper functions to each file

- [ ] https://github.com/tektoncd/pipeline/pull/6799
- [ ] https://github.com/tektoncd/pipeline/pull/6800

2. Clean up the test code to meet go style

Yongxuanzhang:
# Changes

This commit fixes #5820. It refactors the trusted resouces testing functions into a testing pkg. This will avoid running unit test on these functions.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


lbernick:
Why does this PR remove unit tests? I'm confused about the commit message-- why would moving code into a different package avoid running unit tests, and why would we not want to run unit tests?

Yongxuanzhang:
> Why does this PR remove unit tests? I'm confused about the commit message-- why would moving code into a different package avoid running unit tests, and why would we not want to run unit tests?

They are testing helper functions, like https://github.com/tektoncd/pipeline/tree/main/pkg/reconciler/testing, I think we don't do unit testing on these functions right?

lbernick:
> > Why does this PR remove unit tests? I'm confused about the commit message-- why would moving code into a different package avoid running unit tests, and why would we not want to run unit tests?
> 
> They are testing helper functions, like https://github.com/tektoncd/pipeline/tree/main/pkg/reconciler/testing, I think we don't do unit testing on these functions right?

It looks like there's a private helper function, `signInterface`, that's used in the test helpers `GetSignedPipeline` and `GetSignedTask`, and this PR is removing tests for that helper function. A few comments:

- This seems like a lot of logic to live in test helpers. Any time there's a lot of setup logic in tests, it makes the test harder to understand and gives us less confidence that the code being tested works as intended. Needing tests for our test helpers seems to me like a code smell.
- Why not test the publicly exported functions instead, i.e. GetSignedPipeline and GetSignedTask, since those are the functions used by other packages?
- There's no reason to treat "helper" functions differently from other functions. It makes sense to unit test a function if there's enough logic in it to be worth testing (e.g. I probably wouldn't test a one-liner), and this one does look like it has a significant amount of logic. See https://bobcatfish.com/post/helpers/

Yongxuanzhang:
> > > Why does this PR remove unit tests? I'm confused about the commit message-- why would moving code into a different package avoid running unit tests, and why would we not want to run unit tests?
> > 
> > 
> > They are testing helper functions, like https://github.com/tektoncd/pipeline/tree/main/pkg/reconciler/testing, I think we don't do unit testing on these functions right?
> 
> It looks like there's a private helper function, `signInterface`, that's used in the test helpers `GetSignedPipeline` and `GetSignedTask`, and this PR is removing tests for that helper function. A few comments:
> 
> * This seems like a lot of logic to live in test helpers. Any time there's a lot of setup logic in tests, it makes the test harder to understand and gives us less confidence that the code being tested works as intended. Needing tests for our test helpers seems to me like a code smell.
> * Why not test the publicly exported functions instead, i.e. GetSignedPipeline and GetSignedTask, since those are the functions used by other packages?
> * There's no reason to treat "helper" functions differently from other functions. It makes sense to unit test a function if there's enough logic in it to be worth testing (e.g. I probably wouldn't test a one-liner), and this one does look like it has a significant amount of logic. See https://bobcatfish.com/post/helpers/

Thanks so much for the suggestions!
- What are you suggestions for setup functions if they seem a code smell? What's the best practice for this?
- Yeah we can do that. Are you suggesting: instead of moving them to `testing` pkg, we just leave them in the previous pkg and add more unit tests?
- Sure!

lbernick:
> Thanks so much for the suggestions!
> 
> * What are you suggestions for setup functions if they seem a code smell? What's the best practice for this?

I don't think having test setup functions in general is a code smell, but having one so complex that it needs tests itself is a code smell. Best practices are to test the functions that are publicly exported as part of the package API (i.e. the tests should use the same functions as the code that imports the package), and to use dependency injection. I'll take some time to try to make some more specific suggestions but I'm not as familiar with this part of the codebase.

> * Yeah we can do that. Are you suggesting: instead of moving them to `testing` pkg, we just leave them in the previous pkg and add more unit tests?

I don't have a strong opinion on what package `GetSignedPipeline` and `GetSignedTask` should be in, but if you're exporting them for use in other packages they should have unit tests (rather than testing the unexported function `signInterface`. This is why we recommend having unit tests in a separate package from the code they're testing. I'd also reconsider whether it makes sense to have these functions as test helpers. Why do tests need signed pipelines, but non-test code doesn't need these functions? Is there a better API that would make sense for non-test code, that would eliminate the need to do this extra setup in tests?

Side note: I'm also noticing the test helper functions `GetUnsignedPipeline` and `GetUnsignedTask` that just seem to return test data. I'd suggest defining the test data where it's actually used, as this makes the test much easier to understand. It doesn't make a lot of sense to me to have these functions exported in a test package, since they're very specific to the test cases rather than test setup.

Yongxuanzhang:
> > Thanks so much for the suggestions!
> > 
> > * What are you suggestions for setup functions if they seem a code smell? What's the best practice for this?
> 
> I don't think having test setup functions in general is a code smell, but having one so complex that it needs tests itself is a code smell. Best practices are to test the functions that are publicly exported as part of the package API (i.e. the tests should use the same functions as the code that imports the package), and to use dependency injection. I'll take some time to try to make some more specific suggestions but I'm not as familiar with this part of the codebase.
> 
> > * Yeah we can do that. Are you suggesting: instead of moving them to `testing` pkg, we just leave them in the previous pkg and add more unit tests?
> 
> I don't have a strong opinion on what package `GetSignedPipeline` and `GetSignedTask` should be in, but if you're exporting them for use in other packages they should have unit tests (rather than testing the unexported function `signInterface`. This is why we recommend having unit tests in a separate package from the code they're testing. I'd also reconsider whether it makes sense to have these functions as test helpers. Why do tests need signed pipelines, but non-test code doesn't need these functions? Is there a better API that would make sense for non-test code, that would eliminate the need to do this extra setup in tests?
> 
> Side note: I'm also noticing the test helper functions `GetUnsignedPipeline` and `GetUnsignedTask` that just seem to return test data. I'd suggest defining the test data where it's actually used, as this makes the test much easier to understand. It doesn't make a lot of sense to me to have these functions exported in a test package, since they're very specific to the test cases rather than test setup.

Thanks! I will think about it and try to refactor in a better way

Yongxuanzhang:
will open later

PR #312, opened 2018-12-05
Add ClusterTask new CRD

pivotal-nader-ziada:
Fixes: #251 

## Proposed changes:
 - new type for ClusterTask
 - Add kind to TaskRef to differentiate between Task kinds
 - TaskInterface to be used by controllers, implemented by Task and ClusterTask
 - update taskRun and pipelineRun to use TaskInterface
 - update docs and tests

PR #5977, opened 2023-01-09
Cleanup resources of ClusterTask

JeromeJu:
# Changes
This commit cleans up the unused functions for ClusterTask. It aims to keep the hygiene of the codebase and gets the repo ready for swapping the storage version to v1.
- Remove the GetClusterTask type which is not referenced anywhere in the codebase, this was first introduced in https://github.com/tektoncd/pipeline/pull/312 and not used after the removal of v1alpha1 directory from `./pkg/reconciler/v1alpha1`

/kind cleanup
part of #5979

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
remove `GetClusterTask` func
```

### DIFF
--- a/pkg/reconciler/taskrun/resources/taskspec.go
+++ b/pkg/reconciler/taskrun/resources/taskspec.go
@@ -32,9 +32,6 @@ type GetTask func(context.Context, string) (v1beta1.TaskObject, *v1beta1.ConfigS
 // GetTaskRun is a function used to retrieve TaskRuns
 type GetTaskRun func(string) (*v1beta1.TaskRun, error)
 
-// GetClusterTask is a function that will retrieve the Task from name and namespace.
-type GetClusterTask func(name string) (v1beta1.TaskObject, error)
-
 // GetTaskData will retrieve the Task metadata and Spec associated with the
 // provided TaskRun. This can come from a reference Task or from the TaskRun's
 // metadata and embedded TaskSpec.


JeromeJu:
I am wondering if those functions are tested then they are not regarded as dead codes?

JeromeJu:
part of #5979

JeromeJu:
/retest
Hub was down

XinruZhang (approved):
Thanks for the change @JeromeJu!

Yongxuanzhang:
👍 I think release note is needed?

JeromeJu:
> 👍 I think release note is needed?

I think we might not as this is not part of the API but the reconciler?

Yongxuanzhang:
> > 👍 I think release note is needed?
> 
> I think we might not as this is not part of the API but the reconciler?

It could be a using facing change, since it is public type?
https://pkg.go.dev/github.com/tektoncd/pipeline@v0.43.1/pkg/reconciler/taskrun/resources#GetClusterTask

JeromeJu:
> > > 👍 I think release note is needed?
> > 
> > 
> > I think we might not as this is not part of the API but the reconciler?
> 
> It could be a using facing change, since it is public type? https://pkg.go.dev/github.com/tektoncd/pipeline@v0.43.1/pkg/reconciler/taskrun/resources#GetClusterTask

Makes sense! Thanks for pointing this out!

lbernick:
/approve

nit on the release note: the function is being removed, not deprecated ("deprecated" would make me think it's still usable)

Issue #5833, opened 2022-12-05
Switch the default for "embedded-status" feature flag to "minimal" on or after Jan 25, 2023


abayer:
The [deprecations list entry for `PipelineRun.Status.TaskRuns` and `PipelineRun.Status.Runs`](https://github.com/tektoncd/pipeline/blob/09c7285609717d72a318058806b18fc6da4e9643/docs/deprecations.md?plain=1#L23) has Jan 25, 2023 as the earliest date for removing those fields. I'd like to propose that we switch the default to `minimal` on Jan 25, 2023, and do the actual removal later.

Technically, we could switch the default now and actually remove `PipelineRun.Status.TaskRuns`, `PipelineRun.Status.Runs`, and the `embedded-status` flag (more specifically, the `full` and `both` options for the flag) on or after Jan 25, 2023, without violating the deprecation/removal policy, but I feel like changing the default to `minimal` only on/after the deprecation/removal date is more in the spirit of the policy. 

Additionally, when we remove the `full` and `both` options, we may went to keep the `PipelineRun.Status` fields in `v1beta1` for the rest of `v1beta1`s lifetime - once we remove the `full`/`both` options, those fields will no longer get written to by new `PipelineRun`s, and we can remove all of the internal code from the reconciler for handling the `PipelineRun.Status` fields, but any `PipelineRun` created before the switch to `minimal` by default or before the removal of `full`/`both` where someone manually sets `embedded-status` to one of those values which still exists in etcd would run into problems when, for example, getting deleted (based on past experiences I've had with fields I deleted). It feels like it'll make the most sense to just make those fields read-only and never actually remove them.

Thoughts, @tektoncd/core-maintainers, @JeromeJu?

JeromeJu:
Thanks Andrew for creating this issue. Marking `PipelineRun.Status.TaskRuns` and `PipelineRun.Status.Runs` as read-only instead of entirely removing them sounds reasonable.

> someone manually sets embedded-status to one of those values which still exists in etcd would run into problems when
I think we shall also include the case where the `embedded-status` is flipped back and forth. (added as my TODO)

/assign

jerop:
related to https://github.com/tektoncd/pipeline/issues/4954

lbernick:
[...]

Issue #4954, opened 2022-06-09
TEP-0100: `PipelineRun` Status Migration


jerop:
[TEP-0100: Embedded TaskRuns and Runs Status in PipelineRuns][tep-0100] proposed changes to `PipelineRun` status to reduce the amount of information stored about the status of `TaskRuns` and `Runs` to improve performance, reduce memory bloat and improve extensibility. 

This feature was released in [v0.35.0][35] gated behind the `embedded-status` feature flag, and this issue tracks the migration and removal of that feature flag.

- [x] [v0.35.0][35]: `embedded-status` flag defaults to `full`
- [x] [v0.44.0][44]: `embedded-status` flag defaults to `minimal`
- [ ] [v0.45.0][45]: `embedded-status` flag is removed

[tep-0100]: https://github.com/tektoncd/community/blob/main/teps/0100-embedded-taskruns-and-runs-status-in-pipelineruns.md
[35]: https://github.com/tektoncd/pipeline/releases/tag/v0.35.0
[44]: https://github.com/tektoncd/pipeline/releases/tag/v0.44.0
[45]: https://github.com/tektoncd/pipeline/releases/tag/v0.45.0

jerop:
06.24 update: we are dogfooding the minimal status in the alpha integration tests
- https://github.com/tektoncd/pipeline/pull/5032
- https://github.com/tektoncd/plumbing/pull/1096

JeromeJu:
Planning on how to remove the embedded-status feature flag entirely:
- Start from the pipelinerun reconciler? or the test cases?
- Finally cleanup the docs

afrittoli:
One part of this will be addressed in v0.44, but I think we will only close in v0.45, so perhaps we should move this to that milestone?

JeromeJu:
> One part of this will be addressed in v0.44, but I think we will only close in v0.45, so perhaps we should move this to that milestone?

The TEP100 and this issue shall close only in v0.45, sounds good to me. Thanks @afrittoli

JeromeJu:
- [ ] [v0.45.0](https://github.com/tektoncd/pipeline/releases/tag/v0.45.0): embedded-status flag is removed

   - For the changes in v0.45, the test cases with FullEmbeddedStatus shall be removed along with the flag value.

JeromeJu:
[...]

PR #5934, opened 2023-01-03
[TEP0100] Default to minimal embedded-status

JeromeJu:
# Changes
This commit changes the default of embedded-status to minimal status.

Prior to this PR, the `DefaultEmbeddedStatus` is `full`, the pipelineRun reconciler
will populate both `taskruns` and `runs` for `pipelineRunStatus`. With the change
to `minimal`, the reconciler populates `childReferences` instead.

fixes: #5833 
part of: #4954 

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
switch the default value of `embedded-status` to `minimal`
the `embedded-status` flag along with the `both` and `full` functionalities are deprecated, and will be removed in v0.45.
```

PR #5989, opened 2023-01-12
Fix PipelineRunStatus Reconciler for EmbeddedStatus Switch

JeromeJu:
# Changes
This commit fixes the updates for reconciling PipelineRunStatus when switching
the EmbeddedStatus feature flag. It resets the `status.runs` and `status.taskruns` 
to `nil` with `minimal` EmbeddedStatus and `status.childReferences` to `nil` with
`full` EmbeddedStatus.

Prior to this change, the `childReferences` `runs` and `taskruns` in pipelineRunStatus
would not have been reset to nil and could have led to validation error when the feature
flag is being switched. The following issue could help prevent such bugs in the future:
- https://github.com/tektoncd/pipeline/issues/5999

/kind bug

fixes: - #5991  
part of: #4954 #5833 

Thanks to @abayer @XinruZhang @lbernick 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -1331,20 +1331,32 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
 
 func updatePipelineRunStatusFromChildObjects(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, taskRuns []*v1beta1.TaskRun, runObjects []v1beta1.RunObject) error {
 	cfg := config.FromContextOrDefaults(ctx)
-	fullEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.FullEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
-	minimalEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.MinimalEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
 
-	if minimalEmbedded {
-		updatePipelineRunStatusFromChildRefs(logger, pr, taskRuns, runObjects)
-	}
-	if fullEmbedded {
+	switch cfg.FeatureFlags.EmbeddedStatus {
+	case config.FullEmbeddedStatus:
+		// Clear any ChildReferences that are present.
+		// This can occur if the value of "embedded-status" was modified during PipelineRun execution.
+		pr.Status.ChildReferences = nil
 		updatePipelineRunStatusFromTaskRuns(logger, pr, taskRuns)
 		updatePipelineRunStatusFromCustomRunsOrRuns(logger, pr, runObjects)
+	case config.BothEmbeddedStatus:
+		updatePipelineRunStatusFromTaskRuns(logger, pr, taskRuns)
+		updatePipelineRunStatusFromCustomRunsOrRuns(logger, pr, runObjects)
+		updatePipelineRunStatusFromChildRefs(logger, pr, taskRuns, runObjects)
+	default:
+		// Clear any TaskRuns and Runs present in the status.
+		// This can occur if the value of "embedded-status" was modified during PipelineRun execution.
+		pr.Status.Runs = nil
+		pr.Status.TaskRuns = nil
+		updatePipelineRunStatusFromChildRefs(logger, pr, taskRuns, runObjects)
 	}
 
 	return validateChildObjectsInPipelineRunStatus(ctx, pr.Status)
 }
 
+// TODO: https://github.com/tektoncd/pipeline/issues/5999 integration tests for changing
+// feature flags could help prevent validation errors when the feature-flag is being switched.
+// In such case, the error could keep the pipelineRun run indefinitely.
 func validateChildObjectsInPipelineRunStatus(ctx context.Context, prs v1beta1.PipelineRunStatus) error {
 	cfg := config.FromContextOrDefaults(ctx)
 
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -5161,7 +5161,7 @@ status:
 		expectedPr = expectedPrMinimalStatus
 	}
 
-	if d := cmp.Diff(expectedPr, reconciledRun, ignoreResourceVersion, ignoreLastTransitionTime, ignoreCompletionTime, ignoreStartTime); d != "" {
+	if d := cmp.Diff(expectedPr, reconciledRun, ignoreResourceVersion, ignoreLastTransitionTime, ignoreCompletionTime, ignoreStartTime, cmpopts.EquateEmpty()); d != "" {
 		t.Errorf("expected to see pipeline run results created. Diff %s", diff.PrintWantGot(d))
 	}
 }
@@ -5482,7 +5482,7 @@ status:
 		expectedPr = expectedPrMinimalStatus
 	}
 
-	if d := cmp.Diff(expectedPr, reconciledRun, ignoreResourceVersion, ignoreLastTransitionTime, ignoreCompletionTime, ignoreStartTime); d != "" {
+	if d := cmp.Diff(expectedPr, reconciledRun, ignoreResourceVersion, ignoreLastTransitionTime, ignoreCompletionTime, ignoreStartTime, cmpopts.EquateEmpty()); d != "" {
 		t.Errorf("expected to see pipeline run results created. Diff %s", diff.PrintWantGot(d))
 	}
 }
@@ -8947,7 +8947,7 @@ spec:
 			if err != nil {
 				t.Fatalf("Got an error getting reconciled run out of fake client: %s", err)
 			}
-			if d := cmp.Diff(tt.expectedPipelineRun, pipelineRun, ignoreResourceVersion, ignoreTypeMeta, ignoreLastTransitionTime, ignoreStartTime, ignoreFinallyStartTime); d != "" {
+			if d := cmp.Diff(tt.expectedPipelineRun, pipelineRun, ignoreResourceVersion, ignoreTypeMeta, ignoreLastTransitionTime, ignoreStartTime, ignoreFinallyStartTime, cmpopts.EquateEmpty()); d != "" {
 				t.Errorf("expected PipelineRun was not created. Diff %s", diff.PrintWantGot(d))
 			}
 		})
@@ -9555,7 +9555,7 @@ spec:
 			if err != nil {
 				t.Fatalf("Got an error getting reconciled run out of fake client: %s", err)
 			}
-			if d := cmp.Diff(tt.expectedPipelineRun, pipelineRun, ignoreResourceVersion, ignoreTypeMeta, ignoreLastTransitionTime, ignoreStartTime, ignoreFinallyStartTime); d != "" {
+			if d := cmp.Diff(tt.expectedPipelineRun, pipelineRun, ignoreResourceVersion, ignoreTypeMeta, ignoreLastTransitionTime, ignoreStartTime, ignoreFinallyStartTime, cmpopts.EquateEmpty()); d != "" {
 				t.Errorf("expected PipelineRun was not created. Diff %s", diff.PrintWantGot(d))
 			}
 		})
@@ -10015,7 +10015,7 @@ status:
 			if err != nil {
 				t.Fatalf("Got an error getting reconciled run out of fake client: %s", err)
 			}
-			if d := cmp.Diff(tt.expectedPipelineRun, pipelineRun, ignoreResourceVersion, ignoreTypeMeta, ignoreLastTransitionTime, ignoreStartTime, cmpopts.SortSlices(lessChildReferences)); d != "" {
+			if d := cmp.Diff(tt.expectedPipelineRun, pipelineRun, ignoreResourceVersion, ignoreTypeMeta, ignoreLastTransitionTime, ignoreStartTime, cmpopts.SortSlices(lessChildReferences), cmpopts.EquateEmpty()); d != "" {
 				t.Errorf("expected PipelineRun was not created. Diff %s", diff.PrintWantGot(d))
 			}
 		})
@@ -10549,7 +10549,7 @@ spec:
 			if err != nil {
 				t.Fatalf("Got an error getting reconciled run out of fake client: %s", err)
 			}
-			if d := cmp.Diff(tt.expectedPipelineRun, pipelineRun, ignoreResourceVersion, ignoreTypeMeta, ignoreLastTransitionTime, ignoreStartTime, ignoreFinallyStartTime); d != "" {
+			if d := cmp.Diff(tt.expectedPipelineRun, pipelineRun, ignoreResourceVersion, ignoreTypeMeta, ignoreLastTransitionTime, ignoreStartTime, ignoreFinallyStartTime, cmpopts.EquateEmpty()); d != "" {
 				t.Errorf("expected PipelineRun was not created. Diff %s", diff.PrintWantGot(d))
 			}
 		})


JeromeJu:
I also thought https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L207-L214 shall be wrapped in a condition check of Full or Both embeddedStatus reading the comments at https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L826

 But that doesn't seem to be the fix, 
```
if cfg.FeatureFlags.EmbeddedStatus == config.FullEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus {
```

Would you mind taking a look @abayer ?

abayer:
Ah-ha! It's test fixtures with `.status.runs: {}` and/or `.status.taskRuns: {}` that's the problem. With your change, when the embedded status is `minimal`, those fields will no longer be empty maps, but will instead be `nil`. A commit that fixes this can be found at https://github.com/abayer/tektoncd-pipeline/commit/ca0dd7d0586d44cb8c2a3d3b7a92502f36f24369

JeromeJu:
Thanks @abayer for the fix, I will integrate this with defaulting the `minimal` embeddedStatus in https://github.com/tektoncd/pipeline/pull/5934.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1334,6 +1334,11 @@ func updatePipelineRunStatusFromChildObjects(ctx context.Context, logger *zap.Su
 	fullEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.FullEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
 	minimalEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.MinimalEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
 
+	if cfg.FeatureFlags.EmbeddedStatus == config.MinimalEmbeddedStatus {
```
chitrangpatel, line 1337: should the RHS of == be `minimalEmbedded` or `config.MinimalEmbeddedStatus` (double checking)

JeromeJu: Thank you for the checking, I think this should be the `MinimalEmbeddedStatus` which is string 'minimal' in the `config-feature-flags.yaml` because the minimalEmbedded above is a union of `Both` and `Minimal` EmbeddedStatus.

chitrangpatel: Sounds good!

chitrangpatel (approved):

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1334,6 +1334,11 @@ func updatePipelineRunStatusFromChildObjects(ctx context.Context, logger *zap.Su
...
+		pr.Status.TaskRuns = nil
+		pr.Status.Runs = nil
+	}
```
lbernick, line 1340: I'm not sure I understand this change. Why are `pr.status.taskruns` and `pr.status.runs` set in the first place if embedded-status is "minimal"?

Is the change just from an empty slice to a nil slice? This is my interpretation based on the tests. If that's the case, we should change the tests to not care whether an empty slice or a nil slice is used.

If these fields are set incorrectly (i.e. to something other than an empty slice or nil pointer), can we change the part of the code that sets them, add tests, and link a bug that gives more context on the problem being encountered?

lbernick (changes requested):

pkg/reconciler/pipelinerun/pipelinerun.go
abayer: Yeah, it's empty map to nil map. I remember having weird hassles with those maps being empty sometimes and nil sometimes when I added the minimal status in the first place, but can't remember details.

JeromeJu: > Why are pr.status.taskruns and pr.status.runs set in the first place if embedded-status is "minimal"?
In the 'normal' cases, it should be the case as described. But since the context was switching the defaultEmbeddedStatus from `full` to `minimal`, there could be edge cases where an existing piplineRun has `.stats.taskruns` or `.status.runs`. So in this case, it would error out at https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L1362 or https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L1362

JeromeJu: This case happened to cover this case:
https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_test.go#L1891. But I am not sure if a similar one shall be added to this PR as the 'DefaultEmbeddedStatus`  to minimal would be in #5934 . If preferred, I can add this edge case in that PR?

XinruZhang: > Why are pr.status.taskruns and pr.status.runs set in the first place if embedded-status is "minimal"?

Great question, this could be a very rare edge case that @JeromeJu  wants to catch: say a PipelienRun contains a TaskRun that could run for a period time. While the TaskRun is running, we switched the embedded status feature flag from `full` to `minimum`, at this time, `pr.status.runs` are set while embedded-status is "minimal".

JeromeJu: Thanks @lbernick , updated with the comment. But I feel like setting them as `nil` rather than having any empty `status.runs` or `status.taskruns` would be more appropriate since we are going to remove them eventually according to TEP100?

lbernick: Can you elaborate a bit? having `status.runs` set to nil is meaningfully the same as having it set to an empty map. No client should be differentiating between the two cases, which is why I suggest our tests do the same. I'm not sure why setting these fields to nil would make the transition to removing these fields easier.

Also, do you think it is possible to write an integration test for this scenario?

JeromeJu: > Hm I'm wondering if it would be possible/worthwhile to try to write an integration test for this scenario?

I would like to confirm, by an integration test, what are the scenarios that need to be covered?

I think for the case where there could be a pipelineRunStatus with existing `status.runs` or `status.taskruns` could have been covered by [TestReconcileForCustomTaskWithPipelineRunTimedOut](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_test.go#L1891). I am willing to create a thorough one if there're more complicated scenarios 🤔

lbernick: It sounds like the goal of this PR is to avoid errors in the case where someone changes the value of embedded-status while a pipelineRun is running. How do we know if it works or not?

JeromeJu: > It sounds like the goal of this PR is to avoid errors in the case where someone changes the value of embedded-status while a pipelineRun is running. How do we know if it works or not?

Agreed. Then I think the integration test shall be added in the same PR as https://github.com/tektoncd/pipeline/pull/5934?

And I am trying to picture the cases to be covered: probably one PipelineRun with all the fields in `status.taskruns` being switched from `Full` or `Both` to `Minimal` shall be sufficient?

JeromeJu: > Can you elaborate a bit? having status.runs set to nil is meaningfully the same as having it set to an empty map. No client should be differentiating between the two cases, which is why I suggest our tests do the same. I'm not sure why setting these fields to nil would make the transition to removing these fields easier.

I hope the following makes sense to elaborate:

- After the `PipelineRunStatus` migration:
  The existence of `status.runs{}` and `status.taskruns{}` could suggest using them in some ways. I think removing them entirely could be mitigating the chances for those to be exposed to users.
- During the switch of the `DefaultEmbeddedStatus`:
  There should not be `status.taskruns` or `status.runs` populated if embedded-status is "minimal", removing them could also be a guarantee that these fields are not existent at all when defaulting to minimal.
- For now:
  It could avoid potential hassles of `status.taskruns` or `status.runs` as suggested

JeromeJu: > This should be in our code standards, but tests should be included in the same pull request as the code that's being tested.

Thanks Lee! Got it.

lbernick: > * After the `PipelineRunStatus` migration:
>   The existence of `status.runs{}` and `status.taskruns{}` could suggest using them in some ways. I think removing them entirely could be mitigating the chances for those to be exposed to users.
> * During the switch of the `DefaultEmbeddedStatus` to `minimal`:
>   There should not be `status.taskruns` or `status.runs` populated if embedded-status is "minimal", removing them could also be a guarantee that these fields are not existent at all when defaulting to minimal.
> * For now:
>   It could avoid potential hassles of `status.taskruns` or `status.runs` as suggested

I'm still a bit confused-- I'm not too concerned that users will think they should be using taskruns and runs if set to an empty map, when childrefs are populated. I think it's ok for taskruns and runs to be set to an empty map if embedded status is minimal.

It still sounds like the root problem is concern about users migrating to minimal embedded status experiencing bugs with their pipelinerun. Can you try to reproduce this issue on your own cluster, and if you're able to reproduce it, file a bug?

Based on the previous conversation on this PR, it seems like this PR might have been intended to fix a test failure on https://github.com/tektoncd/pipeline/pull/5934? If that's the case, you can fix this by using `cmp.Diff(want, got, cmpopts.EquateEmpty()`.

JeromeJu: Thanks Lee for the reminder, created https://github.com/tektoncd/pipeline/issues/5991. Could you help take a look if that makes sense for addressing the issue that's intended for this PR?

> Based on the previous conversation on this PR, it seems like this PR might have been intended to fix a test failure on https://github.com/tektoncd/pipeline/pull/5934? If that's the case, you can fix this by using cmp.Diff(want, got, cmpopts.EquateEmpty().

I don't think that case could be fixed with `cmp.Diff(want, got, cmpopts.EquateEmpty().`

JeromeJu:
Ideally, an integration test that includes the process of switching `Embedded status` is needed,

- [ ] Create an issue tracking this need and update in https://github.com/tektoncd/pipeline/issues/4954
- [ ] Update test plan of TEP100

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1334,6 +1334,14 @@ func updatePipelineRunStatusFromChildObjects(ctx context.Context, logger *zap.Su
 	fullEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.FullEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
 	minimalEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.MinimalEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
 
+	if cfg.FeatureFlags.EmbeddedStatus == config.MinimalEmbeddedStatus {
```
lbernick, line 1337: I think this set of conditionals will be more readable with a `switch` statement, since some of them overlap.
e.g.

```
switch cfg.FeatureFlags.EmbeddedStatus {
  case  config.MinimalEmbeddedStatus:
    // clear taskruns and runs
   // add childrefs
  case config.FullEmbeddedStatus:
    // add taskruns and runs
   // clear childrefs
  case config.BothEmbeddedStatus:
   // add both
}
```

pkg/reconciler/pipelinerun/pipelinerun.go
lbernick: #5991 won't be fixed with `cmpopts.EquateEmpty` but I think it's still best practice and worth adding here.

lbernick (commented):
The errors [here](https://github.com/tektoncd/pipeline/blob/12457e22680fb558720cf4997c9c84568bf27959/pkg/reconciler/pipelinerun/pipelinerun.go#L1353-L1367) don't apply anymore, can you please update this section?

In addition, please restore the comments about why the objects are cleared from the pipelinerun status. I created https://github.com/tektoncd/pipeline/issues/5999, so you can just add a TODO (in comments) for integration tests. No updates are needed to TEP-0100.

Lastly, can you please update the commit message to describe the "why" of the changes in a bit more detail? e.g. "If a cluster operator switches the value of `embedded-status` while a PipelineRun is running, the reconciler previously returned an error; with these changes, the reconciler updates the PipelineRun's status to reflect the new value of embedded-status."

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1350,6 +1355,8 @@ func validateChildObjectsInPipelineRunStatus(ctx context.Context, prs v1beta1.Pi
 
 	var err error
 
+	// TODO: Integration test to be introduced for switching in embeddedStatus
+	// https://github.com/tektoncd/pipeline/issues/5999
 	// Verify that we don't populate child references for "full"
```
JeromeJu, line 1360: > The errors [here](https://github.com/tektoncd/pipeline/blob/12457e22680fb558720cf4997c9c84568bf27959/pkg/reconciler/pipelinerun/pipelinerun.go#L1353-L1367) don't apply anymore, can you please update this section?

Sorry I haven't fully understood here why does it not apply anymore? I thought though we are solving the bugs for the context that the `embeddedStatus` feature flag is switching these validation shall still apply?

lbernick: The error conditions are that we have "full" embedded-status + childRefs, or "minimal" status with taskruns + runs. With your changes, those conditions can't happen anymore, so I think it would be OK to remove this error.
(However, I could see the argument for not doing so, which is that `validateChildObjects...` shouldn't understand the details of how it's called, so I'll leave this up to you.)

JeromeJu: Thanks @lbernick  for elaborating. I might leave it until the integration test was implemented to have more thorough testing.

lbernick:
@JeromeJu can you please add code comments that explain the reasons for the change and link to the TODO? That way, people don't have to context switch to the pull request when they're looking at the code in order to find the reason for the change (although we still want to keep this info in the pull request description)

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1331,15 +1331,20 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
-	fullEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.FullEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
-	minimalEmbedded := cfg.FeatureFlags.EmbeddedStatus == config.MinimalEmbeddedStatus || cfg.FeatureFlags.EmbeddedStatus == config.BothEmbeddedStatus
 
-	if minimalEmbedded {
-		updatePipelineRunStatusFromChildRefs(logger, pr, taskRuns, runObjects)
-	}
-	if fullEmbedded {
+	switch cfg.FeatureFlags.EmbeddedStatus {
+	case config.FullEmbeddedStatus:
+		pr.Status.ChildReferences = nil
```
lbernick, line 1337:
```suggestion
                 // Clear any ChildReferences that are present.
                 // This can occur if the value of "embedded-status" was modified during PipelineRun execution.
		pr.Status.ChildReferences = nil
```

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1331,15 +1331,20 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
+		updatePipelineRunStatusFromTaskRuns(logger, pr, taskRuns)
+		updatePipelineRunStatusFromCustomRunsOrRuns(logger, pr, runObjects)
+	case config.BothEmbeddedStatus:
 		updatePipelineRunStatusFromTaskRuns(logger, pr, taskRuns)
 		updatePipelineRunStatusFromCustomRunsOrRuns(logger, pr, runObjects)
+		updatePipelineRunStatusFromChildRefs(logger, pr, taskRuns, runObjects)
+	default:
+		pr.Status.Runs = nil
+		pr.Status.TaskRuns = nil
```
lbernick, line 1346:
```suggestion
                 // Clear any TaskRuns and Runs present in the status.
                 // This can occur if the value of "embedded-status" was modified during PipelineRun execution.
		pr.Status.Runs = nil
		pr.Status.TaskRuns = nil
```

lbernick (approved):

pkg/reconciler/pipelinerun/pipelinerun.go
JeromeJu: Thanks, done

lbernick:
@JeromeJu can you please update the release notes with details of the bug fix?

JeromeJu:
> @JeromeJu can you please update the release notes with details of the bug fix?

Updated.

Issue #5967, opened 2023-01-06
Handle `PipelineResources` when swapping `v1` to the storage version


JeromeJu:
`PipelineResources` was deprecated but not removed from `v1beta1` according to https://github.com/tektoncd/pipeline/issues/5785 and the WG decisions at [Apr 11, 2022](https://drive.google.com/file/d/1tqMBe_dHY2GV9_yZXfZzZaqUhRjJSjvz/view?usp=sharing) and confirmed at Nov 28, 2022.

However when swapping v1 to be the storage version, the `PipelineResources` related logics could not be handled by changing `v1beta1` to `v1`. eg. https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/taskrun/validate_resources.go#L217 would not be valid if we swap to `v1.TaskSpec` as `PipelineResources` are removed during the migration to v1.

One approach would need to be decided about handling the deprecated but not removed fields as `PipelineResources` for reconciling.

/kind misc

vdemeester:
One thing to consider, because those fields are in `v1beta1`, we can remove them from `v1beta1` (given the proper warning, deprecation notice, …). Doing this "just before" we switch to `v1` would help us a lot 👼🏼 to not make too much complexity for a deprecated feature.

JeromeJu:
subTask of https://github.com/tektoncd/pipeline/issues/5541

dibyom:
Yeah pipeline resources have been deprecated since v0.30 so I think it can be removed

pritidesai:
Yes please, let's remove from the code base which will give us an opportunity to comply the [use cases](https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md#use-cases) mentioned in the TEP. At the same time, an opportunity to reconsider and design features which could potentially replace resources - https://github.com/tektoncd/community/blob/main/teps/0074-deprecate-pipelineresources.md#features-that-will-replace-pipelineresources-functionality

JeromeJu:
Also updated TEP0074 at https://github.com/tektoncd/community/pull/922
Thanks for the reminder cc @lbernick

JeromeJu:
[...]

PR #5995, opened 2023-01-16
[TEP074] Remove CloudEvent pipelineResources

JeromeJu:
# Changes
This commit removes the CloudEvent Resources support.
Removal of `pipelineResources`, as in #5967 has been broken up into removal of each
resources packages for the `small PR` code standard.
This PR removes `github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1/cloudevent`
the respective example test and docs for cloud-event resources.

Part of #5967 
/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
action required: please migrate off of `cloudevent` `pipelineresources` as it is removed, please refer to the doc at https://github.com/tektoncd/pipeline/blob/main/docs/pipelineresources.md#replacing-a-cloudevent-resource
```


third_party/github.com/cloudflare/circl/LICENSE
```diff
@@ -0,0 +1,57 @@
+Copyright (c) 2019 Cloudflare. All rights reserved.
```
JeromeJu, line 1: Running update-codegen on main to investigate these unexpected changes for the licenses.

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -3149,102 +3149,6 @@ spec:
   taskRef:
```
JeromeJu, line 3149: I am under the impression that this whole test and CloudEvent could be removed along with the removal of CloudEvent resources.

pkg/reconciler/events/cloudevent/cloud_event_controller.go
```diff
@@ -40,24 +38,6 @@ import (
...
-				targets = append(targets, cer.TargetURI)
-			}
-		}
-		if len(targets) > 0 {
-			tr.Status.CloudEvents = cloudEventDeliveryFromTargets(targets)
-		}
-	}
-}
-
 func cloudEventDeliveryFromTargets(targets []string) []v1beta1.CloudEventDelivery {
```
lbernick, line 41: Are any other functions in this file now dead code?

pkg/reconciler/taskrun/taskrun_test.go
lbernick: yes, I think so

lbernick (approved):
@JeromeJu could you please add "action required" to the release notes, and a bit more detail on what's expected of users upgrading to the new release with these changes? (i.e. migrate off of cloudevent pipelineresources, with links to docs on how to do that)

pkg/reconciler/events/cloudevent/cloud_event_controller.go
JeromeJu: Thanks for the reminder, looking into it!

JeromeJu: Done, not any dead code that could be removed with CloudEvent resources.

afrittoli: There are two code paths for CloudEvents, one is associated with the pipeline resource and the other one with the CloudEvents feature.  If I remember correctly, `SendCloudEvents` is used by the former and `SendCloudEventsWithRetries` is used by the latter. 
I need to spend a bit more time reviewing the code to remember exactly what can be removed. I would recommend to try to remove the  `SendCloudEvents` and related usages and check if all cloud events test cases are still passing.

docs/resources.md
```diff
@@ -927,67 +927,7 @@ service account.
...
-          "name": "source-to-image",
-          "kind": "Task"
-        },
-        "timeout": "1h0m0s"
-      },
-      "status": {...}
-    }
-  }
-```
+CloudEvent Resource is deprecated and please refer to the doc at https://github.com/tektoncd/pipeline/blob/main/docs/pipelineresources.md#replacing-a-cloudevent-resource for migration.
```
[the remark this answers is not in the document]
JeromeJu: Thanks for the suggestion. Done.

pkg/reconciler/events/cloudevent/cloud_event_controller.go
JeromeJu: Thanks Andrea, deleted:
- cloudEventDeliveryFromTargets
- SendCloudEvents

afrittoli (commented):
Thank you @JeromeJu, this looks great. 

The 'cloudevent_count' metric counts cloudevent sent via the `PipelineResource`, so it doesn't make any more sense. I don't think we added metrics to the API stability policy, so it should be safe to remove. Besides, it wouldn't report anything at all.

I noticed that `taskRun.status.CloudEvents` has only just been marked as deprecated and is not planned for removal until Oct 2023. 

I though that when we deprecated `PipelineResources` we also deprecated all parts of the API associated with them, which would include resource inputs/outputs, resource results and cloudevents status. It may be that we were not explicit about the parts of the API affected, but I'm not sure we should keep around an API which is not used by anything.

@tektoncd/core-maintainers thoughts about this?

I don't think this discussion should block this PR, so it's fine to move ahead and possibly further cleanup in a follow-up PR. Perhaps here we could remove the metric at least?

JeromeJu:
Many thanks to the detailed look into this removal and the respective effects on the related APIs. I have done the following 
so this PR might get in more safely for now. :)
- Tracked this down in the original issue: https://github.com/tektoncd/pipeline/pull/6002
- Put into the pipeline WG as a discussion

JeromeJu:
Many thanks @afrittoli for the detailed look into this removal and the respective effects on the related APIs. I have done the following 
so this PR might get in more safely for now. :)
- Tracked this down in the original issue: https://github.com/tektoncd/pipeline/pull/6002
- Put into the pipeline WG as a discussion

vdemeester:
> I though that when we deprecated `PipelineResources` we also deprecated all parts of the API associated with them, which would include resource inputs/outputs, resource results and cloudevents status. It may be that we were not explicit about the parts of the API affected, but I'm not sure we should keep around an API which is not used by anything.
> 
> @tektoncd/core-maintainers thoughts about this?

That was my thought as well.

vdemeester (approved):

Issue #5727, opened 2022-11-07
Remote resolution for kaniko catalog examples


JeromeJu:
The compatible remote resolution task referred in the catalog repo needs to be updated in  https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/pipelinerun.yaml#L39-L40.

The yaml file to reproduce:
```
# This demo modifies the cluster (deploys to it) you must use a service
# account with permission to admin the cluster (or make your default user an admin
# of the `default` namespace with default-cluster-admin.
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  generateName: default-cluster-admin-
subjects:
  - kind: ServiceAccount
    name: default
    namespace: default
roleRef:
  kind: ClusterRole
  name: cluster-admin
  apiGroup: rbac.authorization.k8s.io
---
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: "unit.tests"
spec:
  workspaces:
  - name: source
    mountPath: /workspace/source/go/src/github.com/GoogleContainerTools/skaffold
  steps:
  - name: run-tests
    image: golang
    env:
    - name: GOPATH
      value: /workspace/go
    workingDir: $(workspaces.source.path)
    script: |
      # The intention behind this example Task is to run unit test, however we
      # currently do nothing to ensure that a unit test issue doesn't cause this example
      # to fail unnecessarily. In the future we could re-introduce the unit tests (since
      # we are now pinning the version of Skaffold we pull) or use Tekton Pipelines unit tests.
      echo "pass"
---
# This task deploys with kubectl apply -f <filename>
apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: demo-deploy-kubectl
spec:
  params:
  - name: path
    description: Path to the manifest to apply
  - name: yqArg
    description: Okay this is a hack, but I didn't feel right hard-coding `-d1` down below
  - name: yamlPathToImage
    description: The path to the image to replace in the yaml manifest (arg to yq)
  - name: imageURL
    description: The URL of the image to deploy
  workspaces:
  - name: source
  steps:
  - name: replace-image
    image: mikefarah/yq:3
    command: ['yq']
    args:
    - "w"
    - "-i"
    - "$(params.yqArg)"
    - "$(params.path)"
    - "$(params.yamlPathToImage)"
    - "$(params.imageURL)"
  - name: run-kubectl
[...]

PR #6002, opened 2023-01-17
[TEP074] Remove Image pipelineResources

JeromeJu:
# Changes
This commit removes the Image Resources support.

This PR removes `github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1/image`
the imagedigestexporter and the docs for image resources.It also removes the
 integration tests `helm_task_test` and`hermetic_taskrun_test` as these tests 
are testing building against image resources.

Removal of `pipelineResources`, as in https://github.com/tektoncd/pipeline/issues/5967 has been broken up into removal of each resources according to small PRs.

Part of #5967 
/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
action required: please migrate off of `image` `pipelineresources` as it is removed, please refer to the doc at https://github.com/tektoncd/pipeline/blob/main/docs/pipelineresources.md#replacing-an-image-resource
`imagedigestexporter` image will not be built nor maintained from now on
```


docs/resources.md
```diff
@@ -584,89 +584,7 @@ spec:
...
-    resourcesResult:
-      - key: "digest"
-        value: "sha256:eed29cd0b6feeb1a92bc3c4f977fd203c63b376a638731c88cacefe3adb1c660"
-        resourceName: skaffold-image-leeroy-web
-    # ...
-```
-
-If the `index.json` file is not produced, the image digest will not be included
-in the `taskRun` output.
+Image Resource is deprecated and please refer to the doc at https://github.com/tektoncd/pipeline/blob/main/docs/pipelineresources.md#replacing-an-image-resource for migration.
```
lbernick, line 587:
```suggestion
Support for image resources has been removed. Please refer to the doc at https://github.com/tektoncd/pipeline/blob/main/docs/pipelineresources.md#replacing-an-image-resource for migration.
```

suggesting since "deprecation" implies that it still works, while "removal" implies that it does not.

lbernick (approved):

vdemeester (approved):

test/yamls/v1beta1/pipelineruns/pipelinerun.yaml
```diff
@@ -89,18 +89,6 @@ spec:
     # makes it explicit that it needs to run as root.
     securityContext:
       runAsUser: 0
-  - name: write-digest
```
JeromeJu, line 92: Tried deleting these steps, but I have been wondering if this would make the whole test case meaningless without the image digest?

JeromeJu:
[TestHelmDeployPipelineRun](https://github.com/tektoncd/pipeline/blob/main/test/helm_task_test.go#L42) and [TestKanikoTaskRun](https://github.com/tektoncd/pipeline/blob/main/test/kaniko_task_test.go#L44) are highly dependent on `imagedigestexporter`, I am trying to figure out in which way shall we refactor these tests?

lbernick:
> [TestHelmDeployPipelineRun](https://github.com/tektoncd/pipeline/blob/main/test/helm_task_test.go#L42) and [TestKanikoTaskRun](https://github.com/tektoncd/pipeline/blob/main/test/kaniko_task_test.go#L44) are highly dependent on `imagedigestexporter`, I am trying to figure out in which way shall we refactor these tests?

Can you replace the PipelineResources in these test cases with Tasks, following the migration guide?

lbernick:
> > [TestHelmDeployPipelineRun](https://github.com/tektoncd/pipeline/blob/main/test/helm_task_test.go#L42) and [TestKanikoTaskRun](https://github.com/tektoncd/pipeline/blob/main/test/kaniko_task_test.go#L44) are highly dependent on `imagedigestexporter`, I am trying to figure out in which way shall we refactor these tests?
> 
> Can you replace the PipelineResources in these test cases with Tasks, following the migration guide?

I want to change my feedback here-- I don't think these tests actually have value without the image pipelineresource, so you can go ahead and remove them.

JeromeJu:
In this PR, one question regarding removing the image resource at https://github.com/tektoncd/pipeline/pull/6002 is that dependencies is within the apis at https://github.com/tektoncd/pipeline/pull/6002. But the image resources api itself is being removed in this PR. Then in this case, there are 2 options to move forward:
- etracting the image resources logics to /pkg/image to keep the imagedigestexporter image without image resources apis
- remove the imageDigestExporter with this removal of image resources

dibyom:
Isn't imagedigestexporter part of the image resource?

JeromeJu:
> Isn't imagedigestexporter part of the image resource?

I think so, so according to the dependencies we shall remove it along with the removal of the image `pipelineResources` apis?

vdemeester (approved):

examples/v1beta1/pipelineruns/pipelinerun.yaml
```diff
@@ -90,7 +90,7 @@ spec:
       runAsUser: 0
   - name: write-digest
     workingDir: $(workspaces.source.path)
-    image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/imagedigestexporter:latest
+    image: gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/imagedigestexporter:v0.11.1
```
lbernick, line 93: Can this example be updated to not use the imagedigestexporter image? I believe we mentioned in the WG yesterday that there's a suitable kaniko flag that can replace it.

dibyom: yeah we should just replace this kaniko task with the latest version of kaniko from the Hub

JeromeJu: > yeah we should just replace this kaniko task with the latest version of kaniko from the Hub
For v1beta1 I think we can but resolvers are behind beta feature flag in v1.

lbernick: Thanks for opening the issue! even if we can't use remote resolution here we should still migrate off usage of this image though

JeromeJu: Following up on the migration off the dependencies of the imageDigestExporter image, @QuanZhang-William has helped me with the investigation of errors with the latest knaiko task at https://github.com/tektoncd/pipeline/issues/5727#issuecomment-1306169832. According to the findings that the cataglog latest kaniko task works fine, it is the old `url` provided at https://github.com/GoogleContainerTools/skaffold/blob/main/examples/microservices/leeroy-web/Dockerfile was not compatible with the latest kaniko task. So here are two options moving forward to unblock this:

- [ ] delete the population of the image digest if the purpose of the `pipelinerun.yaml` was not to deploy the microservice from skaffold
- [ ] not to use the skaffold images examples for params, which would change the microservices to be built in `pipelinerun.yaml` also

JeromeJu: Do we have any preference on maintaining the existent pipelinerun.yaml examples?

QuanZhang-William: It looks to me that the goal of the example is to showcase a typical CICD pipeline (i.e. clone -> test -> build & push -> deploy). If that is the case I think we should go for #2.

test/yamls/v1beta1/pipelineruns/pipelinerun.yaml
pritidesai: I reviewed the examples and testes that are deleted so far as part of this PR. I would have preferred to replace the `image` resource with something equivalent but I am not sure what is that equivalent. 😞

examples/v1beta1/pipelineruns/pipelinerun.yaml
pritidesai: `pipelinerun.yaml` is one of the original examples and has been a great resource showcasing a basic CI/CD use case, please preserve it if possible, thanks!

JeromeJu:
/hold
This shall target to be the very last type of the removal of piplineResources according to https://github.com/tektoncd/pipeline/issues/5967#issuecomment-1428252386.

Blockers:
- [Migrate Kaniko Task off ImageDigestExporter](https://github.com/tektoncd/pipeline/pull/6094)
- [[TEP074] Remove Storage and Git PipelineResources](https://github.com/tektoncd/pipeline/pull/6150)

afrittoli:
> Blockers:

> https://github.com/tektoncd/pipeline/pull/6094
> https://github.com/tektoncd/pipeline/pull/6150

Thanks @JeromeJu, it sounds like those two should be part of the milestone too, or this should be removed. wdyt?

JeromeJu:
> > Blockers:
> 
> > #6094
> > #6150
> 
> Thanks @JeromeJu, it sounds like those two should be part of the milestone too, or this should be removed. wdyt?

Yeah it'd be appreciated if they could be added to the milestone @afrittoli  😺 . And sorry haven't been able to do so as there has been changes on the sequence of how the implementations should be planned out.

JeromeJu:
/unhold
Not that #6150 is awaiting review, I think we could get this removed first.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
```diff
@@ -2520,23 +2500,19 @@ func TestValidateResourceBindings_Missing(t *testing.T) {
...
-				},
 			},
 		},
 	}
 	pr := &v1beta1.PipelineRun{
 		ObjectMeta: metav1.ObjectMeta{Name: "pipelinerun"},
 		Spec: v1beta1.PipelineRunSpec{
 			PipelineRef: &v1beta1.PipelineRef{Name: "pipeline"},
-			Resources:   []v1beta1.PipelineResourceBinding{gitSweetResourceBinding},
+			Resources:   []v1beta1.PipelineResourceBinding{},
```
lbernick, line 2510: why remove this?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -933,71 +918,6 @@ spec:
 			name:  "sa-step",
 			cmd:   "/mycmd",
 		}}),
-	}, {
-		name:    "params",
-		taskRun: taskRunSubstitution,
```
lbernick, line 938: `taskRunSubstitution` hasn't been deleted; I think this test case should remain

test/ko_test.go
```diff
@@ -1,53 +0,0 @@
-//go:build e2e
-// +build e2e
```
lbernick, line 2: why remove this file?

examples/v1beta1/taskruns/optional-resources-with-clustertask.yaml
```diff
@@ -1,36 +0,0 @@
...
-  params:
-  - name: filename
-    type: string
-    default: "README.md"
-  resources:
-    inputs:
-      - name: git-repo
-        type: git
-        optional: true
-    outputs:
```
lbernick, line 17: I'd prefer just removing image resources from the examples that don't rely on building images rather than removing the examples entirely.

JeromeJu: Thanks I will restore them.
Previously I saw them as cases that would only be meaningful if the outputs resources could be used as a digest. But that makes sense to leave the rest of git resources until removal of all resources.

pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
JeromeJu: In this case the image-resource was the missing one but now since it is being removed, we would have to change to the other resource that should be missing but required by the pipeline resource declaration.

pkg/reconciler/taskrun/taskrun_test.go
JeromeJu: I think for this case, the expected steps and volumes will be (almost 🤔  )completely changed without imageResourceType as the outputs resources. Would it make sense to delete the `taskRunSubstitution` also since they are only referenced here?

test/ko_test.go
JeromeJu: These were just the helper functions moved here earlier (they should have been deleted in previous PR also.)

pkg/reconciler/taskrun/taskrun_test.go
lbernick: It looks like it gets some volumes from the image resource and some from the git resource. Can you just remove the ones that it gets from the image resource? (and same for params)

JeromeJu: That makes sense for maintaining code coverage:) Thanks!

PR #5737, opened 2022-11-09
Add prow env for beta integration test

JeromeJu:
This commit adds the prow env for beta integration test

# Changes
/kind feature
fixes #5734 

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [n/a] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
Adds pull-tekton-pipelines-beta-integration-test
```

Issue #5734, opened 2022-11-09
V1 pull-tekton-pipeline-beta-integration-tests


JeromeJu:
### Feature request

Prow env needs to be updated in the pipeline repo.

As v1 got introduced in #5579  , the feature flag gets a new valid value of `beta`. Therefore a new CI job `pull-tekton-pipeline-beta-integration-tests` to test features that are marked as beta in the v1 API will be needed. 

Since during the gap between v1 gets `served` in 0.42 release and when it gets stored in the next release (0.43 expected), the v1beta1 api is still served and needs to be tested.

### Use case
The beta features will be on by default in v1beta1 testing (which is now `pull-tekton-pipeline-integration-tests`) and `beta` features in v1 would be tested with a feature-flag of `beta`.

JeromeJu:
Shall be in plumbing repo

JeromeJu:
Link with https://github.com/tektoncd/plumbing/issues/1255

Issue #5865, opened 2022-12-12
Switch all the MustParseV1beta1TaskRun to MustParseV1TaskRun after v1beta1 deprecation


JeromeJu:
This is to track that once we deprecate v1beta1, we should just modify MustParseTaskRun to use v1 by default.

JeromeJu:
This could be a sub task when we are deprecating v1beta1.

PR #6031, opened 2023-01-24
Beta Example Tests

chitrangpatel:
This PR is runs example tests of beta features if the feature flag `enable-api-fields` is set to `beta`. This is an extension to PR https://github.com/tektoncd/pipeline/pull/5737. It also addresses a part of https://github.com/tektoncd/pipeline/issues/5734

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Run beta example tests when `enable-api-fields: beta` is set.
```
/kind feature

### DIFF
--- a/test/path_filtering.go
+++ b/test/path_filtering.go
@@ -56,6 +56,8 @@ func getPathFilter(t *testing.T) (pathFilter, error) {
 		f = stablePathFilter
 	case "alpha":
 		f = alphaPathFilter
+	case "beta":
+		f = betaPathFilter
 	}
 	if f == nil {
 		return nil, fmt.Errorf("unable to create path filter from feature gate %q", enabledFeatureGate)
@@ -90,11 +92,17 @@ func getFeatureGate(namespace string) (string, error) {
 // stablePathFilter returns true for any example that should be allowed to run
 // when "enable-api-fields" is "stable".
 func stablePathFilter(p string) bool {
-	return !(strings.Contains(p, "/alpha/") || strings.Contains(p, "/beta/"))
+	return !strings.Contains(p, "/alpha/") && !strings.Contains(p, "/beta/")
 }
 
 // alphaPathFilter returns true for any example that should be allowed to run
 // when "enable-api-fields" is "alpha".
 func alphaPathFilter(p string) bool {
-	return strings.Contains(p, "/alpha/") || strings.Contains(p, "/beta/") || stablePathFilter(p)
+	return true
+}
+
+// betaPathFilter returns true for any example that should be allowed to run
+// when "enable-api-fields" is "beta".
+func betaPathFilter(p string) bool {
+	return !strings.Contains(p, "/alpha/")
 }
--- a/test/path_filtering_test.go
+++ b/test/path_filtering_test.go
@@ -103,3 +103,50 @@ func TestAlphaPathFilter(t *testing.T) {
 		})
 	}
 }
+
+func TestBetaPathFilter(t *testing.T) {
+	for _, tc := range []struct {
+		path    string
+		allowed bool
+	}{{
+		path:    "/test.yaml",
+		allowed: true,
+	}, {
+		path:    "/alpha/test.yaml",
+		allowed: false,
+	}, {
+		path:    "/beta/test.yaml",
+		allowed: true,
+	}, {
+		path:    "/foo/test.yaml",
+		allowed: true,
+	}, {
+		path:    "/v1alpha1/taskruns/test.yaml",
+		allowed: true,
+	}, {
+		path:    "/v1alpha1/taskruns/alpha/test.yaml",
+		allowed: false,
+	}, {
+		path:    "/v1beta1/taskruns/test.yaml",
+		allowed: true,
+	}, {
+		path:    "/v1beta1/taskruns/alpha/test.yaml",
+		allowed: false,
+	}, {
+		path:    "/v1beta1/taskruns/alpha/test.yaml",
+		allowed: false,
+	}, {
+		path:    "/v1alpha1/pipelineruns/beta/test.yaml",
+		allowed: true,
+	}, {
+		path:    "/v1alpha1/pipelineruns/beta/test.yaml",
+		allowed: true,
+	}} {
+		name := strings.Replace(tc.path, "/", " ", -1)
+		t.Run(name, func(t *testing.T) {
+			if got := betaPathFilter(tc.path); got != tc.allowed {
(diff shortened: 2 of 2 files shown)


lbernick:
@chitrangpatel @JeromeJu does this address https://github.com/tektoncd/pipeline/issues/5734?

test/path_filtering.go
```diff
@@ -98,3 +100,9 @@ func stablePathFilter(p string) bool {
 func alphaPathFilter(p string) bool {
 	return strings.Contains(p, "/alpha/") || strings.Contains(p, "/beta/") || stablePathFilter(p)
 }
+
+// betaPathFilter returns true for any example that should be allowed to run
+// when "enable-api-fields" is "beta".
+func betaPathFilter(p string) bool {
```
JeromeJu, line 106: I have been trying to understand if there are any thing we need to do with the stable Path? Previously we regard `beta` as stable now that we have `v1` which is stable, otherwise it feels the beta testing just a union of stable tests?

JeromeJu:
There has not been yet any examples tests for now but I wonder if it might be helpful to include addition of paths in https://github.com/tektoncd/pipeline/blob/main/test/examples_test.go#L248-L269 as for the purpose/title of this PR?

test/path_filtering.go
chitrangpatel: I think:
`Alpha`: tests all examples inside `v1beta1/, v1beta1/alpha, v1beta1/beta` (basically everything)
`Beta`: tests all examples inside `v1beta1/ and v1beta1/beta` but not `v1beta1/alpha` (skip examples in alpha directory).
`Stable`: tests examples inside `v1beta1` (skip the `alpha` and `beta` folders).

chitrangpatel: So I don't think we need to make any changes to stablePath or alphaPath.

chitrangpatel:
> Thanks @chitrangpatel for setting up the examples tests for beta:)
> 
> I am aware that there has not yet been any examples tests but I wonder if it might be helpful to include addition of paths in https://github.com/tektoncd/pipeline/blob/main/test/examples_test.go#L248-L269 as for the purpose/title of this PR?

Hmm. I'm not sure anything needs to be updated there. What were you thinking of in particular?
The function you referenced is called by `TestExamples` which does the path filtering before calling it.

test/path_filtering.go
JeromeJu: I think previously `beta` is `stable` according to
https://github.com/tektoncd/community/blob/main/teps/0033-tekton-feature-gates.md#new-api-fields. But now that `GA` is stable, `beta` should not be in `stablePath`?

JeromeJu:
> https://github.com/tektoncd/pipeline/blob/main/test/examples_test.go#L248-L269

Right! sorry I was thinking ahead of when, in the future, we deprecate `v1beta1` but for now there should be nothing to change maybe we can have a similar TODO with https://github.com/tektoncd/pipeline/issues/5865 for [waitValidatePipelineRunDone](https://github.com/tektoncd/pipeline/blob/main/test/examples_test.go#L253) and `waitValidateTaskRunDone` here 😺

test/path_filtering.go
chitrangpatel: I was confused by that for a while too but if you look carefully at the stable path it's filtering on 

**!(** string.contains(alpha) || string.contains(beta) **)**

So its picking only the standard ones.

JeromeJu: Ahhhhh 👓 was in my shopping cart.
Thanks.

JeromeJu (approved):
lgtm

test/path_filtering.go
XinruZhang: Ohhh this is confusing 😵‍💫, wondering it would be better if we change the condition to !string.contains(alpha) && !string.contains(beta)

dibyom (approved):

test/path_filtering.go
```diff
@@ -90,11 +92,17 @@ func getFeatureGate(namespace string) (string, error) {
...
 // alphaPathFilter returns true for any example that should be allowed to run
 // when "enable-api-fields" is "alpha".
 func alphaPathFilter(p string) bool {
 	return strings.Contains(p, "/alpha/") || strings.Contains(p, "/beta/") || stablePathFilter(p)
 }
+
+// betaPathFilter returns true for any example that should be allowed to run
+// when "enable-api-fields" is "beta".
+func betaPathFilter(p string) bool {
+	return strings.Contains(p, "/beta/") || stablePathFilter(p)
```
lbernick, line 107: I think this can be simplified to `!strings.Contains(p, "/alpha/")`

test/path_filtering.go
```diff
@@ -90,11 +92,17 @@ func getFeatureGate(namespace string) (string, error) {
...
 // when "enable-api-fields" is "stable".
 func stablePathFilter(p string) bool {
-	return !(strings.Contains(p, "/alpha/") || strings.Contains(p, "/beta/"))
+	return !strings.Contains(p, "/alpha/") && !strings.Contains(p, "/beta/")
 }
 
 // alphaPathFilter returns true for any example that should be allowed to run
 // when "enable-api-fields" is "alpha".
 func alphaPathFilter(p string) bool {
 	return strings.Contains(p, "/alpha/") || strings.Contains(p, "/beta/") || stablePathFilter(p)
```
lbernick, line 101: I think this can be simplified to `return true`

test/path_filtering_test.go
```diff
@@ -103,3 +103,50 @@ func TestAlphaPathFilter(t *testing.T) {
...
+
+func TestBetaPathFilter(t *testing.T) {
+	for _, tc := range []struct {
+		path    string
+		allowed bool
+	}{{
+		path:    "/test.yaml",
+		allowed: true,
+	}, {
+		path:    "/alpha/test.yaml",
```
lbernick, line 115: Do paths ever contain "alpha" (i.e. missing a slash at the beginning or end)?

test/path_filtering.go
chitrangpatel: Yes! Thanks

test/path_filtering_test.go
chitrangpatel: I don't think so. These cases are the same as those from the stablePath tests. 
If we do encounter a case, we can update the filters later. But I doubt it.

test/path_filtering.go
chitrangpatel: Agreed! That's way simpler :)

chitrangpatel:
> /hold

@lbernick can we cancel the hold now?

PR #6049, opened 2023-01-26
[TEP0100] Remove `EmbeddedStatus` Feature Flag

JeromeJu:
# Changes

This commit removes the embedded-status flag following up on the
[Default to Minimal EmbeddedStatus](https://github.com/tektoncd/pipeline/pull/5934) in the v0.44 release. It removes all
the test cases related with the `full` and `both` `EmbeddedStatus` and
only kept the `minimal` test cases as default.

In detail, the test cases that checks the `taskruns` or `runs` status from
previous was changed to checking referenced `taskruns` or `runs` from
the `ChildReferences`.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
`embedded-status` feature flag is removed. TaskRun and Run status will be populated only in `pipelineRun.status.childReferences`.
```


lbernick:
> Put up to decide if the ones for the plumbing repo should go in first.

There's no need to change anything in plumbing afaik. Clients should be able to upgrade to the version of pipelines with these changes without taking any additional steps, unless for some reason they rely on `pipelinerun.status.taskruns`.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -6607,9 +5860,9 @@ func TestReconcilePipeline_FinalTasks(t *testing.T) {
 					t.Errorf("Expected PipelineRun status to be failed, but was %v for %s",
 						reconciledRun.Status.GetCondition(apis.ConditionSucceeded), tt.name)
 				}
-				if d := cmp.Diff(reconciledRun.Status.TaskRuns, tt.expectedTaskRuns); d != "" {
-					t.Fatalf("Expected PipelineRunTaskRun status to match TaskRun(s) status, but got a mismatch for %s: %s", tt.name, d)
-				}
+				// if d := cmp.Diff(reconciledRun.Status.TaskRuns, tt.expectedTaskRuns); d != "" {
```
JeromeJu, line 5863: Wondering if it is just safe here to remove the expected TaskRun, not sure if any checks needs to be added functionailty-wise?

lbernick: you could use the tekton client to query the status of the taskruns named in the pipelinerun.status.childreferences and assert that their statuses are as expected.

JeromeJu:
For the unit test [`TestReconcilePipeline_FinalTasks`](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_test.go#L6209), I tried to use the queried taskruns in pipelinerun.status.childreferences to assert but found when we deleted the [updateTaskRunsStatusDirectly](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L849), the Conditions will not be updated, which corresponds to the TEP100, since it has been deprecated:
```
This struct, and ChildStatusReferences.ConditionChecks, will be removed once Conditions, which have been deprecated, are removed completely. We are not using child references for the conditions' statuses, because ConditionCheckStatus, the only thing in PipelineRunConditionCheckStatus other than the ConditionName, isn't replicated anywhere else, and contains a fairly minimal amount of data - the pod name, start and completion times, and a corev1.ContainerState. See [the issue for deprecating Conditions](https://github.com/tektoncd/community/blob/main/teps/issue-3377) for more information on the planned removal of Conditions.
```

So I would like to confirm that this is the case and I can just remove the Condition check for `taskruns` status in the first 3 test cases of [`TestReconcilePipeline_FinalTasks`](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_test.go#L6209)?

lbernick:
> For the unit test [`TestReconcilePipeline_FinalTasks`](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_test.go#L6209), I tried to use the queried taskruns in pipelinerun.status.childreferences to assert but found when we deleted the [updateTaskRunsStatusDirectly](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L849), the Conditions will not be updated, which corresponds to the TEP100, since it has been deprecated:
> 
> ```
> This struct, and ChildStatusReferences.ConditionChecks, will be removed once Conditions, which have been deprecated, are removed completely. We are not using child references for the conditions' statuses, because ConditionCheckStatus, the only thing in PipelineRunConditionCheckStatus other than the ConditionName, isn't replicated anywhere else, and contains a fairly minimal amount of data - the pod name, start and completion times, and a corev1.ContainerState. See [the issue for deprecating Conditions](https://github.com/tektoncd/community/blob/main/teps/issue-3377) for more information on the planned removal of Conditions.
> ```
> 
> So I would like to confirm that this is the case and I can just remove the Condition check for `taskruns` status in the first 3 test cases of [`TestReconcilePipeline_FinalTasks`](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_test.go#L6209)?

Are you sure that the conditions here are the CRD that we removed, rather than the conditions in the taskrun status? There's a bit of name overloading which can be confusing :)

You're correct that the childReferences don't have the status of the child TaskRuns, which is needed in the test. What you can do is get the names of the taskruns from the childReferences, and use a fake tekton client to get these taskruns via a fake api call, asserting that the taskruns returned have the correct status. Here's [an example](https://github.com/tektoncd/pipeline/blob/66cdefa9ce7485a0c39dd9db83d88f2c4f31cba0/pkg/reconciler/pipelinerun/pipelinerun_test.go#L7247) of using the fake tekton client to "fetch" a pipelinerun.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -1610,18 +1533,22 @@ status:
...
-			Status: duckv1.Status{
-				Conditions: []apis.Condition{{Type: apis.ConditionSucceeded}},
-			},
-		},
+	if d := cmp.Diff(reconciledRun.Status.ChildReferences, expectedChildReferences); d != "" {
+		t.Fatalf("Expected PipelineRun status to match ChildReference(s) status, but got a mismatch %s", diff.PrintWantGot(d))
+	}
+
+	// Check the status of taskruns from ChildReferences to be Succeeded
+	trFromChildRef, err := clients.Pipeline.TektonV1beta1().TaskRuns("foo").Get(prt.TestAssets.Ctx, reconciledRun.Status.ChildReferences[0].Name, metav1.GetOptions{})
```
JeromeJu, line 1541: This is for comparing with the FinalTask testcase.

JeromeJu:
> > For the unit test [`TestReconcilePipeline_FinalTasks`](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_test.go#L6209), I tried to use the queried taskruns in pipelinerun.status.childreferences to assert but found when we deleted the [updateTaskRunsStatusDirectly](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L849), the Conditions will not be updated, which corresponds to the TEP100, since it has been deprecated:
> > ```
> > This struct, and ChildStatusReferences.ConditionChecks, will be removed once Conditions, which have been deprecated, are removed completely. We are not using child references for the conditions' statuses, because ConditionCheckStatus, the only thing in PipelineRunConditionCheckStatus other than the ConditionName, isn't replicated anywhere else, and contains a fairly minimal amount of data - the pod name, start and completion times, and a corev1.ContainerState. See [the issue for deprecating Conditions](https://github.com/tektoncd/community/blob/main/teps/issue-3377) for more information on the planned removal of Conditions.
> > ```
> > 

> > So I would like to confirm that this is the case and I can just remove the Condition check for `taskruns` status in the first 3 test cases of [`TestReconcilePipeline_FinalTasks`](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_test.go#L6209)?
> 
> Are you sure that the conditions here are the CRD that we removed, rather than the conditions in the taskrun status? There's a bit of name overloading which can be confusing :)
> 
> You're correct that the childReferences don't have the status of the child TaskRuns, which is needed in the test. What you can do is get the names of the taskruns from the childReferences, and use a fake tekton client to get these taskruns via a fake api call, asserting that the taskruns returned have the correct status. Here's [an example](https://github.com/tektoncd/pipeline/blob/66cdefa9ce7485a0c39dd9db83d88f2c4f31cba0/pkg/reconciler/pipelinerun/pipelinerun_test.go#L7247) of using the fake tekton client to "fetch" a pipelinerun.

Thanks @lbernick , I think it is able to compare with the taskRun status referenced as in ChildReference's Name: one changed I made was at https://github.com/tektoncd/pipeline/pull/6049#discussion_r1092064877 if that looks correct as you pointed.

But for the [TestReconcilePipeline_FinalTasks](https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_test.go#L6209), it felt different since when looking in the stacktrace, if it's with test case `pipelineRunStatusFalse`, it hasn't reached https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L230 where `ChildReferences` are updated, so we could not check the Status by ChildReference?

JeromeJu:
PTAL 👼
/assign @abayer @jerop @lbernick

jerop (commented):
@JeromeJu I think this change should also remove `TaskRuns` and `Runs` in `PipelineRun.Status`, but it's not clear if that's the case 🤔

pkg/apis/pipeline/v1beta1/pipelinerun_conversion_test.go
```diff
@@ -570,14 +507,13 @@ func TestPipelineRunConversionEmbeddedStatusRoundTrip(t *testing.T) {
 	}
 }
 
-func TestPipelineRunConversionEmbeddedStatusConvertTo(t *testing.T) {
+func TestPipelineRunConversionConvertTo(t *testing.T) {
```
lbernick, line 510: I think this test could probably be removed

pkg/reconciler/pipelinerun/cancel.go
```diff
@@ -192,43 +191,24 @@ func cancelPipelineTaskRunsForTaskNames(ctx context.Context, logger *zap.Sugared
 }
 
 // getChildObjectsFromPRStatusForTaskNames returns taskruns, customruns, and runs in the PipelineRunStatus's ChildReferences or TaskRuns/Runs,
```
lbernick, line 193:
```suggestion
// getChildObjectsFromPRStatusForTaskNames returns taskruns, customruns, and runs in the PipelineRunStatus's ChildReferences,
```

JeromeJu:
> 

Thanks @jerop.
Sorry I thought we were only removing the feature flag in this release(v0.45). 
Are we targeting to also remove the deprecated fields `status.taskruns` and `status.runs` within the same release? I will open up a separate PR for doing so right now.

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -38,196 +38,15 @@ import (
...
-		customRuns     []*v1beta1.CustomRun
-		wantErr        bool
+		name string
+
+		pipelineRun *v1beta1.PipelineRun
+		taskRuns    []*v1beta1.TaskRun
+		runs        []*v1alpha1.Run
+		customRuns  []*v1beta1.CustomRun
+		wantErr     bool
 	}{{
```
lbernick, line 48: I don't think all of these test cases can be deleted; we need to make sure all these cases are still covered. For example, I don't think there's currently a test case for "no resolved taskrun", since "no-resolved-taskrun-with-full" was deleted. Can you take a closer look at these to figure out which are redundant without an embedded-status flag, and which need to be updated to use childreferences?

JeromeJu:
It would be appreciated if we could add this to the [v0.45](https://github.com/tektoncd/pipeline/milestone/64)?

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -458,98 +272,7 @@ func TestGetChildObjectsFromPRStatusForTaskNames(t *testing.T) {
...
-			}},
-			expectedTRNames:  nil,
-			expectedRunNames: []string{"r1"},
-			hasError:         false,
-		}, {
-			name:           "default minimal embedded",
-			embeddedStatus: config.MinimalEmbeddedStatus,
+			name: "default",
 			prStatus: v1beta1.PipelineRunStatus{PipelineRunStatusFields: v1beta1.PipelineRunStatusFields{
 				TaskRuns: map[string]*v1beta1.PipelineRunTaskRunStatus{
```
lbernick, line 277: No tests should have pipelineruns with taskruns in the status anymore. This test name should also be updated to reflect what it's testing

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1369,26 +1318,13 @@ func (c *Reconciler) updatePipelineRunStatusFromInformer(ctx context.Context, pr
...
-		updatePipelineRunStatusFromChildRefs(logger, pr, taskRuns, runObjects)
-	default:
-		// Clear any TaskRuns and Runs present in the status.
-		// This can occur if the value of "embedded-status" was modified during PipelineRun execution.
-		pr.Status.Runs = nil
-		pr.Status.TaskRuns = nil
-		updatePipelineRunStatusFromChildRefs(logger, pr, taskRuns, runObjects)
-	}
+	// Clear any TaskRuns and Runs present in the status.
+	// This can occur if the value of "embedded-status" was modified during PipelineRun execution prior to its removal.
```
lbernick, line 1322: Is this still necessary without an embedded-status flag?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -203,40 +202,10 @@ func getPipelineRunUpdates(t *testing.T, actions []ktesting.Action) []*v1beta1.P
...
-	}
-}
-
-// runTestReconcileWithEmbeddedStatus runs "Reconcile" on a PipelineRun with one
+// runTestReconcile runs "Reconcile" on a PipelineRun with one
 // Task that has not been started yet.  It verifies that the TaskRun is created,
 // it checks the resulting API actions, status and events.
-func runTestReconcileWithEmbeddedStatus(t *testing.T, embeddedStatus string) {
+func TestReconcile(t *testing.T) {
 	t.Helper()
```
lbernick, line 209: t.Helper should be removed, here and in all the other test classes that are no longer helpers.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -4943,13 +4661,13 @@ spec:
...
 		CustomRuns:   crs,
-		ConfigMaps:   []*corev1.ConfigMap{withEmbeddedStatus(newFeatureFlagsConfigMap(), embeddedStatus)},
+		ConfigMaps:   []*corev1.ConfigMap{newFeatureFlagsConfigMap()},
 	}
 	prt := newPipelineRunTest(t, d)
 	defer prt.Cancel()
 	reconciledRun, _ := prt.reconcileRun("foo", "test-pipeline-run-finally-results", []string{}, false)
 
-	expectedPrFullStatus := parse.MustParseV1beta1PipelineRun(t, `
+	expectedPrMinimalStatus := parse.MustParseV1beta1PipelineRun(t, `
```
lbernick, line 4670:
```suggestion
	expectedPrStatus := parse.MustParseV1beta1PipelineRun(t, `
```

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -6607,27 +5918,55 @@ func TestReconcilePipeline_FinalTasks(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("Failure to get TaskRun from ChildReference %s: %s", childRef.Name, err)
+		}
+
+		if expectedTaskRunStatus, ok := expectedTaskRuns[trName]; !ok {
+			t.Fatalf("Expected not to have TaskRun status from ChildReferences %s", trName)
+		} else if d := cmp.Diff(*expectedTaskRunStatus.Status, trFromChildRef.Status); d != "" {
+			t.Fatalf("Expected TaskRun Status to match ChildReferences TaskRun Status, but got a mismatch %s", diff.PrintWantGot(d))
+		}
+		delete(expectedTaskRuns, trName)
```
lbernick, line 5962: Is this necessary? I prefer to avoid modifying inputs unless the function is designed to do so

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -10963,39 +10126,30 @@ spec:
...
-				if len(pr.Status.ChildReferences) > 1 {
-					t.Errorf("Expected one CustomRun or TaskRun in child references, but found %d", len(pr.Status.ChildReferences))
-				}
-			}
+	if len(pr.Status.Runs) > 1 {
+		t.Errorf("Expected one CustomRun in status, but found %d", len(pr.Status.Runs))
+	}
+	if len(pr.Status.TaskRuns) > 0 {
+		t.Errorf("Expected no TaskRuns in status, but found %d", len(pr.Status.TaskRuns))
+	}
```
lbernick, line 10148: This doesn't look right, shouldn't we be checking childreferences?

pkg/reconciler/pipelinerun/timeout_test.go
```diff
@@ -33,135 +33,14 @@ import (
...
 	testCases := []struct {
 		name                  string
-		embeddedStatus        string
 		useV1Beta1CustomTasks bool
 		pipelineRun           *v1beta1.PipelineRun
 		taskRuns              []*v1beta1.TaskRun
 		customRuns            []*v1beta1.CustomRun
 		runs                  []*v1alpha1.Run
 		wantErr               bool
 	}{{
```
lbernick, line 42: Same comment as for cancellation tests, I'm not sure all of these test cases can be safely removed

test/conversion_test.go
```diff
@@ -1013,14 +986,6 @@ func TestPipelineRunCRDConversion(t *testing.T) {
...
-	featureFlags, _ := config.NewFeatureFlagsFromMap(map[string]string{
-		"embedded-status": config.FullEmbeddedStatus,
-	})
-	cfg := &config.Config{FeatureFlags: featureFlags}
-	return config.ToContext(context.Background(), cfg)
-}
-
 // validatePipelineRunTaskRunStatusAnnotations validates the `status.taskRun` in Annotations by converting
 // the json string into a struct and compare the valid status fields
 func validatePipelineRunTaskRunStatusAnnotations(pipelineRunName string, annotations map[string]string) error {
```
lbernick, line 991: Is this still needed?

JeromeJu: Thanks for spotting this. I hesitated to remove this since there could be still users using `taskruns` / `runs` for the time being? If so I could remove this in the PR that does the removal?

pkg/reconciler/pipelinerun/cancel_test.go
JeromeJu: Thanks, updated.

pkg/reconciler/pipelinerun/pipelinerun.go
JeromeJu: I was trying to be safe earlier to not to delete the guards before removing the `status.taskruns` or `status.runs` from the apis and also to create the cleanup issue as TODO. I am happy to clean them up if we are doing so in this release.

JeromeJu:
I might have misunderstanding on the goal of [v0.45 release for the implementations for TEP0100: ](https://github.com/tektoncd/pipeline/issues/4954): 

- Are we also going to remove `status.taskruns` or `status.runs` within this release v0.45?
- If so, will it be better practice to put it in another PR?

pkg/reconciler/pipelinerun/cancel_test.go
JeromeJu: restored, sorry that I thought the ones were relied on taskruns status.

pkg/reconciler/pipelinerun/pipelinerun_test.go
JeromeJu: Thanks for the suggestion, I delete it so as to makes sure that all expected status are compared.

I changed to using a var for counting if that is better.

pkg/reconciler/pipelinerun/pipelinerun.go
JeromeJu: I think this could still be valuable before we remove the `status.runs` `status.taskruns` corresponding to the test case we might want at https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun_updatestatus_test.go#L1028-L1058?

pkg/reconciler/pipelinerun/pipelinerun_test.go
JeromeJu: Thanks for catching:)

JeromeJu:
I also noticed https://github.com/tektoncd/pipeline/pull/4850 for changes in `pkg/status`, I think it might be reasonable to remove those in the following up PR that removes the taskruns and runs fields?

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -10963,39 +10110,39 @@ spec:
...
-			}
+	for _, childRef := range pr.Status.ChildReferences {
+		var runsCount, taskrunsCount int
+		switch childRef.Kind {
+		case "TaskRun":
+			taskrunsCount++
+		case "Run":
+			runsCount++
+		}
+		if runsCount > 1 {
```
JeromeJu, line 10135: TODO: figure out if this is the intended cases or I shall change to case "CustomRun"

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -38,196 +38,131 @@ import (
...
-		taskRuns: []*v1beta1.TaskRun{
-			{ObjectMeta: metav1.ObjectMeta{Name: "t1"}},
-		},
-		runs: []*v1alpha1.Run{
-			{ObjectMeta: metav1.ObjectMeta{Name: "r2"}},
-		},
-	}, {
-		name:           "child-references-with-minimal",
-		embeddedStatus: config.DefaultEmbeddedStatus,
+		name: "child-references",
```
lbernick, line 165: can you rename the test cases with "child-references" in the name to something more descriptive? All of the test cases now have child references, so putting it in the test name isn't very informative.

pkg/reconciler/pipelinerun/cancel_test.go
```diff
@@ -458,102 +388,8 @@ func TestGetChildObjectsFromPRStatusForTaskNames(t *testing.T) {
...
-					PipelineTaskName: "run-1",
-				}},
-			}},
-			expectedTRNames:  []string{"t1"},
-			expectedRunNames: nil,
-			hasError:         false,
-		}, {
-			name:           "both embedded",
-			embeddedStatus: config.BothEmbeddedStatus,
+			name: "default",
```
lbernick, line 391: can you rename this test to something more descriptive as well?

pkg/reconciler/pipelinerun/pipelinerun_updatestatus_test.go
```diff
@@ -935,99 +926,71 @@ pipelineTaskName: task-6
...
+			actualTaskRuns := actualPrStatus.PipelineRunStatusFields.TaskRuns
+			if actualTaskRuns != nil {
+				fixedTaskRuns := make(map[string]*v1beta1.PipelineRunTaskRunStatus)
+				re := regexp.MustCompile(`^[a-z\-]*?-task-[0-9]`)
+				for k, v := range actualTaskRuns {
+					newK := re.FindString(k)
+					fixedTaskRuns[newK+"-xxyyy"] = v
 				}
+				actualPrStatus.PipelineRunStatusFields.TaskRuns = fixedTaskRuns
+			}
```
lbernick, line 958: Can this be removed?

pkg/reconciler/pipelinerun/pipelinerun.go
lbernick: I think we should remove or rewrite test cases with taskruns/runs

pkg/reconciler/pipelinerun/timeout_test.go
```diff
@@ -98,70 +112,28 @@ func TestTimeoutPipelineRun(t *testing.T) {
...
-			{ObjectMeta: metav1.ObjectMeta{Name: "t2"}},
-		},
-		runs: []*v1alpha1.Run{
-			{ObjectMeta: metav1.ObjectMeta{Name: "r1"}},
-			{ObjectMeta: metav1.ObjectMeta{Name: "r2"}},
-		},
-	}, {
-		name:           "child-references-with-minimal",
-		embeddedStatus: config.MinimalEmbeddedStatus,
+		name: "child-references",
```
lbernick, line 136: same comment here about descriptive test names

pkg/status/status.go
```diff
@@ -77,9 +77,9 @@ func GetRunStatusForPipelineTask(ctx context.Context, client versioned.Interface
 	return runStatus, nil
 }
 
-// GetFullPipelineTaskStatuses returns populated TaskRun and Run status maps for a PipelineRun from its ChildReferences.
+// GetPipelineTaskStatuses returns populated TaskRun and Run status maps for a PipelineRun from its ChildReferences.
 // If the PipelineRun has no ChildReferences, its .Status.TaskRuns and .Status.Runs will be returned instead.
-func GetFullPipelineTaskStatuses(ctx context.Context, client versioned.Interface, ns string, pr *v1beta1.PipelineRun) (map[string]*v1beta1.PipelineRunTaskRunStatus,
+func GetPipelineTaskStatuses(ctx context.Context, client versioned.Interface, ns string, pr *v1beta1.PipelineRun) (map[string]*v1beta1.PipelineRunTaskRunStatus,
```
lbernick, line 82: @abayer is this meant to be a helper function for package importers outside tektoncd/pipeline? It seems like dead code.

@JeromeJu can you please update this function and its docstring to not use taskruns/runs?

pkg/status/status_test.go
```diff
@@ -295,7 +295,7 @@ status:
 			expectedErr:         nil,
 		},
 		{
-			name: "minimal embedded",
+			name: "default",
```
lbernick, line 298: can you rename this to something more descriptive?

pkg/status/status.go
JeromeJu: I was not sure if this is used for populating taskruns/runs as `full` embbededStatus alike pipelineRunStatus to prepare for the removal of this feature flag? Was also wondering about this at
https://github.com/tektoncd/pipeline/issues/6090#issuecomment-1412730427

pkg/status/status_test.go
JeromeJu: Thanks, renamed.

pkg/reconciler/pipelinerun/pipelinerun_updatestatus_test.go
JeromeJu: thanks, removed!

pkg/reconciler/pipelinerun/pipelinerun_test.go
JeromeJu: Thanks for checking @XinruZhang

pkg/status/status.go
JeromeJu: Updated and removed the populating from taskruns/runs. Thanks for the pointers

lbernick (approved):

lbernick:
@JeromeJu just edited the release notes since we don't really have a way for people to migrate off of the feature flag; they just can't use it anymore

JeromeJu:
> @JeromeJu just edited the release notes since we don't really have a way for people to migrate off of the feature flag; they just can't use it anymore

Ohhh thanks for correcting that:)

chitrangpatel (approved):
Looks good! Just wanted to ask what happens if users provide `embedded-status` flag. Do they get an error or does it just silently ignore it?

JeromeJu:
> Looks good! Just wanted to ask what happens if users provide `embedded-status` flag. Do they get an error or does it just silently ignore it?

I don't think they will get an error as it shall be just an unknown feature flag in config/config-feature-flag

chitrangpatel:
> > Looks good! Just wanted to ask what happens if users provide `embedded-status` flag. Do they get an error or does it just silently ignore it?
> 
> I don't think they will get an error as it shall be just an unknown feature flag in config/config-feature-flag

I think thats ok.

jerop (commented):
@JeromeJu please also need to remove `embedded-status` feature flag from dogfooding - https://github.com/tektoncd/plumbing/blob/main/tekton/cd/pipeline/overlays/dogfooding/feature-flags.yaml#L8

thank you!

/lgtm

JeromeJu:
/retest
to retrigger ci for build test

JeromeJu:
Thanks Jerop, raised https://github.com/tektoncd/plumbing/pull/1339 for followup at plumbing repo.

Issue #6023, opened 2023-01-24
taskrunvalidationfailed - array results - results initialized with ambiguous failure message


pritidesai:
# Expected Behavior

As a user, I expect the type of a result in the `taskRun` status matches with the type specified in the task definition. I do not expect to have the result set to `string` type by default. 

# Actual Behavior

I see two issues here when a result is defined as an array but it is initialized using a string value:

Issue 1: The taskRun results section lists the result of type string whereas it was originally defined as array in the task definition.

Issue 2: The error message in the status section of a taskRun is ambiguous. The error message needs to be a little more verbose and explicit. Also, the error message needs to list down the result names and their specified types. 

# Steps to Reproduce the Problem

1. Deploy Tekton Pipelines with the latest commit
2. Apply task, pipeline, and pipelineRun to the cluster.

```
$ kubectl create -f pipelinerun.yaml
pipeline.tekton.dev/mismatch-type-in-result created
task.tekton.dev/mismatch-type-task created
pipelinerun.tekton.dev/mismatch-type-in-result-pipelinerun-jq8mh created
```

```
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: mismatch-type-in-result
spec:
  tasks:
    - name: mismatch-type
      taskRef:
        name: mismatch-type-task
---

apiVersion: tekton.dev/v1beta1
kind: Task
metadata:
  name: mismatch-type-task
spec:
  results:
    - name: result1
      type: array
  steps:
    - name: write-invalid-result
      image: alpine:latest
      script: |
        #!/bin/sh
        echo -n "this is a string for a result with array type" | tee $(results.result1.path);
---

apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  generateName: mismatch-type-in-result-pipelinerun-
spec:
  pipelineRef:
    name: mismatch-type-in-result
```

3. Check the taskRun status:

```
$ kubectl get tr
NAME                                                      SUCCEEDED   REASON                    STARTTIME   COMPLETIONTIME
mismatch-type-in-result-pipelinerun-jq8mh-mismatch-type   False       TaskRunValidationFailed   8s          3s
```

![Screen Shot 2023-01-23 at 6 44 18 PM](https://user-images.githubusercontent.com/206285/214203071-30c8866b-56f7-4187-89eb-987d404dd7f2.png)

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

```
[...]

PR #6056, opened 2023-01-26
Add support for array task results as matrix parameters

### LINKED ISSUE #5925: Support `Array` and `Strings` for Matrix param validation
Hey,

Currently validation for the `matrix` alpha feature is partly done in this code:

```go
func validateParametersInTaskMatrix(matrix *Matrix) (errs *apis.FieldError) {
	if matrix != nil {
		for _, param := range matrix.Params {
			if param.Value.Type != ParamTypeArray {
				errs = errs.Also(apis.ErrInvalidValue("parameters of type array only are allowed in matrix", "").ViaFieldKey("matrix", param.Name))
			}
		}
	}
	return errs
}
```

Unfortunately this doesn't yet support passing in Tekton `Arrays` as kubernetes type `String` as described here in [TEP-0090](https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#specifying-results-in-the-matrix)

For example the following is not currently supported

```yaml
tasks:
...
- name: task-5
  taskRef:
    name: task-5
  matrix:
  - name: values
    value: (tasks.task-4.results.foo) # array
```

Err:
```
for: "pipeline/pipeline-test.yaml": error when patching "pipeline/pipeline-test.yaml": admission webhook "validation.webhook.pipeline.tekton.dev" denied the request: validation failed: invalid value: parameters of type array only are allowed in matrix: spec.tasks[1].matrix[k8s-versions], spec.tasks[1].matrix[machine-versions]
```

Is that something that is being worked on?

abayer:
# Changes

fixes #5925

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Array task results can now be used as the value for matrix parameters.
```

### DIFF
--- a/examples/v1beta1/pipelineruns/alpha/pipelinerun-with-matrix-and-results.yaml
+++ b/examples/v1beta1/pipelineruns/alpha/pipelinerun-with-matrix-and-results.yaml
@@ -39,16 +39,13 @@ spec:
       - name: get-browsers
         taskSpec:
           results:
-            - name: one
-            - name: two
-            - name: three
+            - name: browsers
           steps:
             - name: echo
               image: alpine
               script: |
-                printf chrome | tee /tekton/results/one
-                printf safari | tee /tekton/results/two
-                printf firefox | tee /tekton/results/three
+                #!/usr/bin/env bash
+                echo -n "[\"chrome\",\"safari\",\"firefox\"]" | tee $(results.browsers.path)
       - name: platforms-and-browsers-dag
         matrix:
           params:
@@ -58,8 +55,6 @@ spec:
                 - $(tasks.get-platforms.results.two)
                 - $(tasks.get-platforms.results.three)
             - name: browser
-              value:
-                - $(tasks.get-browsers.results.one)
-                - $(tasks.get-browsers.results.two)
+              value: $(tasks.get-browsers.results.browsers[*])
         taskRef:
           name: platform-browsers
--- a/pkg/reconciler/pipelinerun/resources/apply.go
+++ b/pkg/reconciler/pipelinerun/resources/apply.go
@@ -178,7 +178,7 @@ func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResul
 			pipelineTask := resolvedPipelineRunTask.PipelineTask.DeepCopy()
 			pipelineTask.Params = replaceParamValues(pipelineTask.Params, stringReplacements, arrayReplacements, objectReplacements)
 			if pipelineTask.IsMatrixed() {
-				pipelineTask.Matrix.Params = replaceParamValues(pipelineTask.Matrix.Params, stringReplacements, nil, nil)
+				pipelineTask.Matrix.Params = replaceParamValues(pipelineTask.Matrix.Params, stringReplacements, arrayReplacements, nil)
 			}
 			pipelineTask.WhenExpressions = pipelineTask.WhenExpressions.ReplaceWhenExpressionsVariables(stringReplacements, arrayReplacements)
 			if pipelineTask.TaskRef != nil && pipelineTask.TaskRef.Params != nil {
--- a/pkg/reconciler/pipelinerun/resources/apply_test.go
+++ b/pkg/reconciler/pipelinerun/resources/apply_test.go
@@ -2331,6 +2331,38 @@ func TestApplyTaskResults_MinimalExpression(t *testing.T) {
 			},
 		}},
 	}, {
+		name: "Test array result substitution on minimal variable substitution expression - matrix",
+		resolvedResultRefs: ResolvedResultRefs{{
+			Value: *v1beta1.NewStructuredValues("arrayResultValueOne", "arrayResultValueTwo"),
+			ResultReference: v1beta1.ResultRef{
+				PipelineTask: "aTask",
+				Result:       "aResult",
+			},
+			FromTaskRun: "aTaskRun",
+		}},
+		targets: PipelineRunState{{
+			PipelineTask: &v1beta1.PipelineTask{
+				Name:    "bTask",
+				TaskRef: &v1beta1.TaskRef{Name: "bTask"},
+				Matrix: &v1beta1.Matrix{
+					Params: []v1beta1.Param{{
+						Name:  "bParam",
+						Value: *v1beta1.NewStructuredValues(`$(tasks.aTask.results.aResult[*])`),
+					}}},
+			},
+		}},
+		want: PipelineRunState{{
+			PipelineTask: &v1beta1.PipelineTask{
+				Name:    "bTask",
+				TaskRef: &v1beta1.TaskRef{Name: "bTask"},
+				Matrix: &v1beta1.Matrix{
+					Params: []v1beta1.Param{{
+						Name:  "bParam",
+						Value: *v1beta1.NewStructuredValues("arrayResultValueOne", "arrayResultValueTwo"),
+					}}},
+			},
+		}},
+	}, {
 		name: "Test array result substitution on minimal variable substitution expression - when expressions",
 		resolvedResultRefs: ResolvedResultRefs{{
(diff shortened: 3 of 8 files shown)


abayer:
/hold
Until I add an example test. =)

pkg/reconciler/pipelinerun/resources/apply.go
```diff
@@ -178,7 +178,7 @@ func ApplyTaskResults(targets PipelineRunState, resolvedResultRefs ResolvedResul
 			pipelineTask := resolvedPipelineRunTask.PipelineTask.DeepCopy()
 			pipelineTask.Params = replaceParamValues(pipelineTask.Params, stringReplacements, arrayReplacements, objectReplacements)
 			if pipelineTask.IsMatrixed() {
-				pipelineTask.Matrix.Params = replaceParamValues(pipelineTask.Matrix.Params, stringReplacements, nil, nil)
+				pipelineTask.Matrix.Params = replaceParamValues(pipelineTask.Matrix.Params, stringReplacements, arrayReplacements, nil)
```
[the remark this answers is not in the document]
abayer: I'm not sure if objects actually make sense in a matrix params context - thoughts, @jerop?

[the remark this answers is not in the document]
jerop: I don't think we should handle object results - unless we're speaking of getting a particular value from the object as an item in the matrix array, but even then I'd like to leave that out for when we have use cases and explore it in detail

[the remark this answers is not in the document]
afrittoli: Thank you, never mind, it was a silly comment. 
We might support objects in future there, to make it easier for a task to trigger a specific set of combinations of parameters through results, but that would be a new feature.

abayer:
/hold cancel

@afrittoli I agree - #6061 may also make sense to backport.

jerop (commented):
thank you @abayer! 🙏🏾 

only request is to add some details to the commit message, specifically that array results was supported after matrix was added and add a link to the section of TEP-0090 discussing this feature - https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#specifying-results-in-the-matrix

jerop:
@abayer it seems some changes are needed in the validation: 
`"validation.webhook.pipeline.tekton.dev" denied the request: validation failed: invalid value: parameters of type array only are allowed in matrix: spec.pipelineSpec.tasks[2].matrix[browser]`

pkg/reconciler/pipelinerun/resources/apply_test.go
```diff
@@ -2330,6 +2330,38 @@ func TestApplyTaskResults_MinimalExpression(t *testing.T) {
 					}}},
 			},
 		}},
+	}, {
+		name: "Test array result substitution on minimal variable substitution expression - matrix",
```
lbernick, line 2334: can you also add a test case for array result indexing in a matrix?

abayer: That already exists.

lbernick: my bad!

abayer:
> it seems some changes are needed in the validation: `"validation.webhook.pipeline.tekton.dev" denied the request: validation failed: invalid value: parameters of type array only are allowed in matrix: spec.pipelineSpec.tasks[2].matrix[browser]`

Oh my, yes, this is a rathole. =) So the problem is that we're checking what the raw parsed type is - which will be a string, not an array, since the value is `$(tasks.get-browsers.results.browsers[*])`. If we turn off the validation check in `validateParametersInTaskMatrix`, things do just work (or rather, I thought they did, but in my local test, the controller is panicking in `ValidateResolvedTaskResources`, so there might be something more busted). Aaaanyway - the validation problem is that we're requiring the param value to be an array, but we don't actually know that the initially-parsed string `$(tasks.get-browsers.results.browsers[*])` will end up being an array. And now it also appears that there are other issues with resolved task resources, but I haven't dug into _that_ yet.

abayer:
Ok, I have a fix for the validation that I'm fairly happy with, but the next problem is that by the time `GetNamesOfTaskRuns(pipelineRun.Status.ChildReferences, pipelineTask.Name, pipelineRun.Name, pipelineTask.GetMatrixCombinationsCount())` gets called within `ResolvePipelineTask`, the array results parameter is still a string, meaning `GetMatrixCombinationsCount()` ends up returning 0 (because the "string" param has `len(param.Value.ArrayVal) == 0`). Digging further.

abayer:
I've updated the PR with my fix for the validation issue, and with a unit test tweak to use an array result as a matrix parameter, to make it easier to dig into the problem.

afrittoli:
Thanks @abayer for the detailed investigation on this.

So it sounds like this:
- this is not something we should backport to v0.44 as it will be a non-trivial change
- we could perhaps detect and have a clearer error message for users that attempt to pass results into parameters, and backport that fix instead

wdyt?

abayer:
@afrittoli Yeah, that sounds good. I’ll try to get a PR up tomorrow for the better error message.

abayer:
/hold

Putting this on hold until we most likely just close it. =)

From a chat this morning with @jerop, I feel like this is a reasonable path forward:
* First, add a check in `Pipeline` validation to verify that matrix parameters are not just arrays, but that they are not _empty_ arrays. This won't change existing functionality/capabilities, but will get rid of the panic in the reconciler. I created #6071 for this.
* Next, tackle adding support for using `PipelineRun` parameters as matrix parameter values. This will require making changes to the validation to allow `$(params.foo[*])`, rather than an explicit YAML array, to be used as a matrix parameter, and will require the `PipelineRun` reconciler to handle replacing that params reference with the actual value, and better error handling around the possibility of that parameter's value being an empty array.
* Finally, go after using array results as matrix parameter values. Everything relevant to `$(params.foo[*])` as matrix parameter value applies here too, but there's the additional problem that, at the time that the `PipelineRun` reconciler is generating the list of matrixed `PipelineTask`s it will eventually execute, we don't yet know what `$(tasks.someTask.results.anArrayResult[*])` actually will contain, so we can't calculate the matrix combination count, and therefore can't create the `PipelineTask`s for each matrix combination. I feel like this is going to require some fairly deep changes to how matrix is handled in `PipelineRun` reconciliation, so yeah, this'll be A Bunch Of Work. =)

pritidesai:
@abayer @jerop Issue #6023 explains what is going on with array results when the task specification does not include square brackets which is the core identification piece in general but [this](https://github.com/tektoncd/pipeline/issues/6023#issuecomment-1409300278) might be related here.

examples/v1beta1/pipelineruns/alpha/pipelinerun-with-matrix-and-results.yaml
```diff
@@ -39,16 +39,13 @@ spec:
       - name: get-browsers
         taskSpec:
           results:
-            - name: one
-            - name: two
-            - name: three
+            - name: browsers
```
pritidesai, line 42: this will need an `array` type, right? Default type of a param is `string`.

pritidesai:
Pipelines WG - @abayer do you think we want to keep this in this milestone or move it the next milestone?

afrittoli:
@abayer do you think there is still a chance for this to happen in the next milestone or should we move it to the following one?

abayer:
Moved to next milestone - I need to think about this a bit more, but I'm leaning towards likely closing it in favor of what's described in https://github.com/tektoncd/pipeline/pull/6056#issuecomment-1408936114.

pritidesai:
@EmmaMunley @jerop is this PR still applicable and needed after all the efforts from @EmmaMunley? Shall we keep this open? 

This did not make it to 0.46, moving it to 0.47.

jerop:
yes, it's still applicable, let's leave it open for now

lbernick:
@abayer my understanding is that @EmmaMunley is now working on this feature in a separate PR, do you mind if I close this?

PR #6063, opened 2023-01-27
increase the timer for reading k8s event in test

Yongxuanzhang:
# Changes

This commits extends the current timer from 10 millisecond to wait.ForeverTestTimeout and tried to fix flake test #5748. The 10 ms timer may be too short to receive events from channel in tests.

/kind flake

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/events/k8sevent/events.go
+++ b/pkg/reconciler/events/k8sevent/events.go
@@ -20,14 +20,14 @@ import (
 	"regexp"
 	"testing"
 	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 // CheckEventsOrdered checks that the events received via the given chan are the same as wantEvents,
 // in the same order.
 func CheckEventsOrdered(t *testing.T, eventChan chan string, testName string, wantEvents []string) error {
 	t.Helper()
-	// Sleep 50ms to make sure events have delivered
-	time.Sleep(50 * time.Millisecond)
 	err := eventsFromChannel(eventChan, wantEvents)
 	if err != nil {
 		return fmt.Errorf("error in test %s: %v", testName, err)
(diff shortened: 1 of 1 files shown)


pkg/reconciler/events/k8sevent/events.go
```diff
@@ -65,12 +62,16 @@ func eventsFromChannel(c chan string, wantEvents []string) error {
...
 			}
-		case <-timer.C:
-			if len(foundEvents) != len(wantEvents) {
-				return fmt.Errorf("received %d events but %d expected. Found events: %#v", len(foundEvents), len(wantEvents), foundEvents)
-			}
+		case <-timer:
+			return fmt.Errorf("received %d events but %d expected. Found events: %#v", len(foundEvents), len(wantEvents), foundEvents)
+		}
+	}
+	for {
```
lbernick, line 69: Can you add a comment to this section?

lbernick (approved):

pkg/reconciler/events/k8sevent/events.go
Yongxuanzhang: sure!

pkg/reconciler/events/k8sevent/events.go
```diff
@@ -20,14 +20,14 @@ import (
 	"regexp"
 	"testing"
 	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 // CheckEventsOrdered checks that the events received via the given chan are the same as wantEvents,
 // in the same order.
 func CheckEventsOrdered(t *testing.T, eventChan chan string, testName string, wantEvents []string) error {
```
pritidesai, line 29: Wondering why this function is not part of `_test.go` when its written to be used for only **tests**? 🤔 This way it can be consumed outside of `tests`.  Shall we make the wait interval configurable such that the user is responsible for choosing the appropriate wait time?

pritidesai:
/lgtm

thanks @Yongxuanzhang I am happy to merge this as is since its addressing a flake but the wait time could have been made configurable 🙃

pkg/reconciler/events/k8sevent/events.go
Yongxuanzhang: > Wondering why this function is not part of `_test.go` when its written to be used for only **tests**? 🤔 This way it can be consumed outside of `tests`. Shall we make the wait interval configurable such that the user is responsible for choosing the appropriate wait time?

It's because the functions in `_test.go` cannot be used in other pkgs

Yongxuanzhang:
> /lgtm
> 
> thanks @Yongxuanzhang I am happy to merge this as is since its addressing a flake but the wait time could have been made configurable 🙃

yeah sure! I think that makes sense!

PR #6065, opened 2023-01-27
Clean up cloudevents package

lbernick:
This commit makes several changes:
- removes capitalization of error strings, following the go style guide (https://google.github.io/styleguide/go/decisions#error-strings)
- removes dead code: the functions `eventForTaskRun`, `eventForPipelineRun`, `eventForRun`, and `eventForCustomRun`
- moves tests into a separate package so only exported functions are tested
- avoids exporting FakeClient.events, since this exposes implementation details
and isn't required for implementing the cloudevents.Client interface

No expected functional changes.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -130,10 +107,33 @@ func (c *FakeClient) CheckCloudEventsUnordered(t *testing.T, testName string, wa
...
+	var gotEvents []string
+	timeout := time.After(wait.ForeverTestTimeout)
+	// Read len(wantEvents) events from the channel, blocking until they have all been received
+	// or the test reaches its (long) timeout
+	for i := 0; i < numExpectedEvents; i++ {
+		select {
+		case got := <-c:
+			gotEvents = append(gotEvents, got)
+		case <-timeout:
+			t.Errorf("Expected event, got nothing")
```
Yongxuanzhang, line 127: this may not be very accurate? Since we may receive some events?

abayer:
/approve

oooo, nice!

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
lbernick: good point, fixed!

Yongxuanzhang:
the coverage drop of [pkg/reconciler/events/cloudevent/cloudevent.go](https://storage.cloud.google.com/tekton-prow/pr-logs/pull/tektoncd_pipeline/6065/pull-tekton-pipeline-go-coverage-df/1619088271806042112/artifacts/line-cov.html#file1) should be the removal of unused functions

lbernick:
/hold
the FakeRecorder.events doesn't actually need to be a channel, going to refactor to just use a slice

afrittoli:
> This commit makes several changes:
> 
> * replaces the wrapper around the cloudevent client with improved code to read events from the channel, modeled on https://github.com/kubernetes/client-go/blob/cbb60162c0d5ef255261d9ee5fe8b5f829c95baa/tools/leaderelection/leaderelection_test.go#L1240-L1262

Thanks @lbernick - about this first part, I'm sure you're aware, this removes a lot of the changes introduced in https://github.com/tektoncd/pipeline/commit/fff3fcda2774f07d6787bbd0496c58610d381f9d by @Yongxuanzhang to remove a unit test flake, so we should be extra careful not to reintroduce any flake there 🙏

lbernick:
> Thanks @lbernick - about this first part, I'm sure you're aware, this removes a lot of the changes introduced in [fff3fcd](https://github.com/tektoncd/pipeline/commit/fff3fcda2774f07d6787bbd0496c58610d381f9d) by @Yongxuanzhang to remove a unit test flake, so we should be extra careful not to reintroduce any flake there 🙏

Definitely! I've been communicating with Yongxuan a lot about these changes and how we do concurrency in our events tests. This isn't a guarantee of course, but `go test ./pkg/reconciler/events/cloudevent --count 1000` passes on my machine. To clarify why I believe this won't be flaky:

- Before Yongxuan's changes, tests were sleeping a short amount of time "until all events were received". This was flaky because it didn't block until the expected number of events were sent.
- The tests previously also timed out reading from the channel after a short amount of time. This was because in order to check that no extra events were received, they blocked until either an additional event was received, or the timeout was reached. The assumption was that all events would be present before reading and that this short amount of time was sufficient to read them, but these assumptions didn't hold.

The new tests take a different approach:
- Read `numExpectedEvents` from the channel. Each read blocks until an event is ready (from https://go.dev/tour/concurrency/5). The (long) timeout here is intended only to prevent the test from blocking forever in case something goes wrong with sending events. It's not intended to be reached during normal execution of the test, and the test fails if it's reached.
- There's a check afterwards that no additional events were received. The "default" block here is only executed if an extra event can't be read from the channel (from https://go.dev/tour/concurrency/6).

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -130,10 +107,34 @@ func (c *FakeClient) CheckCloudEventsUnordered(t *testing.T, testName string, wa
...
-	if len(expected) != 0 {
-		t.Errorf("%d events %#v are not received", len(expected), expected)
+	// Ensure that no extra events have been sent on the channel
+	for {
+		select {
+		case a := <-c:
+			t.Errorf("Unexpected event: %q", a)
+		default:
+			return gotEvents
+		}
```
afrittoli, line 138: This will return immediately unless there is an extra event in the channel.

I'm not sure why in the first loop we block until timeout, but here we can return right away.
Unless we introduce some kind of "done writing" channel/message, shouldn't we would always have to wait for a timeout (which would not be ideal)?

afrittoli: That said, we have a limited size buffer, so until we start reading messages from the channel, it won't be possible to write extra unexpected messages to the channel. Do you know if the channel will block then in case of overflow?

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
```diff
@@ -37,18 +40,15 @@ type FakeClientBehaviour struct {
...
 
 // newFakeClient is a FakeClient factory, it returns a client for the target
-func newFakeClient(behaviour *FakeClientBehaviour, expectedEventCount int) CEClient {
+func newFakeClient(behaviour *FakeClientBehaviour, expectedEventCount int) cloudevents.Client {
 	return FakeClient{
 		behaviour: behaviour,
 		// set buffersize to length of want events to make sure no extra events are sent
-		Events:    make(chan string, expectedEventCount),
-		waitGroup: &sync.WaitGroup{},
+		events: make(chan string, expectedEventCount),
```
afrittoli, line 51: We set the client buffer to `expectedEventCount` which means that we cannot receive more events than expected, at least until we start reading messages from the channel.

afrittoli (commented):
Thanks for this, it looks great, but I have a question about catching the case of more messages than expected.

vdemeester (approved):

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
lbernick: Ah I think you're right, nice catch. I tried running a few test cases with extra events and they passed 😱 It looks like the referenced example is not quite right! FYI @Yongxuanzhang

lbernick:
/hold
nice catch from @afrittoli, need to figure out how to handle cases w/ extra events sent

pkg/reconciler/events/cloudevent/cloudeventsfakeclient.go
lbernick: I can revert the concurrency changes for now and just have the cleanup-only changes in this PR if that's helpful?

Yongxuanzhang: > Ah I think you're right, nice catch. I tried running a few test cases with extra events and they passed 😱 It looks like the referenced example is not quite right! FYI @Yongxuanzhang we may need to fix the k8s events tests to not pass when they shouldn't

For k8s event, this may not be an issue. Since we don't initialize the channel with size=expected events, it is set to 1000.

Yongxuanzhang: IIUC, @afrittoli helped to find out this issue as well in https://github.com/tektoncd/pipeline/commit/fff3fcda2774f07d6787bbd0496c58610d381f9d, 

The comment is [here](https://github.com/tektoncd/pipeline/pull/5690#discussion_r1019626059). To address it I added a check when sending cloud event in FakeClient and will return error if extra events are to be sent

lbernick: I also tried increasing the channel size, and the tests still passed when they shouldn't. I think Andrea is right that the "default" branch gets picked up here since there may not be an event immediately ready on the channel.

lbernick: I think the check you added in response to Andrea's comment works because we waited until all events had been sent before reading any events from the channel, whereas this pattern starts reading from the channel without waiting for all the events to be available. I'm going to try to bring back some of the wait group logic in this PR but I'm not sure our logic is correct for k8s events.

Yongxuanzhang: > I think the check you added in response to Andrea's comment works because we waited until all events had been sent before reading any events from the channel, whereas this pattern starts reading from the channel without waiting for all the events to be available. I'm going to try to bring back some of the wait group logic in this PR but I'm not sure our logic is correct for k8s events.

emmm let me take a look

Yongxuanzhang: Oh I think there is similar situation in k8s event, but it should be very rare to happen. 

For k8s event, the channel size is set to 1000, so for our tests it's of low possibility that we can reach the buffer size and block the extra events. And since we have a large timer(equals to timeout of the test), we should be able to read the events even if it is sent in a goroutine and not in channel when we run `CheckEventsOrdered`.
 
I think maybe we can do the same for cloud events? Instead of setting the buffersize=expected events, we set it to a larger value (it was 100 before my previous PR)

lbernick:
/hold cancel
should be fixed now

lbernick:
@afrittoli interestingly it looks like the data race being reported previously in the unit tests was due to checking the cloud event cache within the goroutine. I'm not sure why this didn't show up as a data race previously? I think I should trim this PR down to only the cleanup changes and open a separate one with the concurrency changes, and maybe that can help us better understand if this is a bug in the fake client or the real code, WDYT?

lbernick:
I've reverted the concurrency changes and will put them in a separate PR.

Yongxuanzhang:
/lgtm
Thanks!

Issue #6076, opened 2023-01-31
Tekton unable to handle PipelineRuns too big


RafaeLeal:
# Expected Behavior

Tekton should be able to run PipelineRuns that are big or at least give a proper error.
# Actual Behavior

PipelineRun gets stuck in the same status in the cluster, not respecting any timeouts. 

# Steps to Reproduce the Problem

1. Create a PipelineRun with a lot of embedded tasks and status
2. Watch the tekton controller logs
3.

# Versions Info

<details>
  <summary><b>Kubernetes Version</b> (output of `kubectl version -o yaml`)</summary>

```yaml
$ kubectl version -o yaml
clientVersion:
  buildDate: "2022-10-12T10:47:25Z"
  compiler: gc
  gitCommit: 434bfd82814af038ad94d62ebe59b133fcb50506
  gitTreeState: clean
  gitVersion: v1.25.3
  goVersion: go1.19.2
  major: "1"
  minor: "25"
  platform: darwin/amd64
kustomizeVersion: v4.5.7
serverVersion:
  buildDate: "2022-11-29T18:41:42Z"
  compiler: gc
  gitCommit: 52e500d139bdef42fbc4540c357f0565c7867a81
  gitTreeState: clean
  gitVersion: v1.22.16-eks-ffeb93d
  goVersion: go1.16.15
  major: "1"
  minor: 22+
  platform: linux/amd64
```

</details>

<details>
  <summary><b>Tekton Pipeline Version</b> (output of `tkn version`)</summary>

```
Client version: 0.26.0
Pipeline version: v0.35.1
Triggers version: v0.20.0
Dashboard version: v0.26.0
```

</details>

# Additional Info
- We (Nubank) manage a very large CICD cluster (we reached over 700k taskruns/week), and we've built some abstractions over Tekton to make it easier for our users to declare new pipelines
- We started to observe that some pipelineruns were getting stuck in our cluster, not finishing even hours after the timeout.
- Investigating the logs, we found these (im formatting to become easier to read):
```yaml
{ 
   caller: pipelinerun/reconciler.go:268
   commit: 422a468
   error: etcdserver: request is too large
   knative.dev/controller: github.com.tektoncd.pipeline.pkg.reconciler.pipelinerun.Reconciler
   knative.dev/key: pipelines/itaipu-tests-main-run-ie3c7
   knative.dev/kind: tekton.dev.PipelineRun
[...]

PR #6095, opened 2023-02-01
Fix for PipelineRuns getting stuck in the running state in the cluster

RafaeLeal:
# Changes

Closes: https://github.com/tektoncd/pipeline/issues/6076
This PR tries to fix scenarios where in the same reconciliation cycle the PipelineRun status gets too big to be updated and the status is changed to timeout. This means that the [UpdateStatus](https://github.com/tektoncd/pipeline/blob/main/pkg/client/injection/reconciler/pipeline/v1beta1/pipelinerun/reconciler.go#L267-L272) fails, so the PipelineRun can't get out from the running status, even after the timeout.
The PR introduces an arbitrary threshold (2*timeout), and if the reconciliation notice that threshold is reached, it skips any other update, and updates only the status, making the request as small as possible, avoiding etcd request size errors.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
Fix a bug that made big PipelineRuns get stuck in the running state in the cluster
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -183,6 +183,20 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 	// Read the initial condition
 	before := pr.Status.GetCondition(apis.ConditionSucceeded)
 
+	// Check if we are failing to mark this as timed out for a while. If we are, mark immediately and finish the
+	// reconcile. We are assuming here that if the PipelineRun has timed out for a long time, it had time to run
+	// before and it kept failing. One reason that can happen is exceeding etcd request size limit. Finishing it early
+	// makes sure the request size is manageable
+	if pr.HasTimedOutForALongTime(ctx, c.Clock) && !pr.IsTimeoutConditionSet() {
+		if err := timeoutPipelineRun(ctx, logger, pr, c.PipelineClientSet); err != nil {
+			return err
+		}
+		if err := c.finishReconcileUpdateEmitEvents(ctx, pr, before, nil); err != nil {
+			return err
+		}
+		return controller.NewPermanentError(errors.New("PipelineRun has timed out for a long time"))
+	}
+
 	if !pr.HasStarted() && !pr.IsPending() {
 		pr.Status.InitializeConditions(c.Clock)
 		// In case node time was not synchronized, when controller has been scheduled to other nodes.
--- a/pkg/reconciler/pipelinerun/pipelinerun_test.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun_test.go
@@ -28,6 +28,8 @@ import (
 	"testing"
 	"time"
 
+	"sigs.k8s.io/yaml"
+
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/google/go-containerregistry/pkg/registry"
@@ -1385,7 +1387,7 @@ status:
     reason: Running
     status: Unknown
     type: Succeeded
-  startTime: "2021-12-31T00:00:00Z"
+  startTime: "2021-12-31T11:00:00Z"
   childReferences:
   - name: test-pipeline-run-custom-task-hello-world-1
     pipelineTaskName: hello-world-1
@@ -2039,7 +2041,7 @@ spec:
   serviceAccountName: test-sa
   timeout: 12h0m0s
 status:
-  startTime: "2021-12-31T00:00:00Z"
+  startTime: "2021-12-31T11:00:00Z"
 `)}
 	ts := []*v1beta1.Task{simpleHelloWorldTask}
 
@@ -2101,7 +2103,7 @@ spec:
   timeouts:
     pipeline: 12h0m0s
 status:
-  startTime: "2021-12-31T00:00:00Z"
+  startTime: "2021-12-31T11:00:00Z"
   childReferences:
   - name: test-pipeline-run-with-timeout-hello-world-1
     pipelineTaskName: hello-world-1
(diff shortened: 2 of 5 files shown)


pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -176,6 +176,18 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 	// Read the initial condition
 	before := pr.Status.GetCondition(apis.ConditionSucceeded)
 
+	// Check if we are failing to mark this as timed out for a while. If we are, mark immediately and finish the
+	// reconcile
+	if pr.HasTimedOutForALongTime(ctx, c.Clock) && !isPipelineRunTimeoutConditionSet(pr) {
+		setPipelineRunTimeoutCondition(ctx, pr)
```
afrittoli, line 182: Why not call `timeoutPipelineRun` instead to ensure the child objects have been attempted at least?
That won't cause an increase in the size of `PipelineRun` and will behave better in cases where the long timeout is caused by a different situation like for instance the controller being unresponsive or down for some time.

The only extra bit of data that `timeoutPipelineRun` would write is the completion time, which is unlikely to cause the etcd failure I think. If you would prefer to not write a completion time, we should make it obvious in the comment what is going on.

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -176,6 +176,18 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
...
+
+		if err := c.finishReconcileUpdateEmitEvents(ctx, pr, before, nil); err != nil {
+			return err
+		}
+
+		return nil
```
afrittoli, line 188: Rather than returning `nil` I think we could return a "PermanentError" instead, which means that the key won't be re-queued to the controller but we'll still track this in an error situation and we can provide a context in the error about what is going on - i.e. that the pipeline run condition is still not updated after  2*timeout so we apply a different logic here.

pkg/reconciler/pipelinerun/timeout.go
```diff
@@ -81,20 +81,29 @@ func init() {
...
+
+// isPipelineRunTimeoutConditionSet returns true when the pipelinerun has the pipelinerun timed out reason
+func isPipelineRunTimeoutConditionSet(pr *v1beta1.PipelineRun) bool {
+	condition := pr.Status.GetCondition(apis.ConditionSucceeded)
+	return condition.IsFalse() && condition.Reason == v1beta1.PipelineRunReasonTimedOut.String()
+}
+
 // timeoutPipelineRun marks the PipelineRun as timed out and any resolved TaskRun(s) too.
 func timeoutPipelineRun(ctx context.Context, logger *zap.SugaredLogger, pr *v1beta1.PipelineRun, clientSet clientset.Interface) error {
 	errs := timeoutPipelineTasks(ctx, logger, pr, clientSet)
```
afrittoli, line 102: Note: this would not use extra space in etcd for the pipelinerun

pkg/reconciler/pipelinerun/timeout.go
```diff
@@ -81,20 +81,29 @@ func init() {
...
-
-		pr.Status.SetCondition(&apis.Condition{
-			Type:    apis.ConditionSucceeded,
-			Status:  corev1.ConditionFalse,
-			Reason:  reason,
-			Message: fmt.Sprintf("PipelineRun %q failed to finish within %q", pr.Name, pr.PipelineTimeout(ctx).String()),
-		})
+		setPipelineRunTimeoutCondition(ctx, pr)
 		// update pr completed time
 		pr.Status.CompletionTime = &metav1.Time{Time: time.Now()}
```
afrittoli, line 108: This will use a tiny bit extra space in etcd for the `PipelineRun`, but it's important to have always associated with a condition change to `ConditionTrue` or `ConditionFalse`.

afrittoli (commented):
Thanks for this, I think it looks good and the tests are really nice.
I have a couple questions / comments on the logic.

afrittoli:
Thanks, @RafaeLeal for this PR!

Could you provide a bit more context about the issue in the commit message / PR description, and fill in the PR checklist from the template? Also, tide does not automatically squash commits before merging, so I would ask you to please squash them at least once the PR has been approved.

Thank you!

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -176,6 +176,18 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 	// Read the initial condition
 	before := pr.Status.GetCondition(apis.ConditionSucceeded)
 
+	// Check if we are failing to mark this as timed out for a while. If we are, mark immediately and finish the
+	// reconcile
```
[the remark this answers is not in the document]
RafaeLeal: 👍

pkg/reconciler/pipelinerun/pipelinerun.go
RafaeLeal: I didn't thought about it, to be honest. Fair enough.

pkg/reconciler/pipelinerun/pipelinerun.go
RafaeLeal: 👍  I'll change it, but should we consider returning a `controller.NewRequeueImmediately()`? Since we are testing for `!isPipelineRunTimeoutConditionSet(pr)`, it would not be in an infinite loop.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -2554,6 +2556,152 @@ spec:
...
+		}
+
+		if len(bytes) > limit {
+			t.Logf("request size: %d\nrequest limit: %d\n", len(bytes), limit)
+			t.Logf("payload:\n%s\n", string(bytes))
+			// actionBytes, err := yaml.Marshal(action)
+			// if err != nil {
+			// 	t.Fatalf("returned a unserializable action: %+v", action)
+			// }
+			// t.Logf("action:\n%s\n", string(actionBytes))
```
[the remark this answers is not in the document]
RafaeLeal: Nope, sorry.

pkg/reconciler/pipelinerun/timeout.go
RafaeLeal: 👍

pkg/reconciler/pipelinerun/timeout.go
RafaeLeal: Fair enough

RafaeLeal:
> Could you provide a bit more context about the issue in the commit message / PR description, and fill in the PR checklist from the template? Also, tide does not automatically squash commits before merging, so I would ask you to please squash them at least once the PR has been approved.

Sure, I was not done yet actually. I added another test on HasTimedOutForALongTime func. I had some ideas on how to avoid arbitrary thresholds, that I shared in the issue, which I quote here.

> While implementing it, I was considering maybe we could always do a two-step timing out, this way we could avoid arbitrary thresholds. The first reconciliation would check `pr.HasTimedOut()` and mark the status and return `controller.NewRequeueImmediately()`. This would trigger a UpdateStatus with only the condition change, then in the second reconciliation, we could try to update the rest of the status (the `childReferences`, for example).

> This could work, I think. But the problem I had is that we still depend a lot on the order of the execution to make everything work properly. In the second reconciliation, we already have the timed-out condition, which means the `pr.IsDone()` is true and this changes the whole reconciliation process.
https://github.com/tektoncd/pipeline/blob/main/pkg/reconciler/pipelinerun/pipelinerun.go#L205-L225

> Looking at this, does it make sense to "fail fast" the reconciliation process?
For example, even inside in this `pr.IsDone()` branch, it runs several cleanup processes. If one fails, should we really prevent the next one from executing? 🤔

afrittoli:
> > While implementing it, I was considering maybe we could always do a two-step timing out, this way we could avoid arbitrary thresholds. The first reconciliation would check `pr.HasTimedOut()` and mark the status and return `controller.NewRequeueImmediately()`. This would trigger a UpdateStatus with only the condition change, then in the second reconciliation, we could try to update the rest of the status (the `childReferences`, for example).
> 

It could work, but I have two concerns:
- orchestrating updates across reconciles can be tricky - if there is no external input it's generally best to try and do everything in one reconcile cycle
- but I think it would be confusing for clients. If a client reads that the pipelinerun has completed, how does it know whether there is an update missing somewhere? Even with `NewRequeueImmediately`, the controller queue can be quite busy and it may take some time before the new reconcile happens

RafaeLeal:
> orchestrating updates across reconciles can be tricky - if there is no external input it's generally best to try and do everything in one reconcile cycle

> but I think it would be confusing for clients. If a client reads that the pipelinerun has completed, how does it know whether there is an update missing somewhere? Even with NewRequeueImmediately, the controller queue can be quite busy and it may take some time before the new reconcile happens

I agree with these concerns. I think we might need a more robust solution long-term for this, but it seems pretty good for now. WDYT?

RafaeLeal:
@afrittoli let me know if there's anything missing to merge this :)

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -183,6 +183,20 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 	// Read the initial condition
 	before := pr.Status.GetCondition(apis.ConditionSucceeded)
 
+	// Check if we are failing to mark this as timed out for a while. If we are, mark immediately and finish the
+	// reconcile. We are assuming here that if the PipelineRun has timed out for a long time, it had time to run the
```
lbernick, line 187:
```suggestion
	// reconcile. We are assuming here that if the PipelineRun has timed out for a long time, it had time to run
```

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -183,6 +183,20 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
...
+	// before and it kept failing. One reason that can happen is exceeding etcd request size limit. Finishing it early
+	// makes sure the request size is manageable
+	if pr.HasTimedOutForALongTime(ctx, c.Clock) && !isPipelineRunTimeoutConditionSet(pr) {
+		if err := timeoutPipelineRun(ctx, logger, pr, c.PipelineClientSet); err != nil {
+			return err
+		}
+		if err := c.finishReconcileUpdateEmitEvents(ctx, pr, before, nil); err != nil {
```
lbernick, line 194: I'm a bit confused why this works. From the PR description and linked issue, it sounds like the controller is trying to update the pipelinerun status based on the taskrun status, and the request (which contains the full pipelinerun status) is too large to be handled by etcd. However, this calls finishReconcileUpdateEmitEvents, which calls updateLabelsAndAnnotations, which updates the pr with its current status. Why would this request not get stuck for being too big?

Is the idea that we skip including info from the child TaskRun that would make the status too large? If this is the case, it seems like this fix would only help for embedded-status=full (or both), since the childReferences are added as soon as the taskruns are created, and don't increase in size when the taskruns complete.

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -2351,6 +2353,148 @@ spec:
...
+			if reconciledRun.Status.GetCondition(apis.ConditionSucceeded).Reason != "PipelineRunTimeout" {
+				t.Errorf("Expected PipelineRun to be timed out, but condition reason is %s", reconciledRun.Status.GetCondition(apis.ConditionSucceeded))
+			}
+		})
+	}
+}
+
+// withEtcdRequestSizeLimit calculates the yaml marshal of the payload and gives an `etcdserver: request too large` when
+// the limit is reached
+func withEtcdRequestSizeLimit(t *testing.T, limit int) ktesting.ReactionFunc {
```
lbernick, line 2480:
```suggestion
func withEtcdRequestSizeLimit(t *testing.T, limitBytes int) ktesting.ReactionFunc {
```

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -2351,6 +2353,148 @@ spec:
...
+			prt := newPipelineRunTest(t, d)
+			defer prt.Cancel()
+
+			wantEvents := []string{
+				"Warning Failed PipelineRun \"test-pipeline-run-with-timeout\" failed to finish within \"12h0m0s\"",
+			}
+
+			// this limit is just enough to set the timeout condition, but not enough for extra metadata.
+			etcdRequestSizeLimit := 650
+			prt.TestAssets.Clients.Pipeline.PrependReactor("update", "pipelineruns", withEtcdRequestSizeLimit(t, etcdRequestSizeLimit))
```
lbernick, line 2442: nice!

pkg/reconciler/pipelinerun/pipelinerun.go
RafaeLeal: Yes, I was thinking about embedded-status=full when implementing this.. but I think this scenario you mention would be covered too. If we receive a `request is too large` error while adding `childReferences`, wouldn't the PipelineRun get stuck in the `running` state in the cluster too? I believe this would prevent that too.

Even if the `updateLabelsAndAnnotations` fails with `request is too large`, I believe that wouldn't be a problem. Sure, the reconciliation would fail, but the UpdateStatus request would succeed in making the PipelineRun condition transition, so the PipelineRun wouldn't be stuck in a running state. That's because the UpdateStatus only apply changes in the status field.

lbernick (approved):
@RafaeLeal do you think there is any work remaining to address https://github.com/tektoncd/pipeline/issues/6076 once this is merged? (If not I'll link the issue directly to the PR)

RafaeLeal:
> @RafaeLeal do you think there is any work remaining to address #6076 once this is merged? (If not I'll link the issue directly to the PR)

I don't think so, you can link the issue.

pkg/reconciler/pipelinerun/timeout.go
```diff
@@ -81,20 +81,29 @@ func init() {
...
+		Reason:  v1beta1.PipelineRunReasonTimedOut.String(),
+		Message: fmt.Sprintf("PipelineRun %q failed to finish within %q", pr.Name, pr.PipelineTimeout(ctx).String()),
+	})
+}
+
+// isPipelineRunTimeoutConditionSet returns true when the pipelinerun has the pipelinerun timed out reason
+func isPipelineRunTimeoutConditionSet(pr *v1beta1.PipelineRun) bool {
+	condition := pr.Status.GetCondition(apis.ConditionSucceeded)
+	return condition.IsFalse() && condition.Reason == v1beta1.PipelineRunReasonTimedOut.String()
+}
```
jerop, line 98: consider defining this as a member function in [pipelinerun_types.go](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L159) with tests:
```suggestion
// IsTimeoutConditionSet returns true when the pipelinerun has the pipelinerun timed out reason
func (pr *v1beta1.PipelineRun) IsTimeoutConditionSet() bool {
	condition := pr.Status.GetCondition(apis.ConditionSucceeded)
	return condition.IsFalse() && condition.Reason == v1beta1.PipelineRunReasonTimedOut.String()
}
```

pkg/reconciler/pipelinerun/timeout.go
```diff
@@ -81,20 +81,29 @@ func init() {
...
 
+// setPipelineRunTimeoutCondition sets the status of the PipelineRun to timed out.
+func setPipelineRunTimeoutCondition(ctx context.Context, pr *v1beta1.PipelineRun) {
+	pr.Status.SetCondition(&apis.Condition{
+		Type:    apis.ConditionSucceeded,
+		Status:  corev1.ConditionFalse,
+		Reason:  v1beta1.PipelineRunReasonTimedOut.String(),
+		Message: fmt.Sprintf("PipelineRun %q failed to finish within %q", pr.Name, pr.PipelineTimeout(ctx).String()),
+	})
+}
```
jerop, line 92: consider defining this as a member function in [pipelinerun_types.go](https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/pipelinerun_types.go#L159) with tests:
```suggestion
// SetTimeoutCondition sets the status of the PipelineRun to timed out.
func (pr *v1beta1.PipelineRun) SetTimeoutCondition(ctx context.Context) {
	pr.Status.SetCondition(&apis.Condition{
		Type:    apis.ConditionSucceeded,
		Status:  corev1.ConditionFalse,
		Reason:  v1beta1.PipelineRunReasonTimedOut.String(),
		Message: fmt.Sprintf("PipelineRun %q failed to finish within %q", pr.Name, pr.PipelineTimeout(ctx).String()),
	})
}
```

pkg/reconciler/pipelinerun/timeout.go
```diff
@@ -81,20 +81,29 @@ func init() {
...
 	if len(errs) == 0 {
-		reason := v1beta1.PipelineRunReasonTimedOut.String()
-
-		pr.Status.SetCondition(&apis.Condition{
-			Type:    apis.ConditionSucceeded,
-			Status:  corev1.ConditionFalse,
-			Reason:  reason,
-			Message: fmt.Sprintf("PipelineRun %q failed to finish within %q", pr.Name, pr.PipelineTimeout(ctx).String()),
-		})
+		setPipelineRunTimeoutCondition(ctx, pr)
```
jerop, line 106:
```suggestion
		pr.SetTimeoutCondition(ctx)
```

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -183,6 +183,20 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 	// Read the initial condition
 	before := pr.Status.GetCondition(apis.ConditionSucceeded)
 
+	// Check if we are failing to mark this as timed out for a while. If we are, mark immediately and finish the
+	// reconcile. We are assuming here that if the PipelineRun has timed out for a long time, it had time to run the
+	// before and it kept failing. One reason that can happen is exceeding etcd request size limit. Finishing it early
+	// makes sure the request size is manageable
+	if pr.HasTimedOutForALongTime(ctx, c.Clock) && !isPipelineRunTimeoutConditionSet(pr) {
```
jerop, line 190:
```suggestion
	if pr.HasTimedOutForALongTime(ctx, c.Clock) && !pr.IsTimeoutConditionSet() {
```

jerop (commented):
Thanks for this fix @RafaeLeal, left some suggestions to make the code cleaner

pkg/reconciler/pipelinerun/pipelinerun.go
RafaeLeal: 👍

RafaeLeal:
@jerop @lbernick  accepted all suggestions and rebased 👍

PR #6096, opened 2023-02-01
[WIP] Remove Artifacts Package

JeromeJu:
# Changes
This commit cleans up the artifacts package.
It prepares for the removal storage resources to mitigate dependencies.

/kind cleanup

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```


JeromeJu:
I have been trying to cleanup the artifacts first before moving the entire chunk of storage resources.

Some questions I encountered during the cleanup:
- It feels that the output_resources paths are affected by the removal of such setup for pvc, shall we remove them along with the artifacts package?

lbernick:
I think artifacts are probably required for the storage resource to work correctly, so removing artifacts before removing storage resources would put us in an unreleasable state.

JeromeJu:
> I think artifacts are probably required for the storage resource to work correctly (although haven't looked into this thoroughly), so removing artifacts before removing storage resources would put us in an unreleasable state.

Thanks for the instant replies. If this is not preferred then let's remove both storage resources and artifacts in one shot.

afrittoli:
Given that this is already a large PR, I'm not sure we should bundle it with more changes, but I agree with @lbernick that we must first delete the remaining PipelineResources and then this.

PR #6092, opened 2023-01-31
when expressions with array results (doc and example)

pritidesai:
# Changes

Adding an example showcasing the usage of array results in `when` expressions, including indexing into an array results and array results with star notation. This usage includes both `input` and `values` components of `when` expressions.

Also, added doc with explicit syntax for `input` and `values`.

This is for the preparation of a potential promotion of https://github.com/tektoncd/pipeline/issues/5688.

Signed-off-by: pritidesai <pdesai@us.ibm.com>

/kind documentation

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ ] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
Doc update for - How to refer to an array result in when expressions?
```

PR #6060, opened 2023-01-27
updating an example pipelinerun with array indexing

pritidesai:
# Changes

* Validate param values indexing into an array which are mapped to a pipeline params instead of echoing.
* Add an example of how to reference a single element from an array. 
 
This is for the preparation of a potential promotion of https://github.com/tektoncd/pipeline/issues/5688.

Signed-off-by: pritidesai <pdesai@us.ibm.com>

/kind misc

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ ] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

PR #6055, opened 2023-01-26
updating an existing example pipelinerun with array results

pritidesai:
# Changes

Updating a PR to validate the length of the array results including each element in the array.

Also, added an example to validate an array when one of the elements is empty string.

This is for the preparation of a potential promotion of https://github.com/tektoncd/pipeline/issues/5688.

Signed-off-by: pritidesai <pdesai@us.ibm.com>

/kind misc

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [ ] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [ ] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

Issue #5688, opened 2022-10-26
Promote TEP-0075 and TEP-0076


sravanlakkimsetti:
### Feature looking for Promotion

These features are currently behind alpha. Requesting them to be promoted to beta

- [TEP-0075 Propose object (dicitonary) param and result types](https://github.com/tektoncd/community/blob/main/teps/0075-object-param-and-result-types.md) and 
- [TEP-0076 Propose array results and indexing](https://github.com/tektoncd/community/commit/e0e7d73e08113ac82e5fe8bc6467d31db51fbec7)

### Current Feature Version

`enable-api-fields: alpha`

### Raise Feature Version

`enable-api-fields: stable`

### Pipeline Release

0.38

### Issues Reported by the Users

None

### Usage in Dogfooding Cluster

None

### Examples/Tests

- [Validate Pipeline results array index](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/alpha/pipeline-emitting-results.yaml)
- [A pipeline run example with both object param and result](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/alpha/pipeline-object-param-and-result.yaml)
- [Support Object Results substitution](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/alpha/pipeline-object-results.yaml)
- [Array Results substitution](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/alpha/pipelinerun-array-results-substitution.yaml)
- [Indexing into array for params](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/alpha/pipelinerun-param-array-indexing.yaml)

dibyom:
We may have to resolve https://github.com/tektoncd/pipeline/issues/5803 and https://github.com/tektoncd/chains/issues/636 before moving this to beta

Yongxuanzhang:
> We have teams that have gone through testing of the alpha features here and in particular would like to use the array support in production - e.g. TEP76. FWIW we do not currently have production users asking for dictionary support. Are there any issues that are holding up making the array support at least beta?

I will bring this up next Monday!

pritidesai:
@jerop brought up a great point on validating array and object types consumption in `whenexpression` [functionlities ](https://github.com/tektoncd/pipeline/pull/6055#pullrequestreview-1272826458). An **action item** here is to make sure we cover this.

[...]

PR #6103, opened 2023-02-02
TEP-0076 (array results and indexing into array) promoted to beta 🎓

pritidesai:
# Changes

[Pipelines 0.38](https://github.com/tektoncd/pipeline/releases/tag/v0.38.0) introduced two alpha features - producing array results and indexing into array param (proposed in [TEP-0076](https://github.com/tektoncd/community/blob/main/teps/0076-array-result-types.md)). This commit is promoting these two features to `beta` such that these features can be used by the task authors and pipeline authors in a cluster when `enable-api-fields` is either set to `beta` or `alpha`.

These features are still not available in stable API i.e. these features will not work when `enable-api-fields` set to `stable`

Before we can merge this PR, I have following three PRs to update the doc and examples. Please review the outstanding PRs first. **These PRs are merged now.**

- PR #6055 
- PR #6060 
- PR #6092 

Fixes part of https://github.com/tektoncd/pipeline/issues/5688.

Signed-off-by: pritidesai <pdesai@us.ibm.com>

/kind feature

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
TEP-0076 promoted to beta - Array results and indexing into an array is now possible with enable-api-fields set to beta.
```


pritidesai:
@afrittoli @JeromeJu @Yongxuanzhang I appreciate if you can review the changes in this PR, my team is looking for this feature as part of 0.45. Thanks!

pkg/apis/config/feature_flags_test.go
```diff
@@ -287,6 +287,37 @@ func TestCheckWarnResourceVerificationMode(t *testing.T) {
 	}
 }
 
+// TestCheckAlphaOrBetaAPIFields validates CheckAlphaOrBetaAPIFields for alpha, beta, and stable context
+func TestCheckAlphaOrBetaAPIFields(t *testing.T) {
```
JeromeJu, line 291: Thanks for the PR @pritidesai this looks good.

Just one NIT question, curious why we would explicitly check this is not Stable instead of testing if the intended flags are set correctly eg. verify stable for `config.EnableStableAPIFields(ctx)`?

JeromeJu (approved):

Yongxuanzhang (commented):
/lgtm
Thanks! Sorry I thought we want to merge other issues fixes before this one. I think it's ok to get this merged in v0.45

pkg/apis/config/feature_flags_test.go
pritidesai: thank you @JeromeJu for the review 👍 

I am not sure if I followed your question 😜 

This is a test function to validate `CheckAlphaOrBetaAPIFields` i.e. return true if `enable-api-fields` set to either `alpha` or `beta` and false otherwise. The test cases below include testing for all three flags.

JeromeJu: I do agree that for `CheckAlphaOrBetaAPIFields` this makes sense and covers all the api feature flags. 

I was curious at first sight about the naming for `CheckAlphaOrBetaAPIFields` but now I think this is fine as an intermediate stage to beta.

vdemeester (approved):

vdemeester:
Cc @abayer @afrittoli @jerop to remove the hold 😛 but yay 🎉

docs/pipelines.md
```diff
@@ -942,17 +942,17 @@ Sharing `Results` between `Tasks` in a `Pipeline` happens via
 a `Result` and another receives it as a `Parameter` with a variable such as
 `$(tasks.<task-name>.results.<result-name>)`. Pipeline support two new types of
 results and parameters: array `[]string` and object `map[string]string`.
-Both are alpha features and can be enabled by setting `enable-api-fields` to `alpha`.
+Object result is alpha feature and can be enabled by setting `enable-api-fields` to `alpha`.
```
afrittoli, line 945: NIT: shouldn't we say that "array results are a beta feature. They can be enabled in the `v1` API by setting the `enable-api-fields` to `beta`

docs/tasks.md
```diff
@@ -833,7 +833,7 @@ or [at the `Pipeline` level](./pipelines.md#emitting-results-from-a-pipeline).
 #### Emitting Array `Results`
 
 Tekton Task also supports defining a result of type `array` and `object` in addition to `string`.
-Emitting a task result of type `array` is an `alpha` feature implemented based on the
+Emitting a task result of type `array` is implemented based on the
```
afrittoli, line 836: Ditto: shouldn't me mention that arrays are a beta feature?

pkg/apis/config/feature_flags_test.go
```diff
@@ -287,6 +287,37 @@ func TestCheckWarnResourceVerificationMode(t *testing.T) {
...
+		name:           "when enable-api-fields is set to alpha",
+		c:              config.EnableAlphaAPIFields(ctx),
+		expectedResult: true,
+	}, {
+		name:           "when enable-api-fields is set to beta",
+		c:              config.EnableBetaAPIFields(ctx),
+		expectedResult: true,
+	}, {
+		name:           "when enable-api-fields is set to stable",
+		c:              config.EnableStableAPIFields(ctx),
```
afrittoli, line 308: NIT: stable is the default (and will always be), so this feels a bit redundant.

pkg/apis/pipeline/v1/result_validation.go
```diff
@@ -33,21 +33,23 @@ func (tr TaskResult) Validate(ctx context.Context) (errs *apis.FieldError) {
...
 	// and should be considered valid
-	if tr.Type == "" {
+	case tr.Type == "":
 		return nil
-	}
-
-	// By default the result type is string
-	if tr.Type != ResultsTypeString {
+	// By default, the result type is string
+	case tr.Type != ResultsTypeString:
```
afrittoli, line 52: NIT: I find it a bit brittle to rely on the sum of case condition to provide full coverage of all cases, I think it would be preferable to do something like:

```go
	default:
            if tr.Type != ResultsTypeString:
               return apis.ErrInvalidValue(tr.Type, "type", "type must be string")
```

afrittoli (commented):
Thanks @pritidesai 🎉 - this looks mostly good to me, only a couple of minor things - it would be nice to have the docs updated, especially given the fact that we don't have v1/v1beta separate docs, so the docs should default to the latest API which is v1

docs/pipelines.md
pritidesai: I am not sure, with `v1beta1` API, for this feature to work, `enable-api-fields` does not have to be `alpha` anymore which means it can be set to either `beta` or `stable`. 

@chitrangpatel how did you approach this when we promoted propagated params/workspaces  features? 🤔 Or these features were not classified as `beta`? will appreciate any guidance. Thanks!

abayer:
/approve

Good by me!

docs/pipelines.md
chitrangpatel: Propagated Params was moved straight to stable since stable and beta were same at the time.
Propagated workspaces is currently beta meaning that it will work if `enable-api-fields` is `alpha` or `beta` but **not** `stable`.

pkg/apis/pipeline/v1/result_validation.go
pritidesai: thank you @afrittoli I am not able to follow the concern here, sorry 🙃 😕

pritidesai: the `default` is always guaranteed to execute, not sure if it make sense to introduce default with a `condition` 🤔 v/s its better to be explicit with the `case` statement ...

pkg/apis/config/feature_flags_test.go
pritidesai: Yup I completely agree, `stable` is a default. its just added to be consistent with the 

> /approve
> 
> Good by me!

thanks @abayer 🙏 

@vdemeester can we cancel the hold? I do not think there is anything pending here now. If there is a concern of array params in a `matrix` and `when` expressions, we had examples and doc updated for both features in PR #6092 and PR #6077. @Yongxuanzhang please confirm!

pkg/apis/pipeline/v1beta1/result_validation_test.go
```diff
@@ -74,6 +83,9 @@ func TestResultsValidate(t *testing.T) {
 			if tt.apiFields == "alpha" {
 				ctx = config.EnableAlphaAPIFields(ctx)
 			}
+			if tt.apiFields == config.BetaAPIFields {
```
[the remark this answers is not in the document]
pritidesai: yup completely agree, will create a separate PR. There are many other such cases in the same test file. Thanks!

docs/pipelines.md
jerop: yea, we should clarify that array results is an beta feature

@pritidesai it seems that in both v1 and v1beta1, we check that beta feature flag is enabled -- @lbernick we may need clarification in the TEP about whether the flag should be the same in both versions

pkg/apis/config/feature_flags.go
```diff
@@ -337,6 +342,12 @@ func CheckWarnResourceVerificationMode(ctx context.Context) bool {
 	return cfg.FeatureFlags.ResourceVerificationMode == WarnResourceVerificationMode
 }
 
+// CheckAlphaOrBetaAPIFields return true if the enable-api-fields is either set to alpha or set to beta
+func CheckAlphaOrBetaAPIFields(ctx context.Context) bool {
+	cfg := FromContextOrDefaults(ctx)
+	return cfg.FeatureFlags.EnableAPIFields == AlphaAPIFields || cfg.FeatureFlags.EnableAPIFields == BetaAPIFields
+}
+
```
jerop, line 350: not related to this PR -- @chitrangpatel what do you think of cleaning up propagated workspaces to reuse this function? just to make the logic clearer?

chitrangpatel: Yes, can do! I think it's best to stay consistent. Once this is merged, I will open an issue and implement a fix.

docs/pipelines.md
pritidesai: > Propagated workspaces is currently beta meaning that it will work if enable-api-fields is alpha or beta but not stable.

Thanks @chitrangpatel for getting back. How did we document that propagated workspace is `beta` feature? At the same time, I am checking docs. I just want to make sure its consistent across many different features.

pkg/apis/config/feature_flags.go
jerop: sounds good, thanks @chitrangpatel!

docs/pipelines.md
chitrangpatel: https://github.com/tektoncd/pipeline/pull/6030/files is all I had done.

pritidesai: So I just checked the PR #6030, there is no specific guidelines other than the replacing `alpha` with `beta` in the docs 🙃 I will update the docs here but it's really very confusing for folks using `v1beta1` APIs.

pritidesai:
thank you @Yongxuanzhang @afrittoli @vdemeester @abayer @JeromeJu @jerop and @chitrangpatel 🙏 

I have updated this PR to document that this feature is `beta` feature. I am looking for lgtm unless there is anything else pending. Thanks a bunch!

jerop (approved):

Yongxuanzhang:
/lgtm
Thanks!

pkg/apis/pipeline/v1/result_validation.go
afrittoli: The `default` is not always guaranteed to execute, it won't be executed if any of the options before matches. 
But it guarantees execution if the value was not caught by any of the `case` statements. 
My proposed alternative is not very different indeed, I think a better one would be to explicitly succeed on "string" and fail with a different error message in the default case.
Not blocking for this PR anyways, I can make a small update once this is merged.

Issue #6111, opened 2023-02-03
introduce type validation in task results (array and object)


pritidesai:
I created a pipeline with a task producing an array result:

```
echo -n "[\"$image1\", \"$image2\", \"$image3\",]}]" | tee $(results.array-results-1.path)
```

Notice an extra `,` after `$image3`.

The pipelineRun returned an error with reason `TaskRunValidationFailed` and an error message:

```
"mismatched types:  \"array-results-1\": task result is expected to be \"array\" type but was initialized to a different type \"string\"
```

It's great that we have such error reporting to notify the users that there was some problem with the result initialization. But at the same time this is not enough. An extra `,` in the task result returns an error that the type of the result does not match. 

I am opening this issue here to experiment whether it is possible to introduce some kind of data validation and to report a little more precise error message such that the user can easily spot the issue in the task.

Yongxuanzhang:
I think one reason for this case is that the JSON format data cannot be unmarshalled as array, so it is unmarshalled to string instead. 
I think we could add more precise error message as @pritidesai suggested. Will open a PR

PR #6119, opened 2023-02-07
add more error messages for results validation

Yongxuanzhang:
# Changes
This commit adds more error messages for results validation. Before this commit we only log the types, but the reasons lead to type mismatch could be the JSON data is not of correct format, or the type defined is not correct. This commit adds this hint to make it easier for users to debug.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/taskrun/validate_resources.go
+++ b/pkg/reconciler/taskrun/validate_resources.go
@@ -307,7 +307,7 @@ func validateTaskRunResults(tr *v1beta1.TaskRun, resolvedTaskSpec *v1beta1.TaskS
 			s = append(s, fmt.Sprintf(" \"%v\": %v", k, v))
 		}
 		sort.Strings(s)
-		return fmt.Errorf("mismatched types: %v", strings.Join(s, ","))
+		return fmt.Errorf("Provided results don't match declared results; may be invalid JSON or missing result declaration: %v", strings.Join(s, ","))
 	}
 
 	// When get the results, for object value need to check if they have missing keys.
--- a/pkg/reconciler/taskrun/taskrun_test.go
+++ b/pkg/reconciler/taskrun/taskrun_test.go
@@ -1881,19 +1881,19 @@ status:
   - reason: ToBeRetried
     status: Unknown
     type: Succeeded
-    message: "mismatched types:  \"aResult\": task result is expected to be \"array\" type but was initialized to a different type \"string\""
+    message: "Provided results don't match declared results; may be invalid JSON or missing result declaration:  \"aResult\": task result is expected to be \"array\" type but was initialized to a different type \"string\""
   sideCars:
   retriesStatus:
   - conditions:
     - reason: TaskRunValidationFailed
       status: "False"
       type: "Succeeded"
-      message: "mismatched types:  \"aResult\": task result is expected to be \"array\" type but was initialized to a different type \"string\""
+      message: "Provided results don't match declared results; may be invalid JSON or missing result declaration:  \"aResult\": task result is expected to be \"array\" type but was initialized to a different type \"string\""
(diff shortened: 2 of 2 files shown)


Yongxuanzhang:
Hi @pritidesai, is this ok for https://github.com/tektoncd/pipeline/issues/6111? I know that in https://github.com/tektoncd/pipeline/issues/6111 you're suggesting to do data validation for array and object.  I wonder if adding this can also meet the need?

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -307,7 +307,7 @@ func validateTaskRunResults(tr *v1beta1.TaskRun, resolvedTaskSpec *v1beta1.TaskS
 			s = append(s, fmt.Sprintf(" \"%v\": %v", k, v))
 		}
 		sort.Strings(s)
-		return fmt.Errorf("mismatched types: %v", strings.Join(s, ","))
+		return fmt.Errorf("mismatched types, please check and make sure results are emitted in correct JSON format or the type in taskspec is correct: %v", strings.Join(s, ","))
```
pritidesai, line 310: suggest replacing this to:

```
please check and make sure the results are written in valid JSON escaped format or the result type in taskSpec is correct:
```

pritidesai:
> Hi @pritidesai, is this ok for #6111? I know that in #6111 you're suggesting to do data validation for array and object. I wonder if adding this can also meet the need?

This will definitely add more value to the troubleshooting efforts. I still prefer to introduce data validation if possible, thanks!

pritidesai (approved):

Yongxuanzhang:
> > Hi @pritidesai, is this ok for #6111? I know that in #6111 you're suggesting to do data validation for array and object. I wonder if adding this can also meet the need?
> 
> This will definitely add more value to the troubleshooting efforts. I still prefer to introduce data validation if possible, thanks!

Sure. I will take a look at the data validation. 😄

chitrangpatel (approved):

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -307,7 +307,7 @@ func validateTaskRunResults(tr *v1beta1.TaskRun, resolvedTaskSpec *v1beta1.TaskS
 			s = append(s, fmt.Sprintf(" \"%v\": %v", k, v))
 		}
 		sort.Strings(s)
-		return fmt.Errorf("mismatched types: %v", strings.Join(s, ","))
+		return fmt.Errorf("mismatched types, please check and make sure the results are written in valid JSON escaped format or the result type in taskSpec is correct: %v", strings.Join(s, ","))
```
lbernick, line 310: any way to make this a bit more concise?

Yongxuanzhang: you mean divide the "invalida JSON" and "task not defined"?

Yongxuanzhang: I revisit the code and think you're right. I added a hold to this PR. 
We should try to check them separately as @pritidesai suggested, but I haven't implemented the "Data Validation" yet

Yongxuanzhang: Maybe once we could check the JSON data, we don't need this PR.

lbernick: It would be nice to check "invalid json" and "task results not defined" separately, but I think this is OK to merge if you can just condense it a bit, e.g. "Provided results don't match declared results; may be invalid JSON or missing result declaration"

Yongxuanzhang: sure! Thanks for your suggestion!

Issue #5297, opened 2022-08-10
Pipelines controller removing label from PipelineRun on reconcilation


sm43:
In [Pipelines As Code](https://github.com/openshift-pipelines/pipelines-as-code), we create PipelineRun and then `Update` the PipelineRun with a Label.

Recently we noticed the label is missing, on debugging it more, we found when pipelines controller adds `tekton.dev/pipeline` label on the PipelineRun, it is removing the other label.

```
  --- PipelineRun pac-app-pipelines/pac-app-7-fp9f2 v392837 (2022-08-08T13:53:33+05:30) (gen. 1)
    +++ PipelineRun pac-app-pipelines/pac-app-7-fp9f2 v392871 (2022-08-08T13:53:33+05:30) (gen. 1)
    @@ -32,7 +32,6 @@
  pipelinesascode.tekton.dev/branch: main
  -    pipelinesascode.tekton.dev/check-run-id: "772154487234"
       pipelinesascode.tekton.dev/event-type: pull_request
       pipelinesascode.tekton.dev/git-provider: github
       pipelinesascode.tekton.dev/original-prname: pac-app-7
    @@ -42,9 +41,10 @@
  pipelinesascode.tekton.dev/state: started
  pipelinesascode.tekton.dev/url-org: sm43
  pipelinesascode.tekton.dev/url-repository: pac-app
  +    tekton.dev/pipeline: pac-app-7-fp9f2
    name: pac-app-7-fp9f2
    namespace: pac-app-pipelines
    -  resourceVersion: "392837"
  +  resourceVersion: "392871"
    uid: bcc2a3dc-e9a8-4ba0-9257-2b9308fc2d43
    spec:
      params:
```
How can we avoid this from happening? if this is a race condition

Slack Thread https://tektoncd.slack.com/archives/CLCCEBUMU/p1659949444059689

sm43:
cc @imjasonh @afrittoli @vdemeester

vdemeester:
I think this got fixed be https://github.com/tektoncd/pipeline/pull/5597 and will not even be relevent once https://github.com/tektoncd/pipeline/issues/5146 is fixed, so I will go ahead and close this issue.

PR #6127, opened 2023-02-08
Remove Updating labels and annotations…

vdemeester:
# Changes

Labels and Annotations on PipelineRun and TaskRun won't be mutated by the controller anymore. Usually, a controller doesn't modify the metadata of the object it reconcile.

This effectively remove the propagation of annotations and labels from Task to TaskRun, and Pipeline to PipelineRun.

This is an "attempt" to fix https://github.com/tektoncd/pipeline/issues/5146 **but** it is mainly opened to start some discussion around it. Mainly the following :

One feature that we "want" is that labels and annotations from a `Task` and a `TaskRun` are both inherited by the `Pod`. Same goes for `PipelineRun`. Prior to this change, we get away with it by merging it at the `TaskRun` or `PipelineRun` level — which cause issues in the past (and still might), like https://github.com/tektoncd/pipeline/issues/5297.
Even if this PR propose to remove this "feature", we still want the same behavior (aka a `Pod` gets its labels from both the `Task` and the `TaskRun`). Because we are only "caching" the spec of the `Task` in `TaskRun.status` (and same goes for `PipelineRun`), we would have to query the metadata for the `Task` each time. Using informers, it wouldn't cost that much, but we would run the same possible race as we did before and that's the reason we use `TaskRun.status` and `PipelineRun.status` as source of truth for the spec in all reconciler loop for a given object. 

So the question is the following : **Where should we store the `Task` (and `Pipeline` for `PipelineRun`) labels and annotations in a `TaskRun` object ?** Because we have already release `v1`, we can only do "additive" changes, so, most likely, the only way would be to have a `taskMetadata` field in the `status` in addition to `taskSpec`.
But I am curious if anyone has any other ideas 👼🏼.

/cc @imjasonh @afrittoli @abayer @lbernick @jerop @chuangw6 

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/pipelinerun/pipelinerun.go
+++ b/pkg/reconciler/pipelinerun/pipelinerun.go
@@ -22,7 +22,6 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
-	"reflect"
 	"regexp"
 	"strings"
 
@@ -39,6 +38,7 @@ import (
 	"github.com/tektoncd/pipeline/pkg/internal/affinityassistant"
 	resolutionutil "github.com/tektoncd/pipeline/pkg/internal/resolution"
 	"github.com/tektoncd/pipeline/pkg/pipelinerunmetrics"
+	podconvert "github.com/tektoncd/pipeline/pkg/pod"
 	tknreconciler "github.com/tektoncd/pipeline/pkg/reconciler"
 	"github.com/tektoncd/pipeline/pkg/reconciler/events"
 	"github.com/tektoncd/pipeline/pkg/reconciler/events/cloudevent"
@@ -67,6 +67,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/utils/clock"
 	"knative.dev/pkg/apis"
+	"knative.dev/pkg/changeset"
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/kmap"
 	"knative.dev/pkg/kmeta"
@@ -299,17 +300,12 @@ func (c *Reconciler) durationAndCountMetrics(ctx context.Context, pr *v1.Pipelin
 func (c *Reconciler) finishReconcileUpdateEmitEvents(ctx context.Context, pr *v1.PipelineRun, beforeCondition *apis.Condition, previousError error) error {
 	ctx, span := c.tracerProvider.Tracer(TracerName).Start(ctx, "finishReconcileUpdateEmitEvents")
 	defer span.End()
-	logger := logging.FromContext(ctx)
 
 	afterCondition := pr.Status.GetCondition(apis.ConditionSucceeded)
 	events.Emit(ctx, beforeCondition, afterCondition, pr)
-	_, err := c.updateLabelsAndAnnotations(ctx, pr)
-	if err != nil {
-		logger.Warn("Failed to update PipelineRun labels/annotations", zap.Error(err))
-		events.EmitError(controller.GetEventRecorder(ctx), err, pr)
-	}
 
-	merr := multierror.Append(previousError, err).ErrorOrNil()
+	// Note(vdemeester): This is still there because.. otherwise there is way too many test to rewrite
+	merr := multierror.Append(previousError).ErrorOrNil()
 	if controller.IsPermanentError(previousError) {
 		return controller.NewPermanentError(merr)
 	}
@@ -322,7 +318,8 @@ func (c *Reconciler) resolvePipelineState(
 	ctx context.Context,
 	tasks []v1.PipelineTask,
 	pipelineMeta *metav1.ObjectMeta,
-	pr *v1.PipelineRun) (resources.PipelineRunState, error) {
+	pr *v1.PipelineRun,
+) (resources.PipelineRunState, error) {
 	ctx, span := c.tracerProvider.Tracer(TracerName).Start(ctx, "resolvePipelineState")
 	defer span.End()
 	pst := resources.PipelineRunState{}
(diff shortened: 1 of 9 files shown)


pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -1225,40 +1220,20 @@ func addMetadataByPrecedence(metadata map[string]string, addedMetadata map[strin
...
 			return nil
 		}
 
-		// Propagate labels from Pipeline to PipelineRun. PipelineRun labels take precedences over Pipeline.
-		pr.ObjectMeta.Labels = kmap.Union(meta.Labels, pr.ObjectMeta.Labels)
-		pr.ObjectMeta.Labels[pipeline.PipelineLabelKey] = meta.Name
-
-		// Propagate annotations from Pipeline to PipelineRun. PipelineRun annotations take precedences over Pipeline.
-		pr.ObjectMeta.Annotations = kmap.Union(meta.Annotations, pr.ObjectMeta.Annotations)
+		// // Propagate labels from Pipeline to PipelineRun. PipelineRun labels take precedences over Pipeline.
```
imjasonh, line 1231: These can just be dropped?

imjasonh (commented):
Love it!

/lgtm

pkg/reconciler/pipelinerun/pipelinerun.go
vdemeester: From there yes, overall no as written in the PR description.

> Even if this PR propose to remove this "feature", we still want the same behavior (aka a `Pod` gets its labels from both the `Task` and the `TaskRun`). Because we are only "caching" the spec of the `Task` in `TaskRun.status` (and same goes for `PipelineRun`), we would have to query the metadata for the `Task` each time. Using informers, it wouldn't cost that much, but we would run the same possible race as we did before and that's the reason we use `TaskRun.status` and `PipelineRun.status` as source of truth for the spec in all reconciler loop for a given object. 
> 
> So the question is the following : **Where should we store the `Task` (and `Pipeline` for `PipelineRun`) labels and annotations in a `TaskRun` object ?** Because we have already release `v1`, we can only do "additive" changes, so, most likely, the only way would be to have a `taskMetadata` field in the `status` in addition to `taskSpec`.
But I am curious if anyone has any other ideas 👼🏼.

lbernick:
I'm curious why we can't separate out the labels/annotations propagation from the fix to https://github.com/tektoncd/pipeline/issues/5146? i.e. if we just remove the call to update, and modify the labels of the TR/PR object passed to ReconcileKind, will knative update the object with the new metadata? Or does it only handle status updates?

Also want to better understand the race condition you're referring to. My understanding is that if a Task's labels change, we will re-reconcile the TaskRun, and update the TaskRun with the Task's new labels (and likewise for PipelineRuns). The issue described in #5297 was the PipelineRun controller replacing labels, but we've updated it to merge labels. Can you give an example scenario where the race condition you're describing occurs?

vdemeester:
> I'm curious why we can't separate out the labels/annotations propagation from the fix to #5146? i.e. if we just remove the call to update, and modify the labels of the TR/PR object passed to ReconcileKind, will knative update the object with the new metadata? Or does it only handle status updates?

@lbernick As you "hinted" in one of your question, this is because it only update the status 👼🏼. It doesn't update anything else (metadata or spec). The take on updating only the status is that, users are supposed to interact with `metadata` and `spec` and controllers with `status` (separation of concern, …).

> Also want to better understand the race condition you're referring to. My understanding is that if a Task's labels change, we will re-reconcile the TaskRun, and update the TaskRun with the Task's new labels (and likewise for PipelineRuns). The issue described in https://github.com/tektoncd/pipeline/issues/5297 was the PipelineRun controller replacing labels, but we've updated it to merge labels. Can you give an example scenario where the race condition you're describing occurs?

If a Task label's change, it changes, nothing get's "reconciled" whatsoever. #5297 was not really "removing" things but overriding the content of the labels we "got" at the begining of the reconciler loop, even if, in the meantime, the `PipelineRun` itself was updated.  But this is not the race I am talking about here.
The race I am talking about is if we choose "to query the metadata for the Task each time". If we decide to do that, we would be in the same situation as prior to use `status.taskrunSpec` as a source of truth. At time T, the Task would have labels "a, b and c", we start something and then we are in a new reconciler loop (new event), the labels could have been changed on the `Task` (now only "a and b"), and *this* could be seen as a race. It's less of a problem on metadata than on the spec, but still it could lead to confusing behaviour (if something acts on labels, …).

lbernick:
> @lbernick As you "hinted" in one of your question, this is because it only update the status 👼🏼. It doesn't update anything else (metadata or spec). The take on updating only the status is that, users are supposed to interact with `metadata` and `spec` and controllers with `status` (separation of concern, …). If it did, that PR would be green already (and the code I'm deleting here wouldn't have been required in the first place).

Makes sense! I understand why you wouldn't want a controller to update the object's spec but why would it be bad for the controller to update labels? @imjasonh is there a recommended alternative pattern to use?


> If a Task label's change, it changes, nothing get's "reconciled" whatsoever. 

I'm not sure why I said this 🤦‍♀️  Curious if label changes in general trigger re-reconciles though?

> #5297 was not really "removing" things but overriding the content of the labels we "got" at the begining of the reconciler loop, even if, in the meantime, the `PipelineRun` itself was updated. But this is not the race I am talking about here. The race I am talking about is if we choose "to query the metadata for the Task each time". If we decide to do that, we would be in the same situation as prior to use `status.taskrunSpec` as a source of truth. At time T, the Task would have labels "a, b and c", we start something and then we are in a new reconciler loop (new event), the labels could have been changed on the `Task` (now only "a and b"), and _this_ could be seen as a race. It's less of a problem on metadata than on the spec, but still it could lead to confusing behaviour (if something acts on labels, …).

Makes sense, thanks for the explanation! I'm not sure off the top of my head of any solutions other than adding `status.taskMetadata` or something along those lines but can think a bit more

vdemeester:
> I'm not sure why I said this  Curious if label changes in general trigger re-reconciles though?

It would on the object(s) that are watched, so `TaskRun` or `PipelineRun`. There is no reconciler watching `Task` or `Pipeline`.

vdemeester:
/remove-lifecycle rotten
Still need to rebase and rework a bit that one.. 😅

vdemeester:
Updated this, aside from some tests to be rewritten, it should be ready for review.
This removes the extra update we do during reconcilation, which means we rely solely on knative/pkg for updating the object (see https://github.com/tektoncd/pipeline/issues/5146) and removes one possible "update" call during the loop.

/kind enhancement

jimmyjones2:
Label propagation also causes issues with ArgoCD e.g
https://github.com/argoproj/argo-cd/issues/5792#issuecomment-801858189
https://github.com/argoproj/argo-cd/issues/10982

PR #6134, opened 2023-02-08
Mark config-trusted-resources as deprecated

Yongxuanzhang:
# Changes

This commit marks config-trusted-resources as deprecated, VerificationPolicy has covered the features of config-trusted-resources so there's no need to keep it.

/kind misc

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
action required: config-trusted-resources of trusted resources to store public key is deprecated and will be removed in release 0.46, please use VerificationPolicy instead 
```


pkg/apis/config/trusted_resources.go
```diff
@@ -29,6 +29,7 @@ import (
 // TrustedResources holds the collection of configurations that we attach to contexts.
 // Configmap named with "config-trusted-resources" where cosign pub key path and
 // KMS pub key path can be configured
+// Deprecated.
```
JeromeJu, line 32: Maybe we would like to add 'Use verification policy instead' for pointing users to the alternative.

JeromeJu (approved):
Thanks Yongxuan

JeromeJu:
Maybe we would like to add 'Use verification policy instead' for pointing users to the alternative.
^ I think this might also apply to the release note

pkg/apis/config/trusted_resources.go
Yongxuanzhang: Sure! Thanks for the suggestion!

docs/deprecations.md
```diff
@@ -26,7 +26,7 @@ being deprecated.
...
 | [`pipelineRef.bundle` and `taskRef.bundle` are deprecated](https://github.com/tektoncd/pipeline/issues/5514)                                                                                                                 | v0.41.0                                                              | Alpha                                                                                                  | July 13, 2023                       |
-
+| [`config-trusted-resources` is deprecated](https://github.com/tektoncd/pipeline/issues/5852)                                                                                                                 | v0.45.0                                                              | Alpha                                                                                                  | April 10, 2023 or v0.47.0                       |
```
lbernick, line 29: Do we need to update the alpha features table to explain that this feature is going to be removed? I'm also confused what the "config-trusted-resources" feature actually is. Is that the feature enabled by the flag "resource-verification-mode"? How would someone use VerificationPolicy, if that's the replacement feature?
Lastly, why are we deciding to support this for two more months?

Yongxuanzhang: - This feature stores public keys in config map, it's a subset feature of VerificationPolicy. 
- I will update the docs as @JeromeJu suggested. 
- The announcement is 0.45, are we ok to remove it in 0.46? If so I think we could change it to 0.46

lbernick: sg, please lmk when you've updated the PR to make it more clear how to use the replacement feature. re: when can we remove it? we can remove alpha features at any time according to our compatibility policy, so I'd put N/A (alpha)

pkg/apis/config/trusted_resources.go
lbernick, line 32: I'm surprised you don't need to re-run apigen here, what's going on?

docs/deprecations.md
```diff
@@ -26,7 +26,7 @@ being deprecated.
...
 | [`pipelineRef.bundle` and `taskRef.bundle` are deprecated](https://github.com/tektoncd/pipeline/issues/5514)                                                                                                                 | v0.41.0                                                              | Alpha                                                                                                  | July 13, 2023                       |
-
+| [`config-trusted-resources` is deprecated](https://github.com/tektoncd/pipeline/issues/5852)                                                                                                                 | v0.45.0                                                              | Alpha                                                                                                  | v0.46.0                       |
```
lbernick, line 29:
```suggestion
| [The `config-trusted-resources` configMap is deprecated](https://github.com/tektoncd/pipeline/issues/5852)                                                                                                                 | v0.45.0                                                              | Alpha                                                                                                  | v0.46.0                       |
```

lbernick (approved):

pkg/apis/config/trusted_resources.go
Yongxuanzhang: I think this is config not api in v1beta1 or v1? when I run `update-openapigen.sh` nothing happens

pritidesai:
please add a little more details in the release note, I do no think so but could anyone be using this configmap? Also, how about we also update the doc at https://github.com/tektoncd/pipeline/blob/main/docs/trusted-resources.md#config-key-at-configmap-will-be-deprecated?

Yongxuanzhang:
> please add a little more details in the release note, I do no think so but could anyone be using this configmap? Also, how about we also update the doc at https://github.com/tektoncd/pipeline/blob/main/docs/trusted-resources.md#config-key-at-configmap-will-be-deprecated?

Thanks! I have updated the release notes. For the docs section, I think I have updated? What else am I missing?

pkg/apis/config/trusted_resources.go
lbernick: Ah I see, we only generate deepcopy functions, thanks!

XinruZhang:
It would be great if we can create an issue to track the removal of this ConfigMap :)
Thank you @Yongxuanzhang

PR #6135, opened 2023-02-08
[TEP074] Remove Git PipelineResources

JeromeJu:
This is targeted to be the last PR for the removal of PipelineResources along with the removal of generic pipelineresources functions.

# Changes
This commit removes the `Git` pipelineResources, it removes `github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1/git` and the corresponding test cases.

The coverage drops are expected as it is one of the last few PRs for removing the resources.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
action required: please migrate off of `git` `pipelineresources` as it is removed, please refer to the doc at https://github.com/tektoncd/pipeline/blob/main/docs/pipelineresources.md#replacing-an-git-resource
```


pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -507,11 +413,9 @@ spec:
 	tr1Name := "test-pipeline-run-success-unit-test-1"
 	tr2Name := "test-pipeline-run-success-unit-test-cluster-task"
 
-	verifyTaskRunStatusesCount(t, reconciledRun.Status, 2)
+	// TODO
+	// verifyTaskRunStatusesCount(t, reconciledRun.Status, 2)
```
JeromeJu, line 417: I was trying to verify if this shall be changed to 3

lbernick: It doesn't look like you've changed the number of tasks in the input data (but I could be wrong) so I'd be surprised if this is changed. Does the output of the test align with your expectations, given the input of the test?

JeromeJu: I have been trying to understand https://github.com/tektoncd/pipeline/blob/2da10fc5ff9d99571b6055bcb86aca12805cb9e0/pkg/reconciler/pipelinerun/pipelinerun_test.go#L161 about how the taskruns are fetched from actions, but not sure how to investigate in the ktesting.CreateAction here

lbernick: I'm confused; is the problem with `getTaskRunCreations` on line 375, or with `verifyTaskRunStatusesCount` on line 417?

It looks like `verifyTaskRunStatusesCount` just counts the number of TaskRuns in the PipelineRun.Status.ChildReferences, which shouldn't have changed.

If the problem is with `getTaskRunCreations`, that's a bit more complicated. I don't like the usage of that function in our test suite, as it verifies implementation details (API calls) rather than the outputs we care about (TaskRuns themselves), making the tests harder to understand. We should refactor our tests to not rely on this, but that's out of scope for this PR.

If you want to better understand the Action class, you could try stepping through in a debugger (or use print statements) to figure out what api calls are being made (aka the actions), and if anything has changed that you wouldn't expect to change? Unfortunately, the Action class is undocumented, as is the client-go testing package (https://pkg.go.dev/k8s.io/client-go/testing#Action), but perhaps you could look for example usages of this package to better understand what the actions are.

JeromeJu: Thanks the clarifications Lee. Yes the issue is with `verifyTaskRunStatusesCount `, and it changed to 3 after I deleted some of the resources (which feels wrong). And that's when I looked into how the taskrun status are being created and led me to the confusion at `getTaskRunCreations`.

And appreciate it for the pointers to the doc

lbernick: Can you print the output pipelinerun and make sure it has only the taskruns that are expected? This might be easier than trying to dig into `getTaskRunCreations`.

JeromeJu: Thanks Lee! I figured this out earlier following your suggestions above. 

Sorry haven't been updating this PR yet. I will check again before bring this ready for review, and create an followup issue on the cleanups for `getTaskRunCreations`.

JeromeJu: Openend https://github.com/tektoncd/pipeline/issues/6143 and this shall be ready to review:) Thanks!

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -33,13 +33,6 @@ import (
 	"knative.dev/pkg/apis"
 )
 
-var validResource = v1beta1.TaskResource{
-	ResourceDeclaration: v1beta1.ResourceDeclaration{
-		Name: "validsource",
-		Type: "git",
```
lbernick, line 39: technically it would be better to replace this with the image resource as you've done elsewhere (aligns better with our code standards on always being in a releasable state)

pkg/artifacts/artifact_storage_test.go
```diff
@@ -469,31 +173,6 @@ func TestCleanupArtifactStorage(t *testing.T) {
 	}
 }
 
-func TestInitializeArtifactStorageWithoutConfigMap(t *testing.T) {
```
lbernick, line 472: is the function InitializeArtifactStorage deleted somewhere?

pkg/reconciler/taskrun/resources/apply_test.go
```diff
@@ -1103,19 +1086,6 @@ func TestApplyResources(t *testing.T) {
 			spec.Steps[8].Image = "/foo/builtImage"
 			spec.Steps[9].Image = "/workspace/foo/builtImage"
 		}),
-	}, {
-		name: "input resource specified",
```
lbernick, line 1107: I think this is OK to delete fully because despite the test name it's really a test for the git pipelineresource mutations it looks like

pkg/reconciler/taskrun/resources/input_resource_test.go
```diff
@@ -45,107 +43,16 @@ var (
...
-			Type: "git",
-		}},
-	}
 	gcsInputs = []v1beta1.TaskResource{{
 		ResourceDeclaration: v1beta1.ResourceDeclaration{
 			Name:       "workspace",
 			Type:       "gcs",
 			TargetPath: "gcs-dir",
 		}}}
-	multipleGcsInputs = []v1beta1.TaskResource{
```
lbernick, line 69: should this be restored?

pkg/reconciler/taskrun/resources/input_resource_test.go
```diff
@@ -746,111 +299,6 @@ gsutil cp gs://fake-bucket/rules.zip /workspace/gcs-dir
...
-							ResourceRef: &v1beta1.PipelineResourceRef{
-								Name: "storage-gcs-invalid",
-							},
-							Name: "workspace",
-						},
-					}},
-				},
-			},
-		},
-		wantErr: true,
```
lbernick, line 792: should this be restored?

pkg/reconciler/taskrun/resources/input_resource_test.go
```diff
@@ -1101,242 +549,6 @@ gsutil rsync -d -r gs://fake-bucket/rules.zip /workspace/gcs-input-resource
...
-			Name:      "task-with-targetpath",
-			Namespace: "marshmallow",
-		},
-		Spec: v1beta1.TaskSpec{
-			Resources: &v1beta1.TaskResources{
-				Inputs: gcsInputs,
-			},
-		},
-	}
-	taskWithMultipleGcsInputs := &v1beta1.Task{
```
lbernick, line 1127: curious why this PR deletes a lot of GCS tests?

pkg/reconciler/taskrun/resources/taskresourceresolution_test.go
```diff
@@ -222,29 +209,6 @@ func TestResolveTaskRun_noResources(t *testing.T) {
...
-func TestResolveTaskRun_InvalidBothSpecified(t *testing.T) {
-	inputs := []v1beta1.TaskResourceBinding{{
-		PipelineResourceBinding: v1beta1.PipelineResourceBinding{
-			Name: "repoToBuildFrom",
-			// Can't specify both ResourceRef and ResourceSpec
-			ResourceRef: &v1beta1.PipelineResourceRef{
-				Name: "git-repo",
-			},
-			ResourceSpec: &resourcev1alpha1.PipelineResourceSpec{
-				Type: resourcev1alpha1.PipelineResourceTypeGit,
```
lbernick, line 234: I'd swap to a different resource type here

pkg/apis/pipeline/v1beta1/task_validation_test.go
JeromeJu: Thanks for the suggestion, I was also debating this to see what would be a best solution since we are going to be in a stage where all the resourceTypes are removed before we start to remove the generic resourceTypes functions. Does that mean the removal of the last resourceTypes should also include the removal of the generic functions? But until then we need to change them all to the very last resource type?

JeromeJu: Thanks for the answer offline with Lee.

I am hoping to put this PR as the last removal along with the generic functions of PipelineResources eg. inputresources etc.

lbernick: It might make sense to put a hold on this pr or mark it as a draft in that case

JeromeJu:
/hold
This will be the last removal of pipelineResources along with the generic functions.

JeromeJu:
Closing as https://github.com/tektoncd/pipeline/pull/6150 supersedes this PR

Issue #6071, opened 2023-01-30
PipelineRun reconciler panics for a matrix parameter with an empty array value


abayer:
# Expected Behavior

`Pipeline` validation should fail at creation time when a matrix parameter is set to an empty array.

# Actual Behavior

The `PipelineRun` reconciler will panic when attempting to create the planned tasks for a matrix with an empty array parameter, because the count of matrix combinations becomes 0 (the combination count is generated by multiplying the length of each of the matrix parameters together, and 0...well, yeah). The controller then crashes. The panic itself looks like this (line numbers may be slightly different, 'cos this crash was on a branch of mine):
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1dbebe6]

goroutine 264 [running]:
github.com/tektoncd/pipeline/pkg/reconciler/taskrun.ValidateResolvedTaskResources({0x2899fd0?, 0xc000ccf800?}, {0x0?, 0xc0005581f0?, 0x3?}, 0x3a37e40?, 0x0)
	github.com/tektoncd/pipeline/pkg/reconciler/taskrun/validate_resources.go:212 +0x46
github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun.(*Reconciler).reconcile(0xc0003b5680, {0x2899fd0, 0xc000bcdb60}, 0xc00057e000, 0xc000bcdb60?, 0xc000354af0)
	github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun/pipelinerun.go:599 +0x2c45
github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun.(*Reconciler).ReconcileKind(0xc0003b5680, {0x2899fd0, 0xc000bcda40}, 0xc00057e000)
	github.com/tektoncd/pipeline/pkg/reconciler/pipelinerun/pipelinerun.go:249 +0x161e
github.com/tektoncd/pipeline/pkg/client/injection/reconciler/pipeline/v1beta1/pipelinerun.(*reconcilerImpl).Reconcile(0xc0001e4a00, {0x2899fd0, 0xc000bcd980}, {0xc00012d540, 0x19})
	github.com/tektoncd/pipeline/pkg/client/injection/reconciler/pipeline/v1beta1/pipelinerun/reconciler.go:235 +0x582
```

# Steps to Reproduce the Problem

[...]

Issue #6137, opened 2023-02-09
`timeouts`  is not working as expected. 


VeereshAradhya:
# Expected Behavior
When provided timeouts for pipeline and task in `timeout` section of pipelinerun, both pipelinerun and taskrun timeout should get updated

# Actual Behavior
When provided timeouts for pipeline and task in `timeouts` section of pipelinerun, only the pipelinerun timeout is updated and the taskrun gets created with default timeout

# Steps to Reproduce the Problem

1. Create a task and pipeline
2. Create pipelierun with timeouts section defined

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

```
 Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.4", GitCommit:"872a965c6c6526caa949f0c6ac028ef7aff3fb78", GitTreeState:"archive", BuildDate:"2022-11-11T00:00:00Z", GoVersion:"go1.19.3", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
 Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.6+263df15", GitCommit:"eddac29feb4bb46b99fb570999324e582d761a66", GitTreeState:"clean", BuildDate:"2023-01-23T21:00:20Z", GoVersion:"go1.18.7", Compiler:"gc", Platform:"linux/amd64"}

```

- Tekton Pipeline version:

  **Output of `tkn version` or `kubectl get pods -n tekton-pipelines -l app=tekton-pipelines-controller -o=jsonpath='{.items[0].metadata.labels.version}'`**

```
Client version: 0.28.0
Pipeline version: v0.41.0
Triggers version: v0.22.0
Operator version: v0.63.0
```

```
[varadhya@fedora plumbing]$ oc get tektonconfig config -o yaml | grep timeout 
    default-timeout-minutes: 5
[varadhya@fedora plumbing]$ cat pr.yaml
apiVersion: tekton.dev/v1beta1
kind: PipelineRun
metadata:
  generateName: test-timeout-run-
spec:
  timeouts:
    pipeline: 2h
    tasks: 1h
  pipelineRef:
    name: test-timeout
[varadhya@fedora plumbing]$ oc create -f pr.yaml
pipelinerun.tekton.dev/test-timeout-run-n64qf created
[varadhya@fedora plumbing]$ tkn pr describe -L
Name:              test-timeout-run-n64qf
[...]

PR #6140, opened 2023-02-09
Added Skip Logic: Matrix parameters cannot be empty arrays

EmmaMunley:
# Changes

Added a check in Pipeline validation to verify that matrix parameters are not just arrays, but that they are not empty arrays. This won't change existing functionality/capabilities, but will get rid of the panic in the reconciler. 
This addresses issues: #6071 

/kind bug

# Submitter Checklist

  functionality, content, code)
- N/A Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- N/A Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/matrix.md
+++ b/docs/matrix.md
@@ -188,6 +188,7 @@ Note that:
 - The names of the `Parameters` in the `Matrix` must be unique. Specifying the same parameter multiple times
 will result in a validation error.
 - A `Parameter` can be passed to either the `matrix` or `params` field, not both.
+- If the `Matrix` has an empty array `Parameter`, then the `PipelineTask` will be skipped.
 
 For further details on specifying `Parameters` in the `Pipeline` and passing them to
 `PipelineTasks`, see [documentation](pipelines.md#specifying-parameters).
--- a/examples/v1beta1/pipelineruns/alpha/pipelinerun-with-matrix.yaml
+++ b/examples/v1beta1/pipelineruns/alpha/pipelinerun-with-matrix.yaml
@@ -20,7 +20,7 @@ kind: PipelineRun
 metadata:
   generateName: matrixed-pr-
 spec:
-  serviceAccountName: 'default'
+  serviceAccountName: "default"
   pipelineSpec:
     tasks:
       - name: platforms-and-browsers
@@ -51,3 +51,28 @@ spec:
             value: chrome
         taskRef:
           name: platform-browsers
+      - name: matrix-params-with-empty-array-skipped
+        matrix:
+          params:
+            - name: version
+              value: []
+        taskSpec:
+          params:
+            - name: version
+          steps:
+            - name: echo
+              image: ubuntu
+              script: exit 1
+    finally:
+      - name: matrix-params-with-empty-array-skipped-in-finally
+        matrix:
+          params:
+            - name: version
+              value: []
+        taskSpec:
+          params:
+            - name: version
+          steps:
+            - name: echo
+              image: ubuntu
+              script: exit 1
--- a/pkg/apis/pipeline/v1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1/pipelinerun_types.go
@@ -469,6 +469,8 @@ const (
 	TasksTimedOutSkip SkippingReason = "PipelineRun Tasks timeout has been reached"
 	// FinallyTimedOutSkip means the task was skipped because the PipelineRun has passed its Timeouts.Finally.
 	FinallyTimedOutSkip SkippingReason = "PipelineRun Finally timeout has been reached"
+	// EmptyArrayInMatrixParams means the task was skipped because Matrix parameters contain empty array.
+	EmptyArrayInMatrixParams SkippingReason = "Matrix Parameters have an empty array"
 	// None means the task was not skipped
 	None SkippingReason = "None"
 )
--- a/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
+++ b/pkg/apis/pipeline/v1beta1/pipelinerun_types.go
@@ -495,6 +495,8 @@ const (
 	TasksTimedOutSkip SkippingReason = "PipelineRun Tasks timeout has been reached"
 	// FinallyTimedOutSkip means the task was skipped because the PipelineRun has passed its Timeouts.Finally.
 	FinallyTimedOutSkip SkippingReason = "PipelineRun Finally timeout has been reached"
+	// EmptyArrayInMatrixParams means the task was skipped because Matrix parameters contain empty array.
+	EmptyArrayInMatrixParams SkippingReason = "Matrix Parameters have an empty array"
 	// None means the task was not skipped
 	None SkippingReason = "None"
 )
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
@@ -383,6 +383,8 @@ func (t *ResolvedPipelineTask) skip(facts *PipelineRunFacts) TaskSkipStatus {
 		skippingReason = v1beta1.PipelineTimedOutSkip
 	case t.skipBecausePipelineRunTasksTimeoutReached(facts):
 		skippingReason = v1beta1.TasksTimedOutSkip
+	case t.skipBecauseEmptyArrayInMatrixParams():
+		skippingReason = v1beta1.EmptyArrayInMatrixParams
 	default:
 		skippingReason = v1beta1.None
 	}
@@ -499,6 +501,19 @@ func (t *ResolvedPipelineTask) skipBecausePipelineRunFinallyTimeoutReached(facts
 	return false
 }
 
+// skipBecauseEmptyArrayInMatrixParams returns true if the matrix parameters contain an empty array
+func (t *ResolvedPipelineTask) skipBecauseEmptyArrayInMatrixParams() bool {
+	if t.PipelineTask.IsMatrixed() {
+		for _, ps := range t.PipelineTask.Matrix.Params {
+			if len(ps.Value.ArrayVal) == 0 {
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
 // IsFinalTask returns true if a task is a finally task
 func (t *ResolvedPipelineTask) IsFinalTask(facts *PipelineRunFacts) bool {
 	return facts.isFinalTask(t.PipelineTask.Name)
@@ -521,6 +536,8 @@ func (t *ResolvedPipelineTask) IsFinallySkipped(facts *PipelineRunFacts) TaskSki
 			skippingReason = v1beta1.PipelineTimedOutSkip
 		case t.skipBecausePipelineRunFinallyTimeoutReached(facts):
 			skippingReason = v1beta1.FinallyTimedOutSkip
+		case t.skipBecauseEmptyArrayInMatrixParams():
+			skippingReason = v1beta1.EmptyArrayInMatrixParams
 		default:
 			skippingReason = v1beta1.None
 		}
--- a/pkg/reconciler/taskrun/validate_taskrun.go
+++ b/pkg/reconciler/taskrun/validate_taskrun.go
@@ -33,6 +33,9 @@ import (
 // validateParams validates that all Pipeline Task, Matrix.Params and Matrix.Include parameters all have values, match the specified
 // type and object params have all the keys required
 func validateParams(ctx context.Context, paramSpecs []v1beta1.ParamSpec, params v1beta1.Params, matrixParams v1beta1.Params) error {
+	if paramSpecs == nil {
+		return nil
+	}
 	neededParamsNames, neededParamsTypes := neededParamsNamesAndTypes(paramSpecs)
 	providedParams := params
 	providedParams = append(providedParams, matrixParams...)
@@ -99,6 +102,7 @@ func wrongTypeParamsNames(params []v1beta1.Param, matrix v1beta1.Params, neededP
 			// passed to the task that aren't being used.
 			continue
 		}
+		// Matrix param replacements must be of type String
 		if neededParamsTypes[param.Name] != v1beta1.ParamTypeString {
 			wrongTypeParamNames = append(wrongTypeParamNames, param.Name)
 		}
@@ -160,7 +164,11 @@ func findMissingKeys(neededKeys, providedKeys map[string][]string) map[string][]
 // It also validates that all parameters have values, parameter types match the specified type and
 // object params have all the keys required
 func ValidateResolvedTask(ctx context.Context, params v1beta1.Params, matrix *v1beta1.Matrix, rtr *resources.ResolvedTask) error {
-	if err := validateParams(ctx, rtr.TaskSpec.Params, params, matrix.GetAllParams()); err != nil {
+	var paramSpecs v1beta1.ParamSpecs
+	if rtr != nil {
+		paramSpecs = rtr.TaskSpec.Params
+	}
+	if err := validateParams(ctx, paramSpecs, params, matrix.GetAllParams()); err != nil {
 		return fmt.Errorf("invalid input params for task %s: %w", rtr.TaskName, err)
 	}
 	return nil
--- a/docs/pipeline-api.md
+++ b/docs/pipeline-api.md
@@ -3868,7 +3868,10 @@ SkippingReason
 <th>Description</th>
 </tr>
 </thead>
-<tbody><tr><td><p>&#34;PipelineRun Finally timeout has been reached&#34;</p></td>
+<tbody><tr><td><p>&#34;Matrix Parameters have an empty array&#34;</p></td>
+<td><p>EmptyArrayInMatrixParams means the task was skipped because Matrix parameters contain empty array.</p>
+</td>
+</tr><tr><td><p>&#34;PipelineRun Finally timeout has been reached&#34;</p></td>
 <td><p>FinallyTimedOutSkip means the task was skipped because the PipelineRun has passed its Timeouts.Finally.</p>
 </td>
 </tr><tr><td><p>&#34;PipelineRun was gracefully cancelled&#34;</p></td>
--- a/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
+++ b/pkg/reconciler/pipelinerun/resources/pipelinerunresolution_test.go
@@ -1159,6 +1159,77 @@ func TestIsSkipped(t *testing.T) {
 			"mytask1": false,
 			"mytask2": true,
 		},
+	}, {
+		name: "matrix-params-contain-empty-arr",
+		state: PipelineRunState{{
+			// not skipped no empty arrs
+			PipelineTask: &v1beta1.PipelineTask{
+				Name:    "mytask1",
+				TaskRef: &v1beta1.TaskRef{Name: "matrix-1"},
+				Matrix: &v1beta1.Matrix{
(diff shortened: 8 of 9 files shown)


jerop (commented):
Thank you @EmmaMunley!

- Please update the documentation:  https://github.com/tektoncd/pipeline/blob/main/docs/matrix.md
- You referenced issue https://github.com/tektoncd/pipeline/issues/6137 but it should be https://github.com/tektoncd/pipeline/issues/6071

jerop:
cc @abayer

pritidesai:
thanks @EmmaMunley for the PR, I have updated the PR description with the right issue number.

@abayer and @jerop, please review the comment at https://github.com/tektoncd/pipeline/issues/6071#issuecomment-1424705562

EmmaMunley:
> thanks @EmmaMunley for the PR, I have updated the PR description with the right issue number.
> 
> @abayer and @jerop, please review the comment at [#6071 (comment)](https://github.com/tektoncd/pipeline/issues/6071#issuecomment-1424705562)

Thank you @pritidesai. I've updated the matrix documentation.

docs/matrix.md
```diff
@@ -69,6 +69,7 @@ The `Matrix` will take `Parameters` of type `"array"` only, which will be suppli
 `PipelineTask` by substituting `Parameters` of type `"string"` in the underlying `Task`.
 The names of the `Parameters` in the `Matrix` must match the names of the `Parameters`
 in the underlying `Task` that they will be substituting.
+Note: The `Paramaters` of type `array` cannot be empty.
```
pritidesai, line 72: thank you @EmmaMunley for updating the doc, I appreciate it but this is not my expected behavior is. 

As a user, I would want to:

1. With a single matrix param - `skip` the pipelineTask if matrix param is an empty array. 
2. With multiple matrix params - If all params are empty array, follow (1) else create param combination with non-empty matrix params for the underlying pipelinesTask.

There is a distinction between an empty array and uninitialized array. Uninitialized array is treated as an error but empty array could have been chosen by design. 

@jerop thoughts?

jerop: @pritidesai do you have an example or use case for why an empty array should be valid? tried to think of something but couldn't come up with any because I assume that a value has to be passed to the taskrun for that param

also, if there's no non-empty array, I'd expect `failure`, instead of `skipping`, because we can't create a taskrun from that `pipelinetask`

pritidesai: thank you @jerop!

We had a discussion in the [API WG](https://docs.google.com/document/d/1XTN44T_AzAXDRHS6relY-Au2DzoaHSu2MpT0UkH01TA/edit#heading=h.ah8u3qqpv18t) with @EmmaMunley and @afrittoli on this topic.

Use case: A pipeline is generating SBOMs using many different tools and uploading them to an object store. Here, a list of SBOMs is a matrix param. It is possible that none of the tools produced SBOM for some valid reason. As a pipeline author, I would like to continue with the rest of the pipeline instead of failing the entire run. Generating these SBOMs could be just a part of one pipeline with other tasks. As a pipeline author,

One more use case I documented in the WG notes with the list of images. As a pipeline author, test the list of images which were published based on some criteria. It is possible that there were no images published. In that case, the pipeline will not run tests but will continue running the rest of the tasks of why the images failed to meet the required criteria.

Failure is fatal and will cause the entire pipeline to fail. The pipeline could have spent hours on a list of tasks. 

>also, if there's no non-empty array, I'd expect failure, instead of skipping, because we can't create a taskrun from that pipelinetask

Please correct me if I am wrong but skipping anyways does not create a `taskRun`.

Also, this might need a little more discussion.

Now, the whole purpose of `matrix` is to generate exhaustive list of combinations for a given set of parameters.

In case of a matrix with one param and that one param being empty, `matrix` generates 0 combination. So, the natural choice is to skip the pipelineTask `my-matrix-task` because pipeline controller does not have enough information to decide whether to create one taskRun or 10 taskRuns and what should be the value of that param `param1`. Now, this param `param1` could be  associated to optional or required param of the task. We could run the underlying task once with a default value of that param. And in case of a required param, fail the pipelineRun since the required param is not initialized. But its contracting `matrix` contract of initiating taskRuns based on the number of combinations of matrix params.  

```
  - name: my-matrix-task
    matrix:
      params:
        - name: param1
          value: "[]"
```

Therefore, the best way forward could be skip creating a taskRun and therefore declare that pipelineTask as skipped. 

But for a matrix with more than one params and one of the params is an empty array:

```
  - name: my-matrix-task
    matrix:
      params:
        - name: param1
          value: "[a, b, c]"
        - name: param2
          value: "[]"
```

This `matrix` can result in three combinations with `param2` as an empty string for all three combinations or pick a default value from the task for this param. Or even when a `matrix` resulted in three combinations, skip the pipelineTask `my-matrix-task` because one of the matrix params was an empty array.

And, both matrix params set to an empty array:

```
  - name: my-matrix-task
    matrix:
      params:
        - name: param1
          value: "[]"
        - name: param2
          value: "[]"
```

This article has a few examples on `matrix` configuration based on some use cases: https://ncorti.com/blog/howto-github-actions-build-matrix

To summarize, `matrix` must not cause a pipelineRun to fail when a matrix param is set to an empty array. In case of a `matrix` without `include` from [TEP-0118](https://github.com/tektoncd/community/blob/main/teps/0118-matrix-with-explicit-combinations-of-parameters.md), skip if any one of the matrix param is an empty array.

I am putting on hold until we can come to conclusion (hope very soon)

/hold

afrittoli: Thanks, @pritidesai - it seems to me that there are two levels to the issue, and I think the decision for both should be consistent regardless of the number of matrixed params and number of zero length arrays

1.  Are empty arrays in a matrix valid? i.e. `PipelineRun` continues running or not in case of empty arrays in matrix. Because we want to support dynamic arrays in future, I think it would be preferable *not* to fail, i.e. consider an empty array a valid alternative
2. Does the `PipelineRun` controller create `TaskRuns` in case of empty arrays in matrix? If one or more dimensions are empty the controller could still create `TaskRuns` without the specific parameter and let the `TaskRun` controller decide whether to run it (if there is a default set) or not. This could be a desirable behaviour in some cases, but I feel like it could be confusing for users, so my recommendation would be to skip if any of the matrixed arrays are empty

In case of skip, we should use apply the same logic that we apply for `when` expressions - the default skip logic would be to continue with following tasks, unless the skip logic is set to branch.

I'm not sure if we support results from a task in a matrix today, but if we do, the resulting array of results should be empty, so that the following tasks could decide how to react based on the size. 
I believe we don't support the `${#array}` syntax yet to evaluate the size of an array, I believe that would be a nice addition for these cases.

jerop: thanks for sharing the use cases and example scenarios @pritidesai!

agree with @afrittoli about consistency - let's start by skipping the pipelinetask if any param in a matrix is empty...we can revisit the idea of fanning out using non-empty params only (and ignoring the empty params) in the future so that we can keep things simple for now, what do you think @pritidesai?

@afrittoli - we do not yet support results from a matrixed pipelinetask - [TEP-0090](https://github.com/tektoncd/community/blob/main/teps/0090-matrix.md#results-from-fanned-out-pipelinetasks )

jerop: Discussed with @pritidesai at API WG on 3/6 and agreed to move forward with @afrittoli's suggestion that we skip the PipelineTask if there's an empty array in the Matrix -- cc @EmmaMunley

pritidesai: thanks a bunch @jerop @afrittoli and others for all the input! Very excited to see we have come to the conclusion 🎉 

To summarize:

The initial proposal was to `fail` the `pipelineRun` when a `matrix.params` has an empty array. My initial suggestion was to `skip` in two scenarios (1) a single `matrix.params` set to empty array (2) all specified `matrix.params` are set to empty array but continue execution in third scenario (3) all specified `matrix.params` are not empty i.e. some of them are initialized to a valid array. @afrittoli suggested to keep it consistent and `skip` in all three scenarios (thank you @afrittoli for the great suggestion).

@EmmaMunley please feel free to ping me if you have any questions. I will be very happy to guide you further.

pkg/apis/pipeline/v1/pipelinerun_types.go
```diff
@@ -470,6 +470,8 @@ const (
 	// FinallyTimedOutSkip means the task was skipped because the PipelineRun has passed its Timeouts.Finally.
 	FinallyTimedOutSkip SkippingReason = "PipelineRun Finally timeout has been reached"
 	// None means the task was not skipped
+	MatrixParamsContainEmptyArrSkip SkippingReason = "PipelineRun tasks were skipped because matrix parameters contain empty arr"
```
pritidesai, line 473: at first I thought there is a typo in `Arr` in `MatrixParamsContainEmptyArrSkip` 🙃 

NIT: Suggest to keep it simple, `EmptyArrayInMatrixParams` and set it to `"PipelineRun tasks were skipped because matrix parameters contain empty array"`

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -388,6 +388,8 @@ func (t *ResolvedPipelineTask) skip(facts *PipelineRunFacts) TaskSkipStatus {
 		skippingReason = v1beta1.PipelineTimedOutSkip
 	case t.skipBecausePipelineRunTasksTimeoutReached(facts):
 		skippingReason = v1beta1.TasksTimedOutSkip
+	case t.skipBecauseMatrixParamsContainEmptyArr(facts):
+		skippingReason = v1beta1.TasksTimedOutSkip
```
pritidesai, line 392: `skippingReason` is set to wrong reason here, set to `TasksTimedOutSkip` instead of `MatrixParamsContainEmptyArrSkip`

pkg/apis/pipeline/v1/pipelinerun_types.go
EmmaMunley: Updated naming

pkg/apis/pipeline/v1/pipelinerun_types.go
```diff
@@ -469,6 +469,8 @@ const (
 	TasksTimedOutSkip SkippingReason = "PipelineRun Tasks timeout has been reached"
 	// FinallyTimedOutSkip means the task was skipped because the PipelineRun has passed its Timeouts.Finally.
 	FinallyTimedOutSkip SkippingReason = "PipelineRun Finally timeout has been reached"
+	// EmptyArrayInMatrixParams means the task was skipped because Matrix parameters contain empty array.
+	EmptyArrayInMatrixParams SkippingReason = "PipelineRun tasks were skipped because matrix parameters contain empty array"
```
jerop, line 473:
```suggestion
	EmptyArrayInMatrixParams SkippingReason = "Matrix Parameters have an empty array"
```

pkg/apis/pipeline/v1beta1/pipelinerun_types.go
```diff
@@ -479,6 +479,8 @@ const (
 	TasksTimedOutSkip SkippingReason = "PipelineRun Tasks timeout has been reached"
 	// FinallyTimedOutSkip means the task was skipped because the PipelineRun has passed its Timeouts.Finally.
 	FinallyTimedOutSkip SkippingReason = "PipelineRun Finally timeout has been reached"
+	// EmptyArrayInMatrixParams means the task was skipped because Matrix parameters contain empty array.
+	EmptyArrayInMatrixParams SkippingReason = "PipelineRun tasks were skipped because matrix parameters contain empty array"
```
jerop, line 483:
```suggestion
	EmptyArrayInMatrixParams SkippingReason = "Matrix Parameters have an empty array"
```

pkg/apis/pipeline/v1/pipelinerun_types.go
jerop: the previous reasons reference "pipelinerun tasks / pipelinerun finally" because they are talking about the timeouts set in pipelineruns

jerop: see `MissingResultsSkip` or `WhenExpressionsSkip` for other examples

EmmaMunley: Ah I see. Thanks!

pritidesai:
thanks @EmmaMunley for resolving all the comments 👍 Can I please request to add an example as well? A similar example to check if the task is skipped when a matrix param is an empty array - https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/pipelinerun-with-when-expressions.yaml#L83

examples/v1beta1/pipelineruns/alpha/pipelinerun-with-matrix.yaml
```diff
@@ -51,3 +51,8 @@ spec:
             value: chrome
         taskRef:
           name: platform-browsers
+      - name: matrix-with-skipped-tasks # matrix parameters cannot have empty array values
+        matrix:
+          params:
+            - name: platform
+              value: '[]'
```
jerop, line 58: this is not a fully formed pipelinetask - needs a taskspec or taskref

added a taskspec, and implemented it such that this task should fail if it's executed (to ensure that it is always skipped)
```suggestion
      - name: matrix-params-with-empty-array-skipped
        matrix:
          params:
            - name: versions
              value: []
        taskSpec:
          params:
            - name: version
          steps:
            - name: echo
              image: ubuntu
              script: exit 1
```

also, in another PR, let's add the same examples under finally 🙏🏾

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -388,6 +388,8 @@ func (t *ResolvedPipelineTask) skip(facts *PipelineRunFacts) TaskSkipStatus {
 		skippingReason = v1beta1.PipelineTimedOutSkip
 	case t.skipBecausePipelineRunTasksTimeoutReached(facts):
 		skippingReason = v1beta1.TasksTimedOutSkip
+	case t.skipBecauseEmptyArrayInMatrixParams(facts):
+		skippingReason = v1beta1.EmptyArrayInMatrixParams
```
jerop, line 392: a similar check needs to be added to [IsFinallySkipped](https://github.com/tektoncd/pipeline/blob/57577006b054c3bc45f3f668dfba890533322100/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L513)

jerop (commented):
also please update the commit message and PR description (that we're skipping, not validating)

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
EmmaMunley: Added to [IsFinallySkipped](https://github.com/tektoncd/pipeline/blob/57577006b054c3bc45f3f668dfba890533322100/pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go#L513) as well

jerop:
@EmmaMunley please run hack/update-codegen.sh

```Diffing /home/prow/go/src/github.com/tektoncd/pipeline against freshly generated codegen
--- /home/prow/go/src/github.com/tektoncd/pipeline/docs/pipeline-api.md	2023-03-13 18:37:50.472529682 +0000
+++ /home/prow/go/src/github.com/tektoncd/pipeline/tmpdiffroot.ShZ0o8/docs/pipeline-api.md	2023-03-13 18:28:51.567470101 +0000
@@ -3811,10 +3811,7 @@ SkippingReason
 <th>Description</th>
 </tr>
 </thead>
-<tbody><tr><td><p>&#34;Matrix Parameters have an empty array&#34;</p></td>
-<td><p>EmptyArrayInMatrixParams means the task was skipped because Matrix parameters contain empty array.</p>
-</td>
-</tr><tr><td><p>&#34;PipelineRun Finally timeout has been reached&#34;</p></td>
+<tbody><tr><td><p>&#34;PipelineRun Finally timeout has been reached&#34;</p></td>
 <td><p>FinallyTimedOutSkip means the task was skipped because the PipelineRun has passed its Timeouts.Finally.</p>
 </td>
 </tr><tr><td><p>&#34;PipelineRun was gracefully cancelled&#34;</p></td>
/home/prow/go/src/github.com/tektoncd/pipeline is out of date. Please run hack/update-codegen.sh.
```

EmmaMunley:
> please add docs for this in matrix.md

@jerop Docs have been updated

docs/matrix.md
```diff
@@ -70,11 +70,11 @@ For more information, see [installation customizations](install.md#customizing-b
 
 The `Matrix` will take `Parameters` of type `"array"` only, which will be supplied to the
 `PipelineTask` by substituting `Parameters` of type `"string"` in the underlying `Task`.
-The names of the `Parameters` in the `Matrix` must match the names of the `Parameters`
+A `Parameter`that contains an empty array is considered invalid. The names of the
+`Parameters` in the `Matrix` must match the names of the `Parameters`
```
jerop, line 74: it is invalid implies that it'd fail, let's make it clear that it'd be skipped

```suggestion
If the `Matrix` has an empty array `Parameter`, then the `PipelineTask` will be skipped.
The names of the `Parameters` in the `Matrix` must match the names of the `Parameters`
```

pkg/reconciler/pipelinerun/resources/pipelinerunresolution.go
```diff
@@ -504,6 +506,19 @@ func (t *ResolvedPipelineTask) skipBecausePipelineRunFinallyTimeoutReached(facts
 	return false
 }
 
+// skipBecauseEmptyArrayInMatrixParams returns true if the matrix parameters contain an empty array
+func (t *ResolvedPipelineTask) skipBecauseEmptyArrayInMatrixParams(facts *PipelineRunFacts) bool {
```
Yongxuanzhang, line 510: Is the `facts` used in this function? Maybe we can remove it if not used.

EmmaMunley: Yes, good catch. Thanks!

pritidesai:
This PR is actively reviewed and updated. It did not make it to 0.46, moving it to 0.47.

dibyom:
Overall LGTM! Just one question around adding another unit test

pkg/reconciler/taskrun/validate_taskrun.go
```diff
@@ -159,8 +162,12 @@ func findMissingKeys(neededKeys, providedKeys map[string][]string) map[string][]
 // ValidateResolvedTask validates that all parameters declared in the TaskSpec are present in the taskrun
 // It also validates that all parameters have values, parameter types match the specified type and
 // object params have all the keys required
-func ValidateResolvedTask(ctx context.Context, params []v1beta1.Param, matrix *v1beta1.Matrix, rtr *resources.ResolvedTask) error {
-	if err := validateParams(ctx, rtr.TaskSpec.Params, params, matrix.GetAllParams()); err != nil {
+func ValidateResolvedTask(ctx context.Context, params v1beta1.Params, matrix *v1beta1.Matrix, rtr *resources.ResolvedTask) error {
```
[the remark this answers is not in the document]
EmmaMunley: @dibyom Sorry, disregard. The existing logic is correct. I was confused why we were hardcoding the type string, but it's because it is checking the type for string replacements not the param type itself. Added a comment to clarify why we are checking type string.

dibyom (approved):

jerop (commented):
/lgtm

PR #6014, opened 2023-01-20
[TEP074] Remove Storage pipelineResources

JeromeJu:
# Changes

Remove Storage PipelineResources

This commit removes the Storage Resources support.
This PR removes `github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1/storage`,
`pkg/artifacts` and the respective test cases.

# Submitter Checklist

As the author of this PR, please check off the items in this checklist:

- [x] Has [Docs](https://github.com/tektoncd/community/blob/main/standards.md#docs) included if any changes are user facing
- [x] Has [Tests](https://github.com/tektoncd/community/blob/main/standards.md#tests) included if any functionality added or changed
- [x] Follows the [commit message standard](https://github.com/tektoncd/community/blob/main/standards.md#commits)
- [x] Meets the [Tekton contributor standards](https://github.com/tektoncd/community/blob/main/standards.md) (including
  functionality, content, code)
- [x] Has a kind label. You can add one by adding a comment on this PR that contains `/kind <type>`. Valid types are bug, cleanup, design, documentation, feature, flake, misc, question, tep
- [x] Release notes block below has been updated with any user facing changes (API changes, bug fixes, changes requiring upgrade notices or deprecation warnings)
- [x] Release notes contains the string "action required" if the change requires additional action from users switching to the new release

# Release Notes

```release-note
action required: please migrate off of `storage` `pipelineresources` as it is removed, please refer to the doc at https://github.com/tektoncd/pipeline/blob/main/docs/pipelineresources.md#replacing-an-storage-resource
`artifacts` apis are removed with the setup for pvc/buckets
```

PR #6150, opened 2023-02-12
[TEP074] Remove Storage, Git and Generic PipelineResources

JeromeJu:
_This is going to be the last PR for PipelineResources removal._

# Changes

This commit removes the Git, Storage and Generic PipelineResources:

- removes the Storage resources support at `pkg/apis/resource/v1alpha1/storage` API
- removes the Git resources support at `pkg/apis/resource/v1alpha1/git` API
- removes the generic logics for input and output resources
- removes the `pkg/artifacts` API
- their corresponding unit test cases, examples and integration tests
- renames:
  - ResolvedTaskResources Struct renamed to ResolvedTask
  - `pkg/reconciler/taskrun/validate_resources.go` to `pkg/reconciler/taskrun/validate_taskrun.go`
  - `pkg/reconciler/taskrun/resources/taskresourceresolution.go` to `pkg/reconciler/taskrun/resources/taskresolution.go`

Reasons why removing `Git`, `Storage` Resources and`pkg/artifacts` together:
- Why bundle Git and Storage resources removal?
  The Git and Storage PipelineResources both uses artifact storage, which made removing Storage PipelineResources along to be intervened by changes on Git Resources that are inappropriate to separate. See [the decision made after attempting to remove Storage Resources along](https://github.com/tektoncd/pipeline/pull/6014#discussion_r1104486051) for more detail.
- Why bundle removal of Storage resources with the `artifacts` package?
  The artifacts package has also [dependencies on storage resources for the artifacts storage setup](https://github.com/tektoncd/pipeline/blob/505d1cf57c627c705f3c953ba849fcafebad8d19/pkg/artifacts/artifacts_storage.go#L28), which made it hard to separate the removal of storage resources first. See [the decision after the attempt to remove artifacts package first](https://github.com/tektoncd/pipeline/pull/6096#issuecomment-1412412615).
 
 ---
 * The coverage loss is expected since the `pipelineResources`.

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
action required: please migrate off of `git` and `storage` `pipelineresources` as they are removed, please refer to the doc at https://github.com/tektoncd/pipeline/blob/main/docs/pipelineresources.md#replacing-an-git-resource  https://github.com/tektoncd/pipeline/blob/main/docs/pipelineresources.md#replacing-an-storage-resource

installing this release on a cluster with tasks and pipelines that use pipeline resources may break them

pipelineresources are removed in this release
```


JeromeJu:
git tag -d ok-to-test

pkg/apis/pipeline/v1beta1/task_validation_test.go
```diff
@@ -33,13 +33,6 @@ import (
 	"knative.dev/pkg/apis"
 )
 
-var validResource = v1beta1.TaskResource{
-	ResourceDeclaration: v1beta1.ResourceDeclaration{
-		Name: "validsource",
-		Type: "git",
```
lbernick, line 39: can you change this to an image resource instead of removing these tests?

pkg/apis/resource/v1alpha1/pipelineresource_validation_test.go
```diff
@@ -32,86 +32,26 @@ func TestResourceValidation_Invalid(t *testing.T) {
...
-					Type: "not-supported",
-				},
-			},
-			want: apis.ErrInvalidValue("spec.type", "not-supported"),
-		}, {
-			name: "missing spec",
-			res: &v1alpha1.PipelineResource{
-				ObjectMeta: metav1.ObjectMeta{
-					Name: "temp",
-				},
```
lbernick, line 110: let's keep these test cases (missing spec and invalid type) around

docs/variables.md
```diff
@@ -108,14 +93,6 @@ variable via `resources.inputs.<resourceName>.<variableName>` or
...
 
-#### Variables for the `GCS` type
-
-| Variable | Description |
-| -------- | ----------- |
-| `name` | The name of the resource. |
-| `type` | Type value of `"gcs"`. |
-| `location` | The fully qualified address of the blob storage. |
-
 #### Variables for the `Cluster` type
```
lbernick, line 96: Can you please open a separate PR removing the docs for cluster and cloudevent resources?

pkg/pipelinerunmetrics/metrics_test.go
```diff
@@ -78,7 +78,7 @@ func TestMetricsOnStore(t *testing.T) {
 	}
 
 	// We check that there's no change when incorrect config is passed
-	MetricsOnStore(logger)(config.GetMetricsConfigName(), &config.ArtifactBucket{})
+	MetricsOnStore(logger)(config.GetMetricsConfigName(), &config.Store{})
```
lbernick, line 81: just curious what's config.Store and how is it different from config.ArtifactBucket?

pkg/reconciler/pipelinerun/resources/input_output_steps.go
```diff
@@ -24,15 +24,15 @@ import (
...
+func GetOutputSteps(outputs map[string]*resourcev1alpha1.PipelineResource, taskName string) []v1beta1.TaskResourceBinding {
 	var taskOutputResources []v1beta1.TaskResourceBinding
 
 	for name, outputResource := range outputs {
 		taskOutputResource := v1beta1.TaskResourceBinding{
 			PipelineResourceBinding: v1beta1.PipelineResourceBinding{
 				Name: name,
 			},
-			Paths: []string{filepath.Join(storageBasePath, taskName, name)},
+			Paths: []string{filepath.Join(taskName, name)},
```
lbernick, line 35: This doesn't look like it would produce a valid file path. Is this function still needed? It seems like git and storage are the only remaining output resources

pkg/apis/resource/v1alpha1/pipelineresource_validation_test.go
JeromeJu: sorry i think the (missing spec and invalid type) two are retained, these changes might have made it confusing after rebasing and merging multiple PRs from branches:/

pkg/apis/pipeline/v1beta1/task_validation_test.go
JeromeJu: Thanks for catching!

Sorry that some of the restoring of git type test cases are removed during multi-state rebasing.

docs/variables.md
JeromeJu: Appreciate the catch, updated at https://github.com/tektoncd/pipeline/pull/6158

docs/resources.md
```diff
@@ -290,171 +271,6 @@ You can refer to different examples demonstrating usage of optional resources in
 
 ## Resource Types
 
-### Git Resource
```
pritidesai, line 293: There are a few other places in this doc where the instructions refer to `git PipelineResource`. Its not a blocker in merging this PR but at this point, `resources.md` does not bring much value. If we are not too far from deleting `resources.md`, its reasonable to make these changes.

examples/v1beta1/pipelineruns/output-pipelinerun.yaml
```diff
@@ -1,106 +0,0 @@
-# `PipelineResources` are deprecated, consider using `Tasks` and other replacement features instead
-# https://github.com/tektoncd/pipeline/blob/main/docs/migrating-v1alpha1-to-v1beta1.md#replacing-pipelineresources-with-tasks
-apiVersion: tekton.dev/v1alpha1
-kind: PipelineResource
```
pritidesai, line 4: In addition to a git resource, this example demonstrates an ability to write to workspace and then read the content written before. Please make sure there is a similar example included as part of our CI before removing it. Or convert this to not include git pipelineresource.

JeromeJu: Thanks @pritidesai , I think the concern does make much sense. I’ve taken another look at output-pipeline example and I think it mainly targeted to show that how the output resources is being used.  As for the ability to write to the workspace and to use it afterwards I think the [pipelinerun.yaml](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/pipelinerun.yaml#L100-L118) could cover that and the git resolver used here shall correspond to the replacement of our git resources according to the guide https://github.com/tektoncd/pipeline/blob/main/docs/pipelineresources.md.

pkg/reconciler/pipelinerun/resources/input_output_steps.go
```diff
@@ -17,45 +17,13 @@ limitations under the License.
...
-				Params:       outputResource.Spec.Params,
-				SecretParams: outputResource.Spec.SecretParams,
-			}
-		}
-		taskOutputResources = append(taskOutputResources, taskOutputResource)
-	}
-	return taskOutputResources
-}
-
 // GetInputSteps will add the correct `path` to the input resources for pt. If the resources are provided by
```
lbernick, line 24: Suggest removing the line about resources being provided by previous tasks, since that functionality is being removed.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -744,18 +744,6 @@ func (c *Reconciler) createPod(ctx context.Context, ts *v1beta1.TaskSpec, tr *v1
 		return nil, err
 	}
 
-	ts, err = resources.AddInputResource(ctx, c.KubeClientSet, c.Images, rtr.TaskName, ts, tr, inputResources)
-	if err != nil {
-		logger.Errorf("Failed to create a pod for taskrun: %s due to input resource error %v", tr.Name, err)
-		return nil, err
-	}
-
```
lbernick, line 752: why is this removed? I don't think it should be

lbernick (approved):

pkg/reconciler/taskrun/taskrun.go
JeromeJu: Thanks Lee. Restored as we still have image resources to be removed but can be an input.

pkg/reconciler/taskrun/taskrun.go
```diff
@@ -743,19 +743,12 @@ func (c *Reconciler) createPod(ctx context.Context, ts *v1beta1.TaskSpec, tr *v1
...
 		return nil, err
 	}
-
 	ts, err = resources.AddInputResource(ctx, c.KubeClientSet, c.Images, rtr.TaskName, ts, tr, inputResources)
 	if err != nil {
 		logger.Errorf("Failed to create a pod for taskrun: %s due to input resource error %v", tr.Name, err)
 		return nil, err
 	}
 
-	ts, err = resources.AddOutputResources(ctx, c.KubeClientSet, c.Images, rtr.TaskName, ts, tr, outputResources)
```
pritidesai, line 753: why are we deleting `AddOutputResources` but not `AddInputResource`?At the same time, `ApplyResources` on L761 ⬇️ is as is.

pkg/reconciler/taskrun/validate_resources_test.go
```diff
@@ -82,28 +66,7 @@ func TestValidateResolvedTaskResources_ValidResources(t *testing.T) {
...
-				Spec: resourcev1alpha1.PipelineResourceSpec{
-					Type: resourcev1alpha1.PipelineResourceTypeGit,
-					Params: []v1beta1.ResourceParam{{
-						Name:  "foo",
-						Value: "bar",
-					}},
-				},
-			},
-		},
+		Inputs:   map[string]*resourcev1alpha1.PipelineResource{},
```
pritidesai, line 69: this is not necessary 🙃

pkg/apis/resource/resource.go
```diff
@@ -22,22 +22,15 @@ import (
...
-	"github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1/storage"
 )
 
 // FromType returns an instance of the correct PipelineResource object type which can be
 // used to add input and output containers as well as volumes to a TaskRun's pod in order to realize
 // a PipelineResource in a pod.
 func FromType(name string, r *resourcev1alpha1.PipelineResource, images pipeline.Images) (pipelinev1beta1.PipelineResourceInterface, error) {
-	switch r.Spec.Type {
-	case resourcev1alpha1.PipelineResourceTypeGit:
-		return git.NewResource(name, images.GitImage, r)
```
imjasonh, line 36: Does removing this mean we can remove all references to `GitImage`, including the `--git-image` flag passed to the controller in config/controller.yaml?

pkg/apis/resource/v1alpha1/storage/gcs.go
```diff
@@ -1,172 +0,0 @@
...
-		return nil, fmt.Errorf("GCSResource: Need Location to be specified in order to create GCS resource %s", r.Name)
-	}
-	return &GCSResource{
-		Name:        name,
-		Type:        r.Spec.Type,
-		Location:    location,
-		TypeDir:     dir,
-		Secrets:     r.Spec.SecretParams,
-		ShellImage:  images.ShellImage,
-		GsutilImage: images.GsutilImage,
```
imjasonh, line 85: Removing this might also let us remove `GsutilImage`, all the way up to the flag in controller.yaml.

JeromeJu: Thanks Jason for catching this. Removed 😺

pkg/reconciler/taskrun/taskrun.go
JeromeJu: I think the reason is that we've removed all the allowedOutput types here in this PR but still Image resource could be the left input resources according to https://github.com/tektoncd/pipeline/pull/6150#discussion_r1105034369. But these should be all targeted to be removed in https://github.com/tektoncd/pipeline/pull/6167. 

Sorry for the confusion it might led to in this PR. I would love to follow if there're route that could make this more explicit/ less confusing 😇

pkg/reconciler/taskrun/validate_resources_test.go
JeromeJu: Thanks Priti:)

pkg/apis/resource/resource.go
JeromeJu: Thanks @imjasonh .  We might want to move them to catalog and leave it here for now according to the updated plans at [tep074](https://github.com/tektoncd/community/commit/5f65433ebd7869dcfbd8c4f0db2d0cfe6331d8c6)

docs/additional-configs.md
```diff
@@ -82,9 +82,6 @@ use a `ConfigMap` with the name `config-artifact-bucket` and the following attri
 
 **Important:** Configure your bucket's retention policy to delete all files after your `Tasks` finish running.
 
-**Note:** You can only use an S3 bucket located in the `us-east-1` region. This is a limitation of [`gsutil`](https://cloud.google.com/storage/docs/gsutil) running a `boto` configuration behind the scenes to access the S3 bucket.
```
afrittoli, line 85: Why removing this note?

JeromeJu: Thanks for catching, I think instead this whole section shall be removed for `##Configuring PipelineResource storage`

config/config-artifact-bucket.yaml
```diff
@@ -1,33 +0,0 @@
-# Copyright 2019 The Tekton Authors
```
afrittoli, line 1: config-artifact-pvc.yaml should go too

examples/v1beta1/taskruns/task-output-image.yaml
```diff
@@ -1,97 +0,0 @@
-# `PipelineResources` are deprecated, consider using `Tasks` and other replacement features instead
-# https://github.com/tektoncd/pipeline/blob/main/docs/migrating-v1alpha1-to-v1beta1.md#replacing-pipelineresources-with-tasks
-apiVersion: tekton.dev/v1alpha1
-kind: PipelineResource
-metadata:
-  name: skaffold-image-leeroy-web-output-image
-spec:
-  type: image
```
afrittoli, line 8: Image pipeline resources are not really removed yet, but I guess they'll go away soon enough, so probably ok to remove this... I would have preferred a neat separation though

examples/v1beta1/taskruns/task-multiple-output-image.yaml
```diff
@@ -1,124 +0,0 @@
-# `PipelineResources` are deprecated, consider using `Tasks` and other replacement features instead
-# https://github.com/tektoncd/pipeline/blob/main/docs/migrating-v1alpha1-to-v1beta1.md#replacing-pipelineresources-with-tasks
```
afrittoli, line 2: ditto about image resources

examples/v1beta1/taskruns/optional-resources.yaml
```diff
@@ -1,133 +0,0 @@
-apiVersion: tekton.dev/v1beta1
-kind: Task
-metadata:
-  name: task-check-optional-resources
```
afrittoli, line 4: optional resources have not been removed yet, but since all resources are going away it's not worth migrating this test

config/config-artifact-bucket.yaml
JeromeJu: Thanks @afrittoli removed config-artifact-pvc.yaml

JeromeJu:
Replying to https://github.com/tektoncd/pipeline/pull/6150#discussion_r1108699137, https://github.com/tektoncd/pipeline/pull/6150#discussion_r1108698448, https://github.com/tektoncd/pipeline/pull/6150#discussion_r1108697583 since they are about the examples removal.

Thanks @afrittoli for being considerately on covering the functionalities of the examples that are removed accordingly.

I tried to refactor(separate) those 2 examples (output-pipeline) but there doesn’t seem an easy way to replace git resources with resolver and then separate out the image resources part.  https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/pipelinerun.yaml which uses the image digest could cover such intended functionalities: 
- Replacing the image resources: using [results of image digest and url](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/pipelinerun.yaml#L60-L62)
- Replacing the git resources: using [git resolver](https://github.com/tektoncd/pipeline/blob/main/examples/v1beta1/pipelineruns/pipelinerun.yaml#L102-L109)

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -897,7 +886,7 @@ func (c *Reconciler) createTaskRun(ctx context.Context, taskRunName string, para
 		tr.Annotations[workspace.AnnotationAffinityAssistantName] = getAffinityAssistantName(pipelinePVCWorkspaceName, pr.Name)
 	}
 
-	resources.WrapSteps(&tr.Spec, rpt.PipelineTask, rpt.ResolvedTaskResources.Inputs, rpt.ResolvedTaskResources.Outputs, storageBasePath)
+	resources.WrapSteps(&tr.Spec, rpt.PipelineTask, rpt.ResolvedTaskResources.Inputs, rpt.ResolvedTaskResources.Outputs)
```
afrittoli, line 889: I don't think we need outputs here anymore

pkg/reconciler/pipelinerun/resources/input_output_steps.go
```diff
@@ -79,26 +46,13 @@ func GetInputSteps(inputs map[string]*resourcev1alpha1.PipelineResource, inputRe
...
-			taskInputResource.Paths = append(taskInputResource.Paths, stepSourceNames...)
-		}
 		taskInputResources = append(taskInputResources, taskInputResource)
 	}
 	return taskInputResources
 }
 
 // WrapSteps will add the correct `paths` to all of the inputs and outputs for pt
-func WrapSteps(tr *v1beta1.TaskRunSpec, pt *v1beta1.PipelineTask, inputs, outputs map[string]*resourcev1alpha1.PipelineResource, storageBasePath string) {
+func WrapSteps(tr *v1beta1.TaskRunSpec, pt *v1beta1.PipelineTask, inputs, outputs map[string]*resourcev1alpha1.PipelineResource) {
```
afrittoli, line 55: Outputs are not used anymore here

pkg/reconciler/pipelinerun/pipelinerun_test.go
```diff
@@ -281,56 +253,25 @@ spec:
       value: somethingfun
     - name: bar
       value: $(params.bar)
-    - name: templatedparam
```
afrittoli, line 284: Do you know if this functionality (templated param) is only available for PipelineResources? In case it's generally available we should switch this to a regular parameter instead of removing it, or else we'll lose test coverage

afrittoli (commented):
Thanks @JeromeJu - I'm halfway through the review, and I'm not feeling very confident at this stage that we should release this, without the rest of the pipeline resource removal.

Image resources do not rely on the shared artifact storage, but they can be used as output resources - they won't use the storage but they cause an extra step to be added to the task which writes the image digest to the results.

It looks like the logic for image resources is handled on a separate code path, so the removal of `output_resources.go` may be ok, but many tests have been removed because they would be really hard to migrate (and not worth it), so I'm not 100% confident that everything works as it should.

My proposal would be to focus on this and related PRs at the beginning of the next release cycle, to give us enough time to remove everything associated with pipeline resources.

pkg/reconciler/pipelinerun/pipelinerun_test.go
JeromeJu: sorry i might be mistaken here but it felt to me `templatedparam` should be the name of a param?

pkg/apis/pipeline/v1beta1/pipeline_types.go
```diff
@@ -586,36 +563,6 @@ type PipelineTaskParam struct {
...
-	// DeclaredPipelineResources to the input PipelineResources required by the Task.
-	// +listType=atomic
-	Inputs []PipelineTaskInputResource `json:"inputs,omitempty"`
-	// Outputs holds the mapping from the PipelineResources declared in
-	// DeclaredPipelineResources to the input PipelineResources required by the Task.
-	// +listType=atomic
-	Outputs []PipelineTaskOutputResource `json:"outputs,omitempty"`
-}
-
 // PipelineTaskInputResource maps the name of a declared PipelineResource input
```
dibyom, line 566: Can we not also delete PipelineTaskInputResource and PipelineTaskOutputResource?

config/controller.yaml
```diff
@@ -71,9 +71,6 @@ spec:
           "-sidecarlogresults-image", "ko://github.com/tektoncd/pipeline/cmd/sidecarlogresults",
           "-workingdirinit-image", "ko://github.com/tektoncd/pipeline/cmd/workingdirinit",
 
-          # This is gcr.io/google.com/cloudsdktool/cloud-sdk:302.0.0-slim
-          "-gsutil-image", "gcr.io/google.com/cloudsdktool/cloud-sdk@sha256:27b2c22bf259d9bc1a291e99c63791ba0c27a04d2db0a43241ba0f1f20f4067f",
```
dibyom, line 75: We can also remove https://github.com/tektoncd/pipeline/blob/main/config/300-resource.yaml as well right?

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -157,23 +157,16 @@ func TestBuild_JoinMultipleRoots(t *testing.T) {
 	b := v1beta1.PipelineTask{Name: "b"}
 	c := v1beta1.PipelineTask{Name: "c"}
 	xDependsOnA := v1beta1.PipelineTask{
```
dibyom, line 159: nice to have: There are two kinds of dependencies before - using From and using RunAfter. In this PR, we are getting rid of From. So, RunAfter is the only option. Can we simplify these tests instead of just replacing each instance of From with RunAfter?

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -456,24 +423,15 @@ func TestBuild_InvalidDAG(t *testing.T) {
...
 
 	tcs := []struct {
 		name string
 		spec v1beta1.PipelineSpec
 		err  string
 	}{{
-		// a
-		// |
-		// a ("a" depends on resource from "a")
-		name: "self-link-from",
```
dibyom, line 473: is there another test that detects cycles using runAfter?

pkg/reconciler/pipelinerun/pipelinerun_test.go
dibyom: I think Andrea was referring to the `$(inputs.workspace.$(params.rev-param))` usage above. I'm not 100% sure if this is supported elsewhere. cc @jerop do you know?

pkg/reconciler/taskrun/resources/taskresourceresolution.go
```diff
@@ -30,66 +26,16 @@ type ResolvedTaskResources struct {
...
-	Inputs map[string]*resourcev1alpha1.PipelineResource
-	// Outputs is a map from the name of the output required by the Task
-	// to the actual Resource to use for it
-	Outputs map[string]*resourcev1alpha1.PipelineResource
 }
 
-// GetResource is a function used to retrieve PipelineResources.
-type GetResource func(string) (*resourcev1alpha1.PipelineResource, error)
-
 // ResolveTaskResources looks up PipelineResources referenced by inputs and outputs and returns
```
dibyom, line 31: This docstring is not accurate anymore

pkg/reconciler/taskrun/resources/taskresourceresolution.go
```diff
@@ -30,66 +26,16 @@ type ResolvedTaskResources struct {
...
 // a structure that unites the resolved references and the Task Spec. If referenced PipelineResources
 // can't be found, an error is returned.
-func ResolveTaskResources(ts *v1beta1.TaskSpec, taskName string, kind v1beta1.TaskKind, inputs []v1beta1.TaskResourceBinding, outputs []v1beta1.TaskResourceBinding, gr GetResource) (*ResolvedTaskResources, error) {
+func ResolveTaskResources(ts *v1beta1.TaskSpec, taskName string, kind v1beta1.TaskKind) (*ResolvedTaskResources, error) {
```
dibyom, line 34: nice to have: this function is just returning its inputs in a specific struct. Can we simplify and get rid of it?

pkg/reconciler/taskrun/taskrun.go
dibyom: I think this has been addressed now? Both Output and Input are removed?

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -878,17 +716,15 @@ spec:
 
 	taskruns := []*v1beta1.TaskRun{
 		taskRunSuccess, taskRunWithSaSuccess,
-		taskRunSubstitution, taskRunInputOutput,
```
dibyom, line 881: why was taskRunSubstitution remvoed?

pkg/reconciler/taskrun/validate_resources.go
```diff
@@ -25,53 +25,13 @@ import (
 	"github.com/hashicorp/go-multierror"
```
dibyom, line 25: nit: should we rename this file since this no longer has anything to do with validating resources?

test/custom_task_test.go
```diff
@@ -1204,3 +1205,30 @@ func tearDownV1Beta1CustomTask(ctx context.Context, t *testing.T, c *clients, na
 
 	tearDown(ctx, t, c, namespace)
 }
+
+// updateConfigMap updates the config map for specified @name with values. We can't use the one from knativetest because
```
dibyom, line 1209: why do we need to add this in this PR?

test/pipelinerun_test.go
```diff
@@ -179,44 +175,15 @@ func TestPipelineRun(t *testing.T) {
...
-		testSetup              func(ctx context.Context, t *testing.T, c *clients, namespace string, index int) (map[string]*v1alpha1.PipelineResource, *v1beta1.Pipeline)
+		testSetup              func(ctx context.Context, t *testing.T, c *clients, namespace string, index int) *v1beta1.Pipeline
 		expectedTaskRuns       []string
 		expectedNumberOfEvents int
-		pipelineRunFunc        func(*testing.T, int, string, string, map[string]*v1alpha1.PipelineResource) *v1beta1.PipelineRun
+		pipelineRunFunc        func(*testing.T, int, string, string) *v1beta1.PipelineRun
 	}
 
 	tds := []tests{{
-		name: "fan-in and fan-out",
```
dibyom, line 189: question: do we have another test that is e2e testing fan-in and fan-out?

test/resolvers_test.go
```diff
@@ -55,6 +55,7 @@ const (
 	scmRemoteUserPassword = "ab_d1234HIJKL"
 	// Defined in git-resolver/gitea.yaml's "gitea" StatefulSet, in the env for the "configure-gitea" init container
 	scmGiteaAdminPassword = "giteaPassword1234"
+	systemNamespace       = "tekton-pipelines"
```
dibyom, line 58: why do we need this in this PR?

dibyom (commented):
I did a first pass of all the changes - I think I need to come back and review the changes in pkg/reconciler again

pkg/apis/pipeline/images.go
```diff
@@ -56,7 +54,6 @@ func (i Images) Validate() error {
 		{i.GitImage, "git-image"},
```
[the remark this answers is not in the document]
JeromeJu: I might be mistaken but I was thinking the git image is needed for the git-init image?

pkg/apis/pipeline/v1beta1/pipeline_types.go
JeromeJu: Sorry I left the wrong struct here to be tombstoned. 

So to confirm that for `PipelineResourecs` do we need to tombstone fields since it has long been deprecated:)

test/resolvers_test.go
JeromeJu: It is moved here from the deleted artifact_bucket_test.go

test/custom_task_test.go
JeromeJu: It was also deleted from the deleted 'test/artifact_bucket_test' and moved here since it is used in this file.

pkg/apis/pipeline/v1beta1/resource_types.go
```diff
@@ -174,23 +85,6 @@ type PipelineResourceRef struct {
...
-	// GetOutputTaskModifier returns the TaskModifier instance that should be used on a Task
-	// in order to add this kind of resource when it is being used as an output.
-	GetOutputTaskModifier(ts *TaskSpec, path string) (TaskModifier, error)
-	// GetInputTaskModifier returns the TaskModifier instance that should be used on a Task
-	// in order to add this kind of resource when it is being used as an input.
-	GetInputTaskModifier(ts *TaskSpec, path string) (TaskModifier, error)
-}
-
 // TaskModifier is an interface to be implemented by different PipelineResources
 type TaskModifier interface {
```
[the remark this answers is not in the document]
afrittoli: I don't see any of the TaskModifier bits used anywhere, shall we remove it for good, along with its tests

pkg/apis/pipeline/images.go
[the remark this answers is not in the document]
afrittoli: I was going to comment the same as Dibyo - I'm not sure I follow.
This module is used to pass a list of images to the controller, and without the git pipeline resource the controller should not use the git-init image as far as I'm aware.

pkg/reconciler/pipeline/dag/dag_test.go
afrittoli: Indeed - another option could be to use a different kind of dependency - replace the `From` with a result -> param type of dependency.

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -526,10 +484,6 @@ func TestBuild_InvalidDAG(t *testing.T) {
 		name: "duplicate-tasks",
 		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{a, a}},
 		err:  "duplicate pipeline task",
-	}, {
-		name: "invalid-task-name-from",
-		spec: v1beta1.PipelineSpec{Tasks: []v1beta1.PipelineTask{invalidTaskFrom}},
```
afrittoli, line 531: Do we test the result -> param dependency type elsewhere?
We could replace the From dependency with that and keep the test

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -208,10 +203,6 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, pr *v1beta1.PipelineRun)
 	if pr.IsDone() {
 		pr.SetDefaults(ctx)
 
-		if err := artifacts.CleanupArtifactStorage(ctx, pr, c.KubeClientSet); err != nil {
-			logger.Errorf("Failed to delete PVC for PipelineRun %s: %v", pr.Name, err)
-			return c.finishReconcileUpdateEmitEvents(ctx, pr, before, err)
-		}
 		if err := c.cleanupAffinityAssistants(ctx, pr); err != nil {
```
afrittoli, line 206: NIT: we could change this to

```golang
err := c.cleanupAffinityAssistants(ctx, pr)
if err != nil {
    logger.Errorf("Failed to delete StatefulSet for PipelineRun %s: %v", pr.Name, err)
}
```

pkg/reconciler/pipelinerun/pipelinerun.go
```diff
@@ -59,7 +56,6 @@ import (
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
```
afrittoli, line 62: L101 and comment can be removed 
```golang
	ReasonCouldntGetResource = "CouldntGetResource"
```

afrittoli (commented):
@JeromeJu - thanks for this!
I'm at 117 / 156 files viewed - sharing my comments so far. I will finish the review after lunch.

pkg/reconciler/taskrun/taskrun_test.go
JeromeJu: I thought it mainly was related with the git resources, when deleting the resources these taskRuns just become meaningless?

pkg/reconciler/pipelinerun/pipelinerun.go
JeromeJu: Thanks Andrea, removed

pkg/apis/pipeline/v1beta1/resource_types.go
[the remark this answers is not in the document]
JeromeJu: Thanks I removed it along with the related functions.

pkg/apis/pipeline/v1beta1/resource_types.go
```diff
@@ -174,23 +85,6 @@ type PipelineResourceRef struct {
 	APIVersion string `json:"apiVersion,omitempty"`
 }
 
-// PipelineResourceInterface interface to be implemented by different PipelineResource types
```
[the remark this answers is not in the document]
JeromeJu: Done

pkg/apis/pipeline/v1beta1/pipeline_conversion.go
```diff
@@ -36,9 +34,6 @@ func (p *Pipeline) ConvertTo(ctx context.Context, to apis.Convertible) error {
 	switch sink := to.(type) {
 	case *v1.Pipeline:
 		sink.ObjectMeta = p.ObjectMeta
-		if err := serializePipelineResources(&sink.ObjectMeta, &p.Spec); err != nil {
```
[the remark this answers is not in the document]
JeromeJu: Thank you, updated in the release note

config/controller.yaml
JeromeJu: Thanks, removed:)

pkg/apis/pipeline/images.go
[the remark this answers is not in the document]
JeromeJu: Thanks for the pointers, removed as well since they are just used for pipeline resources.

pkg/reconciler/pipeline/dag/dag_test.go
JeromeJu: Thanks Dibyo and Andrea, I modified to only have 2 parallel tasks using `runAfter` with another one using result of both.

pkg/apis/pipeline/v1beta1/pipeline_types.go
dibyom: don't we need to keep the other types around as well if we want to tombstone?

(Given how long pipeline resouces have been deprecated for, I'm not sure if we need to tombstone)

pkg/reconciler/pipeline/dag/dag_test.go
dibyom: yeah that's a good point - do we have result based dependencies captured in dag?

XinruZhang:

Flake: `read tcp 10.52.129.2:33306->185.199.108.154:443: read: connection reset by peer`

pkg/reconciler/pipeline/dag/dag_test.go
JeromeJu: Yes there is and i replaced this one with results and params as @afrittoli suggested.

pkg/reconciler/pipeline/dag/dag_test.go
JeromeJu: Thanks for the suggestion, replaced.

pkg/reconciler/pipeline/dag/dag_test.go
JeromeJu: Thanks both, I replaced all the dependencies that previously based on resources to be using result-params.

pkg/reconciler/taskrun/resources/taskresourceresolution.go
JeromeJu: By removing that, is it okay to leave `ResolvedTaskResources` in `taskresourceresolution.go`?

test/pipelinerun_test.go
JeromeJu: I don't think we do. But for this case I tried to restore it but would not have been able to without the resources.

pkg/reconciler/taskrun/resources/taskresourceresolution.go
```diff
@@ -17,11 +17,7 @@ limitations under the License.
...
 
 import (
-	"errors"
-
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
-	resourcev1alpha1 "github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // ResolvedTaskResources contains all the data that is needed to execute
```
afrittoli, line 23: NIT: ResolvedTaskResources could be renamed to ResolvedTask but we don't have to

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -245,20 +212,6 @@ var (
...
-						Type: resourcev1alpha1.PipelineResourceTypeGit,
-					},
-				}},
-				Outputs: []v1beta1.TaskResource{{
-					ResourceDeclaration: v1beta1.ResourceDeclaration{
-						Name: "workspace",
-						Type: resourcev1alpha1.PipelineResourceTypeGit,
-					},
-				}},
-			},
```
afrittoli, line 261: Line 279 includes a reference to resources

pkg/reconciler/taskrun/taskrun_test.go
```diff
@@ -919,61 +754,6 @@ spec:
 			name:  "sa-step",
 			cmd:   "/mycmd",
 		}}),
-	}, {
-		name:    "params",
-		taskRun: taskRunSubstitution,
```
afrittoli, line 924: Why is this test removed?
`templatedTask` is still around and not tested (one line still includes a reference to a pipeline resource).
The other substituions are still valid tests value, various variables defined in args and elsewhere

test/dag_test.go
```diff
@@ -81,110 +74,56 @@ spec:
...
-        - pipeline-task-2-parallel-1
-        - pipeline-task-2-parallel-2
-        name: repo
-        resource: repo
-      outputs:
-      - name: repo
-        resource: repo
     taskRef:
       name: %s
+    runAfter:
```
afrittoli, line 96: NIT: I think an inplicit type of DAG ordering (e.g. param -> result) would match better with the original test as opposed to an explicit one like runAfter. 
That said, without an e2e coverage report is hard to say.
Once we switch to golang 1.20, we could benefit from https://go.dev/testing/coverage/

test/pipelinerun_test.go
afrittoli: Could you elaborate on why this is not possible?
It is possible to fan-in / fan-out in a pipeline, not with PipelineResources anymore but with results -> params it definitely is.

afrittoli (commented):
Thanks for all the updates, just a few more comments, it's getting very close to ready to merge now.

test/dag_test.go
JeromeJu: Thanks, I replaced the tasks to be using result and params so as to be implicit.

pkg/reconciler/taskrun/taskrun_test.go
JeromeJu: ty for the catch, restored and replaced the resources parts

pkg/reconciler/taskrun/validate_resources.go
JeromeJu: renamed to `validate_taskrun`

pkg/reconciler/taskrun/taskrun_test.go
JeromeJu: restored, thanks for the catch

test/pipelinerun_test.go
dibyom: Chatted w/ @JeromeJu offline about this, I think we can use a few tasks with params and results to do the fan in/fan out behavior

pkg/reconciler/taskrun/resources/taskresourceresolution.go
dibyom: minor but since there are no resources involved anymore, we could rename the func and the struct (and probably the file too)

JeromeJu: make sense, renamed to `taskresolution`

pkg/reconciler/taskrun/resources/taskresourceresolution.go
JeromeJu: renamed and updated in the commit message:)

test/pipelinerun_test.go
JeromeJu: Thanks @dibyom , I replaced it with params and results for fan-in-fan-out test case.

pkg/reconciler/taskrun/resources/taskresolution.go
```diff
@@ -14,16 +14,16 @@ See the License for the specific language governing permissions and
...
-
-	"github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1"
-	"knative.dev/pkg/webhook/resourcesemantics"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 )
 
-func TestTypes(t *testing.T) {
-	// Assert that types satisfy webhook interface.
-	var _ resourcesemantics.GenericCRD = (*v1alpha1.PipelineResource)(nil)
+// ResolvedTask contains all the data that is needed to execute
```
dibyom, line 23: this comment is not accurate - there is no mention of a TaskRun here and no reference to a resource

dibyom (commented):
Need to go through dag_test and pipelinerun_test but the rest looks good to me!

pkg/reconciler/taskrun/resources/taskresolution.go
JeromeJu: Thanks, I also moved it to `pkg/reconciler/taskrun/resources/taskspec.go` since the original file is mostly removed.

afrittoli:
I tried out templated params, and something like `$(param.$(param.paramName))` does not pass validation, so the test should be safe to remove. The PR looks good to me, holding off merge to let Dibyo complete his review.

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -153,27 +153,43 @@ func TestBuild_Parallel(t *testing.T) {
...
 		RunAfter: []string{"b"},
-		Resources: &v1beta1.PipelineTaskResources{
-			Inputs: []v1beta1.PipelineTaskInputResource{{From: []string{"a"}}},
+		Params: []v1beta1.Param{
+			{
+				Value: *v1beta1.NewStructuredValues("$(tasks.a.results.result)"),
+			},
 		},
 	}
 	zDependsOnX := v1beta1.PipelineTask{
```
dibyom, line 190: nit: `zRunsAfterX`

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -219,22 +235,43 @@ func TestBuild_JoinMultipleRoots(t *testing.T) {
...
+		TaskSpec: &v1beta1.EmbeddedTask{TaskSpec: v1beta1.TaskSpec{
+			Results: []v1beta1.TaskResult{{
+				Name: "resultFromD",
+			}},
+		}},
+		Params: []v1beta1.Param{{
+			Value: *v1beta1.NewStructuredValues("$(tasks.a.results.result)"),
+		}},
 	}
 	eRunsAfterA := v1beta1.PipelineTask{
```
dibyom, line 257: nit: the var name suggests eRunsAfterA but the struct suggest it depends on the result of A

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -409,43 +463,69 @@ func TestBuild_InvalidDAG(t *testing.T) {
...
+			Results: []v1beta1.TaskResult{{
+				Name: "resultb",
+			}},
+		}},
 	}
 	cRunsAfterA := v1beta1.PipelineTask{
-		Name:     "c",
-		RunAfter: []string{"a"},
+		Name: "c",
+		Params: []v1beta1.Param{{
```
dibyom, line 513: use runAfterA instead of the param mapping

pkg/reconciler/pipeline/dag/dag_test.go
```diff
@@ -456,10 +536,8 @@ func TestBuild_InvalidDAG(t *testing.T) {
 		RunAfter: []string{"d"},
 	}
 	gDependsOnF := v1beta1.PipelineTask{
-		Name: "g",
-		Resources: &v1beta1.PipelineTaskResources{
-			Inputs: []v1beta1.PipelineTaskInputResource{{From: []string{"f"}}},
-		},
+		Name:     "g",
+		RunAfter: []string{"f"},
```
dibyom, line 540: change to gRunAfterF or use a param/result dependency

JeromeJu:
Thanks @dibyom and @afrittoli for the reviews, just resolved the rest of the comments :)

wlynch:
Heads up, I think this may be a similar problematic change as https://github.com/tektoncd/pipeline/pull/6206 where we're removing both the functionality and API client types in the same change.

Can we consider reverting API fields back to tombstones at least for the v1beta1 client? 🙏

lbernick:
> Heads up, I think this may be a similar problematic change as #6206 where we're removing both the functionality and API client types in the same change.
> 
> Can we consider reverting API fields back to tombstones at least for the v1beta1 client? 🙏

There may have been some discussion around this already but I'm having trouble finding it (https://github.com/tektoncd/pipeline/pull/6150#discussion_r1122197209 is the closest thing I'm seeing). I think this is reasonable; happy to open a PR

PR #6153, opened 2023-02-13
If pod was evicted, prefer pod status message.

drewbailey:
If a pod was evicted, use the pods status message instead of trying to determine a failure message from the pod's container statuses. A container may have a generic exit code that contains less information, such as an exit code and message related to not being located.

```
state:
  terminated:
    exitCode: 137
    finishedAt: null
    message: The container could not be located when the pod was terminated
    reason: ContainerStatusUnknown
    startedAt: null
```

This leads to pipeline status conditions that are less helpful than the eviction message from the pod.

```
status:
  completionTime: "2023-02-08T18:45:37Z"
  conditions:
  - lastTransitionTime: "2023-02-08T18:45:37Z"
    message: |
      "step" exited with code 137 (image: "docker-pullable://...."); for logs run: kubectl -n namespace pod logs step
```

# Changes

# Submitter Checklist

  functionality, content, code)

# Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/pod/status.go
+++ b/pkg/pod/status.go
@@ -423,7 +423,10 @@ func areStepsComplete(pod *corev1.Pod) bool {
 }
 
 func getFailureMessage(logger *zap.SugaredLogger, pod *corev1.Pod) string {
-	// If the pod was evicted return the eviction message.
+	// If a pod was evicted, use the pods status message before trying to
+	// determine a failure message from the pod's container statuses. A
+	// container may have a generic exit code that contains less information,
+	// such as an exit code and message related to not being located.
 	if pod.Status.Reason == evicted {
 		return pod.Status.Message
 	}
--- a/pkg/pod/status.go
+++ b/pkg/pod/status.go
@@ -81,7 +81,10 @@ const (
 	timeFormat = "2006-01-02T15:04:05.000Z07:00"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/pod/status.go
```diff
@@ -420,6 +423,11 @@ func areStepsComplete(pod *corev1.Pod) bool {
 }
 
 func getFailureMessage(logger *zap.SugaredLogger, pod *corev1.Pod) string {
+	// If the pod was evicted return the eviction message.
```
wlynch, line 426:
```suggestion
	// If a pod was evicted, use the pods status message before trying to determine
	// a failure message from the pod's container statuses.
	// A container may have a generic exit code that contains less information,
	// such as an exit code and message related to not being located.
```

Your PR description did a good job explaining why we're doing this instead of waiting for the fallthrough after the container status search. Let's include it here!

wlynch (changes requested):
Looks good! Just want to add some more context in the code.

vdemeester (approved):
LGTM with @wlynch's proposal 👼🏼

pkg/pod/status.go
drewbailey: sounds good to me, added

wlynch (approved):
/lgtm

Thanks!

PR #6155, opened 2023-02-13
Fixed Resolver Timeout

EmmaMunley:
Added global default resolution timeout field in the config map for all remote resolution timeouts.
Changed the default resolution timeout to 2m instead of 1m as requested by @abayer.
Removed git-specific configuration for "fetch-timeout".

/kind bug

  functionality, content, code)

# Release Notes

```release-note
Deprecated the "fetch-timeout" field in the git resolver config
```


pkg/reconciler/resolutionrequest/resolutionrequest.go
```diff
@@ -78,5 +80,5 @@ func requestDuration(rr *v1beta1.ResolutionRequest) time.Duration {
 }
 
 func timeoutMessage() string {
```
abayer, line 82: I'd change this to take `timeout` as an argument, so we don't need the top-level `var timeout ...`.

pkg/resolution/resolver/git/config.go
```diff
@@ -17,10 +17,9 @@ limitations under the License.
 package git
 
 const (
-	// defaultTimeoutKey is the configuration field name for controlling
+	// (deprecated) defaultTimeoutKey is the configuration field name for controlling
 	// the maximum duration of a resolution request for a file from git.
-	defaultTimeoutKey = "fetch-timeout"
-
+	defaultTimeoutKey = "0m"
```
abayer, line 22: Ah, the key should stay the same - it's the value we want to default to `0m`.

pkg/resolution/resolver/git/resolver.go
```diff
@@ -279,10 +279,17 @@ func (r *Resolver) GetConfigName(context.Context) string {
...
-// fetch-timeout field in the git-resolver-config configmap.
+// single git fetch may take. Prior to being deprecated, the defaultTimeout
+// was mapped to the "fetch-timeout" field in the git-resolver-config configmap.
+// Currently, this is defaulted to 0 since there is now a global
+// "DefaultRemoteResolutionTimeoutMinutes" field to configure the timeout for all
+// remote resolution requests including the "git" resolver
 func (r *Resolver) GetResolutionTimeout(ctx context.Context, defaultTimeout time.Duration) time.Duration {
 	conf := framework.GetResolverConfigFromContext(ctx)
+
+	if defaultTimeoutKey == "0m" {
```
abayer, line 290: Remove this bit, since `defaultTimeoutKey` should stay the same.

abayer: ...and then change `defaultMaximumResolutionDuration` in `pkg/resolution/resolver/framework/resolver.go` to 0.

config/resolvers/git-resolver-config.yaml
```diff
@@ -22,8 +22,6 @@ metadata:
     app.kubernetes.io/instance: default
     app.kubernetes.io/part-of: tekton-pipelines
 data:
-  # The maximum amount of time a single anonymous cloning resolution may take.
```
vdemeester, line 25: Do we need to deprecate this prior to removing it ? 🙃 resolvers being beta, we *may* have to follow the 9month rule 😬

abayer: So we're not actually _removing_ the functionality of `fetch-timeout` - anyone who already has it set in their `git-resolver-config` will still have that timeout apply. What we're doing is (a) setting the default to 0, so that the global timeout applies when `fetch-timeout` isn't explicitly set, and (b) removing it from `git-resolver-config` for new installs, to prevent confusion, since users should be using the global timeout, not the git-specific one.

We could leave it in `git-resolver-config.yaml`, set to 0 by default and adding a `DEPRECATED: use global timeout instead blah blah` comment, if need be, but given that the point of deprecation vs removal is to nudge users to use the replacement while not breaking existing users, I think removing `fetch-timeout` from new installs is perfectly fine.

vdemeester: Ah thanks for the explanation, makes sense 👍🏼

pkg/apis/config/default.go
```diff
@@ -118,6 +124,14 @@ func NewDefaultsFromMap(cfgMap map[string]string) (*Defaults, error) {
...
 	}
 
+	if DefaultResolutionTimeoutMinutes, ok := cfgMap[defaultResolutionTimeoutMinutesKey]; ok {
+		timeout, err := strconv.ParseInt(DefaultResolutionTimeoutMinutes, 10, 0)
+		if err != nil {
+			return nil, fmt.Errorf("failed parsing tracing config %q", defaultResolutionTimeoutMinutesKey)
+		}
+		tc.DefaultResolutionTimeoutMinutes = int(timeout)
+	}
+
```
jerop, line 134: please add tests

